PR #891, opened 2017-07-18
Get rid of eni statemanager package

aaithal:
### Summary

Get rid of eni statemanager package

### Implementation details

* a5d142c agent/eni: remove the dependency of eni watcher on eni state manager
* fd9cf21 agent/eni/statemanager: get rid of this package

### Testing

None

### Licensing


agent/eni/watcher_linux.go
```diff
@@ -42,50 +42,43 @@ type UdevWatcher struct {
...
-	// Return on empty list
-	if len(links) == 0 {
-		return errors.New("udev watcher init: no network interfaces discovered for initialization")
-	}
-
-	// Initialize eni state manager with a list of network interface
-	// devices discovered.
-	go udevWatcher.state.Init(links)
-	return nil
+	return udevWatcher.reconcileOnce()
```
vsiddharth: Just to confirm, if Init() errors out, the agent would exit. 

Is this the behavior we want ??
I'd assumed that we want the agent to be up and running even if the eni watcher setup fails as the state can be reconciled upon the next attempt.

agent/eni/watcher_linux.go
```diff
@@ -131,8 +124,48 @@ func (udevWatcher *UdevWatcher) reconcileOnce() {
 	// As we postulate the netlinkClient.LinkList() call to be expensive, we allow
 	// the race here. The state would be corrected during the next reconciliation loop.
 
-	udevWatcher.state.Reconcile(currentState)
-	log.Debug("Udev watcher reconciliation: end")
+	// Add new interfaces next
+	for mac, _ := range currentState {
+		udevWatcher.sendENIStateChange(mac)
```
vsiddharth: Why not make this a goroutine ?

agent/eni/watcher_linux.go
```diff
@@ -111,18 +106,16 @@ func (udevWatcher *UdevWatcher) performPeriodicReconciliation(ctx context.Contex
...
 
 // reconcileOnce is used to reconcile the state of ENIs attached to the instance
-func (udevWatcher *UdevWatcher) reconcileOnce() {
-	log.Debug("Udev watcher reconciliation: begin")
+func (udevWatcher *UdevWatcher) reconcileOnce() error {
 	links, err := udevWatcher.netlinkClient.LinkList()
 	if err != nil {
-		log.Warnf("Udev watcher reconciliation: error retrieving network interfaces: %v", err)
-		return
+		return errors.Wrapf(err, "uder watcher: unable to retrieve network interfaces")
```
richardpen: typo: `udev watcher`.

agent/eni/watcher_linux.go
```diff
@@ -131,8 +124,48 @@ func (udevWatcher *UdevWatcher) reconcileOnce() {
...
+// and if its status should be sent to backend
+func (udevWatcher *UdevWatcher) shouldSendENIStateChange(macAddress string) (*api.ENIAttachment, bool) {
+	if macAddress == "" {
+		log.Warn("ENI state manager: device with empty mac address")
+		return nil, false
+	}
+	// check if this is an eni required by a task
+	eni, ok := udevWatcher.agentState.ENIByMac(macAddress)
+	if !ok {
+		log.Infof("ENI state manager: eni not managed by ecs: %s", macAddress)
```
richardpen: I was wondering should we change this to `Debug` level, since there will be lots of interface created by docker running into this condition.

agent/eni/watcher_linux.go
aaithal: I specifically want the blocking behavior here as any issues with the channel receiver will result in a deadlock rather than a go routine leak.

agent/eni/watcher_linux.go
aaithal: The only path that leads to an error here is the `LinkList`, which can fail because of failed syscalls and I think it's fine to not start the Agent if we are running into that. Moreover, this is sticking to the previous behavior anyway.

agent/eni/watcher_linux.go
aaithal: Yeah, I want to more here to reduce the verbosity of logs in this package. For example, do not scan non `ethernet` devices, ignore the primary ENI etc. Will follow up with another PR.

agent/eni/watcher_linux.go
vsiddharth: Hmm, okay.

agent/eni/watcher_linux.go
vsiddharth: Alright, thanks

PR #892, opened 2017-07-18
Reset agent state if the instance id changed on agent restart

### LINKED ISSUE #859: ECS agent doesn't reset task state when instance id is modified
Repro steps: 
1. Launch instance with Amazon Linux and ecsInstanceRole in default ECS cluster 
2. Install ecs-init package 
3. Start ecs, and let agent register 
4. Run a task
5. Stop ecs
6. Modify the ec2 instance id in the state file
7. Start ecs
8. Inspect the state file/introspection API to illustrate that task ID from previous instance id is persisted in the state file

richardpen:
### Summary

Fix #859 

### Implementation details

Rest the agent state when detected the instance id is different from the one stored in the agent state file.

### Testing

- [x] Manually tested
New tests cover the changes: 

### Licensing

yes

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -232,6 +232,8 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 		log.Warnf(instanceIDMismatchErrorFormat,
 			previousEC2InstanceID, currentEC2InstanceID)
 
+		// Reset agent state as a new container instance
+		state.Reset()
 		// Reset taskEngine; all the other values are still default
 		return engine.NewTaskEngine(agent.cfg, agent.dockerClient, credentialsManager,
 			containerChangeEventStream, imageManager, state), currentEC2InstanceID, nil
--- a/agent/engine/dockerstate/docker_task_engine_state.go
+++ b/agent/engine/dockerstate/docker_task_engine_state.go
@@ -52,6 +52,8 @@ type TaskEngineState interface {
 	AddImageState(imageState *image.ImageState)
 	// RemoveTask removes a task from the state
 	RemoveTask(task *api.Task)
+	// Reset resets all the fileds in the state
+	Reset()
 	// RemoveImageState removes an image.ImageState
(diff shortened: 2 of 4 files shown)


agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -96,6 +98,18 @@ func newDockerTaskEngineState() *DockerTaskEngineState {
 	}
 }
 
+// Reset resets all the states
+func (state *DockerTaskEngineState) Reset() {
+	state.lock.RLock()
```
aaithal: This should be `Lock` and `Unlock` instead of `RLock` and `RUnlock`

agent/app/agent.go
```diff
@@ -232,6 +232,8 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 		log.Warnf(instanceIDMismatchErrorFormat,
 			previousEC2InstanceID, currentEC2InstanceID)
 
+		// Reset agent state as a new container instance
+		state.Reset()
```
aaithal: It seems like you'd need to invoke `statemanager.Save()` here as well? Otherwise, an agent restart here may still result in the wrong state, yes?

richardpen: I don't think so, the file will be override later. And if agent restart at this point, it will run into the same code path `state.Reset` will be executed again.

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -96,6 +98,18 @@ func newDockerTaskEngineState() *DockerTaskEngineState {
 	}
 }
 
+// Reset resets all the states
+func (state *DockerTaskEngineState) Reset() {
```
aaithal: Can you modify this so that you extract all the `state.xxx=make()` bits into a new method, and invoke it here and from `newDockerTaskEngineState()` so that in future if we modify this struct to persist more state, there's no cognitive load of updating `Reset` method as well?

Issue #732, opened 2017-03-15
Unable to provide EngineAuthData from /etc/ecs_container_agent/config.json


ejholmes:
When trying to move from environment variables to a config.json file, I noticed that the agent kept coming up with a bad docker auth configuration (logging `[WARN] Could not parse 'dockercfg' type auth config`) regardless of having a properly formatted config.json, like this:

```json
{
  "Cluster": "<redacted>",
  "EngineAuthType": "dockercfg",
  "EngineAuthData": {
    "https://index.docker.io/v1/":{
      "auth":"<redacted>",
      "email":"<redacted>"
    }
  },
  "Checkpoint": true,
  "DataDir": "/var/run/ecs_agent",
  "TaskIAMRoleEnabled": true,
  "InstanceAttributes": {
    "hostgroup": "<redacted>"
  }
}
```

I tried a little Go program to figure out what the agent was using for the `EngineAuthData`:

```go
package main

import (
	"fmt"
	"log"

	"github.com/aws/amazon-ecs-agent/agent/config"
	"github.com/aws/amazon-ecs-agent/agent/ec2"
	"github.com/aws/amazon-ecs-agent/agent/engine/dockerauth"
)

func main() {
	if err := run(); err != nil {
		log.Fatal(err)
	}
}

func run() error {
	ec2MetadataClient := ec2.NewBlackholeEC2MetadataClient()
	cfg, err := config.NewConfig(ec2MetadataClient)
	if err != nil {
		return err
	}
	fmt.Printf("%#v\n", cfg)
	fmt.Println(string(cfg.EngineAuthData.Contents()))
	auth := dockerauth.NewDockerAuthProvider(cfg.EngineAuthType, cfg.EngineAuthData.Contents())
	fmt.Println(auth)
	return nil
}
```

Which was printing an empty string for `cfg.EngineAuthData.Contents()`. I believe this happens because the Config object returned from pulling values out of environment variables always sets a non-nil value: https://github.com/aws/amazon-ecs-agent/blob/v1.14.0/agent/config/config.go#L333.

A simple fix for this particular instance would be to set `EngineAuthData` to nil of `ECS_ENGINE_AUTH_DATA` is an empty string, which would then fallback to the value provided by the config. However, as a user, I'd prefer to be able to pass a path to a config file, and not have the agent try to merge in environment variables.

richardpen:
[...]

PR #893, opened 2017-07-19
Allow instance attributes to be provided from config file.

ejholmes:
This is basically another case of https://github.com/aws/amazon-ecs-agent/issues/732, where instance attributes from the config file get overridden by the empty instance attributes from the environment.

I think the underlying problem was that `ZeroOrNil` in the `utils` package didn't handle map's, so I fixed the issue there. Based on `git grep ZeroOrNil` it looks like this wouldn't have any unintended side effects.

### Licensing


petderek:
@ejholmes Thank you!! Can you confirm that this contribution is licensed under the terms of the Apache 2.0 license?

ejholmes:
@petderek sorry about that! Update the PR description with the license.

petderek:
@ejholmes Could you respond "yes" after the colon in the template?

Also, I'd like to make a few additional changes to the unit tests. I'll open another PR in addition to this one and carry your commit into it (assuming the license is squared away). Thank you!

ejholmes:
@petderek updated.

petderek:
Merged with #908

PR #879, opened 2017-07-07
Added a reaper for handling exits of child processes

aaithal:
### Summary

Added a reaper for handling exits of child processes and fixed a bunch of minor bugs

### Implementation details

* agent/app: wait for child processes and SIGCHLD
* Makefile: run test-in-docker target with net=none
* agent/ecscni: modify the mount path for the container netns handle
* agent/ecscni: set loglevel for plugins to reflect Agent's log levels
* agent/ecscni: modify the netns path to be Agent specific, rather than reading from /proc directly
* agent/eni: remove dependency on DockerTaskEngine and make udev watcher init asynchronous
* agent/ecscni: modify the cni plugin version string to exclude the asterisk
* misc/pause-container: remove user id for pause container

### Testing

- [X] Builds on Linux (`make release`)
- [X] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [X] Unit tests on Linux (`make test`) pass
- [X] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [X] Integration tests on Linux (`make run-integ-tests`) pass
- [X] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [X] Functional tests on Linux (`make run-functional-tests`) pass
- [x] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: No, tested manually 

### Description for the changelog

None
### Licensing

This contribution is under the terms of the Apache 2.0 License: Yes

PR #895, opened 2017-07-19
agent/app: get rid of the reaper.

aaithal:
### Summary

Gets rid of the process reaper in the ECS Agent

### Implementation details

* b841186 agent/app: get rid of the reaper

The CNI plugin invocation by the ECS Agent fails if the Agent gets a `SIGCHLD` while the CNI plugin invocation is in progress because of the overarching `waitpid(-1)` reaper code. In this scenario, the catch-all `waitpid(-1)` causes CNI's  call to return with `waitid` `syscall` error because it relies
on [cmd.Run()](https://github.com/containernetworking/cni/blob/master/pkg/invoke/raw_exec.go#L42), which in turn calls [cmd.Wait](https://golang.org/pkg/os/exec/#Cmd.Wait), leading to the failure described above.

It could be argued that Agent can be smarter and wait only for `dhclient`  processes by starting them in a different process group and hence not needing to do `waitpid(-1)`. However, I'm not sure if it's feasible for a child process to  be created with a process group that's different from that of the parent (Agent). An `init` system would still be needed to accomplish that. 

It is much cleaner to to instead run the ECS Agent with host's pid namespace. In the future, we can revisit this decision if we want to run the ECS Agent with an init system. The host's  directory will still mounted as  within the Agent container as per https://github.com/aws/amazon-ecs-agent/blob/enis/agent/ecscni/types.go#L32 and that's how it accesses the netns handle for the  container. The only change in the future would be to get rid of the  when
an Agent is configured to run with an `init` system.

### Testing

New tests cover the changes: 

### Licensing


samuelkarp:
> It is much cleaner to to instead run the ECS Agent with host's pid namespace. In the future, we can revisit this decision if we want to run the ECS Agent with an init system.

I'd like to avoid using more things in the host namespace as much as we can.  If it's just for process reaping, I'd much prefer packaging tini or dumb-init inside the agent container and using that.

aaithal:
> I'd like to avoid using more things in the host namespace as much as we can.

I think `pid=host` is kind of unrelated to that. Because we need the pid handle for a process to access it's network namespace, the `/proc` mount is already leaking things into the Agent. That's not going to change even if we package tini or dumb-init with the Agent. I'd like to avoid the complexity of packaging more things into the Agent for this particular use-case.

samuelkarp:
> I'd like to avoid the complexity of packaging more things into the Agent for this particular use-case.

What about just using `--init` then?

The other thing that makes me unhappy with this: neither `--pid=host` nor `--init` are available in the task definition today.  It looks like we really want one of those here in the agent today, and that there's a decently strong need for those features.  If we do start to use either of those flags, I'd like to figure out how we can get the task definition to properly support those as well; it doesn't make a lot of sense for the agent to take advantage of capabilities that we're not making available to our customers via the task definition.

aaithal:
> What about just using `--init` then?

I'm all for that. If you read the commit message, that's what is said in fact. However, that does mean that the Task ENIs feature would be gated on particular versions of docker and I'd like to discuss that offline if that's viable.  

In either of these cases, it doesn't make sense to have the reaper code in the agent, which is what this PR is about.

> The other thing that makes me unhappy with this: neither `--pid=host` nor `--init` are available in the task definition today. 

Agreed, but that's out of scope for this PR.

adnxn:
this is interesting and prompted me to learn about docker and reaping zombie processes.

just to clarify - we need some kind of reaping process specifically to handle dhclient child process? this is something that we need to move forward with the task networking related changes? 

i was looking at pr #879, and the original need for a reaper process in the agent was not clear to me.

samuelkarp:
> I'm all for that. If you read the commit message, that's what is said in fact. However, that does mean that the Task ENIs feature would be gated on particular versions of docker and I'd like to discuss that offline if that's viable.

My opinion is that this is viable; a new feature can depend on a newer version of Docker.

> In either of these cases, it doesn't make sense to have the reaper code in the agent, which is what this PR is about.

If that's the case, I'd like to have the task networking feature gated on some detection code in the agent.  Specifically, we should be able to read our PID and determine if we're running under a different init process (ensure that our PID != 1).  If we're not running under a different init, we should emit a warning or error into the log and disable the feature.

aaithal:
@samuelkarp  done. There's still some work for actually disabling the capability. But as per https://github.com/aaithal/amazon-ecs-agent/blob/rmReaper/agent/api/ecsclient/client.go#L165 it'd be better to do that with some engine refactor where we'll move `Capabilities()` into the `agent/app` package (in a future PR).

agent/api/ecsclient/client.go
```diff
@@ -230,6 +236,34 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
...
+		seelog.Infof("Unable to determine if Agent is started with an init system; Disabling Task ENI capability: %v", err)
+		// TODO: We still need to remove the "task-eni" capability from the
+		// capabilities list. This will be done in the subsequent PR, where
+		// the `TaskEngine.Capabilities()` method will also be refactored
+		// into the `app` package
+		client.config.TaskENIEnabled = false
+		return
+	}
+
+	if !ok {
```
jhaynes: This seems like backwards logic. Why not `if ok { client.config.TaskENIEnabled = false }`? When I'm thinking through this, the nested negations get confused quickly.

agent/api/ecsclient/client_unspecified.go
```diff
@@ -13,11 +13,11 @@
 // express or implied. See the License for the specific language governing
 // permissions and limitations under the License.
 
-package app
+package ecsclient
```
jhaynes: Dunno if this is github or git being weird, but this isn't a file move as much as a deletion of cruft combined with addition of a new function. Would be nice if the git history looked like that too.

aaithal: This seems to be just github weirdness.

agent/api/ecsclient/client.go
```diff
@@ -230,6 +236,34 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
...
 }
 
+// disableTaskENIEnabledIfNoInit disables the Task ENI capability of the
+// Agent if the Agent was not started with an init system. We need the init system
+// to properly cleanup the dhclient processes when Task ENI capability is
+// enabled
+func (client *APIECSClient) disableTaskENIEnabledIfNoInit() {
+	ok, err := isAgentAlsoTheInit()
+	if err != nil {
+		seelog.Infof("Unable to determine if Agent is started with an init system; Disabling Task ENI capability: %v", err)
```
samuelkarp: I think this should be either WARN or ERROR.  This is actually a failure of being able to detect.

agent/api/ecsclient/client.go
```diff
@@ -230,6 +236,34 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
...
+		// the `TaskEngine.Capabilities()` method will also be refactored
+		// into the `app` package
+		client.config.TaskENIEnabled = false
+		return
+	}
+
+	if !ok {
+		return
+	}
+	seelog.Info("Agent is not started with an init system; Disabling Task ENI capability")
```
samuelkarp: If it was explicitly enabled, this should probably be a WARN, since we're now disabling it.

agent/api/ecsclient/client.go
aaithal: done.

agent/api/ecsclient/client.go
aaithal: done.

PR #897, opened 2017-07-19
Improve the error message when reserved memory is larger than the available memory

### LINKED ISSUE #888: Poor error handling when ECS_RESERVED_MEMORY > system memory
In the case where ECS_RESERVED_MEMORY is set to a value higher than system memory, the agent log shows:

```
[ERROR] Error registering: ClientException: integerValue must be between 0 and 2147483647         status code: 400, request id: [SNIP UUID HERE]
```

This is unhelpful until one inspects the client registration api to see that all the integerValues are to do with resources.  

The agent should check if the value to be reported for the resources is negative and either refuse to register with an appropriate error message indicating the resource overly constrained, or truncate the value up to 0 and register.

richardpen:
### Summary

Fix #888 

### Implementation details

Check the value of memory trying to registered before calling ecs RegisterContainerInstance. And if the value is negative then just emit an error and exit the agent.

### Testing

New tests cover the changes: 
yes

### Licensing

yes

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -15,6 +15,7 @@ package ecsclient
 
 import (
 	"errors"
+	"fmt"
 	"runtime"
 	"strings"
 	"time"
@@ -172,7 +173,12 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
 	integerStr := "INTEGER"
 
 	cpu, mem := getCpuAndMemory()
-	mem = mem - int64(client.config.ReservedMemory)
+	remainingMem := mem - int64(client.config.ReservedMemory)
+	if remainingMem < 0 {
+		return "", fmt.Errorf(
+			"api register-container-instance: reserved memory is higher than available memory on the host, total memory: %d, reserved: %d",
+			mem, client.config.ReservedMemory)
+	}
 
 	cpuResource := ecs.Resource{
 		Name:         utils.Strptr("CPU"),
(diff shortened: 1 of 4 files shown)


agent/api/ecsclient/client.go
```diff
@@ -172,7 +173,10 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
 	integerStr := "INTEGER"
 
 	cpu, mem := getCpuAndMemory()
-	mem = mem - int64(client.config.ReservedMemory)
+	registeredMem := mem - int64(client.config.ReservedMemory)
+	if registeredMem < 0 {
+		return "", fmt.Errorf("Resource memory is less than reserved, total memory: %d, reserved: %d", mem, client.config.ReservedMemory)
```
aaithal: Can you modify the error string to be of the `"context: message"` format? 

`"api register-container-instance: reserved memory higher than available memory on the host ..."`

agent/api/ecsclient/client.go
```diff
@@ -172,7 +173,12 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
 	integerStr := "INTEGER"
 
 	cpu, mem := getCpuAndMemory()
-	mem = mem - int64(client.config.ReservedMemory)
+	registeredMem := mem - int64(client.config.ReservedMemory)
```
petderek: Would 'remainingMem' be more appropriate?

PR #899, opened 2017-07-21
Support HTTP Websocket Connections in wsclient

dastbe:
### Summary

This pull requests allows the wsclient (and therefore acs and tcs) to connect over http and upgrade to a plain websocket connection.

### Implementation details
wss (secure websocket) was hardcoded. This change makes a determination based on http or https. Maybe there will be some crazy httpX in the future so I made it a switch.

### Testing

- [X] Builds on Linux (`make release`)
- [X] Unit tests on Linux (`make test`) pass

Enhancement - Allow plain HTTP connections through wsclient

### Licensing

### DIFF
--- a/agent/wsclient/client.go
+++ b/agent/wsclient/client.go
@@ -54,9 +54,6 @@ const (
 	// writeBufSize is the size of the write buffer for the ws connection.
 	writeBufSize = 32768
 
-	// gorilla/websocket expects the websocket scheme (ws[s]://)
-	wsScheme = "wss"
-
 	// Default NO_PROXY env var IP addresses
 	defaultNoProxyIP = "169.254.169.254,169.254.170.2"
 )
@@ -142,6 +139,10 @@ func (cs *ClientServerImpl) Connect() error {
 		return err
 	}
 
+	wsScheme, err := websocketScheme(parsedURL.Scheme)
+	if err != nil {
+		return err
+	}
 	parsedURL.Scheme = wsScheme
 
 	// NewRequest never returns an error if the url parses and we just verified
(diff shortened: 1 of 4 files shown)


agent/wsclient/client.go
```diff
@@ -349,6 +350,19 @@ func (cs *ClientServerImpl) handleMessage(data []byte) {
 	}
 }
 
+func websocketScheme(httpScheme string) (wsScheme string, err error) {
```
petderek: In general, we've avoided using named returns. Please change this.

agent/wsclient/client.go
```diff
@@ -349,6 +350,19 @@ func (cs *ClientServerImpl) handleMessage(data []byte) {
...
+	// gorilla/websocket expects the websocket scheme (ws[s]://)
+	switch httpScheme {
+	case "http":
+		wsScheme = "ws"
+	case "https":
+		wsScheme = "wss"
+	default:
+		err = fmt.Errorf("Unknown httpScheme %s", httpScheme)
```
petderek: Use format `$context: msg` and package `errors.New()`

```errors.New("wsclient: Unknown httpScheme %s", httpScheme)```

File: agent/wsclient/mock/utils/utils.go [hunk 3]
```diff
@@ -22,9 +22,9 @@ import (
 	"github.com/gorilla/websocket"
 )
 
-// StartMockServer starts a mock websocket server.
+// GetMockServer retuns a mock websocket server that can be started up as TLS or not.
 // TODO replace with gomock
-func StartMockServer(t *testing.T, closeWS <-chan []byte) (*httptest.Server, chan<- string, <-chan string, <-chan error, error) {
+func GetMockServer(t *testing.T, closeWS <-chan []byte) (*httptest.Server, chan<- string, <-chan string, <-chan error, error) {
```
petderek: Nice!!

agent/wsclient/client.go
samuelkarp: https://github.com/samuelkarp/amazon-ecs-agent/commit/adcca2ccf68fd42b0e54c13e184cbc9218a31886

agent/wsclient/client.go
dastbe: Done.

agent/wsclient/client.go
dastbe: Done

agent/wsclient/mock/utils/utils.go
dastbe: "Should we gate this code path with an 'allowUnsecureACS' option?"

So my argument against is that we do allow using an http endpoint to talk to ECS via the SDK. If there were a flag, it should apply to both.

samuelkarp:
> Should we gate this code path with an 'allowUnsecureACS' option? I don't think we want this enabled by default.

I'm somewhat okay with it being ungated since the endpoint is not actually configurable; you only get the endpoint from the result of a `DiscoverPollEndpoint` call.

agent/wsclient/client.go
samuelkarp: https://github.com/samuelkarp/amazon-ecs-agent/commit/d57d983d856ddbb3f7eb71fa7ada1df4bba3872f

PR #900, opened 2017-07-24
Refactor the task engine to move the capability function into agent/app package

richardpen:
### Summary

Move the capability function from engine package to the agent/app package.

### Testing

New tests cover the changes: 

### Licensing

yes


samuelkarp:
What's the motivation for this?

richardpen:
@samuelkarp This will make the implementation easier to not register `task-eni` capability when `vpc` or `subnet` acquired failed from the instance metadata, see some comments [here](https://github.com/aws/amazon-ecs-agent/pull/895/files#diff-018bd0da2d0bae369174b684bcec7faeR247). Since app has the ec2 metadata client, we can check the `vpc` and `subnet` before checking the `task-eni` capability. Also since we have `agent/app` package now, this may make more sense to move the the `agent/app` package, as the attributes/capabilities is in the instance level and this was only registered when the agent start.

samuelkarp:
I'm not really convinced that `app` is the right place long-term either; what might actually make more sense is a subsystem capability registration and prerequisite system where different components can register their ability to provide capabilities if the prerequisites are present.  Of course that's a lot more complicated, so if it's simpler for now to just do the move and it doesn't introduce additional debt I don't really have a major problem with it.

richardpen:
@samuelkarp thanks for the suggestion, I will create an issue for that. But for now let's use the simpler way and deliver this feature.

PR #910, opened 2017-08-02
Fix type assertion bug

nmeyerhans:
A type assertion of event.Error assumed that the variable being tested was a
reference, when we actually weren't consistent in how we assigned that
variable. This caused the assertion to fail in places where it should have
passed, ultimately causing us to leak containers in the case where docker
returned an API error in response to the 'stop' call.

### Summary
Fix a bug in which containers can be leaked in the event that docker returns an error in response to a `stop` API call.

### Implementation details

This is a minimal fix that ensures that we're performing assertions of the same type that we're actually assigning. It additionally makes a minor change to log output to add some additional detail.

### Testing

Using a custom build of Docker to force the `stop` API call to return a 500 error. Verify that without this change, agent logs `Error while pulling container; will try to run anyways` and leaks a container. (Leak, in this case, means that it has indicated to the ECS service that the task has successfully transitioned to `STOPPED` even though at least one associated container was still running, and no further efforts would be made to terminate the container.)

With the change, Agent exhibits the expected behavior, which is that it retries the `stop` call indefinitely and does not signal that the task has transitioned to `STOPPED`.

New tests cover the changes:

Bug -- Fix a situation in which containers may be falsely reported as STOPPED in the case of a Docker "stop" API failure.

### Licensing

### DIFF
--- a/agent/engine/errors.go
+++ b/agent/engine/errors.go
@@ -30,6 +30,11 @@ type engineError interface {
 	ErrorName() string
 }
 
+type cannotStopContainerError interface {
+	engineError
+	IsUnretriableError() bool
+}
+
 // impossibleTransitionError is an error that occurs when an event causes a
 // container to try and transition to a state that it cannot be moved to
 type impossibleTransitionError struct {
--- a/agent/engine/errors.go
+++ b/agent/engine/errors.go
@@ -32,7 +32,7 @@ type engineError interface {
 
 type cannotStopContainerError interface {
 	engineError
-	IsUnretriableError() bool
+	IsRetriableError() bool
 }
 
 // impossibleTransitionError is an error that occurs when an event causes a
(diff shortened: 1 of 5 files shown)


nmeyerhans:
@aaithal we already have an `engineError` interface for our custom types. `CannotStopContainerError` is a special case, as it's the only error type that defines additional methods beyond what are already specified by `engineError`.

What we really need here is a new interface type for the `CannotStopContainerError`, then type assertion will work a bit more like it's expected to. This should eliminate the possibility of future bugs introduced by mixing pointers with direct references. I'll put together another PR to implement that change, and we can decide between the two.

agent/engine/task_manager.go
```diff
@@ -254,7 +254,7 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 			}
 			// If docker returned a transient error while trying to stop a container,
 			// reset the known status to the current status and return
-			cannotStopContainerError, ok := event.Error.(*CannotStopContainerError)
+			cannotStopContainerError, ok := event.Error.(CannotStopContainerError)
```
samuelkarp: While this fixes it for now, this doesn't prevent the same problem from happening again.  I would much prefer if we defined an interface for this and type-asserted to the interface.  Interface type-assertions don't matter whether you have a struct or a pointer since either can satisfy the interface.  See [here](https://play.golang.org/p/T07I-lGfiP).

Something like this:

```go
type maybeUnretriableError interface {
	IsUnretriableError() bool
}
```

You should be able to just define this above `handleContainerChange` and use it here.

nmeyerhans: Pretty much exactly what I'm doing... :1st_place_medal: 

Except that I'm taking the opportunity to change that method name!

agent/engine/errors.go
```diff
@@ -154,15 +159,19 @@ func (err CannotStopContainerError) ErrorName() string {
 	return "CannotStopContainerError"
 }
 
-func (err CannotStopContainerError) IsUnretriableError() bool {
+// When stopping a container, most errors that we can get should be
```
samuelkarp: You have this in a godoc location, but it's not godoc format.  Change the beginning to read "IsRetriableError returns ..."

nmeyerhans: :squid: :octopus: :ocean: :whale2: :fish: :ocean: :shell: 

:white_check_mark:

PR #848, opened 2017-06-14
Container Metadata

huerdong:
### Summary

Adds the Container Metadata feature to the agent, exposing the metadata from ECS and from Docker to agent managed containers, such as the container's DockerID and the Container Instance ARN. 

### Implementation details

43932c90a4d6631b11081ee035a319d92dcfb37c Implementation of container metadata on Linux. 
Adds containermetadata package which includes the following:
metadata_manager.go - Provides the MetadataManager interface and implementations for the metadata operations of create, update, and cleanup. 
acquire_metadata.go - Contains functions to parse input from the agent and the raw Docker Container object. 
write_metadata.go - Contains functions to parse the collected metadata and write it into a file. 
types.go - Provides the Metadata type that is populated after parsing inputs and other support structures. 
In the engine package we add calls to MetadataManager methods docker_task_engine.go at the appropriate entry points (Before CreateContainer, after StartContainer, and during sweepTask). In the app package we add a MetadataManager element to the ecsAgent struct so we can pass this to all TaskEngine created by the agent.   

87f7b0bb94ae138f0793fefda25db49eec227946 Added Windows support
Split metadata_manager.go into metadata_manager_unix.go and metadata_manager_windows.go to add platform specific features to the windows build, i.e. different file path formats, to allow container metadata service to function in Windows.

9701a8189f4cc226aef166d6aec6fdeba0e5c35c Added container metadata nable parameter and added dynamic client version 
Added parameter in configs called ContainerMetadataEnabled which enables these features (Default false). Also added logic in docker_container_engine.go to use 1.21 version Docker Client if CotnainerMetadataEnabled is set true as certain networking metadata is unavailable in older versions. 

c800c9f9cd60f708e516f44a6834e0b5d5df8b83 Final set of implementation changes
Refactored containermetadata/metadata_manager.go, app/agent.go, and engine/docker_container_engine.go to allow addition of Container Instance Arn and Docker Client version as parameters in the MetadataManager as these are not readily available previously. These should be the last of any metadata unavailable directly through the configuration or through a Client inspection. 

### Testing

- [x] Builds on Linux (`make release`)
- [x] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [x] Unit tests on Linux (`make test`) pass
- [x] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [x] Integration tests on Linux (`make run-integ-tests`) pass
[...]

Issue #198, opened 2015-09-10
Discovering host port mapping when host port isn't specified.


bcwp:
If a task definition specifies a containerPort, but no host port, ecs automatically picks a random port on the host and maps it to the specified containerPort. There doesn't seem to be any documented way to discover which port it picked from within the container. This seems like a flaw.

euank:
The `describe-tasks` API call returns a `networkBindings` field specifying container/host ports. Once the container transitions to `RUNNING`, the port assignment should be visible there.

I'll see about getting the documentation here improved.

Thanks for the feedback,
Euan

bcwp:
Hey @euank,

How would you recommend doing running the call from within the container (for registering with service discovery)?

euank:
Unfortunately, there's not a great way to get your dynamic port binding from within the container right now. Having extra information in the environment relates to #3 and that specific item is #151.

A different approach to service registration is to instead have whatever process started the task also wait for it to become available and register it rather than have the container register itself.

aaithal:
Another use-case for #709

nick-wellinghoff-sp:
Is there any update on this?  Is there a better way that has evolved over the last 2 years?

richardpen:
@nick-wellinghoff-sp Thanks for reaching out, our engineer is working on the feature and the PR #913 is almost done. Please track this PR for updates.

Thanks,
Peng

samuelkarp:
This will be supported as part of https://github.com/aws/amazon-ecs-agent/issues/710, which is going out with our next release.

samuelkarp:
We have released [ECS agent v1.15.0](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.15.0), which adds this feature.  Please see [the documentation](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-metadata.html) for information on how to use it.

miskiw:
@samuelkarp not sure the ECS v1.15.0 helps in this situation. The "HostPort" value that comes back in the ECS_CONTAINER_METADATA_FILE file is set to "0" (not specified) and not the dynamic port that we are hoping for :-(

samuelkarp:
@miskiw We fixed a few bugs related to that in the [v1.15.2](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.15.2) release.  Can you try that one?

miskiw:
[...]

PR #913, opened 2017-08-04
[WIP] Container Metadata

huerdong:
### Summary

Adds the Container Metadata feature to the agent, exposing the metadata from ECS and from Docker to agent managed containers, such as the container's DockerID and the Container Instance ARN, and then exposes this metadata for users to access from within the container. 

### Implementation details

See #848 

### Testing

Feature - Add metadata service for ECS-created containers. 

### Licensing

This contribution is under the terms of the Apache 2.0 License:  Yes


sharanyad:
Shouldn't this be merged into dev instead of master?

vsiddharth:
@huerdong please change the base branch

huerdong:
Changing the base branch seems to be causing major problems in the commit history.

samuelkarp:
You'll need to change the base branch before we can review this.  Right now, this pull request shows commits that are already merged into the dev branch from different authors; it's not practical to review that mix of commits.

vsiddharth:
@huerdong Please update the summary block and check the appropriate boxes. 
Please double check the license attribution too. 

Thanks

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,144 @@
...
+	// Create task and container directories if they do not yet exist
+	metadataDirectoryPath, err := getMetadataFilePath(task, container, manager.dataDir)
+	// Stop metadata creation if path is malformed for any reason
+	if err != nil {
+		return fmt.Errorf("container metadata create: %v", err)
+	}
+
+	err = os.MkdirAll(metadataDirectoryPath, os.ModePerm)
+	if err != nil {
+		return err
```
aaithal: Please add more context for this error. At the very least something like `fmt.Errorf("creating metadata directory: %v", err)`.

README.md
```diff
@@ -175,6 +175,8 @@ configure them as something other than the defaults.
 | `ECS_IMAGE_MINIMUM_CLEANUP_AGE` | 30m | The minimum time interval between when an image is pulled and when it can be considered for automated image cleanup. | 1h | 1h |
 | `ECS_NUM_IMAGES_DELETE_PER_CYCLE` | 5 | The maximum number of images to delete in a single automated image cleanup cycle. If set to less than 1, the value is ignored. | 5 | 5 |
 | `ECS_INSTANCE_ATTRIBUTES` | `{"stack": "prod"}` | These attributes take effect only during initial registration. After the agent has joined an ECS cluster, use the PutAttributes API action to add additional attributes. For more information, see [Amazon ECS Container Agent Configuration](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) in the Amazon ECS Developer Guide.| `{}` | `{}` |
+| `ECS_ENABLE_CONTAINER_METADATA` | `<true|false>` | Whether to enable creation of a metadata file in the container | `false` | `false` |
```
vsiddharth: Please update this line with more information

README.md
```diff
@@ -175,6 +175,8 @@ configure them as something other than the defaults.
...
+| `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted | `/var/lib/ecs` | `Not used` |
```
vsiddharth: Update this line too

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	// Create metadata file
+	err = createMetadataFile(metadataDirectoryPath)
+	if err != nil {
+		return err
```
aaithal: Same as above. This error needs context. You can even use `errors.Wrapf()` from https://github.com/pkg/errors

agent/containermetadata/parse_metadata.go
```diff
@@ -0,0 +1,153 @@
...
+		containerName: containerName,
+		taskARN:       taskARNFromConfig,
+	}
+}
+
+// parseMetadata gathers metadata from a docker container, and task
+// configuration and data then packages it for JSON Marshaling
+// Since we accept incomplete metadata fields, we should not return
+// errors here and handle them at this or the above stage.
+func (manager *metadataManager) parseMetadata(dockerContainer *docker.Container, task *api.Task, container *api.Container) Metadata {
```
aaithal: It seems like all you need here are task's arn and container's name. Can you just pass those instead of `task` and `container`?

agent/containermetadata/parse_metadata.go
```diff
@@ -0,0 +1,153 @@
...
+	taskMD := parseTaskMetadata(task, container)
```
aaithal: Along the same lines, I don't see the value of creating a function which just initializes a struct. It would be lot helpful to anyone who reads the code to just do this instead:
` return Metadata { taskMetadata: TaskMetadata{...} ...}`

agent/containermetadata/types.go
```diff
@@ -0,0 +1,109 @@
...
+// our functions without creating import cycles
+type dockerMetadataClient interface {
+	InspectContainer(string, time.Duration) (*docker.Container, error)
+}
+
+// Network is a struct that keeps track of metadata of a network interface
+type Network struct {
+	NetworkMode string `json:"NetworkMode,omitempty"`
+	IPv4Address string `json:"IPv4Address,omitempty"`
+}
```
aaithal: Please add the `IPv6Address` field here as well with `omitempty`. You don't need to do anything for it right now. But, it'll be useful when "Task ENIs" is enabled.

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,144 @@
...
+type MetadataManager interface {
+	SetContainerInstanceARN(string)
+	CreateMetadata(*docker.Config, *docker.HostConfig, *api.Task, *api.Container) error
+	UpdateMetadata(string, *api.Task, *api.Container) error
+	CleanTaskMetadata(*api.Task) error
+}
+
+// metadataManager implements the MetadataManager interface
+type metadataManager struct {
+	client               dockerMetadataClient
```
vsiddharth: I'd like to see some comments here for each attribute

agent/containermetadata/write_metadata.go
```diff
@@ -0,0 +1,64 @@
...
+		return "", fmt.Errorf("get metdata file path of task %s container %s: %v", task, container, err)
+	}
+	return filepath.Join(dataDir, metadataJoinSuffix, taskID, container.Name), nil
+}
+
+// getTaskMetadataDir acquires the directory with all of the metadata
+// files of a given task
+func getTaskMetadataDir(task *api.Task, dataDir string) (string, error) {
+	taskID, err := getTaskIDfromARN(task.Arn)
+	return filepath.Join(dataDir, metadataJoinSuffix, taskID), err
```
aaithal: You need to deal with `err` here.

agent/containermetadata/write_metadata_unix.go
```diff
@@ -0,0 +1,84 @@
...
+
+// createMetadataFile initializes the metadata file
+func createMetadataFile(metadataDirectoryPath string) error {
+	metadataFilePath := filepath.Join(metadataDirectoryPath, metadataFile)
+	temp, err := ioutil.TempFile(metadataDirectoryPath, "temp_metadata_file")
+	if err != nil {
+		return err
+	}
+	defer temp.Close()
+	if err != nil {
```
aaithal: This is redundant

agent/containermetadata/write_metadata.go
```diff
@@ -0,0 +1,64 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/aws/amazon-ecs-agent/agent/api"
+)
+
+const (
+	metadataJoinSuffix = "metadata"
+	metadataFile       = "metadata.json"
```
aaithal: How about "ecs-container-metadata.json" ?

agent/engine/docker_task_engine.go
```diff
@@ -303,6 +308,14 @@ func (engine *DockerTaskEngine) sweepTask(task *api.Task) {
 			seelog.Errorf("Error removing container reference from image state: %v", err)
 		}
 	}
+
+	// Clean metadata directory for task
+	if engine.cfg.ContainerMetadataEnabled {
+		err := engine.metadataManager.CleanTaskMetadata(task)
+		if err != nil {
+			seelog.Errorf("clean task metadata failed for task %s: %v", task, err)
```
aaithal: Logs need to start with Upper case letters.

agent/containermetadata/parse_metadata.go
```diff
@@ -0,0 +1,153 @@
...
+// available prior to container creation
+// Since we accept incomplete metadata fields, we should not return
+// errors here and handle them at this or the above stage.
+func (manager *metadataManager) parseMetadataAtContainerCreate(task *api.Task, container *api.Container) Metadata {
+	taskMD := parseTaskMetadata(task, container)
+	return Metadata{
+		cluster:              manager.cluster,
+		taskMetadata:         taskMD,
+		containerInstanceARN: manager.containerInstanceARN,
+		metadataStatus:       "NOT_READY",
```
vsiddharth: Could we have a new type for metadata status ?
I'd like to have it be one amongst a set of well defined values

agent/containermetadata/write_metadata_unix.go
```diff
@@ -0,0 +1,84 @@
...
+
+// writeToMetadata puts the metadata into JSON format and writes into
+// the metadata file
+func writeToMetadataFile(data []byte, task *api.Task, container *api.Container, dataDir string) error {
+	metadataFileDir, err := getMetadataFilePath(task, container, dataDir)
+	// Boundary case if file path is bad (Such as if task arn is incorrectly formatted)
+	if err != nil {
+		return fmt.Errorf("write to metadata file for task %s container %s: %v", task, container, err)
+	}
+	metadataFilePath := filepath.Join(metadataFileDir, metadataFile)
```
vsiddharth: This seems to be duplicated code from the `createMetadataFile` method

agent/containermetadata/write_metadata_windows.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/pborman/uuid"
+)
+
+const (
+	mountPoint                     = `C:\ProgramData\Amazon\ECS\metadata`
+	ContainerMetadataClientVersion = dockerclient.Version_1_24
+)
+
+// createMetadataFile initializes the metadata file
```
vsiddharth: Update this comment

agent/containermetadata/write_metadata_windows.go
```diff
@@ -0,0 +1,79 @@
...
+func createMetadataFile(metadataDirectoryPath string) error {
+	metadataFilePath := filepath.Join(metadataDirectoryPath, metadataFile)
+	file, err := os.Create(metadataFilePath)
+	if err != nil {
```
vsiddharth: Why didn't you go with the "Tempfile" approach here ?

agent/app/agent.go
```diff
@@ -108,6 +110,26 @@ func newAgent(
 		return nil, err
 	}
 
+	var metadataManager containermetadata.MetadataManager
```
aaithal: `containermetadata.Manager` stutters less than  `containermetadata.MetadataManager`

agent/app/agent.go
```diff
@@ -108,6 +110,26 @@ func newAgent(
...
+	if cfg.ContainerMetadataEnabled {
+		// Get the oldest Docker Client version with up to date inspect API
```
aaithal: Please move this to a method. Also, this needs a test in `agent/app/agent_test.go`

agent/config/config.go
```diff
@@ -315,6 +315,9 @@ func environmentConfig() (Config, error) {
 		seelog.Debugf("Setting instance attribute %v: %v", attributeKey, attributeValue)
 	}
 
+	containerMetadataEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_CONTAINER_METADATA"), false)
```
aaithal: needs a test

agent/config/config_unix.go
```diff
@@ -28,6 +28,7 @@ func DefaultConfig() Config {
 		ReservedPorts:               []uint16{SSHPort, DockerReservedPort, DockerReservedSSLPort, AgentIntrospectionPort, AgentCredentialsPort},
 		ReservedPortsUDP:            []uint16{},
 		DataDir:                     "/data/",
+		DataDirOnHost:               "/var/lib/ecs",
```
aaithal: needs a test

agent/config/config_windows.go
```diff
@@ -44,6 +44,7 @@ const (
 func DefaultConfig() Config {
 	programData := utils.DefaultIfBlank(os.Getenv("ProgramData"), `C:\ProgramData`)
 	ecsRoot := filepath.Join(programData, "Amazon", "ECS")
+	dataDir := filepath.Join(ecsRoot, "data")
```
aaithal: needs a test

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aaithal: This can be renamed as `manager.go`

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
...
+
+const (
+	metadataEnvironmentVariable = "ECS_CONTAINER_METADATA_FILE"
+	inspectContainerTimeout     = 30 * time.Second
+)
+
+type MetadataStatus string
+
+const (
+	Metadata_NOT_READY = "NOT_READY"
```
aaithal: lint: `MetadataNotReady` and `MetadataReady`.

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
...
+
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+const (
+	metadataEnvironmentVariable = "ECS_CONTAINER_METADATA_FILE"
+	inspectContainerTimeout     = 30 * time.Second
+)
+
+type MetadataStatus string
```
aaithal: lint: needs `// MetadataStatus is ...` or `// MetadataStatus specifies..` etc

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
...
+const (
+	Metadata_NOT_READY = "NOT_READY"
+	Metadata_READY     = "READY"
+)
+
+// MetadataManager is an interface that allows us to abstract away the metadata
+// operations
+type MetadataManager interface {
+	SetContainerInstanceARN(string)
+	CreateMetadata(*docker.Config, *docker.HostConfig, *api.Task, *api.Container) error
```
aaithal: As per https://github.com/aws/amazon-ecs-agent/pull/913#discussion_r131703627, you should only require task arn and container name in here and all the way down.

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
...
+	UpdateMetadata(string, *api.Task, *api.Container) error
```
aaithal: Same as above. As per https://github.com/aws/amazon-ecs-agent/pull/913#discussion_r131703627, you should only require task arn and container name in here and all the way down.

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	err = os.MkdirAll(metadataDirectoryPath, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("creating metadata directory for task %s: %v", task, err)
+	}
+
+	// Acquire the metadata then write it in JSON format to the file
+	metadata := manager.parseMetadataAtContainerCreate(task.Arn, container.Name)
+	data, err := json.MarshalIndent(metadata, "", "\t")
```
aaithal: You need to handle this `err` as well (points to a lacking unit test).

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	// Ensure we do not update a container that is invalid or is not running
+	if dockerContainer == nil || !dockerContainer.State.Running {
+		return fmt.Errorf("container metadata update: container not running or invalid")
+	}
+
+	// Acquire the metadata then write it in JSON format to the file
+	metadata := manager.parseMetadata(dockerContainer, task.Arn, container.Name)
+	data, err := json.MarshalIndent(metadata, "", "\t")
```
aaithal: You need to handle this `err` as well (points to a lacking unit test).

agent/containermetadata/types.go
```diff
@@ -0,0 +1,110 @@
...
+	cluster                 string
+	taskMetadata            TaskMetadata
+	dockerContainerMetadata DockerContainerMetadata
+	containerInstanceARN    string
+	metadataStatus          MetadataStatus
+}
+
+// MetadataSerializer is an intermediate struct that converts the information
+// in Metadata into information to encode into JSOn
+type MetadataSerializer struct {
```
aaithal: Can this not be global scoped (`metadataSerializer `)?

agent/containermetadata/write_metadata.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return "", fmt.Errorf("get metdata file path of task %s container %s: %v", task, container, err)
+	}
+	return filepath.Join(dataDir, metadataJoinSuffix, taskID, container.Name), nil
+}
+
+// getTaskMetadataDir acquires the directory with all of the metadata
+// files of a given task
+func getTaskMetadataDir(task *api.Task, dataDir string) (string, error) {
+	taskID, err := getTaskIDfromARN(task.Arn)
```
aaithal: You need handle this error (points to a lack of unit test).

agent/containermetadata/metadata_manager.go
aaithal: You can name this as `Create` instead.

agent/containermetadata/metadata_manager.go
aaithal: You can name this as `Update` instead.

agent/containermetadata/metadata_manager.go
```diff
@@ -0,0 +1,151 @@
...
+	Metadata_READY     = "READY"
+)
+
+// MetadataManager is an interface that allows us to abstract away the metadata
+// operations
+type MetadataManager interface {
+	SetContainerInstanceARN(string)
+	CreateMetadata(*docker.Config, *docker.HostConfig, *api.Task, *api.Container) error
+	UpdateMetadata(string, *api.Task, *api.Container) error
+	CleanTaskMetadata(*api.Task) error
```
aaithal: You can name this as `Clean` instead.

aaithal:
@huerdong can you also paste the output of `go test -cover ./agent/containermetadata`? You may need to make some modifications with the command. I'm mostly looking for test coverage data.

agent/app/agent.go
huerdong: I understand your purpose but I am reluctant to make this change because there are so many structs in the agent with the word Manager in them. Having a single struct called Manager, even if preceded by package containermetadata, feels confusing to me.

aaithal: That's fine. As long as variables are named correctly, it shouldn't matter what the struct name is.

agent/containermetadata/metadata_manager.go
vsiddharth: `NOT_READY` is not entirely correct here as we have some data which we consider as as metadata when this status is used. 

I'd prefer statuses like 
* `CREATED`
* `UPDATED`

and documenting what each of them implies

vsiddharth:
@huerdong Could you also resolve the conflicts when you update #913 ?

agent/app/agent.go
```diff
@@ -108,6 +110,26 @@ func newAgent(
...
+		// as some metadata is unavailable in older API client versions
+		// If this version is unavailable we use default
+		// Due to limitations of interfaces and import cycles
```
samuelkarp: I think your only import cycle issue is with the `containermetadata.ContainerMetadataClientVersion`.  I'd be in favor of moving this into the client/factory code directly and just moving the constant over there as well so you don't have an import cycle; that'll keep the impact localized to just the code that needs to be refactored anyway.

agent/app/agent.go
```diff
@@ -108,6 +110,26 @@ func newAgent(
 		return nil, err
 	}
 
+	var metadataManager containermetadata.MetadataManager
+	if cfg.ContainerMetadataEnabled {
+		// Get the oldest Docker Client version with up to date inspect API
+		// as some metadata is unavailable in older API client versions
+		// If this version is unavailable we use default
```
samuelkarp: It'd be reasonable to just disable the feature if Docker is too old.

agent/containermetadata/parse_metadata.go
```diff
@@ -0,0 +1,136 @@
...
+		ports:               ports,
+		networkInfo:         networkMetadata,
+	}
+}
+
+// parseMetadata gathers metadata from a docker container, and task
+// configuration and data then packages it for JSON Marshaling
+// Since we accept incomplete metadata fields, we should not return
+// errors here and handle them at this or the above stage.
+func (manager *metadataManager) parseMetadata(dockerContainer *docker.Container, taskARN string, containerName string) Metadata {
```
samuelkarp: Generally, I'd like to see the functions in a file organized with caller-above-callee so that you start with abstractions at a high level and get deeper into details as you read further in the file.  That would mean that the entrypoints (`parseMetadata` and `parseMetadataAtContainerCreate`) are at the top, then `parseDockerContainerMetadata`, and last `parseNetworkMetadata`.

agent/containermetadata/write_metadata_windows.go
```diff
@@ -0,0 +1,68 @@
...
+	metadataFileDir, err := getMetadataFilePath(taskARN, containerName, dataDir)
+	// Boundary case if file path is bad (Such as if task arn is incorrectly formatted)
+	if err != nil {
+		return fmt.Errorf("write to metadata file for task %s container %s: %v", taskARN, containerName, err)
+	}
+	metadataFileName := filepath.Join(metadataFileDir, metadataFile)
+
+	file, err := os.OpenFile(metadataFileName, os.O_WRONLY, 644)
+	if err != nil {
+		// Retry if file does not exist
```
samuelkarp: It's worth checking what this `err` is before blindly retrying.

agent/containermetadata/metadata_manager.go
aaithal: If there's a second update, `CREATED` and `UPDATED` are not sufficient to convey that intent. `NOT_READY` indicates helps with this regard.

agent/engine/docker_task_engine.go
```diff
@@ -625,7 +647,21 @@ func (engine *DockerTaskEngine) startContainer(task *api.Task, container *api.Co
 			Error: CannotStartContainerError{fmt.Errorf("Container not recorded as created")},
 		}
 	}
-	return client.StartContainer(dockerContainer.DockerID, startContainerTimeout)
+	dockerContainerMD := client.StartContainer(dockerContainer.DockerID, startContainerTimeout)
+
+	// Get metadata through container inspection and available task information then write this to the metadata file
+	// Performs this in the background to avoid delaying container start
```
samuelkarp: It almost feels like this should be a new container state.  As is, if the agent restarts while this goroutine is running (after the container has started and we've recorded that in the state file, but before we've completed updating the metadata), you may never get metadata properly updated (`startContainer` won't be called again).

agent/containermetadata/metadata_manager.go
vsiddharth: I still don't clearly understand the reason behind the second update. 
However, in this case we could then switch to these states:
* `INITIAL`
* `READY`

I'm not in favor of `NOT_READY` as a status here

README.md
sharanyad: Could you please add information on why this is not used for Windows?

agent/containermetadata/write_metadata.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sharanyad: Why is this file named 'write_metadata.go'? This does not really write metadata as such.

sharanyad:
Having some unit tests for metadata manager would be good.

petderek:
Why do we need INITIAL, READY, NOT_READY, etc? As a user, do I care about the state of the metadata?

aaithal:
@petderek  it'd have been easier if you had just added to the threaded discussion than a comment here. Regardless, as a user, I'd care as it'd help me distinguish between "will i ever get this data" and "i'll eventually get this data if i keep polling"

agent/containermetadata/metadata_manager.go
petderek: >> Why do we need INITIAL, READY, NOT_READY, etc? As a user, do I care about the state of the metadata?

> @petderek it'd have been better if you had just added to the threaded discussion than a comment here. Regardless, as a user, I'd care as it'd help me distinguish between "will i ever get this data" and "i'll eventually get this data if i keep polling"



Without split updates:
[ does data exist? ] => [ does the field I want exist? ] => [Yay, data!]

With:
[ does data exist? ] => [ is the 'ready' field set? ] => [ does the field i want exist? ] => [Yay, data!]

I'm not convinced that splitting up updates is simplifying. Either way, you will still need to poll for metadata. However, now anyone calling this needs to be aware of a different field that guarantees completeness.

aaithal: Because the update is inherently racy, you need a way to determine if the file is in a consumable format. Data not being present and not being updated are two different things and this is what the field helps with. Alternatively, we only create the file when we have all of the data and get rid of the state stuff.

agent/app/agent.go
huerdong: The feature still works for older versions of docker. The differences in the versions do not actually matter previously because I get the relevant networking metadata through an alternative route for older versions. With the ENI feature added, this difference is significant because older versions of Docker API won't display the additional network modes. Again, I am not very familiar with ENI project except that I was asked to make the Network information into a list to support multiple network modes. I would think if a user wants to use this feature even with an older version of Docker, we should not disable it if it works for that older client.

agent/engine/docker_task_engine.go
huerdong: I think I previously discussed making this a separate container state with @aaithal. Given the timeline of my ability to work on this project, I do not think I can feasibly make this change right now.

agent/containermetadata/metadata_manager.go
huerdong: The most general way to do this is to have some kind of update counter, but that information feels too abstract and not informative. I think the best course of action here is to agree on what to name the current two states and then when we need additional layers of updates, make changes to these names.

agent/engine/docker_task_engine.go
richardpen: This could definite happen in the agent, to make it simple than adding a new state is to add a new state in the container struct to indicate the metadata state.

agent/containermetadata/metadata_manager.go
petderek: >  Alternatively, we only create the file when we have all of the data and get rid of the state stuff.

That's what I'd suggest. I think it would simplify our code and make it easier to consume.

huerdong: The issue with getting rid of the state stuff is that we are getting rid a way to detect the metadata race condition where the user wants this data but it is not ready. The container has to start before we get all the data, so it needs a way to check from within itself that the metadata is not ready. This means that the status of the metadata file itself is a container metadata for which we need to provide lookup.

agent/engine/docker_task_engine.go
```diff
@@ -597,6 +610,15 @@ func (engine *DockerTaskEngine) createContainer(task *api.Task, container *api.C
 	seelog.Infof("Created container name mapping for task %s - %s -> %s", task, container, containerName)
 	engine.saver.ForceSave()
 
+	// Create metadata directory and file then populate it with common metadata of all containers of this task
```
richardpen: Looks like the changes in this file are not unit tested.

agent/engine/docker_task_engine.go
```diff
@@ -303,6 +308,14 @@ func (engine *DockerTaskEngine) sweepTask(task *api.Task) {
 			seelog.Errorf("Error removing container reference from image state: %v", err)
 		}
 	}
+
+	// Clean metadata directory for task
+	if engine.cfg.ContainerMetadataEnabled {
```
richardpen: This check seems weak, as writing the metadata file could fail. The safer way is to add a new state of this file and only delete this file if it's created.

Also please add a debug log here to indicate this happened.

agent/containermetadata/write_metadata_windows.go
```diff
@@ -0,0 +1,68 @@
...
+// the metadata file
+func writeToMetadataFile(data []byte, taskARN string, containerName string, dataDir string) error {
+	metadataFileDir, err := getMetadataFilePath(taskARN, containerName, dataDir)
+	// Boundary case if file path is bad (Such as if task arn is incorrectly formatted)
+	if err != nil {
+		return fmt.Errorf("write to metadata file for task %s container %s: %v", taskARN, containerName, err)
+	}
+	metadataFileName := filepath.Join(metadataFileDir, metadataFile)
+
+	file, err := os.OpenFile(metadataFileName, os.O_WRONLY, 644)
```
richardpen: can you make `644` a constant, probably move the one defined in `write_metadata_unix.go` to `metadatamanager.go`.

agent/app/agent.go
```diff
@@ -108,6 +110,26 @@ func newAgent(
...
+		// into containermetadata.NewMetadataManager()
+		supportedClients := dockerClient.SupportedVersions()
+		metadataDockerClient := dockerClient
+		for index := range supportedClients {
+			if supportedClients[index] == containermetadata.ContainerMetadataClientVersion {
+				metadataDockerClient = dockerClient.WithVersion(containermetadata.ContainerMetadataClientVersion)
+				break
+			}
+		}
+		metadataManager = containermetadata.NewMetadataManager(metadataDockerClient, cfg)
```
huerdong: The only missing metadata is in the network settings with the old client. However because we have specific checks in the parsing of the networking metadata to address this issue we will not actually be missing any information unless there are multiple network modes present.

agent/engine/docker_task_engine.go
huerdong: I am not sure why an additional check is needed. If the file is not created, the cleanup would simply do nothing (Not even return an error)
https://golang.org/pkg/os/#RemoveAll

agent/containermetadata/metadata_manager.go
petderek: > The container has to start before we get all the data, so it needs a way to check from within itself that the metadata is not ready.

What makes this a need? 

I don't think we 'need' any of this state stuff. We can just guarantee that metadata is available once the container is running. Adding metadata state introduces complexity that warrants further changes ( more container states, etc).

My concern is that we have started trying to solve a problem that doesn't need to exist in the first place.

agent/containermetadata/write_metadata.go
huerdong: Most of my writing operations was originally here and has since been moved to other places. It is probably better to name it utils.go

agent/containermetadata/metadata_manager.go
huerdong: I understand what you are saying and certainly we do not need the states stuff if we perform a single metadata file write after the container starts and we have all the data. The problem is for users to check that this file is ready in this case, they would either need a delay after container start or a file exist check after the container starts to ensure the file is valid. This puts unnecessary burden on the user to address the race condition when we can simply provide the availability of the metadata as a metadata field as well.

petderek: >  unnecessary burden on the user to address the race condition

What is the race condition?

agent/app/agent.go
huerdong: Right now I am cautious about making changes here because I have encountered in the past import cycles and/or "interface is not implemented" errors when dealing with this section of code. I will see what I can do here.

agent/containermetadata/metadata_manager.go
huerdong: If I do not rename MetadataManager, is it still appropriate to rename this file? On one hand, there is no confusion with renaming as this file will be isolated in its own package. On the other this breaks the convention of all the other "_manager" files in the code base.

agent/app/agent.go
huerdong: I'm not quite sure what you mean here. The problem I am encountering is that adding the WithVersion function to the internal dockerMetadataClient interface in the containermetadata package led to "interface not implemented" errors in the past. I am very uncertain about this section of code as a result, but I will try to see if I can minimize the exposure in this section.

agent/containermetadata/parse_metadata.go
huerdong: I have made this reorganization (I'll also use this to remind myself to reorganize the Ports in the metadata which was not addressed yet from a previous comment).

agent/app/agent.go
samuelkarp: If we handle old clients, do we even care about the client version anymore?  Should we just use the default client and get rid of this logic entirely?

agent/containermetadata/metadata_manager.go
huerdong: While I have handled this error, I'm not sure if it is possible for an error to occur here and whether it is worthwhile to test for it. Since all my fields in the Metadata struct are strings, there should never be an error in the Marshaling. https://stackoverflow.com/questions/33903552/what-input-will-cause-golangs-json-marshal-to-return-an-error

agent/containermetadata/metadata_manager.go
huerdong: See above.

huerdong:
I have added unit tests for the containermetadata package, addressing comments by @sharanyad @aaithal. 
In addition as requested by @aaithal here is the output of `go test -cover ./agent/containermetadata`

`ok  	github.com/aws/amazon-ecs-agent/agent/containermetadata	0.012s	coverage: 74.5% of statements
`

agent/containermetadata/metadata_manager.go
aaithal: That's not a convention, just a sign of us being immature golang coders when that code was written. I don't see a strong argument against renaming this. Please rename both as per my earlier comment.

agent/app/agent.go
aaithal: I don't see how moving this to a method would result in import cycles. On the other hand, moving this to a method improves readability and makes it more testable.

aaithal:
@huerdong `74.5%` is a relatively low number for a new package that's being written. Please take another look at what code paths are not being covered in tests and add coverage for the same. Here's a blog on how you can generate html reports to visualize the missing test coverage in code: https://blog.golang.org/cover

agent/app/agent.go
huerdong: The client version matters because with Task ENI multiple network modes are now supported. The default client does not give information on additional network modes which is only available in v1.21.

agent/app/agent.go
huerdong: Do you mean moving it into a method but still within the app package? Maybe I misconstrued your request.

aaithal: > Do you mean moving it into a method but still within the app package?

Yeah.

agent/app/agent.go
samuelkarp: @richardpen and @aaithal can correct me here, but I don't believe that `docker inspect` will return information about the network interfaces added through a CNI plugin.

aaithal: > but I don't believe that docker inspect will return information about the network interfaces added through a CNI plugin

That's correct. It's also worth nothing that we invoke 2 top level CNI plugins and each of them results in a network interface being added to the container (in effect, being part of multiple "network"s)

huerdong: In that case, I will remove the 1.21 client check and use the default. I will leave the Networks field in the metadata stored as a list in case we use a newer API version as a default later.

agent/config/config.go
huerdong: For this and the "needs a test" comments below for the config package, I have added tests to check for my new configuration variables.

agent/engine/docker_task_engine.go
huerdong: I have added unit tests for the docker_task_engine_test.go for the cases where metadata service is enabled.

agent/config/config_unix.go
huerdong: see above

agent/config/config_windows.go
huerdong: see above

agent/engine/docker_task_engine.go
huerdong: I have added a TODO here for adding a metadata file status in api.Container and for doing metadata updates for containers running after agent restart.

opalelement:
@aaithal Looks like @huerdong mentioned the code coverage has been bumped to 96.2%

vsiddharth:
@opalelement The code coverage was bumped by @huerdong after we introduced him to using wrappers for mocks.

mariuszluciow:
Hi here, any news on this? We are waiting for #198 quite some time now...

aaithal:
closing this in favor of https://github.com/aws/amazon-ecs-agent/pull/981

PR #915, opened 2017-08-04
Disable task eni capability on errors

aaithal:
### Summary

Disable task eni capability on errors

### Implementation details

* 2608468 agent/app: combine all of ENI dependency initialization logic
* 260cb75 agent/ec2: Get rid of the 'LaunchedWithoutVPC' method
* 3775d75 agent/eni: fixed a bug where the udev monitor was not beig cleanly torn down
* 16df36f agent/app/oswrapper: add a package to wrap methods in the stdlib's os package that are used by the app package
* ae06fcb agent/api: remove task eni specific logic

### Testing

None

### Licensing


agent/app/oswrapper/oswrapper.go
```diff
@@ -11,26 +11,24 @@
 // express or implied. See the License for the specific language governing
 // permissions and limitations under the License.
 
-package ec2
+package oswrapper
```
richardpen: Can this package moved to the `agent/utils`, which possibly will be reused in the future?

agent/app/agent_unspecified.go
```diff
@@ -13,15 +13,15 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/engine"
+	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
+)
 
-// isAgentAlsoTheInit returns true if the Agent process's pid is
-// 1, which means it's running without an init system
-func isAgentAlsoTheInit() (bool, error) {
-	return os.Getpid() == initPID, nil
+func (agent *ecsAgent) initializeTaskENIDependencies(state dockerstate.TaskEngineState, taskEngine engine.TaskEngine) (error, bool) {
+	return errors.New("unsupported platform"), true
```
richardpen: Can we stick with the `github.com/pkg/errors` package?

agent/app/agent_unix.go
```diff
@@ -0,0 +1,156 @@
...
+	return true
+}
+
+// queryCNIPluginsCapabilities returns an error if there's an error querying
+// capabilities or if the required capability is absent from the capabilies
+// of the following plugins:
+// a. ecs-eni
+// b. ecs-bridge
+// c. ecs-ipam
+func (agent *ecsAgent) queryCNIPluginsCapabilities() error {
```
richardpen: Would the name`checkCNIPluginsCapabilities` or `verifyCNIPluginsCapabilities` better than `queryCNIPluginsCapabilities`?

agent/app/agent_unix.go
```diff
@@ -0,0 +1,156 @@
...
+		if pause.IsNoSuchFileError(err) || pause.UnsupportedPlatform(err) {
+			// If the pause container's image tarball doesn't exist or if the
+			// invocation is done for an unsupported platform, we cannot recover.
+			// Return the error as terminal for these cases
+			return fmt.Errorf("unable to load pause container image: %v", err), true
+		}
+		return fmt.Errorf("unable to load pause container image: %v", err), false
+	}
+
+	if _, err := enisetup.New(agent.ctx, state, taskEngine.StateChangeEvents()); err != nil {
```
richardpen: I was wondering should we change the method name to `enisetup.Start` or something like this, `enisetup.New` seems a little confusing to me.

agent/app/agent.go
```diff
@@ -176,8 +188,30 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+					return exitcodes.ExitTerminal
+				}
+				return exitcodes.ExitError
+			}
+			// We have ascertained that the EC2 Instance is not running in a VPC
+			// No need to stop the ECS Agent in this case; all we need to do is
+			// to not update the config to disabled the TaskENIEnabled flag and
+			// move on
+			seelog.Warnf("Unable to detect VPC ID for the Instance, disabling Task ENI capability: %v", err)
+			agent.cfg.TaskENIEnabled = false
```
aaithal: I think the use-case here is that a customer running an instance with ECS Agent installed via the `ECS Init` package in non-VPC land would have their instance never register because it's enabled via ECS Init. The right thing to do here may be to disable this using the config variable via ECS Init. But then, you'd have to deal with failure cases where customers might have created the `ecs.config` file with this turned on. So, disabling this on instances not running on VPC seems like the right answer. Let me know if you have alternate suggestions here.

agent/app/agent_unix.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// Load the pause container's image from the 'disk'
+	if _, err := agent.pauseLoader.LoadImage(agent.cfg, agent.dockerClient); err != nil {
+		if pause.IsNoSuchFileError(err) || pause.UnsupportedPlatform(err) {
+			// If the pause container's image tarball doesn't exist or if the
+			// invocation is done for an unsupported platform, we cannot recover.
+			// Return the error as terminal for these cases
+			return fmt.Errorf("unable to load pause container image: %v", err), true
+		}
+		return fmt.Errorf("unable to load pause container image: %v", err), false
```
aaithal: I did not want to use that here because I don't think an error stack trace here provides much value beyond the `fmt.Errorf` wrapping.

agent/app/agent_unix.go
aaithal: Sure. That seems like a better name.

agent/app/agent_unix.go
aaithal: Sure. Those seem like better names.

agent/app/agent_unspecified.go
aaithal: I don't see a solid reason here for using that package.

agent/app/oswrapper/oswrapper.go
aaithal: May be. But, I'd rather have it here until there's a use-case for it.

agent/app/agent.go
```diff
@@ -176,8 +188,30 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+			if err != instanceNotLaunchedInVPCError {
+				seelog.Criticalf("Unable to initialize Task ENI dependencies: %v", err)
+				if terminal {
+					return exitcodes.ExitTerminal
+				}
+				return exitcodes.ExitError
+			}
+			// We have ascertained that the EC2 Instance is not running in a VPC
+			// No need to stop the ECS Agent in this case; all we need to do is
+			// to not update the config to disabled the TaskENIEnabled flag and
```
samuelkarp: nit: "to disable"

aaithal:
@richardpen I've addressed your comments. Can you please take another look?

agent/app/agent_unix.go
```diff
@@ -0,0 +1,156 @@
...
+// to configure the ENI for a task
+var awsVPCCNIPlugins = []string{ecscni.ECSENIPluginName,
+	ecscni.ECSBridgePluginName,
+	ecscni.ECSIPAMPluginName,
+}
+
+// initializeTaskENIDependencies initializes all of the dependencies required by
+// the Agent to support the 'awsvpc' networking mode. A non nil error is returned
+// if an error is encountered during this process. An additional boolean flag to
+// indicate if this error is considered terminal is also returned
```
samuelkarp: I'm wondering if we should make the return something like `ok, err` rather than `err, notOK`.  I think having `true` mean that it's not okay is weird to me.  I can be convinced that what you have here is reasonable enough.

agent/app/agent_unix.go
```diff
@@ -0,0 +1,156 @@
...
+		// the ENI plugin when there's no init process in the pid namesapce.
+		// Specifically, the DHClient processes that are started as children
+		// of the Agent will not be reaped leading to the ENI device
+		// disappearing until the Agent is killed.
+		return errors.New("agent is not started with an init system"), true
+	}
+
+	// Set VPC and Subnet IDs for the instance
+	if err, ok := agent.setVPCSubnet(); err != nil {
+		return err, ok
```
samuelkarp: (same as above) Using `ok` here where `true` indicates terminal is pretty confusing to me.

agent/app/agent_unix.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	return nil, false
+}
+
+// setVPCSubnet sets the vpc and subnet ids for the agent by querying the
+// instance metadata service
+func (agent *ecsAgent) setVPCSubnet() (error, bool) {
+	mac, err := agent.ec2MetadataClient.PrimaryENIMAC()
+	if err != nil {
```
samuelkarp: Aren't these errors all terminal?  Doesn't this indicate that we're not running in a VPC?

aaithal: even if the instance is running in classic EC2, it'll have a primary network interface and Instance Metadata Services (IMDS) should return a value for the primary network interfaces' mac request. I hesitate to treat all errors as terminal errors here because IMDS's availability can also lead to an error here, and retrying should (at some point) fix that.

agent/app/agent_unix.go
aaithal: I can see how `ok, err` is less confusing here. I can make that change.

aaithal: On second thought, `ok` has no meaning if `err` is not set. I take back my earlier statement :) Let's keep it `error, bool`

samuelkarp: `true` = terminal is still kinda confusing when a `bool` usually seems to be `ok` in go.  However...don't change it if you feel differently.  This is minor.

aaithal: We can make it a typed error, which will convey the intent more clearly. 

```
type initializeDependencyError {
 error
 terminal bool
}
```

agent/app/agent.go
```diff
@@ -176,8 +188,30 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		return exitcodes.ExitTerminal
 	}
 
+	var vpcSubnetAttributes []*ecs.Attribute
+	// Check if Task ENI is enabled
+	if agent.cfg.TaskENIEnabled {
+		if err, terminal := agent.initializeTaskENIDependencies(state, taskEngine); err != nil {
```
richardpen: Will use `switch` here makes it looks better?
```
err, terminal := agent.initializeTaskENIDependencies(state, taskEngine);
switch err{
	case nil:
		vpcSubnetAttributes = agent.constructVPCSubnetAttributes()
	case instanceNotLaunchedInVPCError:
		seelog.Criticalf("Unable to initialize Task ENI dependencies: %v", err)
		if terminal {
			return exitcodes.ExitTerminal
		}
		return exitcodes.ExitError
	default:
	seelog.Warnf("Unable to detect VPC ID for the Instance, disabling Task ENI capability: %v", err)
	agent.cfg.TaskENIEnabled = false
}
```

agent/app/agent_unix.go
```diff
@@ -0,0 +1,175 @@
...
+	if metadataErr, ok := err.(*ec2.MetadataError); ok &&
+		metadataErr.GetStatusCode() == http.StatusNotFound {
+		return false
+	}
+
+	return true
+}
+
+// verifyCNIPluginsCapabilities returns an error if there's an error querying
+// capabilities or if the required capability is absent from the capabilies
```
adnxn: typo: "capabilies"

agent/app/agent_unspecified.go
adnxn: out of curiosity - how are we deciding to use one over the other? are we using `github.com/pkg/errors` only when we want to include additional context?

agent/app/agent_unix.go
```diff
@@ -0,0 +1,175 @@
...
+			// If the pause container's image tarball doesn't exist or if the
+			// invocation is done for an unsupported platform, we cannot recover.
+			// Return the error as terminal for these cases
+			return err, true
+		}
+		return err, false
+	}
+
+	if err := agent.startUdevWatcher(state, taskEngine.StateChangeEvents()); err != nil {
+		// Inability to initialize the udev watcher could succeed if the error
```
aaithal: That's tech debt :P 

I'll modify it so that it makes more sense

agent/app/agent_unix.go
```diff
@@ -0,0 +1,175 @@
...
+	if err != nil {
+		if isInstanceLaunchedInVPC(err) {
+			return fmt.Errorf("unable to get vpc id from instance metadata: %v", err), true
+		}
+		return instanceNotLaunchedInVPCError, false
+	}
+
+	subnetID, err := agent.ec2MetadataClient.SubnetID(mac)
+	if err != nil {
+		return fmt.Errorf("unable to get vpc id from instance metadata: %v", err), false
```
aaithal: yes, Ctrl + C and Ctrl + V is hard

agent/app/agent_unspecified.go
aaithal: It depends is the short answer here. Please read https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully for a better explanation for this.

As a rule of thumb, if the error needs to be propagated multiple layers up, I tend to use `errors.Wrapf`.

agent/app/agent.go
aaithal: yeah, sure.

PR #917, opened 2017-08-07
Task Cgroups: Initial Changes

vsiddharth:
### Summary

Initial changes for task cgroups

### Testing

* Initial changes for task cgroups
### Licensing

This contribution is under the terms of the Apache 2.0 License:  Yes


vsiddharth:
We also need to fix [Appveyor](https://github.com/aws/amazon-ecs-agent/commit/01fccf076b8d674743c39f8e5cc4fdd1eaabcad8)

samuelkarp:
> We also need to fix Appveyor

I think you just mean fix the code to work on Windows.  AppVeyor looks to be failing correctly; those symbols don't exist on Windows.

agent/api/task_unix.go
```diff
@@ -15,10 +15,48 @@
...
+		return errors.Wrapf(err, "task set cgroup parent: unable to get valid cgroup spec")
+	}
+
+	// Check for empty cgroup root
+	if cgroupSpec.Root == "" {
+		return errors.New("task set cgroup parent: empty cgroup root")
+	}
+
+	// Set cgroup parent
+	hostConfig.CgroupParent = cgroupSpec.Root
```
petderek: This is still odd. You shouldn't be modifying external configuration from within a function on an unrelated type.

vsiddharth:
Appveyor build + test has been fixed.

agent/resources/cgroup/cgroup_controller_linux_test.go
```diff
@@ -0,0 +1,96 @@
...
+package cgroup
+
+import (
+	"testing"
+
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/stretchr/testify/assert"
+)
+
+// TODO: Add tests to cover happy paths
```
petderek: We may have duplicated a small amount of work here -- I was reworking and writing unit/integ tests for cgroup_controller_linux.

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,95 @@
...
+func validateCgroupSpec(cgroupSpec *Spec) error {
+	if cgroupSpec == nil {
+		return errors.New("cgroup spec validator: empty cgroup spec")
+	}
+
+	if cgroupSpec.Root == "" {
+		return errors.New("cgroup spec validator: invalid cgroup root")
+	}
+
+	if !filepath.HasPrefix(cgroupSpec.Root, config.DefaultTaskCgroupPrefix) {
```
petderek: I removed this in a branch -- I don't think it makes sense to have this configurable and then validate against the default.

agent/api/task_unix.go
aaithal: This definitely seems like reverse of how it should be. I don't think it should be a member of `Task` struct at all. Also, there's some more refactoring that you can do here to break this down and better convey what it's trying to do:
```
func (tas *Task) getCgSpecWitRoot() (cgroup.Spec, error) {
 cgroupSpec, err := task.GetCgroupSpec()
 if err != nil {
  ..
 }
 if cgroupSpec.Root == "" {
  return errors.New()
 }

 return cgroupSpec, nil
}
```

Then, you can just invoke this in `task.dockerHostConfig()` and not depend on this confusing method at all.

agent/engine/task_manager_linux.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+// SetupCgroup sets up the cgroup for each managed task
+func (mtask *managedTask) SetupCgroup() error {
+	// Grab cgroup spec
+	cgroupSpec, err := mtask.Task.GetCgroupSpec()
+	if err != nil {
+		return errors.Wrapf(err, "cgroup setup: unable to obtain cgroup spec")
+	}
+	seelog.Debugf("Setting up task cgroup %s", cgroupSpec.Root)
```
aaithal: You should also emit task arn here.

agent/engine/task_manager_linux.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// CleanupCgroup removes the task cgroup
+func (mtask *managedTask) CleanupCgroup() error {
+	// Grab cgroup spec
+	cgroupSpec, err := mtask.Task.GetCgroupSpec()
+	if err != nil {
+		return errors.Wrapf(err, "cgroup cleanup: unable to obtain cgroup spec")
+	}
+	seelog.Debugf("Cleaning up task cgroup %s", cgroupSpec.Root)
```
aaithal: Same here. Task arn needs to be logged.

agent/api/task.go
```diff
@@ -669,3 +688,11 @@ func (t *Task) String() string {
 	}
 	return res + "]"
 }
+
+// CgroupEnabled returns a bool when cgroupSpec is available
+func (task *Task) CgroupEnabled() bool {
+	task.cgroupSpecLock.RLock()
+	defer task.cgroupSpecLock.RUnlock()
+
+	return task.CgroupSpec != nil
```
aaithal: When/where is this being set?

agent/engine/task_manager.go
```diff
@@ -148,6 +149,22 @@ func (mtask *managedTask) overseeTask() {
...
 			llog.Debug("Task not steady state or terminal; progressing it")
+
+			// TODO:
+			// 1) Determine better way to check feature enablement and
+			// cgroup hierarchy setup
+			// 2) Add new task resources provisioned state
+			if mtask.Task.CgroupEnabled() {
+				err := mtask.SetupCgroup()
+				if err != nil {
+					seelog.Errorf("Unable to set up task cgroup: %v", err)
```
aaithal: Task arn needs to be logged here.

agent/engine/task_manager.go
```diff
@@ -148,6 +149,22 @@ func (mtask *managedTask) overseeTask() {
...
+					// If task cgroup creation fails, set task status to stopped
+					// with valid reason
+					mtask.SetKnownStatus(api.TaskStopped)
```
aaithal: We should not be modifying the task's known status to 'STOPPED' without also modifying it's desired status (which we can safely presume is 'RUNNING' here)

agent/engine/task_manager.go
```diff
@@ -535,6 +552,14 @@ func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 	// discard events while the task is being removed from engine state
 	go mtask.discardEventsUntil(handleCleanupDone)
 	mtask.engine.sweepTask(mtask.Task)
+
+	// Remove task cgroup
+	if mtask.Task.CgroupEnabled() {
+		err := mtask.CleanupCgroup()
+		if err != nil {
+			seelog.Warnf("Failed to cleanup task cgroup: %v", err)
```
aaithal: Task arn needs to be logged here.

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// load is used to load the cgroup based off the spec post validation
+func load(cgroupSpec *Spec) (cgroups.Cgroup, error) {
+	seelog.Debugf("Loading cgroup %s", cgroupSpec.Root)
+
+	// Validate incoming spec
+	err := validateCgroupSpec(cgroupSpec)
+	if err != nil {
+		return nil,
```
aaithal: These can be in a single line.

agent/engine/task_manager_linux_test.go
```diff
@@ -0,0 +1,47 @@
...
+func TestCleanupCgroupErrorPath(t *testing.T) {
+	mtask := managedTask{
+		Task: &api.Task{},
+	}
+
+	err := mtask.CleanupCgroup()
+	assert.Error(t, err)
+}
+
+// TODO: Add tests to cover happy paths
```
aaithal: When are you planning to do this? Can this be addressed in this PR so that we don't check in code that lacks test coverage? Doing it in another PR would diminish our view into the coverage of happy/unhappy paths i think.

agent/engine/task_manager.go
vsiddharth: Ok

agent/engine/task_manager_linux_test.go
vsiddharth: @petderek  has started work on this and already has some changes.

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: @petderek Sure, I'll remove this here too.

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: Ok

agent/resources/cgroup/cgroup_controller_linux_test.go
vsiddharth: @petderek I apologize. We can get rid of redundant things. Please feel free to remove things here that are unnecessary.

agent/api/task_unix.go
vsiddharth: I'll make this better as it's causing unnecessary confusion

agent/engine/task_manager.go
vsiddharth: Alright

agent/engine/task_manager_linux_test.go
samuelkarp: We should still not merge this until we have tests.

agent/Godeps/Godeps.json
```diff
@@ -327,10 +341,20 @@
 			"Rev": "2940d2e2e9365b4b3fec8ae24822435195f56b61"
 		},
 		{
+			"ImportPath": "github.com/opencontainers/runtime-spec/specs-go",
```
samuelkarp: I don't see attribution for this in the diff above.

agent/Godeps/Godeps.json
```diff
@@ -327,10 +341,20 @@
...
+			"Comment": "v1.0.0-rc5-255-gd349388",
+			"Rev": "d349388c43b01b2ea695965ae561b5bddb81318f"
+		},
+		{
 			"ImportPath": "github.com/pborman/uuid",
 			"Rev": "ca53cad383cad2479bbba7f7a1a05797ec1386e4"
 		},
 		{
+			"ImportPath": "github.com/pkg/errors",
```
samuelkarp: I don't see attribution for this in the diff above.

agent/api/task.go
```diff
@@ -110,6 +111,13 @@ type Task struct {
 	// used to look up the credentials for task in the credentials manager
 	credentialsID     string
 	credentialsIDLock sync.RWMutex
+
+	// cgroupSpecLock to reliably update the cgroup spec
+	cgroupSpecLock sync.RWMutex
+
+	// CgroupSpec attribute to capture the cgroup spec
+	// TODO: Inspect upon model changes
+	CgroupSpec *cgroup.Spec `json:omitempty`
```
samuelkarp: Can you order this like everything else in the file?  That means moving `cgroupSpecLock` directly below `CgroupSpec`.

Also, we discussed putting the `CgroupSpec` inside a platform-specific wrapper struct such that it wouldn't compile in on Windows.  Is there a reason you went away from that?

agent/api/task.go
```diff
@@ -409,12 +417,30 @@ func (task *Task) dockerHostConfig(container *Container, dockerContainerMap map[
...
+	// Populate hostConfig
 	hostConfig := &docker.HostConfig{
 		Links:        dockerLinkArr,
 		Binds:        binds,
 		PortBindings: dockerPortMap,
 		VolumesFrom:  volumesFrom,
 	}
+	// Update cgroup parent
+	// TODO: Feature gating
+	if task.CgroupEnabled() {
```
samuelkarp: This is another good place for platform-specific wrappers instead of a plain feature-gate.

agent/engine/task_manager_windows.go
```diff
@@ -0,0 +1,28 @@
...
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package engine
+
+// SetupCgroup sets up the cgroup for each managed task
+func (mtask *managedTask) SetupCgroup() error {
```
samuelkarp: Again, these should be in platform-specific wrappers such that we don't even compile in these symbols on Windows.

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Validate the linux resource specs
+	if cgroupSpec.Specs == nil {
+		return errors.New("cgroup spec validator: empty linux resource spec")
+	}
+	return nil
+}
+
+// Create creates a new cgroup based off the spec post validation
+func Create(cgroupSpec *Spec) error {
```
samuelkarp: Can you move `Create` above `validateCgroupSpec` so we have caller-above-callee?

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "cgroup load: failed to validate spec")
+	}
+
+	// Load cgroup
+	return cgroups.Load(cgroups.V1, cgroups.StaticPath(cgroupSpec.Root))
+}
+
+// Remove is used to delete the cgroup
+func Remove(cgroupSpec *Spec) error {
```
samuelkarp: Same here: `Remove` should be above `load`.

agent/resources/cgroup/init_unsupported.go
```diff
@@ -0,0 +1,22 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package cgroup
```
samuelkarp: The whole `cgroup` package should not be compiled on Windows at all.

agent/api/task.go
vsiddharth: I'll rearrange things here. 

I think `omitempty` should suffice here.

agent/engine/task_manager_windows.go
vsiddharth: This is a wrapper and I want these to be noop's on unsupported platforms.

agent/resources/cgroup/init_unsupported.go
vsiddharth: Sure, I can get rid of the unnecessary cruft here

agent/Godeps/Godeps.json
vsiddharth: Will be updated in the following revision, thanks

agent/Godeps/Godeps.json
vsiddharth: Will be updated in the following revision, thanks

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: Sure

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: Sure

agent/api/task.go
samuelkarp: `omitempty` still leaves the cgroup-specific things (which are Linux-specific) in common, platform-independent code.  Please encapsulate this in a Linux-specific wrapper.

petderek: @samuelkarp I'm curious -- why isn't this sufficient? The task def isn't going to be platform-dependent.

agent/engine/task_manager_windows.go
vsiddharth: @samuelkarp thanks

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,86 @@
...
+
+	// Validate incoming spec
+	err := validateCgroupSpec(cgroupSpec)
+	if err != nil {
+		return errors.Wrapf(err, "cgroup create: failed to validate spec")
+	}
+
+	// Create cgroup
+	_, err = cgroups.New(cgroups.V1, cgroups.StaticPath(cgroupSpec.Root), cgroupSpec.Specs)
+
```
richardpen: this blank line can be removed.

agent/resources/cgroup/cgroup_controller_linux_test.go
```diff
@@ -0,0 +1,86 @@
...
+// TestValidateCgroupSpecWithMissingRoot checks for missing cgroup root
+func TestValidateCgroupSpecWithMissingRoot(t *testing.T) {
+	cgroupSpec := Spec{}
+	err := validateCgroupSpec(&cgroupSpec)
+	assert.Error(t, err, "missing cgroup root")
+}
+
+// TestValidateCgroupSpecWithMissingResourceSpecs checks for cgroup spec with
+// missing linux resource specs
+func TestValidateCgroupSpecWithMissingResourceSpecs(t *testing.T) {
```
richardpen: It's a good place to use table-driven tests.

agent/api/task_test.go
```diff
@@ -800,3 +801,36 @@ func assertSetStructFieldsEqual(t *testing.T, expected, actual interface{}) {
 		}
 	}
 }
+
+// TestCgroupEnabledWithMissingSpec checks if task has cgroup
+func TestCgroupEnabledWithMissingSpec(t *testing.T) {
```
richardpen: This can be table-driven tests.

agent/config/config.go
```diff
@@ -80,6 +80,9 @@ const (
 	// minimumNumImagesToDeletePerCycle specifies the minimum number of images that to be deleted when
 	// performing image cleanup.
 	minimumNumImagesToDeletePerCycle = 1
+
+	// Default cgroup prefix for ECS tasks
+	DefaultTaskCgroupPrefix = "/ecs"
```
richardpen: Will this be configurable by environment variable or task definition, I didn't see that in the code?

agent/engine/task_manager.go
```diff
@@ -148,6 +149,18 @@ func (mtask *managedTask) overseeTask() {
...
+			// TODO: Add new task resources provisioned state ?
+			err := mtask.SetupPlatformResources()
+			if err != nil {
+				seelog.Errorf("Unable to set up resources for task %s: %v", mtask.Task.Arn, err)
+				// If task cgroup creation fails, set task status to stopped
+				// with valid reason
+				mtask.SetKnownStatus(api.TaskStopped)
+				mtask.SetDesiredStatus(api.TaskStopped)
+				mtask.engine.emitTaskEvent(mtask.Task, taskUnableToCreateCgroup)
+				return
```
richardpen: We shouldn't return here, I think this will cause the task status not reported to backend? You just need to set the desired status to be stopped.

agent/api/task.go
aaithal: As per @samuelkarp  comment, we definitely need to move this block to `task_unix.go`. We should not evaluate `if task.CgroupEnabled()` if running on windows.

agent/engine/task_manager.go
```diff
@@ -148,6 +149,18 @@ func (mtask *managedTask) overseeTask() {
 		if !mtask.GetKnownStatus().Terminal() {
 			// If we aren't terminal and we aren't steady state, we should be able to move some containers along
 			llog.Debug("Task not steady state or terminal; progressing it")
+
+			// TODO: Add new task resources provisioned state ?
+			err := mtask.SetupPlatformResources()
+			if err != nil {
```
aaithal: Do we have a unit test that's covering the task clean up behavior when unable to setup platform resources?

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,86 @@
...
+func Remove(cgroupSpec *Spec) error {
+	seelog.Debugf("Removing cgroup %s", cgroupSpec.Root)
+
+	control, err := load(cgroupSpec)
+	if err != nil {
+		return errors.Wrapf(err, "cgroup remove: unable to obtain controller")
+	}
+
+	// Delete cgroup
+	return control.Delete()
```
aaithal: Since you're wrapping other errors here, it makes sense to wrap this error as well.

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,86 @@
...
+	seelog.Debugf("Loading cgroup %s", cgroupSpec.Root)
+
+	// Validate incoming spec
+	err := validateCgroupSpec(cgroupSpec)
+	if err != nil {
+		return nil, errors.Wrapf(err, "cgroup load: failed to validate spec")
+	}
+
+	// Load cgroup
+	return cgroups.Load(cgroups.V1, cgroups.StaticPath(cgroupSpec.Root))
```
aaithal: Since you're wrapping other errors here, it makes sense to wrap this error as well.

agent/config/config.go
vsiddharth: This is the not going to be configurable now. 
The agent will create this prefix and stick all the tasks under here when the feature is enabled. 

The tasks that are started without this feature will be under the default cgroup root.

agent/engine/task_manager.go
vsiddharth: Alright, should we also emit an event here ?

agent/config/config.go
samuelkarp: This symbol should only exist on Linux.  Can you move it to `config_linux.go`?

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,73 @@
...
+
+package cgroup
+
+import (
+	"github.com/cihub/seelog"
+	"github.com/containerd/cgroups"
+	"github.com/pkg/errors"
+)
+
+var factory CgroupFactory = &GlobalCgroupFactory{}
```
samuelkarp: I don't have a different suggestion yet, but package-level variables like this are an anti-pattern I'd like to avoid as much as possible.

If you're only doing this to make testing possible, I think I can live with it but it needs a clear indication that it's only for tests.

agent/resources/cgroup/cgroup_controller_linux_test.go
```diff
@@ -0,0 +1,129 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package cgroup
+
+import (
+	"testing"
+
```
samuelkarp: nit: remove empty line

agent/resources/cgroup/cgroup_controller_linux.go
petderek: This is explicitly for testing, and it doesn't maintain any state. We can add a comment.

agent/config/config.go
vsiddharth: Done

agent/resources/cgroup/cgroup_controller_linux_test.go
vsiddharth: Done

agent/config/config.go
aaithal: I think @richardpen's asking if we should let it be configurable, which seems like a reasonable thing to do from an extensibility perspective.

agent/engine/task_manager_linux.go
```diff
@@ -0,0 +1,44 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package engine
+
+// SetupPlatformResources sets up platform level resources
+func (mtask *managedTask) SetupPlatformResources() error {
+	if mtask.Task.CgroupEnabled() {
```
aaithal: Can you move this check to https://github.com/aws/amazon-ecs-agent/pull/917/files#diff-20346f7af6202828de353f0137b6e43aR154 (to agent/engine/task_manager.go line 154)?

agent/engine/task_manager_linux.go
```diff
@@ -0,0 +1,44 @@
...
+		return mtask.setupCgroup()
+	}
+	return nil
+}
+
+// CleanupPlatformResources cleans up platform level resources
+func (mtask *managedTask) CleanupPlatformResources() error {
+	if mtask.Task.CgroupEnabled() {
```
aaithal: Same here. Can you move this check to wherever `CleanupPlatformResources()` is being invoked?

agent/resources/cgroup/cgroup_controller_linux.go
aaithal: Why can't this be a parameter that `Create` and `Remove` accept? It seems very weird and an anti-pattern as @samuelkarp  says.

agent/resources/cgroup/types_linux.go
```diff
@@ -0,0 +1,47 @@
...
+// Spec captures the abstraction for a creating a new
+// cgroup based on root and the runtime specifications
+type Spec struct {
+	// Root is the cgroup path
+	Root string
+	// Specs are for all the linux resources including cpu, memory, etc...
+	Specs *specs.LinuxResources
+}
+
+//go:generate go run ../../../scripts/generate/mockgen.go github.com/containerd/cgroups Cgroup mock/mock_cgroups_linux.go
```
aaithal: Can you move this to a `generate_mocks.go`? We're (at least I am) very used to looking at that file for any mock generation logic.

agent/resources/cgroup/types_linux.go
```diff
@@ -0,0 +1,47 @@
...
+type CgroupFactory interface {
```
aaithal: lint: `CgroupFactory is ...`

agent/resources/cgroup/types_linux.go
```diff
@@ -0,0 +1,47 @@
...
+	New(hierarchy cgroups.Hierarchy, path cgroups.Path, specs *specs.LinuxResources) (cgroups.Cgroup, error)
```
aaithal: lint: `New does things ...`

agent/resources/cgroup/types_linux.go
```diff
@@ -0,0 +1,47 @@
...
+	Load(hierarchy cgroups.Hierarchy, path cgroups.Path) (cgroups.Cgroup, error)
```
aaithal: lint: `Load does other things ...`

agent/resources/cgroup/types_linux.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// GlobalCgroupFactory calls the cgroups library global functions
+type GlobalCgroupFactory struct{}
+
+func (c *GlobalCgroupFactory) Load(hierarchy cgroups.Hierarchy, path cgroups.Path) (cgroups.Cgroup, error) {
```
aaithal: Comments here as well

agent/resources/cgroup/types_linux.go
```diff
@@ -0,0 +1,47 @@
...
+	return cgroups.Load(hierarchy, path)
+}
+
+func (c *GlobalCgroupFactory) New(hierarchy cgroups.Hierarchy, path cgroups.Path, specs *specs.LinuxResources) (cgroups.Cgroup, error) {
```
aaithal: Comments here as well

agent/engine/task_manager_linux.go
petderek: Its within this block because it is a platform-specific call. Is there a reason you'd rather have it outside this block?

agent/api/task_unix.go
```diff
@@ -22,3 +22,8 @@ const (
 func (task *Task) adjustForPlatform() {}
 
 func getCanonicalPath(path string) string { return path }
+
+func (task *Task) CgroupEnabled() bool {
```
petderek: I implemented this on my branch -- @vsiddharth only needed the stub.

agent/engine/task_manager_linux.go
petderek: Talked offline -- `CgroupEnabled()` is platform specific and needs this wrapper.

agent/engine/task_manager_linux.go
petderek: It may be useful to define what we mean by resource here, or use another term. The way we've been thinking about it is something like  "task-level resource that isn't managed by docker". 

We don't currently have an example outside of cgroups, though.

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/cihub/seelog"
+	"github.com/containerd/cgroups"
+	"github.com/pkg/errors"
+)
+
+// factory here is only to ease testing and it does not maintain any state
+var factory CgroupFactory = &GlobalCgroupFactory{}
+
+// Create creates a new cgroup based off the spec post validation
+func Create(cgroupSpec *Spec) (cgroups.Cgroup, error) {
```
richardpen: Can we have the [GlobalCgroupFactory](https://github.com/aws/amazon-ecs-agent/pull/917/files#diff-ab18a83c62201eea885f981fc8b94475R41) implement the `Create` and `Remove`, then we can have an obj that's responsible for creating/deleting the cgroup, also we can get rid of the global variable `CgroupFactory`? Though I don't have strong opposition of the current pattern, it just makes me feels better to have an obj to manage these process instead of calling it anywhere directly like the c style?

petderek: CgroupFactory makes it possible to test cgroup_controller_linux. We could (and probably should) have another obj thats responsible for create / remove, but we still can't get rid of CgroupFactory.

agent/engine/task_manager.go
vsiddharth: We only have stubs with this PR. 
I'll add tests to validate the setup and cleanup paths once we add `platformResources` to managedTask

agent/api/task_unix.go
vsiddharth: `CgroupEnabled` will perform sanity checks and check whether the cgroup code paths needs to be activated. We only have a stub here and @petderek probably meant that he's changed this method and implemented the actual functionality in his branch.

agent/engine/task_manager.go
aaithal: That still doesn't obviate the need for a test here. You have a code path that's dealing with an error being returned. So, having a test case to cover that is necessary.

agent/api/task_unix.go
aaithal: Alright, but why would that prevent a comment being added here?

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// control is used to implement the cgroup Control interface
+type control struct {
+	factory.CgroupFactory
+}
+
+// New is used to obtain a new cgroup control object
+func New() Control {
+	return _new(&factory.GlobalCgroupFactory{})
```
aaithal: Why is this `_new` and not `new`?

vsiddharth: It can be new but I'm used to the convention of using `_` for local functions.

aaithal: Please change this to `new` as we prefer not to use names that begin with `_` if we can help it in the Agent code base.

agent/api/task_unix.go
vsiddharth: Added comment

agent/engine/task_manager.go
vsiddharth: I'm sure that we need to test this path but we only have stubs with this PR and we don't have a way to obtain a `non-nil` condition here at this moment. 

Changes to api.Task will result in changes to task.CgroupEnabled() and all the stubs will be replaced with actual implementation. I'd like to focus this PR on `resources/cgroup`. 

This PR would be followed up with another with changes to managedTask and api.Task.

aaithal: >  we don't have a way to obtain a `non-nil` condition here at this moment. 

That probably points to either the lack of abstractions or wrong abstractions where-in `SetupPlatformResources` can return what we expect it to return to test the failure scenarios.  

> I'd like to focus this PR on `resources/cgroup`. 

If you're not planning to add tests to this code path in this PR, please remove that and add it back with tests in the PR where you plan to add tests.

petderek: This logic won't change, right? Regardless of how 'setupPlatformResources' works. We can test the logic within this function and mock 'setupPlatformResources'.

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: `new` is a keyword and I'd like to avoid using it here to mangle the meaning. 
`_new` just makes is a better fit imho.

agent/engine/task_manager.go
vsiddharth: @petderek I'd vouch for getting rid of the changes to the task_manager now and following up  the changes to the task_manager once we define `platformResources` in a separate PR.

agent/resources/cgroup/cgroup_controller_linux.go
samuelkarp: Use `newControl`.  Don't use `_new`.

samuelkarp: We don't have a convention of using `_` for local functions.

agent/engine/task_manager.go
petderek: 👍 works for me

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: @samuelkarp thanks! I've updated `_new` -> `newControl`

agent/engine/task_manager.go
vsiddharth: @petderek thanks!

agent/resources/cgroup/cgroup_controller_linux_test.go
```diff
@@ -0,0 +1,146 @@
...
+func TestCreateHappyCase(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockCgroup := mock_cgroups.NewMockCgroup(ctrl)
+	mockCgroupFactory := mock_factory.NewMockCgroupFactory(ctrl)
+
+	mockCgroupFactory.EXPECT().New(gomock.Any(), gomock.Any(), gomock.Any()).Return(mockCgroup, nil)
+
+	testString := "/ecs/foo"
```
richardpen: can you make this a constant, since it was used in multiple places.

agent/resources/cgroup/init_linux_test.go
```diff
@@ -0,0 +1,56 @@
...
+	defer ctrl.Finish()
+
+	mockCgroupFactory := mock_factory.NewMockCgroupFactory(ctrl)
+
+	mockCgroupFactory.EXPECT().New(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, errors.New("cgroup error"))
+
+	control := newControl(mockCgroupFactory)
+
+	err := control.Init()
+	assert.Error(t, err)
```
richardpen: minor: these two lines can be combined together as `assert.Error(t, control.Init())`, multiple places.

agent/engine/task_manager_linux_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"testing"
+
+	"github.com/aws/amazon-ecs-agent/agent/api"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestSetupCgroupWithInvalidSpec checks if managedTask can be setup with an
+// invalid cgroup spec
+func TestSetupCgroupWithInvalidSpec(t *testing.T) {
+	t.Skip()
```
vsiddharth: Yes, we plan to have a cgroupSpec builder in the following PR. 
I can get rid of this in this PR if its bothering you

agent/resources/cgroup/cgroup_controller_linux_test.go
vsiddharth: Sure

agent/resources/cgroup/init_linux_test.go
vsiddharth: Sure, but I just did not want to combine them.

agent/engine/task_manager_linux_test.go
vsiddharth: @richardpen removed these for now

agent/resources/cgroup/cgroup_controller_linux_test.go
vsiddharth: Done

agent/resources/cgroup/init_linux_test.go
vsiddharth: Combined these

agent/engine/task_manager_linux_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+// TestSetupPlatformResourcesWithCgroupDisabled checks if platform resources
+// can be setup without errors when task cgroups are disabled
+func TestSetupPlatformResourcesWithCgroupDisabled(t *testing.T) {
+	mtask := managedTask{
+		Task: &api.Task{},
+	}
+
+	err := mtask.SetupPlatformResources()
+	assert.Nil(t, err)
```
aaithal: Please change this to `assert.NoError`

agent/engine/task_manager_linux_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+// TestCleanupPlatformResourcesWithCgroupDisabled checks if platform resources
+// can be cleaned up without errors when task cgroups are disabled
+func TestCleanupPlatformResourcesWithCgroupDisabled(t *testing.T) {
+	mtask := managedTask{
+		Task: &api.Task{},
+	}
+
+	err := mtask.CleanupPlatformResources()
+	assert.Nil(t, err)
```
aaithal: Please change this to `assert.NoError`

agent/engine/task_manager_linux_test.go
vsiddharth: Done

agent/engine/task_manager_linux_test.go
vsiddharth: Done

agent/engine/task_manager_linux.go
samuelkarp: > We don't currently have an example outside of cgroups, though.

ENIs are, but are being handled differently right now.  We'll want to converge this in the future so that we can have something a bit more consistent.  (We're cheating with ENIs because we *do* have a Docker container involved that's necessary for Docker to share namespaces between containers.)

agent/resources/cgroup/factory/factory_linux.go
```diff
@@ -0,0 +1,42 @@
...
+// CgroupFactory wraps around the global functions exposed by the cgroups library
+type CgroupFactory interface {
+	// New is used to create a new cgroup hierarchy
+	New(hierarchy cgroups.Hierarchy, path cgroups.Path, specs *specs.LinuxResources) (cgroups.Cgroup, error)
+	// Load is used to load the cgroup based off the cgroup path
+	Load(hierarchy cgroups.Hierarchy, path cgroups.Path) (cgroups.Cgroup, error)
+}
+
+// GlobalCgroupFactory calls the cgroups library global functions
+type GlobalCgroupFactory struct{}
```
samuelkarp: I believe this can be private if you expose something like this:

```go
func Default() CgroupFactory {
	return &globalCgroupFactory{}
}
```

PR #918, opened 2017-08-08
A set of changes aimed at improving readability

samuelkarp:
### Summary
Cleans up the agent code a bit

### Implementation details
Added comments, extracted sections into functions, added some tests

### Testing

None

### Licensing
This contribution is under the terms of the Apache 2.0 License: yes (Amazon employee)

### DIFF
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -242,34 +242,34 @@ func (task *Task) UpdateMountPoints(cont *Container, vols map[string]string) {
 	}
 }
 
-// updateTaskKnownState updates the given task's status based on its container's status.
+// updateTaskKnownStatus updates the given task's status based on its container's status.
 // It updates to the minimum of all containers no matter what
-// It returns a TaskStatus indicating what change occured or TaskStatusNone if
+// It returns a TaskStatus indicating what change occurred or TaskStatusNone if
 // there was no change
+// Invariant: task known status is the minimum of container known status
 func (task *Task) updateTaskKnownStatus() (newStatus TaskStatus) {
-	llog := log.New("task", task)
-	llog.Debug("Updating task")
+	seelog.Debug("Updating task %v", task)
 
 	// Set to a large 'impossible' status that can't be the min
 	earliestStatus := ContainerZombie
 	essentialContainerStopped := false
-	for _, cont := range task.Containers {
-		contKnownStatus := cont.GetKnownStatus()
-		if contKnownStatus == ContainerStopped && cont.Essential {
+	for _, container := range task.Containers {
+		containerKnownStatus := container.GetKnownStatus()
+		if containerKnownStatus == ContainerStopped && container.Essential {
 			essentialContainerStopped = true
 		}
-		if contKnownStatus < earliestStatus {
-			earliestStatus = contKnownStatus
+		if containerKnownStatus < earliestStatus {
+			earliestStatus = containerKnownStatus
 		}
 	}
 
 	// If the essential container is stopped while other containers may be running
 	// don't update the task status until the other containers are stopped.
 	if earliestStatus == ContainerRunning && essentialContainerStopped {
-		llog.Debug("Essential container is stopped while other containers are running, not update task status")
+		seelog.Debug("Essential container is stopped while other containers are running, not updating task status, task: %v", task)
 		return TaskStatusNone
 	}
-	llog.Debug("Earliest status is " + earliestStatus.String())
+	seelog.Debug("Earliest status is %q for task %v", earliestStatus.String(), task)
 	if task.GetKnownStatus() < earliestStatus.TaskStatus() {
 		task.SetKnownStatus(earliestStatus.TaskStatus())
 		return task.GetKnownStatus()
@@ -550,7 +550,8 @@ func (task *Task) UpdateDesiredStatus() {
 	task.updateContainerDesiredStatus()
 }
 
-// updateTaskDesiredStatus determines what status the task should properly be at based on its container's statuses
+// updateTaskDesiredStatus determines what status the task should properly be at based on the containers' statuses
+// Invariant: task desired status must be stopped if any essential container is stopped
 func (task *Task) updateTaskDesiredStatus() {
 	llog := log.New("task", task)
 	llog.Debug("Updating task")
@@ -567,6 +568,7 @@ func (task *Task) updateTaskDesiredStatus() {
 
 // updateContainerDesiredStatus sets all container's desired status's to the
 // task's desired status
+// Invariant: container desired status is <= task desired status converted to container status
 func (task *Task) updateContainerDesiredStatus() {
 	for _, c := range task.Containers {
 		taskDesiredStatus := task.GetDesiredStatus()
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -248,7 +248,7 @@ func (task *Task) UpdateMountPoints(cont *Container, vols map[string]string) {
 // there was no change
 // Invariant: task known status is the minimum of container known status
 func (task *Task) updateTaskKnownStatus() (newStatus TaskStatus) {
-	seelog.Debug("Updating task %v", task)
+	seelog.Debug("Updating task: %s", task.String())
 
 	// Set to a large 'impossible' status that can't be the min
 	earliestStatus := ContainerZombie
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -87,7 +87,7 @@ func dependenciesCanBeResolved(target *api.Container, by []*api.Container) bool 
 
 // DependenciesAreResolved validates that the `target` container can be started
 // given the current known state of the containers in `by`. If this function
-// returns true, `target` should be technically able to launch with on issues
+// returns true, `target` should be technically able to launch without issues
 func DependenciesAreResolved(target *api.Container, by []*api.Container) bool {
 	nameMap := make(map[string]*api.Container)
 	for _, cont := range by {
@@ -153,35 +153,33 @@ func linkIsResolved(target *api.Container, link *api.Container) bool {
 
 func volumeCanResolve(target *api.Container, volume *api.Container) bool {
 	targetDesiredStatus := target.GetDesiredStatus()
-	volumeDesiredStatus := volume.GetDesiredStatus()
-	if targetDesiredStatus == api.ContainerCreated || targetDesiredStatus == api.ContainerRunning {
-		return volumeDesiredStatus == api.ContainerCreated ||
-			volumeDesiredStatus == api.ContainerRunning ||
-			volumeDesiredStatus == api.ContainerStopped
+	if targetDesiredStatus != api.ContainerCreated && targetDesiredStatus != api.ContainerRunning {
+		log.Error("Unexpected desired status", "target", target)
+		return false
 	}
 
-	log.Error("Unexpected desired status", "target", target)
-	return false
+	volumeDesiredStatus := volume.GetDesiredStatus()
+	return volumeDesiredStatus == api.ContainerCreated ||
+		volumeDesiredStatus == api.ContainerRunning ||
+		volumeDesiredStatus == api.ContainerStopped
 }
 
 func volumeIsResolved(target *api.Container, volume *api.Container) bool {
 	targetDesiredStatus := target.GetDesiredStatus()
-	if targetDesiredStatus == api.ContainerCreated || targetDesiredStatus == api.ContainerRunning {
-		knownStatus := volume.GetKnownStatus()
-		return knownStatus == api.ContainerCreated ||
-			knownStatus == api.ContainerRunning ||
-			knownStatus == api.ContainerStopped
+	if targetDesiredStatus != api.ContainerCreated && targetDesiredStatus != api.ContainerRunning {
+		log.Error("Unexpected desired status", "target", target)
+		return false
 	}
 
-	log.Error("Unexpected desired status", "target", target)
-	return false
+	knownStatus := volume.GetKnownStatus()
+	return knownStatus == api.ContainerCreated ||
+		knownStatus == api.ContainerRunning ||
+		knownStatus == api.ContainerStopped
 }
 
 // onRunIsResolved defines a relationship where a target cannot be created until
 // 'run' has reached a running state.
 func onRunIsResolved(target *api.Container, run *api.Container) bool {
-	if target.GetDesiredStatus() >= api.ContainerCreated {
-		return run.GetKnownStatus() >= api.ContainerRunning
-	}
-	return false
+	return target.GetDesiredStatus() >= api.ContainerCreated &&
+		run.GetKnownStatus() >= api.ContainerRunning
 }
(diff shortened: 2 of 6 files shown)


agent/engine/task_manager.go
```diff
@@ -100,53 +100,32 @@ func (engine *DockerTaskEngine) newManagedTask(task *api.Task) *managedTask {
...
-			if mtask.GetDesiredStatus().Terminal() {
-				// If we end up here, that means we recieved a start then stop for this
-				// task before a task that was expected to stop before it could
-				// actually stop
-				break
-			}
-		}
-		llog.Debug("Wait over; ready to move towards status: " + mtask.GetDesiredStatus().String())
-	}
+	// Wait for	host resources required by this task to become available
```
richardpen: I think this tab is not intended.

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
+	maxWait := make(chan bool, 1)
+	timer := mtask.time().After(steadyStateTaskVerifyInterval)
+	go func() {
+		<-timer
+		maxWait <- true
+	}()
+	timedOut := mtask.waitEvent(maxWait)
+
+	if timedOut {
+		llog.Debug("Checking task to make sure it's still at steadystate")
```
aaithal: I know this is cruft from the way this code used to be earlier. But, should this log statement be moved to the `CheckTaskState` method itself?

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
+		go mtask.engine.CheckTaskState(mtask.Task)
+	}
+	return
```
aaithal: We can get rid of this blank `return`

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
+}
+
+// cleanupCredentials removes credentials for a stopped task
+func (mtask *managedTask) cleanupCredentials() {
+	taskCredentialsID := mtask.GetCredentialsID()
+	if taskCredentialsID != "" {
+		mtask.engine.credentialsManager.RemoveCredentials(taskCredentialsID)
+	}
+	return
```
aaithal: We can get rid of this blank `return`

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
 	for _, container := range mtask.Containers {
 		mtask.engine.emitContainerEvent(mtask.Task, container, "")
 	}
 	mtask.engine.emitTaskEvent(mtask.Task, "")
 }
 
+// waitForResources waits for host resources to become available to start the
+// task. This involves waiting for previous stops to complete so the resources
+// become free.
+func (mtask *managedTask) waitForResources() {
```
aaithal: Can we be more explicit and name this `waitForHostResources()`?

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
+	llog := log.New("task", mtask.Task)
+	if mtask.StartSequenceNumber != 0 && !mtask.GetDesiredStatus().Terminal() {
+		llog.Debug("Waiting for any previous stops to complete", "seqnum", mtask.StartSequenceNumber)
```
aaithal: This should be an `Info` I think

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
+		}()
+		for !mtask.waitEvent(othersStopped) {
+			if mtask.GetDesiredStatus().Terminal() {
+				// If we end up here, that means we received a start then stop for this
+				// task before a task that was expected to stop before it could
+				// actually stop
+				break
+			}
+		}
+		llog.Debug("Wait over; ready to move towards status: " + mtask.GetDesiredStatus().String())
```
aaithal: `Info` is better here as well

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
+// event, or a timeout.
+func (mtask *managedTask) waitSteady() {
+	llog := log.New("task", mtask.Task)
+	llog.Debug("Task at steady state", "state", mtask.GetKnownStatus().String())
+
+	maxWait := make(chan bool, 1)
+	timer := mtask.time().After(steadyStateTaskVerifyInterval)
+	go func() {
+		<-timer
+		maxWait <- true
```
petderek: can we just use timer? using a goroutine feels odd.

agent/api/task.go
```diff
@@ -242,34 +242,34 @@ func (task *Task) UpdateMountPoints(cont *Container, vols map[string]string) {
...
+// updateTaskKnownStatus updates the given task's status based on its container's status.
 // It updates to the minimum of all containers no matter what
-// It returns a TaskStatus indicating what change occured or TaskStatusNone if
+// It returns a TaskStatus indicating what change occurred or TaskStatusNone if
 // there was no change
+// Invariant: task known status is the minimum of container known status
 func (task *Task) updateTaskKnownStatus() (newStatus TaskStatus) {
-	llog := log.New("task", task)
-	llog.Debug("Updating task")
+	seelog.Debug("Updating task %v", task)
```
samuelkarp: The prior code dumped the whole struct.  There's value in it to see what's being updated, but I could also be convinced to make it more targeted if you felt strongly about it.

agent/engine/task_manager.go
samuelkarp: (this is moved code)

This does use a timer: `time.After`.  What's different is that `time.After` returns a `chan time.Time` but we need a `chan bool` for the call below this.  The goroutine converts the `time.Time` to a `bool` when the channel receives.

agent/engine/task_manager.go
samuelkarp: Existing code has it at `Debug`.  I agree that `Info` makes more sense, will change.

agent/engine/task_manager.go
samuelkarp: Existing code has it at `Debug`.  I agree that `Info` makes more sense, will change.

agent/engine/task_manager.go
samuelkarp: No, I don't think so. The log line makes sense to me here, because it explains the reason as to why we're invoking `CheckTaskState`. While `CheckTaskState` today is only invoked from here, it doesn't actually have any logic that special-cases steady-state versus any other state; it's really just about reading the current state from Docker regardless of what the agent already knows.

agent/engine/task_manager.go
```diff
@@ -327,77 +346,76 @@ func (mtask *managedTask) handleStoppedToRunningContainerTransition(status api.C
...
 }
 
-type containerTransitionFunc func(container *api.Container, nextStatus api.ContainerStatus)
-
 // progressContainers tries to step forwards all containers that are able to be
 // transitioned in the task's current state.
 // It will continue listening to events from all channels while it does so, but
 // none of those changes will be acted upon until this set of requests to
 // docker completes.
+// If container changes
```
jhaynes: What?

agent/engine/task_manager.go
```diff
@@ -327,77 +346,76 @@ func (mtask *managedTask) handleStoppedToRunningContainerTransition(status api.C
...
 func (mtask *managedTask) progressContainers() {
-	log.Debug("Progressing task", "task", mtask.Task)
+	seelog.Debug("Progressing task %v", mtask.Task)
```
jhaynes: :+1:

agent/engine/task_manager.go
```diff
@@ -420,15 +438,19 @@ func (mtask *managedTask) progressContainers() {
 	// complete, but keep reading events as we do.. in fact, we have to for
 	// transitions to complete
 	mtask.waitForContainerTransitions(transitions, transitionChange, transitionChangeContainer)
-	log.Debug("Done transitioning all containers for task", "task", mtask.Task)
+	seelog.Debug("Done transitioning all containers for task %v", mtask.Task)
 
-	if mtask.UpdateStatus() {
-		log.Debug("Container change also resulted in task change")
+	// update the task status
+	changed := mtask.UpdateStatus()
```
jhaynes: Why split this into two lines? `UpdateStatus` is just a boolean without side effects right?

agent/engine/task_manager.go
```diff
@@ -451,6 +473,50 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
 }
 
+type containerTransitionFunc func(container *api.Container, nextStatus api.ContainerStatus)
+
+// containerNextState determines the next state a container should go to.
+// It returns: The state it should transition to, a bool indicating whether any
+// action is required, and a bool indicating whether a known status change is
+// possible.
+//
+// 'Stopped, false, true' -> "You can move it to known stopped, but you don't have to call a transition function"
```
jhaynes: Are these all the cases? It isn't clear if these are examples or an exhaustive list. In either case, super helpful to have!

agent/engine/task_manager.go
```diff
@@ -451,6 +473,50 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+// 'Running, true, true' -> "You can move it to running and you need to call the transition function"
+// 'None, false, false' -> "This should not be moved; it has unresolved dependencies or is complete; no knownstatus change"
+//
+// Next status is determined by whether the known and desired statuses are
+// equal, the next numerically greater (iota-wise) status, and whether
+// dependencies are fully resolved.
+func (mtask *managedTask) containerNextState(container *api.Container) (api.ContainerStatus, bool, bool) {
+	clog := log.New("task", mtask.Task, "container", container)
```
jhaynes: `clog` and `seelog` sounds like a confusing plan. Per my comment above, can we standardize on one at least where we're touching code?

agent/engine/task_manager.go
samuelkarp: Thanks, fixed.

agent/engine/task_manager.go
samuelkarp: Done.

agent/engine/task_manager.go
samuelkarp: Done.

agent/engine/task_manager.go
samuelkarp: Done.

agent/api/task.go
```diff
@@ -550,7 +550,8 @@ func (task *Task) UpdateDesiredStatus() {
 	task.updateContainerDesiredStatus()
 }
 
-// updateTaskDesiredStatus determines what status the task should properly be at based on its container's statuses
+// updateTaskDesiredStatus determines what status the task should properly be at based on the containers' statuses
+// Invariant: task desired status should be stopped if any essential container is stopped
```
samuelkarp: Done.

agent/engine/task_manager.go
```diff
@@ -159,31 +138,110 @@ func (mtask *managedTask) overseeTask() {
...
+		mtask.engine.credentialsManager.RemoveCredentials(taskCredentialsID)
+	}
+	return
+}
+
+// waitEvent waits for any event to occur. If an event occurs, the appropriate
+// handler is called. If the event is the passed in channel, it will return the
+// value written to the channel, otherwise it will return false.
+func (mtask *managedTask) waitEvent(stopWaiting <-chan bool) bool {
+	log.Debug("Waiting for event for task", "task", mtask.Task)
```
samuelkarp: I'd rather not. I replaced things with `seelog` when it was easy but didn't want to get too far down that rabbit hole in this specific PR.

jhaynes: Understandable, but I'd argue that a PR that is addressing readability is a great place to include work on standardizing some of the logging chaos.

agent/engine/task_manager.go
samuelkarp: `mtask.UpdateStatus()` returns a `bool` indicating whether the task status changed or not, but it was difficult for me to see that.  I think the `changed` variable makes that more clear.

agent/engine/task_manager.go
samuelkarp: Examples, not exhaustive.

agent/engine/task_manager.go
samuelkarp: Fixed.

agent/api/task.go
```diff
@@ -242,34 +242,34 @@ func (task *Task) UpdateMountPoints(cont *Container, vols map[string]string) {
...
+		if containerKnownStatus < earliestStatus {
+			earliestStatus = containerKnownStatus
 		}
 	}
 
 	// If the essential container is stopped while other containers may be running
 	// don't update the task status until the other containers are stopped.
 	if earliestStatus == ContainerRunning && essentialContainerStopped {
-		llog.Debug("Essential container is stopped while other containers are running, not update task status")
+		seelog.Debug("Essential container is stopped while other containers are running, not update task status, task: %v", task)
```
samuelkarp: Done.

agent/engine/dependencygraph/graph.go
```diff
@@ -87,7 +87,7 @@ func dependenciesCanBeResolved(target *api.Container, by []*api.Container) bool
 
 // DependenciesAreResolved validates that the `target` container can be started
 // given the current known state of the containers in `by`. If this function
-// returns true, `target` should be technically able to launch with on issues
+// returns true, `target` should be technically able to launch with out issues
```
samuelkarp: Done.

agent/engine/task_manager.go
samuelkarp: I'd rather not do that at this point.  I made these changes as I was reading through the engine code in preparation for a different set of changes and not as a general cleanup; I don't want to spend more time on this right now.

agent/engine/task_manager.go
petderek: [this isn't blocking, I'm just curious if this is a common idiom]
I understand what's happening, but its weird to me that we need to have two async things happening to trigger one event.

agent/api/task.go
nmeyerhans: I'm not a fan of dumping a whole struct into the logs. At least this is at debug severity, but if we can limit the output to fields that are actually going to be meaningful, it would be preferable.

aaithal: `... :%s ", task.String()` is an acceptable middle ground here I think.

samuelkarp: I can do that.

samuelkarp: Done.

agent/engine/task_manager.go
samuelkarp: It isn't exactly a common idiom.  Part of the complication is that `waitEvent` is actually doing a `select` on three different channels in order to see if either an event occurred (a message from ACS or from Docker) or if it's canceled. If canceled, it returns the `bool` that was passed through the channel; otherwise it returns `false`.

I looked at moving this over to `context.Context` entirely and using a deadline, but that ended up being more work than I wanted to put in (we have some tests that simulate a timeout and I'd have to change those too).

petderek: Got it. Thanks for the explanation! :shipit:

PR #924, opened 2017-08-11
Windows MemorySwappiness Work Around

cabbruzzese:
### Summary
Adds workaround to set MemorySwappiness to -1 for Windows

### Implementation details
HostConfig is retrieved from the task and deep copied to a new struct. Platform based values are then overridden for this copy of the HostConfig struct.

### Testing
New tests added:
task_windows_test.go - TestWindowsPlatformHostConfigOverride
task_windows_test.go - TestWindowsMemorySwappinessOption

Bug - Workaround for MemorySwapiness bug for Windows

### Licensing


agent/api/task_windows.go
```diff
@@ -50,3 +53,14 @@ func (task *Task) downcaseAllVolumePaths() {
...
 }
+
+// platformHostConfigOverride Sets up default HostConfig options to be passed to Docker API.
+// Docker for Windows treats some options different than Unix and some are not supported at all.
+// Version 1.12.x of Docker for Windows would ignore the unsupported option MemorySwappiness.
+// Version 17.03.x will cause an error if any value other than -1 is passed in for MemorySwappiness.
+// This bug is not noticed when no value is passed in. However, the go-dockerclient client version
+// we are using removed the json option omitempty causing this parameter to default to 0 if empty.
+// https://github.com/fsouza/go-dockerclient/commit/72342f96fabfa614a94b6ca57d987eccb8a836bf
+func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig) {
```
samuelkarp: nit: I'd probably nest this just a little bit deeper so that the comment you have can be on the part that actually talks about what it does.  Something like this:

```go
// platformHostConfigOverrides applies platform-specific overrides to HostConfig options.
func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig) {
	task.forceDefaultMemorySwappiness(hostConfig)
}

// forceDefaultMemorySwappiness sets the MemorySwappiness field to be -1
// Docker for Windows treats some options differently than Linux [...]
func (task *Task) forceDefaultMemorySwappiness(hostConfig *docker.HostConfig) {
	hostConfig.MemorySwappiness = memorySwappinessDefault
}

File: agent/api/task_windows.go [hunk 2]
```diff
@@ -50,3 +53,14 @@ func (task *Task) downcaseAllVolumePaths() {
...
+	hostConfig.MemorySwappiness = memorySwappinessDefault
```
petderek: I don't like this pattern of passing a reference to data and then modifying it, particularly when we can avoid it.

agent/api/task_windows_test.go
```diff
@@ -99,3 +103,60 @@ func TestPostUnmarshalWindowsCanonicalPaths(t *testing.T) {
 	assert.Equal(t, expectedTask.Containers, task.Containers, "Containers should be equal")
 	assert.Equal(t, expectedTask.Volumes, task.Volumes, "Volumes should be equal")
 }
+
+func TestWindowsPlatformHostConfigOverride(t *testing.T) {
```
petderek: You could use subtests here.

petderek: actually, i lied. don't do that here :)

agent/api/task_windows.go
cabbruzzese: Would you recommend returning a new copy that has been modified, or moving the method to be attached to hostConfig (i.e. hostConfig.platformOverride())

agent/api/task_unix.go
```diff
@@ -15,10 +15,17 @@
 
 package api
 
+import (
+	"github.com/fsouza/go-dockerclient"
+)
+
 const (
 	portBindingHostIP = "0.0.0.0"
+	memorySwappinessDefault = 0
```
cabbruzzese: It's required to be defined for a test in task_test.go. Adding a comment

agent/api/task_windows.go
petderek: We talked offline -- this is fine. It'd be nice to refactor this to have a translator class for task api -> host config... but thats outside the scope here.

samuelkarp:
```
Abstracted platformHostConfigOverride in task_windows.go to another
method to clarify the intention of the override. Updated related
comments regarding injection point for HostConfig override values.

Commented on task_unix.go to explain the use of memorySwappinessDefault
constant.

Cleaned up extra lines in task_windows_test.go. Removed unneeded
properties in task object use in TestWindowsPlatformHostConfigOverride.
```

Your current commit message describes what the changes are, but not why.  I'd prefer it was the other way around. :smile:

PR #925, opened 2017-08-14
add repoint-latest script

jhaynes:
### Summary
This adds a script to move the latest pointers from our S3 bucket and from Docker Hub to the prior agent release version

### Implementation details
Done via a bash script

### Testing
Tested against a private S3 bucket and a private Docker Hub repo

Add script to un-publish an agent version.

### Licensing

### DIFF
--- a/scripts/repoint-latest
+++ b/scripts/repoint-latest
@@ -96,6 +96,9 @@ unpublish_docker() {
 		exit 1
 	fi
 
+	echo "Removing any amazon/amazon-ecs-agent images for safety"
+	docker images -q amazon/amazon-ecs-agent | xargs -r -L 1 docker rmi -f
+
 	docker load < "${tarball}"
 
 	echo "Tagging as ${IMAGE_NAME}:${IMAGE_TAG_LATEST}"
--- a/scripts/repoint-latest
+++ b/scripts/repoint-latest
@@ -1,0 +1,175 @@
+#!/bin/bash
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the
+# "License"). You may not use this file except in compliance
+#  with the License. A copy of the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+DRYRUN=true
+NUCLEAR=false
+
+IMAGE_NAME="amazon/amazon-ecs-agent"
+
+AWS_PROFILE=""
+PUBLISH_S3_BUCKET="amazon-ecs-agent"
+S3_ACL_OVERRIDE=""
+
+source $(dirname "${0}")/publishing-common.sh
+
+TARGET_OS_LIST=${SUPPORTED_OSES[@]}
+
+usage() {
+	echo "Usage: ${0} -s BUCKET [OPTIONS]"
+	echo
+	echo "This script is responsible for un-publishing a given release of the ECS Agent."
+	echo		
+	echo "Options"
+	echo "  -d  true|false      Dryrun (default is true)"
+	echo "  -b  BAD_VERSION     Version that will be unpublished"
+	echo "  -g  GOOD_VERSION    Version that will be re-tagged as latest"
+	echo "  -p  PROFILE         AWS CLI Profile (default is none)"
+	echo "  -n                  Nuclear option: delete the current version rather than just moving the latest pointer"
+	echo "  -h                  Display this help message"
+}
+
+unpublish_s3() {
+	if [ "$1" == "windows" ]; then
+	     	base_name="ecs-agent-windows"
+		suffix="zip"
+	else
+		base_name="ecs-agent"
+		suffix="tar"
+	fi
+
+	tarball_bad="$(mktemp)"
+	tarball_latest="$(mktemp)"
+
+	s3_cp "s3://${PUBLISH_S3_BUCKET}/ecs-agent-${BAD_VERSION}.tar" "${tarball_bad}"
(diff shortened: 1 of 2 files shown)


scripts/repoint-latest
```diff
@@ -0,0 +1,175 @@
...
+	echo "Usage: ${0} -s BUCKET [OPTIONS]"
+	echo
+	echo "This script is responsible for un-publishing a given release of the ECS Agent."
+	echo		
+	echo "Options"
+	echo "  -d  true|false      Dryrun (default is true)"
+	echo "  -b  BAD_VERSION     Version that will be unpublished"
+	echo "  -g  GOOD_VERSION    Version that will be re-tagged as latest"
+	echo "  -p  PROFILE         AWS CLI Profile (default is none)"
+	echo "  -n                  Nuclear option: delete the current version rather than just moving the latest pointer"
```
aaithal: When/why would we want to do this?

jahkeup: Make the user spell out `--nuclear` please, we don't want any confusion or accidental apocalypses. Better yet would be to make this more clear with a name like `--remove`.

scripts/repoint-latest
```diff
@@ -0,0 +1,175 @@
...
+	if ! check_md5 "${tarball_bad}" test_md5="$(md5sum ${tarball_latest} | sed 's/ .*//')" ; then
+		echo "The bad version ${BAD_VERSION} is not currently tagged as latest"
+		rm "${tarball_bad}"
+		rm "${tarball_latest}"
+		exit 1
+	fi
+
+	set +e
+	s3_ls "s3://${PUBLISH_S3_BUCKET}/${base_name}-${GOOD_VERSION}.${suffix}" &> /dev/null
+	if [ $? -ne 0 ]; then
```
aaithal: Isn't this redundant because of `set +e` ?

scripts/repoint-latest
```diff
@@ -0,0 +1,175 @@
...
+		echo "Failed to validate integrity of s3://${PUBLISH_S3_BUCKET}/ecs-agent-${GOOD_VERSION}.tar"
+		rm "${tarball}"
+		rm "${tarball_md5}"
+		exit 1
+	fi
+
+	docker load < "${tarball}"
+
+	echo "Tagging as ${IMAGE_NAME}:${IMAGE_TAG_LATEST}"
+	docker tag amazon/amazon-ecs-agent:latest "${IMAGE_NAME}:${IMAGE_TAG_LATEST}"
```
aaithal: Should we also have something here to remove all of `amazon/amazon-ecs-agent:*` images from the local file system?

scripts/repoint-latest
jhaynes: Potentially never and we should remove it. But we've discussed the possibility of wanting fully remove a bad agent that is in the wild if blast radius is know and small and/or the issue(s) with that release are very critical. However, since we don't have those details defined, maybe the right thing is to leave it out entirely for now.

scripts/repoint-latest
jhaynes: Agreed.

jhaynes: Probably should to be safe.

scripts/repoint-latest
jhaynes: We had discussed the possibility that an agent bug could be critical enough that we would want this functionality. However, until we have definitions for what constitutes a bug this big, maybe we should remove this functionality altogether.

scripts/repoint-latest
jhaynes: `s3_ls` exits non-zero if the object isn't present. I want to catch that and give some helpful info rather than just dying. `set +e` lets `s3_ls` exit non-zero without exiting the script.

scripts/repoint-latest
```diff
@@ -0,0 +1,163 @@
...
+		suffix="tar"
+	fi
+
+	tarball_bad="$(mktemp)"
+	tarball_latest="$(mktemp)"
+
+	s3_cp "s3://${PUBLISH_S3_BUCKET}/ecs-agent-${BAD_VERSION}.tar" "${tarball_bad}"
+	s3_cp "s3://${PUBLISH_S3_BUCKET}/ecs-agent-${IMAGE_TAG_LATEST}.tar.md5" "${tarball_latest}"
+
+	if ! check_md5 "${tarball_bad}" test_md5="$(md5sum ${tarball_latest} | sed 's/ .*//')" ; then
```
nmeyerhans: This usage of check_md5 doesn't look right. I think you just want to pass `${tarball_latest}` as the second parameter. The function is defined as:

```
 check_md5() {
 	test_md5="$(md5sum ${1} | sed 's/ .*//')"
        expected_md5="$(cat ${2})"
...
```

scripts/repoint-latest
```diff
@@ -0,0 +1,163 @@
...
+			TARGET_OS_LIST+=(${os})
+		else
+			echo "Unsupported target OS ${os}" >&2
+			exit 1
+		fi
+	done
+fi
+
+echo "Current bad version: ${BAD_VERSION}"
+echo "New \"latest\" version: ${GOOD_VERSION}"
```
nmeyerhans: We should at least do some cursory validation of `${BAD_VERSION}` and `${GOOD_VERSION}`. At least that they have nonzero length.

scripts/repoint-latest
jhaynes: Right you are. Fixed.

scripts/repoint-latest
```diff
@@ -0,0 +1,173 @@
...
+	s3_cp "s3://${PUBLISH_S3_BUCKET}/ecs-agent-${GOOD_VERSION}.tar.md5" "${tarball_md5}"
+
+	if ! check_md5 "${tarball}" "${tarball_md5}" ; then
+		echo "Failed to validate integrity of s3://${PUBLISH_S3_BUCKET}/ecs-agent-${GOOD_VERSION}.tar"
+		rm "${tarball}"
+		rm "${tarball_md5}"
+		exit 1
+	fi
+
+	docker images -q amazon/amazon-ecs-agent | xargs -L 1 docker rmi -f
```
aaithal: Is `xargs --no-run-if-empty` better here? Also an `echo "Removing amazon/amazon-ecs-agent images from the disk"` might be useful as well.

scripts/repoint-latest
aaithal: never mind. i misunderstood `set -e`

PR #929, opened 2017-08-16
ENI workflow bug fixes

aaithal:
### Summary

Various bug fixes related to ENI attachment for a Task

### Implementation details

* 01e1707 agent/api: fix failing test in appveyor by using common time format for ENIAttachment.ExpiresAt
* 91d7095 agent/eventhandler: fix bug where eni attachment would not be submitted
* f8206d6 agent/api,agent/acs/handler: fix bug where eni attachment state could not be saved

### Testing

None

### Licensing

### DIFF
--- a/agent/acs/handler/attach_eni_handler.go
+++ b/agent/acs/handler/attach_eni_handler.go
@@ -102,13 +102,18 @@ func (handler *attachENIHandler) handleSingleMessage(message *ecsacs.AttachTaskN
 		return errors.Wrapf(err,
 			"attach eni message handler: error validating AttachTaskNetworkInterface message received from ECS")
 	}
-	if err := handler.acsClient.MakeRequest(&ecsacs.AckRequest{
-		Cluster:           message.ClusterArn,
-		ContainerInstance: message.ContainerInstanceArn,
-		MessageId:         message.MessageId,
-	}); err != nil {
-		seelog.Warnf("Failed to ack request with messageId: %s, error: %v", aws.StringValue(message.MessageId), err)
-	}
+
+	// Send ACK
+	go func(clusterArn *string, containerInstanceArn *string, messageID *string) {
+		if err := handler.acsClient.MakeRequest(&ecsacs.AckRequest{
+			Cluster:           clusterArn,
+			ContainerInstance: containerInstanceArn,
+			MessageId:         messageID,
+		}); err != nil {
+			seelog.Warnf("Failed to ack request with messageId: %s, error: %v", aws.StringValue(messageID), err)
+		}
+	}(message.ClusterArn, message.ContainerInstanceArn, message.MessageId)
+
 	// Check if this is a duplicate message
 	mac := aws.StringValue(message.ElasticNetworkInterfaces[0].MacAddress)
 	if eniAttachment, ok := handler.state.ENIByMac(mac); ok {
--- a/agent/acs/handler/attach_eni_handler.go
+++ b/agent/acs/handler/attach_eni_handler.go
@@ -100,7 +100,7 @@ func (handler *attachENIHandler) handleSingleMessage(message *ecsacs.AttachTaskN
 	// Validate fields in the message
 	if err := validateAttachTaskNetworkInterfacesMessage(message); err != nil {
 		return errors.Wrapf(err,
-			"attach eni message handler: error validating AttachTaskNetworkInterfac message received from ECS")
+			"attach eni message handler: error validating AttachTaskNetworkInterface message received from ECS")
 	}
 	if err := handler.acsClient.MakeRequest(&ecsacs.AckRequest{
 		Cluster:           message.ClusterArn,
--- a/agent/acs/handler/attach_eni_handler.go
+++ b/agent/acs/handler/attach_eni_handler.go
@@ -85,7 +85,9 @@ func (attachENIHandler *attachENIHandler) handleMessages() {
 	for {
 		select {
 		case message := <-attachENIHandler.messageBuffer:
-			attachENIHandler.handleSingleMessage(message)
+			if err := attachENIHandler.handleSingleMessage(message); err != nil {
+				seelog.Warnf("Unable to handle ENI Attachment message [%s]: %v", message.String(), err)
+			}
 		case <-attachENIHandler.ctx.Done():
 			return
 		}
@@ -93,56 +95,57 @@ func (attachENIHandler *attachENIHandler) handleMessages() {
 }
 
 // handleSingleMessage acks the message received
-// TODO: Send response to validate ENI attachment
-func (attachENIHandler *attachENIHandler) handleSingleMessage(message *ecsacs.AttachTaskNetworkInterfacesMessage) error {
+func (handler *attachENIHandler) handleSingleMessage(message *ecsacs.AttachTaskNetworkInterfacesMessage) error {
+	receivedAt := time.Now()
 	// Validate fields in the message
-	err := validateAttachTaskNetworkInterfacesMessage(message)
-	if err != nil {
-		return errors.Wrapf(err, "attach eni message handler: error validating AttachTaskNetworkInterfac message received from ECS")
+	if err := validateAttachTaskNetworkInterfacesMessage(message); err != nil {
+		return errors.Wrapf(err,
+			"attach eni message handler: error validating AttachTaskNetworkInterfac message received from ECS")
 	}
-
-	err = attachENIHandler.acsClient.MakeRequest(&ecsacs.AckRequest{
+	if err := handler.acsClient.MakeRequest(&ecsacs.AckRequest{
 		Cluster:           message.ClusterArn,
 		ContainerInstance: message.ContainerInstanceArn,
 		MessageId:         message.MessageId,
-	})
-	if err != nil {
+	}); err != nil {
 		seelog.Warnf("Failed to ack request with messageId: %s, error: %v", aws.StringValue(message.MessageId), err)
 	}
-
 	// Check if this is a duplicate message
-	eniMac := aws.StringValue(message.ElasticNetworkInterfaces[0].MacAddress)
-	if _, ok := attachENIHandler.state.ENIByMac(eniMac); ok {
-		seelog.Infof("Duplicate ENI attachment message for ENI with MAC address: %s", eniMac)
-		return nil
+	mac := aws.StringValue(message.ElasticNetworkInterfaces[0].MacAddress)
+	if eniAttachment, ok := handler.state.ENIByMac(mac); ok {
+		seelog.Infof("Duplicate ENI attachment message for ENI with MAC address: %s", mac)
+		eniAckTimeoutHandler := ackTimeoutHandler{mac: mac, state: handler.state}
+		return eniAttachment.StartTimer(eniAckTimeoutHandler.handle)
 	}
-
-	attachENIHandler.addENIAttachmentToState(message)
-	err = attachENIHandler.saver.Save()
-	if err != nil {
-		return errors.Wrapf(err, "attach eni message handler: error save agent state")
+	if err := handler.addENIAttachmentToState(message, receivedAt); err != nil {
+		return errors.Wrapf(err, "attach eni message handler: unable to add eni attachment to engine state")
+	}
+	if err := handler.saver.Save(); err != nil {
+		return errors.Wrapf(err, "attach eni message handler: unable to save agent state")
 	}
 	return nil
 }
 
 // addENIAttachmentToState adds the eni info to the state
-func (handler *attachENIHandler) addENIAttachmentToState(message *ecsacs.AttachTaskNetworkInterfacesMessage) {
-	attachmentArn := aws.StringValue(message.ElasticNetworkInterfaces[0].AttachmentArn)
-	eniMac := aws.StringValue(message.ElasticNetworkInterfaces[0].MacAddress)
-	// Stop tracking the eni attachment after timeout
-	eniAckTimeout := time.Duration(aws.Int64Value(message.WaitTimeoutMs)) * time.Millisecond
-	eniAckTimeoutHandler := ackTimeoutHandler{mac: eniMac, state: handler.state}
-	ackTimeoutTimer := time.AfterFunc(eniAckTimeout, eniAckTimeoutHandler.handle)
-
-	seelog.Info("Adding eni info to state, eni: %s", attachmentArn)
+func (handler *attachENIHandler) addENIAttachmentToState(message *ecsacs.AttachTaskNetworkInterfacesMessage, receivedAt time.Time) error {
+	attachmentARN := aws.StringValue(message.ElasticNetworkInterfaces[0].AttachmentArn)
+	mac := aws.StringValue(message.ElasticNetworkInterfaces[0].MacAddress)
+	taskARN := aws.StringValue(message.TaskArn)
 	eniAttachment := &api.ENIAttachment{
-		TaskArn:          aws.StringValue(message.TaskArn),
-		AttachmentArn:    attachmentArn,
+		TaskARN:          taskARN,
+		AttachmentARN:    attachmentARN,
 		AttachStatusSent: false,
-		MacAddress:       eniMac,
-		AckTimer:         ackTimeoutTimer,
+		MACAddress:       mac,
+		// Stop tracking the eni attachment after timeout
+		ExpiresAt: receivedAt.Add(time.Duration(aws.Int64Value(message.WaitTimeoutMs)) * time.Millisecond),
 	}
+	eniAckTimeoutHandler := ackTimeoutHandler{mac: mac, state: handler.state}
+	if err := eniAttachment.StartTimer(eniAckTimeoutHandler.handle); err != nil {
+		return err
+	}
+	seelog.Info("Adding eni info for task '%s' to state, attachment=%s mac=%s",
+		taskARN, attachmentARN, mac)
 	handler.state.AddENIAttachment(eniAttachment)
+	return nil
 }
 
 // ackTimeoutHandler remove ENI attachment from agent state after the ENI ack timeout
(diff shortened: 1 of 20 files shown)


agent/api/eniattachment.go
```diff
@@ -0,0 +1,94 @@
...
+	// TaskARN is the task identifier from ecs
+	TaskARN string `json:"taskArn"`
+	// AttachmentARN is the identifier for the eni attachment
+	AttachmentARN string `json:"attachmentArn"`
+	// AttachStatusSent indicates whether the attached status has been sent to backend
+	AttachStatusSent bool `json:"attachSent"`
+	// MACAddress is the mac address of eni
+	MACAddress string `json:"macAddress"`
+	// Status is the status of the eni: none/attached/detached
+	Status         ENIAttachmentStatus `json:"status"`
```
vsiddharth: Is this spacing intentional ?

agent/api/eniattachment.go
```diff
@@ -0,0 +1,94 @@
...
+	eni.ackTimer.Stop()
+}
+
+// String returns a string representation of the ENI Attachment
+func (eni *ENIAttachment) String() string {
+	eni.sentStatusLock.RLock()
+	defer eni.sentStatusLock.RUnlock()
+
+	return fmt.Sprintf(
+		"ENI Attachment; task=%s;attachment=%s;attachmentSent=%t;mac=%s;status=%s;expiresAt=%s",
```
vsiddharth: This is awfully dense. 
Can we do away with lesser information here ?

agent/eventhandler/handler_test.go
```diff
@@ -271,27 +271,30 @@ func TestENISentStatusChange(t *testing.T) {
...
+	assert.NoError(t, eniAttachment.StartTimer(timeoutFunc))
 
 	sendableTaskEvent := newSendableTaskEvent(api.TaskStateChange{
-		Attachments: eniAttachment,
-		TaskArn:     "taskarn",
-		Status:      api.TaskRunning,
-		Task:        task,
+		Attachment: eniAttachment,
+		TaskArn:    "taskarn",
+		// TODO: This should be api.TaskStatusNone
```
vsiddharth: Remove this line

agent/acs/handler/attach_eni_handler_test.go
```diff
@@ -28,9 +28,9 @@ import (
 )
 
 const (
-	eniMessageId = "123"
-	randomMAC    = "00:0a:95:9d:68:16"
-	waitTimeout  = 10
+	eniMessageId      = "123"
+	randomMAC         = "00:0a:95:9d:68:16"
+	waitTimeoutMillis = 10
```
aaithal: why is it strange?

agent/api/eniattachment.go
aaithal: It's whatever `go fmt` does.

agent/api/eniattachment.go
aaithal: What do you think is not needed here?

agent/acs/handler/attach_eni_handler_test.go
vsiddharth: waitTimeoutMillis doesn't read like an unit of time

agent/api/eniattachment.go
vsiddharth: Alright, just looked strange to me

agent/api/eniattachment.go
vsiddharth: Can we get rid of attachmentSent,  attachmentArn and expiresAt here ?

agent/acs/handler/attach_eni_handler_test.go
aaithal: I still don't get it. It's not a unit of time. It only gets converted to a unit of time when multiplied with `time.Millisecond`

agent/api/eniattachment.go
aaithal: All of these are important information that would help with debugging if logged.

vsiddharth: It just appears to be a lot of information.
We can keep them all if they are essential.

agent/acs/handler/attach_eni_handler_test.go
samuelkarp: Millis definitely reads like a unit of time to me.

agent/api/eniattachment.go
aaithal: To quote myself from the previous [comment](https://github.com/aws/amazon-ecs-agent/pull/929#discussion_r133577062):

> All of these are important information that would help with debugging if logged.

agent/acs/handler/attach_eni_handler_test.go
vsiddharth: I understand that but this name is just confusing me. 
I would prefer `waitTimeout` or `waitTimeoutMs`.

samuelkarp: If it were `waitTimeout`, it would need to be equal to `10 * time.Millisecond` rather than its current value.  I don't see any reason to change the name.

aaithal: It was bugging me that `waitTimeout` was set to `10` without any unit information. `GoRename` seemed simpler than changing all lines where this was referred to. Honestly, I don't think the name of this variable is that important in the larger context of this PR. I'm sticking with what I have now unless some one feels way too strongly about it.

agent/acs/handler/attach_eni_handler.go
```diff
@@ -85,64 +85,67 @@ func (attachENIHandler *attachENIHandler) handleMessages() {
...
 	// Check if this is a duplicate message
-	eniMac := aws.StringValue(message.ElasticNetworkInterfaces[0].MacAddress)
-	if _, ok := attachENIHandler.state.ENIByMac(eniMac); ok {
-		seelog.Infof("Duplicate ENI attachment message for ENI with MAC address: %s", eniMac)
-		return nil
+	mac := aws.StringValue(message.ElasticNetworkInterfaces[0].MacAddress)
+	if eniAttachment, ok := handler.state.ENIByMac(mac); ok {
+		seelog.Infof("Duplicate ENI attachment message for ENI with MAC address: %s", mac)
+		eniAckTimeoutHandler := ackTimeoutHandler{mac: mac, state: handler.state}
+		return eniAttachment.StartTimer(eniAckTimeoutHandler.handle)
```
richardpen: Do we need this, through it's handled inside the `StartTimer` where it checks for `nil`, do you have any concerns remove this?

agent/acs/handler/attach_eni_handler.go
```diff
@@ -154,11 +157,11 @@ type ackTimeoutHandler struct {
 func (handler *ackTimeoutHandler) handle() {
 	eniAttachment, ok := handler.state.ENIByMac(handler.mac)
 	if !ok {
-		seelog.Warnf("Timed out waiting for ENI ack; ignoring unmanaged ENI attachment with MAC address:%s", handler.mac)
+		seelog.Warnf("Timed out waiting for ENI ack; ignoring unmanaged ENI attachment with MAC address: %s", handler.mac)
 		return
 	}
 	if !eniAttachment.IsSent() {
```
richardpen: minor: Would it better `ISSent()`?

agent/acs/handler/attach_eni_handler.go
```diff
@@ -154,11 +157,11 @@ type ackTimeoutHandler struct {
...
-		seelog.Info("Timed out waiting for ENI ack; removing ENI attachment record with MAC address: %s", handler.mac)
+		seelog.Infof("Timed out waiting for ENI ack; removing ENI attachment record with MAC address: %s", handler.mac)
 		handler.state.RemoveENIAttachment(handler.mac)
```
richardpen: I can see there may be race condition here in edge case where a eni attachment event was sent to eventhandler but before submitstatechange return, at this time the timer was triggered. This should be handled in another PR, just make sure we don't forget.

agent/api/eniattachment.go
```diff
@@ -0,0 +1,94 @@
...
+	TaskARN string `json:"taskArn"`
+	// AttachmentARN is the identifier for the eni attachment
+	AttachmentARN string `json:"attachmentArn"`
+	// AttachStatusSent indicates whether the attached status has been sent to backend
+	AttachStatusSent bool `json:"attachSent"`
+	// MACAddress is the mac address of eni
+	MACAddress string `json:"macAddress"`
+	// Status is the status of the eni: none/attached/detached
+	Status         ENIAttachmentStatus `json:"status"`
+	sentStatusLock sync.RWMutex
```
richardpen: I'd suggest we have a general lock for all the fields of this struct.

agent/api/eniattachment.go
```diff
@@ -0,0 +1,94 @@
...
+		// The timer has already been initialized, do nothing
+		return nil
+	}
+	now := time.Now()
+	duration := eni.ExpiresAt.Sub(now)
+	if duration <= 0 {
+		return errors.Errorf("eni attachment: timer expiration is in the past; expiration [%s] < now [%s]",
+			eni.ExpiresAt.String(), now.String())
+	}
+	seelog.Infof("Starting ENI ack timer with duration=%s, %s", duration.String(), eni.String())
```
richardpen: Shouldn't this be `debug` level?

agent/api/eniattachment.go
```diff
@@ -0,0 +1,94 @@
...
+	// Status is the status of the eni: none/attached/detached
+	Status         ENIAttachmentStatus `json:"status"`
+	sentStatusLock sync.RWMutex
+	// ExpiresAt is the timestamp past which the ENI Attachment is considered
+	// unsuccessful. The SubmitTaskStateChange API, with the attachment information
+	// should be invoked before this timestamp.
+	ExpiresAt time.Time `json:"expiresAt"`
+	// ackTimer is used to register the expirtation timeout callback for unsuccessful
+	// ENI attachments
+	ackTimer ttime.Timer
```
richardpen: Is `json:"-"` needed here?

agent/api/eniattachment.go
richardpen: I was wondering if breaking that into multiple lines will make it more readable?

agent/acs/handler/attach_eni_handler.go
aaithal: Are you asking if it's ok to get rid of this line? We need to make sure that the expiration timer starts if the Agent was restarted, which is why I added this line.

agent/acs/handler/attach_eni_handler.go
aaithal: Can you elaborate that comment?

agent/acs/handler/attach_eni_handler.go
aaithal: If the eventhandler was unable to submit the state change for the attachment within the timeout, there's no need maintaining the attachment state locally anyway, which I think is fine.

agent/api/eniattachment.go
aaithal: Yeah, i'll make that change

agent/api/eniattachment.go
aaithal: No, it's not globally scoped.

agent/api/eniattachment.go
aaithal: I'd prefer it be `Infof` as there shouldn't be too many of these lines and at least for now, it seems important for the ENI workflow.

agent/api/eniattachment.go
aaithal: hmm, alright. I can make that change. FWIW, here are some sample log lines, which I don't think is too verbose.

```
2017-08-17T15:30:07Z [INFO] Starting ENI ack timer with duration=59.999783182s, ENI Attachment; task=arn:aws:ecs:us-east-1:xxxx:task/xxxxx;attachment=arn:aws:ecs:us-east-1:xxxx:attachment/xxxx;attachmentSent=false;mac=xx:xx:xx:70:f4:24;status=NONE;expiresAt=2017-08-17 15:31:07.161392737 +0000 UTC
2017-08-17T15:30:11Z [INFO] Emitting ENI change event for: ENI Attachment; task=arn:aws:ecs:us-east-1:xxx:task/xxx;attachment=arn:aws:ecs:us-east-1:xxx:attachment/xxx;attachmentSent=false;mac=xx:xx:xx:70:f4:24;status=ATTACHED;expiresAt=2017-08-17 15:31:07.161392737 +0000 UTC
```

agent/api/ecsclient/client.go
```diff
@@ -288,13 +288,13 @@ func (client *APIECSClient) getCustomAttributes() []*ecs.Attribute {
 
 func (client *APIECSClient) SubmitTaskStateChange(change api.TaskStateChange) error {
 	// Submit attachment state change
-	if change.Attachments != nil {
+	if change.Attachment != nil {
```
adnxn: something that wasnt clear to me and probably included in a different PR - where are these eni attachment events being emitted from? is it only out of the `eni/watcher` package?

aaithal: yeah, it's coming from https://github.com/aws/amazon-ecs-agent/blob/enis/agent/eni/watcher/watcher_linux.go#L141

agent/api/ecsclient/client.go
```diff
@@ -288,13 +288,13 @@ func (client *APIECSClient) getCustomAttributes() []*ecs.Attribute {
...
 		var attachments []*ecs.AttachmentStateChange
 
-		eniStatus := change.Attachments.Status.String()
+		eniStatus := change.Attachment.Status.String()
 		attachments = []*ecs.AttachmentStateChange{
 			{
-				AttachmentArn: &change.Attachments.AttachmentArn,
+				AttachmentArn: &change.Attachment.AttachmentARN,
```
aaithal: We should. I'll modify that.

agent/api/eniattachment.go
aaithal: Not `ExpiresAt`. I was referring to `ackTimer`. I meant to say `unexported` by that. From https://golang.org/pkg/encoding/json/#Marshal: 

> Struct values encode as JSON objects. Each exported struct field becomes a member of the object, using the field name as the object key, unless the field is omitted for one of the reasons given below.

agent/acs/handler/attach_eni_handler_test.go
```diff
@@ -253,21 +255,75 @@ func TestENIAckSingleMessage(t *testing.T) {
...
+		ContainerInstanceArn: aws.String(containerInstanceArn),
+		ElasticNetworkInterfaces: []*ecsacs.ElasticNetworkInterface{
+			&mockNetInterface1,
+		},
+		TaskArn:       aws.String(taskArn),
+		WaitTimeoutMs: aws.Int64(waitTimeoutMillis),
+	}
+
+	// Expect an error starting the timer because of <=0 duration
+	err := eniAttachHandler.handleSingleMessage(message)
```
richardpen: if you are manually call this, do you still need line 283 `go eniAttachHandler.start()`, and then you can get rid of the `context`?

agent/acs/handler/attach_eni_handler_test.go
```diff
@@ -253,21 +255,75 @@ func TestENIAckSingleMessage(t *testing.T) {
...
+		ClusterArn:           aws.String(clusterName),
+		ContainerInstanceArn: aws.String(containerInstanceArn),
+		ElasticNetworkInterfaces: []*ecsacs.ElasticNetworkInterface{
+			&mockNetInterface1,
+		},
+		TaskArn:       aws.String(taskArn),
+		WaitTimeoutMs: aws.Int64(waitTimeoutMillis),
+	}
+
+	// Expect an error starting the timer because of <=0 duration
```
richardpen: To make this comments easier to understand, can you add a negative value of WaitTimeoutMs in the returned ENIAttachment in line 280?

agent/acs/handler/attach_eni_handler.go
richardpen: Just the naming, never mind, I don't have a strong feeling about this.

agent/acs/handler/attach_eni_handler_test.go
aaithal: done.

agent/acs/handler/attach_eni_handler_test.go
aaithal: @richardpen  I have fixed this and other comments. Can you please take another look?

agent/acs/handler/attach_eni_handler.go
```diff
@@ -85,64 +85,67 @@ func (attachENIHandler *attachENIHandler) handleMessages() {
...
-	err := validateAttachTaskNetworkInterfacesMessage(message)
-	if err != nil {
-		return errors.Wrapf(err, "attach eni message handler: error validating AttachTaskNetworkInterfac message received from ECS")
+	if err := validateAttachTaskNetworkInterfacesMessage(message); err != nil {
+		return errors.Wrapf(err,
+			"attach eni message handler: error validating AttachTaskNetworkInterface message received from ECS")
 	}
-
-	err = attachENIHandler.acsClient.MakeRequest(&ecsacs.AckRequest{
+	if err := handler.acsClient.MakeRequest(&ecsacs.AckRequest{
```
aaithal: May be. We have that behavior with rest of the ack sending logic. Can you think of reasons to make this asynchronous?

richardpen: One concern is that if the make request takes long time, it will block the rest of the process, like start the timer, save the state.

agent/api/ecsclient/client.go
```diff
@@ -288,20 +288,20 @@ func (client *APIECSClient) getCustomAttributes() []*ecs.Attribute {
 
 func (client *APIECSClient) SubmitTaskStateChange(change api.TaskStateChange) error {
 	// Submit attachment state change
-	if change.Attachments != nil {
+	if change.Attachment != nil {
```
samuelkarp: Doesn't need to be this particular PR right now, but I'd prefer if the attachment logic were factored out into a separate function; it took me a bit to see that this block has an early return and doesn't invoke the other state-updating logic.

```go
if change.Attachment != nil {
	return client.submitAttachmentTaskStateChange(change)
}
```

aaithal: We can pick it up in the next PR

PR #934, opened 2017-08-21
Ignore destroy event of container to get rid of "no such container" error in docker log

richardpen:
### Summary

This PR change the task engine to ignore docker remove events, thus there won't be additional calls to docker for the removed container which will result "no such container" in docker logs.

### Implementation details

When received destroy event from docker, just ignore it.

### Testing

New tests cover the changes: 

* Bug - Fix the issue that causes "no such container" error in docker.

### Licensing

yes

### DIFF
--- a/agent/engine/docker_container_engine.go
+++ b/agent/engine/docker_container_engine.go
@@ -708,11 +708,14 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 				continue
 			case "restart":
 			case "resize":
-			case "destroy":
 			case "unpause":
+			case "destroy":
+				// container remove events doesn't need to be handled by ecs agent
+				// as it has already moved to stopped before this event
+				continue
+
 			// These result in us falling through to inspect the container, some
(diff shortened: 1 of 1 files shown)


agent/engine/docker_container_engine.go
```diff
@@ -708,10 +708,13 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 				continue
 			case "restart":
 			case "resize":
-			case "destroy":
 			case "unpause":
 			// These result in us falling through to inspect the container, some
```
samuelkarp: This comment is now misplaced.  It should move below `case "destroy"` since it doesn't apply to that case.

richardpen: thanks, updated!

richardpen:
@aaithal Confirmed with all the docker versions listed [here](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html#ecs-optimized-ami-agent-versions).

PR #937, opened 2017-08-23
Adding task.GetID

vsiddharth:
### Summary

Adding a method to retrieve taskID from taskARN

### Testing

### Licensing


agent/api/task.go
```diff
@@ -674,3 +678,18 @@ func (t *Task) String() string {
...
 }
+
+func (task *Task) GetID() (string, error) {
+	parsedARN, err := arn.Parse(task.Arn)
+	if err != nil {
+		return "", errors.Wrapf(err, "task get-id: unable to parse taskARN")
+	}
+
+	resource := parsedARN.Resource
+	resourceSplit := strings.SplitN(resource, arnResourceDelimiter, arnResourceSections)
```
sharanyad: Isn't Split enough here instead of SplitN, since taskARN has only single '/'? In that case, `arnResourceSections` may not be necessary.

samuelkarp: SplitN is safer here, because you'll end up with a  known number of sections (or less, if there are not enough delimiters); it leaves any extra unparsed.  Here, it will simplify error handling because you only need to check whether there are enough delimited sections rather than also having to handle too many.

vsiddharth: True but I'm just being extra cautious here.

agent/api/task.go
```diff
@@ -674,3 +678,27 @@ func (t *Task) String() string {
 	}
 	return res + "]"
 }
+
+// GetID is used to retrieve the taskID from taskARN
+// Reference: http://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arn-syntax-ecs
+func (task *Task) GetID() (string, error) {
```
petderek: Neat function! The error messages could offer more hints about what the underlying problem is. Maybe explicitly say that the arn associated with the task is malformed.

ex: `resource missing delimiter` isn't going to be clear in a logfile without the context of the surrounding functions.

vsiddharth: @petderek I've updated the error messages with more context. 
I want to use `errors.New` but I've combined it as `errors.New(fmt.Sprintf(...))`

Should we fallback to `fmt.Errorf` instead ?

PR #938, opened 2017-08-23
resources/cgroup: Adding 'Exists' to cgroup interface

vsiddharth:
### Summary

Adding a new method to verify the existence of a cgroup at the given
root. This method accepts a cgroup root and returns a boolean.

Signed-off-by: Vinothkumar Siddharth <sidvin@amazon.com

### Testing

### Licensing

### DIFF
--- a/agent/resources/cgroup/cgroup_controller_linux.go
+++ b/agent/resources/cgroup/cgroup_controller_linux.go
@@ -50,30 +50,42 @@ func (c *control) Create(cgroupSpec *Spec) (cgroups.Cgroup, error) {
 
 	// Create cgroup
 	seelog.Infof("Creating cgroup %s", cgroupSpec.Root)
-	control, err := c.New(cgroups.V1, cgroups.StaticPath(cgroupSpec.Root), cgroupSpec.Specs)
+	controller, err := c.New(cgroups.V1, cgroups.StaticPath(cgroupSpec.Root), cgroupSpec.Specs)
 
 	if err != nil {
 		return nil, errors.Wrapf(err, "cgroup create: unable to create controller")
 	}
 
-	return control, nil
+	return controller, nil
 }
 
 // Remove is used to delete the cgroup
 func (c *control) Remove(cgroupPath string) error {
 	seelog.Debugf("Removing cgroup %s", cgroupPath)
 
-	control, err := c.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
+	controller, err := c.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
 	if err != nil {
 		return errors.Wrapf(err, "cgroup remove: unable to obtain controller")
 	}
 
 	// Delete cgroup
-	err = control.Delete()
+	err = controller.Delete()
 	if err != nil {
 		return errors.Wrapf(err, "cgroup remove: unable to delete cgroup")
 	}
 	return nil
+}
+
+// Exists is used to verify the existence of a cgroup
+func (c *control) Exists(cgroupPath string) bool {
+	seelog.Debugf("Checking existence of cgroup: %s", cgroupPath)
+
+	controller, err := c.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
+	if err != nil || controller == nil {
+		return false
+	}
+
+	return true
 }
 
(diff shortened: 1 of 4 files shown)


agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -50,32 +50,44 @@ func (c *control) Create(cgroupSpec *Spec) (cgroups.Cgroup, error) {
...
 
-	return control, nil
+	return controller, nil
 }
 
 // Remove is used to delete the cgroup
 func (c *control) Remove(cgroupPath string) error {
 	seelog.Debugf("Removing cgroup %s", cgroupPath)
 
-	control, err := c.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
```
petderek: nit:
can you change this to 'cgroup' to further distinguish it from 'control'? The return type for Load is a `cgroups.Cgroup`.

agent/resources/cgroup/cgroup_controller_linux_test.go
```diff
@@ -140,3 +140,43 @@ func TestRemoveErrorCase(t *testing.T) {
...
+	assert.False(t, control.Exists(testCgroupRoot))
+}
+
+func TestExistsErrorPathWithLoadError(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockCgroupFactory := mock_factory.NewMockCgroupFactory(ctrl)
+
+	mockCgroupFactory.EXPECT().Load(gomock.Any(), gomock.Any()).Return(nil, nil)
```
petderek: When does this path happen?

agent/resources/cgroup/cgroup_controller_linux.go
```diff
@@ -50,32 +50,44 @@ func (c *control) Create(cgroupSpec *Spec) (cgroups.Cgroup, error) {
...
 	}
 	return nil
 }
 
+// Exists is used to verify the existence of a cgroup
+func (c *control) Exists(cgroupPath string) bool {
+	seelog.Debugf("Checking existence of cgroup: %s", cgroupPath)
+
+	controller, err := c.Load(cgroups.V1, cgroups.StaticPath(cgroupPath))
+	if err != nil || controller == nil {
```
sharanyad: When would controller be nil?

agent/resources/cgroup/cgroup_controller_linux_test.go
vsiddharth: This should never ever happen but I just wanted to be extra cautious here

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: This should never ever happen but I just wanted to be extra cautious here

agent/resources/cgroup/cgroup_controller_linux.go
vsiddharth: I wanted to avoid any naming `confusion|conflicts`.
`cgroup` is also the name of the package and we also have a `containerd/cgroups` package. 

Should we shrink it down to something like `ctl` or `cgrp` or `cg` ?

petderek: `group`?

vsiddharth: `group` totally takes away the `control` aspect but I'm open to other suggestions if `controller` isn't clear.

agent/resources/cgroup/cgroup_controller_linux.go
sharanyad: Question for my understanding: Is it possible that the cgroup path exists, but for some reason, it wouldn't load it? In that case, will load return an error and this method could return false?

vsiddharth: The load call here would fail iff the cgroup is deleted prior to the load. 
The situation you anticipate shouldn't ever happen.

agent/resources/cgroup/cgroup_controller_linux.go
petderek: Looks like we've already consumed all of the appropriate names then? I don't have a strong preference if the good names are all taken :)

PR #939, opened 2017-08-23
Fix the issue where agent could create the same container twice 

richardpen:
### Summary

Fix the issue if agent was restarted before the docker create API returned, the agent could create duplicate containers for the task.

### Implementation details

Originally the agent stores the generated container name in memory before actually calling CreateContainer API, see [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine.go#L596). And this information was only stored in the taskToID struct, due to lacking of docker ID [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/dockerstate/docker_task_engine_state.go#L236). But the `taskToID` struct wasn't stored in the [state file](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/dockerstate/json.go#L26) by `statemanager`, so when the agent restart it will lose this information and then it will try to create the container with [another name](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine.go#L590) again.

This PR make sure the the generated container name can be restored from agent state file if agent restarted before the `CreateContainer` API returned.  Thus ensure the container won't be created twice. Also the agent would detect the container if it was created during the period of restarting, see [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine.go#L232).

### Testing

- [x] Manually  verified the state when agent was restarted at the create.

New tests cover the changes: 
yes

Bug - Fix an issue where a container could be created twice.
### Licensing

yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## UNRELEASED
 * Feature - Support for provisioning Tasks with ENIs
 * Bug - Fixed a memory leak issue when submitting the task state change [#967](https://github.com/aws/amazon-ecs-agent/pull/967)
+* Bug - Fix a race condition where a container can be created twice when agent restarts. [#939](https://github.com/aws/amazon-ecs-agent/pull/939)
 
 ## 1.14.4
 * Enhancement - Batch container state change events. [#867](https://github.com/aws/amazon-ecs-agent/pull/867)
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -252,6 +252,7 @@ func (engine *DockerTaskEngine) synchronizeState() {
 					log.Warn("Could not find matching container for expected", "name", cont.DockerName)
 				} else {
 					cont.DockerID = describedCont.ID
+					cont.Container.SetKnownStatus(dockerStateToState(describedCont.State))
 					// update mappings that need dockerid
 					engine.state.AddContainer(cont, task)
 					engine.imageManager.RecordContainerReference(cont.Container)
@@ -577,15 +578,24 @@ func (engine *DockerTaskEngine) createContainer(task *api.Task, container *api.C
 		client = client.WithVersion(dockerclient.DockerVersion(*container.DockerConfig.Version))
 	}
 
+	dockerContainerName := ""
+	containerMap, ok := engine.state.ContainerMapByArn(task.Arn)
+	if !ok {
+		containerMap = make(map[string]*api.DockerContainer)
+	} else {
+		// looking for container that has docker name but not created
+		for _, v := range containerMap {
+			if v.Container.Name == container.Name {
+				dockerContainerName = v.DockerName
+				break
+			}
+		}
+	}
+
 	// Resolve HostConfig
 	// we have to do this in create, not start, because docker no longer handles
 	// merging create config with start hostconfig the same; e.g. memory limits
 	// get lost
-	containerMap, ok := engine.state.ContainerMapByArn(task.Arn)
-	if !ok {
-		containerMap = make(map[string]*api.DockerContainer)
-	}
-
 	hostConfig, hcerr := task.DockerHostConfig(container, containerMap)
 	if hcerr != nil {
 		return DockerContainerMetadata{Error: api.NamedError(hcerr)}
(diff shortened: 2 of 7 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -236,6 +236,7 @@ func (engine *DockerTaskEngine) synchronizeState() {
 					log.Warn("Could not find matching container for expected", "name", cont.DockerName)
 				} else {
 					cont.DockerID = describedCont.ID
+					cont.Container.SetKnownStatus(api.ContainerCreated)
```
samuelkarp: I think we need to look here and set the state based on what comes back.  It could be created, running, or stopped.

agent/engine/docker_task_engine.go
```diff
@@ -562,9 +563,17 @@ func (engine *DockerTaskEngine) createContainer(task *api.Task, container *api.C
 	// we have to do this in create, not start, because docker no longer handles
```
samuelkarp: Can you move this comment about HostConfig above line 579 instead of here?  It's less clear what the comment is applying to now.

agent/engine/docker_task_engine.go
```diff
@@ -562,9 +563,17 @@ func (engine *DockerTaskEngine) createContainer(task *api.Task, container *api.C
...
+	dockerContainerName := ""
 	containerMap, ok := engine.state.ContainerMapByArn(task.Arn)
 	if !ok {
 		containerMap = make(map[string]*api.DockerContainer)
+	} else {
+		// looking for container that has docker name but not created
+		for _, v := range containerMap {
+			if v.Container.Name == container.Name {
+				dockerContainerName = v.DockerName
+			}
```
aaithal: you should/can add a `break` here as well.

agent/engine/docker_task_engine.go
richardpen: If it falls into this condition `dockerID==""`, the container status shouldn't be > created. Otherwise the `dockerID` field won't be empty. But have a check here will definitely add some confidence.

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -250,18 +250,25 @@ func (state *DockerTaskEngineState) AddContainer(container *api.DockerContainer,
 	}
 
 	if container.DockerID != "" {
+		// Update the container id to the state
```
aaithal: Is there a unit test covering the `if` and `else if` code paths?

jhaynes:
Can you update the CHANGELOG.md also as part of this PR?

agent/engine/dockerstate/docker_task_engine_state.go
richardpen: added.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+##
```
aaithal: Should we leave this as blank? We could consider adding a tentative version string with a `*` annotation to indicate that this is not yet released (Example: `1.14.5*` or `1.15.0*`). Or, we can add a placeholder release tag (Example: `UNRELEASED`). What do you think? 

cc @jhaynes

agent/engine/dockerstate/dockerstate_test.go
```diff
@@ -258,3 +258,44 @@ func TestRemoveNonExistingImageState(t *testing.T) {
 		t.Error("Error removing incorrect image state")
 	}
 }
+
+// TestAddContainer tests first add container with docker name and
+// then add the container with dockerID
+func TestAddContainer(t *testing.T) {
```
aaithal: Can you rename the test to be more descriptive? You can use what you have written in the comments as well. Example:`TestAddContainerNameAndID`.

CHANGELOG.md
nmeyerhans: I'd go with a placeholder for now, and we'll fill in the actual version as part of the release process. "UNRELEASED", "IN-PROGRESS", or "PENDING" would be my suggestions.

aaithal: I vote `"UNRELEASED"`

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## UNRELEASED
+* Bug - Fix a race condition where a container can be created twice when agent restart. [#939](https://github.com/aws/amazon-ecs-agent/pull/939)
```
aaithal: nit: restarts

agent/engine/docker_task_engine.go
```diff
@@ -579,28 +589,30 @@ func (engine *DockerTaskEngine) createContainer(task *api.Task, container *api.C
...
-	// name
-	engine.state.AddContainer(&api.DockerContainer{DockerName: containerName, Container: container}, task)
-	seelog.Infof("Created container name mapping for task %s - %s -> %s", task, container, containerName)
-	engine.saver.ForceSave()
+		// Pre-add the container in case we stop before the next, more useful,
+		// AddContainer call. This ensures we have a way to get the container if
+		// we die before 'createContainer' returns because we can inspect by
+		// name
+		engine.state.AddContainer(&api.DockerContainer{DockerName: dockerContainerName, Container: container}, task)
+		seelog.Infof("Created container name mapping for task %s - %s -> %s", task, container, dockerContainerName)
```
aaithal: Can you change `task` to `task.Arn`? We don't need the whole task to be logged here.

PR #944, opened 2017-08-24
Task cgroup limit models and config

petderek:
### Summary

Adds model and config for task limits. 

### Implementation details
* Limit values set at task level
* Capabilities code enhanced to check for docker api 1.22
* Config enhanced to allow users to disable features

### Testing

### Licensing

### DIFF
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -69,7 +69,11 @@ type Task struct {
 	Containers []*Container
 	// Volumes are the volumes for the task
 	Volumes []TaskVolume `json:"volumes"`
-
+	// VCPULimit is a task-level limit for compute resources. A value of 1 means that
+	// the task may access 100% of 1 vCPU on the instance
+	VCPULimit float64 `json:"CPULimit,omitempty"`
+	// MemoryLimit is a task-level limit for memory resources in bytes
+	MemoryLimit int64 `json:"Memory,omitempty"`
 	// DesiredStatusUnsafe represents the state where the task should go. Generally,
 	// the desired status is informed by the ECS backend as a result of either
 	// API calls made to ECS or decisions made by the ECS service scheduler.
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -288,6 +288,7 @@ func environmentConfig() (Config, error) {
 	appArmorCapable := utils.ParseBool(os.Getenv("ECS_APPARMOR_CAPABLE"), false)
 	taskIAMRoleEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE"), false)
 	taskIAMRoleEnabledForNetworkHost := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST"), false)
+	taskCPUMemLimitEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT"), true)
 
 	credentialsAuditLogFile := os.Getenv("ECS_AUDIT_LOGFILE")
 	credentialsAuditLogDisabled := utils.ParseBool(os.Getenv("ECS_AUDIT_LOGFILE_DISABLED"), false)
@@ -341,6 +342,7 @@ func environmentConfig() (Config, error) {
 		AppArmorCapable:                  appArmorCapable,
 		TaskCleanupWaitDuration:          taskCleanupWaitDuration,
 		TaskIAMRoleEnabled:               taskIAMRoleEnabled,
+		TaskCPUMemLimit:                  taskCPUMemLimitEnabled,
 		DockerStopTimeout:                dockerStopTimeout,
 		CredentialsAuditLogFile:          credentialsAuditLogFile,
 		CredentialsAuditLogDisabled:      credentialsAuditLogDisabled,
@@ -484,5 +486,5 @@ func (config *Config) validateAndOverrideBounds() error {
 // String returns a lossy string representation of the config suitable for human readable display.
 // Consequently, it *should not* return any sensitive information.
 func (config *Config) String() string {
-	return fmt.Sprintf("Cluster: %v, Region: %v, DataDir: %v, Checkpoint: %v, AuthType: %v, UpdatesEnabled: %v, DisableMetrics: %v, ReservedMem: %v, TaskCleanupWaitDuration: %v, DockerStopTimeout: %v", config.Cluster, config.AWSRegion, config.DataDir, config.Checkpoint, config.EngineAuthType, config.UpdatesEnabled, config.DisableMetrics, config.ReservedMemory, config.TaskCleanupWaitDuration, config.DockerStopTimeout)
+	return fmt.Sprintf("Cluster: %v, Region: %v, DataDir: %v, Checkpoint: %v, AuthType: %v, UpdatesEnabled: %v, DisableMetrics: %v, ReservedMem: %v, TaskCleanupWaitDuration: %v, DockerStopTimeout: %v, TaskCPUMemLimit: %v", config.Cluster, config.AWSRegion, config.DataDir, config.Checkpoint, config.EngineAuthType, config.UpdatesEnabled, config.DisableMetrics, config.ReservedMemory, config.TaskCleanupWaitDuration, config.DockerStopTimeout, config.TaskCPUMemLimit)
 }
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -16,7 +16,6 @@ package config
 import (
 	"errors"
 	"os"
-	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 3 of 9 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -830,6 +832,15 @@ func (engine *DockerTaskEngine) Capabilities() []string {
 		}
 	}
 
+	if engine.cfg.TaskResourceLimitEnabled {
+		if _, ok := supportedVersions[dockerclient.Version_1_22]; ok {
+			capabilities = append(capabilities, capabilityPrefix+capabilityTaskResourceLimit)
```
vsiddharth: We need to ensure that we register this capability only on linux.

agent/api/task.go
```diff
@@ -65,7 +65,11 @@ type Task struct {
 	Containers []*Container
 	// Volumes are the volumes for the task
 	Volumes []TaskVolume `json:"volumes"`
-
+	// CPULimit is a task-level limit for compute resources. A value of 1 means that
+	// the task may access 100% of 1 vCPU on the instance
+	CPULimit float64 `json:"CPULimit"`
```
vsiddharth: Should we have `omitempty` as these aren't mandatory

vsiddharth:
@petderek Could you please update the summary block's testing section ?

agent/engine/docker_task_engine.go
petderek: TaskResourceLimitEnabled will always be overridden to false on windows.

agent/api/task.go
petderek: Yep! I'll fix that.

petderek:
> ...testing...

I need to verify the integ tests on linux. I'll keep the summary block updated.

> What happens when old agent reads new state file ?

I haven't addressed that case yet. If `omitempty` is set it should be fine because I didn't change any of the required structures, but I can look at that. I'm also concerned about agent creating a hierarchy after it reads an old state file -- but I'm not addressing that second situation in this PR.

agent/api/task.go
```diff
@@ -65,7 +65,11 @@ type Task struct {
 	Containers []*Container
 	// Volumes are the volumes for the task
 	Volumes []TaskVolume `json:"volumes"`
-
+	// CPULimit is a task-level limit for compute resources. A value of 1 means that
```
samuelkarp: Should this be `VCPULimit`?

agent/config/config_test.go
```diff
@@ -124,41 +110,33 @@ func TestEnvironmentConfig(t *testing.T) {
 	assert.Equal(t, (90 * time.Second), conf.TaskCleanupWaitDuration)
 }
 
-func TestTrimWhitespace(t *testing.T) {
+func TestTrimWhitespaceWhenCreating(t *testing.T) {
+	os.Setenv("AWS_DEFAULT_REGION", "us-west-2")
```
samuelkarp: Why was this added?

agent/config/config_windows.go
```diff
@@ -84,4 +84,7 @@ func (config *Config) platformOverrides() {
 		}
 		config.ReservedPorts = append(config.ReservedPorts, httpPort)
 	}
+
+	// ensure TaskResourceLimit is disabled
+	config.TaskResourceLimitEnabled = false
```
samuelkarp: Can you add this to `DefaultConfig()` above too?

agent/engine/docker_task_engine.go
```diff
@@ -830,6 +832,15 @@ func (engine *DockerTaskEngine) Capabilities() []string {
 		}
 	}
 
+	if engine.cfg.TaskResourceLimitEnabled {
```
samuelkarp: Can you add a TODO here talking about ensuring compatibility with existing, already-running tasks (upgrade scenario)?  If there are already tasks running without a cgroup parent, we need to disable this feature.

petderek: I'm tracking that separately; I will also add a TODO

agent/api/task.go
petderek: I don't know. From agent's perspective its just applying a period and a quota. I'll change it if you think its more clear.

agent/config/config_test.go
petderek: Which one?

* `TestTrimWhitespace` was testing two different things, so I split it up.
* `os.Setenv` is there because the test was getting mad without the region explicitly set

agent/config/types.go
```diff
@@ -109,6 +109,9 @@ type Config struct {
 	// tasks with IAM Roles.
 	TaskIAMRoleEnabled bool
 
+	// TaskResourceLimitEnabled specifies if Agent can launch a task with a hierarchical cgroup
+	TaskResourceLimitEnabled bool
```
aaithal: This requires a change to `config.String()` as well.

agent/engine/docker_task_engine.go
```diff
@@ -45,6 +45,7 @@ const (
 	capabilityPrefix             = "com.amazonaws.ecs.capability."
 	capabilityTaskIAMRole        = "task-iam-role"
 	capabilityTaskIAMRoleNetHost = "task-iam-role-network-host"
+	capabilityTaskResourceLimit  = "task-resource-limit"
```
aaithal: Can you modify this to be `task-cpu-mem-limit`? `resource` has a pretty broad definition and seems like we're overloading it here. This applies to other names as well, let's make it very explicit by modifying `TaskResourceLimit` to be `TaskCPUMemLimit` everywhere.

agent/config/config_test.go
samuelkarp: Why does it require `os.Setenv` now?

agent/api/task.go
samuelkarp: I'd like it to match whatever we're going to have in the task definition.  If we don't know that yet, go ahead with either one and then we can fix it once we have an answer.

agent/config/config_test.go
petderek: I looked. The tests are depending on envs being set in other tests. I don't like that. I'll fix it.

agent/engine/docker_task_engine.go
vsiddharth: Given that memory is optional, should we split this up into two distinct capabilities ?

aaithal: @vsiddharth the capabilities do not reflect fields that are set in tasks. They are a way of letting agent register if it's capable of handling "certain things". In this particular scenario, it can handle both "cpu" and "memory" at the "task" level. Since we'll be releasing support for both of these with the same agent version, we don't need them to be 2 distinct capabilities.

vsiddharth: @aaithal thanks for the explanation.

agent/config/config_test.go
```diff
@@ -211,170 +188,146 @@ func TestInvalideValueDockerStopTimeout(t *testing.T) {
...
-	}
+	assert.NoError(t, err)
+	assert.Equal(t, cfg.NumImagesToDeletePerCycle, DefaultNumImagesToDeletePerCycle, "Wrong value for NumImagesToDeletePerCycle")
+}
 
-	if cfg.NumImagesToDeletePerCycle != DefaultNumImagesToDeletePerCycle {
-		t.Errorf("Wrong value for NumImagesToDeletePerCycle: %v", cfg.NumImagesToDeletePerCycle)
-	}
+func TestTaskResourceLimitsOverride(t *testing.T) {
+	defer os.Clearenv()
```
aaithal: I'm not a fan of all the `os.Clearenv` directives here. Can we reduce the scope of the cleanup to be just 
```
defer os.Unsetenv("SOME_SPECIFIC_THING_1")
defer os.Unsetenv("SOME_SPECIFIC_THING_2_IF_NEEDED")
``` 
? My main concern is that it might have some unintended consequences for other unit tests.

agent/engine/docker_task_engine.go
```diff
@@ -830,6 +832,16 @@ func (engine *DockerTaskEngine) Capabilities() []string {
 		}
 	}
 
+	// TODO we also need to disable this if agent already managing older tasks
+	if engine.cfg.TaskCPUMemLimit {
+		if _, ok := supportedVersions[dockerclient.Version_1_22]; ok {
+			capabilities = append(capabilities, capabilityPrefix+capabilityTaskCPUMemLimit)
+		} else {
+			seelog.Warn("Task CPU + Mem Limit disabled due to unsupported Docker version")
```
aaithal: can this be `"Setting task CPU and Memory Limit disabled .."` instead?

agent/engine/docker_task_engine.go
```diff
@@ -771,6 +772,7 @@ func (engine *DockerTaskEngine) State() dockerstate.TaskEngineState {
 //    com.amazonaws.ecs.capability.ecr-auth
 //    com.amazonaws.ecs.capability.task-iam-role
 //    com.amazonaws.ecs.capability.task-iam-role-network-host
+//    com.amazonaws.ecs.capability.task-resource-limit
```
aaithal: nit: this should be `ecs.task-cpu-mem-limit`

agent/engine/docker_task_engine.go
```diff
@@ -830,6 +832,16 @@ func (engine *DockerTaskEngine) Capabilities() []string {
 		}
 	}
 
+	// TODO we also need to disable this if agent already managing older tasks
+	if engine.cfg.TaskCPUMemLimit {
+		if _, ok := supportedVersions[dockerclient.Version_1_22]; ok {
+			capabilities = append(capabilities, capabilityPrefix+capabilityTaskCPUMemLimit)
```
aaithal: this shouldn't use `capabilityPrefix`. It should use `"ecs.capability."` instead (refer https://github.com/aws/amazon-ecs-agent/blob/enis/agent/app/agent_capability.go#L29)

agent/engine/docker_task_engine.go
```diff
@@ -830,6 +832,16 @@ func (engine *DockerTaskEngine) Capabilities() []string {
...
+		} else {
+			seelog.Warn("Task CPU + Mem Limit disabled due to unsupported Docker version")
+			engine.cfg.TaskCPUMemLimit = false
```
aaithal: The approach we've taken recently is to kill the agent with a terminal error code if it can't register with the required capability when asked to do so via the config. We should rethink if we have to take the same approach here. The argument is that it's better to not register the instance than register it without the capability that was asked for as it leads to a more consistent troubleshooting experience.

petderek: I think its slightly different here. In this case, 'enabled' is the default and it can be manually overridden to false.

Since it is a default, it would be disruptive to not register the instance (in case of updates, old versions of docker, etc).

petderek: After looking over some things further we think that the best way to handle this is to make enhancements to the config package. We need to fail if the value was set explicitly, but we should only fall back to old behavior gracefully if the value was derived from a default. 

Its a significant change, so I'd like to do that on a separate PR and keep progress moving forward on this branch.

agent/engine/docker_task_engine.go
```diff
@@ -45,7 +45,9 @@ const (
 	capabilityPrefix             = "com.amazonaws.ecs.capability."
 	capabilityTaskIAMRole        = "task-iam-role"
 	capabilityTaskIAMRoleNetHost = "task-iam-role-network-host"
+	capabilityTaskCPUMemLimit    = "task-cpu-mem-limit"
 	labelPrefix                  = "com.amazonaws.ecs."
+	attributePrefix              = "ecs.capability."
```
aaithal: If/when you rebase from `dev`, this is going to cause a merge conflict. I'd suggest that soon after merging this PR as the capabilities code has been moved to "agent/app/agent_capabilities.go"

agent/config/config.go
```diff
@@ -288,6 +288,7 @@ func environmentConfig() (Config, error) {
 	appArmorCapable := utils.ParseBool(os.Getenv("ECS_APPARMOR_CAPABLE"), false)
 	taskIAMRoleEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE"), false)
 	taskIAMRoleEnabledForNetworkHost := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST"), false)
+	taskCPUMemLimitEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT"), true)
```
richardpen: Can you add this to the readme file?

PR #946, opened 2017-08-24
engine: add test for ContainerNextState with deps

samuelkarp:
### Summary
Adds a test that specifically covers dependency states with `ContainerNextState`

### Implementation details
New test!

### Testing

New tests cover the changes: yes!

None

### Licensing
This contribution is under the terms of the Apache 2.0 License: yes, Amazon employee

### DIFF
--- a/agent/engine/task_manager_test.go
+++ b/agent/engine/task_manager_test.go
@@ -130,7 +130,7 @@ func TestContainerNextState(t *testing.T) {
 		expectedTransitionPossible   bool
 	}{
 		// NONE -> RUNNING transition is allowed and actionable, when desired is Running
-		// The exptected next status is Pulled
+		// The expected next status is Pulled
 		{api.ContainerStatusNone, api.ContainerRunning, api.ContainerPulled, true, true},
 		// NONE -> NONE transition is not be allowed and is not actionable,
 		// when desired is Running
@@ -151,7 +151,7 @@ func TestContainerNextState(t *testing.T) {
 		// actionable, when desired is STOPPED
 		{api.ContainerPulled, api.ContainerStopped, api.ContainerStopped, false, true},
 		// CREATED -> RUNNING transition is allowed and actionable, when desired is Running
-		// The exptected next status is Running
+		// The expected next status is Running
 		{api.ContainerCreated, api.ContainerRunning, api.ContainerRunning, true, true},
 		// CREATED -> CREATED transition is not allowed and not actionable,
 		// when desired is Running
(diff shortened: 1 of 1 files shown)


agent/engine/task_manager_test.go
```diff
@@ -201,8 +201,61 @@ func TestContainerNextState(t *testing.T) {
...
+}
+
+func TestContainerNextStateWithDependencies(t *testing.T) {
+	testCases := []struct {
+		containerCurrentStatus       api.ContainerStatus
+		containerDesiredStatus       api.ContainerStatus
+		dependencyCurrentStatus      api.ContainerStatus
+		expectedContainerStatus      api.ContainerStatus
+		expectedTransitionActionable bool
+		expectedTransitionPossible   bool
```
petderek: Possible and actionable are always the same in your test cases -- is that coincidence or are there situations where they differ?

agent/engine/task_manager_test.go
```diff
@@ -201,8 +201,61 @@ func TestContainerNextState(t *testing.T) {
...
+		// The expected next status is Pulled
+		{api.ContainerStatusNone, api.ContainerRunning, api.ContainerRunning, api.ContainerPulled, true, true},
+		// NONE -> RUNNING transition is allowed and actionable, when desired is Running and dependency is Stopped
+		// The expected next status is Pulled
+		{api.ContainerStatusNone, api.ContainerRunning, api.ContainerStopped, api.ContainerPulled, true, true},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("%s to %s Transition",
+			tc.containerCurrentStatus.String(), tc.containerDesiredStatus.String()), func(t *testing.T) {
```
petderek: is .String() implied by %s? Or am I thinking too much in Java?

aaithal: wait, what's the question?

samuelkarp: `.String()` is implied: https://play.golang.org/p/KpcJEdaPKy

I was following the existing code.  Can change it if you want.

agent/engine/task_manager_test.go
samuelkarp: See here:

https://github.com/samuelkarp/amazon-ecs-agent/blob/0b451f9583ecfcd3959890f03662aa8554dbab8e/agent/engine/task_manager.go#L480-L482

> // 'Stopped, false, true' -> "You can move it to known stopped, but you don't have to call a transition function"
> // 'Running, true, true' -> "You can move it to running and you need to call the transition function"
> // 'None, false, false' -> "This should not be moved; it has unresolved dependencies or is complete; no knownstatus change"

It's not covered in my test cases because I don't need it to be right now.

agent/engine/task_manager_test.go
petderek: I'm not sure what the convention is, but I'll typically vote for less code. Up to you.

agent/engine/task_manager_test.go
petderek: Are those paths covered in other tests? I'm not sure what I can take away from that commented block.

samuelkarp: Covered in `TestContainerNextState`.

PR #964, opened 2017-09-01
Add more logging during websocket connections.

jhaynes:
### Summary

There is frequent confusion regarding our logging of websocket disconnects. This adds logging for the subsequent reconnection.

### Implementation details
NA

### Testing
NA

NA

### Licensing

This contribution is under the terms of the Apache 2.0 License:  yes

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -200,6 +200,7 @@ func (acsSession *session) Start() error {
 			if shouldReconnectWithoutBackoff(acsError) {
 				// If ACS closed the connection, there's no need to backoff,
 				// reconnect immediately
+				seelog.Info("ACS Websocket connection closed for a valid reason")
 				acsSession.backoff.Reset()
 				sendEmptyMessageOnChannel(connectToACS)
 			} else {
@@ -293,6 +294,8 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer, timer t
 		seelog.Errorf("Error connecting to ACS: %v", err)
 		return err
 	}
+	seelog.Info("Connected to ACS endpoint")
+
 	acsSession.resources.connectedToACS()
 
 	backoffResetTimer := time.AfterFunc(
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -383,11 +383,11 @@ func newSessionResources(credentialsProvider *credentials.Credentials) sessionRe
 
 // acsWsURL returns the websocket url for ACS given the endpoint
 func acsWsURL(endpoint, cluster, containerInstanceArn string, taskEngine engine.TaskEngine, acsSessionState sessionState) string {
-	acsUrl := endpoint
+	acsURL := endpoint
 	if endpoint[len(endpoint)-1] != '/' {
-		acsUrl += "/"
+		acsURL += "/"
 	}
-	acsUrl += "ws"
+	acsURL += "ws"
 	query := url.Values{}
 	query.Set("clusterArn", cluster)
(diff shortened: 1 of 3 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -293,6 +293,8 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer, timer t
 		seelog.Errorf("Error connecting to ACS: %v", err)
 		return err
 	}
+	seelog.Infof("Connected to ACS endpoint")
```
aaithal: nit: this, and the other instance can be `Info`

agent/wsclient/client.go
```diff
@@ -132,6 +132,7 @@ type ClientServerImpl struct {
 // 'MakeRequest' can be made after calling this, but responss will not be
 // receivable until 'Serve' is also called.
 func (cs *ClientServerImpl) Connect() error {
+	seelog.Debugf("Establishing a Websocket connection to ", cs.URL)
```
aaithal: Can you use the format string here instead?  `"Establishing a Websocket connection to: %s", cs.URL`

agent/wsclient/client.go
```diff
@@ -132,6 +132,7 @@ type ClientServerImpl struct {
 // 'MakeRequest' can be made after calling this, but responss will not be
 // receivable until 'Serve' is also called.
 func (cs *ClientServerImpl) Connect() error {
+	seelog.Debugf("Establishing a Websocket connection to %s", cs.URL)
```
richardpen: With the log here, I think we should delete this [line](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/tcs/handler/handler.go#L92), they are duplicate.

jhaynes: I think I got all the duplicate lines [here](https://github.com/aws/amazon-ecs-agent/pull/964/files#diff-19e06c1031fd70ac815de965c115b519L92) but if not, I'll go find the others.

agent/acs/handler/acs_handler.go
```diff
@@ -200,6 +200,7 @@ func (acsSession *session) Start() error {
 			if shouldReconnectWithoutBackoff(acsError) {
 				// If ACS closed the connection, there's no need to backoff,
 				// reconnect immediately
+				seelog.Info("ACS Websocket connection closed for a valid reason")
```
samuelkarp: Is it worth logging the error?

agent/tcs/handler/handler.go
```diff
@@ -86,10 +87,9 @@ func StartSession(params TelemetrySessionParams, statsEngine stats.Engine) error
 func startTelemetrySession(params TelemetrySessionParams, statsEngine stats.Engine) error {
 	tcsEndpoint, err := params.ECSClient.DiscoverTelemetryEndpoint(params.ContainerInstanceArn)
 	if err != nil {
-		log.Errorf("Unable to discover poll endpoint: ", err)
+		seelog.Errorf("Unable to discover poll endpoint: ", err)
```
samuelkarp: This either needs a format string (probably `%v`) or should be `Error` instead of `Errorf`.

agent/acs/handler/acs_handler.go
jhaynes: I opted to remove it from Info level logging since it seemed to lead to confusion more than anything. However, it is still logged at Debug in the wsclient code [here](https://github.com/aws/amazon-ecs-agent/pull/964/files#diff-944db589e66167aa4670b399248d81d0R295)

agent/tcs/handler/handler.go
jhaynes: Agreed. Fixing

PR #968, opened 2017-09-06
Add blackhole imds route for awsvpc network mode

aaithal:
### Summary

Add AWSVPCBlockInstanceMetdata config (`ECS_AWSVPC_BLOCK_IMDS`) for blocking access to instance metdata for awsvpc tasks

### Implementation details

* 5e8001c agent/app: register task-eni-block-instance-metadata capability when AWSVPCBlockInstanceMetdata config is enabled
* 63734d3 agent/ecscni,agent/engine: wire in AWSVPCBlockInstanceMetdata config to CNI config
* 5f0e4ed agent/config: add AWSVPCBlockInstanceMetdata config for blocking access to instance metdata for awsvpc tasks

### Testing

None

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ configure them as something other than the defaults.
 | `ECS_INSTANCE_ATTRIBUTES` | `{"stack": "prod"}` | These attributes take effect only during initial registration. After the agent has joined an ECS cluster, use the PutAttributes API action to add additional attributes. For more information, see [Amazon ECS Container Agent Configuration](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) in the Amazon ECS Developer Guide.| `{}` | `{}` |
 | `ECS_ENABLE_TASK_ENI` | `false` | Whether to enable task networking for task to be launched with its own network interface | `false` | `false` |
 | `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | `/amazon-ecs-cni-plugins` |
-| `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to Instance Metdata for Tasks started with `awsvpc` network mode | `false` | `false`|
+| `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to [Instance Metdata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for Tasks started with `awsvpc` network mode | `false` | `false`|
 
 ### Persistence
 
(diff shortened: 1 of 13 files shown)


aaithal:
@nmeyerhans I decided not to do a changelog entry as I think this should be covered under  https://github.com/aws/amazon-ecs-agent/commit/0e782a542e63ad9de52a13fc57052d8a0c987731. Do you think this warrants an entry by itself?

agent/app/agent_capability.go
```diff
@@ -124,6 +126,14 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 		capabilities = append(capabilities, taskENIVersionAttribute)
 	}
 
+	if agent.cfg.AWSVPCBlockInstanceMetdata {
```
richardpen: You may want to move this above inside the check for `TaskENIEnabled` since this relies on the `awsvpc` network mode and cni plugins.

README.md
```diff
@@ -177,6 +177,7 @@ configure them as something other than the defaults.
 | `ECS_INSTANCE_ATTRIBUTES` | `{"stack": "prod"}` | These attributes take effect only during initial registration. After the agent has joined an ECS cluster, use the PutAttributes API action to add additional attributes. For more information, see [Amazon ECS Container Agent Configuration](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) in the Amazon ECS Developer Guide.| `{}` | `{}` |
 | `ECS_ENABLE_TASK_ENI` | `false` | Whether to enable task networking for task to be launched with its own network interface | `false` | `false` |
 | `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | `/amazon-ecs-cni-plugins` |
+| `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to Instance Metdata for Tasks started with `awsvpc` network mode | `false` | `false`|
```
richardpen: Do you we need to add a link to the instance metadata here, I'm not sure if all the customers know Instance Metadata?

agent/config/config_test.go
```diff
@@ -380,3 +380,11 @@ func TestImageCleanupMinimumNumImagesToDeletePerCycle(t *testing.T) {
 		t.Errorf("Wrong value for NumImagesToDeletePerCycle: %v", cfg.NumImagesToDeletePerCycle)
 	}
 }
+
+func TestAWSVPCBlockInstanceMetadata(t *testing.T) {
```
richardpen: Please add a simple comment here?

README.md
aaithal: Sure, I can do that.

agent/config/config_test.go
aaithal: Isn't the name itself descriptive?

agent/app/agent_capability.go
aaithal: I thought about and ended up not doing it as it's essentially a nop to do this if Task ENI is not enabled. I'll change it if you insist.

agent/config/config_test.go
richardpen: Only if I know that this test was in the config package, it will have different meaning in `engine` package, where the test may be testing the process to block the metadata. Ignore my comments if we are supposed to have the context when reading the code.

agent/app/agent_capability.go
richardpen: Move this above makes the logic more clear to me, let's do it, thanks.

aaithal:
@nmeyerhans that makes me happy as well. I thought the changes here were small enough and didn't need additional context. Let me know if you had any particular commit(s) in mind and I can add some context.

nmeyerhans:
One thing that'd help the commit messages here is to ensure that the "subject" (first line of the message) is no longer than about 50 characters. When displaying a commit log, Github truncates at 70 characters, so there's a little room to be flexible here, but most of the commits in this PR don't even fit within the 70 character limit. Any context or details that don't fit within ~50 characters should be expanded upon in the "body" of the commit message.

Note for others who might be reading: Do as I say, not as I do. I've certainly pushed some poor commit messages in the past. More recently, though, we've had internal conversations about improving our commit logs, and I'm trying to drive that effort forward...

PR #969, opened 2017-09-07
api, engine: Adding task cgroup hooks

vsiddharth:

### Summary

Hooks the cgroup changes into `api` and `engine`

### Testing

### Licensing


agent/api/task_unix.go
```diff
@@ -15,8 +15,21 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+	"github.com/cihub/seelog"
+	docker "github.com/fsouza/go-dockerclient"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
 const (
 	portBindingHostIP = "0.0.0.0"
+	sepForwardSlash   = "/"
```
richardpen: Can you rename this to something like `pathSeparator`?

agent/api/task_unix.go
```diff
@@ -25,6 +38,70 @@ func getCanonicalPath(path string) string { return path }
...
+
+	cgroupRoot := strings.Join([]string{config.DefaultTaskCgroupPrefix, taskID}, sepForwardSlash)
+	return cgroupRoot, nil
+}
+
+// BuildLinuxResourceSpec returns a linuxResources object for the task cgroup
+func (task *Task) BuildLinuxResourceSpec() specs.LinuxResources {
+	linuxResourceSpec := specs.LinuxResources{}
+
+	if !(utils.ZeroOrNil(task.VCPULimit)) {
```
richardpen: Will this task be started with `cpu limit` zero in this case? Should this return an error if the `task.VCPULimit` is zero? The same for the `memory limit`.

agent/api/task_unix.go
```diff
@@ -25,6 +38,70 @@ func getCanonicalPath(path string) string { return path }
...
+
+// overrideCgroupParent updates hostconfig with cgroup parent when task cgroups
+// are enabled
+func (task *Task) overrideCgroupParent(hostConfig *docker.HostConfig) {
+	if !task.CgroupEnabled() {
+		return
+	}
+
+	cgroupRoot, err := task.BuildCgroupRoot()
+	if err != nil {
```
richardpen: Why this error was ignored, shouldn't  the task moved to be stopped when it can't get the group root?

agent/engine/task_manager.go
```diff
@@ -606,6 +611,10 @@ func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 	mtask.engine.sweepTask(mtask.Task)
 
 	// TODO: Remove platform task resources
+	err := mtask.CleanupPlatformResources()
+	if err != nil {
+		seelog.Warnf("Unable to cleanup platform resources: %v", err)
```
richardpen: Can you change `platform resource` to be more specific? `platform resource` isn't very clear to me.

agent/engine/task_manager_linux.go
```diff
@@ -15,9 +15,19 @@
...
 package engine
 
+import (
+	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+// control is used to manipulate cgroups and ease testing
+var control cgroup.Control
```
richardpen: Is there anyway to get rid of the global variable, if it is specific for each task, can you add it to the `mtask` struct?

agent/engine/task_manager_linux.go
```diff
@@ -33,12 +43,57 @@ func (mtask *managedTask) CleanupPlatformResources() error {
...
 func (mtask *managedTask) setupCgroup() error {
-	// Stub for now
+	cgroupRoot, err := mtask.Task.BuildCgroupRoot()
+	if err != nil {
+		return errors.Wrap(err, "mtask setup cgroup: unable to determine cgroup root")
+	}
+
+	seelog.Debugf("Setting up cgroup at: %s", cgroupRoot)
+
+	if control.Exists(cgroupRoot) {
```
richardpen: Under which scenario will this happen, what if two tasks accidentally have the same `cgroupRoot`? Should this error just be ignored?

agent/api/task_unix.go
samuelkarp: You should really just use `filepath.Join` here.  If you need to manually construct for some strange reason, use `filepath.Separator`.

vsiddharth: @richardpen I'll remove this const
@samuelkarp thanks!

agent/api/task_unix.go
```diff
@@ -25,6 +38,70 @@ func getCanonicalPath(path string) string { return path }
 
 // CgroupEnabled checks whether cgroups need be enabled at the task level
 func (task *Task) CgroupEnabled() bool {
```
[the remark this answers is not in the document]
vsiddharth: Not all tasks need cgroups even when the feature is enabled at the agent level

agent/api/task_unix.go
vsiddharth: The `taskDef` validation happens elsewhere. 
This method would be called only when task cgroups are required.

agent/api/task_unix.go
vsiddharth: This should fail only when the `taskARN` is invalid and I thought that to be highly infrequent. However, I can return an error to make it more semantically correct.

agent/engine/task_manager.go
petderek: I don't know that we can. There isn't a windows equivalent to this, but we need a generic hook here. I think we are all open to suggestions though 😄

vsiddharth: I'm considering cgroups to be a platform specific resource at the task level.
Everything worked upon here has that idea. 

I'd like to hear of alternatives that you think would make this clearer.

agent/engine/task_manager_linux.go
vsiddharth: `mtask` is not platform specific and cgroups don't apply to windows. 
To get rid of this global would require some sort of encapsulation and since we only have one resource here, I'm not sure if its necessary.

agent/engine/task_manager_linux.go
vsiddharth: Two tasks on the same instance should never have the same taskARN. 
This is primarily to avoid task cgroup recreation

agent/api/task_unix.go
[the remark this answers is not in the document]
richardpen: thanks for clarifying that.

agent/api/task_unix.go
```diff
@@ -25,6 +38,70 @@ func getCanonicalPath(path string) string { return path }
...
+
+// BuildCgroupRoot helps build the task cgroup prefix
+// Example: /ecs/task-id
+func (task *Task) BuildCgroupRoot() (string, error) {
+	taskID, err := task.GetID()
+	if err != nil {
+		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id")
+	}
+
+	cgroupRoot := strings.Join([]string{config.DefaultTaskCgroupPrefix, taskID}, sepForwardSlash)
```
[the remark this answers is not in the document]
samuelkarp: It would be better to just use `filepath.Join` here.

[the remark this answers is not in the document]
petderek: Agreed. I didn't realize it worked like that

agent/engine/task_manager_linux.go
```diff
@@ -33,12 +43,57 @@ func (mtask *managedTask) CleanupPlatformResources() error {
...
+		seelog.Debugf("Cgroup at %s already exists, skipping creation", cgroupRoot)
+		return nil
+	}
+
+	linuxResourceSpec := mtask.Task.BuildLinuxResourceSpec()
```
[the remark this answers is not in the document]
vsiddharth: I wanted to do something along those lines but I'll address that in a different PR.

agent/api/task.go
```diff
@@ -432,6 +432,11 @@ func (task *Task) dockerHostConfig(container *Container, dockerContainerMap map[
 		}
 	}
 
+	err = task.platformHostConfigOverride(hostConfig)
+	if err != nil {
```
sharanyad: Can you please add a test for this?

vsiddharth: Done

agent/api/task_unix.go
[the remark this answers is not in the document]
samuelkarp: @vsiddharth We'll need to have cgroups for every task when the feature is enabled.  We'll also need to disable the feature if there are any running tasks that were created with an old version of the agent.

Because cgroup subsystem settings are inherited (including things like memory and relative weights like CPU shares), we need to bring some of the values up from the container-level cgroups to the parent task-level cgroup or ensure that the parent task-level cgroup has no limit.

This is somewhat easy to reason about with memory, as a limit that is lower than the limit on a child cgroup means that the child will not be able to even reach its limit.  However, the naive approach of summing the child limits to make the parent limit is inflexible and does not allow for overcommit.  Instead, we'll likely want to ensure two things:

1. When no memory limit is specified at the task level, ensure that the parent cgroup has no limit
2. When a memory limit is specified at the task level, ensure that it is greater or equal to the specified hard limit of each individual container (when no hard limit is specified, the task-level limit can be the only limit in place)

For CPU shares, the necessity to bring up container-level CPU shares is a bit less obvious.  There are two things that are in play here:

1. ECS uses the `cpu` field in the task definition as a placement/scheduling constraint, which requires that each container's CPU share is relative to the other containers placed by ECS
2. When a parent cgroup is present, the child cgroup's CPU share is relative to the other children of the same parent and the aggregate of those children is inside the parent's CPU share which is relative to other cgroups at its level.  Another way to think about this is that cgroups' CPU shares are only relative to other cgroups that share the same direct parent

Thus, to preserve the placement property that ECS relies upon, we need to ensure that the aggregate of the containers' CPU share is relative at the task level to other tasks.  We can do this by summing the CPU share of all containers in a task and assigning that CPU share value to the task-level cgroup.

agent/api/task.go
```diff
@@ -432,9 +433,11 @@ func (task *Task) dockerHostConfig(container *Container, dockerContainerMap map[
 		}
 	}
 
-	err = task.platformHostConfigOverride(hostConfig)
-	if err != nil {
-		return nil, &HostConfigError{err.Error()}
+	if cfg.TaskCPUMemLimit {
```
sharanyad: If we are going to simply override other host config fields too in the future (as the method name suggests), having this check inside platformHostConfigOverride method makes more sense.

vsiddharth: I was hesitant about passing `cfg` down to the platformOverrides but I think its a better call for future overrides.

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+		// TODO: DefaultCPUPeriod only permits 10VCPUs.
+		// Adaptive calculation of CPUPeriod required for further support
+		linuxResourceSpec.CPU = &specs.LinuxCPU{
+			Quota:  &taskCPUQuota,
+			Period: &taskCPUPeriod,
+		}
+	} else {
+		taskCPUShares := uint64(0)
+		for _, container := range task.Containers {
+			if !(utils.ZeroOrNil(container.CPU)) {
```
samuelkarp: nit: extra parens

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+
+		linuxResourceSpec.CPU = &specs.LinuxCPU{
+			Shares: &taskCPUShares,
+		}
+	}
+
+	// If task memory limit is not present, cgroup parent memory is not set
+	// If task memory limit is set, ensure that no container
+	// of this task has a greater request
+	if !(utils.ZeroOrNil(task.MemoryLimit)) {
```
samuelkarp: nit: extra parens

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+		// Adaptive calculation of CPUPeriod required for further support
+		linuxResourceSpec.CPU = &specs.LinuxCPU{
+			Quota:  &taskCPUQuota,
+			Period: &taskCPUPeriod,
+		}
+	} else {
+		taskCPUShares := uint64(0)
+		for _, container := range task.Containers {
+			if !(utils.ZeroOrNil(container.CPU)) {
+				taskCPUShares += uint64(container.CPU)
```
samuelkarp: Why += and not just =?

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+		taskCPUQuota := int64(task.VCPULimit * defaultCPUPeriod)
+
+		// TODO: DefaultCPUPeriod only permits 10VCPUs.
+		// Adaptive calculation of CPUPeriod required for further support
+		linuxResourceSpec.CPU = &specs.LinuxCPU{
+			Quota:  &taskCPUQuota,
+			Period: &taskCPUPeriod,
+		}
+	} else {
+		taskCPUShares := uint64(0)
```
samuelkarp: I think it would be slightly more idiomatic to write this as `var taskCPUShares uint64`.

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+	}
+
+	// If task memory limit is not present, cgroup parent memory is not set
+	// If task memory limit is set, ensure that no container
+	// of this task has a greater request
+	if !(utils.ZeroOrNil(task.MemoryLimit)) {
+		taskMemoryLimit := int64(task.MemoryLimit)
+		for _, container := range task.Containers {
+			containerMemoryLimit := int64(container.Memory)
+			if !(utils.ZeroOrNil(containerMemoryLimit)) && containerMemoryLimit > taskMemoryLimit {
```
samuelkarp: nit: extra parens

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+		linuxResourceSpec.Memory = &specs.LinuxMemory{
+			Limit: &taskMemoryLimit,
+		}
+	}
+
+	return linuxResourceSpec, nil
+}
+
+// platformHostConfigOverride to override platform specific feature sets
+func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig, cfg *config.Config) error {
```
samuelkarp: I think we usually order it so `config.Config` would be first in the argument list.

samuelkarp: It looks like this function is an entrypoint to the file; I think it'd be worth moving it up to the top.  In general, I'd prefer code to be ordered such that caller is above callee, so you can read through the code in the order that functions are called (you start at the most abstract and go to the most concrete).

agent/engine/task_manager_linux.go
```diff
@@ -15,30 +15,88 @@
...
+
+	err = control.Remove(cgroupRoot)
+	if err != nil {
+		return errors.Wrapf(err, "mtask cleanup cgroup: unable to remove cgroup at %s", cgroupRoot)
+	}
+
 	return nil
 }
+
+// setControl is used to set the cgroup controller
```
samuelkarp: Is this function just for testing?  If so, please add that to the comment (and you might also think about changing the name to `unsafeSetControl`).

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
-// CgroupEnabled checks whether cgroups need be enabled at the task level
-func (task *Task) CgroupEnabled() bool {
-	//stub for now
-	return false
+// BuildCgroupRoot helps build the task cgroup prefix
+// Example: /ecs/task-id
+func (task *Task) BuildCgroupRoot() (string, error) {
+	taskID, err := task.GetID()
+	if err != nil {
+		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id")
```
richardpen: can you also log the taskarn, either here or where this error was logged?

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+
+	cgroupRoot := filepath.Join(config.DefaultTaskCgroupPrefix, taskID)
+	return cgroupRoot, nil
+}
+
+// BuildLinuxResourceSpec returns a linuxResources object for the task cgroup
+func (task *Task) BuildLinuxResourceSpec() (specs.LinuxResources, error) {
+	linuxResourceSpec := specs.LinuxResources{}
+
+	if !(utils.ZeroOrNil(task.VCPULimit)) {
```
richardpen: Question here, do we need to check the value of the `task.VCPULimit`, say can I set it to `1000000000`?

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+		}
+		linuxResourceSpec.Memory = &specs.LinuxMemory{
+			Limit: &taskMemoryLimit,
+		}
+	}
+
+	return linuxResourceSpec, nil
+}
+
+// platformHostConfigOverride to override platform specific feature sets
```
richardpen: nit: `platformHostConfigOverride overrides`

agent/api/task_unix.go
```diff
@@ -15,16 +15,105 @@
...
+	}
+	return nil
+}
+
+// overrideCgroupParent updates hostconfig with cgroup parent when task cgroups
+// are enabled
+func (task *Task) overrideCgroupParent(hostConfig *docker.HostConfig) error {
+	cgroupRoot, err := task.BuildCgroupRoot()
+	if err != nil {
+		seelog.Debugf("Unable to obtain task cgroup root: %v", err)
```
richardpen: task arn is needed in the log. See the comments above, the task arn can be added here or wrapped in the err.

agent/engine/task_manager.go
```diff
@@ -130,8 +130,13 @@ func (mtask *managedTask) overseeTask() {
 			llog.Debug("Task not steady state or terminal; progressing it")
 
 			// TODO:
-			// 1) Add platform resources setup and failure handling
-			// 2) Add new task resources provisioned state ?
+			// 1) Add new task resources provisioned state ?
```
richardpen: I'd recommend create an issue for this if we plan to address this before the release.

agent/engine/task_manager.go
```diff
@@ -606,6 +611,10 @@ func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 	mtask.engine.sweepTask(mtask.Task)
 
 	// TODO: Remove platform task resources
+	err := mtask.CleanupPlatformResources()
```
richardpen: Is the todo here still needed? Also will this impact other new task if we don't remove this right after the task is stopped, since the cleanup happens after `TASK_CLEANUP_DURATION`?

agent/engine/task_manager.go
richardpen: Please log at least the taskarn here.

agent/engine/task_manager.go
```diff
@@ -130,8 +130,13 @@ func (mtask *managedTask) overseeTask() {
 			llog.Debug("Task not steady state or terminal; progressing it")
 
 			// TODO:
-			// 1) Add platform resources setup and failure handling
-			// 2) Add new task resources provisioned state ?
+			// 1) Add new task resources provisioned state ?
+			err := mtask.SetupPlatformResources()
+			if err != nil {
+				seelog.Criticalf("Unable to setup task platform resources: %v", err)
```
richardpen: this also requires task.arn to be logged.

agent/engine/testdata/test_tasks/sleep5TaskCgroup.json
```diff
@@ -0,0 +1,40 @@
+{
```
richardpen: If the `CPULimit` and `Memory` are the only missing fields in the `sleep5.json`, can we reuse the `sleep5.json` and modify the field after loading the task. We don't want every kind of task to have a json file in this directory for a unit test.

agent/engine/task_manager_linux_test.go
```diff
@@ -16,28 +16,381 @@
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockControl := mock_cgroup.NewMockControl(ctrl)
+	setControl(mockControl)
+	assert.Equal(t, mockControl, control)
+}
+
+// Cgroup Cleanup Tests
+
```
richardpen: Please fix these two line, whether remove or rewrite the comments in the format `Testxxxx tests xxx`.

agent/api/task.go
```diff
@@ -395,11 +396,11 @@ func (task *Task) dockerConfigVolumes(container *Container) (map[string]struct{}
 	return volumeMap, nil
 }
 
-func (task *Task) DockerHostConfig(container *Container, dockerContainerMap map[string]*DockerContainer) (*docker.HostConfig, *HostConfigError) {
-	return task.Overridden().dockerHostConfig(container.Overridden(), dockerContainerMap)
+func (task *Task) DockerHostConfig(container *Container, dockerContainerMap map[string]*DockerContainer, cfg *config.Config) (*docker.HostConfig, *HostConfigError) {
```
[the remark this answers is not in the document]
samuelkarp: I know why you added `cfg`.  I think you should use it in `PostUnmarshalTask` instead.  That's a more-appropriate place for this logic to happen.

agent/api/task_unix.go
samuelkarp: If you're aggregating, you should avoid the logic with the `defaultCPUShare`.  I don't think we want to include that right now.

agent/engine/task_manager_linux.go
samuelkarp: This is an anti-pattern.  I would like you to avoid doing this if possible.

agent/api/task.go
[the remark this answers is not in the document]
vsiddharth: I don't understand your comment here, `PostUnmarshalTask` is called from `AddTask` for the task level overrides while `DockerHostConfig` is called from `createContainer`. 

Could you please provide more context here ?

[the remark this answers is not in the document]
samuelkarp: You have logic that depends on `cfg`.  You should only read `cfg` from `PostUnmarshalTask`.  You can use `PostUnmarshalTask` to add some values to the `api.Task` struct that you then read in `DockerHostConfig`.  This avoids coupling `DockerHostConfig` to `cfg`, which is why I prefer this approach.

agent/api/task_unix.go
vsiddharth: Could you please provide more context on why we don't want to include `defaultCPUShare` now?

agent/engine/testdata/test_tasks/sleep5TaskCgroup.json
vsiddharth: I would prefer to keep this `json` as it makes testing cleaner. 
It's also going to be used by integ tests

agent/api/task.go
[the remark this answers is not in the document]
vsiddharth: I don't like the approach of updating `api.Task` from `engine`. 
I prefer the current approach but if you are against the idea of passing the config around, I'd prefer to pass pass the config boolean `TaskCPUMemLimit`.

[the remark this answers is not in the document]
samuelkarp: I'm not suggesting updating `api.Task` from `engine`; I'm suggesting doing it from `PostUnmarshalTask` which is specifically for things like this.

agent/api/task_unix.go
samuelkarp: `defaultCPUShare` isn't something we schedule on and we don't make any guarantees around the weight when you assign `"cpu":0`.  If you sum `defaultCPUShare`, a task with 10 containers each with `"cpu":0` will have a weight of 20 compared to a task with 1 container that has `"cpu":0` that will end up with a weight of 2.  In each of these cases, they should have a weight of 2.

petderek: @samuelkarp  My understanding is that summing to 20 is needed in order to preserve the current experience, even though we don't guarantee scheduling.

samuelkarp: When you set `"cpu":0`, you're saying "I don't want any guaranteed capacity, but I want to be able to burst."  We initially set `CpuShares` in the Docker HostConfig to 0, which got silently converted inside Docker to be 2 in the actual cgroup, because 2 is the actual minimum that the kernel enforces.  Later, a newer version of Docker started rejecting a specification of 0 and we added some compatibility code in the agent to handle that.  However, the intent here is still to get as little as possible; if you sum, you're starting to actually give a capacity guarantee (or, alternately, steal capacity under contention from containers that are *actually* intended to have a capacity guarantee).

agent/api/task.go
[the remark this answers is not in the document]
vsiddharth: `PostUnmarshalTask` is invoked from the `engine`(AddTask) and I don't like the idea of using it to update specific attributes of the `Task` struct. `PostUnmarshalTask` looks like an assimilation of hacks including the empty volume initialization. 

I'm also not keen on adding new attributes to the task struct just for this purpose. 
I prefer the current approach or passing the config boolean to resolve this.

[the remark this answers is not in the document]
samuelkarp: `PostUnmarshalTask` is *specifically* for this purpose.  Yes, it's a collection of hacks but it at least keeps all the hacks in one place.

agent/api/task_unix.go
vsiddharth: Hmm, so if we have a task (without task limits) but with `n containers` each without any `cpu` limits. Should we just set the task level cgroup with `CPUShares := 2`?

samuelkarp: Yeah, if nothing has defined limits you should set the task CPU shares to 2.

vsiddharth: @samuelkarp thanks for clarifying the ambiguity here

vsiddharth:
Superseded by #984

Issue #426, opened 2016-06-14
Bind task to specific IP address


abatsakis:
Hi all,

I was wondering if there are any plans in supporting the binding of a task to a specific IP address on the host that can e.g. provided at RunTask time similar to docker's  -p IP:host_port:container_port

This is very handy for instances with multiple interfaces on different subnets

kiranmeduri:
IIUC you are asking on how can ECS tasks running on EC2 instances support ENIs. We are currently gathering requirements to support this and would love to understand more about your use-case. Is your case to assign one IP per task and have that IP constrained by policies defined in the VPC/Subnet/Security Groups? 

In the current system even if we support passing in IP in RunTask API, you need to know IP beforehand which is not true in RunTask case which doesn't take containerInstance as input. May be you meant StartTask API. Even in that scenario it is interesting to understand more about the end-to-end flow. 

Thanks for bringing this up.

-kiran

samuelkarp:
@abatsakis We're working on attaching ENIs to tasks and allowing IP addresses per task.  Does the proposal in https://github.com/aws/amazon-ecs-agent/pull/701 address your use-case?

adnxn:
@abatsakis We've released [task networking for containers](https://aws.amazon.com/about-aws/whats-new/2017/11/amazon-ecs-introduces-awsvpc-networking-mode-for-containers-to-support-full-networking-capabilities/) as proposed in #701. I'm closing this issue since we haven't heard back, but feel free to let us know if this addresses your use case.

PR #972, opened 2017-09-07
Do not explicitly set host IP to 0.0.0.0 for Linux, defer to Docker

goodhart:
### Summary

By hardcoding the host IP as 0.0.0.0, the agent is removing the ability for customers to specify specific host IP addresses (such as when there are multiple network interfaces). By default, the docker daemon defaults to 0.0.0.0, but this value may be overridden by launching the daemon with a --ip flag. [Reference](https://docs.docker.com/engine/reference/commandline/dockerd/)

This also has an implication for the CloudWatch Events published by ECS. Specifically, prior to this change the 'bindIP' field in the ECS Task Status events always contains '0.0.0.0'. By explicitly setting this value, customers can control the value that appears in these events, removing the need to query the ECS and/or EC2 control plane to obtain this value.

This change will benefit the following customer use cases:
[Issues-426](https://github.com/aws/amazon-ecs-agent/issues/426)
[AWS ECS Forum Thread 173250](https://forums.aws.amazon.com/thread.jspa?threadID=173250)
[Serverfault.com Question 695393](https://serverfault.com/questions/695393/amazon-ecs-docker-binding-container-to-specific-ip-address)

### Implementation details
Remove HostIP field when calling to the Docker daemon to create a container.

### Testing
Tests have been updated to no longer assert the host IP. Testing was performed by launching tasks against my personal EC2 hosts with this forked version of the agent, both with the default IP binding and an explicitly specified host binding IP set in the /etc/docker/daemon.json file.  Also verified that published ECS Task State events contain the new bound IP.

I have verified that:
* make release, make test, make run-integ-tests, and make run-functional-tests runs cleanly on 2017.03.f
* make release, make test, make run-integ-tests, and make run-functional-tests runs cleanly on 2015.09.d.  The tests do not build on versions of the AMI prior to this because the Debian image used by the test lives in a 2+ repository that is incompatible with older docker versions.
* Manually launched a server on 2017.03.f and verified that it correctly maintains the default behavior of binding to 0.0.0.0
* Manually launched a server on 2015.03.c and verified that it correctly maintains the default behavior of binding to 0.0.0.0. The agent certificates container does not build on versions of the AMI prior to this because the Debian repository related problems.
* Manually launched a server on 2017.03.f and verified that it uses the Docker –ip flag if specified.

- [X] Functional tests on Linux (`make run-functional-tests`) pass

New tests cover the changes: N/A - Tests have been updated to reflect change.

* Enhancement - Stop explicitly binding containers to 0.0.0.0 and defer to Docker daemon

### Licensing

### DIFF
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -616,9 +616,9 @@ func (task *Task) dockerPortMap(container *Container) map[docker.Port][]docker.P
 		dockerPort := docker.Port(strconv.Itoa(int(portBinding.ContainerPort)) + "/" + portBinding.Protocol.String())
 		currentMappings, existing := dockerPortMap[dockerPort]
 		if existing {
-			dockerPortMap[dockerPort] = append(currentMappings, docker.PortBinding{HostIP: portBindingHostIP, HostPort: strconv.Itoa(int(portBinding.HostPort))})
+			dockerPortMap[dockerPort] = append(currentMappings, docker.PortBinding{HostPort: strconv.Itoa(int(portBinding.HostPort))})
 		} else {
-			dockerPortMap[dockerPort] = []docker.PortBinding{{HostIP: portBindingHostIP, HostPort: strconv.Itoa(int(portBinding.HostPort))}}
+			dockerPortMap[dockerPort] = []docker.PortBinding{{HostPort: strconv.Itoa(int(portBinding.HostPort))}}
 		}
 	}
 	return dockerPortMap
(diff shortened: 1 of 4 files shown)


samuelkarp:
The code for this looks fine.  We're still figuring out whether we believe this will have any backwards-incompatible impact.

aaithal:
Verified that all tests [pass](https://github.com/aaithal/amazon-ecs-agent/commits/goodhart-972). Also, manually verified that on our oldest supported AMI, setting `--ip=<instance private ip>` and leaving it as is doesn't result in any unexpected behavior. 

Without setting `--ip` in `/etc/sysconfig/docker`, `networkBindings` looks like this:
```
                    "networkBindings": [
                        {
                            "protocol": "tcp",
                            "bindIP": "0.0.0.0",
                            "containerPort": 80,
                            "hostPort": 32768
                        }
                    ],
```

With `--ip=so.me.ip.addr`, `networkBindings` looks like this:
```
                    "networkBindings": [
                        {
                            "protocol": "tcp",
                            "bindIP": "so.me.ip.addr",
                            "containerPort": 80,
                            "hostPort": 32768
                        }
                    ],
```

So, we're all ready to merge this with the next release (after https://github.com/aws/amazon-ecs-agent/pull/999).

PR #973, opened 2017-09-07
app: Update agent startup for task cgroups

vsiddharth:

### Summary

Update agent startup for task cgroups

### Testing

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -108,7 +108,7 @@ func newAgent(
 		return nil, err
 	}
 
-	return &ecsAgent{
+	agent := &ecsAgent{
 		ctx:               ctx,
 		ec2MetadataClient: ec2MetadataClient,
 		cfg:               cfg,
@@ -119,7 +119,10 @@ func newAgent(
 		credentialProvider:    defaults.CredChain(defaults.Config(), defaults.Handlers()),
 		stateManagerFactory:   factory.NewStateManager(),
 		saveableOptionFactory: factory.NewSaveableOption(),
-	}, nil
+	}
+
+	agent.initPlatformResources()
+	return agent, nil
 }
 
 // printVersion prints the ECS Agent version string
@@ -164,9 +167,12 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		log.Criticalf("Error creating state manager: %v", err)
 		return exitcodes.ExitTerminal
 	}
-	// TODO:
-	// 1) Feature gate ECS cgroup root init
-	// 2) Update mocks and unit tests
+
+	err = agent.setupPlatformResources()
+	if err != nil {
+		log.Criticalf("Unable to setup platform resources: %v", err)
+		return exitcodes.ExitTerminal
+	}
 
 	// Register the container instance
 	err = agent.registerContainerInstance(taskEngine, stateManager, client)
--- a/agent/app/agent_unix.go
+++ b/agent/app/agent_unix.go
@@ -1,0 +1,57 @@
+// +build linux
+
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package app
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
+	"github.com/cihub/seelog"
+)
+
+// control is here to help create the cgroup root '/ecs' when
+// the feature is enabled. It also aids in unit testing
+var control cgroup.Control
+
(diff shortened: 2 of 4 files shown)


agent/app/agent_unix.go
```diff
@@ -0,0 +1,57 @@
...
+	if agent.cfg.TaskCPUMemLimit {
+		setControl(cgroup.New())
+	}
+}
+
+// setupPlatformResources helps set up the platform resource controllers
+// Currently used to setup '/ecs' cgroup
+func (agent *ecsAgent) setupPlatformResources() error {
+	if agent.cfg.TaskCPUMemLimit {
+		return setupTaskCgroupPrefix()
```
richardpen: Can you add a log here denotes the process is happening?

agent/app/agent.go
```diff
@@ -164,9 +167,12 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		log.Criticalf("Error creating state manager: %v", err)
 		return exitcodes.ExitTerminal
 	}
-	// TODO:
-	// 1) Feature gate ECS cgroup root init
-	// 2) Update mocks and unit tests
+
+	err = agent.setupPlatformResources()
```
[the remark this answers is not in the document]
vsiddharth: I think we need both `initPlatformResources()` and `setupPlatformResources()` as they both have well defined functionalities. This also helps ease testing quite a bit.

agent/app/agent_unix.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
+	"github.com/cihub/seelog"
+)
+
+// control is here to help create the cgroup root '/ecs' when
+// the feature is enabled. It also aids in unit testing
+var control cgroup.Control
```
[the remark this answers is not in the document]
vsiddharth: I prefer the current pattern as it clearly separates the functionalities. It is also consistent with the way similar things are done elsewhere within the agent code-base. 

`setupTaskCgroupPrefix()` is only going to be conditionally invoked.

agent/app/agent.go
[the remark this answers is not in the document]
samuelkarp: I think @petderek is saying that the distinction is not clear.  I agree with that comment; I don't think they have well-defined purposes yet.  Can you clarify the comments on those two functions?

agent/app/agent_unix.go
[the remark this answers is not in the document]
samuelkarp: Package-level variables are used to make it easier to test things without refactoring a bunch of old code.  I'd prefer not to use them when we're writing new code, if at all possible.

Not everything in this codebase is an example worth following.

agent/app/agent.go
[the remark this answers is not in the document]
aaithal: I'd even be in favor of merging the functionality of the two, if possible into just one method and retaining that.

agent/app/agent_unix.go
```diff
@@ -0,0 +1,57 @@
...
+func setupTaskCgroupPrefix() error {
+	if control.Exists(config.DefaultTaskCgroupPrefix) {
+		seelog.Debugf("Cgroup at %s already exists, skipping creation", config.DefaultTaskCgroupPrefix)
+		return nil
+	}
+	return control.Init()
+}
+
+// setControl sets the cgroup controller
+func setControl(controller cgroup.Control) {
```
aaithal: I'm really really really apprehensive about this pattern:
1. Using global variables even if it aids unit testing seems like an anti-pattern to me
2. This leads to all kinds of (potential) race conditions without being protected with a lock. 

Can you instead make `control` a member of `ecsAgent` and pass it around instead?

agent/app/agent_unix.go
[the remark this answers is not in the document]
vsiddharth: The only reason not to objectify this is because we did not want to wrap the cgroup within a platform specific object which was called out as an overkill.

agent/app/agent_unix.go
vsiddharth: `ecsAgent` is platform agnostic while `control` here is linux only. 
Embedding the `control` object within another was called out as an overkill.

aaithal: Would an interface help you here?

vsiddharth: I'm not sure if another interface is going to add significant value here. 
Also, if I eliminate the `init` items here, testing `doStart()` is going to be harder. 

I'm totally not in favor of using global objects. It just simplifies things here.
Could you please comment on what races you foresee with this approach ?

aaithal: > I'm not sure if another interface is going to add significant value here.
It will help you with not having to use this global variable

> Also, if I eliminate the `init` items here, testing `doStart()` is going to be harder. 
That should be fine. 

> Could you please comment on what races you foresee with this approach ?
As per my earlier comment, the access to `control` variable is not protected with a lock. Someone in the future will write a test where this inadvertently  gets invoked at the same time when `setupTaskCgroupPrefix` gets invoked.

vsiddharth: `control` is already a `linux only` interface as it does not apply to `windows`.

aaithal: So, you can have a dummy object in `ecsAgent`, which can remain uninitialized in windows.

vsiddharth: This is exactly the kind of problem we wanted to avoid in the first place. 

If we plan to embed this object into another or create dummies for unsupported platforms, we could have gone with a `platformResources` object.

vsiddharth:
Superseded by #984

PR #974, opened 2017-09-08
Task cgroup config

petderek:
### Summary

Agent will now fail when TaskCPUMemLimit is explicitly enabled.

### Implementation details

Agent is now aware of if the value was set by user or as a default. See change to config/types.go.

### Testing

### Licensing

### DIFF
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -14,12 +14,14 @@
 package app
 
 import (
+	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	"github.com/aws/amazon-ecs-agent/agent/ecscni"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerclient"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
 )
 
 const (
@@ -53,7 +55,7 @@ const (
 //    com.amazonaws.ecs.capability.task-iam-role-network-host
 //    ecs.capability.task-eni
 //    ecs.capability.task-eni-block-instance-metadata
-func (agent *ecsAgent) capabilities() []*ecs.Attribute {
+func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	var capabilities []*ecs.Attribute
 
 	if !agent.cfg.PrivilegedDisabled {
@@ -113,12 +115,17 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 		}
 	}
 
-	if agent.cfg.TaskCPUMemLimit {
+	// TODO we also need to disable this if agent already managing older tasks
+	if agent.cfg.TaskCPUMemLimit.Enabled() {
 		if _, ok := supportedVersions[dockerclient.Version_1_22]; ok {
 			capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+capabilityTaskCPUMemLimit)
+		} else if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+			// explicitly enabled -- return an error because we cannot fulfil an explicit request
+			return nil, errors.New("engine: Task CPU + Mem limit cannot be enabled due to unsupported Docker version")
 		} else {
-			seelog.Warn("Task CPU + Mem Limit disabled due to unsupported Docker version")
-			agent.cfg.TaskCPUMemLimit = false
+			// implicitly enabled -- don't register the capability, but degrade gracefully
+			seelog.Warn("Task CPU + Mem Limit disabled due to unsupported Docker version. API version 1.22 or greater is required.")
+			agent.cfg.TaskCPUMemLimit = config.ExplicitlyDisabled
 		}
 	}
 
@@ -131,7 +138,7 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 		})
 		taskENIVersionAttribute, err := agent.getTaskENIPluginVersionAttribute()
 		if err != nil {
-			return capabilities
+			return capabilities, nil
 		}
 		capabilities = append(capabilities, taskENIVersionAttribute)
 		// We only care about AWSVPCBlockInstanceMetdata if Task ENI is enabled
@@ -145,7 +152,7 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 
 	}
 
-	return capabilities
+	return capabilities, nil
 }
 
 // getTaskENIPluginVersionAttribute returns the version information of the ECS
--- a/agent/config/conditional.go
+++ b/agent/config/conditional.go
@@ -1,0 +1,64 @@
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package config
+
+import (
+	"encoding/json"
+	"strconv"
+)
+
+// Conditional makes it possible to understand if a variable was set explicitly or relies on a default setting
+type Conditional int
+
+const (
+	_ Conditional = iota
+	ExplicitlyEnabled
+	ExplicitlyDisabled
+	DefaultEnabled
+)
+
+// Enabled is a convenience function for when consumers don't care if the value is implicit or explicit
+func (b Conditional) Enabled() bool {
+	return b == ExplicitlyEnabled || b == DefaultEnabled
+}
+
+// MarshalJSON is used to serialize the type to json, per the Marshaller interface
+func (b Conditional) MarshalJSON() ([]byte, error) {
+	switch b {
+	case ExplicitlyEnabled:
+		return json.Marshal(true)
+	case ExplicitlyDisabled:
+		return json.Marshal(false)
+	default:
+		return json.Marshal(nil)
+	}
+}
+
+// UnmarshalJSON is used to deserialize json types into Conditional, per the Unmarshaller interface
+func (b *Conditional) UnmarshalJSON(jsonData []byte) error {
+	jsonString := string(jsonData)
+	jsonBool, err := strconv.ParseBool(jsonString)
+	if err != nil && jsonString != "null" {
+		return err
+	}
+
+	if jsonString == "" || jsonString == "null" {
+		*b = DefaultEnabled
+	} else if jsonBool {
+		*b = ExplicitlyEnabled
+	} else {
+		*b = ExplicitlyDisabled
+	}
+
+	return nil
+}
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -311,7 +311,19 @@ func environmentConfig() (Config, error) {
 	taskENIEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_ENI"), false)
 	taskIAMRoleEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE"), false)
 	taskIAMRoleEnabledForNetworkHost := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST"), false)
-	taskCPUMemLimitEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT"), true)
+
+	var taskCPUMemLimitEnabled Conditional
+	taskCPUMemLimitConfigString := os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT")
+
+	// We only want to set taskCPUMemLimit if it is explicitly set to true or false.
+	// We can do this by checking against the ParseBool default
+	if taskCPUMemLimitConfigString != "" {
+		if utils.ParseBool(taskCPUMemLimitConfigString, false) {
+			taskCPUMemLimitEnabled = ExplicitlyEnabled
+		} else {
+			taskCPUMemLimitEnabled = ExplicitlyDisabled
+		}
+	}
 
 	credentialsAuditLogFile := os.Getenv("ECS_AUDIT_LOGFILE")
 	credentialsAuditLogDisabled := utils.ParseBool(os.Getenv("ECS_AUDIT_LOGFILE_DISABLED"), false)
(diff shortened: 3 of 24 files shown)


samuelkarp:
Can you rebase this on top of `dev`?  The capability logic has changed and I'd rather not review it multiple times(once now and once after rebasing).

agent/config/types.go
```diff
@@ -110,7 +110,7 @@ type Config struct {
 	TaskIAMRoleEnabled bool
 
 	// TaskCPUMemLimit specifies if Agent can launch a task with a hierarchical cgroup
-	TaskCPUMemLimit bool
+	TaskCPUMemLimit DefaultBool
```
aaithal: Does setting this from config JSON work as well? I have a feeling that `DefaultBool` may need to implement the `Unmarshaler` interface: https://golang.org/pkg/encoding/json/#Unmarshaler

agent/config/types.go
```diff
@@ -210,3 +210,24 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
...
 	return nil
 }
+
+// DefaultBool makes it possible to understand if a variable was set explicitly or relies on a default setting
+type DefaultBool struct {
+	value    bool
+	explicit bool
+}
+
+func NewDefaultBool(setting, explicit bool) DefaultBool {
```
aaithal: lint: this and other global methods need documentation

agent/config/types.go
samuelkarp: Why use a new type and not just `*bool`?

agent/config/types.go
petderek: looking over this again... should I have this be a private function and expose constants?

IE, ExplicitlyEnabled, ImplicitlyEnabled, ExplicitlyDisabled...

agent/config/types.go
petderek: Using a `*bool` to effectively make a boolean have 3 states made me feel sad. I don't have a good technical reason. Is there a precedent?

petderek: Probably not! I'll fix this

agent/config/types.go
```diff
@@ -15,11 +15,20 @@ package config
...
 	"encoding/json"
+	"log"
+	"strconv"
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerclient"
 )
 
+const (
+	_ DefaultBool = iota
```
vsiddharth: `DefaultBool` doesn't capture the abstraction intuitively imho.

agent/config/types.go
```diff
@@ -210,3 +219,42 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
 	data.contents = json.RawMessage(jsonData)
 	return nil
 }
+
+// DefaultBool makes it possible to understand if a variable was set explicitly or relies on a default setting
+type DefaultBool int
```
vsiddharth: Could this name something along the lines of `TriBool` ?
Could you move this above the `iota` ?

agent/config/types.go
```diff
@@ -210,3 +219,42 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
...
+
+// Enabled is a convenience function for when consumers don't care if the value is implicit or explicit
+func (b DefaultBool) Enabled() bool {
+	return b == ExplicitlyEnabled || b == DefaultEnabled
+}
+
+func (b DefaultBool) MarshalJSON() ([]byte, error) {
```
vsiddharth: nit: Simple comments above

agent/config/types.go
```diff
@@ -210,3 +219,42 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
...
+	case ExplicitlyEnabled:
+		return json.Marshal(true)
+	case ExplicitlyDisabled:
+		return json.Marshal(false)
+	default:
+		return json.Marshal(nil)
+	}
+}
+
+func (b *DefaultBool) UnmarshalJSON(jsonData []byte) error {
```
vsiddharth: nit: simple comments above

agent/config/types.go
```diff
@@ -210,3 +219,42 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
...
+	jsonString := string(jsonData)
+	jsonBool, err := strconv.ParseBool(jsonString)
+
```
vsiddharth: This pattern appears to be a bit strange to me. 
* Why isn't the error check immediately performed ?
* Combining multiple checks within the switch is confusing ?

agent/config/types.go
```diff
@@ -210,3 +219,42 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
...
+	switch {
+	case jsonString == "" || jsonString == "null":
+		*b = DefaultEnabled
+	case err != nil:
+		return err
+	case jsonBool:
+		log.Print("exp en")
```
vsiddharth: nit: Is this intentional ?

agent/config/types.go
```diff
@@ -210,3 +219,42 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
...
+		*b = ExplicitlyEnabled
+	case !jsonBool:
+		log.Print("exp off")
```
vsiddharth: nit: is this intentional ?

agent/config/types.go
```diff
@@ -210,3 +219,42 @@ func (data *SensitiveRawMessage) UnmarshalJSON(jsonData []byte) error {
...
+		return json.Marshal(nil)
+	}
+}
+
+func (b *DefaultBool) UnmarshalJSON(jsonData []byte) error {
+	jsonString := string(jsonData)
+	jsonBool, err := strconv.ParseBool(jsonString)
+
+	switch {
+	case jsonString == "" || jsonString == "null":
```
vsiddharth: When would the `null` case happen?

petderek: `null` happens if the struct is serialized to json without `omitempty`. Note that this is the json `null` value and not a json string `"null"`.

I don't know if the first case `""` can happen in practice -- I'll double check

agent/config/types.go
petderek: Nope! I'll remove it

agent/config/types.go
petderek: Same -- oops

agent/config/types.go
petderek: > Why isn't the error check immediately performed ?

`ParseBool` returns an error for the `null` case, so there can be happy failure cases. But you're right, there is a more idiomatic way to express this.

> Combining multiple checks within the switch is confusing ?

[`switch` defaults to `switch(true)`. ](https://github.com/golang/go/wiki/Switch#missing-expression)  The idea was to put all of the conditions in one place. 

Given both of your comments, this may be better:

```golang

        // follow golang idiom closer with immediate nil check
        if err != nil && jsonString != "null" {
              return err
        }

        // I can swap this out with if / else 
	switch {
	case jsonString == "" || jsonString == "null":
		*b = DefaultEnabled
	case jsonBool:
		*b = ExplicitlyEnabled
        case !jsonBool:
		*b = ExplicitlyDisabled
	}
```

agent/config/types.go
petderek: How about moving the type and iota into their own file?

vsiddharth: I like the idea of making this more widely available

agent/config/types.go
vsiddharth: I prefer the proposed approach and the `if/else` over `switch`

agent/config/types.go
petderek: `DefaultBool` is a terrible and ambiguous name. I'm not sure we would want to name it `Tri*` that on the chance we decide to add fourth state `DefaultDisabled`. 

Is `Conditional` reasonable here?

vsiddharth: `Conditional` appears to be reasonable here 👍

agent/config/conditional.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"encoding/json"
+	"strconv"
+)
+
+// Conditional makes it possible to understand if a variable was set explicitly or relies on a default setting
+type Conditional int
+
+const (
+	_ Conditional = iota
```
samuelkarp: Does anything use this identifier?

agent/config/config.go
```diff
@@ -311,7 +311,18 @@ func environmentConfig() (Config, error) {
...
 	taskIAMRoleEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE"), false)
 	taskIAMRoleEnabledForNetworkHost := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST"), false)
-	taskCPUMemLimitEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT"), true)
+
+	var taskCPUMemLimitEnabled Conditional
+	taskCPUMemLimitConfigString := os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT")
+
+	switch {
+	case taskCPUMemLimitConfigString == "":
+		taskCPUMemLimitEnabled = DefaultEnabled
```
samuelkarp: This logic shouldn't be here.  If you see an explicit enable or explicit disable, use that.  Defaults should be in the existing defaults method though, and you should go through the merge for that.

agent/config/conditional.go
petderek: I wanted to avoid undesired/unexpected behavior when the iota is zero.

agent/config/config.go
petderek: Ah, good point!

agent/config/config.go
```diff
@@ -311,7 +311,16 @@ func environmentConfig() (Config, error) {
 	taskENIEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_ENI"), false)
 	taskIAMRoleEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE"), false)
 	taskIAMRoleEnabledForNetworkHost := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST"), false)
-	taskCPUMemLimitEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT"), true)
+
+	var taskCPUMemLimitEnabled Conditional
+	taskCPUMemLimitConfigString := os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT")
+
+	switch {
+	case utils.ParseBool(taskCPUMemLimitConfigString, true) && utils.ParseBool(taskCPUMemLimitConfigString, false):
```
vsiddharth: This is going to be confusing for folks without the context. 
A comment to explain the logic would be helpful imho. 

Also, isn't this check a bit excessive.

petderek: Yes, I over-thought this.  I'll make it more clear

agent/config/config.go
```diff
@@ -542,3 +551,7 @@ func (cfg *Config) String() string {
 		cfg.platformString(),
 	)
 }
+
```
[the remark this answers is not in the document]
petderek: 🤷‍♂️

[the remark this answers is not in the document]
vsiddharth: Oops, I meant it for the next line `555`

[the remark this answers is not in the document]
petderek: Oh. Yeah that shouldn't be there. removing it.

agent/app/agent_capability.go
```diff
@@ -113,12 +115,17 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
...
 		if _, ok := supportedVersions[dockerclient.Version_1_22]; ok {
 			capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+capabilityTaskCPUMemLimit)
+		} else if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+			// explicitly enabled -- return an error because we cannot fulfil an explicit request
+			return nil, errors.New("engine: Task CPU + Mem limit cannot be enabled due to unsupported Docker version")
 		} else {
-			seelog.Warn("Task CPU + Mem Limit disabled due to unsupported Docker version")
-			agent.cfg.TaskCPUMemLimit = false
+			// implicitly enabled -- don't register the capability, but degrade gracefully
+			seelog.Warn("Task CPU + Mem Limit disabled due to unsupported Docker version. Version 1.22 is required.")
```
samuelkarp: nit: "or greater"

Or maybe indicate which Docker version is the minimum rather than just the API?

agent/config/config_test.go
```diff
@@ -101,9 +101,8 @@ func TestEnvironmentConfig(t *testing.T) {
 	assert.True(t, conf.TaskIAMRoleEnabled, "Wrong value for TaskIAMRoleEnabled")
 	assert.True(t, conf.TaskIAMRoleEnabledForNetworkHost, "Wrong value for TaskIAMRoleEnabledForNetworkHost")
 	assert.True(t, conf.ImageCleanupDisabled, "Wrong value for ImageCleanupDisabled")
-	assert.True(t, conf.TaskCPUMemLimit, "Wrong value for TaskCPUMemLimit")
-	assert.True(t, conf.TaskENIEnabled, "Wrong value for TaskNetwork")
 
+	assert.True(t, conf.TaskENIEnabled, "Wrong value for TaskNetwork")
```
samuelkarp: What happened to the assertion on `conf.TaskCPUMemLimit`?

agent/config/config_unix_test.go
```diff
@@ -73,6 +73,7 @@ func TestConfigFromFile(t *testing.T) {
   "EngineAuthData": %s,
   "DataDir": "/var/run/ecs_agent",
   "TaskIAMRoleEnabled": true,
+  "TaskCPUMemLimit": true,
```
samuelkarp: (I can't comment on other lines because GitHub's diff is weird) Can you add an assertion to `TestConfigDefault` that verifies it's `DefaultEnabled`?

agent/config/config_test.go
petderek: It isn't set at this point because it isn't set until merged with the default.

agent/config/config_unix_test.go
petderek: Sure

agent/app/agent_capability.go
petderek: I think the API is more appropriate in this context

agent/config/config.go
```diff
@@ -311,7 +311,20 @@ func environmentConfig() (Config, error) {
 	taskENIEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_ENI"), false)
 	taskIAMRoleEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE"), false)
 	taskIAMRoleEnabledForNetworkHost := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST"), false)
-	taskCPUMemLimitEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT"), true)
+
+	var taskCPUMemLimitEnabled Conditional
+	taskCPUMemLimitConfigString := os.Getenv("ECS_ENABLE_TASK_CPU_MEM_LIMIT")
+
+	// We only want to set taskCPUMemLimit if it is explicitly set to true or false.
```
samuelkarp: I think it'd be a lot simpler to do this:

```go
if taskCPUMemLimitConfigString != "" {
	if utils.ParseBool(taskCPUMemLimitConfigString, false) {
		taskCPUMemLimitEnabled = ExplicitlyEnabled
	} else {
		taskCPUMemLimitEnabled = ExplicitlyDisabled
	}
}
```

petderek: I have no idea why I didn't see it that way in the first place.

agent/config/conditional.go
```diff
@@ -0,0 +1,64 @@
...
+	ExplicitlyEnabled
```
aaithal: How do I set a variable to be `ExplicitlyDisabled` from `ecs-init`?

PR #975, opened 2017-09-08
Pull retries and configurable routes for `awsvpc` network mode

samuelkarp:
### Summary
This series of commits adds retries for failed pulls (compensating for transient network issues) and allows additional routes to be configured for tasks launched with the `awsvpc` network mode.

### Implementation details
Retries are implemented with the new `utils.RetryNWithBackoffCtx` (based on existing `utils.RetryNWithBackoff`) inside the `dockerGoClient` and excludes one expected non-transient error (missing IAM permissions to call `ecr:GetAuthorizationToken`).  Other non-transient errors could be excluded from retries by extending `CannotPullContainerError` to implement `utils.Retrier`, but has not been done yet (I have not assembled a list of expected non-transient error codes from Docker).

Configurable routes were added by depending on the `types.IPNet` type from `github.com/containernetworking/cni/pkg/types`, which wraps `net.IPNet` so it can be serialized and deserialized in CIDR notation.

### Testing

Ran in an environment with manually configured additional routes and with transient network failures.

* Enhancement - Retry failed container image pull operations.

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## UNRELEASED
 * Feature - Support for provisioning Tasks with ENIs
+* Enhancement - Retry failed container image pull operations.
 * Bug - Fixed a memory leak issue when submitting the task state change [#967](https://github.com/aws/amazon-ecs-agent/pull/967)
 
 ## 1.14.4
--- a/README.md
+++ b/README.md
@@ -175,9 +175,10 @@ configure them as something other than the defaults.
 | `ECS_IMAGE_MINIMUM_CLEANUP_AGE` | 30m | The minimum time interval between when an image is pulled and when it can be considered for automated image cleanup. | 1h | 1h |
 | `ECS_NUM_IMAGES_DELETE_PER_CYCLE` | 5 | The maximum number of images to delete in a single automated image cleanup cycle. If set to less than 1, the value is ignored. | 5 | 5 |
 | `ECS_INSTANCE_ATTRIBUTES` | `{"stack": "prod"}` | These attributes take effect only during initial registration. After the agent has joined an ECS cluster, use the PutAttributes API action to add additional attributes. For more information, see [Amazon ECS Container Agent Configuration](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) in the Amazon ECS Developer Guide.| `{}` | `{}` |
-| `ECS_ENABLE_TASK_ENI` | `false` | Whether to enable task networking for task to be launched with its own network interface | `false` | `false` |
-| `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | `/amazon-ecs-cni-plugins` |
-| `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to [Instance Metdata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for Tasks started with `awsvpc` network mode | `false` | `false`|
+| `ECS_ENABLE_TASK_ENI` | `false` | Whether to enable task networking for task to be launched with its own network interface | `false` | Not applicable |
+| `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | Not applicable |
+| `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to [Instance Metdata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for Tasks started with `awsvpc` network mode | `false` | Not applicable |
+| `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES` | `["10.0.15.0/24"]` | In `awsvpc` network mode, traffic to these prefixes will be routed via the host bridge instead of the task ENI | `[]` | Not applicable |
 
 ### Persistence
 
--- a/agent/api/errors.go
+++ b/agent/api/errors.go
@@ -46,7 +46,8 @@ type NamedError interface {
 	ErrorName() string
 }
 
-// NamedError is a wrapper type for 'error' which adds an optional name and provides a symetric marshal/unmarshal
+// DefaultNamedError is a wrapper type for 'error' which adds an optional name and provides a symmetric
+// marshal/unmarshal
 type DefaultNamedError struct {
 	Err  string `json:"error"`
 	Name string `json:"name"`
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -218,7 +218,8 @@ func fileConfig() (Config, error) {
 
 	err = json.Unmarshal(data, &cfg)
 	if err != nil {
-		seelog.Errorf("Error reading cfg json data, err %v", err)
+		seelog.Criticalf("Error reading cfg json data, err %v", err)
+		return cfg, err
 	}
 
 	// Handle any deprecated keys correctly here
@@ -293,6 +294,7 @@ func environmentConfig() (Config, error) {
 	}
 
 	taskCleanupWaitDuration := parseEnvVariableDuration("ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION")
+
 	availableLoggingDriversEnv := os.Getenv("ECS_AVAILABLE_LOGGING_DRIVERS")
 	loggingDriverDecoder := json.NewDecoder(strings.NewReader(availableLoggingDriversEnv))
 	var availableLoggingDrivers []dockerclient.LoggingDriver
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -300,7 +300,7 @@ func environmentConfig() (Config, error) {
 	loggingDriverDecoder := json.NewDecoder(strings.NewReader(availableLoggingDriversEnv))
 	var availableLoggingDrivers []dockerclient.LoggingDriver
 	err = loggingDriverDecoder.Decode(&availableLoggingDrivers)
-	// EOF means the string was blank as opposed to UnexepctedEof which means an
+	// EOF means the string was blank as opposed to UnexpectedEof which means an
 	// invalid parse
 	// Blank is not a warning; we have sane defaults
 	if err != io.EOF && err != nil {
@@ -330,15 +330,15 @@ func environmentConfig() (Config, error) {
 	cniPluginsPath := os.Getenv("ECS_CNI_PLUGINS_PATH")
 	awsVPCBlockInstanceMetadata := utils.ParseBool(os.Getenv("ECS_AWSVPC_BLOCK_IMDS"), false)
 
-	instanceAttributesEnv := os.Getenv("ECS_INSTANCE_ATTRIBUTES")
-	attributeDecoder := json.NewDecoder(strings.NewReader(instanceAttributesEnv))
 	var instanceAttributes map[string]string
-
-	err = attributeDecoder.Decode(&instanceAttributes)
-	if err != io.EOF && err != nil {
-		err := fmt.Errorf("Invalid format for ECS_INSTANCE_ATTRIBUTES. Expected a json hash")
-		seelog.Warn(err)
-		errs = append(errs, err)
+	instanceAttributesEnv := os.Getenv("ECS_INSTANCE_ATTRIBUTES")
+	err = json.Unmarshal([]byte(instanceAttributesEnv), &instanceAttributes)
+	if instanceAttributesEnv != "" {
+		if err != nil {
+			wrappedErr := fmt.Errorf("Invalid format for ECS_INSTANCE_ATTRIBUTES. Expected a json hash: %v", err)
+			seelog.Error(wrappedErr)
+			errs = append(errs, wrappedErr)
+		}
 	}
 	for attributeKey, attributeValue := range instanceAttributes {
 		seelog.Debugf("Setting instance attribute %v: %v", attributeKey, attributeValue)
@@ -349,7 +349,7 @@ func environmentConfig() (Config, error) {
 	if additionalLocalRoutesEnv != "" {
 		err := json.Unmarshal([]byte(additionalLocalRoutesEnv), &additionalLocalRoutes)
 		if err != nil {
-			seelog.Warnf("Invalid format for ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES, expected a json array of CIDRs: %v", err)
+			seelog.Errorf("Invalid format for ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES, expected a json array of CIDRs: %v", err)
 			errs = append(errs, err)
 		}
 	}
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/cihub/seelog"
+	cnitypes "github.com/containernetworking/cni/pkg/types"
 )
 
 const (
@@ -343,6 +344,16 @@ func environmentConfig() (Config, error) {
 		seelog.Debugf("Setting instance attribute %v: %v", attributeKey, attributeValue)
 	}
 
+	var additionalLocalRoutes []cnitypes.IPNet
+	additionalLocalRoutesEnv := os.Getenv("ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES")
+	if additionalLocalRoutesEnv != "" {
+		err := json.Unmarshal([]byte(additionalLocalRoutesEnv), &additionalLocalRoutes)
+		if err != nil {
+			seelog.Warnf("Invalid format for ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES, expected a json array of CIDRs: %v", err)
+			errs = append(errs, err)
+		}
+	}
+
 	if len(errs) > 0 {
 		err = utils.NewMultiError(errs...)
 	} else {
@@ -381,6 +392,7 @@ func environmentConfig() (Config, error) {
 		InstanceAttributes:               instanceAttributes,
 		CNIPluginsPath:                   cniPluginsPath,
 		AWSVPCBlockInstanceMetdata:       awsVPCBlockInstanceMetadata,
+		AWSVPCAdditionalLocalRoutes:      additionalLocalRoutes,
 	}, err
 }
 
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -14,6 +14,7 @@
 package config
 
 import (
+	"encoding/json"
 	"errors"
 	"os"
 	"reflect"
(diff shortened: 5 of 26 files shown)


agent/engine/docker_container_engine.go
```diff
@@ -218,38 +225,63 @@ func (dg *dockerGoClient) time() ttime.Time {
 }
 
 func (dg *dockerGoClient) PullImage(image string, authData *api.RegistryAuthenticationData) DockerContainerMetadata {
+	// TODO Switch to just using context.WithDeadline and get rid of this funky code
 	timeout := dg.time().After(pullImageTimeout)
+	ctx, cancel := context.WithCancel(context.TODO())
 
 	response := make(chan DockerContainerMetadata, 1)
-	go func() { response <- dg.pullImage(image, authData) }()
+	go func() {
```
richardpen: Should we distinguish the errors generated by the agent from the errors generated by docker, and we only retry for errors from docker? For the errors generated by the agent like `CannotPullECRContainerError` and `CreateEmptyVolumeError` probably won't recover with retry.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## UNRELEASED
 * Feature - Support for provisioning Tasks with ENIs
+* Enhancement - Retry failed container image pull operations.
```
richardpen: This also added another feature that add customized route in the container namespace for task using `awsvpc`.  If I read the code correctly, this only can be configured by the config file, would it be more convenient to use environment variable. Also we may need to check the route before adding it, as the task traffic inside the container namespace can be changed by adding some route?

agent/engine/engine_mocks.go
```diff
@@ -17,6 +17,7 @@
 package engine
 
 import (
+	context0 "context"
```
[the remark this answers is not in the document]
samuelkarp: I'm using the `context` package in my code.

I have no idea why `golang.org/x/net/context` is still imported here...

agent/ecscni/plugin.go
```diff
@@ -24,16 +24,16 @@ import (
...
 	"github.com/pkg/errors"
 )
 
 // CNIClient defines the method of setting/cleaning up container namespace
 type CNIClient interface {
 	Version(string) (string, error)
 	Capabilities(string) ([]string, error)
-	SetupNS(*Config) error
-	CleanupNS(*Config) error
+	SetupNS(*Config, []cnitypes.IPNet) error
```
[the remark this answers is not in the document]
samuelkarp: I could do that if you feel strongly about it.

agent/engine/docker_container_engine.go
samuelkarp: `CannotPullECRContainerError` is already excluded from retries.  I can extend `CreateEmptyVolumeError` to exclude it from retries as well.  The other errors wrapped by `CannotPullContainerError` are not excluded right now; I only covered ECR because it was easy.

agent/engine/engine_mocks.go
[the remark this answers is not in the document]
richardpen: Looks like the `golang.org/x/net/context` is used by the `task_engine.Init()`, I think that's why the reason it's imported. I asked because I saw two context imported, we can't get rid of this until go docker client updated this.

CHANGELOG.md
samuelkarp: You're right; I had intended to make it configurable by environment variable as well but apparently never wrote the code.  I'll do that.

I don't think we need to check the route.  This is a sufficiently advanced, optional feature that is used in fairly narrow use-cases.

agent/engine/docker_task_engine.go
```diff
@@ -553,7 +553,23 @@ func (engine *DockerTaskEngine) pullAndUpdateContainerReference(task *api.Task,
...
+	var metadata DockerContainerMetadata
+	imagePullBackoff := utils.NewSimpleBackoff(250*time.Millisecond, time.Second, 0.20, 1.5)
+	_ = utils.RetryNWithBackoff(imagePullBackoff, 10, func() error {
+		metadata = engine.client.PullImage(container.Image, container.RegistryAuthentication)
+		if container.IsInternal() {
+			// "internal" containers are created by the agent and we don't expect to actually pull them
+			return nil
+		}
+		if imagePullError, ok := metadata.Error.(CannotPullContainerError); ok {
+			seelog.Warnf("When pulling %s: %s", container.Image, pullError)
```
nmeyerhans: `pullError` and `imagePullError` aren't supposed to be distinct names. Can we fix this commit before the PR is merged? Either name is fine to use.

(I realize that this code doesn't exist as written in the final PR, but I do want bisects, etc, to work and we shouldn't have unbuildable code in the repo.)

samuelkarp: Yeah, I'll fix this.

agent/engine/docker_task_engine.go
```diff
@@ -687,14 +687,21 @@ func (engine *DockerTaskEngine) cleanupPauseContainerNetwork(task *api.Task, con
...
+	return engine.cniClient.CleanupNS(cniConfig, engine.cfg.AWSVPCAdditionalLocalRoutes)
 }
 
 func (engine *DockerTaskEngine) buildCNIConfigFromTaskContainer(task *api.Task, container *api.Container) (*ecscni.Config, error) {
 	cfg, err := task.BuildCNIConfig()
 	if err != nil {
 		return nil, errors.Wrapf(err, "engine: build cni configuration from taskfailed")
 	}
+
+	if engine.cfg.OverrideAWSVPCLocalIPv4Address != nil &&
```
petderek: [nit]
consider moving this expression to a smaller function. ie, `hasValidAWSVPCLocalOverrides(cfg *Config)`

agent/utils/utils.go
```diff
@@ -109,11 +110,28 @@ var _time ttime.Time = &ttime.DefaultTime{}
...
+// retried
+func RetryWithBackoffCtx(ctx context.Context, backoff Backoff, fn func() error) error {
 	var err error
-	for err = fn(); true; err = fn() {
-		retriable, isRetriable := err.(Retriable)
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		default:
```
petderek: Neat!! 

Why keep the default block empty? Would it make sense to put `err = fn()`.. etc there?

agent/ecs_client/model/ecs/service.go
```diff
@@ -70,8 +70,8 @@ func New(p client.ConfigProvider, cfgs ...*aws.Config) *ECS {
 
 // newClient creates, initializes and returns a new service client instance.
 func newClient(cfg aws.Config, handlers request.Handlers, endpoint, signingRegion, signingName string) *ECS {
-	if signingName == "" {
-		signingName = ServiceName
+	if len(signingName) == 0 {
```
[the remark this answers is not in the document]
samuelkarp: So, I didn't make this change...this is what happened when I ran `make gogenerate`.

But the logic here is actually a little better; even though the zero-value of a string is `""`, this also would handle a `*string` where the zero-value is `nil` with no changes.

agent/engine/docker_task_engine.go
samuelkarp: Okay.

agent/utils/utils.go
samuelkarp: Seemed clearer to me to have it out.  This way it reads that we early return if the context is done, but otherwise keep going.

agent/engine/docker_task_engine.go
samuelkarp: I think I'm going to leave this as-is for now unless you feel strongly about it.

samuelkarp:
@richardpen @nmeyerhans I've addressed your comments, can you take a look?

agent/api/errors.go
```diff
@@ -46,7 +46,7 @@ type NamedError interface {
 	ErrorName() string
 }
 
-// NamedError is a wrapper type for 'error' which adds an optional name and provides a symetric marshal/unmarshal
+// NamedError is a wrapper type for 'error' which adds an optional name and provides a symmetric marshal/unmarshal
```
richardpen: Looks like this comment should be moved above to the `NamedError`.

agent/config/config.go
```diff
@@ -218,7 +219,8 @@ func fileConfig() (Config, error) {
 
 	err = json.Unmarshal(data, &cfg)
 	if err != nil {
-		seelog.Errorf("Error reading cfg json data, err %v", err)
+		seelog.Criticalf("Error reading cfg json data, err %v", err)
```
richardpen: Is it worth to logging the `string(data)` here?

agent/config/config.go
```diff
@@ -341,6 +344,16 @@ func environmentConfig() (Config, error) {
 		seelog.Debugf("Setting instance attribute %v: %v", attributeKey, attributeValue)
 	}
 
+	var additionalLocalRoutes []cnitypes.IPNet
+	additionalLocalRoutesEnv := os.Getenv("ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES")
+	if additionalLocalRoutesEnv != "" {
+		err := json.Unmarshal([]byte(additionalLocalRoutesEnv), &additionalLocalRoutes)
+		if err != nil {
+			seelog.Warnf("Invalid format for ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES, expected a json array of CIDRs: %v", err)
```
richardpen: same here, should we log the `additionalLocalRoutesEnv`?

agent/config/config_test.go
```diff
@@ -80,53 +81,75 @@ func TestBrokenEC2MetadataEndpoint(t *testing.T) {
...
 	os.Setenv("ECS_IMAGE_MINIMUM_CLEANUP_AGE", "30m")
+	defer os.Unsetenv("ECS_IMAGE_MINIMUM_CLEANUP_AGE")
 	os.Setenv("ECS_NUM_IMAGES_DELETE_PER_CYCLE", "2")
+	defer os.Unsetenv("ECS_NUM_IMAGES_DELETE_PER_CYCLE")
 	os.Setenv("ECS_INSTANCE_ATTRIBUTES", "{\"my_attribute\": \"testing\"}")
+	defer os.Unsetenv("ECS_INSTANCE_ATTRIBUTES")
 	os.Setenv("ECS_ENABLE_TASK_ENI", "true")
+	defer os.Unsetenv("ECS_ENABLE_TASK_ENI")
+	additionalLocalRoutesJSON := `["1.2.3.4/22","5.6.7.8/32"]`
+	os.Setenv("ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES", additionalLocalRoutesJSON)
```
richardpen: you missed `defer` here

agent/config/config_unix_test.go
```diff
@@ -135,21 +146,38 @@ func TestDockerAuthMergeFromFile(t *testing.T) {
...
-	assert.Equal(t, dockerAuthType, config.EngineAuthType, "docker auth type not as expected from file")
-	assert.Equal(t, dockerAuth, string(config.EngineAuthData.Contents()), "docker auth data not as expected from file")
-	assert.Equal(t, map[string]string{"attribute1": "value1"}, config.InstanceAttributes)
+	assert.Equal(t, cluster, cfg.Cluster, "cluster name not as expected from environment variable")
+	assert.Equal(t, dockerAuthType, cfg.EngineAuthType, "docker auth type not as expected from file")
+	assert.Equal(t, dockerAuth, string(cfg.EngineAuthData.Contents()), "docker auth data not as expected from file")
+	assert.Equal(t, map[string]string{"attribute1": "value1"}, cfg.InstanceAttributes)
+}
+
+// TestDockerAuthMergeFromFile tests docker auth read from file correctly after merge
```
richardpen: the comment is incorrect.

agent/config/config_windows_test.go
```diff
@@ -25,6 +25,8 @@ import (
 )
 
 func TestConfigDefault(t *testing.T) {
+	os.Setenv("AWS_DEFAULT_REGION", "foo-bar-1")
+	defer os.Unsetenv("AWS_DEFAULT")
```
richardpen: nit: `AWS_DEFAULT_REGION`

agent/config/config_windows_test.go
```diff
@@ -64,10 +66,12 @@ func TestConfigDefault(t *testing.T) {
 }
 
 func TestConfigIAMTaskRolesReserves80(t *testing.T) {
+	os.Setenv("AWS_DEFAULT_REGION", "foo-bar-1")
+	defer os.Unsetenv("AWS_DEFAULT")
```
richardpen: same as above.

agent/api/errors.go
samuelkarp: I'll change it to talking about `DefaultNamedError`, since that's the implementation that actually provides the marshal/unmarshal behavior.

agent/config/config.go
samuelkarp: Yeah, not a bad idea.

agent/config/config.go
samuelkarp: Yes, can do.

agent/config/config_test.go
samuelkarp: Oops!

agent/config/config_windows_test.go
samuelkarp: Oops!

agent/config/config_windows_test.go
samuelkarp: Oops!

agent/config/config_unix_test.go
samuelkarp: Oops!

samuelkarp:
> Also have you manually tested the route add part?

Yes, but in an earlier revision.  I will re-run my tests with the latest changes.

README.md
```diff
@@ -175,9 +175,10 @@ configure them as something other than the defaults.
...
-| `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | `/amazon-ecs-cni-plugins` |
-| `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to [Instance Metdata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for Tasks started with `awsvpc` network mode | `false` | `false`|
+| `ECS_ENABLE_TASK_ENI` | `false` | Whether to enable task networking for task to be launched with its own network interface | `false` | Not applicable |
+| `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | Not applicable |
+| `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to [Instance Metdata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for Tasks started with `awsvpc` network mode | `false` | Not applicable |
+| `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES` | `["10.0.15.0/24"]` | Additional routes that cause traffic to travel by the default ENI | `[]` | Not applicable |
```
nmeyerhans: Although the name strongly implies it, the description should be explicit that this is used in `awsvpc` mode.

It's also not entirely accurate to say that traffic to these routes goes out via the default ENI; it may never leave the host at all. I might rewrite the description as

> In `awsvpc` network mode, traffic to these prefixes will be routed via the host bridge instead of the task ENI

or something like that.

agent/config/config.go
nmeyerhans: Why is `warn` the right severity here?

agent/config/config.go
samuelkarp: After further thought, I'm going to leave this out for now.  I'm concerned about the case where a customer adds credentials (`EngineAuthData`) to the file and then we have a parse error; that would cause credentials to be written out to the log.

agent/config/config.go
samuelkarp: I'm also going to leave this out.  We don't log the read environment variable for any of the other ones that we parse.  I'd rather leave this consistent for now and then revisit it later when we can change all of them.

README.md
samuelkarp: I'll use your description.

agent/config/config.go
samuelkarp: We're using `warn` for all of them right now.  I'd be happy to change all of them to `error` if you'd like.

samuelkarp: Actually, we're using `warn` for both non-fatal conditions (actually should be warnings) and for parsing errors (should be errors).  I'll change the ones that actually cause the agent to exit to `error`.

nmeyerhans: Thanks, that's pretty much what I was getting at. If `NewConfig()` returns an error, agent panics. We should probably log the condition that cause the panic at an appropriately high severity.

samuelkarp:
@richardpen @nmeyerhans I believe I have addressed your comments, including additional manual testing.  Can you take a look?

agent/config/config.go
```diff
@@ -327,20 +330,30 @@ func environmentConfig() (Config, error) {
...
-	err = attributeDecoder.Decode(&instanceAttributes)
-	if err != io.EOF && err != nil {
-		err := fmt.Errorf("Invalid format for ECS_INSTANCE_ATTRIBUTES. Expected a json hash")
-		seelog.Warn(err)
-		errs = append(errs, err)
+	instanceAttributesEnv := os.Getenv("ECS_INSTANCE_ATTRIBUTES")
+	err = json.Unmarshal([]byte(instanceAttributesEnv), &instanceAttributes)
+	if instanceAttributesEnv != "" {
+		if err != nil {
+			err := fmt.Errorf("Invalid format for ECS_INSTANCE_ATTRIBUTES. Expected a json hash")
```
aaithal: Can we log/wrap the original error as well?

agent/ecscni/plugin_test.go
```diff
@@ -26,7 +27,11 @@ func TestSetupNS(t *testing.T) {
 		mockResult.EXPECT().String().Return(""),
 	)
 
-	err := ecscniClient.SetupNS(&Config{})
+	additionalRoutesJson := `["169.254.172.1/32", "10.11.12.13/32"]`
+	additionalRoutes := &[]cnitypes.IPNet{}
```
aaithal: nit: this reads weird to me. Can it be just `var AdditionalLocalRoutes []cnitypes.IPNet{}` here instead and you can unmarshal it as `&additionalRoutes`?

agent/utils/utils.go
```diff
@@ -109,11 +110,28 @@ var _time ttime.Time = &ttime.DefaultTime{}
...
+		}
 
-		if err == nil || isRetriable && !retriable.Retry() {
+		err = fn()
+
+		retriableErr, isRetriableErr := err.(Retriable)
+
+		if err == nil || isRetriableErr && !retriableErr.Retry() {
```
aaithal: Can you use parenthesis here to illustrate/enforce the operator precedence? Otherwise, readers like me will always have to look up the operator precedence documentation for how this is evaluated.

agent/config/types.go
```diff
@@ -167,6 +168,18 @@ type Config struct {
 	// AWSVPCBlockInstanceMetdata specifies if InstanceMetadata endpoint should be blocked
 	// for tasks that are launched with network mode "awsvpc" when ECS_AWSVPC_BLOCK_IMDS=true
 	AWSVPCBlockInstanceMetdata bool
+
+	// OverrideAWSVPCLocalIPv4Address overrides the local IPv4 address chosen
+	// for a task using the `awsvpc` networking mode. Using this configuration
+	// will limit you to running one `awsvpc` task at a time. IPv4 addresses
```
aaithal: > Using this configuration will limit you to running one `awsvpc` task

Since we don't enforce this anywhere in the code (iiuc), may be we should add a `TODO` so that we don't lose sight of this? Even an issue would be fine.

agent/utils/utils.go
samuelkarp: It's just short-circuiting evaluation; any time a left-condition evaluates false, nothing proceeds.

Are you asking me to change this to `err == nil || (isRetriableErr && !retriableErr.Retry())`?  That's logically equivalent, but (to me) harder to read.

agent/config/types.go
samuelkarp: Do you want a TODO even if we don't have any short-term plans to change it?  If you statically-assign an IP address, you'll always be limited to one or else you'll have IP address conflicts and things won't work right.

agent/utils/utils.go
aaithal: > Are you asking me to change this to `err == nil || (isRetriableErr && !retriableErr.Retry())`?  That's logically equivalent, but (to me) harder to read.

Yeah, I prefer that. Because, then I wouldn't have to wonder if `isRetriableErr` is associated with the `||` or `&&` operator.

agent/config/types.go
aaithal: Hmm, may be the better thing to do here is to update the description for the `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES ` field in the README file with some of what you have here as well?

samuelkarp: `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES` won't limit you to running one `awsvpc` task; it just adds routes.  It's specifically the static IP assignment that is limiting here.

aaithal: So, `OverrideAWSVPCLocalIPv4Address` gets set just from the JSON config and nowhere else? If that's the case, it's probably fine to leave it be. I'm just looking for ways to highlight this behavior so that it doesn't lead to post `start-task` failures.

samuelkarp:
@aaithal I've addressed your comments, except for the one about `OverrideAWSVPCLocalIPv4Address`.  Let me know if you want me to make any additional changes.

agent/config/types.go
samuelkarp: Exactly.  The intent here is that this is meant to serve a very limited set of use-cases and is not really a generally useful feature, so it's excluded from the more-common way to configure the agent (environment variables).

Issue #970, opened 2017-09-07
IP address not released when `docker stop` is called with awsvpc network mode


samuelkarp:
### Summary
The IP address allocated to a task is not released in the IPAM plugin's database when a task moves to stopped through an essential container exiting.

### Expected Behavior
`docker stop` of an essential container in a task should cause all resources to be cleaned up the same as if `aws ecs stop-task` is called.

### Observed Behavior
The IP address allocated to the task remains allocated, even after the task is cleaned up.  This can lead to exhaustion of the available IP addresses locally.

### Environment Details
Agent built from the `dev` branch running against a pre-production environment.

### Supporting Log Snippets
```
2017-09-07T17:08:43Z [ERROR] Set up pause container namespace failed, err: bridge ipam ADD: failed to execute plugin: ecs-ipam: getIPV4Address commands: failed to mark this ip 169.254.172.42/22 as used: Previous K/V pair exists, cannot complete Atomic operation, task: nginx:1 arn:aw
s:ecs:us-west-2:123456789012:task/f7803046-75b2-450d-833f-9ed57f3dc151, TaskStatus: (NONE->RUNNING) Containers: [nginx (NONE->RUNNING),~internal~ecs~pause (RUNNING->RESOURCES_PROVISIONED),]
```

samuelkarp:
Fixed in [1.14.5](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.14.5).

PR #976, opened 2017-09-11
Make the IPAM cleanup at the end of the task

richardpen:
### Summary

Fixing #970 

### Implementation details

Previously, the pause container namespace cleanup was invoked in `StopContainer` of task engine. This pr separate the cleanup into two process: clean up the network namespace and release the ip from ipam. Also this pr moves the ipam clean up out of the `StopContainer` to the end of a task life cycle. So that the `ipam del` command is always called when a task moved to stop.

### Testing

- [x] Manual test the scenario where the pause container is killed outside agent

New tests cover the changes: 
yes

### Licensing

yes

### DIFF
--- a/agent/ecscni/plugin.go
+++ b/agent/ecscni/plugin.go
@@ -34,6 +34,7 @@ type CNIClient interface {
 	Capabilities(string) ([]string, error)
 	SetupNS(*Config) error
 	CleanupNS(*Config) error
+	ReleaseIPResource(*Config) error
 }
 
 // cniClient is the client to call plugin and setup the network
@@ -67,15 +68,15 @@ func (client *cniClient) SetupNS(cfg *Config) error {
 		IfName:      defaultEthName,
 	}
 
-	netConfigList, err := client.constructNetworkConfig(cfg)
+	networkConfigList, err := client.createNetworkConfig(cfg, client.createBridgeConfigWithIPAM)
 	if err != nil {
-		return errors.Wrap(err, "cni invocation: Failed to construct network configuration for confirguring namespace")
+		return errors.Wrap(err, "cni invocation: Failed to construct network configuration for configuring namespace")
 	}
 
 	seelog.Debugf("Starting setup the ENI (%s) in container namespace: %s", cfg.ENIID, cfg.ContainerID)
 	os.Setenv("ECS_CNI_LOGLEVEL", logger.GetLevel())
 	defer os.Unsetenv("ECS_CNI_LOGLEVEL")
-	result, err := client.libcni.AddNetworkList(netConfigList, cns)
+	result, err := client.libcni.AddNetworkList(networkConfigList, cns)
 	if err != nil {
 		return err
 	}
@@ -84,8 +85,32 @@ func (client *cniClient) SetupNS(cfg *Config) error {
 	return nil
 }
 
+// ReleaseIPResource marks the ip available in the ipam db
+func (client *cniClient) ReleaseIPResource(cfg *Config) error {
+	cns := &libcni.RuntimeConf{
+		ContainerID: cfg.ContainerID,
+		NetNS:       fmt.Sprintf(netnsFormat, cfg.ContainerPID),
+		IfName:      defaultEthName,
+	}
+
+	ipamConfig, err := client.createIPAMConfig(cfg)
+	if err != nil {
+		return err
+	}
+
+	networkConfigList := &libcni.NetworkConfigList{
+		CNIVersion: client.cniVersion,
+		Plugins:    []*libcni.NetworkConfig{ipamConfig},
+	}
+
+	seelog.Debugf("Releaseing the ip resource from ipam db, id: [%s], ip: [%s]", cfg.ID, cfg.IPAMV4Address)
+	os.Setenv("ECS_CNI_LOGLEVEL", logger.GetLevel())
+	defer os.Unsetenv("ECS_CNI_LOGLEVEL")
+	return client.libcni.DelNetworkList(networkConfigList, cns)
+}
+
 // CleanupNS will clean up the container namespace, including remove the veth
-// pair, release the ip address in ipam
+// pair and stop the dhclient
 func (client *cniClient) CleanupNS(cfg *Config) error {
 	cns := &libcni.RuntimeConf{
 		ContainerID: cfg.ContainerID,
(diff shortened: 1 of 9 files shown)


agent/ecscni/plugin.go
```diff
@@ -135,6 +162,48 @@ func (client *cniClient) constructNetworkConfig(cfg *Config) (*libcni.NetworkCon
...
+		Network: &types.NetConf{
+			Type: ECSBridgePluginName,
+		},
+		Bytes: bridgeConfBytes,
+	}
+
+	return networkConfig, nil
+}
+
+func (client *cniClient) createBridgeConfig(cfg *Config) (*libcni.NetworkConfig, error) {
```
samuelkarp: It looks like there's a bunch of duplicate code between `createBridgeConfig` and `createBridgeConfigWithIPAM`.  Can we share instead of duplicating?

agent/ecscni/plugin.go
```diff
@@ -135,6 +162,48 @@ func (client *cniClient) constructNetworkConfig(cfg *Config) (*libcni.NetworkCon
...
+		Network: &types.NetConf{
+			Type: ECSBridgePluginName,
+		},
+		Bytes: bridgeConfBytes,
+	}
+
+	return networkConfig, nil
+}
+
+func (client *cniClient) createtENIConfig(cfg *Config) (*libcni.NetworkConfig, error) {
```
samuelkarp: s/createt/create/

agent/ecscni/plugin.go
```diff
@@ -145,37 +214,70 @@ func (client *cniClient) constructNetworkConfig(cfg *Config) (*libcni.NetworkCon
...
-		Bytes: eniConfBytes,
-	})
+		Bytes: ipamConfigBytes,
+	}
 
-	netconf := &libcni.NetworkConfigList{
+	return networkConfig, nil
+}
+
+// createAddNetworkConfig creates the configuration for invoking plugins
```
samuelkarp: nit: createNetworkConfig

agent/ecscni/plugin.go
```diff
@@ -145,37 +214,70 @@ func (client *cniClient) constructNetworkConfig(cfg *Config) (*libcni.NetworkCon
...
+func (client *cniClient) createNetworkConfig(cfg *Config, bridgeConfigFunc func(*Config) (*libcni.NetworkConfig, error)) (*libcni.NetworkConfigList, error) {
```
samuelkarp: It looks like this function is the entrypoint for both `createBridgeConfig` and `createBridgeConfigWithENI`.  Can you move it above those functions?

agent/ecscni/plugin.go
```diff
@@ -145,37 +214,70 @@ func (client *cniClient) constructNetworkConfig(cfg *Config) (*libcni.NetworkCon
...
+		Name:       ECSIPAMPluginName,
+		CNIVersion: client.cniVersion,
+		IPAM:       ipam,
+	}
+
+	ipamConfigBytes, err := json.Marshal(ipamConfig)
 	if err != nil {
-		seelog.Errorf("Marshal eni configuration error: %v", err)
-		return nil, err
+		return nil, errors.Wrapf(err, "Marshal ipam plugin configuration failed")
```
aaithal: nit: error message should begin with lower case letter. please fix this everywhere.

agent/engine/docker_container_engine.go
```diff
@@ -39,6 +40,7 @@ import (
 
 const (
 	dockerDefaultTag = "latest"
+	ImageNameFormat  = "%s:%s"
```
aaithal: lint: `ImageNameFormat ...` Or better yet, make it local scoped.

agent/engine/task_manager.go
```diff
@@ -323,6 +324,19 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
+// releaseIPInIPAM releases the ip used by the task for awsvpc
+func (mtask *managedTask) releaseIPInIPAM() {
+	if mtask.ENI == nil {
+		return
+	}
+	seelog.Infof("Releasing ip in IPAM for task used eni, task: %s", mtask.Task.Arn)
+
+	err := mtask.engine.releaseIPInIPAM(mtask.Task)
+	if err != nil {
+		seelog.Infof("Releasing the ip in IPAM failed, task: [%s], err: %v", mtask.Task.Arn, err)
```
aaithal: Why is this `Infof`? And, why is it ok to consume this error?

richardpen: For this kind of error, it's fine because there will be  more available ips than the task, so if one ip was failed to release the next task can still get an available ip. But in rare condition all the release failed, and all the ip has been used then no task can be started, that's why I marked it as `Infof`, any better suggestion here?

aaithal: So, it's probably fine that we consume this error and move on. I'd still like this to be logged as a warning.

richardpen:
@aaithal @samuelkarp I have addressed your comments, please take another review, thanks.

agent/engine/docker_container_engine.go
```diff
@@ -271,14 +274,19 @@ func (dg *dockerGoClient) pullImage(image string, authData *api.RegistryAuthenti
...
-	if image == emptyvolume.Image+":"+emptyvolume.Tag {
+	if image == fmt.Sprintf(imageNameFormat, emptyvolume.Image, emptyvolume.Tag) {
 		scratchErr := dg.createScratchImageIfNotExists()
 		if scratchErr != nil {
 			return CreateEmptyVolumeError{scratchErr}
 		}
 		return nil
 	}
 
+	// Skip pull pause container image, it's already loaded on agent restart
```
samuelkarp: Can you remove this logic for now?  I think I'd rather have this occur at a different layer in the codebase (probably in DockerTaskEngine rather than here, so we can trigger on the container type rather than the image name) and it's orthogonal to the rest of the PR.  I was planning on looking at this on Monday.

agent/engine/task_manager.go
```diff
@@ -151,6 +151,7 @@ func (mtask *managedTask) overseeTask() {
 		llog.Debug("Marking done for this sequence", "seqnum", mtask.StopSequenceNumber)
 		mtask.engine.taskStopGroup.Done(mtask.StopSequenceNumber)
 	}
+	go mtask.releaseIPInIPAM()
```
samuelkarp: Is this idempotent?  You may end up with multiple calls/partial completion if there's an agent restart in the middle.

agent/ecscni/plugin.go
```diff
@@ -84,62 +85,127 @@ func (client *cniClient) SetupNS(cfg *Config) error {
 	return nil
 }
 
+// ReleaseIPResource marks the ip available in the ipam db
+func (client *cniClient) ReleaseIPResource(cfg *Config) error {
```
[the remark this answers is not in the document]
richardpen: No, just randomly placed, I will fix that.

agent/engine/task_manager.go
richardpen: This will ensure the ip is removed if agent restarted, but could emit an error in the ipam logs if  this was called twice before and after restart. One possible solution would be remove the eniattachment from state here and check the eniattachment before calling releasing. But there is also a race condition where restart happened in the middle of saving state. Any better suggestion, otherwise I will just add a TODO in the code and we can come back to that later.

agent/engine/docker_container_engine.go
richardpen: removed

agent/engine/task_manager.go
samuelkarp: If the only downside is an error in the IPAM logs, I'm okay with adding a TODO.

richardpen:
@samuelkarp I have addressed all your comments, merge this now. Feel free to continue comment on this if you have more comments.

agent/engine/engine_integ_test.go
```diff
@@ -89,6 +89,7 @@ func setup(cfg *config.Config, t *testing.T) (TaskEngine, func(), credentials.Ma
 	}, credentialsManager
 }
 
+/*
```
samuelkarp: Why was this commented out?

agent/engine/docker_container_engine.go
```diff
@@ -271,7 +274,7 @@ func (dg *dockerGoClient) pullImage(image string, authData *api.RegistryAuthenti
 
 	// Special case; this image is not one that should be pulled, but rather
 	// should be created locally if necessary
-	if image == emptyvolume.Image+":"+emptyvolume.Tag {
+	if image == fmt.Sprintf(imageNameFormat, emptyvolume.Image, emptyvolume.Tag) {
```
samuelkarp: Why was this changed?

agent/engine/engine_integ_test.go
richardpen: This part should be removed when we merge the enis branch, as it is moved to `test_common.go`,
 see [here](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/engine/test_common.go#L26), but l forgot to remove this part, I will remove this part later.

agent/engine/docker_container_engine.go
richardpen: I thought this was better the previous way of doing this, and we will use this format for more internal image names.

Issue #978, opened 2017-09-11
Makefile can't clean up after itself


nmeyerhans:
### Summary

`make clean` fails after `make release` due to permission failures.

### Description

The `docker-release` make target (or its dependencies) runs processes as root in a container that write to subdirectories of the source directory via a mounted volume. Thus the subdirectories and files that it creates are owned by root and can't be cleaned up by `make clean`

### Expected Behavior

`make clean` deletes files created by `make release`

### Observed Behavior

Following `make release`, `make clean` fails as follows:

```
$ make release
...
Successfully built 4b5485b82cc6
Built Docker image "amazon/amazon-ecs-agent:latest"
$ make clean
# ensure docker is running and we can talk to it, abort if not:
docker ps > /dev/null
rm -f misc/certs/ca-certificates.crt &> /dev/null
rm -rf out/*
rm: cannot remove 'out/cni-plugins/ecs-eni': Permission denied
rm: cannot remove 'out/cni-plugins/ecs-ipam': Permission denied
rm: cannot remove 'out/cni-plugins/ecs-bridge': Permission denied
Makefile:169: recipe for target 'clean' failed
make: *** [clean] Error 1

```

jhaynes:
This change was release with [Agent v1.14.5](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.14.5)

Issue #977, opened 2017-09-11
Makefile default rule doesn't work as documented due to undeclared dependencies


nmeyerhans:
### Summary

Default `make` rule does not declare adequate dependencies. The `release` makefile target appears to be missing a dependency on `get-cni-sources`.

### Description

README.md indicates that an agent container image can be built with `make`, but the default makefile rule is missing dependencies and the build fails.

### Expected Behavior

`make` should generate an `amazon/ecs-container-agent:make` docker image, per the README.

### Observed Behavior

`make` results in a build failure:

```
mkdir -p "/home/local/ANT/nmeyerha/src/go/src/github.com/aws/amazon-ecs-agent/out"
Sending build context to Docker daemon 57.63 MB
Step 1 : FROM golang:1.7
 ---> f69c27b2f59a
Step 2 : MAINTAINER Amazon Web Services, Inc.
 ---> Using cache
 ---> 79102635ad24
Step 3 : RUN mkdir -p /go/src/github.com/aws/
 ---> Using cache
 ---> cee2a83cb6ed
Step 4 : WORKDIR /go/src/github.com/aws/amazon-ecs-cni-plugins
 ---> Using cache
 ---> baec03feb959
Step 5 : CMD make plugins
 ---> Using cache
 ---> 650612a2d7e1
Successfully built 650612a2d7e1
make: *** No rule to make target 'plugins'.  Stop.
Makefile:126: recipe for target 'cni-plugins' failed
make: *** [cni-plugins] Error 2
```

jhaynes:
This change was release with [Agent v1.14.5](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.14.5)

PR #980, opened 2017-09-12
Makefile fixes

nmeyerhans:
### Summary

Makefile improvements to address #977 and #978 

### Implementation details

General makefile cleanup to provide better handling of dependencies.

### Testing

Multiple runs of `make clean`, `make get-deps`, etc. Confirm that subsequent runs don't do anything, e.g. 

```
ulap:amazon-ecs-agent‹makefile-fixes›$ make get-deps    
if [ -d amazon-ecs-cni-plugins/plugins ]; then \
        cd amazon-ecs-cni-plugins && git fetch && git checkout master ; \
else \
        git clone https://github.com/aws/amazon-ecs-cni-plugins.git --branch master ; \
fi
Cloning into 'amazon-ecs-cni-plugins'...
remote: Counting objects: 1867, done.
remote: Total 1867 (delta 0), reused 0 (delta 0), pack-reused 1867
Receiving objects: 100% (1867/1867), 2.91 MiB | 1.88 MiB/s, done.
Resolving deltas: 100% (808/808), done.
Checking connectivity... done.
mkdir -p amazon-ecs-cni-plugins/bin/plugins
touch .cni-sources-stamp
go get github.com/tools/godep
go get golang.org/x/tools/cmd/cover
go get github.com/golang/mock/mockgen
go get golang.org/x/tools/cmd/goimports
touch .get-deps-stamp
ulap:amazon-ecs-agent‹makefile-fixes›$ make get-deps  
make: Nothing to be done for 'get-deps'.
```

New tests cover the changes: There are no specific tests for the makefile.

This change doesn't introduce any behavior changes into the agent and can be left out of the changelog.

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,16 +28,16 @@ all: docker
 
 # Dynamic go build; useful in that it does not have -a so it won't recompile
 # everything every time
-gobuild: get-deps
+gobuild:
 	./scripts/build false
 
 # Basic go build
-static: get-deps
+static:
 	./scripts/build
 
 # 'build-in-docker' builds the agent within a dockerfile and saves it to the ./out
 # directory
-build-in-docker: get-deps
+build-in-docker:
 	@docker build -f scripts/dockerfiles/Dockerfile.build -t "amazon/amazon-ecs-agent-build:make" .
 	@docker run --net=none \
 	  -e TARGET_OS="${TARGET_OS}" \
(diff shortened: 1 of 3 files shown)


samuelkarp:
So I think I'm not really convinced that I want all of these dependencies grouped together into the same target.  Things like the `go get` commands really shouldn't ever be automatically run (they're not necessary for building and only for development) and *should* be run on-demand.  On the other hand, the `get-cni-sources` target *should* be run automatically and for builds, and shouldn't be unnecessarily re-run.

What do you think about making these just separate targets entirely?  You can have target dependencies on `get-cni-sources` instead of `get-deps` and remove the stamping for `get-deps`.

nmeyerhans:
@samuelkarp Yes, I think I agree with you. We should download the CNI sources as a dependency for our agent targets.

I wonder also if we should avoid deleting the CNI sources in the `clean` target. That target can delete any actual build but leave the sources in place. We could add an additional `distclean` target that goes further to reset the source tree to a pristine state, including blowing away the CNI sources.

samuelkarp:
`get-cni-sources` seems like a poorly implemented copy of something like a git submodule, so it might just be better to use a real submodule.

If I remember correctly, we did this because the CNI plugin sources were moving very quickly as they were under active development.  Things are mostly stable now, so I think the benefit of using a system that's designed for this (embedding one git repo inside another) and tightly controls the version of the thing that's there outweighs the annoyance of changing the revision.  One of the Go vendoring tools might also work here, but we'll have to figure out how to get it to work in a nice way because this isn't a typical use-case for those tools.

nmeyerhans:
Added another commit that converts the `amazon-ecs-cni-plugins` to a submodule.

Makefile
```diff
@@ -97,83 +97,88 @@ test-registry: netkitten volumes-test squid awscli image-cleanup-test-images flu
...
 	rm -rf out/*
-	rm -rf $(shell pwd)/amazon-ecs-cni-plugins
+	$(MAKE) -C $(ECS_CNI_REPOSITORY_SRC_DIR) clean
 	rm -rf agent/Godeps/_workspace/pkg/
 	-$(MAKE) -C misc/netkitten $(MFLAGS) clean
 	-$(MAKE) -C misc/volumes-test $(MFLAGS) clean
 	-$(MAKE) -C misc/gremlin $(MFLAGS) clean
 	-$(MAKE) -C misc/testnnp $(MFLAGS) clean
 	-$(MAKE) -C misc/image-cleanup-test-images $(MFLAGS) clean
+	-rm -f .get-deps-stamp .cni-sources-stamp
```
samuelkarp: `.cni-sources-stamp` doesn't exist anywhere.

nmeyerhans: Fixed, thanks.

PR #981, opened 2017-09-12
metadata service continued

adnxn:
### Summary

carried over work from #913 with following additions:
* fix mismatch of generated mocks between metadata service and dev branch
* includes dev changes up to 6f11735d92bbf1ce7bbe099721a17384095fdff7
* updated gomock to latest version and regenerated all mocks
* added code to handle agent restarts and unit test
* functional testing for unix

### Implementation details
n/a

### Testing

New tests cover the changes: 

### Licensing


README.md
```diff
@@ -175,10 +175,16 @@ configure them as something other than the defaults.
 | `ECS_IMAGE_MINIMUM_CLEANUP_AGE` | 30m | The minimum time interval between when an image is pulled and when it can be considered for automated image cleanup. | 1h | 1h |
 | `ECS_NUM_IMAGES_DELETE_PER_CYCLE` | 5 | The maximum number of images to delete in a single automated image cleanup cycle. If set to less than 1, the value is ignored. | 5 | 5 |
 | `ECS_INSTANCE_ATTRIBUTES` | `{"stack": "prod"}` | These attributes take effect only during initial registration. After the agent has joined an ECS cluster, use the PutAttributes API action to add additional attributes. For more information, see [Amazon ECS Container Agent Configuration](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html) in the Amazon ECS Developer Guide.| `{}` | `{}` |
+<<<<<<< 6f11735d92bbf1ce7bbe099721a17384095fdff7
```
aaithal: Seems like a remnant of merge resolution. please get rid of this.

agent/config/config.go
```diff
@@ -393,6 +395,8 @@ func environmentConfig() (Config, error) {
 		CNIPluginsPath:                   cniPluginsPath,
 		AWSVPCBlockInstanceMetdata:       awsVPCBlockInstanceMetadata,
 		AWSVPCAdditionalLocalRoutes:      additionalLocalRoutes,
+		ContainerMetadataEnabled:         containerMetadataEnabled,
```
aaithal: Can you also update `DefaultConfig` to set `ContainerMetadataEnabled` to `false`?

agent/config/config.go
```diff
@@ -353,6 +353,8 @@ func environmentConfig() (Config, error) {
 			errs = append(errs, err)
 		}
 	}
+	containerMetadataEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_CONTAINER_METADATA"), false)
+	dataDirOnHost := os.Getenv("ECS_HOST_DATA_DIR")
```
aaithal: What happens when ECS Init/whatever's starting the Agent mounts some other directory as Agent's data directory (`-v /tmp:/data`) sets `ECS_HOST_DATA_DIR` to `"/var/lib/ecs"`

It doesn't seem like a strong enough abstraction to be dependent on Agent configuration options to expect `ECS_HOST_DATA_DIR` to be the same as whatever's being mounted as `/data` on the host. My main concern is the late-binding/asynchronous failures this could lead to if the wrong host mount was specified. Wondering if there are better alternatives here. Can we at the very least inspect the Agent container and check if `ECS_HOST_DATA_DIR` is mounted?

agent/containermetadata/types.go
```diff
@@ -0,0 +1,114 @@
...
+// The problems described above are indications engine.DockerClient needs to be moved
+// outside the engine package
+type DockerMetadataClient interface {
+	InspectContainer(string, time.Duration) (*docker.Container, error)
+}
+
+// Network is a struct that keeps track of metadata of a network interface
+type Network struct {
+	NetworkMode string `json:"NetworkMode,omitempty"`
+	IPv4Address string `json:"IPv4Address,omitempty"`
```
aaithal: Can you change this to `IPv4Addresses` instead? For now it can be a list of length 1. It'd be helpful to surface this info when we start containers with multiple IP addresses.

agent/containermetadata/types.go
```diff
@@ -0,0 +1,114 @@
...
+	IPv6Address string `json:"IPv6Address,omitempty"`
```
aaithal: Same as above. Can you change this to `IPv6Addresses` instead (`[]string`)?

agent/functional_tests/testdata/taskdefinitions/mdservice-validator-unix/task-definition.json
```diff
@@ -0,0 +1,10 @@
+{
+  "family": "ecsftest-mdservice-validator",
+  "containerDefinitions": [{
+    "image": "127.0.0.1:51670/busybox:latest",
+    "name": "mdservice-validator-unix",
+    "cpu": 10,
+    "memory": 10,
+    "command": ["sh", "-c", "sleep 1 && if cat $ECS_CONTAINER_METADATA_FILE | grep \"READY\"; then exit 42; else exit 1; fi"]
```
aaithal: `sleep 1` may not always be sufficient. Can you increase this duration to something like `10s`? Or, you can check for the existence in a loop and poll on the file for a maximum of `1m`.

agent/containermetadata/utils_test.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// TestGetTaskIDFailDueToInvalidID checks the case where task ARN has 6 parts but last section is invalid
+func TestGetTaskIDFailDueToInvalidID(t *testing.T) {
+	mockTaskARN := invalidSplitTaskARN
+
+	_, err := getTaskIDfromARN(mockTaskARN)
+	expectErrorMessage := fmt.Sprintf("get task ARN: cannot find TaskID for TaskARN %s", mockTaskARN)
+
+	if err.Error() != expectErrorMessage {
```
richardpen: Please use `assert` instead, same for other packages.

agent/engine/default.go
```diff
@@ -26,6 +27,6 @@ import (
 var log = logger.ForModule("TaskEngine")
 
 // NewTaskEngine returns a default TaskEngine
-func NewTaskEngine(cfg *config.Config, client DockerClient, credentialsManager credentials.Manager, containerChangeEventStream *eventstream.EventStream, imageManager ImageManager, state dockerstate.TaskEngineState) TaskEngine {
-	return NewDockerTaskEngine(cfg, client, credentialsManager, containerChangeEventStream, imageManager, state)
+func NewTaskEngine(cfg *config.Config, client DockerClient, credentialsManager credentials.Manager, containerChangeEventStream *eventstream.EventStream, imageManager ImageManager, state dockerstate.TaskEngineState, metadataManager containermetadata.Manager) TaskEngine {
```
richardpen: Can you break this into multiple lines.

agent/engine/docker_task_engine.go
```diff
@@ -92,13 +93,14 @@ type DockerTaskEngine struct {
...
 	containerStatusToTransitionFunction map[api.ContainerStatus]transitionApplyFunc
+	metadataManager                     containermetadata.Manager
 }
 
 // NewDockerTaskEngine returns a created, but uninitialized, DockerTaskEngine.
 // The distinction between created and initialized is that when created it may
 // be serialized/deserialized, but it will not communicate with docker until it
 // is also initialized.
-func NewDockerTaskEngine(cfg *config.Config, client DockerClient, credentialsManager credentials.Manager, containerChangeEventStream *eventstream.EventStream, imageManager ImageManager, state dockerstate.TaskEngineState) *DockerTaskEngine {
+func NewDockerTaskEngine(cfg *config.Config, client DockerClient, credentialsManager credentials.Manager, containerChangeEventStream *eventstream.EventStream, imageManager ImageManager, state dockerstate.TaskEngineState, metadataManager containermetadata.Manager) *DockerTaskEngine {
```
richardpen: please break this into multiple lines

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,170 @@
...
+)
+
+// MetadataStatus specifies the current update status of the metadata file.
+// The purpose of this status is for users to check if the metadata file has
+// reached the stage they need before they read the rest of the file to avoid
+// race conditions (Since the final stage will need to be after the container
+// starts up
+// In the future the metadata may require multiple stages of update and these
+// statuses should amended/appended accordingly.
+type MetadataStatus string
```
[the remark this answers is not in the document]
huerdong: I believe this was a section I wrote and the reason I used a string rather than enum was to display the actual name of the status in the JSON Marshaling. I believe using an enum would produce numerical codes in the Marshaled output. which I think is less helpful to users checking for this status.

agent/containermetadata/manager_test.go
```diff
@@ -0,0 +1,187 @@
...
+	newManager := &metadataManager{}
+	err := newManager.Clean(mockTaskARN)
+	expectErrorMessage := fmt.Sprintf("clean task %s: get task ARN: invalid TaskARN invalidARN", mockTaskARN)
+
+	if err.Error() != expectErrorMessage {
+		t.Error("Got unexpected error: " + err.Error())
+	}
+}
+
+// TestClean is the mainline case for metadata create
```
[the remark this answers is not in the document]
huerdong: As with the comment below, I wrote this comment intending it to mean "TestClean is the happy path case for metadata file cleanup".

agent/containermetadata/manager_unix_test.go
```diff
@@ -0,0 +1,96 @@
...
+package containermetadata
+
+import (
+	"testing"
+
+	docker "github.com/fsouza/go-dockerclient"
+	"github.com/golang/mock/gomock"
+)
+
+// TestCreate is the mainline case for metadata create
```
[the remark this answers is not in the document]
huerdong: Same as above

agent/containermetadata/manager_test.go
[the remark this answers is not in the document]
richardpen: Then we may want these comments to be more specific to the test, as the current comment doesn't have useful and clear meaning for the test.

agent/config/config.go
huerdong: I think we can do away with `ECS_HOST_DATA_DIR` if we directly inspect the Agent container for whichever directory is mounted as `/data`.

adnxn: hm you're right, there's a failure case here. i agree it makes sense to validate ECS_HOST_DATA_DIR is mounted by inspecting the Agent container.

however, i'm not sure if it makes more sense to fail during agent creation with a configuration error since the user will be expecting the metadata feature to be enabled, or if it would make more sense for us to surface a warning log and just disable the metadata feature.

i think failing explicitly during agent creation, `newAgent(...)`, makes more sense since this would be a configuration error. whereas logging, disabling the feature and moving forward seems like a poor experience since the only ways for the user to find out about these errors would be to catch it in the log, or through containers failing because they expect a metadata file.

README.md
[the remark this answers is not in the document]
adnxn: fixed.

agent/containermetadata/manager_test.go
```diff
@@ -0,0 +1,187 @@
...
+// TestSetContainerInstanceARN checks whether the container instance ARN is set correctly.
+func TestSetContainerInstanceARN(t *testing.T) {
+	_, _, _, _, done := managerSetup(t)
+	defer done()
+
+	mockARN := containerInstanceARN
+
+	newManager := &metadataManager{}
+	newManager.SetContainerInstanceARN(mockARN)
+	if newManager.containerInstanceARN != mockARN {
```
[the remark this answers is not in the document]
adnxn: there are a few these i hadn't noticed. fixed.

agent/config/config_test.go
```diff
@@ -120,6 +120,10 @@ func TestEnvironmentConfig(t *testing.T) {
 	additionalLocalRoutesJSON := `["1.2.3.4/22","5.6.7.8/32"]`
 	os.Setenv("ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES", additionalLocalRoutesJSON)
 	defer os.Unsetenv("ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES")
+	os.Setenv("ECS_ENABLE_CONTAINER_METADATA", "true")
```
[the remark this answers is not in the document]
adnxn: fixed. added check below.

agent/engine/docker_task_engine.go
adnxn: fixed.

agent/engine/default.go
adnxn: fixed.

README.md
```diff
@@ -175,10 +175,16 @@ configure them as something other than the defaults.
...
 | `ECS_ENABLE_TASK_ENI` | `false` | Whether to enable task networking for task to be launched with its own network interface | `false` | Not applicable |
 | `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | Not applicable |
 | `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to [Instance Metdata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for Tasks started with `awsvpc` network mode | `false` | Not applicable |
 | `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES` | `["10.0.15.0/24"]` | In `awsvpc` network mode, traffic to these prefixes will be routed via the host bridge instead of the task ENI | `[]` | Not applicable |
+| `ECS_ENABLE_TASK_ENI` | `false` | Whether to enable task networking for task to be launched with its own network interface | `false` | `false` |
```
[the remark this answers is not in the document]
adnxn: it made it on to aws:dev from 9241ed4894aa829d16fdb8b158ea2ad2bee4ab85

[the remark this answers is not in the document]
adnxn: ugh it's related to your last comment, left over from merge resolution. fixed.

agent/config/config.go
aaithal: > i think failing explicitly during agent creation, newAgent(...), makes more sense since this would be a configuration error.

I agree

adnxn: @aaithal so as per our discussion earlier, we will call this out in the README under usage instructions for the metadata feature.

agent/containermetadata/manager.go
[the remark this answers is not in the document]
samuelkarp: You can still have status names in marshaled JSON output with iotas (Go doesn't actually have enums), but you'd need to implement the [`TextMarshaler`](https://golang.org/pkg/encoding/#TextMarshaler) and [`TextUnmarshaler`](https://golang.org/pkg/encoding/#TextUnmarshaler) interfaces.

I would recommend doing this, as you then get stronger assurances about what will be marshaled/unmarshaled.

agent/config/config.go
samuelkarp: There are a few assumptions here that I think are problematic:

* Assuming that `ECS_HOST_DATA_DIR` is `/var/lib/ecs` on the host
* Assuming that the ECS agent is running inside a container (it runs *outside* a container on Windows)
* Assuming that the ECS agent container has the name `ecs-agent` so you can inspect the container (true on ECS-optimized AMIs and other places where `ecs-init` is used, but not elsewhere)
* Assuming that you can get the ECS agent container ID through its hostname so you can inspect the container (won't be true if the UTS namespace is shared)

huerdong: Adding to samuelkarp's second point, I believe the metadata feature does not work when the agent is run outside a container on Linux environments. Initially I was told this is a minor concern, but do we have use use cases for the agent being run outside a a container on Linux? For Windows, this situation is explicitly handled (As noted in the current documentation).

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -636,3 +636,27 @@ func fluentdDriverTest(taskDefinition string, t *testing.T) {
 	err = SearchStrInDir(fluentdLogPath, "ecsfts", logTag)
 	assert.NoError(t, err, "failed to find the log tag specified in the task definition")
 }
+
+// TestMetadataServiceValidator Tests that the metadata file can be accessed from the
```
richardpen: Can we also add this test for windows?

agent/statemanager/state_manager.go
```diff
@@ -48,6 +48,7 @@ const (
 	//   b) Add 'ContainerResourcesProvisioned' as a new 'ContainerStatus' enum
 	//   c) Add 'SteadyStateStatus' field to 'Container' struct
 	//   d) Add 'ENIAttachments' struct
+	//	 e) Add 'MetadataUpdated' field to 'api.Container'
```
richardpen: nit: this should be space instead of tab.

agent/api/container.go
```diff
@@ -306,3 +310,17 @@ func (c *Container) IsInternal() bool {
 func (c *Container) IsRunning() bool {
 	return c.GetKnownStatus().IsRunning()
 }
+
+func (c *Container) IsMetadataFileUpdated() bool {
```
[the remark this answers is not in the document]
adnxn: fixed

agent/config/config_windows.go
```diff
@@ -59,7 +60,9 @@ func DefaultConfig() Config {
 			netBIOSPort,
 		},
 		ReservedPortsUDP: []uint16{},
-		DataDir:          filepath.Join(ecsRoot, "data"),
+		DataDir:          dataDir,
+		// DataDirOnHost is identical to DataDir for Windows because we do not run on a container
```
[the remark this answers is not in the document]
adnxn: fixed.

agent/containermetadata/write_metadata_windows.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return fmt.Errorf("write to metadata file for task %s container %s: %v", taskARN, containerName, err)
+	}
+	metadataFileName := filepath.Join(metadataFileDir, metadataFile)
+
+	file, err := osWrap.OpenFile(metadataFileName, os.O_WRONLY, metadataPerm)
+	if err != nil {
+		// Retry if file does not exist
+		if osWrap.IsNotExist(err) {
+			file, err = osWrap.Create(metadataFileName)
```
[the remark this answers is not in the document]
adnxn: looking at this again, we can remove the retry block with `osWrap.IsNotExist(err) {...` with changing the open file line to `file, err := osWrap.OpenFile(metadataFileName, os.O_WRONLY | os.O_CREATE, metadataPerm)`.

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,169 @@
...
+
+const (
```
aaithal: This is more of a code structure thing. I like to see consts at the top of the file (you'll find that to be the case in rest of this code base as well). Can you move this block to the top of the file?

agent/statemanager/state_manager.go
adnxn: fixed.

agent/engine/docker_task_engine.go
```diff
@@ -326,6 +337,14 @@ func (engine *DockerTaskEngine) sweepTask(task *api.Task) {
 			seelog.Errorf("Error removing container reference from image state: %v", err)
 		}
 	}
+
+	// Clean metadata directory for task
+	if engine.cfg.ContainerMetadataEnabled {
+		err := engine.metadataManager.Clean(task.Arn)
+		if err != nil {
+			seelog.Errorf("Clean task metadata failed for task %s: %v", task, err)
```
aaithal: Can you please change this to a `Warn`? I'd prefer `Error` to be used if it correlates with a customer facing error, which this is not (Applies to the rest of the file(s) as well).

adnxn: fixed.

agent/containermetadata/manager.go
adnxn: fixed.

agent/containermetadata/write_metadata_windows.go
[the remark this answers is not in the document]
adnxn: actually nvm. from godocs,

>"On Windows, the mode must be non-zero but otherwise only the 0200 bit (owner writable) of mode is used; it controls whether the file's read-only attribute is set or cleared. attribute. The other bits are currently unused. Use mode 0400 for a read-only file and 0600 for a readable+writable file."

so we might not be able to have the exact same permission as we do on linux. i'll follow up with someone on the team that knows more about windows file permissions.

aaithal:
@adnxn an you investigate why appveyor is failing and resolve merge conflicts as well?

agent/containermetadata/manager.go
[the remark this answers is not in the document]
aaithal: @adnxn can  you address @samuelkarp's comment as well? Also, can this be moved to `containermetadata/types.go` ?

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,166 @@
...
+		return fmt.Errorf("container metadata create for task %s container %s: %v", taskARN, containerName, err)
+	}
+
+	err = manager.osWrap.MkdirAll(metadataDirectoryPath, os.ModePerm)
+	if err != nil {
+		return fmt.Errorf("creating metadata directory for task %s: %v", taskARN, err)
+	}
+
+	// Acquire the metadata then write it in JSON format to the file
+	metadata := manager.parseMetadataAtContainerCreate(taskARN, containerName)
```
aaithal: The logic of parsing, marshaling and writing the metadata file is duplicated between this and the `Update` method. Please extract that to a method of its own.

agent/containermetadata/utils.go
```diff
@@ -0,0 +1,64 @@
...
+	metadataJoinSuffix = "metadata"
+)
+
+// getTaskIDfromARN parses a task ARN and produces the task ID
+// A task ARN has format arn:aws:ecs:[region]:[account-id]:task/[task-id]
+// For a correctly formatted ARN we split it over ":" into 6 parts, the last part
+// containing the task-id which we extract by splitting it by "/".
+// This function should be eventually be replaced by a standardized ARN parsing
+// module
+func getTaskIDfromARN(taskARN string) (string, error) {
```
aaithal: We should be able to get rid of this if we use the updated SDK and the `arn.Parse()` method from https://github.com/aws/aws-sdk-go/commit/1fb10cfa0eea8817b40942b70889cb1509479527

agent/containermetadata/write_metadata_unix.go
```diff
@@ -0,0 +1,68 @@
...
+// the metadata file
+func writeToMetadataFile(osWrap oswrapper.OS, ioutilWrap ioutilwrapper.IOUtil, data []byte, taskARN string, containerName string, dataDir string) error {
+	metadataFileDir, err := getMetadataFilePath(taskARN, containerName, dataDir)
+	// Boundary case if file path is bad (Such as if task arn is incorrectly formatted)
+	if err != nil {
+		return fmt.Errorf("write to metadata file for task %s container %s: %v", taskARN, containerName, err)
+	}
+	metadataFileName := filepath.Join(metadataFileDir, metadataFile)
+
+	temp, err := ioutilWrap.TempFile(metadataFileDir, "temp_metadata_file")
```
aaithal: Can you const `"temp_metadata_file"`?

agent/engine/docker_task_engine.go
```diff
@@ -326,6 +337,14 @@ func (engine *DockerTaskEngine) sweepTask(task *api.Task) {
 			seelog.Errorf("Error removing container reference from image state: %v", err)
 		}
 	}
+
+	// Clean metadata directory for task
+	if engine.cfg.ContainerMetadataEnabled {
+		err := engine.metadataManager.Clean(task.Arn)
+		if err != nil {
+			seelog.Warnf("Clean task metadata failed for task %s: %v", task, err)
```
aaithal: This should be `task.Arn`. Also, can  you add unit tests for `sweepTask` and exercise this code path?

agent/containermetadata/manager.go
adnxn: fixed.

agent/containermetadata/write_metadata_unix.go
adnxn: fixed.

agent/containermetadata/manager.go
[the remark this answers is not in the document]
adnxn: gotcha, fixed. using iotas and moved to `containermetadata/types.go`.

agent/engine/docker_task_engine.go
adnxn: fixed the task.Arn thing. and the existing happy path [test](https://github.com/aws/amazon-ecs-agent/pull/981/files/b208d502e4abee9890d82c5056d2e333638a77c5#diff-3aedbbecbf5a79074e46dafde46c1536R401) exercises that code path.

agent/containermetadata/utils.go
adnxn: fixed. updating sdk here https://github.com/aws/amazon-ecs-agent/pull/1006 and added the arn to this set of changes.

adnxn: not including this change since godep is doing [this](https://github.com/tools/godep/issues/498#issuecomment-238946586).

agent/containermetadata/parse_metadata.go
```diff
@@ -0,0 +1,137 @@
...
+	imageNameFromConfig := ""
+	if dockerContainer.Config != nil {
+		imageNameFromConfig = dockerContainer.Config.Image
+	} else {
+		seelog.Warnf("Failed to parse container metadata for task %s container %s: container has no configuration", taskARN, containerName)
+	}
+
+	// Get Port bindings from docker configurations
+	var ports []api.PortBinding
+	var err error
```
aaithal: minor: if there's no specific reason to declare this here, can you get rid of it?

agent/containermetadata/types.go
```diff
@@ -0,0 +1,165 @@
...
+	case MetadataReady:
+		text = []byte(MetadataReadyText)
+	default:
+		return nil, fmt.Errorf("failed marshalling MetadataStatus %v", status)
+	}
+	return
+}
+
+func (status *MetadataStatus) UnmarshalText(text []byte) error {
+	var err error
```
aaithal: You can get rid of this line and `return nil` at the end of the method.

agent/containermetadata/types.go
```diff
@@ -0,0 +1,165 @@
...
+// MetadataStatus specifies the current update status of the metadata file.
+// The purpose of this status is for users to check if the metadata file has
+// reached the stage they need before they read the rest of the file to avoid
+// race conditions (Since the final stage will need to be after the container
+// starts up
+// In the future the metadata may require multiple stages of update and these
+// statuses should amended/appended accordingly.
+type MetadataStatus int32
+
+func (status MetadataStatus) MarshalText() (text []byte, err error) {
```
aaithal: We have unit test for this and the unmarshal method?

agent/engine/docker_task_engine.go
```diff
@@ -675,7 +703,24 @@ func (engine *DockerTaskEngine) startContainer(task *api.Task, container *api.Co
...
+
+	// Get metadata through container inspection and available task information then write this to the metadata file
+	// Performs this in the background to avoid delaying container start
+	// TODO: Add a state to the api.Container for the status of the metadata file (Whether it needs update) and
+	// add logic to engine state restoration to do a metadata update for containers that are running after the agent was restarted
+	if dockerContainerMD.Error == nil && engine.cfg.ContainerMetadataEnabled && !container.IsInternal() {
+		go func() {
+			err := engine.metadataManager.Update(dockerContainer.DockerID, task.Arn, container.Name)
+			if err != nil {
+				seelog.Warnf("Update metadata file failed for container %s of task %s: %v", container, task, err)
```
aaithal: minor: you can `return` here and get rid of `else`.

agent/engine/docker_task_engine.go
```diff
@@ -675,7 +703,24 @@ func (engine *DockerTaskEngine) startContainer(task *api.Task, container *api.Co
...
+			} else {
+				container.SetMetadataFileUpdated()
+				seelog.Debugf("Updated metadata file for container %s of task %s", container, task)
```
aaithal: Here and in the rest of the file, let's log less: `task.Arn`, `container.Name`

agent/containermetadata/types.go
adnxn: oh i see why, fixed.

agent/engine/docker_task_engine.go
adnxn: fixed.

agent/engine/docker_task_engine.go
adnxn: gotcha. cleaned instances of this in the file.

agent/containermetadata/types.go
adnxn: added tests for this.

agent/containermetadata/parse_metadata.go
adnxn: im going to leave this one as it is, the nested branch needs both, and we return ports even if its empty.

agent/statemanager/state_manager.go
```diff
@@ -49,6 +49,7 @@ const (
 	//   c) Add 'SteadyStateStatus' field to 'Container' struct
 	//   d) Add 'ENIAttachments' struct
 	//   e) Deprecate 'SteadyStateDependencies' in favor of 'TransitionDependencySet'
+	//	 f) Add 'MetadataUpdated' field to 'api.Container'
 	ECSDataVersion = 6
```
aaithal: This should be `7` now.

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
+	"github.com/aws/amazon-ecs-agent/agent/utils/oswrapper"
+
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+const (
+	metadataEnvironmentVariable = "ECS_CONTAINER_METADATA_FILE"
```
aaithal: Please add documentation for this variable stating that this will be set in the container's host config and this is what customers would use to locate/access metadata.

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	// Ensure we do not update a container that is invalid or is not running
+	if dockerContainer == nil || !dockerContainer.State.Running {
+		return fmt.Errorf("container metadata update for task %s container %s: container not running or invalid", taskARN, containerName)
+	}
+
+	// Acquire the metadata then write it in JSON format to the file
+	metadata := manager.parseMetadata(dockerContainer, taskARN, containerName)
+	err = manager.marshalAndWrite(metadata, taskARN, containerName)
```
aaithal: nit: if you aren't doing anything with the error returned, you can just do `return manager.marshalAndWrite(metadata, taskARN, containerName)` here.

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,159 @@
...
+		return err
+	}
+	return nil
+}
+
+// Clean removes the metadata files of all containers associated with a task
+func (manager *metadataManager) Clean(taskARN string) error {
+	metadataPath, err := getTaskMetadataDir(taskARN, manager.dataDir)
+	if err != nil {
+		return fmt.Errorf("clean task %s: %v", taskARN, err)
```
aaithal: please add more context here: `"clean task metadata: unable to get metadata directory for task %s: %v"`

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	return manager.osWrap.RemoveAll(metadataPath)
+}
+
+func (manager *metadataManager) marshalAndWrite(metadata Metadata, taskARN string, containerName string) error {
+	data, err := json.MarshalIndent(metadata, "", "\t")
+	if err != nil {
+		return fmt.Errorf("create metadata for task %s container %s: %v", taskARN, containerName, err)
```
aaithal: same here, needs more context: `"create metadata for container %s in task %s: failed to marshal metadata"`

agent/containermetadata/manager.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	// Write the metadata to file
+	err = writeToMetadataFile(manager.osWrap, manager.ioutilWrap, data, taskARN, containerName, manager.dataDir)
```
aaithal: here too, you could just do `return writeMetadata...`

agent/statemanager/state_manager.go
adnxn: right, fixed.

agent/containermetadata/manager.go
adnxn: makes sense, fixed.

agent/containermetadata/manager.go
adnxn: makes sense, fixed.

agent/containermetadata/manager.go
adnxn: fixed.

agent/containermetadata/manager.go
adnxn: fixed.

agent/containermetadata/manager.go
adnxn: added documentation for `ECS_CONTAINER_METADATA_FILE` in the readme, also we should discuss if we need to include more detailed documentation wrt format of the file and the specific fields.

README.md
```diff
@@ -181,6 +181,9 @@ configure them as something other than the defaults.
 | `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES` | `["10.0.15.0/24"]` | In `awsvpc` network mode, traffic to these prefixes will be routed via the host bridge instead of the task ENI | `[]` | Not applicable |
 | `ECS_ENABLE_CONTAINER_METADATA` | `<true|false>` | Whether to enable creation of a metadata file in the containers. If enabled the agent will inject a file with the container's various metadata and the user can access the container's metadata internally by accessing the environment variable $ECS_CONTAINER_METADATA_FILE. | `false` | `false` |
 | `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the Agent is run in a container. We do not use this value in Windows because the Agent is not run on a container in Windows. | `/var/lib/ecs` | `Not used` |
+| `ECS_CONTAINER_METADATA_FILE` | set by the agent | This is variable is set in
```
aaithal: I don't think we need to document this in README.md. The variables we have here are the Agent config settings used to tune the Agent. It may make sense to add an example at the end of the readme or something demonstrating how to use the metadata feature.

adnxn: okay, i'll add a separate section for this altogether.

adnxn: not including this change in favor of separate docs pr

agent/containermetadata/manager.go
adnxn: not including this change in favor of separate docs pr

aaithal: Why not? You can still document what this variable does `// metadataEnvironmentVariable is ...`

adnxn: i didn't realise you meant in the code earlier. fixed.

adnxn: didnt realise you meant document in the code earlier. fixed now.

agent/api/container.go
```diff
@@ -317,3 +321,20 @@ func (c *Container) IsInternal() bool {
 func (c *Container) IsRunning() bool {
 	return c.GetKnownStatus().IsRunning()
 }
+
+// IsMetadataFileUpdated() returns true if the metadata file has been once the
```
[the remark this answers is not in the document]
adnxn: yep gone.

README.md
```diff
@@ -179,6 +179,8 @@ configure them as something other than the defaults.
 | `ECS_CNI_PLUGINS_PATH` | `/ecs/cni` | The path where the cni binary file is located | `/amazon-ecs-cni-plugins` | Not applicable |
 | `ECS_AWSVPC_BLOCK_IMDS` | `true` | Whether to block access to [Instance Metdata](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html) for Tasks started with `awsvpc` network mode | `false` | Not applicable |
 | `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES` | `["10.0.15.0/24"]` | In `awsvpc` network mode, traffic to these prefixes will be routed via the host bridge instead of the task ENI | `[]` | Not applicable |
+| `ECS_ENABLE_CONTAINER_METADATA` | `true` | When `true`, the agent will create a file describing the container's metadata and the file can be located and consumed by using the container enviornment variable `$ECS_CONTAINER_METADATA_FILE` | `false` | `false` |
+| `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the Agent is run in a container. We do not use this value in Windows because the Agent is not run on a container in Windows. | `/var/lib/ecs` | `Not used` |
```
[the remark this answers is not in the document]
adnxn: fixed.

PR #984, opened 2017-09-15
Task Cgroup Hooks

vsiddharth:
### Summary

* Adds cgroup hooks into `engine` and `app`

### Implementation details

* Introduces a new platform agnostic abstraction called `Resource` 

### Testing

### Licensing


agent/resources/error.go
```diff
@@ -0,0 +1,31 @@
...
+package resources
+
+// UnsupportedPlatformError indicates an error on unsupported platforms
+type UnsupportedPlatformError struct {
+	error
+}
+
+// UnsupportedPlatform returns true if the error is of UnsupportedPlatformError
+// type
+func UnsupportedPlatform(err error) bool {
```
richardpen: Please add a unit test for this

agent/api/task.go
```diff
@@ -118,15 +119,19 @@ type Task struct {
 	// used to look up the credentials for task in the credentials manager
 	credentialsID     string
 	credentialsIDLock sync.RWMutex
+
+	// memoryCPULimits to determine if task supports CPU, memory limits
+	memoryCPULimits     bool
```
sharanyad: This doesn't come to me as a bool variable. May be names like `memoryCPULimitsSet` work?

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,119 @@
...
+		Root:  cgroupRoot,
+		Specs: &linuxResourceSpec,
+	}
+
+	cgrp, err := c.control.Create(&cgroupSpec)
+	if err != nil {
+		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s", cgroupRoot)
+	}
+
+	// NOTE: This should be impossible
```
richardpen: if this is impossible, then do you really need the check here?

agent/api/task_windows.go
```diff
@@ -18,14 +18,21 @@ package api
...
 
 const (
 	portBindingHostIP = ""
 )
 
 // adjustForPlatform makes Windows-specific changes to the task after unmarshal
-func (task *Task) adjustForPlatform() {
+func (task *Task) adjustForPlatform(cfg *config.Config) {
+	task.memoryCPULimitsLock.Lock()
+	task.memoryCPULimits = cfg.TaskCPUMemLimit
```
sharanyad: Is it necessary to explicitly set this to false since its already the default?

agent/api/task_unix.go
```diff
@@ -15,16 +15,119 @@
...
+		// Adaptive calculation of CPUPeriod required for further support
+		linuxResourceSpec.CPU = &specs.LinuxCPU{
+			Quota:  &taskCPUQuota,
+			Period: &taskCPUPeriod,
+		}
+	} else {
+		var taskCPUShares uint64
+		for _, container := range task.Containers {
+			if !utils.ZeroOrNil(container.CPU) {
+				taskCPUShares += uint64(container.CPU)
```
sharanyad: A comment in the code here would improve understandability

agent/api/task.go
vsiddharth: Maybe `memoryCPULimitsEnabled`

agent/api/task_unix.go
vsiddharth: Sure

agent/api/task_unix.go
```diff
@@ -15,16 +15,119 @@
...
+
+	// If task memory limit is not present, cgroup parent memory is not set
+	// If task memory limit is set, ensure that no container
+	// of this task has a greater request
+	if !utils.ZeroOrNil(task.MemoryLimit) {
+		taskMemoryLimit := int64(task.MemoryLimit)
+		for _, container := range task.Containers {
+			containerMemoryLimit := int64(container.Memory)
+			if !utils.ZeroOrNil(containerMemoryLimit) && containerMemoryLimit > taskMemoryLimit {
+				return specs.LinuxResources{}, errors.New("task resource spec builder: invalid memory configuration")
```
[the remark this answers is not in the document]
vsiddharth: sure

agent/app/agent.go
```diff
@@ -164,9 +167,15 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
 		return exitcodes.ExitTerminal
 	}
-	// TODO:
-	// 1) Feature gate ECS cgroup root init
-	// 2) Update mocks and unit tests
+
+	// Conditionally create '/ecs' cgroup root
+	if agent.cfg.TaskCPUMemLimit {
+		err = agent.resource.Init()
+		if err != nil || resources.UnsupportedPlatform(err) {
```
petderek: This check is redundant.

`|| resources.UnsupportedPlatform(err)` will never be true when the first term is false..

agent/api/task_windows.go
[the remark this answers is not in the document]
vsiddharth: The logic in config will ensure that this is `false` on windows but I can explicitly set it here too. 
Thinking once again, we might not even need this chunk.

agent/app/agent.go
```diff
@@ -164,9 +167,15 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		log.Criticalf("Error creating state manager: %v", err)
 		return exitcodes.ExitTerminal
 	}
-	// TODO:
-	// 1) Feature gate ECS cgroup root init
-	// 2) Update mocks and unit tests
+
+	// Conditionally create '/ecs' cgroup root
+	if agent.cfg.TaskCPUMemLimit {
```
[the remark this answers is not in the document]
vsiddharth: I would like to condition things on `config`. 
Moving this to `Init` would mean passing in additional arguments which I'd like to avoid

agent/resources/resources_linux_test.go
```diff
@@ -0,0 +1,210 @@
+// +build linux
```
petderek: Will agent still build on mac?

agent/engine/task_manager.go
```diff
@@ -129,9 +133,14 @@ func (mtask *managedTask) overseeTask() {
 			// able to move some containers along.
 			llog.Debug("Task not steady state or terminal; progressing it")
 
-			// TODO:
-			// 1) Add platform resources setup and failure handling
-			// 2) Add new task resources provisioned state ?
+			if mtask.engine.cfg.TaskCPUMemLimit {
```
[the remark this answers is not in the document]
vsiddharth: I can add the relevant sections back

agent/engine/task_manager.go
```diff
@@ -129,9 +133,14 @@ func (mtask *managedTask) overseeTask() {
...
+				err := mtask.resource.Setup(mtask.Task)
+				if err != nil || resources.UnsupportedPlatform(err) {
+					seelog.Criticalf("Unable to setup task platform resources: %v", err)
```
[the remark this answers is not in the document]
vsiddharth: sure

agent/engine/task_manager.go
```diff
@@ -605,7 +614,12 @@ func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 	go mtask.discardEventsUntil(handleCleanupDone)
 	mtask.engine.sweepTask(mtask.Task)
 
-	// TODO: Remove platform task resources
+	if mtask.engine.cfg.TaskCPUMemLimit {
+		err := mtask.resource.Cleanup(mtask.Task)
+		if err != nil || resources.UnsupportedPlatform(err) {
```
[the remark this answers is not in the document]
vsiddharth: sure

agent/resources/resources_linux.go
vsiddharth: Just being extra cautious. It shouldn't hurt

agent/resources/resources_windows.go
```diff
@@ -0,0 +1,52 @@
+// +build windows
```
petderek: In line with my other comment about building on mac, I have a few suggestions:
* rename this to `resources_unsupported'
* build tag `!linux`
* return nil instead of returning errors

Windows and Unix are unsupported. We want agent to compile on mac (not necessarily for running agent, but testing and compiling is important to folks using IDEs on mac). Hence the change to build tag / filename.

agent/api/task.go
petderek: 'enabled' is what we use in the config variables.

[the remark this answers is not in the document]
vsiddharth: We want to avoid passing in config from `engine` while updating the containers `CgroupParent`. 
Now during `PostUnmarshalTask`, this attribute would be updated and `DockerHostConfig` would use this. 

All hacks are localized

agent/api/task_unix.go
```diff
@@ -15,16 +15,119 @@
...
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
 const (
 	portBindingHostIP = "0.0.0.0"
+	maxTaskVCPULimit  = 10     // 10 VCPUs limit
+	defaultCPUPeriod  = 100000 // 100ms
+	// Reference: http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html
+	defaultCPUShare = 2
```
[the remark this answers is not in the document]
vsiddharth: sure

agent/engine/testdata/test_tasks/sleep5TaskCgroup.json
```diff
@@ -0,0 +1,40 @@
+{
```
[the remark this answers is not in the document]
vsiddharth: This is inline with the new changes to `api.Task`. 
Would also come in handy during `integ` and `functional` tests

agent/api/task_unix.go
```diff
@@ -15,16 +15,126 @@
...
+	"github.com/cihub/seelog"
+	docker "github.com/fsouza/go-dockerclient"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
 const (
 	portBindingHostIP = "0.0.0.0"
+	maxTaskVCPULimit  = 10     // 10 VCPUs limit
+	defaultCPUPeriod  = 100000 // 100ms
```
aaithal: Can you set this const with the appropriate unit (`100 * time.Millisecond`) https://play.golang.org/p/NpNlwVp4ez. You can also look at https://golang.org/pkg/time/#Duration for a reference of how to do various conversion you may need.

agent/api/task_unix.go
```diff
@@ -15,16 +15,126 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+	"github.com/cihub/seelog"
+	docker "github.com/fsouza/go-dockerclient"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
 const (
 	portBindingHostIP = "0.0.0.0"
+	maxTaskVCPULimit  = 10     // 10 VCPUs limit
```
aaithal: please add some documentation here, especially about what it means to set this to 10 VCPUs.

agent/api/task_unix.go
```diff
@@ -15,16 +15,126 @@
...
-	return false
+// BuildCgroupRoot helps build the task cgroup prefix
+// Example: /ecs/task-id
+func (task *Task) BuildCgroupRoot() (string, error) {
+	taskID, err := task.GetID()
+	if err != nil {
+		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id from taskARN: %s", task.Arn)
+	}
+
+	cgroupRoot := filepath.Join(config.DefaultTaskCgroupPrefix, taskID)
```
aaithal: minor: you can combine these to just one line: `return filepath.Join(...), nil`

agent/api/task_unix.go
```diff
@@ -15,16 +15,126 @@
...
+	return cgroupRoot, nil
+}
+
+// BuildLinuxResourceSpec returns a linuxResources object for the task cgroup
+func (task *Task) BuildLinuxResourceSpec() (specs.LinuxResources, error) {
+	linuxResourceSpec := specs.LinuxResources{}
+
+	// If task level CPU limits are requested, set CPU quota + CPU period
+	if !utils.ZeroOrNil(task.VCPULimit) {
+		if task.VCPULimit > maxTaskVCPULimit || task.VCPULimit < 0 {
```
aaithal: Can this be equal to `0`?

agent/app/agent.go
```diff
@@ -164,9 +167,15 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+		err = agent.resource.Init()
+		if err != nil {
+			log.Criticalf("Unable to setup '/ecs' cgroup: %v", err)
+			return exitcodes.ExitTerminal
```
aaithal: Can you add a comment here explaining why this should be treated as `Terminal` ?

agent/api/task_unix.go
```diff
@@ -15,16 +15,126 @@
...
 
 func getCanonicalPath(path string) string { return path }
 
-// CgroupEnabled checks whether cgroups need be enabled at the task level
-func (task *Task) CgroupEnabled() bool {
-	//stub for now
-	return false
+// BuildCgroupRoot helps build the task cgroup prefix
+// Example: /ecs/task-id
+func (task *Task) BuildCgroupRoot() (string, error) {
```
aaithal: Shouldn't this be `buildCgroupRoot`?

agent/api/task_unix.go
```diff
@@ -15,16 +15,126 @@
...
+	if err != nil {
+		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id from taskARN: %s", task.Arn)
+	}
+
+	cgroupRoot := filepath.Join(config.DefaultTaskCgroupPrefix, taskID)
+	return cgroupRoot, nil
+}
+
+// BuildLinuxResourceSpec returns a linuxResources object for the task cgroup
+func (task *Task) BuildLinuxResourceSpec() (specs.LinuxResources, error) {
```
aaithal: Shouldn't this be `buildLinuxResourceSpec`?

agent/api/task_unix.go
vsiddharth: Sure

agent/api/task_unix.go
vsiddharth: Ok

agent/api/task_unix.go
vsiddharth: Called outside this package too

agent/api/task_unix.go
vsiddharth: Sure

agent/api/task_unix.go
vsiddharth: Called outside this package too

agent/api/task_unix.go
vsiddharth: `Task.VCPULimit` can be `0` but in that case we do:
* Aggregate task container's `CPUShares` for the task level setting
* Set `CPUShares := 2` when there are no task level or container level restrictions

agent/app/agent.go
vsiddharth: Sure

agent/resources/resources_linux_test.go
vsiddharth: Will do

agent/api/task_unix.go
```diff
@@ -15,16 +15,127 @@
...
-// CgroupEnabled checks whether cgroups need be enabled at the task level
-func (task *Task) CgroupEnabled() bool {
-	//stub for now
-	return false
+// BuildCgroupRoot helps build the task cgroup prefix
+// Example: /ecs/task-id
+func (task *Task) BuildCgroupRoot() (string, error) {
+	taskID, err := task.GetID()
+	if err != nil {
+		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id from taskARN: %s", task.Arn)
```
aaithal: nit: `taskARN` -> `task ARN`

agent/api/task_unix.go
```diff
@@ -15,16 +15,127 @@
...
+		linuxResourceSpec.CPU = &specs.LinuxCPU{
+			Shares: &taskCPUShares,
+		}
+	}
+
+	// If task memory limit is not present, cgroup parent memory is not set
+	// If task memory limit is set, ensure that no container
+	// of this task has a greater request
+	if !utils.ZeroOrNil(task.MemoryLimit) {
+		taskMemoryLimit := int64(task.MemoryLimit)
```
aaithal: isn't this already `int64` ?

agent/api/task_unix.go
```diff
@@ -15,16 +15,127 @@
...
+		for _, container := range task.Containers {
+			containerMemoryLimit := int64(container.Memory)
+			if !utils.ZeroOrNil(containerMemoryLimit) && containerMemoryLimit > taskMemoryLimit {
```
aaithal: What is the `!utils.ZeroOrNil(containerMemoryLimit)` check giving us here that the `>` is not already doing?

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,119 @@
...
+		return nil
+	}
+	return c.control.Init()
+}
+
+// setupCgroup is used to create the task cgroup
+func (c *cgroupWrapper) setupCgroup(task *api.Task) error {
+	cgroupRoot, err := task.BuildCgroupRoot()
+	if err != nil {
+		return errors.Wrap(err, "resource: setup cgroup: unable to determine cgroup root")
```
aaithal: can you include task's arn here as well?

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	seelog.Debugf("Setting up cgroup at: %s", cgroupRoot)
```
aaithal: can you include task's arn here as well?

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,119 @@
...
+
+	if c.control.Exists(cgroupRoot) {
+		seelog.Debugf("Cgroup at %s already exists, skipping creation", cgroupRoot)
```
aaithal: can you include task's arn here as well?

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	cgroupSpec := cgroup.Spec{
+		Root:  cgroupRoot,
+		Specs: &linuxResourceSpec,
+	}
+
+	cgrp, err := c.control.Create(&cgroupSpec)
+	if err != nil {
+		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s", cgroupRoot)
```
aaithal: can you include task's arn here as well?

agent/resources/resources_linux.go
aaithal: Please add a note for when you anticipate for this to happen. I agree with @richardpen  that if it's not plausible for this condition to occur, it might be better to let the Agent panic rather than handle it as an error.

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"github.com/aws/amazon-ecs-agent/agent/api"
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+// cgroupWrapper implements the Resource interface
+type cgroupWrapper struct {
```
sharanyad: Do we expect to add more linux specific resources here in later? In either case, is it reasonable to name this as `cgroupWrapper` or the file as `resources_linux.go`?

vsiddharth: There is no plan to extend this immediately. 
`cgroupWrapper` just aims to implement the `Resource` interface on `linux`. 
It just wraps up the cgroup object for now!

agent/api/task.go
```diff
@@ -125,6 +125,10 @@ type Task struct {
 	// ENI is the elastic network interface specified by this task
 	ENI     *ENI
 	eniLock sync.RWMutex
+
+	// memoryCPULimitsEnabled to determine if task supports CPU, memory limits
+	memoryCPULimitsEnabled     bool
```
samuelkarp: Does this need to be saved in the state file?  Or are we recomputing when the task is read from the state file?

agent/api/task_test.go
```diff
@@ -579,7 +581,8 @@ func TestPostUnmarshalTaskWithEmptyVolumes(t *testing.T) {
 	task, err := TaskFromACS(&taskFromACS, &ecsacs.PayloadMessage{SeqNum: &seqNum})
 	assert.Nil(t, err, "Should be able to handle acs task")
 	assert.Equal(t, 2, len(task.Containers)) // before PostUnmarshalTask
-	task.PostUnmarshalTask(nil, nil)
+	cfg := config.Config{TaskCPUMemLimit: false}
```
samuelkarp: Do you need `TaskCPUMemLimit` specified here?  Should be false by default.

agent/api/task_unix.go
```diff
@@ -25,16 +33,113 @@ const (
...
+	return filepath.Join(config.DefaultTaskCgroupPrefix, taskID), nil
 }
 
-func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig) {}
+// BuildLinuxResourceSpec returns a linuxResources object for the task cgroup
+func (task *Task) BuildLinuxResourceSpec() (specs.LinuxResources, error) {
+	linuxResourceSpec := specs.LinuxResources{}
+
+	// If task level CPU limits are requested, set CPU quota + CPU period
+	if !utils.ZeroOrNil(task.VCPULimit) {
```
samuelkarp: (multiple occurrences) I would prefer to avoid `ZeroOrNil` in critical code.  I think it would be better if you explicitly checked for those things here.  `ZeroOrNil` is really meant for handling unknown data (`interface{}`) and not really for handling things where the type is known.

agent/api/task_unix.go
```diff
@@ -25,16 +33,113 @@ const (
...
+	if err != nil {
+		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id from task ARN: %s", task.Arn)
+	}
+
+	return filepath.Join(config.DefaultTaskCgroupPrefix, taskID), nil
 }
 
-func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig) {}
+// BuildLinuxResourceSpec returns a linuxResources object for the task cgroup
+func (task *Task) BuildLinuxResourceSpec() (specs.LinuxResources, error) {
```
samuelkarp: I think this function is a bit on the long side with a larger amount of indentation than ideal.  Might be useful to break it up into smaller functions.  Something like:

```go
func (task *Task) BuildLinuxResourceSpec() (specs.LinuxResources, error) {}
func (task *Task) buildExplicitCPUSpec() (specs.LinuxCPU, error) {}
func (task *Task) buildImplicitCPUSpec() (specs.LinuxCPU, error) {}
func (task *Task) buildMemorySpec() (specs.LinuxMemory, error) {}
```

agent/app/agent.go
```diff
@@ -188,9 +191,18 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
-	// TODO:
-	// 1) Feature gate ECS cgroup root init
-	// 2) Update mocks and unit tests
+
+	// Conditionally create '/ecs' cgroup root
+	if agent.cfg.TaskCPUMemLimit {
+		err = agent.resource.Init()
+		// When task CPU and memory limits are enabled, all tasks are placed
+		// under the '/ecs' cgroup root. If we are unable to create this
+		// cgroup, the agent should exit with a terminal error.
```
samuelkarp: This should *only* be fatal if task resource limits are *explicitly* enabled (rather than just implicitly enabled).  The reason is that a customer may call `UpdateContainerAgent` on an instance that is running an old version of `ecs-init` that doesn't mount in the cgroup hierarchy; we should not fail to restart the agent simply because a feature that they're not using doesn't work.

agent/engine/task_manager.go
```diff
@@ -129,9 +133,15 @@ func (mtask *managedTask) overseeTask() {
 			// able to move some containers along.
 			llog.Debug("Task not steady state or terminal; progressing it")
 
-			// TODO:
-			// 1) Add platform resources setup and failure handling
-			// 2) Add new task resources provisioned state ?
+			// TODO: Add new resource provisioned state ?
```
samuelkarp: Yeah, we'll need to do something.  Right now, this will run on every task transition, which isn't what we want.

agent/app/agent.go
```diff
@@ -92,6 +93,7 @@ type ecsAgent struct {
 	vpc                   string
 	subnet                string
 	mac                   string
+	resource              resources.Resource
```
samuelkarp: This feels like the wrong naming to me, but I think we can leave it for now just to move this forward.

agent/app/agent.go
vsiddharth: This is a really good point but this call is conditioned on the `config`. 
We are going to have another PR to conditionally disable the feature and we can ensure that the cgroup mounts are valid at that point.

agent/engine/task_manager.go
vsiddharth: Yes, but we don't recreate the `cgroup` if it already exists.

agent/api/task.go
samuelkarp: Are you going to persist the task limit settings or cgroup spec here?  We need something that indicates to a restarted agent that a task has a parent cgroup and whether it has already been set up.

agent/api/task_test.go
samuelkarp: Can you remove it?

agent/app/agent.go
samuelkarp: Can you add a TODO if you're not going to address this now?  We need to get this done correctly before release, but it doesn't need to be part of this PR.

agent/engine/task_manager.go
samuelkarp: I'm working on expressing transition dependencies a bit more explicitly right now (PR coming either today or tomorrow).  Rather than adding a new state, we might just want to add an explicit model for the resource to the task, and then make each of the containers dependent on that resource.

agent/api/task.go
vsiddharth: We can persist this onto the state file but its still not going to indicate if the cgroup hierarchy has been setup for the task.

agent/api/task_test.go
vsiddharth: I will remove it in my next revision

agent/app/agent.go
vsiddharth: I will add a `TODO` here in my next revision
It will be addressed in a different PR though

agent/engine/task_manager.go
vsiddharth: Yep, that makes sense but I'll leave the `TODO` as is for now and we can update it following your changes

samuelkarp: Check out https://github.com/aws/amazon-ecs-agent/pull/989 and see if that makes sense to you as a possible way to address this.

agent/api/task_unix.go
```diff
@@ -25,16 +32,140 @@ const (
...
+		if err != nil {
+			return specs.LinuxResources{}, err
+		}
+		linuxResourceSpec.CPU = &linuxCPUSpec
+	} else {
+		linuxCPUSpec := task.buildImplicitLinuxCPUSpec()
+		linuxResourceSpec.CPU = &linuxCPUSpec
+	}
+
+	// If task memory limit is not present, cgroup parent memory is not set
```
samuelkarp: This comment should really be in `buildLinuxMemorySpec` at this point.

agent/engine/task_manager.go
vsiddharth: @samuelkarp I've looked at #989 and will update this `TODO` based off it when its been merged.

agent/api/task_unix.go
vsiddharth: Done

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,120 @@
...
+		return errors.Wrapf(err, "resource: setup cgroup: unable to build resource spec for task: %s", task.Arn)
+	}
+
+	cgroupSpec := cgroup.Spec{
+		Root:  cgroupRoot,
+		Specs: &linuxResourceSpec,
+	}
+
+	cgrp, err := c.control.Create(&cgroupSpec)
+	if err != nil {
```
aaithal: I wonder if we need to think about the use-case where there could be an error in one of the subsystems after a number of subsystems have already been initialized (https://github.com/containerd/cgroups/blob/master/cgroup.go#L73). How do we clean up in such scenarios?

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,120 @@
...
+		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s for task: %s", cgroupRoot, task.Arn)
+	}
+
+	// NOTE: This should be impossible
+	// cgrp is an interface and it should never be nil when err == nil
+	if cgrp == nil {
```
aaithal: Please get rid of this if this is impossible (which it seems like from https://github.com/containerd/cgroups/blob/master/cgroup.go#L74).

agent/resources/resources_linux.go
```diff
@@ -0,0 +1,120 @@
...
+func (c *cgroupWrapper) cleanupCgroup(task *api.Task) error {
+	cgroupRoot, err := task.BuildCgroupRoot()
+	if err != nil {
+		return errors.Wrapf(err, "resource: cleanup cgroup: unable to determine cgroup root for task: %s", task.Arn)
+	}
+
+	seelog.Debugf("Cleaning up cgroup at: %s for task: %s", cgroupRoot, task.Arn)
+
+	err = c.control.Remove(cgroupRoot)
+	if err != nil {
```
aaithal: Please modify this to handle `ErrCgroupDeleted` (Ignore and log the error in such a case)

agent/api/task_unix.go
```diff
@@ -25,16 +32,140 @@ const (
...
+	if task.VCPULimit > maxTaskVCPULimit {
+		return specs.LinuxCPU{},
+			errors.Errorf("task CPU spec builder: unsupported CPU limits, requested=%f, max-supported=%d",
+				task.VCPULimit, maxTaskVCPULimit)
+	}
+	taskCPUPeriod := uint64(defaultCPUPeriod / time.Microsecond)
+	taskCPUQuota := int64(task.VCPULimit * float64(taskCPUPeriod))
+
+	// TODO: DefaultCPUPeriod only permits 10VCPUs.
+	// Adaptive calculation of CPUPeriod required for further support
```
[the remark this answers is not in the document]
vsiddharth: Sure, I'll add more details here

agent/api/task_unix.go
```diff
@@ -25,16 +32,142 @@ const (
...
+// platformHostConfigOverride to override platform specific feature sets
+func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig) error {
+	// Override cgroup parent
+	return task.overrideCgroupParent(hostConfig)
+}
+
+// overrideCgroupParent updates hostconfig with cgroup parent when task cgroups
+// are enabled
+func (task *Task) overrideCgroupParent(hostConfig *docker.HostConfig) error {
+	task.memoryCPULimitsEnabledLock.RLock()
```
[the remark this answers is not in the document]
vsiddharth: `task.BuildCgroupRoot` just creates `/ecs/task-id` and should be efficient. 
I also need the lock to determine if `memoryCPULimitsEnabled` is set correctly. 

I don't see any major benefit in reducing the lock's scope here

agent/api/task_unix.go
```diff
@@ -25,16 +32,142 @@ const (
...
+	defer task.memoryCPULimitsEnabledLock.RUnlock()
+	if task.MemoryCPULimitsEnabled {
+		cgroupRoot, err := task.BuildCgroupRoot()
+		if err != nil {
+			seelog.Debugf("Unable to obtain task cgroup root for task %s: %v", task.Arn, err)
```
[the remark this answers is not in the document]
vsiddharth: Sure, it was just additional logging to make things easier to understand.

agent/app/agent.go
vsiddharth: `cgroup` isn't platform agnostic as it does not apply to `windows`

agent/app/agent.go
vsiddharth: `TODO` has been added above on `L196`. I'll refine the TODO's here

agent/resources/resources_linux.go
vsiddharth: The `cgroup` subsystem and associated controllers have been around for a long time and are very efficient. I don't see any particular reason why a specific `subsystem/controller` would misbehave. I don't think we have to concern ourselves with this problem as we don't operate at such a low level within the system. 

If you have any specific concerns, I'd like to hear more about them.

agent/resources/resources_linux.go
vsiddharth: I still hold my case that having code to check for impossible conditions is fair. 
Isn't it better to fail to start a task rather than making the agent panic.  

I can remove it if you feel strongly about it.

agent/resources/resources_linux.go
vsiddharth: Sure, I can handle this explicitly here but its hard to think of how we might end up with such a situation.

agent/resources/resources_linux.go
aaithal: I'd rather panic if we ever hit an impossible condition as that'd mean something else has gone horribly wrong as well. This is dead code now for all practical purposes and I'd rather not have that in code.

agent/resources/resources_linux.go
aaithal: We have seen the Agent trying to clean up the same task twice before because of bugs. No reason why we won't run into it again.

agent/resources/resources_linux.go
aaithal: I am not advocating that we change anything here, but I don't buy the argument that something has been around for a long time and can thus be considered to never error out. We're interacting with the cgroup system using a go library and any number of things can go wrong in that interaction. It's a very useful exercise to look at what the library is actually doing and evaluate if something could go wrong and how we respond to scenarios when it does go wrong. 

Have we verified that the call paths are all thread safe and do not have any blocking routines before making that statement? Are we sure that there are no additional safeguards that we should be putting in place? We have seen such things coming back to bite us when using other 3p packages. 

As an example, what happens if a customer has a configuration where they've disabled/unmounted a particular cgroup system? Or, how does this library behave if the Agent is launched with reduced privileges? Again, these things might not necessarily result in the error in this particular method, but I'm not sure if all of these scenarios will be/are being gracefully handled.

aaithal: Oh, the scenario that I had in mind when I asked you this was let's say Agent gets restarted after a couple of subsystems have been initialized. What happens on recovery?

vsiddharth: I implied that the cgroup subsystem within the Linux kernel is fairly stable and hasn't changed much in a long time. There are plans for `cgroups v2` but its still `WIP` and is missing  controllers last time I investigated. 

I have looked into the 3p library we are using and its well maintained and doesn't have any of the problems you are concerned about. I have experimented with the library and have not identified any major issues. Even when specific cgroup subsystems are disabled, the library still works accurately. 

The features we care about depend primarily on the `cpu` and `memory` subsystems. 

`amazon-ecs-init` is going to launch `amazon-ecs-agent` with `CAP_NET_ADMIN` and `CAP_SYS_ADMIN`. 

If the instance has a working docker environment and the agent container is launched with the required mounts, the library and the agent work in harmony to launch tasks accurately.

vsiddharth: If the agent is restarted and new cgroup subsystems have been initialized, the agent should still function without any issues. The agent would attempt to `Load` the cgroup if it already exists before making any attempts to `recreate` the cgroup.

samuelkarp: Cgroup subsystems can be enabled and disabled independently.  However, Docker generally relies on the same ones being enabled that we rely on (`cpu` and `memory` here), so I'm not too worried about that case.

agent/api/task_unix.go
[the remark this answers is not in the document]
vsiddharth: Done

agent/engine/task_manager.go
```diff
@@ -129,9 +133,15 @@ func (mtask *managedTask) overseeTask() {
...
+			if mtask.engine.cfg.TaskCPUMemLimit {
+				err := mtask.resource.Setup(mtask.Task)
+				if err != nil {
+					seelog.Criticalf("Unable to setup platform resources for task %s: %v", mtask.Task.Arn, err)
+					mtask.SetDesiredStatus(api.TaskStopped)
+					mtask.engine.emitTaskEvent(mtask.Task, taskUnableToCreatePlatformResources)
+				}
+			}
```
aaithal: Please add a log line here indicating that the cgroup hierarchy has been established for the task.

vsiddharth: Can I add it after #989 is merged and we update the transition dependencies as this log line would now appear on every task transition ?

aaithal: Yeah, that's fine. You can add a TODO for the time being so that we won't forget it.

vsiddharth: Done! 
Thanks

PR #987, opened 2017-09-20
Upgrade go docker client

richardpen:
### Summary

Upgrade go docker client

### Implementation details

The only change here is that added buffer for the channel to receive docker events. As go docker client changed the behavior that if the listener buffer is full the event won't be sent. See related code [here](https://github.com/fsouza/go-dockerclient/commit/6854512932fa6bd072204f1f4ed60a4e4c45e2f3)

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/docker_container_engine.go
+++ b/agent/engine/docker_container_engine.go
@@ -42,6 +42,8 @@ const (
 	dockerDefaultTag = "latest"
 	// imageNameFormat is the name of a image may look like: repo:tag
 	imageNameFormat = "%s:%s"
+	// the buffer size will ensure agent doesn't miss any event from docker
+	dockerEventBufferSize = 100
 )
 
 // Timelimits for docker operations enforced above docker
@@ -718,17 +720,24 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 	if err != nil {
 		return nil, err
 	}
+	dockerEvents := make(chan *docker.APIEvents, dockerEventBufferSize)
 	events := make(chan *docker.APIEvents)
+	buffer := NewInfiniteBuffer()
 
-	err = client.AddEventListener(events)
+	err = client.AddEventListener(dockerEvents)
 	if err != nil {
 		log.Error("Unable to add a docker event listener", "err", err)
 		return nil, err
 	}
 	go func() {
 		<-ctx.Done()
-		client.RemoveEventListener(events)
+		client.RemoveEventListener(dockerEvents)
 	}()
+
+	// Cache the event from go docker client
+	go buffer.Serve(dockerEvents)
+	// Read the buffered events and send to task engine
+	go buffer.Consume(events)
 
 	changedContainers := make(chan DockerContainerChangeEvent)
 
--- a/agent/engine/docker_container_engine.go
+++ b/agent/engine/docker_container_engine.go
@@ -743,11 +743,6 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 
 	go func() {
 		for event := range events {
-			// currently only container events type needs to be handled
-			if event.Type != "container" || event.ID == "" {
-				continue
-			}
-
 			containerID := event.ID
 			log.Debug("Got event from docker daemon", "event", event)
 
@@ -761,23 +756,6 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 				fallthrough
 			case "die":
 				status = api.ContainerStopped
-			case "kill":
-				fallthrough
-			case "rename":
-				// TODO, ensure this wasn't one of our containers. This isn't critical
-				// because we typically have the docker id stored too and a wrong name
-				// won't be fatal once we do
-				continue
-			case "restart":
-			case "resize":
-			case "unpause":
-			case "destroy":
-				// container remove events doesn't need to be handled by ecs agent
-				// as it has already moved to stopped before this event
-				continue
-
-			// These result in us falling through to inspect the container, some
-			// out of caution, some because it's a form of state change
 			case "oom":
 				containerInfo := event.ID
 				// events only contain the container's name in newer Docker API
@@ -790,39 +768,8 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 				// "oom" can either means any process got OOM'd, but doesn't always
 				// mean the container dies (non-init processes). If the container also
 				// dies, you see a "die" status as well; we'll update suitably there
-				fallthrough
-			case "pause":
-				// non image events that aren't of interest currently
-				fallthrough
-			case "exec_create":
-				fallthrough
-			case "exec_start":
-				fallthrough
-			case "top":
-				fallthrough
-			case "attach":
-				fallthrough
-			// image events
-			case "export":
-				fallthrough
-			case "pull":
-				fallthrough
-			case "push":
-				fallthrough
-			case "tag":
-				fallthrough
-			case "untag":
-				fallthrough
-			case "import":
-				fallthrough
-			case "delete":
-				// No interest in image events
 				continue
 			default:
-				if strings.HasPrefix(event.Status, "exec_create:") || strings.HasPrefix(event.Status, "exec_start:") {
-					continue
-				}
-
 				// Because docker emits new events even when you use an old event api
 				// version, it's not that big a deal
 				seelog.Debugf("Unknown status event from docker: %s", event.Status)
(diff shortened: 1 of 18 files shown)


agent/engine/docker_container_engine.go
```diff
@@ -39,7 +39,8 @@ import (
 )
 
 const (
-	dockerDefaultTag = "latest"
+	dockerDefaultTag          = "latest"
+	containerEventsBufferSize = 1000
```
aaithal: Why was this changed? Please add a description in either the commit message or as inline documentation. Doesn't this open up the possibility of the ECS Agent processing multiple events from docker simultaneously? Have you analyzed the impact of the same on the container and task engines?

richardpen: The description is added in the Summary, I will add the same in the commit message. 
I don't think ECS Agent can process multiple events simultaneously, based on the [code](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine.go#L403) the events will be handled serially. But I could be wrong, please correct me if that's not true.
>Have you analyzed the impact of the same on the container and task engines?

Can you elaborate here, I don't think this will have impact on the container or task engine since the behavior of handling this doesn't change. The event will still be handled serially in task engine.

vsiddharth: Why `1000` ? 
Could you please add more details here ?

richardpen: 1000 is picked up due to the maximum number of containers can be ran on an instance is ~1000. Let me know if you have better ideas.

samuelkarp: That's not accurate.  There's a limit of 1023 containers using the `bridge` network mode, but there is no limit of containers using `host` or `none`.

vsiddharth: A `note/comment` is here going to be useful. 
It would nice to calculate this value rather than using a fixed constant imho.

richardpen:
@aaithal @samuelkarp @vsiddharth I have changed this to the same behavior as before, where the go routine that was used to send events was moved from go docker client to agent. Let me know if you have other suggestions.

agent/engine/docker_container_engine.go
```diff
@@ -712,13 +715,24 @@ func metadataFromContainer(dockerContainer *docker.Container) DockerContainerMet
 	return metadata
 }
 
+// CopyDockerEventsWithoutBlocking copies the event from input to output without blocking the input
+func CopyDockerEventsWithoutBlocking(in chan *docker.APIEvents, out chan *docker.APIEvents) {
```
samuelkarp: I don't understand how this doesn't block.  This just means that you have an effective buffer of `2*containerEventsBufferSize` and if that's full, you'll still block...

richardpen: Even if the `out` is full, the read of `in` won't be block, as it simply reads and creates a go routine to write. In this case these go routines will be blocked until `out` isn't full, but go docker client can keeps writing to `in` so that agent won't miss any event.

In rare case, if the process to create the go routine slows down, which means producer is much faster than consumer, the `in` could be full, that's why I had a buffer size of 100 for `in`. This is the best way I can think of, please suggest.

samuelkarp: I missed that you're creating a goroutine for *each* event.  I think this is not a good idea; we'll have goroutine explosion and will lose any ordering that was provided by go-dockerclient.

You can implement a non-blocking, infinite (subject to how much memory you have) buffer through appending to a slice.  However, I think the more important question here is to determine what the correct behavior should be.

* Do we rely on every single event being delivered?  We have logic that tolerates event loss and reconciles with explicit `inspect`s.
* What happens when we lose an event? Our reconciliation runs every 10 minutes, so we could have a 10-minute window of inconsistency.
* Do we need every event?  No, there's only a small subset of events that we pay attention to at all.

richardpen: I agree this is not a good idea, but this is the same as how the agent works today. Previously the go routine to send event was created in go docker client code, but now they are created in the agent code.

Currently, the go docker client doesn't grantee the order of the event sent, so I think agent have the tolerance for the order of the event. 

>Do we rely on every single event being delivered? We have logic that tolerates event loss and reconciles with explicit inspects.
What happens when we lose an event? Our reconciliation runs every 10 minutes, so we could have a 10-minute window of inconsistency.

Ten minutes is really a big gap, we probably don't want the task status change report have a 10 minutes delay.

>Do we need every event? No, there's only a small subset of events that we pay attention to at all.

We don't need all these events, but the agent has to do the analysis, currently go docker client doesn't support the filter.

I will try to implement a non-blocking infinite buffer to solve the goroutine explosion problem while keeping the order of the event. Any concerns or suggestions?

samuelkarp: It might make more sense to filter here so we can discard early.  It also might make more sense to do analysis of expected event volume under high load to determine if we can figure out an appropriate buffer size.

richardpen:
@aaithal @samuelkarp @vsiddharth I did some test with ~800 containers start and stop, no event will miss with the buffer size 100, while with 1 there will still be some events missing. Also from the benchmark test, docker can produce the event with at most 60/second. Please let me know if you have other concerns.

aaithal:
@richardpen Awesome! Thanks for doing that. Can you post the details of your experiments here (scripts if you have some) so that we can reproduce/rerun these tests in the future? My concern still is that 60 events per second might be the max tps that docker daemon is hitting today, but it can change with improvements made to that code base. It's not a good place to be constatnly worry about the buffer size that we have with each new release of docker. We should ensure that we have a better reconciliation story for missing events altogether, because at some point of time we will miss these events.

richardpen:
@aithal, I agree with you that this still depends on the tps of docker events, and this isn't the only factor that impact the behavior. It will also be impacted by the speed of `write to a channel`(go docker client write event) and `read from the channel and create a go routine`(agent read the channel and create a go routine to record the event). I'm not sure how better we can do, but this should unblock us for now to upgrade the go docker client. We can probably have some other options like make the reconciliation interval configurable in the future or move to use other client tools?

The test I did for benchmarking docker events spike is launching short-life cycle containers on a high performance instance, and calculate the output of docker events. eg:
```
#!/bin/bash

for i in {1..1000};
do
    docker run -d --rm ubuntu bash sleep 5s &
done
```
The test with ecs-agent is try launch 1000 short-life cycle containers(with sleep 1m) on high performance instance, and count the event sent by go docker client and the event received by the agent.

aaithal:
@richardpen Thanks. I'd be a lot more comfortable with this change if we reduced our steady state polling/reconciliation time to `5m` instead of the `10m` that it's currently set to. I'd even say that something like `2m` or `3m` should be fine. But, at the very least, we should reduce it to `5m` so that the worst case latency of missing an event and reporting the same to the backend is reduced by half of what its set to today.

agent/engine/docker_events_buffer.go
```diff
@@ -0,0 +1,92 @@
...
+
+// NewInfiniteBuffer returns an InfiniteBuffer object
+func NewInfiniteBuffer() *InfiniteBuffer {
+	return &InfiniteBuffer{
+		waitDone: make(chan struct{}),
+	}
+}
+
+// Serve starts reading from the input channel and writes to the buffer
+func (buffer *InfiniteBuffer) Serve(events chan *docker.APIEvents) {
```
aaithal: The very common `Serve()` method from `http` is a blocking call and it might be confusing for us to have a method named `Serve` that's not blocking. Can you rename this to `StartListener()` or something?

agent/engine/docker_events_buffer.go
```diff
@@ -0,0 +1,92 @@
...
+	for event := range events {
+		go buffer.Write(event)
+	}
+}
+
+// Write writes the event into the buffer
+func (buffer *InfiniteBuffer) Write(event *docker.APIEvents) {
```
aaithal: `CopyEvents` might be a better name here?

agent/engine/docker_events_buffer.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+var ignoredDockerEvents = []string{"rename", "kill", "destroy", "pause", "exec_create", "exec_start", "top", "attach", "export", "pull", "push", "tag", "untag", "import", "delete"}
+
+// InfiniteBuffer defines an unlimited buffer, where it reads from
+// input channel and write to output channel.
+type InfiniteBuffer struct {
+	events   []*docker.APIEvents
+	empty    bool
+	waitDone chan struct{}
```
aaithal: Can we use a `sync.Waitgroup` instead for this? It makes it simpler to understand the code.

richardpen:
From the benchmark test, it shows the latencies was reduced and the failure rate wasn't increased. Any other concerns?

agent/engine/docker_container_engine.go
```diff
@@ -42,6 +42,8 @@ const (
 	dockerDefaultTag = "latest"
 	// imageNameFormat is the name of a image may look like: repo:tag
 	imageNameFormat = "%s:%s"
+	// the buffer size will ensure agent doesn't miss any event from docker
```
vsiddharth: nit: `dockerEventBufferSize` is used to ensure ....

Issue #369, opened 2016-04-11
Allow local reservation of CPU resources


seiffert:
In order to limit the CPU resources that are offered to ECS tasks, I would like to discuss whether adding a configuration value `ECS_RESERVED_CPU` would make sense. This option would have an effect analogous to `ECS_RESERVED_MEMORY` and limit the CPU resources reported to ECS. 

For memory, the calculation is done here: (https://github.com/aws/amazon-ecs-agent/blob/304495bc6defdcc98f15a97b271c41127dba5e48/agent/api/api_client.go#L209)

```go
	cpu, mem := getCpuAndMemory()
	mem = mem - int64(client.config.ReservedMemory)
    
	[...]

	memResource := ecs.Resource{
		Name:         utils.Strptr("MEMORY"),
		Type:         &integerStr,
		IntegerValue: &mem,
	}
```

We currently face a situation where a complete reservation of an instance doesn't leave much CPU resources for system processes.
If you agree that such a change would make sense, I'm happy to provide a PR.

samuelkarp:
Hi @seiffert,

Thanks for offering to help out!  While we appreciate what you're offering, unfortunately I don't think creating a "reserved CPU" configuration will have the effect you intend.

The `cpu` field in the task definition ultimately maps to the `CpuShares` setting for the container (like the `--cpu-share` flag on `docker run`).  CPU shares are a relative resource share rather than being a fixed resource allocation or limit; this means that while we specify that there are 1024 CPU units per CPU core on the instance, a task that specifies less than 1024 still has access to the full 1024 when it is either the only task running or there is no CPU contention.  From our [documentation](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html):

> For example, if you run a single-container task on a single-core instance type with 512 CPU units specified for that container, and that is the only task running on the container instance, that container could use the full 1,024 CPU unit share at any given time. However, if you launched another copy of the same task on that container instance, each task would be guaranteed a minimum of 512 CPU units when needed, and each container could float to higher CPU usage if the other container was not using it, but if both tasks were 100% active all of the time, they would be limited to 512 CPU units.

[...]

PR #988, opened 2017-09-21
Support reserving CPU units for the system

rbranson:
Please provide the following information:

### Summary
Support for reserving CPU units for the system, very similar to reserved memory.

### Implementation details
ECS_RESERVED_CPU is a new environment var that accepts CPU units, which are subtracted from the hosts total resource amount. This patch also switches the internals of ECS_RESERVED_MEMORY to use a 32-bit integer (previously 16-bit) so that it can support a reservation of more than 64GB.

### Testing

* Added ECS_RESERVED_CPU for reserving CPU units for the system
* ECS_RESERVED_MEMORY now supports >64GB

### Licensing


rbranson:
This was discussed before in this issue: https://github.com/aws/amazon-ecs-agent/issues/369. @samuelkarp's description of the mechanics of CPU shares is right there. The aim with this is to reduce the amount of tasks that get scheduled on a container instance by ECS, not necessarily to make a hard CPU cut-out for non-ECS processes. It certainly isn't as good as support for Docker CPU quotas, but it's been over a year since that issue was closed, and ECS still doesn't support it.

jhaynes:
Hi @rbranson. Thank you for the contribution. However, we're worried that this change will not have the effect you're looking for. As @skarp mentioned in #369, CPU shares are relative, so decreasing the number presented to our back end will only have the effect of placing tasks with lower cpu requirements. It will not have any effect on the "noisy neighbor" problem presented by host processes.

That said, the work we have begun on nested [cgroups for tasks](https://github.com/aws/amazon-ecs-agent/tree/task-cgroup) is the first step toward CPU quotas, which will allow for reserving some CPU time for host processes.

samuelkarp:
While we appreciate the contribution, we're not intending to merge this at this time.

PR #989, opened 2017-09-21
Make container dependencies more explicit

samuelkarp:
### Summary
Containers in a task may depend on each other for a few reasons: links, shared volumes, "empty" volumes, and network namespace setup.  Prior to this change, "empty" volume and network namespace dependencies were very simplistic, blocking any transition (none->pulled, pulled->created, etc), resulting in some containers needing to move all the way into running before others could even be pulled.  This was hard to see as well, given the opaqueness with which the engine handles container state transitions.

This change seeks to make inter-container dependencies more explicit, with both specified blocked states (dependent states) and specifying when those dependencies are satisfied (satisfied states).

### Implementation details

The `api.Container` struct is extended with a new field `TransitionDependencies`.  The `TransitionDependencies` field is a struct with an slice of `ContainerDependency` structs.  `TransitionDependencies` is a wrapper struct that would enable us to add other non-container dependencies in the future.  `ContainerDependency` specifies a dependency container, the dependent status, and the status of the dependency container that satisfies this dependency.

The old `SteadyStateDependencies` slice is kept, as state files created with old agents will still have that fields.  I've left in code to handle `SteadyStateDependencies` as-is; tasks with that field can coexist with the new `TransitionDependencies` field, but new tasks use `TransitionDependencies` instead of `SteadyStateDependencies`.  Another alternative would be converting the old `SteadyStateDependencies` into `TransitionDependencies` during state load, but I think leaving the compatibility code in place is reasonable.

### Testing
- [x] Tested with an `awsvpc` task and observed pause container dependency
- [x] Tested with an "empty" volume task and observed "empty volume" container dependency

None, this isn't really user-facing right now

### Licensing


agent/api/container.go
```diff
@@ -85,11 +85,20 @@ type Container struct {
...
+	// that depends on the resource, and the state of the dependency that
+	// satisfies.
+	TransitionDependencySet TransitionDependencySet `json:"TransitionDependencySet"`
+
+	// Deprecated.
 	// SteadyStateDependencies is a list of containers that must be in "steady state" before
 	// this one is created
 	// Note: Current logic requires that the containers specified here are run
 	// before this container can even be pulled.
 	SteadyStateDependencies []string `json:"RunDependencies"`
```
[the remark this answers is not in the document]
samuelkarp: `SteadyStateDependencies` is now deprecated; `TransitionDependencySet` replaces it.  We retain `SteadyStateDependencies` for backwards-compatibility with old agent state files that may be read.

agent/engine/dependencygraph/graph_test.go
```diff
@@ -444,3 +444,90 @@ func assertResolved(f func(target *api.Container, dep *api.Container) bool, targ
...
+
+func TestTransitionDependenciesResolved(t *testing.T) {
+	testcases := []struct {
+		Name            string
+		TargetKnown     api.ContainerStatus
+		TargetDesired   api.ContainerStatus
+		DependencyKnown api.ContainerStatus
+		DependentStatus api.ContainerStatus
+		SatisfiedStatus api.ContainerStatus
+		Resolved        bool
```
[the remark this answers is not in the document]
samuelkarp: ExpectedResolved?

agent/api/transitiondependency.go
```diff
@@ -0,0 +1,21 @@
...
+// ContainerDependency defines the relationship between a dependent container
+// and its dependency.
+type ContainerDependency struct {
+	// Container defines the container on which a transition depends
+	Container string `json:"Container"`
+	// SatisfiedStatus defines the status that satisfies the dependency
+	SatisfiedStatus ContainerStatus `json:"SatisfiedStatus"`
+	// DependentStatus defines the status that cannot be reached until the
+	// resource satisfies the dependency
+	DependentStatus ContainerStatus `json:"DependentStatus"`
```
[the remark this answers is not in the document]
samuelkarp: Container A depends on container B.  Container A's `PULLED` status (dependent status) depends on container B's `RUNNING` status (satisfied status).

agent/api/container.go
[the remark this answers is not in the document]
petderek: Got it. For me, the reasoning isn't implicit. Can you change this to something like:

`// Deprecated -- required to deserialize older agent state files.`

agent/engine/dependencygraph/graph_test.go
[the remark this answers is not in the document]
petderek: Sure!

agent/api/transitiondependency.go
```diff
@@ -0,0 +1,21 @@
...
+	// ContainerDependencies is the set of containers on which a transition is
+	// dependent.
+	ContainerDependencies []ContainerDependency `json:"ContainerDependencies"`
+}
+
+// ContainerDependency defines the relationship between a dependent container
+// and its dependency.
+type ContainerDependency struct {
+	// Container defines the container on which a transition depends
+	Container string `json:"Container"`
```
richardpen: Would `ContainerName` be more accurate here, as `Container` is widely used as a container struct?

agent/engine/dependencygraph/graph.go
```diff
@@ -73,7 +73,8 @@ func dependenciesCanBeResolved(target *api.Container, by []*api.Container) bool
 	}
 
 	return verifyStatusResolvable(target, nameMap, neededVolumeContainers, volumeCanResolve) &&
-		verifyStatusResolvable(target, nameMap, linksToContainerNames(target.Links), linkCanResolve)
+		verifyStatusResolvable(target, nameMap, linksToContainerNames(target.Links), linkCanResolve) &&
+		verifyStatusResolvable(target, nameMap, target.SteadyStateDependencies, onSteadyStateCanResolve)
```
richardpen: Can this line be removed, as `SteadyStateDependencies` isn't updated anywhere? Also do we need to check the `TransitionDependencySet` here?

agent/api/task.go
```diff
@@ -251,10 +255,14 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
...
 		}
-		if container.SteadyStateDependencies == nil {
-			container.SteadyStateDependencies = make([]string, 0)
+		if container.TransitionDependencySet.ContainerDependencies == nil {
+			container.TransitionDependencySet.ContainerDependencies = make([]ContainerDependency, 0)
 		}
-		container.SteadyStateDependencies = append(container.SteadyStateDependencies, PauseContainerName)
+		container.TransitionDependencySet.ContainerDependencies = append(container.TransitionDependencySet.ContainerDependencies, ContainerDependency{
+			Container:       PauseContainerName,
+			SatisfiedStatus: ContainerRunning,
```
richardpen: If the `SatisfiedStatus ` means other containers can process after this container reached `SatisfiedStatus`, shouldn't this be `ContainerResourcesProvisioned`?

agent/engine/dependencygraph/graph.go
```diff
@@ -131,6 +133,46 @@ func verifyStatusResolvable(target *api.Container, existingContainers map[string
...
+}
+
+func resolvesContainerTransitionDependency(target *api.Container, resource *api.Container, dependency api.ContainerDependency) bool {
+	targetDesired := target.GetDesiredStatus()
+	if targetDesired < dependency.DependentStatus {
+		// not trying to reach dependent status
+		return true
+	}
+	targetKnown := target.GetKnownStatus()
+	if targetKnown >= dependency.DependentStatus {
```
richardpen: When will this happen, if this happened, isn't that a bug? If a container's dependency isn't satisfied, why it can process and has the `targetKnown >= dependency.DependentStatus`? Or you just want to check the knownstatus is >=stopped?

agent/engine/dependencygraph/graph.go
samuelkarp: No, `SteadyStateDependencies` are still set via old state files if we read them in.

We probably should check `TransitionDependencySet` here too, but I'd rather not block on that right now.

agent/engine/dependencygraph/graph.go
samuelkarp: If `targetKnown >= dependency.DependentStatus`, that means that the dependency was already satisfied in a previous transition and this just skips checking it again.  It's not a bug.

agent/api/task.go
samuelkarp: `SatisfiedStatus` means that this particular dependent container can progress.  Satisfied doesn't necessarily mean `api.ContainerResourcesProvisioned`; it could also be `api.ContainerRunning` (and could be different statuses in the future).

agent/api/transitiondependency.go
samuelkarp: Done

agent/api/task.go
```diff
@@ -140,10 +140,14 @@ func (task *Task) initializeEmptyVolumes() {
...
 			if _, ok := vol.(*EmptyHostVolume); ok {
-				if container.SteadyStateDependencies == nil {
-					container.SteadyStateDependencies = make([]string, 0)
+				if container.TransitionDependencySet.ContainerDependencies == nil {
+					container.TransitionDependencySet.ContainerDependencies = make([]ContainerDependency, 0)
 				}
-				container.SteadyStateDependencies = append(container.SteadyStateDependencies, emptyHostVolumeName)
+				container.TransitionDependencySet.ContainerDependencies = append(container.TransitionDependencySet.ContainerDependencies, ContainerDependency{
+					ContainerName:   emptyHostVolumeName,
+					SatisfiedStatus: ContainerRunning,
```
aaithal: Can this be `ContainerCreated` for the empty host volume container?

agent/api/task.go
```diff
@@ -251,10 +255,14 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
...
 		}
-		if container.SteadyStateDependencies == nil {
-			container.SteadyStateDependencies = make([]string, 0)
+		if container.TransitionDependencySet.ContainerDependencies == nil {
+			container.TransitionDependencySet.ContainerDependencies = make([]ContainerDependency, 0)
 		}
-		container.SteadyStateDependencies = append(container.SteadyStateDependencies, PauseContainerName)
+		container.TransitionDependencySet.ContainerDependencies = append(container.TransitionDependencySet.ContainerDependencies, ContainerDependency{
+			ContainerName:   PauseContainerName,
+			SatisfiedStatus: ContainerRunning,
```
aaithal: This should be `ContainerResourcesReady`

agent/engine/dependencygraph/graph_test.go
```diff
@@ -444,3 +444,90 @@ func assertResolved(f func(target *api.Container, dep *api.Container) bool, targ
...
+		{
+			Name:             "Dependency running, create depends on created",
+			TargetKnown:      api.ContainerPulled,
+			TargetDesired:    api.ContainerRunning,
+			DependencyKnown:  api.ContainerRunning,
+			DependentStatus:  api.ContainerCreated,
+			SatisfiedStatus:  api.ContainerCreated,
+			ExpectedResolved: true,
+		},
+		{
```
aaithal: Really appreciate these test cases. Can you also add some test cases for `SatisfiedStatus` being `api.ContainerResourcesReady`?

agent/api/task.go
samuelkarp: IIRC, it need to be `ContainerRunning` because old versions of Docker (1.5.x era) don't actually create the volumes until `Start` is called on the container.

aaithal: ah, let's leave it as-is then.

agent/api/task.go
samuelkarp: Done

agent/engine/dependencygraph/graph_test.go
samuelkarp: Done

PR #990, opened 2017-09-25
Handle pull for "internal" containers

samuelkarp:
### Summary
Internal containers should be treated differently during the pull step of the container lifecycle.

We currently have two types of internal containers: "empty host volume" containers and pause containers.  "Empty host volume" containers should either be pulled from a remote image (as on Windows) or imported from a locally-generated tarball (as on Linux).  The pause container image is either imported from a locally-embedded tarball inside the agent's rootfs or expected to already be present on the host (if the image is overridden via config).

A second commit here changes the empty host volume image on Windows from `microsoft/windowsservercore:latest` to `microsoft/nanoserver:latest`, as it's a significantly smaller image.

This addresses https://github.com/aws/amazon-ecs-agent/issues/729#issuecomment-331792234.

### Implementation details

New constant exposed in the `emptyvolume` package to control whether or not the image is present locally to be imported, or it should be pulled.  New logic in `DockerTaskEngine` to dispatch to different methods based on the container's `Type` field.

### Testing

* Bug - `microsoft/windowsservercore:latest` was not pulled on Windows under certain conditions

### Licensing


agent/engine/docker_task_engine_test.go
```diff
@@ -1515,3 +1517,75 @@ func TestCreateContainerOnAgentRestart(t *testing.T) {
...
+	container := &api.Container{
+		Type:  api.ContainerEmptyHostVolume,
+		Image: imageName,
+	}
+	task := &api.Task{
+		Containers: []*api.Container{container},
+	}
+
+	// Windows is special-cased here and just performs a normal pull
+	if runtime.GOOS == "windows" {
```
petderek: I'd prefer this test to be in platform specific files if the expectations are different.

agent/engine/docker_task_engine.go
```diff
@@ -508,6 +509,20 @@ func (engine *DockerTaskEngine) GetTaskByArn(arn string) (*api.Task, bool) {
 }
 
 func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Container) DockerContainerMetadata {
+	if container.Type == api.ContainerCNIPause {
+		// ContainerCNIPause images are managed at startup
+		return DockerContainerMetadata{}
```
petderek: This is redundant.

agent/engine/docker_task_engine.go
```diff
@@ -508,6 +509,20 @@ func (engine *DockerTaskEngine) GetTaskByArn(arn string) (*api.Task, bool) {
...
+	}
+	switch container.Type {
+	case api.ContainerCNIPause:
+		// ContainerCNIPause image are managed at startup
+		return DockerContainerMetadata{}
+	case api.ContainerEmptyHostVolume:
+		// ContainerEmptyHostVolume image is either local (must be imported) or remote (must be pulled)
+		if emptyvolume.LocalImage {
+			return engine.client.ImportLocalEmptyVolumeImage()
```
petderek: Did `create` better describe what was happening behind the scenes here?

agent/engine/docker_task_engine_test.go
samuelkarp: Yeah, wasn't sure about that since this is the only difference in all these tests.  I can move it though.

agent/engine/docker_task_engine.go
samuelkarp: What part is redundant?  The comment?  That explains *why* there's an early return.

jhaynes:
Can you edit `CHANGELOG.md` with the description of this change also?

agent/engine/docker_task_engine.go
samuelkarp: `Import` is the verb that is used in the Docker API.

agent/engine/docker_task_engine.go
petderek: The whole `if` block doesn't seem necessary. The `switch` block that runs the same check.

agent/engine/docker_task_engine_test.go
samuelkarp: > Have this check at the top of the function and do a t.Skip() if you don't want to make a separate Linux file?

There's both a Linux- and Windows-test, sharing most of the test structure except some of the expectations.

> You could also make this more OO instead of using platform constants.

Not sure I understand what you mean here.

agent/engine/docker_task_engine.go
samuelkarp: Oh!  Yeah...oops.  Will remove!

agent/engine/docker_task_engine_test.go
petderek: You branch based on a constant boolean that's defined for each platform. You could instead attach these functions to a struct with this configuration embedded.

The tests would then be driven by configuration instead of driven by platform.

samuelkarp:
@petderek @jhaynes Done, please take a look.

agent/engine/docker_task_engine_test.go
samuelkarp: I went and separated them into platform-specific files.

agent/engine/docker_container_engine.go
```diff
@@ -365,6 +358,27 @@ func (dg *dockerGoClient) pullImage(image string, authData *api.RegistryAuthenti
...
+		if err != nil {
+			wrapped = CreateEmptyVolumeError{err}
+		}
+		response <- DockerContainerMetadata{Error: wrapped}
+	}()
+	select {
+	case resp := <-response:
+		return resp
+	case <-timeout:
+		return DockerContainerMetadata{Error: &DockerTimeoutError{pullImageTimeout, "pulled"}}
```
vsiddharth: Can the `transition` string be a const ?

samuelkarp: Any particular reason why?  I generally like constants if there's string reuse, string comparison, or if it aids clarity.  I suppose reuse might apply here, but it's only used in two places right now.

vsiddharth: I was primarily concerned with reuse but it was more of a suggestion.

PR #991, opened 2017-09-26
ecscni: Re-implement removed AdditionalLocalRoutes

samuelkarp:
### Summary
Code handling AdditionalLocalRoutes was inadvertently removed in 10fb083ad7cb23c64e822459a1562d72f58908b7.  This commit adds it back, and adjusts the TestSetupNS unit test to explicitly check for it.

### Implementation details
Added back the code that actually uses the field.  Checked the value in the mock implementation.

### Testing
- [x] Observed additional routes being properly added to the network namespace

None

### Licensing

### DIFF
--- a/agent/ecscni/plugin.go
+++ b/agent/ecscni/plugin.go
@@ -94,7 +94,7 @@ func (client *cniClient) CleanupNS(cfg *Config) error {
 		IfName:      defaultEthName,
 	}
 
-	// clean up the network namespace is separate from releasing the ip from ipam
+	// clean up the network namespace is separate from releasing the IP from IPAM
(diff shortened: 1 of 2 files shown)


agent/ecscni/plugin.go
```diff
@@ -240,17 +241,24 @@ func (client *cniClient) createIPAMConfig(cfg *Config) (IPAMConfig, error) {
 		return IPAMConfig{}, err
 	}
 
+	routes := []*cnitypes.Route{
+		{
+			Dst: *dst,
+		},
+	}
+	for _, route := range cfg.AdditionalLocalRoutes {
+		seelog.Debugf("Adding an additional route for %s", route)
```
aaithal: nit: I think there's a `String()` that we should be able to use for this.

samuelkarp: No, there's no `String()`.

PR #993, opened 2017-09-27
Websocket timeout

aaithal:
### Summary

Implements some websocket connection management improvements. 

### Implementation details

1. Set read and write deadlines for websocket ReadMessage and
    WriteMessage operations. This is to ensure that these methods
    do not hang and result in io timeout if there's issues with
    the connection
2. Reduce the scope of the lock in the Connect() method. The
    lock was being held for the length of Connect() method, which
    meant that it wouldn't be relnquished if there was any delay
    in establishing the connection. The scope of the lock has now
    been reduced to just accessing the cs.conn variable
3. Start ACS heartbeat timer after the connection has been
    established. The timer was being started before a call to
    Connect, which meant that the connection could be prematurely
    terminated for being idle if there was a delay in establishing
    the connection

### Testing

- [x] Manual tests
  - [x] Reconnects to ACS and TCS work as expected when ACS/TCS initiate disconnect
  - [x] Reconnects to ACS and TCS work as expected when Agent initiates disconnect

Covered in changelog.md

### Licensing

### DIFF
--- a/agent/acs/client/acs_client.go
+++ b/agent/acs/client/acs_client.go
@@ -20,14 +20,13 @@ package acsclient
 
 import (
 	"errors"
+	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/config"
-	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/wsclient"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/cihub/seelog"
 )
-
-var log = logger.ForModule("acs client")
 
 // clientServer implements ClientServer for acs.
 type clientServer struct {
@@ -37,7 +36,7 @@ type clientServer struct {
 // New returns a client/server to bidirectionally communicate with ACS
 // The returned struct should have both 'Connect' and 'Serve' called upon it
 // before being used.
-func New(url string, cfg *config.Config, credentialProvider *credentials.Credentials) wsclient.ClientServer {
+func New(url string, cfg *config.Config, credentialProvider *credentials.Credentials, rwTimeout time.Duration) wsclient.ClientServer {
 	cs := &clientServer{}
 	cs.URL = url
 	cs.CredentialProvider = credentialProvider
@@ -45,6 +44,7 @@ func New(url string, cfg *config.Config, credentialProvider *credentials.Credent
 	cs.ServiceError = &acsError{}
 	cs.RequestHandlers = make(map[string]wsclient.RequestHandler)
 	cs.TypeDecoder = NewACSDecoder()
+	cs.RWTimeout = rwTimeout
 	return cs
 }
 
@@ -52,9 +52,9 @@ func New(url string, cfg *config.Config, credentialProvider *credentials.Credent
 // AddRequestHandler). All request handlers should be added prior to making this
 // call as unhandled requests will be discarded.
 func (cs *clientServer) Serve() error {
-	log.Debug("Starting websocket poll loop")
+	seelog.Debug("ACS client starting websocket poll loop")
 	if !cs.IsReady() {
-		return errors.New("Websocket not ready for connections")
+		return errors.New("acs client: websocket not ready for connections")
 	}
 	return cs.ConsumeMessages()
 }
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -246,19 +246,13 @@ func (acsSession *session) startSessionOnce() error {
 	client := acsSession.resources.createACSClient(url, acsSession.agentConfig)
 	defer client.Close()
 
-	// Start inactivity timer for closing the connection
-	timer := newDisconnectionTimer(client, acsSession.heartbeatTimeout(), acsSession.heartbeatJitter())
-	defer timer.Stop()
-
-	return acsSession.startACSSession(client, timer)
+	return acsSession.startACSSession(client)
 }
 
 // startACSSession starts a session with ACS. It adds request handlers for various
 // kinds of messages expected from ACS. It returns on server disconnection or when
 // the context is cancelled
-func (acsSession *session) startACSSession(client wsclient.ClientServer, timer ttime.Timer) error {
-	// Any message from the server resets the disconnect timeout
-	client.SetAnyRequestHandler(anyMessageHandler(timer))
+func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 	cfg := acsSession.agentConfig
 
 	refreshCredsHandler := newRefreshCredentialsHandler(acsSession.ctx, cfg.Cluster, acsSession.containerInstanceARN,
@@ -313,6 +307,11 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer, timer t
 		return err
 	}
 	seelog.Info("Connected to ACS endpoint")
+	// Start inactivity timer for closing the connection
+	timer := newDisconnectionTimer(client, acsSession.heartbeatTimeout(), acsSession.heartbeatJitter())
+	// Any message from the server resets the disconnect timeout
+	client.SetAnyRequestHandler(anyMessageHandler(timer, client))
+	defer timer.Stop()
 
 	acsSession.resources.connectedToACS()
 
@@ -377,7 +376,7 @@ func (acsSession *session) heartbeatJitter() time.Duration {
 
 // createACSClient creates the ACS Client using the specified URL
 func (acsResources *acsSessionResources) createACSClient(url string, cfg *config.Config) wsclient.ClientServer {
-	return acsclient.New(url, cfg, acsResources.credentialsProvider)
+	return acsclient.New(url, cfg, acsResources.credentialsProvider, heartbeatTimeout+heartbeatJitter)
 }
 
 // connectedToACS records a successful connection to ACS
@@ -424,9 +423,8 @@ func acsWsURL(endpoint, cluster, containerInstanceArn string, taskEngine engine.
 func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration, jitter time.Duration) ttime.Timer {
 	timer := time.AfterFunc(utils.AddJitter(timeout, jitter), func() {
 		seelog.Warn("ACS Connection hasn't had any activity for too long; closing connection")
-		closeErr := client.Close()
-		if closeErr != nil {
-			seelog.Warnf("Error disconnecting: %v", closeErr)
+		if err := client.Close(); err != nil {
+			seelog.Warnf("Error disconnecting: %v", err)
 		}
 	})
 
@@ -435,9 +433,15 @@ func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration, 
 
 // anyMessageHandler handles any server message. Any server message means the
 // connection is active and thus the heartbeat disconnect should not occur
-func anyMessageHandler(timer ttime.Timer) func(interface{}) {
+func anyMessageHandler(timer ttime.Timer, client wsclient.ClientServer) func(interface{}) {
 	return func(interface{}) {
-		seelog.Debug("ACS activity occured")
+		seelog.Debug("ACS activity occurred")
+		// Reset read deadline as there's activity on the channel
+		if err := client.SetReadDeadline(time.Now().Add(heartbeatTimeout + heartbeatJitter)); err != nil {
+			seelog.Warn("Unable to extend read deadline for ACS connection: %v", err)
+		}
+
+		// Reset heearbeat timer
 		timer.Reset(utils.AddJitter(heartbeatTimeout, heartbeatJitter))
 	}
 }
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -48,6 +48,9 @@ const (
 	// without disconnecting
 	heartbeatTimeout = 1 * time.Minute
 	heartbeatJitter  = 1 * time.Minute
+	// wsRWTimeout is the duration of read and write deadline for the
+	// websocket connection
+	wsRWTimeout = 2*heartbeatTimeout + heartbeatJitter
 
 	inactiveInstanceReconnectDelay = 1 * time.Hour
 
@@ -376,7 +379,7 @@ func (acsSession *session) heartbeatJitter() time.Duration {
 
 // createACSClient creates the ACS Client using the specified URL
 func (acsResources *acsSessionResources) createACSClient(url string, cfg *config.Config) wsclient.ClientServer {
-	return acsclient.New(url, cfg, acsResources.credentialsProvider, heartbeatTimeout+heartbeatJitter)
+	return acsclient.New(url, cfg, acsResources.credentialsProvider, wsRWTimeout)
 }
 
 // connectedToACS records a successful connection to ACS
@@ -437,11 +440,11 @@ func anyMessageHandler(timer ttime.Timer, client wsclient.ClientServer) func(int
 	return func(interface{}) {
 		seelog.Debug("ACS activity occurred")
 		// Reset read deadline as there's activity on the channel
-		if err := client.SetReadDeadline(time.Now().Add(heartbeatTimeout + heartbeatJitter)); err != nil {
-			seelog.Warn("Unable to extend read deadline for ACS connection: %v", err)
+		if err := client.SetReadDeadline(time.Now().Add(wsRWTimeout)); err != nil {
+			seelog.Warnf("Unable to extend read deadline for ACS connection: %v", err)
 		}
 
-		// Reset heearbeat timer
+		// Reset hearbeat timer
 		timer.Reset(utils.AddJitter(heartbeatTimeout, heartbeatJitter))
 	}
 }
--- a/agent/acs/handler/acs_handler_test.go
+++ b/agent/acs/handler/acs_handler_test.go
@@ -804,9 +804,7 @@ func TestConnectionIsClosedOnIdle(t *testing.T) {
 		_heartbeatJitter:     10 * time.Millisecond,
 	}
 	go func() {
-		timer := newDisconnectionTimer(mockWsClient, acsSession.heartbeatTimeout(), acsSession.heartbeatJitter())
-		defer timer.Stop()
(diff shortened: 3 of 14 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -435,9 +433,15 @@ func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration,
...
 // anyMessageHandler handles any server message. Any server message means the
 // connection is active and thus the heartbeat disconnect should not occur
-func anyMessageHandler(timer ttime.Timer) func(interface{}) {
+func anyMessageHandler(timer ttime.Timer, client wsclient.ClientServer) func(interface{}) {
 	return func(interface{}) {
-		seelog.Debug("ACS activity occured")
+		seelog.Debug("ACS activity occurred")
+		// Reset read deadline as there's activity on the channel
+		if err := client.SetReadDeadline(time.Now().Add(heartbeatTimeout + heartbeatJitter)); err != nil {
+			seelog.Warn("Unable to extend read deadline for ACS connection: %v", err)
```
richardpen: nit: `Warnf`

agent/tcs/handler/handler.go
```diff
@@ -112,7 +118,7 @@ func startSession(url string, cfg *config.Config, credentialProvider *credential
 	timer := time.AfterFunc(utils.AddJitter(heartbeatTimeout, heartbeatJitter), func() {
```
richardpen: Can you do the same as acs to move the timer after the connection was established?

agent/tcs/handler/handler.go
```diff
@@ -35,8 +36,8 @@ const (
 	defaultPublishMetricsInterval = 20 * time.Second
 
 	// The maximum time to wait between heartbeats without disconnecting
-	defaultHeartbeatTimeout            = 5 * time.Minute
-	defaultHeartbeatJitter             = 3 * time.Minute
+	defaultHeartbeatTimeout            = 1 * time.Minute
```
samuelkarp: Why did we change these?

agent/tcs/handler/handler.go
```diff
@@ -144,6 +151,18 @@ func ackPublishMetricHandler(timer *time.Timer) func(*ecstcs.AckPublishMetric) {
 	}
 }
 
+// anyMessageHandler handles any server message. Any server message means the
+// connection is active
+func anyMessageHandler(client wsclient.ClientServer) func(interface{}) {
+	return func(interface{}) {
+		seelog.Trace("TCS activity occured")
+		// Reset read deadline as there's activity on the channel
+		if err := client.SetReadDeadline(time.Now().Add(defaultHeartbeatTimeout + defaultHeartbeatJitter)); err != nil {
```
samuelkarp: As above, I think this should be longer than just timeout + jitter.

agent/acs/handler/acs_handler.go
```diff
@@ -377,7 +376,7 @@ func (acsSession *session) heartbeatJitter() time.Duration {
 
 // createACSClient creates the ACS Client using the specified URL
 func (acsResources *acsSessionResources) createACSClient(url string, cfg *config.Config) wsclient.ClientServer {
-	return acsclient.New(url, cfg, acsResources.credentialsProvider)
+	return acsclient.New(url, cfg, acsResources.credentialsProvider, heartbeatTimeout+heartbeatJitter)
```
samuelkarp: I think this should be longer than timeout + jitter.

agent/tcs/handler/handler.go
```diff
@@ -94,10 +95,15 @@ func startTelemetrySession(params TelemetrySessionParams, statsEngine stats.Engi
...
+func startSession(url string,
+	cfg *config.Config,
+	credentialProvider *credentials.Credentials,
+	statsEngine stats.Engine,
+	heartbeatTimeout, heartbeatJitter,
+	publishMetricsInterval time.Duration,
 	deregisterInstanceEventStream *eventstream.EventStream) error {
-	client := tcsclient.New(url, cfg, credentialProvider, statsEngine, publishMetricsInterval)
+	client := tcsclient.New(url, cfg, credentialProvider, statsEngine,
+		publishMetricsInterval, defaultHeartbeatTimeout+defaultHeartbeatJitter)
```
samuelkarp: I think this should be longer than timeout + jitter.

agent/tcs/handler/handler.go
```diff
@@ -144,6 +151,18 @@ func ackPublishMetricHandler(timer *time.Timer) func(*ecstcs.AckPublishMetric) {
 	}
 }
 
+// anyMessageHandler handles any server message. Any server message means the
+// connection is active
+func anyMessageHandler(client wsclient.ClientServer) func(interface{}) {
+	return func(interface{}) {
+		seelog.Trace("TCS activity occured")
```
richardpen: Is the `Trace` level here intentionally?

agent/wsclient/client.go
```diff
@@ -48,6 +49,9 @@ const (
 	// wsConnectTimeout specifies the default connection timeout to the backend.
 	wsConnectTimeout = 30 * time.Second
 
+	// wsHandshakeTimeout specifies the default handshake timeout for the websocket client
+	wsHandshakeTimeout = wsConnectTimeout
```
richardpen: If we have wsHandshakeTimeout, do we still need wsConnectTimeout? As if the hand shake timed out, the connect will fail, is that right? Or they should be the same value, eg: `wsHandshakeTimeout 
 = 3*wsConnectTimeout`

agent/wsclient/client.go
```diff
@@ -198,35 +207,56 @@ func (cs *ClientServerImpl) Connect() error {
...
+// SetConnection passes a websocket connection object into the client. This is used only in
+// testing and should be avoided in non-test code.
 func (cs *ClientServerImpl) SetConnection(conn WebsocketConn) {
 	cs.conn = conn
 }
 
+// SetReadDeadline sets the read deadline for the websocket connection
+// A read timeout results in an io error if there are any outstanding reads
+// that exceed the deadline
+func (cs *ClientServerImpl) SetReadDeadline(t time.Time) error {
```
richardpen: Would it be better to pass the duration and calculate the time right before calling the `SetReadDeadline`, it's more accurate, as the passed in time could already beyond time.Now() for some wired reason.

agent/acs/client/acs_client.go
```diff
@@ -20,15 +20,14 @@ package acsclient
 
 import (
 	"errors"
+	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/config"
-	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/wsclient"
 	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/cihub/seelog"
```
jhaynes: :+1:

agent/acs/client/acs_client.go
```diff
@@ -37,24 +36,25 @@ type clientServer struct {
...
-func New(url string, cfg *config.Config, credentialProvider *credentials.Credentials) wsclient.ClientServer {
+func New(url string, cfg *config.Config, credentialProvider *credentials.Credentials, rwTimeout time.Duration) wsclient.ClientServer {
 	cs := &clientServer{}
 	cs.URL = url
 	cs.CredentialProvider = credentialProvider
 	cs.AgentConfig = cfg
 	cs.ServiceError = &acsError{}
 	cs.RequestHandlers = make(map[string]wsclient.RequestHandler)
 	cs.TypeDecoder = NewACSDecoder()
+	cs.RWTimeout = rwTimeout
```
petderek: nit:
why not just 'timeout'?

agent/acs/handler/acs_handler.go
jhaynes: Can you expand on why? I would assume you'd want it to be smaller if anything. That way, you could retry a read/write without necessarily having to re-establish the whole connection.

agent/acs/handler/acs_handler.go
```diff
@@ -424,9 +423,8 @@ func acsWsURL(endpoint, cluster, containerInstanceArn string, taskEngine engine.
 func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration, jitter time.Duration) ttime.Timer {
 	timer := time.AfterFunc(utils.AddJitter(timeout, jitter), func() {
 		seelog.Warn("ACS Connection hasn't had any activity for too long; closing connection")
-		closeErr := client.Close()
-		if closeErr != nil {
-			seelog.Warnf("Error disconnecting: %v", closeErr)
+		if err := client.Close(); err != nil {
```
petderek: There is only one return value.. so is this still more idiomatic than:
`if client.Close() != nil {`

agent/acs/handler/acs_handler.go
```diff
@@ -435,9 +433,15 @@ func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration,
 
 // anyMessageHandler handles any server message. Any server message means the
 // connection is active and thus the heartbeat disconnect should not occur
-func anyMessageHandler(timer ttime.Timer) func(interface{}) {
+func anyMessageHandler(timer ttime.Timer, client wsclient.ClientServer) func(interface{}) {
 	return func(interface{}) {
-		seelog.Debug("ACS activity occured")
+		seelog.Debug("ACS activity occurred")
```
petderek: Is this debug line useful?

agent/acs/handler/acs_handler.go
```diff
@@ -435,9 +433,15 @@ func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration,
...
+		// Reset read deadline as there's activity on the channel
+		if err := client.SetReadDeadline(time.Now().Add(heartbeatTimeout + heartbeatJitter)); err != nil {
```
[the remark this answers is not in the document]
jhaynes: Can you explain again here too? Doesn't doubling that have the net effect of doubling your pre-established `heartbeatTimeout` the first time there is some traffic over the websocket?

agent/acs/handler/acs_handler_test.go
```diff
@@ -804,9 +804,7 @@ func TestConnectionIsClosedOnIdle(t *testing.T) {
 		_heartbeatJitter:     10 * time.Millisecond,
 	}
 	go func() {
-		timer := newDisconnectionTimer(mockWsClient, acsSession.heartbeatTimeout(), acsSession.heartbeatJitter())
-		defer timer.Stop()
-		acsSession.startACSSession(mockWsClient, timer)
+		acsSession.startACSSession(mockWsClient)
```
petderek: nit: remove the function wrapper
 `go acsSession.startACSSession(mockWsClient)`

agent/tcs/client/client.go
```diff
@@ -46,7 +46,12 @@ type clientServer struct {
 // New returns a client/server to bidirectionally communicate with the backend.
 // The returned struct should have both 'Connect' and 'Serve' called upon it
 // before being used.
-func New(url string, cfg *config.Config, credentialProvider *credentials.Credentials, statsEngine stats.Engine, publishMetricsInterval time.Duration) wsclient.ClientServer {
+func New(url string,
+	cfg *config.Config,
```
jhaynes: :+1: I much prefer this style for functions with more than a few arguments.

agent/tcs/client/client.go
```diff
@@ -67,11 +73,11 @@ func New(url string, cfg *config.Config, credentialProvider *credentials.Credent
 func (cs *clientServer) Serve() error {
 	seelog.Debug("TCS client starting websocket poll loop")
 	if !cs.IsReady() {
-		return fmt.Errorf("Websocket not ready for connections")
+		return fmt.Errorf("tcs client: websocket not ready for connections")
```
jhaynes: Should TCS be capitalized here? We capitalize ACS in most places.

agent/tcs/handler/handler.go
```diff
@@ -35,8 +36,8 @@ const (
 	defaultPublishMetricsInterval = 20 * time.Second
 
 	// The maximum time to wait between heartbeats without disconnecting
```
petderek: Can we go ahead and add the default rwtimeout here too? even if expresed as:

`defaultRWTimeout = defaultHeartBeatJitter + defaultHeartbeatTimeout`

agent/tcs/client/client.go
```diff
@@ -67,11 +73,11 @@ func New(url string, cfg *config.Config, credentialProvider *credentials.Credent
...
 	}
 
 	if cs.statsEngine == nil {
-		return fmt.Errorf("uninitialized stats engine")
+		return fmt.Errorf("tcs client: uninitialized stats engine")
```
jhaynes: As above. Probably should be TCS.

agent/tcs/handler/handler.go
jhaynes: I assume this is just meant to mimic the work that @richardpen did for the ACS client.

agent/tcs/handler/handler.go
```diff
@@ -112,7 +118,7 @@ func startSession(url string, cfg *config.Config, credentialProvider *credential
 	timer := time.AfterFunc(utils.AddJitter(heartbeatTimeout, heartbeatJitter), func() {
 		// Close the connection if there haven't been any messages received from backend
 		// for a long time.
-		seelog.Debug("TCS Connection hasn't had a heartbeat or an ack message in too long of a timeout; disconnecting")
+		seelog.Info("TCS Connection hasn't had a heartbeat or an ack message in too long of a timeout; disconnecting")
```
jhaynes: This wording is confusing. The similar ACS log message is: "ACS Connection hasn't had any activity for too long; closing connection"

agent/tcs/handler/handler.go
jhaynes: The similar ACS log line is `Debug` level.

agent/wsclient/client.go
jhaynes: I would assume the `wsHandshakeTimeout` would be <= `wsConectTimeout`

agent/tcs/handler/handler.go
aaithal: @samuelkarp @jhaynes is correct. This is reflecting the change that @richardpen did.

agent/acs/handler/acs_handler.go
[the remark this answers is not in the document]
aaithal: Because we use [`rand.Int63n()` in `AddJitter`](https://github.com/aws/amazon-ecs-agent/blob/master/agent/utils/backoff.go#L76), `heartbeatTimeout + heartbeatJitter` should always be greater than the heartbeat duration that we compute. I think it should be safe to set it the way it's currently being set.

agent/tcs/handler/handler.go
aaithal: Yeah, I thought I did that. But, obviously didn't. I'll modify this.

agent/tcs/handler/handler.go
aaithal: TCS channel is more chatty, which means we'll see this printed every 20 seconds if this is changed to `Debug`. I could get rid of this altogether if you have a strong preference here. But, somehow this Trace soothes my mind.

agent/tcs/handler/handler.go
samuelkarp: Got it.

agent/acs/client/acs_client.go
aaithal: I wanted it to convey more meaning that just `timeout` would convey.

agent/acs/handler/acs_handler.go
aaithal: how is that? I don't see how `err := something; if err != nil {}` is more idiomatic than `if err := something; err != nil {}`

agent/acs/handler/acs_handler.go
aaithal: Yeah, I think so. Shows us when the timer was reset.

agent/acs/handler/acs_handler_test.go
aaithal: will do.

agent/tcs/handler/handler.go
aaithal: Yeah, I thought about and didn't think we needed a new const for this. But, I think it makes it more explicit. I'll add that.

agent/tcs/client/client.go
aaithal: We're sticking to the style of error strings beginning with a lower case letter.

agent/tcs/handler/handler.go
aaithal: I'll modify that.

agent/acs/handler/acs_handler.go
samuelkarp: Will respond on the other comment.

agent/acs/handler/acs_handler.go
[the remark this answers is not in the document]
samuelkarp: >  Doesn't doubling that have the net effect of doubling your pre-established `heartbeatTimeout` the first time there is some traffic over the websocket?

No, the deadline is set based on `time.Now()`, so it's always in the future by the amount specified here.

> I think it should be safe to set it the way it's currently being set.

I'm worried, because this is a time and not a duration, and because Go 1.7 doesn't have a monotonic clock, that something unexpected can happen here.  I'd feel more comfortable with a longer timeout because we don't expect this timeout to be hit anyway and I don't see any danger of making it longer.

Once we move to Go 1.9 there will be a monotonic clock and this should be safer.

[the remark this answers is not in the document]
aaithal: > I'd feel more comfortable with a longer timeout because we don't expect this timeout to be hit anyway

Oh, you'd be surprised :) I think we'll hit this timeout when there's connection issues. However, I do buy the argument that `2*heartbeatTimeout + heartbeatJitter` will work for most cases here.

agent/acs/handler/acs_handler.go
petderek: Right! I was asking if setting the `err` variable is necessary at all.  On closer inspection I see that we log the error, so you should ignore this :)

aaithal:
@samuelkarp @jhaynes @richardpen @petderek I've modified the code to address your comments. Can you please take a look?

PR #996, opened 2017-09-28
engine/dockerclient: Add new DockerClient version support

sharanyad:
### Summary
Add new Docker API version 1.25 support

### Implementation details
Added 1.25 to supported agent version for init flag support.

### Testing

New tests cover the changes: 

Support for Docker Client version 1.25

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,9 @@
 ## UNRELEASED
 * Feature - Support for provisioning Tasks with ENIs
 * Bug - Fixed a bug where unsupported Docker API client versions could be registered
-[#1014](https://github.com/aws/amazon-ecs-agent/pull/1014)
+  [#1014](https://github.com/aws/amazon-ecs-agent/pull/1014)
+* Enhancement - Support `init` process in containers by adding support for Docker remote API client version 1.25
+  [#996](https://github.com/aws/amazon-ecs-agent/pull/996)
 
 ## 1.14.5
 * Enhancement - Retry failed container image pull operations [#975](https://github.com/aws/amazon-ecs-agent/pull/975)
--- a/agent/engine/dockerclient/versionsupport_unix.go
+++ b/agent/engine/dockerclient/versionsupport_unix.go
@@ -38,6 +38,7 @@ func getAgentVersions() []DockerVersion {
 		Version_1_22,
 		Version_1_23,
 		Version_1_24,
+		Version_1_25,
 	}
 }
 


agent/app/agent_capability.go
```diff
@@ -63,7 +66,12 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 	// Determine API versions to report as supported. Supported versions are also used for capability-enablement, except
 	// logging drivers.
 	for _, version := range agent.dockerClient.SupportedVersions() {
-		capabilities = appendNameOnlyAttribute(capabilities, capabilityPrefix+"docker-remote-api."+string(version))
+		dockerVersion, _ := strconv.ParseFloat(string(version), 64)
+		if dockerVersion >= dockerVersionUsingNewCapabilityPrefix {
```
samuelkarp: Why are we doing this?

agent/app/agent_capability.go
```diff
@@ -63,7 +66,12 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 	// Determine API versions to report as supported. Supported versions are also used for capability-enablement, except
 	// logging drivers.
 	for _, version := range agent.dockerClient.SupportedVersions() {
-		capabilities = appendNameOnlyAttribute(capabilities, capabilityPrefix+"docker-remote-api."+string(version))
+		dockerVersion, _ := strconv.ParseFloat(string(version), 64)
```
samuelkarp: We should probably not parse this as a float.  1.2 is not greater than 1.11.

agent/app/agent_capability.go
sharanyad: I believe the new agent capabilities have the "ecs.capability" prefix. Shouldn't we maintain the same for new remote api versions too?

agent/app/agent_capability.go
sharanyad: aah yes, will change it.

agent/app/agent_capability.go
samuelkarp: I think we should leave all the api version capabilities consistent with each other.

samuelkarp:
Have you verified that the vendored copy of go-dockerclient has all known fields for API version 1.25 in the `Config` and `HostConfig` structs?

sharanyad:
Yes, it does have all the fields in `Config` and `HostConfig`.

agent/app/agent_capability.go
```diff
@@ -63,7 +64,11 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 	// Determine API versions to report as supported. Supported versions are also used for capability-enablement, except
 	// logging drivers.
 	for _, version := range agent.dockerClient.SupportedVersions() {
-		capabilities = appendNameOnlyAttribute(capabilities, capabilityPrefix+"docker-remote-api."+string(version))
+		if dockerLegacyCapabilityPrefix[string(version)] {
```
petderek: This feels like it will be confusing from an administrative perspective. Can we insist that we keep all docker version capabilities using the old style for attributes?

sharanyad:
Revert back to using old capability prefix for Docker remote API versions to maintain uniformity.

aaithal:
@sharanyad doesn't this also need an update to the go-dockerclient lib? If so, please mention that in your PR so that this gets merged only after https://github.com/aws/amazon-ecs-agent/pull/987.

sharanyad:
@aaithal Thanks for bringing it up. Yes, this does depend on https://github.com/aws/amazon-ecs-agent/pull/987

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## UNRELEASED
 * Feature - Support for provisioning Tasks with ENIs
+* Enhancement - Add support for Docker remote API client version 1.25 [#996](https://github.com/aws/amazon-ecs-agent/pull/996)
```
aaithal: Can you reword this make this message more relevant to customers? Something like:

* Feature - Support running init process in containers by adding support for Docker remote API client version 1.25 [#996](https://github.com/aws/amazon-ecs-agent/pull/996)

sharanyad:
Fixed issue https://github.com/aws/amazon-ecs-agent/issues/1010 that was a blocker for this PR.

PR #998, opened 2017-09-29
Don't log websocket messages if there are errors.

jhaynes:
The messages that come over the websockets can potentially contain sensitive
information that shouldn't be put in logs. Separately, if reading the message
results in an error, the content of the message is irrelevant and unreliable.

### Summary
When there is a websocket connection error, we shouldn't log the message because we have no way of knowing what it contains.

### Implementation details
Removed the `message` portion of the log line.

### Testing

NA

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -214,7 +214,7 @@ func (acsSession *session) Start() error {
 				// Disconnected unexpectedly from ACS, compute backoff duration to
 				// reconnect
 				reconnectDelay := acsSession.computeReconnectDelay(isInactiveInstance)
-				seelog.Debugf("Reconnecting to ACS in: %v", reconnectDelay)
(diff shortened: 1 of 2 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -214,7 +214,7 @@ func (acsSession *session) Start() error {
 				// Disconnected unexpectedly from ACS, compute backoff duration to
 				// reconnect
 				reconnectDelay := acsSession.computeReconnectDelay(isInactiveInstance)
-				seelog.Debugf("Reconnecting to ACS in: %v", reconnectDelay)
+				seelog.Infof("Reconnecting to ACS in: %v", reconnectDelay)
```
petderek: Will this include the units?

jhaynes: It will now!

samuelkarp:
:+1:

PR #1004, opened 2017-10-05
ecs_client/model, functional_tests: updated model, functional tests for devices and init fields

sharanyad:
### Summary
Model changes to Container Definition for enabling devices and init for Linux containers.
Functional tests

### Implementation details
Functional tests that verify:
* If device is added to the container with the right cgroup permissions
* If init flag is enabled, then init process is present in the container

### Testing

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/devices/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/devices/task-definition.json
@@ -1,0 +1,19 @@
+{
+  "family": "ecsftest-devices",
+  "containerDefinitions": [{
+    "image": "127.0.0.1:51670/ubuntu:latest",
+    "name": "exit",
+    "cpu": 10,
+    "memory": 10,
+    "linuxParameters": {
+      "devices":[
+        {
+          "hostPath": "/dev/xvda",
+          "containerPath": "/dev/sda",
+          "permissions": ["read"]
+        }
+      ]
+    },
+    "command": ["sh", "-c", "if ls /dev/sda && ! fdisk /dev/sda; then exit 42; else exit 1; fi"]
+  }]
+}
--- a/agent/functional_tests/testdata/taskdefinitions/init-process/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/init-process/task-definition.json
@@ -1,0 +1,13 @@
+{
+  "family": "ecsftest-init-process",
+  "containerDefinitions": [{
+    "image": "127.0.0.1:51670/ubuntu:latest",
+    "name": "exit",
+    "cpu": 10,
+    "memory": 10,
+    "linuxParameters": {
+      "initProcessEnabled":true
+    },
+    "command": ["sh", "-c", "if pidof init == 1; then exit 42; else exit 1; fi"]
+  }]
+}
--- a/agent/ecs_client/model/api/api-2.json
+++ b/agent/ecs_client/model/api/api-2.json
@@ -552,6 +552,18 @@
       "type":"integer",
       "box":true
     },
+    "DeviceCgroupPermission":{
+      "type":"string",
+      "enum":[
+        "read",
+        "write",
+        "mknod"
+      ]
+    },
+    "DeviceCgroupPermissions":{
+      "type":"list",
+      "member":{"shape":"DeviceCgroupPermission"}
+    },
     "ClientException":{
       "type":"structure",
       "members":{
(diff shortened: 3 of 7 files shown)


sharanyad:
This PR depends on the merge of https://github.com/aws/amazon-ecs-agent/pull/996

sharanyad:
Referenced the issues in the commit message.
I mentioned it as `WIP` as there will be more changes added to this PR for the documentation of new fields and change in required agent version for the tests. I would like to get the tests reviewed before that.

agent/functional_tests/testdata/taskdefinitions/devices/task-definition.json
```diff
@@ -0,0 +1,19 @@
...
+    "linuxParameters": {
+      "devices":[
+        {
+          "hostPath": "/dev/xvda",
+          "containerPath": "/dev/sda",
+          "permissions": ["read"]
+        }
+      ]
+    },
+    "command": ["sh", "-c", "if ls /dev/sda && ! fdisk /dev/sda; then exit 42; else exit 1; fi"]
```
aaithal: Can we not do `fdisk` here? Especially because you are mounting `/dev/xvda` into the container?  Or at least `fdisk` with no flags? `fdisk -l` seems reasonable. Or any of `lsblk` or `blockdev`?

agent/functional_tests/testdata/taskdefinitions/init-process/task-definition.json
```diff
@@ -0,0 +1,13 @@
...
+  "family": "ecsftest-init-process",
+  "containerDefinitions": [{
+    "image": "127.0.0.1:51670/ubuntu:latest",
+    "name": "exit",
+    "cpu": 10,
+    "memory": 10,
+    "linuxParameters": {
+      "initProcessEnabled":true
+    },
+    "command": ["sh", "-c", "ps -x | grep \"/dev/init\" && exit 42 || exit 1"]
```
aaithal: This will return false positives I think. Can we do `pidof "/dev/init" == 1`  check instead?

agent/functional_tests/testdata/taskdefinitions/devices/task-definition.json
sharanyad: the `fdisk` call is supposed to fail here, since only `read` permission is added to the device in container. `-l` flag can also be used here.

agent/functional_tests/testdata/taskdefinitions/init-process/task-definition.json
sharanyad: `pidof "/dev/init" == 1` is a better check. will change this.

agent/functional_tests/testdata/taskdefinitions/devices/task-definition.json
aaithal: Right, but it'll also fail if the device wasn't mounted. You need a check to ensure that it's mounted as well. Can't you check the permission with something like `stat --format=%a /dev/sda`?

sharanyad: I think commands like `stat` and `lsblk` take information from `sysfs`, hence do not show the right permissions for the mounted devices. 

I tried the following: 

<pre>
$ ls -l /dev/xvda
brw-rw---- 1 root disk 202, 0 Jul  5 23:37 /dev/xvda
</pre>

<pre>
docker run -it --device /dev/xvda:/dev/dummy ubuntu
# ls -l /dev/dummy
brw-rw---- 1 root disk 202, 0 Oct  6 18:54 /dev/dummy
# stat --format=%a /dev/dummy
660
# fdisk /dev/dummy
Welcome to fdisk (util-linux 2.27.1).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.
Command (m for help): q
</pre>

and 

<pre>
docker run -it --device /dev/xvda:/dev/dummy:r ubuntu
# ls -l /dev/dummy
brw-rw---- 1 root disk 202, 0 Oct  6 18:58 /dev/dummy
# stat --format=%a /dev/dummy
660
# fdisk /dev/dummy
Welcome to fdisk (util-linux 2.27.1).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.
fdisk: cannot open /dev/dummy: Operation not permitted
</pre>

As you mentioned, we can do `ls` to check if the device is mounted and `fdisk` for permissions.

jhaynes:
Can you add entries to the `CHANGELOG.md` for enabling `init` and `devices` and reference this pr as well as #996?

sharanyad:
These changes cannot go in until support for device and init is there in the backend services. Will add changelog and reference https://github.com/aws/amazon-ecs-agent/pull/996.

aaithal:
@sharanyad this is tagged for milestone 1.15.2. Can you please address the comments if this needs to go in the next release? Or, please change the milestone if it doesn't need to go in the next release. Thanks!

sharanyad:
@aaithal The comments have been addressed. The functional tests pass now too. 
Could you please have a look again? Thanks.

PR #1005, opened 2017-10-05
eni attachment state change: ignore expired

aaithal:
### Summary

Ignore expired eni attachment state change events

### Implementation details

This change fixes a bug with the Agent's event submission
logic, where if the ENI attachment is ignored if the
ack time sent by the backend has expired. Not doing this
will result in the Agent forever trying to send the
attachment event to the backend as we can't reliably
distinguish between throttle errors and invalid input
errors from SubmitTaskStateChange response.

### Testing

- [X] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [X] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [X] Functional tests on Linux (`make run-functional-tests`) pass
- [X] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass
- [X] Manual Testing: Verified that Agent stops sending `SubmitTaskStateChange` events on by artificially introducing a delay to simulate timing out expiration time.

None

### Licensing

### DIFF
--- a/agent/api/eniattachment.go
+++ b/agent/api/eniattachment.go
@@ -90,6 +90,15 @@ func (eni *ENIAttachment) StopAckTimer() {
 	eni.ackTimer.Stop()
 }
 
+// HasExpired returns true if the ENI attachment object has exceeded the
+// threshold for notifying the backend of the attachment
+func (eni *ENIAttachment) HasExpired() bool {
+	eni.guard.RLock()
+	defer eni.guard.RUnlock()
+
+	return time.Now().After(eni.ExpiresAt)
+}
+
 // String returns a string representation of the ENI Attachment
 func (eni *ENIAttachment) String() string {
 	eni.guard.RLock()
--- a/agent/eni/watcher/watcher_linux.go
+++ b/agent/eni/watcher/watcher_linux.go
@@ -139,46 +139,46 @@ func (udevWatcher *UdevWatcher) reconcileOnce() error {
 
 	// Add new interfaces next
 	for mac, _ := range currentState {
-		udevWatcher.sendENIStateChange(mac)
+		if err := udevWatcher.sendENIStateChange(mac); err != nil {
+			log.Warnf("Udev watcher reconciliation: unable to send state change: %v", err)
+		}
 	}
 	return nil
 }
 
 // sendENIStateChange handles the eni event from udev or reconcile phase
-func (udevWatcher *UdevWatcher) sendENIStateChange(mac string) {
-	eniAttachment, ok := udevWatcher.shouldSendENIStateChange(mac)
-	if ok {
-		go func(eni *api.ENIAttachment) {
-			eni.Status = api.ENIAttached
-			log.Infof("Emitting ENI change event for: %v", eni)
-			udevWatcher.eniChangeEvent <- api.TaskStateChange{
-				TaskARN:    eni.TaskARN,
-				Attachment: eni,
-			}
-		}(eniAttachment)
-	}
-}
-
-// shouldSendENIStateChange checks whether this eni is managed by ecs
-// and if its status should be sent to backend
-func (udevWatcher *UdevWatcher) shouldSendENIStateChange(macAddress string) (*api.ENIAttachment, bool) {
-	if macAddress == "" {
-		log.Warn("ENI state manager: device with empty mac address")
-		return nil, false
+func (udevWatcher *UdevWatcher) sendENIStateChange(mac string) error {
+	if mac == "" {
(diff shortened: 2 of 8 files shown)


aaithal:
@richardpen That's a good point. It should be doing that now. Can you take another look?

agent/eni/watcher/watcher_linux.go
```diff
@@ -147,16 +147,26 @@ func (udevWatcher *UdevWatcher) reconcileOnce() error {
...
-		go func(eni *api.ENIAttachment) {
-			eni.Status = api.ENIAttached
-			log.Infof("Emitting ENI change event for: %v", eni)
-			udevWatcher.eniChangeEvent <- api.TaskStateChange{
-				TaskARN:    eni.TaskARN,
-				Attachment: eni,
-			}
-		}(eniAttachment)
+	if !ok {
+		if eniAttachment != nil {
```
richardpen: This will also remove the eni attachment that status has been sent, and still managed by agent. We should exclude that whether add a check here or return nil in shouldSendENIStateChange. Also please add this case in your test.

agent/api/eniattachment.go
```diff
@@ -90,6 +90,15 @@ func (eni *ENIAttachment) StopAckTimer() {
 	eni.ackTimer.Stop()
 }
 
+// HasExpired returns true if the ENI attachment object has exceeded the
+// threshold for notifying the backend of the attachment
+func (eni *ENIAttachment) HasExpired() bool {
+	eni.guard.RLock()
```
petderek: When does 'expiresAt' change? Do you need this?

agent/eni/watcher/watcher_linux.go
```diff
@@ -147,16 +147,26 @@ func (udevWatcher *UdevWatcher) reconcileOnce() error {
...
+			// Agent is aware of the ENI, but we decide not to ack it
+			// as it's either already been ack'd or because it's ack
+			// timeout has expired
+			log.Infof("ENI state manager: removing ENI attachment from state, no longer tracking it: %s",
+				eniAttachment.String())
+			udevWatcher.agentState.RemoveENIAttachment(eniAttachment.MACAddress)
```
petderek: why do we remove it if it has already been ack'd?

agent/eni/watcher/watcher_linux.go
aaithal: Isn't that a good thing? I don't think we need to persist the attachment if we've already acknowledged with the backend that Agent has seen the ENI. The only weirdness here is that Attachment data from the state will be cleaned up before the task data, which I still think is not a problem.

richardpen: Remove it from the agent state while it's still managed by ECS sounds not right to me. One rare case I can think of is if customer can know the number of enis managed by ecs by check the eniattachment field in the state file.

aaithal: I'd argue that you should be able to get that `DescribeContainerInstance` API. But, I agree that it may lead to a weird behavior. I'll exclude `IsSent()` from this.

agent/api/eniattachment.go
aaithal: Whenever ACS sends a new message. It's better to be safe than sorry.

agent/eni/watcher/watcher_linux.go
aaithal: We are discussing that here: https://github.com/aws/amazon-ecs-agent/pull/1005/files/a1ff0d886dbf615a68e1c6b29cbb877fd4fb9189#diff-b92a021259580671251a2c3e099e925c

agent/api/eniattachment.go
petderek: That makes sense! I was didn't realize expiresAt was something that would be constantly updated

aaithal: We don't expect it. But, it may change because of bugs in ACS for example.

agent/eni/watcher/watcher_linux.go
```diff
@@ -139,46 +139,47 @@ func (udevWatcher *UdevWatcher) reconcileOnce() error {
...
 	}
-
 	if eni.IsSent() {
-		log.Infof("ENI state manager: eni attach status has already sent: %s", macAddress)
-		return eni, false
+		return errors.Errorf("udev watcher send ENI state change: eni status already sent: %s", eni.String())
+	}
+	if eni.HasExpired() {
+		// Agent is aware of the ENI, but we decide not to ack it
+		// as it's either already been ack'd or because it's ack
```
richardpen: This comment is no longer accurate.

PR #1007, opened 2017-10-05
app: ensure that old tasks are compatible with TaskCPUMemLimits

petderek:
### Summary
This ensures that we dont start old versions of agent with the cgroup hierarchy enabled.

### Implementation
Two main behavior paths:
* If the feature is explicitly enabled by the end user, we will throw an error on startup if old tasks exist in the state file
* If the feature is default enabled, we will degrade gracefully

### Testing

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -193,16 +193,17 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 	}
 
 	// Conditionally create '/ecs' cgroup root
-	// TODO: Ensure that this feature is enabled only when
-	// the cgroup mountpoint is accessible to the agent
 	if agent.cfg.TaskCPUMemLimit.Enabled() {
 		err = agent.resource.Init()
 		// When task CPU and memory limits are enabled, all tasks are placed
 		// under the '/ecs' cgroup root.
-		// TODO: Make this fatal only when explicitly enabled
 		if err != nil {
-			seelog.Criticalf("Unable to setup '/ecs' cgroup: %v", err)
-			return exitcodes.ExitTerminal
+			if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+				seelog.Criticalf("Unable to setup '/ecs' cgroup: %v", err)
+				return exitcodes.ExitTerminal
+			}
+			seelog.Warnf("Disabling TaskCPUMemLimit because agent is unabled to setup '/ecs' cgroup: %v", err)
+			agent.cfg.TaskCPUMemLimit = config.ExplicitlyDisabled
 		}
 	}
 
@@ -280,18 +281,24 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 	previousTaskEngine := engine.NewTaskEngine(agent.cfg, agent.dockerClient,
 		credentialsManager, containerChangeEventStream, imageManager, state)
 
-	// previousState is used to verify that our current runtime configuration is
+	// previousStateManager is used to verify that our current runtime configuration is
 	// compatible with our past configuration as reflected by our state-file
-	previousState, err := agent.newStateManager(previousTaskEngine, &previousCluster,
+	previousStateManager, err := agent.newStateManager(previousTaskEngine, &previousCluster,
 		&previousContainerInstanceArn, &previousEC2InstanceID)
 	if err != nil {
 		seelog.Criticalf("Error creating state manager: %v", err)
 		return nil, "", err
 	}
 
-	err = previousState.Load()
+	err = previousStateManager.Load()
 	if err != nil {
 		seelog.Criticalf("Error loading previously saved state: %v", err)
+		return nil, "", err
+	}
+
+	err = agent.checkCompatibility(previousTaskEngine)
+	if err != nil {
+		seelog.Criticalf("Error checking compatibility with previously saved state: %v", err)
 		return nil, "", err
 	}
 
--- a/agent/app/agent_compatibility_linux.go
+++ b/agent/app/agent_compatibility_linux.go
@@ -1,0 +1,63 @@
+// +build linux
+
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package app
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/engine"
+
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+// checkCompatibility (for linux) ensures that the running task engine is capable for running with the TaskCPUMemLimit
+// flag enabled. It will disable the feature if it determines that it is incompatible and the feature isn't explicitly
+// enabled.
+func (agent *ecsAgent) checkCompatibility(engine engine.TaskEngine) error {
+
+	// We don't need to do these checks if CPUMemLimit is disabled
+	if !agent.cfg.TaskCPUMemLimit.Enabled() {
+		return nil
+	}
+
+	tasks, err := engine.ListTasks()
+	if err != nil {
+		return err
+	}
+
+	// Loop over each task to determine if the state is compatible with the running version of agent and its options
+	compatible := true
+	for _, task := range tasks {
+		if !task.MemoryCPULimitsEnabled {
+			seelog.Warnf("App: task from state file is not compatible with TaskCPUMemLimit setting, because it is not using ecs' cgroup hierarchy: %s", task.Arn)
+			compatible = false
+			break
+		}
+	}
+
+	if compatible {
+		return nil
+	}
+
+	if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+		return errors.New("App: unable to load old tasks because TaskCPUMemLimits setting is incompatible with old state.")
+	}
+
+	seelog.Warn("App: disabling TaskCPUMemLimit.")
+	agent.cfg.TaskCPUMemLimit = config.ExplicitlyDisabled
+
+	return nil
+}
--- a/agent/app/agent_compatibility_linux_test.go
+++ b/agent/app/agent_compatibility_linux_test.go
@@ -1,0 +1,132 @@
+// +build linux
+
+// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package app
+
+import (
+	"context"
+	"testing"
+
+	"github.com/aws/amazon-ecs-agent/agent/api"
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/ec2"
+	"github.com/aws/amazon-ecs-agent/agent/eventstream"
+	"github.com/aws/amazon-ecs-agent/agent/statemanager/mocks"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCompatibilityEnabledSuccess(t *testing.T) {
+	ctrl, creds, state, images, _, _, stateManagerFactory, saveableOptionFactory := setup(t)
+	defer ctrl.Finish()
+	stateManager := mock_statemanager.NewMockStateManager(ctrl)
+
+	cfg := getTestConfig()
+	cfg.Checkpoint = true
+	cfg.TaskCPUMemLimit = config.DefaultEnabled
+
+	agent := &ecsAgent{
+		cfg:                   &cfg,
+		stateManagerFactory:   stateManagerFactory,
+		saveableOptionFactory: saveableOptionFactory,
+		ec2MetadataClient:     ec2.NewBlackholeEC2MetadataClient(),
+	}
+
+	gomock.InOrder(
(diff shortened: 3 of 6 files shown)


agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,48 @@
...
+			seelog.Infof("Task from state file is not compatible with TaskCPUMemLimit setting, %s", task)
+			incompatible = true
+			break
+		}
+	}
+
+	if incompatible && agent.cfg.TaskCPUMemLimit.Enabled() {
+		if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+			return errors.New("app: unable to load old tasks because TaskCPUMemLimits setting is incompatible with old state.")
+		} else {
```
aaithal: nit: you don't need the `else` here.

agent/app/agent.go
```diff
@@ -282,19 +282,25 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 
 	// previousState is used to verify that our current runtime configuration is
 	// compatible with our past configuration as reflected by our state-file
-	previousState, err := agent.newStateManager(previousTaskEngine, &previousCluster,
+	previousStateManager, err := agent.newStateManager(previousTaskEngine, &previousCluster,
```
vsiddharth: * Comment needs to be updated alongside
* `previousStateManager` is confusing, a better name is preferable

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,48 @@
...
+	var incompatible bool
+	for _, task := range tasks {
+		if utils.ZeroOrNil(task.MemoryCPULimitsEnabled) {
+			seelog.Infof("Task from state file is not compatible with TaskCPUMemLimit setting, %s", task)
+			incompatible = true
+			break
+		}
+	}
+
+	if incompatible && agent.cfg.TaskCPUMemLimit.Enabled() {
```
aaithal: I think there's an efficiency gain to be had here if you refactor this a bit. Checking if `agent.cfg.TaskCPUMemLimit.Enabled()` is `false` at the very beginning of the method will result in avoiding all of the `ListTasks` and checking each task's  `MemoryCPULimitsEnabled` field.

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (agent *ecsAgent) checkCompatibility(engine engine.TaskEngine) error {
+	tasks, err := engine.ListTasks()
+	if err != nil {
+		return err
+	}
+
+	var incompatible bool
+	for _, task := range tasks {
+		if utils.ZeroOrNil(task.MemoryCPULimitsEnabled) {
```
aaithal: Why are we checking for `task.MemoryCPULimitsEnabled` here? Shouldn't the mere fact that there tasks that are already running on the Agent prevent us from enabling this feature? As otherwise, there'd be a skewed `cgroup` hierarchy?

vsiddharth: Can you avoid using `utils.ZeroOrNil`

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,48 @@
...
+			seelog.Infof("Task from state file is not compatible with TaskCPUMemLimit setting, %s", task)
```
vsiddharth: The task ARN should suffice here

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,48 @@
...
+			incompatible = true
+			break
+		}
+	}
+
+	if incompatible && agent.cfg.TaskCPUMemLimit.Enabled() {
+		if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+			return errors.New("app: unable to load old tasks because TaskCPUMemLimits setting is incompatible with old state.")
+		} else {
+			agent.cfg.TaskCPUMemLimit = config.ExplicitlyDisabled
```
vsiddharth: Please add a log before explicitly turning of the setting

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,48 @@
+// +build linux
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
vsiddharth: nit: empty line after the build tags

agent/app/agent_compatibility_linux_test.go
```diff
@@ -0,0 +1,129 @@
+// +build linux
+// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
vsiddharth: nit: empty line after build tags

agent/app/agent_compatibility_linux_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/api"
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/ec2"
+	"github.com/aws/amazon-ecs-agent/agent/eventstream"
+	"github.com/aws/amazon-ecs-agent/agent/statemanager/mocks"
+
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+)
+
```
vsiddharth: Can the mocks be ordered in these tests

agent/app/agent_compatibility_unpecified.go
```diff
@@ -0,0 +1,23 @@
+// +build !linux
+// Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
vsiddharth: nit: empty line after build tag

agent/app/agent_compatibility_unpecified.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package app
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/engine"
```
vsiddharth: braces not required here

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/engine"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+func (agent *ecsAgent) checkCompatibility(engine engine.TaskEngine) error {
```
vsiddharth: Can this method also verify that the agent has access to the cgroup mounts?

agent/app/agent_compatibility_linux.go
petderek: If we are restoring from a snapshot of the current or later agent version, with limits enabled, its completely reasonable to proceed.

agent/app/agent_compatibility_linux.go
petderek: Good idea!

agent/app/agent.go
petderek: Good callout.

I'm not sure what is confusing about it; what would you recommend? I added 'manager' because its a manager object and not the underlying state.

agent/app/agent_compatibility_linux.go
petderek: ahhh i forgot about that. will fix

agent/app/agent_compatibility_linux.go
petderek: I'm not sure that fits within the scope of responsibility for this function. Here we are checking the state file for incompatibilities. 

Don't we handle that error in the agent startup process (with cgroups enabled)?

agent/app/agent_compatibility_linux_test.go
petderek: I can, but I don't think it adds any value for these tests in particular.

Is it just to be consistent with other tests in the package? I can agree with that.

agent/app/agent_compatibility_unpecified.go
petderek: Not sure what the standard is/was -- I can remove them.

agent/app/agent_compatibility_unpecified.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func (agent *ecsAgent) checkCompatibility(engine engine.TaskEngine) error {
```
sharanyad: When exactly will this be called? 
also nit - I guess the file name is `agent_compatibility_unspecified.go`?

petderek: This still gets called when we reload the state file in the app startup. On Windows / OS X, it won't do anything.

Also thanks! Spelling is hard.

sharanyad: aah ok, used to seeing `_windows` everywhere, `unspecified` was new. So just confirmed.

agent/app/agent.go
vsiddharth: I was slightly confused; I'm no longer unhappy with the name
The `previousStateManager` uses the `previousTaskEngine` to become the new `taskEngine`

agent/app/agent_compatibility_linux.go
vsiddharth: The agent does not do this yet. We only have a `TODO`. Have a look at `L#196`

agent/app/agent_compatibility_linux_test.go
vsiddharth: Its just for consistency within this package

agent/app/agent_compatibility_unpecified.go
vsiddharth: The task cgroup checks need happen only on supported platforms`(linux)` but we still need to maintain compatibility on `windows`

agent/app/agent.go
petderek: Yeah. Its kind of weird but the whole function has this notion of 'previous' where everything labeled 'previous' is inferred from the state file that has been loaded

agent/app/agent_compatibility_linux.go
petderek: aha! I'll need to think about that.

We want that to happen in more cases than only when this function gets invoked.

vsiddharth: I think a better place for the `mountpoint` check would be the `config` package as we can stick it in the `overrides`

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,55 @@
...
+	compatible := true
+	for _, task := range tasks {
+		if !task.MemoryCPULimitsEnabled {
+			seelog.Warnf("app: task from state file is not compatible with TaskCPUMemLimit setting, %s", task.Arn)
+			compatible = false
+			break
+		}
+	}
+
+	if !compatible {
```
aaithal: nit: you can reduce the indentation characters:

```
if compatible {
 return nil
}

things and stuff ...
```

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,55 @@
...
+		if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+			return errors.New("app: unable to load old tasks because TaskCPUMemLimits setting is incompatible with old state.")
+		}
+		seelog.Warnf("app: disabling TaskCPUMemLimit.")
```
aaithal: nit: `"App ..`, also it would be nice to log why this is being disabled as well.

agent/app/agent_compatibility_linux.go
aaithal: I agree that we should do that check outside of this function, possibly somewhere in this block: https://github.com/aws/amazon-ecs-agent/pull/1007/files#diff-2149975f54c050dd31ac9a37f1ef63bcR198

agent/app/agent_compatibility_linux.go
vsiddharth: nit: `Warnf -> Warn`

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,55 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package app
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/engine"
+	"github.com/cihub/seelog"
```
vsiddharth: nit: separate 3p packages

agent/app/agent_compatibility_linux.go
petderek: The reason will be included with the previous Warn.

agent/app/agent_compatibility_linux.go
petderek: Agreed. I'll address that separately.

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,60 @@
...
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/engine"
+
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+func (agent *ecsAgent) checkCompatibility(engine engine.TaskEngine) error {
```
vsiddharth: might we worth adding a note that this method can update the config too

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,60 @@
...
+
+	if compatible {
+		return nil
+	}
+
+	if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+		return errors.New("app: unable to load old tasks because TaskCPUMemLimits setting is incompatible with old state.")
+	}
+
+	seelog.Warn("app: disabling TaskCPUMemLimit.")
```
aaithal: nit: `"App"`. Please ensure that log messages being with upper-case letters.

agent/app/agent_compatibility_linux.go
```diff
@@ -0,0 +1,60 @@
...
+	tasks, err := engine.ListTasks()
+	if err != nil {
+		return err
+	}
+
+	// Loop over each task to determine if the state is compatible with the running version of agent and its options
+	compatible := true
+	for _, task := range tasks {
+		if !task.MemoryCPULimitsEnabled {
+			seelog.Warnf("app: task from state file is not compatible with TaskCPUMemLimit setting, %s", task.Arn)
```
aaithal: Please expand this to log why it's incompatible as well.

agent/app/agent_compatibility_linux.go
aaithal: This is still not addressed.

petderek: whoops. I’ve been doing the opposite. Will fix!

PR #1009, opened 2017-10-07
move to dep

adnxn:
### Summary
Use [dep](https://github.com/golang/dep) instead of [godep](https://github.com/tools/godep).

$ dep version
dep:
 version     : v0.3.2
 build date  : 2017-10-19
 git hash    : 8ddfc8a
 go version  : go1.9
 go compiler : gc
 platform    : linux/amd64

### Implementation details

See `agent/Gopkg.toml` and `agent/Gopkg.lock`. 
### Testing

New tests cover the changes: 
no
### Licensing

yes


agent/Gopkg.toml
```diff
@@ -0,0 +1,50 @@
...
+#   source = "github.com/myfork/project2"
+#
+# [[override]]
+#  name = "github.com/x/y"
+#  version = "2.4.0"
+
+
+[[constraint]]
+  name = "github.com/aws/aws-sdk-go"
+  version = "1.8.19"
```
petderek: What do you think about locking everything to the commit that is referenced in the old Godeps.json file? It would potentially reduce the blast radius.

adnxn: what about locking dependencies to tags and upgrade to nearest tagged version where it makes sense. ill add a separate commit for each dependency i lock to a new version. cause as we were saying earlier, tagging to revisions doesn't really provide guarantees. we can definitely do some things to reduce blast radius though.

vsiddharth:
Quoting the original document from `dep`
```
dep is a prototype dependency management tool for Go. It requires Go 1.8 or newer to compile.
dep is the official experiment, but not yet the official tool. 
```

Even though `dep` is safe for production, I'm not totally sure of switching to it yet!

adnxn:
> It requires Go 1.8 or newer to compile.

> Even though dep is safe for production, I'm not totally sure of switching to it yet!

hrm that's true, we should consider this.

/cc @aaithal, @jahkeup 

what do you think?

agent/Gopkg.toml
jahkeup: Assuming that the Gopkg.lock contains the same commits as the previous Gopkg.json file had for our deps I think it's okay to migrate this way.

jahkeup:
The Makefile will also need to be updated to account for this change, which will require that we also bump the version of go used for the travis build. Though I'm happy to start using dep, we need to verify we are ready to update go. This looks great otherwise :+1:

samuelkarp:
I don't think we need to bump Go.  I think the dep tool needs Go 1.8 to be compiled itself, but our code should only rely on Go >= 1.5 (with GO15VENDOREXPERIMENT).

samuelkarp:
This now needs a rebase, since `Godeps.json` has been modified.

agent/prune.out
```diff
@@ -0,0 +1,6094 @@
+On branch move-to-dep
```
samuelkarp: Can you remove this file from this intermediate commit?

PR #1012, opened 2017-10-11
app, ecs_client: Terminal error handling for registerContainerInstance api

sharanyad:
### Summary
All the Server and Client exception errors from Register Container Instance call are transient errors.
Making InvalidParameterException error to be non-transient i.e, Agent will not restart on InvalidParameter error.

### Implementation details
Check if the AWS error returned from register container instance API is of the type InvalidParameterException and return it as non-transient error. 

### Testing

Making invalid parameter exception error from register container instance API call non-transient

### Licensing


agent/app/agent.go
```diff
@@ -402,6 +404,10 @@ func (agent *ecsAgent) registerContainerInstance(
 			seelog.Critical("Instance registration attempt with an invalid attribute")
 			return err
 		}
+		awsErr, ok := err.(awserr.Error)
+		if ok && awsErr.Code() == ecs_service.ErrCodeInvalidParameterException {
+			return err
```
aaithal: Can you also add a log line here? Similar to the `"invalid attribute"` one above?

agent/app/agent.go
```diff
@@ -402,6 +404,10 @@ func (agent *ecsAgent) registerContainerInstance(
 			seelog.Critical("Instance registration attempt with an invalid attribute")
 			return err
 		}
+		awsErr, ok := err.(awserr.Error)
```
aaithal: This should be before `utils.AttributeError` comparison. Also, a minor nit here: `if awsErr, ok := ...` pattern leads to lesser locs.

agent/app/agent.go
```diff
@@ -47,6 +47,8 @@ import (
 	aws_credentials "github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/cihub/seelog"
+	"github.com/aws/aws-sdk-go/aws/awserr"
```
vsiddharth: nit: move these above the 3p packages

agent/app/agent.go
sharanyad: done.

agent/app/agent.go
sharanyad: done

agent/app/agent.go
sharanyad: done

agent/app/agent.go
```diff
@@ -46,6 +46,8 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	aws_credentials "github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	ecs_service "github.com/aws/aws-sdk-go/service/ecs"
```
samuelkarp: Please use the private generated client that we have at `github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs` instead.  I'd rather not vendor two copies of the client.  We *may* need to update the [generator](https://github.com/aws/amazon-ecs-agent/blob/v1.14.5/agent/gogenerate/awssdk.go) to generate these constants, but I would prefer that to vendoring an additional copy.

sharanyad:
Generated error types in `ecs_client`, instead of using aws-sdk-go client vendor package. Please have a look again. Thanks!

Issue #1010, opened 2017-10-09
Docker capability testing is broken with Docker >= 1.13.0


samuelkarp:
### Summary
Capability detection is broken because we're not able to detect supported Remote API versions properly since Docker 1.13.0.

### Description
See https://github.com/moby/moby/issues/35145

Docker stopped rejecting API requests with version numbers greater than the maximum version supported by the daemon. This happened in https://github.com/moby/moby/commit/e98e4a71110fd33852bb755a9b8b4ebc9df904db#diff-14051df1b5de1608aaba3a983f2a87e3 as part of https://github.com/moby/moby/pull/27745 and has affected every version of Docker since 1.13.0.

This change breaks the capability detection logic used by the Amazon ECS agent. The Amazon ECS agent is designed to work with all Docker versions >= 1.5.0, and we enable or disable features depending on the capabilities of the Docker version used. We use Remote API version as we expected this to be more stable than the daemon version; the daemon version has gone through a few changes since we started (roughly semver-like to year-month plus edition). We currently call the `/_ping` API with a range of different versions to discover which API versions are supported by the Docker daemon, but this change makes it so that every request with an API version >= 1.12 will result in success. This is causing incorrect behavior where features are being enabled with incompatible Docker versions; see https://github.com/aws/amazon-ecs-agent/issues/1008 .

In https://docs.docker.com/engine/api/version-history/#v125-api-changes , I see a note about a new header returned that specifies the maximum API version of the daemon. However, I do not see a note about the change in behavior to start accepting requests that have a specified version greater than the maximum.

**Steps to reproduce the issue:**
Run `curl --verbose --unix-socket /var/run/docker.sock http://localhost/v9999.9999/_ping`

On Docker versions prior to 1.13.0, you'll see an HTTP 400 with an error message "client is newer than server"
On Docker versions 1.13.0 or greater, you'll see an HTTP 200 OK.

aaithal:
To make the version detection logic in the Agent more generic, let's do this:

1. create an unversioned docker client
2. call `client.Version()`
3. if response contains `MinAPIVersion`, register the versions we care about between [`MinAPIVersion`,`ApiVersion`]
4. else use the versioned client + `client.Ping()` logic we have today

samuelkarp:
LGTM, except I'd just use a versioned client at the default (1.17 for Linux and 1.24 for Windows) in order to make the initial `/version` call.

sharanyad:
[...]

PR #1014, opened 2017-10-12
dockerclient: Agent supported Docker versions from client min,API Versions

sharanyad:
### Summary
Finds Agent supported Docker version clients from client version's min, API Versions returned

### Implementation details
dockerclient: Finding agent supported Docker versions by getting the default client version and if the min API version is present, add all the clients between min API version and API version supported, else follow current logic
utils: reuse the semantic version comparator to compare Docker API versions with only major and minor versions too

### Testing

New tests cover the changes: yes|no

Supported Docker version clients from default client's min supported API and API version returned

### Licensing

### DIFF
--- a/agent/engine/dockerclient/dockerclientfactory.go
+++ b/agent/engine/dockerclient/dockerclientfactory.go
@@ -14,13 +14,23 @@
 package dockerclient
 
 import (
-	"errors"
-
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
 	log "github.com/cihub/seelog"
 	docker "github.com/fsouza/go-dockerclient"
+	"github.com/pkg/errors"
 )
 
+const (
+	// minAPIVersionKey is the docker.Env key for min API version
+	// This is supported in Docker API versions >=1.25
+	// https://docs.docker.com/engine/api/version-history/#v125-api-changes
+	minAPIVersionKey = "MinAPIVersion"
+	// apiVersionKey is the docker.Env key for API version
+	apiVersionKey    = "ApiVersion"
+	// zeroPatch is a string to append patch number zero if the major minor version lacks it
+	zeroPatch = ".0"
+)
 // Factory provides a collection of docker remote clients that include a
 // recommended client version as well as a set of alternative supported
 // docker clients.
@@ -110,19 +120,71 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
 // findDockerVersions loops over all known API versions and finds which ones
 // are supported by the docker daemon on the host
 func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
+	// if the client version returns a MinAPIVersion and APIVersion, then use it to return
+	// all the Docker clients between MinAPIVersion and APIVersion, else try pinging
+	// the clients in getKnownAPIVersions
+	var minAPIVersion, apiVersion string
+	// get a Docker client with the default supported version
+	client, err := newVersionedClient(endpoint, string(minDockerAPIVersion))
+	if err == nil {
+		clientVersion, err := client.Version()
+		if err == nil {
+			// check if the docker.Env obj has MinAPIVersion key
+			if clientVersion.Exists(minAPIVersionKey) {
+				minAPIVersion = clientVersion.Get(minAPIVersionKey)
+			}
+			// check if the docker.Env obj has APIVersion key
+			if clientVersion.Exists(apiVersionKey) {
+				apiVersion = clientVersion.Get(apiVersionKey)
+			}
+		}
+	}
+
 	clients := make(map[DockerVersion]dockeriface.Client)
 	for _, version := range getKnownAPIVersions() {
-		client, err := newVersionedClient(endpoint, string(version))
+		dockerClient, err := getDockerClientForVersion(endpoint, string(version), minAPIVersion, apiVersion)
 		if err != nil {
-			log.Infof("Error while creating client: %v", err)
+			log.Infof("Unable to get Docker client for version %s: %v", version, err)
 			continue
 		}
-
-		err = client.Ping()
-		if err != nil {
-			log.Infof("Failed to ping with Docker version %s: %v", version, err)
-		}
-		clients[version] = client
+		clients[version] = dockerClient
 	}
 	return clients
+}
+
+func getDockerClientForVersion(
+	endpoint string,
+	version string,
+	minAPIVersion string,
+	apiVersion string) (dockeriface.Client, error) {
+	if (minAPIVersion != "" && apiVersion != "") {
+		// Adding patch number zero to Docker versions to reuse the existing semver
+		// comparator
+		// TODO: remove this logic later when non-semver comparator is implemented
+		versionWithPatch := version + zeroPatch
+		lessThanMinCheck := "<" + minAPIVersion + zeroPatch
+		moreThanMaxCheck := ">" + apiVersion + zeroPatch
+		minVersionCheck, err := utils.Version(versionWithPatch).Matches(lessThanMinCheck)
+		if err != nil {
+			return nil, errors.Wrapf(err, "version detection using MinAPIVersion: unable to get min version: %s", minAPIVersion)
+		}
+		maxVersionCheck, err := utils.Version(versionWithPatch).Matches(moreThanMaxCheck)
+		if err != nil {
+			return nil, errors.Wrapf(err, "version detection using MinAPIVersion: unable to get max version: %s", apiVersion)
+		}
+		// do not add the version when it is less than min api version or greater
+		// than api version
+		if minVersionCheck || maxVersionCheck {
+			return nil, errors.Errorf("version detection using MinAPIVersion: unsupported version: %s", version)
+		}
+	}
+	client, err := newVersionedClient(endpoint, string(version))
+	if err != nil {
+		return nil, errors.Wrapf(err, "version detection check: unable to create Docker client for version: %s", version)
+	}
+	err = client.Ping()
+	if err != nil {
+		return nil, errors.Wrapf(err, "version detection check: failed to ping with Docker version: %s", string(version))
+	}
+	return client, nil
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,14 @@
 
 ## UNRELEASED
 * Feature - Support for provisioning Tasks with ENIs
+* Bug - Fixed a bug where unsupported Docker API client versions could be registered
+[#1014](https://github.com/aws/amazon-ecs-agent/pull/1014)
 
 ## 1.14.5
 * Enhancement - Retry failed container image pull operations [#975](https://github.com/aws/amazon-ecs-agent/pull/975)
 * Enhancement - Set read and write timeouts for websocket connectons [#993](https://github.com/aws/amazon-ecs-agent/pull/993)
 * Enhancement - Add support for the SumoLogic Docker log driver plugin
-  [#992](https://github.com/aws/amazon-ecs-agent/pull/992) 
+  [#992](https://github.com/aws/amazon-ecs-agent/pull/992)
 * Bug - Fixed a memory leak issue when submitting the task state change [#967](https://github.com/aws/amazon-ecs-agent/pull/967)
 * Bug - Fixed a race condition where a container can be created twice when agent restarts. [#939](https://github.com/aws/amazon-ecs-agent/pull/939)
 * Bug - Fixed an issue where `microsoft/windowsservercore:latest` was not
--- a/agent/engine/dockerclient/dockerclientfactory_test.go
+++ b/agent/engine/dockerclient/dockerclientfactory_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface/mocks"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	docker "github.com/fsouza/go-dockerclient"
 )
 
 const expectedEndpoint = "expectedEndpoint"
@@ -35,7 +36,8 @@ func TestGetDefaultClientSuccess(t *testing.T) {
 		if version == string(getDefaultVersion()) {
 			mockClient = expectedClient
 		}
-		mockClient.EXPECT().Ping()
+		mockClient.EXPECT().Version().Return(&docker.Env{}, nil).AnyTimes()
+		mockClient.EXPECT().Ping().AnyTimes()
 
 		return mockClient, nil
 	}
@@ -59,7 +61,8 @@ func TestFindSupportedAPIVersions(t *testing.T) {
 	// Ensure that agent pings all known versions of Docker API
 	for i := 0; i < len(allVersions); i++ {
 		mockClients[string(allVersions[i])] = mock_dockeriface.NewMockClient(ctrl)
-		mockClients[string(allVersions[i])].EXPECT().Ping()
+		mockClients[string(allVersions[i])].EXPECT().Version().Return(&docker.Env{}, nil).AnyTimes()
+		mockClients[string(allVersions[i])].EXPECT().Ping().AnyTimes()
 	}
 
 	// Define the function for the mock client
@@ -91,4 +94,59 @@ func TestVerifyAgentVersions(t *testing.T) {
 	for _, agentVersion := range getAgentVersions() {
 		assert.True(t, isKnown(agentVersion))
 	}
+}
+
+func TestFindSupportedAPIVersionsFromMinAPIVersions(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	agentVersions := getAgentVersions()
+	allVersions := getKnownAPIVersions()
+
+	// Set up the mocks and expectations
+	mockClients := make(map[string]*mock_dockeriface.MockClient)
+
+	// Ensure that agent pings all known versions of Docker API
+	for i := 0; i < len(allVersions); i++ {
+		mockClients[string(allVersions[i])] = mock_dockeriface.NewMockClient(ctrl)
+		mockClients[string(allVersions[i])].EXPECT().Version().Return(&docker.Env{"MinAPIVersion=1.12","ApiVersion=1.27"}, nil).AnyTimes()
+		mockClients[string(allVersions[i])].EXPECT().Ping().AnyTimes()
+	}
+
+	// Define the function for the mock client
+	// For simplicity, we will pretend all versions of docker are available
+	newVersionedClient = func(endpoint, version string) (dockeriface.Client, error) {
+		return mockClients[version], nil
+	}
+
+	factory := NewFactory(expectedEndpoint)
+	actualVersions := factory.FindSupportedAPIVersions()
+
+	assert.Equal(t, len(agentVersions), len(actualVersions))
+	for i := 0; i < len(actualVersions); i++ {
+		assert.Equal(t, agentVersions[i], actualVersions[i])
+	}
+}
+
+func TestCompareDockerVersionsWithMinAPIVersion(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	minAPIVersion := "1.12"
+	apiVersion := "1.27"
+	versions := []string{"1.11", "1.29"}
+	rightVersion := "1.25"
+
+	for _, version := range versions {
+		_, err := getDockerClientForVersion("endpoint", version, minAPIVersion, apiVersion)
+		assert.EqualError(t, err, "version detection using MinAPIVersion: unsupported version: " + version)
+	}
+
+	mockClients := make(map[string]*mock_dockeriface.MockClient)
+	newVersionedClient = func(endpoint, version string) (dockeriface.Client, error) {
+		mockClients[version] = mock_dockeriface.NewMockClient(ctrl)
+		mockClients[version].EXPECT().Ping()
+		return mockClients[version], nil
+	}
+	client, _ := getDockerClientForVersion("endpoint", rightVersion, minAPIVersion, apiVersion)
+	assert.Equal(t, mockClients[rightVersion], client)
 }
--- a/agent/engine/dockerclient/dockerclientfactory_unix_test.go
+++ b/agent/engine/dockerclient/dockerclientfactory_unix_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface/mocks"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	docker "github.com/fsouza/go-dockerclient"
 )
 
 func TestGetClientCached(t *testing.T) {
@@ -29,7 +30,8 @@ func TestGetClientCached(t *testing.T) {
 
 	newVersionedClient = func(endpoint, version string) (dockeriface.Client, error) {
 		mockClient := mock_dockeriface.NewMockClient(ctrl)
-		mockClient.EXPECT().Ping()
+		mockClient.EXPECT().Version().Return(&docker.Env{}, nil).AnyTimes()
+		mockClient.EXPECT().Ping().AnyTimes()
 		return mockClient, nil
 	}
 
--- a/agent/engine/dockerclient/dockerclientfactory_windows_test.go
+++ b/agent/engine/dockerclient/dockerclientfactory_windows_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface/mocks"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	docker "github.com/fsouza/go-dockerclient"
 )
 
 func TestGetClientMinimumVersion(t *testing.T) {
@@ -31,10 +32,11 @@ func TestGetClientMinimumVersion(t *testing.T) {
 
 	newVersionedClient = func(endpoint, version string) (dockeriface.Client, error) {
 		mockClient := mock_dockeriface.NewMockClient(ctrl)
-		if version == string(MinDockerAPIWindows) {
+		if version == string(minDockerAPIVersion) {
 			mockClient = expectedClient
 		}
-		mockClient.EXPECT().Ping()
+		mockClient.EXPECT().Version().Return(&docker.Env{}, nil).AnyTimes()
+		mockClient.EXPECT().Ping().AnyTimes()
 		return mockClient, nil
 	}
 
--- a/agent/engine/dockerclient/versionsupport_unix.go
+++ b/agent/engine/dockerclient/versionsupport_unix.go
@@ -18,6 +18,10 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface"
 )
 
+const (
+	// minDockerAPIVersion is the min Docker API version supported by agent
+	minDockerAPIVersion = Version_1_17
+)
 // GetClient on linux will simply return the cached client from the map
 func (f *factory) GetClient(version DockerVersion) (dockeriface.Client, error) {
 	return f.getClient(version)
--- a/agent/engine/dockerclient/versionsupport_windows.go
+++ b/agent/engine/dockerclient/versionsupport_windows.go
@@ -18,14 +18,14 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface"
 )
 
-const MinDockerAPIWindows = Version_1_24
+const minDockerAPIVersion = Version_1_24
 
 // GetClient will replace some versions of Docker on Windows. We need this because
 // agent assumes that it can always call older versions of the docker API.
 func (f *factory) GetClient(version DockerVersion) (dockeriface.Client, error) {
 	for _, v := range getWindowsReplaceableVersions() {
 		if v == version {
-			version = MinDockerAPIWindows
+			version = minDockerAPIVersion
 			break
 		}
 	}
@@ -48,10 +48,10 @@ func getWindowsReplaceableVersions() []DockerVersion {
 
 // getAgentVersions for Windows should return all of the replaceable versions plus additional versions
 func getAgentVersions() []DockerVersion {
-	return append(getWindowsReplaceableVersions(), MinDockerAPIWindows)
+	return append(getWindowsReplaceableVersions(), minDockerAPIVersion)
 }
 
 // getDefaultVersion returns agent's default version of the Docker API
 func getDefaultVersion() DockerVersion {
-	return MinDockerAPIWindows
+	return minDockerAPIVersion
 }


sharanyad:
This addresses https://github.com/aws/amazon-ecs-agent/issues/1010 required for https://github.com/aws/amazon-ecs-agent/pull/996

aaithal:
@sharanyad can you please fix the unit test failure?

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +135,54 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
 		if err != nil {
 			log.Infof("Failed to ping with Docker version %s: %v", version, err)
+			continue
 		}
 		clients[version] = client
 	}
 	return clients
 }
+
+func findDockerVersionsfromMinMaxVersions(endpoint string) (map[DockerVersion]dockeriface.Client, bool) {
```
aaithal: You're using the second parameter type (`bool`) to indicate the success/failure of this method, which means that returning an `error` instead is the better way to do that.

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +135,54 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	clients := make(map[DockerVersion]dockeriface.Client)
```
vsiddharth: This line can be moved down below

agent/utils/compare_versions.go
```diff
@@ -93,6 +96,9 @@ func (lhs Version) Matches(selector string) (bool, error) {
 
 func parseSemver(version string) (semver, error) {
 	var result semver
+	if ok := checkForNoPatch(version); ok {
```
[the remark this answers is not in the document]
sharanyad: done

agent/engine/dockerclient/dockerclientfactory.go
sharanyad: done

agent/engine/dockerclient/dockerclientfactory.go
sharanyad: I am not sure if it can be called a failure if min api version key is not found, nevertheless changed the return type to `error`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +135,54 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	if err != nil {
+		log.Infof("Error while getting client version: %v", err)
+		return nil, false
+	}
+
+	// check if the docker.Env obj has MinAPIVersion key
+	if clientVersion.Exists(minAPIVersionKey) {
+		minAPIVersion := clientVersion.Get(minAPIVersionKey)
+		apiVersion := clientVersion.Get(apiVersionKey)
+		for _, version := range getKnownAPIVersions() {
```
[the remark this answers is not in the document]
sharanyad: Sure.

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -14,13 +14,21 @@
...
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
 	log "github.com/cihub/seelog"
 	docker "github.com/fsouza/go-dockerclient"
+	"github.com/pkg/errors"
 )
 
+const (
+	// minDockerAPILinux is the min Docker API version supported by agent
+	minDockerAPILinux = Version_1_17
```
samuelkarp: You need to specify this differently for Linux and Windows.  Windows won't respond at this API version.  You can change the constants in `versionsupport_windows.go` and `versionsupport_unix.go` to have the same name and then just use those.

agent/utils/compare_versions.go
```diff
@@ -20,6 +20,9 @@ import (
 	"strings"
 )
 
+// zeroPatch is a string to append patch number zero if the major minor version lacks it
+const zeroPatch = ".0"
```
samuelkarp: Instead of doing this, I think I'd actually prefer implementing a different non-semver version comparator.  API versions from Docker are not semver and making them look like semver can just be a source of confusion.

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +137,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+
+	clients := make(map[DockerVersion]dockeriface.Client)
+	// check if the docker.Env obj has MinAPIVersion key
+	if clientVersion.Exists(minAPIVersionKey) {
+		minAPIVersion := clientVersion.Get(minAPIVersionKey)
+		apiVersion := clientVersion.Get(apiVersionKey)
+		for _, version := range getKnownAPIVersions() {
+			dockerClient, err := getDockerClientForVersion(endpoint, string(version), minAPIVersion, apiVersion)
+			if err != nil {
+				log.Infof("Error getting client version: %v", err)
```
samuelkarp: Is INFO the right level for this?  I think you actually want DEBUG and probably don't call it an error (since it is expected that some won't match).

agent/utils/compare_versions.go
sharanyad: I just wanted to reuse the existing semver comparator due to time constraints for this fix. I agree this is not the right way, but I would prefer to add it to my list, come back and implement a non-semver comparator once this goes out.

samuelkarp: That's fair.  Want to add a TODO comment and possibly open an issue to track?

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +137,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
 		clients[version] = client
 	}
 	return clients
 }
+
+func findDockerVersionsfromMinMaxVersions(endpoint string) (map[DockerVersion]dockeriface.Client, error) {
+	// get a Docker client with the default supported version
+	client, err := newVersionedClient(endpoint, string(minDockerAPILinux))
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while creating client")
```
vsiddharth: Please add more context to the error and not use upper case characters in the message. 
The code base has plenty of examples using `errors.Wrap`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +137,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	}
+
+	clientVersion, err := client.Version()
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while getting client version")
```
vsiddharth: same as above

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +137,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	endpoint string,
+	version string,
+	minAPIVersion string,
+	apiVersion string) (dockeriface.Client, error) {
+	lessThanMinCheck := "<"+minAPIVersion
+	moreThanMaxCheck := ">"+apiVersion
+	minVersionCheck, minErr := utils.Version(version).Matches(lessThanMinCheck)
+	maxVersionCheck, maxErr := utils.Version(version).Matches(moreThanMaxCheck)
+	if minErr != nil {
+		return nil, errors.Wrapf(minErr, "Error while comparing version %s with minAPIVersion %s", version, minAPIVersion)
```
vsiddharth: same as above

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +137,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	}
+	if maxErr != nil {
+		return nil, errors.Wrapf(minErr, "Error while comparing version %s with apiVersion %s", version, apiVersion)
```
vsiddharth: same as above

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +137,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	}
+	// do not add the version when it is less than min api version or greater
+	// than api version
+	if minVersionCheck || maxVersionCheck {
+		return nil, errors.Errorf("Min and API versions comparison check failed for version: %s", version)
```
vsiddharth: same as above

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +137,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	}
+	client, err := newVersionedClient(endpoint, string(version))
+	if err != nil {
+		return nil, errors.Wrap(err, "Error while creating client")
```
vsiddharth: same as above

agent/utils/compare_versions.go
sharanyad: Sure, will do that.

agent/engine/dockerclient/dockerclientfactory.go
vsiddharth: I don't think this would work on both `linux` and `windows`. 
Please ensure that this change works well on all supported platforms.

sharanyad:
@aithal @samuelkarp @vsiddharth Addressed the review comments. Could you please have a look again?

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -14,13 +14,19 @@
...
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockeriface"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
 	log "github.com/cihub/seelog"
 	docker "github.com/fsouza/go-dockerclient"
+	"github.com/pkg/errors"
 )
 
+const (
+	// minAPIVersionKey is the docker.Env key for min API version
+	minAPIVersionKey = "MinAPIVersion"
```
aaithal: Can you also document here that this is supported in API versions `>=1.25` as per https://docs.docker.com/engine/api/version-history/#v125-api-changes ?

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -110,7 +116,15 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
 // findDockerVersions loops over all known API versions and finds which ones
 // are supported by the docker daemon on the host
 func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
-	clients := make(map[DockerVersion]dockeriface.Client)
+	// if the client version returns a min version and api version, then use it to return
+	// Docker versions
```
aaithal: Let's make this more explicit: `use it to return versions between MinAPIVersion and APIVerison`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +135,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	}
+
+	clientVersion, err := client.Version()
+	if err != nil {
+		return nil, errors.Wrap(err, "min API version check: error while getting client version")
+	}
+
+	clients := make(map[DockerVersion]dockeriface.Client)
+	// check if the docker.Env obj has MinAPIVersion key
+	if clientVersion.Exists(minAPIVersionKey) {
```
aaithal: let's refactor this a bit more:
```
if !clientVersion.Exists(minAPIVersionKey) {
 return ...
}

minAPIVersion := clientVersion.Get(minAPIVersionKey)
...
```

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +135,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+}
+
+func getDockerClientForVersion(
+	endpoint string,
+	version string,
+	minAPIVersion string,
+	apiVersion string) (dockeriface.Client, error) {
+	lessThanMinCheck := "<"+minAPIVersion
+	moreThanMaxCheck := ">"+apiVersion
+	minVersionCheck, minErr := utils.Version(version).Matches(lessThanMinCheck)
```
aaithal: I don't think you need multiple `error` variables here. If you stick to the policy of creating/initializing a variable only when needed, you'll do something like this:
```
minVersionCheck, err := utils.Version(version).Matches(lessThanMinCheck)
if err != nil {
  ...
}

maxVersionCheck, err := utils.Version(version).Matches(moreThanMaxCheck)
if err != nil {
 ...
}
if minVersionCheck || ...
```

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -110,7 +116,15 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
 // findDockerVersions loops over all known API versions and finds which ones
 // are supported by the docker daemon on the host
 func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
-	clients := make(map[DockerVersion]dockeriface.Client)
+	// if the client version returns a min version and api version, then use it to return
+	// Docker versions
+	clients, err := findDockerVersionsfromMinMaxVersions(endpoint)
+	if err != nil {
+		log.Debugf("Unable to get Docker clients from min API version: %v", err)
```
aaithal: I think this should be `Infof`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +135,66 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+
+	clients := make(map[DockerVersion]dockeriface.Client)
+	// check if the docker.Env obj has MinAPIVersion key
+	if clientVersion.Exists(minAPIVersionKey) {
+		minAPIVersion := clientVersion.Get(minAPIVersionKey)
+		apiVersion := clientVersion.Get(apiVersionKey)
+		for _, version := range getKnownAPIVersions() {
+			dockerClient, err := getDockerClientForVersion(endpoint, string(version), minAPIVersion, apiVersion)
+			if err != nil {
+				log.Debugf("Unable to get Docker client for version %s: %v", version, err)
```
aaithal: I think this should be `Infof`

agent/utils/compare_versions.go
```diff
@@ -40,7 +43,8 @@ type semver struct {
 // * <x.y.z -- Matches a version less than the selector version
 // * x.y.z,a.b.c -- Matches if the version matches either of the two selector versions
 func (lhs Version) Matches(selector string) (bool, error) {
-	lhsVersion, err := parseSemver(string(lhs))
+	lhsVer := appendIfNoPatch(string(lhs))
```
aaithal: Let's not do `appendIfNoPatch` anywhere in this method. Instead, do that in `getDockerClientForVersion` so that you don't have to keep doing that at multiple places here. Let's minimize changes to `Version` type as much as possible.

agent/engine/dockerclient/dockerclientfactory.go
sharanyad: It makes more sense to have it as `Debugf` to me, since we are just trying to obtain docker client for a particular version and it might not always be successful, which is expected. All the older versions' agent logs would contain these messages looking like errors in the beginning. So made it `Debugf`. What do you think?

agent/utils/compare_versions.go
aaithal: If there are asynchronous failures for containers because of a specific versioned docker client  not being there, having these errors printed out during discovery helps us debug the issue. If we make these `Debug`, we'll lose these messages when logging at default `Info` level.

agent/engine/dockerclient/dockerclientfactory.go
aaithal: If there are asynchronous failures for containers because of a specific versioned docker client  not being there, having these errors printed out during discovery helps us debug the issue. If we make these `Debug`, we'll lose these messages when logging at default `Info` level.

samuelkarp: @aaithal Theoretically...we shouldn't have async failures due to missing clients.  This discovery logic is used for capability registration, and if we don't register a capability there shouldn't be a task scheduled that requires an unavailable client.

aaithal: @samuelkarp yeah, but bugs.. Since this is just a once time discovery thing, logging it at Info shouldn't contribute to too much noise IMO.

agent/engine/dockerclient/dockerclientfactory.go
sharanyad: Earlier was doing a check of the error messages together, didn't modify it later. changed now.

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +139,70 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+		return nil, errors.Wrapf(err, "error while comparing version %s with minAPIVersion %s", version, minAPIVersion)
+	}
+	maxVersionCheck, err := utils.Version(version).Matches(moreThanMaxCheck)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while comparing version %s with apiVersion %s", version, apiVersion)
+	}
+	// do not add the version when it is less than min api version or greater
+	// than api version
+	if minVersionCheck || maxVersionCheck {
+		return nil, errors.Errorf("min and API versions comparison check failed for version: %s", version)
```
aaithal: Can you rephrase this? `"version detection using MinAPIVersion: unsupported version: %s"`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +139,70 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	version += zeroPatch
+	lessThanMinCheck := "<" + minAPIVersion + zeroPatch
+	moreThanMaxCheck := ">" + apiVersion + zeroPatch
+	minVersionCheck, err := utils.Version(version).Matches(lessThanMinCheck)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while comparing version %s with minAPIVersion %s", version, minAPIVersion)
+	}
+	maxVersionCheck, err := utils.Version(version).Matches(moreThanMaxCheck)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while comparing version %s with apiVersion %s", version, apiVersion)
```
aaithal: Can you rephrase this? `"version detection using MinAPIVersion: unable to get max version: %s"`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +139,70 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+	apiVersion string) (dockeriface.Client, error) {
+	// Adding patch number zero to Docker versions to reuse the existing semver
+	// comparator
+	// TODO: remove this logic later when non-semver comparator is implemented
+	version += zeroPatch
+	lessThanMinCheck := "<" + minAPIVersion + zeroPatch
+	moreThanMaxCheck := ">" + apiVersion + zeroPatch
+	minVersionCheck, err := utils.Version(version).Matches(lessThanMinCheck)
+	if err != nil {
+		return nil, errors.Wrapf(err, "error while comparing version %s with minAPIVersion %s", version, minAPIVersion)
```
aaithal: Can you rephrase this? `"version detection using MinAPIVersion: unable to get min version: %s"`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +139,70 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+		return nil, errors.Wrapf(err, "error while comparing version %s with apiVersion %s", version, apiVersion)
+	}
+	// do not add the version when it is less than min api version or greater
+	// than api version
+	if minVersionCheck || maxVersionCheck {
+		return nil, errors.Errorf("min and API versions comparison check failed for version: %s", version)
+	}
+	client, err := newVersionedClient(endpoint, string(version))
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to create Docker client for version: %s", version)
```
aaithal: Can you rephrase this? `"version detection using MinAPIVersion: unable to create Docker client for version: %s"`

agent/utils/compare_versions_test.go
```diff
@@ -185,6 +185,11 @@ func TestVersionMatches(t *testing.T) {
 			selector:       ">=1.5.0",
 			expectedOutput: true,
 		},
+		{
+			version:	"1.17",
```
aaithal: Shouldn't you move this to `engine/dockerclient` package as well?

sharanyad: yeah, this test should not be here, will remove it. There is a general test added in `engine/dockerclient` for getting Docker versions from `minAPIVersion`.

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -121,8 +139,70 @@ func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
...
+		return nil, errors.Wrap(err, "min API version check: error while getting client version")
+	}
+
+	clients := make(map[DockerVersion]dockeriface.Client)
+	// check if the docker.Env obj has MinAPIVersion key
+	if !clientVersion.Exists(minAPIVersionKey) {
+		return nil, errors.New("min API version not found in client version")
+	}
+	minAPIVersion := clientVersion.Get(minAPIVersionKey)
+	apiVersion := clientVersion.Get(apiVersionKey)
```
richardpen: Can you add a check here for `minAPIVersion` or `apiVersion` is empty?

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -110,7 +120,15 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
 // findDockerVersions loops over all known API versions and finds which ones
 // are supported by the docker daemon on the host
 func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
-	clients := make(map[DockerVersion]dockeriface.Client)
+	// if the client version returns a MinAPIVersion and APIVersion, then use it to return
+	// all the Docker clients between MinAPIVersion and APIVersion
+	clients, err := findDockerVersionsfromMinMaxVersions(endpoint)
```
petderek: Two concerns here. The first is that this is going to behave differently on windows. Windows only supports API version 1.24 or greater. If we put this check here, it will skip over the capability logic that ensures tasks are schedule-able to windows.

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -110,7 +120,15 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
...
+	if err != nil {
+		log.Infof("Unable to get Docker clients from min API version: %v", err)
+	} else {
+		return clients
```
petderek: The second concern is tangential. Putting this here puts a hard branch in our detection logic, and I think we should consider composing it differently.

```golang
// As written, the code boils down to this
clients, err := findDockerVersionsOneWay()
if err == nil {
    return clients
}
clients, err := findDockerVersionsADifferentWay()
```
I'd rather see us handle the detection in the same place using existing logic instead of branching like this. One suggestion would be to look at the `version_support_{platform}` files and the Min version variable. You could write a function that does something like this:

```golang
 // Use either the docker minimum API version or rely on the version set in version_support_{platform}.go
func getMinVersion() DockerVersion {
    if functionThatChecksIfDockerHasMinAPIParam() {
        return functionThatExtractsMinAPIFromDockerAPI()
    } else {
        return minDockerAPIVersion
    }
}

File: agent/engine/dockerclient/dockerclientfactory.go (hunk 27 above)
Comment (aaithal) [reply agent/engine/dockerclient/dockerclientfactory.go]: Yeah, I agree. I missed this in my review. Good call!

File: agent/engine/dockerclient/dockerclientfactory.go (hunk 28 above)
Comment (petderek) [reply agent/engine/dockerclient/dockerclientfactory.go]: Oh &mdash; You'd want to write a function to check for the max version, too :)

File: agent/engine/dockerclient/dockerclientfactory.go (hunk 27 above)
Comment (sharanyad) [reply agent/engine/dockerclient/dockerclientfactory.go]: According to current logic for `FindSupportedAPIVersions`, if minimum version client `1.17` is not supported in Windows, the tasks should anyway not be scheduled on Windows right?

File: agent/engine/dockerclient/dockerclientfactory.go (hunk 28 above)
Comment (sharanyad) [reply agent/engine/dockerclient/dockerclientfactory.go]: What would be the default max version here be?

File: agent/engine/dockerclient/dockerclientfactory.go (hunk 27 above)
Comment (samuelkarp) [reply agent/engine/dockerclient/dockerclientfactory.go]: No, not quite right.  We register capabilities for API versions below 1.24 on Windows because we still want tasks that use features in those APIs to work on Windows too.  Our capability logic is pretty naive right now: we map a set of features to the API version in which they were introduced and then use that as an indicator of the minimum version of Docker they'll run on.  Because Docker on Windows still supports (many of) the features that work on older versions of Docker, this was an inexpensive way to get that support without rebuilding our capability logic.

I agree that this is confusing and non-obvious.  I think long-term we probably want to revisit this logic and see if we can improve it, given that we now know different things from when we did when it was designed.  However, we haven't done that yet and we still need to register capabilities for older API versions on Windows.

File: agent/engine/dockerclient/dockerclientfactory.go [hunk 29]
```diff
@@ -110,19 +120,72 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
 // findDockerVersions loops over all known API versions and finds which ones
 // are supported by the docker daemon on the host
 func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
+	// if the client version returns a MinAPIVersion and APIVersion, then use it to return
+	// all the Docker clients between MinAPIVersion and APIVersion, else try pinging
+	// the clients in getKnownAPIVersions
+	minAPIVersion := ""
```
vsiddharth: This can be composed more idiomatically
`var minAPIVersion, apiVersion string`

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -110,19 +120,72 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
...
+		lessThanMinCheck := "<" + minAPIVersion + zeroPatch
+		moreThanMaxCheck := ">" + apiVersion + zeroPatch
+		minVersionCheck, err := utils.Version(versionWithPatch).Matches(lessThanMinCheck)
+		if err != nil {
+			return nil, errors.Wrapf(err, "version detection using MinAPIVersion: unable to get min version: %s", minAPIVersion)
+		}
+		maxVersionCheck, err := utils.Version(versionWithPatch).Matches(moreThanMaxCheck)
 		if err != nil {
-			log.Infof("Failed to ping with Docker version %s: %v", version, err)
+			return nil, errors.Wrapf(err, "version detection using MinAPIVersion: unable to get max version: %s", apiVersion)
```
vsiddharth: Should this be `MinAPIVersion`?

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -110,19 +120,72 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
...
+		}
+		// do not add the version when it is less than min api version or greater
+		// than api version
+		if minVersionCheck || maxVersionCheck {
+			return nil, errors.Errorf("version detection using MinAPIVersion: unsupported version: %s", version)
```
vsiddharth: Should this be `MinAPIVersion` again ?

agent/engine/dockerclient/dockerclientfactory.go
sharanyad: Got it. 
From reading today's code, in `findDockerVersions` for windows, we iterate through the `KnownAPIVersions` from 1.17. Wouldn't trying to create a client with version 1.17 or trying to ping, fail in Docker Windows? Eventually, those versions would not be added to the  list of clients that Docker supports and hence the capability for 1.17 would not be added? 
https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/dockerclient/dockerclientfactory.go#L115

samuelkarp: We don't actually want to create a client at 1.17 on Windows or even to attempt to ping it; we know that it'll fail (except on Docker 1.12 beta).  I'm talking strictly about capability registration here in that we still need to register those capabilities (and then override the incoming API version with the minimum of 1.24).

sharanyad: But yes, I agree we should revisit this logic sometime and adapt it for windows.

agent/engine/dockerclient/dockerclientfactory.go
sharanyad: No, we are simply logging the version from from `knownVersions` , which does not fall in the range `[minAPIVersion, apiVersion]`

agent/engine/dockerclient/dockerclientfactory.go
sharanyad: No, this means something was wrong with `apiVersion` and the comparison check failed.

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -110,19 +120,71 @@ func (f *factory) getClient(version DockerVersion) (dockeriface.Client, error) {
 // findDockerVersions loops over all known API versions and finds which ones
 // are supported by the docker daemon on the host
 func findDockerVersions(endpoint string) map[DockerVersion]dockeriface.Client {
+	// if the client version returns a MinAPIVersion and APIVersion, then use it to return
+	// all the Docker clients between MinAPIVersion and APIVersion, else try pinging
+	// the clients in getKnownAPIVersions
+	var minAPIVersion, apiVersion string
+	// get a Docker client with the default supported version
+	client, err := newVersionedClient(endpoint, string(minDockerAPIVersion))
+	if err == nil {
```
samuelkarp: I would find this easier to read if it were refactored a bit and the conditions were inverted, something like this:

```go
func findDockerVersionsFromAPI(endpoint string) (int, int, bool) {
	client, err := newVersionedClient(endpoint, string(minDockerAPIVersion))
	if err != nil {
		seelog.Debug("Failed to query the version API", err)
		return "", "", false
	}

	clientVersion, err := client.Version()
	if err != nil {
		seelog.Debug("Failed to extract version", err)
		return "", "", false
	}

	if !clientVersion.Exists(minAPIVersionKey) || !clientVersion.Exists(apiVersionKey) {
		seelog.Debug("Failed to extract minimum and maximum version")
		return "", "", false
	}

	return clientVersion.Get(minAPIVersionKey), clientVersion.Get(apiVersionKey), true
}
```

This way the error conditions are more clear and there's less to keep in mind as you're reading through.

PR #1022, opened 2017-10-18
Adding task timestamp in task state change

richardpen:
### Summary

Adding new fields in task state change, specifically the task pull started timestamp, pull stopped timestamp and container stop timestamp.

### Implementation details

The timestamp for pulling start and stop was sent together with the task running, while container stop timestamp was sent along with task stopped.

### Testing

- [x] Manual test
New tests cover the changes: 
yes
### Licensing

yes


agent/api/task.go
```diff
@@ -847,6 +857,127 @@ func (task *Task) GetTaskENI() *ENI {
 	return task.ENI
 }
 
+// SetPullStartedAt sets the task pullstartedat timestamp and returns whether
+// this field was updated or not
+func (task *Task) SetPullStartedAt(timestamp time.Time) bool {
+	task.knownStatusTimeLock.Lock()
+	defer task.knownStatusTimeLock.Unlock()
+
+	// Only set this field if it is not set
```
sharanyad: Does this mean we don't need to reset this incase of pull retries? 
If so, can we log this somewhere?

agent/engine/docker_task_engine.go
```diff
@@ -568,10 +581,16 @@ func (engine *DockerTaskEngine) serialPull(task *api.Task, container *api.Contai
...
 	defer ImagePullDeleteLock.Unlock()
 
-	pullStart := time.Now()
+	// Recrod the task pull_started_at timestamp
+	pullStart := engine.time().Now()
 	defer func(startTime time.Time) {
 		seelog.Infof("Finished pulling container %v in %s. Task: %v", container.Image, time.Since(startTime).String(), task)
 	}(pullStart)
+
+	ok := task.SetPullStartedAt(pullStart)
```
sharanyad: Should we set `PullStartedAt` time here or before the actual `client.PullImage` call in `pullAndUpdateContainerReference` method ?

agent/api/ecsclient/client.go
```diff
@@ -312,13 +312,14 @@ func (client *APIECSClient) SubmitTaskStateChange(change api.TaskStateChange) er
...
-	// Submit task state change
-	if change.Status == api.TaskStatusNone {
+	// Submit task state change or ExecutionStoppedAt timestamp
+	if change.Status == api.TaskStatusNone && change.ExecutionStoppedAt == nil {
 		seelog.Warnf("SubmitTaskStateChange called with an invalid change: %s", change.String())
 		return errors.New("ecs api client: SubmitTaskStateChange called with an invalid change")
 	}
 
-	if change.Status != api.TaskRunning && change.Status != api.TaskStopped {
+	// It's possible to submit ExecutionStoppedAt timestamp with task status in NONE
```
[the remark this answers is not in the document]
samuelkarp: This comment is extremely confusing.

agent/engine/docker_task_engine.go
```diff
@@ -538,13 +538,19 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
 			return engine.client.ImportLocalEmptyVolumeImage()
 		}
 	}
+
+	// Record the pullStoppedAt timestamp
```
samuelkarp: What happens if this doesn't happen?  What happens if the agent crashes right before this code executes?

agent/engine/task_manager.go
```diff
@@ -690,3 +699,42 @@ func (mtask *managedTask) waitForStopReported() bool {
 	}
 	return taskStopped
 }
+
+// sendExecutionStoppedAtTimestamp start the timer for sending execution stopped at timestamp
```
samuelkarp: What exactly is this timer used for?

agent/engine/task_manager.go
```diff
@@ -26,6 +26,7 @@ import (
 
 const (
 	steadyStateTaskVerifyInterval         = 5 * time.Minute
+	executionStoppedAtSubmitDelay         = 12 * time.Hour
```
samuelkarp: This constant needs a comment.  I don't know what it represents or why it's set to 12 hours.

agent/engine/task_manager.go
```diff
@@ -690,3 +699,42 @@ func (mtask *managedTask) waitForStopReported() bool {
...
+	seelog.Info("Essential container stopped, start sending timestamp for stopped")
+
+	executionStoppedAtElapsed := time.Now().Sub(*mtask.Task.GetExecutionStoppedAt())
+	maxDelay := executionStoppedAtSubmitDelay - executionStoppedAtElapsed
+
+	if mtask.submitExecutionStoppedTimer != nil {
+		mtask.submitExecutionStoppedTimer.Stop()
+	}
+
+	mtask.submitExecutionStoppedTimer = mtask.time().AfterFunc(maxDelay, func() {
```
samuelkarp: It's possible for this timer to not be active if the agent restarts while the timer has not elapsed.  What happens then?

agent/api/task.go
richardpen: We only need to set this whenever we start the first pull, no need to reset this for retry. It's already logged [here](https://github.com/aws/amazon-ecs-agent/pull/1022/files#diff-b9aa12c3425c34d6f330f28407f779f6R569). Or do you want something else to be logged too?

agent/engine/docker_task_engine.go
richardpen: I don't think so, since there is a lock inside the `serialPull` and `concurrentPull`, it would be more accurate to log here, otherwise if the pull is waiting for this lock, the actual `PullStartedAt` would be incorrect.

agent/api/ecsclient/client.go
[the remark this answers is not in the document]
richardpen: I will try to add more details in the comment to make it more readable.

agent/engine/task_manager.go
richardpen: On agent restart, it will resend current status [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/task_manager.go#L114), the timestamp will be sent at that point.

agent/engine/task_manager.go
richardpen: We want this timestamp to be sent along with the task STOPPED event to save 
 one extra SubmitTaskSTateChange call, but if the task failed to stop within `executionStoppedAtSubmitDelay` this timer will be triggered to send the timestamp.

agent/engine/task_manager.go
richardpen: will add comment, actually this is the maximum delay can be accepted by backend.

agent/engine/docker_task_engine.go
richardpen: If this doesn't happen, this field will be empty and the task won't have `pullstoppedat` timestamps. Are you concerning about container moved to stopped without pulling?  If agent crashes here, agent should try to pull the image again when it start, then it will be set by that time.

agent/api/tasktimestamp.go
```diff
@@ -0,0 +1,82 @@
...
+package api
+
+import (
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+// TaskTimestamp is an enumeration of valid state of task timestamp
+type TaskTimestamp int32
```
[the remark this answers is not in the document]
richardpen: `int32` is picked up because it was used for other "enumeration", any suggestion?

agent/api/task.go
```diff
@@ -847,6 +857,127 @@ func (task *Task) GetTaskENI() *ENI {
...
+	}
+	return false
+}
+
+// GetPullStartedAt returns the PullStartedAt timestamp in UTC format
+func (task *Task) GetPullStartedAt() *time.Time {
+	task.knownStatusTimeLock.RLock()
+	defer task.knownStatusTimeLock.RUnlock()
+
+	if task.PullStartedAt == (time.Time{}) {
```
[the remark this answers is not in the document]
richardpen: That means this field has already been set, that's the pull of the task has already started. i.e:  a task with two containers, we only set this field when start pulling the first container, and when pull the second container this field won't be updated. The `nil` is to indicate whether this field is updated or not( or whether this is pulling the first container or not).

agent/api/task.go
```diff
@@ -847,6 +857,127 @@ func (task *Task) GetTaskENI() *ENI {
...
 }
 
+// SetPullStartedAt sets the task pullstartedat timestamp and returns whether
+// this field was updated or not
+func (task *Task) SetPullStartedAt(timestamp time.Time) bool {
+	task.knownStatusTimeLock.Lock()
+	defer task.knownStatusTimeLock.Unlock()
+
+	// Only set this field if it is not set
+	if task.PullStartedAt == (time.Time{}) {
```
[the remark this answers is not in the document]
richardpen: It won't compile without the parens. [example](https://play.golang.org/p/4v9NGaZJfT)

agent/ecs_client/model/api/api-2.json
```diff
@@ -42,20 +42,6 @@
         {"shape":"UnsupportedFeatureException"}
       ]
     },
-    "CreateServiceLinkedRole":{
```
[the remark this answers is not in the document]
richardpen: Actually this was removed in the latest model, I'll check to see if I can bring it back.

agent/api/task.go
```diff
@@ -97,6 +97,15 @@ type Task struct {
 	KnownStatusTimeUnsafe time.Time `json:"KnownTime"`
 	knownStatusTimeLock   sync.RWMutex
 
+	// PullStartedAt is the timestamp when the task start pulling
```
aaithal: This needs a longer description. Especially because there's no `pull` for the task (same with other timestamp variables you've added to this struct.

agent/api/task.go
```diff
@@ -847,6 +857,127 @@ func (task *Task) GetTaskENI() *ENI {
 	return task.ENI
 }
 
+// SetPullStartedAt sets the task pullstartedat timestamp and returns whether
+// this field was updated or not
+func (task *Task) SetPullStartedAt(timestamp time.Time) bool {
+	task.knownStatusTimeLock.Lock()
```
aaithal: Can we consolidate all of these locks into a single lock? It's hard to keep track of different locks for different methods. Let's have a single `sync.RWMutex` lock for the task. I understand that doing that is beyond the mandate of this PR. But, this is as good a time as any to make that change.

agent/api/task.go
```diff
@@ -847,6 +857,127 @@ func (task *Task) GetTaskENI() *ENI {
...
+
+	task.ExecutionStoppedAt = timestamp
+}
+
+// GetExecutionStoppedAt returns the task executionStoppedAt timestamp in UTC format
+func (task *Task) GetExecutionStoppedAt() *time.Time {
+	task.knownStatusTimeLock.RLock()
+	defer task.knownStatusTimeLock.RUnlock()
+
+	if task.ExecutionStoppedAt == (time.Time{}) {
```
aaithal: Why can we just return `task.ExecutionStoppedAt` here and let the `SubmitTaskStateChange` API handle the UTC part of it? It seems like we're adding extra computation here and increasing the lock scope and I'd prefer to do those cosmetic operations outside of the lock scope.

agent/api/ecsclient/client.go
```diff
@@ -312,13 +312,15 @@ func (client *APIECSClient) SubmitTaskStateChange(change api.TaskStateChange) er
 		return nil
 	}
 
-	// Submit task state change
-	if change.Status == api.TaskStatusNone {
+	if change.Status == api.TaskStatusNone && change.ExecutionStoppedAt == nil {
```
aaithal: Why was this updated?

agent/eventhandler/task_handler.go
```diff
@@ -217,6 +217,17 @@ func (handler *TaskHandler) SubmitTaskEvents(taskEvents *eventList, client api.E
 					seelog.Debugf("TaskHandler, Submitted task state change: %s", event.String())
 					backoff.Reset()
 					taskEvents.events.Remove(eventToSubmit)
+
+					// ExecutionStoppedAt can also be sent here
+					if event.taskChange.ExecutionStoppedAt != nil {
```
aaithal: This logic is getting complicated. If we want to track a couple more timestamps, it'll get even more complicated. Instead of this, let's just have a boolean for each kind of timestamp that we track and have a `Sent` status field to track it.

agent/api/task.go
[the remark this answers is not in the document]
samuelkarp: `time.Time{}` is the zero value.  Why not just use that?  What additional value does `nil` add?

samuelkarp:
Can you rebase and fix the merge conflicts?

agent/api/task.go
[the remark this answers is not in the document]
richardpen: This makes it easier [here](https://github.com/aws/amazon-ecs-agent/pull/1022/files#diff-b9aa12c3425c34d6f330f28407f779f6R363), I think &time.Time{} is different from nil, so I don't need additional check.

agent/api/task.go
```diff
@@ -97,6 +97,15 @@ type Task struct {
 	KnownStatusTimeUnsafe time.Time `json:"KnownTime"`
 	knownStatusTimeLock   sync.RWMutex
 
+	// PullStartedAt is the timestamp when the task start pulling
+	PullStartedAt time.Time `json:"PullStartedAt"`
+	// PullStoppedAt is the timestamp when the task finished pulling
+	PullStoppedAt time.Time `json:"PullStoppedAt"`
```
adnxn: since we're using pointers to the `time.Time` fields in the Get* methods - would it make sense have these fields use pointers as well? 

then we can do nil checks (which i think would be clearer) instead of the `task.PullStartedAt == (time.Time{})` and also we'd avoid going back and forth between using values in some places and the pointers in others. 

also the marshaled zero value for `time.Time` is noisy if we need to debug.

agent/engine/docker_task_engine.go
```diff
@@ -554,10 +560,17 @@ func (engine *DockerTaskEngine) concurrentPull(task *api.Task, container *api.Co
 	defer seelog.Debugf("Released ImagePullDeleteLock after pulling image - %s. Task: %v", container.Image, task)
 	defer ImagePullDeleteLock.RUnlock()
 
-	pullStart := time.Now()
+	// Recrod the task pull_started_at timestamp
```
adnxn: typo: `Record`

agent/engine/docker_task_engine.go
```diff
@@ -568,10 +581,16 @@ func (engine *DockerTaskEngine) serialPull(task *api.Task, container *api.Contai
 	defer seelog.Debugf("Released ImagePullDeleteLock after pulling image - %s. Task: %v", container.Image, task)
 	defer ImagePullDeleteLock.Unlock()
 
-	pullStart := time.Now()
+	// Recrod the task pull_started_at timestamp
```
adnxn: typo: `Record`

agent/api/ecsclient/client.go
```diff
@@ -312,13 +312,15 @@ func (client *APIECSClient) SubmitTaskStateChange(change api.TaskStateChange) er
...
 		seelog.Warnf("SubmitTaskStateChange called with an invalid change: %s", change.String())
 		return errors.New("ecs api client: SubmitTaskStateChange called with an invalid change")
 	}
 
-	if change.Status != api.TaskRunning && change.Status != api.TaskStopped {
+	// Submit task ExecutionStoppedAt timestamp, the status in the change
+	// can be RUNNING, STOPPED or NONE(essential container stopped before task
```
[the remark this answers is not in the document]
richardpen: Yes, to be more specific, it's when task has desiredstatus == stopped while knownstatus < running

agent/api/task.go
richardpen: Record as a pointer in the task struct and return the pointer in getter will cause some data race.  It looks like the time.Time{} won't be saved in Marshal, [example](https://play.golang.org/p/ZeJ0RyyBrk)

agent/api/task.go
```diff
@@ -778,99 +791,237 @@ func (task *Task) SetKnownStatus(status TaskStatus) {
...
 }
 
+// SetPullStartedAt sets the task pullstartedat timestamp and returns whether
+// this field was updated or not
+func (task *Task) SetPullStartedAt(timestamp time.Time) bool {
+	task.statusLock.Lock()
+	defer task.statusLock.Unlock()
+
+	// Only set this field if it is not set
+	if task.PullStartedAt == (time.Time{}) {
```
aaithal: can you just use https://golang.org/pkg/time/#Time.IsZero here instead?

agent/engine/task_manager.go
aaithal: As per offline conversation, you shouldn't need to do this at all. The ticker in `eventhandler.TaskHandler` should take care of batching and sending all unsent events.

agent/api/task.go
```diff
@@ -778,99 +791,237 @@ func (task *Task) SetKnownStatus(status TaskStatus) {
...
+func (task *Task) GetExecutionStoppedAtSent() bool {
+	task.statusLock.RLock()
+	defer task.statusLock.RUnlock()
+
+	return task.ExecutionStoppedAtSent
+}
+
+// ShouldSendExecutionStoppedTime determins whether the container state change
+// should cause ExecutionStoppedAt to be sent
+func (task *Task) ShouldSendExecutionStoppedTime(container *Container, timestamp time.Time) bool {
```
aaithal: You'd need to change this based on my recommendation for getting rid of the timer for this.

agent/engine/task_manager.go
```diff
@@ -690,3 +699,42 @@ func (mtask *managedTask) waitForStopReported() bool {
...
+	mtask.lock.RLock()
+	defer mtask.lock.RUnlock()
+
+	err := mtask.Task.StatusShouldBeSent()
+	if err != nil {
+		seelog.Debugf("Task status not recognized by backend or no need to sent, task: %s, msg: %v", task.String(), err)
+		return
+	}
+
+	// stop the timer for submitting executionStoppedAt if this is a task stopped event
```
aaithal: As per my earlier comments, you shouldn't need to do this at all

agent/api/statechange.go
```diff
@@ -80,24 +90,69 @@ func (c *ContainerStateChange) String() string {
...
+	if timestamp := change.Task.GetPullStoppedAt(); !timestamp.IsZero() {
+		change.PullStoppedAt = aws.Time(timestamp.UTC())
+	}
+	if timestamp := change.Task.GetExecutionStoppedAt(); !timestamp.IsZero() {
+		change.ExecutionStoppedAt = aws.Time(timestamp.UTC())
+	}
+}
+
+// ShouldBeReport checks if the statechange should be reported to backend
+func (change *TaskStateChange) ShouldBeReport() bool {
```
samuelkarp: ShouldReport or ShouldBeReported

But I think we already have a way of determining whether state changes should be reported or not.  Can you explain what's different about this logic?

richardpen: The logic today only allow "RUNNING" and "STOPPED" status to be sent for task state change. While we also want to send the container change if the task is in "PENDING", eg container failed to create.

agent/engine/docker_task_engine.go
```diff
@@ -572,10 +583,16 @@ func (engine *DockerTaskEngine) concurrentPull(task *api.Task, container *api.Co
 	defer seelog.Debugf("Released ImagePullDeleteLock after pulling image - %s. Task: %v", container.Image, task)
 	defer ImagePullDeleteLock.RUnlock()
 
-	pullStart := time.Now()
+	// Record the task pull_started_at timestamp
+	pullStart := engine.time().Now()
 	defer func(startTime time.Time) {
 		seelog.Infof("Finished pulling container %v in %s. Task: %v", container.Image, time.Since(startTime).String(), task)
 	}(pullStart)
+	ok := task.SetPullStartedAt(pullStart)
```
sharanyad: Since we are setting `PullStartedAt` in both `concurrentPull` as well as `serialPull`methods, can we just set once in  `pullAndUpdateContainerReference`, which is called by `concurrentPull` and `serialPull` ?

richardpen: Ideally we don't want to have duplicate code here, but since we want to set the timestamp after acquired the lock, we can't set in the upper level in `pullAndUpdateContainerReference`.

agent/api/statechange.go
```diff
@@ -80,24 +90,69 @@ func (c *ContainerStateChange) String() string {
...
+	if timestamp := change.Task.GetPullStoppedAt(); !timestamp.IsZero() {
+		change.PullStoppedAt = aws.Time(timestamp.UTC())
+	}
+	if timestamp := change.Task.GetExecutionStoppedAt(); !timestamp.IsZero() {
+		change.ExecutionStoppedAt = aws.Time(timestamp.UTC())
+	}
+}
+
+// ShouldBeReported checks if the statechange should be reported to backend
+func (change *TaskStateChange) ShouldBeReported() bool {
```
aaithal: please add a `TODO` here for dealing with network attachments in the future

agent/engine/task_manager.go
```diff
@@ -312,6 +312,15 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 		}
 	}
 
+	// If the essential container is stopped, set the ExecutionStoppedAt timestamp
+	if container.GetKnownStatus() == api.ContainerStopped && container.IsEssential() {
+		now := mtask.time().Now()
+		ok := mtask.Task.SetExecutionStoppedAt(now)
+		if ok {
+			seelog.Infof("Essential container in task stopped, task %s, time: %s", mtask.Task.String(), now)
```
aaithal: shouldn't this be be `now.String()`?

aaithal: Also, let's be more clear here that we're recording the execution stopped time: `"Recording execution stopped time for a task, essential container in task stopped"`

agent/engine/docker_task_engine.go
```diff
@@ -572,10 +583,16 @@ func (engine *DockerTaskEngine) concurrentPull(task *api.Task, container *api.Co
...
+	if ok {
+		seelog.Infof("Task started pull, task %s, time: %s", task.String(), pullStart)
```
aaithal: Let's change this message to be more clear about timestamp being recorded as well: `"Recording timestamp for starting image pull, task: ..."`

agent/engine/docker_task_engine.go
```diff
@@ -586,10 +603,15 @@ func (engine *DockerTaskEngine) serialPull(task *api.Task, container *api.Contai
...
 	defer ImagePullDeleteLock.Unlock()
 
-	pullStart := time.Now()
+	pullStart := engine.time().Now()
 	defer func(startTime time.Time) {
 		seelog.Infof("Finished pulling container %v in %s. Task: %v", container.Image, time.Since(startTime).String(), task)
 	}(pullStart)
+	ok := task.SetPullStartedAt(pullStart)
+	if ok {
+		seelog.Infof("Task started pull, task %s, time: %s", task.String(), pullStart)
```
aaithal: Here too. Let's be more explicit about recording a timestamp here:  `"Recording timestamp for starting image pull, task: ..."`

agent/engine/engine_unix_integ_test.go
```diff
@@ -845,9 +846,9 @@ func TestStartStopWithSecurityOptionNoNewPrivileges(t *testing.T) {
 	assert.Equal(t, event.(api.TaskStateChange).Status, api.TaskRunning, "Expected task to be RUNNING")
 
 	// Kill the existing container now
-	taskUpdate := *testTask
+	taskUpdate := createTestTask(testArn)
```
aaithal: thanks for this change. This was leading to race conditions in my branch.

PR #1025, opened 2017-10-20
Add flag to print capabilities

jahkeup:
### Summary
Print the Agent's capabilities without having to connect it to ECS.

### Implementation details
This uses the Agent's existing capability discovery mechanism and prints the values without starting the Agent.

### Testing
This was run on an ec2 instance with the added flag.

New tests cover the changes: 

Agent offline capability discovery

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -71,6 +71,9 @@ var (
 type agent interface {
 	// printVersion prints the Agent version string
 	printVersion() int
+	// printECSAttributes prints the Agent's capabilities based on
+	// its environment
+	printECSAttributes() int
 	// start starts the Agent execution
 	start() int
 }
@@ -160,6 +163,15 @@ func newAgent(
 // printVersion prints the ECS Agent version string
 func (agent *ecsAgent) printVersion() int {
 	version.PrintVersion(agent.dockerClient)
+	return exitcodes.ExitSuccess
+}
+
+// printECSAttributes prints the Agent's ECS Attributes based on its
+// environment
+func (agent *ecsAgent) printECSAttributes() int {
+	for _, attr := range agent.capabilities() {
+		fmt.Printf("%s\t%s\n", aws.StringValue(attr.Name), aws.StringValue(attr.Value))
+	}
 	return exitcodes.ExitSuccess
 }
 
--- a/agent/app/args/flag.go
+++ b/agent/app/args/flag.go
@@ -21,9 +21,11 @@ const (
 	acceptInsecureCertUsage  = "Disable SSL certificate verification. We do not recommend setting this option"
 	licenseUsage             = "Print the LICENSE and NOTICE files and exit"
 	blacholeEC2MetadataUsage = "Blackhole the EC2 Metadata requests. Setting this option can cause the ECS Agent to fail to work properly.  We do not recommend setting this option"
+	ecsAttributesUsage       = "Print the Agent's ECS Attributes based on its environment"
 
 	versionFlagName              = "version"
 	logLevelFlagName             = "loglevel"
+	ecsAttributesFlagName        = "ecs-attributes"
 	acceptInsecureCertFlagName   = "k"
 	licenseFlagName              = "license"
 	blackholeEC2MetadataFlagName = "blackhole-ec2-metadata"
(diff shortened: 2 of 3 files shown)


petderek:
Thers

agent/app/agent.go
```diff
@@ -71,6 +71,8 @@ var (
 type agent interface {
 	// printVersion prints the Agent version string
 	printVersion() int
+	// printCapabilities prints the Agent's capabilities
+	printCapabilities() int
```
petderek: I like the overall interface here. I'm a bit concerned that when future consumers of this interface see it, they will assume that capabilities are strictly bound to agent version.  We should make it clear that capabilities aren't only dependent on agent version, but also the environment and runtime version.

agent/app/args/flag.go
```diff
@@ -21,9 +21,11 @@ const (
 	acceptInsecureCertUsage  = "Disable SSL certificate verification. We do not recommend setting this option"
 	licenseUsage             = "Print the LICENSE and NOTICE files and exit"
 	blacholeEC2MetadataUsage = "Blackhole the EC2 Metadata requests. Setting this option can cause the ECS Agent to fail to work properly.  We do not recommend setting this option"
+	capabilitiesUsage        = "Print the agent capabilities and exit"
```
petderek: I'd change this to read 'print the capabilities given the current environment and agent' or something like that.

agent/app/args/flag.go
```diff
@@ -21,9 +21,11 @@ const (
...
 
 	versionFlagName              = "version"
 	logLevelFlagName             = "loglevel"
+	capabilitiesFlagName         = "capabilities"
```
petderek: Likewise, I'd consider naming the flag `-environmentCapabilities`, `-hostCapabilities` or something similar that makes that association explicit.

agent/app/run.go
```diff
@@ -49,11 +49,15 @@ func Run(arguments []string) int {
 		return exitcodes.ExitError
 	}
 
-	if *parsedArgs.Version {
-		// Print version and exit
+	switch {
```
petderek: 👍

agent/app/args/flag.go
sharanyad: Since we also support Linux Capabilities in ECS now, naming this to agentCapabilities or something different can make it less confusing.

jahkeup: I would vote for `agentCapabilities`, which, with a few docstring changes, would accurately describe how and what these capabilities are IMO.

jahkeup: How do we feel about dashes? (ie: `agent-capabilities`) I personally don't like camelCased flags.

agent/app/args/flag.go
jahkeup: How about this: `Prints the supported Agent capabilites based on its environment`

agent/app/args/flag.go
petderek: Naming is hard. What about `runtime-capabilities`? Its invocation would look like:
```sh
$> ./agent -runtime-capabilites
```
camelCased flags are terrible; I don't know what I was thinking.

jahkeup: That name implies unintended meaning as `runtime` is a very overloaded term. In this case it could refer to runtime resolution of capabilities or the _runtime_'s capabilties (ie: docker). I'd favor something that is clearly the Agent's capabilities. 

    ./agent -agent-capabilities 

This one still looks the most accurate to me, ignoring the studder in this case. The help message for the flag itself would clarify how the agent-capabilities are determined and their significance.

aaithal: why not `./agent -capabilities`

jahkeup: The issue raised is that the term "capabilities" is grossly overloaded in our technical domain. The flag name `agent-capabilities` makes as much sense and has the added benefit of being scoped both in name and in operation to the agent's idea of a "capability".

aaithal: Makes sense. Might I also suggest  `./agent -ecs-attributes`?

jahkeup: If we're going that direction, I'd prefer to use a combination: `ecs-capabilities`. Though the data is represented as attributes these attributes are presented and understood as capabilities (they're prefixed with `com.amazonaws.ecs.capability`). Please correct me if I've misunderstood this data. I'm going to prepare a change with the name `ecs-capabilities` for final submission in the meantime.

agent/app/agent.go
```diff
@@ -163,6 +166,27 @@ func (agent *ecsAgent) printVersion() int {
...
+// environment
+func (agent *ecsAgent) printECSAttributes() int {
+	for _, attr := range agent.capabilities() {
+		var (
+			name  string
+			value string
+		)
+
+		if attr.Name != nil {
+			name = *attr.Name
```
aaithal: minor: please use `aws.StringValue()` here

agent/app/agent.go
```diff
@@ -163,6 +166,27 @@ func (agent *ecsAgent) printVersion() int {
...
+		}
+		if attr.Value != nil {
+			value = *attr.Value
```
aaithal: minor: please use `aws.StringValue()` here

PR #1026, opened 2017-10-20
Non essential container state change bug fix

aaithal:
### Summary

Fixes the bug wherein non-essential container state changes were not being submitted to ECS

### Implementation details

Added a ticker to drain the event queue every minute. If
there are any container state changes not submitted,
the ticker ensures that they do get submitted.

### Testing

Will add an entry in CHANGELOG.md

### Licensing

### DIFF
--- a/agent/eventhandler/task_handler.go
+++ b/agent/eventhandler/task_handler.go
@@ -50,8 +50,7 @@ type TaskHandler struct {
 	// submitSemaphore for the number of tasks that may be handled at once
 	submitSemaphore utils.Semaphore
 	// taskToEvents is arn:*eventList map so events may be serialized per task
-	//TODO: fix leak, currently items are never removed from this map
-	tasksToEvents map[string]*eventList
+	tasksToEvents map[string]*taskSendableEvents
 	// tasksToContainerStates is used to collect container events
 	// between task transitions
 	tasksToContainerStates map[string][]api.ContainerStateChange
@@ -71,8 +70,8 @@ type TaskHandler struct {
 	ctx                  context.Context
 }
 
-// eventList is used to group all events for a task
-type eventList struct {
+// taskSendableEvents is used to group all events for a task
+type taskSendableEvents struct {
 	// events is a list of *sendableEvents
 	events *list.List
 	// sending will check whether the list is already being handlerd
@@ -81,13 +80,8 @@ type eventList struct {
 	lock sync.Mutex
 	// createdAt is a timestamp for when the event list was created
 	createdAt time.Time
-	// taskARN is the task arn that the event list is associated
+	// taskARN is the task arn that the event list is associated with
 	taskARN string
-}
-
-func (events *eventList) toStringUnsafe() string {
-	return fmt.Sprintf("Task event list [taskARN: %s, sending: %t, createdAt: %s]",
-		events.taskARN, events.sending, events.createdAt.String())
 }
 
 // NewTaskHandler returns a pointer to TaskHandler
@@ -98,7 +92,7 @@ func NewTaskHandler(ctx context.Context,
 	// Create a handler and start the periodic event drain loop
 	taskHandler := &TaskHandler{
 		ctx:                    ctx,
-		tasksToEvents:          make(map[string]*eventList),
+		tasksToEvents:          make(map[string]*taskSendableEvents),
 		submitSemaphore:        utils.NewSemaphore(concurrentEventCalls),
 		tasksToContainerStates: make(map[string][]api.ContainerStateChange),
 		stateSaver:             stateManager,
@@ -109,6 +103,33 @@ func NewTaskHandler(ctx context.Context,
 	go taskHandler.startDrainEventsTicker()
 
 	return taskHandler
+}
+
+// AddStateChangeEvent queues up a state change for sending using the given client.
+func (handler *TaskHandler) AddStateChangeEvent(change statechange.Event, client api.ECSClient) error {
+	handler.lock.Lock()
+	defer handler.lock.Unlock()
+
+	switch change.GetEventType() {
+	case statechange.TaskEvent:
+		event, ok := change.(api.TaskStateChange)
+		if !ok {
+			return errors.New("eventhandler: unable to get task event from state change event")
+		}
+		handler.flushBatchUnsafe(&event, client)
+		return nil
+
+	case statechange.ContainerEvent:
+		event, ok := change.(api.ContainerStateChange)
+		if !ok {
+			return errors.New("eventhandler: unable to get container event from state change event")
+		}
+		handler.batchContainerEventUnsafe(event)
+		return nil
+
+	default:
+		return errors.New("eventhandler: unable to determine event type from state change event")
+	}
 }
 
 // startDrainEventsTicker starts a ticker that periodically drains the events queue
@@ -139,7 +160,7 @@ func (handler *TaskHandler) getBatchedContainerEvents() []api.TaskStateChange {
 	defer handler.lock.RUnlock()
 
 	var events []api.TaskStateChange
-	for taskARN, _ := range handler.tasksToContainerStates {
+	for taskARN := range handler.tasksToContainerStates {
 		if task, ok := handler.state.TaskByArn(taskARN); ok {
 			events = append(events, api.TaskStateChange{
 				TaskARN: taskARN,
@@ -151,84 +172,40 @@ func (handler *TaskHandler) getBatchedContainerEvents() []api.TaskStateChange {
 	return events
 }
 
-// AddStateChangeEvent queues up a state change for sending using the given client.
-func (handler *TaskHandler) AddStateChangeEvent(change statechange.Event, client api.ECSClient) error {
-	switch change.GetEventType() {
-	case statechange.TaskEvent:
-		event, ok := change.(api.TaskStateChange)
-		if !ok {
-			return errors.New("eventhandler: unable to get task event from state change event")
-		}
-		handler.flushBatch(&event)
-		handler.addEvent(newSendableTaskEvent(event), client)
-		return nil
-
-	case statechange.ContainerEvent:
-		event, ok := change.(api.ContainerStateChange)
-		if !ok {
-			return errors.New("eventhandler: unable to get container event from state change event")
-		}
-		handler.batchContainerEvent(event)
-		return nil
-
-	default:
-		return errors.New("eventhandler: unable to determine event type from state change event")
-	}
-}
-
-// batchContainerEvent collects container state change events for a given task arn
-func (handler *TaskHandler) batchContainerEvent(event api.ContainerStateChange) {
-	handler.lock.Lock()
-	defer handler.lock.Unlock()
-
+// batchContainerEventUnsafe collects container state change events for a given task arn
+func (handler *TaskHandler) batchContainerEventUnsafe(event api.ContainerStateChange) {
 	seelog.Infof("TaskHandler: batching container event: %s", event.String())
 	handler.tasksToContainerStates[event.TaskArn] = append(handler.tasksToContainerStates[event.TaskArn], event)
 }
 
-// flushBatch attaches the task arn's container events to TaskStateChange event that
-// is being submittied to the backend
-func (handler *TaskHandler) flushBatch(event *api.TaskStateChange) {
-	handler.lock.Lock()
-	defer handler.lock.Unlock()
-
+// flushBatchUnsafe attaches the task arn's container events to TaskStateChange event that
+// is being submittied to the backend and sends it to the backend
+func (handler *TaskHandler) flushBatchUnsafe(event *api.TaskStateChange, client api.ECSClient) {
 	event.Containers = append(event.Containers, handler.tasksToContainerStates[event.TaskARN]...)
 	delete(handler.tasksToContainerStates, event.TaskARN)
+
+	// Prepare a given event to be sent by adding it to the handler's appropriate
+	// eventList and remove the entry in tasksToEvents map
+	change := newSendableTaskEvent(*event)
+	taskEvents := handler.getTaskEventsUnsafe(change)
+
+	taskEvents.sendChange(change, client, handler)
 }
 
-// addEvent prepares a given event to be sent by adding it to the handler's appropriate
-// eventList and remove the entry in tasksToEvents map
-func (handler *TaskHandler) addEvent(change *sendableEvent, client api.ECSClient) {
-	taskEvents := handler.getTaskEventList(change)
-
-	taskEvents.lock.Lock()
-	defer taskEvents.lock.Unlock()
-
-	// Update taskEvent
-	seelog.Infof("TaskHandler: Adding event: %s", change.String())
-	taskEvents.events.PushBack(change)
-
-	if !taskEvents.sending {
-		taskEvents.sending = true
-		go handler.submitTaskEvents(taskEvents, client, change.taskArn())
-	}
-}
-
-// getTaskEventList gets the eventList from taskToEvent map
-func (handler *TaskHandler) getTaskEventList(change *sendableEvent) (taskEvents *eventList) {
-	handler.lock.Lock()
-	defer handler.lock.Unlock()
-
+// getTaskEventsUnsafe gets the eventList from taskToEvent map
+func (handler *TaskHandler) getTaskEventsUnsafe(change *sendableEvent) *taskSendableEvents {
 	taskARN := change.taskArn()
 	taskEvents, ok := handler.tasksToEvents[taskARN]
 	if !ok {
-		taskEvents = &eventList{events: list.New(),
+		taskEvents = &taskSendableEvents{
+			events:    list.New(),
 			sending:   false,
 			createdAt: time.Now(),
 			taskARN:   taskARN,
 		}
 		handler.tasksToEvents[taskARN] = taskEvents
 		seelog.Debugf("TaskHandler: collecting events for new task; event: %s; events: %s ",
-			change.String(), taskEvents.toStringUnsafe())
+			change.toString(), taskEvents.toStringUnsafe())
 	}
 
 	return taskEvents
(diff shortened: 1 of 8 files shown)


agent/eventhandler/task_handler.go
```diff
@@ -256,3 +279,77 @@ func (handler *TaskHandler) SubmitTaskEvents(taskEvents *eventList, client api.E
...
+	}
+}
+
+// setTaskChangeSent sets the event's task change object as sent
+func setTaskChangeSent(event *sendableEvent) {
+	if event.taskChange.Task != nil {
+		event.taskChange.Task.SetSentStatus(event.taskChange.Status)
+	}
+	for _, container := range event.taskChange.Containers {
+		container.Container.SetSentStatus(event.taskChange.Status.ContainerStatus(container.Container.GetSteadyStateStatus()))
```
richardpen: Why not just `container.Status` instead of `event.taskChange.Status.ContainerStatus(container.Container.GetSteadyStateStatus())`?

agent/eventhandler/task_handler_types.go
```diff
@@ -71,16 +72,27 @@ func (event *sendableEvent) taskArn() string {
...
 	if tevent.Status == api.TaskStatusNone {
 		return false // defensive programming :)
 	}
-	if event.taskSent || (tevent.Task != nil && tevent.Task.GetSentStatus() >= tevent.Status) {
+	if event.taskSent {
 		return false // redundant event
 	}
+	if tevent.Task != nil && tevent.Task.GetSentStatus() >= tevent.Status {
+		for _, containerStateChange := range tevent.Containers {
+			container := containerStateChange.Container
```
richardpen: Please add some comments here to makes it easier to read. And this part should have a unit test to cover.

agent/eventhandler/task_handler.go
```diff
@@ -188,57 +254,14 @@ func (handler *TaskHandler) SubmitTaskEvents(taskEvents *eventList, client api.E
 			event := eventToSubmit.Value.(*sendableEvent)
 
 			if event.containerShouldBeSent() {
```
richardpen: We probably don't need this part of code now, which can be in another pr.

agent/eventhandler/task_handler.go
```diff
@@ -58,16 +67,73 @@ type TaskHandler struct {
...
+
+// getBatchedContainerEvents gets a list task state changes for container events that
+// have been batched and not sent beyond the drainEventsFrequency threshold
+func (handler *TaskHandler) getBatchedContainerEvents() []api.TaskStateChange {
+	handler.taskHandlerLock.RLock()
+	defer handler.taskHandlerLock.RUnlock()
+
+	var events []api.TaskStateChange
+	for taskARN, _ := range handler.tasksToContainerStates {
+		if task, ok := handler.state.TaskByArn(taskARN); ok {
```
richardpen: I'm not sure if the `else` situation will happen, but since you are checking here, we may want to remove it from the `handler.tasksToContainerStates` map if it happened. And probably a log will help.

agent/eventhandler/task_handler.go
```diff
@@ -58,16 +67,73 @@ type TaskHandler struct {
...
+				handler.AddStateChangeEvent(event, handler.client)
+			}
+		}
+	}
+}
+
+// getBatchedContainerEvents gets a list task state changes for container events that
+// have been batched and not sent beyond the drainEventsFrequency threshold
+func (handler *TaskHandler) getBatchedContainerEvents() []api.TaskStateChange {
+	handler.taskHandlerLock.RLock()
```
richardpen: There may be a race condition where we are sending in the ticker and received the task state change also, can you move this lock inside the ticker line 110?

agent/eventhandler/task_handler.go
```diff
@@ -58,16 +67,73 @@ type TaskHandler struct {
 	// stateSaver is a statemanager which may be used to save any
 	// changes to a task or container's SentStatus
 	stateSaver statemanager.Saver
+
+	drainEventsFrequency time.Duration
```
samuelkarp: Looks like this needs gofmt.

aaithal: Why do you think so? It is formatted code.

agent/eventhandler/task_handler.go
aaithal: I missed that we have this in `ContainerStateChange`. I've modified this.

agent/eventhandler/task_handler.go
```diff
@@ -58,16 +67,73 @@ type TaskHandler struct {
...
+func (handler *TaskHandler) startDrainEventsTicker() {
+	ticker := time.NewTicker(handler.drainEventsFrequency)
+	for {
+		select {
+		case <-handler.ctx.Done():
+			seelog.Infof("TaskHandler: Stopping periodic container state change submission ticker")
+			ticker.Stop()
+			return
+		case <-ticker.C:
+			seelog.Info("TaskHandler tick")
```
jhaynes: Should this be `Debug` level? (and potentially line 105 too?) If we keep it at `Info`, I think we should make it more descriptive.

agent/eventhandler/task_handler.go
```diff
@@ -58,16 +67,73 @@ type TaskHandler struct {
...
+			for _, event := range handler.getBatchedContainerEvents() {
+				seelog.Infof(
```
jhaynes: Same question as above. Should this be `Debug`?

agent/eventhandler/task_handler.go
```diff
@@ -256,3 +279,77 @@ func (handler *TaskHandler) SubmitTaskEvents(taskEvents *eventList, client api.E
...
+	eventType string,
+	client api.ECSClient,
+	eventToSubmit *list.Element,
+	stateSaver statemanager.Saver,
+	backoff *utils.SimpleBackoff,
+	taskEvents *eventList) {
+
+	// Extract the wrapped event from the list element
+	event := eventToSubmit.Value.(*sendableEvent)
+	seelog.Infof("TaskHandler, Sending %s change: %s", eventType, event.String())
```
jhaynes: This changed from the prior if blocks from `Debug` to `Info`. I like this increase in verbosity better than some of the other `Info` level additions. However, I might flip the level on to `Debug` and log only the success message (line 309) as `Info`.

agent/eventhandler/task_handler.go
aaithal: Good catch. As per offline discussion, I don't think moving the lock itself resolves the race condition. However, moving https://github.com/aws/amazon-ecs-agent/blob/dev/agent/eventhandler/task_handler.go#L114 to https://github.com/aws/amazon-ecs-agent/blob/dev/agent/eventhandler/task_handler.go#L154 does resolve it I think. I'll modify that.

agent/eventhandler/task_handler.go
samuelkarp: I...have no idea what I was looking at.  Please ignore.

agent/eventhandler/task_handler.go
aaithal: That's a miss on my part. It should be removed.

agent/eventhandler/task_handler.go
aaithal: it makes sense for this to be `Info` as it's a trigger for submitting state changes to the backend and we'd want to know about it without having to elevate log level to debug

agent/eventhandler/task_handler.go
richardpen: Looks like this is still not fixed, did you miss this in the commit?

aaithal: Seems so. Working on it.

agent/eventhandler/task_handler.go
```diff
@@ -53,21 +59,96 @@ type TaskHandler struct {
...
+		ctx:                    ctx,
 		tasksToEvents:          make(map[string]*eventList),
 		submitSemaphore:        utils.NewSemaphore(concurrentEventCalls),
 		tasksToContainerStates: make(map[string][]api.ContainerStateChange),
 		stateSaver:             stateManager,
+		state:                  state,
+		client:                 client,
+		drainEventsFrequency:   drainEventsFrequency,
 	}
+	go taskHandler.startDrainEventsTicker()
```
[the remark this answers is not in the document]
aaithal: I think we have this pattern elsewhere as well. An object creation method doing all the things for completing the object creation doesn't seem like an unwanted thing to me. I don't have strong opinions here. I can change it if you really want me to.

agent/eventhandler/task_handler.go
```diff
@@ -117,142 +198,218 @@ func (handler *TaskHandler) flushBatch(event *api.TaskStateChange) {
 // addEvent prepares a given event to be sent by adding it to the handler's appropriate
 // eventList and remove the entry in tasksToEvents map
 func (handler *TaskHandler) addEvent(change *sendableEvent, client api.ECSClient) {
-	handler.taskHandlerLock.Lock()
```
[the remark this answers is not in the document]
aaithal: Yeah, that's a valid race. Please look at the latest revision where this has been fixed as well.

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
 	}
 }
 
-// batchContainerEvent collects container state change events for a given task arn
-func (handler *TaskHandler) batchContainerEvent(event api.ContainerStateChange) {
-	handler.taskHandlerLock.Lock()
-	defer handler.taskHandlerLock.Unlock()
+// startDrainEventsTicker starts a ticker that periodically drains the events queue
+// by submitting state change events to the ECS backend
+func (handler *TaskHandler) startDrainEventsTicker() {
```
richardpen: nit: I'd love to move this function before `AddStateChangeEvent`.

agent/eventhandler/task_handler.go
```diff
@@ -53,21 +59,96 @@ type TaskHandler struct {
...
+}
+
+// getBatchedContainerEvents gets a list task state changes for container events that
+// have been batched and not sent beyond the drainEventsFrequency threshold
+func (handler *TaskHandler) getBatchedContainerEvents() []api.TaskStateChange {
+	handler.lock.RLock()
+	defer handler.lock.RUnlock()
+
+	var events []api.TaskStateChange
+	for taskARN, _ := range handler.tasksToContainerStates {
```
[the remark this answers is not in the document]
richardpen: Do you want to modify this?

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
-				}
-			} else {
-				// Shouldn't be sent as either a task or container change event; must have been already sent
-				seelog.Infof("TaskHandler, Not submitting redundant event; just removing: %s", event.String())
-				taskEvents.events.Remove(eventToSubmit)
-			}
+// sendChange adds the change to the sendable events list. If the
+// event for the task is in the process of being sent, it doesn't trigger
+// the async submit state api method
+func (taskEvents *taskSendableEvents) sendChange(change *sendableEvent,
```
richardpen: Same as above, can you move this above `submitTaskEvents`.

agent/eventhandler/task_handler.go
aaithal: I wanted the exported functions to be at the top of the file as that's a reasonable way to organize the code.

agent/eventhandler/task_handler.go
aaithal: The order is `TaskHandler` methods and then `taskEvents` methods (exported first, non exported afterwards), which is again an organization structure that we've followed everywhere in the code base

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
+// taskSendableEvents is used to group all events for a task
+type taskSendableEvents struct {
+	// events is a list of *sendableEvents
+	events *list.List
+	// sending will check whether the list is already being handlerd
+	sending bool
+	//eventsListLock locks both the list and sending bool
+	lock sync.Mutex
+	// createdAt is a timestamp for when the event list was created
+	createdAt time.Time
```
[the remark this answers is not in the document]
aaithal: both `createdAt` and `taskARN` fields are to help with debugging/logging.

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
+	}
+	return events
 }
 
-// flushBatch attaches the task arn's container events to TaskStateChange event that
-// is being submittied to the backend
-func (handler *TaskHandler) flushBatch(event *api.TaskStateChange) {
-	handler.taskHandlerLock.Lock()
-	defer handler.taskHandlerLock.Unlock()
+// batchContainerEventUnsafe collects container state change events for a given task arn
```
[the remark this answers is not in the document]
aaithal: we're moving it up to `AddStateChangeEvent`. More details in the commit message: https://github.com/aws/amazon-ecs-agent/pull/1026/commits/d959f31b6f1a0870a4ce1f80a755a13b0d40e72e

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
-	// events is a list of *sendableEvents
-	events *list.List
-	// sending will check whether the list is already being handlerd
-	sending bool
-	//eventsListLock locks both the list and sending bool
-	eventListLock sync.Mutex
-}
+	// drainEventsFrequency is the frequency at the which unsent events batched
+	// by the task handler are sent to the backend
+	drainEventsFrequency = time.Minute
```
samuelkarp: Should we go faster than a minute?  A minute seems pretty slow to me.

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
-				}
-			} else {
-				// Shouldn't be sent as either a task or container change event; must have been already sent
-				seelog.Infof("TaskHandler, Not submitting redundant event; just removing: %s", event.String())
-				taskEvents.events.Remove(eventToSubmit)
-			}
+// sendChange adds the change to the sendable events list. If the
+// event for the task is in the process of being sent, it doesn't trigger
+// the async submit state api method
+func (taskEvents *taskSendableEvents) sendChange(change *sendableEvent,
```
samuelkarp: Can you add a bit to this comment here explaining that the actual task submission (API request) occurs asynchronously, so that this function doesn't end up causing the lock to be held for a long time?

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
+		ctx:                    ctx,
+		tasksToEvents:          make(map[string]*taskSendableEvents),
 		submitSemaphore:        utils.NewSemaphore(concurrentEventCalls),
 		tasksToContainerStates: make(map[string][]api.ContainerStateChange),
 		stateSaver:             stateManager,
+		state:                  state,
+		client:                 client,
+		drainEventsFrequency:   drainEventsFrequency,
 	}
+	go taskHandler.startDrainEventsTicker()
```
samuelkarp: I'd like a comment somewhere explaining the flow for batched events.  This is roughly what I got through reading this:

```
startDrainEventsTicker->
	getBatchedContainerEvents->
		read from tasksToContainerStates and returns TaskEvent
	AddStateChangeEvent->
		flushBatchUnsafe
```

The part that took me a while is that `AddStateChangeEvent` also calls `batchContainerUnsafe` for `ContainerEvent`s and the key difference is that `getBatchedContainerEvents` returns a `TaskEvent` even though it has "ContainerEvents" in the name.

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
+	defer handler.removeTaskEvents(taskARN)
+
+	backoff := utils.NewSimpleBackoff(submitStateBackoffMin, submitStateBackoffMax,
+		submitStateBackoffJitterMultiple, submitStateBackoffMultiple)
 
 	// Mirror events.sending, but without the need to lock since this is local
 	// to our goroutine
 	done := false
-
+	// TODO: wire in the context here. Else, we have go routine leaks in tests
```
samuelkarp: Is this TODO expected to be fixed before merging?  goroutine leaks in tests can cause flaky tests.

agent/eventhandler/task_handler.go
```diff
@@ -15,244 +15,323 @@ package eventhandler
...
-		})
+	// Update taskEvent
+	seelog.Infof("TaskHandler: Adding event: %s", change.toString())
+	taskEvents.events.PushBack(change)
+
+	if !taskEvents.sending {
+		taskEvents.sending = true
+		go handler.submitTaskEvents(taskEvents, client, change.taskArn())
+	} else {
+		seelog.Debugf("TaskHandler: Not submitting change as the task is already being sent: %s", change.toString())
```
samuelkarp: So I think there's a race here, but the impact of the race is probably only a delay of `drainEventsFrequency`.  If we evaluate `taskEvents.sending` as true we won't cause a new goroutine to be created for `handler.submitTaskEvents`.  However, if the existing goroutine is just about done sending, it's possible it won't pick up the new event that we just added on line 263.  I think this is a reasonable tradeoff here, but it would be worth recording it somewhere that the maximum length of time we can have an unsubmitted change is really effectively `2*drainEventsFrequency`.

agent/eventhandler/task_handler.go
aaithal: As per offline conversation, i'll change this to `20s`

agent/eventhandler/task_handler.go
aaithal: Not with this PR :(

agent/eventhandler/task_handler.go
aaithal: I'll rename `getBatchedContainerEvents` -> `taskStateChangesToSend`. Will add some comments in `startDrainEventsTicker` as well.

aaithal:
@samuelkarp  can you take a look at the latest commit? I think I've addressed your comments. Thanks!

Issue #1024, opened 2017-10-20
ECS starting agent fails with: Could not unmarshal state


tsonev:
### Summary
After enabling docker debugging as per this article:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/docker-debug-mode.html

and running :

sudo service docker restart
sudo start ecs

the ECS agent failed to start with :

 [CRITICAL] Error loading previously saved state: Could not unmarshal state; incomplete save. There was no task for arn arn:aws:ecs:us-west-2:888922156537:task/bd886aa4-d3d3-4cbd-8588-6f02b793d4a9

### Expected Behavior
The agent and services container should start.

### Observed Behavior
Agent container fails to start

### Environment Details
$ docker info
Containers: 4
 Running: 0
 Paused: 0
 Stopped: 4
Images: 4
Server Version: 17.03.2-ce
Storage Driver: devicemapper
 Pool Name: docker-docker--pool
 Pool Blocksize: 524.3 kB
 Base Device Size: 10.74 GB
 Backing Filesystem: ext4
 Data file:
 Metadata file:
 Data Space Used: 2.659 GB
 Data Space Total: 23.33 GB
 Data Space Available: 20.67 GB
 Metadata Space Used: 1.241 MB
 Metadata Space Total: 25.17 MB
 Metadata Space Available: 23.92 MB
 Thin Pool Minimum Free Space: 2.333 GB
 Udev Sync Supported: true
 Deferred Removal Enabled: true
 Deferred Deletion Enabled: true
 Deferred Deleted Device Count: 0
 Library Version: 1.02.135-RHEL7 (2016-11-16)
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 4ab9917febca54791c5f071a9d1f404867857fcc
runc version: 54296cf40ad8143b62dbcaa1d90e520a2136ddfe
init version: 949e6fa
Security Options:
 seccomp
  Profile: default
Kernel Version: 4.9.43-17.39.amzn1.x86_64
Operating System: Amazon Linux AMI 2017.03
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 3.86 GiB
Name: ip-172-31-10-150
[...]

PR #1033, opened 2017-10-25
dockerstate: map keys differ before create

samuelkarp:
The state.idToTask and state.idToContainer maps initially have the keys
set to the generated DockerName rather than the DockerID.  Once
containers are created, the keys are changed to the DockerID.  This
changed as part of this commit:
https://github.com/aws/amazon-ecs-agent/commit/49c36c7855d14155469a7796b35318da34743cf8#diff-464fc7e15d1a6b818ceca89e7b68cd4e.

### Summary
Fixes https://github.com/aws/amazon-ecs-agent/issues/1024

If a task is run where the image fails to pull, the agent will get into an unrecoverable state after it cleans up the task.  This happens because it fails to delete an entry in the `state.idToTask` and `state.idToContainer` maps as it attempts to delete with the wrong key.

### Implementation details
The state.idToTask and state.idToContainer maps initially have the keys set to the generated DockerName rather than the DockerID.  Once containers are created, the keys are changed to the DockerID.  This changed as part of this commit: https://github.com/aws/amazon-ecs-agent/commit/49c36c7855d14155469a7796b35318da34743cf8#diff-464fc7e15d1a6b818ceca89e7b68cd4e.

### Testing

Bug - Fixed a bug where tasks that fail to pull containers can cause the agent to fail to restore properly after a restart. [#1024](https://github.com/aws/amazon-ecs-agent/issues/1024)

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
   [#1014](https://github.com/aws/amazon-ecs-agent/pull/1014)
 * Enhancement - Support `init` process in containers by adding support for Docker remote API client version 1.25
   [#996](https://github.com/aws/amazon-ecs-agent/pull/996)
+* Bug - Fixed a bug where tasks that fail to pull containers can cause the
+  agent to fail to restore properly after a restart.
+  [#1033](https://github.com/aws/amazon-ecs-agent/pull/1033)
 
 ## 1.14.5
 * Enhancement - Retry failed container image pull operations [#975](https://github.com/aws/amazon-ecs-agent/pull/975)
--- a/agent/engine/dockerstate/docker_task_engine_state.go
+++ b/agent/engine/dockerstate/docker_task_engine_state.go
@@ -318,19 +318,7 @@ func (state *DockerTaskEngineState) AddContainer(container *api.DockerContainer,
 		state.tasks[task.Arn] = task
 	}
 
-	if container.DockerID != "" {
-		// Update the container id to the state
-		state.idToContainer[container.DockerID] = container
-		state.idToTask[container.DockerID] = task.Arn
-
-		// Remove the previously added name mapping
-		delete(state.idToContainer, container.DockerName)
-		delete(state.idToTask, container.DockerName)
-	} else if container.DockerName != "" {
-		// Update the container name mapping to the state when the ID isn't available
-		state.idToContainer[container.DockerName] = container
-		state.idToTask[container.DockerName] = task.Arn
-	}
+	state.storeIDToContainerTaskUnsafe(container, task)
 
 	existingMap, exists := state.taskToID[task.Arn]
 	if !exists {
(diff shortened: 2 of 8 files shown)


samuelkarp:
> LGTM but 'RemoveTask" should have a unit test that validates this change.

Yep, that's why this is marked WIP right now.  Tests are forthcoming (probably tomorrow).

samuelkarp:
I've modified some unit tests and also added a new functional test to ensure we don't regress behavior here.  This test fails with the v1.14.5 agent.

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -89,6 +89,43 @@ func TestSavedState(t *testing.T) {
...
+	require.NoError(t, err, "failed to start task")
+	assert.NoError(t, testTask.ExpectErrorType("error", "CannotPullContainerError", 1*time.Minute))
+
+	resp, err := agent.CallTaskIntrospectionAPI(testTask)
+	assert.NoError(t, err, "should be able to introspect the task")
+	assert.NotNil(t, resp, "should receive a response")
+	assert.Equal(t, *testTask.TaskArn, resp.Arn, "arn should be equal")
+
+	// wait two minutes for it to be cleaned up
+	fmt.Println("Sleeping...")
```
aaithal: minor: remove this? Or use a logger?

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -364,18 +364,27 @@ func (state *DockerTaskEngineState) RemoveTask(task *api.Task) {
...
 		return
 	}
 	delete(state.taskToID, task.Arn)
 
 	for _, dockerContainer := range containerMap {
-		delete(state.idToTask, dockerContainer.DockerID)
-		delete(state.idToContainer, dockerContainer.DockerID)
+		// The key to these maps is either the Docker ID or agent-generated name.  We use the agent-generated name
+		// before a Docker ID is available.
+		key := dockerContainer.DockerID
```
aaithal: it might be safer to move this out of this method into a `getKey(container)` method (so that future modifications in this method do not touch that critical piece of business logic). what do you think?

agent/functional_tests/util/utils_unix.go
```diff
@@ -166,8 +166,12 @@ func (agent *TestAgent) StartAgent() error {
 		Links: agent.Options.ContainerLinks,
 	}
 
+	if os.Getenv("ECS_FTEST_FORCE_NET_HOST") != "" {
```
richardpen: Where is this environment variable being set?

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
   [#1014](https://github.com/aws/amazon-ecs-agent/pull/1014)
 * Enhancement - Support `init` process in containers by adding support for Docker remote API client version 1.25
   [#996](https://github.com/aws/amazon-ecs-agent/pull/996)
+* Bug - Fixed a bug where tasks that fail to pull containers can cause the
+  agent to fail to restore properly after a restart.
+  [#1024](https://github.com/aws/amazon-ecs-agent/issues/1024)
```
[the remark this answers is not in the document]
samuelkarp: I wrote this before the PR was created.  I can update it.

agent/functional_tests/util/utils_unix.go
samuelkarp: It's not automatically set.  I added this because I was developing the test in an environment where the docker0 bridge did not have Internet access (for unrelated reasons) and wanted the agent to actually be able to run.

agent/engine/dockerstate/docker_task_engine_state.go
samuelkarp: The code that adds to the map also removes references during a subsequent call.  I'll extract both operations out to explicit functions with comments explaining why they're doing this.

agent/functional_tests/tests/functionaltests_test.go
samuelkarp: `fmt.Println` will print out immediately (when running tests for a single package) while `t.Log` buffers until the test ends.  I had wanted to print this out during the run so I could inspect the agent logs while it was sleeping.  I'm inclined to leave this in right now, but I can remove it if you feel strongly.

samuelkarp:
@aaithal @richardpen Can you take a look?

agent/functional_tests/util/utils_unix.go
richardpen: Then can you add this into the `agent/functional_tests/README.d`.

agent/functional_tests/tests/functionaltests_test.go
aaithal: Nah, you can leave this be. It's functional tests and it doesn't matter if log out to stdout here immediately.

agent/functional_tests/util/utils_unix.go
samuelkarp: Done.  I rewrote a bit more of the README to make it more clear.

PR #1036, opened 2017-10-26
Fix the issue where task does not wait for resource on restart

richardpen:
### Summary

Initially agent should wait for instance resource(port, mem) to be
released by waiting for previous task to be stopped, but this will
broken when task is waiting and agent was restarted. This fix will add
task into waitgroup on restart.
### Implementation details

On agent restart, check if the task has the stop sequence number(which is set when agent receives payload message to stop a task), and add the stop sequence number into the waitStopGroup.

### Testing

- [x] Manually tests with the scenario
New tests cover the changes: 
yes
### Licensing

yes

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -245,52 +245,72 @@ func (engine *DockerTaskEngine) synchronizeState() {
 	}
 
 	tasks := engine.state.AllTasks()
+	var tasksToStart []*api.Task
 	for _, task := range tasks {
 		conts, ok := engine.state.ContainerMapByArn(task.Arn)
 		if !ok {
-			engine.startTask(task)
+			// task hasn't started processing, no need to check container status
+			tasksToStart = append(tasksToStart, task)
 			continue
 		}
-		for _, cont := range conts {
-			if cont.DockerID == "" {
-				log.Debug("Found container potentially created while we were down", "name", cont.DockerName)
-				// Figure out the dockerid
-				describedCont, err := engine.client.InspectContainer(cont.DockerName, inspectContainerTimeout)
-				if err != nil {
-					log.Warn("Could not find matching container for expected", "name", cont.DockerName)
-				} else {
-					cont.DockerID = describedCont.ID
-					cont.Container.SetKnownStatus(dockerStateToState(describedCont.State))
-					// update mappings that need dockerid
-					engine.state.AddContainer(cont, task)
-					engine.imageManager.RecordContainerReference(cont.Container)
-				}
-			}
-			if cont.DockerID != "" {
-				currentState, metadata := engine.client.DescribeContainer(cont.DockerID)
-				if metadata.Error != nil {
-					currentState = api.ContainerStopped
-					if !cont.Container.KnownTerminal() {
-						cont.Container.ApplyingError = api.NewNamedError(&ContainerVanishedError{})
-						log.Warn("Could not describe previously known container; assuming dead", "err", metadata.Error, "id", cont.DockerID, "name", cont.DockerName)
-						engine.imageManager.RemoveContainerReferenceFromImageState(cont.Container)
-					}
-				} else {
-					engine.imageManager.RecordContainerReference(cont.Container)
-					if engine.cfg.ContainerMetadataEnabled && !cont.Container.IsMetadataFileUpdated() {
-						go engine.updateMetadataFile(task, cont)
-					}
 
-				}
-				if currentState > cont.Container.GetKnownStatus() {
-					cont.Container.SetKnownStatus(currentState)
-				}
-			}
+		for _, cont := range conts {
+			engine.synchronizeContainerStatus(cont, task)
 		}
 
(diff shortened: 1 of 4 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -287,8 +288,19 @@ func (engine *DockerTaskEngine) synchronizeState() {
 				}
 			}
 		}
+
+		tasksToStart = append(tasksToStart, task)
+
+		// Put tasks that stopped by acs but not stopped in wait group
```
aaithal: nit: `"that are stopped"`

agent/engine/docker_task_engine.go
```diff
@@ -245,6 +245,7 @@ func (engine *DockerTaskEngine) synchronizeState() {
 	}
 
 	tasks := engine.state.AllTasks()
+	var tasksToStart []*api.Task
```
aaithal: shouldn't you also append to this list here: https://github.com/aws/amazon-ecs-agent/pull/1036/files#diff-b9aa12c3425c34d6f330f28407f779f6R252

agent/engine/docker_task_engine.go
```diff
@@ -245,6 +245,7 @@ func (engine *DockerTaskEngine) synchronizeState() {
...
 	for _, task := range tasks {
```
aaithal: please refactor contents of this `for` loop into a method while you're here. Same goes for https://github.com/aws/amazon-ecs-agent/pull/1036/files#diff-b9aa12c3425c34d6f330f28407f779f6R255

agent/engine/docker_task_engine.go
```diff
@@ -245,53 +245,73 @@ func (engine *DockerTaskEngine) synchronizeState() {
...
+		} else {
+			container.DockerID = describedContainer.ID
+			container.Container.SetKnownStatus(dockerStateToState(describedContainer.State))
+			// update mappings that need dockerid
+			engine.state.AddContainer(container, task)
+			engine.imageManager.RecordContainerReference(container.Container)
+		}
+	}
+
+	if container.DockerID != "" {
```
aaithal: please refactor this as 

```
if container.DockerID == "" { 
 return 
}

currentState, ...
```

agent/engine/docker_task_engine.go
```diff
@@ -245,53 +245,73 @@ func (engine *DockerTaskEngine) synchronizeState() {
...
+		currentState, metadata := engine.client.DescribeContainer(container.DockerID)
+		if metadata.Error != nil {
+			currentState = api.ContainerStopped
```
aaithal: This is a bug in the Agent today, where if we can't inspect a container we mark it as stopped. Please add a `TODO` here to come back and fix this (better yet, create an issue). @samuelkarp would have more details about the bug.

agent/engine/docker_task_engine.go
```diff
@@ -245,53 +245,73 @@ func (engine *DockerTaskEngine) synchronizeState() {
...
 }
 
+// synchronizeContainerStatus checks and updates the container status with docker
+func (engine *DockerTaskEngine) synchronizeContainerStatus(container *api.DockerContainer, task *api.Task) {
+	if container.DockerID == "" {
+		log.Debug("Found container potentially created while we were down", "name", container.DockerName)
+		// Figure out the dockerid
+		describedContainer, err := engine.client.InspectContainer(container.DockerName, inspectContainerTimeout)
+		if err != nil {
+			log.Warn("Could not find matching container for expected", "name", container.DockerName)
```
aaithal: What happens to the task at this point of time?

agent/engine/docker_task_engine.go
samuelkarp: I don't think we want to assume that it's dead.  However, I do think it would make more sense to mark it as desired=STOPPED and then let it progress through the task engine.  I'm looking at a bug right now where we do something very similar and assume that a container is stopped when it isn't, which can cause problems (resource conflicts, unintended processes still running, etc).

agent/engine/docker_task_engine.go
richardpen: Nothing will happen at this point, task won't update the status of this container for this cycle of synchronizing. It looks like the behavior is a little different for containers with dockerID and containers without dockerID. I have created an issue for this, based on your comments below.

exp0nge:
Any idea when this will be merged?

samuelkarp:
@richardpen Please let me know when you've addressed @aaithal's comments.

agent/engine/docker_task_engine_test.go
```diff
@@ -1967,3 +1967,79 @@ func TestMetadataFileUpdatedAgentRestart(t *testing.T) {
 
 	wg.Wait()
 }
+
+// TestTakWaitFroHostResourceOnRestart tests task stopped by acs but hasn't
+// reached stopped should block the later task to start
+func TestTakWaitFroHostResourceOnRestart(t *testing.T) {
```
samuelkarp: nit: s/Tak/Task/ and s/Fro/For/

PR #1037, opened 2017-10-26
DNS overrides for 'awsvpc' tasks

aaithal:
### Summary

Apply DNS overrides in container's hostconfig

### Implementation details

If the container belongs to a task with an ENI, and if the ENI object has custom DNS overrides specified, override it in the container's host config as well.

### Testing

### Licensing


agent/api/task.go
```diff
@@ -590,6 +596,27 @@ func (task *Task) shouldOverrideNetworkMode(container *Container, dockerContaine
...
+// 2. ENI has custom DNS IPs and search list associated with it
+// This should only be done for the pause container as other containers inherit
+// /etc/resolv.conf of this container (they share the network namespace)
+func (task *Task) overrideDNS(hostConfig *docker.HostConfig) *docker.HostConfig {
+	eni := task.GetTaskENI()
+	if eni == nil {
+		return hostConfig
+	}
+	if len(eni.DomainNameServers) > 0 {
+		hostConfig.DNS = eni.DomainNameServers
```
richardpen: Not sure if we need add logs for these two check, if not, then do we still need the check?

agent/statemanager/state_manager.go
```diff
@@ -50,7 +50,8 @@ const (
 	//   d) Add 'ENIAttachments' struct
 	//   e) Deprecate 'SteadyStateDependencies' in favor of 'TransitionDependencySet'
 	// 7)
-	//	 a) Add 'MetadataUpdated' field to 'api.Container'
+	//   a) Add 'MetadataUpdated' field to 'api.Container'
+	//   b) Add 'DomainNameServers' and 'DomainNameSearchList' in `api.ENI`
```
richardpen: Are we sure that this is going together with metadata?

agent/api/task.go
aaithal: I don't think we need to log anything here. Why wouldn't we need these checks though?

agent/statemanager/state_manager.go
aaithal: We'll see :) If not, its a merge conflict that'll need resolution

agent/api/task.go
richardpen: Won't this has the same effect, we can even assign the empty value when the `len(xxx)==0`, any side effect?
```
hostConfig.DNS = eni.DomainNameServers
hostConfig.DNSSearch = eni.DomainNameSearchList
```

aaithal: Oh I didn't consider that. Yeah, that's effectively doing the same thing. I'll update this.

aaithal:
@vsiddharth oops! should be fixed now. Thanks for catching it.

PR #1042, opened 2017-10-30
Fix timing issue with hostsetup script

kalpik:
### Summary
I noticed that sometimes the port forwarding that we do using netsh doesn't work. The ECS agent does listen on port 51679, but we cannot connect to it using port 80, and hence IAM task roles don't work. I figured out that this happens when the netsh command is executed before the Credential IP is up. Maybe this is because we do the IP allocation in powershell, and then use netsh (cmd) to do the power forwarding. This change makes sure that the IP is up and responding before we try to set up port forwarding. And with this, I haven't noticed the issue anymore.

### Implementation details
Just wait for ping to succeed to the Credential IP before setting up port forwarding

### Testing
N/A

Fix timing issue with setting up port forwarding

### Licensing

### DIFF
--- a/misc/windows-deploy/amazon-ecs-agent.ps1
+++ b/misc/windows-deploy/amazon-ecs-agent.ps1
@@ -61,7 +61,7 @@ try {
 
     LogMsg "Docker is running and ready."
 
-    if([System.Environment]::GetEnvironmentVariable("ECS_ENABLE_TASK_IAM_ROLE", "Machine") -eq "true") {
+    if([System.Environment]::GetEnvironmentVariable("ECS_ENABLE_TASK_IAM_ROLE") -eq "true") {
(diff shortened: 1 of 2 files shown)


adnxn:
@kalpik Thanks for the PR. Just wanted to let you know we've been reviewing this internally and trying to figure out where it fits in with our recent agent changes for windows.

kalpik:
@adnxn Thanks! Yes, I saw that PR, but I wanted it to work reliably for me in the mean time, so put it out here so that it might benefit others too :)

adnxn:
@kalpik thanks again for bringing our attention to this. I'm closing this PR in favor of work being proposed in #1198. We're moving the `*.ps1` scripts out of the agent's repo altogether.

PR #1048, opened 2017-11-02
Remove task/container overridden in DockerHostConfig

### LINKED ISSUE #1047: Task can be blocked due to a race condition in task/container overridden
### Summary

Task is blocked in creating state and will stay there forever.

### Description

When the task is trying to create container, it will call [task.Overrridden](https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task.go#L390) and [container.Overridden](https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/container.go#L172). In those two methods, it's copying the struct that has some lock, which is anti-pattern as from the [document](https://golang.org/pkg/sync/#RWMutex), this will cause a deadlock if it's copying while some one acquired the lock.

### Expected Behavior
Task should move to running without blocking.

### Observed Behavior
Task is blocked in creating forever.

richardpen:
In task/container overridden, it's copying the task/container struct
which has a lock, this could cause deadlock if some other go routine
acquired the lock right before copying and released the lock after the
copy. This commit moved the overridden part to ACS when unmarshal the
task from payload message.

### Summary

Fix #1047 

### Implementation details

Removed the Overridden function, and added the same in TaskFromACS.

### Testing

New tests cover the changes: 
yes
### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
-# Changelog
+#Changelog
+
+## Unreleased
+* Bug - Fixed a bug where a task can be blocked in creating state. [#1048](https://github.com/aws/amazon-ecs-agent/pull/1048)
 
 ## 1.15.0
 * Feature - Support for provisioning tasks with ENIs.
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -167,19 +167,6 @@ func NewContainerWithSteadyState(steadyState ContainerStatus) *Container {
 	}
 }
 
-// Overriden applies the overridden command and returns the resulting
-// container object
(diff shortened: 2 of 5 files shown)


samuelkarp:
Can you add a test case that covers reading a task with overrides from the state file and unmarshalling it correctly so that the overrides are properly applied?

richardpen:
@samuelkarp Actually the `overrides` field of the container won't be set in the agent today. We don't even need the code [here](https://github.com/aws/amazon-ecs-agent/pull/1048/files#diff-e008975819a900a8494cf91881a8304aR711) in this PR, we shall remove this part of code when the model is updated.

PR #906, opened 2017-07-27
Handle the new field ecrpullcredentials in the task definition

richardpen:
### Summary

Handle the new field "ecrpullcredentials" in task definition.

### Implementation details

Check for the pullcredentials stored in container struct when pulling the image, if it exists, then use the credentials provided to get the docker auth from ecr.

### Testing

- [x] Builds on Linux (`make release`)
- [x] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [x] Unit tests on Linux (`make test`) pass
- [x] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [x] Integration tests on Linux (`make run-integ-tests`) pass
- [x] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [x] Functional tests on Linux (`make run-functional-tests`) pass
- [x] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: 
yes
### Description for the changelog

Add support for ECRPullCredentials for container.
### Licensing

This contribution is under the terms of the Apache 2.0 License: 
yes

PR #1062, opened 2017-11-09
add support for awslogs authentication over execution role

adnxn:
### Summary

This PR includes changes to the following packages,
acs/handler: specify the credentials role type
api: add changes required for awslogs authentication
app: add execution-role-awslogs capability
credentials: add fields for RoleType
engine: apply execution role auth for awslogs in hostconfig
handlers/credentials: extract role type for audit logs from credentials
logger/audit: add new event types for credential role types

If the task has an execution role + awslogs, the payload message will include `logsAuthStrategy` specifying `ExecutionRole` in the `Container` struct. The agent checks for this `ExecutionRole` value and then passes the generated credentials endpoint to the container during container create using `hostConfig.LogConfig`.

related aws-ecs-init change: https://github.com/aws/amazon-ecs-init/pull/128
### Testing

New tests cover the changes: 

### Licensing

yes


agent/api/task.go
```diff
@@ -500,6 +504,26 @@ func (task *Task) DockerHostConfig(container *Container, dockerContainerMap map[
...
+	id := task.GetExecutionCredentialsID()
+	if id == "" {
+		// No execution credentials set for the task. Do not inject the endpoint environment variable.
+		return &HostConfigError{"No execution credentials set for the task"}
+	}
+
+	executionRoleCredentials, ok := credentialsManager.GetTaskCredentials(id)
+	if !ok {
+		// Task has credentials id set, but credentials manager is unaware of
+		// the id. This should never happen as the payload handler sets
```
samuelkarp: "should never happen" in a comment is asking for it to happen. 😄 

It looks like this gets called during CreateContainer.  We need to verify (and probably write a test to enforce) that credentials are present before CreateContainer gets invoked.  @richardpen did work here in https://github.com/aws/amazon-ecs-agent/pull/906 for pull, but we need to make sure that the same dependency condition is applied for create as well.

agent/app/agent_capability.go
```diff
@@ -136,6 +137,9 @@ func (agent *ecsAgent) capabilities() []*ecs.Attribute {
 			})
 		}
 	}
+	if agent.cfg.OverrideAWSLogsExecutionRole {
```
samuelkarp: Can you make a TODO and an issue to come back to this later?  We'll want to have non-overridden enabled with an appropriate Docker version once that ships.

agent/api/task.go
aaithal: I agree on the sentiment here. However, I'd also advise against changing this logic significantly at this point of time. 

We follow a similar logic in `api.Task`'s `PostUnmarshalTask` method. I think returning a `HostConfigError` seems like an acceptable way to handle that edge case.

agent/config/types.go
```diff
@@ -188,6 +188,8 @@ type Config struct {
 	// ContainerMetadataEnabled specifies if the agent should provide a metadata
 	// file for containers.
 	ContainerMetadataEnabled bool
+
+	OverrideAWSLogsExecutionRole bool
```
aaithal: Please add documentation for this exported field. There are a bunch of these that you've added where there's no accompanying comment. Please fix all such occurrences.

agent/logger/audit/entry_types.go
```diff
@@ -34,11 +36,11 @@ const (
 	// 4. url
 	// 5. user agent
 	// 6. arn for the entity associated with credentials
-	// 7. event type ('GetCredentials')
+	// 7. event type ('GetCredentials, GetCredentialsExecutionRole')
```
aaithal: In line number 32, it still says version "1". Can you add changes for version "2" under a new comments section? After line 42 perhaps? Similar to how we document state file revision changes?

agent/acs/handler/refresh_credentials_handler.go
```diff
@@ -127,16 +127,23 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
...
+		seelog.Errorf("Unknown RoleType for task in credentials message, roleType: %s arn: %s, messageId: %s", roleType, taskArn, messageId)
+	} else {
+
+		taskCredentials := credentials.TaskIAMRoleCredentials{
+			ARN:                taskArn,
+			IAMRoleCredentials: credentials.IAMRoleCredentialsFromACS(message.RoleCredentials, roleType),
+		}
+		err = refreshHandler.credentialsManager.SetTaskCredentials(taskCredentials)
+		if err != nil {
+			seelog.Errorf("Error updating credentials, err: %v messageId: %s", err, messageId)
```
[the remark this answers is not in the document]
adnxn: gotcha. fixed.

agent/api/task.go
```diff
@@ -500,6 +504,26 @@ func (task *Task) DockerHostConfig(container *Container, dockerContainerMap map[
 	return task.dockerHostConfig(container, dockerContainerMap)
 }
 
+func (task *Task) ApplyExecutionRoleLogsAuth(hostConfig *docker.HostConfig, credentialsManager credentials.Manager) *HostConfigError {
```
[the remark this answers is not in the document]
adnxn: fixed.

agent/config/types.go
adnxn: ah missed these. fixed.

agent/logger/audit/entry_types.go
adnxn: added.

agent/app/agent_capability.go
adnxn: sure sounds good.

agent/api/task.go
adnxn: leaving this logic as is for now.

agent/acs/handler/refresh_credentials_handler.go
```diff
@@ -125,18 +125,25 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
...
+		taskCredentials := credentials.TaskIAMRoleCredentials{
+			ARN:                taskArn,
+			IAMRoleCredentials: credentials.IAMRoleCredentialsFromACS(message.RoleCredentials, roleType),
+		}
+		err = refreshHandler.credentialsManager.SetTaskCredentials(taskCredentials)
+		if err != nil {
+			seelog.Errorf("Unable to update credentials for task, err: %v messageId: %s", err, messageId)
+			return fmt.Errorf("unable to update credentials %v", err)
+		}
+		task.SetCredentialsID(aws.StringValue(message.RoleCredentials.CredentialsId))
```
richardpen: `SetCredentialsID` is used to set the id for task iam role, where we have two type of roles now. You may need to set this based on the type of the role.

agent/credentials/manager.go
```diff
@@ -38,6 +38,10 @@ const (
 
 	v1CredentialsEndpointRelativeURIFormat = "%s?" + CredentialsIDQueryParameterName + "=%s"
 	v2CredentialsEndpointRelativeURIFormat = "%s/%s"
+
+	// The role types specified in the credentials payload msg from the backend
+	ApplicationRoleType = "TaskApplication"
```
aaithal: should be `"// ApplicationRoleType specifies ..."`

agent/credentials/manager.go
```diff
@@ -38,6 +38,10 @@ const (
...
+	ExecutionRoleType   = "TaskExecution"
```
aaithal: This needs a comment as well:
`"// ExecutionRoleType is ..."`

agent/credentials/manager.go
```diff
@@ -51,6 +55,9 @@ type IAMRoleCredentials struct {
 	// while marshalling/unmarshalling this field in the agent. The agent just echo's
 	// whatever is sent by the backend.
 	Expiration string `json:"Expiration"`
+	//RoleType distinguishes between TaskRole and ExecutionRole for the
```
aaithal: spacing: `// RoleType`

agent/logger/audit/entry_types.go
```diff
@@ -38,7 +40,11 @@ const (
 	// 8. version
 	// 9. cluster
 	// 10. container instance arn
-	getCredentialsAuditLogVersion = 1
+
+	// Version '2', following fields were modified
+	// 1. event type ('GetCredentials, GetCredentialsExecutionRole')
```
aaithal: would be better use `7` here, instead of `1`?

agent/acs/handler/refresh_credentials_handler.go
adnxn: you're right, ill add this check.

agent/logger/audit/entry_types.go
adnxn: that was my initial thought too, but also thought it might be confusing. i can change it, i dont feel strongly either way.

agent/credentials/manager.go
adnxn: fixed.

agent/credentials/manager.go
adnxn: fixed.

agent/credentials/manager.go
adnxn: fixed.

PR #1069, opened 2017-11-10
Set CPU/Memory in both config and hostconfig

### LINKED ISSUE #616: Memory limit not enforced on Windows
Containers on Windows are showing up with `"Memory": 0` even when a memory limit is configured in the task definition.  This appears to be related to the `Memory` field moving from `Config` to `HostConfig` in API version 1.18.

On Linux, compatibility is maintained through [reading from both `Config` fields and `HostConfig`](https://github.com/docker/docker/blob/b248de7e332b6e67b08a8981f68060e6ae629ccf/runconfig/config_unix.go), but [Windows omits that code](https://github.com/docker/docker/blob/b248de7e332b6e67b08a8981f68060e6ae629ccf/runconfig/config_windows.go).  See a [1.17 example](https://github.com/docker/docker/blob/b248de7e332b6e67b08a8981f68060e6ae629ccf/runconfig/fixtures/unix/container_config_1_17.json#L5) versus a [1.19 example](https://github.com/docker/docker/blob/b248de7e332b6e67b08a8981f68060e6ae629ccf/runconfig/fixtures/unix/container_config_1_19.json#L35) to see how the field moved.

This is related to other API compatibility issues as noted in https://github.com/docker/docker/issues/27601.

richardpen:
### Summary

Fix #616 
### Implementation details

Set the CPU and memory in both Docker Config and HostConfig
### Testing

- [x] Manually tested on windows

New tests cover the changes: 

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * Feature - Enable support for task level CPU and memory constraints.
 * Bug - Fixed a bug where a task can be blocked in creating state. [#1048](https://github.com/aws/amazon-ecs-agent/pull/1048)
 * Bug - Fixed dynamic HostPort in container metadata. [#1052](https://github.com/aws/amazon-ecs-agent/pull/1052)
+* Bug - Fixed bug on Windows where container memory limits are not enforced. [#1069](https://github.com/aws/amazon-ecs-agent/pull/1069)
 
 ## 1.15.0
 * Feature - Support for provisioning tasks with ENIs.
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -485,12 +485,7 @@ func (task *Task) SetConfigHostconfigBasedOnVersion(container *Container, config
 		return err
 	}
 
-	dockerAPIVersion_1_18, err := docker.NewAPIVersion("1.18")
-	if err != nil {
-		seelog.Errorf("Creating docker api version 1.18 failed, err: %v", err)
-		return err
-	}
-
+	dockerAPIVersion_1_18 := docker.APIVersion([]int{1, 18})
 	if dockerAPIVersion.GreaterThanOrEqualTo(dockerAPIVersion_1_18) {
 		// Set the memory and cpu in host config
 		if hc != nil {
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/ecscni"
+	"github.com/aws/amazon-ecs-agent/agent/engine/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/engine/emptyvolume"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
 	"github.com/aws/aws-sdk-go/aws/arn"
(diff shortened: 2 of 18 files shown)


agent/api/task.go
```diff
@@ -464,6 +464,25 @@ func (task *Task) dockerConfig(container *Container) (*docker.Config, *DockerCli
 	return config, nil
 }
 
+// SetConfigHostConfigCommonFileds sets the same fields in both Config and HostConfig for backward compatibility
+func (task *Task) SetConfigHostConfigCommonFileds(container *Container, config *docker.Config, hc *docker.HostConfig) {
```
vsiddharth: nit: typo s/SetConfigHostConfigCommonFileds/SetConfigHostConfigCommonFields

agent/api/task.go
```diff
@@ -464,6 +464,25 @@ func (task *Task) dockerConfig(container *Container) (*docker.Config, *DockerCli
...
+	// Convert MB to B
+	dockerMem := int64(container.Memory * 1024 * 1024)
```
vsiddharth: I think we have a const (`bytesPerMegabyte`) for this multiplier in `task_unix.go`. 
Could be moved and reused if necessary.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Feature - Enable support for task level CPU and memory constraints.
 * Bug - Fixed a bug where a task can be blocked in creating state. [#1048](https://github.com/aws/amazon-ecs-agent/pull/1048)
 * Bug - Fixed dynamic HostPort in container metadata. [#1052](https://github.com/aws/amazon-ecs-agent/pull/1052)
+* Bug - Fixed bug on windows where container memory limits isn't enforced. [#1069](https://github.com/aws/amazon-ecs-agent/pull/1069)
```
samuelkarp: s/windows/Windows/
s/isn't/are not/

agent/api/task.go
```diff
@@ -464,6 +464,25 @@ func (task *Task) dockerConfig(container *Container) (*docker.Config, *DockerCli
...
+	if dockerMem != 0 && dockerMem < DockerContainerMinimumMemoryInBytes {
+		dockerMem = DockerContainerMinimumMemoryInBytes
+	}
+	cpuShare := task.dockerCPUShares(container.CPU)
```
samuelkarp: Do we really want to set CPU in both?

richardpen: Based on this [commit](https://github.com/moby/moby/commit/837eec064d2d40a4d86acbc6f47fada8263e0d4c), it looks like the memory and cpu are both moved to HostConfig, I will modify this only be set when the api version is <1.18

agent/api/task_windows.go
```diff
@@ -28,6 +29,8 @@ const (
 	memorySwappinessDefault = -1
 )
 
+var cpus = runtime.NumCPU() * 1024
```
samuelkarp: How about "cpuShareScaleFactor"?

agent/api/task.go
```diff
@@ -464,6 +468,47 @@ func (task *Task) dockerConfig(container *Container) (*docker.Config, *DockerCli
...
+
+	// Docker copied Memory and cpu field into hostconfig in 1.6 with api version(1.18)
+	// https://github.com/moby/moby/commit/837eec064d2d40a4d86acbc6f47fada8263e0d4c
+	dockerAPIVersion, err := docker.NewAPIVersion(string(apiVersion))
+	if err != nil {
+		seelog.Errorf("Creating docker api version failed, err: %v", err)
+		return err
+	}
+
+	dockerAPIVersion_1_18, err := docker.NewAPIVersion("1.18")
```
samuelkarp: `APIVersion` is just `[]int`.  I think you could do this:

```go
var dockerAPIVersion_1_18 := APIVersion([]int{1,18})
```

That gets rid of your `err`.

agent/api/task.go
```diff
@@ -464,6 +468,47 @@ func (task *Task) dockerConfig(container *Container) (*docker.Config, *DockerCli
...
+	if err != nil {
+		seelog.Errorf("Creating docker api version 1.18 failed, err: %v", err)
+		return err
+	}
+
+	if dockerAPIVersion.GreaterThanOrEqualTo(dockerAPIVersion_1_18) {
```
petderek: should we log in an `else` case?

agent/engine/dockerclient/dockerclientfactory.go
```diff
@@ -106,6 +110,18 @@ func (f *factory) FindKnownAPIVersions() []DockerVersion {
 	return knownVersions
 }
 
+// FindClientAPIVersion returns the version of the client from the map
+// TODO we should let go docker client return this version information
```
petderek: the TODO seems like a better option, whats stopping us from doing that?

richardpen: go dockerclient doesn't support this today, we need contribute and update.

PR #1070, opened 2017-11-12
Windows service

samuelkarp:
### Summary
This set of commits enables the agent to be run as a Windows service and write logs into the Windows event log.

amazon-ecs-agent-windows-scm
amazon-ecs-agent-windows-event-viewer

~Note: This is marked WIP as it will fail to build because I haven't vendored the new dependencies (two more packages inside `golang.org/x/sys/windows`).  I'm waiting for https://github.com/aws/amazon-ecs-agent/pull/1009 to be merged so I can manage dependencies with dep.~

### Implementation details
* Added a new command-line argument to run as a Windows service
* Refactored the termination handler to use a function provided as part of the `agent` struct
* Implemented a new log receiver for seelog

### TODO
- [x] Vendor dependencies
- [ ] Set restart settings
- [x] Set dependency on Docker
- [ ] Determine correct recovery settings
- [ ] Determine if these need to be set in Go code or in PowerShell startup?

### Testing
- [x] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`) (as long as you have `golang.org/x/sys/windows on your `GOPATH)
- [x] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass (as long as you have `golang.org/x/sys/windows on your `GOPATH)
- [x] Starts the agent
- [x] Can be Started, Stopped, and Restarted from PowerShell and services.msc
- [x] Writes to Windows event log
- [x] Service stops gracefully on its own when Windows is shutting down or rebooting
- [x] Service starts correctly on boot when service is set to automatic startup or delayed start

Feature - The Amazon ECS agent can now be installed as a Windows service

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,23 +58,27 @@ To install the service, you can do the following:
 
 ```powershell
 PS C:\> # Set up directories the agent uses
-PS C:\> New-Item -Type directory -Path $ProgramFiles\Amazon\ECS
-PS C:\> New-Item -Type directory -Path $ProgramData\Amazon\ECS
+PS C:\> New-Item -Type directory -Path ${env:ProgramFiles}\Amazon\ECS -Force
+PS C:\> New-Item -Type directory -Path ${env:ProgramData}\Amazon\ECS -Force
 PS C:\> # Set up configuration
-PS C:\> $ecsExeDir = "$env:ProgramFiles\Amazon\ECS"
+PS C:\> $ecsExeDir = "${env:ProgramFiles}\Amazon\ECS"
 PS C:\> [Environment]::SetEnvironmentVariable("ECS_CLUSTER", "my-windows-cluster", "Machine")
-PS C:\> [Environment]::SetEnvironmentVariable("ECS_LOGFILE", "$ProgramData\Amazon\ECS\log\ecs-agent.log", "Machine")
-PS C:\> [Environment]::SetEnvironmentVariable("ECS_DATADIR", "$ProgramData\Amazon\ECS\data", "Machine")
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_LOGFILE", "${env:ProgramData}\Amazon\ECS\log\ecs-agent.log", "Machine")
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_DATADIR", "${env:ProgramData}\Amazon\ECS\data", "Machine")
 PS C:\> # Download the agent
 PS C:\> $agentVersion = "latest"
 PS C:\> $agentZipUri = "https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-windows-$agentVersion.zip"
-PS C:\> $zipFile = "$env:TEMP\ecs-agent.zip"
+PS C:\> $zipFile = "${env:TEMP}\ecs-agent.zip"
 PS C:\> Invoke-RestMethod -OutFile $zipFile -Uri $agentZipUri
 PS C:\> # Put the executables in the executable directory.
 PS C:\> Expand-Archive -Path $zipFile -DestinationPath $ecsExeDir -Force
-PS C:\> # Uncomment the following line to enable task IAM roles.
+PS C:\> Set-Location ${ecsExeDir}
+PS C:\> # Set $EnableTaskIAMRoles to $true to enable task IAM roles
 PS C:\> # Note that enabling IAM roles will make port 80 unavailable for tasks.
-PS C:\> # cd '$ecsExeDir'; .\hostsetup.ps1
+PS C:\> [bool]$EnableTaskIAMRoles = $false
+PS C:\> if (${EnableTaskIAMRoles} {
+>> .\hostsetup.ps1
+>> }
 PS C:\> # Install the agent service
 PS C:\> New-Service -Name "AmazonECS" `
         -BinaryPathName "$ecsExeDir\amazon-ecs-agent.exe -windows-service" `
@@ -93,19 +97,19 @@ Start-Service AmazonECS
 
 ```powershell
 PS C:\> # Set up directories the agent uses
-PS C:\> New-Item -Type directory -Path $ProgramFiles\Amazon\ECS
-PS C:\> New-Item -Type directory -Path $ProgramData\Amazon\ECS
+PS C:\> New-Item -Type directory -Path ${env:ProgramFiles}\Amazon\ECS -Force
+PS C:\> New-Item -Type directory -Path ${env:ProgramData}\Amazon\ECS -Force
 PS C:\> # Set up configuration
-PS C:\> $ecsExeDir = "$env:ProgramFiles\Amazon\ECS"
+PS C:\> $ecsExeDir = "${env:ProgramFiles}\Amazon\ECS"
 PS C:\> [Environment]::SetEnvironmentVariable("ECS_CLUSTER", "my-windows-cluster", "Machine")
-PS C:\> [Environment]::SetEnvironmentVariable("ECS_LOGFILE", "$ProgramData\Amazon\ECS\log\ecs-agent.log", "Machine")
-PS C:\> [Environment]::SetEnvironmentVariable("ECS_DATADIR", "$ProgramData\Amazon\ECS\data", "Machine")
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_LOGFILE", "${env:ProgramData}\Amazon\ECS\log\ecs-agent.log", "Machine")
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_DATADIR", "${env:ProgramData}\Amazon\ECS\data", "Machine")
 PS C:\> # Set this environment variable to "true" to enable IAM roles.  Note that enabling IAM roles will make port 80 unavailable for tasks.
 PS C:\> [Environment]::SetEnvironmentVariable("ECS_ENABLE_TASK_IAM_ROLE", "false", "Machine")
 PS C:\> # Download the agent
 PS C:\> $agentVersion = "latest"
 PS C:\> $agentZipUri = "https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-windows-$agentVersion.zip"
-PS C:\> $zipFile = "$env:TEMP\ecs-agent.zip"
+PS C:\> $zipFile = "${env:TEMP}\ecs-agent.zip"
 PS C:\> Invoke-RestMethod -OutFile $zipFile -Uri $agentZipUri
 PS C:\> # Put the executables in the executable directory.
 PS C:\> Expand-Archive -Path $zipFile -DestinationPath $ecsExeDir -Force
--- a/README.md
+++ b/README.md
@@ -48,10 +48,48 @@ See also the Advanced Usage section below.
 
 ### On Windows Server 2016
 
-On Windows Server 2016, the Amazon ECS Container Agent runs as a process on the
-host. Unlike Linux, the agent may not run inside a container as it uses the
-host's registry and the named pipe at `\\.\pipe\docker_engine` to communicate
-with the Docker daemon.
+On Windows Server 2016, the Amazon ECS Container Agent runs as a process or
+service on the host. Unlike Linux, the agent may not run inside a container as
+it uses the host's registry and the named pipe at `\\.\pipe\docker_engine` to
+communicate with the Docker daemon.
+
+#### As a Service
+To install the service, you can do the following:
+
+```powershell
+PS C:\> # Set up directories the agent uses
+PS C:\> New-Item -Type directory -Path $ProgramFiles\Amazon\ECS
+PS C:\> New-Item -Type directory -Path $ProgramData\Amazon\ECS
+PS C:\> # Set up configuration
+PS C:\> $ecsExeDir = "$env:ProgramFiles\Amazon\ECS"
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_CLUSTER", "my-windows-cluster", "Machine")
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_LOGFILE", "$ProgramData\Amazon\ECS\log\ecs-agent.log", "Machine")
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_DATADIR", "$ProgramData\Amazon\ECS\data", "Machine")
+PS C:\> # Download the agent
+PS C:\> $agentVersion = "latest"
+PS C:\> $agentZipUri = "https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-windows-$agentVersion.zip"
+PS C:\> $zipFile = "$env:TEMP\ecs-agent.zip"
+PS C:\> Invoke-RestMethod -OutFile $zipFile -Uri $agentZipUri
+PS C:\> # Put the executables in the executable directory.
+PS C:\> Expand-Archive -Path $zipFile -DestinationPath $ecsExeDir -Force
+PS C:\> # Uncomment the following line to enable task IAM roles.
+PS C:\> # Note that enabling IAM roles will make port 80 unavailable for tasks.
+PS C:\> # cd '$ecsExeDir'; .\hostsetup.ps1
+PS C:\> # Install the agent service
+PS C:\> New-Service -Name "AmazonECS" `
+        -BinaryPathName "$ecsExeDir\amazon-ecs-agent.exe -windows-service" `
+        -DisplayName "Amazon ECS" `
+        -Description "Amazon ECS service runs the Amazon ECS agent" `
+        -DependsOn Docker `
+        -StartupType Manual
+```
+
+To run the service, you can do the following:
+```powershell
+Start-Service AmazonECS
+```
+
+#### As a Process
 
 ```powershell
 PS C:\> # Set up directories the agent uses
--- a/agent/app/agent_windows.go
+++ b/agent/app/agent_windows.go
@@ -175,14 +175,8 @@ func (h *handler) runAgent(ctx context.Context) {
 
 // sleepCtx provides a cancelable sleep
 func sleepCtx(ctx context.Context, duration time.Duration) {
-	done := make(chan struct{})
-	time.AfterFunc(duration, func() {
-		close(done)
-	})
-	select {
-	case <-ctx.Done():
-	case <-done:
-	}
+	derivedCtx, _ := context.WithDeadline(ctx, time.Now().Add(duration))
+	<-derivedCtx.Done()
 }
 
 type termHandlerIndicator struct {
--- a/agent/app/agent_windows.go
+++ b/agent/app/agent_windows.go
@@ -78,7 +78,8 @@ func (h *handler) Execute(args []string, requests <-chan svc.ChangeRequest, resp
 	select {
 	case <-windowsDone:
 		// Service was told to stop by the Windows API.  This happens either through manual intervention (i.e.,
-		// "Stop-Service ECS") or through system shutdown.  Regardless, this is a normal exit event and not an error.
+		// "Stop-Service AmazonECS") or through system shutdown.  Regardless, this is a normal exit event and not an
+		// error.
 		seelog.Info("Received normal signal from Windows to exit")
 	case <-agentDone:
 		// This means that the agent stopped on its own.  This is where it's appropriate to light the event log on fire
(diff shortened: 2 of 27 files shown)


agent/logger/eventlog_windows.go
```diff
@@ -0,0 +1,84 @@
...
+References to MSDN guidelines can be found here:
+* https://msdn.microsoft.com/en-us/library/windows/desktop/aa363632(v=vs.85).aspx
+* https://msdn.microsoft.com/en-us/library/windows/desktop/aa363648(v=vs.85).aspx
+* https://msdn.microsoft.com/en-us/library/windows/desktop/aa363661(v=vs.85).aspx
+* https://msdn.microsoft.com/en-us/library/windows/desktop/aa363669(v=vs.85).aspx
+*/
+
+const (
+	eventLogName = "AmazonECSAgent"
+	eventLogID   = 999
```
tehshwann: Can we add a comment here explaining that this is a default value so others know its overridden later on a case-by-case basis?

agent/logger/log.go
```diff
@@ -86,7 +87,7 @@ func SetLevel(logLevel string) {
 // GetLevel gets the log level
 func GetLevel() string {
 	levelLock.RLock()
-	defer levelLock.RLock()
+	defer levelLock.RUnlock()
```
tehshwann: Nit: RUnlock -> Runlock or RunLock??

README.md
```diff
@@ -48,10 +48,48 @@ See also the Advanced Usage section below.
...
+it uses the host's registry and the named pipe at `\\.\pipe\docker_engine` to
+communicate with the Docker daemon.
+
+#### As a Service
+To install the service, you can do the following:
+
+```powershell
+PS C:\> # Set up directories the agent uses
+PS C:\> New-Item -Type directory -Path $ProgramFiles\Amazon\ECS
+PS C:\> New-Item -Type directory -Path $ProgramData\Amazon\ECS
```
tehshwann: New-Item -Type directory -Path **$ENV:**ProgramData\Amazon\ECS **-Force**

README.md
```diff
@@ -48,10 +48,48 @@ See also the Advanced Usage section below.
...
+service on the host. Unlike Linux, the agent may not run inside a container as
+it uses the host's registry and the named pipe at `\\.\pipe\docker_engine` to
+communicate with the Docker daemon.
+
+#### As a Service
+To install the service, you can do the following:
+
+```powershell
+PS C:\> # Set up directories the agent uses
+PS C:\> New-Item -Type directory -Path $ProgramFiles\Amazon\ECS
```
tehshwann: New-Item -Type directory -Path **$ENV:**ProgramFiles\Amazon\ECS **-Force**

README.md
```diff
@@ -48,10 +48,48 @@ See also the Advanced Usage section below.
...
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_LOGFILE", "$ProgramData\Amazon\ECS\log\ecs-agent.log", "Machine")
+PS C:\> [Environment]::SetEnvironmentVariable("ECS_DATADIR", "$ProgramData\Amazon\ECS\data", "Machine")
+PS C:\> # Download the agent
+PS C:\> $agentVersion = "latest"
+PS C:\> $agentZipUri = "https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-windows-$agentVersion.zip"
+PS C:\> $zipFile = "$env:TEMP\ecs-agent.zip"
+PS C:\> Invoke-RestMethod -OutFile $zipFile -Uri $agentZipUri
+PS C:\> # Put the executables in the executable directory.
+PS C:\> Expand-Archive -Path $zipFile -DestinationPath $ecsExeDir -Force
+PS C:\> # Uncomment the following line to enable task IAM roles.
```
tehshwann: This isnt super clear cause the following line is a comment. Maybe add a bool here for users to change so its more clear? Also... I think you want to cd into the exec dir in either case??

Example:

Set-Location $ecsExeDir
/# Set '$EnableTaskIAMRoles' to '$true' if you wish to use Task IAM Roles.
[bool]$EnableTaskIAMRoles = $false
if ($EnableTaskIAMRoles) {
    .\hostsetup.ps1
}

README.md
```diff
@@ -48,10 +48,48 @@ See also the Advanced Usage section below.
...
+        -DependsOn Docker `
+        -StartupType Manual
+```
+
+To run the service, you can do the following:
+```powershell
+Start-Service AmazonECS
+```
+
+#### As a Process
```
tehshwann: Do we want to continue to support this? Whats the thought here?

agent/app/agent_windows.go
```diff
@@ -0,0 +1,188 @@
...
+		defer wg.Done()
+		h.runAgent(ctx)
+	}()
+
+	// Wait until one of the goroutines is either told to stop or fails spectacularly.  Under normal conditions we will
+	// be waiting here for a long time.
+	select {
+	case <-windowsDone:
+		// Service was told to stop by the Windows API.  This happens either through manual intervention (i.e.,
+		// "Stop-Service ECS") or through system shutdown.  Regardless, this is a normal exit event and not an error.
```
tehshwann: Stop-Service AmazonECS

agent/app/agent_windows.go
```diff
@@ -0,0 +1,188 @@
...
+// handleWindowsRequests is a loop intended to run in a goroutine.  It handles bidirectional communication with the
+// Windows service manager.  This function works by pretty much immediately moving to running and then waiting for a
+// stop or shut down message from Windows or to be canceled (which could happen if the agent exits by itself and the
+// calling function cancels the context).
+func (h *handler) handleWindowsRequests(ctx context.Context, requests <-chan svc.ChangeRequest, responses chan<- svc.Status) {
+	// Immediately tell Windows that we are pending service start.
+	responses <- svc.Status{State: svc.StartPending}
+	seelog.Info("Starting Windows service")
+
+	// TODO: Pre-start hooks go here (unclear if we need any yet)
```
tehshwann: Do we need log initialization? For example, we may want to put some sort of break in the log so users know when the service was restarted when reading?

agent/app/args/flag.go
```diff
@@ -18,17 +18,19 @@ import "flag"
 const (
 	versionUsage             = "Print the agent version information and exit"
 	logLevelUsage            = "Loglevel: [<crit>|<error>|<warn>|<info>|<debug>]"
+	ecsAttributesUsage       = "Print the Agent's ECS Attributes based on its environment"
 	acceptInsecureCertUsage  = "Disable SSL certificate verification. We do not recommend setting this option"
 	licenseUsage             = "Print the LICENSE and NOTICE files and exit"
 	blacholeEC2MetadataUsage = "Blackhole the EC2 Metadata requests. Setting this option can cause the ECS Agent to fail to work properly.  We do not recommend setting this option"
-	ecsAttributesUsage       = "Print the Agent's ECS Attributes based on its environment"
```
tehshwann: Why are we removing this?

agent/logger/eventlog_windows.go
```diff
@@ -0,0 +1,84 @@
...
+
+// platformLogConfig exposes log configuration for the event log receiver
+func platformLogConfig() string {
+	return `<custom name="wineventlog" formatid="windows" />`
+}
+
+// ReceiveMessage receives a log line from seelog and emits it to the Windows event log
+func (r *eventLogReceiver) ReceiveMessage(message string, level seelog.LogLevel, context seelog.LogContextInterface) error {
+	switch level {
+	case seelog.DebugLvl, seelog.InfoLvl:
```
tehshwann: nit: should this be indented?

agent/sighandlers/termination_handler.go
```diff
@@ -61,11 +66,11 @@ func FinalSave(saver statemanager.Saver, taskEngine engine.TaskEngine) error {
 	engineDisabled := make(chan error)
 
 	disableTimer := time.AfterFunc(engineDisableTimeout, func() {
-		engineDisabled <- errors.New("Timed out waiting for TaskEngine to settle")
+		engineDisabled <- errors.New("final save: timed out waiting for TaskEngine to settle")
```
tehshwann: nit: **F**inal save:...

agent/sighandlers/termination_handler.go
```diff
@@ -75,10 +80,10 @@ func FinalSave(saver statemanager.Saver, taskEngine engine.TaskEngine) error {
 
 	stateSaved := make(chan error)
 	saveTimer := time.AfterFunc(finalSaveTimeout, func() {
-		stateSaved <- errors.New("Timed out trying to save to disk")
+		stateSaved <- errors.New("final save: timed out trying to save to disk")
```
tehshwann: nit: **F**inal...

agent/sighandlers/termination_handler.go
```diff
@@ -75,10 +80,10 @@ func FinalSave(saver statemanager.Saver, taskEngine engine.TaskEngine) error {
...
 	})
 	go func() {
-		log.Debug("Saving state before shutting down")
+		seelog.Debug("Saving state before shutting down")
 		stateSaved <- saver.ForceSave()
```
tehshwann: Excuse my not-Go-savviness... Should this have a trap/catch with a log so in the event that the state was not saved correctly... its at least logged?

adnxn:
@samuelkarp #1009 merged =]

README.md
jahkeup: Is `New-Item -Type dir -Path ... -Force` like `mkdir -p`?

agent/app/agent_windows.go
```diff
@@ -0,0 +1,188 @@
...
+	// be waiting here for a long time.
+	select {
+	case <-windowsDone:
+		// Service was told to stop by the Windows API.  This happens either through manual intervention (i.e.,
+		// "Stop-Service ECS") or through system shutdown.  Regardless, this is a normal exit event and not an error.
+		seelog.Info("Received normal signal from Windows to exit")
+	case <-agentDone:
+		// This means that the agent stopped on its own.  This is where it's appropriate to light the event log on fire
+		// and set off all the alarms.
+		seelog.Error("Exiting")
```
jahkeup: Can we add more context in this exit log message?

agent/app/agent_windows.go
```diff
@@ -0,0 +1,188 @@
...
+	sleepCtx(agentCtx, time.Minute) // give the agent a minute to start and invoke terminationHandler
+
+	// wait for the termination handler to run.  Once the termination handler runs, we can safely exit.  If the agent
+	// exits by itself, the termination handler doesn't need to do anything and skips.  If the agent exits before the
+	// termination handler is invoked, we can exit immediately.
+	wg.Wait()
+}
+
+// sleepCtx provides a cancelable sleep
+func sleepCtx(ctx context.Context, duration time.Duration) {
```
jahkeup: This can instead use the stdlib (or golang.net/x) [`context.WithDeadline`](https://godoc.org/golang.org/x/net/context#WithDeadline) function and wait on `ctx.Done()` of that context.

README.md
tehshwann: Assuming you meant -d, (dir) yes.

agent/logger/eventlog_windows.go
jahkeup: Nope, this is [Go's spec](https://golang.org/ref/spec#Switch_statements) :)

agent/logger/log.go
jahkeup: This look like it's a https://godoc.org/sync#RWMutex, so RUnlock is a reader taking a lock.

agent/sighandlers/termination_handler.go
jahkeup: This is conventional idiomatic go casing for the error message to be lower cased. Here's an example:

`agent error: this package errored fetching data: I couldn't reach the server.`

This can come from 3 different sources and would look weird if it was capitalized after each `:`.

agent/sighandlers/termination_handler.go
jahkeup: See above: idiomatic error messages.

agent/sighandlers/termination_handler.go
jahkeup: The `saver.ForceSave` sends its results on down the `stateSaved` channel and its up to the other side to determine what to do about that. However.. we could be better on how we log failures and its context.

agent/logger/log.go
tehshwann: cool

agent/app/args/flag.go
jahkeup: This got reordered.

agent/logger/eventlog_windows.go
samuelkarp: Right now it's a constant without being overridden.  I'm going to leave it as-is for now and we can come back to making it play more nicely with Windows later.

agent/logger/eventlog_windows.go
samuelkarp: `gofmt` does this.  I recommend running `gofmt` on save every time you modify a Go file.

agent/logger/log.go
samuelkarp: RUnlock is unlocking the read-lock.  This is unrelated to Windows; I just saw gometalinter complain about it while I was working in the package.

agent/sighandlers/termination_handler.go
samuelkarp: The convention is described [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

agent/sighandlers/termination_handler.go
samuelkarp: The log line is above, on line 83.  That function will get triggered on a timeout so we have a record that the save failed.

agent/app/agent_windows.go
samuelkarp: There isn't a ton of context here except that the agent stopped on its own... What would you like me to add here?

README.md
samuelkarp: Yes, I'd like to continue documenting this at least until we vend an easier way to install the agent.

README.md
```diff
@@ -48,26 +48,68 @@ See also the Advanced Usage section below.
...
 
 ```powershell
 PS C:\> # Set up directories the agent uses
-PS C:\> New-Item -Type directory -Path $ProgramFiles\Amazon\ECS
-PS C:\> New-Item -Type directory -Path $ProgramData\Amazon\ECS
+PS C:\> New-Item -Type directory -Path ${env:ProgramFiles}\Amazon\ECS -Force
```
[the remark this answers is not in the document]
samuelkarp: I'm going to leave this as-is for now; we can come back and fix it later.

agent/app/agent_windows_test.go
```diff
@@ -81,3 +88,242 @@ func TestDoStartHappyPath(t *testing.T) {
...
+		cancel()
+		resp = <-responses
+		assert.Equal(t, svc.StopPending, resp.State, "Send StopPending after Cancel")
+		wg.Done()
+	}()
+
+	wg.Wait()
+}
+
+func TestHandler_Execute_WindowsStops(t *testing.T) {
```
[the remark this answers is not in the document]
samuelkarp: Unit tests are run with a 25s timeout right now.  This completes in about a second.  Our unit tests aren't really able to simulate the worst-case right now, so I'd rather not add an assert that won't tell us anything or try to make the unit test different.  Instead, I think this can be addressed in a higher-level acceptance test (probably outside the agent itself).

samuelkarp:
@cabbruzzese Not sure why I can't seem to reply inline to your comment about the event log ID.  Is it important to include it in the README even though it's a constant right now?

cabbruzzese:
Was just looking for the event id to have better visibility on search engines for future debugging. Not a blocker.

README.md
```diff
@@ -48,26 +48,68 @@ See also the Advanced Usage section below.
...
+PS C:\> Expand-Archive -Path $zipFile -DestinationPath $ecsExeDir -Force
+PS C:\> Set-Location ${ecsExeDir}
+PS C:\> # Set $EnableTaskIAMRoles to $true to enable task IAM roles
+PS C:\> # Note that enabling IAM roles will make port 80 unavailable for tasks.
+PS C:\> [bool]$EnableTaskIAMRoles = $false
+PS C:\> if (${EnableTaskIAMRoles} {
+>> .\hostsetup.ps1
+>> }
+PS C:\> # Install the agent service
+PS C:\> New-Service -Name "AmazonECS" `
```
ofiliz: Isn't it confusing the call the ECS Agent service "AmazonECS"? This is the agent. Would it be better to call it "AmazonECSAgent"?

Sorry if this topic came up before, feel free to ignore & point me to any existing discussion.

samuelkarp: I named it "AmazonECS" so that we'd have the flexibility to swap out a bit in terms of implementation without requiring customers to change the name that they're using or making the name confusing.  Specifically, this enables us to potentially transition to running the agent inside a container and using a service on the host to manage the container (similar to what we do with ecs-init on Linux).  If we called it "AmazonECSAgent", we could still do that but it could create confusion because the service is separate from the agent.

samuelkarp:
> Service starts correctly on boot when service is set to automatic startup or delayed start

I had success when setting the service Startup Type to "Automatic (Delayed Start)".  Setting it to "Automatic" caused it to try to start before networking was available and then fail because it couldn't retrieve the instance identity document.  For now, I would say that we want to recommend "Automatic (Delayed Start)" and as a follow up later we should implement retries for getting the instance identity document.

README.md
petderek: AmazonECSService?

agent/logger/eventlog_windows.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+// platformLogConfig exposes log configuration for the event log receiver
+func platformLogConfig() string {
+	return `<custom name="wineventlog" formatid="windows" />`
+}
+
+// ReceiveMessage receives a log line from seelog and emits it to the Windows event log
+func (r *eventLogReceiver) ReceiveMessage(message string, level seelog.LogLevel, context seelog.LogContextInterface) error {
+	switch level {
```
petderek: How tightly is this coupled to seelog? Are we adding significant work for if we want to swap out logging implementations?

agent/app/agent_windows.go
```diff
@@ -0,0 +1,226 @@
...
+	return t.agentRunning
+}
+
+func (t *termHandlerIndicator) agentStopped() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.agentRunning = false
+}
+
+func (t *termHandlerIndicator) done() {
```
petderek: Would recommend this be called 'finish' or 'end'. This isn’t a strong preference, just that 'done()' is used by the context class for listening instead of terminating.

samuelkarp:
Merging this into the `windows` branch for now.  I will have a follow-up PR to adjust what we're doing with the failure/recovery policy and address any additional comments.

agent/app/agent_windows.go
```diff
@@ -0,0 +1,226 @@
...
+			seelog.Criticalf("Error saving state before final shutdown: %v", err)
+		}
+	}
+	h.ecsAgent.setTerminationHandler(terminationHandler)
+
+	go func() {
+		h.ecsAgent.start() // should block forever, unless there is an error
+		// TODO: distinguish between recoverable and unrecoverable errors
+		indicator.agentStopped()
+		cancel()
```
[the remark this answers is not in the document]
samuelkarp: Possibly.  I will be changing some of this in the next PR.

agent/app/agent_windows.go
```diff
@@ -0,0 +1,226 @@
...
+	indicator.wait()
+}
+
+// sleepCtx provides a cancelable sleep
+func sleepCtx(ctx context.Context, duration time.Duration) {
+	derivedCtx, _ := context.WithDeadline(ctx, time.Now().Add(duration))
+	<-derivedCtx.Done()
+}
+
+type termHandlerIndicator struct {
```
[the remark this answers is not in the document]
samuelkarp: Not that I'm aware of, but I didn't do a very deep search yet.

agent/app/agent_windows.go
samuelkarp: Sure, will change in the next PR.

agent/app/agent_windows.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (t *termHandlerIndicator) setInvoked() {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	t.handlerInvoked = true
+}
+
+func (t *termHandlerIndicator) wait() {
+	t.mu.Lock()
+	invoked := t.handlerInvoked
```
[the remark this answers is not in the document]
samuelkarp: Yep.  This is just to determine whether we need to wait on the channel or not.

agent/logger/eventlog_windows.go
samuelkarp: This is Seelog's interface, so it's somewhat tightly coupled.  However, the Windows event log interface is pretty simple so it should be easy to change this if we switch to a different logger as long as that logger has a pluggable backend.

PR #1076, opened 2017-11-14
Fix container change not reported when task status is NONE

### LINKED ISSUE #1075: "Network Bindings - not configured" although there are port mappings in the container.
This is a duplicate of #1057, #1063, and possibly #1064. #1063 was closed as fixed in version 1.15.1, however, I had a chance to test today and the problem is still present in that version.

richardpen:
### Summary

Fix #1075 

### Implementation details

Task status with NONE is dropped in taskShouldBeSent, this commit makes the taskShouldBeSent more clear and stronger to handle both container  event and task event.

### Testing

- [x] Manual test.

New tests cover the changes: 
yes
### Licensing

yes

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -312,12 +312,6 @@ func (client *APIECSClient) SubmitTaskStateChange(change api.TaskStateChange) er
 		return nil
 	}
 
-	// Submit task state change
-	if change.Status == api.TaskStatusNone {
-		seelog.Warnf("SubmitTaskStateChange called with an invalid change: %s", change.String())
-		return errors.New("ecs api client: SubmitTaskStateChange called with an invalid change")
-	}
-
 	if change.Status != api.TaskRunning && change.Status != api.TaskStopped && len(change.Containers) == 0 {
 		seelog.Debugf("Not submitting unsupported upstream task state: %s", change.Status.String())
 		// Not really an error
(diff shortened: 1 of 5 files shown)


nmeyerhans:
I've tested this with my repro case for #1075 and it passes. Task shows `lastStatus: PENDING` while some containers are RUNNING and others are PENDING and blocked on image pulls. Containers within the task that are RUNNING have fully populated metadata, while PENDING containers have empty metadata. As containers start, the task object returned by DescribeTasks is populated, such that when the task itself finally transitions to RUNNING all of the containers are fully described. This matches the desired behavior, so from the outside this looks good.

agent/api/ecsclient/client.go
```diff
@@ -312,12 +312,6 @@ func (client *APIECSClient) SubmitTaskStateChange(change api.TaskStateChange) er
 		return nil
 	}
 
-	// Submit task state change
-	if change.Status == api.TaskStatusNone {
-		seelog.Warnf("SubmitTaskStateChange called with an invalid change: %s", change.String())
-		return errors.New("ecs api client: SubmitTaskStateChange called with an invalid change")
-	}
-
 	if change.Status != api.TaskRunning && change.Status != api.TaskStopped && len(change.Containers) == 0 {
```
aaithal: let's also get rid of this block altogether. This determination should be happening in `sendableEvent .taskShouldBeSent` method: https://github.com/richardpen/amazon-ecs-agent/blob/fix-1075/agent/eventhandler/task_handler_types.go#L67

PR #1077, opened 2017-11-15
Enable metrics for windows

### LINKED ISSUE #608: Metrics for Windows not being sent to ECS
Docker stats was added to Windows with the merge of this PR https://github.com/docker/docker/pull/25737 and was back ported to the 1.12.x branch. However it still remains unavailable for calls using pre v1.21 API.  However, Agent currently doesn't report metrics for from Windows Container instance nodes. This is in large part due to the fact that docker stats does not return stats in docker API versions pre v1.21 ( [docker/docker/daemon/stats.go#77](https://github.com/docker/docker/blob/master/daemon/stats.go#L77 ) ). 

Agent needs to be updated to the current API version in order to make use of Windows docker stats and re-enable sending of metrics for Windows.

richardpen:
### Summary

Fix #608
### Implementation details

Change the default configuration of metrics on windows to be enabled by default.

### Testing

- [x] Manual test.

New tests cover the changes: 
yes

### Licensing

yes

### DIFF
--- a/agent/stats/utils_unix.go
+++ b/agent/stats/utils_unix.go
@@ -24,9 +24,9 @@ import (
 // dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
 func dockerStatsToContainerStats(dockerStats *docker.Stats) (*ContainerStats, error) {
 	// The length of PercpuUsage represents the number of cores in an instance.
-	if len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0 {
-		seelog.Debug("Invalid container statistics reported, invalid stats payload from docker")
-		return nil, fmt.Errorf("Invalid container statistics reported")
+	if len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0 || numCores == uint64(0) {
+		seelog.Debug("Invalid container statistics reported, no cpu core usage reported")
+		return nil, fmt.Errorf("Invalid container statistics reported, no cpu core usage reported")
 	}
 
 	cpuUsage := dockerStats.CPUStats.CPUUsage.TotalUsage / numCores
--- a/agent/stats/utils_unix.go
+++ b/agent/stats/utils_unix.go
@@ -1,0 +1,39 @@
+// +build !windows
+// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package stats
+
+import (
+	"fmt"
+
+	"github.com/cihub/seelog"
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+// dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
+func dockerStatsToContainerStats(dockerStats *docker.Stats) (*ContainerStats, error) {
+	// The length of PercpuUsage represents the number of cores in an instance.
+	if len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0 {
+		seelog.Debug("Invalid container statistics reported, invalid stats payload from docker")
+		return nil, fmt.Errorf("Invalid container statistics reported")
+	}
+
+	cpuUsage := dockerStats.CPUStats.CPUUsage.TotalUsage / numCores
+	memoryUsage := dockerStats.MemoryStats.Usage - dockerStats.MemoryStats.Stats.Cache
+	return &ContainerStats{
+		cpuUsage:    cpuUsage,
+		memoryUsage: memoryUsage,
+		timestamp:   dockerStats.Read,
+	}, nil
+}
--- a/agent/stats/utils_windows.go
+++ b/agent/stats/utils_windows.go
@@ -24,8 +24,8 @@ import (
 // dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
 func dockerStatsToContainerStats(dockerStats *docker.Stats) (*ContainerStats, error) {
 	if numCores == uint64(0) {
-		seelog.Error("Invalid number of cpu cores")
-		return nil, fmt.Errorf("Invalid number of cpu cores")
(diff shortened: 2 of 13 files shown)


agent/stats/utils_windows.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+
+	"github.com/cihub/seelog"
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+// dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
+func dockerStatsToContainerStats(dockerStats *docker.Stats) (*ContainerStats, error) {
+	if numCores == uint64(0) {
+		seelog.Error("Invalid number of cpu cores")
```
aaithal: Please add more context here: "Stats: invalid number of cpu cores in response"

agent/stats/utils_windows.go
```diff
@@ -0,0 +1,38 @@
...
+		return nil, fmt.Errorf("Invalid number of cpu cores")
```
aaithal: should start with a lower-case letter and conform to https://github.com/golang/go/wiki/CodeReviewComments#error-strings

agent/stats/utils_unix.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/cihub/seelog"
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+// dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
+func dockerStatsToContainerStats(dockerStats *docker.Stats) (*ContainerStats, error) {
+	// The length of PercpuUsage represents the number of cores in an instance.
+	if len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0 {
+		seelog.Debug("Invalid container statistics reported, invalid stats payload from docker")
+		return nil, fmt.Errorf("Invalid container statistics reported")
```
aaithal: please fix the error message here as well.

agent/stats/utils_windows.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"fmt"
+
+	"github.com/cihub/seelog"
+	docker "github.com/fsouza/go-dockerclient"
+)
+
+// dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
+func dockerStatsToContainerStats(dockerStats *docker.Stats) (*ContainerStats, error) {
```
aaithal: Can you add a unit test for this?

richardpen: It should be covered [here](https://github.com/aws/amazon-ecs-agent/pull/1077/files#diff-13afca01e3a8339ae7856f996a5b1859R85)

agent/stats/utils_windows.go
```diff
@@ -0,0 +1,38 @@
...
+	if numCores == uint64(0) {
```
sharanyad: When would this happen?

agent/stats/utils_windows.go
aaithal: it doesn't cover `numCores` block, right?

agent/stats/utils_windows.go
richardpen: theory it shouldn't happen, as it's acquired from [runtime](https://golang.org/pkg/runtime/#NumCPU). But I don't know when will it happen.

agent/stats/utils_unix.go
```diff
@@ -0,0 +1,39 @@
...
+
+// dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
+func dockerStatsToContainerStats(dockerStats *docker.Stats) (*ContainerStats, error) {
+	// The length of PercpuUsage represents the number of cores in an instance.
+	if len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0 {
+		seelog.Debug("Invalid container statistics reported, no cpu core usage reported")
+		return nil, fmt.Errorf("Invalid container statistics reported, no cpu core usage reported")
+	}
+
+	cpuUsage := dockerStats.CPUStats.CPUUsage.TotalUsage / numCores
```
sharanyad: Should we do a zero check for `numCores` here too?

agent/stats/utils_windows.go
aaithal: @sharanyad we should always remember that the runtime can send weird unexpected values to us (which has happened in the past). Since in this case, it'll lead to a panic, it's better to be protected

agent/stats/utils_unix.go
richardpen: I will add that.

agent/stats/utils_windows.go
```diff
@@ -0,0 +1,38 @@
...
+		seelog.Error("Invalid number of cpu cores acquired from the system")
+		return nil, fmt.Errorf("invalid number of cpu cores acquired from the system")
```
sharanyad: Can we still send `memoryUsage` and `timestamp` information since they don't require `numCores` ?

agent/stats/utils_windows.go
richardpen: yes, you're right, adding it.

agent/stats/utils_windows.go
richardpen: I don't have a strong opinion here, but my suggestion is that we shouldn't send part of the metrics if something was wrong, not sending wrong metrics is better than sending wrong metrics I think.

agent/stats/utils_windows_test.go
```diff
@@ -0,0 +1,40 @@
+// +build windows,!integration
+// Copyright 2014-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sharanyad: nit: 2017

agent/stats/utils_windows.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	cpuUsage := dockerStats.CPUStats.CPUUsage.TotalUsage / numCores
```
samuelkarp: Can we verify that dividing by `numCores` is appropriate on Windows?  From when I was playing around with it on a 4-core machine, `docker stats` showed `100%` usage on Windows when using all 4 cores but showed `400%` on Linux for the same workload.

petderek: Does it mean different things on each platform?

richardpen: @samuelkarp  confirmed in that case, the metrics will be the same as docker stats, it will show 100%.

samuelkarp: So do we need to divide by numCores or should we remove that?

richardpen: we should keep it, also this [code](https://github.com/moby/moby/blob/master/integration-cli/docker_api_stats_test.go#L50) indicates the same.

samuelkarp: As discussed offline: I don't have an objection to merging this as long as we close out on the right math before we release anything.

PR #1090, opened 2017-11-17
adding functional test for oom on windows

richardpen:
### Implementation details

Running a task has memory limit set, and try to use more memory than the limit should cause the container to fail.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/functional_tests/tests/functionaltests_windows_test.go
+++ b/agent/functional_tests/tests/functionaltests_windows_test.go
@@ -307,3 +307,17 @@ func TestTelemetry(t *testing.T) {
 	err = VerifyMetrics(cwclient, params, true)
 	assert.NoError(t, err, "Task stopped, verify metrics for memory utilization failed")
 }
+
+// TestOOMContainer verifies that an OOM container returns an error
+func TestOOMContainer(t *testing.T) {
+	agent := RunAgent(t, nil)
+	defer agent.Cleanup()
+
+	testTask, err := agent.StartTask(t, "oom-windows")
+	require.NoError(t, err, "Expected to start invalid-image task")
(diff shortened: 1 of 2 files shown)


agent/functional_tests/tests/functionaltests_windows_test.go
```diff
@@ -307,3 +307,17 @@ func TestTelemetry(t *testing.T) {
...
+	agent := RunAgent(t, nil)
+	defer agent.Cleanup()
+
+	testTask, err := agent.StartTask(t, "oom-windows")
+	require.NoError(t, err, "Expected to start invalid-image task")
+	err = testTask.WaitRunning(waitTaskStateChangeDuration)
+	assert.NoError(t, err, "Expect task to be running")
+	err = testTask.WaitStopped(waitTaskStateChangeDuration)
+	assert.NoError(t, err, "Expect task to be stopped")
+	assert.NotEqual(t, 0, testTask.Containers[0].ExitCode, "container should fail with memory error")
```
samuelkarp: Do we get a stopped reason indicating out of memory?

richardpen: Nothing except the exit code is 1, and docker logs shows `MemeryError` which is generated by python.

samuelkarp: Does `docker inspect` show the OOM flag to be true?

richardpen: No, it's still false, there may be something different on windows and linux which I hadn't dig into. But the behavior looks like is expected, if the container try to get more memory it will fail.

PR #1095, opened 2017-11-20
Set the cpu percent to minimum 1 if cpu shares is too low on windows

richardpen:
### Implementation details

When convert the cpu shares to cpu percent on windows, if the cpushares is too low the cpu percent will be 0, the commit will ensure the minimum cpu percent will be 1.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/task_test.go
+++ b/agent/api/task_test.go
@@ -16,6 +16,7 @@ package api
 import (
 	"encoding/json"
 	"reflect"
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


agent/api/task_test.go
```diff
@@ -174,99 +174,6 @@ func TestDockerHostConfigVolumesFrom(t *testing.T) {
 	}
 }
 
-func TestDockerHostConfigRawConfig(t *testing.T) {
```
samuelkarp: Instead of moving the actual tests, can you just define constants that are different for Windows and Linux and keep the tests here?

richardpen: Done

agent/api/task_test.go
```diff
@@ -214,6 +214,11 @@ func TestDockerHostConfigRawConfig(t *testing.T) {
 	assert.Nil(t, configErr)
 
 	expectedOutput := rawHostConfigInput
+	expectedOutput.CPUPercent = minimumCPUPercent
+	if minimumCPUPercent == 1 {
```
samuelkarp: You could check `runtime.GOOS == "windows"` if you want to make an OS assertion.

PR #1108, opened 2017-11-22
statemanager: handle non-existent file on startup

### LINKED ISSUE #722: On first startup, agent tries to remove old state file that does not exist
> 2017-03-03T02:41:51Z [INFO] Created new registry key: HKEY_LOCAL_MACHINE\SOFTWARE\Amazon\ECS Agent\State File
2017-03-03T02:41:51Z [ERROR] Error removing old file ; err remove : The system cannot find the path specified.

There's no old state file to remove, but the agent tries to remove `""` which does not exist.

There is no impact to this bug except the log line above.

adnxn:
### Summary

Merging @vsiddharth's fix into dev: https://github.com/vsiddharth/amazon-ecs-agent/commit/23dfb18aa66bcc791f70ec546003ec0c99776ced

This patch ensures that the statemanager does not attempt to remove the
old("")
state file on startup.

Fixes #722

### Testing

New tests cover the changes: 

### Licensing

yes

### DIFF
--- a/agent/statemanager/state_manager_windows.go
+++ b/agent/statemanager/state_manager_windows.go
@@ -146,9 +146,14 @@ func (manager *basicStateManager) writeFile(data []byte) error {
 		seelog.Errorf("Failed to save the data file path: %v", err)
 		return err
 	}
+	if oldFile == "" {
+		return nil
+	}
+
 	err = deps.fs.Remove(oldFile)
 	if err != nil {
 		seelog.Errorf("Error removing old file %s; err %v", oldFile, err)
+		return err
 	}
 	return nil
 }
--- a/agent/statemanager/state_manager_windows_test.go
+++ b/agent/statemanager/state_manager_windows_test.go
@@ -211,4 +211,27 @@ func TestStateManagerSave(t *testing.T) {
 	assert.Nil(t, err)
 }
 
+func TestStateManagerNoOldStateRemoval(t *testing.T) {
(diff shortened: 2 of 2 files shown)


agent/statemanager/state_manager_windows.go
```diff
@@ -146,9 +146,12 @@ func (manager *basicStateManager) writeFile(data []byte) error {
 		seelog.Errorf("Failed to save the data file path: %v", err)
 		return err
 	}
-	err = deps.fs.Remove(oldFile)
-	if err != nil {
-		seelog.Errorf("Error removing old file %s; err %v", oldFile, err)
+	if oldFile != "" {
```
samuelkarp: Can you invert this so that the normal case is unindented?

```go

if oldFile == "" {
	return nil
}
// do the removal
```

agent/statemanager/state_manager_windows.go
```diff
@@ -146,11 +146,15 @@ func (manager *basicStateManager) writeFile(data []byte) error {
...
+		return nil
+	}
+
 	err = deps.fs.Remove(oldFile)
 	if err != nil {
 		seelog.Errorf("Error removing old file %s; err %v", oldFile, err)
+		return err
 	}
-	return nil
 }
```
aaithal: i think you still need `return nil` here.

adnxn: oops. appveyor was telling me the same thing -> `agent\statemanager\state_manager_windows.go:158: missing return at end of function
`

PR #1109, opened 2017-11-22
adding a 'awsvpc' functional test

aaithal:
The test starts a task with an 'nginx' container, gets the ip address of
the container by descrining the task and queries the ip address for a
response from the nginx container.

### Summary

Adding a 'awsvpc' functional test

### Implementation details

The `TestNetworkModeAWSVPC ` test starts a task with an 'nginx' container, gets the ip address of the container by describing the task and queries the ip address for a response from the nginx container.

### Testing

New tests cover the changes: Yes, the test is the test

### Licensing

### DIFF
--- a/agent/functional_tests/util/utils.go
+++ b/agent/functional_tests/util/utils.go
@@ -18,7 +18,6 @@ package util
 import (
 	"crypto/md5"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
@@ -39,6 +38,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
 	"github.com/aws/aws-sdk-go/service/iam"
 	docker "github.com/fsouza/go-dockerclient"
+	"github.com/pkg/errors"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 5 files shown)


agent/functional_tests/util/utils.go
```diff
@@ -220,6 +221,48 @@ func (agent *TestAgent) StartTaskWithTaskDefinitionOverrides(t *testing.T, task
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	if len(resp.Failures) != 0 || len(resp.Tasks) == 0 {
+		return nil, errors.New("Failure starting task: " + *resp.Failures[0].Reason)
+	}
+
+	task := resp.Tasks[0]
+	agent.t.Logf("Started task: %s\n", *task.TaskArn)
```
richardpen: Why `agent.t` instead of `t`?

agent/functional_tests/util/utils.go
```diff
@@ -533,6 +577,14 @@ func (task *TestTask) Stop() error {
 	return err
 }
 
+func (task *TestTask) GetAttachmentInfo() ([]*ecs.KeyValuePair, error) {
```
richardpen: Can you add comment for this method?

agent/functional_tests/util/utils.go
aaithal: because it's there :) I'll modify this so as to not pass the `testing.T` object again

agent/functional_tests/util/utils.go
aaithal: sure

PR #1116, opened 2017-11-28
Adding functional tests for execution roles

richardpen:
### Summary

Add functional test for using execution roles with ECR pull and awslogs

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,7 +99,7 @@ test-in-docker:
 	# Privileged needed for docker-in-docker so integ tests pass
 	docker run --net=none -v "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" --privileged "amazon/amazon-ecs-agent-test:make"
 
-run-functional-tests: testnnp test-registry
+run-functional-tests: testnnp test-registry ecr-execution-role-image
 	. ./scripts/shared_env && go test -tags functional -timeout=30m -v ./agent/functional_tests/...
 
 testnnp:
@@ -144,7 +144,7 @@ volumes-test:
 
 # TODO, replace this with a build on dockerhub or a mechanism for the
 # functional tests themselves to build this
-.PHONY: squid awscli fluentd gremlin taskmetadata-validator image-cleanup-test-images
+.PHONY: squid awscli fluentd gremlin taskmetadata-validator image-cleanup-test-images ecr-execution-role-image
 squid:
 	$(MAKE) -C misc/squid $(MFLAGS)
 
@@ -162,6 +162,8 @@ image-cleanup-test-images:
 
 taskmetadata-validator:
 	$(MAKE) -C misc/taskmetadata-validator $(MFLAGS)
+ecr-execution-role-image:
+	$(MAKE) -C misc/ecr $(MFLAGS)
 
 .get-deps-stamp:
 	go get golang.org/x/tools/cmd/cover
--- a/agent/functional_tests/README.md
+++ b/agent/functional_tests/README.md
@@ -59,6 +59,12 @@ done first:
   `iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679`
   and `export TEST_TASK_IAM_ROLE_NET_HOST=true`
 
+#### Additional setup for EXECUTION roles
+In order to run ExecutionRole functional tests, the following environment variable needs to be set:
+`TEST_EXECUTION_ROLE`: true, whether to run the test `TestExecutionRole`.
+`ECS_FTS_EXECUTIONROLE_ECR_IMAGE`: container image from ECR, the container image
+used for this test, which should output `hello world`.
+
 ### Windows
 
 Before running these tests, you should build the ECS agent (as `agent.exe`) and
--- a/agent/functional_tests/README.md
+++ b/agent/functional_tests/README.md
@@ -42,12 +42,10 @@ execution behavior:
 * `ECS_FTEST_FORCE_NET_HOST`: Run the agent with `--net=host`
 
 #### Additional setup for IAM roles
-In order to run TaskIamRole functional tests, the following steps should b
+In order to run TaskIamRole functional tests, the following steps should be
 done first:
 * Run command: `sysctl -w net.ipv4.conf.all.route_localnet=1` and
   `iptables -t nat -A PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-destination 127.0.0.1:51679`.
-* Set the environment variable to enable the test under default network mode:
-  `export TEST_TASK_IAM_ROLE=true`.
 * Set the environment variable of IAM roles the test will use:
   `export TASK_IAM_ROLE_ARN="iam role arn"`. The role should have the
   `ec2:DescribeRegions` permission and have a trust relationship with
@@ -55,13 +53,16 @@ done first:
   isn't set, the test will use the IAM role attached to the instance profile.
   In this case, the IAM role should additionally have the
   `iam:GetInstanceProfile` permission.
+* To skip the test `TestTaskIAMRolesDefaultNetworkMode`, please set `TEST_DISABLE_TASK_IAM_ROLE` to be true.
 * Testing under net=host network mode requires additional commands:
   `iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679`
-  and `export TEST_TASK_IAM_ROLE_NET_HOST=true`
+* To skip the test `TestTaskIAMRolesNetHostMode`, please set `TEST_DISABLE_TASK_IAM_ROLE_NET_HOST` to be true.
+
 
 #### Additional setup for EXECUTION roles
-In order to run ExecutionRole functional test, the following environment variable needs to be set:
-`TEST_EXECUTION_ROLE`: true, whether to run the test `TestExecutionRole`.
+* In order to run ExecutionRole functional test, the following environment variable needs to be set:
+* `ECS_FTS_EXECUTION_ROLE`: execution role arn. The role should have permission to cloudwatch logs and ECR pull.
+* To skip the test `TestExecutionRole`, please set `TEST_DISABLE_EXECUTION_ROLE` to be true.
 
 ### Windows
 
--- a/agent/functional_tests/README.md
+++ b/agent/functional_tests/README.md
@@ -60,10 +60,8 @@ done first:
   and `export TEST_TASK_IAM_ROLE_NET_HOST=true`
 
 #### Additional setup for EXECUTION roles
-In order to run ExecutionRole functional tests, the following environment variable needs to be set:
+In order to run ExecutionRole functional test, the following environment variable needs to be set:
 `TEST_EXECUTION_ROLE`: true, whether to run the test `TestExecutionRole`.
-`ECS_FTS_EXECUTIONROLE_ECR_IMAGE`: container image from ECR, the container image
-used for this test, which should output `hello world`.
 
 ### Windows
 
--- a/agent/functional_tests/tests/functionaltests_unix_test.go
+++ b/agent/functional_tests/tests/functionaltests_unix_test.go
@@ -704,3 +704,63 @@ func TestTaskMetadataValidator(t *testing.T) {
 
 	assert.Equal(t, 42, exitCode, fmt.Sprintf("Expected exit code of 42; got %d", exitCode))
 }
+
+// TestExecutionRole verifies that task can use the execution credentials to pull from ECR and
+// send logs to cloudwatch with awslogs driver
+func TestExecutionRole(t *testing.T) {
+	if os.Getenv("TEST_EXECUTION_ROLE") != "true" {
+		t.Skip("TEST_EXECUTION_ROLE was not set to true")
+	}
+
+	RequireDockerVersion(t, ">=17.06.2-ce") // awslogs drivers with execution role available from docker 17.06.2
+	cwlClient := cloudwatchlogs.New(session.New(), aws.NewConfig().WithRegion(*ECS.Config.Region))
+
+	agentOptions := AgentOptions{
+		ExtraEnvironment: map[string]string{
+			"ECS_AVAILABLE_LOGGING_DRIVERS":             `["awslogs"]`,
+			"ECS_ENABLE_AWSLOGS_EXECUTIONROLE_OVERRIDE": "true",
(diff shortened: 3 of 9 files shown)


agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -669,3 +669,65 @@ func TestMetadataServiceValidator(t *testing.T) {
...
+	testTask, err := agent.StartTaskWithTaskDefinitionOverrides(t, "execution-role", tdOverrides)
+	require.NoError(t, err, "Expected to start task using awslogs driver failed")
+
+	// Wait for the container to start
+	testTask.WaitRunning(waitTaskStateChangeDuration)
+	strs := strings.Split(*testTask.TaskArn, "/")
+	taskId := strs[len(strs)-1]
+
+	// Delete the log stream after the test
+	defer func() {
```
aaithal: minor: you can just do `defer cwlClient.DeleteLogStream(...)` here

agent/functional_tests/README.md
```diff
@@ -59,6 +59,12 @@ done first:
   `iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679`
   and `export TEST_TASK_IAM_ROLE_NET_HOST=true`
 
+#### Additional setup for EXECUTION roles
+In order to run ExecutionRole functional tests, the following environment variable needs to be set:
+`TEST_EXECUTION_ROLE`: true, whether to run the test `TestExecutionRole`.
+`ECS_FTS_EXECUTIONROLE_ECR_IMAGE`: container image from ECR, the container image
```
aaithal: Instead of this, can we build the ecr image and upload it to ecr in the test itself?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -669,3 +669,65 @@ func TestMetadataServiceValidator(t *testing.T) {
...
+	tdOverrides["$$$$EXECUTION_ROLE$$$$"] = os.Getenv("ECS_FTS_EXECUTION_ROLE")
+	tdOverrides["$$$$IMAGE$$$$"] = os.Getenv("ECS_FTS_EXECUTIONROLE_ECR_IMAGE")
+
+	testTask, err := agent.StartTaskWithTaskDefinitionOverrides(t, "execution-role", tdOverrides)
+	require.NoError(t, err, "Expected to start task using awslogs driver failed")
+
+	// Wait for the container to start
+	testTask.WaitRunning(waitTaskStateChangeDuration)
+	strs := strings.Split(*testTask.TaskArn, "/")
+	taskId := strs[len(strs)-1]
```
adnxn: we could use the `aws-sdk-go/aws/arn`'s parser here. similar to how we use it in `task.go` [here](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/api/task.go#L1033)

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -669,3 +669,65 @@ func TestMetadataServiceValidator(t *testing.T) {
...
+	params := &cloudwatchlogs.GetLogEventsInput{
+		LogGroupName:  aws.String(awslogsLogGroupName),
+		LogStreamName: aws.String(fmt.Sprintf("ecs-functional-tests/executionrole-awslogs-test/%s", taskId)),
+	}
+
+	resp, err := waitCloudwatchLogs(cwlClient, params)
+	require.NoError(t, err, "CloudWatchLogs get log failed")
+	assert.Len(t, resp.Events, 1, fmt.Sprintf("Get unexpected number of log events: %d", len(resp.Events)))
+	assert.Equal(t, *resp.Events[0].Message, "hello world", fmt.Sprintf("Got log events message unexpected: %s", *resp.Events[0].Message))
+	// Search the audit log to verify the credential request from awslogs driver
```
adnxn: oh this is a good idea, i didn't think to check the audit log in the functional test as well.

Makefile
```diff
@@ -160,6 +160,9 @@ fluentd:
 image-cleanup-test-images:
 	$(MAKE) -C misc/image-cleanup-test-images $(MFLAGS)
 
+functional-tests-image:
```
aaithal: can you rename this to something more descriptive? Example: `ecr-execution-role-image`

Makefile
```diff
@@ -22,7 +22,7 @@ ECS_CNI_REPOSITORY_REVISION=master
 ECS_CNI_REPOSITORY_SRC_DIR=$(PWD)/amazon-ecs-cni-plugins
 
 
-.PHONY: all gobuild static docker release certs test clean netkitten test-registry run-functional-tests gremlin benchmark-test gogenerate run-integ-tests image-cleanup-test-images pause-container get-cni-sources cni-plugins
+.PHONY: all gobuild static docker release certs test clean netkitten test-registry run-functional-tests gremlin benchmark-test gogenerate run-integ-tests image-cleanup-test-images pause-container get-cni-sources cni-plugins functional-tests-image
```
aaithal: instead of adding the new target here, can you please add it in line 147?

agent/functional_tests/README.md
```diff
@@ -59,6 +59,11 @@ done first:
   `iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679`
   and `export TEST_TASK_IAM_ROLE_NET_HOST=true`
 
+#### Additional setup for EXECUTION roles
+In order to run ExecutionRole functional tests, the following environment variable needs to be set:
```
aaithal: nit: %s/tests/test/

agent/functional_tests/README.md
```diff
@@ -59,6 +59,11 @@ done first:
...
+`TEST_EXECUTION_ROLE`: true, whether to run the test `TestExecutionRole`.
+used for this test, which should output `hello world`.
```
aaithal: nit: which is expected to print "hello world" to stdout

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -669,3 +669,66 @@ func TestMetadataServiceValidator(t *testing.T) {
...
 }
+
+// TestExecutionRole verifies that task can use the execution credentials to pull from ECR and
+// send logs to cloudwatch with awslogs driver
+func TestExecutionRole(t *testing.T) {
+	if os.Getenv("TEST_EXECUTION_ROLE") != "true" {
+		t.Skip("TEST_EXECUTION_ROLE was not set to true")
+	}
+
+	RequireDockerVersion(t, ">=17.06.2-ce") // awslogs drivers with execution role available from docker 17.06.2
```
aaithal: didn't you mention in another PR that `-ce` would not be expected? do we need it here?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -669,3 +669,66 @@ func TestMetadataServiceValidator(t *testing.T) {
...
+			"51679/tcp": {
+				"HostIP":   "0.0.0.0",
+				"HostPort": "51679",
+			},
+		},
+	}
+	agent := RunAgent(t, &agentOptions)
+	defer agent.Cleanup()
+	tdOverrides := make(map[string]string)
+	accountID, err := GetAccountID()
```
aaithal: it seems like it'd be better to do this before even running the agent so that we fail fast.

agent/functional_tests/tests/functionaltests_unix_test.go
aaithal: let's use sdk's `arn` package here

misc/ecr/build.sh
```diff
@@ -0,0 +1,29 @@
...
+eval $(aws ecr get-login --region ${region} --no-include-email)
+repository=${accountID}.dkr.ecr.${region}.amazonaws.com/executionrole:fts
+
+if ! aws ecr describe-repositories --region ${region}| grep -q 'repositoryName": "executionrole"' ;
+then
+    aws ecr create-repository --region ${region} --repository-name executionrole
+fi
+
+docker build -t ${repository} .
+docker push ${repository}
```
aaithal: I wonder if we can avoid uploading the image to ECR on each build (using md5sum of the executable as a tag or something). If not, we should at least set the life cycle policy on this repo.

agent/functional_tests/tests/functionaltests_unix_test.go
richardpen: Actually we needed it, I tested that the `17.06.2-ce>=17.06.2` will return false, I have modified in line 674 also.

samuelkarp:
> Actually we needed it, I tested that the `17.06.2-ce>=17.06.2` will return false, I have modified in line 674 also.

The comparator we're using follows [SemVer 2.0.0](https://semver.org/) where a `-` indicates a prerelease (which would imply that `-ce` is a prerelease).  Docker doesn't follow SemVer, so using that as our comparator is probably wrong.  Instead, we should probably treat `-ce` and `-ee*` as special.

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -704,3 +704,67 @@ func TestTaskMetadataValidator(t *testing.T) {
...
+			},
+		},
+	}
+	agent := RunAgent(t, &agentOptions)
+	defer agent.Cleanup()
+	tdOverrides := make(map[string]string)
+	testImage := fmt.Sprintf("%s.dkr.ecr.%s.amazonaws.com/executionrole:fts", accountID, *ECS.Config.Region)
+
+	tdOverrides["$$$$TEST_REGION$$$$"] = *ECS.Config.Region
+	tdOverrides["$$$$EXECUTION_ROLE$$$$"] = os.Getenv("ECS_FTS_EXECUTION_ROLE")
```
adnxn: we should have some documentation on how `ECS_FTS_EXECUTION_ROLE` is set.

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -704,3 +704,67 @@ func TestTaskMetadataValidator(t *testing.T) {
...
 }
+
+// TestExecutionRole verifies that task can use the execution credentials to pull from ECR and
+// send logs to cloudwatch with awslogs driver
+func TestExecutionRole(t *testing.T) {
+	if os.Getenv("TEST_EXECUTION_ROLE") != "true" {
+		t.Skip("TEST_EXECUTION_ROLE was not set to true")
+	}
+
+	RequireDockerVersion(t, ">=17.06.2-ce") // awslogs drivers with execution role available from docker 17.06.2
```
sharanyad: Is there a remote API version which supports execution role, that we can use to compare, instead of the Docker version itself?

richardpen: It's the same as version in this case, since this version has a patch so it will have the same API version as the one without patch.

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -322,20 +322,20 @@ func TestAWSLogsDriver(t *testing.T) {
 
 	// Wait for the container to start
 	testTask.WaitRunning(waitTaskStateChangeDuration)
-	strs := strings.Split(*testTask.TaskArn, "/")
-	taskId := strs[len(strs)-1]
+	taskID, err := GetTaskID(*testTask.TaskArn)
```
aaithal: please use `aws.StringValue()` here

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -704,3 +704,67 @@ func TestTaskMetadataValidator(t *testing.T) {
...
+
+	cwlClient := cloudwatchlogs.New(session.New(), aws.NewConfig().WithRegion(*ECS.Config.Region))
+
+	agentOptions := AgentOptions{
+		ExtraEnvironment: map[string]string{
+			"ECS_AVAILABLE_LOGGING_DRIVERS":             `["awslogs"]`,
+			"ECS_ENABLE_AWSLOGS_EXECUTIONROLE_OVERRIDE": "true",
+			"ECS_ENABLE_TASK_IAM_ROLE":                  "true",
+		},
+		PortBindings: map[docker.Port]map[string]string{
```
aaithal: why not just launch the agent in `host` networking mode here?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -704,3 +704,67 @@ func TestTaskMetadataValidator(t *testing.T) {
...
+				"HostPort": "51679",
+			},
+		},
+	}
+	agent := RunAgent(t, &agentOptions)
+	defer agent.Cleanup()
+	tdOverrides := make(map[string]string)
+	testImage := fmt.Sprintf("%s.dkr.ecr.%s.amazonaws.com/executionrole:fts", accountID, *ECS.Config.Region)
+
+	tdOverrides["$$$$TEST_REGION$$$$"] = *ECS.Config.Region
```
aaithal: please use `aws.StringValue()` here (and in other places where you're dereferencing pointers in this file)

agent/functional_tests/util/utils.go
```diff
@@ -740,3 +746,38 @@ func getSubnetID() (string, error) {
...
+	parsedARN, err := arn.Parse(taskARN)
+	if err != nil {
+		return "", errors.Wrapf(err, "task get-id: malformed taskARN: %s", taskARN)
+	}
+
+	// Get task resource section
+	resource := parsedARN.Resource
+
+	if !strings.Contains(resource, arnResourceDelimiter) {
+		return "", errors.New(fmt.Sprintf("task get-id: malformed task resource: %s", resource))
```
aaithal: you can just do `errors.Errorf()` here

agent/functional_tests/util/utils.go
```diff
@@ -740,3 +746,38 @@ func getSubnetID() (string, error) {
...
+	}
+
+	resourceSplit := strings.SplitN(resource, arnResourceDelimiter, arnResourceSections)
+	if len(resourceSplit) != arnResourceSections {
+		return "", errors.New(fmt.Sprintf("task get-id: invalid task resource split: %s, expected=%d, actual=%d", resource, arnResourceSections, len(resourceSplit)))
```
aaithal: you can just do `errors.Errorf()` here

agent/functional_tests/tests/functionaltests_unix_test.go
richardpen: It should also work, I'll make the change.

PR #1118, opened 2017-11-29
Bug - Fixed a bug where -version fails due to its dependency on docker client 

yunhee-l:
### Testing

New tests cover the changes: 

Making -version flag for agent independent of docker client, and removing docker version from output.
% docker run --rm --env-file ~/env.txt amazon/amazon-ecs-agent:make -version
Amazon ECS Agent:
	Version: 1.16.0
	Commit: c2836d6
    	Dirty: true

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 #Changelog
 
+## Unreleased
+* Bug - Fixed a bug where `-version` fails due to its dependency on docker client. [#1118](https://github.com/aws/amazon-ecs-agent/pull/1118)
+
 ## 1.16.0
 * Feature - Support pulling from Amazon ECR with specified IAM role in task definition
 * Feature - Enable support for task level CPU and memory constraints.
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -44,7 +44,6 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
 	"github.com/aws/amazon-ecs-agent/agent/tcs/handler"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
-	"github.com/aws/amazon-ecs-agent/agent/version"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	aws_credentials "github.com/aws/aws-sdk-go/aws/credentials"
@@ -71,8 +70,6 @@ var (
 // object. Its purpose is to mostly demonstrate how to interact with the
 // ecsAgent type.
 type agent interface {
-	// printVersion prints the Agent version string
-	printVersion() int
 	// printECSAttributes prints the Agent's capabilities based on
 	// its environment
 	printECSAttributes() int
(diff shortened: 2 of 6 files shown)


agent/version/formatting.go
```diff
@@ -13,7 +13,11 @@
...
 package version
 
-import "fmt"
+import (
+	"fmt"
+
+	"github.com/aws/amazon-ecs-agent/agent/sighandlers/exitcodes"
+)
 
 type Versioner interface {
```
jhaynes: Is there a reason to leave this interface around?

agent/app/agent.go
```diff
@@ -170,12 +167,6 @@ func newAgent(
 	}, nil
 }
 
-// printVersion prints the ECS Agent version string
-func (agent *ecsAgent) printVersion() int {
```
jhaynes: It looks like `agent/app/agent_integ_test.go` calls `printVersion`. There is also a windows mock that can probably go away here: `agent/app/agent_windows_test.go`.

yunhee-l: For the agent_integ_test, I'll replace the call to printVersion with printECSAttributes, now that printVersion doesn't actually depend on newAgent call. 
Removing the mock from agent_window_test.

agent/version/formatting.go
yunhee-l: Versioner interface isn't used anywhere regardless of this change, removing it.

aaithal:
@samuelkarp why would prefer to keep docker version being printed with Agent version? Can't you get it from other sources? I'm not sure if we really need to print docker version when printing agent version.

agent/version/formatting.go
```diff
@@ -26,7 +26,7 @@ type Versioner interface {
 //        Version: 0.0.3
 //        Commit: 4bdc7fc
 //        DockerVersion: 1.5.0
-func PrintVersion(extra ...Versioner) {
+func PrintVersion() int {
```
richardpen: Can you update the comments also.

nmeyerhans:
IMO the -version flag should get us the agent version and little, if anything, else.

We could consider adding a -info flag or similar that gives us additional info, similar to `docker info`. This -info flag could provide some additional information about the agent's runtime environment, including some Docker details.

samuelkarp:
@aaithal I'd like an easy way for us to ask customers to run one command that gives enough information for troubleshooting.  @nmeyerhans's suggestion of `-info` satisfies that, so I'd be okay removing Docker version if we implement something like `-info`.

aaithal:
makes sense. just noting that we don't usually ask customers to run the `-version` command in our trouble-shooting guide. let's get rid of Docker version from `-version` with a plan to add `-info` in the future.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 #Changelog
 
+## Unreleased
+* Bug - Fixed a bug where -version fails due to its dependency on docker client. [#1118](https://github.com/aws/amazon-ecs-agent/pull/1118)
```
aaithal: nit: `-version` would be better

yunhee-l: fixed

agent/version/formatting.go
yunhee-l: fixed

nmeyerhans:
LGTM. My only request is that the commit message be fleshed out a little bit. https://chris.beams.io/posts/git-commit/ has some good guidelines for how to structure the text.

PR #1125, opened 2017-12-02
Persist container exit code in agent state file

### LINKED ISSUE #986: Container exit code not persisted properly
### Summary
It looks like the container exit code may not be properly persisted in the agent's state file, leading to data loss  potentially impacting the call to submit state changes to ECS.

### Description
Container data is stored in an `api.Container` struct, which is then serialized to JSON and stored in a file.  During agent restarts, this file is read and the `api.Container` structs are rebuilt based on the serialized data.  With Go's default JSON marshaling library (`encoding/json`), only visible fields (public) are serialized.  In fee2053a4afb48884e35c8ce9cff882b1d683851, `KnownExitCode` was changed to `knownExitCode` which resulted in the field being a non-visible, private field.

### Expected Behavior
Exit code should be serialized and saved in the agent state file.

### Observed Behavior
Exit code is not present in the agent state file.

ofiliz:
### Summary

Container exit code needs to be persisted in the agent state file so
that api.Container structs can be rebuilt after agent restarts. Go's
default JSON encoder marshals only public fields. This PR renames
and exports that field, updates access functions and adds a test case
for exercising the new logic.

Future work: We should write a custom marshaler for container class
so that we don't have to expose internal state. Since this improvement
is not specific to exit code, I chose to fix it first, and implement the
custom marshaler in a later commit.

### Implementation details

* Rename and export KnownExitCodeUnsafe with proper JSON marking.
* Add test to ensure exit code is persisted and catch any regressions.
* Fix existing test panics caused by continuing after fatal nil checks.

### Testing

- [x] Confirmed the new test case fails before this commit, and succeeds after it.

Bug - Persist container exit code in agent state file

### Licensing

Fixes #986

### DIFF
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -135,7 +135,14 @@ type Container struct {
 	// metadata file
 	MetadataFileUpdated bool `json:"metadataFileUpdated"`
 
-	knownExitCode     *int
+	// KnownExitCodeUnsafe specifies the exit code for the container.
+	// It is exposed outside of the package so that it's marshalled/unmarshalled in
+	// the JSON body while saving the state.
+	// NOTE: Do not access KnownExitCodeUnsafe directly. Instead, use `GetKnownExitCode`
+	// and `SetKnownExitCode`.
+	KnownExitCodeUnsafe *int `json:"KnownExitCode"`
+
+	// KnownPortBindings is an array of port bindings for the container.
 	KnownPortBindings []PortBinding
 
 	// SteadyStateStatusUnsafe specifies the steady state status for the container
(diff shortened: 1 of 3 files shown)


aaithal:
I missed this in my comments. Can you please add a changelog entry as well? Example: https://github.com/aws/amazon-ecs-agent/commit/6527d7d324fabb1fa2de9651071b07faba587299#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR4

agent/api/container.go
```diff
@@ -135,7 +135,14 @@ type Container struct {
 	// metadata file
 	MetadataFileUpdated bool `json:"metadataFileUpdated"`
 
-	knownExitCode     *int
+	// KnownExitCodeUnsafe specifies the exit code for the container.
+	// It is exposed outside of the package so that it's marshalled/unmarshalled in
+	// the JSON body while saving the state.
+	// NOTE: Do not access KnownExitCodeUnsafe directly. Instead, use `GetKnownExitCode`
+	// and `SetKnownExitCode`.
+	KnownExitCodeUnsafe *int `json:"KnownExitCode"`
```
richardpen: You may also bump the state file version [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/statemanager/state_manager.go#L60)

Issue #1111, opened 2017-11-23
awslogs driver's multiline options not supported in ECS windows 


sharanyad:
### Summary
awslogs driver multiline options require remote api version 1.30. This version is currently not supported for windows.

### Description
Bug found while testing in https://github.com/aws/amazon-ecs-agent/pull/1101 
Agent would not return an attribute for remote API version 1.30 since it is not present [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/dockerclient/versionsupport_windows.go#L51)

### Expected Behavior
Tasks with awslogs driver multiline options `awslogs-multiline-pattern` and `awslogs-datetime-format` set should run in windows

### Observed Behavior
They do not

PR #1129, opened 2017-12-05
Adding support for Docker API version 1.30 in Windows

sharanyad:
### Summary
Currently, agent only supports Docker remote API version 1.24 in Windows. Extending this to support other versions such as 1.30.

### Testing

### Licensing

### DIFF
--- a/agent/engine/dockerclient/dockerclientfactory_windows_test.go
+++ b/agent/engine/dockerclient/dockerclientfactory_windows_test.go
@@ -53,6 +53,18 @@ func TestFindClientAPIVersion(t *testing.T) {
 	for _, version := range getAgentVersions() {
 		client, err := factory.GetClient(version)
 		assert.NoError(t, err)
-		assert.Equal(t, Version_1_24, factory.FindClientAPIVersion(client))
+		if isWindowsReplaceableVersion(version) {
+			version = minDockerAPIVersion
+		}
+		assert.Equal(t, version, factory.FindClientAPIVersion(client))
 	}
+}
+
+func isWindowsReplaceableVersion(version DockerVersion) bool {
+	for _, v := range getWindowsReplaceableVersions() {
+		if v == version {
+			return true
+		}
+	}
+	return false
 }
--- a/agent/engine/dockerclient/versionsupport_unix.go
+++ b/agent/engine/dockerclient/versionsupport_unix.go
@@ -40,6 +40,10 @@ func getAgentVersions() []DockerVersion {
 		Version_1_23,
 		Version_1_24,
 		Version_1_25,
+		Version_1_26,
+		Version_1_27,
+		Version_1_28,
+		Version_1_29,
 		Version_1_30,
 	}
 }
--- a/agent/engine/dockerclient/versionsupport_windows.go
+++ b/agent/engine/dockerclient/versionsupport_windows.go
@@ -46,9 +46,23 @@ func getWindowsReplaceableVersions() []DockerVersion {
 	}
 }
 
-// getAgentVersions for Windows should return all of the replaceable versions plus additional versions
+// getWindowsSupportedVersions returns the set of remote api versions that are
+// supported by agent in windows
+func getWindowsSupportedVersions() []DockerVersion {
+	return []DockerVersion{
+		Version_1_24,
+		Version_1_25,
+		Version_1_26,
+		Version_1_27,
+		Version_1_28,
+		Version_1_29,
+		Version_1_30,
+	}
+}
+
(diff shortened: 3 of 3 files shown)


agent/engine/dockerclient/versionsupport_windows.go
```diff
@@ -46,9 +46,18 @@ func getWindowsReplaceableVersions() []DockerVersion {
...
+	}
+}
+
+// getAgentVersions for Windows should return all of the replaceable versions plus supported versions
 func getAgentVersions() []DockerVersion {
-	return append(getWindowsReplaceableVersions(), minDockerAPIVersion)
+	return append(getWindowsReplaceableVersions(), getWindowsSupportedVersions()...)
 }
 
 // getDefaultVersion returns agent's default version of the Docker API
```
adnxn: do we keep the default version as `minDockerAPIVersion`?

sharanyad: yes, since we know that should at least be the minimum version supported in the instance.

sharanyad:
This PR addresses issue https://github.com/aws/amazon-ecs-agent/issues/1111.

agent/engine/dockerclient/versionsupport_windows.go
```diff
@@ -46,9 +46,18 @@ func getWindowsReplaceableVersions() []DockerVersion {
 	}
 }
 
-// getAgentVersions for Windows should return all of the replaceable versions plus additional versions
+// getWindowsSupportedVersions returns the set of remote api versions that are
+// supported by agent in windows
+func getWindowsSupportedVersions() []DockerVersion {
+	return []DockerVersion{
+		Version_1_24,
+		Version_1_30,
```
samuelkarp: Is there any reason we don't have the versions between 1.24 and 1.30?

sharanyad: We add versions here when those api versions' features are used in ECS. Hence, only 1.24 and 1.30 are added here.

samuelkarp: I haven't read the surrounding code recently enough to know whether there has been a change here I'm forgetting, but...

The initial intent of reporting API versions up to ECS was twofold: First, to enable capability-based placement logic such that we only place a task on an instance that is capable of implementing all the requested configuration options.  Second, it was to enable us to add additional configuration options to the task definition over time and not require updates in the ECS agent.  This second bit is accomplished through accepting a serialized `Config` and `HostConfig` from the ECS backend and applying it to our container creation calls through the Docker API.  We signal all of the available API versions (regardless of whether or not they're used) so that we have this flexibility going forward; we may make use of any of them in the future without requiring further changes to the agent.

sharanyad: Got it. If thats the case, we should probably add all of the versions between 1.25 and 1.30 [here](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/engine/dockerclient/versionsupport_unix.go#L32) for Linux too.

agent/engine/dockerclient/dockerclientfactory_windows_test.go
```diff
@@ -53,6 +53,18 @@ func TestFindClientAPIVersion(t *testing.T) {
 	for _, version := range getAgentVersions() {
```
petderek: I'd split TestFindClientAPIVersion into two different tests, since the assertions are different.

One that iterates over getWindowsReplaceableVersions(), and one that iterates over getWindowsSupportedVersions().

agent/engine/dockerclient/versionsupport_unix.go
```diff
@@ -40,6 +40,10 @@ func getAgentVersions() []DockerVersion {
 		Version_1_23,
 		Version_1_24,
 		Version_1_25,
+		Version_1_26,
```
richardpen: Did you tested that these capabilities are registered as expected, `com.amazonaws.ecs.capability.docker-remote-api.1.[26,27,28,29]` for the container instance attributes?

sharanyad: Yes, the capabilities for the api versions(if supported) are registered as expected.

agent/engine/dockerclient/dockerclientfactory_windows_test.go
sharanyad: This test gets clients for all the agent supported versions i.e `getAgentVersions()` (this now includes replaceable and supported versions), hence extended this for `getWindowsSupportedVersions()` too.

Issue #1096, opened 2017-11-20
Windows failing integ and functional tests


samuelkarp:
`TestImageWithSameNameAndDifferentID` is timing out with Docker 1.12 beta on Windows.

`TestTaskIamRolesDefaultNetworkMode` is failing with Docker 17.06.2-ee-5 on Windows. This looks to me like I need to adjust the timeout.

```
=== RUN   TestTaskIamRolesDefaultNetworkMode
Container InstanceArn: arn:aws:ecs:us-west-2:123456789012:container-instance/6a7416ad-33ca-416d-8c37-271085b8118e
--- FAIL: TestTaskIamRolesDefaultNetworkMode (166.00s)
        utils_windows.go:86: Registry location test_path_1511205926629440700
        utils_windows.go:96: datadir C:\work\amazon-ecs-agent\ftest_temp\ecs_integ_testdata429920973\data
        utils_windows.go:104: Created directory C:\work\amazon-ecs-agent\ftest_temp\ecs_integ_testdata429920973 to store test data in
        utils.go:149: Found agent metadata: {Cluster:ecs-functional-tests ContainerInstanceArn:0xc04244e140 Version:Amazon ECS Agent - v1.15.0 (f8c4efa)}
        utils.go:170: Task definition: ecsftest-iamrole-test-16b7aada7b892f7909cf0a8985aac8c2:1
        utils.go:190: Started task: arn:aws:ecs:us-west-2:123456789012:task/0b243a36-16b3-4969-a6f5-34d1e28f14e1
        functionaltests_windows_test.go:182: Waiting task to stop error : Timed out waiting for task to reach STOPPED: arn:aws:ecs:us-west-2:123456789012:task-definition/ecsftest-iamrole-test-16b7aada7b892f7909cf0a8985aac8c2:1, arn:aws:ecs:us-west-2:123456789012:task/0b243a36-16b3-4969-a6f5-34d1e28f14e1
        utils.go:157: Preserving test dir for failed test C:\work\amazon-ecs-agent\ftest_temp\ecs_integ_testdata429920973
```

`TestMetadataServiceValidator` is failing with Docker 17.06.2-ee-5 on Windows.  This looks like another timeout.
```
=== RUN   TestMetadataServiceValidator
Container InstanceArn: arn:aws:ecs:us-west-2:123456789012:container-instance/adf85770-8cce-493a-a5d6-e39006e47c4a
--- FAIL: TestMetadataServiceValidator (120.80s)
[...]

PR #1138, opened 2017-12-08
windows functional tests for awslogs multiline options

sharanyad:
### Summary
Functional tests that verify logs from CloudWatch events for 'awslogs-multiline-pattern' and 'awslogs-datetime-format' options. This is for Windows.

### Implementation details
Test for awslogs option 'awslogs-datetime-format' verifies that multiple log lines with a certain pattern of timestamp are sent to a single CloudWatch log event.
Test for awslogs option 'awslogs-multiline-pattern' verifies that multiple log lines with a certain pattern are sent to a single CloudWatch log event.

### Testing

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/awslogs-datetime-windows/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/awslogs-datetime-windows/task-definition.json
@@ -1,0 +1,21 @@
+{
+  "family": "ecs-awslogs-datetime-windows-test",
+  "containerDefinitions": [{
+    "essential": true,
+    "memory": 512,
+    "name": "awslogs-datetime-windows",
+    "cpu": 1024,
+    "image": "microsoft/windowsservercore:latest",
+    "entryPoint": ["powershell"],
+    "command": ["echo", "\"May 01, 2017 19:00:01 ECS\nMay 01, 2017 19:00:04 Agent\nRunning\nin the instance\""],
+    "logConfiguration": {
+        "logDriver": "awslogs",
+        "options": {
+            "awslogs-group":"ecs-functional-tests",
+            "awslogs-region":"$$$TEST_REGION$$$",
+            "awslogs-stream-prefix":"ecs-functional-tests",
+            "awslogs-datetime-format":"%b %d, %Y %H:%M:%S"
+        }
+    }
+  }]
+}
--- a/agent/functional_tests/tests/functionaltests_windows_test.go
+++ b/agent/functional_tests/tests/functionaltests_windows_test.go
@@ -28,6 +28,7 @@ import (
 	. "github.com/aws/amazon-ecs-agent/agent/functional_tests/util"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudwatch"
 	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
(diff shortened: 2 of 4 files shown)


agent/functional_tests/tests/functionaltests_windows_test.go
```diff
@@ -332,3 +333,141 @@ func TestOOMContainer(t *testing.T) {
...
+	assert.Len(t, resp.Events, 2, fmt.Sprintf("Got unexpected number of log events: %d", len(resp.Events)))
+	assert.Equal(t, *resp.Events[0].Message, "INFO: ECS Agent\nRunning\n", fmt.Sprintf("Got log events message unexpected: %s", *resp.Events[0].Message))
+	assert.Equal(t, *resp.Events[1].Message, "INFO: Instance\r\n", fmt.Sprintf("Got log events message unexpected: %s", *resp.Events[1].Message))
+}
+
+// TestAWSLogsDriverDatetimeFormat verifies that multiple log lines with a certain
+// pattern of timestamp, specified using 'awslogs-datetime-format' option, are sent
+// to a single CloudWatch log event
+func TestAWSLogsDriverDatetimeFormat(t *testing.T) {
+	RequireDockerAPIVersion(t, ">=1.30")
```
samuelkarp: Any reason not to just use `RequireDockerVersion` here?

agent/functional_tests/tests/functionaltests_windows_test.go
```diff
@@ -332,3 +333,141 @@ func TestOOMContainer(t *testing.T) {
...
+	logGroupExists := false
+	for i := 0; i < len(respDescribeLogGroups.LogGroups); i++ {
+		if *respDescribeLogGroups.LogGroups[i].LogGroupName == awslogsLogGroupName {
+			logGroupExists = true
+			break
+		}
+	}
+
+	if !logGroupExists {
+		_, err := cwlClient.CreateLogGroup(&cloudwatchlogs.CreateLogGroupInput{
```
aaithal: (I know you're using existing code as an example. This is an improvement we can make everywhere I think) You should be able to just invoke `CreateLogGroup` and ignore `ResourceAlreadyExistsException` error instead of the describe + create logic. Here's the [API documentation](http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_CreateLogGroup.html) for the same.

agent/functional_tests/tests/functionaltests_windows_test.go
sharanyad: ECS checks for awslogs multiline support from the API version, which makes sense to use in the functional tests too.
Since Docker versions does not comply with semver standards, it would be better to compare API versions here. What do you think?

agent/functional_tests/testdata/taskdefinitions/awslogs-datetime-windows/task-definition.json
```diff
@@ -0,0 +1,21 @@
+{
+  "family": "ecs-awslogs-datetime-windows-test",
+  "containerDefinitions": [{
+    "essential": true,
+    "memory": 10,
```
richardpen: I'd suggest you increase the memory to >256, as low memory could cause the test failure due to slow start/stop, related to this [issue](https://github.com/aws/amazon-ecs-agent/issues/1096). Same for the other task definition.

agent/functional_tests/tests/functionaltests_windows_test.go
```diff
@@ -332,3 +331,105 @@ func TestOOMContainer(t *testing.T) {
 	assert.NoError(t, err, "Expect task to be stopped")
 	assert.NotEqual(t, 0, testTask.Containers[0].ExitCode, "container should fail with memory error")
 }
+
+// TestAWSLogsDriverMultilinePattern verifies that multiple log lines with a certain
+// pattern, specified using 'awslogs-multiline-pattern' option, are sent to a single
+// CloudWatch log event
+func TestAWSLogsDriverMultilinePattern(t *testing.T) {
+	RequireDockerAPIVersion(t, ">=1.30")
+	cwlClient := cloudwatchlogs.New(session.New(), aws.NewConfig().WithRegion(*ECS.Config.Region))
```
richardpen: Can you make this consistent to use `aws.StringValue`?

PR #1140, opened 2017-12-12
WIP: Systemd Cgroup support

rickard-von-essen:
### Summary
When running docker under systemd docker requires that `--cgroup-parent` matches the systemd hierarchy. Otherwise it docker run will fail with the message:

```
cgroup-parent for systemd cgroup should be a valid slice named as "xxx.slice"
```

This now happens after this commit (2a3c4a46923353f3a963deb81b9ddb091f5e84c4) introduced in v1.16.0.

### Implementation details

This creates the root cgroup `ecs.slice` and runs each task in a subslice ecs-<task-id>.slice (with `-` in task-id replaced with `_`).

This also adds `ECS_CGROUP_PREFIX` to change the name of the parent slice, defaults to `ecs`.

An example cgroup hierarchy:

```
├─ecs.slice
│ └─ecs-fd6c1fe9_a5e6_425f_938c_457556f12050.slice
│   └─docker-a4d5069457bd9e0fccc45f132c1f1b78c911344787b25a7f60edde4e0e470b14.scope
│     └─16466 java -Xms512M -Xmx512M -jar server.jar server
```

For more information see [Red Hat Enterprise Linux 7 Documentation - 1.2. Default Cgroup Hierarchies](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies)

### Testing

New tests cover the changes: 

### Licensing

This contribution is under the terms of the Apache 2.0 License:  yes

### DIFF
--- a/agent/api/task_unix.go
+++ b/agent/api/task_unix.go
@@ -45,6 +45,7 @@ const (
 func (task *Task) adjustForPlatform(cfg *config.Config) {
 	task.memoryCPULimitsEnabledLock.Lock()
 	defer task.memoryCPULimitsEnabledLock.Unlock()
+	task.CgroupDriver = cfg.CgroupDriver
 	task.MemoryCPULimitsEnabled = cfg.TaskCPUMemLimit.Enabled()
 }
 
@@ -58,7 +59,7 @@ func (task *Task) BuildCgroupRoot() (string, error) {
 	if err != nil {
 		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id from task ARN: %s", task.Arn)
 	}
-	task.CgroupDriver = "systemd" // TODO FIX
+
 	if task.CgroupDriver == "systemd" {
 		// See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies
 		return fmt.Sprintf("%s-%s.slice", config.DefaultTaskCgroupPrefixSystemd, strings.Replace(taskID, "-", "_", -1)), nil
--- a/agent/api/task_unix.go
+++ b/agent/api/task_unix.go
@@ -17,6 +17,7 @@ package api
 
 import (
 	"fmt"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -45,21 +46,24 @@ func (task *Task) adjustForPlatform(cfg *config.Config) {
 	task.memoryCPULimitsEnabledLock.Lock()
 	defer task.memoryCPULimitsEnabledLock.Unlock()
 	task.MemoryCPULimitsEnabled = cfg.TaskCPUMemLimit.Enabled()
-	task.cgroupPrefix = cfg.CgroupPrefix
 }
 
 func getCanonicalPath(path string) string { return path }
 
 // BuildCgroupRoot helps build the task cgroup prefix
-// Example: /ecs/task-id
+// Cgroupfs example: /ecs/task-id
+// Systemd example: ecs.slice/ecs-task_id.slice
 func (task *Task) BuildCgroupRoot() (string, error) {
 	taskID, err := task.GetID()
 	if err != nil {
 		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id from task ARN: %s", task.Arn)
 	}
-
-	// See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies
-	return fmt.Sprintf("%s-%s.slice", task.cgroupPrefix, strings.Replace(taskID, "-", "_", -1)), nil
+	task.CgroupDriver = "systemd" // TODO FIX
+	if task.CgroupDriver == "systemd" {
+		// See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies
+		return fmt.Sprintf("%s-%s.slice", config.DefaultTaskCgroupPrefixSystemd, strings.Replace(taskID, "-", "_", -1)), nil
+	}
+	return filepath.Join(config.DefaultTaskCgroupPrefixCgroupFS, taskID), nil
 }
 
 // BuildLinuxResourceSpec returns a linuxResources object for the task cgroup
--- a/agent/api/task_unix.go
+++ b/agent/api/task_unix.go
@@ -16,9 +16,6 @@
 package api
 
 import (
-	"fmt"
-	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/config"
@@ -45,7 +42,6 @@ const (
 func (task *Task) adjustForPlatform(cfg *config.Config) {
 	task.memoryCPULimitsEnabledLock.Lock()
 	defer task.memoryCPULimitsEnabledLock.Unlock()
-	task.CgroupDriver = cfg.CgroupDriver
 	task.MemoryCPULimitsEnabled = cfg.TaskCPUMemLimit.Enabled()
 }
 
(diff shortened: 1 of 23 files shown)


rickard-von-essen:
Whit this change ECS runs on _RHEL/CentOS/Fedora Atomic_ using this unit file (adapted from https://github.com/aws/amazon-ecs-init/pull/109):

```
# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
#
# Licensed under the Apache License, Version 2.0 (the
# "License"). You may not use this file except in compliance
# with the License. A copy of the License is located at
#
#     http://aws.amazon.com/apache2.0/
#
# or in the "license" file accompanying this file. This file is
# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
# CONDITIONS OF ANY KIND, either express or implied. See the
# License for the specific language governing permissions and
#
# A systemd unit file to run `ecs-agent` in a Docker container. This service
# attempts to duplicate the functionality of the `ecs-init` Golang package. The
# notable differences currently are:
# 1. The unit file does an unconditional pull of the latest ECS Agent from
# Docker Hub at startup.
# 2. The unit file does not currently handle the Agent's self-upgrade
# functionality.

[Unit]
Description=Amazon ECS Agent
Documentation=http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ECS_agent.html
Requires=docker.service
After=docker.service

[Service]
Type=simple
ExecStartPre=/usr/bin/docker pull amazon/amazon-ecs-agent:v1.15.2
ExecStartPre=/sbin/iptables -t nat -A PREROUTING -d 169.254.170.2/32 \
 -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:51679
ExecStartPre=/sbin/iptables -t nat -A OUTPUT -d 169.254.170.2/32 \
 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679
ExecStartPre=/sbin/sysctl -w net.ipv4.conf.all.route_localnet=1
ExecStart=/usr/bin/docker run --privileged \
  --restart=on-failure:10 \
  --volume=/var/run:/var/run \
  --volume=/var/log/ecs/:/log \
  --volume=/var/lib/ecs/data:/data \
  --volume=/etc/ecs:/etc/ecs \
  --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro \
  --volume=/run/docker/execdriver/native:/var/lib/docker/execdriver/native:ro \
  --publish=127.0.0.1:51678:51678 \
  --net=host \
  --env ECS_SELINUX_CAPABLE=true \
  --env ECS_LOGFILE=/log/ecs-agent.log \
  --env ECS_DATADIR=/data \
  --env ECS_UPDATES_ENABLED=false \
  --env ECS_ENABLE_TASK_IAM_ROLE=true \
  --env ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST=true \
  --env ECS_CGROUP_PREFIX=ecs \
  --env-file=/etc/ecs/ecs.config \
  amazon/amazon-ecs-agent:make
ExecStop=/usr/bin/docker stop ecs-agent
ExecStopPost=-/sbin/iptables -t nat -D PREROUTING -d 169.254.170.2/32 \
 -p tcp -m tcp --dport 80 -j DNAT --to-destination 127.0.0.1:51679
ExecStopPost=-/sbin/iptables -t nat -D OUTPUT -d 169.254.170.2/32 \
 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679
ExecStopPost=/bin/sh -c \
 "/sbin/sysctl -w net.ipv4.conf.all.route_localnet=$(/sbin/sysctl -q -n net.ipv4.conf.default.route_localnet)"
#Restart=on-failure
RestartPreventExitStatus=5
```

petderek:
@rickard-von-essen Thanks for submitting this! 

Please confirm this contribution is under the terms of the Apache 2.0 license.

rickard-von-essen:
I hereby confirm this contribution is under the terms of the Apache 2.0 license.

rickard-von-essen:
I didn't find any documentation on how to update the mocks. When I tried it resulted in alot of updated mocks.

Could someone give me the commands that you use to update them?

adnxn:
> I hereby confirm this contribution is under the terms of the Apache 2.0 license.

Do you mind adding your confirmation to the PR description under the licensing heading? It's part of the PR template.

> Could someone give me the commands that you use to update them?

Running the `make gogenerate` should generate the mocks that you need. Although you may need to also get `mockgen` if you don't already have that. 

Otherwise, we're evaluating how this fits in with our init story since Amazon Linux does not use systemd.

agent/api/task_unix.go
```diff
@@ -56,7 +58,8 @@ func (task *Task) BuildCgroupRoot() (string, error) {
 		return "", errors.Wrapf(err, "task build cgroup root: unable to get task-id from task ARN: %s", task.Arn)
 	}
 
-	return filepath.Join(config.DefaultTaskCgroupPrefix, taskID), nil
+	// See https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies
+	return fmt.Sprintf("%s-%s.slice", task.cgroupPrefix, strings.Replace(taskID, "-", "_", -1)), nil
```
samuelkarp: I'm concerned about this change since you're unconditionally using the systemd slice.  Non-systemd distributions like Amazon Linux shouldn't use this format, so I think we'd need to turn this into conditional behavior.

rickard-von-essen: Yes maybe it should. But then it should probably auto detect if it's running under systemd. I didn't implement it in the because it should work even with non systemd distros and I saw that Amazon Linux 2 is using systemd.

I can continue this after the holidays, but if this is i prior for Amazon Linux 2 feel free to take this over.

nmeyerhans: @rickard-von-essen Which distros are enabling Docker's systemd cgroups support by default? Docker's official packaging for [RPM based](https://github.com/docker/docker-ce/blob/master/components/packaging/rpm/systemd/docker.service) and [Debian based](https://github.com/docker/docker-ce/blob/master/components/packaging/deb/systemd/docker.service) distros does not, and comments in the unit files suggest that there are issues with doing so. They don't provide any additional details about the issues, though. Given this, it's unclear that we'll enable the systemd cgroups driver when we ship our Amazon Linux 2 ECS Optimized AMI.

In any case, the [/info](https://docs.docker.com/engine/api/v1.33/#operation/SystemInfo) Docker api provides the cgroup driver details in the `CgroupDriver` key, and we should be able to use that properly autodetect the appropriate configuration. Feel free to update this PR when convenient, or we'll get to this when we can.

rickard-von-essen: > Which distros are enabling Docker's systemd cgroups support by default? 

At least RHEL/CentOS/Fedora Atomic, and expect that the normal server flavor does it too.

> /info Docker api provides the cgroup driver details in the CgroupDriver

Excellent, I'll look into enabling this automatically.

rickard-von-essen:
> Which distros are enabling Docker's systemd cgroups support by default?

Fedora:
https://src.fedoraproject.org/rpms/docker/blob/master/f/docker.service#_20

CentOS/RHEL:
https://git.centos.org/blob/rpms!docker.git/c7-alt/SOURCES!docker.service#L23
This is on the _c7-alt_ branch, but at least that matches what you get with _Atomic_.

adnxn:
@rickard-von-essen: are you still interested in getting this merged? i'm changing the base branch to `dev` since that is what we merge against. additionally, tests are failing and there are merge conflicts with `dev`, we need to get those fixed before reviewing further. please let us know though.


@aws/aws-ecs-agent: fwiw we should open an issue to track systemd cgroup driver support.

rickard-von-essen:
Yes, I'm still planning on finishing this, but currently I'm on paternity leave so it might take a while before I get back to it.

I think everything is working now but the tests needs to be fixed and a rebase. But I have to double check.

adnxn:
> Yes, I'm still planning on finishing this, but currently I'm on paternity leave so it might take a while before I get back to it.

Okay that sounds great! I'm closing this for now, but please open a new PR when you're ready.

Also - Please make sure to clean up the WIP commits in your new PR. Thanks, and again, we appreciate the PR. Looking forward to getting this reviewed.

phekno:
This PR was closed, but was any of it actually merged?

rickard-von-essen:
See above. (No, I never finished it)

Issue #534, opened 2016-09-17
Feature Request: Support for Docker Health Checks when bumping a task definition revision


CpuID:
Currently ECS utilises ELB/ALB health checks to verify when a task is ready to accept traffic, and also when it is safe to terminate additional tasks as part of a rolling replacement/upgrade when bumping a task definition revision (to align with the service deployment configuration parameters).

Would it be possible when an ELB is not in use for an ECS service, to also look at the Docker health check status? There are some scenarios when you may not want an ELB in use, but need to gracefully rolling replace the containers as part of an upgrade.

Details on the feature introduced in Docker 1.12.x:

https://docs.docker.com/engine/reference/builder/#/healthcheck
https://docs.docker.com/engine/reference/run/#/healthcheck

CpuID:
@samuelkarp any feedback re this request?

samuelkarp:
@CpuID Thanks for the feature request!  I'm interested in as much detail as you (or anyone else also interested in this) can provide, as that helps us both prioritize and make sure we're actually addressing the use-cases appropriately.

Some thoughts on what you've provided already:
* Docker does not implement a health check policy yet (it was rejected in their initial discussions), but defining policies (what to do when the health check fails) is required for the behavior that you're looking for.  Are the types of settings that ELB/ALB health checks already support (timeout, interval, healthy threshold, unhealthy threshold) sufficient, or are you looking for something different?
* You're asking about services here, but ECS also supports one-off tasks.  Would you also want a health check to be performed/enforced for those tasks?
* For a service that does have a load balancer, would you want both the load balancer health check and the Docker health check to be enforced?  If you want both and one health check fails but the other succeeds, how important is it to know which health check failed?

CpuID:
@samuelkarp thanks for the response.

Replies inline below:

> Docker does not implement a health check policy yet (it was rejected in their initial discussions), but defining policies (what to do when the health check fails) is required for the behavior that you're looking for. Are the types of settings that ELB/ALB health checks already support (timeout, interval, healthy threshold, unhealthy threshold) sufficient, or are you looking for something different?

I think those thresholds that exist for an ALB/ELB currently are perfectly sufficient for most requirements for now.

> You're asking about services here, but ECS also supports one-off tasks. Would you also want a health check to be performed/enforced for those tasks?

[...]

PR #1141, opened 2017-12-12
Enable container health check

richardpen:
### Summary

Enable docker container health check in the agent and report container health metrics to backend

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -48,8 +48,6 @@ type DockerConfig struct {
 	HostConfig *string `json:"hostConfig"`
 	// Version specifies the docker client API version to use
 	Version *string `json:"version"`
-	// HealthCheck is the configuration of docker health check
-	HealthCheck *string `json:"healthCheck,omitempty"`
 }
 
 // HealthStatus contains the health check result returned by docker
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -32,15 +32,19 @@ const (
 	// that specifies that the log driver should be authenticated using the
 	// execution role
 	awslogsAuthExecutionRole = "ExecutionRole"
+
+	// dockerHealthCheckType is the type of container health check provided by docker
+	dockerHealthCheckType = "docker"
 )
 
 // DockerConfig represents additional metadata about a container to run. It's
 // remodeled from the `ecsacs` api model file. Eventually it should not exist
 // once this remodeling is refactored out.
 type DockerConfig struct {
-	Config     *string `json:"config"`
-	HostConfig *string `json:"hostConfig"`
-	Version    *string `json:"version"`
+	Config      *string `json:"config"`
+	HostConfig  *string `json:"hostConfig"`
+	Version     *string `json:"version"`
+	HealthCheck *string `json:"healthCheck,omitempty"`
 }
 
 // Container is the internal representation of a container in the ECS agent
@@ -65,11 +69,14 @@ type Container struct {
 	Overrides              ContainerOverrides          `json:"overrides"`
 	DockerConfig           DockerConfig                `json:"dockerConfig"`
 	RegistryAuthentication *RegistryAuthenticationData `json:"registryAuthentication"`
-
+	HealthCheckType        string                      `json:"healthCheckType,omitempty"`
+	// HealthStatus is the status of container health check
+	HealthStatus ContainerHealthStatus `json:"healthStatus,omitempty"`
+	// LastHealthKnownTime is the time when the container health status changed
+	LastHealthKnownTime time.Time `json:"lastUpdatedTime,omitempty"`
 	// LogsAuthStrategy specifies how the logs driver for the container will be
 	// authenticated
 	LogsAuthStrategy string
-
 	// lock is used for fields that are accessed and updated concurrently
 	lock sync.RWMutex
 
@@ -457,4 +464,10 @@ func (c *Container) GetLabels() map[string]string {
 	defer c.lock.RUnlock()
 
 	return c.labels
+}
+
+// HealthCheckShouldBeReported returns true if the health check is defined in
+// the task definition
+func (c *Container) HealthCheckShouldBeReported() bool {
+	return c.HealthCheckType == dockerHealthCheckType
 }
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
+	"github.com/aws/aws-sdk-go/aws"
 )
 
 const (
@@ -47,6 +48,18 @@ type DockerConfig struct {
 	HealthCheck *string `json:"healthCheck,omitempty"`
 }
 
+// HealthStatus contains the health check result returned by docker
+type HealthStatus struct {
+	// Status is the container health status
+	Status ContainerHealthStatus `json:"status,omitempty"`
+	// Since is the timestamp when container health status changed
+	Since *time.Time `json:"statusSince,omitempty"`
+	// ExitCode is the exitcode of health check if failed
+	ExitCode int `json:"exitCode,omitempty"`
+	// Output is the output of health check
+	Output string `json:"output,omitempty"`
+}
+
 // Container is the internal representation of a container in the ECS agent
 type Container struct {
 	// Name is the name of the container specified in the task definition
@@ -70,10 +83,8 @@ type Container struct {
 	DockerConfig           DockerConfig                `json:"dockerConfig"`
 	RegistryAuthentication *RegistryAuthenticationData `json:"registryAuthentication"`
 	HealthCheckType        string                      `json:"healthCheckType,omitempty"`
-	// HealthStatus is the status of container health check
-	HealthStatus ContainerHealthStatus `json:"healthStatus,omitempty"`
-	// LastHealthKnownTime is the time when the container health status changed
-	LastHealthKnownTime time.Time `json:"lastUpdatedTime,omitempty"`
+	// Health contains the health check information of
+	Health HealthStatus `json:"health,omitempty"`
 	// LogsAuthStrategy specifies how the logs driver for the container will be
 	// authenticated
 	LogsAuthStrategy string
@@ -470,4 +481,38 @@ func (c *Container) GetLabels() map[string]string {
 // the task definition
 func (c *Container) HealthCheckShouldBeReported() bool {
 	return c.HealthCheckType == dockerHealthCheckType
+}
+
+// SetHealthStatus sets the container health status
+func (c *Container) SetHealthStatus(health HealthStatus) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.Health.Status == health.Status {
+		return
+	}
+
+	c.Health.Status = health.Status
+	c.Health.Since = aws.Time(time.Now())
+	c.Health.Output = health.Output
+
+	// Set the health exit code if the health check failed
+	if c.Health.Status == ContainerUnhealthy {
+		c.Health.ExitCode = health.ExitCode
+	}
+}
+
+// GetHealthStatus returns the container health information
+func (c *Container) GetHealthStatus() HealthStatus {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	// Copy the pointer to avoid race condition
+	copyHealth := c.Health
+
+	if c.Health.Since != nil {
+		copyHealth.Since = aws.Time(*c.Health.Since)
+	}
+
+	return copyHealth
 }
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -42,9 +42,13 @@ const (
 // remodeled from the `ecsacs` api model file. Eventually it should not exist
 // once this remodeling is refactored out.
 type DockerConfig struct {
-	Config      *string `json:"config"`
-	HostConfig  *string `json:"hostConfig"`
-	Version     *string `json:"version"`
+	// Config is the configuration used to create container
+	Config *string `json:"config"`
+	// HostConfig is the configuration of container related to host resource
+	HostConfig *string `json:"hostConfig"`
+	// Version specifies the docker client API version to use
+	Version *string `json:"version"`
+	// HealthCheck is the configuration of docker health check
 	HealthCheck *string `json:"healthCheck,omitempty"`
 }
 
@@ -68,23 +72,37 @@ type Container struct {
 	Image string
 	// ImageID is the local ID of the image used in the container
 	ImageID string
-
-	Command                []string
-	CPU                    uint `json:"Cpu"`
-	Memory                 uint
-	Links                  []string
-	VolumesFrom            []VolumeFrom  `json:"volumesFrom"`
-	MountPoints            []MountPoint  `json:"mountPoints"`
-	Ports                  []PortBinding `json:"portMappings"`
-	Essential              bool
-	EntryPoint             *[]string
-	Environment            map[string]string           `json:"environment"`
-	Overrides              ContainerOverrides          `json:"overrides"`
-	DockerConfig           DockerConfig                `json:"dockerConfig"`
+	// Command is the command to run in the container which is specified in the task definition
+	Command []string
+	// CPU is the cpu limitation of the container which is specified in the task definition
+	CPU uint `json:"Cpu"`
+	// Memory is the memory limitation of the container which is specified in the task definition
+	Memory uint
+	// Links contains a list of containers to link, corresponding to docker option: --link
+	Links []string
+	// VolumesFrom contains a list of container's volume to use, corresponding to docker option: --volumes-from
+	VolumesFrom []VolumeFrom `json:"volumesFrom"`
+	// MountPoints contains a list of volume mount paths
+	MountPoints []MountPoint `json:"mountPoints"`
+	// Ports contains a list of ports binding configuration
+	Ports []PortBinding `json:"portMappings"`
+	// Essential denotes whether the container is essential or not
+	Essential bool
+	// EntryPoint is entrypoint of the container, corresponding to docker option: --entrypoint
+	EntryPoint *[]string
+	// Environment is the environment variable set in the container
+	Environment map[string]string `json:"environment"`
+	// Overrides contains the configuration to override of a container
+	Overrides ContainerOverrides `json:"overrides"`
+	// DockerConfig is the configuration used to create the container
+	DockerConfig DockerConfig `json:"dockerConfig"`
+	// RegistryAuthentication is the auth data used to pull image
 	RegistryAuthentication *RegistryAuthenticationData `json:"registryAuthentication"`
-	HealthCheckType        string                      `json:"healthCheckType,omitempty"`
-	// Health contains the health check information of
-	Health HealthStatus `json:"health,omitempty"`
+	// HealthCheckType is the mechnism to use for the container health check
+	// currently it only supports 'DOCKER'
+	HealthCheckType string `json:"healthCheckType,omitempty"`
+	// Health contains the health check information of container health check
+	Health HealthStatus `json:"-"`
 	// LogsAuthStrategy specifies how the logs driver for the container will be
 	// authenticated
 	LogsAuthStrategy string
@@ -477,9 +495,9 @@ func (c *Container) GetLabels() map[string]string {
 	return c.labels
 }
 
-// HealthCheckShouldBeReported returns true if the health check is defined in
+// HealthStatusShouldBeReported returns true if the health check is defined in
 // the task definition
-func (c *Container) HealthCheckShouldBeReported() bool {
+func (c *Container) HealthStatusShouldBeReported() bool {
 	return c.HealthCheckType == dockerHealthCheckType
 }
 
@@ -511,7 +529,7 @@ func (c *Container) GetHealthStatus() HealthStatus {
 	copyHealth := c.Health
 
 	if c.Health.Since != nil {
-		copyHealth.Since = aws.Time(*c.Health.Since)
+		copyHealth.Since = aws.Time(aws.TimeValue(c.Health.Since))
 	}
 
 	return copyHealth
--- a/agent/api/container_test.go
+++ b/agent/api/container_test.go
@@ -168,3 +168,36 @@ func TestSetupExecutionRoleFlag(t *testing.T) {
 		})
 	}
 }
+
+func TestSetHealtStatus(t *testing.T) {
+	container := Container{}
+
+	// set the container status to be healthy
+	container.SetHealthStatus(HealthStatus{Status: ContainerHealthy, Output: "test"})
+	health := container.GetHealthStatus()
+	assert.Equal(t, health.Status, ContainerHealthy)
+	assert.Equal(t, health.Output, "test")
+	assert.NotEmpty(t, health.Since)
+
+	// set the health status again shouldn't update the timestamp
+	container.SetHealthStatus(HealthStatus{Status: ContainerHealthy})
+	health2 := container.GetHealthStatus()
+	assert.Equal(t, health2.Status, ContainerHealthy)
+	assert.Equal(t, health2.Since, health.Since)
+
+	// change the container health status
+	container.SetHealthStatus(HealthStatus{Status: ContainerUnhealthy, ExitCode: 1})
+	health3 := container.GetHealthStatus()
+	assert.Equal(t, health3.Status, ContainerUnhealthy)
+	assert.Equal(t, health3.ExitCode, 1)
+	assert.NotEqual(t, health3.Since, health2.Since)
+}
+
+func TestHealthCheckShouldBeReported(t *testing.T) {
+	container := Container{}
+	assert.False(t, container.HealthCheckShouldBeReported(), "Health status of container that does not have HealthCheckType set should not be reported")
+	container.HealthCheckType = dockerHealthCheckType
+	assert.True(t, container.HealthCheckShouldBeReported(), "Health status of container that has docker HealthCheckType set should be reported")
+	container.HealthCheckType = "unknown"
+	assert.False(t, container.HealthCheckShouldBeReported(), "Health status of container that has non-docker HealthCheckType set should not be reported")
+}
--- a/agent/api/container_test.go
+++ b/agent/api/container_test.go
@@ -16,6 +16,7 @@ package api
 import (
 	"fmt"
 	"testing"
+	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/fsouza/go-dockerclient"
@@ -185,6 +186,9 @@ func TestSetHealtStatus(t *testing.T) {
 	assert.Equal(t, health2.Status, ContainerHealthy)
 	assert.Equal(t, health2.Since, health.Since)
 
+	// the sleep is to ensure the different of the two timestamp returned by time.Now()
+	// is big enough to pass asser.NotEqual
+	time.Sleep(10 * time.Millisecond)
 	// change the container health status
 	container.SetHealthStatus(HealthStatus{Status: ContainerUnhealthy, ExitCode: 1})
 	health3 := container.GetHealthStatus()
--- a/agent/api/container_test.go
+++ b/agent/api/container_test.go
@@ -193,11 +193,11 @@ func TestSetHealtStatus(t *testing.T) {
 	assert.NotEqual(t, health3.Since, health2.Since)
 }
 
-func TestHealthCheckShouldBeReported(t *testing.T) {
+func TestHealthStatusShouldBeReported(t *testing.T) {
 	container := Container{}
-	assert.False(t, container.HealthCheckShouldBeReported(), "Health status of container that does not have HealthCheckType set should not be reported")
+	assert.False(t, container.HealthStatusShouldBeReported(), "Health status of container that does not have HealthCheckType set should not be reported")
 	container.HealthCheckType = dockerHealthCheckType
-	assert.True(t, container.HealthCheckShouldBeReported(), "Health status of container that has docker HealthCheckType set should be reported")
+	assert.True(t, container.HealthStatusShouldBeReported(), "Health status of container that has docker HealthCheckType set should be reported")
 	container.HealthCheckType = "unknown"
-	assert.False(t, container.HealthCheckShouldBeReported(), "Health status of container that has non-docker HealthCheckType set should not be reported")
+	assert.False(t, container.HealthStatusShouldBeReported(), "Health status of container that has non-docker HealthCheckType set should not be reported")
 }
--- a/agent/api/containerstatus.go
+++ b/agent/api/containerstatus.go
@@ -43,9 +43,9 @@ const (
 const (
 	// ContainerHealthUnknown is the initial status of container health
 	ContainerHealthUnknown ContainerHealthStatus = iota
-	// ContainerHealthy represents the container health check returned healthy
+	// ContainerHealthy represents the status of container health check when returned healthy
 	ContainerHealthy
-	// ContainerUnhealthy represents the container health check failed
+	// ContainerUnhealthy represents the status of container health check when returned unhealthy
 	ContainerUnhealthy
 )
 
@@ -150,7 +150,7 @@ func (cs *ContainerStatus) UnmarshalJSON(b []byte) error {
 	}
 	if b[0] != '"' || b[len(b)-1] != '"' {
 		*cs = ContainerStatusNone
-		return errors.New("ContainerStatus must be a string or null; Got " + string(b))
+		return errors.New("container status unmarshal: status must be a string or null; Got " + string(b))
 	}
 	strStatus := string(b[1 : len(b)-1])
 	// 'UNKNOWN' and 'DEAD' for Compatibility with v1.0.0 state files
@@ -166,7 +166,7 @@ func (cs *ContainerStatus) UnmarshalJSON(b []byte) error {
 	stat, ok := containerStatusMap[strStatus]
 	if !ok {
 		*cs = ContainerStatusNone
-		return errors.New("Unrecognized ContainerStatus")
+		return errors.New("container status unmarshal: unrecognized status")
 	}
 	*cs = stat
 	return nil
@@ -182,31 +182,26 @@ func (cs *ContainerStatus) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON overrides the logic for parsing the JSON-encoded container health data
 func (healthStatus *ContainerHealthStatus) UnmarshalJSON(b []byte) error {
+	*healthStatus = ContainerHealthUnknown
+
 	if strings.ToLower(string(b)) == "null" {
-		*healthStatus = ContainerHealthUnknown
 		return nil
 	}
-
 	if b[0] != '"' || b[len(b)-1] != '"' {
-		*healthStatus = ContainerHealthUnknown
-		return errors.New("ContainerHealthStatus must be a string or null; Got " + string(b))
+		return errors.New("container health status unmarshal: status must be a string or null; Got " + string(b))
 	}
+
 	strStatus := string(b[1 : len(b)-1])
-	if strStatus == "UNKNOWN" {
-		*healthStatus = ContainerHealthUnknown
-		return nil
-	}
-	if strStatus == "HEALTHY" {
+	switch strStatus {
+	case "UNKNOWN":
+	case "HEALTHY":
 		*healthStatus = ContainerHealthy
-		return nil
-	}
-
-	if strStatus == "UNHEALTHY" {
+	case "UNHEALTHY":
 		*healthStatus = ContainerUnhealthy
-		return nil
+	default:
+		return errors.New("container health status unmarshal: unrecognized status: " + string(b))
 	}
-
-	return errors.New("Unrecognized container health status: " + string(b))
+	return nil
 }
 
 // MarshalJSON overrides the logic for JSON-encoding the ContainerHealthStatus type
--- a/agent/api/containerstatus.go
+++ b/agent/api/containerstatus.go
@@ -194,6 +194,7 @@ func (healthStatus *ContainerHealthStatus) UnmarshalJSON(b []byte) error {
 	strStatus := string(b[1 : len(b)-1])
 	switch strStatus {
 	case "UNKNOWN":
+	// The health status is already set to ContainerHealthUnknown initially
 	case "HEALTHY":
 		*healthStatus = ContainerHealthy
 	case "UNHEALTHY":
--- a/agent/api/containerstatus.go
+++ b/agent/api/containerstatus.go
@@ -40,8 +40,20 @@ const (
 	ContainerZombie
 )
 
+const (
+	// ContainerHealthUnknown is the initial status of container health
+	ContainerHealthUnknown ContainerHealthStatus = iota
+	// ContainerHealthy represents the container health check returned healthy
+	ContainerHealthy
+	// ContainerUnhealthy represents the container health check failed
+	ContainerUnhealthy
+)
+
 // ContainerStatus is an enumeration of valid states in the container lifecycle
 type ContainerStatus int32
+
+// ContainerHealthStatus is an enumeration of container health check status
+type ContainerHealthStatus int32
 
 var containerStatusMap = map[string]ContainerStatus{
 	"NONE":                  ContainerStatusNone,
@@ -50,6 +62,23 @@ var containerStatusMap = map[string]ContainerStatus{
 	"RUNNING":               ContainerRunning,
 	"RESOURCES_PROVISIONED": ContainerResourcesProvisioned,
 	"STOPPED":               ContainerStopped,
+}
+
+// BackendStatus returns the container health status recognized by backend
+func (healthStatus ContainerHealthStatus) BackendStatus() string {
+	switch healthStatus {
+	case ContainerHealthy:
+		return "HEALTHY"
+	case ContainerUnhealthy:
+		return "UNHEALTHY"
+	default:
+		return "UNKNOWN"
+	}
+}
+
+// String returns the readable description of the container health status
+func (healthStatus ContainerHealthStatus) String() string {
+	return healthStatus.BackendStatus()
 }
 
 // String returns a human readable string representation of this object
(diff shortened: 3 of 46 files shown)


agent/app/agent.go
```diff
@@ -540,10 +540,8 @@ func (agent *ecsAgent) startAsyncRoutines(
 		StatsEngine:                   statsEngine,
 	}
 
-	if !agent.cfg.DisableMetrics {
-		// Start metrics session in a go routine
-		go tcshandler.StartMetricsSession(telemetrySessionParams)
-	}
+	// Start metrics session in a go routine
+	go tcshandler.StartMetricsSession(telemetrySessionParams)
```
aaithal: This is problematic because today, disabling metrics also means disabling the metrics collection engine that gets initialized (stats engine's init) from tcs handler. Customers might choose to do that because they want lighter cpu/memory footprint for the  ECS Agent. If `agent.cfg.DisableMetrics` is `false`, we should disable both metrics collection and reporting. I think with this change, we're only disabling the reporting pat. We should also disable metrics collection

agent/engine/docker_task_engine.go
```diff
@@ -489,28 +489,38 @@ func (engine *DockerTaskEngine) handleDockerEvents(ctx context.Context) {
...
+	}
+
 	engine.processTasks.RLock()
 	managedTask, ok := engine.managedTasks[task.Arn]
 	// hold the lock until the message is sent so we don't send on a closed channel
 	defer engine.processTasks.RUnlock()
 	if !ok {
-		log.Crit("Could not find managed task corresponding to a docker event", "event", event, "task", task)
-		return true
+		seelog.Criticalf("Could not find managed task corresponding to a docker event, event: %s, task: %s", event, task)
```
aaithal: I don't think we need the whole task here (or in other log lines you've changed). Can you use `task.Arn` instead?

agent/engine/types.go
```diff
@@ -63,6 +65,8 @@ type DockerContainerMetadata struct {
 	StartedAt time.Time
 	// FinishedAt is the timestamp of container stop
 	FinishedAt time.Time
+	//Health contains the result of a container health check
```
aaithal: lint: `// Health`

agent/engine/docker_container_engine.go
```diff
@@ -802,6 +822,10 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 				// mean the container dies (non-init processes). If the container also
 				// dies, you see a "die" status as well; we'll update suitably there
 				continue
+			case "health_status: healthy":
```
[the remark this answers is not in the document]
richardpen: yes, the event's status is `"health_status: healthy"`

agent/tcs/client/client.go
```diff
@@ -241,3 +327,20 @@ func copyTaskMetrics(from []*ecstcs.TaskMetric) []*ecstcs.TaskMetric {
 	copy(to, from)
 	return to
 }
+
+// copyHealthMetadata performs a deep copy of HealthMetadata object
+func copyHealthMetadata(metadata *ecstcs.HealthMetadata, fin bool) *ecstcs.HealthMetadata {
+	return &ecstcs.HealthMetadata{
+		Cluster:           aws.String(*metadata.Cluster),
```
aaithal: I think you can just do `Cluster: metadata.Cluster` here (same for other string fields in this block.

agent/tcs/client/client.go
```diff
@@ -221,6 +228,85 @@ func (cs *clientServer) metricsToPublishMetricRequests() ([]*ecstcs.PublishMetri
...
+
+	var requests []*ecstcs.PublishHealthRequest
+	var taskHealths []*ecstcs.TaskHealth
+	numOfTasks := len(taskHealthMetrics)
+	for i, taskHealth := range taskHealthMetrics {
+		taskHealths = append(taskHealths, taskHealth)
+		// create a request if the number of task reaches the maximum
+		if (i+1)%tasksInMessage == 0 {
+			requestMetadata := copyHealthMetadata(metadata, (i+1) == numOfTasks)
+			requests = append(requests, ecstcs.NewPublishHealthMetricsRequest(requestMetadata, copyTaskHealthMetrics(taskHealths)))
```
aaithal: please split this into multiple lines so that it's easier to read

agent/tcs/client/client.go
```diff
@@ -221,6 +228,85 @@ func (cs *clientServer) metricsToPublishMetricRequests() ([]*ecstcs.PublishMetri
...
+		seelog.Debug("No container health metrics to report")
+		return nil, nil
+	}
+
+	var requests []*ecstcs.PublishHealthRequest
+	var taskHealths []*ecstcs.TaskHealth
+	numOfTasks := len(taskHealthMetrics)
+	for i, taskHealth := range taskHealthMetrics {
+		taskHealths = append(taskHealths, taskHealth)
+		// create a request if the number of task reaches the maximum
```
aaithal: minor: ".. reaches the maximum page size"

agent/tcs/client/client.go
richardpen: No, `metadata.xx` is a pointer, but we want a deep copy here, so that the actual value wasn't changed elsewhere.

aaithal: sorry, missed that. please use aws.StringValue here as well. Avoid doing pointer dereferencing as much as possible.

agent/stats/engine.go
```diff
@@ -296,13 +318,109 @@ func (engine *DockerStatsEngine) GetInstanceMetrics() (*ecstcs.MetricsMetadata,
...
+	for taskARN := range engine.tasksToHealthCheckContainers {
+		taskHealth := engine.getTaskHealthUnsafe(taskARN)
+		if taskHealth == nil {
+			continue
+		}
+		taskHealths = append(taskHealths, taskHealth)
+	}
+
+	if len(taskHealths) == 0 {
+		return nil, nil, EmptyMetricsError
```
aaithal: please return a different error here. `EmptyMetricsError` is not the same as `EmptyHealthStatusError` for example

agent/stats/engine.go
```diff
@@ -445,6 +563,17 @@ func (engine *DockerStatsEngine) doRemoveContainerUnsafe(container *StatsContain
 		delete(engine.tasksToDefinitions, taskArn)
 		seelog.Debugf("Deleted task from tasks, arn: %s", taskArn)
 	}
+
+	// Remove the container from health container watch list
+	if _, ok := engine.tasksToHealthCheckContainers[taskArn][dockerID]; !ok {
```
aaithal: I don't think you need this. `delete` is a no-op for non-existentkeys

agent/stats/types.go
```diff
@@ -38,7 +38,7 @@ type UsageStats struct {
 
 // ContainerMetadata contains meta-data information for a container.
 type ContainerMetadata struct {
-	DockerID string `json:"-"`
+	DockerID string
```
aaithal: why did you change this?

agent/tcs/client/client.go
```diff
@@ -221,6 +228,87 @@ func (cs *clientServer) metricsToPublishMetricRequests() ([]*ecstcs.PublishMetri
...
+		return nil, nil
+	}
+
+	var requests []*ecstcs.PublishHealthRequest
+	var taskHealths []*ecstcs.TaskHealth
+	numOfTasks := len(taskHealthMetrics)
+	for i, taskHealth := range taskHealthMetrics {
+		taskHealths = append(taskHealths, taskHealth)
+		// create a request if the number of task reaches the maximum page size
+		if (i+1)%tasksInMessage == 0 {
```
aaithal: Please define a new const for `tasksInHealthMessage` and rename `tasksInMessage` to `tasksInMetricsMessage`. Also add a TODO for determining the proper value for `tasksInHealthMessage`. Because of changes in payload structure, we'd have differing values for these I think

agent/tcs/client/client.go
aaithal: This is still an outstanding item

agent/api/containerstatus.go
```diff
@@ -52,6 +64,18 @@ var containerStatusMap = map[string]ContainerStatus{
...
 
+// BackendStatus returns the container health status recognized by backend
+func (healthStatus ContainerHealthStatus) BackendStatus() string {
+	switch healthStatus {
+	case ContainerHealthy:
+		return "HEALTHY"
+	case ContainerUnhealthy:
+		return "UNHEALTHY"
+	default:
+		return "UNKNOWN"
```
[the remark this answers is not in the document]
richardpen: > When do you expect this to be the case?

When the docker inspect api failed(eg, timeout), agent wasn't able to get the health status of the container.

> Is this a valid backend status? That is, what will the health of the container/task be if we report it as "UNKNOWN"?

Yes, backend expect 'UNKNOWN' when agent can't get container health status. In that case the container health status will be reported as unknown.

agent/engine/docker_container_engine.go
```diff
@@ -743,7 +747,26 @@ func metadataFromContainer(dockerContainer *docker.Container) DockerContainerMet
...
+	if dockerContainer.State.Health.Status == "" {
+		return metadata
+	}
+
+	// Record the health check information if exists
+	health := api.HealthStatus{}
+	logLength := len(dockerContainer.State.Health.Log)
+	if logLength != 0 {
+		// Only save the last log from the health check
+		health.Output = dockerContainer.State.Health.Log[logLength-1].Output
```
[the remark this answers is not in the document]
richardpen: > Is this only the last line from the health check log? will this have enough context?
Yes, since the log is the health check command output, and if the command failed the output should be the same. So I think the last line is enough.

>is this the only way to get visibility into why the health check command may have failed?

You can also get the failure reason from exitcode. the output is the health check command output, which is more understandable than the exitcode.

agent/api/container_test.go
```diff
@@ -168,3 +168,36 @@ func TestSetupExecutionRoleFlag(t *testing.T) {
...
+	container := Container{}
+
+	// set the container status to be healthy
+	container.SetHealthStatus(HealthStatus{Status: ContainerHealthy, Output: "test"})
+	health := container.GetHealthStatus()
+	assert.Equal(t, health.Status, ContainerHealthy)
+	assert.Equal(t, health.Output, "test")
+	assert.NotEmpty(t, health.Since)
+
+	// set the health status again shouldn't update the timestamp
```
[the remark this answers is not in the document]
richardpen: yes, the timestamps means the timestamp when the status changed.

agent/api/containerstatus.go
```diff
@@ -121,7 +150,7 @@ func (cs *ContainerStatus) UnmarshalJSON(b []byte) error {
 	}
 	if b[0] != '"' || b[len(b)-1] != '"' {
 		*cs = ContainerStatusNone
-		return errors.New("ContainerStatus must be a string or null; Got " + string(b))
+		return errors.New("containerStatus must be a string or null; Got " + string(b))
```
aaithal: Can you reformat this as: `"container status unmarshal: status must be a string or null.."`

agent/api/containerstatus.go
```diff
@@ -137,7 +166,7 @@ func (cs *ContainerStatus) UnmarshalJSON(b []byte) error {
 	stat, ok := containerStatusMap[strStatus]
 	if !ok {
 		*cs = ContainerStatusNone
-		return errors.New("Unrecognized ContainerStatus")
+		return errors.New("unrecognized ContainerStatus")
```
aaithal: Can you reformat this as: `"container status unmarshal: unrecognized status"`

agent/api/containerstatus.go
```diff
@@ -151,6 +180,43 @@ func (cs *ContainerStatus) MarshalJSON() ([]byte, error) {
...
+// UnmarshalJSON overrides the logic for parsing the JSON-encoded container health data
+func (healthStatus *ContainerHealthStatus) UnmarshalJSON(b []byte) error {
+	if strings.ToLower(string(b)) == "null" {
+		*healthStatus = ContainerHealthUnknown
+		return nil
+	}
+
+	if b[0] != '"' || b[len(b)-1] != '"' {
+		*healthStatus = ContainerHealthUnknown
+		return errors.New("containerHealthStatus must be a string or null; Got " + string(b))
```
aaithal: Same here. Please reformat this as per go standard for error strings:
`"container health status: ..."`

agent/api/containerstatus.go
```diff
@@ -151,6 +180,43 @@ func (cs *ContainerStatus) MarshalJSON() ([]byte, error) {
...
+		*healthStatus = ContainerHealthy
+		return nil
+	}
+
+	if strStatus == "UNHEALTHY" {
+		*healthStatus = ContainerUnhealthy
+		return nil
+	}
+
+	return errors.New("unrecognized container health status: " + string(b))
```
aaithal: Same here too: `"container health status: unrecognized status..."`

agent/engine/docker_task_engine.go
```diff
@@ -489,28 +489,40 @@ func (engine *DockerTaskEngine) handleDockerEvents(ctx context.Context) {
...
-	log.Debug("Handling a docker event", "event", event)
+func (engine *DockerTaskEngine) handleDockerEvent(event DockerContainerChangeEvent) {
+	seelog.Debugf("Handling a docker event: %s", event.String())
 
 	task, taskFound := engine.state.TaskByID(event.DockerID)
 	cont, containerFound := engine.state.ContainerByID(event.DockerID)
 	if !taskFound || !containerFound {
-		log.Debug("Event for container not managed", "dockerId", event.DockerID)
-		return false
+		seelog.Debugf("Event for container not managed, container: %s", event.DockerID)
```
aaithal: can you also print the values of taskFound and containerFound in the log?

agent/app/agent_capability.go
```diff
@@ -99,6 +100,11 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 		capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+"execution-role-ecr-pull")
 	}
 
+	if _, ok := supportedVersions[dockerclient.Version_1_24]; ok {
```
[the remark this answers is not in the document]
richardpen: Docker health check was available from 1.24 and afterwards. `StartPeriod` was added in 1.29. We only check the 1.24 here, and backend will check 1.29 capability if the `StartPeriod` is specified.

agent/engine/types.go
```diff
@@ -73,3 +78,41 @@ type ListContainersResponse struct {
...
 	Error error
 }
+
+// String returns a human readable string of the container change event
+func (event *DockerContainerChangeEvent) String() string {
+	res := fmt.Sprintf("Status: %s, DockerID: %s", event.Status.String(), event.DockerID)
+	res += ", health: " + event.Health.Status.String()
+
+	if event.ExitCode != nil {
+		res += fmt.Sprintf(", ExitCode: %v", aws.IntValue(event.ExitCode))
```
aaithal: please use `'%d'` here

agent/stats/engine.go
```diff
@@ -40,9 +40,20 @@ const (
...
+	// statsMetricsMap represents the map 'tasksToContainers'
+	statsMetricsMap statsEngineMapType = iota
+	// statsHealthmetricsMap represents the map 'tasksToHealthCheckContainers'
+	statsHealthmetricsMap
+)
+
 // EmptyMetricsError indicates an error for a task when there are no container
 // metrics to report
 var EmptyMetricsError = errors.New("stats engine: no task metrics to report")
+var EmptyHealthMetricsError = errors.New("stats engine: no task health metrics to report")
```
aaithal: missing `// EmptyHealth...` here. Also, you can do this to organize this better:

```
var (
 EmptyMetricsError = ...
 EmptyHealthMetricsError = ...
)
```

agent/stats/engine.go
```diff
@@ -163,77 +200,83 @@ func (engine *DockerStatsEngine) removeAll() {
...
-	defer engine.lock.Unlock()
-
+func (engine *DockerStatsEngine) addContainer(dockerID string) *StatsContainer {
 	// Make sure that this container belongs to a task and that the task
 	// is not terminal.
 	task, err := engine.resolver.ResolveTask(dockerID)
 	if err != nil {
 		seelog.Debugf("Could not map container to task, ignoring, err: %v, id: %s", err, dockerID)
-		return
+		return nil
```
aaithal: please return `error` and `*StatsContainer` instead of returning just an empty object (here and everywhere in this method).

agent/stats/engine.go
```diff
@@ -163,77 +200,83 @@ func (engine *DockerStatsEngine) removeAll() {
...
+		engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, statsHealthmetricsMap)
+	}
+
+	return statsContainer
+}
+
+func (engine *DockerStatsEngine) addToStatsContainerMapUnsafe(
+	taskARN, containerID string,
+	statsContainer *StatsContainer,
+	mapType statsEngineMapType) *StatsContainer {
```
aaithal: as per offline conversation, this still seems a bit clunky to refer to a map that you have access from an enum. Can you consider using a function pointer here instead, which returns you the appropriate map to manipulate?

agent/stats/engine.go
```diff
@@ -296,13 +339,108 @@ func (engine *DockerStatsEngine) GetInstanceMetrics() (*ecstcs.MetricsMetadata,
...
+}
+
 func (engine *DockerStatsEngine) isIdle() bool {
 	engine.lock.RLock()
 	defer engine.lock.RUnlock()
 
 	return len(engine.tasksToContainers) == 0
 }
 
+func (engine *DockerStatsEngine) isHealthContainerIdle() bool {
```
aaithal: can you rename this to `containerHealthsToMonitor()`? we can then do something like 
```
if ok:=engine.containerHealthsToMonitor(); ok {
 // we have containers whose healths are being monitored
} else {
 // no containers' healths are being monitored
}
```

It reads better than `isHealthContainerIdle`, which doesn't exactly convey the intent of the method.

agent/tcs/client/client.go
```diff
@@ -30,15 +30,20 @@ import (
 	"github.com/cihub/seelog"
 )
 
-// tasksInMessage is the maximum number of tasks that can be sent in a message to the backend
-// This is a very conservative estimate assuming max allowed string lengths for all fields.
-const tasksInMessage = 10
+const (
+	// tasksInMessage is the maximum number of tasks that can be sent in a message to the backend
+	// This is a very conservative estimate assuming max allowed string lengths for all fields.
+	tasksInMessage = 10
```
aaithal: you can rename this to `tasksInMetricMessage`

agent/tcs/client/client.go
```diff
@@ -83,7 +90,11 @@ func (cs *clientServer) Serve() error {
 	// Start the timer function to publish metrics to the backend.
 	cs.publishTicker = time.NewTicker(cs.publishMetricsInterval)
 	cs.endPublish = make(chan struct{})
-	go cs.publishMetrics()
+
+	if !cs.disableResourceMetrics {
+		go cs.publishMetrics()
+	}
+	go cs.publishHelathMetrics()
```
aaithal: typo: `publishHelathMetrics` -> `publishHealthMetrics`

agent/tcs/client/client.go
```diff
@@ -30,15 +30,20 @@ import (
...
 
-// tasksInMessage is the maximum number of tasks that can be sent in a message to the backend
-// This is a very conservative estimate assuming max allowed string lengths for all fields.
-const tasksInMessage = 10
+const (
+	// tasksInMessage is the maximum number of tasks that can be sent in a message to the backend
+	// This is a very conservative estimate assuming max allowed string lengths for all fields.
+	tasksInMessage = 10
+	// tasksInHealthMessage is the maximum number of tasks that can be sent in a message to the backend
+	tasksInHealthMessage = 10
```
aaithal: please add a TODO for evaluating what this number should be

agent/engine/docker_container_engine.go
```diff
@@ -45,6 +45,10 @@ const (
 	imageNameFormat = "%s:%s"
 	// the buffer size will ensure agent doesn't miss any event from docker
 	dockerEventBufferSize = 100
+	// healthCheckHealthy is the health status returned from docker container health check
+	healthCheckHealthy = "healthy"
+	// healthCheckUnhealthy is unhealth status returned from docker container health check
```
aaithal: nit: 'unhealthy'

agent/engine/docker_task_engine.go
```diff
@@ -489,28 +489,41 @@ func (engine *DockerTaskEngine) handleDockerEvents(ctx context.Context) {
...
+		return
 	}
+
+	// Container health status change doesnot affect the container status
+	// no need to process this in task manager
+	if event.Type == api.ContainerHealthEvent {
+		if cont.Container.HealthStatusShouldBeReported() {
+			seelog.Debugf("Updating container health status: %s", event.DockerContainerMetadata.Health)
+			cont.Container.SetHealthStatus(event.DockerContainerMetadata.Health)
+			engine.saver.Save()
```
aaithal: I don't think we need to persist this in the state file. In the event of a restart, inspecting the container should be sufficient to get the health status of the container. wdyt? imo, we can delete this line.

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
-	return nil
 }
 
 // addContainer adds a container to the map of containers being watched.
-// It also starts the periodic usage data collection for the container.
-func (engine *DockerStatsEngine) addContainer(dockerID string) {
-	engine.lock.Lock()
-	defer engine.lock.Unlock()
-
+func (engine *DockerStatsEngine) addContainer(dockerID string) (*StatsContainer, error) {
```
aaithal: please rename this to `addContainerUnsafe`

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
+	dockerContainer, err := engine.resolver.ResolveContainer(dockerID)
+	if err == nil && dockerContainer.Container.HealthStatusShouldBeReported() {
+		seelog.Debugf("Adding container to stats health check watch list, id: %s, task: %s", dockerID, task.Arn)
+		engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, engine.healthCheckContainerMetricsMap)
 	}
 
+	return statsContainer, nil
+}
+
+func (engine *DockerStatsEngine) containerMetricsMap() map[string]map[string]*StatsContainer {
```
aaithal: please rename this as `containerMetricsMapUnsafe`

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
+	return engine.tasksToContainers
+}
+func (engine *DockerStatsEngine) healthCheckContainerMetricsMap() map[string]map[string]*StatsContainer {
```
aaithal: please rename this as `healthCheckContainerMetricsMapUnsafe`

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
+	return engine.tasksToHealthCheckContainers
+}
+
+func (engine *DockerStatsEngine) addToStatsContainerMapUnsafe(
+	taskARN, containerID string,
+	statsContainer *StatsContainer,
+	fun func() map[string]map[string]*StatsContainer) *StatsContainer {
```
aaithal: `fun` is a very generic, non descriptive name. Please rename that to be something more meaningful

agent/api/container.go
```diff
@@ -32,15 +33,35 @@ const (
...
 type DockerConfig struct {
-	Config     *string `json:"config"`
+	// Config is the configuration used to create container
+	Config *string `json:"config"`
+	// HostConfig is the configuration of container related to host resource
 	HostConfig *string `json:"hostConfig"`
-	Version    *string `json:"version"`
+	// Version specifies the docker client API version to use
+	Version *string `json:"version"`
+	// HealthCheck is the configuration of docker health check
```
samuelkarp: Is this another json-serialized object?  Is this part of Config/HostConfig?

agent/api/containerstatus.go
```diff
@@ -151,6 +180,43 @@ func (cs *ContainerStatus) MarshalJSON() ([]byte, error) {
...
+		*healthStatus = ContainerHealthUnknown
+		return nil
+	}
+
+	if b[0] != '"' || b[len(b)-1] != '"' {
+		*healthStatus = ContainerHealthUnknown
+		return errors.New("container health status unmarshal: status must be a string or null; Got " + string(b))
+	}
+	strStatus := string(b[1 : len(b)-1])
+	if strStatus == "UNKNOWN" {
```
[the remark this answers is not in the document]
samuelkarp: Please use `switch` for this.

agent/engine/docker_container_engine.go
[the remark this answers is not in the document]
samuelkarp: Why would you only have the last line?  Also: I don't see any limit here to the length of the line.  We should probably limit this in terms of bytes and maybe select the last N bytes instead of the last line.

agent/functional_tests/tests/generated/simpletests_unix/simpletests_generated_unix_test.go
```diff
@@ -397,7 +397,7 @@ func TestInitProcessEnabled(t *testing.T) {
 	}
 	agent := RunAgent(t, nil)
 	defer agent.Cleanup()
-	agent.RequireVersion(">=1.14.5")
+	agent.RequireVersion(">=1.15.0")
```
samuelkarp: Why did this change?

richardpen: I believe this was changed to `1.14.5` unexpectedly [here](https://github.com/aws/amazon-ecs-agent/commit/c268ea09b58d6eacf704ccf648e5ae6817a2b8a0), so I changed it back.

agent/api/container.go
richardpen: No and it's not part of the Config/HostConfig from the payload message. It was sent as a string and agent will unmarshal it into the docker recognized struct [api/task.go](https://github.com/aws/amazon-ecs-agent/pull/1141/files/dfdc730c25029805e1b602ecb7b64a35f70a6a74#diff-e008975819a900a8494cf91881a8304aR458)

richardpen:
@samuelkarp Not sure why github doesn't allow me to response your comment in place, so I put it here. We only have the last line because the line should be the same in most cases for the failure and it should be enough for debugging purpose, and it saves memory to save the additional output especially when the retry number is very big. 
For your suggestion to add limit, I think that's a good idea, I'll make the change. Thanks.

agent/functional_tests/tests/generated/simpletests_unix/simpletests_generated_unix_test.go
samuelkarp: @sharanyad Can you clarify whether that was intended?

agent/api/container.go
samuelkarp: Why isn't it part of the `Config` from the payload message?  It's being stuck in the `Config` anyway, so why do we need a separate unmarshal for it?

samuelkarp:
> the line should be the same in most cases for the failure

I don't understand what you mean by that.

> it should be enough for debugging purpose

This seems like a pretty big assumption, unless I'm misunderstanding who creates the output.

agent/functional_tests/tests/generated/simpletests_unix/simpletests_generated_unix_test.go
sharanyad: @richardpen is right. According to the releases, `1.15.0` has the `init` support in agent. Not sure how the testing change got slipped in.

richardpen:
@samuelkarp Sorry for the confusing, the last line isn't the last line of the output, it is the last entry of the outputs which is the whole output of the latest health check command.

samuelkarp:
@richardpen Thanks, that does help.  I think that makes sense, the only thing now is to look at making sure we implement a size limit for the output.

richardpen:
@samuelkarp @aaithal I have addressed all the comments, please take another look.

agent/api/task.go
```diff
@@ -455,11 +456,15 @@ func (task *Task) dockerConfig(container *Container, apiVersion dockerclient.Doc
...
 
 	if container.DockerConfig.Config != nil {
-		err := json.Unmarshal([]byte(*container.DockerConfig.Config), &config)
+		err := json.Unmarshal([]byte(aws.StringValue(container.DockerConfig.Config)), &config)
 		if err != nil {
 			return nil, &DockerClientConfigError{"Unable decode given docker config: " + err.Error()}
 		}
 	}
+	if container.HealthCheckType == dockerHealthCheckType && config.Healthcheck == nil {
+		return nil, &DockerClientConfigError{"docker health check is nil while container health check type is DOCKER"}
```
aaithal: nit: please split this into multiple lines for readability

agent/api/task.go
```diff
@@ -1085,12 +1090,12 @@ func (task *Task) GetID() (string, error) {
...
 
 	if !strings.Contains(resource, arnResourceDelimiter) {
-		return "", errors.New(fmt.Sprintf("task get-id: malformed task resource: %s", resource))
+		return "", errors.Errorf("task get-id: malformed task resource: %s", resource)
 	}
 
 	resourceSplit := strings.SplitN(resource, arnResourceDelimiter, arnResourceSections)
 	if len(resourceSplit) != arnResourceSections {
-		return "", errors.New(fmt.Sprintf("task get-id: invalid task resource split: %s, expected=%d, actual=%d", resource, arnResourceSections, len(resourceSplit)))
+		return "", errors.Errorf("task get-id: invalid task resource split: %s, expected=%d, actual=%d", resource, arnResourceSections, len(resourceSplit))
```
aaithal: nit: please split this into multiple lines for readability

agent/engine/docker_container_engine.go
```diff
@@ -45,6 +45,12 @@ const (
 	imageNameFormat = "%s:%s"
 	// the buffer size will ensure agent doesn't miss any event from docker
 	dockerEventBufferSize = 100
+	// healthCheckHealthy is the healthy status returned from docker container health check
+	healthCheckHealthy = "healthy"
+	// healthCheckUnhealthy is unhealthy status returned from docker container health check
+	healthCheckUnhealthy = "unhealthy"
+	// maxHealthCheckOutputSize is the maximum size of healthcheck command output that agent will save
```
aaithal: nit: please change this to "maximum length" instead of "size"

agent/stats/engine.go
```diff
@@ -123,20 +164,15 @@ func (engine *DockerStatsEngine) MustInit(taskEngine ecsengine.TaskEngine, clust
...
 	// Subscribe to the container change event stream
-	err := engine.containerChangeEventStream.Subscribe(containerChangeHandler, engine.handleDockerEvents)
+	err = engine.containerChangeEventStream.Subscribe(containerChangeHandler, engine.handleDockerEvents)
 	if err != nil {
 		return fmt.Errorf("Failed to subscribe to container change event stream, err %v", err)
 	}
 
-	go engine.listContainersAndStartEventHandler()
-	go engine.waitToStop()
+	engine.synchronizeState()
```
aaithal: this returns an error. please make sure that you log a warning if an error is returned.

agent/stats/engine.go
```diff
@@ -102,14 +111,46 @@ func (resolver *DockerContainerMetadataResolver) ResolveContainer(dockerID strin
...
+	return nil
+}
+
+// addAndStartStatsContainer add the container into stats engine and start collecting the container stats
+func (engine *DockerStatsEngine) addAndStartStatsContainer(containerID string) {
+	engine.lock.Lock()
+	defer engine.lock.Unlock()
+	statsContainer, err := engine.addContainerUnsafe(containerID)
+	if err != nil {
+		seelog.Debugf("Adding container to stats watch list failed, err: %v", err)
```
aaithal: please log container id here.

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
+	statsContainer := newStatsContainer(dockerID, engine.client, engine.resolver)
+	engine.tasksToDefinitions[task.Arn] = &taskDefinition{family: task.Family, version: task.Version}
+
+	if !engine.disableMetrics {
+		// Adding container to the map for collecting stats
+		statsContainer = engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, engine.containerMetricsMapUnsafe)
+	}
+
+	dockerContainer, err := engine.resolver.ResolveContainer(dockerID)
+	if err == nil && dockerContainer.Container.HealthStatusShouldBeReported() {
```
aaithal: i think this will make it hard for us to debug this in the future. Can you please restructure this like below:

```
if dockerContainer, err := engine.resolver.ResolveContainer(dockerID); err != nil {
 // log something, probably at debug
} else {
 if dockerContainer.Container. HealthStatusShouldBeReported() {
  // add to the map
 }
}
```

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
+		engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, engine.healthCheckContainerMetricsMapUnsafe)
 	}
 
+	return statsContainer, nil
+}
+
+func (engine *DockerStatsEngine) containerMetricsMapUnsafe() map[string]map[string]*StatsContainer {
+	return engine.tasksToContainers
+}
+func (engine *DockerStatsEngine) healthCheckContainerMetricsMapUnsafe() map[string]map[string]*StatsContainer {
```
aaithal: please add a newline here and rename this to `healthCheckContainerMapUnsafe`

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
+	return engine.tasksToHealthCheckContainers
+}
+
+func (engine *DockerStatsEngine) addToStatsContainerMapUnsafe(
+	taskARN, containerID string,
+	statsContainer *StatsContainer,
+	statsMapToUpdate func() map[string]map[string]*StatsContainer) *StatsContainer {
```
aaithal: why is this method returning `*StatsContainer`? iiuc, you're just returning `statsContainer` here and that's not changing any state. It may make more sense to return a bool indicating if we should start watching this container or not based on its existence in the map.

agent/tcs/client/client.go
```diff
@@ -221,6 +232,86 @@ func (cs *clientServer) metricsToPublishMetricRequests() ([]*ecstcs.PublishMetri
...
+		seelog.Debug("Skipping publishing health metrics. Publish ticker is uninitialized")
+		return
+	}
+
+	// Publish metrics immediately after we connect and wait for ticks. This makes
+	// sure that there is no data loss when a scheduled metrics publishing fails
+	// due to a connection reset.
+	err := cs.publishHealthMetricsOnce()
+	if err != nil {
+		seelog.Warnf("Error publishing health metrics: %v", err)
```
aaithal: nit: `"Unable to publish health metrics"` would be a better log message

agent/tcs/client/client.go
```diff
@@ -221,6 +232,86 @@ func (cs *clientServer) metricsToPublishMetricRequests() ([]*ecstcs.PublishMetri
...
+	}
+	for {
+		select {
+		case <-cs.publishTicker.C:
+			err := cs.publishHealthMetricsOnce()
+			if err != nil {
+				seelog.Warnf("Error publishing health metrics: %v", err)
```
aaithal: nit: `"Unable to publish health metrics"` would be a better log message

agent/tcs/handler/handler.go
```diff
@@ -145,6 +147,15 @@ func ackPublishMetricHandler(timer *time.Timer) func(*ecstcs.AckPublishMetric) {
 	}
 }
 
+// ackPublishHealthMetricHandler consumes the ack message from backend. The backend sends
+// the ack each time it processes a health message
+func ackPublishHealthMetricHandler(timer *time.Timer) func(*ecstcs.AckPublishHealth) {
+	return func(*ecstcs.AckPublishHealth) {
+		seelog.Debug("Received ACKPublishHealth from tcs")
```
aaithal: Can you check if the message id can be logged here? That'd improve the message tracking if we need to debug anything with lossy messages in the future.

agent/stats/engine.go
richardpen: I was trying to avoid an extra `if` check, but I can modify that if this makes the code easier to read.

agent/tcs/handler/handler.go
richardpen: The message ID isn't available alone here, we can only log the whole message.

agent/stats/engine.go
```diff
@@ -163,77 +199,78 @@ func (engine *DockerStatsEngine) removeAll() {
...
-	return nil
 }
 
-// addContainer adds a container to the map of containers being watched.
-// It also starts the periodic usage data collection for the container.
-func (engine *DockerStatsEngine) addContainer(dockerID string) {
-	engine.lock.Lock()
-	defer engine.lock.Unlock()
-
+// addContainerUnsafe adds a container to the map of containers being watched.
```
adnxn: maybe i missed this. why are we getting rid of the locks here and using `*Unsafe`?

agent/tcs/handler/handler.go
aaithal: oh, never mind then

agent/app/agent.go
richardpen: This has already been addressed. Disable the metrics won't affect the health check part.

agent/stats/engine.go
richardpen: No, the lock was just moved to line 138.

agent/api/container.go
```diff
@@ -51,25 +70,40 @@ type Container struct {
...
+	// MountPoints contains a list of volume mount paths
+	MountPoints []MountPoint `json:"mountPoints"`
+	// Ports contains a list of ports binding configuration
+	Ports []PortBinding `json:"portMappings"`
+	// Essential denotes whether the container is essential or not
+	Essential bool
+	// EntryPoint is entrypoint of the container, corresponding to docker option: --entrypoint
+	EntryPoint *[]string
+	// Environment is the environment variable set in the container
+	Environment map[string]string `json:"environment"`
```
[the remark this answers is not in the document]
richardpen: I think this is pretty straightforward, as it's like setting any environment variable. Not sure what to put here. I'll leave it as for now, unless you have a better suggestion what to put here.

agent/stats/engine.go
adnxn: oops not sure how i missed that. thanks.

agent/api/containerstatus.go
```diff
@@ -151,6 +180,38 @@ func (cs *ContainerStatus) MarshalJSON() ([]byte, error) {
...
+	if strings.ToLower(string(b)) == "null" {
+		return nil
+	}
+	if b[0] != '"' || b[len(b)-1] != '"' {
+		return errors.New("container health status unmarshal: status must be a string or null; Got " + string(b))
+	}
+
+	strStatus := string(b[1 : len(b)-1])
+	switch strStatus {
+	case "UNKNOWN":
```
aaithal: nit: please add a comment here stating that the status is already set to `ContainerHealthUnknown`.

agent/engine/docker_task_engine.go
```diff
@@ -489,28 +498,40 @@ func (engine *DockerTaskEngine) handleDockerEvents(ctx context.Context) {
...
+		seelog.Debugf("Event for container not managed, container: %s, task found: %v, container found: %v",
+			event.DockerID, taskFound, containerFound)
+		return
 	}
+
+	// Container health status change doesnot affect the container status
+	// no need to process this in task manager
+	if event.Type == api.ContainerHealthEvent {
+		if cont.Container.HealthStatusShouldBeReported() {
+			seelog.Debugf("Updating container health status: %v", event.DockerContainerMetadata.Health)
```
aaithal: please log container name/id and task arn here as well

agent/stats/engine.go
```diff
@@ -163,77 +202,87 @@ func (engine *DockerStatsEngine) removeAll() {
...
+
+	if dockerContainer, err := engine.resolver.ResolveContainer(dockerID); err != nil {
+		seelog.Debugf("Could not map container ID to container, container: %s, err: %s", dockerID, err)
+	} else if dockerContainer.Container.HealthStatusShouldBeReported() {
+		// Track the container health status
+		engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, engine.healthCheckContainerMapUnsafe)
+		seelog.Debugf("Adding container to stats health check watch list, id: %s, task: %s", dockerID, task.Arn)
+	}
+
+	if !shouldCollectStats {
```
aaithal: can you move this to line 238, before invoking `ResolveContainer`?

richardpen: line 238-243 should be always executed for collecting container health status. So move this to line 238 will skip this part if `shouldCollectStats` was false.

aaithal: Ah, I misunderstood. Can you please rename `shouldCollectStats` to `untrackedStatsContainer` or something similar? `shouldCollectStats` is somewhat misleading in this context.

richardpen:
@aaithal I will add the integration test in a separate PR, please take another look, thanks.

agent/stats/engine.go
```diff
@@ -163,77 +202,87 @@ func (engine *DockerStatsEngine) removeAll() {
...
+
+	if dockerContainer, err := engine.resolver.ResolveContainer(dockerID); err != nil {
+		seelog.Debugf("Could not map container ID to container, container: %s, err: %s", dockerID, err)
+	} else if dockerContainer.Container.HealthStatusShouldBeReported() {
+		// Track the container health status
+		engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, engine.healthCheckContainerMapUnsafe)
+		seelog.Debugf("Adding container to stats health check watch list, id: %s, task: %s", dockerID, task.Arn)
+	}
+
+	if !statsContainerTracked {
```
aaithal: this still doesn't read right. if stats container is not tracked, we should start tracking it. However, that's not what this variable is conveying. Can you rename this to be `watchStatsContainer`?

agent/stats/engine.go
```diff
@@ -163,77 +202,87 @@ func (engine *DockerStatsEngine) removeAll() {
...
+	return engine.tasksToContainers
+}
+
+func (engine *DockerStatsEngine) healthCheckContainerMapUnsafe() map[string]map[string]*StatsContainer {
+	return engine.tasksToHealthCheckContainers
+}
+
+// addToStatsContainerMapUnsafe adds the statscontainer into stats for tracking and returns whether
+// stats should start the statscontainer to collect metrics
+func (engine *DockerStatsEngine) addToStatsContainerMapUnsafe(
```
aaithal: please update the documentation for with a description of what's returned from this method.

damiencarol:
Any new on this one?

richardpen:
@damiencarol Thanks for tracking this, we are still actively working on this. I'm going to merge this PR, and we still need to add some tests, please track the #534 for updates.

PR #1142, opened 2017-12-12
taskmetadata: bump throttles

aaithal:
### Summary

Because all containers in a task share the same IP address in an
"awsvpc" task, and a task can be constituted of up to 10 containers, the
steady state rate is set at 10 rps. Burst rate is set to 15 rps.

### Implementation details

NA

### Testing

New tests cover the changes: 

NA

### Licensing

### DIFF
--- a/agent/handlers/taskmetadata/handler.go
+++ b/agent/handlers/taskmetadata/handler.go
@@ -41,9 +41,20 @@ const (
 	apiVersion1 = 1
 	apiVersion2 = 2
 
-	// Rate limits for the metadata endpoint(s) and APIs by request ip addresses
-	rateLimitPerSecond      = 3
-	rateLimitBurstPerSecond = 6
+	// Rate limits for the metadata endpoint(s) and APIs by request ip addresses. Today, we
+	// support the following endpoints:
+	// 1. /v2/credentials: For serving IAM role credentials
+	// 2. /v2/metadata: For serving task and container metadata information (for "awsvpc" tasks)
+	// 3. /v2/stats: For serving task and container stats information (for "awsvpc" tasks)
+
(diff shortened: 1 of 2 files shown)


agent/handlers/taskmetadata/handler.go
```diff
@@ -41,9 +41,20 @@ const (
...
+	// support the following endpoints:
+	// 1. /v2/credentials: For serving IAM role credentials
+	// 2. /v2/metadata: For serving task and container metadata information (for "awsvpc" tasks)
+	// 3. /v2/stats: For serving task and container stats information (for "awsvpc" tasks)
+
+	// rateLimitPerSecond specifies the steady state throttle for the task metadata endpoint.
+	// Because all containers in a task share the same IP address in an "awsvpc" task, and a
+	// task can be constituted of up to 10 containers, the steady state rate is set at 10
+	// per second
+	rateLimitPerSecond = 10
```
richardpen: I was wondering if 10 is too small when there are 10 containers requesting this metadata. 
Since the maximum tasks with `awsvpc` mode is limited by number of [enis(<=15)](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html), that means the maximum request rate is 15 * 15, would it possible to increase the number to something bigger comparing with the maximum requests agent can handler per second, eg: `rateLimitPerSecond = 20; rateLimitBurstPerSecond = 30`

aaithal: This is 10 per task as it's filtered by the ip address, not an aggregate throttle for all tasks on the instance

richardpen: Sorry, I misunderstood that this is for all the containers but not just the container launched in `awsvpc` mode.

PR #1147, opened 2017-12-19
[WIP] Fix an error case with task limits

petderek:
### Summary
This fixes a bug with task limits

### Implementation details
If the host doesn't have hierarchical memory flag  (`use_hierarchy`) enabled, it can cause tasks to not enforce task limits correctly. This doesn't affect container size limits.

The agent will now manually set the `use_hierarchy` flag.

### Testing

New tests cover the changes: indeed

* Bug - Task limit enforcement breaks in some conditions

### Licensing

### DIFF
--- a/agent/resources/resources_linux.go
+++ b/agent/resources/resources_linux.go
@@ -16,27 +16,39 @@
 package resources
 
 import (
+	"os"
+	"path/filepath"
+	"strconv"
+
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
+	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
 	"github.com/cihub/seelog"
 	"github.com/containerd/cgroups"
 	"github.com/pkg/errors"
 )
 
+const (
+	memorySubsystem    = "/sys/fs/cgroup/memory/"
+	memoryUseHierarchy = "memory.use_hierarchy"
+)
+
 // cgroupWrapper implements the Resource interface
 type cgroupWrapper struct {
 	control cgroup.Control
+	ioutil  ioutilwrapper.IOUtil
 }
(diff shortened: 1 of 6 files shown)


agent/resources/resources_linux.go
```diff
@@ -16,27 +16,39 @@
...
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
+	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
 	"github.com/cihub/seelog"
 	"github.com/containerd/cgroups"
 	"github.com/pkg/errors"
 )
 
+const (
+	memorySubsystem    = "/sys/fs/cgroup/memory/"
```
nmeyerhans: This feels like tech debt to me. If we're running in a container managed by ecs-init, we can assume the existence of this path, but otherwise we can't. Additionally, we probably don't want to set memory.use_hierarchy system-wide, but only for a specific cgroup hierarchy. Probably this should simply be the /ecs hierarchy, but it's conceivable we might want it only at the task level. Really we only need it at the task level, but setting it for /ecs will get us that and I don't think there are any drawbacks.

petderek: You're right, I was only considering the container as managed by ecs-init. However, the /ecs/ hierarchy will be brought in on line 109 with the `filepath.Join(memorySubsystem, cgroupRoot, memoryUseHierarchy)`. cgroupRoot should contain ` /ecs/`.

I think I can make a couple improvements:
- check for the other common prefix for cgroup hierarchy
- check to see if the value is 1 before writing

adnxn: > check to see if the value is 1 before writing

this makes sense

@nmeyerhans do you think it makes sense to have this be configurable through ecs.config? and for our case default to `/sys/fs/cgroup/memory/`. let me know what you think please.

adnxn:
closing in favor of https://github.com/aws/amazon-ecs-agent/pull/1195

PR #1148, opened 2017-12-21
eni watcher: add backoff-retry

aaithal:
### Summary

eni watcher: add backoff-retry

### Implementation details

- [x] This commit adds back-off retry logic in:
* udev watcher: It takes a few milliseconds for the ENI's mac address to
show up on the instance once udev watcher detects that a new device has
been attached
* eni state lookup: If for whatever reason, there's a delay in ACS
sending the ENI attachment message to the agent, the agent's state
remains unaware of the ENI
- [x] changelog entry

This should lead to improved task start latencies of awsvpc tasks.

### Testing

- [x] Manual tests: 
* Time between `AttachTaskNetworkInterfacesMessage` and `PayloadMessage` without this change: 23s
* Time between `AttachTaskNetworkInterfacesMessage` and `PayloadMessage` with this change: 11s

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Feature - Support a HTTP endpoint for `awsvpc` tasks to query metadata
 * Bug - Fixed a bug where `-version` fails due to its dependency on docker client. [#1118](https://github.com/aws/amazon-ecs-agent/pull/1118)
 * Bug - Persist container exit code in agent state file [#1125](https://github.com/aws/amazon-ecs-agent/pull/1125)
+* Bug - Fixed a bug where the agent could miss sending an ENI attachment to ECS because of address propagation delays [#1148](https://github.com/aws/amazon-ecs-agent/pull/1148)
 
 ## 1.16.0
 * Feature - Support pulling from Amazon ECR with specified IAM role in task definition
--- a/agent/eni/networkutils/utils_linux.go
+++ b/agent/eni/networkutils/utils_linux.go
@@ -17,22 +17,113 @@
 package networkutils
 
 import (
+	"context"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/eni/netlinkwrapper"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+	"github.com/pkg/errors"
 
-	log "github.com/cihub/seelog"
+	"github.com/cihub/seelog"
 )
 
+const (
+	// macAddressBackoffMin specifies the mimimum duration for the backoff
+	// when looking for an ENI's mac address on the host
+	macAddressBackoffMin = 2 * time.Millisecond
+
+	// macAddressBackoffMax specifies the maximum duration for the backoff
+	// when looking for an ENI's mac address on the host
+	macAddressBackoffMax = 200 * time.Millisecond
+
+	// macAddressBackoffJitter specifies the jitter multiple percentage when
+	// looking for an ENI's mac address on the host
+	macAddressBackoffJitter = 0.2
+
+	// macAddressBackoffMultiple specifies the backoff duration multiplier
+	// when looking for an ENI's mac address on the host
+	macAddressBackoffMultiple = 1.5
+)
+
+// macAddressRetriever is used to retrieve the mac address of a device. It collects
+// all the information necessary to start this operation and stores the result in
+// the 'macAddress' attribute
+type macAddressRetriever struct {
+	dev           string
+	netlinkClient netlinkwrapper.NetLink
+	macAddress    string
+	// timeout specifies the timeout duration before giving up when
+	// looking for an ENI's mac address on the host
+	timeout time.Duration
+	ctx     context.Context
+}
+
 // GetMACAddress retrieves the MAC address of a device using netlink
-func GetMACAddress(dev string, netlinkClient netlinkwrapper.NetLink) (string, error) {
-	dev = filepath.Base(dev)
-	link, err := netlinkClient.LinkByName(dev)
+func GetMACAddress(ctx context.Context,
+	timeout time.Duration,
+	dev string,
+	netlinkClient netlinkwrapper.NetLink) (string, error) {
+	retriever := &macAddressRetriever{
+		dev:           dev,
+		netlinkClient: netlinkClient,
+		ctx:           ctx,
+		timeout:       timeout,
+	}
+	return retriever.retrieve()
+}
+
+// retrieve retrives the mac address of a network device. If the retrieved mac
+// address is empty, it retries the operation with a timeout specified by the
+// caller
+func (retriever *macAddressRetriever) retrieve() (string, error) {
+	backoff := utils.NewSimpleBackoff(macAddressBackoffMin, macAddressBackoffMax,
+		macAddressBackoffJitter, macAddressBackoffMultiple)
+	ctx, cancel := context.WithTimeout(retriever.ctx, retriever.timeout)
+	defer cancel()
+
+	err := utils.RetryWithBackoffCtx(ctx, backoff, func() error {
+		retErr := retriever.retrieveOnce()
+		if retErr != nil {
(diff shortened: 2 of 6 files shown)


agent/eni/networkutils/utils_linux.go
```diff
@@ -17,22 +17,113 @@
...
+				retriever.dev)
+		}
+
+		return nil
+	})
 	if err != nil {
 		return "", err
 	}
-	return link.Attrs().HardwareAddr.String(), err
+	// RetryWithBackoffCtx returns nil when the context is cancelled. Check if there was
```
richardpen: Would it be better to check `retriever.macAddress != ""`, though I don't strongly opposite this.

agent/eni/watcher/watcher_linux.go
```diff
@@ -57,6 +88,17 @@ type UdevWatcher struct {
 	primaryMAC           string
 }
 
+// unmanagedENIError is used to indicate that the agent found an ENI, but the agent isn't
```
richardpen: Can this be moved to another file?

agent/eni/networkutils/utils_linux.go
aaithal: Hmm, I don't know if it gives us any additional benefits. i like dealing with errors rather than empty strings or nils. So, will keep it this way

agent/eni/watcher/watcher_linux.go
aaithal: why?

agent/eni/networkutils/utils_linux.go
```diff
@@ -43,23 +134,23 @@ func IsValidNetworkDevice(devicePath string) bool {
 	* eth1 -> /devices/pci0000:00/0000:00:05.0/net/eth1
 	* eth0 -> ../../devices/pci0000:00/0000:00:03.0/net/eth0
 	* lo   -> ../../devices/virtual/net/lo
-	*/
+	 */
 	splitDevLink := strings.SplitN(devicePath, "devices/", 2)
 	if len(splitDevLink) != 2 {
-		log.Warnf("Cannot determine device validity: %s", devicePath)
+		seelog.Warnf("Cannot determine device validity: %s", devicePath)
```
adnxn: its not clear to me why `len(splitDevLink) != 2` means we cannot determine device validity. can we add some documentation? should we add the `splitDevLink` to log output as well? although, maybe i'm just missing context.

agent/eni/watcher/watcher_linux.go
```diff
@@ -57,6 +88,17 @@ type UdevWatcher struct {
...
+// aware if this ENI is being managed by ECS
+type unmanagedENIError struct {
+	mac string
+}
+
+// Error returns the error string for the unmanagedENIError type
+func (err *unmanagedENIError) Error() string {
+	return fmt.Sprintf("udev watcher send ENI state change: eni not managed by ecs: %s", err.mac)
```
adnxn: it's not clear to me what we mean by `udev watcher send ENI state change`. the second part of the error message is fine though.

agent/utils/errors.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2014-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
adnxn: thanks for documenting all these =]

agent/eni/watcher/watcher_linux.go
richardpen: I was thinking we may need a `type.go` for the struct and const, to split this file into smaller pieces which are easier to read.

agent/eni/networkutils/utils_linux.go
```diff
@@ -17,22 +17,113 @@
...
+
+	return retriever.macAddress, nil
+}
+
+// retrieveOnce retrieves the MAC address of a device using netlink.LinkByName
+func (retriever *macAddressRetriever) retrieveOnce() error {
+	dev := filepath.Base(retriever.dev)
+	link, err := retriever.netlinkClient.LinkByName(dev)
+	if err != nil {
+		return utils.NewRetriableError(utils.NewRetriable(false), err)
```
nmeyerhans: Are you sure that we should never retry in this case?

agent/eni/watcher/watcher_linux.go
aaithal: i'm not a fan of the `type.go` model. It's much more easier to navigate/organize the code where the struct and the methods are defined in the same file. We could move this into an `errors.go` file. But, seeing how it's just one type, I'd rather keep it here than to create a new file for it.

agent/eni/networkutils/utils_linux.go
```diff
@@ -17,22 +17,113 @@
...
 
 	"github.com/aws/amazon-ecs-agent/agent/eni/netlinkwrapper"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+	"github.com/pkg/errors"
 
-	log "github.com/cihub/seelog"
+	"github.com/cihub/seelog"
 )
 
+const (
```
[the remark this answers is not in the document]
aaithal: with experiments. It took an average of 10ms for the ENI's mac address to show up on the instance. The one for attachment message was based on my conversations with the ECS backend team.

agent/eni/networkutils/utils_linux.go
aaithal: yeah, an error means a syscall failed for describing the device. at this point of time, it doesn't make sense to retry it as it really should be visible on the instance

agent/eni/networkutils/utils_linux.go
aaithal: The example entries in the comment above are meant to illustrate that:

`eth1 -> /devices/pci0000:00/0000:00:05.0/net/eth1`

This means that we expect it to be of "devices/some-string" format.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Feature - Support a HTTP endpoint for `awsvpc` tasks to query metadata
 * Bug - Fixed a bug where `-version` fails due to its dependency on docker client. [#1118](https://github.com/aws/amazon-ecs-agent/pull/1118)
 * Bug - Persist container exit code in agent state file [#1125](https://github.com/aws/amazon-ecs-agent/pull/1125)
+* Bug - Fixed a bug where the agent could miss sending an ENI attachment to ECS because of address propagation delays [#1148](https://github.com/aws/amazon-ecs-agent/pull/1148)
```
samuelkarp: Can you wrap this at 80 characters?

agent/eni/networkutils/utils_linux.go
```diff
@@ -17,22 +17,113 @@
...
+	// when looking for an ENI's mac address on the host
+	macAddressBackoffMin = 2 * time.Millisecond
+
+	// macAddressBackoffMax specifies the maximum duration for the backoff
+	// when looking for an ENI's mac address on the host
+	macAddressBackoffMax = 200 * time.Millisecond
+
+	// macAddressBackoffJitter specifies the jitter multiple percentage when
+	// looking for an ENI's mac address on the host
+	macAddressBackoffJitter = 0.2
```
samuelkarp: I don't think jitter is really necessary here since you're not making remote calls or accessing a resource with high concurrency.  On the other hand, I don't think it really hurts either except to make the timing a bit more unpredictable.

agent/eni/networkutils/utils_linux.go
```diff
@@ -17,22 +17,113 @@
...
+		}
+
+		return nil
+	})
 	if err != nil {
 		return "", err
 	}
-	return link.Attrs().HardwareAddr.String(), err
+	// RetryWithBackoffCtx returns nil when the context is cancelled. Check if there was
+	// a timeout here. TODO: Fix RetryWithBackoffCtx to return ctx.Err() on context Done()
```
samuelkarp: oops, I think that's on me...

aaithal:
closing this as this is already merged to the master branch with 1.16.1 release

Issue #787, opened 2017-05-03
Add support for `--shm-size` `docker run` flag


aaithal:
Has [7 +1s](https://github.com/aws/amazon-ecs-agent/issues/502#issuecomment-245119957) and the following mentions:
* [502](https://github.com/aws/amazon-ecs-agent/issues/502#issuecomment-264976915)
* [655](https://github.com/aws/amazon-ecs-agent/issues/655#issuecomment-279770935)

sdeprez:
Hi ,

Any ETA for this feature? Or any workaround for it in the meantime?

Basically, I need to run some version of Chromium headless inside containers, and that requires more than the default 64M of shared memory so at the moment it's simply impossible to run it.

Thanks,

csvan:
+1, also running Chromium inside ECS here. Next to impossible with default shm size.

linuxfreakus:
+1, ditto on headless chromium.

matthew-cox:
+1, I would really prefer support for this option vs implementing the possible work arounds

trstephen:
+1, headless chromium is a shrinking violet with default 64mb shm

AlexMarchini:
+1, also stuck until this is implemented.

mkleint:
we are also waiting for the Amazon ECS AMi to contain latest docker daemon to at least be able to change the default size of shm to be bigger than 64M

alinpa:
+1, I am waiting to see this feature implemented.

zackarychapple:
Any update on when this will be a reality? We are running into issues on this as well.

tobypinder:
@alinpa we use

```
"Volumes": [
    {
      "HostDirectory": "/dev/shm",
      "ContainerDirectory": "/dev/shm"
    }
  ],
```

in the `Dockerrun.aws.json` to effectively hand the guest container the host's `shm`. I am unclear of any side effects here but it appears to work well for our single container environment. I'm not an expert but you and others here may find this a suitable workaround and we haven't noticed any issues arising from this yet.

aatreya:
Anyone interested in this for use with Chrome/Chromium may also want to star / comment on these bugs:
https://bugs.chromium.org/p/chromium/issues/detail?id=715363
https://bugs.chromium.org/p/chromium/issues/detail?id=736452

zackarychapple:
This article is amazing. https://medium.com/@marco.luethy/running-headless-chrome-on-aws-lambda-fa82ad33a9eb

[...]

Issue #469, opened 2016-08-03
Request: Support for tmpfs parameter when launching containers


CameronGo:
There is now support in the docker engine for a tmpfs flag that mounts a tmpfs file system into the running container.  I believe this was added with docker v1.10.0.  I would very much like to use this as a mechanism to store config secrets (such as PGP keys, DB passwords, etc...)  I do not see a way currently to pass this option in to the ECS TaskDefinition.  The syntax for the docker run command with this parameter looks like this:

```
docker run -ti --tmpfs /secrets:rw,noexec,nosuid,size=100m ubuntu:14.04 bash

```

ceharris:
While the use of tmpfs mounts for secrets is also interesting, our main use case for this feature would be to allow containers running under ECS to more easily run in read-only mode; e.g. by mounting `tmpfs` filesystems on `/tmp`, `/var/run`, etc.

Overbryd:
This would be highly appreciated!

I would like to run some containers **read only** and use **--tmpfs** to give them some read/write space they can use (it is ephermal data anyways). This would **substantially** lower the pressure on EBS io.

+1

raygunsix:
+1

rhuddleston:
+1 I would also like to run read only and need this to move forward with that

cyrilchampier:
+1, nice to have

ceharris:
Running containers with a read-only root filesystem is now part of the CIS benchmark, and thus it won't be long before organizations start insisting on having this Docker functionality when running on ECS.

philenz:
+1 such a simple and effective security option to use read-only with tmpfs

osiegmar:
+1

nomnux:
+1

Prendo93:
+1

seren:
This would be useful for us as well

jhmartin:
A specific use case for this is the storage of the [Varnish VSM file](https://varnish-cache.org/docs/trunk/reference/vsm.html), where Varnish is very sensitive to slowdowns in writes to this temporary file. Mounting this directory into tmpfs would insulate it against EBS latencies.

piotrbulinski:
In some cases, it's sufficient with using /dev/shm (which is tmpfs and has 64M by default).

jhmartin:
@piotrbulinski I looked at that, but unfortunately my Varnish VSM file is about 80MB.

piotrbulinski:
[...]

Issue #549, opened 2016-10-04
Containers need additional storage


blaines:
I'm looking in to how to modify the `dm.basesize` setting. It's not incredibly clear where to push these kinds of configuration settings. Could someone please follow up with a stable place to modify some storage defaults? Thanks!

I'm seeing related settings in /etc/sysconfig/docker-storage-setup, /etc/sysconfig/docker-storage, and other files.

More info about `dm.basesize` can be found here: https://docs.docker.com/v1.10/engine/reference/commandline/daemon/

```
[ec2-user@debug ~]$ cat /etc/sysconfig/docker-storage
DOCKER_STORAGE_OPTIONS="--storage-driver devicemapper --storage-opt dm.thinpooldev=/dev/mapper/docker-docker--pool --storage-opt dm.use_deferred_removal=true --storage-opt dm.use_deferred_deletion=true --storage-opt dm.fs=ext4"
[ec2-user@debug ~]$ cat /etc/sysconfig/docker-storage-setup
# Edit this file to override any configuration options specified in
# /usr/lib/docker-storage-setup/docker-storage-setup.
#
# For more details refer to "man docker-storage-setup"
DEVS=/dev/xvdcz
VG=docker
DATA_SIZE=99%FREE
AUTO_EXTEND_POOL=yes
LV_ERROR_WHEN_FULL=yes
EXTRA_DOCKER_STORAGE_OPTIONS="--storage-opt dm.fs=ext4"
[ec2-user@debug ~]$
```

richardpen:
@blaines ECS Optimized Ami reads the information in `/etc/sysconfig/docker-storage-setup` and write configuration to the file  "/etc/sysconfig/docker-storage" which will be used for docker daemon. 

You need add the options in "/etc/sysconfig/docker" which docker use to config the storage. As the change requires docker to restart if docker is already running, ECS Optimized AMI start the docker automatically when it start, it may not be safe to restart docker in your normal user data(at which time docker may not have finished initializing). One way you can do is to use the [multi-part cloud-init](http://cloudinit.readthedocs.io/en/latest/topics/format.html) to add the option before docker daemon starts, in which you put your user data like this:
```
Content-Type: multipart/mixed; boundary="===============BOUNDARY=="
MIME-Version: 1.0

--===============BOUNDARY==
MIME-Version: 1.0
Content-Type: text/x-shellscript; charset="us-ascii"
Content-Transfer-Encoding: 7bit
[...]

PR #1149, opened 2017-12-22
Adding proposal for shmsize and tmpfs flags support

sharanyad:
### Summary
Adds a proposal for https://github.com/aws/amazon-ecs-agent/issues/469 and https://github.com/aws/amazon-ecs-agent/issues/787

### Implementation details
NA

### Testing
NA

NA

### Licensing

### DIFF
--- a/proposals/shmsize-tmpfs.md
+++ b/proposals/shmsize-tmpfs.md
@@ -1,0 +1,72 @@
+<!--
+Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License"). You may
+not use this file except in compliance with the License. A copy of the
+License is located at
+
+     http://aws.amazon.com/apache2.0/
+
+or in the "license" file accompanying this file. This file is distributed
+on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+express or implied. See the License for the specific language governing
+permissions and limitations under the License.
+-->
+
+# Support for ShmSize and Tmpfs Docker Parameters in ECS
+
+## Background
+
+Docker supports the parameters `--shm-size` and `--tmpfs` which sets the size of
+`/dev/shm` volume and creates an empty tmpfs mount within a container respectively.
+More information on these flags:
+* ShmSize: https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources
+* Tmpfs: https://docs.docker.com/engine/reference/run/#tmpfs-mount-tmpfs-filesystems
+
+## ECS Support for the flags
+
+### ShmSize
+A new field will be added for Docker's --shm-size flag under the linuxParameters
+field in ContainerDefinition of the Task definition. Through this field, customers
+will be able to set the value for `ShmSize` parameter in Docker's HostConfig that
+is passed during create container call.
+
+### Tmpfs
+A new field will be added for Docker's --tmpfs flag under the linuxParameters field
+in ContainerDefinition of the Task definition.  Through this field, customers will
+be able to set the value for `Tmpfs` parameter in Docker's HostConfig that is passed
+during the create container call.
+
+Tmpfs parameter takes in a container path as the mount point and other mount options
+for tmpfs including the mount size. Other options can be found in the man page for
+mount [here](http://man7.org/linux/man-pages/man8/mount.8.html).
+
+## Implications of using these fields during task placement decisions:
+
+Since both the flags consume memory in the instance in which their corresponding
+tasks are placed, their memory sizes will be accounted against the total memory
+available in the instance. As a result, the memory set for both the ShmSize and
+Tmpfs parameters will affect placement decisions for other tasks in the same instances.
+Size will be a mandatory parameter in the Tmpfs field since it is essential for
+making task placement decisions.
+
+Memory reservation metric would now include the `shmsize` and size of `tmpfs` mounts added
+to the containers through the task definition. The memory utilization metric remains
+the same as it is today.
+Example:
+If 80 MiB of container memory is allocated through `memory` parameter and 20 MiB
+for `ShmSize`, then the reservation metric would reflect the total memory reserved
+as 100 MiB. Suppose the container uses the entire 80 MiB of container memory and
+none of the shared memory, the utilization metrics would still show the usage as
+100%, since it is not affected by the memory reserved/used for shared memory (`ShmSize`).
+
+## Validation of the new fields:
+
+* ShmSize: This should be a non negative integer value in MiB.
+* Tmpfs:
+    * Container Path: This should be an absolute file path and will be a required
+    field. This path should not conflict with any of the container mount paths for
+    volumes.
+    * Size: Non negative integer value in MiB. This will be a required field.
+    * Mount options: This should be one of the mount options for `tmpfs` allowed
+    by Docker (except size).


proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,46 @@
...
+on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+express or implied. See the License for the specific language governing
+permissions and limitations under the License.
+-->
+
+# Support for ShmSize and Tmpfs Docker Parameters in ECS
+
+## Background
+
+Docker supports the parameters --shm-size and --tmpfs which sets the size of /dev/shm volume and creates an empty tmpfs mount within a container respectively. More information on these flags:
```
aaithal: it would be good to quote `--shm-size` and `--tmpfs`. Also `/dev/shm`

proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,46 @@
...
+* ShmSize: https://docs.docker.com/engine/reference/run/#runtime-constraints-on-resources
+* Tmpfs: https://docs.docker.com/engine/reference/run/#tmpfs-mount-tmpfs-filesystems
+
+## ECS Support for the flags
```
adnxn: should we include a snippet of the proposed changes to the task definition? examples of what the `shm-size` and `tmpfs` fields in the task definition.

proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,46 @@
...
+
+CloudWatch Metrics would now include additional metrics on memory reserved for these parameters.
+
+## Validation of the new fields:
+
+* ShmSize: This should be a non negative integer value in MiB.
+* Tmpfs:
+    * Container Path: This should be an absolute file path and will be a required field.
+    * Size: Non negative integer value in MiB. This will be a required field.
+    * Mount options: This should be one of the mount options allowed by Linux's mount command(except size). The list can be found in man [page](http://man7.org/linux/man-pages/man8/mount.8.html) of mount command.
```
adnxn: is it *exactly* the options allowed by Linux's mount command? or does Docker only handle a subset? are we also constrained by the set of options that are supported by the docker client in the agent?

sharanyad: It seems like Docker accepts only a subset of the `mount` options. We should probably support only those. 
go-dockerclient does not perform any validation of the mount options passed for `tmpfs`.

proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,46 @@
...
+on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+express or implied. See the License for the specific language governing
+permissions and limitations under the License.
+-->
+
+# Support for ShmSize and Tmpfs Docker Parameters in ECS
+
+## Background
+
+Docker supports the parameters `--shm-size` and `--tmpfs` which sets the size of `/dev/shm` volume and creates an empty tmpfs mount within a container respectively. More information on these flags:
```
samuelkarp: Can you wrap all the lines in this doc at either 80 or 120 characters?

proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,46 @@
...
+### Tmpfs
+A new field will be added for Docker's --tmpfs flag under the linuxParameters field in ContainerDefinition of the Task definition.  Through this field, customers will be able to set the value for `Tmpfs` parameter in Docker's HostConfig that is passed during the create container call.
+
+Tmpfs parameter takes in a container path as the mount point and other mount options for tmpfs including the mount size. Other options can be found in the man page for mount [here](http://man7.org/linux/man-pages/man8/mount.8.html).
+
+## Implications of using these fields during task placement decisions:
+
+Since both the flags consume memory in the instance in which their corresponding tasks are placed, their memory sizes will be accounted against the total memory available in the instance. As a result, the memory set for both the ShmSize and Tmpfs parameters will affect placement decisions for other tasks in the same instances. Size will be a mandatory parameter in the Tmpfs field since it is essential for making task placement decisions.
+
+CloudWatch Metrics would now include additional metrics on memory reserved for these parameters.
```
samuelkarp: Are you just talking about the reserved memory metric?  Does the utilization metric also get affected by this or does utilization just cover Memory/MemoryReservation fields in the task def?

proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Validation of the new fields:
+
+* ShmSize: This should be a non negative integer value in MiB.
+* Tmpfs:
+    * Container Path: This should be an absolute file path and will be a required field.
```
samuelkarp: Do we need to validate that this doesn't conflict with any of the volume mount paths?

proposals/shmsize-tmpfs.md
sharanyad: It is just the reservation metrics for both the new parameters.

samuelkarp: Here's a hypothetical situation: If I have 80 MiB allocated through the "memory" parameter (and I'm using all of it) and I have 20 MiB through ShmSize, what do I see?

I expect that reservation metrics would show 100 MiB (80 + 20).
I'm not sure whether utilization metrics would show 100% (80 / 80) or 80% (80 / 100).

sharanyad: In this case, the utilization metrics would still show 100%. If we explicitly specify that the utilization metrics do not include ShmSize and Tmpfs size consumed, it would make sense.

proposals/shmsize-tmpfs.md
sharanyad: This is a good point. I think we need to have this validation.

proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,65 @@
...
+## Implications of using these fields during task placement decisions:
+
+Since both the flags consume memory in the instance in which their corresponding
+tasks are placed, their memory sizes will be accounted against the total memory
+available in the instance. As a result, the memory set for both the ShmSize and
+Tmpfs parameters will affect placement decisions for other tasks in the same instances.
+Size will be a mandatory parameter in the Tmpfs field since it is essential for
+making task placement decisions.
+
+CloudWatch Metrics would now include additional metrics on memory reserved for these
```
poojamaiya: Summarizing offline discussion - I think we'd want to introduce new reservation metrics only if we can have corresponding utilization metrics. If we're not going to track Shm and Tmpfs utilization independently, it might make sense to update existing metrics and not introduce new metrics.

samuelkarp: I don't think we want to create a new metric.  However, I do think we need to include shm and tmpfs memory in the existing reservation metric.  Customers use the reservation metric as a triggering metric in auto scaling policies; if we don't include shm and tmpfs memory, customers will be unable to use this metric as a way to accurately indicate the amount of available memory for future tasks in their clusters.

aaithal: `Tmpfs` and `/dev/shm` get used by processes as filesystem volumes that can be written to and not as their run-time memory and as such are treated differently by the kernel. For example, the OOM killer is not invoked if you try to create/extend a file beyond the tmfs volume size that was allocated as that's associated with the memory cgroup (specified by the `--memory` parameter). For these reasons, it doesn't make much sense to account for these in utilization metrics. However, as these will have an impact on task placement, we should definitely account for these in reservation metrics as pointed out by @samuelkarp's comment.

poojamaiya: Makes sense, thanks!

proposals/shmsize-tmpfs.md
```diff
@@ -0,0 +1,66 @@
...
+Since both the flags consume memory in the instance in which their corresponding
+tasks are placed, their memory sizes will be accounted against the total memory
+available in the instance. As a result, the memory set for both the ShmSize and
+Tmpfs parameters will affect placement decisions for other tasks in the same instances.
+Size will be a mandatory parameter in the Tmpfs field since it is essential for
+making task placement decisions.
+
+Memory reservation metric would now include the `shmsize` and size of `tmpfs` mounts added
+to the containers through the task definition. The memory utilization metric remains
+the same as it is today.
```
samuelkarp: Can you add an illustrative example to this doc?  I'd like something like https://github.com/aws/amazon-ecs-agent/pull/1149#discussion_r158569602 to be here so that we make it explicit.

sharanyad: Done!

lbod:
@sharanyad  I'd like to ask how long it takes for a proposal like this to actually hit production and AWS update their ECS agents? 

If it's a rough estimate that's ok, we're currently unable to reliably run Chrome on ECS and looking at workarounds till this lands.

sharanyad:
@lbod We understand your use case for this feature in ECS. We generally don't comment on the ETA of our future roadmap. We are actively working on this feature and shall notify as soon as it’s released. Thanks for the patience.

lbod:
@sharanyad thanks for the prompt reply. Without knowing an ETA then we have to consider workarounds. 

Am I correct thinking we could map an additional volume in ECS from the container to the host i.e. `/dev/shm:/local/mount/directory` which would get round the limitation?

richardpen:
@lbod If you read the comments in #787, there are several workarounds for this:
* Mount the `/dev/shm` in the container
* Modify the default shm size of docker `--default-shm-size`
* Run container as privileged and mount the /dev/shm inside the container

But these may not be the ideal way as `--shm-size` to share the memory with the container. Again we are working hard to make this happen, will let you know whenever we have any updates.

Thanks

lbod:
@richardpen Thanks for the reply, it's why I asked about mounting /dev/shm in the last reply.

I did miss the `--default-shm-size` comment in #787 , is that just another argument to the agent? I can't see any docs on it. The effort and help are both appreciated.

richardpen:
@lbod `--default-shm-size` is an argument for docker daemon, you can find some information [here](https://github.com/moby/moby/pull/29692). if you are using ECS Optimized AMI, you can add this in `/etc/sysconfig/docker`, and then restart the docker:
```
OPTIONS="${OPTIONS} --default-shm-size 10240"
```

If you want to do this in your user data, you need to modify this before docker daemon start use multi part cloud init as an example [here](https://github.com/aws/amazon-ecs-agent/issues/549#issuecomment-251517887):
```
Content-Type: multipart/mixed; boundary="===============BOUNDARY=="
MIME-Version: 1.0

--===============BOUNDARY==
MIME-Version: 1.0
Content-Type: text/x-shellscript; charset="us-ascii"
Content-Transfer-Encoding: 7bit
Content-Disposition: attachment; filename="standard_userdata.txt"

#!/bin/bash
echo "normal user data script part"


--===============BOUNDARY==
MIME-Version: 1.0
Content-Type: text/cloud-boothook; charset="us-ascii"
Content-Transfer-Encoding: 7bit
Content-Disposition: attachment; filename="boothook.txt"

#cloud-boothook
echo 'OPTIONS="${OPTIONS} --default-shm-size 10240"' >> /etc/sysconfig/docker

--===============BOUNDARY==--
```

lbod:
@richardpen We're on the ECS AMI so this is great to know, our user data is getting bigger but we'll have a few kb left. 
**It'd appear currently this isn't a blocker for the majority of people and just needs more official docs?** 
I'd also like to try /dev/shm mount because it seems a bit less config, would we have to factor in the extra shm size when we size the ECS instance (it's maybe not important in our case as we're trying to only run one task at a time)?

jonreiter:
i have some containers that run out of batch in managed compute environments and need shm-size set.  is it possible to get these new parameters added to the batch job definition in addition to the ecs task definition?  or to expose (some) ecs container controls upstream in batch?

currently i need to run separate managed compute environments with modified ecs amis that set shm-size as required.  and i need a bunch of job queues to keep the environments separate.  i can see the batch tasks running in ecs...surely the container definitions exist somewhere under the covers.

PR #1168, opened 2018-01-08
Log the error when unmarshal task from acs failed

richardpen:
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/acs/handler/payload_handler.go
+++ b/agent/acs/handler/payload_handler.go
@@ -179,7 +179,7 @@ func (payloadHandler *payloadRequestHandler) addPayloadTasks(payload *ecsacs.Pay
 	validTasks := make([]*api.Task, 0, len(payload.Tasks))
 	for _, task := range payload.Tasks {
 		if task == nil {
-			seelog.Criticalf("Received nil task for messageId: %s", *payload.MessageId)
+			seelog.Criticalf("Received nil task for messageId: %s", aws.StringValue(payload.MessageId))
 			allTasksOK = false
 			continue
 		}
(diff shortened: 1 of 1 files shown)


agent/acs/handler/payload_handler.go
```diff
@@ -325,6 +325,8 @@ func isTaskStatusNotStopped(status api.TaskStatus) bool {
 // handleUnrecognizedTask handles unrecognized tasks by sending 'stopped' with
 // a suitable reason to the backend
 func (payloadHandler *payloadRequestHandler) handleUnrecognizedTask(task *ecsacs.Task, err error, payload *ecsacs.PayloadMessage) {
+	seelog.Warnf("Received unexpected acs message, messageID: %s, task: %v, err: %v", *payload.MessageId, task, err)
```
aaithal: please use aws.StringValue for printing message id. Also, let's not print the entire task here. Can you just print just the arn instead?

sharanyad: I think task arn can be nil here.

aaithal: `aws.StringValue(task.Arn)` should cover that.

agent/acs/handler/payload_handler.go
```diff
@@ -325,8 +325,11 @@ func isTaskStatusNotStopped(status api.TaskStatus) bool {
 // handleUnrecognizedTask handles unrecognized tasks by sending 'stopped' with
 // a suitable reason to the backend
 func (payloadHandler *payloadRequestHandler) handleUnrecognizedTask(task *ecsacs.Task, err error, payload *ecsacs.PayloadMessage) {
+	seelog.Warnf("Received unexpected acs message, messageID: %s, task: %v, err: %v",
+		aws.StringValue(payload.MessageId), aws.StringValue(task.Arn), err)
+
 	if task.Arn == nil {
```
aaithal: Let's also modify this to be `if aws.StringValue(task.Arn) == ""`. That is a more complete check, that'll check both for the field not being set (`nil`) and set to an empty string.

richardpen:
@aaithal Removed the `amazon-ecs-cni-plugins` from this PR.

agent/acs/handler/payload_handler.go
```diff
@@ -325,8 +325,11 @@ func isTaskStatusNotStopped(status api.TaskStatus) bool {
 // handleUnrecognizedTask handles unrecognized tasks by sending 'stopped' with
 // a suitable reason to the backend
 func (payloadHandler *payloadRequestHandler) handleUnrecognizedTask(task *ecsacs.Task, err error, payload *ecsacs.PayloadMessage) {
-	if task.Arn == nil {
-		seelog.Criticalf("Received task with no arn, messageId: %s, task: %v", *payload.MessageId, task)
+	seelog.Warnf("Received unexpected acs message, messageID: %s, task: %v, err: %v",
+		aws.StringValue(payload.MessageId), aws.StringValue(task.Arn), err)
+
+	if aws.StringValue(task.Arn) == "" {
+		seelog.Criticalf("Received task with no arn, messageId: %s, task: %v", aws.StringValue(payload.MessageId), aws.StringValue(task.Arn))
```
aaithal: nit: we don't need to print `task.Arn` here since we know it's not set.

PR #1171, opened 2018-01-09
eventhandler: add case for taskShouldBeSent (missing exit code fix)

adnxn:
### Summary

This change adds a new condition for determining whether a task state
change event should be sent to the backend. If a task state change is in
STOPPED and does not contain any container state change events, we do
not do not send it to the backend.

### Implementation details

Add a new check to the `taskShouldBeSent()` function

### Testing

New tests cover the changes: 

### Licensing


agent/eventhandler/handler_test.go
```diff
@@ -35,6 +35,76 @@ import (
 
 const taskARN = "taskarn"
 
+func TestDoNotSendTaskStoppedEventEmptyContainerEventsFirst(t *testing.T) {
```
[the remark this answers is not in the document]
adnxn: fixed. also added another unit test that exercises the new check in `taskShouldBeSent()` directly.

agent/eventhandler/task_handler_types.go
```diff
@@ -81,6 +81,12 @@ func (event *sendableEvent) taskShouldBeSent() bool {
 		return false
 	}
 
+	// task event should not be sent if status is STOPPED and includes no
+	// container state change events
+	if tevent.Status == api.TaskStopped && len(tevent.Containers) == 0 {
```
richardpen: If I understand the code correctly, this won't work if the task moved directly to stopped eg: have a circular dependency in the task definition. Please correct me if I'm wrong.

aaithal: @richardpen  What are some of the other scenarios where tasks can transition directly to STOPPED?

richardpen: Currently, the one that we have in code is the circular dependency [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine.go#L533).

aaithal: right. but, what are the other scenarios in which this kind of thing can happen? is that the only one?

richardpen: Checking the code for `emitTaskEvent`, looks like this is the only one scenario today.

Issue #681, opened 2017-01-26
Update to Go 1.8


samuelkarp:
Go 1.8 will be out soon and we should update.

The AWS SDK for Go will also [need to be updated](https://aws.amazon.com/blogs/developer/aws-sdk-for-go-update-needed-for-go-1-8/).

samuelkarp:
Go 1.8 [is out](https://blog.golang.org/go1.8).

samuelkarp:
We may need https://github.com/Microsoft/go-winio/pull/48.

cabbruzzese:
Saw https://github.com/Microsoft/go-winio/issues/41 when building in go 1.8. We should include https://github.com/Microsoft/go-winio/pull/48

samuelkarp:
Go 1.9 [is out](https://blog.golang.org/go1.9).

yunhee-l:
is there something I need to do for using go-winio#48? (re: pr)

samuelkarp:
@yunhee-l You'll want to make sure our vendored version of go-winio has that commit.  I'd probably just recommend updating it to the most recent version.

Same thing for the AWS SDK for Go.

yunhee-l:
> ✗ dep status
PROJECT                                 CONSTRAINT     VERSION        REVISION  LATEST   PKGS USED
...
github.com/Microsoft/go-winio           *              v0.4.7         7da180e   7da180e  1   
...
github.com/aws/aws-sdk-go               ^1.12.67       v1.12.67       decd990   decd990  33  
...

Both are latest of their respective repo/branch.

PR #1172, opened 2018-01-10
Adding gocyclo to travis config, excluding test and mock files

yunhee-l:
### Summary
Adding gocyclo to travis ci config, to run as a part of required checks for any commits. 

### Implementation details
Adding gocyclo installation and run of it to fail on any files with cyclomatic complexity over 15, excluding test and mock files, and vender/ testutils/ directories. 

### Testing

Tested on my forked repo:
```
$ make static-check
gocyclo -over 15 <go files as filtered by go list command>
The command "make static-check" exited with 0.
```

Did not update the changelog, not a feature implementation or bug fix that are customer facing.

### Licensing

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -243,16 +243,8 @@ func environmentConfig() (Config, error) {
 	engineAuthType := os.Getenv("ECS_ENGINE_AUTH_TYPE")
 	engineAuthData := os.Getenv("ECS_ENGINE_AUTH_DATA")
 
-	var checkpoint bool
 	dataDir := os.Getenv("ECS_DATADIR")
-	if dataDir != "" {
-		// if we have a directory to checkpoint to, default it to be on
-		checkpoint = utils.ParseBool(os.Getenv("ECS_CHECKPOINT"), true)
-	} else {
-		// if the directory is not set, default to checkpointing off for
-		// backwards compatibility
-		checkpoint = utils.ParseBool(os.Getenv("ECS_CHECKPOINT"), false)
-	}
+	checkpoint := getCheckpoint(dataDir)
 
 	// Format: json array, e.g. [1,2,3]
 	reservedPortEnv := os.Getenv("ECS_RESERVED_PORTS")
@@ -286,13 +278,7 @@ func environmentConfig() (Config, error) {
 
 	reservedMemory := parseEnvVariableUint16("ECS_RESERVED_MEMORY")
 
-	var dockerStopTimeout time.Duration
-	parsedStopTimeout := parseEnvVariableDuration("ECS_CONTAINER_STOP_TIMEOUT")
-	if parsedStopTimeout >= minimumDockerStopTimeout {
-		dockerStopTimeout = parsedStopTimeout
-	} else if parsedStopTimeout != 0 {
-		seelog.Warnf("Discarded invalid value for docker stop timeout, parsed as: %v", parsedStopTimeout)
-	}
+	dockerStopTimeout := getDockerStopTimeout()
 
 	taskCleanupWaitDuration := parseEnvVariableDuration("ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION")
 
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -5,5 +5,10 @@ go:
 go_import_path: github.com/aws/amazon-ecs-agent
 
 install: make get-deps
+before_script:
+  # All the .go files in the agent, excluding vendor/ and testutils/ directories, and all *_test.go and *_mocks.go files
+  - GO_FILES=$(find ./agent -iname '*.go' -type f | grep -v /vendor/ | grep -v /testutils/ | grep -v _test\.go$ | grep -v _mocks\.go$)
```
aaithal: instead of `find`, it would be better to `go list ./agent/...`. Example: https://github.com/aws/amazon-ecs-agent/blob/73758e34dfc45d98ec6c3aaf623dd3bd8d6572b7/Makefile#L85

.travis.yml
```diff
@@ -5,5 +5,10 @@ go:
...
+  - go get github.com/fzipp/gocyclo
```
aaithal: it's better to do this under `get-deps` imo. We already grab a bunch of tools under that target amd `install` is already invoking it. wdyt?

.travis.yml
```diff
@@ -5,5 +5,10 @@ go:
 go_import_path: github.com/aws/amazon-ecs-agent
 
 install: make get-deps
+before_script:
```
aaithal: Instead of having this just in travis, how about we add a `Makefile` target for running cyclo? Then, we can run it locally and in travis by just invoking that target. If we added something like `make static-analysis`, which invoked `go-cyclo` today and other tools in the future, all we need to do is just wire that in here.

yunhee-l: You are right, I should've thought of using gocyclo in development context, I focused only on adding to the PR/commit checks. Will fix as suggested.

.travis.yml
yunhee-l: "go list" lists the packages named by the import paths, but gocyclo needs local file/directory of go files. Looking through go list options to see if there's any I can use.

aaithal: yeah, that's not super optimal. I just assumed it'd be aware of packages.  I didn't want us to use `find` as the tool might not always be available in all environments. If it ran successfully in travis, it's probably fine for now.

yunhee-l: Bit messy but found a way to use "go list" instead of "find"

.travis.yml
yunhee-l: Fixed.

Makefile
```diff
@@ -165,10 +165,18 @@ taskmetadata-validator:
 ecr-execution-role-image:
 	$(MAKE) -C misc/ecr $(MFLAGS)
 
+gocyclo:
+    # Run gocyclo over all .go files in the agent, excluding vendor/ and testutils/ directories, and all *_test.go and *_mocks.go files
+	gocyclo -over 15 $(shell go list -f "{{$p := .}}{{range $f := .GoFiles}}{{$p.Dir}}/{{$f}} {{end}}" ./agent/... | grep -v /vendor/ | grep -v /testutils/ | grep -v _test\.go$ | grep -v _mocks\.go$)
```
petderek: why 15 and not lower?

.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: go
 sudo: false
 go:
-  - 1.7
+  - 1.9
```
richardpen: Is `1.9` required by gocyclo? As the agent currently is built with go  `1.7`, we probably should also update it to build with `1.9`.

samuelkarp: https://github.com/aws/amazon-ecs-agent/issues/681 has some of the work necessary for us to upgrade.  Beyond that, 72056b389af566c0ae2bec675cde85369cc004dd was the last time that we upgraded and shows a bit more of the actual change.

Makefile
yunhee-l: goreportcard reports ones over 15. I can make it lower but lot more work needs to be done to refactor more code.

.travis.yml
yunhee-l: This was me testing if the later go version will work with my use of "go list" command in makefile. It seems like it's a quotation using " or ' that's causing the problem, I'm going to revert this back to 1.7 and fix the command.

yunhee-l:
Fixed the go list command in gocyclo makefile target, make static-check succeeds now.

PR #1173, opened 2018-01-10
Generate binaries and tests using AWS Codebuild

petderek:
### Summary 
Provides base layer of infrastructure for cross-compiling windows binaries on linux. This is only used in testing for now, but its in a good place where we could extend this to automate builds.

### Implementation
This provides the following:
- A makefile target that does the following:
  - compiles linux tarball and cni plugins
  - compiles windows zipfile
  - compiles all functional and integration test binaries
- A codebuild buildspec used to automate the compile step
- Some fixes to the build tooling to make sure that 'root' doesn't own the artifacts

When run with `make codebuild`, your out directory will result:

```
out
├── amazon-ecs-agent
├── amazon-ecs-agent.exe
├── amazon-ecs-pause.tar
├── cni-plugins
│   ├── ecs-bridge
│   ├── ecs-eni
│   └── ecs-ipam
├── ecs-agent.tar
├── ecs-agent-windows.zip
└── test-artifacts
    ├── unix-app-tests
    ├── unix-engine-tests
    ├── unix-handwritten-tests
    ├── unix-simple-tests
    ├── unix-stats-tests
    ├── windows-app-tests.exe
    ├── windows-engine-tests.exe
    ├── windows-handwritten-tests.exe
    ├── windows-simple-tests.exe
    └── windows-stats-tests.exe
```

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,18 +11,9 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
-PAUSE_CONTAINER_IMAGE = "amazon/amazon-ecs-pause"
-PAUSE_CONTAINER_TAG = "0.1.0"
-PAUSE_CONTAINER_TARBALL = "amazon-ecs-pause.tar"
+USERID=$(shell id -u)
 
-# Variable to determine branch/tag of amazon-ecs-cni-plugins
-ECS_CNI_REPOSITORY_REVISION=master
-
-# Variable to override cni repository location
-ECS_CNI_REPOSITORY_SRC_DIR=$(PWD)/amazon-ecs-cni-plugins
-
-
-.PHONY: all gobuild static docker release certs test clean netkitten test-registry run-functional-tests benchmark-test gogenerate run-integ-tests pause-container get-cni-sources cni-plugins
+.PHONY: all gobuild static docker release certs test clean netkitten test-registry run-functional-tests benchmark-test gogenerate run-integ-tests pause-container get-cni-sources cni-plugins test-artifacts
 
 all: docker
 
@@ -31,40 +22,55 @@ all: docker
 gobuild:
 	./scripts/build false
 
+# create output directories
+.out-stamp:
+	mkdir -p ./out/test-artifacts ./out/cni-plugins
+	touch .out-stamp
+
 # Basic go build
 static:
 	./scripts/build
 
+BUILDER_IMAGE="amazon/amazon-ecs-agent-build:make"
+.builder-image-stamp: scripts/dockerfiles/Dockerfile.build
+	@docker build -f scripts/dockerfiles/Dockerfile.build -t $(BUILDER_IMAGE) .
+	touch .builder-image-stamp
+
 # 'build-in-docker' builds the agent within a dockerfile and saves it to the ./out
 # directory
-build-in-docker:
-	@docker build -f scripts/dockerfiles/Dockerfile.build -t "amazon/amazon-ecs-agent-build:make" .
+# TODO: make this idempotent
+build-in-docker: .builder-image-stamp .out-stamp
 	@docker run --net=none \
-	  -e TARGET_OS="${TARGET_OS}" \
-	  -e LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
-	  -X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
-	  -v "$(PWD)/out:/out" \
-	  -v "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
-	  "amazon/amazon-ecs-agent-build:make"
+		--env TARGET_OS="${TARGET_OS}" \
+		--env LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
+			-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
+		--volume "$(PWD)/out:/out" \
+		--volume "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
+		--user "$(USERID)" \
+		--rm \
+		$(BUILDER_IMAGE)
 
 # 'docker' builds the agent dockerfile from the current sourcecode tree, dirty
 # or not
-docker: certs build-in-docker pause-container-release cni-plugins
+docker: certs build-in-docker pause-container-release cni-plugins .out-stamp
 	@cd scripts && ./create-amazon-ecs-scratch
 	@docker build -f scripts/dockerfiles/Dockerfile.release -t "amazon/amazon-ecs-agent:make" .
 	@echo "Built Docker image \"amazon/amazon-ecs-agent:make\""
 
 # 'docker-release' builds the agent from a clean snapshot of the git repo in
 # 'RELEASE' mode
-docker-release: pause-container-release cni-plugins
+# TODO: make this idempotent
+docker-release: pause-container-release cni-plugins .out-stamp
 	@docker build -f scripts/dockerfiles/Dockerfile.cleanbuild -t "amazon/amazon-ecs-agent-cleanbuild:make" .
 	@docker run --net=none \
-	  -e TARGET_OS="${TARGET_OS}" \
-	  -v "$(PWD)/out:/out" \
-	  -e LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
-	  -X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
-	  -v "$(PWD):/src/amazon-ecs-agent" \
-	  "amazon/amazon-ecs-agent-cleanbuild:make"
+		--env TARGET_OS="${TARGET_OS}" \
+		--env LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
+			-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
+		--user "$(USERID)" \
+		--volume "$(PWD)/out:/out" \
+		--volume "$(PWD):/src/amazon-ecs-agent" \
+		--rm \
+		"amazon/amazon-ecs-agent-cleanbuild:make"
 
 # Release packages our agent into a "scratch" based dockerfile
 release: certs docker-release
@@ -90,6 +96,80 @@ test-silent:
 benchmark-test:
 	. ./scripts/shared_env && go test -run=XX -bench=. $(shell go list ./agent/... | grep -v /vendor/)
 
+define dockerbuild
+	docker run \
+		--net none \
+		--user "$(USERID)" \
+		--volume "$(PWD)/out:/out" \
+		--volume "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
+		--rm \
+		$(BUILDER_IMAGE) \
+		$(1)
+endef
+
+define win-cgo-dockerbuild
+	docker run --net=none \
+		--user "$(USERID)" \
+		--volume "$(PWD)/out:/out" \
+		--volume "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
+		--env "GOOS=windows" \
+		--env "CGO_ENABLED=1" \
+		--env "CC=x86_64-w64-mingw32-gcc" \
+		--rm \
+		$(BUILDER_IMAGE) \
+		$(1)
+endef
+
+# TODO: use `go list -f` to target the test files more directly
+ALL_GO_FILES = $(shell find . -name "*.go" -print | tr "\n" " ")
+GO_INTEG_TEST = go test -race -tags integration -c -o
+out/test-artifacts/linux-engine-tests: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call dockerbuild,$(GO_INTEG_TEST) $@ ./agent/engine)
+
+out/test-artifacts/linux-app-tests: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call dockerbuild,$(GO_INTEG_TEST) $@ ./agent/app)
+
+out/test-artifacts/linux-stats-tests: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call dockerbuild,$(GO_INTEG_TEST) $@ ./agent/stats)
+
+out/test-artifacts/windows-engine-tests.exe: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call win-cgo-dockerbuild,$(GO_INTEG_TEST) $@ ./agent/engine)
+
+out/test-artifacts/windows-app-tests.exe: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call win-cgo-dockerbuild,$(GO_INTEG_TEST) $@ ./agent/app)
+
+out/test-artifacts/windows-stats-tests.exe: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call win-cgo-dockerbuild,$(GO_INTEG_TEST) $@ ./agent/stats)
+
+GO_FUNCTIONAL_TEST = go test -tags functional -c -o
+out/test-artifacts/linux-simple-tests: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests/generated/simpletests_unix)
+
+out/test-artifacts/linux-handwritten-tests: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests)
+
+out/test-artifacts/windows-simple-tests.exe: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call win-cgo-dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests/generated/simpletests_windows)
+
+out/test-artifacts/windows-handwritten-tests.exe: $(ALL_GO_FILES) .out-stamp .builder-image-stamp
+	$(call win-cgo-dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests)
+
+##.PHONY: test-artifacts-windows test-artifacts-linux test-artifacts
+
+WINDOWS_ARTIFACTS_TARGETS := out/test-artifacts/windows-engine-tests.exe out/test-artifacts/windows-stats-tests.exe
+WINDOWS_ARTIFACTS_TARGETS += out/test-artifacts/windows-app-tests.exe out/test-artifacts/windows-simple-tests.exe
+WINDOWS_ARTIFACTS_TARGETS += out/test-artifacts/windows-handwritten-tests.exe
+
+LINUX_ARTIFACTS_TARGETS := out/test-artifacts/linux-engine-tests out/test-artifacts/linux-stats-tests
+LINUX_ARTIFACTS_TARGETS += out/test-artifacts/linux-app-tests out/test-artifacts/linux-simple-tests
+LINUX_ARTIFACTS_TARGETS += out/test-artifacts/linux-handwritten-tests
+
+test-artifacts-windows: $(WINDOWS_ARTIFACTS_TARGETS)
+
+test-artifacts-linux: $(LINUX_ARTIFACTS_TARGETS)
+
+test-artifacts: test-artifacts-windows test-artifacts-linux
+
 # Run our 'test' registry needed for integ and functional tests
 test-registry: netkitten volumes-test squid awscli image-cleanup-test-images fluentd taskmetadata-validator
 	@./scripts/setup-test-registry
@@ -102,12 +182,14 @@ test-in-docker:
 run-functional-tests: testnnp test-registry ecr-execution-role-image
 	. ./scripts/shared_env && go test -tags functional -timeout=30m -v ./agent/functional_tests/...
 
-testnnp:
-	$(MAKE) -C misc/testnnp $(MFLAGS)
+PAUSE_CONTAINER_IMAGE = "amazon/amazon-ecs-pause"
+PAUSE_CONTAINER_TAG = "0.1.0"
+PAUSE_CONTAINER_TARBALL = "amazon-ecs-pause.tar"
 
-pause-container:
+pause-container: .out-stamp
 	@docker build -f scripts/dockerfiles/Dockerfile.buildPause -t "amazon/amazon-ecs-build-pause-bin:make" .
 	@docker run --net=none \
+		-u "$(USERID)" \
 		-v "$(PWD)/misc/pause-container:/out" \
 		-v "$(PWD)/misc/pause-container/buildPause:/usr/src/buildPause" \
 		"amazon/amazon-ecs-build-pause-bin:make"
@@ -116,18 +198,23 @@ pause-container:
 	@docker rmi -f "amazon/amazon-ecs-build-pause-bin:make"
 
 pause-container-release: pause-container
-	mkdir -p "$(PWD)/out"
 	@docker save ${PAUSE_CONTAINER_IMAGE}:${PAUSE_CONTAINER_TAG} > "$(PWD)/out/${PAUSE_CONTAINER_TARBALL}"
+
+# Variable to determine branch/tag of amazon-ecs-cni-plugins
+ECS_CNI_REPOSITORY_REVISION=master
+
+# Variable to override cni repository location
+ECS_CNI_REPOSITORY_SRC_DIR=$(PWD)/amazon-ecs-cni-plugins
 
 get-cni-sources:
 	git submodule update --init --checkout
 
-cni-plugins: get-cni-sources
+cni-plugins: get-cni-sources .out-stamp
 	@docker build -f scripts/dockerfiles/Dockerfile.buildCNIPlugins -t "amazon/amazon-ecs-build-cniplugins:make" .
-	mkdir -p out/cni-plugins
 	docker run --rm --net=none \
 		-e GIT_SHORT_HASH=$(shell cd $(ECS_CNI_REPOSITORY_SRC_DIR) && git rev-parse --short HEAD) \
 		-e GIT_PORCELAIN=$(shell cd $(ECS_CNI_REPOSITORY_SRC_DIR) && git status --porcelain 2> /dev/null | wc -l | sed 's/^ *//') \
+		-u "$(USERID)" \
 		-v "$(PWD)/out/cni-plugins:/go/src/github.com/aws/amazon-ecs-cni-plugins/bin/plugins" \
 		-v "$(ECS_CNI_REPOSITORY_SRC_DIR):/go/src/github.com/aws/amazon-ecs-cni-plugins" \
 		"amazon/amazon-ecs-build-cniplugins:make"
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -35,10 +35,12 @@ gobuild:
 static:
 	./scripts/build
 
+.dockerbuild:
```
jahkeup: Is the intent to cache the builder image with this target? I think it would be better named as `builder-image` or something similar as a dependent target name if that's the case.

scripts/build-test-artifacts
```diff
@@ -0,0 +1,22 @@
...
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# This script generates artifacts used to test the agent. It only has windows
+# functional tests, but will eventaully be extended to support others.
+
+mkdir -p ./out/test-artifacts
+
+GOOS=windows go test -tags functional -o ./out/test-artifacts/windows-simple-tests.exe -c ./agent/functional_tests/tests/generated/simpletests_windows/
```
jahkeup: We should to build all the test artifacts or take in a GOOS to build for. When we are ready to, we can take advantage of all the test artifacts built.

petderek: > take in a GOOS

The problem with this is that the output files are going to be different (.exe) and I'd rather be explicit instead of switching on the supplied operating system.

> build all the test artifacts

I went back and forth on whether or not to do this yet &mdash;   you're the second person to ask me about it so I'll go ahead and do that :smiley:

Makefile
samuelkarp: Why `.`?

buildspec.yml
```diff
@@ -0,0 +1,17 @@
...
+phases:
+  build:
+    commands:
+      - make test-artifacts-in-docker
+  post_build:
+    commands:
+      - cp out/test-artifacts/*.exe .
+artifacts:
+  files:
+      - 'agent.exe'
```
samuelkarp: I think instead of doing `agent.exe` and `hostsetup.ps1` like this, can you just use the [zip file](https://github.com/aws/amazon-ecs-agent/blob/v1.16.1/scripts/stage#L74-L91) (the logic will probably need to be extracted from there, and then you can do it in `post_build`)?

scripts/build-test-artifacts
```diff
@@ -0,0 +1,22 @@
...
+GOOS=windows go test -tags functional -o ./out/test-artifacts/windows-handwritten-tests.exe -c ./agent/functional_tests/tests/
+GOOS=windows go build -o out/test-artifacts/agent.exe ./agent
```
samuelkarp: I'd prefer to use the same [build script](https://github.com/aws/amazon-ecs-agent/blob/v1.16.1/scripts/build) for the actual agent that we use elsewhere.

Makefile
petderek: I had thought .target means 'target' is usually executed as a dependency of another target. I can change this.

samuelkarp: I haven't seen this convention, but @nmeyerhans might know more than me about this.

buildspec.yml
petderek: I was letting codebuild do the zip logic for me (this preference is stored in the project settings rather than the buildspec).

I'm inclined to continue letting codebuild zip, but that's only because I don't want to write/copy code.

buildspec.yml
```diff
@@ -0,0 +1,12 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - make codebuild 
+artifacts:
+  files:
+      - 'out/**/*'
+      - 'misc/windows-deploy/hostsetup.ps1'
```
samuelkarp: I think instead of doing `agent.exe` and `hostsetup.ps1` like this, can you just use the [zip file](https://github.com/aws/amazon-ecs-agent/blob/v1.16.1/scripts/stage#L74-L91) (the logic will probably need to be extracted from there, and then you can do it in `post_build`)?

Makefile
```diff
@@ -90,6 +92,22 @@ test-silent:
...
+	go test -tags functional -o ./out/test-artifacts/unix-simple-tests -c ./agent/functional_tests/tests/generated/simpletests_unix/
+	go test -tags functional -o ./out/test-artifacts/unix-handwritten-tests -c ./agent/functional_tests/tests/
+
+	# Race detector on windows requires CGO
+	GOOS=windows CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc go test -race -tags integration -o ./out/test-artifacts/windows-engine-tests.exe -c ./agent/engine
+	GOOS=windows CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc go test -race -tags integration -o ./out/test-artifacts/windows-stats-tests.exe -c ./agent/stats
+	GOOS=windows CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc go test -race -tags integration -o ./out/test-artifacts/windows-app-tests.exe -c ./agent/app
+	GOOS=windows go test -tags functional -o ./out/test-artifacts/windows-simple-tests.exe -c ./agent/functional_tests/tests/generated/simpletests_windows/
+	GOOS=windows go test -tags functional -o ./out/test-artifacts/windows-handwritten-tests.exe -c ./agent/functional_tests/tests/
+	GOOS=windows go build -o out/test-artifacts/agent.exe ./agent
```
samuelkarp: I'd prefer to use the same [build script](https://github.com/aws/amazon-ecs-agent/blob/v1.16.1/scripts/build) for the actual agent that we use elsewhere.

scripts/dockerfiles/Dockerfile.build
```diff
@@ -11,10 +11,13 @@
...
 # permissions and limitations under the License.
 
-FROM golang:1.7
+FROM golang:1.9
 MAINTAINER Amazon Web Services, Inc.
 
-RUN mkdir /out
+RUN apt-get update &&\
+    apt-get install -y gcc-mingw-w64 &&\
+    mkdir /out
```
jhaynes: I'd rather this `mkdir` be in a separate `RUN` directive since it isn't related to package installation.

buildspec.yml
samuelkarp: Well, the point of what I'm saying is I'd like the artifact we use for testing to better match what we're going to be giving to customers.  Building it two separate ways doesn't do that.

scripts/dockerfiles/Dockerfile.build
petderek: I'm thinking about simply putting it in the makefile itself, since several different targets expect this folder to exist.

Makefile
petderek: Doing that now. I hadn't addressed your set of comments before pushing my updates.

buildspec.yml
petderek: I think its valuable to build the same way, but the on-host unpacking and setup will still be different. For example, the working directory for the tests here will be `\go\src\...\amazon-ecs-agent` instead of the normal runtime `\Amazon\ECS`.

scripts/cleanbuild
```diff
@@ -27,7 +28,7 @@ EOW
 [ ! -z "$(git status --porcelain)" ] && echo "$DIRTY_WARNING"
 
 # Fresh clone to ensure our build doesn't rely on anything outside of vcs
-git clone --quiet /src/amazon-ecs-agent /go/src/github.com/aws/amazon-ecs-agent
+git clone -c user.name="cleanbuild" -c user.email="cleanbuild" /src/amazon-ecs-agent /go/src/github.com/aws/amazon-ecs-agent
```
jahkeup: What are these values used for? Can you comment on why we need to configure these?

scripts/local-stage
```diff
@@ -0,0 +1,49 @@
...
+# License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+IMAGE_NAME="amazon/amazon-ecs-agent:latest"
+
+stage() {
+	local tarball="./out/ecs-agent.tar"
+  docker save -o "${tarball}" "${IMAGE_NAME}"
```
jahkeup: Indentation got a bit funny here.

Makefile
```diff
@@ -79,7 +90,7 @@ gogenerate:
 certs: misc/certs/ca-certificates.crt
 misc/certs/ca-certificates.crt:
 	docker build -t "amazon/amazon-ecs-agent-cert-source:make" misc/certs/
-	docker run "amazon/amazon-ecs-agent-cert-source:make" cat /etc/ssl/certs/ca-certificates.crt > misc/certs/ca-certificates.crt
+	docker run -u "$(USERID)" "amazon/amazon-ecs-agent-cert-source:make" cat /etc/ssl/certs/ca-certificates.crt > misc/certs/ca-certificates.crt
```
jahkeup: The `-u` here shouldn't be necessary as the output is redirected by the user's shell.

Makefile
```diff
@@ -136,6 +166,12 @@ cni-plugins: get-cni-sources
 run-integ-tests: test-registry gremlin
 	. ./scripts/shared_env && go test -race -tags integration -timeout=5m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
 
+codebuild: out get-deps test-artifacts-in-docker
```
jahkeup: Absolutely outstanding to have codebuild producing artifacts! :clap:

Makefile
```diff
@@ -136,6 +166,12 @@ cni-plugins: get-cni-sources
...
+	$(MAKE) release TARGET_OS="linux"
```
jahkeup: These target names (`release`, `docker-release`) seem a little backwards to me after reading them, `release` depends on `docker-release` but `release` is actually the target that bundles the agent into a Docker container image (according to the comments). Likely a conversation for another PR.

scripts/cleanbuild
petderek: `git clone` will try to infer the user name to guess at these values for the initial git repo config. However, when running as a UID in a docker container it is unable to lookup the user name in the passwd file. It panics and exits.

Giving git clone this configuration prevents that error case.

scripts/local-stage
petderek: I don'
                    t know what you are talking about

Makefile
petderek: Yep. Theres a lot of simplification and cleanup we could do -- but for now I was only trying to mirror what the staging and building scripts do.

Makefile
```diff
@@ -31,25 +32,34 @@ all: docker
 gobuild:
 	./scripts/build false
 
+# create output directories
+out:
+	mkdir ./out
```
samuelkarp: I'd recommend `mkdir -p` for all of these.  Plus, you can do it on the same line:

`mkdir -p ./out/test-artifacts ./out/cni-plugins`

Makefile
```diff
@@ -31,25 +32,34 @@ all: docker
...
-	@docker build -f scripts/dockerfiles/Dockerfile.build -t "amazon/amazon-ecs-agent-build:make" .
+build-in-docker: builder-image out
 	@docker run --net=none \
-	  -e TARGET_OS="${TARGET_OS}" \
-	  -e LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
-	  -X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
-	  -v "$(PWD)/out:/out" \
-	  -v "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
-	  "amazon/amazon-ecs-agent-build:make"
+		-e TARGET_OS="${TARGET_OS}" \
```
samuelkarp: Would you be willing to change to the long flags for these?  `--env`, `--volume`, etc

Makefile
```diff
@@ -31,25 +32,34 @@ all: docker
...
+		-e LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
+		-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
```
samuelkarp: Can you indent the `-X` here and line it up with the one above?  It's not a top-level flag but actually part of the `LDFLAGS` environment variable.

Makefile
```diff
@@ -59,12 +69,13 @@ docker: certs build-in-docker pause-container-release cni-plugins
 docker-release: pause-container-release cni-plugins
 	@docker build -f scripts/dockerfiles/Dockerfile.cleanbuild -t "amazon/amazon-ecs-agent-cleanbuild:make" .
 	@docker run --net=none \
-	  -e TARGET_OS="${TARGET_OS}" \
-	  -v "$(PWD)/out:/out" \
-	  -e LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
-	  -X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
-	  -v "$(PWD):/src/amazon-ecs-agent" \
-	  "amazon/amazon-ecs-agent-cleanbuild:make"
+		-e TARGET_OS="${TARGET_OS}" \
```
samuelkarp: Same comments as above.

Makefile
```diff
@@ -90,6 +101,28 @@ test-silent:
...
 
+test-artifacts: out
+	go test -race -tags integration -o ./out/test-artifacts/unix-engine-tests -c ./agent/engine
+	go test -race -tags integration -o ./out/test-artifacts/unix-stats-tests -c ./agent/stats
+	go test -race -tags integration -o ./out/test-artifacts/unix-app-tests -c ./agent/app
+	go test -tags functional -o ./out/test-artifacts/unix-simple-tests -c ./agent/functional_tests/tests/generated/simpletests_unix/
+	go test -tags functional -o ./out/test-artifacts/unix-handwritten-tests -c ./agent/functional_tests/tests/
+
+	# Race detector on windows requires CGO
+	GOOS=windows CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc go test -race -tags integration -o ./out/test-artifacts/windows-engine-tests.exe -c ./agent/engine
```
samuelkarp: Do we need to install mingw32-gcc to run this?  It'd be good to skip these if it's not available.

Makefile
```diff
@@ -90,6 +101,28 @@ test-silent:
 benchmark-test:
 	. ./scripts/shared_env && go test -run=XX -bench=. $(shell go list ./agent/... | grep -v /vendor/)
 
+test-artifacts: out
+	go test -race -tags integration -o ./out/test-artifacts/unix-engine-tests -c ./agent/engine
```
samuelkarp: I think it'd be nicer to split these into targets that control the files, or at least have the generated files in the target.  Kinda like this:

```make
SOURCEDIR=./agent

.PHONY: test-artifacts
test-artifacts: linux-integration-artifacts linux-functional-artifacts

.PHONY: linux-integration-artifacts
linux-integration-artifacts: out/test-artifacts/linux-stats-tests out/test-artifacts/unix-app-tests

INTEG_STATS_SOURCES := $(shell find $(SOURCEDIR)/stats -name '*.go')
INTEG_APP_SOURCES := $(shell find $(SOURCEDIR)/app -name '*.go')
out/test-artifacts/linux-stats-tests out/test-artifacts/linux-app-tests: $(INTEG_STATS_SOURCES) $(INTEG_APP_SOURCES)
	go test -race -tags integration -o ./out/test-artifacts/linux-stats-tests -c ./agent/stats
	go test -race -tags integration -o ./out/test-artifacts/linux-app-tests -c ./agent/app
```

scripts/dockerfiles/Dockerfile.cleanbuild
```diff
@@ -14,12 +14,9 @@
...
 
-RUN mkdir /out
-VOLUME ['/out']
-
-RUN mkdir -p /go/src/github.com/aws/
-
 COPY ./scripts/cleanbuild /scripts/cleanbuild
 WORKDIR /go/src/github.com/aws/amazon-ecs-agent
 
+# Preserve permissions.. https://github.com/docker-library/golang/blob/2f2f3b620d61f533484f24a568c2ca46e4fda91c/1.9/stretch/Dockerfile#L49
```
samuelkarp: That's not preserving, but setting.  Those perms are fine, but the comment is inaccurate.

scripts/local-stage
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
samuelkarp: I wouldn't call this `stage`.  Maybe `local-save` or `local-artifacts`?

Makefile
petderek: Yep, its in the build dockerfile

scripts/dockerfiles/Dockerfile.cleanbuild
petderek: WRITEABLE EVERYWHERE

Makefile
petderek: neat

Makefile
```diff
@@ -116,18 +203,17 @@ pause-container:
 	@docker rmi -f "amazon/amazon-ecs-build-pause-bin:make"
 
 pause-container-release: pause-container
-	mkdir -p "$(PWD)/out"
 	@docker save ${PAUSE_CONTAINER_IMAGE}:${PAUSE_CONTAINER_TAG} > "$(PWD)/out/${PAUSE_CONTAINER_TARBALL}"
 
 get-cni-sources:
 	git submodule update --init --checkout
```
petderek: This is why I keep updating the submodule.

Makefile
```diff
@@ -90,6 +110,75 @@ test-silent:
...
+
+out/test-artifacts/windows-simple-tests.exe: out $(ALL_GO_FILES)
+	$(call win-cgo-dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests/generated/simpletests_windows)
+
+out/test-artifacts/windows-handwritten-tests.exe: out $(ALL_GO_FILES)
+	$(call win-cgo-dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests)
+
+.PHONY: test-artifacts-windows test-artifacts-linux test-artifacts
+
+test-artifacts-windows: out/test-artifacts/windows-engine-tests.exe out/test-artifacts/windows-stats-tests.exe out/test-artifacts/windows-app-tests.exe out/test-artifacts/windows-simple-tests.exe out/test-artifacts/windows-handwritten-tests.exe
```
samuelkarp: I think you can break this into a few lines so it doesn't go so long: https://stackoverflow.com/questions/38086542/break-long-dependencies-in-makefile-into-several-lines

scripts/local-save
```diff
@@ -0,0 +1,45 @@
...
+IMAGE_NAME="amazon/amazon-ecs-agent:latest"
+
+stage() {
+	local tarball="./out/ecs-agent.tar"
+	docker save -o "${tarball}" "${IMAGE_NAME}"
+}
+
+stage_windows() {
+	local ziproot="$(mktemp -d)"
+	local ziptgtdir="$(mktemp -d)"
```
samuelkarp: nit: I like "target" instead of "tgt"

scripts/local-stage
samuelkarp: Did you mean to leave this file here?

petderek: version control is hard

scripts/local-save
petderek: i had copied it from the other file -- will fix

Makefile
```diff
@@ -31,40 +32,59 @@ all: docker
 gobuild:
 	./scripts/build false
 
+# create output directories
+.out-stamp:
```
aaithal: This is not a `PHONY` target on purpose?

Makefile
```diff
@@ -31,40 +32,59 @@ all: docker
...
+	mkdir -p ./out/test-artifacts ./out/cni-plugins
+	touch .out-stamp
+
+out: .out-stamp
+
 # Basic go build
 static:
 	./scripts/build
 
+BUILDER_IMAGE="amazon/amazon-ecs-agent-build:make"
```
aaithal: nit: can you move this to the top of the file?

Makefile
```diff
@@ -31,40 +32,59 @@ all: docker
...
-	  -X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
-	  -v "$(PWD)/out:/out" \
-	  -v "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
-	  "amazon/amazon-ecs-agent-build:make"
+		--env TARGET_OS="${TARGET_OS}" \
+		--env LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
+			-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
+		--volume "$(PWD)/out:/out" \
+		--volume "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
+		--user "$(USERID)" \
```
aaithal: Can you please document why the `--user` param is required here?

Makefile
```diff
@@ -31,40 +32,59 @@ all: docker
...
+	touch .out-stamp
+
+out: .out-stamp
+
 # Basic go build
 static:
 	./scripts/build
 
+BUILDER_IMAGE="amazon/amazon-ecs-agent-build:make"
+.builder-image-stamp: scripts/dockerfiles/Dockerfile.build
```
aaithal: This is not a `PHONY` target on purpose?

Makefile
```diff
@@ -31,40 +32,59 @@ all: docker
 gobuild:
 	./scripts/build false
 
+# create output directories
+.out-stamp:
+	mkdir -p ./out/test-artifacts ./out/cni-plugins
+	touch .out-stamp
+
+out: .out-stamp
```
aaithal: Seems like `out` is a nop, why is it needed?

Makefile
```diff
@@ -31,40 +32,59 @@ all: docker
...
 # Basic go build
 static:
 	./scripts/build
 
+BUILDER_IMAGE="amazon/amazon-ecs-agent-build:make"
+.builder-image-stamp: scripts/dockerfiles/Dockerfile.build
+	@docker build -f scripts/dockerfiles/Dockerfile.build -t $(BUILDER_IMAGE) .
+	touch .builder-image-stamp
+
+builder-image: .builder-image-stamp
```
aaithal: Seems like `builder-image` is a nop, why is it needed?

Makefile
```diff
@@ -31,40 +32,59 @@ all: docker
...
-	  -X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
-	  -v "$(PWD):/src/amazon-ecs-agent" \
-	  "amazon/amazon-ecs-agent-cleanbuild:make"
+		--env TARGET_OS="${TARGET_OS}" \
+		--env LDFLAGS="-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerTag=$(PAUSE_CONTAINER_TAG) \
+			-X github.com/aws/amazon-ecs-agent/agent/config.DefaultPauseContainerImageName=$(PAUSE_CONTAINER_IMAGE)" \
+		--user "$(USERID)" \
+		--volume "$(PWD)/out:/out" \
+		--volume "$(PWD):/src/amazon-ecs-agent" \
+		--rm \
```
aaithal: Thank you!

Makefile
```diff
@@ -90,6 +110,80 @@ test-silent:
...
+		--env "GOOS=windows" \
+		--env "CGO_ENABLED=1" \
+		--env "CC=x86_64-w64-mingw32-gcc" \
+		--rm \
+		$(BUILDER_IMAGE) \
+		$(1)
+endef
+
+# TODO: use `go list -f` to target the test files more directly
+ALL_GO_FILES = $(shell find . -name "*.go" -print | tr "\n" " ")
```
aaithal: Can you please address this `TODO` in this PR? Also, as per earlier comments, please move these const's to the top of the file

Makefile
```diff
@@ -90,6 +110,80 @@ test-silent:
...
+GO_INTEG_TEST = go test -race -tags integration -c -o
+out/test-artifacts/linux-engine-tests: out $(ALL_GO_FILES)
```
aaithal: All of these `out/` targets are deliberately not `PHONY` targets, yeah?

Makefile
```diff
@@ -90,6 +110,80 @@ test-silent:
...
+
+out/test-artifacts/windows-stats-tests.exe: out $(ALL_GO_FILES)
+	$(call win-cgo-dockerbuild,$(GO_INTEG_TEST) $@ ./agent/stats)
+
+GO_FUNCTIONAL_TEST = go test -tags functional -c -o
+out/test-artifacts/linux-simple-tests: out $(ALL_GO_FILES)
+	$(call dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests/generated/simpletests_unix)
+
+out/test-artifacts/linux-handwritten-tests: out $(ALL_GO_FILES)
+	$(call dockerbuild,$(GO_FUNCTIONAL_TEST) $@ ./agent/functional_tests/tests)
```
aaithal: this is not `./agent/functional_tests/tests/...` because ?

Makefile
samuelkarp: They're not `.PHONY` because they're producing real files that make is then able to track.

aaithal: right, wanted to confirm that, thanks

Makefile
```diff
@@ -90,6 +110,80 @@ test-silent:
...
+WINDOWS_ARTIFACTS_TARGETS += out/test-artifacts/windows-app-tests.exe out/test-artifacts/windows-simple-tests.exe
+WINDOWS_ARTIFACTS_TARGETS += out/test-artifacts/windows-handwritten-tests.exe
+
+LINUX_ARTIFACTS_TARGETS := out/test-artifacts/linux-engine-tests out/test-artifacts/linux-stats-tests
+LINUX_ARTIFACTS_TARGETS += out/test-artifacts/linux-app-tests out/test-artifacts/linux-simple-tests
+LINUX_ARTIFACTS_TARGETS += out/test-artifacts/linux-handwritten-tests
+
+test-artifacts-windows: $(WINDOWS_ARTIFACTS_TARGETS)
+
+test-artifacts-linux: $(LINUX_ARTIFACTS_TARGETS)
```
nmeyerhans: Running `make test-artifacts-linux` twice results in the builds happening twice. This is probably the case with `test-artifacts-windows` as well. It would be nice if this was not the case.

Makefile
petderek: I want to ensure that the `./out/*` is created by the user running `make`.  If docker creates a folder structure, it will be owned by root and it breaks `make clean`.

I'm putting this here so that it can be reused by the multiple targets that need it.

Makefile
petderek: Yep! It creates a file called `.out-stamp` that tells `make` not to rerun this step.

Makefile
petderek: Same reasoning as `.out-stamp`

Makefile
petderek: Its not a `NOP`. It builds a dockerfile thats shared between multiple targets.

Makefile
petderek: I'll figure out the best place to put it -- we use it in several places.

The reason we need it is so that the build artifacts are owned by the user running `make` instead of root. This means `make clean` will work.

Makefile
petderek: In short, I tried and its a little complicated. I wanted more time to think about it -- but I'd love to make another pass and clean up more of the Makefile, including this `TODO`.

petderek: Regarding moving this constant, It seemed more standard to have makefiles in 'sections'. But I'm largely indifferent. If you still want me to move it I will.

Makefile
petderek: The `-c` test flag only works on a single package. Also, the only test files it needs to be aware of are in this folder.

Makefile
aaithal: But, that's done in the `.out-stamp` target. Isn't it? `out` itself isn't doing anything apart from depending on `.out-stamp`. Is there some obvious `Makefile` logic in play here that I'm missing?

petderek: Oh right right, I see what you're asking.

The make target needs to be the same name as the output, which is why .out-stamp creates .out-stamp. However, `.out-stamp` doesn't convey much to the targets that depend on it, so we use `out` to have a more meaningful name.

I'm not just using `out` because I read some advice online that suggets not using directories as makefile targets.

Makefile
petderek: Just kidding -- this is a nop. Its just a more meaningful name that other targets depend on.

Makefile
petderek: The reason is that I didn't understand [order-only prerequisites](https://www.gnu.org/savannah-checkouts/gnu/make/manual/html_node/Prerequisite-Types.html). Because `out` is labeled phony, it gets executed every time, leading make to believe that test-artifacts-linux needs to be executed as well. 

I can solve this by marking `out` as order-only.

Makefile
petderek: We talked about this offline and decided to keep variables near the targets that use them.

Makefile
nmeyerhans: Do you really need order-only prereqs, or can you use stamp files and non-PHONY targets?

PR #1176, opened 2018-01-11
More go vet fixes, needed for adding to static check target

yunhee-l:
### Summary
Fixes for go vet violations (there are more).

### Implementation details
Fixes for the following violations: 
- agent/api/eni.go:40: struct field DomainNameSearchList repeats json tag "omitempty" also at agent/api/eni.go:37  **<- see https://golang.org/pkg/encoding/json/#Marshal "note the leading comma"**
- agent/containermetadata/utils.go:61: result of fmt.Errorf call not used
- agent/handlers/taskmetadata/helpers.go:32: struct field httpErrorCode has json tag but is not exported  **<- I do not have context on why this field was not exported**
- agent/handlers/v1_handlers.go:73: github.com/aws/amazon-ecs-agent/agent/handlers/types/v1.ContainerResponse composite literal uses unkeyed fields

### Testing

Did not update changelog for this change

### Licensing

### DIFF
--- a/agent/containermetadata/utils.go
+++ b/agent/containermetadata/utils.go
@@ -58,7 +58,7 @@ func getMetadataFilePath(taskARN string, containerName string, dataDir string) (
 func getTaskMetadataDir(taskARN string, dataDir string) (string, error) {
 	taskID, err := getTaskIDfromARN(taskARN)
 	if err != nil {
-		fmt.Errorf("get task metadata directory: %v", err)
+		return "", fmt.Errorf("get task metadata directory: %v", err)
 	}
 	return filepath.Join(dataDir, metadataJoinSuffix, taskID), err
 }
--- a/agent/handlers/taskmetadata/helpers.go
+++ b/agent/handlers/taskmetadata/helpers.go
@@ -29,7 +29,7 @@ const (
 type errorMessage struct {
 	Code          string `json:"code"`
 	Message       string `json:"message"`
-	httpErrorCode int    `json:"-"`
+	httpErrorCode int
 }
 
 func writeJSONToResponse(w http.ResponseWriter, httpStatusCode int, jsonMessage []byte, requestType string) {
--- a/agent/api/eni.go
+++ b/agent/api/eni.go
@@ -34,10 +34,10 @@ type ENI struct {
 	MacAddress string
 	// DomainNameServers specifies the nameserver IP addresses for
 	// the eni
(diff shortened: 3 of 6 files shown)


agent/containermetadata/utils.go
```diff
@@ -58,7 +58,7 @@ func getMetadataFilePath(taskARN string, containerName string, dataDir string) (
 func getTaskMetadataDir(taskARN string, dataDir string) (string, error) {
 	taskID, err := getTaskIDfromARN(taskARN)
 	if err != nil {
-		fmt.Errorf("get task metadata directory: %v", err)
+		_ = fmt.Errorf("get task metadata directory: %v", err)
```
samuelkarp: This should probably return an error or be logged, not sure which one.  We don't want to just assign it to `_`.

agent/handlers/taskmetadata/helpers.go
```diff
@@ -29,7 +29,7 @@ const (
 type errorMessage struct {
 	Code          string `json:"code"`
 	Message       string `json:"message"`
-	httpErrorCode int    `json:"-"`
+	HttpErrorCode int    `json:"-"`
```
samuelkarp: I don't think we need to export it.  `json:"-"` should ignore the field when marshaling.  If you leave it unexported you can remove the `json` tag.

agent/containermetadata/utils.go
aaithal: this seems like a bug. seems like we should be bubbling it up.

agent/handlers/taskmetadata/credentials.go
```diff
@@ -54,7 +54,7 @@ func credentialsV1V2RequestHandler(credentialsManager credentials.Manager, audit
 		jsonResponse, arn, roleType, errorMessage, err := processCredentialsV1V2Request(credentialsManager, r, credentialsID, apiVersion)
 		if err != nil {
 			jsonMsg, _ := json.Marshal(errorMessage)
-			writeCredentialsV1V2RequestResponse(w, r, errorMessage.httpErrorCode, audit.GetCredentialsEventType(roleType), arn, auditLogger, jsonMsg)
+			writeCredentialsV1V2RequestResponse(w, r, errorMessage.HttpErrorCode, audit.GetCredentialsEventType(roleType), arn, auditLogger, jsonMsg)
```
aaithal: minor: can you split this into multiple lines?

PR #1183, opened 2018-01-15
Fix container metadata information missing during agent restart

richardpen:
### Summary

Fix the issue where agent could miss the container metadata information(portmapping/exitcode), if the container is started/stopped when the agent isn't running(eg: restart).

### Implementation details

Record the metadata information during the reconciliation on agent restart.

### Testing

- [x] Manual test:
  - [x] Reproducible with the previous agent, but not with this fix.
  - [x] Container information(port mapping/exit code) correctly submitted to backend.
  - [x] Container information was updated correctly after agent restart while container is created/started/stopped when agent isn't running.

New tests cover the changes: 
yes
### Licensing

yes

### DIFF
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -1103,3 +1103,21 @@ func (task *Task) GetID() (string, error) {
 
 	return resourceSplit[1], nil
 }
+
+func (task *Task) RecordExecutionStoppedAt(container *Container) {
+	if !container.Essential {
+		return
+	}
+	if container.GetKnownStatus() != ContainerStopped {
+		return
+	}
+	// If the essential container is stopped, set the ExecutionStoppedAt timestamp
+	now := time.Now()
+	ok := task.SetExecutionStoppedAt(now)
+	if !ok {
+		// ExecutionStoppedAt was already recorded. Nothing to left to do here
+		return
+	}
+	seelog.Infof("Task [%s]: recording execution stopped time. Essential container [%s] stopped at: %s",
+		task.Arn, container.Name, now.String())
+}
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -1104,6 +1104,8 @@ func (task *Task) GetID() (string, error) {
 	return resourceSplit[1], nil
 }
 
+// RecordExecutionStoppedAt checks if this is an essential container stopped
+// and set the task executionStoppedAt timestamps
 func (task *Task) RecordExecutionStoppedAt(container *Container) {
 	if !container.Essential {
 		return
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -37,6 +37,7 @@ import (
 	utilsync "github.com/aws/amazon-ecs-agent/agent/utils/sync"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
 
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
 	"github.com/pkg/errors"
 	"golang.org/x/net/context"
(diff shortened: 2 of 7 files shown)


aaithal:
@richardpen can you add more details about what `Manual test`s you're planning to run and if they have passed/failed as well?

agent/engine/docker_task_engine.go
```diff
@@ -294,32 +322,37 @@ func (engine *DockerTaskEngine) synchronizeContainerStatus(container *api.Docker
...
 		return
 	}
 
 	currentState, metadata := engine.client.DescribeContainer(container.DockerID)
 	if metadata.Error != nil {
 		currentState = api.ContainerStopped
-		if !container.Container.KnownTerminal() {
-			container.Container.ApplyingError = api.NewNamedError(&ContainerVanishedError{})
+		// If this is a Docker API error
+		if metadata.Error.ErrorName() == "CannotDescribeContainerError" {
```
aaithal: minor: Can you const `"CannotDescribeContainerError" `?

agent/engine/docker_task_engine.go
```diff
@@ -329,6 +362,16 @@ func (engine *DockerTaskEngine) synchronizeContainerStatus(container *api.Docker
 		// update the container known status
 		container.Container.SetKnownStatus(currentState)
 	}
+	// Update task ExecutionStoppedAt timestamp
+	if container.Container.GetKnownStatus() == api.ContainerStopped && container.Container.IsEssential() {
+		now := engine.time().Now()
+		ok := task.SetExecutionStoppedAt(now)
+		if ok {
+			seelog.Infof("Recording execution stopped time for a task, essential container in task stopped, task %s, time: %s",
```
aaithal: Once #1208 is merged, please rebase this on top of that and don't forget to update the log string as per the format being followed in that:
* `"Recording execution ` -> `"Task engine [%s] Recording execution `
* instead of `task.String()`, use `task.Arn`

agent/engine/docker_task_engine.go
```diff
@@ -332,6 +365,15 @@ func (engine *DockerTaskEngine) synchronizeContainerStatus(container *api.Docker
 		// update the container known status
 		container.Container.SetKnownStatus(currentState)
 	}
+	// Update task ExecutionStoppedAt timestamp
```
aaithal: Please rebase this from dev once #1217 is merged. This is now moved into its own method. May be it's better to review this after that PR is merged

agent/engine/docker_task_engine.go
```diff
@@ -285,6 +286,33 @@ func (engine *DockerTaskEngine) synchronizeState() {
...
+	}
+
+	// Update Volume
+	if metadata.Volumes != nil {
+		task.UpdateMountPoints(container, metadata.Volumes)
+	}
+
+	// Set Exitcode if it's not set
+	if metadata.ExitCode != nil && (container.GetKnownExitCode() == nil ||
+		aws.IntValue(metadata.ExitCode) != aws.IntValue(container.GetKnownExitCode())) {
```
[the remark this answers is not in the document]
richardpen: It should have the same effect, but not sure which way is better. I can change that if no one is on the opposite side.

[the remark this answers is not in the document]
aaithal: I vote for setting it anyway and getting rid of this check. nice catch!

agent/api/task.go
```diff
@@ -1103,3 +1103,21 @@ func (task *Task) GetID() (string, error) {
 
 	return resourceSplit[1], nil
 }
+
+func (task *Task) RecordExecutionStoppedAt(container *Container) {
```
aaithal: lint: `// RecordExecutionStoppedAt ...`

agent/engine/docker_task_engine_test.go
```diff
@@ -1720,3 +1720,170 @@ func TestHandleDockerHealthEvent(t *testing.T) {
 	})
 	assert.Equal(t, testContainer.Health.Status, api.ContainerHealthy)
 }
+
+func TestCreatedContainerMetadataUpdateOnRestart(t *testing.T) {
```
aaithal: I think you can combine all of these into a single test. Either a table based test or a task having all of these containers.

agent/engine/docker_task_engine.go
```diff
@@ -290,6 +291,36 @@ func (engine *DockerTaskEngine) synchronizeState() {
 	engine.saver.Save()
 }
 
+func updateContainerMetadata(metadata *DockerContainerMetadata, container *api.Container, task *api.Task) {
```
adnxn: nit: `// updateContainerMetadata ...`

agent/engine/docker_task_engine.go
```diff
@@ -301,51 +332,49 @@ func (engine *DockerTaskEngine) synchronizeContainerStatus(container *api.Docker
 			seelog.Warnf("Task engine [%s]: could not find matching container for expected name [%s]: %v",
 				task.Arn, container.DockerName, err)
 		} else {
+			// update the container metadata in case the container status/metadata changed during agent restart
```
adnxn: so if i understand correctly, this codepath is for containers that were ` potentially created while we were down` (debug msg above). is that right? maybe update the comment to reflect that.

PR #1186, opened 2018-01-17
Adding integration test for container health check

richardpen:
### Summary

Adding integration tests to cover the code path in `agent/stats` for container health check metrics.
### Testing

New tests cover the changes: 
yes
### Licensing

yes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-# Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You
 # may not use this file except in compliance with the License. A copy of
@@ -133,7 +133,7 @@ cni-plugins: get-cni-sources
 		"amazon/amazon-ecs-build-cniplugins:make"
 	@echo "Built amazon-ecs-cni-plugins successfully."
 
-run-integ-tests: test-registry gremlin
+run-integ-tests: test-registry gremlin container-health-check-image
 	. ./scripts/shared_env && go test -race -tags integration -timeout=5m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
 
 netkitten:
@@ -144,7 +144,7 @@ volumes-test:
 
 # TODO, replace this with a build on dockerhub or a mechanism for the
 # functional tests themselves to build this
-.PHONY: squid awscli fluentd gremlin taskmetadata-validator image-cleanup-test-images ecr-execution-role-image
+.PHONY: squid awscli fluentd gremlin taskmetadata-validator image-cleanup-test-images ecr-execution-role-image container-health-check-image
 squid:
 	$(MAKE) -C misc/squid $(MFLAGS)
 
@@ -164,6 +164,9 @@ taskmetadata-validator:
 	$(MAKE) -C misc/taskmetadata-validator $(MFLAGS)
 ecr-execution-role-image:
 	$(MAKE) -C misc/ecr $(MFLAGS)
+
+container-health-check-image:
+	$(MAKE) -C misc/container-health $(MFLAGS)
 
 .get-deps-stamp:
 	go get golang.org/x/tools/cmd/cover
@@ -185,4 +188,5 @@ clean:
 	-$(MAKE) -C misc/testnnp $(MFLAGS) clean
 	-$(MAKE) -C misc/image-cleanup-test-images $(MFLAGS) clean
 	-$(MAKE) -C misc/taskmetadata-validator $(MFLAGS) clean
+	-$(MAKE) -C misc/container-health $(MFLAGS) clean
 	-rm -f .get-deps-stamp
--- a/agent/stats/common_test.go
+++ b/agent/stats/common_test.go
@@ -131,7 +131,7 @@ func validateInstanceMetrics(t *testing.T, engine *DockerStatsEngine) {
 
 	taskMetric := taskMetrics[0]
 	assert.Equal(t, aws.StringValue(taskMetric.TaskDefinitionFamily), taskDefinitionFamily, "unexpected task definition family")
-	assert.Equal(t, aws.StringValue(taskMetric.TaskDefinitionVersion), taskDefinitionVersion, "unexcpected task definition version")
+	assert.Equal(t, aws.StringValue(taskMetric.TaskDefinitionVersion), taskDefinitionVersion, "unexpected task definition version")
 	assert.NoError(t, validateContainerMetrics(taskMetric.ContainerMetrics, 1), "validating container metrics failed")
 }
 
@@ -231,6 +231,8 @@ func validateTaskHealthMetrics(t *testing.T, engine *DockerStatsEngine) {
 }
 
 func validateEmptyTaskHealthMetrics(t *testing.T, engine *DockerStatsEngine) {
+	// If the metrics is empty, no health metrics will be send, the metadata won't be used
+	// no need to check the metadata here
 	_, healthMetrics, err := engine.GetTaskHealthMetrics()
 	assert.NoError(t, err, "getting task health failed")
 	assert.Len(t, healthMetrics, 0, "no health metrics was expected")
--- a/agent/stats/common_test.go
+++ b/agent/stats/common_test.go
@@ -15,20 +15,16 @@ package stats
 
 import (
 	"fmt"
-	"os"
 	"testing"
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
-	ecsengine "github.com/aws/amazon-ecs-agent/agent/engine"
-	"github.com/aws/amazon-ecs-agent/agent/engine/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/eventstream"
 	"github.com/aws/amazon-ecs-agent/agent/statechange"
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
 	"github.com/aws/amazon-ecs-agent/agent/tcs/model/ecstcs"
-	"github.com/aws/amazon-ecs-agent/agent/utils"
 
 	"github.com/aws/aws-sdk-go/aws"
 	docker "github.com/fsouza/go-dockerclient"
@@ -41,7 +37,6 @@ const (
 	// checkPointSleep is the sleep duration in milliseconds between
 	// starting/stopping containers in the test code.
 	checkPointSleep              = 5 * SleepBetweenUsageDataCollection
-	testImageName                = "amazon/amazon-ecs-gremlin:make"
 	testContainerHealthImageName = "amazon/amazon-ecs-containerhealthcheck:make"
 
 	// defaultDockerTimeoutSeconds is the timeout for dialing the docker remote API.
@@ -57,14 +52,11 @@ const (
 	containerName         = "gremlin-container"
 )
 
-var endpoint = utils.DefaultIfBlank(os.Getenv(ecsengine.DockerEndpointEnvVariable), ecsengine.DockerDefaultEndpoint)
-
-var client, _ = docker.NewClient(endpoint)
-var clientFactory = dockerclient.NewFactory(endpoint)
-var cfg = config.DefaultConfig()
-
-var defaultCluster = "default"
-var defaultContainerInstance = "ci"
+var (
+	cfg                      = config.DefaultConfig()
+	defaultCluster           = "default"
+	defaultContainerInstance = "ci"
+)
 
 func init() {
 	cfg.EngineAuthData = config.NewSensitiveRawMessage([]byte{})
--- a/agent/stats/common_test.go
+++ b/agent/stats/common_test.go
@@ -1,4 +1,4 @@
-// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"). You may
 // not use this file except in compliance with the License. A copy of the
@@ -28,16 +28,18 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
 	"github.com/aws/amazon-ecs-agent/agent/tcs/model/ecstcs"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
-	docker "github.com/fsouza/go-dockerclient"
 
+	"github.com/aws/aws-sdk-go/aws"
+	docker "github.com/fsouza/go-dockerclient"
 	"golang.org/x/net/context"
 )
 
 const (
 	// checkPointSleep is the sleep duration in milliseconds between
 	// starting/stopping containers in the test code.
-	checkPointSleep = 5 * SleepBetweenUsageDataCollection
-	testImageName   = "amazon/amazon-ecs-gremlin:make"
+	checkPointSleep              = 5 * SleepBetweenUsageDataCollection
+	testImageName                = "amazon/amazon-ecs-gremlin:make"
+	testContainerHelathImageName = "amazon/amazon-ecs-containerhealthcheck:make"
 
 	// defaultDockerTimeoutSeconds is the timeout for dialing the docker remote API.
 	defaultDockerTimeoutSeconds uint = 10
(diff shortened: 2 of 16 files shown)


Makefile
```diff
@@ -165,6 +165,9 @@ taskmetadata-validator:
 ecr-execution-role-image:
 	$(MAKE) -C misc/ecr $(MFLAGS)
 
+container-health-check-image:
```
aaithal: Can you also make sure that `clean` target for this gets invoked under the `clean:` section in the makefile?

agent/stats/common_test.go
```diff
@@ -28,16 +28,18 @@ import (
...
 )
 
 const (
 	// checkPointSleep is the sleep duration in milliseconds between
 	// starting/stopping containers in the test code.
-	checkPointSleep = 5 * SleepBetweenUsageDataCollection
-	testImageName   = "amazon/amazon-ecs-gremlin:make"
+	checkPointSleep              = 5 * SleepBetweenUsageDataCollection
+	testImageName                = "amazon/amazon-ecs-gremlin:make"
+	testContainerHelathImageName = "amazon/amazon-ecs-containerhealthcheck:make"
```
aaithal: typo: `testContainerHelathImageName` -> `testContainerHealthImageName`

agent/stats/common_test.go
```diff
@@ -170,6 +182,43 @@ func validateMetricsMetadata(metadata *ecstcs.MetricsMetadata) error {
 	return nil
 }
 
+func validateHealthMetricsMetadata(metadata *ecstcs.HealthMetadata) error {
+	if metadata == nil {
+		return fmt.Errorf("metadata is nil")
+	}
+
+	if aws.StringValue(metadata.Cluster) != defaultCluster {
+		return fmt.Errorf("expected cluster in metadata to be: %s, got %s", defaultCluster, aws.StringValue(metadata.Cluster))
```
aaithal: please split this and other long lines of code to multiple lines.

agent/stats/engine.go
```diff
@@ -390,6 +390,27 @@ func (engine *DockerStatsEngine) containerHealthsToMonitor() bool {
...
+	if err != nil {
+		// Error determining if the container is terminal. This means that the container
+		// id could not be resolved to a container that is being tracked by the
+		// docker task engine. If the docker task engine has already removed
+		// the container from its state, there's no point in stats engine tracking the
+		// container. So, clean-up anyway.
+		seelog.Warnf("Error determining if the container %s is terminal, removing from stats, err: %v", container.containerMetadata.DockerID, err)
+		engine.doRemoveContainerUnsafe(container, taskARN)
+		return true
+	} else if terminal {
```
aaithal: can you make just an `if terminal` instead? You don't need an `else if` because of the `return true` statement

agent/stats/engine.go
```diff
@@ -390,6 +390,27 @@ func (engine *DockerStatsEngine) containerHealthsToMonitor() bool {
 	return len(engine.tasksToHealthCheckContainers) != 0
 }
 
+func (engine *DockerStatsEngine) stopTrackingContainer(container *StatsContainer, taskARN string) bool {
```
aaithal: Please add `Unsafe` suffix for this method as it invokes `doRemoveContainerUnsafe` without a lock

aaithal: Can you also add documentation for this method explaining what it is that method is returning?

sdemontfort:
@richardpen when is this feature likely to be available on ECS? 😃

agent/stats/engine_integ_test.go
```diff
@@ -220,89 +191,275 @@ func TestStatsEngineWithNewContainers(t *testing.T) {
...
-	if *taskMetric.TaskDefinitionFamily != taskDefinitionFamily {
-		t.Errorf("Excpected task definition family to be: %s, got: %s", taskDefinitionFamily, *taskMetric.TaskDefinitionFamily)
+	assert.Equal(t, aws.StringValue(taskMetric.TaskDefinitionFamily), "docker-container-health", "task definition family not expected")
+	assert.Equal(t, aws.StringValue(taskMetric.TaskDefinitionVersion), taskDefinitionVersion, "task definition family version not expected")
+	assert.NoError(t, validateContainerMetrics(taskMetric.ContainerMetrics, 1), "error validating container metrics")
+
+	// Verify the health metrics of container
+	healthMetadata, healthMetrics, err := engine.GetTaskHealthMetrics()
+	assert.NoError(t, err, "getting task health metrics failed")
+	require.Len(t, healthMetrics, 1)
```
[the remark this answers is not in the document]
aaithal: :+1: for @sharanyad 's comment. Let's re-use as much code as possible in `*_test.go` files so that we don't have repeating lines of code in all of our tests. Can you take another shot at refactoring this and moving commonly used assertions/methods into methods and/or switching to table based tests?

agent/stats/engine_integ_test.go
```diff
@@ -220,89 +191,275 @@ func TestStatsEngineWithNewContainers(t *testing.T) {
...
+	testTask := &api.Task{
+		Arn:                 "container-health-task",
+		DesiredStatusUnsafe: api.TaskRunning,
+		KnownStatusUnsafe:   api.TaskRunning,
+		Family:              "docker-container-health",
+		Version:             "1",
+		Containers:          containers,
 	}
+	// Populate Tasks and Container map in the engine.
+	dockerTaskEngine, _ := taskEngine.(*ecsengine.DockerTaskEngine)
```
aaithal: if you're ignoring the `bool` returned here, you can switch to the single value return version of type-casting. `dockerTaskEngine := taskEngine.(*ecsengine.DockerTaskEngine)` should work i think

agent/stats/engine_integ_test.go
```diff
@@ -220,89 +191,275 @@ func TestStatsEngineWithNewContainers(t *testing.T) {
...
+	assert.NoError(t, err, "getting task health metrics failed")
+	require.Len(t, healthMetrics, 1)
+	assert.NoError(t, validateHealthMetricsMetadata(healthMetadata))
+	assert.Equal(t, aws.StringValue(healthMetrics[0].TaskArn), testTask.Arn, "task arn not expected")
+	assert.Equal(t, aws.StringValue(healthMetrics[0].TaskDefinitionFamily), "docker-container-health", "task definition family not expected")
+	assert.Equal(t, aws.StringValue(healthMetrics[0].TaskDefinitionVersion), taskDefinitionVersion, "task definition version not expected")
+	assert.NoError(t, validateTaskHealth(healthMetrics[0].Containers, 1))
+
+	err = client.StopContainer(container.ID, defaultDockerTimeoutSeconds)
+	assert.NoError(t, err, "stopping container failed")
```
aaithal: I think all errors from docker client should instead be handled by `require` than `assert`.

agent/stats/engine_integ_test.go
```diff
@@ -220,89 +191,275 @@ func TestStatsEngineWithNewContainers(t *testing.T) {
...
+	})
+	assert.NoError(t, err, "write to container change event stream failed")
+
+	time.Sleep(waitForCleanupSleep)
+
+	// Should not contain any metrics after cleanup.
+	err = validateIdleContainerMetrics(engine)
+	assert.NoError(t, err, "validating idle metrics failed")
+
+	_, healthMetrics, err = engine.GetTaskHealthMetrics()
```
aaithal: don't we need assertions for the metadata here as well?

agent/stats/engine_integ_test.go
```diff
@@ -220,89 +191,275 @@ func TestStatsEngineWithNewContainers(t *testing.T) {
...
+	defer engine.removeAll()
+
+	container, err := createHealthContainer(client)
+	require.NoError(t, err, "creating container failed")
+	defer client.RemoveContainer(docker.RemoveContainerOptions{
+		ID:    container.ID,
+		Force: true,
+	})
+
+	// Wait for containers from previous tests to transition states.
```
aaithal: Why is this needed? Aren't we cleaning up containers properly at the end of the tests? This seems like a race condition that will lead to flaky tests because of failure on slower machines.

agent/stats/engine_integ_test.go
```diff
@@ -220,89 +191,275 @@ func TestStatsEngineWithNewContainers(t *testing.T) {
...
+		},
+		testTask)
+
+	err = engine.MustInit(taskEngine, defaultCluster, defaultContainerInstance)
+	require.NoError(t, err, "initializing stats engine failed")
+	defer engine.containerChangeEventStream.Unsubscribe(containerChangeHandler)
+
+	err = client.StartContainer(container.ID, nil)
+	defer client.StopContainer(container.ID, defaultDockerTimeoutSeconds)
+	require.NoError(t, err, "starting container failed")
```
aaithal: `require` should precede `defer`

agent/stats/engine_test.go
```diff
@@ -318,6 +319,40 @@ func TestGetTaskHealthMetrics(t *testing.T) {
...
+	containerToStats[containerID] = newStatsContainer(containerID, nil, resolver)
+	engine.tasksToHealthCheckContainers["t1"] = containerToStats
+	engine.tasksToDefinitions["t1"] = &taskDefinition{
+		family:  "f1",
+		version: "1",
+	}
+
+	engine.resolver = resolver
+	_, _, err := engine.GetTaskHealthMetrics()
+	assert.Error(t, err, "empty metrics should causing error")
```
aaithal: nit: `causing` -> `cause an`

aaithal:
Hi @sdemontfort , we do not comment on the release dates and the roadmap. However, this is marked for the [1.17.0](https://github.com/aws/amazon-ecs-agent/milestone/16) milestone, which should should give you some indication of the same.

agent/stats/engine_integ_test.go
```diff
@@ -220,278 +161,323 @@ func TestStatsEngineWithNewContainers(t *testing.T) {
...
+	// Create a container to get the container id.
+	container, err := createHealthContainer(client)
+	require.NoError(t, err, "creating container failed")
+	defer client.RemoveContainer(docker.RemoveContainerOptions{
+		ID:    container.ID,
+		Force: true,
+	})
+
+	// Wait for containers from previous tests to transition states.
+	time.Sleep(checkPointSleep)
```
aaithal: I had a question regarding this here as well: https://github.com/aws/amazon-ecs-agent/pull/1186/#discussion_r162112427

Can you please answer why we need to wait for containers from previous tests to transition states at all? Why aren't the tests cleaning up things properly after themselves?

agent/stats/engine_integ_test.go
aaithal: This is still outstanding

agent/stats/engine_integ_test.go
aaithal: This is still outstanding

agent/tcs/client/client.go
```diff
@@ -89,7 +93,8 @@ func (cs *clientServer) Serve() error {
 
 	// Start the timer function to publish metrics to the backend.
 	cs.publishTicker = time.NewTicker(cs.publishMetricsInterval)
-	cs.endPublish = make(chan struct{})
+	cs.publishHealthTicker = time.NewTicker(cs.publishMetricsInterval)
+	cs.ctx, cs.cancel = context.WithCancel(context.TODO())
```
aaithal: please add a `TODO` here for this to be populated from handler (eventually via the app package)

agent/tcs/client/client.go
```diff
@@ -137,9 +142,14 @@ func (cs *clientServer) signRequest(payload []byte) []byte {
 func (cs *clientServer) Close() error {
 	if cs.publishTicker != nil {
 		cs.publishTicker.Stop()
-		cs.endPublish <- struct{}{}
+	}
+	if cs.publishHealthTicker != nil {
+		cs.publishHealthTicker.Stop()
 	}
 
+	if cs.cancel != nil {
```
aaithal: It's awesome that we're using contexts for this. However, i think we should get rid of this check. If needed we can populate this in our test files. But, checking `cancelFunc` is nil before invoking it doesn't seem like a clean way to do that imo

agent/stats/engine_integ_test.go
richardpen: I'm not sure about that, it exists in the previous test. Looking at the code, it looks to me that the previous test should impact the current test, I'll remove it and see if it works.

agent/stats/engine_integ_test.go
```diff
@@ -32,27 +35,28 @@ func init() {
...
-func (resolver *IntegContainerMetadataResolver) addToMap(containerID string) {
-	resolver.containerIDToTask[containerID] = &api.Task{
-		Arn:     taskArn,
-		Family:  taskDefinitionFamily,
-		Version: taskDefinitionVersion,
-	}
-	resolver.containerIDToDockerContainer[containerID] = &api.DockerContainer{
-		DockerID:  containerID,
-		Container: &api.Container{},
+func createTestTask() *api.Task {
```
adnxn: nit: naming `createTestTaskRunnning()`

agent/stats/engine_integ_test.go
adnxn: @aaithal, so i understand reasoning: why require vs assert?

aaithal:
@adnxn  `require` will halt the execution of the test if the assertion fails. There's not much point in continuing with the rest of the test there if we're unable to start/stop a container.

agent/stats/common_test.go
```diff
@@ -170,6 +179,63 @@ func validateMetricsMetadata(metadata *ecstcs.MetricsMetadata) error {
...
+	require.Len(t, healthMetrics, 1)
+	assert.NoError(t, validateHealthMetricsMetadata(healthMetadata), "validating health metedata failed")
+	assert.Equal(t, aws.StringValue(healthMetrics[0].TaskArn), taskArn, "task arn not expected")
+	assert.Equal(t, aws.StringValue(healthMetrics[0].TaskDefinitionFamily), taskDefinitionFamily, "task definition family not expected")
+	assert.Equal(t, aws.StringValue(healthMetrics[0].TaskDefinitionVersion), taskDefinitionVersion, "task definition version not expected")
+	assert.NoError(t, validateContainerHealthMetrics(healthMetrics[0].Containers, 1))
+}
+
+func validateEmptyTaskHealthMetrics(t *testing.T, engine *DockerStatsEngine) {
+	_, healthMetrics, err := engine.GetTaskHealthMetrics()
```
aaithal: can you also validate the metadata here? Or, please document why you aren't validating it.

agent/stats/engine_integ_test.go
```diff
@@ -32,186 +35,119 @@ func init() {
...
-func (resolver *IntegContainerMetadataResolver) addToMap(containerID string) {
-	resolver.containerIDToTask[containerID] = &api.Task{
-		Arn:     taskArn,
-		Family:  taskDefinitionFamily,
-		Version: taskDefinitionVersion,
-	}
-	resolver.containerIDToDockerContainer[containerID] = &api.DockerContainer{
-		DockerID:  containerID,
-		Container: &api.Container{},
+func createTaskRunning() *api.Task {
```
aaithal: nit: can you rename this to `createRunningTask`?

agent/stats/common_test.go
richardpen: It shouldn't matter, because if the metrics is empty, then we don't care the metadata as we won't send health metrics in this case.

aaithal: thanks for that explanation. Can you please document the same in the method?

richardpen: done

agent/stats/common_test.go
```diff
@@ -128,6 +131,18 @@ func (resolver *IntegContainerMetadataResolver) ResolveContainer(containerID str
 	return container, nil
 }
 
+func validateInstanceMetrics(t *testing.T, engine *DockerStatsEngine) {
+	metadata, taskMetrics, err := engine.GetInstanceMetrics()
+	assert.NoError(t, err, "gettting instance metrics failed")
```
sharanyad: should we use `require`, instead of `assert` here and in other tests?

agent/stats/common_test.go
```diff
@@ -170,6 +179,65 @@ func validateMetricsMetadata(metadata *ecstcs.MetricsMetadata) error {
...
+	}
+	if len(aws.StringValue(metadata.MessageId)) == 0 {
+		return fmt.Errorf("empty MessageId")
+	}
+
+	return nil
+}
+
+func validateContainerHealthMetrics(metrics []*ecstcs.ContainerHealth, expected int) error {
+	if len(metrics) != expected {
```
sharanyad: nit: can use `require.Len` here

agent/stats/engine.go
```diff
@@ -390,6 +390,31 @@ func (engine *DockerStatsEngine) containerHealthsToMonitor() bool {
 	return len(engine.tasksToHealthCheckContainers) != 0
 }
 
+// stopTrackingContainerUnsafe remove the StatsContaine from stats engine and
```
sharanyad: nit: `removes` , `StatsContainer`

agent/tcs/client/client.go
```diff
@@ -42,7 +43,9 @@ const (
 type clientServer struct {
 	statsEngine            stats.Engine
 	publishTicker          *time.Ticker
-	endPublish             chan struct{}
+	publishHealthTicker    *time.Ticker
```
sharanyad: What's the motivation for introducing a new ticker for health metrics?

richardpen: If we reuse the ticker, the interval of each metrics will mess and can be 2*interval = 40s.

agent/stats/common_test.go
richardpen: The `testing.T` isn't passed in this method, though we probably should, let's take it out of this PR, I will refactor this part with another pr.

agent/stats/common_test.go
richardpen: For this kind of test, we may not want to terminate, it still has value to continue the test.

PR #1195, opened 2018-01-19
enable hierarchy memory limit,  config cgroup root

adnxn:
### Summary

Includes the following changes:
1. enable `memory.use_hierarchy` scoped to the task's cgroup root
2. make the cgroup root path configurable through ecs.config and resolved at runtime, let default be `/sys/fs/cgroup/memory/`. this addresses https://github.com/aws/amazon-ecs-agent/pull/1147#pullrequestreview-85597853
3. refactored `DockerTaskEngine` to receive `agent.resource` instead of creating it's own instance.

We enforce the `memory.use_hierarchy` flag to the task level and not the `/ecs` root because the kernel does not allow modifications to anything under the `/ecs/` tree if child cgroups have already been created. This means that we will not be able to toggle the `memory.use_hierarchy` flag for existing ECS instances. Reports of this issue are also tracked here https://github.com/lxc/lxc/issues/236 and https://github.com/linuxkit/linuxkit/issues/373.

### Implementation details

1. echo 1 > memory.use_hierarchy scoped to task cgroup
2. added `ApplyConfigDependencies(cfg *config.Config)` to `type Resource interface` 
3. added `memorySubsystemPath string` to `type cgroupWrapper struct`
4. refactored function signatures for `NewTaskEngine` and `NewDockerTaskEngine`
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,7 @@ additional details on each available environment variable.
 | `ECS_ENABLE_CONTAINER_METADATA` | `true` | When `true`, the agent will create a file describing the container's metadata and the file can be located and consumed by using the container enviornment variable `$ECS_CONTAINER_METADATA_FILE` | `false` | `false` |
 | `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the ECS Agent is running as a container. We do not use this value in Windows because the ECS Agent is not running as container in Windows. | `/var/lib/ecs` | `Not used` |
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
+| `ECS_CGROUP_MEMORY_SUBSYSTEM_PATH` | `/sys/fs/cgroup/memory/` | The root path of the cgroup memory subsystem  | `/sys/fs/cgroup/memory/` | Not applicable |
 
 ### Persistence
 
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ additional details on each available environment variable.
 | `ECS_ENABLE_CONTAINER_METADATA` | `true` | When `true`, the agent will create a file describing the container's metadata and the file can be located and consumed by using the container enviornment variable `$ECS_CONTAINER_METADATA_FILE` | `false` | `false` |
 | `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the ECS Agent is running as a container. We do not use this value in Windows because the ECS Agent is not running as container in Windows. | `/var/lib/ecs` | `Not used` |
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
-| `ECS_CGROUP_MEMORY_SUBSYSTEM_PATH` | `/sys/fs/cgroup/memory/` | The root path of the cgroup memory subsystem  | `/sys/fs/cgroup/memory/` | Not applicable |
+| `ECS_CGROUP_PATH` | `/sys/fs/cgroup` | The root cgroup path that is expected by the ECS agent. This is the path that accessible from the agent mount. | `/sys/fs/cgroup` | Not applicable |
 
 ### Persistence
 
--- a/agent/acs/update_handler/updater_test.go
+++ b/agent/acs/update_handler/updater_test.go
@@ -1,4 +1,4 @@
-// Copyright 2014-2016 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"). You may
 // not use this file except in compliance with the License. A copy of the
@@ -105,7 +105,8 @@ func TestPerformUpdateWithUpdatesDisabled(t *testing.T) {
 		Reason:            ptr("Updates are disabled").(*string),
 	}})
 
-	u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
+	taskEngine := engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil)
+	msg := &ecsacs.PerformUpdateMessage{
 		ClusterArn:           ptr("cluster").(*string),
 		ContainerInstanceArn: ptr("containerInstance").(*string),
 		MessageId:            ptr("mid").(*string),
@@ -113,7 +114,9 @@ func TestPerformUpdateWithUpdatesDisabled(t *testing.T) {
 			Location:  ptr("https://s3.amazonaws.com/amazon-ecs-agent/update.tar").(*string),
 			Signature: ptr("c54518806ff4d14b680c35784113e1e7478491fe").(*string),
 		},
-	})
+	}
+
+	u.performUpdateHandler(statemanager.NewNoopStateManager(), taskEngine)(msg)
 }
 
 func TestFullUpdateFlow(t *testing.T) {
@@ -159,7 +162,8 @@ func TestFullUpdateFlow(t *testing.T) {
 
 			require.Equal(t, "update-tar-data", writtenFile.String(), "incorrect data written")
 
-			u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
+			taskEngine := engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil)
+			msg := &ecsacs.PerformUpdateMessage{
 				ClusterArn:           ptr("cluster").(*string),
 				ContainerInstanceArn: ptr("containerInstance").(*string),
 				MessageId:            ptr("mid2").(*string),
@@ -167,7 +171,9 @@ func TestFullUpdateFlow(t *testing.T) {
 					Location:  ptr("https://" + host + "/amazon-ecs-agent/update.tar").(*string),
 					Signature: ptr("c54518806ff4d14b680c35784113e1e7478491fe").(*string),
 				},
-			})
+			}
+
+			u.performUpdateHandler(statemanager.NewNoopStateManager(), taskEngine)(msg)
 		})
 	}
 }
@@ -224,11 +230,14 @@ func TestUndownloadedUpdate(t *testing.T) {
 		MessageId:         ptr("mid").(*string),
 	}})
 
-	u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
+	taskEngine := engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil)
+	msg := &ecsacs.PerformUpdateMessage{
 		ClusterArn:           ptr("cluster").(*string),
 		ContainerInstanceArn: ptr("containerInstance").(*string),
 		MessageId:            ptr("mid").(*string),
-	})
+	}
+
+	u.performUpdateHandler(statemanager.NewNoopStateManager(), taskEngine)(msg)
 }
 
 func TestDuplicateUpdateMessagesWithSuccess(t *testing.T) {
@@ -282,7 +291,8 @@ func TestDuplicateUpdateMessagesWithSuccess(t *testing.T) {
 
 	require.Equal(t, "update-tar-data", writtenFile.String(), "incorrect data written")
 
-	u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
+	taskEngine := engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil)
+	msg := &ecsacs.PerformUpdateMessage{
 		ClusterArn:           ptr("cluster").(*string),
 		ContainerInstanceArn: ptr("containerInstance").(*string),
 		MessageId:            ptr("mid3").(*string),
@@ -290,7 +300,9 @@ func TestDuplicateUpdateMessagesWithSuccess(t *testing.T) {
 			Location:  ptr("https://s3.amazonaws.com/amazon-ecs-agent/update.tar").(*string),
 			Signature: ptr("c54518806ff4d14b680c35784113e1e7478491fe").(*string),
 		},
-	})
+	}
+
+	u.performUpdateHandler(statemanager.NewNoopStateManager(), taskEngine)(msg)
 }
 
 func TestDuplicateUpdateMessagesWithFailure(t *testing.T) {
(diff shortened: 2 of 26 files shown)


agent/config/config.go
```diff
@@ -280,6 +280,8 @@ func environmentConfig() (Config, error) {
 
 	dockerStopTimeout := getDockerStopTimeout()
 
+	cgroupMemorySubsystemPath := os.Getenv("ECS_CGROUP_MEMORY_SUBSYSTEM_PATH")
```
richardpen: This needs to be updated in the readme file.

agent/config/config_unix.go
```diff
@@ -27,6 +27,8 @@ const (
 	defaultCredentialsAuditLogFile = "/log/audit.log"
 	// Default cgroup prefix for ECS tasks
 	DefaultTaskCgroupPrefix = "/ecs"
+
+	defaultCgroupMemorySubsystemPath = "/sys/fs/cgroup/memory/"
```
richardpen: Can you add a comment for this?

agent/resources/resources_linux.go
```diff
@@ -93,6 +109,14 @@ func (c *cgroupWrapper) setupCgroup(task *api.Task) error {
 		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s for task: %s", cgroupRoot, task.Arn)
 	}
 
+	// echo 1 > memory.use_hierarchy
+	memorySubsystem := c.memorySubsystemPath
+
+	err = c.ioutil.WriteFile(filepath.Join(memorySubsystem, cgroupRoot, memoryUseHierarchy), []byte(strconv.Itoa(1)), os.FileMode(0))
```
richardpen: Can you make the `1` and `0` constant and add some comment?

adnxn:
@richardpen righttt the header dates. will do.

agent/resources/resources_linux.go
adnxn: Unless you feel strongly about it, I don't think this needs to have its own constants since they're only being used in this specific instance, also I think the comment ` echo 1 > ...` conveys context. Let me know though.

agent/config/config.go
adnxn: fixed.

agent/acs/update_handler/updater_test.go
```diff
@@ -105,7 +105,7 @@ func TestPerformUpdateWithUpdatesDisabled(t *testing.T) {
 		Reason:            ptr("Updates are disabled").(*string),
 	}})
 
-	u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
+	u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
```
aaithal: minor: this is an opportunity for you to split these long lines into multiple lines :)

agent/app/agent.go
```diff
@@ -308,13 +311,13 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 	if !agent.cfg.Checkpoint {
 		seelog.Info("Checkpointing not enabled; a new container instance will be created each time the agent is run")
 		return engine.NewTaskEngine(agent.cfg, agent.dockerClient,
-			credentialsManager, containerChangeEventStream, imageManager, state, agent.metadataManager), "", nil
+			credentialsManager, containerChangeEventStream, imageManager, state, agent.metadataManager, agent.resource), "", nil
```
aaithal: minor: again, this is an opportunity for you to split these long lines into multiple lines :)

README.md
```diff
@@ -176,6 +176,7 @@ additional details on each available environment variable.
 | `ECS_ENABLE_CONTAINER_METADATA` | `true` | When `true`, the agent will create a file describing the container's metadata and the file can be located and consumed by using the container enviornment variable `$ECS_CONTAINER_METADATA_FILE` | `false` | `false` |
 | `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the ECS Agent is running as a container. We do not use this value in Windows because the ECS Agent is not running as container in Windows. | `/var/lib/ecs` | `Not used` |
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
+| `ECS_CGROUP_MEMORY_SUBSYSTEM_PATH` | `/sys/fs/cgroup/memory/` | The root path of the cgroup memory subsystem  | `/sys/fs/cgroup/memory/` | Not applicable |
```
aaithal: is this the path on the host or in the agent mount? We should make that more clear here as well.

agent/app/agent.go
```diff
@@ -145,6 +145,9 @@ func newAgent(
 		metadataManager = containermetadata.NewManager(dockerClient, cfg)
 	}
 
+	resource := resources.New()
+	resource.ApplyConfigDependencies(cfg)
```
aaithal: shouldn't this be happening only if cgroup feature is enabled?

agent/resources/resources_linux.go
aaithal: minor: can you split this into multiple lines?

aaithal: should we do this for each task? Isn't it enough if we did this once at the top level `ecs` cgroup?

README.md
petderek: I would prefer `ECS_CGROUP_PATH` and not include the memory subsystem at this level.

It makes it more open for reuse, and I think its important given that it is a configuration variable.

agent/app/agent.go
petderek: That logic is within resources, I believe.

The idea was that 'resources' could hold on to more things than just cgroups.

adnxn: i can put it behind after the `agent.cfg.TaskCPUMemLimit.Enabled` check.

agent/resources/resources_linux.go
adnxn: @aaithal  you're right, ill move this up a level and make it part of `cgroupInit()`

README.md
adnxn: this sounds reasonable.

agent/config/types.go
```diff
@@ -194,4 +194,6 @@ type Config struct {
 	// OverrideAWSLogsExecutionRole is config option used to enable awslogs
 	// driver authentication over the task's execution role
 	OverrideAWSLogsExecutionRole bool
+
+	CgroupPath string
```
aaithal: Needs a comment

agent/resources/interface.go
```diff
@@ -26,4 +27,6 @@ type Resource interface {
 	Setup(task *api.Task) error
 	// Cleanup removes the resource
 	Cleanup(task *api.Task) error
+
+	ApplyConfigDependencies(cfg *config.Config)
```
aaithal: Needs a comment

agent/acs/update_handler/updater_test.go
```diff
@@ -159,7 +160,7 @@ func TestFullUpdateFlow(t *testing.T) {
 
 			require.Equal(t, "update-tar-data", writtenFile.String(), "incorrect data written")
 
-			u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
+			u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
```
aaithal: Can you also split this and other long lines as well?

agent/acs/update_handler/updater_test.go
```diff
@@ -105,7 +105,8 @@ func TestPerformUpdateWithUpdatesDisabled(t *testing.T) {
 		Reason:            ptr("Updates are disabled").(*string),
 	}})
 
-	u.performUpdateHandler(statemanager.NewNoopStateManager(), engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil))(&ecsacs.PerformUpdateMessage{
+	u.performUpdateHandler(statemanager.NewNoopStateManager(),
```
petderek: I think we are trying to do a little too much in one line, even with line breaks. Maybe assign some variables above?

agent/config/config_unix.go
```diff
@@ -30,7 +30,7 @@ const (
 
 	// Default cgroup memory system root path, this is the default used if the
 	// path has not been configured through ECS_CGROUP_MEMORY_SUBSYSTEM_PATH
```
petderek: change the comment

agent/resources/resources_linux.go
```diff
@@ -55,13 +68,31 @@ func (c *cgroupWrapper) Cleanup(task *api.Task) error {
 	return c.cleanupCgroup(task)
 }
 
+func (c *cgroupWrapper) ApplyConfigDependencies(cfg *config.Config) {
+	c.cgroupPath = cfg.CgroupPath
+}
+
 // cgroupInit is used to create the root '/ecs/ cgroup
```
richardpen: Can you update this comment?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -98,6 +98,17 @@ func TestOOMContainer(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+// TestOOMTask verifies that a task with a memory limit returns an error
```
richardpen: Can you add the `agent.RequireVersion(">=1.16.0")` for this test?

agent/config/types.go
```diff
@@ -194,4 +194,7 @@ type Config struct {
 	// OverrideAWSLogsExecutionRole is config option used to enable awslogs
 	// driver authentication over the task's execution role
 	OverrideAWSLogsExecutionRole bool
+
+	// The expected Cgroup root path in the agent
```
richardpen: Please use the format  `CgroupPath is xxx`.

agent/resources/resources_linux_test.go
```diff
@@ -39,13 +42,18 @@ func TestInitHappyPath(t *testing.T) {
 	defer ctrl.Finish()
 
 	mockControl := mock_cgroup.NewMockControl(ctrl)
+	mockIO := mock_ioutilwrapper.NewMockIOUtil(ctrl)
+	cgroupPath := fmt.Sprintf("/sys/fs/cgroup/memory/ecs/memory.use_hierarchy")
```
richardpen: you don't need the `fmt.Sprintf` here.

agent/functional_tests/tests/functionaltests_unix_test.go
adnxn: fixed.

agent/config/types.go
```diff
@@ -194,4 +194,7 @@ type Config struct {
 	// OverrideAWSLogsExecutionRole is config option used to enable awslogs
 	// driver authentication over the task's execution role
 	OverrideAWSLogsExecutionRole bool
+
+	// The expected Cgroup path in the agent is /sys/fs/cgroup
```
aaithal: lint: `CgroupPath is ...`

agent/resources/resources_linux.go
```diff
@@ -55,13 +68,32 @@ func (c *cgroupWrapper) Cleanup(task *api.Task) error {
...
+	err := c.control.Init()
+
+	if err != nil {
+		return err
+	}
+
+	// echo 1 > memory.use_hierarchy
+	cgroupRoot := c.cgroupPath
+
+	err = c.ioutil.WriteFile(filepath.Join(cgroupRoot, memorySubsystem, config.DefaultTaskCgroupPrefix, memoryUseHierarchy), []byte(strconv.Itoa(1)), os.FileMode(0))
```
aaithal: nit: please split this into multiple lines. Also, please add documentation for all those magic numbers:
* `[]byte(strconv.Itoa(1)`
* `os.FileMode(0)`

adnxn: oops missed this one for splits, and will doc numbers.

aaithal:
@adnxn can you also update this PR with a commit with a changelog entry as well?

agent/resources/resources_linux.go
```diff
@@ -16,27 +16,41 @@
...
+	memorySubsystem    = "/memory"
+	memoryUseHierarchy = "memory.use_hierarchy"
+	rootReadOnly       = 400
+)
+
 // cgroupWrapper implements the Resource interface
 type cgroupWrapper struct {
-	control cgroup.Control
+	control    cgroup.Control
+	ioutil     ioutilwrapper.IOUtil
```
sharanyad: Why should `ioutil` be part of the `cgroupWrapper` object? How about creating the object, when its actually used? (only once?)

agent/engine/default.go
```diff
@@ -21,13 +21,15 @@ import (
...
 // NewTaskEngine returns a default TaskEngine
 func NewTaskEngine(cfg *config.Config, client DockerClient,
 	credentialsManager credentials.Manager,
 	containerChangeEventStream *eventstream.EventStream,
 	imageManager ImageManager, state dockerstate.TaskEngineState,
-	metadataManager containermetadata.Manager) TaskEngine {
-	return NewDockerTaskEngine(cfg, client, credentialsManager, containerChangeEventStream, imageManager, state, metadataManager)
+	metadataManager containermetadata.Manager,
+	resource resources.Resource) TaskEngine {
+	return NewDockerTaskEngine(cfg, client, credentialsManager, containerChangeEventStream, imageManager, state, metadataManager, resource)
```
aaithal: nit: we can probably split this into multiple lines as well

agent/resources/resources_linux.go
```diff
@@ -93,6 +113,12 @@ func (c *cgroupWrapper) setupCgroup(task *api.Task) error {
 		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s for task: %s", cgroupRoot, task.Arn)
 	}
 
+	// echo 1 > memory.use_hierarchy
+	err = c.ioutil.WriteFile(filepath.Join(c.cgroupPath, memorySubsystem, cgroupRoot, memoryUseHierarchy), []byte(strconv.Itoa(1)), os.FileMode(0))
```
aaithal: this too, please split this into multiple lines

adnxn: how do i keep missing this one. FIXED 😬

agent/resources/resources_linux.go
```diff
@@ -93,6 +113,13 @@ func (c *cgroupWrapper) setupCgroup(task *api.Task) error {
 		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s for task: %s", cgroupRoot, task.Arn)
 	}
 
+	// echo 1 > memory.use_hierarchy
+	memoryHierarchyPath := filepath.Join(c.cgroupPath, memorySubsystem, cgroupRoot, memoryUseHierarchy)
+	err = c.ioutil.WriteFile(memoryHierarchyPath, []byte(strconv.Itoa(1)), os.FileMode(0))
```
aaithal: these are still magic numbers. you should be able to const both of these with descriptive names. sorry for missing this in the last round.

adnxn: ah it's fine. i made this change earlier and i think it got lost while trying to revert back to per task `memory.use_hierarchy`

agent/resources/resources_linux.go
```diff
@@ -55,12 +70,18 @@ func (c *cgroupWrapper) Cleanup(task *api.Task) error {
 	return c.cleanupCgroup(task)
 }
 
-// cgroupInit is used to create the root '/ecs/ cgroup
+func (c *cgroupWrapper) ApplyConfigDependencies(cfg *config.Config) {
+	c.cgroupPath = cfg.CgroupPath
+}
+
+// cgroupInit is used to create the root '/ecs/ cgroup and enable
+// memory.use_hierarchy at the '/ecs/' level
```
sharanyad: Can you edit this comment to reflect the new changes?

adnxn: good catch.

agent/resources/resources_linux.go
```diff
@@ -93,6 +114,13 @@ func (c *cgroupWrapper) setupCgroup(task *api.Task) error {
 		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s for task: %s", cgroupRoot, task.Arn)
 	}
 
+	// echo 1 > memory.use_hierarchy
+	memoryHierarchyPath := filepath.Join(c.cgroupPath, memorySubsystem, cgroupRoot, memoryUseHierarchy)
+	err = c.ioutil.WriteFile(memoryHierarchyPath, []byte(strconv.Itoa(setMemoryHierarchy)), os.FileMode(rootReadOnly))
```
aaithal: Apologies for not being more clear about this. I think you can do something like to make it more readable:
```
const (
	rootReadOnlyPermissions = os.FileMode(400)
)

var (
	enableMemoryHierarchy = []byte(strconv.Itoa(1))
)
```

PR #1206, opened 2018-01-22
Overrides security group for awsvpc task

richardpen:
### Testing

New tests cover the changes: 

### Licensing

yes

### DIFF
--- a/agent/functional_tests/tests/functionaltests_test.go
+++ b/agent/functional_tests/tests/functionaltests_test.go
@@ -18,7 +18,6 @@ package functional_tests
 import (
 	"context"
 	"fmt"
-	"net/http"
 	"os"
 	"reflect"
 	"strconv"
@@ -28,7 +27,6 @@ import (
 	ecsapi "github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	. "github.com/aws/amazon-ecs-agent/agent/functional_tests/util"
 
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


agent/functional_tests/util/utils.go
```diff
@@ -246,14 +246,20 @@ func (agent *TestAgent) startAWSVPCTask(taskDefinition string) (*TestTask, error
...
 	agent.t.Logf("Starting 'awsvpc' task in subnet: %s", subnet)
 	resp, err := ECS.StartTask(&ecs.StartTaskInput{
 		Cluster:            &agent.Cluster,
 		ContainerInstances: []*string{&agent.ContainerInstanceArn},
 		TaskDefinition:     &taskDefinition,
 		NetworkConfiguration: &ecs.NetworkConfiguration{
 			AwsvpcConfiguration: &ecs.AwsVpcConfiguration{
-				Subnets: []*string{&subnet},
+				Subnets:        []*string{&subnet},
+				SecurityGroups: []*string{&securityGroup},
```
samuelkarp: Do we accept the security group name here as well as the ID?  I think IMDS is returning the name.

richardpen: Looks like it doesn't accept the name, I will modify this.

aaithal:
I have a different take on this. I think we can get rid of the part of the test where we test for reachability into the task by querying the task's IP address. It seems like we are testing if SG's are working at this point of time rather than is the ECS agent able to launch tasks with `"awsvpc"` mode. If we're able to prove that the task was launched successfully when networking mode is `"awsvpc"`, that should be enough. What do you think about getting rid of that check altogether? 

cc @richardpen  @samuelkarp

richardpen:
@aaithal Sam had the same opinion offline, my opinion is that the IP address of the eni is set by the plugin, this check also ensures the IP address is set up correctly?

agent/engine/docker_container_engine_test.go
```diff
@@ -355,7 +355,7 @@ func TestCreateContainerTimeout(t *testing.T) {
 		warp <- time.Now()
 		wait.Wait()
 		// Don't return, verify timeout happens
-	})
+	}).MaxTimes(1)
```
aaithal: why was this added?

richardpen: Because this test sometimes fails due to this missing call. Checking the code, this could happen if the `xContainerShortTimeout` is too short, where the `CreateContainer` hasn't been called but the context has returned [here](https://github.com/richardpen/amazon-ecs-agent/blob/master/agent/engine/docker_container_engine.go#L458).

aaithal: This means that sometimes, the code-path where the context might be done while a `createContainer` call is in progress will not be verified by this test. I think the right way to fix this is by wiring in a context to the `CreateContainer` [call](https://github.com/richardpen/amazon-ecs-agent/blob/ea58fddf8462034ff2034aa9788b02341d7702e3/agent/engine/docker_container_engine.go#L440).  Thinking more about it, all of the APIs in the `dockerGoClient` struct need `context`s to be wired in from the caller. Can you please add a TODO and open an issue for that?

richardpen: Issue #1212 was created.

PR #1217, opened 2018-01-25
fix issue1043

aaithal:
### Testing

- [x] Builds on Linux (`make release`)
- [x] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [x] Unit tests on Linux (`make test`) pass
- [x] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [x] Integration tests on Linux (`make run-integ-tests`) pass
- [x] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [x] Functional tests on Linux (`make run-functional-tests`) pass
- [x] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass
- [X] Manual tests: 
  - [X] Simulated inspect error in start container and verified that other containers and the started container in a task with ephemeral volumes were stopped and cleaned up properly
  - [X] Simulated docker timeout error in start container and verified that other containers and the started container in a task with ephemeral volumes were stopped and cleaned up properly
New tests cover the changes: yes!

### Licensing

This contribution is under the terms of the Apache 2.0 License:

Issue #1191, opened 2018-01-18
TestTaskMetadataValidator fails with older agents


samuelkarp:
### Summary
TestTaskMetadataValidator fails with older agents

### Description
[TestTaskMetadataValidator](https://github.com/aws/amazon-ecs-agent/blob/27a1a4de898f3d4a402327ddcfe77408e0eb9ef8/agent/functional_tests/tests/functionaltests_unix_test.go#L673-L674) does not `agent.RequireVersion(>=1.17.0)`, so it fails when run with an old agent.

### Expected Behavior
Test is skipped for an agent that is to old

### Observed Behavior
Test runs and fails

richardpen:
This has been fixed in #1218

PR #1218, opened 2018-01-26
Adding container health information into task metadata

richardpen:
### Summary

Providing container health information in the task metadata.

### Testing

New tests cover the changes: 

### Licensing

yes


agent/functional_tests/util/utils.go
```diff
@@ -246,14 +247,20 @@ func (agent *TestAgent) startAWSVPCTask(taskDefinition string) (*TestTask, error
 		return nil, err
 	}
 
-	agent.t.Logf("Starting 'awsvpc' task in subnet: %s", subnet)
+	securityGroup, err := getSecurityGroupID()
```
aaithal: as per #1206 , you don't need this anymore, right?

misc/taskmetadata-validator/Makefile
```diff
@@ -17,7 +17,7 @@ all: taskmetadata-validator image
 taskmetadata-validator: taskmetadata-validator.go
 	@./build-in-docker
 
-image: taskmetadata-validator
+image: taskmetadata-validator hello.go
```
aaithal: this should be moved to line num 17

misc/taskmetadata-validator/hello.go
```diff
@@ -0,0 +1,22 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aaithal: I don't feel strongly about this. But, you're probably better off using a base image like `alpine` and using one of the commands available there as a proxy instead of doing something like this. Means less code and less chance for bugs.

agent/engine/task_manager.go
```diff
@@ -262,7 +262,7 @@ func (mtask *managedTask) waitSteady() {
 	timedOut := mtask.waitEvent(maxWait)
 
 	if timedOut {
-		seelog.Debugf("Managed task [%s]: checking to make sure it's still at steadystate")
+		seelog.Debugf("Managed task [%s]: checking to make sure it's still at steadystate", mtask.Arn)
```
aaithal: this is fixed in #1217

agent/engine/docker_task_engine.go
```diff
@@ -522,7 +522,7 @@ func (engine *DockerTaskEngine) handleDockerEvent(event DockerContainerChangeEve
 	}
 	seelog.Debugf("Task engine [%s]: writing docker event to the task: %v", task.Arn, event)
 	managedTask.dockerMessages <- dockerContainerChange{container: cont.Container, event: event}
-	seelog.Debugf("Task engine [%s]: wrote docker event to the task", task.Arn, event)
+	seelog.Debugf("Task engine [%s]: wrote docker event to the task: %v", task.Arn, event)
```
aaithal: This is fixed with #1217

agent/functional_tests/util/utils.go
richardpen: ah, for some reason I lost my mind, the metadata endpoint doesn't need to go through eni. I'll remove that.

misc/taskmetadata-validator/hello.go
richardpen: yeah, I thought the same. I'll modify that.

misc/taskmetadata-validator/taskmetadata-validator.go
```diff
@@ -107,6 +115,8 @@ func containerMetadata(client *http.Client, id string) (*ContainerResponse, erro
 		return nil, err
 	}
 
+	fmt.Println("Received data: \n ", string(body))
```
aaithal: minor: redundant `\n`

misc/taskmetadata-validator/taskmetadata-validator.go
```diff
@@ -209,11 +222,15 @@ func main() {
...
 	}
 
-	_, err = containerMetadata(client, containerID)
+	containerMetadata, err := containerMetadata(client, containerID)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Unable to get container metadata for '%s': %v", containerID, err)
 		os.Exit(1)
 	}
+	if containerMetadata.Health.Status != "HEALTHY" || containerMetadata.Health.Output != "hello" {
+		fmt.Fprintf(os.Stderr, "Container health metadata unexpected, got: %s", containerMetadata.Health)
```
aaithal: this, and the other `fmt.Fprintf` probably needs a `\n` at the end

agent/handlers/types/v2/response.go
```diff
@@ -58,6 +58,7 @@ type ContainerResponse struct {
 	FinishedAt    *time.Time `json:",omitempty"`
 	Type          string
 	Networks      []containermetadata.Network `json:",omitempty"`
+	Health        *api.HealthStatus           `json:",omitempty"`
```
aaithal: `api.HealthStatus`'s  `Status` field is an enum type (`ContainerHealthStatus`) and we should emit a `string` for that instead of a number.

richardpen: It actually implemented the `MarshalJson` [here](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/api/containerstatus.go#L209), so it will appear as `HEALTHY`, `UNHEALTHY` or  `UNKNOWN` in the response.

aaithal: nice! thanks for confirming that.

aaithal:
Just documenting that #1191 should be resolved when this gets merged

PR #1226, opened 2018-02-02
engine: mtask wire in derived context

adnxn:
### Summary

Wire in `context.Context` where possible for `mtask.waitEvent()` calls, else use `<-chan struct{}` signalling channel instead of `<-chan bool`
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/common_test.go
+++ b/agent/engine/common_test.go
@@ -182,7 +182,6 @@ func addTaskToEngine(t *testing.T,
 	sleepTask *api.Task,
 	mockTime *mock_ttime.MockTime,
 	createStartEventsReported sync.WaitGroup) {
-
 	// steadyStateCheckWait is used to force the test to wait until the steady-state check
 	// has been invoked at least once
 	steadyStateVerify := make(chan time.Time, 1)
@@ -248,5 +247,4 @@ func waitForStopEvents(t *testing.T, stateChangeEvents <-chan statechange.Event,
 		t.Fatal("Should be out of events")
 	default:
 	}
-
 }
--- a/agent/engine/common_test.go
+++ b/agent/engine/common_test.go
@@ -182,22 +182,18 @@ func addTaskToEngine(t *testing.T,
 	sleepTask *api.Task,
 	mockTime *mock_ttime.MockTime,
 	createStartEventsReported sync.WaitGroup) {
+
 	// steadyStateCheckWait is used to force the test to wait until the steady-state check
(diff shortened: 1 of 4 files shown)


agent/engine/common_test.go
```diff
@@ -184,20 +184,15 @@ func addTaskToEngine(t *testing.T,
 	createStartEventsReported sync.WaitGroup) {
 	// steadyStateCheckWait is used to force the test to wait until the steady-state check
 	// has been invoked at least once
-	steadyStateCheckWait := sync.WaitGroup{}
```
aaithal: why was this removed? steadyState check should not have any bearing on wiring contexts to `waitEvent` method, right?

adnxn: I removed this because further down we do a `steadyStateCheckWait.Done()` for a `mocktime.EXPECT()...`, since this mocktime maps to this removed [code](https://github.com/aws/amazon-ecs-agent/pull/1226/files/e42e6b13f0813b3d5ce6f8115ef6751b304d46ba#diff-20346f7af6202828de353f0137b6e43aL264) in the `waitSteady()` method. I don't *think* we're losing any coverage by getting rid of this here. Let me know if you think I'm missing something.

agent/engine/task_manager.go
```diff
@@ -238,12 +241,16 @@ func (mtask *managedTask) waitForHostResources() {
 
 	seelog.Infof("Managed task [%s]: waiting for any previous stops to complete. Sequence number: %d",
 		mtask.Arn, mtask.StartSequenceNumber)
-	othersStopped := make(chan bool, 1)
+
+	othersStoppedCtx, othersStoppedCancel := context.WithCancel(mtask.ctx)
```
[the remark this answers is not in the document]
adnxn: makes sense, it's a small block of code.

agent/engine/common_test.go
richardpen: Since we are getting rid of the `mocktime` for `waitSteady`, how are making sure the `waitSteady` is called? Also can the line 194 be removed?

adnxn: > Also can the line 194 be removed?

We can remove it.

> how are making sure the waitSteady is called? 

Are we losing anything by not ensuring that waitSteady is being called? Since it is a private method and an implementation detail, I'm not sure we need to be explicitly checking if waitSteady is called. Please let me know what you think though.

PR #1227, opened 2018-02-07
Temporary fix for #1127

### LINKED ISSUE #1127: CPU Percent for containers set to 1 on Windows when CPU not defined in task
### Summary
CPU Percent for containers set to 1 on Windows when CPU not defined in task

### Description
Prior to version 1.16.0 Windows tasks definition with CPU not set or set to 0 would receive a value for CpuPercent of 0.

Now those values seem to be getting translated to a percentage of 1 which is causing existing tasks not to launch correctly on that version (fine on versions <=1.15.2).

### Expected Behavior
If CPU is not set in the definition of a container or set to 0, then CpuPercent should not be set either.

### Observed Behavior
When CPU is not set in the definition of a container or set to 0, CpuPercent is set to 1.

### Environment Details
AMI: `Windows_Server-2016-English-Full-ECS_Optimized-2017.11.24 (ami-0b60bc73)`

user data for container instance:
```
<powershell>
[Environment]::SetEnvironmentVariable("ECS_DISABLE_IMAGE_CLEANUP", "true", "Machine")
[Environment]::SetEnvironmentVariable("ECS_IMAGE_MINIMUM_CLEANUP_AGE", "6h", "Machine")
[Environment]::SetEnvironmentVariable("ECS_ENABLE_TASK_CPU_MEM_LIMIT", "false", "Machine")
[Environment]::SetEnvironmentVariable("ECS_ENABLE_CONTAINER_METADATA", "true", "Machine")
Import-Module ECSTools
Initialize-ECSAgent -Cluster 'windows-nonprod-v3' -EnableTaskIAMRole -LogLevel debug
</powershell>
```

docker info:
```
Containers: 2
 Running: 0
 Paused: 0
 Stopped: 2
Images: 3
Server Version: 17.06.2-ee-5
Storage Driver: windowsfilter
 Windows:
Logging Driver: json-file
Plugins:
 Volume: local
 Network: l2bridge l2tunnel nat null overlay transparent
 Log: awslogs etwlogs fluentd json-file logentries splunk syslog
Swarm: inactive
Default Isolation: process
Kernel Version: 10.0 14393 (14393.1794.amd64fre.rs1_release.171008-1615)
Operating System: Windows Server 2016 Datacenter
OSType: windows
Architecture: x86_64
CPUs: 1
Total Memory: 2GiB
Name: EC2AMAZ-N8320N0
ID: WED3:ICYW:F56U:OR3Y:HS7E:N42U:Y4CP:RTLR:QR7V:S23I:GPEN:CBT7
Docker Root Dir: C:\ProgramData\docker
Debug Mode (client): false
Debug Mode (server): false
Registry: https://index.docker.io/v1/
Experimental: false
Insecure Registries:
 127.0.0.0/8
Live Restore Enabled: false
```

Instance Metadata:
```
PS C:\Users\Administrator> Invoke-RestMethod http://localhost:51678/v1/metadata

Cluster            ContainerInstanceArn                                                                       Version
-------            --------------------                                                                       -------
windows-nonprod-v3 arn:aws:ecs:us-west-2:108110855944:container-instance/054fad8f-f648-4a06-b1fa-63b8342f189a Amazon ECS Agent - v1.16.0 (1ca656c)
```

### Supporting Log Snippets
snippet of task definition:
```
        "containerDefinitions": [
            {
                "name": "app",
                "image": "108110855944.dkr.ecr.us-west-2.amazonaws.com/collectionsach/dev:2.1.0.4",
                "cpu": 0,
                "memory": 256,
                "portMappings": [],
                "essential": true,
                "environment": [
                    {
                        "name": "EXIT_DELAY",
                        "value": "60"
                    },
                    {
                        "name": "ENVIRONMENT",
                        "value": "dev"
                    }
                ],
                "mountPoints": [],
                "volumesFrom": [],
                "dnsServers": []
            }
        ]
```

snippet of log showing conversion for CPU shares:
```
2017-12-04T17:31:13Z [INFO] Creating container module="TaskEngine" task="CollectionsACH-dev:26 arn:aws:ecs:us-west-2:108110855944:task/ae7d60f0-1d51-449d-a8e4-83ade6cadce4, TaskStatus: (NONE->RUNNING) Containers: [app (PULLED->RUNNING),]" container="app(108110855944.dkr.ecr.us-west-2.amazonaws.com/collectionsach/dev:2.1.0.4) (PULLED->RUNNING)"
2017-12-04T17:31:13Z [DEBUG] Converting CPU shares to allowed minimum of 2 for task arn: [arn:aws:ecs:us-west-2:108110855944:task/ae7d60f0-1d51-449d-a8e4-83ade6cadce4] and cpu shares: 0
```

snippet of docker inspect for created container:
```
            "CpuShares": 0,
            "Memory": 268435456,
            "NanoCpus": 0,
...
            "CpuPeriod": 0,
            "CpuQuota": 0,
            "CpuRealtimePeriod": 0,
            "CpuRealtimeRuntime": 0,
            "CpusetCpus": "",
            "CpusetMems": "",
...
            "MemoryReservation": 0,
            "MemorySwap": 0,
            "MemorySwappiness": -1,
...
            "CpuCount": 0,
            "CpuPercent": 1,
```

sharanyad:
### Summary
Allow a mix of CPU unbounded and bounded tasks for windows

### Implementation details
- Introduce an agent config variable to allow this behavior 
- if CPU is not set or set to zero, CPU percent and shares of the container will also be zero, else set to 1 if CPU percent calculates to zero 

### Testing

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.17.1-dev
 * Bug - Fixed a bug that was causing a runtime panic by accessing negative index in the health check log slice. [#1239](https://github.com/aws/amazon-ecs-agent/pull/1239)
+* Bug - Workaround for an issue where CPU percent was set to 1 when CPU was not
+  set or set to zero(unbounded) in Windows [#1227](https://github.com/aws/amazon-ecs-agent/pull/1227)
 
 ## 1.17.0
 * Feature - Support a HTTP endpoint for `awsvpc` tasks to query metadata
--- a/README.md
+++ b/README.md
@@ -177,6 +177,7 @@ additional details on each available environment variable.
 | `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the ECS Agent is running as a container. We do not use this value in Windows because the ECS Agent is not running as container in Windows. | `/var/lib/ecs` | `Not used` |
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
 | `ECS_CGROUP_PATH` | `/sys/fs/cgroup` | The root cgroup path that is expected by the ECS agent. This is the path that accessible from the agent mount. | `/sys/fs/cgroup` | Not applicable |
+| `ECS_ENABLE_CPU_UNBOUNDED_WINDOWS_WORKAROUND` | `true` | When `true`, ECS will allow CPU unbounded(CPU=`0`) tasks to run along with CPU bounded tasks in Windows. | Not applicable | `false` |
 
 ### Persistence
 
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -143,6 +143,9 @@ type Task struct {
 	// MemoryCPULimitsEnabled to determine if task supports CPU, memory limits
 	MemoryCPULimitsEnabled bool `json:"MemoryCPULimitsEnabled,omitempty"`
 
+	// platformFields consists of fields specific to linux/windows for a task
+	platformFields platformFields
+
 	// lock is for protecting all fields in the task struct
 	lock sync.RWMutex
 }
@@ -509,20 +512,6 @@ func (task *Task) SetConfigHostconfigBasedOnVersion(container *Container, config
 	}
 
 	return nil
-}
-
-// dockerCPUShares converts containerCPU shares if needed as per the logic stated below:
-// Docker silently converts 0 to 1024 CPU shares, which is probably not what we
-// want.  Instead, we convert 0 to 2 to be closer to expected behavior. The
-// reason for 2 over 1 is that 1 is an invalid value (Linux's choice, not Docker's).
-func (task *Task) dockerCPUShares(containerCPU uint) int64 {
-	if containerCPU <= 1 {
-		seelog.Debugf(
-			"Converting CPU shares to allowed minimum of 2 for task arn: [%s] and cpu shares: %d",
-			task.Arn, containerCPU)
-		return 2
-	}
-	return int64(containerCPU)
 }
 
 func (task *Task) dockerExposedPorts(container *Container) map[docker.Port]struct{} {
--- a/agent/api/task_unix.go
+++ b/agent/api/task_unix.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/cihub/seelog"
 	docker "github.com/fsouza/go-dockerclient"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
@@ -39,6 +40,9 @@ const (
 	minimumCPUPercent = 0
 	bytesPerMegabyte  = 1024 * 1024
 )
+
+// platformFields consists of fields specific to Linux for a task
+type platformFields struct {}
 
 func (task *Task) adjustForPlatform(cfg *config.Config) {
 	task.lock.Lock()
@@ -177,4 +181,18 @@ func (task *Task) overrideCgroupParent(hostConfig *docker.HostConfig) error {
 		hostConfig.CgroupParent = cgroupRoot
 	}
 	return nil
+}
+
+// dockerCPUShares converts containerCPU shares if needed as per the logic stated below:
+// Docker silently converts 0 to 1024 CPU shares, which is probably not what we
+// want.  Instead, we convert 0 to 2 to be closer to expected behavior. The
+// reason for 2 over 1 is that 1 is an invalid value (Linux's choice, not Docker's).
+func (task *Task) dockerCPUShares(containerCPU uint) int64 {
+	if containerCPU <= 1 {
+		seelog.Debugf(
+			"Converting CPU shares to allowed minimum of 2 for task arn: [%s] and cpu shares: %d",
+			task.Arn, containerCPU)
+		return 2
+	}
+	return int64(containerCPU)
 }
--- a/agent/api/task_windows.go
+++ b/agent/api/task_windows.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/cihub/seelog"
 	docker "github.com/fsouza/go-dockerclient"
 )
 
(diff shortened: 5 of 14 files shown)


sharanyad:
this addresses https://github.com/aws/amazon-ecs-agent/issues/1127

agent/api/task.go
```diff
@@ -145,6 +145,8 @@ type Task struct {
 
 	// lock is for protecting all fields in the task struct
 	lock sync.RWMutex
+
+	cpuUnboundedWindowsEnabled bool
```
aaithal: `cpuUnboundedWindowsEnabled` is quite a long name. Can you make it shorter? `cpuUnboundedWindows` perhaps? `Enabled` seems redundant since its already a `bool`.

Also, the right thing to do here is to probably create `Task` struct in `task_windows.go` and make this a part of that struct. But, we can do that in a separate PR.

agent/api/task_windows.go
```diff
@@ -69,8 +71,13 @@ func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig) erro
 	// Convert the CPUShares to CPUPercent
 	hostConfig.CPUPercent = hostConfig.CPUShares * percentageFactor / int64(cpuShareScaleFactor)
 	if hostConfig.CPUPercent == 0 {
```
aaithal: Can you refactor this as `if hostConfig.CPUPercent == 0 && hostConfig.CPUShares != 0 `?

agent/config/config_windows.go
```diff
@@ -94,6 +95,8 @@ func (cfg *Config) platformOverrides() {
 
 	// ensure TaskResourceLimit is disabled
 	cfg.TaskCPUMemLimit = ExplicitlyDisabled
+
+	cfg.CPUUnboundedWindowsEnabled = utils.ParseBool(os.Getenv("ECS_ENABLE_CPU_UNBOUNDED_WINDOWS_WORKAROUND"), false)
```
aaithal: do we really need the string `WORKAROUND` in the name here? Seems unnecessary.

sharanyad: suggestion by @samuelkarp to name it this way was to convey this could be a temporary fix/hack, until cpu shares are fixed in windows. Moreover allowing both cpu unbounded and bounded tasks in an instance can have serious consequences like bounded tasks not getting their allotted shares and 'workaround' may warn the customers to use it with caution. wdyt?

aaithal: Just having a flag named `ECS_ENABLE_CPU_UNBOUNDED_WINDOWS` would probably indicate that this is a workaround imo. I don't have strong opinions here. The env var name does seem too long, however.

samuelkarp: I'd like the env var name to convey that we might remove it as a configuration option at some point (i.e., when we fix the CPU bursting behavior).  I think `WORKAROUND` or `HACK` is helpful in pointing out that this is not an intended long-term approach.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
```
aaithal: We're going with the convention of `next-version.dev` instead of the `Unreleased` tag. So, this should probably be `1.17.1-dev`

README.md
```diff
@@ -177,6 +177,7 @@ additional details on each available environment variable.
 | `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the ECS Agent is running as a container. We do not use this value in Windows because the ECS Agent is not running as container in Windows. | `/var/lib/ecs` | `Not used` |
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
 | `ECS_CGROUP_PATH` | `/sys/fs/cgroup` | The root cgroup path that is expected by the ECS agent. This is the path that accessible from the agent mount. | `/sys/fs/cgroup` | Not applicable |
+| `ECS_ENABLE_CPU_UNBOUNDED_WINDOWS_HACK` | `true` | When `true`, ECS will allow CPU unbounded(CPU=`0`) tasks to run along with CPU bounded tasks in Windows. | Not applicable | `false` |
```
aaithal: let's not have the word `HACK` anywhere in these flags.

agent/api/task_unix.go
```diff
@@ -40,6 +42,89 @@ const (
 	bytesPerMegabyte  = 1024 * 1024
 )
 
+// Task is the internal representation of a task in the ECS agent
+type Task struct {
```
aaithal: All of this code duplication is leading me to think that may be this is not the best way to do this. Because, adding a new field to the task now requires modifications in 2 different files. Also, 2 * code = 2 * bugs. How about something like this:
```
// task.go, no build flags
type Task struct {
 ... // all common fields
 additionalFields
}

// task_unix.go, with the negative windows platform flag
type additionalFields struct {
 // empty for now
}

// task_windows.go, with the windows build flag
type additionalFields struct {
 cpuUnboundedEnabled bool
}
```

If that sounds reasonable, we should rewrite `config.go` in a similar vein as well. I'm sure you can come with a better name for `additionalFields` :)

agent/api/task_windows.go
```diff
@@ -35,9 +37,98 @@ const (
...
+
+	// ENI is the elastic network interface specified by this task
+	ENI *ENI
+
+	// MemoryCPULimitsEnabled to determine if task supports CPU, memory limits
+	MemoryCPULimitsEnabled bool `json:"MemoryCPULimitsEnabled,omitempty"`
+
+	// cpuUnboundedEnabled determines whether a mix of unbounded and bounded CPU tasks
+	// are allowed to run in the instance
+	cpuUnboundedEnabled bool
```
aaithal: `cpuUnbounded` is good enough i think.

agent/api/task_unix.go
petderek: I think we should try to use the same struct. The Task API from the end user perspective is going to be the same, and forking it to have different implementations internally may become confusing down the road for us. What would you think about simply ignoring unrelated / unimplemented fields on each platform? This is the pattern used in the Docker API, and it works pretty well.

agent/config/types_windows.go
```diff
@@ -0,0 +1,22 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package config
+
+// PlatformVariables consists of configuration variables specific to Windows
+type PlatformVariables struct {
+		// CPUUnbounded specifies if agent can run a mix of CPU bounded and
```
aaithal: nit: tabbing seems off here.

PR #1238, opened 2018-02-16
#1144 - > Do not pass the MemoryReservation parameter to docker on windows

### LINKED ISSUE #1144: Soft limit should not pass the memory reservation parameter on Windows
### Summary
Since the new ECS version (1.16), the memory reservation parameter is passed to docker when setting a soft memory limit. However, on Windows, docker sends the following error: 
docker: Error response from daemon: invalid option: Windows does not support MemoryReservation.

### Description
This is a breaking change for our use case since we are now forced to set a hard limit which is something we not want to do.

### Environment Details
ECS agent v1.16
Docker 17.06.2-ee-5
Windows

julienduchesne:
### Summary
This fixes #1144

### Implementation details
task_windows.go removes the MemoryReservation HostConfig parameter.

### Testing

### DIFF
--- a/agent/api/task_windows.go
+++ b/agent/api/task_windows.go
@@ -73,6 +73,10 @@ func (task *Task) platformHostConfigOverride(hostConfig *docker.HostConfig) erro
 		hostConfig.CPUPercent = minimumCPUPercent
 	}
 	hostConfig.CPUShares = 0
+
+	// As of version  17.06.2-ee-6 of docker. MemoryReservation is not supported on windows. This ensures that
+	// this parameter is not passed, allowing to launch a container without a hard limit.
+	hostConfig.MemoryReservation = 0
 	return nil
 }
 
--- a/agent/api/task_windows_test.go
+++ b/agent/api/task_windows_test.go
@@ -39,6 +39,9 @@ const (
 	expectedEmptyVolumeContainerCmd   = "not-applicable"
 
 	expectedMemorySwappinessDefault = memorySwappinessDefault
+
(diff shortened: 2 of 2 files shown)


adnxn:
@julienduchesne thanks for the PR. I'm changing the base branch diff be against `dev`

richardpen:
@julienduchesne Thanks for the PR, I think without the soft limit memory support from Docker on Windows, it's hard to satisfy all the use cases in issue #1144. As some of the parameters in task definition are platform specific, ideally we shouldn't send tasks to the platform that are not supported. This is currently a gap we are trying to improve. After we make this improvement, the task definition with `memoryReservation` will not be placed on the Windows platform until we support this parameter on Windows. Setting it to zero in the agent today will let you get rid of the error from Docker `docker: Error response from daemon: invalid option: Windows does not support MemoryReservation`, but it also means there is no memory limit for the container, it can consume all the available memory on the instance which may lead to unexpected and sometimes undesirable behavior of applications/containers being starved of memory due to contention.

In ECS task definition, one or both of the `memory`(hard limit) and `memoryReservation`(soft limit) is required, and both of them are used to make the decision in task placement. It will cause confusion for the customer if we use `memoryReservation` only for task placement while leaving the container with unlimited memory. Some of the options that might be better in this scenario are: [1] disable the per container memory limit on Windows while adding another parameter for scheduling. Or, [2] wait for the memoryReservation to be supported on Windows. In either case, we need to make some change in the task definition and the agent.

I'm closing this PR and we will update on the #1144 when we have a more generic solution.

Thanks,
Peng

Issue #1237, opened 2018-02-16
runtime error: index out of range


cgouteix:
### Summary

After upgrading to 1.17.0, the agent container is enable to start on some randomly instances.

### Description

I'm using ECS and an EC2 Autoscaling Group. All my containers have docker's healthcheck on.

I'm using  ami-63cbae0c (amzn-ami-2017.09.h-amazon-ecs-optimized) in eu-central-1. I was on ami-eacf5d85 (amzn-ami-2017.09.f-amazon-ecs-optimized) with no such trouble.

After upgrading to 1.17.0 I get this behavior randomly (some EC2 instances and all):

- Tasks could not be started with this event message : `service xxx was unable to place a task because no container instance met all of its requirements. The closest matching container-instance xxx encountered error "AGENT". For more information, see the Troubleshooting section.`
- In the ECS Instances tab, the "Agent Connected" column says `false` for the previously mentionned instance
- After login in the instance, I can see that the `ecs-agent` container is perpetually restarting. Its last log message is `panic: runtime error: index out of range` (see the trace bellow)

### Environment Details
```
$ docker info
Containers: 35
 Running: 29
 Paused: 0
 Stopped: 6
Images: 6
Server Version: 17.09.1-ce
Storage Driver: devicemapper
 Pool Name: docker-docker--pool
 Pool Blocksize: 524.3kB
 Base Device Size: 10.74GB
 Backing Filesystem: ext4
 Data file: 
 Metadata file: 
 Data Space Used: 2.904GB
 Data Space Total: 23.33GB
 Data Space Available: 20.42GB
 Metadata Space Used: 2.707MB
 Metadata Space Total: 25.17MB
 Metadata Space Available: 22.46MB
 Thin Pool Minimum Free Space: 2.333GB
 Udev Sync Supported: true
 Deferred Removal Enabled: true
 Deferred Deletion Enabled: true
 Deferred Deleted Device Count: 0
 Library Version: 1.02.135-RHEL7 (2016-11-16)
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 06b9cb35161009dcb7123345749fef02f7cea8e0
[...]

PR #1239, opened 2018-02-16
engine: docker healthcheck logLength index fix

adnxn:
### Summary

This change addresses https://github.com/aws/amazon-ecs-agent/issues/1237 by adding a zero check before accessing the `Health.Log` slice. Also adds corresponding test.

The issue was surfaced when the agent recieved a Docker health check message indicating the container was `unhealthy` but sent an empty `[]HealthCheck` in `Health.Log`. 

We have not reproduced the issue, but are handling this case by doing a length check on `Health.Log`.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.17.1-dev
+* Bug - Fixed a bug that was causing a runtime panic by accessing negative index in the health check log slice. [#1239](https://github.com/aws/amazon-ecs-agent/pull/1239)
+
 ## 1.17.0
 * Feature - Support a HTTP endpoint for `awsvpc` tasks to query metadata
 * Feature - Support Docker health check
--- a/agent/engine/docker_client.go
+++ b/agent/engine/docker_client.go
@@ -45,6 +45,8 @@ const (
 	imageNameFormat = "%s:%s"
 	// the buffer size will ensure agent doesn't miss any event from docker
 	dockerEventBufferSize = 100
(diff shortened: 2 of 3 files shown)


jahkeup:
I was able to reproduce this state of the health check using this Dockerfile:

```dockerfile
FROM debian:latest
HEALTHCHECK --interval=3s --timeout=1s --start-period=56s CMD /bin/true
CMD ["/bin/sleep", "60"]
```

And running it using the following commands:

```bash
docker build -t healthchecked .
docker run -d --name healthchecked healthchecked
# wait for container to start
sleep 2 
# check the container's state for health checks observed:
docker inspect -f '{{.State.Health}}' healthchecked
```

Which shows the health state with no checks performed and still in the starting state:

```
{starting 0 []}
```

aaithal:
@jahkeup  in your example, the container's health status is `starting` and not actually `unhealthy` right?

jahkeup:
Yep, I see where we're checking the logs now. Let me try this out with the Agent running to see if there's some other way that the logs could be empty.

jahkeup:
Doesn't appear that we could easily get into this state, the Docker daemon [always appends to its Log field when handling a check result](https://github.com/docker/docker-ce/blob/4db928a87dfbff4f97f30e31f3178f99147c2163/components/engine/daemon/health.go#L138-L142) and this is also where it's setting the unhealthy/healthy status.

aaithal:
@adnxn  don't forget to add an entry in the chanelog before merging this.

Issue #1231, opened 2018-02-13
Aws::Errors::MissingCredentialsError: unable to sign request without credentials set


jcarr-sailthru:
There is an issue with the latest ECS Agent 1.17.0 AMI which is breaking authentication for one of our Ruby applications in an inconsistent manner.

This fault only occurred for us after upgrading to the ECS agent 1.17.0 using AWS AMI amzn-ami-2017.09.h-amazon-ecs-optimized (ami-5e414e24) but does not occur on earlier images, such as amzn-ami-2017.09.e-amazon-ecs-optimized (ami-13401669) with ECS agent 1.16.1.

Note that we're not sure yet if this is the agent's fault (could be something else in the AMI, eg Docker, Kernel, or some bug in the Ruby SDK that is only triggered with the newer ECS Agent).

We have AWS Business Support, so have logged a more formal support ticket, but I wanted to log an issue here incase anyone else is experiencing issues and looking online using this error string to combined knowledge.

Demonstration of the issue:
```
1 Uploading s3://example/530b1815-01dc-4add-b89a-4db0cf3da660...
2 Uploading s3://example/251ea219-c480-4857-94cb-0736a2a66ac8...
3 Uploading s3://example/0949cec1-e6fa-4a3a-b4bc-2d2da5a6a024...
4 Uploading s3://example/31efdb30-1a60-4243-b4c6-0d46ea09fe57...
5 Uploading s3://example/1c78dbbc-aa01-42c7-994a-8b06ece260c5...
6 Uploading s3://example/e0a3c434-25e7-4a36-85e0-ca193c90b546...
7 Uploading s3://example/238a685d-325b-4925-8b52-751847534681...
8 Uploading s3://example/e1561427-16d0-444a-ad78-4e884fc5ab28...
9 Uploading s3://example/b83e9cd4-f54c-499b-a038-445d8a30d115...
10 Uploading s3://example/de1fdd18-1b92-46e9-8961-f6fd670148c5...
11 Uploading s3://example/d5183b91-3b21-412f-b370-7d53408e7479...
12 Uploading s3://example/18ea17df-8ed5-4572-bb32-ecfcf92eb06d...
13 Uploading s3://example/e23bcdc7-5105-4ae0-8e16-74737b2fc093...
14 Uploading s3://example/cdfe7b07-0e7e-489a-b560-1299734f25bb...
15 Uploading s3://example/57a6e6eb-3858-4e25-a5d0-371198cf79d3...
[...]

PR #1240, opened 2018-02-17
Increase steady state and burst throttles for task endpoints

sharanyad:
## Summary
This addresses https://github.com/aws/amazon-ecs-agent/issues/1231 

### Implementation details
default steady state throttle of 40 rps and burst of 60 rps.
values configurable through `ECS_TASK_METADATA_RPS_LIMIT` environment variable.

### Testing

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,7 @@ additional details on each available environment variable.
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
 | `ECS_CGROUP_PATH` | `/sys/fs/cgroup` | The root cgroup path that is expected by the ECS agent. This is the path that accessible from the agent mount. | `/sys/fs/cgroup` | Not applicable |
 | `ECS_ENABLE_CPU_UNBOUNDED_WINDOWS_WORKAROUND` | `true` | When `true`, ECS will allow CPU unbounded(CPU=`0`) tasks to run along with CPU bounded tasks in Windows. | Not applicable | `false` |
+| `ECS_TASK_METADATA_RPS_LIMIT` | `100,150` | Comma separated integer values for steady state and burst throttle limits for task metadata endpoint | `40,60` | `40,60` |
 
 ### Persistence
 
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -93,6 +93,14 @@ const (
 
 	// pauseContainerTarball is the path to the pause container tarball
 	pauseContainerTarballPath = "/images/amazon-ecs-pause.tar"
+
+	// DefaultTaskMetadataSteadyStateRate is set as 40. This is arrived from our benchmarking
+	// results where task endpoint can handle 4000 rps effectively. Here, 100 containers
+	// will be able to send out 40 rps.
+	DefaultTaskMetadataSteadyStateRate = 40
+
+	// DefaultTaskMetadataBurstRate is set to handle 60 burst requests at once
+	DefaultTaskMetadataBurstRate = 60
 )
 
 var (
@@ -346,6 +354,7 @@ func environmentConfig() (Config, error) {
 	}
 	containerMetadataEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_CONTAINER_METADATA"), false)
 	dataDirOnHost := os.Getenv("ECS_HOST_DATA_DIR")
+	steadyStateRate, burstRate := getTaskMetadataThrottles()
 
 	if len(errs) > 0 {
 		err = utils.NewMultiError(errs...)
@@ -391,6 +400,8 @@ func environmentConfig() (Config, error) {
 		DataDirOnHost:                    dataDirOnHost,
 		OverrideAWSLogsExecutionRole:     overrideAWSLogsExecutionRoleEnabled,
 		CgroupPath:                       cgroupPath,
+		TaskMetadataSteadyStateRate:      steadyStateRate,
+		TaskMetadataBurstRate:            burstRate,
 	}, err
 }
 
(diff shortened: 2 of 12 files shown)


yunhee-l:
looks like gocyclo failed? if/else statements with || tends to push the cyclomatic complexity up, which we currently fail if a method has complexity > 15.

agent/config/config.go
```diff
@@ -434,6 +445,25 @@ func getTaskCPUMemLimitEnabled() Conditional {
 	return taskCPUMemLimitEnabled
 }
 
+func getTaskMetatadataRpsLimits() (int, int) {
+	var rpsLimit, rpsBurstLimit int
+	rpsLimitEnvVal := os.Getenv("ECS_TASK_METADATA_RPS_LIMIT")
+	rpsLimitSplits := strings.Split(rpsLimitEnvVal, ",")
+	if len(rpsLimitSplits) != 2 {
+		seelog.Warnf("Invalid format for \"ECS_TASK_METADATA_RPS_LIMIT\", expected: \"rateLimit,burst\"")
```
aaithal: nit: you can also do something like this to avoid these escapes:
```
seelog.Warnf(`Invalid format for "ECS_TASK_METADATA_RPS_LIMIT", expected: "rateLimit,burst"`)
```

Also, you can just use `seelog.Warn()` here.

agent/config/config.go
```diff
@@ -434,6 +445,25 @@ func getTaskCPUMemLimitEnabled() Conditional {
...
+	} else {
```
aaithal: just `return` here so that you don't need an additional `else` block. Avoids one level of indentation

agent/config/config.go
```diff
@@ -434,6 +445,25 @@ func getTaskCPUMemLimitEnabled() Conditional {
...
+		rpsLimitVal, err := strconv.Atoi(strings.TrimSpace(rpsLimitSplits[0]))
+		rpsBurstLimitVal, err1 := strconv.Atoi(strings.TrimSpace(rpsLimitSplits[1]))
+		if err != nil || err1 != nil {
```
aaithal: This kind of error handling makes it harder to know which field exactly caused the error.  Doing something like this is much simpler imo:

```
rpsLimit, err := strconv.Atoi(strings.TrimSpace(rpsLimitSplits[0]))
if err != nil {
 ...
 return 0, 0
}

rpsBurstLimit, err := strconv.Atoi(strings.TrimSpace(rpsLimitSplits[1]))
if err != nil {
 ...
 return 0, 0
}

return rpsLimit, rpsBurstLimit
```

agent/config/config.go
```diff
@@ -93,6 +93,14 @@ const (
 
 	// pauseContainerTarball is the path to the pause container tarball
 	pauseContainerTarballPath = "/images/amazon-ecs-pause.tar"
+
+	// DefaultRpsLimit is set as 40. This is arrived from our benchmarking
+	// results where task endpoint can handle 4000 rps effectively. Here, 100 containers
+	// will be able to send out 40 rps.
+	DefaultRpsLimit = 40
```
aaithal: `DefaultRpsLimit` is a bit vague. Similar to above.  Can you please rename this to something like `DefaultTaskMetadataSteadyStateRate`?

agent/config/config.go
```diff
@@ -93,6 +93,14 @@ const (
...
+
+	// DefaultRpsBurstLimit is set to handle 60 burst requests at once
+	DefaultRpsBurstLimit = 60
```
aaithal: Similar to above. Can you please rename this to something like `DefaultTaskMetadataBurstRate`?

agent/config/config.go
```diff
@@ -434,6 +445,25 @@ func getTaskCPUMemLimitEnabled() Conditional {
 	return taskCPUMemLimitEnabled
 }
 
+func getTaskMetatadataRpsLimits() (int, int) {
```
aaithal: `getTaskMetadataThrottles()` is a better name for this.

agent/config/types.go
```diff
@@ -198,4 +198,10 @@ type Config struct {
 	// CgroupPath is the path expected by the agent, defaults to
 	// '/sys/fs/cgroup'
 	CgroupPath string
+
+	// RpsLimit specifies the steady state throttle for the task metadata endpoint
+	RpsLimit int
```
aaithal: Similar to above.  Can you please rename this to something like `TaskMetadataSteadyStateRate`?

agent/config/types.go
```diff
@@ -198,4 +198,10 @@ type Config struct {
...
+
+	// RpsBurstLimit specifies the burst rate throttle for the task metadata endpoint
+	RpsBurstLimit int
```
aaithal: Similar to above.  Can you please rename this to something like `TaskMetadataBurstRate`?

agent/config/config.go
```diff
@@ -346,6 +354,7 @@ func environmentConfig() (Config, error) {
 	}
 	containerMetadataEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_CONTAINER_METADATA"), false)
 	dataDirOnHost := os.Getenv("ECS_HOST_DATA_DIR")
+	rpsLimit, rpsBurstLimit := getTaskMetatadataRpsLimits()
```
aaithal: `steadyStateRate` and `burstRate` are better names imo

agent/config/config.go
```diff
@@ -434,6 +445,25 @@ func getTaskCPUMemLimitEnabled() Conditional {
 	return taskCPUMemLimitEnabled
 }
 
+func getTaskMetatadataRpsLimits() (int, int) {
+	var rpsLimit, rpsBurstLimit int
+	rpsLimitEnvVal := os.Getenv("ECS_TASK_METADATA_RPS_LIMIT")
```
aaithal: Can you also add an entry in README and changelog for this?

agent/config/config.go
adnxn: maybe i missed an earlier conversation. why don't we want two separate env variables for this? `ECS_TASK_METADATA_RPS_LIMIT` and `ECS_TASK_METADATA_RPS_BURST_LIMIT`?

agent/config/config.go
sharanyad: This block got moved into this method from the original `environmentConfig()` and was never refactored. Will make this change.

agent/config/config.go
sharanyad: Since steady state and burst throttles both are needed and related, it would be convenient for customers to have a single environment variable.

adnxn: hm i don't feel strongly either way, my concern was mostly if it would add extra overhead of enforcing ordering of the throttle limits (for the customer) and extra validation code (for us). but what you said about "needed and related" also makes sense. okay thanks.

agent/config/config.go
```diff
@@ -434,6 +445,27 @@ func getTaskCPUMemLimitEnabled() Conditional {
 	return taskCPUMemLimitEnabled
 }
 
+func getTaskMetadataThrottles() (int, int) {
+	var steadyStateRate, burstRate int
+	rpsLimitEnvVal := os.Getenv("ECS_TASK_METADATA_RPS_LIMIT")
+	rpsLimitSplits := strings.Split(rpsLimitEnvVal, ",")
+	if len(rpsLimitSplits) != 2 {
+		seelog.Warn(`Invalid format for "ECS_TASK_METADATA_RPS_LIMIT", expected: "rateLimit,burst"`)
+		return steadyStateRate, burstRate
```
adnxn: nit: for consistency, should we explicitly `return 0, 0` as you do for errors reading input in the two cases below?

README.md
```diff
@@ -177,6 +177,8 @@ additional details on each available environment variable.
 | `ECS_HOST_DATA_DIR` | `/var/lib/ecs` | The source directory on the host from which ECS_DATADIR is mounted. We use this to determine the source mount path for container metadata files in the case the ECS Agent is running as a container. We do not use this value in Windows because the ECS Agent is not running as container in Windows. | `/var/lib/ecs` | `Not used` |
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
 | `ECS_CGROUP_PATH` | `/sys/fs/cgroup` | The root cgroup path that is expected by the ECS agent. This is the path that accessible from the agent mount. | `/sys/fs/cgroup` | Not applicable |
+| `ECS_TASK_METADATA_RPS_LIMIT` | `100,150` | Comma separated values for steady state and burst throttle limits for task metadata endpoint | `40,60` | `40,60` |
```
adnxn: nit: specify integer value w/ "Comma separated integer values for..."

PR #1251, opened 2018-02-21
Enforce the telemetry functional tests to check resource usage

richardpen:
### Implementation details

Run a task that uses specific amount of resource(cpu/memory), and verify the metrics from cloudwatch matches the actually usage.

### Testing

New tests cover the changes: 

### Licensing


agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -429,6 +435,98 @@ func containerHealthWithStartPeriodTest(t *testing.T, taskDefinition string) {
...
+
+	params.MetricName = aws.String("MemoryUtilization")
+	_, err = VerifyMetrics(cwclient, params, true)
+	assert.NoError(t, err, "Before task running, verify metrics for memory utilization failed")
+
+	cpuNum := runtime.NumCPU()
+
+	tdOverrides := make(map[string]string)
+	// Set the container cpu percentage 25%
+	tdOverrides["$$$$CPUSHARE$$$$"] = strconv.Itoa(int(float64(cpuNum*cpuSharesPerCore) * 0.25))
```
sharanyad: This would still be CPU shares and not percent in Linux right?

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -429,6 +435,98 @@ func containerHealthWithStartPeriodTest(t *testing.T, taskDefinition string) {
...
+	require.NoError(t, err, "Error wait telemetry task running")
+
+	time.Sleep(waitMetricsInCloudwatchDuration)
+	params.EndTime = aws.Time(RoundTimeUp(time.Now(), time.Minute).UTC())
+	params.StartTime = aws.Time((*params.EndTime).Add(-waitMetricsInCloudwatchDuration).UTC())
+	params.MetricName = aws.String("CPUUtilization")
+	metrics, err := VerifyMetrics(cwclient, params, false)
+	assert.NoError(t, err, "Task is running, verify metrics for CPU utilization failed")
+	// Also verify the cpu usage is around 25%
+	assert.InDelta(t, 25, *metrics.Average, 5)
```
sharanyad: in Linux, CPU shares are respected only when there is bottleneck in CPU usage. 
Here since this is the only task running in the instance and we are trying to consume 100% CPU (in stress), wouldn't the CPU be fully utilized (100%)?

agent/functional_tests/tests/functionaltests_test.go
richardpen: No, this isn't the cpu limitation in docker, it is the cpu unit in cgroup documented [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size).

agent/functional_tests/tests/functionaltests_test.go
richardpen: As per comments above, this isn't cpu shares, the cpu limitation will be converted to something similar to "percent" limitation in cgroup.

sharanyad: Got it, missed the change where cpu was moved to task level. Thanks for clarifying.

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -429,6 +435,98 @@ func containerHealthWithStartPeriodTest(t *testing.T, taskDefinition string) {
...
+	err = testTask.WaitRunning(waitTaskStateChangeDuration)
+	require.NoError(t, err, "Error wait telemetry task running")
+
+	time.Sleep(waitMetricsInCloudwatchDuration)
+	params.EndTime = aws.Time(RoundTimeUp(time.Now(), time.Minute).UTC())
+	params.StartTime = aws.Time((*params.EndTime).Add(-waitMetricsInCloudwatchDuration).UTC())
+	params.MetricName = aws.String("CPUUtilization")
+	metrics, err := VerifyMetrics(cwclient, params, false)
+	assert.NoError(t, err, "Task is running, verify metrics for CPU utilization failed")
+	// Also verify the cpu usage is around 25%
```
adnxn: nit, very minor: mind adding to comment about `+/- 5%`

adnxn: nit, very minor: same with the `.InDelta()` call below for memory utilization.

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -429,6 +435,98 @@ func containerHealthWithStartPeriodTest(t *testing.T, taskDefinition string) {
...
+	params.StartTime = aws.Time((*params.EndTime).Add(-waitMetricsInCloudwatchDuration).UTC())
+	params.MetricName = aws.String("CPUUtilization")
+	_, err = VerifyMetrics(cwclient, params, true)
+	assert.NoError(t, err, "Task stopped: verify metrics for CPU utilization failed")
+
+	params.MetricName = aws.String("MemoryUtilization")
+	metrics, err = VerifyMetrics(cwclient, params, true)
+	memInfo, err := system.ReadMemInfo()
+	require.NoError(t, err, "Acquiring system info failed")
+	totalMemory := memInfo.MemTotal / 1024 / 1024
```
aaithal: nit: can you please const this?

PR #1254, opened 2018-02-23
engine: container health integ test

sharanyad:
### Summary
Container health check integ test

### Implementation details
Start task with docker health check enabled 
Verify health status set

### Testing

### Licensing

### DIFF
--- a/agent/engine/engine_integ_test.go
+++ b/agent/engine/engine_integ_test.go
@@ -15,14 +15,13 @@
 package engine
 
 import (
+	"context"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
 	"time"
-
-	"context"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/config"
@@ -40,8 +39,9 @@ import (
 )
 
 const (
-	testDockerStopTimeout  = 2 * time.Second
-	credentialsIDIntegTest = "credsid"
+	testDockerStopTimeout       = 2 * time.Second
+	credentialsIDIntegTest      = "credsid"
+	waitTaskStateChangeDuration = 2 * time.Minute
 )
 
 func init() {
@@ -283,4 +283,47 @@ func TestStartStopWithCredentials(t *testing.T) {
 	// credentials id set in the task
 	_, ok := credentialsManager.GetTaskCredentials(credentialsIDIntegTest)
 	assert.False(t, ok, "Credentials not removed from credentials manager for stopped task")
+}
+
+func TestContainerHealthCheck(t *testing.T) {
+	taskEngine, done, _ := setupWithDefaultConfig(t)
+	defer done()
+
+	stateChangeEvents := taskEngine.StateChangeEvents()
+
+	taskArn := "testTaskWithHealthCheck"
+	testTask := createTestHealthCheckTask(taskArn)
+
+	go taskEngine.AddTask(testTask)
+
+	verifyContainerRunningStateChange(t, taskEngine)
+	verifyTaskIsRunning(stateChangeEvents, testTask)
+
+	waitForContainerHealthStatus(t, testTask)
+	healthStatus := testTask.Containers[0].GetHealthStatus()
+	assert.Equal(t, "HEALTHY", healthStatus.Status.BackendStatus(), "container health status is not HEALTHY")
+
+	taskUpdate := createTestTask(taskArn)
+	taskUpdate.SetDesiredStatus(api.TaskStopped)
+	go taskEngine.AddTask(taskUpdate)
+
+	verifyContainerStoppedStateChange(t, taskEngine)
+	verifyTaskIsStopped(stateChangeEvents, testTask)
+}
+
+func waitForContainerHealthStatus(t *testing.T, testTask *api.Task) {
+	ctx, _ := context.WithTimeout(context.TODO(), waitTaskStateChangeDuration)
+	for {
+		select {
+		case <-ctx.Done():
+			t.Error("Timed out waiting for container health status")
+		default:
+			healthStatus := testTask.Containers[0].GetHealthStatus()
+			if healthStatus.Status.BackendStatus() == "UNKNOWN" {
+				time.Sleep(time.Second)
+				continue
+			}
+			return
+		}
+	}
 }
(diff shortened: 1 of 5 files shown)


agent/engine/engine_integ_test.go
```diff
@@ -284,3 +283,32 @@ func TestStartStopWithCredentials(t *testing.T) {
...
+	defer done()
+
+	stateChangeEvents := taskEngine.StateChangeEvents()
+
+	taskArn := "testTaskWithHealthCheck"
+	testTask := createTestHealthCheckTask(taskArn)
+
+	go taskEngine.AddTask(testTask)
+
+	event := <-stateChangeEvents
```
aaithal: you can use the `verify[Container/Task][Running|Stopped]` here and in the subsequent block.

agent/engine/engine_integ_test.go
```diff
@@ -284,3 +283,32 @@ func TestStartStopWithCredentials(t *testing.T) {
...
+
+	event = <-stateChangeEvents
+	assert.Equal(t, event.(api.TaskStateChange).Status, api.TaskRunning, "Expected task to be RUNNING")
+
+	// Health check startPeriod is 100 ms and interval is 100 ms, sleep for 200 ms to check status
+	time.Sleep(200 * time.Millisecond)
+	healthStatus := testTask.Containers[0].GetHealthStatus()
+	assert.Equal(t, "HEALTHY", healthStatus.Status.BackendStatus(), "container health status is not HEALTHY")
+
+	taskUpdate := *testTask
```
aaithal: instead of copying task struct, it's much more preferable to use `createTestHealthCheckTask(taskArn)` here as well IMO. avoids things like copying mutexes etc.

agent/engine/engine_integ_test.go
```diff
@@ -295,20 +295,16 @@ func TestContainerHealthCheck(t *testing.T) {
 
 	go taskEngine.AddTask(testTask)
 
-	event := <-stateChangeEvents
-	assert.Equal(t, event.(api.ContainerStateChange).Status, api.ContainerRunning, "Expected container to be RUNNING")
-
-	event = <-stateChangeEvents
-	assert.Equal(t, event.(api.TaskStateChange).Status, api.TaskRunning, "Expected task to be RUNNING")
+	verifyTaskIsRunning(stateChangeEvents, testTask)
```
aaithal: Let's not use `verifyTaskIsRunning` anymore since it's slated for deprecation. Please use https://github.com/aws/amazon-ecs-agent/blob/1b9439dda781fd8aa4b5f6e5750b23774f7a930a/agent/engine/engine_integ_test.go#L100 and https://github.com/aws/amazon-ecs-agent/blob/1b9439dda781fd8aa4b5f6e5750b23774f7a930a/agent/engine/engine_integ_test.go#L107 instead

Same goes for `verifyTaskIsStopped` later in the test.

agent/engine/engine_integ_test.go
```diff
@@ -284,3 +283,29 @@ func TestStartStopWithCredentials(t *testing.T) {
...
+
+	stateChangeEvents := taskEngine.StateChangeEvents()
+
+	taskArn := "testTaskWithHealthCheck"
+	testTask := createTestHealthCheckTask(taskArn)
+
+	go taskEngine.AddTask(testTask)
+
+	verifyContainerRunningStateChange(t, taskEngine)
+	verifyTaskIsRunning(stateChangeEvents, testTask)
```
adnxn: i think we want to use [verifyTaskRunningStateChange](https://github.com/aws/amazon-ecs-agent/blob/1b9439dda781fd8aa4b5f6e5750b23774f7a930a/agent/engine/engine_integ_test.go#L107) here.

agent/engine/engine_integ_test.go
```diff
@@ -284,3 +283,29 @@ func TestStartStopWithCredentials(t *testing.T) {
...
+	time.Sleep(healthCheckWaitTime)
+	healthStatus := testTask.Containers[0].GetHealthStatus()
+	assert.Equal(t, "HEALTHY", healthStatus.Status.BackendStatus(), "container health status is not HEALTHY")
+
+	taskUpdate := createTestTask(taskArn)
+	taskUpdate.SetDesiredStatus(api.TaskStopped)
+	go taskEngine.AddTask(taskUpdate)
+
+	verifyContainerStoppedStateChange(t, taskEngine)
+	verifyTaskIsStopped(stateChangeEvents, testTask)
```
adnxn: i think we want to use [verifyTaskStoppedStateChange](https://github.com/aws/amazon-ecs-agent/blob/1b9439dda781fd8aa4b5f6e5750b23774f7a930a/agent/engine/engine_integ_test.go#L121) here.

adnxn: nit: since the `verifyTaskIs*` functions are slated for deprecation, should we tag the function name to reflect that in a separate commit? might prevent use of `verifyTaskIs*` slipping through reviews. @aaithal, what do you think?

agent/engine/engine_integ_test.go
sharanyad: `verifyTaskIsRunning` does the same state change event check as `verifyTaskRunningStateChange` and more like checking the task's ARN. So it would be preferable to use the former.

agent/engine/engine_integ_test.go
sharanyad: I am not sure if they should be deprecated since they also do a check for the right task's state change events by verifying the ARN. We should probably decide on a standard for using either the verify[Container/Task][Started/Stopped]StateChange or verifyTaskIs* methods, or keep the verifyTaskIs* and create similar ones for container state change events too.

agent/engine/engine_integ_test.go
```diff
@@ -284,3 +283,29 @@ func TestStartStopWithCredentials(t *testing.T) {
...
+
+	taskArn := "testTaskWithHealthCheck"
+	testTask := createTestHealthCheckTask(taskArn)
+
+	go taskEngine.AddTask(testTask)
+
+	verifyContainerRunningStateChange(t, taskEngine)
+	verifyTaskIsRunning(stateChangeEvents, testTask)
+
+	time.Sleep(healthCheckWaitTime)
```
richardpen: Can you change this to a periodically check with a timeout, blindly wait here isn't efficient in integration test, as if there is some delay it will fail?

sharanyad: Done

PR #1267, opened 2018-02-28
Removing driverOptions from internal volume struct

yunhee-l:
go-dockerclient's volume doesn't return this field (see https://tinyurl.com/y7e4nllt)

### Summary
go-dockerclient's volume struct never returns the driver options that was passed when volume is created.

### Implementation details
Removed driver options map from internal volume struct, as we won't have access to this field except when we are creating the volume.

### Testing

### Licensing

### DIFF
--- a/agent/taskresource/volume.go
+++ b/agent/taskresource/volume.go
@@ -24,7 +24,6 @@ type VolumeResource struct {
 	Name                string
 	Mountpoint          string
 	Driver              string
-	DriverOptions       map[string]string
 	Labels              map[string]string
 	createdAtUnsafe     time.Time
 	desiredStatusUnsafe VolumeStatus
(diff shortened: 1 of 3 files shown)


agent/taskresource/volume.go
```diff
@@ -24,7 +24,6 @@ type VolumeResource struct {
 	Name                string
 	Mountpoint          string
 	Driver              string
-	DriverOptions       map[string]string
```
sharanyad: My understanding was we use `VolumeResource` obj to call createVolume (equivalent to CreateResource) with its attributes. In that case, `DriverOptions` would still remain here?

yunhee-l: There's a separate createVolumeOptions obj: https://tinyurl.com/ya4uoo4n
This VolumeResource struct is an internal representation of volume. We may need this field back later, but we might need a PR against go-dockerclient to surface this info.

PR #1270, opened 2018-02-28
cgroup resource model

sharanyad:
### Summary
Implementation of cgroup as a task resource

### Implementation details
Modeling cgroup as a resource by mostly reusing the cgroup implementation available today in `agent/resources`

### Testing

### Licensing


agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
...
+		return errors.Wrapf(err, "resource: cleanup cgroup: unable to remove cgroup at %s", c.CgroupRoot)
+	}
+	return nil
+}
+
+// cgroupResourceJSON duplicates CgroupResource fields, only for marshalling and unmarshalling purposes
+type cgroupResourceJSON struct {
+	CgroupRoot      string `json:"CgroupRoot"`
+	CgroupMountPath string `json:"CgroupMountPath"`
+	CreatedAt       time.Time
```
aaithal: i think this needs to be a pointer with omitempty tag. Then, you can set it only if it's been created. 

Also, why do we need this timestamp at all?

agent/taskresource/cgroupstatus.go
```diff
@@ -0,0 +1,92 @@
...
+	// CgroupCreated represents a task resource which has been created
+	CgroupCreated
+	// CgroupCleaned represents a task resource which has been cleaned up
+	CgroupCleaned
+)
+
+var cgroupStatusMap = map[string]CgroupStatus{
+	"NONE":    CgroupStatusNone,
+	"CREATED": CgroupCreated,
+	"CLEANED": CgroupCleaned,
```
aaithal: "REMOVED" might be a better name. `CgroupCleaned` -> `CgroupRemoved` as well.

agent/taskresource/cgroupstatus.go
```diff
@@ -0,0 +1,92 @@
...
+	for k, v := range cgroupStatusMap {
+		if v == cs {
+			return k
+		}
+	}
+	return "NONE"
+}
+
+// TaskStatus maps the resource status to the corresponding task status.
+func (cs CgroupStatus) TaskStatus() api.TaskStatus {
```
aaithal: I'm not sure if we need this method at all. Is there a similar method in `VolumeResource` too? We have this for `ContainerStatus` because task's status is somewhat tied to container's status. But, that shouldn't be the case here.

cc @yunhee-l

agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
...
+
+var (
+	enableMemoryHierarchy = []byte(strconv.Itoa(1))
+)
+
+// CgroupResource represents Cgroup resource
+type CgroupResource struct {
+	control             cgroup.Control
+	CgroupRoot          string
+	CgroupMountPath     string
```
aaithal: Does this need to be part of `CgroupResource` object? This should be just read from the config object, no?

sharanyad: Yes, this should be read from the config object and passed while creating the `CgroupResource`object. In that case, we can directly use the `CgroupResource`'s `CgroupMountPath ` while `Create()`. Else, we might have to pass the config object to the `Create()` resource method, which might be unnecessary for other resources. 
The idea is to make the resource object mostly self-sufficient to create cgroups itself. Thoughts?

agent/taskresource/cgroupstatus.go
sharanyad: We have this for `VolumeResource`. Thinking about this, it should probably be `ContainerStatus`, since container status can be tied to the resource states. What do you think?

yunhee-l: I was quite a bit confused initially on whether the resource status would map to task status or not. I think it makes sense that containers depend on resource status, but I'm not sure map like this for containerStatus is necessary or not. We can remove this for now and add one for containerStatus if needed.

agent/taskresource/cgroup.go
sharanyad: This could be used for debugging purposes. Also `TaskResource` interface has setters and getters for this field, hence `CgroupResource` has these as well.

agent/taskresource/cgroup.go
petderek: I prefer this as written  -- this module is decoupled from the config package.  Its easier to test and we don't hold a pointer to global state.

agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
...
+	memoryUseHierarchy      = "memory.use_hierarchy"
+	rootReadOnlyPermissions = os.FileMode(400)
+)
+
+var (
+	enableMemoryHierarchy = []byte(strconv.Itoa(1))
+)
+
+// CgroupResource represents Cgroup resource
+type CgroupResource struct {
```
petderek: What's the intended difference between taskresource/cgroupresource and resources/cgroup? why are we using another layer of abstraction? Can we combine?

sharanyad: Sure. I left the package `resources/cgroup` as is, since the current flow of cgroup creation/deletion in engine is dependent on it. Also in `taskresource/cgroup.go`, I am just reusing the methods and mocks under `resources/cgroup`. Once the cgroup create/cleanup methods in `taskresource/cgroup.go` are hooked into engine, the plan is to move everything under `resources/cgroup` to `taskresource` package.

agent/taskresource/interface.go
```diff
@@ -35,6 +37,10 @@ type TaskResource interface {
 	SetCreatedAt(time.Time)
 	// GetCreatedAt sets the timestamp for resource's creation time
 	GetCreatedAt() time.Time
+	// Create performs resource creation
+	Create(task *api.Task)
```
richardpen: No return value here?

agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
+// +build linux
```
richardpen: Would it be better to move this into a subdirectory, as there will be lots of files in the `taskresource` directory if more resource was added, like:
```
agent/taskresource/cgroup/cgroup.go
agent/taskresource/volume/volume.go
agent/taskresource/interface.go
```

agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
...
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.createdAt
+}
+
+func (c *CgroupResource) Create(task *api.Task) error {
+	err := c.cgroupInit()
+	if err != nil {
+		seelog.Criticalf("unable to setup '/ecs' cgroup: %v", err)
```
richardpen: The log message should be capitalized, same below.

agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
...
+	err = c.setupTaskCgroup(task)
+	if err != nil {
+		seelog.Criticalf("unable to setup platform resources: %v", err)
+		return err
+	}
+	return nil
+}
+
+func (c *CgroupResource) cgroupInit() error {
+	if c.control.Exists(config.DefaultTaskCgroupPrefix) {
```
richardpen: The cgroup prefix route will be created at the agent level, right?  Which means it's only created once when the agent restart, do we still need this?

agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
...
+		Root:  cgroupRoot,
+		Specs: &linuxResourceSpec,
+	}
+
+	_, err = c.control.Create(&cgroupSpec)
+	if err != nil {
+		return errors.Wrapf(err, "resource: setup cgroup: unable to create cgroup at %s for task: %s", cgroupRoot, task.Arn)
+	}
+
+	// echo 1 > memory.use_hierarchy
```
richardpen: Can you change this comment to be more descriptive, `enable cgroup memory hierarchy`?

agent/taskresource/cgroup.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	return nil
+}
+
+// cgroupResourceJSON duplicates CgroupResource fields, only for marshalling and unmarshalling purposes
+type cgroupResourceJSON struct {
+	CgroupRoot      string `json:"CgroupRoot"`
+	CgroupMountPath string `json:"CgroupMountPath"`
+	CreatedAt       time.Time
+	DesiredStatus   *CgroupStatus `json:"DesiredStatus"`
```
richardpen: Why does this need to be a pointer if it's only used in the Marshal and Unmarshal? So you can use `c.GetDesiredStatus()` below without additional work to get the reference.

agent/taskresource/cgroup.go
sharanyad: Sure

agent/taskresource/cgroup.go
sharanyad: Sure

agent/taskresource/interface.go
sharanyad: Thanks, missed it. shall add here and in `Cleanup`.

agent/taskresource/cgroup.go
sharanyad: As per our offline discussion, cgroup controller can be created on agent level and passed to `CgroupResource` object, and that can be used to create the cgroup prefix in agent as well.

agent/taskresource/cgroup.go
sharanyad: I shall go ahead and add cgroup into a folder and leave volume as is, since it requires more refactoring, which can be addressed in a separate PR.

agent/taskresource/cgroup.go
yunhee-l: It needs to be a pointer because otherwise it'll try and unmarshall int. I think we discussed about this offline.

agent/taskresource/cgroup.go
petderek: Would it be reasonable to include both? I put that comment in originally because it is the command you run on host to get the same effect.

sharanyad:
addressed comments

agent/taskresource/cgroup/cgroupstatus.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// UnmarshalJSON overrides the logic for parsing the JSON-encoded ResourceStatus data
+func (cs *CgroupStatus) UnmarshalJSON(b []byte) error {
+	if strings.ToLower(string(b)) == "null" {
+		*cs = CgroupStatusNone
+		return nil
+	}
+
+	if b[0] != '"' || b[len(b)-1] != '"' {
```
adnxn: nit: should we put this bit of logic somewhere more generic? maybe factored out behind a `getStrStatus()` function? other resources will probably have to implement this as well.

agent/taskresource/cgroup/cgroup.go
```diff
@@ -0,0 +1,212 @@
...
+	defer c.lock.RUnlock()
+
+	return c.createdAt
+}
+
+// Create creates cgroup root for the task
+func (c *CgroupResource) Create(task *api.Task) error {
+	err := c.setupTaskCgroup(task)
+	if err != nil {
+		seelog.Criticalf("Unable to setup platform resources: %v", err)
```
aaithal: It'd be even better if we log task arn here and everywhere. Also, you can use this unofficial log format being followed in the engine package: `"Cgroup resource [%s] Unable to ..: %v", task.Arn, err"`. Can you please modify all log lines here to conform to that?

Example: https://github.com/aws/amazon-ecs-agent/blob/cc9fcff8e6977895cfafc49e04bdb4df3945b55a/agent/engine/docker_task_engine.go#L328

agent/taskresource/cgroup/cgroup.go
```diff
@@ -0,0 +1,212 @@
...
+	CgroupRoot      string        `json:"CgroupRoot"`
+	CgroupMountPath string        `json:"CgroupMountPath"`
+	CreatedAt       time.Time     `json:",omitempty"`
+	DesiredStatus   *CgroupStatus `json:"DesiredStatus"`
+	KnownStatus     *CgroupStatus `json:"KnownStatus"`
+}
+
+// MarshalJSON marshals CgroupResource object using duplicate struct CgroupResourceJSON
+func (c *CgroupResource) MarshalJSON() ([]byte, error) {
+	return json.Marshal(cgroupResourceJSON{
```
aaithal: You need to do a `nil` check here. Example: https://github.com/aws/amazon-ecs-agent/blob/f4d6e452cf23a15cdc94cc3ee9f1ee5fb27679e5/agent/taskresource/volume.go#L114

agent/taskresource/cgroup/cgroup.go
```diff
@@ -0,0 +1,212 @@
...
+		func() *CgroupStatus { desiredState := c.GetDesiredStatus(); return &desiredState }(),
+		func() *CgroupStatus { knownState := c.GetKnownStatus(); return &knownState }(),
+	})
+}
+
+// UnmarshalJSON unmarshals CgroupResource object using duplicate struct CgroupResourceJSON
+func (c *CgroupResource) UnmarshalJSON(b []byte) error {
+	temp := &cgroupResourceJSON{}
+
+	if err := json.Unmarshal(b, &temp); err != nil {
```
aaithal: Doesn't this mean that you're passing a pointer to a pointer? `json.Unmarshal(b, temp)` should be good enough i think

agent/taskresource/cgroup/cgroup.go
sharanyad: Sure. I don't understand why go lets you call a method on a nil object.

aaithal: https://groups.google.com/forum/#!topic/golang-nuts/wcrZ3P1zeAk

agent/taskresource/cgroup/cgroupstatus.go
sharanyad: I think that's a good idea. Can be addressed in a future PR.

agent/taskresource/cgroup/cgroup.go
sharanyad: Looks like go is really confused between being an non-object oriented and object oriented language. 😄

PR #1279, opened 2018-03-06
Refactor Container Dependency model

sharanyad:
### Summary
Deprecate `TransitionDependencySet` type in `api.Container` used to specify container dependencies for state transitions. The new map type is called `TransitionDependencies`.

### Implementation details
This PR focuses on adding the new `TransitionDependencies` type which is a map of the dependent container status to the existing `TransitionDependencySet` type, which is a list of `ContainerDependency` with the status of the container that satisfies the dependency (excluding the `DependentStatus` field).
Note: This PR does not yet construct the `TransitionDependencies` object from `TransitionDependencySet` type, if present in old state files. This will be handled in a separate PR. 

### Testing

### Licensing

### DIFF
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -538,3 +538,23 @@ func (c *Container) GetHealthStatus() HealthStatus {
 
 	return copyHealth
 }
+
+// BuildContainerDependency adds a new dependency container and satisfied status
+// to the dependent container
+func (c *Container) BuildContainerDependency(contName string,
+	satisfiedStatus ContainerStatus,
+	dependentStatus ContainerStatus) {
+	if c.TransitionDependenciesMap == nil {
+		c.TransitionDependenciesMap = make(map[ContainerStatus]TransitionDependencySet)
+	}
+	contDep := ContainerDependency{
+		ContainerName:   contName,
+		SatisfiedStatus: satisfiedStatus,
+	}
+	if _, ok := c.TransitionDependenciesMap[dependentStatus]; !ok {
+		c.TransitionDependenciesMap[dependentStatus] = TransitionDependencySet{}
+	}
+	deps := c.TransitionDependenciesMap[dependentStatus]
+	deps.ContainerDependencies = append(deps.ContainerDependencies, contDep)
+	c.TransitionDependenciesMap[dependentStatus] = deps
+}
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -128,11 +128,17 @@ type Container struct {
 	// is handled properly so that the state storage continues to work.
 	KnownStatusUnsafe ContainerStatus `json:"KnownStatus"`
 
+	// TransitionDependenciesMap is a map of the dependent container status to other
+	// dependencies that must be satisfied in order for this container to transition.
+	TransitionDependenciesMap map[ContainerStatus]TransitionDependencySet
+
 	// TransitionDependencySet is a set of dependencies that must be satisfied
 	// in order for this container to transition.  Each transition dependency
 	// specifies a resource upon which the transition is dependent, a status
 	// that depends on the resource, and the state of the dependency that
 	// satisfies.
+	// Deprecated: Use TransitionDependenciesMap instead. TransitionDependencySet is
+	// retained for compatibility with old state files.
 	TransitionDependencySet TransitionDependencySet `json:"TransitionDependencySet"`
 
 	// SteadyStateDependencies is a list of containers that must be in "steady state" before
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -171,14 +171,19 @@ func (task *Task) initializeEmptyVolumes() {
 				continue
 			}
 			if _, ok := vol.(*EmptyHostVolume); ok {
-				if container.TransitionDependencySet.ContainerDependencies == nil {
-					container.TransitionDependencySet.ContainerDependencies = make([]ContainerDependency, 0)
+				if container.TransitionDependenciesMap == nil {
+					container.TransitionDependenciesMap = make(map[ContainerStatus]TransitionDependencySet)
 				}
-				container.TransitionDependencySet.ContainerDependencies = append(container.TransitionDependencySet.ContainerDependencies, ContainerDependency{
+				contDep := ContainerDependency{
 					ContainerName:   emptyHostVolumeName,
 					SatisfiedStatus: ContainerRunning,
-					DependentStatus: ContainerCreated,
-				})
+				}
+				if _, ok := container.TransitionDependenciesMap[ContainerCreated]; !ok {
+					container.TransitionDependenciesMap[ContainerCreated] = TransitionDependencySet{}
+				}
+				deps := container.TransitionDependenciesMap[ContainerCreated]
+				deps.ContainerDependencies = append(deps.ContainerDependencies, contDep)
+				container.TransitionDependenciesMap[ContainerCreated] = deps
 				requiredEmptyVolumes = append(requiredEmptyVolumes, mountPoint.SourceVolume)
 			}
 		}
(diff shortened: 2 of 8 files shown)


agent/engine/dependencygraph/graph_test.go
```diff
@@ -445,124 +445,148 @@ func assertResolved(f func(target *api.Container, dep *api.Container) bool, targ
...
 			TargetKnown:      api.ContainerRunning,
 			TargetDesired:    api.ContainerRunning,
+			TargetNext:       api.ContainerRunning,
+			DependencyName:   "container",
 			DependencyKnown:  api.ContainerRunning,
-			DependentStatus:  api.ContainerRunning,
 			SatisfiedStatus:  api.ContainerCreated,
 			ExpectedResolved: true,
 		},
-		// Note: Not all possible situations are tested here.  The only situations tested here are ones that are
```
adnxn: has this changed? are we testing "all possible situations" now?

agent/engine/dependencygraph/graph.go
```diff
@@ -183,38 +183,30 @@ func verifyTransitionDependenciesResolved(target *api.Container, existingContain
 		return true
 	}
 
-	for _, containerDependency := range target.TransitionDependencySet.ContainerDependencies {
-		maybeResolves, exists := existingContainers[containerDependency.ContainerName]
+	if targetGoal < target.GetKnownStatus()+1 {
```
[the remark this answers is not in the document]
sharanyad: This condition would never be reached (since the check is done in `progressContainers`). I shall remove this.

agent/api/container.go
```diff
@@ -128,11 +128,19 @@ type Container struct {
 	// is handled properly so that the state storage continues to work.
 	KnownStatusUnsafe ContainerStatus `json:"KnownStatus"`
 
+	// ContainerDependenciesMap is a map of the dependent container status to other container
+	// dependencies that must be satisfied in order for this container to transition.
+	// Each container dependency specifies a container upon which the transition is
+	// dependent and its state that satisfies the dependency.
+	ContainerDependenciesMap map[ContainerStatus][]ContainerDependency
```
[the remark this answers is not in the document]
sharanyad: Done

agent/api/task.go
```diff
@@ -171,14 +171,21 @@ func (task *Task) initializeEmptyVolumes() {
...
-					container.TransitionDependencySet.ContainerDependencies = make([]ContainerDependency, 0)
-				}
-				container.TransitionDependencySet.ContainerDependencies = append(container.TransitionDependencySet.ContainerDependencies, ContainerDependency{
+				contDep := ContainerDependency{
 					ContainerName:   emptyHostVolumeName,
 					SatisfiedStatus: ContainerRunning,
-					DependentStatus: ContainerCreated,
-				})
+				}
+				if container.TransitionDependenciesMap == nil {
```
aaithal: why not do this outside of the for loop or when you construct the `Container` object? Let's move this outside of the `for` loop

agent/api/task.go
```diff
@@ -171,14 +171,21 @@ func (task *Task) initializeEmptyVolumes() {
...
+					container.TransitionDependenciesMap = make(map[ContainerStatus]TransitionDependencySet)
+				}
+				if deps, ok := container.TransitionDependenciesMap[ContainerCreated]; ok {
```
aaithal: Can you refactor this as:

```
if _, ok := container.TransitionDependenciesMap[ContainerCreated]; !ok {
 container.TransitionDependenciesMap[ContainerCreated] = TransitionDependencySet{}
}
deps = container.TransitionDependenciesMap[ContainerCreated]
deps.ContainerDependencies = append(deps.ContainerDependencies, contDep)
container.TransitionDependenciesMap[ContainerCreated] = deps
```

agent/api/task.go
```diff
@@ -286,14 +293,21 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
...
-			container.TransitionDependencySet.ContainerDependencies = make([]ContainerDependency, 0)
-		}
-		container.TransitionDependencySet.ContainerDependencies = append(container.TransitionDependencySet.ContainerDependencies, ContainerDependency{
+		contDep := ContainerDependency{
 			ContainerName:   PauseContainerName,
 			SatisfiedStatus: ContainerResourcesProvisioned,
-			DependentStatus: ContainerPulled,
-		})
+		}
+		if container.TransitionDependenciesMap == nil {
```
aaithal: same comments as in the `initializeEmptyVolumes` method.

agent/api/transitiondependency.go
```diff
@@ -17,5 +17,5 @@ type ContainerDependency struct {
 	SatisfiedStatus ContainerStatus `json:"SatisfiedStatus"`
 	// DependentStatus defines the status that cannot be reached until the
 	// resource satisfies the dependency
-	DependentStatus ContainerStatus `json:"DependentStatus"`
+	DependentStatus ContainerStatus `json:"DependentStatus,omitempty"`
```
aaithal: Is adding `omitempty` absolutely necessary here? Are there conditions in which it could break the JSON marshaling/unmarshaling logic? I'd rather not add this if it's only giving us marginal gains.

agent/engine/dependencygraph/graph.go
```diff
@@ -183,38 +183,25 @@ func verifyTransitionDependenciesResolved(target *api.Container, existingContain
...
 	}
 
-	for _, containerDependency := range target.TransitionDependencySet.ContainerDependencies {
-		maybeResolves, exists := existingContainers[containerDependency.ContainerName]
+	// TODO: add verifyResourceDependenciesResolved check here
+	return verifyContainerDependenciesResolved(target, existingContainers)
+}
+
+func verifyContainerDependenciesResolved(target *api.Container, existingContainers map[string]*api.Container) bool {
+	targetNext := target.GetKnownStatus() + 1
```
aaithal: A potential improvement here (not required at all in this PR, something to think about) is to instead get the next status from the "Container" object or some transition map. This will mean that containers/resources/tasks can potentially skip transitions if required as they wouldn't rely on the "+1" logic. Might be worth adding a TODO for.

agent/api/container.go
```diff
@@ -128,11 +128,17 @@ type Container struct {
...
+	// TransitionDependenciesMap is a map of the dependent container status to other
+	// dependencies that must be satisfied in order for this container to transition.
+	TransitionDependenciesMap map[ContainerStatus]TransitionDependencySet
+
 	// TransitionDependencySet is a set of dependencies that must be satisfied
 	// in order for this container to transition.  Each transition dependency
 	// specifies a resource upon which the transition is dependent, a status
 	// that depends on the resource, and the state of the dependency that
 	// satisfies.
+	// Deprecated: Use TransitionDependenciesMap instead. TransitionDependencySet is
```
aaithal: A better way to do this (in your future PR) would be to:

1. get rid of  `TransitionDependencySet` altogether
2. tag `TransitionDependenciesMap` as `"TransitionDependencySet"` in JSON
3. update the JSON marshaler to deal with `"TransitionDependencySet"` tag. If it's a map, no big deal. Else, construct the `TransitionDependenciesMap` map.

Doing that would mean that we don't need to have duplicate code, names for this.

agent/api/transitiondependency.go
sharanyad: We won't be using this field going forward, since `DependentStatus` will actually be the key in `TransitionDependenciesMap`. Hence added a `omitEmpty` since we are reusing `ContainerDependency` and this field would be empty. But we can just remove this field in here and take care of it in JSON Marshaler in a future PR?

aaithal: yeah, that'd be better.

agent/engine/dependencygraph/graph.go
sharanyad: I think you're referring to:
https://github.com/aws/amazon-ecs-agent/blob/edc3e26016e363d7e8e6c044b11bde9bbb8853a8/agent/api/container.go#L367
Shall add in this PR.

agent/api/task.go
sharanyad: The container dependency is specified only in cases of empty Volumes, and this can be detected only inside the for loop. Hence we are constructing the map when required.

agent/api/task.go
```diff
@@ -171,14 +171,19 @@ func (task *Task) initializeEmptyVolumes() {
...
 					ContainerName:   emptyHostVolumeName,
 					SatisfiedStatus: ContainerRunning,
-					DependentStatus: ContainerCreated,
-				})
+				}
+				if _, ok := container.TransitionDependenciesMap[ContainerCreated]; !ok {
+					container.TransitionDependenciesMap[ContainerCreated] = TransitionDependencySet{}
+				}
+				deps := container.TransitionDependenciesMap[ContainerCreated]
+				deps.ContainerDependencies = append(deps.ContainerDependencies, contDep)
```
richardpen: I think you need a nil check for `deps.ContainerDependencies`?

agent/api/task.go
```diff
@@ -286,14 +291,19 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
...
+		contDep := ContainerDependency{
 			ContainerName:   PauseContainerName,
 			SatisfiedStatus: ContainerResourcesProvisioned,
-			DependentStatus: ContainerPulled,
-		})
+		}
+		if container.TransitionDependenciesMap == nil {
+			container.TransitionDependenciesMap = make(map[ContainerStatus]TransitionDependencySet)
+		}
+		if _, ok := container.TransitionDependenciesMap[ContainerPulled]; !ok {
```
richardpen: It would be good if you can wrap this as a method of container struct. So you can use it in both the `empty volume` and the `pause container`.

agent/api/task.go
richardpen: I think you need a nil check for `deps.ContainerDependencies`?

sharanyad: Looks like we might not? 
https://play.golang.org/p/1v0JJSknXQX

agent/api/container.go
```diff
@@ -532,3 +538,21 @@ func (c *Container) GetHealthStatus() HealthStatus {
 
 	return copyHealth
 }
+
+func (c *Container) BuildContainerDependency(contName string,
```
richardpen: Lint: please add a comment for this method.

aaithal:
@sharanyad  are you planning on addressing my comments in a subsequent PR?

sharanyad:
@aaithal Yes, I will have a separate PR out with those changes

PR #1286, opened 2018-03-08
Add integration test for engine restart

richardpen:
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/common_test.go
+++ b/agent/engine/common_test.go
@@ -33,7 +33,8 @@ import (
 )
 
 const (
-	containerID = "containerID"
+	containerID                 = "containerID"
+	waitTaskStateChangeDuration = 2 * time.Minute
 )
 
 var (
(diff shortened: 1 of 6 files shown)


agent/engine/task_manager.go
```diff
@@ -277,6 +289,21 @@ func (mtask *managedTask) waitSteady() {
 	}
 }
 
+// getWaitSteadyCheckCancelFunc returns the cancel func for the steadyStateCheck wait
+func (mtask *managedTask) getWaitSteadyCheckCancelFunc() context.CancelFunc {
```
aaithal: rather than surfacing the cancel func like this to an entity that's external to the task, i'd rather have a listening channel and trigger a cancel based on an event being received from that channel

richardpen: yes, that would be a better approach. But this part of code is only useful in the test, having a listening channel will add some complexity into the code(we will need a delicated goroutine listener or reuse the waitEvent) and I don't see a use case where we need this beyond the test. What's your opinion?

aaithal: The current approach feels hacky even for testing IMO. Another alternative is to implement a new `cancelSteadyStateWait()` method and wrap all of this logic within that. Also, for all new members (fields and methods) that you added specifically for testing, please make sure that you add documentation stating the same

agent/engine/task_manager.go
```diff
@@ -277,6 +289,30 @@ func (mtask *managedTask) waitSteady() {
 	}
 }
 
+// cancelSteadyStateWait checks if the task is waiting for steady state and stop
+// the waiting if it it.
+// This is currently only used in test
+func (mtask *managedTask) cancelSteadyStateWait() bool {
```
aaithal: I think returning an `error` here is more appropriate than returning a `bool`

agent/engine/common_test.go
```diff
@@ -244,3 +245,32 @@ func waitForStopEvents(t *testing.T, stateChangeEvents <-chan statechange.Event,
...
+			// Wait for the waitSteady to be invoked
+			time.Sleep(10 * time.Millisecond)
+			continue
+		}
+		i++
+	}
+}
+
+func waitForContainerHealthStatus(t *testing.T, testTask *api.Task) {
+	ctx, _ := context.WithTimeout(context.TODO(), waitTaskStateChangeDuration)
```
aaithal: can you add `defer cancel()` here?

PR #1287, opened 2018-03-08
Update build tools for 8 char short sha

jahkeup:
### Summary

Use 8 character sha in publishing common for scripts.

### Implementation details
I changed `--short` to `--short=8`.

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,7 +212,7 @@ get-cni-sources:
 cni-plugins: get-cni-sources .out-stamp
 	@docker build -f scripts/dockerfiles/Dockerfile.buildCNIPlugins -t "amazon/amazon-ecs-build-cniplugins:make" .
 	docker run --rm --net=none \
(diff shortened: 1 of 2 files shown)


jhaynes:
@aaithal Should we pin https://github.com/aws/amazon-ecs-agent/blob/master/Makefile#L215 to 8 chars also?

aaithal:
can't think of a reason not to do that. would also make sense to apply the same change here as well: https://github.com/aws/amazon-ecs-cni-plugins/blob/0c6216c60401232805e50e31d4040ae84b0b23cf/Makefile#L14

jahkeup:
As long as there are no assumptions for the CNI plugins in this or its own package I'd be happy to have this consistently applied everywhere.

jahkeup:
PTAL, there's one more `--short` that was updated.

Issue #1245, opened 2018-02-19
testing: flakey, TestGetMACAddressNotFoundRetry


adnxn:
### Summary

make targets `test` or `test-in-docker` will sometimes fail with:

```
=== RUN   TestGetMACAddressNotFoundRetry
2018-02-14T21:26:17Z [WARN] Unable to retrieve mac address for device 'eth1': Dummy Netlink Error
--- FAIL: TestGetMACAddressNotFoundRetry (0.01s)
	Error Trace:	utils_linux_test.go:101
	Error:		Received unexpected error "eni mac address: timed out waiting for eni device 'eth1': context deadline exceeded"
		
	Error Trace:	utils_linux_test.go:102
	Error:		Not equal: "" (expected)
			        != "00:0a:95:9d:68:16" (actual)
		
	<autogenerated>:1: missing call(s) to *mock_netlinkwrapper.MockNetLink.LinkByName(is equal to eth1)
	<autogenerated>:1: aborting test due to missing call(s)
```

adnxn:
closing issue, fix merged to dev.

PR #1289, opened 2018-03-09
eni: increase test retry duration, swap assert() params

adnxn:
### Summary
* increase retry duration for test, see https://github.com/aws/amazon-ecs-agent/issues/1245
* swapped params on calls to assert.* as per [expected ordering](https://godoc.org/github.com/stretchr/testify/assert#Equal)
* ran `go test -race . -run=TestGetMACAddressNotFoundRetry -count=1000`
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/eni/networkutils/utils_linux_test.go
+++ b/agent/eni/networkutils/utils_linux_test.go
@@ -54,7 +54,7 @@ func TestGetMACAddress(t *testing.T) {
 	ctx := context.TODO()
 	mac, err := GetMACAddress(ctx, time.Millisecond, randomDevice, mockNetlink)
 	assert.Nil(t, err)
-	assert.Equal(t, mac, validMAC)
+	assert.Equal(t, validMAC, mac)
 }
 
 // TestGetMACAddressWithNetlinkError attempts to test the netlinkClient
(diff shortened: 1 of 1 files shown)


richardpen:
Code looks good to me, did you run the test thousands of times?

adnxn:
@richardpen yea ran with `-count=1000`

agent/eni/networkutils/utils_linux_test.go
```diff
@@ -95,11 +95,11 @@ func TestGetMACAddressNotFoundRetry(t *testing.T) {
 		}, nil),
 	)
 	ctx := context.TODO()
-	// Set max retry duration to twice that of the min backoff to ensure that there's
+	// Set max retry duration to twice that of the max backoff to ensure that there's
 	// enough time to retry
-	mac, err := GetMACAddress(ctx, 2*macAddressBackoffMin, randomDevice, mockNetlink)
+	mac, err := GetMACAddress(ctx, 2*macAddressBackoffMax, randomDevice, mockNetlink)
```
aaithal: can it be just `macAddressBackoffMax` instead of `2* macAddressBackoffMax`?

richardpen: Probably not, the test is expecting two calls(retries), so in the worst case if the first retry has a backoff `macAddressBackoffMax`, then the second call wouldn't happen due to the timeout.

adnxn: correct me if i'm wrong - but we were thinking that since there are two calls, and the [backoff](https://github.com/aws/amazon-ecs-agent/blob/master/agent/utils/backoff.go#L42) could actually be `max + max*jitterMultiple`, which would put us past the `macAddressBackoffMax` upper bound for the retry call to be invoked.

Issue #1249, opened 2018-02-21
image pull: measure and introduce InactivityTimeout parameter


adnxn:
Today image pulls could potentially hang due to underlying issues with docker and the agent will only fall back on the [pullImageTimeout](https://github.com/aws/amazon-ecs-agent/blob/761937f7fb98a06bea00c4f55f9c594560e73725/agent/engine/docker_client.go#L66). We should evaluate reasonable duration for an [InactivityTimeout](https://github.com/fsouza/go-dockerclient/blob/ca33ff277b527ce11b793e62f9ba244129b01caf/image.go#L301) and add it to the pull image options [here](https://github.com/aws/amazon-ecs-agent/blob/c051f10b219cac94c14cc533c050902bce354219/agent/engine/docker_client.go).

aaithal:
We've released a new [version of the ECS agent](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.17.3), which should fix this issue. Closing this for now.

PR #1290, opened 2018-03-12
engine: add inactivity timeout for image pulling

fenxiong:
### Summary

This pull request is for issue 1249 here: https://github.com/aws/amazon-ecs-agent/issues/1249, which is to introduce InactivityTimeout for image pulling.

### Implementation details

By using the inactivity timeout parameter supported by go-dockerclient.

### Testing

Three manual testings are performed as follow, with timeout set to 5 seconds:
(1) Start pulling a large image, send SIGSTOP to Docker daemon process, see whether inactivity timeout is triggered; send SIGCONT to Docker daemon process, see whether pulling resumes.
Result: after SIGSTOP, timeout is triggered; after SIGCONT, pulling resumes

(2) Start pulling a large image (from ECR), blackhole ECR, see whether timeout is triggered; remove blackhole, see whether pulling resumes
Result: after adding blackhole, the pulling fails immediately due to connection failure, and doesn't reach the timeout limit.

(3) Start pulling a large image, randomly drop 50% of packets on the host during the pull
Result: inactivity timeout is not triggered; pulling succeeds normally

Enhancement - Introduce InactivityTimeout to image pulling.

### Licensing

### DIFF
--- a/agent/engine/docker_client.go
+++ b/agent/engine/docker_client.go
@@ -83,6 +83,10 @@ const (
 	// around a docker bug which sometimes results in pulls not progressing.
 	dockerPullBeginTimeout = 5 * time.Minute
 
+	// dockerPullInactivityTimeout is the amount of time that we will
+	// wait when the pulling does not progress
+	dockerPullInactivityTimeout = 1 * time.Minute
+
 	// pullStatusSuppressDelay controls the time where pull status progress bar
 	// output will be suppressed in debug mode
 	pullStatusSuppressDelay = 2 * time.Second
@@ -311,6 +315,7 @@ func (dg *dockerGoClient) pullImage(image string, authData *api.RegistryAuthenti
 	opts := docker.PullImageOptions{
 		Repository:   repository,
 		OutputStream: pullWriter,
+		InactivityTimeout: dockerPullInactivityTimeout,
 	}
 	timeout := dg.time().After(dockerPullBeginTimeout)
 	// pullBegan is a channel indicating that we have seen at least one line of data on the 'OutputStream' above.
--- a/agent/engine/docker_client_test.go
+++ b/agent/engine/docker_client_test.go
@@ -167,6 +167,19 @@ func TestPullImageGlobalTimeout(t *testing.T) {
 	wait.Done()
 }
 
+func TestPullImageInactivityTimeout(t *testing.T) {
+	mockDocker, client, testTime, _, _, done := dockerClientSetup(t)
+	defer done()
+
+	testTime.EXPECT().After(gomock.Any()).AnyTimes()
+	mockDocker.EXPECT().PullImage(&pullImageOptsMatcher{"image:latest"}, gomock.Any()).Return(
+		docker.ErrInactivityTimeout).Times(maximumPullRetries) // expected number of retries
+
+	metadata := client.PullImage("image", nil)
+	assert.Error(t, metadata.Error, "Expected error for pull inactivity timeout")
+	assert.Equal(t, "CannotPullContainerError", metadata.Error.(api.NamedError).ErrorName(), "Wrong error type")
+}
+
 func TestPullImage(t *testing.T) {
 	mockDocker, client, testTime, _, _, done := dockerClientSetup(t)
 	defer done()


agent/engine/docker_client.go
```diff
@@ -83,6 +83,10 @@ const (
 	// around a docker bug which sometimes results in pulls not progressing.
 	dockerPullBeginTimeout = 5 * time.Minute
 
+	// dockerPullInactivityTimeout is the amount of time that we will
+	// wait when the pulling does not progress
+	dockerPullInactivityTimeout = 5 * time.Minute
```
yunhee-l: how did we decide on 5 mins? is there any data/metrics available for us? Ideally p99 would be a good value to use.

fenxiong: currently it's set arbitrarily. i'm not sure if there's any data/metrics for this

adnxn: hm we havent figured out a good way to measure p99 in this case. the only real visibility we've had so far is from root causing hanging tasks, containers which will get stuck for arbitrary periods of time. we want to introduce as a fail fast for docker inactivity at least during image pulls.

do you know if we could get a better idea of what to set here from ecr side? 😕

yunhee-l: From ECR side, P99 is in seconds not minutes (I can show you the dashboard in person) so 5 mins is more than long enough timeout. Thought 5 mins was a bit too long in general but if it's only meant to be used for hanging tasks I suppose it is not a big issue. However flip side is whether this might adversely impact large image pull that can take longer than 5 mins? Is ECR the only repo we pull from?

adnxn: >However flip side is whether this might adversely impact large image pull that can take longer than 5 mins?

The `dockerPullInactivityTimeout ` shouldn't adversely affect large image pulls. This is explicitly to checking to timeout if docker is hanging during the download and large image pulls should show continuous activity during the download. 

> Is ECR the only repo we pull from?

No, the agent could be pulling from anywhere.

yunhee-l: thanks for the explanation! if timeout is specifically for the case where docker is hanging and inactive, then I don't think any of my concerns are relevant.

fenxiong: just as a check, i tested it manually with a large image (3GB) and small timeout (5 secs), and the pulling worked fine with no timeout.

aaithal:
@fenxiong can you please incorporate the following tests for this as well? 

1. Start pulling a large image, blackhole ECR/Dockerhub (wherever the image is being pulled from), ensure that this timeout works
2. Start pulling a large image, send SIGSTOP to Docker daemon process, ensure that timeout works
3. Same as [1] or [2], but remove the blackhole/send SIGCONT and ensure that pull resumes
4. Randomly drop 50% of packets on the host, ensure that pull inactivity timeout behavior works

fenxiong:
@aaithal I have finished the above tests, with timeout set to 5 seconds. For the Docker daemon test, i'm able to observe the timeout and able to restore it. For the blackhole test, i'm not able to observe the timeout because after i added the ip to blackhole the pulling failed immediately and wouldn't reach timeout. For the random drop case, i saw no timeout happen, and i think this is expected right? if 50% packets are not dropped the pulling won't stop

agent/engine/docker_client_test.go
```diff
@@ -167,6 +167,23 @@ func TestPullImageGlobalTimeout(t *testing.T) {
...
+func TestPullImageInactivityTimeout(t *testing.T) {
+	mockDocker, client, testTime, _, _, done := dockerClientSetup(t)
+	defer done()
+
+	testTime.EXPECT().After(gomock.Any()).AnyTimes()
+	mockDocker.EXPECT().PullImage(&pullImageOptsMatcher{"image:latest"}, gomock.Any()).Return(
+		docker.ErrInactivityTimeout).Times(maximumPullRetries) // expected number of retries
+
+	metadata := client.PullImage("image", nil)
+	if metadata.Error == nil {
```
richardpen: Please use the `assert` instead. Same below.

agent/engine/docker_client.go
aaithal: `5m` seems too long here. I don't think this should be greater than `1m`.

fenxiong: The timeout is set to 1 minute now. I have commented on the SageMaker ticket to see if there's any input at their side.

aaithal:
@fenxiong  can you please make sure to submit a new PR that modifies `CHANGELOG.md` file with a reference to this PR?

Thanks,
Anirudh

PR #1296, opened 2018-03-15
Adding ListPlugins wrapper

yunhee-l:
fixing broken build due to wrong printf format

### Summary
Adding ListPlugins wrapper in docker_client

### Implementation details

Note that go-dockerclient currently does not handle filters (for enabled=bool and capability=*driver, see https://docs.docker.com/engine/reference/commandline/plugin_ls/#filtering) in their implementation of listPlugins. I've added "ListPluginsWithFilters" as a convenience method for now, but am planning on submitting PR to go-dockerclient for this, after which we can clean up our implementation. Also, note that ListPlugins currently returns a list of full plugin objects, which is a pretty big nested structure (see response schema in https://docs.docker.com/engine/api/v1.30/#operation/PluginList). I've simplified our interface to deal with only the names of the plugins for now, but this is also up for discussion.

### Testing

Changelog not updated

### Licensing

### DIFF
--- a/agent/engine/docker_client.go
+++ b/agent/engine/docker_client.go
@@ -54,6 +54,9 @@ const (
 	healthCheckUnhealthy = "unhealthy"
 	// maxHealthCheckOutputLength is the maximum length of healthcheck command output that agent will save
 	maxHealthCheckOutputLength = 1024
+
+	// VolumeDriverType is one of the plugin capabilities see https://docs.docker.com/engine/reference/commandline/plugin_ls/#filtering
+	VolumeDriverType = "volumedriver"
 )
 
 // Timelimits for docker operations enforced above docker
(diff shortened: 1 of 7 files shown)


agent/engine/docker_client.go
```diff
@@ -74,6 +77,8 @@ const (
 	inspectContainerTimeout = 30 * time.Second
 	removeImageTimeout      = 3 * time.Minute
 
+	ListPluginsTimeout = 1 * time.Minute
```
yunhee-l: This is arbitrary value, where did we get other timeout values from? 
One thing to note is that this will be used during agent startup to detect installed plugins as its capability, and I don't know if there's any implication of how long we can take to timeout during the startup.

yunhee-l:
going with a different approach, docker plugin ls doesn't show plugins that weren't installed with docker plugin install.

PR #1304, opened 2018-03-19
Adding ListPlugins wrapper

yunhee-l:
### Summary
Adding wrapper method for go-dockerclient's ListPlugins method, and a convenience method (ListPluginsWithFilters) to apply search filters (see https://docs.docker.com/engine/reference/commandline/plugin_ls/#filtering). go-dockerclient's ListPlugins method currently does not handle filters, but when we or someone else submits PR for the fix we will refactor these methods.

### Implementation details
ListPlugins wrapper method and ListPluginsWithFilters on top of it.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/docker_client.go
+++ b/agent/engine/docker_client.go
@@ -54,6 +54,8 @@ const (
 	healthCheckUnhealthy = "unhealthy"
 	// maxHealthCheckOutputLength is the maximum length of healthcheck command output that agent will save
 	maxHealthCheckOutputLength = 1024
+	// VolumeDriverType is one of the plugin capabilities see https://docs.docker.com/engine/reference/commandline/plugin_ls/#filtering
+	VolumeDriverType = "volumedriver"
 )
 
 // Timelimits for docker operations enforced above docker
@@ -73,6 +75,8 @@ const (
 	removeContainerTimeout  = 5 * time.Minute
 	inspectContainerTimeout = 30 * time.Second
 	removeImageTimeout      = 3 * time.Minute
+	// ListPluginsTimeout is the timout for ListPlugins API.
+	ListPluginsTimeout      = 1 * time.Minute
 
 	// Parameters for caching the docker auth for ECR
 	tokenCacheSize = 100
@@ -157,6 +161,14 @@ type DockerClient interface {
 
 	// RemoveVolume removes a volume by its name. A timeout value should be provided for the request
 	RemoveVolume(string, time.Duration) error
+
+	// ListPluginsWithFilters returns the set of docker plugins installed on the host, filtered by options provided.
+	// A timeout value should be provided for the request.
+	ListPluginsWithFilters(bool, []string, time.Duration) ([]string, error)
+
+	// ListPlugins returns the set of docker plugins installed on the host. A timeout value should be provided for
+	// the request.
+	ListPlugins(time.Duration) listPluginsResponse
 
 	// Stats returns a channel of stat data for the specified container. A context should be provided so the request can
 	// be canceled.
@@ -974,7 +986,7 @@ func (dg *dockerGoClient) Version() (string, error) {
 
 type volumeResponse struct {
 	Volume *taskresource.VolumeResource
-	Error       error
+	Error  error
 }
 
 func (dg *dockerGoClient) CreateVolume(name string,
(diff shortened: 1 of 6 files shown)


agent/engine/docker_client_test.go
```diff
@@ -1435,3 +1435,104 @@ func TestRemoveVolume(t *testing.T) {
...
+	wait.Add(1)
+	mockDocker.EXPECT().ListPlugins(gomock.Any()).Do(func(x interface{}) {
+		warp <- time.Now()
+		wait.Wait()
+		// Don't return, verify timeout happens
+		// TODO remove the MaxTimes by cancel the context passed to CreateVolume
+		// when issue #1212 is resolved
+	}).MaxTimes(1)
+	response := client.ListPlugins(timeout)
+	assert.Error(t, response.Error, "expected error for timeout")
```
petderek: I think you could replace these assertions with [EqualError](https://godoc.org/github.com/stretchr/testify/assert#EqualError).

agent/engine/docker_client.go
```diff
@@ -1134,6 +1145,85 @@ func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
...
+					filteredPluginNames = append(filteredPluginNames, pluginDetail.Name)
+					break
+				}
+			}
+		}
+	}
+	return filteredPluginNames, nil
+}
+
+func (dg *dockerGoClient) ListPlugins(timeout time.Duration) listPluginsResponse {
```
petderek: What do you think about making timeout a property of the docker client struct? That should still make it reasonably simple from a testing perspective.

agent/engine/docker_client.go
```diff
@@ -1134,6 +1145,85 @@ func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
...
+	// Injecting the 'timeout' makes it easier to write tests.
+	// Eventually, the context should be initialized from a parent root context
+	// instead of TODO.
+	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
+	defer cancel()
+
+	// Buffered channel so in the case of timeout it takes one write, never gets
+	// read, and can still be GC'd
+	response := make(chan listPluginsResponse, 1)
+	go func() { response <- dg.listPlugins(ctx) }()
```
petderek: Why does this need to happen asynchronously? As far as I can tell, we immediately wait on the goroutine we start.

agent/engine/docker_client_test.go
```diff
@@ -1435,3 +1435,104 @@ func TestRemoveVolume(t *testing.T) {
...
+		Name: pluginName,
+		Tag: pluginTag,
+		Active: true,
+	}
+
+	mockDocker.EXPECT().ListPlugins(gomock.Any()).Return([]docker.PluginDetail{pluginDetail}, nil)
+
+	response := client.ListPlugins(inspectContainerTimeout)
+	assert.NoError(t, response.Error)
+	assert.Equal(t, pluginDetail.ID, response.Plugins[0].ID)
```
petderek: assertEqual _should_ do an equality check on the whole struct... so one line should be fine?

agent/engine/docker_client_test.go
```diff
@@ -1435,3 +1435,104 @@ func TestRemoveVolume(t *testing.T) {
...
+	assert.Equal(t, pluginDetail.Name, response.Plugins[0].Name)
+	assert.Equal(t, pluginDetail.Tag, response.Plugins[0].Tag)
+	assert.True(t, response.Plugins[0].Active)
+}
+
+func TestListPluginsWithFilter(t *testing.T) {
```
petderek: Nice!

agent/engine/docker_client.go
```diff
@@ -73,6 +75,7 @@ const (
 	removeContainerTimeout  = 5 * time.Minute
 	inspectContainerTimeout = 30 * time.Second
 	removeImageTimeout      = 3 * time.Minute
+	ListPluginsTimeout      = 1 * time.Minute
```
aaithal: lint: needs a comment. You can move this to be with rest of the global const's as well

agent/engine/docker_client.go
```diff
@@ -158,6 +161,14 @@ type DockerClient interface {
 	// RemoveVolume removes a volume by its name. A timeout value should be provided for the request
 	RemoveVolume(string, time.Duration) error
 
+	// ListPluginsWithFilters returns the set of docker plugins installed on the host, filtered by options provided.
+	// A timeout value should be provided for the request.
+	ListPluginsWithFilters(bool, []string, time.Duration) ([]string, error)
```
aaithal: Just for my own understanding, what are the use-cases for invoking `ListPluginsWithFilters(false, ...)`?

yunhee-l: this is where capability=volumedriver (and enabled=true) filter is supposed to be applied, but go-dockerclient's ListPlugins currently doesn't handle filters. I am going to submit PR for this fix but didn't want to block our work.

aaithal: ah, makes sense. so, more for the sake of completion than anything else.

agent/engine/docker_client.go
```diff
@@ -1134,6 +1145,85 @@ func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
...
 }
 
+type listPluginsResponse struct {
+	Plugins []docker.PluginDetail
+	Error   error
+}
+
+// ListPluginsWithFilters currently is a convenience method as go-dockerclient doesn't implement fitered list. When we or someone else submits
+// PR for the fix we will refactor this to pass in the fiters.
+func (dg *dockerGoClient) ListPluginsWithFilters(enabledFilter bool, capabilityFilter []string, timeout time.Duration) ([]string, error) {
```
aaithal: nit: `enabled` and `capabilities` might be enough here for param names.

agent/engine/docker_client.go
```diff
@@ -1134,6 +1145,85 @@ func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
 	return nil
 }
 
+type listPluginsResponse struct {
+	Plugins []docker.PluginDetail
+	Error   error
+}
+
+// ListPluginsWithFilters currently is a convenience method as go-dockerclient doesn't implement fitered list. When we or someone else submits
```
sharanyad: Comments explaining what the arguments do here might be helpful for readability.

agent/engine/docker_client.go
```diff
@@ -1134,6 +1145,85 @@ func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
...
+
+	var filteredPluginNames []string
+	response := dg.ListPlugins(timeout)
+
+	if response.Error != nil {
+		return nil, response.Error
+	}
+
+	for _, pluginDetail := range response.Plugins {
+		if pluginDetail.Active != enabledFilter {
```
sharanyad: What does `pluginDetail.Active` denote here? Don't we always expect the plugin to be active?

agent/engine/docker_client_test.go
yunhee-l: this checks for the error message, and for timeout errors it's "Could not transition to ${err.transition}; timed out after waiting ${err.duration}". see https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/errors.go#L59
I'll keep this as is, it's confusing to mention transition here (which we should fix, not all timeouts relate to transitioning container).

agent/engine/docker_client.go
yunhee-l: I initially did, but looking at our DockerClient interface, all timeouts for APIs are passed in as a param. I don't know the design decision behind this, so I just went with being consistent. Given that we are using constants for these timeouts, maybe we could refactor it for all APIs - but let's do that as a separate PR.

agent/engine/docker_client.go
yunhee-l: it was the pattern for all docker API calls (or so I was told). I couldn't find out whether plugin ls was a synchronous or asynchronous call though.

agent/engine/docker_client_test.go
yunhee-l: Fixed.

agent/engine/docker_client.go
yunhee-l: Where do we keep global constants?

agent/engine/docker_client.go
yunhee-l: Fixed

agent/engine/docker_client.go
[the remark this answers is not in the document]
yunhee-l: this is arbitrary, but given the fact that listPlugins basically returns the contents of in-memory map (see https://tinyurl.com/yaep53uw) it shouldn't need to be too long.

agent/engine/docker_client.go
yunhee-l: it is the same as "Enabled". You can install a plugin and then disable it.

agent/engine/docker_client.go
yunhee-l: Fixed

PR #1306, opened 2018-03-19
Make the progresscontainer independent of each other

richardpen:
### Summary

Change the waitForContainerTransitions to only wait for one transition instead of all of the transitions so that each container can still move forward on its own without waiting for other transitions.

### Implementation details

Remove the logic of waiting for all transitions in waitForContainerTransitions, and added a transition status to indicate the container is in the middle of a processing to avoid duplicate action on the same container.

### Testing

- [x] Manual test with task that has multiple containers, containers are progressed independently.
  * Task with multiple containers
  * Task that uses volume (including empty volume and non-empty volume)
  * Task that uses volumesFrom
  * Task that uses links
  * Set a cluster to periodically run a bunch of tasks

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/container.go
+++ b/agent/api/container.go
@@ -153,7 +153,9 @@ type Container struct {
 
 	// AppliedStatus is the status that has been "applied" (e.g., we've called Pull,
 	// Create, Start, or Stop) but we don't yet know that the application was successful.
-	AppliedStatus ContainerStatus
+	// No need to save it in the state file, as agent will synchronize the container status
+	// on restart and for some operation eg: pull, it has to be recalled again.
+	AppliedStatus ContainerStatus `json:"-"`
 	// ApplyingError is an error that occurred trying to transition the container
 	// to its desired state. It is propagated to the backend in the form
 	// 'Name: ErrorString' as the 'reason' field.
@@ -241,12 +243,14 @@ func (c *Container) GetKnownStatus() ContainerStatus {
 	return c.KnownStatusUnsafe
 }
 
-// SetKnownStatus sets the known status of the container
+// SetKnownStatus sets the known status of the container and update the container
+// applied status
 func (c *Container) SetKnownStatus(status ContainerStatus) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
 	c.KnownStatusUnsafe = status
+	c.updateAppliedStatusUnsafe(status)
 }
 
 // GetDesiredStatus gets the desired status of the container
@@ -547,4 +551,39 @@ func (c *Container) GetHealthStatus() HealthStatus {
 	}
 
 	return copyHealth
+}
+
+// updateAppliedStatusUnsafe updates the container transitioning status
+func (c *Container) updateAppliedStatusUnsafe(knownStatus ContainerStatus) {
+	if c.AppliedStatus == ContainerStatusNone {
+		return
+	}
+
+	// Check if the container transition has already finished
+	if c.AppliedStatus <= knownStatus {
+		c.AppliedStatus = ContainerStatusNone
+	}
+}
+
+// SetAppliedStatus sets the applied status of container and returns whether
+// the container is already in a transition
+func (c *Container) SetAppliedStatus(status ContainerStatus) bool {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.AppliedStatus != ContainerStatusNone {
+		// return false to indicate the set operation failed
+		return false
+	}
+
+	c.AppliedStatus = status
+	return true
+}
+
+// GetAppliedStatus returns the transitioning status of container
+func (c *Container) GetAppliedStatus() ContainerStatus {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.AppliedStatus
 }
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -618,8 +618,8 @@ func (mtask *managedTask) progressContainers() {
 	// We've kicked off one or more transitions, wait for them to
 	// complete, but keep reading events as we do. in fact, we have to for
 	// transitions to complete
-	mtask.waitForContainerTransitions(transitions, transitionChange, transitionChangeContainer)
-	seelog.Debugf("Managed task [%s]: done transitioning all containers", mtask.Arn)
+	mtask.waitForContainerTransition(transitions, transitionChange, transitionChangeContainer)
+	seelog.Debugf("Managed task [%s]: wait for container transition done", mtask.Arn)
 
 	// update the task status
 	changed := mtask.UpdateStatus()
@@ -663,16 +663,30 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
 			reasons = append(reasons, transition.reason)
 			continue
 		}
+
+		// If the container is already in a transition, skip
+		if transition.actionRequired && !cont.SetAppliedStatus(transition.nextState) {
+			// At least one container is able to be moved forwards, so we're not deadlocked
+			anyCanTransition = true
+			continue
+		}
+
 		// At least one container is able to be moved forwards, so we're not deadlocked
 		anyCanTransition = true
 
 		if !transition.actionRequired {
-			mtask.handleContainerChange(dockerContainerChange{
-				container: cont,
-				event: DockerContainerChangeEvent{
-					Status: transition.nextState,
-				},
-			})
+			// Updating the container status without calling any docker API, send in
+			// a goroutine so that it won't block here before the waitForContainerTransition
+			// was called after this function. And all the events sent to mtask.dockerMessages
+			// will be handled by handleContainerChange.
+			go func(cont *api.Container, status api.ContainerStatus) {
+				mtask.dockerMessages <- dockerContainerChange{
+					container: cont,
+					event: DockerContainerChangeEvent{
+						Status: status,
+					},
+				}
+			}(cont, transition.nextState)
 			continue
 		}
 		transitions[cont.Name] = transition.nextState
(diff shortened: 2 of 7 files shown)


agent/api/container.go
```diff
@@ -548,3 +553,38 @@ func (c *Container) GetHealthStatus() HealthStatus {
...
+
+// UpdateTransitioningStatusUnsafe updates the container transitioning status
+func (c *Container) UpdateTransitioningStatusUnsafe(knownStatus ContainerStatus) {
+	if c.transitioningStatus == ContainerNotTransition {
+		return
+	}
+
+	// Check if the container transition has already finished
+	if containerTransitioningStatusMapping[c.transitioningStatus] <= knownStatus {
+		c.transitioningStatus = ContainerNotTransition
```
adnxn: do we need a lock for this? also should we just be using the setter here?

richardpen:
@adnxn For question 1: functional tests only test the happy path, I will need to test the failure case where a container failed at different stage:
* Task with multiple containers
* Task that uses volume (including empty volume and non-empty volume)
* Task that uses volumesFrom
* Task that uses links
* Set a cluster to periodically run a bunch of tasks

Let me know if you can think of other test scenario that was missed here.

For 2, we don't need to save the state, as agent has already has the logic to handle duplicate create/start, for pull/stop agent will have to pull/stop again on agent restart.

agent/api/container.go
richardpen: This method should be called within SetKnownStatus, which has already held the lock. Setter is holding a lock where we don't need a lock here, that's why the name xxxUnsafe.

agent/api/container.go
```diff
@@ -548,3 +553,38 @@ func (c *Container) GetHealthStatus() HealthStatus {
...
+		// return false to indicate the set operation failed
+		return false
+	}
+
+	c.transitioningStatus = status
+	return true
+}
+
+// GetTransitioning returns the transitioning status of container
+func (c *Container) GetTransitioning() ContainerTransitioningStatus {
```
aaithal: can you please rename this to `TransitionStatus()`?

agent/api/containerstatus.go
```diff
@@ -49,9 +49,28 @@ const (
 	ContainerUnhealthy
 )
 
+const (
+	// ContainerNotTransition means the container isn't in the middle of a transition
+	ContainerNotTransition ContainerTransitioningStatus = iota
```
aaithal: Can this be `ContainerTransitionNone`?

agent/api/containerstatus.go
```diff
@@ -49,9 +49,28 @@ const (
...
+	// ContainerPulling means the container is in the process of pulling
+	ContainerPulling
```
aaithal: Can you make sure that `ContainerTransition` is a prefix for all of these? Example: `ContainerTransitionPulling`

agent/api/containerstatus.go
```diff
@@ -49,9 +49,28 @@ const (
...
+	// ContainerStopping means the container is in the process of stopping
+	ContainerStopping
+)
+
 // ContainerStatus is an enumeration of valid states in the container lifecycle
 type ContainerStatus int32
 
+// ContainerTransitioningStatus is an enumeration of valid container processing
+// status which indicates which status the container is being processed
+type ContainerTransitioningStatus int32
```
aaithal: Can you move all of Transition statuses and methods to a separate file?

agent/api/container.go
```diff
@@ -548,3 +553,38 @@ func (c *Container) GetHealthStatus() HealthStatus {
...
+
+	// Check if the container transition has already finished
+	if containerTransitioningStatusMapping[c.transitioningStatus] <= knownStatus {
+		c.transitioningStatus = ContainerNotTransition
+	}
+}
+
+// SetTransitioningStatus sets the transitioning status of container and returns whether
+// the container is already in a transition
+func (c *Container) SetTransitioningStatus(status ContainerTransitioningStatus) bool {
```
aaithal: `UpdateTransitioningTo` might be a better name here.

agent/engine/task_manager.go
```diff
@@ -663,16 +663,26 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
 		anyCanTransition = true
 
 		if !transition.actionRequired {
-			mtask.handleContainerChange(dockerContainerChange{
-				container: cont,
-				event: DockerContainerChangeEvent{
-					Status: transition.nextState,
-				},
-			})
+			go func(cont *api.Container, status api.ContainerStatus) {
```
aaithal: why was this logic changed? Can you please add documentation/comments for that as well?

agent/engine/task_manager.go
```diff
@@ -763,39 +773,21 @@ func (mtask *managedTask) onContainersUnableToTransitionState() {
...
-			if allWaitingOnPulled {
-				// We don't actually care to wait for 'pull' transitions to finish if
-				// we're just heading to stopped since those resources aren't
-				// inherently linked to this task anyways for e.g. gc and so on.
-				seelog.Debugf("Managed task [%s]: all containers are waiting for pulled transition; exiting early: %v",
-					mtask.Arn, transitions)
-				break
-			}
-		}
+	// Wait for one transition/acs/docker message
```
aaithal: Same comment here. Why was this changed? What's the impact? Can we please add documentation explaining the change?

agent/api/container.go
```diff
@@ -166,6 +166,9 @@ type Container struct {
 	// handled properly so that the state storage continues to work.
 	SentStatusUnsafe ContainerStatus `json:"SentStatus"`
 
+	// transitioningStatus is the status of the container
+	transitioningStatus ContainerTransitioningStatus
```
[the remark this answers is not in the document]
richardpen: Thanks for pointing this out, I don't know there is already `AppliedStatus` for the same purpose. After some investigation, I think we can just use the `AppliedStatus` and the Transitioning status related should be removed. I will update the PR.

agent/engine/task_manager.go
```diff
@@ -663,16 +663,29 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
-				container: cont,
-				event: DockerContainerChangeEvent{
-					Status: transition.nextState,
-				},
-			})
+			// Updating the container status without calling any docker API, send in
+			// a goroutine so that it won't block here before the waitForContainerTransitions
+			// was called after this function.
+			go func(cont *api.Container, status api.ContainerStatus) {
+				mtask.dockerMessages <- dockerContainerChange{
```
sharanyad: For my understanding - how do we still handle the container status changes here that `handleContainerChange` does when no action is required for the transition ?

agent/api/container.go
```diff
@@ -548,3 +552,38 @@ func (c *Container) GetHealthStatus() HealthStatus {
...
 }
+
+// updateAppliedStatusUnsafe updates the container transitioning status
+func (c *Container) updateAppliedStatusUnsafe(knownStatus ContainerStatus) {
+	if c.AppliedStatus == ContainerStatusNone {
+		return
+	}
+
+	// Check if the container transition has already finished
+	if c.AppliedStatus <= knownStatus {
```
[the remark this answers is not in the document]
richardpen: If the agent received duplicate docker events for some reason, this could happen. As the event can be a past status of the container.

agent/engine/task_manager.go
```diff
@@ -763,39 +776,23 @@ func (mtask *managedTask) onContainersUnableToTransitionState() {
 		mtask.emitTaskEvent(mtask.Task, taskUnableToTransitionToStoppedReason)
 		// TODO we should probably panic here
 	} else {
-		seelog.Criticalf("Managed task [%s]: voving task to stopped due to bad state", mtask.Arn)
+		seelog.Criticalf("Managed task [%s]: moving task to stopped due to bad state", mtask.Arn)
 		mtask.handleDesiredStatusChange(api.TaskStopped, 0)
 	}
 }
 
 func (mtask *managedTask) waitForContainerTransitions(transitions map[string]api.ContainerStatus,
```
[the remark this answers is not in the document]
richardpen: But it's still waiting for container transition. Let me know if you have a better suggestion.

agent/engine/task_manager.go
richardpen: All the events sent to `mtask.dockerMessage` will be handled by `handleContainerChange` which works the same as before except that we don't call `handleContainerChange` directly now. Is that what you are asking?

aaithal:
This requires a CHANGELOG entry as well. Also, should this be tagged for 1.17.3 milestone?

agent/engine/task_manager.go
aaithal: >All the events sent to `mtask.dockerMessage` will be handled by `handleContainerChange` which works the same as before except that we don't call `handleContainerChange` directly now

It'd be awesome if you added that as code comment

agent/statemanager/state_manager.go
```diff
@@ -60,7 +60,8 @@ const (
 	// 9) Add 'ipToTask' map to state file
 	// 10) Add 'healthCheckType' field in 'api.Container'
 	// 11) Add 'PrivateDNSName' field to 'api.ENI'
-	ECSDataVersion = 11
+	// 12) Remove `AppliedStatus` field form 'api.Container'
+	ECSDataVersion = 12
```
aaithal: This should still be `11`. `10` is what's released out there.

agent/engine/task_manager.go
sharanyad: @richardpen Got it, thanks

agent/engine/task_manager.go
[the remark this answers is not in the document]
sharanyad: oh it just makes it look like it is waiting for ALL the containers' transitions, while it is waiting for just one to finish. may be just call it `waitForContainerTransition`?

agent/engine/task_manager.go
```diff
@@ -763,39 +777,23 @@ func (mtask *managedTask) onContainersUnableToTransitionState() {
...
-				// inherently linked to this task anyways for e.g. gc and so on.
-				seelog.Debugf("Managed task [%s]: all containers are waiting for pulled transition; exiting early: %v",
-					mtask.Arn, transitions)
-				break
-			}
-		}
+	// There could be multiple transitions, but we just need to wait for one of them
+	// to ensure that there is at least one container can be processed in the next
+	// progressContainers. This is done by waiting for one transition/acs/docker message.
+	if mtask.waitEvent(transitionChange) {
```
sharanyad: If we do not wait for a container transition here, next call of `progressContainers` would happen, which may be a no-op. is that the reason we are waiting here?

richardpen: yes, otherwise it will exhaust the cpu resource with an empty for loop.

agent/api/container.go
```diff
@@ -548,3 +552,38 @@ func (c *Container) GetHealthStatus() HealthStatus {
...
+}
+
+// SetAppliedStatus sets the applied status of container and returns whether
+// the container is already in a transition
+func (c *Container) SetAppliedStatus(status ContainerStatus) bool {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	if c.AppliedStatus != ContainerStatusNone {
+		// return false to indicate the set operation failed
```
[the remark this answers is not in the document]
richardpen: No, failed means the container is already in a transition(where the appliedstatus isn't none). This ensures that agent won't call the same API(pull/create/start/stop) twice for the same container.

agent/engine/task_manager.go
```diff
@@ -763,39 +777,23 @@ func (mtask *managedTask) onContainersUnableToTransitionState() {
...
-			seelog.Debugf("Managed task [%s]: still waiting for: %v", mtask.Arn, transitions)
-		}
-		if mtask.GetDesiredStatus().Terminal() || mtask.GetKnownStatus().Terminal() {
-			allWaitingOnPulled := true
-			for _, desired := range transitions {
-				if desired != api.ContainerPulled {
-					allWaitingOnPulled = false
-				}
-			}
-			if allWaitingOnPulled {
```
sharanyad: How is this condition handled in this new change?

richardpen: In previously the pull will block the whole transitions, that's why we want to break the wait if allWaitingOnPulled. Where the new change makes it non-blocked, which is handled by default.

agent/engine/task_manager.go
```diff
@@ -763,39 +777,23 @@ func (mtask *managedTask) onContainersUnableToTransitionState() {
...
-				seelog.Debugf("Managed task [%s]: all containers are waiting for pulled transition; exiting early: %v",
-					mtask.Arn, transitions)
-				break
-			}
-		}
+	// There could be multiple transitions, but we just need to wait for one of them
+	// to ensure that there is at least one container can be processed in the next
+	// progressContainers. This is done by waiting for one transition/acs/docker message.
+	if mtask.waitEvent(transitionChange) {
+		changedContainer := <-transitionChangeContainer
```
aaithal: minor: `transitionedContainer` is a better name for this

agent/engine/task_manager.go
```diff
@@ -763,39 +777,23 @@ func (mtask *managedTask) onContainersUnableToTransitionState() {
...
-				// inherently linked to this task anyways for e.g. gc and so on.
-				seelog.Debugf("Managed task [%s]: all containers are waiting for pulled transition; exiting early: %v",
-					mtask.Arn, transitions)
-				break
-			}
-		}
+	// There could be multiple transitions, but we just need to wait for one of them
+	// to ensure that there is at least one container can be processed in the next
+	// progressContainers. This is done by waiting for one transition/acs/docker message.
+	if mtask.waitEvent(transitionChange) {
```
aaithal: minor: `transition` is a better name for this

agent/engine/task_manager.go
```diff
@@ -618,8 +618,8 @@ func (mtask *managedTask) progressContainers() {
 	// We've kicked off one or more transitions, wait for them to
 	// complete, but keep reading events as we do. in fact, we have to for
 	// transitions to complete
-	mtask.waitForContainerTransitions(transitions, transitionChange, transitionChangeContainer)
-	seelog.Debugf("Managed task [%s]: done transitioning all containers", mtask.Arn)
+	mtask.waitForContainerTransition(transitions, transitionChange, transitionChangeContainer)
+	seelog.Debugf("Managed task [%s]: wait for container transition done", mtask.Arn)
```
aaithal: I don't think we need this line. You're logging in `waitForContainerTransition()` anyway

agent/engine/task_manager.go
aaithal: This also needs an `else` part, where we can log that we were interrupted or the transition did not finish or something like that.

PR #1308, opened 2018-03-20
Deprecate 'TransitionDependencySet' for 'TransitionDependenciesMap'

sharanyad:
### Summary
Handle the unmarshaling of old 'TransitionDependencySet' in the state files to the new type 'TransitionDependenciesMap' introduced in https://github.com/aws/amazon-ecs-agent/pull/1279

### Implementation details
- Use the old tag name 'TransitionDependencySet' for the new 'TransitionDependenciesMap' type. In JSON unmarshal, check if the type of the json blob is a map. If not, unmarshal to the old type and convert it to the new map type. 
- Bump state file version

### Testing

### Licensing

### DIFF
--- a/agent/api/transitiondependency.go
+++ b/agent/api/transitiondependency.go
@@ -1,4 +1,24 @@
+// Copyright 2017-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
 package api
+
+import (
+	"encoding/json"
+
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
 
 // TransitionDependencySet contains dependencies that impact transitions of
 // containers.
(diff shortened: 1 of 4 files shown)


agent/api/transitiondependency.go
```diff
@@ -19,3 +26,35 @@ type ContainerDependency struct {
...
+// TransitionDependenciesMap is a map of the dependent container status to other
+// dependencies that must be satisfied.
+type TransitionDependenciesMap map[ContainerStatus]TransitionDependencySet
+
+// UnmarshalJSON decodes the TransitionDependencySet tag in the JSON encoded string
+// into the TransitionDependenciesMap object
+func (td *TransitionDependenciesMap) UnmarshalJSON(b []byte) error {
+	validMap := make(map[ContainerStatus]TransitionDependencySet)
+	if err := json.Unmarshal(b, &validMap); err != nil {
+		seelog.Debugf("Unmarshal 'TransitionDependencySet': not a map: %v", err)
```
richardpen: Can we log the content here?

agent/api/transitiondependency.go
```diff
@@ -19,3 +26,35 @@ type ContainerDependency struct {
...
 }
+
+// TransitionDependenciesMap is a map of the dependent container status to other
+// dependencies that must be satisfied.
+type TransitionDependenciesMap map[ContainerStatus]TransitionDependencySet
+
+// UnmarshalJSON decodes the TransitionDependencySet tag in the JSON encoded string
+// into the TransitionDependenciesMap object
+func (td *TransitionDependenciesMap) UnmarshalJSON(b []byte) error {
+	validMap := make(map[ContainerStatus]TransitionDependencySet)
```
richardpen: Won't `make(TransitionDependenciesMap)` work? Also, can you use the name `dependencyMap` or similar?

agent/api/transitiondependency.go
```diff
@@ -1,5 +1,12 @@
 package api
 
+import (
```
richardpen: Can you add a header for this file, not sure how we missed it?

agent/api/transitiondependency_test.go
```diff
@@ -0,0 +1,57 @@
...
+	assert.NotNil(t, unmarshalledTdMap[ContainerRunning])
+	dep := unmarshalledTdMap[ContainerRunning].ContainerDependencies
+	assert.Len(t, dep, 1)
+	assert.Equal(t, dep[0].ContainerName, "container")
+	assert.Equal(t, dep[0].SatisfiedStatus, ContainerRunning)
+	assert.Equal(t, dep[0].DependentStatus, ContainerStatusNone)
+}
+
+func TestUnmarshalNewTransitionDependencySet(t *testing.T) {
+	bytes := []byte(`{"1":{"ContainerDependencies":[{"ContainerName":"container","SatisfiedStatus":"RUNNING"}]}}`)
```
richardpen: Can you confirm this will not be `PULLED` instead of `1`?

agent/api/transitiondependency.go
sharanyad: just noticing this. thanks, will add.

agent/api/transitiondependency_test.go
sharanyad: The marshal keeps the `ContainerStatus` as it is, and this was `"1"` instead of `"PULLED"`

agent/api/transitiondependency.go
sharanyad: Ideally, `make(TransitionDependenciesMap)` should work. It's weird that I'm running into stack overflow error during unmarshal JSON when I have this as `make(TransitionDependenciesMap)`. thinking if it's due to `ContainerStatus` being present as key in the map as well as in `TransitionDependencySet` (cyclic data?)

agent/api/transitiondependency_test.go
sharanyad: https://play.golang.org/p/Qg4gemS1oup. Not sure about why golang wouldn't use the custom marshal for the object when it is the map's key

sharanyad: I am not sure if that's the case, because that field in struct is still int(not a pointer) and hits the custom marshaler. This is only when the field is the key in a map.

richardpen: I think this is because golang will use the `TextMarshaler` for encoding the key of the map, so the key object will need to implement the `TextMarshaler` to customize the name. I think it's fine to have an integer here.

agent/api/transitiondependency.go
```diff
@@ -19,3 +39,35 @@ type ContainerDependency struct {
...
+
+// TransitionDependenciesMap is a map of the dependent container status to other
+// dependencies that must be satisfied.
+type TransitionDependenciesMap map[ContainerStatus]TransitionDependencySet
+
+// UnmarshalJSON decodes the TransitionDependencySet tag in the JSON encoded string
+// into the TransitionDependenciesMap object
+func (td *TransitionDependenciesMap) UnmarshalJSON(b []byte) error {
+	depMap := make(map[ContainerStatus]TransitionDependencySet)
+	if err := json.Unmarshal(b, &depMap); err != nil {
```
aaithal: can you deal with `err == nil` here instead and return early so that we can reduce indentation levels?

agent/api/transitiondependency.go
```diff
@@ -19,3 +39,35 @@ type ContainerDependency struct {
...
+		seelog.Debugf("Unmarshal 'TransitionDependencySet': %s, not a map: %v", string(b), err)
```
aaithal: I'm not sure if we really need a `Debugf` statement here.

sharanyad: It just helps know when we are dealing with old state files.

agent/api/transitiondependency_test.go
amogh09: Won't this cause an issue when a new container state is introduced in the future?

Issue #1292, opened 2018-03-12
Error getting message from ws backend


combor:
### Summary
ECS agent disconnects under heavy load. 

### Description
When I put my ECS instance under high load, like I scale my container instances from 2 to 12 the ecs agent disconnects with following errors:

```
2018-03-12T22:58:52Z [DEBUG] ACS activity occurred
2018-03-12T22:58:52Z [WARN] Unable to extend read deadline for ACS connection: set tcp 10.0.2.236:52020: use of closed network connection
2018-03-12T22:58:52Z [WARN] Unable to set read deadline for websocket connection: set tcp 10.0.2.236:52020: use of closed network connection for https://ecs-a-1.eu-west-1.amazonaws.com/ws?agentHash=edc3e260&agentVersion=1.17.2&clusterArn=redacted&containerInstanceArn=arn%3Aaws%3Aecs%3Aeu-west-1%3A936492824651%3Acontainer-instance%2Fca59a874-33ae-484b-9f51-654c5940b037&dockerVersion=DockerVersion%3A+17.12.0-ce&sendCredentials=true&seqNum=1
2018-03-12T22:58:52Z [ERROR] Error getting message from ws backend: error: [read tcp 10.0.2.236:52020->54.239.32.153:443: use of closed network connection], messageType: [-1]
```

After that is's marked as `Agent Connected: False` in ECS console until I restart the instance. 

I've got debug logs if you need more info.

richardpen:
@combor Can you share the logs with me via email: penyin (at) amazon.com

Thanks,
Peng

sharanyad:
@combor How long did the disconnection last before you restarted the agent? 
From the logs, it looks like agent is trying to reconnect after `274.540326ms`. 
`2018-03-12T22:58:52Z [INFO] Reconnecting to ACS in: 274.540326ms` 
Immediately agent receives a `SIGTERM` and shuts down, before restarting in ~30s.
`2018-03-12T22:58:52Z [DEBUG] Termination handler received termination signal: terminated`
.
`2018-03-12T22:59:35Z [INFO] Loading configuration`

combor:
I have restarted this one straight after it reported it's faulty as I knew it's not gonna reconnect. We can repeat this if you want and wait. How long you think it will be enough to wait?

sharanyad:
@combor 5 minutes would be good enough to determine if/why agent is not reconnecting.

combor:
Thanks @sharanyad 

[...]

PR #1310, opened 2018-03-21
Handle SetReadDeadline error for websocket connections (#1292)

aaithal:
### Summary
Added error handling for `SetReadDeadline()`. This should take care or scenarios where stale websocket connections linger (#1292)

### Implementation details

1. Forcefully terminate connection if there's any error with the
    SetReadDeadline() call. Not doing this leads to unpredictable behavior
    with stale websocket connections.

    Since SetWriteDeadline always returns `nil`, errors from that are not
    handled.

2. Move WebsocketConn to its own package. This allows us to mock it
    out in the "wsclient" package and write unit tests on it.

### Testing

### Licensing


agent/wsclient/client.go
```diff
@@ -230,15 +223,51 @@ func (cs *ClientServerImpl) IsReady() bool {
...
+	seelog.Warnf("Unable to set read deadline for websocket connection: %v for %s", err, cs.URL)
+	// If we get connection closed error from SetReadDeadline, break out of the for loop and
+	// return an error
+	if opErr, ok := err.(*net.OpError); ok && strings.Contains(opErr.Err.Error(), errClosed) {
+		seelog.Errorf("Stopping redundant reads on closed network connection: %s", cs.URL)
+		return opErr
+	}
+	// An unhandled error has occurred while trying to extend read deadline.
+	// Try asynchronously closing the connection. We don't want to be blocked on stale connections
+	// taking too long to close. The flip side is that we might start accumulating stale connections.
```
adnxn: mainly for my understanding - what exactly happens if we "start accumulating stale connections"?

agent/wsclient/client.go
```diff
@@ -230,15 +223,51 @@ func (cs *ClientServerImpl) IsReady() bool {
...
+	ctx, cancel := context.WithTimeout(context.TODO(), wsConnectTimeout)
+	defer cancel()
+	for {
+		select {
+		case closeErr := <-closeChan:
+			if closeErr != nil {
+				seelog.Warnf("Unable to close websocket connection: %v for %s",
+					closeErr, cs.URL)
+			}
+			return err
```
adnxn: again, mainly for my understanding. what exactly happens if we're "unable to close websocket connection" here? seems like we just move along. what side effects in the agent should we expect if we're unable to close lots of these websocket  connections here?

agent/wsclient/client.go
aaithal: at some point of time, the OS is going to stop the ECS agent from creating new connections if there's a chronic issue because of file descriptor limits.

agent/wsclient/client.go
aaithal: same as above. lmk if you have more questions.

agent/acs/handler/acs_handler.go
```diff
@@ -429,6 +429,7 @@ func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration,
 		if err := client.Close(); err != nil {
 			seelog.Warnf("Error disconnecting: %v", err)
 		}
+		seelog.Info("Disconnected from ACS")
```
yunhee-l: nit: I think copyright line in this file needs updating too.

misc/testnnp/Dockerfile
```diff
@@ -11,7 +11,7 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-FROM gcc:4.9
+FROM gcc:6.4
```
richardpen: Can we just use the latest here, in case we will need to update again in the future?

agent/wsclient/client.go
```diff
@@ -230,15 +225,51 @@ func (cs *ClientServerImpl) IsReady() bool {
...
+	// Try asynchronously closing the connection. We don't want to be blocked on stale connections
+	// taking too long to close. The flip side is that we might start accumulating stale connections.
+	// But, that still seems more desirable than waiting for ever for the connection to close
+	closeChan := make(chan error)
+	go func() {
+		closeChan <- cs.Close()
+	}()
+	ctx, cancel := context.WithTimeout(context.TODO(), wsConnectTimeout)
+	defer cancel()
+	for {
```
richardpen: looks like this doesn't need to be in the for loop, as both of the cases will return?

agent/wsclient/client_test.go
```diff
@@ -36,6 +40,11 @@ import (
 
 const dockerEndpoint = "/var/run/docker.sock"
 
+// Close closes the underlying connection
+func (cs *ClientServerImpl) Close() error {
```
richardpen: Why is this implemented in the test file?

agent/wsclient/client.go
petderek: Why do we need a loop if this only runs once? Select blocks forever and both of the cases return.

agent/wsclient/client.go
```diff
@@ -230,15 +225,51 @@ func (cs *ClientServerImpl) IsReady() bool {
...
+		return err
+	}
+	seelog.Warnf("Unable to set read deadline for websocket connection: %v for %s", err, cs.URL)
+	// If we get connection closed error from SetReadDeadline, break out of the for loop and
+	// return an error
+	if opErr, ok := err.(*net.OpError); ok && strings.Contains(opErr.Err.Error(), errClosed) {
+		seelog.Errorf("Stopping redundant reads on closed network connection: %s", cs.URL)
+		return opErr
+	}
+	// An unhandled error has occurred while trying to extend read deadline.
```
petderek: Two things stick out to me here. They're stylistic things, so FFTI:
* This doesn't follow the standard `err != nil` pattern where only good things happen on the left side.
* In my opinion, all of the second half of this should be in a separate method from `SetReadDeadline`. 

Lets say you extracted the server close logic into `closeCSAsync`. I'd maybe go with a structure like this:
```
err := cs.Conn.SetReadDeadline()
if err != nil {
   // check for opErr
       // return opErr
   servErr := closeCSAsync()
   // check for servErr
}
```

misc/testnnp/Dockerfile
aaithal: I don't want to use `latest` as it makes it hard to audit and track changes

agent/wsclient/client.go
```diff
@@ -60,6 +63,8 @@ const (
 
 	// Default NO_PROXY env var IP addresses
 	defaultNoProxyIP = "169.254.169.254,169.254.170.2"
+
+	errClosed = "use of closed network connection"
```
sharanyad: Can we have the error name more descriptive like `errClosedConnection` or similar?

agent/wsclient/client.go
aaithal: you're right. will modify that

agent/wsclient/client_test.go
aaithal: `ClientServerImpl` is a partial implementation of the interface, which lacks the `Close` method. If this is not there, there'll be NPEs. I can document that to make it more clear.

agent/wsclient/client.go
```diff
@@ -230,15 +225,51 @@ func (cs *ClientServerImpl) IsReady() bool {
...
+				seelog.Warnf("Unable to close websocket connection: %v for %s",
+					closeErr, cs.URL)
+			}
+			return err
+		case <-ctx.Done():
+			if ctx.Err() != nil {
+				seelog.Warnf("Context canceled waiting for termination of websocket connection: %v for %s",
+					ctx.Err(), cs.URL)
+			}
+			return err
```
petderek: nit:
Its a bit annoying to figure out what err is all the way down here. Maybe make it clear which err is getting returned in a comment? (Its the original deadline err, right?)

agent/wsclient/client.go
```diff
@@ -230,15 +223,49 @@ func (cs *ClientServerImpl) IsReady() bool {
...
 func (cs *ClientServerImpl) SetReadDeadline(t time.Time) error {
-	return cs.conn.SetReadDeadline(t)
+	err := cs.conn.SetReadDeadline(t)
+	if err == nil {
+		return nil
+	}
+	seelog.Warnf("Unable to set read deadline for websocket connection: %v for %s", err, cs.URL)
+	// If we get connection closed error from SetReadDeadline, break out of the for loop and
+	// return an error
+	if opErr, ok := err.(*net.OpError); ok && strings.Contains(opErr.Err.Error(), errClosed) {
```
sharanyad: Should we do a `cs.Close()` somewhere in this case too?

agent/wsclient/client.go
aaithal: `errClosed` is pretty descriptive, no? I can change it you have a *strong* preference

agent/wsclient/client.go
aaithal: I don't think so since it's already closed.

agent/acs/update_handler/os/mock/filesystem.go
```diff
@@ -11,146 +11,172 @@
...
 import (
 	io "io"
 	os "os"
+	reflect "reflect"
 
 	gomock "github.com/golang/mock/gomock"
 )
 
-// Mock of FileSystem interface
+// MockFileSystem is a mock of FileSystem interface
```
tehshwann: Nit: // MockFileSystem is a mock of **the** FileSystem interface

agent/api/mocks/api_mocks.go
```diff
@@ -11,184 +11,213 @@
...
 import (
+	reflect "reflect"
+
 	api "github.com/aws/amazon-ecs-agent/agent/api"
 	ecs "github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	gomock "github.com/golang/mock/gomock"
 )
 
-// Mock of ECSSDK interface
+// MockECSSDK is a mock of ECSSDK interface
```
tehshwann: Nit: **the* ECSSDK interface

agent/wsclient/client.go
tehshwann: How do we know that? Do we have a test that ensures that the expected behavior happens when we hit the limit?

PR #1311, opened 2018-03-21
Plugin auto detection to add to agent capability

yunhee-l:
### Summary
Auto-detection of installed plugins by agent to add to capabilities.

### Implementation details
Docker pkg's scan() and plugin list API are called to list installed volume driver plugins and appends them to capabilities list

### Testing

New tests cover the changes: 

### Licensing


samuelkarp:
I'd strongly recommend against using `github.com/docker/docker/pkg/plugins` and auto-detecting plugins that can't be listed with `plugin ls`.  `github.com/docker/docker/pkg/plugins` is an implementation detail of Docker rather than an API, and I'd prefer that we don't couple ourselves to the implementation details.  We've had issues with this in the past, specifically around coupling ourselves to the implementation details that powered `docker stats` instead of using the `docker stats` API; when the implementation changed, we had to rewrite our integration to use the API.

yunhee-l:
@samuelkarp we discussed that specific issue during a design meeting, but ultimately chose a solution that provides more consistent user experience. So it is a potential technical debt we are knowingly taking on.

agent/app/agent_capability.go
```diff
@@ -197,6 +201,28 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
 	return capabilities
 }
 
+func (agent *ecsAgent) appendVolumeDriverCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	// for non-standardized plugins, call docker pkg's plugins.Scan()
+	nonStandardizedPlugins, err1 := agent.scanPlugins()
+	if err1 == nil {
```
petderek: nit: I don't think we need different error names since they are in the same scope and will be overwritten.

agent/app/agent_capability_test.go
```diff
@@ -68,6 +68,8 @@ func TestCapabilities(t *testing.T) {
 			dockerclient.Version_1_19,
 		}),
 		cniClient.EXPECT().Version(ecscni.ECSENIPluginName).Return("v1", nil),
+		client.EXPECT().ListPluginsWithFilters(gomock.Any(), gomock.Any(), gomock.Any()).Return(
+			[]string{"coolvolumedriver"}, nil),
```
petderek: I want to use this volume driver.

yunhee-l:
Integ test will be added, but we need ecs-init update for it to work (to mount the plugin dir). This is in the todo list but for later.

agent/app/agent_capability.go
```diff
@@ -54,6 +55,7 @@ const (
 //    com.amazonaws.ecs.capability.ecr-auth
 //    com.amazonaws.ecs.capability.task-iam-role
 //    com.amazonaws.ecs.capability.task-iam-role-network-host
+//    com.amazonaws.ecs.capability.docker-volume-driver.${driverName}
```
aaithal: we've deprecated `com.*` style for attribute names. Can you please use the `ecs.docker-volume-driver.*` instead?

agent/app/agent_capability.go
```diff
@@ -197,6 +201,28 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
...
+			capabilities = appendNameOnlyAttribute(capabilities, capabilityPrefix+"docker-volume-driver."+string(pluginName))
+		}
+	}
+
+	// for standardized plugins, call docker's plugin ls API
+	pluginEnabled := true
+	volumeDriverType := []string{ecsengine.VolumeDriverType}
+	standardizedPlugins, err := agent.dockerClient.ListPluginsWithFilters(pluginEnabled, volumeDriverType, ecsengine.ListPluginsTimeout)
+
+	if err == nil {
```
aaithal: a minor refactor to reduce indentation levels:

```golang
if err != nil {
 log.Warn(...)
 return capabilities
}

for _, ...
```

agent/app/agent_capability.go
```diff
@@ -197,6 +201,28 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
 	return capabilities
 }
 
+func (agent *ecsAgent) appendVolumeDriverCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	// for non-standardized plugins, call docker pkg's plugins.Scan()
+	nonStandardizedPlugins, err := agent.scanPlugins()
+	if err  == nil {
```
aaithal: both here and a bit lower in the file, can you make sure that error gets logged when set?

agent/app/agent_compatibility_linux.go
```diff
@@ -61,3 +62,13 @@ func (agent *ecsAgent) checkCompatibility(engine engine.TaskEngine) error {
 
 	return nil
 }
+
+// scanPlugins returns the list of plugin names from docker pkg's plugin utilities method.
+func (agent *ecsAgent) scanPlugins() ([]string, error) {
+	pluginNames, err := plugins.Scan()
```
aaithal: can you make sure that `plugins.Scan` is behind an interface so that we can unit test that path as well? on that note, we need unit test coverage for this i think

yunhee-l: did you mean agent.scanPlugins to be behind an interface? plugins.Scan() is in docker's pkg.

agent/app/agent_capability.go
yunhee-l: scanPlugins() already logs when it fails to scan the directory so I didn't do so here. It won't hurt to log again here though. Will add logging.

agent/app/agent_capability.go
yunhee-l: Fixed, and added comment to not use capabilityPrefix for new capabilities.

agent/app/agent_compatibility_linux.go
aaithal: Capturing the offline conversation here for the sake of completeness. I meant defining an interface and having that implemented with `plugins.Scan()` in production code and mocking it out in test code. Example: https://github.com/aws/amazon-ecs-agent/blob/edc3e26016e363d7e8e6c044b11bde9bbb8853a8/agent/utils/oswrapper/os.go#L19

agent/app/agent_capability.go
```diff
@@ -25,6 +26,7 @@ import (
 )
 
 const (
+	// capabilityPrefix is deprecated. For new capabilities, use attributePrefix.
```
aaithal: thank you!

agent/app/agent_capability.go
```diff
@@ -197,6 +202,32 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
 	return capabilities
 }
 
+func (agent *ecsAgent) appendVolumeDriverCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	// for non-standardized plugins, call docker pkg's plugins.Scan()
+	nonStandardizedPlugins, err := agent.scanPlugins()
+	if err  != nil {
+		seelog.Warnf("scanning plugins failed: %v", err)
```
aaithal: minor: Please make sure that log lines being with uppercase letters

agent/app/agent_compatibility_unspecified.go
```diff
@@ -20,3 +20,7 @@ import "github.com/aws/amazon-ecs-agent/agent/engine"
 func (agent *ecsAgent) checkCompatibility(engine engine.TaskEngine) error {
 	return nil
 }
+
+func (agent *ecsAgent) scanPlugins() ([]string, error) {
```
richardpen: Doesn't the `Scan()` work on windows?

yunhee-l: https://docs.docker.com/engine/extend/plugin_api/#plugin-discovery they are specifically scanning for UNIX domain sockets, which is why I've put it under _unix only. Not entirely sure how it works on Windows, I couldn't find any answers googling.

richardpen: https://github.com/moby/moby/pull/25903 I think this solves part of the problem where spec file is supported for now.

agent/app/agent_compatibility_linux.go
yunhee-l: Added interface, open to suggestions on better naming of it and its mock.

agent/app/agent_capability.go
```diff
@@ -197,6 +202,33 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
...
+	pluginEnabled := true
+	volumeDriverType := []string{ecsengine.VolumeDriverType}
+	standardizedPlugins, err := agent.dockerClient.ListPluginsWithFilters(pluginEnabled, volumeDriverType, ecsengine.ListPluginsTimeout)
+	if err != nil {
+		seelog.Warnf("Listing plugins with filters enabled=%t, capabilities=%v failed: %v", pluginEnabled, volumeDriverType, err)
+		return capabilities
+	}
+
+	for _, pluginName := range standardizedPlugins {
+		capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+"docker-volume-driver."+string(pluginName))
```
richardpen: Can you make `docker-volume-driver` a constant?

agent/utils/mobypkgwrapper/Plugins.go
```diff
@@ -0,0 +1,37 @@
...
+type plugins struct {
+}
+
+// NewPlugins creates a new Plugins object
+func NewPlugins() Plugins {
+	return &plugins{}
+}
+
+func (*plugins) Scan() ([]string, error) {
+	fmt.Println("real scan")
```
sharanyad: unwanted print statement?

agent/utils/mobypkgwrapper/Plugins.go
```diff
@@ -0,0 +1,37 @@
...
+	return mobyplugins.Scan()
+}
```
sharanyad: nit: empty line

agent/utils/mobypkgwrapper/Plugins.go
```diff
@@ -0,0 +1,37 @@
...
+
+package mobypkgwrapper
+
+import (
+	"fmt"
+	mobyplugins "github.com/docker/docker/pkg/plugins"
+)
+
+// Plugins wraps moby/pkg/plugins methods for testing
+type Plugins interface {
```
richardpen: Does it make sense to make the Scan a method of the `DockerClient`, so you don't need to create a new interface and also you don't need a new obj in agent struct?

agent/utils/mobypkgwrapper/Plugins.go
yunhee-l: oops, I will remove this!

agent/utils/mobypkgwrapper/Plugins.go
adnxn: i was wondering the same thing, but i also think we're trying to keep this isolated so we can deprecate with minimal fuss.

yunhee-l: No DockerClient is a wrapper for go-dockerclient (fsouza/go-dockerclient repo), this is for Moby (moby/moby repo). I didn't want to mix them up, and we should be more careful of adding more moby/moby dependency.

PR #1314, opened 2018-03-22
taskresource/cgroup: remove task dependency

sharanyad:
### Summary
Minor refactor - Remove task dependency for resource creation.

### Implementation details
Changing method `Create(task *api.Task) error`  to `Create() error`. `api.Task` would contain list of resources in the future. Hence, removing dependency of task in resources package. 
Passing all the fields needed for resource creation to `CgroupResource` object. 

### Testing

### Licensing

### DIFF
--- a/agent/taskresource/cgroup/cgroup.go
+++ b/agent/taskresource/cgroup/cgroup.go
@@ -22,11 +22,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
 	"github.com/cihub/seelog"
 	"github.com/containerd/cgroups"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 )
 
@@ -42,9 +42,11 @@ var (
 
 // CgroupResource represents Cgroup resource
 type CgroupResource struct {
+	taskARN             string
 	control             cgroup.Control
-	CgroupRoot          string
-	CgroupMountPath     string
+	cgroupRoot          string
+	cgroupMountPath     string
+	resourceSpec        specs.LinuxResources
 	ioutil              ioutilwrapper.IOUtil
 	createdAt           time.Time
 	desiredStatusUnsafe CgroupStatus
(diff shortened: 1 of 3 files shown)


agent/taskresource/cgroup/cgroup.go
```diff
@@ -54,11 +56,13 @@ type CgroupResource struct {
 }
 
 // NewCgroupResource is used to return an object that implements the Resource interface
-func NewCgroupResource(control cgroup.Control, cgroupRoot string, cgroupMountPath string) *CgroupResource {
+func NewCgroupResource(taskArn string, control cgroup.Control, cgroupRoot string, cgroupMountPath string, resourceSpec specs.LinuxResources) *CgroupResource {
```
aaithal: minor: can you split this into multiple lines?

agent/taskresource/cgroup/cgroup.go
```diff
@@ -42,9 +42,11 @@ var (
 
 // CgroupResource represents Cgroup resource
 type CgroupResource struct {
+	taskArn             string
```
aaithal: lint: taskARN

agent/taskresource/cgroup/cgroup.go
```diff
@@ -22,11 +22,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
 	"github.com/cihub/seelog"
 	"github.com/containerd/cgroups"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
```
aaithal: Is this a new dependency? Does this require a dep update?

adnxn: looks like we track a [revision](https://github.com/aws/amazon-ecs-agent/blob/master/agent/Gopkg.toml#L62) for this dependency. since they have [versioned releases](https://github.com/opencontainers/runtime-spec/releases), we can track one of those. although this could be a separate PR against dev i guess? but ya, we should do this.

petderek: we were using this already in the cgroup codebase

sharanyad: This does need update. As @adnxn mentioned, we can make it track the latest release, which I shall do in another PR.

PR #1317, opened 2018-03-22
add resources to container's transition dependencies

sharanyad:
### Summary
This PR adds task resources as transition dependencies to the task's containers

### Implementation details
- add `[]TaskResource` to `api.Task`
- add `ResourceDependencies` to `api.Container`'s `TransitionDependenciesMap`
- `dependencygraph`: add check for dependency on resources to verify if all the container transition dependencies are resolved

### Testing

### Licensing


agent/api/container_unix_test.go
```diff
@@ -0,0 +1,35 @@
+// +build !windows,!integration
```
richardpen: Why this is not enabled on windows, the `BuildResourceDependency` isn't specific to linux?

agent/api/task.go
```diff
@@ -76,6 +77,8 @@ type Task struct {
 	Version string
 	// Containers are the containers for the task
 	Containers []*Container
+	// Resources are the resources for the task
+	Resources []taskresource.TaskResource
```
richardpen: Shouldn't this be a pointer/reference?

agent/api/container_unix_test.go
sharanyad: I am using `CgroupResource` for this test, which is specific to linux. hence added the test here. When we enable a common resource, we can add tests to `container_test.go`.

agent/api/container_unix_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"testing"
+
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/cgroup"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBuildResourceDependency(t *testing.T) {
```
adnxn: does `BuildResourceDependency` need to be tested by each resource we implement?

agent/api/task.go
adnxn: i think its an interface which are always a pointers

agent/engine/dependencygraph/graph.go
```diff
@@ -115,6 +117,11 @@ func DependenciesAreResolved(target *api.Container,
 		neededVolumeContainers[i] = volume.SourceContainer
 	}
 
+	resourcesMap := make(map[string]taskresource.TaskResource)
+	for _, resource := range resources {
```
adnxn: question: L117 above. the volume dependencies will get tracked through this resources dependency code path eventually yea?

agent/api/task.go
sharanyad: `TaskResource` is an interface, hence not adding reference.

agent/engine/dependencygraph/graph_test.go
```diff
@@ -133,28 +133,28 @@ func TestDependenciesAreResolvedWhenSteadyStateIsRunning(t *testing.T) {
 		},
 	}
 
-	err = DependenciesAreResolved(php, task.Containers, "", nil)
+	err = DependenciesAreResolved(php, task.Containers, "", nil, nil)
```
adnxn: nit: do you want to put these tests in a table? 😁

agent/taskresource/cgroup/cgroup.go
```diff
@@ -49,8 +51,8 @@ type CgroupResource struct {
 	resourceSpec        specs.LinuxResources
 	ioutil              ioutilwrapper.IOUtil
 	createdAt           time.Time
-	desiredStatusUnsafe CgroupStatus
-	knownStatusUnsafe   CgroupStatus
+	desiredStatusUnsafe taskresource.ResourceStatus
+	knownStatusUnsafe   taskresource.ResourceStatus
```
adnxn: please let me know if i've missed something. why are we naming this *Unsafe? its private field and also the accessor methods have locks around them. same question for both fields added here.

adnxn: actually just noticed the naming was already there. should we change it?

yunhee-l: They are named unsafe because they need to be accessed via get* methods that handle locks.

agent/engine/dependencygraph/graph.go
sharanyad: Those are dependencies on volume containers and can eventually be modelled as container dependencies.

sharanyad: The new volume driver based dependencies will follow the resource dependencies code path.

agent/api/container_unix_test.go
sharanyad: That should not be the case since `BuildResourceDependency` has the same behavior for all types of resources.

agent/api/transitiondependency.go
```diff
@@ -40,6 +45,15 @@ type ContainerDependency struct {
 	DependentStatus ContainerStatus `json:"DependentStatus,omitempty"`
 }
 
+// ResourceDependency defines the relationship between a dependent container
+// and its resource dependency.
+type ResourceDependency struct {
+	// ResourceName defines the Resource on which a transition depends
+	ResourceName string `json:"ResourceName"`
+	// SatisfiedStatus defines the status that satisfies the dependency
+	SatisfiedStatus taskresource.ResourceStatus `json:"SatisfiedStatus"`
```
aaithal: It'd be better to add a getter for this and use that method instead so that there's some flexibility in the future to potentially tweak that logic without changing code at a lot of places. something like `RequiredStatus() taskresource.ResourceStatus`. Also, `RequiredStatus` sounds like a better name.

agent/api/transitiondependency.go
```diff
@@ -40,6 +45,15 @@ type ContainerDependency struct {
 	DependentStatus ContainerStatus `json:"DependentStatus,omitempty"`
 }
 
+// ResourceDependency defines the relationship between a dependent container
+// and its resource dependency.
+type ResourceDependency struct {
+	// ResourceName defines the Resource on which a transition depends
+	ResourceName string `json:"ResourceName"`
```
aaithal: `Name` could be sufficient here.

agent/engine/dependencygraph/graph.go
```diff
@@ -202,6 +211,21 @@ func verifyContainerDependenciesResolved(target *api.Container, existingContaine
...
 
+func verifyResourceDependenciesResolved(target *api.Container, existingResources map[string]taskresource.TaskResource) bool {
+	targetNext := target.GetNextKnownStateProgression()
+	resourceDependencies := target.TransitionDependenciesMap[targetNext].ResourceDependencies
+	for _, resourceDependency := range resourceDependencies {
+		dep, exists := existingResources[resourceDependency.ResourceName]
+		if !exists {
+			return false
+		}
+		if dep.GetKnownStatus() < resourceDependency.SatisfiedStatus {
```
aaithal: please use a getter here (and everywhere else) instead of accessing the field directly.

agent/api/container.go
```diff
@@ -565,3 +566,23 @@ func (c *Container) BuildContainerDependency(contName string,
 	deps.ContainerDependencies = append(deps.ContainerDependencies, contDep)
 	c.TransitionDependenciesMap[dependentStatus] = deps
 }
+
+// BuildResourceDependency adds a new resource dependency and status that satisfies
+// the dependency
+func (c *Container) BuildResourceDependency(resourceName string,
```
aaithal: The intent is to invoke this in `PostUnmarshalTask()`?

agent/engine/dependencygraph/graph.go
```diff
@@ -124,7 +131,7 @@ func DependenciesAreResolved(target *api.Container,
 	}
 
 	if !verifyStatusResolvable(target, nameMap, target.SteadyStateDependencies, onSteadyStateIsResolved) ||
-		!verifyTransitionDependenciesResolved(target, nameMap) {
+		!verifyTransitionDependenciesResolved(target, nameMap, resourcesMap) {
```
aaithal: You'd need `verifyTransitionDependenciesResolved` to return a `bool` and an `error`. Otherwise, both container and resource dependencies not being satisfied will be clubbed under `DependentContainerNotResolvedErr`, which would be misleading

agent/api/container.go
```diff
@@ -565,3 +566,23 @@ func (c *Container) BuildContainerDependency(contName string,
...
+	satisfiedStatus taskresource.ResourceStatus,
+	dependentStatus ContainerStatus) {
+	if c.TransitionDependenciesMap == nil {
```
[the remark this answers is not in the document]
sharanyad: Container/Resource dependencies are optional for a container. Hence, not unnecessarily initializing the map. Also, we modify/add to this map only through `BuildContainerDependency` or `BuildResourcesDependency`. Let me know if that's not right.

agent/api/container.go
sharanyad: Yes, we would add the cgroup/volume dependencies in `PostUnmarshalTask()`.

agent/api/container.go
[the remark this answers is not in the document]
aaithal: > Container/Resource dependencies are optional for a container.

Sure. But, we can still create an empty map, right? 

> Also, we modify/add to this map only through `BuildContainerDependency` or `BuildResourceDependency`.

Exactly. We can avoid that duplicate code by moving that to a single place.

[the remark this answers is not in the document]
sharanyad: I think it makes sense. Would help avoiding unnecessary nil pointer errors in the future.

agent/engine/dependencygraph/graph_test.go
sharanyad: It looks like these tests have a single task and container dependencies, and the containers' known statuses are changed each time and checked if the dependencies are resolved. I don't think this can be put into a table since there is no pattern associated.

agent/api/container.go
```diff
@@ -565,3 +563,26 @@ func (c *Container) BuildContainerDependency(contName string,
 	deps.ContainerDependencies = append(deps.ContainerDependencies, contDep)
 	c.TransitionDependenciesMap[dependentStatus] = deps
 }
+
+// BuildResourceDependency adds a new resource dependency and status that satisfies
+// the dependency
+func (c *Container) BuildResourceDependency(resourceName string,
```
aaithal: Can you please document what each of these fields mean, with an example as well?

agent/api/container_unix_test.go
aaithal: why not create a fake resource and make this platform independent?

agent/api/task.go
```diff
@@ -156,12 +159,19 @@ type Task struct {
...
 	// TODO, add rudimentary plugin support and call any plugins that want to
 	// hook into this
+	task.initializeContainerDependenciesMap()
 	task.adjustForPlatform(cfg)
 	task.initializeEmptyVolumes()
 	task.initializeCredentialsEndpoint(credentialsManager)
 	task.addNetworkResourceProvisioningDependency(cfg)
 }
 
+func (task *Task) initializeContainerDependenciesMap() {
```
aaithal: Do you really need this method? Why not just do it in `UnmarshalJSON` for `api.Container`?

agent/engine/dependencygraph/graph_unix_test.go
```diff
@@ -0,0 +1,75 @@
+// +build !windows,!integration
```
aaithal: Here too, can we remove `!windows` and use a fake/mock dependency instead? Or, if the intent is to make sure that Cgroup dependency works as expected, you can keep this and still add tests around fake resource and dependencies on the same in the other file.

agent/api/task.go
sharanyad: Just realized we are already doing this: 
https://github.com/aws/amazon-ecs-agent/blob/f10cd818e756880d167610ea9489bc44895281c6/agent/api/transitiondependency.go#L50 

Will remove this method.

agent/engine/dependencygraph/graph.go
```diff
@@ -175,16 +191,23 @@ func verifyStatusResolvable(target *api.Container, existingContainers map[string
...
+		return nil
 	}
 
-	// TODO: add verifyResourceDependenciesResolved check here
-	return verifyContainerDependenciesResolved(target, existingContainers)
+	if !verifyContainerDependenciesResolved(target, existingContainers) {
+		return ErrContainerDependencyNotResolved
+	}
+	if !verifyResourceDependenciesResolved(target, existingResources) {
+		return ErrResourceDependencyNotResolved
```
aaithal: Can you point me to the unit test where this path is being covered? I'm mostly looking for the place where the test verifies that this error (`ErrResourceDependencyNotResolved`) is returned when resource dependency is not satisfied.

sharanyad: There are tests for the method `verifyResourceDependenciesResolved` itself which returns a bool. I can move them up a method to check for the returned error.

aaithal: I think there's value in having a test that verifies the error returned as well. Feel free to either modify the existing one or to add a new one scoped just for that.

sharanyad: updated.

Issue #1212, opened 2018-01-24
Wire context into methods that invokes Docker API Call


richardpen:
### Summary

Agent wrapped all the Docker API call into an [interface](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_container_engine.go#L95), where a timeout duration was passed in as a parameter. A context was created inside the method itself to cancel the call, but there is no way to cancel the context from the caller. So instead of passing the timeout duration, a context should be created in the caller and passed to the call so that it can be canceled from the caller.

### Expected Behavior
The Docker API call can be canceled from the caller.

### Observed Behavior
Docker API can only be canceled with a timeout, there isn't a way to cancel it before the timeout in the caller.

aaithal:
Related: https://travis-ci.org/aws/amazon-ecs-agent/builds/348472578?utm_source=github_status&utm_medium=notification

```
--- FAIL: TestRemoveImageTimeout (0.01s)
	<autogenerated>:1: missing call(s) to *mock_dockeriface.MockClient.RemoveImage(is equal to image)
	<autogenerated>:1: aborting test due to missing call(s)
=== RUN   TestRemoveImage
--- FAIL: TestRemoveImage (0.02s)
	docker_client_test.go:970: Did not expect error, err: Could not transition to removing image; timed out after waiting 2ms
	<autogenerated>:1: missing call(s) to *mock_dockeriface.MockClient.RemoveImage(is equal to image)
	<autogenerated>:1: aborting test due to missing call(s)
```

richardpen:
```
--- FAIL: TestInspectContainerTimeout (0.00s)
 	<autogenerated>:1: missing call(s) to *mock_dockeriface.MockClient.InspectContainerWithContext(is equal to id, is anything)
 	<autogenerated>:1: aborting test due to missing call(s)
```

PR #1329, opened 2018-04-04
engine: wire context into methods that invoke Docker API call

haikuoliu:
This pr is to fix issue #1212.

### Summary

Pass the context into the Docker API call instead of creating context inside Docker API call. 

### Testing

### Licensing

### DIFF
--- a/agent/containermetadata/manager.go
+++ b/agent/containermetadata/manager.go
@@ -14,15 +14,16 @@
 package containermetadata
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
 	"github.com/aws/amazon-ecs-agent/agent/utils/oswrapper"
-	"github.com/aws/amazon-ecs-agent/agent/api"
 
 	docker "github.com/fsouza/go-dockerclient"
 )
@@ -41,7 +42,7 @@ const (
 type Manager interface {
 	SetContainerInstanceARN(string)
 	Create(*docker.Config, *docker.HostConfig, *api.Task, string) error
-	Update(string, *api.Task, string) error
+	Update(context.Context, string, *api.Task, string) error
 	Clean(string) error
 }
 
@@ -116,9 +117,9 @@ func (manager *metadataManager) Create(config *docker.Config, hostConfig *docker
 }
 
 // Update updates the metadata file after container starts and dynamic metadata is available
-func (manager *metadataManager) Update(dockerID string, task *api.Task, containerName string) error {
+func (manager *metadataManager) Update(ctx context.Context, dockerID string, task *api.Task, containerName string) error {
 	// Get docker container information through api call
-	dockerContainer, err := manager.client.InspectContainer(dockerID, inspectContainerTimeout)
+	dockerContainer, err := manager.client.InspectContainer(ctx, dockerID, inspectContainerTimeout)
 	if err != nil {
 		return err
 	}
--- a/agent/engine/docker_image_manager.go
+++ b/agent/engine/docker_image_manager.go
@@ -1,4 +1,4 @@
-// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"). You may
 // not use this file except in compliance with the License. A copy of the
@@ -275,7 +275,7 @@ func (imageManager *dockerImageManager) performPeriodicImageCleanup(ctx context.
 	for {
 		select {
 		case <-imageManager.imageCleanupTicker.C:
-			go imageManager.removeUnusedImages()
+			go imageManager.removeUnusedImages(ctx)
 		case <-ctx.Done():
 			imageManager.imageCleanupTicker.Stop()
 			return
@@ -283,7 +283,7 @@ func (imageManager *dockerImageManager) performPeriodicImageCleanup(ctx context.
 	}
 }
 
-func (imageManager *dockerImageManager) removeUnusedImages() {
+func (imageManager *dockerImageManager) removeUnusedImages(ctx context.Context) {
 	seelog.Debug("Attempting to obtain ImagePullDeleteLock for removing images")
 	ImagePullDeleteLock.Lock()
 	seelog.Debug("Obtained ImagePullDeleteLock for removing images")
@@ -299,7 +299,7 @@ func (imageManager *dockerImageManager) removeUnusedImages() {
 		imageManager.imageStatesConsideredForDeletion[imageState.Image.ImageID] = imageState
 	}
 	for i := 0; i < imageManager.numImagesToDelete; i++ {
-		err := imageManager.removeLeastRecentlyUsedImage()
+		err := imageManager.removeLeastRecentlyUsedImage(ctx)
 		if err != nil {
 			seelog.Infof("End of eligible images for deletion: %v; Still have %d image states being managed", err, len(imageManager.getAllImageStates()))
 			break
@@ -307,12 +307,12 @@ func (imageManager *dockerImageManager) removeUnusedImages() {
 	}
 }
 
-func (imageManager *dockerImageManager) removeLeastRecentlyUsedImage() error {
+func (imageManager *dockerImageManager) removeLeastRecentlyUsedImage(ctx context.Context) error {
 	leastRecentlyUsedImage := imageManager.getUnusedImageForDeletion()
 	if leastRecentlyUsedImage == nil {
 		return fmt.Errorf("No more eligible images for deletion")
 	}
-	imageManager.removeImage(leastRecentlyUsedImage)
+	imageManager.removeImage(ctx, leastRecentlyUsedImage)
 	return nil
 }
 
@@ -326,28 +326,28 @@ func (imageManager *dockerImageManager) getUnusedImageForDeletion() *image.Image
 	return imageManager.getLeastRecentlyUsedImage(candidateImageStatesForDeletion)
 }
 
-func (imageManager *dockerImageManager) removeImage(leastRecentlyUsedImage *image.ImageState) {
+func (imageManager *dockerImageManager) removeImage(ctx context.Context, leastRecentlyUsedImage *image.ImageState) {
 	// Handling deleting while traversing a slice
 	imageNames := make([]string, len(leastRecentlyUsedImage.Image.Names))
 	copy(imageNames, leastRecentlyUsedImage.Image.Names)
 	if len(imageNames) == 0 {
 		// potentially untagged image of format <none>:<none>; remove by ID
-		imageManager.deleteImage(leastRecentlyUsedImage.Image.ImageID, leastRecentlyUsedImage)
+		imageManager.deleteImage(ctx, leastRecentlyUsedImage.Image.ImageID, leastRecentlyUsedImage)
 	} else {
 		// Image has multiple tags/repos. Untag each name and delete the final reference to image
 		for _, imageName := range imageNames {
-			imageManager.deleteImage(imageName, leastRecentlyUsedImage)
+			imageManager.deleteImage(ctx, imageName, leastRecentlyUsedImage)
 		}
 	}
 }
 
-func (imageManager *dockerImageManager) deleteImage(imageID string, imageState *image.ImageState) {
+func (imageManager *dockerImageManager) deleteImage(ctx context.Context, imageID string, imageState *image.ImageState) {
 	if imageID == "" {
 		seelog.Errorf("Image ID to be deleted is null")
 		return
 	}
 	seelog.Infof("Removing Image: %s", imageID)
-	err := imageManager.client.RemoveImage(imageID, dockerapi.RemoveImageTimeout)
+	err := imageManager.client.RemoveImage(ctx, imageID, dockerapi.RemoveImageTimeout)
 	if err != nil {
 		if err.Error() == imageNotFoundForDeletionError {
 			seelog.Errorf("Image already removed from the instance: %v", err)
--- a/agent/engine/docker_image_manager_integ_test.go
+++ b/agent/engine/docker_image_manager_integ_test.go
@@ -19,6 +19,7 @@ package engine
 
 import (
 	"container/list"
+	"context"
 	"errors"
 	"fmt"
 	"testing"
@@ -123,7 +124,9 @@ func TestIntegImageCleanupHappyCase(t *testing.T) {
 	}
 
 	// Call Image removal
-	imageManager.removeUnusedImages()
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	imageManager.removeUnusedImages(ctx)
 
 	// Verify top 2 LRU images are deleted from image manager
 	err = verifyImagesAreRemoved(imageManager, imageState1ImageID, imageState2ImageID)
@@ -241,7 +244,9 @@ func TestIntegImageCleanupThreshold(t *testing.T) {
 	}
 
 	// Call Image removal
-	imageManager.removeUnusedImages()
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	imageManager.removeUnusedImages(ctx)
 
 	// Verify Image1 & Image3 are removed from ImageManager as they are beyond the minimumAge threshold
 	err = verifyImagesAreRemoved(imageManager, imageState1ImageID, imageState3ImageID)
@@ -395,7 +400,9 @@ func TestImageWithSameNameAndDifferentID(t *testing.T) {
 	err = verifyTaskIsCleanedUp("task3", taskEngine)
 	assert.NoError(t, err, "task3")
 
-	imageManager.removeUnusedImages()
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	imageManager.removeUnusedImages(ctx)
 
 	// Verify all the three images are removed from image manager
 	err = verifyImagesAreRemoved(imageManager, imageID1, imageID2, imageID3)
@@ -524,7 +531,9 @@ func TestImageWithSameIDAndDifferentNames(t *testing.T) {
 	err = verifyTaskIsCleanedUp("task3", taskEngine)
 	assert.NoError(t, err, "task3")
 
-	imageManager.removeUnusedImages()
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	imageManager.removeUnusedImages(ctx)
 
 	// Verify all the images are removed from image manager
 	err = verifyImagesAreRemoved(imageManager, imageID1)
@@ -668,19 +677,23 @@ func verifyImagesAreNotRemoved(imageManager *dockerImageManager, imageIDs ...str
 }
 
 func cleanupImagesHappy(imageManager *dockerImageManager) {
-	imageManager.client.RemoveContainer("test1", dockerapi.RemoveContainerTimeout)
-	imageManager.client.RemoveContainer("test2", dockerapi.RemoveContainerTimeout)
-	imageManager.client.RemoveContainer("test3", dockerapi.RemoveContainerTimeout)
-	imageManager.client.RemoveImage(test1Image1Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test1Image2Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test1Image3Name, imageRemovalTimeout)
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	imageManager.client.RemoveContainer(ctx, "test1", dockerapi.RemoveContainerTimeout)
+	imageManager.client.RemoveContainer(ctx, "test2", dockerapi.RemoveContainerTimeout)
+	imageManager.client.RemoveContainer(ctx, "test3", dockerapi.RemoveContainerTimeout)
+	imageManager.client.RemoveImage(ctx, test1Image1Name, imageRemovalTimeout)
+	imageManager.client.RemoveImage(ctx, test1Image2Name, imageRemovalTimeout)
+	imageManager.client.RemoveImage(ctx, test1Image3Name, imageRemovalTimeout)
 }
 
 func cleanupImagesThreshold(imageManager *dockerImageManager) {
-	imageManager.client.RemoveContainer("test1", dockerapi.RemoveContainerTimeout)
-	imageManager.client.RemoveContainer("test2", dockerapi.RemoveContainerTimeout)
-	imageManager.client.RemoveContainer("test3", dockerapi.RemoveContainerTimeout)
-	imageManager.client.RemoveImage(test2Image1Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test2Image2Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test2Image3Name, imageRemovalTimeout)
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	imageManager.client.RemoveContainer(ctx, "test1", dockerapi.RemoveContainerTimeout)
+	imageManager.client.RemoveContainer(ctx, "test2", dockerapi.RemoveContainerTimeout)
+	imageManager.client.RemoveContainer(ctx, "test3", dockerapi.RemoveContainerTimeout)
+	imageManager.client.RemoveImage(ctx, test2Image1Name, imageRemovalTimeout)
+	imageManager.client.RemoveImage(ctx, test2Image2Name, imageRemovalTimeout)
+	imageManager.client.RemoveImage(ctx, test2Image3Name, imageRemovalTimeout)
 }
--- a/agent/engine/docker_image_manager_test.go
+++ b/agent/engine/docker_image_manager_test.go
@@ -15,6 +15,7 @@
 package engine
 
 import (
+	"context"
 	"errors"
 	"reflect"
 	"sync"
@@ -29,8 +30,6 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/engine/image"
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
-
-	"context"
 
 	docker "github.com/fsouza/go-dockerclient"
 	"github.com/golang/mock/gomock"
@@ -79,8 +78,10 @@ func TestImagePullRemoveDeadlock(t *testing.T) {
 		wg.Done()
 	}()
 
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
 	go func() {
-		imageManager.(*dockerImageManager).removeUnusedImages()
+		imageManager.(*dockerImageManager).removeUnusedImages(ctx)
 		wg.Done()
 	}()
 	wg.Wait()
@@ -730,8 +731,8 @@ func TestImageCleanupHappyPath(t *testing.T) {
 	imageState.LastUsedAt = time.Now().AddDate(0, -2, 0)
 	imageState.AddImageName("anotherImage")
 
-	client.EXPECT().RemoveImage(container.Image, dockerapi.RemoveImageTimeout).Return(nil)
-	client.EXPECT().RemoveImage("anotherImage", dockerapi.RemoveImageTimeout).Return(nil)
+	client.EXPECT().RemoveImage(gomock.Any(), container.Image, dockerapi.RemoveImageTimeout).Return(nil)
+	client.EXPECT().RemoveImage(gomock.Any(), "anotherImage", dockerapi.RemoveImageTimeout).Return(nil)
 	parent := context.Background()
 	ctx, cancel := context.WithCancel(parent)
 	go imageManager.performPeriodicImageCleanup(ctx, 2*time.Millisecond)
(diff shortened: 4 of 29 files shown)


agent/engine/docker_client.go
```diff
@@ -125,31 +125,32 @@ type DockerClient interface {
...
-	CreateContainer(*docker.Config, *docker.HostConfig, string, time.Duration) DockerContainerMetadata
+	// and a context should be provided for the request.
+	CreateContainer(*docker.Config, *docker.HostConfig, string, time.Duration, context.Context) DockerContainerMetadata
 
-	// StartContainer starts the container identified by the name provided. A timeout value should be provided for the
-	// request.
-	StartContainer(string, time.Duration) DockerContainerMetadata
+	// StartContainer starts the container identified by the name provided. A timeout value and a context should be
+	// provided for the request.
+	StartContainer(string, time.Duration, context.Context) DockerContainerMetadata
```
petderek: Actually, this is all repetitive as far as I can tell. Lets get rid of time.Duration in all of the signatures.

agent/engine/docker_client_test.go
```diff
@@ -438,20 +440,24 @@ func TestStartContainerTimeout(t *testing.T) {
...
 	defer done()
 
-	testDone := make(chan struct{})
 	wait := &sync.WaitGroup{}
 	wait.Add(1)
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+	mockDone := make(chan struct{})
+	mockContext := NewMockContext(mockCtrl)
+	mockContext.EXPECT().Done().Return(mockDone).AnyTimes()
```
richardpen: Why those are `AnyTimes`, shouldn't this be called just once or twice?

agent/engine/docker_client.go
```diff
@@ -474,15 +475,8 @@ func (dg *dockerGoClient) getAuthdata(image string, authData *api.RegistryAuthen
...
-	// Create a context that times out after the 'timeout' duration
-	// This is defined by the const 'createContainerTimeout'. Injecting the 'timeout'
-	// makes it easier to write tests.
-	// Eventually, the context should be initialized from a parent root context
-	// instead of TODO.
-	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
-	defer cancel()
-
+	timeout time.Duration,
+	ctx context.Context) DockerContainerMetadata {
```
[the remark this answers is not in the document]
haikuoliu: Sounds reasonable, so can I just return the timeout error here without checking the error type from the context?

agent/engine/docker_client.go
```diff
@@ -474,15 +475,8 @@ func (dg *dockerGoClient) getAuthdata(image string, authData *api.RegistryAuthen
...
-	timeout time.Duration) DockerContainerMetadata {
-	// Create a context that times out after the 'timeout' duration
-	// This is defined by the const 'createContainerTimeout'. Injecting the 'timeout'
-	// makes it easier to write tests.
-	// Eventually, the context should be initialized from a parent root context
-	// instead of TODO.
-	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
-	defer cancel()
-
+	timeout time.Duration,
```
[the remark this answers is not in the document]
haikuoliu: When returning a timeout error, the timeout value is needed, how to get this value?

agent/engine/docker_client.go
haikuoliu: Please see the comments above.

agent/engine/docker_client_test.go
haikuoliu: I think it should be called twice, because the `CreateContainer` will also call `InspectContainer` where a new child context will be generated, but test using mock context will be deleted anyway.

agent/engine/docker_client.go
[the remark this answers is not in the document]
petderek: If we really want to print the timeout as part of the error, we should assemble it in the caller (which already has both of these pieces of information).

agent/engine/docker_client.go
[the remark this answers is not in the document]
petderek: Precisely!

agent/engine/docker_client_test.go
petderek: My opinion: The number of times context is invoked is an implementation detail and not part of tested function's contract.  IE, its not something we would want the test to fail for.

agent/engine/docker_client.go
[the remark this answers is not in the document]
haikuoliu: I think logging the timeout is important, because most of the timeout values are not defined by users, so they have no idea how long they've waited for the API call in the most of the time. 

But assembling the error from the caller requires at least three lines of the same code, and this same piece of code will be spread wherever the Dock client API is called, and will require more code change, what do you think?

haikuoliu:
Situation when the go routine will not return using mock context:
Let's take the `TestCreateContainerTimeout` for example. 

In `CreateContainer` (call it `goroutine1`) it will create another go routine (call it `goroutine2`) to execute `createContainer`, and `goroutine2` will execute the real Docker API call, the one that we've mocked out. So instead of doing the real Docker API call, `goroutine2` will close the `mockDone` channel, and wait until the `wait.Done()` is executed. After the `mockDone` is closed, the `goroutine1` will exit and return the err as expected. Then the `wait.Done()` is executed, the mocked real Docker API call will return with the return value specified in the call chain of mock method. 

However, returned from mocked real Docker API call, `goroutine2` will continue to execute (it will not execute when using real context), and thus call `containerMetadata` and `InspectContainer`, which will result in a missing call error. But since the error happens after the test is done, the golang runtime will not give you the missing call error directly, it will give you error like 'some error happen after some test has completed' (don't remember the exact error text), so it's hard to observe this kind of bug in the test. 

The good part is I do learn a lot from this. Let me know if anything is unclear.

agent/engine/docker_client.go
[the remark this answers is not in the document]
petderek: You're right. I looked at the calling class (which is embedded in the transition function map) and I don't think it makes sense to change that at this point.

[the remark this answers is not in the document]
haikuoliu: OK, cool, I'll make the change accordingly.

haikuoliu:
Update: use real context in the test; for docker api call, return timeout error directly when context is done

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 1.17.3-dev
+* Enhancement - Wire context into methods that invoke Docker API call to fix flaky tests [#1329](https://github.com/aws/amazon-ecs-agent/pull/1321)
```
yunhee-l: I don't think we normally update Changelog for internal text fixes like this. Also, 1.17.3 just got released so it shouldn't be under 1.17.3-dev.

agent/containermetadata/manager.go
```diff
@@ -117,8 +118,10 @@ func (manager *metadataManager) Create(config *docker.Config, hostConfig *docker
 
 // Update updates the metadata file after container starts and dynamic metadata is available
 func (manager *metadataManager) Update(dockerID string, task *api.Task, containerName string) error {
+	ctx, cancel := context.WithTimeout(context.TODO(), inspectContainerTimeout)
```
aaithal: You should modify `Update(` to expect an arg for `context.Context` as well. The `DockerTaskEngine` struct already has a context, which it can pass in here.

agent/engine/docker_client.go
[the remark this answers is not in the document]
aaithal: For now, lets update the documentation for all of these methods that `timeout` is only used to generate the error and shouldn't be used for anything else. We can also add a `TODO` for changing these methods to instead accept an error type or error string as a param to use when timeout error occurs.

agent/engine/docker_client.go
```diff
@@ -493,15 +487,9 @@ func (dg *dockerGoClient) CreateContainer(config *docker.Config,
...
-		// Context has either expired or canceled. If it has timed out,
-		// send back the DockerTimeoutError
-		err := ctx.Err()
-		if err == context.DeadlineExceeded {
-			return DockerContainerMetadata{Error: &DockerTimeoutError{timeout, "created"}}
-		}
-		// Context was canceled even though there was no timeout. Send
-		// back an error.
-		return DockerContainerMetadata{Error: &CannotCreateContainerError{err}}
+		// Context has expired, send back the DockerTimeoutError
```
aaithal: Why was this modified. I think there's a case to be made for the previous code, where we distinguish between cancelled and timed out contexts.

agent/engine/docker_client.go
```diff
@@ -125,31 +125,32 @@ type DockerClient interface {
...
+	// StartContainer starts the container identified by the name provided. A timeout value and a context should be
+	// provided for the request.
+	StartContainer(string, time.Duration, context.Context) DockerContainerMetadata
 
-	// StopContainer stops the container identified by the name provided. A timeout value should be provided for the
-	// request.
-	StopContainer(string, time.Duration) DockerContainerMetadata
+	// StopContainer stops the container identified by the name provided. A timeout value and a context should be provided
+	// for the request.
+	StopContainer(string, time.Duration, context.Context) DockerContainerMetadata
```
aaithal: here, and in all methods, please make sure that `context.Context` is the first parameter to all the methods (which is the convention in golang)

CHANGELOG.md
haikuoliu: OK, I will remove it.

agent/engine/docker_client.go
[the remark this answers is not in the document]
haikuoliu: OK, sounds good.

agent/engine/docker_client.go
haikuoliu: OK

agent/containermetadata/manager.go
haikuoliu: OK, will change accordingly.

agent/engine/docker_client.go
haikuoliu: As Derek said, when the caller wants to cancel the context, the caller already know the reason,  the logic to handle the timeout or cancel is nothing but report the reason of the error.

haikuoliu:
Update: some function signatures and comments of Docker client API, change log, Update function in manager.go

agent/engine/docker_client_test.go
```diff
@@ -361,18 +360,20 @@ func TestCreateContainerTimeout(t *testing.T) {
 	mockDocker, client, _, _, _, done := dockerClientSetup(t)
 	defer done()
 
-	warp := make(chan time.Time)
+	ctx, cancel := context.WithTimeout(context.TODO(), xContainerLongTimeout)
+	defer cancel()
```
sharanyad: Since we expect the mock call to happen and we do a `cancel()` there, this `cancel()` can be removed.

agent/engine/docker_client.go
```diff
@@ -686,33 +643,25 @@ func (dg *dockerGoClient) stopContainer(ctx context.Context, dockerID string) Do
...
 	case resp := <-response:
 		return resp
 	case <-ctx.Done():
-		err := ctx.Err()
-		// Context has either expired or canceled. If it has timed out,
-		// send back the DockerTimeoutError
-		if err == context.DeadlineExceeded {
-			return &DockerTimeoutError{removeContainerTimeout, "removing"}
-		}
-		return &CannotRemoveContainerError{err}
```
sharanyad: Since we are not using this error type anymore, it can be removed.

sharanyad: Giving it some thought, you can capture the error [here](https://github.com/aws/amazon-ecs-agent/blob/487e699ce931a374d5e92a2ea6ddda702a637a69/agent/engine/docker_client.go#L720) and do
```
if err != nil {
    return &CannotRemoveContainerError{err}
}
```

agent/engine/docker_client.go
aaithal: That's not strictly true. If you look at https://github.com/aws/amazon-ecs-agent/pull/1329/files#diff-b9aa12c3425c34d6f330f28407f779f6R835, you'll see that the caller (owner of the context) is not the entity that's dealing with the error. I think there's value in the way we were previously handling errors until we figure out a better alternative.

haikuoliu: OK, cool, will change accordingly.

agent/engine/docker_image_manager.go
```diff
@@ -345,7 +345,9 @@ func (imageManager *dockerImageManager) deleteImage(imageID string, imageState *
 		return
 	}
 	seelog.Infof("Removing Image: %s", imageID)
-	err := imageManager.client.RemoveImage(imageID, removeImageTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), removeImageTimeout)
```
aaithal: You shouldn't be using `context.Background()` here. Instead, it should be wired all the way from `performPeriodicImageCleanup()`.

agent/engine/docker_task_engine.go
```diff
@@ -346,8 +348,9 @@ func (engine *DockerTaskEngine) synchronizeContainerStatus(container *api.Docker
 		}
 		return
 	}
-
-	currentState, metadata := engine.client.DescribeContainer(container.DockerID)
+	ctx, cancel := context.WithCancel(engine.ctx)
```
aaithal: why are you doing this? Can't you just use `engine.ctx` directly?

agent/engine/docker_task_engine.go
```diff
@@ -392,7 +395,9 @@ func (engine *DockerTaskEngine) CheckTaskState(task *api.Task) {
 		if !ok {
 			continue
 		}
-		status, metadata := engine.client.DescribeContainer(dockerContainer.DockerID)
+		ctx, cancel := context.WithCancel(engine.ctx)
```
aaithal: Same question as above. Can't you just use `engine.ctx` directly?

agent/engine/docker_task_engine.go
```diff
@@ -825,8 +830,9 @@ func (engine *DockerTaskEngine) createContainer(task *api.Task, container *api.C
 				task.Arn, container.Name, mderr)
 		}
 	}
-
-	metadata := client.CreateContainer(config, hostConfig, dockerContainerName, createContainerTimeout)
+	ctx, cancel := context.WithTimeout(engine.ctx, createContainerTimeout)
```
aaithal: This is not an optimal way of doing this. All of the responsibility of creating the child context for timing out should be happening within the docker_client methods. Please change all of  these methods so that context creation happens within docker_client methods. Just that the parent for those contexts need to flow from here.

agent/eni/pause/pause_linux.go
```diff
@@ -48,7 +49,9 @@ func loadFromFile(path string, dockerClient engine.DockerClient, fs os.FileSyste
 		return errors.Wrapf(err,
 			"pause container load: failed to read pause container image: %s", path)
 	}
-	if err := dockerClient.LoadImage(pauseContainerReader, engine.LoadImageTimeout); err != nil {
+	ctx, cancel := context.WithTimeout(context.TODO(), engine.LoadImageTimeout)
```
aaithal: You shouldn't need to create a new context here, using `context.TODO()`. You should instead wire it from the "app" package (via loader. LoadImage)

agent/stats/engine.go
```diff
@@ -123,7 +124,9 @@ func NewDockerStatsEngine(cfg *config.Config, client ecsengine.DockerClient, con
 
 // synchronizeState goes through all the containers on the instance to synchronize the state on agent start
 func (engine *DockerStatsEngine) synchronizeState() error {
-	listContainersResponse := engine.client.ListContainers(false, ecsengine.ListContainersTimeout)
+	ctx, cancel := context.WithTimeout(context.Background(), ecsengine.ListContainersTimeout)
```
aaithal: Here too, you shouldn't need to create a new context. Just use `engine.ctx` as the parent and let the child context creation happen in `ListContainers`

haikuoliu: There is no ctx attribute for stats engine, and I think creating a context for just one method seems to be redundant. So instead of creating a timeout context from background context, I will pass the background context directly to the method.

aaithal: You can create one :) Let it be set from app.ctx.

haikuoliu: ok, cool

agent/eni/pause/pause_linux.go
haikuoliu: ok, sounds good.

agent/engine/docker_image_manager.go
haikuoliu: Previously, in the RemoveImage function of docker client, it uses context.Background() directly, 
so I use context.Background() in the caller, you want me to do the change?

agent/engine/docker_client.go
```diff
@@ -164,9 +165,9 @@ type DockerClient interface {
 	InspectImage(string) (*docker.Image, error)
 
 	// RemoveImage removes the metadata associated with an image and may remove the underlying layer data. A timeout
-	// value should be provided for the request.
-	RemoveImage(string, time.Duration) error
-	LoadImage(io.Reader, time.Duration) error
+	// value and a context should be provided for the request.
+	RemoveImage(context.Context, string, time.Duration) error
+	LoadImage(context.Context, io.Reader, time.Duration) error
```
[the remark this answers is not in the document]
haikuoliu: ok, added.

agent/engine/docker_client.go
```diff
@@ -471,18 +472,13 @@ func (dg *dockerGoClient) getAuthdata(image string, authData *ecr.RegistryAuthen
 	return authConfig, nil
 }
 
-func (dg *dockerGoClient) CreateContainer(config *docker.Config,
+// The timeout parameter here is only used to assemble the return error message
```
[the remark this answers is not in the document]
haikuoliu: ok, done.

agent/engine/docker_task_engine.go
haikuoliu: Will change them all to engine.ctx.

agent/engine/docker_task_engine.go
haikuoliu: See comments above.

agent/engine/docker_task_engine.go
haikuoliu: See comments above.

agent/engine/docker_task_engine.go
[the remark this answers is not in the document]
haikuoliu: After the update for this PR is done, there is no need to pass the inspectContainerTimeout anymore.

agent/engine/docker_image_manager.go
aaithal: yes please

haikuoliu:
Update: add ctx attribute to the stats engine; move the creation of timeout context from caller to docker client API; wire context from caller instead of using context.TODO() or context.Background().

agent/engine/docker_client.go
```diff
@@ -588,32 +582,28 @@ func dockerStateToState(state docker.State) api.ContainerStatus {
...
+	dockerContainer, err := dg.InspectContainer(ctx, dockerID, inspectContainerTimeout)
 	if err != nil {
 		return api.ContainerStatusNone, DockerContainerMetadata{Error: CannotDescribeContainerError{err}}
 	}
 	return dockerStateToState(dockerContainer.State), metadataFromContainer(dockerContainer)
 }
 
-func (dg *dockerGoClient) InspectContainer(dockerID string, timeout time.Duration) (*docker.Container, error) {
+// InspectContainer uses the timeout parameter only to assemble the return error message
+// TODO: replace the timeout with a string type error message
```
aaithal: Can you get rid of this TODO (and others that were added)? It's out of date now.

agent/engine/docker_task_engine.go
```diff
@@ -957,7 +955,7 @@ func (engine *DockerTaskEngine) buildCNIConfigFromTaskContainer(task *api.Task,
 	if !ok {
 		return nil, errors.New("engine: failed to find the pause container")
 	}
-	containerInspectOutput, err := engine.client.InspectContainer(pauseContainer.DockerName, inspectContainerTimeout)
+	containerInspectOutput, err := engine.client.InspectContainer(engine.ctx, pauseContainer.DockerName, inspectContainerTimeout)
```
aaithal: nit: this can be split into multiple lines to improve readability

agent/engine/docker_client_test.go
```diff
@@ -363,28 +362,19 @@ func TestCreateContainerTimeout(t *testing.T) {
 	defer done()
 
 	warp := make(chan time.Time)
```
aaithal: Is `warp` used/effective anymore? Can we get rid of that as well (here and in other methods)?

agent/engine/docker_client_test.go
```diff
@@ -479,21 +468,17 @@ func TestStopContainerTimeout(t *testing.T) {
...
-	wait := &sync.WaitGroup{}
-	wait.Add(1)
 	mockDocker.EXPECT().StopContainerWithContext("id", uint(client.config.DockerStopTimeout/time.Second), gomock.Any()).Do(func(x, y, z interface{}) {
```
aaithal: I think this needs `.MaxTimes(1)` as well. So do all other `EXPECT()`, where we are actually letting these timeouts time out (wherever `xContainerShortTimeout` is used).

agent/engine/docker_client_test.go
```diff
@@ -363,28 +362,19 @@ func TestCreateContainerTimeout(t *testing.T) {
...
 	config := docker.CreateContainerOptions{Config: &docker.Config{Memory: 100}, Name: "containerName"}
 	mockDocker.EXPECT().CreateContainer(gomock.Any()).Do(func(x interface{}) {
 		warp <- time.Now()
-		wait.Wait()
-		// Don't return, verify timeout happens
-		// TODO remove the MaxTimes by cancel the context passed to CreateContainer
-		// when issue #1212 is resolved
 	}).MaxTimes(1)
-	metadata := client.CreateContainer(config.Config, nil, config.Name, xContainerShortTimeout)
+	metadata := client.CreateContainer(context.TODO(), config.Config, nil, config.Name, xContainerShortTimeout)
```
aaithal: In all of the tests, where you have passed `context.TODO()` directly, please modify those to be something like this:

```golang
ctx, cancel := context.WithCancel(context.TODO())
defer cancel()

// pass ctx
```

Otherwise, we'll have go routine leaks because of uncancelled context in edge cases.

agent/engine/docker_image_manager_integ_test.go
```diff
@@ -667,19 +668,19 @@ func verifyImagesAreNotRemoved(imageManager *dockerImageManager, imageIDs ...str
 }
 
 func cleanupImagesHappy(imageManager *dockerImageManager) {
-	imageManager.client.RemoveContainer("test1", removeContainerTimeout)
-	imageManager.client.RemoveContainer("test2", removeContainerTimeout)
-	imageManager.client.RemoveContainer("test3", removeContainerTimeout)
-	imageManager.client.RemoveImage(test1Image1Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test1Image2Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test1Image3Name, imageRemovalTimeout)
+	imageManager.client.RemoveContainer(context.TODO(), "test1", removeContainerTimeout)
```
aaithal: Just create context once as per my earlier comments and use it everywhere.

agent/engine/docker_image_manager_integ_test.go
```diff
@@ -667,19 +668,19 @@ func verifyImagesAreNotRemoved(imageManager *dockerImageManager, imageIDs ...str
...
 }
 
 func cleanupImagesThreshold(imageManager *dockerImageManager) {
-	imageManager.client.RemoveContainer("test1", removeContainerTimeout)
-	imageManager.client.RemoveContainer("test2", removeContainerTimeout)
-	imageManager.client.RemoveContainer("test3", removeContainerTimeout)
-	imageManager.client.RemoveImage(test2Image1Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test2Image2Name, imageRemovalTimeout)
-	imageManager.client.RemoveImage(test2Image3Name, imageRemovalTimeout)
+	imageManager.client.RemoveContainer(context.TODO(), "test1", removeContainerTimeout)
```
aaithal: Just create context once as per my earlier comments and use it everywhere.

agent/engine/docker_image_manager_test.go
```diff
@@ -776,8 +776,8 @@ func TestImageCleanupCannotRemoveImage(t *testing.T) {
 	imageState.PulledAt = time.Now().AddDate(0, -2, 0)
 	imageState.LastUsedAt = time.Now().AddDate(0, -2, 0)
 
-	client.EXPECT().RemoveImage(container.Image, removeImageTimeout).Return(errors.New("error removing image")).AnyTimes()
-	imageManager.removeUnusedImages()
+	client.EXPECT().RemoveImage(gomock.Any(), container.Image, removeImageTimeout).Return(errors.New("error removing image")).AnyTimes()
+	imageManager.removeUnusedImages(context.TODO())
```
aaithal: Same comment as in the other file about using cancellable context everywhere in this file

agent/eni/pause/pause_linux_test.go
```diff
@@ -52,7 +53,7 @@ func TestLoadFromFileWithReaderError(t *testing.T) {
 	mockfs := mock_os.NewMockFileSystem(ctrl)
 	mockfs.EXPECT().Open(pauseTarballPath).Return(nil, errors.New("Dummy Reader Error"))
 
-	err = loadFromFile(pauseTarballPath, client, mockfs)
+	err = loadFromFile(context.TODO(), pauseTarballPath, client, mockfs)
```
aaithal: Same comment as in the other file about using cancellable context here

agent/stats/engine_integ_test.go
```diff
@@ -86,7 +87,7 @@ func TestStatsEngineWithExistingContainersWithoutHealth(t *testing.T) {
 
 	// Simulate container start prior to listener initialization.
 	time.Sleep(checkPointSleep)
-	err = engine.MustInit(taskEngine, defaultCluster, defaultContainerInstance)
+	err = engine.MustInit(context.TODO(), taskEngine, defaultCluster, defaultContainerInstance)
```
aaithal: Same comment as in the other file about using cancellable context here

agent/stats/engine_test.go
```diff
@@ -52,6 +52,7 @@ func TestStatsEngineAddRemoveContainers(t *testing.T) {
 	mockDockerClient.EXPECT().Stats(gomock.Any(), gomock.Any()).Return(mockStatsChannel, nil).AnyTimes()
 
 	engine := NewDockerStatsEngine(&cfg, nil, eventStream("TestStatsEngineAddRemoveContainers"))
+	engine.ctx = context.TODO()
```
aaithal: Same comment as in the other file about using cancellable context here

agent/engine/docker_client.go
haikuoliu: This is the one that I added, we have talked about changing the timeout parameter into some kind of error message, and add a TODO for this purpose.

aaithal: Right, that was when timeout was only being used for generating error messages. But, now its actually used for creating the context. So, that's not needed anymore

haikuoliu: Oh I see, you are right!

agent/engine/docker_task_engine.go
haikuoliu: OK

agent/engine/docker_client_test.go
haikuoliu: It's used to make the go routine not stop so that we don't have to handle the logic after the mocked function return. I will remove the warp add return value for the mocked function instead.

agent/engine/docker_client_test.go
haikuoliu: OK, added `.MaxTimes(1)` to all timeout tests in docker client.

agent/engine/docker_client_test.go
haikuoliu: OK, sure.

agent/engine/docker_image_manager_integ_test.go
haikuoliu: OK

agent/engine/docker_image_manager_integ_test.go
haikuoliu: OK

haikuoliu:
Update: addressed all comments about the comments and tests.

agent/engine/docker_image_manager_test.go
```diff
@@ -71,7 +71,9 @@ func TestImagePullRemoveDeadlock(t *testing.T) {
 	}()
 
 	go func() {
-		imageManager.(*dockerImageManager).removeUnusedImages()
+		ctx, cancel := context.WithCancel(context.TODO())
```
aaithal: can you move the context creation and `defer cancel()` methods outside of the go routine?

agent/engine/docker_image_manager_test.go
```diff
@@ -1055,7 +1073,9 @@ func TestConcurrentRemoveUnusedImages(t *testing.T) {
 	for i := 0; i < numRoutines; i++ {
 		go func() {
 			<-ok
-			imageManager.removeUnusedImages()
+			ctx, cancel := context.WithCancel(context.TODO())
```
aaithal: can you move the context creation and `defer cancel()` methods outside of the `for` loop?

agent/stats/engine_integ_test.go
```diff
@@ -143,7 +146,7 @@ func TestStatsEngineWithNewContainersWithoutHealth(t *testing.T) {
 		},
 		testTask)
 
-	err = engine.MustInit(taskEngine, defaultCluster, defaultContainerInstance)
+	err = engine.MustInit(context.TODO(), taskEngine, defaultCluster, defaultContainerInstance)
```
aaithal: You should use a cancellable context here similar to other test files (applies throughout this file)

agent/engine/docker_image_manager_test.go
haikuoliu: OK

agent/engine/docker_image_manager_test.go
haikuoliu: OK, here only one context should created and used.

agent/stats/engine_integ_test.go
haikuoliu: ok, there are too many context creations in the test, I missed some, I will double check.

haikuoliu:
Just pushed the commit, thank you guys!👍

PR #1332, opened 2018-04-05
Progress task resources

sharanyad:
### Summary
This PR addresses transition of task resources for creation and deletion. 

### Implementation details
Resource transition follows similar logic as today's container transition. 

Some key implementation changes: 
- `progressTask` (earlier called `progressContainers`) will now wait for at least one of either container or resource transition to complete
- If resource creation fails, the desired status of task is marked as STOPPED.  
- No transition function is called for a resource's terminal status (resource removed/cleaned). Instead resources will be cleaned up in `deleteTask` operation. This ensures that we do not block task stop (waiting for resources to be cleaned). 

TODO
- [x] Implement `String()` for `ResourceStatus` 
- [x] Add unit tests for new methods 

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -866,36 +866,55 @@ func (task *Task) UpdateStatus() bool {
 
 // UpdateDesiredStatus sets the known status of the task
 func (task *Task) UpdateDesiredStatus() {
-	task.updateTaskDesiredStatus()
-	task.updateContainerDesiredStatus()
+	task.lock.Lock()
+	defer task.lock.Unlock()
+	task.updateTaskDesiredStatusUnsafe()
+	task.updateContainerDesiredStatusUnsafe(task.DesiredStatusUnsafe)
+	task.updateResourceDesiredStatusUnsafe(task.DesiredStatusUnsafe)
 }
 
-// updateTaskDesiredStatus determines what status the task should properly be at based on the containers' statuses
+// updateTaskDesiredStatusUnsafe determines what status the task should properly be at based on the containers' statuses
 // Invariant: task desired status must be stopped if any essential container is stopped
-func (task *Task) updateTaskDesiredStatus() {
-	seelog.Debugf("Updating task: [%s]", task.String())
+func (task *Task) updateTaskDesiredStatusUnsafe() {
+	seelog.Debugf("Updating task: [%s]", task.stringUnsafe())
 
 	// A task's desired status is stopped if any essential container is stopped
 	// Otherwise, the task's desired status is unchanged (typically running, but no need to change)
 	for _, cont := range task.Containers {
 		if cont.Essential && (cont.KnownTerminal() || cont.DesiredTerminal()) {
 			seelog.Debugf("Updating task desired status to stopped because of container: [%s]; task: [%s]",
-				cont.Name, task.String())
-			task.SetDesiredStatus(TaskStopped)
+				cont.Name, task.stringUnsafe())
+			task.DesiredStatusUnsafe = TaskStopped
 		}
 	}
 }
 
-// updateContainerDesiredStatus sets all container's desired status's to the
+// updateContainerDesiredStatusUnsafe sets all container's desired status's to the
 // task's desired status
 // Invariant: container desired status is <= task desired status converted to container status
 // Note: task desired status and container desired status is typically only RUNNING or STOPPED
-func (task *Task) updateContainerDesiredStatus() {
+func (task *Task) updateContainerDesiredStatusUnsafe(taskDesiredStatus TaskStatus) {
 	for _, c := range task.Containers {
-		taskDesiredStatus := task.GetDesiredStatus()
 		taskDesiredStatusToContainerStatus := taskDesiredStatus.ContainerStatus(c.GetSteadyStateStatus())
 		if c.GetDesiredStatus() < taskDesiredStatusToContainerStatus {
 			c.SetDesiredStatus(taskDesiredStatusToContainerStatus)
+		}
+	}
+}
+
+// updateResourceDesiredStatusUnsafe sets all resources' desired status depending on the
+// task's desired status
+// TODO: Create a mapping of resource status to the corresponding task status and use it here
+func (task *Task) updateResourceDesiredStatusUnsafe(taskDesiredStatus TaskStatus) {
+	for _, r := range task.Resources {
+		if taskDesiredStatus == TaskRunning {
+			if r.GetDesiredStatus() < r.SteadyState() {
+				r.SetDesiredStatus(r.SteadyState())
+			}
+		} else {
+			if r.GetDesiredStatus() < r.TerminalStatus() {
+				r.SetDesiredStatus(r.TerminalStatus())
+			}
 		}
 	}
 }
@@ -1094,7 +1113,11 @@ func (task *Task) GetExecutionStoppedAt() time.Time {
 func (task *Task) String() string {
 	task.lock.Lock()
 	defer task.lock.Unlock()
+	return task.stringUnsafe()
+}
 
+// stringUnsafe returns a human readable string representation of this object
+func (task *Task) stringUnsafe() string {
 	res := fmt.Sprintf("%s:%s %s, TaskStatus: (%s->%s)",
 		task.Family, task.Version, task.Arn,
 		task.KnownStatusUnsafe.String(), task.DesiredStatusUnsafe.String())
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -266,6 +266,14 @@ func (engine *DockerTaskEngine) Disable() {
 	engine.processTasks.Lock()
 }
 
+// isTaskManaged checks if task for the corresponding arn is present
+func (engine *DockerTaskEngine) isTaskManaged(arn string) bool {
+	engine.processTasks.RLock()
+	defer engine.processTasks.RUnlock()
+	_, ok := engine.managedTasks[arn]
+	return ok
+}
+
 // synchronizeState explicitly goes through each docker container stored in
 // "state" and updates its KnownStatus appropriately, as well as queueing up
 // events to push upstream.
@@ -458,6 +466,17 @@ func (engine *DockerTaskEngine) deleteTask(task *api.Task) {
 		if err != nil {
 			seelog.Warnf("Task engine [%s]: unable to cleanup platform resources: %v",
 				task.Arn, err)
+		}
+	}
+
+	for _, resource := range task.Resources {
+		err := resource.Cleanup()
+		if err != nil {
+			seelog.Warnf("Task engine [%s]: unable to cleanup resource %s: %v",
+				task.Arn, resource.GetName(), err)
+		} else {
+			seelog.Debugf("Task engine [%s]: resource %s cleanup complete", task.Arn,
+				resource.GetName())
 		}
 	}
 
@@ -1066,7 +1085,7 @@ func (engine *DockerTaskEngine) updateTaskUnsafe(task *api.Task, update *api.Tas
 }
 
 // transitionContainer calls applyContainerState, and then notifies the managed
-// task of the change. transitionContainer is called by progressContainers and
+// task of the change. transitionContainer is called by progressTask and
 // by handleStoppedToRunningContainerTransition.
 func (engine *DockerTaskEngine) transitionContainer(task *api.Task, container *api.Container, to api.ContainerStatus) {
 	// Let docker events operate async so that we can continue to handle ACS / other requests
--- a/agent/engine/docker_task_engine_unix_test.go
+++ b/agent/engine/docker_task_engine_unix_test.go
@@ -17,16 +17,28 @@ package engine
 import (
 	"context"
 	"errors"
+	"fmt"
+	"strings"
+	"sync"
 	"testing"
+	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
 	"github.com/aws/amazon-ecs-agent/agent/emptyvolume"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate/mocks"
+	"github.com/aws/amazon-ecs-agent/agent/engine/testdata"
+	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup/mock_control"
 	"github.com/aws/amazon-ecs-agent/agent/resources/mock_resources"
 	"github.com/aws/amazon-ecs-agent/agent/statemanager/mocks"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/cgroup"
+	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper/mocks"
+	"github.com/aws/aws-sdk-go/aws"
+	docker "github.com/fsouza/go-dockerclient"
 	"github.com/golang/mock/gomock"
+	specs "github.com/opencontainers/runtime-spec/specs-go"
 
 	"github.com/stretchr/testify/assert"
 )
@@ -38,6 +50,7 @@ const (
 	// isParallelPullCompatible is invoked during engine intialization
 	// on linux. Docker client's Version() call needs to be mocked
 	dockerVersionCheckDuringInit = true
+	cgroupMountPath              = "/sys/fs/cgroup"
 )
 
 func TestPullEmptyVolumeImage(t *testing.T) {
@@ -75,7 +88,7 @@ func TestDeleteTask(t *testing.T) {
 		},
 	}
 
-	cfg := &defaultConfig
+	cfg := defaultConfig
 	cfg.TaskCPUMemLimit = config.ExplicitlyEnabled
 	mockState := mock_dockerstate.NewMockTaskEngineState(ctrl)
 	mockSaver := mock_statemanager.NewMockStateManager(ctrl)
@@ -83,7 +96,7 @@ func TestDeleteTask(t *testing.T) {
 	taskEngine := &DockerTaskEngine{
 		state:    mockState,
 		saver:    mockSaver,
-		cfg:      &defaultConfig,
+		cfg:      &cfg,
 		resource: mockResource,
 	}
 
@@ -114,4 +127,125 @@ func TestEngineDisableConcurrentPull(t *testing.T) {
 	dockerTaskEngine, _ := taskEngine.(*DockerTaskEngine)
 	assert.False(t, dockerTaskEngine.enableConcurrentPull,
 		"Task engine should not be able to perform concurrent pulling for version < 1.11.1")
+}
+
+// TestResourceContainerProgression tests the container progression based on a
+// resource dependency
+func TestResourceContainerProgression(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	ctrl, client, mockTime, taskEngine, _, imageManager, _ := mocks(t, ctx, &defaultConfig)
+	defer ctrl.Finish()
+
+	sleepTask := testdata.LoadTask("sleep5")
+	sleepContainer := sleepTask.Containers[0]
+	sleepContainer.TransitionDependenciesMap = make(map[api.ContainerStatus]api.TransitionDependencySet)
+	sleepContainer.BuildResourceDependency("cgroup", taskresource.ResourceCreated, api.ContainerPulled)
+
+	mockControl := mock_cgroup.NewMockControl(ctrl)
+	mockIO := mock_ioutilwrapper.NewMockIOUtil(ctrl)
+	taskID, err := sleepTask.GetID()
+	assert.NoError(t, err)
+	cgroupMemoryPath := fmt.Sprintf("/sys/fs/cgroup/memory/ecs/%s/memory.use_hierarchy", taskID)
+	cgroupRoot := fmt.Sprintf("/ecs/%s", taskID)
+	cgroupResource := cgroup.NewCgroupResource(sleepTask.Arn, mockControl, cgroupRoot, cgroupMountPath, specs.LinuxResources{})
+	cgroupResource.SetIOUtil(mockIO)
+
+	sleepTask.Resources = []taskresource.TaskResource{cgroupResource}
+	sleepTask.Resources[0].SetDesiredStatus(taskresource.ResourceCreated)
+
+	eventStream := make(chan dockerapi.DockerContainerChangeEvent)
+	// containerEventsWG is used to force the test to wait until the container created and started
+	// events are processed
+	containerEventsWG := sync.WaitGroup{}
+	if dockerVersionCheckDuringInit {
+		client.EXPECT().Version()
+	}
+	client.EXPECT().ContainerEvents(gomock.Any()).Return(eventStream, nil)
+	gomock.InOrder(
+		// Ensure that the resource is created first
+		mockControl.EXPECT().Exists(gomock.Any()).Return(false),
+		mockControl.EXPECT().Create(gomock.Any()).Return(nil, nil),
+		mockIO.EXPECT().WriteFile(cgroupMemoryPath, gomock.Any(), gomock.Any()).Return(nil),
+		imageManager.EXPECT().AddAllImageStates(gomock.Any()).AnyTimes(),
+		client.EXPECT().PullImage(sleepContainer.Image, nil).Return(dockerapi.DockerContainerMetadata{}),
+		imageManager.EXPECT().RecordContainerReference(sleepContainer).Return(nil),
+		imageManager.EXPECT().GetImageStateFromImageName(sleepContainer.Image).Return(nil),
+		client.EXPECT().APIVersion().Return(defaultDockerClientAPIVersion, nil),
+		client.EXPECT().CreateContainer(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Do(
+			func(ctx interface{}, config *docker.Config, hostConfig *docker.HostConfig, containerName string, z time.Duration) {
+				assert.True(t, strings.Contains(containerName, sleepContainer.Name))
+				containerEventsWG.Add(1)
+				go func() {
+					eventStream <- createDockerEvent(api.ContainerCreated)
+					containerEventsWG.Done()
+				}()
+			}).Return(dockerapi.DockerContainerMetadata{DockerID: containerID + ":" + sleepContainer.Name}),
+		// Next, the sleep container is started
+		client.EXPECT().StartContainer(gomock.Any(), containerID+":"+sleepContainer.Name, defaultConfig.ContainerStartTimeout).Do(
+			func(ctx interface{}, id string, timeout time.Duration) {
+				containerEventsWG.Add(1)
+				go func() {
+					eventStream <- createDockerEvent(api.ContainerRunning)
+					containerEventsWG.Done()
+				}()
+			}).Return(dockerapi.DockerContainerMetadata{DockerID: containerID + ":" + sleepContainer.Name}),
+	)
+
+	addTaskToEngine(t, ctx, taskEngine, sleepTask, mockTime, containerEventsWG)
+
+	cleanup := make(chan time.Time, 1)
+	mockTime.EXPECT().After(gomock.Any()).Return(cleanup).AnyTimes()
+
+	// Simulate a container stop event from docker
+	eventStream <- dockerapi.DockerContainerChangeEvent{
+		Status: api.ContainerStopped,
+		DockerContainerMetadata: dockerapi.DockerContainerMetadata{
+			DockerID: containerID + ":" + sleepContainer.Name,
+			ExitCode: aws.Int(exitCode),
+		},
+	}
+	waitForStopEvents(t, taskEngine.StateChangeEvents(), true)
+}
+
+// TestResourceContainerProgressionFailure ensures that task moves to STOPPED when
+// resource creation fails
+func TestResourceContainerProgressionFailure(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	ctrl, client, mockTime, taskEngine, _, _, _ := mocks(t, ctx, &defaultConfig)
+	defer ctrl.Finish()
+	sleepTask := testdata.LoadTask("sleep5")
+	sleepContainer := sleepTask.Containers[0]
+	sleepContainer.TransitionDependenciesMap = make(map[api.ContainerStatus]api.TransitionDependencySet)
+	sleepContainer.BuildResourceDependency("cgroup", taskresource.ResourceCreated, api.ContainerPulled)
+
+	mockControl := mock_cgroup.NewMockControl(ctrl)
+	taskID, err := sleepTask.GetID()
+	assert.NoError(t, err)
+	cgroupRoot := fmt.Sprintf("/ecs/%s", taskID)
+	cgroupResource := cgroup.NewCgroupResource(sleepTask.Arn, mockControl, cgroupRoot, cgroupMountPath, specs.LinuxResources{})
+
+	sleepTask.Resources = []taskresource.TaskResource{cgroupResource}
+	sleepTask.Resources[0].SetDesiredStatus(taskresource.ResourceCreated)
+
+	eventStream := make(chan dockerapi.DockerContainerChangeEvent)
+	if dockerVersionCheckDuringInit {
+		client.EXPECT().Version()
+	}
+	client.EXPECT().ContainerEvents(gomock.Any()).Return(eventStream, nil)
+	gomock.InOrder(
+		// resource creation failure
+		mockControl.EXPECT().Exists(gomock.Any()).Return(false),
+		mockControl.EXPECT().Create(gomock.Any()).Return(nil, errors.New("cgroup create error")),
+	)
+	mockTime.EXPECT().Now().Return(time.Now()).AnyTimes()
+
+	err = taskEngine.Init(ctx)
+	assert.NoError(t, err)
+
+	taskEngine.AddTask(sleepTask)
+	cleanup := make(chan time.Time, 1)
+	mockTime.EXPECT().After(gomock.Any()).Return(cleanup).AnyTimes()
+	waitForStopEvents(t, taskEngine.StateChangeEvents(), true)
 }
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/resources"
 	"github.com/aws/amazon-ecs-agent/agent/statechange"
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	utilsync "github.com/aws/amazon-ecs-agent/agent/utils/sync"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
 
@@ -63,6 +64,13 @@ type dockerContainerChange struct {
 	event     dockerapi.DockerContainerChangeEvent
 }
 
+// resourceStateChange represents the required status change after resource transition
+type resourceStateChange struct {
+	resource  taskresource.TaskResource
+	nextState taskresource.ResourceStatus
+	err       error
+}
+
 type acsTransition struct {
 	seqnum        int64
 	desiredStatus api.TaskStatus
@@ -73,6 +81,17 @@ type containerTransition struct {
 	nextState      api.ContainerStatus
 	actionRequired bool
 	reason         error
+}
+
+// resourceTransition defines the struct for a resource to transition.
+type resourceTransition struct {
+	// nextState represents the next known status that the resource can move to
+	nextState taskresource.ResourceStatus
+	// status is the string value of nextState
+	status string
+	// actionRequired indicates if the transition function needs to be called for
+	// the transition to be complete
+	actionRequired bool
 }
 
 // managedTask is a type that is meant to manage the lifecycle of a task.
@@ -110,6 +129,7 @@ type managedTask struct {
 
 	acsMessages                chan acsTransition
 	dockerMessages             chan dockerContainerChange
+	resourceStateChangeEvent   chan resourceStateChange
 	stateChangeEvents          chan statechange.Event
 	containerChangeEventStream *eventstream.EventStream
 
@@ -142,11 +162,12 @@ type managedTask struct {
 func (engine *DockerTaskEngine) newManagedTask(task *api.Task) *managedTask {
 	ctx, cancel := context.WithCancel(engine.ctx)
 	t := &managedTask{
-		ctx:                        ctx,
-		cancel:                     cancel,
-		Task:                       task,
-		acsMessages:                make(chan acsTransition),
-		dockerMessages:             make(chan dockerContainerChange),
+		ctx:                      ctx,
+		cancel:                   cancel,
+		Task:                     task,
+		acsMessages:              make(chan acsTransition),
+		dockerMessages:           make(chan dockerContainerChange),
+		resourceStateChangeEvent: make(chan resourceStateChange),
 		engine:                     engine,
 		resource:                   engine.resource,
 		cfg:                        engine.cfg,
@@ -207,7 +228,7 @@ func (mtask *managedTask) overseeTask() {
 				}
 				seelog.Infof("Managed task [%s]: Cgroup resource set up for task complete", mtask.Arn)
 			}
-			mtask.progressContainers()
+			mtask.progressTask()
 		}
 
 		// If we reach this point, we've changed the task in some way.
@@ -329,6 +350,12 @@ func (mtask *managedTask) waitEvent(stopWaiting <-chan struct{}) bool {
 			mtask.Arn, dockerChange.container.Name, dockerChange.event.Status.String())
 		mtask.handleContainerChange(dockerChange)
 		return false
+	case resChange := <-mtask.resourceStateChangeEvent:
+		res := resChange.resource
+		seelog.Debugf("Managed task [%s]: got resource [%s] event: [%s]",
+			mtask.Arn, res.GetName(), res.StatusString(resChange.nextState))
+		mtask.handleResourceStateChange(resChange)
+		return false
 	case <-stopWaiting:
 		seelog.Debugf("Managed task [%s]: no longer waiting", mtask.Arn)
 		return true
@@ -423,6 +450,50 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 		mtask.Arn, container.Name, mtask.GetDesiredStatus().String())
 }
 
+// handleResourceStateChange attempts to update resource's known status depending on
+// the current status and errors during transition
+func (mtask *managedTask) handleResourceStateChange(resChange resourceStateChange) {
+	// locate the resource
+	res := resChange.resource
+	if !mtask.isResourceFound(res) {
+		seelog.Criticalf("Managed task [%s]: state error; invoked with another task's resource [%s]",
+			mtask.Arn, res.GetName())
+		return
+	}
+
+	status := resChange.nextState
+	err := resChange.err
+	currentKnownStatus := res.GetKnownStatus()
+
+	if status <= currentKnownStatus {
+		seelog.Infof("Managed task [%s]: redundant resource state change. %s to %s, but already %s",
+			mtask.Arn, res.GetName(), res.StatusString(status), res.StatusString(currentKnownStatus))
+		return
+	}
+
+	if err == nil {
+		res.SetKnownStatus(status)
+		mtask.engine.saver.Save()
+		return
+	}
+	seelog.Debugf("Managed task [%s]: error while transitioning resource %s to %s: %v",
+		mtask.Arn, res.GetName(), res.StatusString(status), err)
+	if status == res.SteadyState() {
+		seelog.Errorf("Managed task [%s]: error while creating resource %s, setting the task's desired status to STOPPED",
+			mtask.Arn, res.GetName())
+		mtask.SetDesiredStatus(api.TaskStopped)
+		mtask.engine.saver.Save()
+	}
+}
+
+func (mtask *managedTask) emitResourceChange(change resourceStateChange) {
+	if mtask.ctx.Err() != nil {
+		seelog.Infof("Managed task [%s]: unable to emit resource state change due to closed context: %v",
+			mtask.Arn, mtask.ctx.Err())
+	}
+	mtask.resourceStateChangeEvent <- change
+}
+
 func (mtask *managedTask) emitTaskEvent(task *api.Task, reason string) {
 	event, err := api.NewTaskStateChangeEvent(task, reason)
 	if err != nil {
@@ -478,6 +549,15 @@ func (mtask *managedTask) isContainerFound(container *api.Container) bool {
 		}
 	}
 	return found
+}
+
+func (mtask *managedTask) isResourceFound(res taskresource.TaskResource) bool {
+	for _, r := range mtask.Resources {
+		if res.GetName() == r.GetName() {
+			return true
+		}
+	}
+	return false
 }
 
 // releaseIPInIPAM releases the ip used by the task for awsvpc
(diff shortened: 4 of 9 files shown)


agent/taskresource/cgroup/cgroup.go
```diff
@@ -63,14 +65,24 @@ func NewCgroupResource(taskARN string,
...
 		ioutil:          ioutilwrapper.NewIOUtil(),
 	}
+	c.initializeResourceStatusToTransitionFunction()
+	return c
+}
+
+func (c *CgroupResource) initializeResourceStatusToTransitionFunction() {
+	resourceStatusToTransitionFunction := map[taskresource.ResourceStatus]func() error{
+		taskresource.ResourceStatus(CgroupCreated): c.Create,
+		taskresource.ResourceStatus(CgroupRemoved): c.Cleanup,
```
richardpen: So this is currently not useful, right? As we don't use the transition function for cleanup.

agent/taskresource/interface.go
```diff
@@ -41,6 +46,21 @@ type TaskResource interface {
 	Cleanup() error
 	// GetName returns the unique name of the resource
 	GetName() string
+	// DesiredTeminal returns true if remove is in terminal state
+	DesiredTerminal() bool
+	// IsCreated returns true if resource state is CREATED
```
richardpen: Lint: KnownCreated.

agent/taskresource/cgroup/cgroup.go
richardpen: So this entry isn't needed, as we don't use the transitonfunction to do cleanup, right?

agent/taskresource/cgroup/cgroup.go
```diff
@@ -97,12 +109,95 @@ func (c *CgroupResource) GetName() string {
 	return resourceName
 }
 
+// DesiredTerminal returns true if the cgroup's desired status is REMOVED
+func (c *CgroupResource) DesiredTerminal() bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.GetDesiredStatus() == taskresource.ResourceStatus(CgroupRemoved)
```
richardpen: I think you can just compare `c.desiredStatusUnsafe == CgroupRemoved` here? same below

agent/taskresource/cgroup/cgroup.go
```diff
@@ -97,12 +109,95 @@ func (c *CgroupResource) GetName() string {
...
+// KnownCreated returns true if the cgroup's known status is CREATED
+func (c *CgroupResource) KnownCreated() bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.GetKnownStatus() == taskresource.ResourceStatus(CgroupCreated)
+}
+
+// TerminalStatus returns the last transition state of cgroup
+func (c *CgroupResource) TerminalStatus() taskresource.ResourceStatus {
```
richardpen: This needs a lock

agent/taskresource/cgroup/cgroup.go
```diff
@@ -97,12 +109,95 @@ func (c *CgroupResource) GetName() string {
...
+	transitionFunc, ok := c.transitionFunctionMap()[nextState]
+	if !ok {
+		seelog.Criticalf("Cgroup Resource [%s]: unsupported desired state transition [%s]: %s",
+			c.taskARN, c.GetName(), nextState.String())
+		return false, nil
+	}
+	return true, transitionFunc()
+}
+
+func (c *CgroupResource) transitionFunctionMap() map[taskresource.ResourceStatus]func() error {
```
richardpen: Why you need this function to return the map, you can just use the `transitionFunctionMap` above?

agent/taskresource/cgroup/cgroup.go
```diff
@@ -97,12 +109,95 @@ func (c *CgroupResource) GetName() string {
...
 func (c *CgroupResource) SetKnownStatus(status taskresource.ResourceStatus) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
 	c.knownStatusUnsafe = status
+	c.updateAppliedStatus(status)
+}
+
+// updateAppliedStatusUnsafe updates the resource transitioning status
+func (c *CgroupResource) updateAppliedStatus(knownStatus taskresource.ResourceStatus) {
```
richardpen: Lint: updatexxUnsafe

agent/engine/task_manager.go
```diff
@@ -624,45 +674,61 @@ func (mtask *managedTask) handleContainerStoppedTransitionError(event DockerCont
...
 		func(container *api.Container, nextStatus api.ContainerStatus) {
 			mtask.engine.transitionContainer(mtask.Task, container, nextStatus)
 			transitionChange <- struct{}{}
-			transitionChangeContainer <- container.Name
+			transitionChangeEntity <- container.Name
 		})
 
-	if !anyCanTransition {
-		if !mtask.waitForExecutionCredentialsFromACS(reasons) {
+	if !anyContainerTransition {
```
richardpen: I think you are missing the case where all containers are waiting for the resource. You can either check the `anyResourceTransition` or add a function like `waitForExecutionCredentialsFromACS` to check whether the containers are waiting for resource.

agent/api/task.go
```diff
@@ -899,6 +900,24 @@ func (task *Task) updateContainerDesiredStatus() {
 	}
 }
 
+// updateResourceDesiredStatus sets all resources' desired status's depending on the
+// task's desired status
+func (task *Task) updateResourceDesiredStatus() {
+	for _, r := range task.Resources {
+		taskDesiredStatus := task.GetDesiredStatus()
+		if taskDesiredStatus == TaskRunning {
```
richardpen: Can you have a map that maps the task status to the resource status, so you don't rely on the actual status here?

agent/engine/docker_task_engine.go
```diff
@@ -1091,6 +1099,39 @@ func (engine *DockerTaskEngine) transitionFunctionMap() map[api.ContainerStatus]
...
+// task of the change. transitionResource is called by progressTask
+func (engine *DockerTaskEngine) transitionResource(task *api.Task, resource taskresource.TaskResource, to taskresource.ResourceStatus) {
+	err := engine.applyResourceState(task, resource, to)
+
+	engine.processTasks.RLock()
+	managedTask, ok := engine.managedTasks[task.Arn]
+	if ok {
+		managedTask.handleResourceChange(resource, to, err)
+	}
+	engine.processTasks.RUnlock()
```
richardpen: Can we release the lock earlier after reading the map?

agent/engine/task_manager.go
```diff
@@ -732,8 +798,42 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+			// At least one resource is able to be moved forwards, so we're not deadlocked
+			anyCanTransition = true
+			continue
+		}
+
+		// At least one resource is able to be move forwards, so we're not deadlocked
+		anyCanTransition = true
+
+		if !transition.actionRequired {
+			mtask.handleResourceChange(res, transition.nextState, nil)
```
richardpen: Can this be a go routine, as this is blocking the transition.

agent/engine/task_manager.go
```diff
@@ -802,6 +902,45 @@ func (mtask *managedTask) containerNextState(container *api.Container) *containe
...
 
+func (mtask *managedTask) resourceNextState(resource taskresource.TaskResource) *resourceTransition {
+	resKnownStatus := resource.GetKnownStatus()
+	resDesiredStatus := resource.GetDesiredStatus()
+
+	if resKnownStatus == resDesiredStatus {
+		seelog.Debugf("Managed task [%s]: resource [%s] at desired status: %s",
+			mtask.Arn, resource.GetName(), resDesiredStatus.String())
+		return &resourceTransition{
+			nextState:      taskresource.ResourceStatus(0), // set to none
```
adnxn: nit: should we create types for these resource status? also i thought we already had done that hm

agent/taskresource/cgroup/cgroup.go
sharanyad: This method just returns `CgroupRemoved`. Hence not using any lock.

agent/api/task.go
sharanyad: To use `TaskStatus` in `taskresource` package(for the mapping), we need to import `api` package. This is causing import cycle since `api` already imports `taskresource`. The plan was to move the status types from `api` into its own package. This is requiring changes in a lot of packages since `TaskStatus` and `ContainerStatus` and their types are used everywhere. I have added a `TODO` in this method, which shall be addressed in a different PR. Does that sound good?

agent/engine/task_manager.go
sharanyad: Each resource has its own status type which is of the type `taskresource.ResourceStatus`. Hence using a generic status type might not solve the problem here.

agent/engine/docker_task_engine.go
```diff
@@ -1091,6 +1099,39 @@ func (engine *DockerTaskEngine) transitionFunctionMap() map[api.ContainerStatus]
...
+// function defined in the transitionFunctionMap for the state
+func (engine *DockerTaskEngine) applyResourceState(task *api.Task, resource taskresource.TaskResource, nextState taskresource.ResourceStatus) error {
+	ok, err := resource.TransitionFunction(nextState)
+	if !ok {
+		seelog.Criticalf("Task engine [%s]: unsupported desired state transition for resource [%s]: %s",
+			task.Arn, resource.GetName(), nextState.String())
+		return errors.New("Impossible to transition to the resource state")
+	}
+	if err != nil {
+		seelog.Infof("Task engine [%s]: error transitioning resource [%s] to [%s]: %v",
```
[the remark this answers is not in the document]
sharanyad: We log this error as `Criticalf` where its returned. So I think that should be good.

agent/api/task.go
```diff
@@ -899,6 +900,25 @@ func (task *Task) updateContainerDesiredStatus() {
 	}
 }
 
+// updateResourceDesiredStatus sets all resources' desired status's depending on the
+// task's desired status
+// TODO: Create a mapping of resource status to the corresponding task status and use it here
+func (task *Task) updateResourceDesiredStatus() {
+	for _, r := range task.Resources {
+		taskDesiredStatus := task.GetDesiredStatus()
```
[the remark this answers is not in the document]
sharanyad: The task's desired status can change anytime. That's the reason why we check it every time here.

agent/api/task.go
```diff
@@ -899,6 +900,25 @@ func (task *Task) updateContainerDesiredStatus() {
...
+		if taskDesiredStatus == TaskRunning {
+			if r.GetDesiredStatus() < r.SteadyState() {
+				r.SetDesiredStatus(r.SteadyState())
+			}
+		} else {
+			// Task desired status is stopped
```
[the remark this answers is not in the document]
sharanyad: `task.UpdateDesiredStatus()` does not always guarantee that task's desired status is moved to stopped. Here, we are setting the resource's desired status depending on the task's status (i.e running/stopped)

agent/engine/task_manager.go
```diff
@@ -624,45 +675,61 @@ func (mtask *managedTask) handleContainerStoppedTransitionError(event DockerCont
...
 	// get GC'd eventually
-	transitionChange := make(chan struct{}, len(mtask.Containers))
-	transitionChangeContainer := make(chan string, len(mtask.Containers))
+	transitionChange := make(chan struct{}, len(mtask.Containers)+len(mtask.Resources))
+	transitionChangeEntity := make(chan string, len(mtask.Containers)+len(mtask.Resources))
 
-	anyCanTransition, transitions, reasons := mtask.startContainerTransitions(
+	anyResourceTransition, resTransitions := mtask.startResourceTransitions(
+		func(resource taskresource.TaskResource, nextStatus taskresource.ResourceStatus) {
+			mtask.engine.transitionResource(mtask.Task, resource, nextStatus)
```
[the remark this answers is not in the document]
sharanyad: I shall move the method to managed task.

agent/engine/task_manager.go
```diff
@@ -802,6 +903,45 @@ func (mtask *managedTask) containerNextState(container *api.Container) *containe
 	}
 }
 
+func (mtask *managedTask) resourceNextState(resource taskresource.TaskResource) *resourceTransition {
+	resKnownStatus := resource.GetKnownStatus()
+	resDesiredStatus := resource.GetDesiredStatus()
+
+	if resKnownStatus == resDesiredStatus {
```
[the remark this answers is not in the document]
sharanyad: These conditions have different debug log messages, hence they are kept differently. We can have a `bool` variable and then return the same `resourceTransition` depending on the conditions.

agent/taskresource/cgroup/cgroup.go
richardpen: Then you are acquiring the lock twice, which may cause deadlock.

agent/engine/task_manager.go
```diff
@@ -818,20 +958,20 @@ func (mtask *managedTask) onContainersUnableToTransitionState() {
 	}
 }
 
-func (mtask *managedTask) waitForContainerTransition(transitions map[string]api.ContainerStatus,
+func (mtask *managedTask) waitForTransition(transitions map[string]string,
 	transition <-chan struct{},
-	transitionChangeContainer <-chan string) {
+	transitionChangeEntity <-chan string) {
```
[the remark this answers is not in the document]
sharanyad: The `transition` channel is used to indicate that either a container/resource transition is complete. It is passed to `waitEvent` method and used here - https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/engine/task_manager.go#L348
To know which container/resource transition that is, we use the `transitionChangeEntity`. It is being used here - 
https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/engine/task_manager.go#L831 (this is now renamed to `transitionChangeEntity`)

agent/taskresource/cgroup/cgroup.go
sharanyad: will change this to `c.desiredStatusUnsafe`

agent/api/task.go
```diff
@@ -867,6 +867,7 @@ func (task *Task) UpdateStatus() bool {
 func (task *Task) UpdateDesiredStatus() {
 	task.updateTaskDesiredStatus()
 	task.updateContainerDesiredStatus()
+	task.updateResourceDesiredStatus()
```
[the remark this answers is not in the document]
sharanyad: I agree this is not the right behavior. We are probably looking at an already existing issue here. 
We should probably hold the task lock for all the three update*Status() here and have a new `setDesiredStatusUnsafe` so that we can set task's desired status safely. Does that sound good?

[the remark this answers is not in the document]
richardpen: that should work, but I'm not sure if `setDesiredStatusUnsafe` is required, you can just reference it in these updates methods, since it's already protected by a lock.

agent/api/task.go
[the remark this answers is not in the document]
aaithal: Right, because of that, you have inconsistent transitions potentially being applied for different resources. It would be lot more straight forward if during one invocation of `updateResourceDesiredStatus()`, you applied one transition. Because, subsequent transitions should be caught in subsequent invocations of `updateResourceDesiredStatus`

agent/api/task.go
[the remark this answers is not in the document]
aaithal: Can you please get rid of the `"// Task desired status is stopped"` comment then?

agent/engine/task_manager.go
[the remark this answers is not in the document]
aaithal: Does logging those things really matter? Can't we just say `>=` and be done with it?

agent/engine/task_manager.go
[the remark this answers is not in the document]
aaithal: Oh, i see it now, thanks.

agent/engine/task_manager.go
```diff
@@ -732,8 +799,42 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
 }
 
+// startResourceTransitions steps through each resource in the task and calls
+// the passed transition function when a transition should occur
+func (mtask *managedTask) startResourceTransitions(transitionFunc resourceTransitionFunc) (bool, map[string]taskresource.ResourceStatus) {
+	anyCanTransition := false
+	transitions := make(map[string]taskresource.ResourceStatus)
+	for _, res := range mtask.Resources {
+		transition := mtask.resourceNextState(res)
+		if transition.reason != nil {
```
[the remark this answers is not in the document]
sharanyad: The `reason` here indicates the error why the resource cannot be transitioned. 
Not all resource transitions require 'action'. This is indicated by the `actionRequired` bool. Hence the two different variables.

agent/engine/task_manager.go
```diff
@@ -732,8 +799,42 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+		if transition.actionRequired && !res.SetAppliedStatus(transition.nextState) {
+			// At least one resource is able to be moved forwards, so we're not deadlocked
+			anyCanTransition = true
+			continue
+		}
+
+		// At least one resource is able to be move forwards, so we're not deadlocked
+		anyCanTransition = true
+
+		if !transition.actionRequired {
```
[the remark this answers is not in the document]
sharanyad: When no action is required for the transition, we just set the known status without calling any transition function. That is also done in `handleResourceChange`.

agent/engine/task_manager.go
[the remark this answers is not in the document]
aaithal: But, it's not being used anywhere. Right now, you can use them interchangeably. I'm not sure what value `reason` is providing right now.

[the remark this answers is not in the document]
sharanyad: These are the possible values of the variables and how they are used now:
`reason`:
`nil`- resource's known status can change(this can happen by applying a transition function and changing the known status OR just changing the known status without applying any transition).
not `nil` - resource's known status need not change and the reason why. 

`actionRequired`:
`true` - known status can be changed (given reason is `nil`) by applying a transition function. 
`false` - known status can be changed (given reason is `nil`) without applying a transition function.

agent/engine/task_manager.go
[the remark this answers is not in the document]
aaithal: i see. thanks for the explanation. Can you please add code comments explaining the same here as well?

agent/api/task.go
```diff
@@ -865,22 +865,25 @@ func (task *Task) UpdateStatus() bool {
...
+	task.lock.Lock()
+	defer task.lock.Unlock()
 	task.updateTaskDesiredStatus()
 	task.updateContainerDesiredStatus()
+	task.updateResourceDesiredStatus()
 }
 
 // updateTaskDesiredStatus determines what status the task should properly be at based on the containers' statuses
 // Invariant: task desired status must be stopped if any essential container is stopped
 func (task *Task) updateTaskDesiredStatus() {
```
richardpen: Please also update the method name to indicate it's not protected by a lock. Same for the others.

agent/api/task.go
```diff
@@ -1093,7 +1114,11 @@ func (task *Task) GetExecutionStoppedAt() time.Time {
 func (task *Task) String() string {
 	task.lock.Lock()
 	defer task.lock.Unlock()
+	return task.stringUnsafe()
+}
 
+// String returns a human readable string representation of this object
```
richardpen: nit: stringUnsafe

agent/engine/docker_task_engine.go
```diff
@@ -447,6 +447,14 @@ func (engine *DockerTaskEngine) deleteTask(task *api.Task, handleCleanupDone cha
 		}
 	}
 
+	for _, resource := range task.Resources {
```
richardpen: Can you add a log to indicate the resource cleanup is done?

agent/engine/task_manager.go
```diff
@@ -480,6 +523,17 @@ func (mtask *managedTask) isContainerFound(container *api.Container) bool {
 	return found
 }
 
+func (mtask *managedTask) isResourceFound(res taskresource.TaskResource) bool {
+	found := false
+	for _, r := range mtask.Resources {
```
richardpen: Is the `Resource` filed in `managedTask` still used?

agent/engine/task_manager.go
```diff
@@ -480,6 +523,17 @@ func (mtask *managedTask) isContainerFound(container *api.Container) bool {
...
+		if res == r {
```
richardpen: Will this work, as the `res.knownStatusUnsafe` can be different with the `r.knownStatusUnsafe`, right? The resource may need another method to compare.

agent/engine/task_manager.go
```diff
@@ -439,6 +448,40 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
+		seelog.Infof("Managed task [%s]: redundant resource state change. %s to %s, but already %s",
+			mtask.Arn, res.GetName(), res.StatusString(status), res.StatusString(currentKnownStatus))
+		return
+	}
+
+	if err == nil {
+		res.SetKnownStatus(status)
+		return
+	}
+	seelog.Debugf("Managed task [%s]: Error while transitioning resource %s: %v",
```
richardpen: Please also log the status here.

agent/engine/task_manager.go
```diff
@@ -624,43 +678,60 @@ func (mtask *managedTask) handleContainerStoppedTransitionError(event DockerCont
...
 	// these and if we exit early transitions can exit the goroutine and it'll
 	// get GC'd eventually
-	transitionChange := make(chan struct{}, len(mtask.Containers))
-	transitionChangeContainer := make(chan string, len(mtask.Containers))
+	transitionChange := make(chan struct{}, len(mtask.Containers)+len(mtask.Resources))
+	transitionChangeEntity := make(chan string, len(mtask.Containers)+len(mtask.Resources))
+
+	// anyResourceTransition is set to true when transition function needs to be called or
+	// known status can be changed
+	anyResourceTransition, resTransitions := mtask.startResourceTransitions(
```
richardpen: Please make a comment that the `startResourceTransitions` has to be called before `startContainerTransitions`.

agent/engine/task_manager.go
sharanyad: This check is done before the known status is changed in `handleResourceChange`. 
Nevertheless, comparing the resource names should be sufficient here.

agent/engine/task_manager.go
sharanyad: That shall be removed once we use `taskresource`'s `cgroup` for creation and deletion.

agent/api/task.go
```diff
@@ -889,16 +892,34 @@ func (task *Task) updateTaskDesiredStatus() {
 // task's desired status
 // Invariant: container desired status is <= task desired status converted to container status
 // Note: task desired status and container desired status is typically only RUNNING or STOPPED
-func (task *Task) updateContainerDesiredStatus() {
+func (task *Task) updateContainerDesiredStatusUnsafe() {
 	for _, c := range task.Containers {
-		taskDesiredStatus := task.GetDesiredStatus()
+		taskDesiredStatus := task.DesiredStatusUnsafe
```
aaithal: It seems like we can extract this line and line num 909 to the higher level method (`task.UpdateDesiredStatus()`) and pass `taskDesiredStatus` as a parameter to both `updateContainerDesiredStatusUnsafe` and `updateResourceDesiredStatusUnsafe` so that the caller doesn't need to worry about taking any locks:

```golang
 func (task *Task) UpdateDesiredStatus() {
 task.updateTaskDesiredStatus()
 taskDesiredStatus := task.GetDesiredStatus()
 task.updateContainerDesiredStatus(taskDesiredStatus)
 task. updateResourceDesiredStatus(taskDesiredStatus)
}

func (task *Task) updateContainerDesiredStatus(taskDesiredStatus api.TaskStatus) {
 ...
}

func (task *Task) updateResourceDesiredStatus(taskDesiredStatus api.TaskStatus) {
 ...
}

```

Unless there's a need to lock the task down when updating container and resource statuses, this seems like a much simpler approach.

agent/api/task.go
```diff
@@ -1093,7 +1114,11 @@ func (task *Task) GetExecutionStoppedAt() time.Time {
 func (task *Task) String() string {
 	task.lock.Lock()
 	defer task.lock.Unlock()
+	return task.stringUnsafe()
```
aaithal: If you agree with my previous comment, this is a redundant method.

agent/engine/task_manager.go
```diff
@@ -73,6 +74,14 @@ type containerTransition struct {
 	reason         error
 }
 
+// resourceTransition defines the struct for a resource to transition
+type resourceTransition struct {
+	nextState      taskresource.ResourceStatus
```
aaithal: please add documentation for what each of these fields are used for

agent/engine/task_manager.go
```diff
@@ -73,6 +74,14 @@ type containerTransition struct {
...
+	statusString   string
```
aaithal: just `status` should be a good enough name for this field imo.

agent/taskresource/cgroup/cgroup.go
```diff
@@ -44,143 +44,230 @@ var (
...
+func (cgroup *CgroupResource) GetKnownStatus() taskresource.ResourceStatus {
+	cgroup.lock.RLock()
+	defer cgroup.lock.RUnlock()
+
+	return cgroup.knownStatusUnsafe
+}
 
-	return c.knownStatusUnsafe
+// StatusString returns the string of the cgroup resource status
+func (cgroup *CgroupResource) StatusString(status taskresource.ResourceStatus) string {
```
aaithal: Why is this method a member of `CgroupResource` struct? Seems like it can be untethered from it.

agent/engine/task_manager.go
```diff
@@ -439,6 +448,39 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 		mtask.Arn, container.Name, mtask.GetDesiredStatus().String())
 }
 
+// handleResourceChange updates a resource's known status
+func (mtask *managedTask) handleResourceChange(res taskresource.TaskResource,
```
aaithal: `updateResourceKnownStatus()` seems like a better name for this based on the description.

agent/engine/task_manager.go
```diff
@@ -480,6 +522,17 @@ func (mtask *managedTask) isContainerFound(container *api.Container) bool {
 	return found
 }
 
+func (mtask *managedTask) isResourceFound(res taskresource.TaskResource) bool {
```
aaithal: This is much simpler:

```golang
for _, r := range mtask.Resources {
 if res.GetName() == r.GetName() {
  return true
 }
 return false
}
```

agent/engine/task_manager.go
```diff
@@ -439,6 +448,39 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 		mtask.Arn, container.Name, mtask.GetDesiredStatus().String())
 }
 
+// handleResourceChange updates a resource's known status
+func (mtask *managedTask) handleResourceChange(res taskresource.TaskResource,
+	status taskresource.ResourceStatus,
+	err error) {
+	// locate the resource
+	found := mtask.isResourceFound(res)
+	if !found {
```
aaithal: why not just do `if !mtask.isResourceFound(res)` here?

agent/engine/task_manager.go
```diff
@@ -439,6 +448,39 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
+		seelog.Criticalf("Managed task [%s]: state error; invoked with another task's resource [%s]!",
```
aaithal: This should be `Errorf()`

agent/engine/task_manager.go
```diff
@@ -439,6 +448,39 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
+	}
+
+	if err == nil {
+		res.SetKnownStatus(status)
+		return
+	}
+	seelog.Debugf("Managed task [%s]: error while transitioning resource %s to %s: %v",
+		mtask.Arn, res.GetName(), res.StatusString(status), err)
+	if status == res.SteadyState() {
+		seelog.Criticalf("Managed task [%s]: error while creating resource %s, setting the task's desired status to STOPPED",
```
aaithal: This should be `Errorf()`.

agent/engine/task_manager.go
```diff
@@ -439,6 +448,39 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
+		seelog.Infof("Managed task [%s]: redundant resource state change. %s to %s, but already %s",
+			mtask.Arn, res.GetName(), res.StatusString(status), res.StatusString(currentKnownStatus))
+		return
+	}
+
+	if err == nil {
+		res.SetKnownStatus(status)
+		return
+	}
+	seelog.Debugf("Managed task [%s]: error while transitioning resource %s to %s: %v",
```
aaithal: Just for my own understanding, what's the logic here? There's an error while transitioning the resource. But, you only set task's desired status to STOPPED is resource's status is equal to it's steady state? Why is that?

agent/engine/task_manager.go
```diff
@@ -732,8 +805,80 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+}
+
+// transitionResource calls applyResourceState, and then notifies the managed
+// task of the change. transitionResource is called by progressTask
+func (mtask *managedTask) transitionResource(resource taskresource.TaskResource,
+	to taskresource.ResourceStatus) {
+	engine := mtask.engine
+	err := mtask.applyResourceState(resource, to)
+
+	engine.processTasks.RLock()
```
aaithal: We should avoid using engine's internal fields as much as possible in managed task. Can you create a new method in engine called `getManagedTask(taskARN string)` and use that instead?

agent/engine/task_manager.go
```diff
@@ -732,8 +805,80 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+	ok, err := resource.ApplyTransition(nextState)
+	if !ok {
+		seelog.Criticalf("Task engine [%s]: unsupported desired state transition for resource [%s]: %s",
+			mtask.Arn, resource.GetName(), resource.StatusString(nextState))
+		return errors.New("impossible to transition to the resource state due to lack of transition function")
+	}
+	if err != nil {
+		seelog.Infof("Task engine [%s]: error transitioning resource [%s] to [%s]: %v",
+			mtask.Arn, resource.GetName(), resource.StatusString(nextState), err)
+	} else {
```
aaithal: just return`err` here, get rid of `else` and return `nil` at the end. Much easier to reason about and one less indentation level in the method.

agent/engine/task_manager.go
```diff
@@ -802,6 +947,24 @@ func (mtask *managedTask) containerNextState(container *api.Container) *containe
 	}
 }
 
+func (mtask *managedTask) resourceNextState(resource taskresource.TaskResource) *resourceTransition {
+	var nextState taskresource.ResourceStatus
```
aaithal: you really don't need this var. It's adding 3 more additional lines of code.

agent/taskresource/cgroup/cgroup.go
```diff
@@ -44,143 +44,230 @@ var (
...
+}
+
+// TerminalStatus returns the last transition state of cgroup
+func (cgroup *CgroupResource) TerminalStatus() taskresource.ResourceStatus {
+	return taskresource.ResourceStatus(CgroupRemoved)
+}
+
+// GetNextKnownStateProgression returns the state that the resource should
+// progress to based on its `KnownState`.
+func (cgroup *CgroupResource) GetNextKnownStateProgression() taskresource.ResourceStatus {
```
aaithal: This can just be `NextKnownState()`, can't it?

agent/engine/task_manager.go
```diff
@@ -732,8 +805,80 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+
+// applyResourceState moves the resource to the given state by calling the
+// function defined in the transitionFunctionMap for the state
+func (mtask *managedTask) applyResourceState(resource taskresource.TaskResource,
+	nextState taskresource.ResourceStatus) error {
+	ok, err := resource.ApplyTransition(nextState)
+	if !ok {
+		seelog.Criticalf("Task engine [%s]: unsupported desired state transition for resource [%s]: %s",
+			mtask.Arn, resource.GetName(), resource.StatusString(nextState))
+		return errors.New("impossible to transition to the resource state due to lack of transition function")
```
aaithal: as per golang's error string convention, this seems like a more appropriate error string(please make this change wherever applicable): `"managed task: resource transition impossible"`. There's context for error's origin before `:` and a description of the error after it.

agent/taskresource/cgroup/cgroup.go
```diff
@@ -44,143 +44,230 @@ var (
...
+	return cgroup.GetKnownStatus() + 1
+}
+
+// ApplyTransition calls the function required to move to the specified status
+func (cgroup *CgroupResource) ApplyTransition(nextState taskresource.ResourceStatus) (bool, error) {
+	transitionFunc, ok := cgroup.resourceStatusToTransitionFunction[nextState]
+	if !ok {
+		seelog.Criticalf("Cgroup Resource [%s]: unsupported desired state transition [%s]: %s",
```
aaithal: This shouldn't be a `Critical` log line either. `Error` seems like a more appropriate level

agent/taskresource/cgroup/cgroup.go
```diff
@@ -44,143 +44,230 @@ var (
...
+	cgroup.lock.Lock()
+	defer cgroup.lock.Unlock()
+
+	cgroup.knownStatusUnsafe = status
+	cgroup.updateAppliedStatusUnsafe(status)
+}
 
-	c.knownStatusUnsafe = status
+// updateAppliedStatusUnsafe updates the resource transitioning status
+func (cgroup *CgroupResource) updateAppliedStatusUnsafe(knownStatus taskresource.ResourceStatus) {
```
aaithal: What's the reasoning behind having this method? Why not just have this in `SetKnownStatus()`?

agent/taskresource/cgroup/cgroup_unsupported.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+)
+
+const (
+	resourceName = "cgroup"
+)
+
+// CgroupResource represents Cgroup resource
+type CgroupResource struct {
+	desiredStatusUnsafe taskresource.ResourceStatus
```
aaithal: Why do you need any members at all in this struct? Everything should be a no-op, right?

agent/taskresource/cgroup/cgroup_unsupported.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// GetNextKnownStateProgression returns the state that the resource should
+// progress to based on its `KnownState`.
+func (c *CgroupResource) GetNextKnownStateProgression() taskresource.ResourceStatus {
+	return c.GetKnownStatus() + 1
+}
+
+// ApplyTransition calls the function required to move to the specified status
+func (c *CgroupResource) ApplyTransition(nextState taskresource.ResourceStatus) (bool, error) {
```
aaithal: All of the methods that return an `error` in this struct should return `unsupported` error.

agent/api/task.go
sharanyad: The intention was to lock task status while updating container and resource statuses. In the suggestion, some method can still modify the status between `task.updateTaskDesiredStatus()` and `taskDesiredStatus := task.GetDesiredStatus()`.
We can have a lock for those two statements.

aaithal: > In the suggestion, some method can still modify the status between `task.updateTaskDesiredStatus()` and `taskDesiredStatus := task.GetDesiredStatus()`.

I think that's fine and something that should be considered normal. What's the concern if that happens?

agent/taskresource/cgroup/cgroup.go
sharanyad: Each resource has its own status type, which only the resource knows of. Since we use resource abstraction in `task_manager` and each status type has its own string implementation, we call `StatusString` to get the string value of the status. Let me know if you have better suggestions.

agent/engine/task_manager.go
sharanyad: once we move the task's desired status to STOPPED, we eventually move every container and resources' desired status to STOPPED [here](https://github.com/aws/amazon-ecs-agent/pull/1332/files#diff-e008975819a900a8494cf91881a8304aR872). I can add a comment.

agent/taskresource/cgroup/cgroup.go
aaithal: I see. Can't think of anything better right now. Will keep this in the back of the mind and let you know if I can think of something better.

agent/engine/task_manager.go
sharanyad: We are reusing `nextState` var in `statusString` field too, hence left it there.

agent/engine/task_manager.go
aaithal: Hmm, I still didn't understand why task's desired status should be conditionally set to STOPPED. Seems like irrespective of resource's status, task's desired status should be set to STOPPED.

agent/engine/task_manager.go
aaithal: Hmm, alright. You can still reduce 1 line by doing `nextState := resource.TerminalStatus()` in both of those places :)

agent/api/task.go
sharanyad: Just inconsistency (not changing statuses according to the task desired status that we want). I can't think of immediate consequences since the only status anything else might change it to is STOPPED. I'll go ahead and make the suggested change.

agent/taskresource/cgroup/cgroup.go
sharanyad: We can have this in `SetKnownStatus ()` too. Having this method would make it more readable :) and if we want to update applied status independently in the future.

agent/taskresource/cgroup/cgroup_unsupported.go
sharanyad: I just left the fields there since this is used in tests (more like a fake resource). so only having the status fields there and not actual implementations like `Create` or `Cleanup`.

agent/engine/task_manager.go
sharanyad: right now, errors during steady state is the only place, where we want to move the task to STOPPED. there may be other states in the future where we might want to handle things differently.  hence the condition check.

agent/taskresource/cgroup/cgroup.go
sharanyad: ok, what do you think about adding a new struct called `TestResource` and use that for testing? we cannot really use the mock resource for tests, since they test for the states progression.

agent/taskresource/cgroup/cgroup_unsupported.go
sharanyad: ok, what do you think about adding a new struct called TestResource and using that for testing? we cannot really use the mock resource for tests, since they test the states progression.

agent/engine/task_manager.go
sharanyad: it's more like attempting to update resource's known status. will update the description.

agent/api/task.go
sharanyad: As per offline discussion, leaving this as it is.

agent/engine/task_manager.go
```diff
@@ -732,8 +805,82 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+}
+
+// transitionResource calls applyResourceState, and then notifies the managed
+// task of the change. transitionResource is called by progressTask
+func (mtask *managedTask) transitionResource(resource taskresource.TaskResource,
+	to taskresource.ResourceStatus) {
+	engine := mtask.engine
+	err := mtask.applyResourceState(resource, to)
+
+	if engine.hasManagedTask(mtask.Arn) {
```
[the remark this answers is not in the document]
sharanyad: If for some reason, the managed task was removed from engine, this check will catch that.

agent/engine/docker_task_engine.go
```diff
@@ -252,6 +252,14 @@ func (engine *DockerTaskEngine) Disable() {
 	engine.processTasks.Lock()
 }
 
+// IsTaskManaged checks if task for the corresponding arn is present
+func (engine *DockerTaskEngine) IsTaskManaged(arn string) bool {
+	engine.processTasks.Lock()
```
richardpen: This should be a read lock.

agent/engine/task_manager.go
```diff
@@ -73,6 +81,17 @@ type containerTransition struct {
 	reason         error
 }
 
+// resourceTransition defines the struct for a resource to transition.
+type resourceTransition struct {
+	// nextState represents the next known status that the resource can move to
+	nextState taskresource.ResourceStatus
+	// status is the string value of nextState
+	status string
```
richardpen: Can we get rid of this field, as it's duplicate with `nextState`?

agent/engine/task_manager.go
```diff
@@ -732,8 +823,83 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+
+// transitionResource calls applyResourceState, and then notifies the managed
+// task of the change. transitionResource is called by progressTask
+func (mtask *managedTask) transitionResource(resource taskresource.TaskResource,
+	to taskresource.ResourceStatus) {
+	engine := mtask.engine
+	err := mtask.applyResourceState(resource, to)
+
+	if engine.IsTaskManaged(mtask.Arn) {
+		mtask.handleResourceStateChange(resourceStateChange{
```
richardpen: This still shouldn't be called directly, an event should be sent to `resourceStateChangeEvent` instead. And instead of check `IsTaskManaged` you can do something like [this](https://github.com/aws/amazon-ecs-agent/pull/1323/files#diff-20346f7af6202828de353f0137b6e43aR454)

agent/engine/task_manager_unix_test.go
```diff
@@ -0,0 +1,266 @@
...
+			}
+			canTransition, transitions := task.startResourceTransitions(
+				func(resource taskresource.TaskResource, nextStatus taskresource.ResourceStatus) {
+					t.Error("Transition function should not be called when no transitions are possible")
+				})
+			assert.Equal(t, tc.CanTransition, canTransition)
+			assert.Empty(t, transitions)
+		})
+	}
+}
```
richardpen: Can you add a test to cover the whole workflow of a task with cgroup resource?

agent/engine/task_manager.go
sharanyad: We use this field [here](https://github.com/aws/amazon-ecs-agent/pull/1332/files#diff-20346f7af6202828de353f0137b6e43aR859) and to track the map of transitions [here](https://github.com/aws/amazon-ecs-agent/pull/1332/files#diff-20346f7af6202828de353f0137b6e43aR741). Without the resource object, we cannot get the status string, hence added it as part of `resourceTransition`.

agent/taskresource/cgroup/cgroup_unsupported.go
```diff
@@ -0,0 +1,116 @@
+// +build !linux
```
richardpen: Just want to confirm, if we make the cgroup as a task resource, can we get rid of this?

agent/taskresource/cgroup/cgroup.go
```diff
@@ -44,143 +44,231 @@ var (
...
-	ioutil              ioutilwrapper.IOUtil
-	createdAt           time.Time
-	desiredStatusUnsafe taskresource.ResourceStatus
-	knownStatusUnsafe   taskresource.ResourceStatus
+	taskARN                            string
+	control                            cgroupres.Control
+	cgroupRoot                         string
+	cgroupMountPath                    string
+	resourceSpec                       specs.LinuxResources
+	IOUtil                             ioutilwrapper.IOUtil
```
richardpen: Why is this changed to be public?

agent/engine/docker_task_engine_unix_test.go
```diff
@@ -115,3 +128,125 @@ func TestEngineDisableConcurrentPull(t *testing.T) {
...
+	)
+	mockTime.EXPECT().Now().Return(time.Now()).AnyTimes()
+
+	err = taskEngine.Init(ctx)
+	assert.NoError(t, err)
+
+	taskEngine.AddTask(sleepTask)
+	cleanup := make(chan time.Time, 1)
+	mockTime.EXPECT().After(gomock.Any()).Return(cleanup).AnyTimes()
+	t.Log("gonna wait for task stop event")
```
richardpen: Do you need to keep this log line?

agent/engine/task_manager.go
```diff
@@ -423,6 +450,50 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
+		mtask.Arn, res.GetName(), res.StatusString(status), err)
+	if status == res.SteadyState() {
+		seelog.Errorf("Managed task [%s]: error while creating resource %s, setting the task's desired status to STOPPED",
+			mtask.Arn, res.GetName())
+		mtask.SetDesiredStatus(api.TaskStopped)
+		mtask.engine.saver.Save()
+	}
+}
+
+func (mtask *managedTask) emitResourceChange(change resourceStateChange) {
```
richardpen: You will need to do something similar to [this](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/task_manager.go#L894) during cleanup, and also close this channel after the clean up.

agent/taskresource/cgroup/cgroup_unsupported.go
sharanyad: cgroup will be enabled as task resource only when `TaskCPUMemLimit` is enabled. This is currently disabled by default in windows. If for some reason, user enables `TaskCPUMemLimit` in windows, we will need cgroup object. In this case, the unsupported one will still be used.

agent/taskresource/cgroup/cgroup.go
sharanyad: so that we could use it in tests [here](https://github.com/aws/amazon-ecs-agent/pull/1332/files#diff-af65229648e7812b105ca712053403b3R152).

agent/engine/task_manager.go
sharanyad: done.

agent/taskresource/cgroup/cgroup.go
richardpen: Then can you add a setter for it?

sharanyad: setter is a good idea, though right now it's used only in the tests. added the same.

agent/taskresource/cgroup/cgroup_unsupported.go
```diff
@@ -0,0 +1,116 @@
...
+
+import (
+	"errors"
+	"time"
+
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+)
+
+// CgroupResource represents Cgroup resource
+type CgroupResource struct{}
```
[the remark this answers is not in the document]
sharanyad: It should ideally not be. But depending on how the flow of cgroup resource creation is (if its in a common file and task cpu limit var is enabled by mistake, windows would look for cgroup obj), this may or may not be used. The next PR should determine that and we can remove or keep this accordingly.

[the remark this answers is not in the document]
aaithal: > (if its in a common file and task cpu limit var is enabled by mistake, windows would look for cgroup obj), this may or may not be used.

Even in that case, it should be handled in the config package. i.e. if windows, explicitly disable it. If that's not being enforced, can you please make that change so that this file is not needed at all?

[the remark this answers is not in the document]
sharanyad: This is to confirm that config explicitly disables task cpu limit for windows. Removed `cgroup_unsupported.go`.

agent/engine/task_manager.go
```diff
@@ -732,8 +832,83 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+	}
+
+	return anyCanTransition, transitions
+}
+
+// transitionResource calls applyResourceState, and then notifies the managed
+// task of the change. transitionResource is called by progressTask
+func (mtask *managedTask) transitionResource(resource taskresource.TaskResource,
+	to taskresource.ResourceStatus) {
+	engine := mtask.engine
```
aaithal: nit: you don't need this line

agent/engine/docker_task_engine.go
```diff
@@ -461,6 +469,16 @@ func (engine *DockerTaskEngine) deleteTask(task *api.Task) {
 		}
 	}
 
+	for _, resource := range task.Resources {
+		err := resource.Cleanup()
+		if err != nil {
+			seelog.Warnf("Task engine [%s]: unable to cleanup resource %s: %v",
+				task.Arn, resource.GetName(), err)
+		}
+		seelog.Debugf("Task engine [%s]: resource %s cleanup complete", task.Arn,
```
aaithal: don't you need an `else` here?

agent/engine/task_manager.go
```diff
@@ -732,8 +832,83 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+			continue
+		}
+		if !transition.actionRequired {
+			// no action is required for the transition, just set the known status without
+			// calling any transition function
+			go func(res taskresource.TaskResource, status taskresource.ResourceStatus, err error) {
+				mtask.emitResourceChange(resourceStateChange{
+					resource:  res,
+					nextState: transition.nextState,
+					err:       nil,
```
aaithal: This should be `err: err`. Or, remove `err error` arg from the function

sharanyad:
comments addressed. 
Previous engine package unit test coverage info: 
```
coverage: 87.4% of statements
ok      github.com/aws/amazon-ecs-agent/agent/engine    5.287s  coverage: 87.4% of statements 
```

Current: 
```
coverage: 88.0% of statements
ok      github.com/aws/amazon-ecs-agent/agent/engine    5.363s  coverage: 88.0% of statements 
```

agent/engine/task_manager.go
```diff
@@ -423,6 +450,50 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 		mtask.Arn, container.Name, mtask.GetDesiredStatus().String())
 }
 
+// handleResourceStateChange attempts to update resource's known status depending on
+// the current status and errors during transition
+func (mtask *managedTask) handleResourceStateChange(resChange resourceStateChange) {
+	// locate the resource
+	res := resChange.resource
+	if !mtask.isResourceFound(res) {
```
aaithal: why do you need this check at all? The call path seems to indicate that this shouldn't be needed. If we need to do something like this, it's a major logic error in the agent and we should just `panic()` in that case. Either get rid of this or add enough tests to ensure that this doesn't ever happen.

agent/engine/task_manager.go
```diff
@@ -732,8 +832,82 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+		transition := mtask.resourceNextState(res)
+		// If the resource is already in a transition, skip
+		if transition.actionRequired && !res.SetAppliedStatus(transition.nextState) {
+			// At least one resource is able to be moved forwards, so we're not deadlocked
+			continue
+		}
+		if !transition.actionRequired {
+			// no action is required for the transition, just set the known status without
+			// calling any transition function
+			go func(res taskresource.TaskResource, status taskresource.ResourceStatus) {
```
aaithal: why not just do `go mtask.emitResourceChange()` here? Also, why is this in a go routine? Or, why is a similar call in `transitionResource()` not in a go routine?

agent/taskresource/interface.go
```diff
@@ -21,6 +21,16 @@ import (
 // ResourceStatus is an enumeration of valid states of task resource lifecycle
 type ResourceStatus int32
 
+// This is currently used for testing purposes
```
aaithal: is this line/comment still accurate?

agent/engine/task_manager.go
sharanyad: This is just another level of check to make sure things are sane. There is a similar check for container today here: https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/engine/task_manager.go#L384 This scenario should not happen. I can remove the check if you have a strong preference.

agent/taskresource/interface.go
sharanyad: no, will remove it.

agent/engine/task_manager.go
aaithal: Let's log this at Critical level then (apologies if i asked you to change it to Error earlier.

PR #1340, opened 2018-04-11
eventhandler: reduce redundant SubmitTaskStateChange api calls

haikuoliu:
### Summary

Update the ecs client api, remove events when the SubmitTaskStateChange api call fails and returns a InvalidParameterException.

### Testing

### Licensing


agent/eventhandler/handler_test.go
```diff
@@ -93,6 +95,35 @@ func TestSendsEventsOneEventRetries(t *testing.T) {
...
+	client.EXPECT().SubmitTaskStateChange(gomock.Any()).Do(func(interface{}) {
+		assert.Equal(t, 1, handler.tasksToEvents[taskARN].events.Len())
+		wg.Done()
+	}).Return(awserr.New(ecs.ErrCodeInvalidParameterException, "", nil))
+
+	handler.AddStateChangeEvent(taskEvent, client)
+
+	wg.Wait()
+	// Require the lock to wait for submitFirstEvent to be finished
+	handler.tasksToEvents[taskARN].lock.Lock()
```
richardpen: There may be a race condition here, as the SubmitTaskStateChange is called in a goroutine, can you add a retry here?

haikuoliu: Can you please explain a bit why the race condition may happen?  I think `handler.tasksToEvents[taskARN].lock.Lock()` will not run until `SubmitTaskStateChange` is called as we have `wg.wait()`. I run the tests for 10000 times and they passed.

richardpen: See the following execution order as an example:
```
1. handler.AddStateChangeEvent(taskEvent, client)
2. SubmitTaskStateChange(this is in go routine)
3. wg.Done
4. wg.Wait
5. assert.Equal(t, 0, handler.tasksToEvents[taskARN].events.Len()) #this will fail, because it's not removed from the list
6. taskEvents.events.Remove(eventToSubmit)
```

haikuoliu: Hi, I used a little trick here, because before `SubmitTaskStateChange ` in `submitFirstEvent` is called,  `taskEvents.lock.Lock()` will be called first, and this lock will be unlocked when `submitFirstEvent` is finished. So before I do `assert.Equal(t, 0, handler.tasksToEvents[taskARN].events.Len())`, first I try to get the lock `handler.tasksToEvents[taskARN].lock.Lock()`, and we will block here until `submitFirstEvent` is finished, in which case we are guarantee that the event will be removed before the assertion (if the remove really happens).

richardpen: Oh, sorry, apparently I missed that.

haikuoliu: It's ok, it works in the same way as doing a retry.

agent/eventhandler/task_handler.go
```diff
@@ -347,11 +349,21 @@ func (taskEvents *taskSendableEvents) submitFirstEvent(handler *TaskHandler, bac
 	} else if event.taskShouldBeSent() {
 		if err := event.send(sendTaskStatusToECS, setTaskChangeSent, "task",
 			handler.client, eventToSubmit, handler.stateSaver, backoff, taskEvents); err != nil {
+			if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == ecs.ErrCodeInvalidParameterException {
```
aaithal: can you move this logic into its own method? you can then avoid code duplication here and in line 362:
```golang
if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == ecs.ErrCodeInvalidParameterException
```

haikuoliu: Ok. What about further generalizing it into IsAwsErrAndEqualToErrCode(error, errorCode) and put this function at agent/utils/error.go? same logic is used in agent/app/agent.go line 470 and we will use more similar logics in the next pr for RegisterContainerInstance API:
```golang
if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == ecs.ErrCodeInvalidParameterException {
```

sharanyad: You can move the error check into a method. Also can you add another method that handles this error type in this file?

haikuoliu: OK, sure.

haikuoliu:
Update: abstract the checking error code if statement out and add corresponding test.

agent/utils/utils.go
```diff
@@ -199,3 +200,11 @@ func ParseBool(str string, default_ bool) bool {
 	}
 	return res
 }
+
+// IsAwsErrAndEqualToEcsErrCode returns true if the err implements
+// the Error interface of awserr and it has the same error code as
+// the passed in ecs error code.
+func IsAwsErrAndEqualToEcsErrCode(err error, ecsErrCode string) bool {
```
aaithal: As per https://github.com/golang/go/wiki/CodeReviewComments#initialisms `Aws` -> `AWS`, `Ecs` -> `ECS`. 

Also, I think `IsErrorCodeEqual(err error, code string)` seems much shorter and meaningful.

agent/eventhandler/task_handler.go
```diff
@@ -382,3 +393,7 @@ func (handler *TaskHandler) getTasksToEventsLen() int {
 
 	return len(handler.tasksToEvents)
 }
+
+func (handler *TaskHandler) isInvalidParameterException(err error) bool {
```
aaithal: This doesn't need to be a member of the `handler` struct

haikuoliu: OK

agent/utils/utils.go
haikuoliu: ok, will change to `IsErrorCodeEqual(err error, code string)`

haikuoliu:
Update: minor refactor.

agent/utils/utils.go
```diff
@@ -199,3 +200,11 @@ func ParseBool(str string, default_ bool) bool {
 	}
 	return res
 }
+
+// IsErrorCodeEqual returns true if the err implements Error
+// interface of awserr and it has the same error code as
+// the passed in error code.
+func IsErrorCodeEqual(err error, code string) bool {
```
sharanyad: Since this method also checks if this is an `awserr`, can you name this `IsAWSErrorCodeEqual`?

agent/eventhandler/task_handler.go
```diff
@@ -347,11 +348,21 @@ func (taskEvents *taskSendableEvents) submitFirstEvent(handler *TaskHandler, bac
...
+				// Remove the event when its parameters are invalid to reduce the redundant API call
+				seelog.Warnf("TaskHandler: Event is sent with invalid parameters; just removing: %s", event.toString())
+				taskEvents.events.Remove(eventToSubmit)
+			}
 			return false, err
 		}
 	} else if event.taskAttachmentShouldBeSent() {
 		if err := event.send(sendTaskStatusToECS, setTaskAttachmentSent, "task attachment",
 			handler.client, eventToSubmit, handler.stateSaver, backoff, taskEvents); err != nil {
+			if isInvalidParameterException(err) {
```
sharanyad: Could you just use `if utils.IsErrorCodeEqual(err, ecs.ErrCodeInvalidParameterException)` here since the method `isInvalidParameterException` is not adding much value?

sharanyad: Instead the below two statements can be moved to a method called `handleInvalidParamException` or something similar.

haikuoliu: oh I see what you mean.

agent/utils/utils.go
haikuoliu: ok

agent/eventhandler/task_handler.go
```diff
@@ -382,3 +385,12 @@ func (handler *TaskHandler) getTasksToEventsLen() int {
 
 	return len(handler.tasksToEvents)
 }
+
+// handleInvalidParamException removes the event from event queue when its parameters are
+// invalid to reduce redundant API call
+func handleInvalidParamException(err error, events *list.List, eventToSubmit *list.Element, event *sendableEvent) {
```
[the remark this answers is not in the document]
haikuoliu: Ok, so I will remove `event`, but that needs an extra conversion in `handleInvalidParamException` to get the event string.

Issue #413, opened 2016-05-26
Deployment successful even though pulling the new image failed


fabfuel:
I redeployed a service on two instances. On one instance, one of the container images could not be pulled (what is actually a ECR problem), but the ECS agent considered the deployment as successful. 
The result is, that the service has reached a steady state, but it's delivering two different versions. That's a blocker in my opinion, because the deployment actually failed and you do not notice.

instance A:
`XXX.dkr.ecr.us-east-1.amazonaws.com/XXX/nginx       develop             c4c0b6fa9ad2        14 minutes ago      182.7 MB`

instance B:
`XXX.dkr.ecr.us-east-1.amazonaws.com/XXX/nginx      develop             e29ed85a1df4        8 days ago          182.7 MB`

This is from the ecs agent log:

> 2016-05-26T10:00:09Z [INFO] Error while pulling container; will try to run anyways module="TaskEngine" task="XXX-staging:38 arn:aws:ecs:eu-central-1:XXX:task/8ce9da0c-51a9-483f-9185-c2c76e4c5dd1, Status: (NONE->RUNNING) Containers: [nginx (PULLED->RUNNING),app (RUNNING->RUNNING),]" err="Error: image XXX/nginx:develop not found"

Both instances are running:
ECS agent: 1.9.0
Docker: 1.9.1

Is this the expected behavior?
Do you need any more information?

Thanks
Fabian

aaithal:
Hi @fabfuel, 

Thank you for reporting this issue. The behavior of the ECS Agent where a container is started even when the pull fails for the image and the image already exists on the host is by design. The intent here is to reduce the impact of not being able to pull an image in case of failures (such as networking events). At this point, we recommend updating your task definition with a new image tag per deployment as a way to ensure the correct image is used. 

However, we do recognize that this workaround could lead to an explosion in the number of tags that need to be maintained. Could you please let us know if the proposal listed below helps you in resolving your issue? 

# Pull Behavior Proposal 
It should be possible to configure the ECS Agent to customize the `pull` behavior. The `ECS_AGENT_PULL_BEHAVIOR` environment variable would be used for this purpose. The following values can be used to modify this behavior: 
* `default`: Every time a container is started as a result of task launch, ECS Agent attempts to pull the container. If a pull fails, ECS Agent tries to start from the Docker image cache anyway, assuming that the image has not changed. 
* `always`: Every time a container is started, ECS Agent attempts to pull the container. If the pull fails, the task fails. 
[...]

PR #1348, opened 2018-04-17
engine: introduce a new env var to distinct pull image behavior

haikuoliu:
### Summary

This PR is to fix issue #413, it introduces a new environment variable `ECS_AGENT_PULL_BEHAVIOR ` to allow users to define different pull image behaviors, please see the design doc in the comments of issue #413, and usage of the `ECS_AGENT_PULL_BEHAVIOR` in README.md.

### Implementation details

1. Add an enum environment variable to config file.
2. Add a new variable `PullSucceeded` in `ImageState`, it will be set to `true` when the pull succeeds.
3. For `prefer-cached` behavior, check if there is cached image by inspecting image, if inspecting image fails, then pull image, otherwise just update container reference. Make task fail when pull fails.
4. For `never` behavior, do nothing but update container reference when pulling container. 
5. For `once` behavior, check if the image has been pulled by inspecting the `IsPullSuccess` of the image state, do nothing when the image has been pulled, otherwise pull as usual. If the pull image API call fails, make the task fail.
6. Abstract the update container reference out in docker_task_engine.go.
7. Some other refactor/bug fix.

### Testing

New tests cover the changes: yes, unit tests, integ tests and some manual tests.

Feature - Introduce a new environment variable `ECS_AGENT_PULL_BEHAVIOR` to make agent pull behavior configurable

### Licensing


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.18.0
```
sharanyad: The convention is to use `1.18.0-dev` before release

haikuoliu: ok.

aaithal:
@haikuoliu  can you please fix failing tests?

haikuoliu:
@aaithal It's the timeout test `TestStartContainerTimeout` and not related to this PR. The timeout tests have a small change to fail, because the timeout value is too small and some methods don't get to execute. 

We talked about this in the wire context PR #1329: to fix it either use a mock context (implemented in the initial PR), or use a cancel context to simulate a timeout context (in one of the revised PR), or don't fix it at all (in final PR), we choose the last method.

aaithal:
>  or don't fix it at all (in final PR), we choose the last method.

That doesn't seem right. I think you 'fix' it by adding `MaxTimes(1)` to the mock docker client call. Can you please make that change it it hasn't already been made?

agent/config/types.go
```diff
@@ -20,6 +20,8 @@ import (
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 )
 
+type AgentPullBehaviorType int8
```
aaithal: this needs a comment

agent/engine/docker_task_engine.go
```diff
@@ -641,6 +641,29 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
 		if emptyvolume.LocalImage {
 			return engine.client.ImportLocalEmptyVolumeImage()
 		}
+	default:
+		// If the agent pull behavior is never, we don't try to pull the image,
+		// try to use local image cache instead.
+		if engine.cfg.AgentPullBehavior == config.NeverAgentPullBehavior {
```
aaithal: can you please refactor these into a method? May be something that returns a boolean to indicate if you should return `dockerapi.DockerContainerMetadata{Error: nil}`?

agent/engine/image/types.go
```diff
@@ -37,11 +37,12 @@ func (image *Image) String() string {
...
-	Image      *Image
-	Containers []*api.Container `json:"-"`
-	PulledAt   time.Time
-	LastUsedAt time.Time
-	updateLock sync.RWMutex
+	Image           *Image
+	Containers      []*api.Container `json:"-"`
+	PulledAt        time.Time
+	LastUsedAt      time.Time
+	IsPulledSuccess bool
```
aaithal: `PullSucceeded` seems like a better name for this. Also, please add documentation for this and all other fields.

agent/engine/task_manager.go
```diff
@@ -543,8 +544,22 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
...
+		// the image pull fails, the task should fail. If we don't fail task here,
+		// then the cached image will probably be used for creating container, and we
+		// don't want to use cached image for both cases.
+		// If the agent pull behavior is never, we won't receive error here, the case
+		// where there is no local image cache will be handled when creating container.
+		if mtask.cfg.AgentPullBehavior == config.AlwaysAgentPullBehavior ||
+			mtask.cfg.AgentPullBehavior == config.OnceAgentPullBehavior {
+			seelog.Criticalf("Managed task [%s]: Error while pulling container %s, task will fail: %v",
+				mtask.Arn, container.Name, event.Error)
+			mtask.SetDesiredStatus(api.TaskStopped)
```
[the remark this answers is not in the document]
aaithal: Essential or non-essential is only effective for `RUNNING` containers. It shouldn't have a bearing on `Pulled` transition failing

aaithal:
@haikuoliu  you also need to update the version and documentation for the agent's state file in https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/statemanager/state_manager.go#L65

haikuoliu:
@aaithal My bad, thanks for pointing out. I will fix it.

agent/engine/image/types.go
haikuoliu: OK, sure.

agent/engine/docker_task_engine.go
haikuoliu: OK, will abstract it out.

agent/config/types.go
haikuoliu: OK

agent/config/config.go
```diff
@@ -86,6 +86,23 @@ const (
 	// performing image cleanup.
 	minimumNumImagesToDeletePerCycle = 1
 
+	// DefaultAgentPullBehavior specficies the behavior that if an image pull API call fails,
+	// agent tries to start from the Docker image cache anyway, assuming that the image has not changed.
+	DefaultAgentPullBehavior AgentPullBehaviorType = 0
```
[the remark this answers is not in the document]
haikuoliu: OK, sure.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
...
+* Enhancement - Introduce a new environment variable `ECS_AGENT_PULL_BEHAVIOR` to make agent pull behavior configurable [#to be added](to be added)
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/engine/task_manager.go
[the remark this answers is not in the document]
haikuoliu: The task will stop for both essential or non-essential containers. I will add some comments for that.

README.md
```diff
@@ -169,6 +169,7 @@ additional details on each available environment variable.
 | `ECS_IMAGE_CLEANUP_INTERVAL` | 30m | The time interval between automated image cleanup cycles. If set to less than 10 minutes, the value is ignored. | 30m | 30m |
 | `ECS_IMAGE_MINIMUM_CLEANUP_AGE` | 30m | The minimum time interval between when an image is pulled and when it can be considered for automated image cleanup. | 1h | 1h |
 | `ECS_NUM_IMAGES_DELETE_PER_CYCLE` | 5 | The maximum number of images to delete in a single automated image cleanup cycle. If set to less than 1, the value is ignored. | 5 | 5 |
+| `ECS_AGENT_PULL_BEHAVIOR` | &lt;default &#124; always &#124; never &#124; once&gt; | Behavior used to customize the pull image process. default: pull image remotely, use cached image if the pull fails; always: only pull image remotely, the task fails when the pull fails; never: only use cached image; once: pull image remotely only if this image has never been pulled before, use cached image otherwise. | default | default |
```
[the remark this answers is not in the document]
haikuoliu: OK, will ask review from him.

agent/config/config_test.go
```diff
@@ -369,6 +371,14 @@ func TestImageCleanupMinimumNumImagesToDeletePerCycle(t *testing.T) {
 	assert.Equal(t, cfg.NumImagesToDeletePerCycle, DefaultNumImagesToDeletePerCycle, "Wrong value for NumImagesToDeletePerCycle")
 }
 
+func TestInvalidAgentPullBehavior(t *testing.T) {
```
aaithal: can you add a test for just `getAgentPullBehavior()` as well? Make it a table driven test so that you can test that all 4 valid values and an invalid value for `ECS_AGENT_PULL_BEHAVIOR` return the expected output.

agent/engine/docker_task_engine.go
```diff
@@ -741,16 +776,23 @@ func (engine *DockerTaskEngine) pullAndUpdateContainerReference(task *api.Task,
 	if container.IsInternal() {
 		return metadata
 	}
+	PullSucceeded := metadata.Error == nil
```
aaithal: why is this `PullSucceeded` and not `pullSucceeded`?

agent/engine/docker_task_engine.go
```diff
@@ -657,6 +663,35 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
...
+		// It's not an internal container, so it's safe to add container reference.
+		engine.updateContainerReference(false, container, taskArn)
+		return true
+	} else if agentPullBehavior == config.OnceAgentPullBehavior {
+		// If this image has been pulled successfully before, don't pull the image,
+		// otherwise pull the image as usual, regardless whether the image exists or not
+		// (the image can be prepopulated with the AMI and never be pulled).
+		imageState := engine.imageManager.GetImageStateFromImageName(container.Image)
+		if imageState != nil && imageState.PullSucceeded {
+			seelog.Infof("Task engine [%s]: image %s has been pulled, use it directly for container %s",
```
aaithal: i'd suggest changing this slightly to :`image %s has been pulled once, not pulling it again`

agent/engine/docker_task_engine.go
```diff
@@ -657,6 +663,35 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
...
 
+// useLocalImageCache returns true if local image cache should be used, or return false to continue
+// pulling image, by inspecting the agent pull behavior variable defined in config. The caller has
+// to make sure the container passed in is not an internal container.
+func (engine *DockerTaskEngine) useLocalImageCache(agentPullBehavior config.AgentPullBehaviorType,
+	container *api.Container,
+	taskArn string) bool {
+	// If the agent pull behavior is never, we don't try to pull the image,
+	// try to use local image cache instead.
+	if agentPullBehavior == config.NeverAgentPullBehavior {
```
aaithal: minor: this'd be more readable with a `switch` and cases.

agent/engine/docker_task_engine.go
```diff
@@ -657,6 +663,35 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
...
+			taskArn, container.Name)
+		// It's not an internal container, so it's safe to add container reference.
+		engine.updateContainerReference(false, container, taskArn)
+		return true
+	} else if agentPullBehavior == config.OnceAgentPullBehavior {
+		// If this image has been pulled successfully before, don't pull the image,
+		// otherwise pull the image as usual, regardless whether the image exists or not
+		// (the image can be prepopulated with the AMI and never be pulled).
+		imageState := engine.imageManager.GetImageStateFromImageName(container.Image)
+		if imageState != nil && imageState.PullSucceeded {
```
aaithal: I think all references to `PullSucceeded` need to be protected with a lock. Please add setters and getters in image state struct for that use those methods everywhere.

agent/engine/task_manager.go
```diff
@@ -543,8 +544,24 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
...
-		// not be fatal if e.g. the image already exists.
+		// If the agent pull behavior is always or once, we receive the error because
+		// the image pull fails, the task should fail. If we don't fail task here,
+		// then the cached image will probably be used for creating container, and we
+		// don't want to use cached image for both cases.
+		// If the agent pull behavior is never, we won't receive error here, the case
+		// where there is no local image cache will be handled when creating container.
+		if mtask.cfg.AgentPullBehavior == config.AlwaysAgentPullBehavior ||
+			mtask.cfg.AgentPullBehavior == config.OnceAgentPullBehavior {
+			seelog.Criticalf("Managed task [%s]: Error while pulling container %s, task will fail: %v",
```
aaithal: you should probably log the container image here as well.

agent/config/config.go
```diff
@@ -455,6 +476,24 @@ func getContainerStartTimeout() time.Duration {
 	return containerStartTimeout
 }
 
+func getAgentPullBehavior() AgentPullBehaviorType {
```
aaithal: you should rebase on top my pr #1353 once it's merged so that this gets moved to `parse.go`.

agent/engine/task_manager.go
```diff
@@ -543,8 +544,24 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
...
+				mtask.Arn, container.Name, event.Error)
+			// The task should be stopped regardless of whether this container is
+			// essential or non-essential.
+			mtask.SetDesiredStatus(api.TaskStopped)
```
richardpen: You may also want to check if the container is essential or not. And why are you emit the `STOPPED` event, the task may not be `STOPPED` at this point.

agent/engine/docker_task_engine.go
```diff
@@ -741,16 +776,23 @@ func (engine *DockerTaskEngine) pullAndUpdateContainerReference(task *api.Task,
 	if container.IsInternal() {
 		return metadata
 	}
+	PullSucceeded := metadata.Error == nil
+	engine.updateContainerReference(PullSucceeded, container, task.Arn)
```
sharanyad: Can this logic be handled here - https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/engine/docker_image_manager.go#L112 
i.e, depending on `InspectImage`, set `PullSucceeded` value?

agent/engine/docker_task_engine_test.go
```diff
@@ -1380,6 +1380,89 @@ func TestPullNormalImage(t *testing.T) {
...
+	}
+
+	imageManager.EXPECT().RecordContainerReference(container)
+	imageManager.EXPECT().GetImageStateFromImageName(imageName).Return(imageState)
+	saver.EXPECT().Save()
+	metadata := taskEngine.pullContainer(task, container)
+	assert.Equal(t, dockerapi.DockerContainerMetadata{}, metadata, "expected empty metadata")
+}
+
+func TestPullImageWithOnceAgentPullBehavior(t *testing.T) {
```
sharanyad: Can you also add a test where config is `once` and `PullSucceeded` is false?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.18.0-dev
+* Feature - Introduce a new environment variable `ECS_AGENT_PULL_BEHAVIOR` to make agent pull behavior configurable [#1348](https://github.com/aws/amazon-ecs-agent/pull/1348)
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/config/config.go
```diff
@@ -325,6 +344,7 @@ func environmentConfig() (Config, error) {
 	imageCleanupInterval := parseEnvVariableDuration("ECS_IMAGE_CLEANUP_INTERVAL")
 	numImagesToDeletePerCycleEnvVal := os.Getenv("ECS_NUM_IMAGES_DELETE_PER_CYCLE")
 	numImagesToDeletePerCycle, err := strconv.Atoi(numImagesToDeletePerCycleEnvVal)
+	agentPullBehavior := getAgentPullBehavior()
```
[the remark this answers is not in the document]
haikuoliu: I wanted to be consistent with other code. I just saw your refactor config commit, will change accordingly.

agent/config/config.go
```diff
@@ -455,6 +476,24 @@ func getContainerStartTimeout() time.Duration {
 	return containerStartTimeout
 }
 
+func getAgentPullBehavior() AgentPullBehaviorType {
+	var agentPullBehavior AgentPullBehaviorType
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/config/config_test.go
haikuoliu: OK

agent/engine/docker_task_engine.go
haikuoliu: OK

agent/engine/docker_task_engine.go
haikuoliu: Yes, it should be `pullSucceeded`.

agent/engine/docker_task_engine.go
haikuoliu: OK

agent/engine/docker_task_engine.go
haikuoliu: OK

agent/engine/task_manager.go
haikuoliu: OK

agent/config/config.go
haikuoliu: OK

agent/config/config.go
```diff
@@ -455,6 +476,24 @@ func getContainerStartTimeout() time.Duration {
...
+	agentPullBehaviorString := os.Getenv("ECS_AGENT_PULL_BEHAVIOR")
+	switch agentPullBehaviorString {
```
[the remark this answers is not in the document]
haikuoliu: ok, Good idea!

agent/engine/task_manager.go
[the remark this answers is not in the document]
sharanyad: Today, when a container fails with desired `CREATED` , we move the task to `STOPPED` only if it is an essential container. We should do the same for `PULLED` too. 
If not, we might want to change the logic to stop the task for failures during `CREATED`, irrespective of the essential status of containers.

agent/engine/docker_image_manager.go
```diff
@@ -154,6 +154,9 @@ func (imageManager *dockerImageManager) addContainerReferenceToNewImageState(con
 			Image:      sourceImage,
 			PulledAt:   time.Now(),
 			LastUsedAt: time.Now(),
+			// The PullSucceeded filed is false by default,
+			// one has to explicitly set it to be true when the pull image succeeds.
+			PullSucceeded: false,
```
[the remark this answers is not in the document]
haikuoliu: This is because even if the image pull fails, the container reference will still be updated, so we set it to be false here. We check if there is no error with image pull, we then set it to be true.

agent/engine/docker_task_engine.go
```diff
@@ -657,6 +663,35 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
 	return engine.serialPull(task, container)
 }
 
+// useLocalImageCache returns true if local image cache should be used, or return false to continue
+// pulling image, by inspecting the agent pull behavior variable defined in config. The caller has
+// to make sure the container passed in is not an internal container.
+func (engine *DockerTaskEngine) useLocalImageCache(agentPullBehavior config.AgentPullBehaviorType,
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/engine/docker_task_engine_test.go
haikuoliu: OK

agent/engine/docker_task_engine.go
```diff
@@ -657,6 +663,35 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
...
+	container *api.Container,
+	taskArn string) bool {
+	// If the agent pull behavior is never, we don't try to pull the image,
+	// try to use local image cache instead.
+	if agentPullBehavior == config.NeverAgentPullBehavior {
+		seelog.Infof("Task engine [%s]: use cached image directly for container %s",
+			taskArn, container.Name)
+		// It's not an internal container, so it's safe to add container reference.
+		engine.updateContainerReference(false, container, taskArn)
```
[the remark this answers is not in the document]
haikuoliu: @richardpen and I discussed about the problem we may face about image cleaning. If we use `never` or `once` agent behavior setup, the image cleanup process could remove the image, and I think for `never` and `once` we should keep the image cache, right?

We think what we can do: 
1. Let customer know that there may be conflicts between image cleaning and agent pull behavior setup, and suggest them when using `never` or `once`, set `NumImagesToDeletePerCycle` to be 0 (or some other image cleanup settings that can prevent image from being removed).

2. Disable image cleanup internally when using `never` or `once`, add some documentation on that so that user knows.

@sharanyad @richardpen @aaithal what do you think about this?

agent/engine/task_manager.go
haikuoliu: yeah, don't notify ACS right now is better.

agent/engine/docker_task_engine.go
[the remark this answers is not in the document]
haikuoliu: yeah you are right, we should.

agent/engine/docker_task_engine.go
haikuoliu: I think inspect image and pull image are two different things, when you inspect an image, the image can be there because it' reconfigured with the AMI, instead of being pulled from user's ECR, this matters when the setting is `once`.

agent/config/config.go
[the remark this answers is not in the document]
haikuoliu: Hi I think this has already been done by `func (cfg *Config) trimWhitespace()`

haikuoliu:
Update: address the comments. Introduce a new behavior "prefer_cached". Will update README, add comments and tests for the new behavior and remove "never" behavior due to the change of the requirement.

haikuoliu:
Update: Update README, add comments and tests for "prefer_cached" behavior and remove "never" behavior due to the change of the requirement. Also do a rebase.

agent/config/config_test.go
```diff
@@ -369,6 +371,56 @@ func TestImageCleanupMinimumNumImagesToDeletePerCycle(t *testing.T) {
...
+func TestInvalidImagePullBehavior(t *testing.T) {
+	defer setTestRegion()()
+	defer setTestEnv("ECS_IMAGE_PULL_BEHAVIOR", "invalid")()
+	cfg, err := NewConfig(ec2.NewBlackholeEC2MetadataClient())
+	assert.NoError(t, err)
+	assert.Equal(t, cfg.ImagePullBehavior, ImagePullDefaultBehavior, "Wrong value for ImagePullBehavior")
+}
+
+func TestParseImagePullBehavior(t *testing.T) {
+	testcases := []struct {
```
petderek: nice!

agent/config/parse.go
```diff
@@ -117,6 +117,22 @@ func parseNumImagesToDeletePerCycle() int {
 	return numImagesToDeletePerCycle
 }
 
+func parseImagePullBehavior() ImagePullBehaviorType {
+	ImagePullBehaviorString := os.Getenv("ECS_IMAGE_PULL_BEHAVIOR")
```
petderek: should we .ToLowerCase() the string?

agent/config/types.go
```diff
@@ -20,6 +20,10 @@ import (
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 )
 
+// ImagePullBehaviorType is an enum variable type corresponding to different agent pull
+// behaviors including default, always, never and once.
+type ImagePullBehaviorType int8
```
petderek: why an int8 over an int?

agent/engine/docker_image_manager.go
```diff
@@ -267,8 +271,12 @@ func (imageManager *dockerImageManager) removeExistingImageNameOfDifferentID(con
 }
 
 func (imageManager *dockerImageManager) StartImageCleanupProcess(ctx context.Context) {
-	// passing the cleanup interval as argument which would help during testing
-	imageManager.performPeriodicImageCleanup(ctx, imageManager.imageCleanupTimeInterval)
+	// If the image pull behavior is prefer cached, don't clean up the image,
+	// because the cached image is needed.
+	if imageManager.imagePullBehavior != config.ImagePullPreferCachedBehavior {
+		// passing the cleanup interval as argument which would help during testing
```
petderek: [do this!]
log that you are disabling this.

[nit]
I think I'd prefer positive comparison instead of !=.
```
if imageManager.imagePullBehavior == config.ImagePullPreferCachedBehavior {
    seelog.Info("Pull behavior is set to always use cache. Disabling cleanup")
    return
}
imageManager.performPeriodicImageCleanup(ctx, imageManager.imageCleanupTimeInterval) 
```

agent/engine/docker_task_engine.go
```diff
@@ -650,6 +657,40 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
...
+		if imageState != nil && imageState.GetPullSucceeded() {
+			seelog.Infof("Task engine [%s]: image %s has been pulled once, not pulling it again",
+				taskArn, container.Image)
+			return true
+		}
+		return false
+	case config.ImagePullPreferCachedBehavior:
+		// If the behavior is prefer_cached, don't pull if we found cached image
+		// by inspecting the image.
+		_, err := engine.client.InspectImage(container.Image)
```
petderek: does this need a timeout?

haikuoliu: No, the docker client API does not need a timeout for inspecting an image.

agent/engine/docker_image_manager.go
haikuoliu: OK, sure.

agent/config/types.go
haikuoliu: Because it's just an enum variable, which doesn't require a large range, there is usage of `int8` elsewhere in agent:
https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/acs/update_handler/updater.go#L65

agent/config/parse.go
haikuoliu: I think we use `trimWhitespace ` to format every env var string and we should not do extra work outside of this method:
https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/config/config.go#L187

agent/config/config_test.go
haikuoliu: Thanks!

agent/config/config.go
```diff
@@ -102,6 +102,25 @@ const (
 	DefaultTaskMetadataBurstRate = 60
 )
 
+const (
+	// ImagePullDefaultBehavior specifies the behavior that if an image pull API call fails,
+	// agent tries to start from the Docker image cache anyway, assuming that the image has not changed.
+	ImagePullDefaultBehavior ImagePullBehaviorType = iota
```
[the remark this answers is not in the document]
haikuoliu: I think `ImagePullDefault` is more general, it describes the `ImagePull` but does not specify which aspect it's describing. I wouldn't know if it's ImagePull default timeout, or ImagePull default behavior, etc. So `ImagePullDefaultBehavior ` makes more sense to me. Let me know if you have more comments on this.

agent/engine/docker_task_engine.go
```diff
@@ -650,6 +657,40 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
 	return engine.serialPull(task, container)
 }
 
+// shouldUseLocalImage returns true if local image cache should be used, or return false to continue
+// pulling image, by inspecting the agent pull behavior variable defined in config. The caller has
+// to make sure the container passed in is not an internal container.
+func (engine *DockerTaskEngine) shouldUseLocalImage(ImagePullBehavior config.ImagePullBehaviorType,
+	container *api.Container,
+	taskArn string) bool {
+	switch ImagePullBehavior {
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/engine/docker_task_engine.go
```diff
@@ -650,6 +657,40 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
...
+	case config.ImagePullOnceBehavior:
+		// If this image has been pulled successfully before, don't pull the image,
+		// otherwise pull the image as usual, regardless whether the image exists or not
+		// (the image can be prepopulated with the AMI and never be pulled).
+		imageState := engine.imageManager.GetImageStateFromImageName(container.Image)
+		if imageState != nil && imageState.GetPullSucceeded() {
```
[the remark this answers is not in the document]
haikuoliu: OK.

agent/engine/docker_task_engine.go
```diff
@@ -734,16 +775,23 @@ func (engine *DockerTaskEngine) pullAndUpdateContainerReference(task *api.Task,
...
 
+func (engine *DockerTaskEngine) updateContainerReference(pullSucceeded bool, container *api.Container, taskArn string) {
 	err := engine.imageManager.RecordContainerReference(container)
 	if err != nil {
 		seelog.Errorf("Task engine [%s]: Unable to add container reference to image state: %v",
-			task.Arn, err)
+			taskArn, err)
 	}
 	imageState := engine.imageManager.GetImageStateFromImageName(container.Image)
+	if imageState != nil && pullSucceeded {
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/engine/task_manager.go
```diff
@@ -543,10 +543,26 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
...
-		// not be fatal if e.g. the image already exists.
-		seelog.Errorf("Managed task [%s]: Error while pulling container %s, will try to run anyway: %v",
-			mtask.Arn, container.Name, event.Error)
+		// If the agent pull behavior is always or once, we receive the error because
+		// the image pull fails, the task should fail. If we don't fail task here,
+		// then the cached image will probably be used for creating container, and we
+		// don't want to use cached image for both cases.
+		if mtask.cfg.ImagePullBehavior == config.ImagePullAlwaysBehavior ||
+			mtask.cfg.ImagePullBehavior == config.ImagePullOnceBehavior {
+			seelog.Criticalf("Managed task [%s]: Error while pulling container %s and image %s, task will fail: %v",
```
sharanyad: This can be `Errorf` and the message can be `Managed task [%s]: Error while pulling image %s for container %s , moving task to STOPPED:`

agent/engine/docker_image_manager.go
```diff
@@ -155,6 +156,9 @@ func (imageManager *dockerImageManager) addContainerReferenceToNewImageState(con
 			Image:      sourceImage,
 			PulledAt:   time.Now(),
 			LastUsedAt: time.Now(),
+			// The PullSucceeded filed is false by default,
+			// one has to explicitly set it to be true when the pull image succeeds.
+			PullSucceeded: false,
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/engine/docker_image_manager.go
```diff
@@ -155,6 +156,9 @@ func (imageManager *dockerImageManager) addContainerReferenceToNewImageState(con
 			Image:      sourceImage,
 			PulledAt:   time.Now(),
 			LastUsedAt: time.Now(),
+			// The PullSucceeded filed is false by default,
```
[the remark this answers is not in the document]
haikuoliu: OK

agent/engine/task_manager.go
haikuoliu: OK

agent/config/parse.go
```diff
@@ -117,6 +117,22 @@ func parseNumImagesToDeletePerCycle() int {
...
+	switch ImagePullBehaviorString {
+	case "always":
+		return ImagePullAlwaysBehavior
+	case "once":
+		return ImagePullOnceBehavior
+	case "prefer_cached":
```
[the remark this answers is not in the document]
aaithal: Those are examples. These are key words that we're defining. All of our APIs are camlCased and don't have underscroes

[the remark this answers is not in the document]
joelbrandenburg: For this, I don't think the APIs matter because this is a value rather than a parameter.  My opinion is to use "prefer-cached".

haikuoliu:
Update: rebase and refactor to address the comments. Currently I am not sure about when this feature should be released, so I keep it 1.17.4 in changelog, will update the changelog before merged.

README.md
```diff
@@ -169,6 +169,7 @@ additional details on each available environment variable.
 | `ECS_IMAGE_CLEANUP_INTERVAL` | 30m | The time interval between automated image cleanup cycles. If set to less than 10 minutes, the value is ignored. | 30m | 30m |
 | `ECS_IMAGE_MINIMUM_CLEANUP_AGE` | 30m | The minimum time interval between when an image is pulled and when it can be considered for automated image cleanup. | 1h | 1h |
 | `ECS_NUM_IMAGES_DELETE_PER_CYCLE` | 5 | The maximum number of images to delete in a single automated image cleanup cycle. If set to less than 1, the value is ignored. | 5 | 5 |
+| `ECS_IMAGE_PULL_BEHAVIOR` | &lt;default &#124; always &#124; once &#124; prefer-cached &gt; | The behavior used to customize the pull image process. If `default` is specified, the image will be pulled remotely, if the pull fails then the cached image will be used. If `always` is specified, the image will be pulled remotely, if the pull fails then the task will fail. If `once` is specified, the image will be pulled remotely if it has not been pulled before, otherwise the cached image will be used. If `prefer-cached` is specified, the image will be pulled remotely if there is no cached image, otherwise the cached image will be used. | default | default |
```
richardpen: Is this reviewed by @nrdlngr?  Also for `once` it should be: the image will be pulled remotely if it has not been pulled or it has been removed by image cleanup.

agent/engine/docker_image_manager.go
```diff
@@ -267,6 +270,12 @@ func (imageManager *dockerImageManager) removeExistingImageNameOfDifferentID(con
 }
 
 func (imageManager *dockerImageManager) StartImageCleanupProcess(ctx context.Context) {
+	// If the image pull behavior is prefer cached, don't clean up the image,
+	// because the cached image is needed.
+	if imageManager.imagePullBehavior == config.ImagePullPreferCachedBehavior {
+		seelog.Info("Pull behavior is set to always use cache. Disabling cleanup")
```
richardpen: Can we point this out in the readme file?

agent/engine/docker_image_manager_integ_test.go
```diff
@@ -81,23 +81,23 @@ func TestIntegImageCleanupHappyCase(t *testing.T) {
...
+	imageState1, ok := imageManager.GetImageStateFromImageName(test1Image1Name)
+	if ok {
 		t.Logf("Found image state for %s", test1Image1Name)
-	}
-	imageState2 := imageManager.GetImageStateFromImageName(test1Image2Name)
-	if imageState2 == nil {
-		t.Fatalf("Could not find image state for %s", test1Image2Name)
 	} else {
-		t.Logf("Found image state for %s", test1Image2Name)
+		t.Fatalf("Could not find image state for %s", test1Image1Name)
```
richardpen: since you are here, can you change these to use `assert` or `require`?

agent/engine/docker_task_engine.go
```diff
@@ -636,18 +636,59 @@ func (engine *DockerTaskEngine) pullContainer(task *api.Task, container *api.Con
...
-
-	if engine.enableConcurrentPull {
-		seelog.Infof("Task engine [%s]: pulling container %s concurrently", task.Arn, container.Name)
-		return engine.concurrentPull(task, container)
-	}
-	seelog.Infof("Task engine [%s]: pulling container %s serially", task.Arn, container.Name)
-	return engine.serialPull(task, container)
+	if engine.imagePullRequired(engine.cfg.ImagePullBehavior, container, task.Arn) {
+		// Record the pullStoppedAt timestamp
+		defer func() {
```
richardpen: In this case what's the value of task PullStartedAt and pullStoppedAt?

agent/engine/image/types.go
```diff
@@ -111,18 +118,36 @@ func (imageState *ImageState) RemoveContainerReference(container *api.Container)
...
+func (imageState *ImageState) SetPullSucceeded(pullSucceeded bool) {
+	imageState.lock.Lock()
+	defer imageState.lock.Unlock()
+
+	imageState.PullSucceeded = pullSucceeded
+}
+
+// GetPullSucceeded safely returns the PullSucceeded of the imageState
+func (imageState *ImageState) GetPullSucceeded() bool {
+	imageState.lock.Lock()
```
richardpen: This should be a read lock.

README.md
haikuoliu: The README and env vars is reviewed by @joelbrandenburg. I think the image cleanup related stuff should be reflected in the tech doc, I will let @joelbrandenburg know about this.

agent/engine/docker_image_manager_integ_test.go
haikuoliu: ok

agent/engine/docker_task_engine.go
haikuoliu: I didn't quite get the question, the logic in `imagePullRequired ` is to pull image, instead of using local cache, hence recording `PullStartedAt` and `pullStoppedAt ` is necessary. (previously it's shouldUseLocalImage, and this is changed to `imagePullRequired` as per @aaithal's comments).

richardpen: So if `engine.imagePullRequired` is false, then the image pullstartedAt and pullStoppedAt won't be set, right? Can you run a task in this scenario and check the pullstoppedAt and pullstartedat from describetask API that it isn't set to something unexpected?

README.md
sharanyad: It would be good to specify that the cached image <i>in the instance</i> will be used.

agent/engine/docker_task_engine.go
haikuoliu: yeah I see what you mean, I saw the comment here: https://github.com/aws/amazon-ecs-agent/blob/159ae5c3067e0a0d7387ff68af2a28bc90f3baaa/agent/api/task.go#L112
And I test the describe task API (which you have seen), it works as expected, the `pull started at` and `pull stopped at` will now show up.

agent/engine/docker_image_manager.go
sharanyad: I think it might be valuable to add a link to the docs in `README`.

agent/engine/docker_task_engine.go
sharanyad: > the pull started at and pull stopped at will now show up.

If image is not pulled, these values shouldn't show up right?

agent/engine/task_manager.go
```diff
@@ -543,10 +543,26 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
...
-		// not be fatal if e.g. the image already exists.
-		seelog.Errorf("Managed task [%s]: Error while pulling container %s, will try to run anyway: %v",
-			mtask.Arn, container.Name, event.Error)
+		// If the agent pull behavior is always or once, we receive the error because
+		// the image pull fails, the task should fail. If we don't fail task here,
+		// then the cached image will probably be used for creating container, and we
+		// don't want to use cached image for both cases.
+		if mtask.cfg.ImagePullBehavior == config.ImagePullAlwaysBehavior ||
+			mtask.cfg.ImagePullBehavior == config.ImagePullOnceBehavior {
+			seelog.Errorf("Managed task [%s]: Error while pulling image %s for container %s , moving task to STOPPED: %v",
```
sharanyad: nit: `Managed task [%s]: error while`
same for below

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## 1.17.4-dev
+* Feature - Configurable agent pull behavior [#1348](https://github.com/aws/amazon-ecs-agent/pull/1348)
```
sharanyad: Could you please change this to `Configurable container image pull behavior`

agent/engine/docker_task_engine.go
haikuoliu: yeah it's not show up

README.md
haikuoliu: OK

agent/engine/task_manager.go
haikuoliu: OK

CHANGELOG.md
haikuoliu: OK

haikuoliu:
@sharanyad I will ask @joelbrandenburg if it's ok to have a link in readme. I will attach the link in readme if he has the link ready before I merge, otherwise I will attach it some time after I merge.

agent/engine/docker_image_manager.go
haikuoliu: I will ask @joelbrandenburg if it's ok to have a link in readme. I will attach the link in readme if he has the link ready before I merge, otherwise I will attach it some time after I merge.

README.md
richardpen: It should also be reflected here, otherwise the description here is not correct and could cause confusion.

agent/engine/task_manager.go
```diff
@@ -543,10 +543,26 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
...
+		if mtask.cfg.ImagePullBehavior == config.ImagePullAlwaysBehavior ||
+			mtask.cfg.ImagePullBehavior == config.ImagePullOnceBehavior {
+			seelog.Errorf("Managed task [%s]: error while pulling image %s for container %s , moving task to STOPPED: %v",
+				mtask.Arn, container.Image, container.Name, event.Error)
+			// The task should be stopped regardless of whether this container is
+			// essential or non-essential.
+			mtask.SetDesiredStatus(api.TaskStopped)
+			return false
+		}
+		// If the agent pull behavior is prefer_cached, we receive the error because
```
sharanyad: nit: `prefer-cached`

PR #1356, opened 2018-04-20
Fixing test failure on mac

yunhee-l:
### Summary
On mac, make test fails with: 
`agent/engine/docker_task_engine_unix_test.go:32:2: build constraints exclude all Go files in /Users/yhlee/go/src/github.com/aws/amazon-ecs-agent/agent/resources/cgroup/mock_control`

### Implementation details
Separated cgroup specific tests to build with linux tag.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine_unix_test.go
+++ b/agent/engine/docker_task_engine_unix_test.go
@@ -17,28 +17,16 @@ package engine
 import (
 	"context"
 	"errors"
-	"fmt"
-	"strings"
-	"sync"
 	"testing"
-	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
 	"github.com/aws/amazon-ecs-agent/agent/emptyvolume"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate/mocks"
-	"github.com/aws/amazon-ecs-agent/agent/engine/testdata"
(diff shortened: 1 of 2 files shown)


agent/engine/docker_task_engine_test.go
```diff
@@ -63,6 +63,13 @@ const (
 	exitCode                    = 1
 	labelsTaskARN               = "arn:aws:ecs:us-east-1:012345678910:task/c09f0188-7f87-4b0f-bfc3-16296622b6fe"
 	taskSteadyStatePollInterval = time.Millisecond
+
+	// dockerVersionCheckDuringInit specifies if Docker client's Version()
+	// API needs to be mocked in engine tests
+	//
+	// isParallelPullCompatible is not invoked during engin intialization
+	// on windows. No need for mock Docker client's Version() call
+	dockerVersionCheckDuringInit = true
```
sharanyad: In windows this value is false, wondering if windows will end up using the value here instead from `_windows_test.go` file.

yunhee-l: I've changed this to a var instead of const, given that we use different value for platforms - shadowing should handle it, although _windows_test.go itself doesn't actually reference it.

agent/engine/docker_task_engine_unix_test.go
```diff
@@ -1,4 +1,4 @@
-// +build !windows,!integration
+// +build linux,!integration
```
petderek: I think there is still value in running these tests on OSX. It might be worth only importing the linux-specific code in a different test file, like `docker_task_engine_linux_test.go`

PR #1357, opened 2018-04-23
adding cgroup as task resource

sharanyad:
### Summary
This PR initializes cgroup as a task resource and gets added to task and as resource dependency to the containers. Still a WIP (tests to be fixed and added).

### Implementation details
- agent uses cgroup controller directly to initialize the root cgroup `/ecs` for the instance. 
- previous cgroup implementation's create and delete function calls are removed since these are now taken care by `progressTask`.
- In `PostUnmarshalTask`, cgroup resource is initialized and added as resource dependency to the task's containers. 

passing cgroup controller object to resource creation:
In the PR, a new structure is introduced called `ResourceFields` which wraps cgroup controller created in agent/engine that is also required for cgroup resource creation in linux. This way, its easier to test by mocking out controller. 
Other ways to do this and why I was skeptical to use them: 
- pass the cgroup control object directly all the way from agent -> task engine -> task -> cgroup resource.
pros: no need for a new struct.
cons: control would now be part of everything though this is meaningless in windows, since they do not support cgroups. `_unsupported.go` files need to be added to support this.
- create `cgroup` task resource in agent.go and pass that to task through engine. 
pro: cgroup control need not be exposed to agent/engine. 
con: other things are needed to create the cgroup resource such as taskArn etc , which is not available at this point. `cgroup` resource object should ideally be created in `PostUnmarshalTask`. this is also useless in windows.
Open to discussion.

### Testing
- [x] Manual testing - 
With this set of changes, a task was placed on the instance. Resource cgroup is first created, and containers do not progress until then due to dependency. Containers then progress and after task is stopped, cgroup is removed later. 

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/errors/errors.go
+++ b/agent/api/errors/errors.go
@@ -14,6 +14,7 @@
 package errors
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -95,3 +96,25 @@ func (err *DockerClientConfigError) Error() string { return err.Msg }
 
 // ErrorName returns the name of the error
 func (err *DockerClientConfigError) ErrorName() string { return "DockerClientConfigError" }
+
+// ResourceInitError is a task error for which a required resource cannot
+// be initialized
+type ResourceInitError struct {
+	taskARN string
+	origErr error
+}
+
+// NewResourceInitError creates an error for resource initialize failure
+func NewResourceInitError(taskARN string, origErr error) *ResourceInitError {
+	return &ResourceInitError{taskARN, origErr}
+}
+
+// Error returns the error as a string
+func (err *ResourceInitError) Error() string {
+	return fmt.Sprintf("resource cannot be initialized for task %s: %v", err.taskARN, err.origErr)
+}
+
+// ErrorName is the name of the error
+func (err *ResourceInitError) ErrorName() string {
+	return "ResourceInitializationError"
+}
--- a/agent/api/task.go
+++ b/agent/api/task.go
@@ -187,13 +187,22 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task, 
 // PostUnmarshalTask is run after a task has been unmarshalled, but before it has been
 // run. It is possible it will be subsequently called after that and should be
 // able to handle such an occurrence appropriately (e.g. behave idempotently).
-func (task *Task) PostUnmarshalTask(cfg *config.Config, credentialsManager credentials.Manager) {
+func (task *Task) PostUnmarshalTask(cfg *config.Config,
+	credentialsManager credentials.Manager, resourceFields *taskresource.ResourceFields) error {
 	// TODO, add rudimentary plugin support and call any plugins that want to
 	// hook into this
 	task.adjustForPlatform(cfg)
+	if task.MemoryCPULimitsEnabled {
+		err := task.initializeCgroupResourceSpec(cfg.CgroupPath, resourceFields)
+		if err != nil {
+			seelog.Errorf("Task [%s]: could not intialize resource: %v", task.Arn, err)
+			return apierrors.NewResourceInitError(task.Arn, err)
+		}
+	}
 	task.initializeEmptyVolumes()
 	task.initializeCredentialsEndpoint(credentialsManager)
 	task.addNetworkResourceProvisioningDependency(cfg)
+	return nil
 }
 
 func (task *Task) initializeEmptyVolumes() {
--- a/agent/api/task_unix.go
+++ b/agent/api/task_unix.go
@@ -19,7 +19,10 @@ import (
 	"path/filepath"
 	"time"
 
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/cgroup"
 	"github.com/cihub/seelog"
 	docker "github.com/fsouza/go-dockerclient"
 	specs "github.com/opencontainers/runtime-spec/specs-go"
@@ -42,12 +45,32 @@ const (
 )
 
 // platformFields consists of fields specific to Linux for a task
-type platformFields struct {}
+type platformFields struct{}
 
 func (task *Task) adjustForPlatform(cfg *config.Config) {
 	task.lock.Lock()
 	defer task.lock.Unlock()
 	task.MemoryCPULimitsEnabled = cfg.TaskCPUMemLimit.Enabled()
+}
+
+func (task *Task) initializeCgroupResourceSpec(cgroupPath string, resourceFields *taskresource.ResourceFields) error {
+	cgroupRoot, err := task.BuildCgroupRoot()
+	if err != nil {
+		return errors.Wrapf(err, "cgroup resource: unable to determine cgroup root for task")
+	}
+	resSpec, err := task.BuildLinuxResourceSpec()
+	if err != nil {
+		return errors.Wrapf(err, "cgroup resource: unable to build resource spec for task")
+	}
+	cgroupResource := cgroup.NewCgroupResource(task.Arn, resourceFields.Control,
+		resourceFields.IOUtil, cgroupRoot, cgroupPath, resSpec)
+	task.Resources = append(task.Resources, cgroupResource)
+	for _, container := range task.Containers {
+		container.BuildResourceDependency(cgroupResource.GetName(),
+			taskresource.ResourceStatus(cgroup.CgroupCreated),
+			apicontainer.ContainerPulled)
+	}
+	return nil
 }
 
 func getCanonicalPath(path string) string { return path }
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -40,11 +40,11 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/eventstream"
 	"github.com/aws/amazon-ecs-agent/agent/handlers"
 	"github.com/aws/amazon-ecs-agent/agent/handlers/taskmetadata"
-	"github.com/aws/amazon-ecs-agent/agent/resources"
 	"github.com/aws/amazon-ecs-agent/agent/sighandlers"
 	"github.com/aws/amazon-ecs-agent/agent/sighandlers/exitcodes"
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
 	"github.com/aws/amazon-ecs-agent/agent/stats"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	"github.com/aws/amazon-ecs-agent/agent/tcs/handler"
 	"github.com/aws/amazon-ecs-agent/agent/utils/mobypkgwrapper"
 	"github.com/aws/amazon-ecs-agent/agent/version"
@@ -104,9 +104,9 @@ type ecsAgent struct {
 	subnet                string
 	mac                   string
 	metadataManager       containermetadata.Manager
-	resource              resources.Resource
 	terminationHandler    sighandlers.TerminationHandler
 	mobyPlugins           mobypkgwrapper.Plugins
+	resourceFields        *taskresource.ResourceFields
 }
 
 // newAgent returns a new ecsAgent object, but does not start anything
@@ -168,7 +168,6 @@ func newAgent(
 		}),
 		os:                 oswrapper.New(),
 		metadataManager:    metadataManager,
-		resource:           resources.New(),
 		terminationHandler: sighandlers.StartDefaultTerminationHandler,
 		mobyPlugins:        mobypkgwrapper.NewPlugins(),
 	}, nil
@@ -202,6 +201,9 @@ func (agent *ecsAgent) start() int {
 	imageManager := engine.NewImageManager(agent.cfg, agent.dockerClient, state)
 	client := ecsclient.NewECSClient(agent.credentialProvider, agent.cfg, agent.ec2MetadataClient)
 
+	if agent.cfg.TaskCPUMemLimit.Enabled() {
+		agent.initializeResourceFields()
+	}
 	return agent.doStart(containerChangeEventStream, credentialsManager, state, imageManager, client)
 }
 
@@ -213,6 +215,14 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 	state dockerstate.TaskEngineState,
 	imageManager engine.ImageManager,
 	client api.ECSClient) int {
+
+	// Conditionally create '/ecs' cgroup root
+	if agent.cfg.TaskCPUMemLimit.Enabled() {
+		if err := agent.cgroupInit(); err != nil {
+			seelog.Criticalf("Unable to initialize cgroup root for ECS: %v", err)
+			return exitcodes.ExitTerminal
+		}
+	}
 
 	// Create the task engine
 	taskEngine, currentEC2InstanceID, err := agent.newTaskEngine(containerChangeEventStream,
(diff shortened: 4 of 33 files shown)


agent/app/agent.go
```diff
@@ -231,17 +231,8 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 
 	// Conditionally create '/ecs' cgroup root
 	if agent.cfg.TaskCPUMemLimit.Enabled() {
```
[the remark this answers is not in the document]
sharanyad: It is enabled by default only in linux. It is explicitly disabled in windows. Agent will be terminated only if the user enables cpu limit in windows.

agent/app/agent_test.go
```diff
@@ -156,48 +154,6 @@ func TestDoStartNewStateManagerError(t *testing.T) {
 	assert.Equal(t, exitcodes.ExitTerminal, exitCode)
 }
 
-func TestDoStartTaskLimitsFail(t *testing.T) {
```
sharanyad: This [test](https://github.com/aws/amazon-ecs-agent/pull/1357/files#diff-4e5ad8207820cdb7351cbdda04bfc94cR534) is same as this removed test here.

agent/api/task.go
```diff
@@ -186,13 +186,22 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
 // PostUnmarshalTask is run after a task has been unmarshalled, but before it has been
 // run. It is possible it will be subsequently called after that and should be
 // able to handle such an occurrence appropriately (e.g. behave idempotently).
-func (task *Task) PostUnmarshalTask(cfg *config.Config, credentialsManager credentials.Manager) {
+func (task *Task) PostUnmarshalTask(cfg *config.Config,
+	credentialsManager credentials.Manager, resourceFields taskresource.ResourceFields) error {
 	// TODO, add rudimentary plugin support and call any plugins that want to
 	// hook into this
 	task.adjustForPlatform(cfg)
+	if task.MemoryCPULimitsEnabled {
```
petderek: can this happen within 'adjustForPlatform' ?

agent/acs/update_handler/updater_test.go
```diff
@@ -435,7 +436,7 @@ func TestNewerUpdateMessages(t *testing.T) {
 
 	require.Equal(t, "newer-update-tar-data", writtenFile.String(), "incorrect data written")
 
-	taskEngine := engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, nil)
+	taskEngine := engine.NewTaskEngine(cfg, nil, nil, nil, nil, nil, nil, taskresource.ResourceFields{})
```
petderek: its really weird that this is the only item in this call that isn't a pointer. What's the reasoning here?

sharanyad: you're right. This should be a pointer. made necessary changes.

agent/api/task.go
sharanyad: Made that a separate method to return a particular error type `ResourceInitializeError` when err from `initializeCgroupResource` is not nil.

agent/api/errors/errors.go
```diff
@@ -95,3 +95,18 @@ func (err *DockerClientConfigError) Error() string { return err.Msg }
 
 // ErrorName returns the name of the error
 func (err *DockerClientConfigError) ErrorName() string { return "DockerClientConfigError" }
+
+// ResourceInitializeError is a task error for which a required resource cannot
+// be initialized
+type ResourceInitializeError struct {
+	TaskArn string
```
aaithal: minor: can this be just `taskARN`? You can have a `New()` method to initialize it instead.

agent/api/errors/errors.go
```diff
@@ -95,3 +95,18 @@ func (err *DockerClientConfigError) Error() string { return err.Msg }
 
 // ErrorName returns the name of the error
 func (err *DockerClientConfigError) ErrorName() string { return "DockerClientConfigError" }
+
+// ResourceInitializeError is a task error for which a required resource cannot
+// be initialized
+type ResourceInitializeError struct {
```
aaithal: nit: `ResourceInitError` is shorter

agent/api/task.go
```diff
@@ -186,13 +186,22 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
...
+func (task *Task) PostUnmarshalTask(cfg *config.Config,
+	credentialsManager credentials.Manager, resourceFields *taskresource.ResourceFields) error {
 	// TODO, add rudimentary plugin support and call any plugins that want to
 	// hook into this
 	task.adjustForPlatform(cfg)
+	if task.MemoryCPULimitsEnabled {
+		err := task.initializeCgroupResource(cfg.CgroupPath, resourceFields)
+		if err != nil {
+			seelog.Errorf("Task [%s]: could not intialize resource: %v", task.Arn, err)
+			return apierrors.ResourceInitializeError{task.Arn}
```
aaithal: We're loosing context of the original error here. Can you please use `errors.Wrapf()` or make sure that `ResourceInitializeError` has an `origError` or some other similar field so that we preserve it?

agent/api/task_unix.go
```diff
@@ -42,14 +44,34 @@ const (
...
-type platformFields struct {}
+type platformFields struct{}
 
 func (task *Task) adjustForPlatform(cfg *config.Config) {
 	task.lock.Lock()
 	defer task.lock.Unlock()
 	task.MemoryCPULimitsEnabled = cfg.TaskCPUMemLimit.Enabled()
 }
 
+func (task *Task) initializeCgroupResource(cgroupPath string, resourceFields *taskresource.ResourceFields) error {
```
aaithal: `initializeCgroupResource` makes it seem like we are actually initializing cgroup here. Are we actually doing that? If we are, this method shouldn't be called in `PostUnmarshalTask`. If we are not, can you please rename this to `injectCgroupResource` or something that conveys that meaning?

agent/app/agent_unix.go
```diff
@@ -183,3 +187,25 @@ func contains(capabilities []string, capability string) bool {
...
+}
+
+func (agent *ecsAgent) cgroupInit() error {
+	err := agent.resourceFields.Control.Init()
+	// When task CPU and memory limits are enabled, all tasks are placed
+	// under the '/ecs' cgroup root.
+	if err != nil {
+		if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+			seelog.Criticalf("Unable to setup '/ecs' cgroup: %v", err)
+			return errors.New("Unable to setup '/ecs' cgroup")
```
aaithal: This should be a `Wrapf()` as well?

agent/app/agent_unix.go
```diff
@@ -183,3 +187,25 @@ func contains(capabilities []string, capability string) bool {
 
 	return false
 }
+
+func (agent *ecsAgent) initializeResourceFields() {
```
aaithal: Why is this separate method needed? Why can't this be done conditionally, in `cgroupInit()`? Is it mostly for testing?

agent/app/agent_unix.go
```diff
@@ -183,3 +187,25 @@ func contains(capabilities []string, capability string) bool {
...
+		Control: cgroup.New(),
+		IOUtil:  ioutilwrapper.NewIOUtil(),
+	}
+}
+
+func (agent *ecsAgent) cgroupInit() error {
+	err := agent.resourceFields.Control.Init()
+	// When task CPU and memory limits are enabled, all tasks are placed
+	// under the '/ecs' cgroup root.
+	if err != nil {
```
aaithal: if you refactor like below, you can reduce an indentation level:

```
if err == nil {
 return nil
}
if gent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
 ..
}
..
return nil
```

agent/api/task_unix.go
sharanyad: This method just initializes the `CgroupResource` object. In`progressTask`, actual creation of cgroup happens by calling `Create()` on `CgroupResource` during resource state transitions. I am not sure if `inject` conveys the right meaning.

agent/app/agent_unix.go
sharanyad: Thats right, it is mainly for testing `doStart` where we mock cgroup control object for init failure.

agent/api/task_unix.go
aaithal: How about `initCgroupResourceSpec()`?

agent/app/agent_unix.go
aaithal: Can you please add documentation for this method explaining the same?

agent/api/errors/errors.go
```diff
@@ -95,3 +96,25 @@ func (err *DockerClientConfigError) Error() string { return err.Msg }
 
 // ErrorName returns the name of the error
 func (err *DockerClientConfigError) ErrorName() string { return "DockerClientConfigError" }
+
+// ResourceInitError is a task error for which a required resource cannot
+// be initialized
+type ResourceInitError struct {
+	taskArn string
```
aaithal: minor: taskARN

agent/api/errors/errors.go
```diff
@@ -95,3 +96,25 @@ func (err *DockerClientConfigError) Error() string { return err.Msg }
...
+	origErr error
+}
+
+// NewResourceInitError creates an error for resource initialize failure
+func NewResourceInitError(taskArn string, origErr error) *ResourceInitError {
```
aaithal: here too: `taksARN`

agent/app/agent.go
```diff
@@ -214,6 +216,14 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 	imageManager engine.ImageManager,
 	client api.ECSClient) int {
 
+	// Conditionally create '/ecs' cgroup root
+	if agent.cfg.TaskCPUMemLimit.Enabled() {
+		if err := agent.cgroupInit(); err != nil {
+			seelog.Criticalf("Cgroup init error: %v", err)
```
aaithal: Can you make this more descriptive and user friendly? Something on the lines `"Unable to initialize cgoup root for ECS: .."`

agent/app/agent.go
```diff
@@ -354,7 +349,7 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 		state.Reset()
 		// Reset taskEngine; all the other values are still default
 		return engine.NewTaskEngine(agent.cfg, agent.dockerClient, credentialsManager,
-			containerChangeEventStream, imageManager, state, agent.metadataManager, agent.resource), currentEC2InstanceID, nil
+			containerChangeEventStream, imageManager, state, agent.metadataManager, agent.resourceFields), currentEC2InstanceID, nil
```
aaithal: minor: part of this can probably be split into a new line

agent/app/agent_unix.go
```diff
@@ -183,3 +187,27 @@ func contains(capabilities []string, capability string) bool {
...
+// initializeResourceFields exists mainly for testing doStart() to use mock Control
+// object
+func (agent *ecsAgent) initializeResourceFields() {
+	agent.resourceFields = &taskresource.ResourceFields{
+		Control: cgroup.New(),
+		IOUtil:  ioutilwrapper.NewIOUtil(),
+	}
+}
+
+func (agent *ecsAgent) cgroupInit() error {
```
aaithal: Is there unit test coverage for this, especially the `ExplicitlyEnabled` path?

agent/engine/docker_task_engine.go
```diff
@@ -612,7 +605,14 @@ func (engine *DockerTaskEngine) StateChangeEvents() chan statechange.Event {
 
 // AddTask starts tracking a task
 func (engine *DockerTaskEngine) AddTask(task *api.Task) error {
-	task.PostUnmarshalTask(engine.cfg, engine.credentialsManager)
+	err := task.PostUnmarshalTask(engine.cfg, engine.credentialsManager, engine.resourceFields)
+	if err != nil {
+		seelog.Errorf("Task engine [%s]: unable to progress task: %v", task.Arn, err)
```
aaithal: `"unable to add task to the engine"` seems more appropriate here

agent/engine/docker_task_engine.go
```diff
@@ -612,7 +605,14 @@ func (engine *DockerTaskEngine) StateChangeEvents() chan statechange.Event {
...
+		task.SetKnownStatus(api.TaskStopped)
+		task.SetDesiredStatus(api.TaskStopped)
+		engine.emitTaskEvent(task, err.Error())
+		return nil
```
aaithal: I don't know what all the `return nil` statements in this method are achieving (it's been there for ever). Can you modify this method to not return any error?

agent/app/agent_unix.go
sharanyad: this is tested from parent method which is `doStart` for `ExplicitlyEnabled` for happy and error path. specifically [here](https://github.com/aws/amazon-ecs-agent/pull/1357/files/22aa850796dc440da28c32d264cd853fd622148b#diff-4e5ad8207820cdb7351cbdda04bfc94cR466).

aaithal: thanks for confirming!

agent/engine/docker_task_engine.go
sharanyad: Sure, how about doing it in a different PR? Changes seem to be in a lot of files including tests.

sharanyad: I can add a `TODO`.

agent/api/errors/errors.go
```diff
@@ -95,3 +96,25 @@ func (err *DockerClientConfigError) Error() string { return err.Msg }
...
+}
+
+// NewResourceInitError creates an error for resource initialize failure
+func NewResourceInitError(taskARN string, origErr error) *ResourceInitError {
+	return &ResourceInitError{taskARN, origErr}
+}
+
+// Error returns the error as a string
+func (err *ResourceInitError) Error() string {
+	return fmt.Sprintf("resource cannot be initialized for task %s: %v", err.taskARN, err.origErr)
```
yunhee-l: should this be done with log instead of fmt? or was this just for debugging?

sharanyad: It is used to return string with use of format specifiers.

agent/engine/docker_task_engine.go
aaithal: Since this is a refactor PR, I'd much rather do it in this one

agent/engine/docker_task_engine.go
```diff
@@ -611,8 +604,16 @@ func (engine *DockerTaskEngine) StateChangeEvents() chan statechange.Event {
...
+// TODO: no error is returned from this method. remove the return type
 func (engine *DockerTaskEngine) AddTask(task *api.Task) error {
-	task.PostUnmarshalTask(engine.cfg, engine.credentialsManager)
+	err := task.PostUnmarshalTask(engine.cfg, engine.credentialsManager, engine.resourceFields)
+	if err != nil {
+		seelog.Errorf("Task engine [%s]: unable to add task to the engine: %v", task.Arn, err)
+		task.SetKnownStatus(api.TaskStopped)
+		task.SetDesiredStatus(api.TaskStopped)
+		engine.emitTaskEvent(task, err.Error())
+		return nil
```
aaithal: Please modify this as per https://github.com/aws/amazon-ecs-agent/pull/1357#discussion_r184716456

agent/engine/docker_task_engine.go
```diff
@@ -611,8 +604,16 @@ func (engine *DockerTaskEngine) StateChangeEvents() chan statechange.Event {
 }
 
 // AddTask starts tracking a task
-func (engine *DockerTaskEngine) AddTask(task *api.Task) error {
-	task.PostUnmarshalTask(engine.cfg, engine.credentialsManager)
+// TODO: no error is returned from this method. remove the return type
```
yunhee-l: this comment can be removed

agent/taskresource/types_unix.go
```diff
@@ -0,0 +1,28 @@
...
+package taskresource
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/resources/cgroup"
+	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
+)
+
+// ResourceFields is the list of fields required for creation of task resources
+// obtained from engine
+type ResourceFields struct {
```
yunhee-l: This won't be needed for other task resources like volume. Is my understanding correct, that if the container doesn't need cgroup, this can be empty in ecsAgent ([here](https://github.com/aws/amazon-ecs-agent/pull/1357/files#diff-2149975f54c050dd31ac9a37f1ef63bcR109))?

sharanyad: yes, this won't be needed for volume, but can be used for other resources in the future.

agent/api/errors/errors.go
```diff
@@ -95,3 +96,25 @@ func (err *DockerClientConfigError) Error() string { return err.Msg }
...
+}
+
+// ErrorName is the name of the error
+func (err *ResourceInitError) ErrorName() string {
+	return "ResourceInitError"
```
aaithal: Since this is customer facing, this should still be "ResourceInitializationError"

agent/engine/docker_task_engine.go
```diff
@@ -632,13 +632,13 @@ func (engine *DockerTaskEngine) AddTask(task *api.Task) error {
...
 		}
-		return nil
+		return
 	}
 
 	// Update task
 	engine.updateTaskUnsafe(existingTask, task)
 
-	return nil
+	return
```
aaithal: please get rid of this line line

PR #1358, opened 2018-04-23
Add timeout for setting up pause container

richardpen:
### Summary

Added timeout for SetupNS.

### Testing

New tests cover the changes: 

### Licensing


agent/ecscni/plugin.go
```diff
@@ -69,7 +70,24 @@ func NewClient(cfg *Config) CNIClient {
...
+	var err error
+	setupDone := make(chan struct{})
+	go func() {
+		result, err = client.setupNS(cfg)
+		setupDone <- struct{}{}
+	}()
+
+	select {
+	case <-ctx.Done():
+		return nil, errors.Errorf("cni setup: setup the container namespace timedout")
```
sharanyad: this can just be `errors.New`? 
also, message can be `"cni: container namespace setup timed out"`

agent/engine/docker_task_engine.go
```diff
@@ -61,6 +61,7 @@ const (
 	labelTaskDefinitionFamily    = labelPrefix + "task-definition-family"
 	labelTaskDefinitionVersion   = labelPrefix + "task-definition-version"
 	labelCluster                 = labelPrefix + "cluster"
+	cniSetupTimeoutDuration      = 1 * time.Minute
```
sharanyad: nit: this can just be `cniSetupTimeout`

richardpen:
Based on the test, it seems this process is really fast on different instance <5s. But in the cni plugin itself, there are some retry and timeout(maximum is 30s) for retrieving instance metadata and ipv6 gateway, so I doubled it to 1mins.

agent/ecscni/plugin.go
```diff
@@ -69,7 +70,24 @@ func NewClient(cfg *Config) CNIClient {
 
 // SetupNS will set up the namespace of container, including create the bridge
 // and the veth pair, move the eni to container namespace, setup the routes
-func (client *cniClient) SetupNS(cfg *Config) (*current.Result, error) {
+func (client *cniClient) SetupNS(cfg *Config, ctx context.Context) (*current.Result, error) {
```
aaithal: Please make sure that `ctx` is the first argument here.

agent/ecscni/plugin.go
```diff
@@ -38,7 +39,7 @@ type CNIClient interface {
 	// Capabilities returns the capabilities supported by a plugin
 	Capabilities(string) ([]string, error)
 	// SetupNS sets up the namespace of container
-	SetupNS(*Config) (*current.Result, error)
+	SetupNS(*Config, context.Context) (*current.Result, error)
```
aaithal: Can you make sure that `CleanupNS()` also has a context wired?

agent/ecscni/plugin.go
```diff
@@ -69,7 +70,24 @@ func NewClient(cfg *Config) CNIClient {
 
 // SetupNS will set up the namespace of container, including create the bridge
 // and the veth pair, move the eni to container namespace, setup the routes
-func (client *cniClient) SetupNS(cfg *Config) (*current.Result, error) {
+func (client *cniClient) SetupNS(cfg *Config, ctx context.Context) (*current.Result, error) {
+	var result *current.Result
+	var err error
+	setupDone := make(chan struct{})
+	go func() {
+		result, err = client.setupNS(cfg)
```
aaithal: What happens if this actually times out to the plugin execution path? Will they still be invoked in an orphan process? Can you please add details about how you tested this? What the behavior is? Essentially, how safe is it to do this?

richardpen: If the timeout happens, agent will move the task to stopped but the plugin was still running until it wasn't stuck. Ideally we should stop the process, but we don't know the plugin process id, I have opened the issue in libcni for adding the functionality to stop the plugin process after a timeout. The test I did is having two awsvpc task, the first was timedout during cni setup, and the second task can still progress successfully.

aaithal: Should we try to invoke `CleanupNS()` in that scenario then (if `SetupNS()` times out invoke `CleanupNS()`)?

richardpen: The `CleanupNS` is already called after the `setupNS` timeout: when setupNS timeout, the task desired status was moved to stop, and agent will try to stop pause container at which time the `CleanupNS` is called.

aaithal: Can you please create an issue in the agent repo to track this as well (and link the cni issue to that)? It doesn't seem super optimal to just leave a process hanging if we decide to give up early on it.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.17.4-dev
+* Bug - Fixed a bug where task stuck in pending due to the cni setup process hangs [#1358](https://github.com/aws/amazon-ecs-agent/pull/1358)
```
aaithal: how about `Fixed a bug where tasks could get stuck waiting for execution of CNI scripts`?

agent/ecscni/plugin.go
```diff
@@ -69,7 +70,29 @@ func NewClient(cfg *Config) CNIClient {
...
+		result, err := client.setupNS(cfg)
+		response <- output{
+			result: result,
+			err:    err,
+		}
+	}(response)
+
+	select {
+	case <-ctx.Done():
+		return nil, errors.Wrap(ctx.Err(), "cni setup: container namespace setup failed")
```
aaithal: I think you have to check if `ctx.Err()` is non nil. I don't think you can blindly do it this way. On second thought, I don't really think you need that async routine at all. If you pass in timeout as an arg to this method, you can do it in a way that's similar to what I'm doing in https://github.com/aws/amazon-ecs-agent/pull/1363 and deal with it in a much better way.

agent/ecscni/plugin.go
```diff
@@ -110,7 +133,21 @@ func (client *cniClient) SetupNS(cfg *Config) (*current.Result, error) {
 
 // CleanupNS will clean up the container namespace, including remove the veth
 // pair and stop the dhclient
-func (client *cniClient) CleanupNS(cfg *Config) error {
+func (client *cniClient) CleanupNS(ctx context.Context, cfg *Config) error {
```
aaithal: Same comment about accepting timeout as a param and dealing with `ctx.Err()` as above.

agent/ecscni/plugin.go
richardpen: Based on the document [here](https://golang.org/pkg/context/#Context), I don't think check `ctx.Err()!=nil` is required.
```
// If Done is not yet closed, Err returns nil.
// If Done is closed, Err returns a non-nil error explaining why:
// Canceled if the context was canceled
// or DeadlineExceeded if the context's deadline passed.
// After Err returns a non-nil error, successive calls to Err return the same error.
```

>A much better model would be to create the derived context within this method

Yes, a timeout instead should be passed in instead of passing the context, will modify that.

aaithal: > Based on the document [here](https://golang.org/pkg/context/#Context), I don't think check `ctx.Err()!=nil` is required.

Thanks! make sense.

PR #1362, opened 2018-04-25
dep: add makefile target to standardize dep usage

adnxn:
### Summary

add Makefile target to track dep version used for dependency management for the repo

### Implementation details

added target for `get-dep`

New tests cover the changes: 

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,6 +288,23 @@ static-check: gocyclo
 
 get-deps: .get-deps-stamp
 
+
+PLATFORM:=$(shell uname -s)
+ifeq (${PLATFORM},Linux)
+		dep_arch=linux-386
+	else ifeq (${PLATFORM},Darwin)
+		dep_arch=darwin-386
+	endif
+
+DEP_VERSION=v0.4.1
+.PHONY: get-dep
+get-dep: bin/dep
+
+bin/dep:
+	mkdir -p ./bin
+	curl -L https://github.com/golang/dep/releases/download/$(DEP_VERSION)/dep-${dep_arch} -o ./bin/dep
+	chmod +x ./bin/dep
+
 clean:
 	# ensure docker is running and we can talk to it, abort if not:
 	docker ps > /dev/null
@@ -306,4 +323,5 @@ clean:
 	-rm -f .get-deps-stamp
 	-rm -f .builder-image-stamp
 	-rm -f .out-stamp
+	-rm -rf $(PWD)/bin
 
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ _bin/
 *-stamp
 /.idea/
 /misc/taskmetadata-validator/taskmetadata-validator
+/bin


Makefile
```diff
@@ -288,6 +288,17 @@ static-check: gocyclo
 
 get-deps: .get-deps-stamp
 
+DEP_VERSION=v0.4.1
+get-dep-linux:
```
yunhee-l: for my understanding really, but could this and -osx targets be added under get-deps-stamp? also, should these be marked phony?

adnxn: > added under get-deps-stamp?

i kept it separate since we want to vendor most of the `get-deps` items as dependencies using the dep tool. whereas the dep tool falls outside of that scope. 

> should these be marked phony?

good question. im not sure. 🤔

petderek: As written, yes they should be.

Makefile
```diff
@@ -288,6 +288,17 @@ static-check: gocyclo
...
+	mkdir -p ./bin
+	curl -L https://github.com/golang/dep/releases/download/$(DEP_VERSION)/dep-linux-386 -o ./bin/dep
+	chmod +x ./bin/dep
```
richardpen: Only downloads the execute file, but it's not added to the system PATH, can we take one more step further and run dep ensure here?

adnxn: > Only downloads the execute file, but it's not added to the system PATH,

why add it to the system path? my intention is for us to use this from within the repo and avoid any conflict with the existing version of dep that might be on the system. please let me know if im missing something though!

> can we take one more step further and run dep ensure here

i'm also not sure about this since `dep ensure` will try to install the project's dependencies and may change the state of the vendor directory. i feel this may create room for confusion. but again, let me know what you think.

Makefile
```diff
@@ -288,6 +288,19 @@ static-check: gocyclo
...
+.PHONY: get-dep-linux
+get-dep-linux:
+	mkdir -p ./bin
+	curl -L https://github.com/golang/dep/releases/download/$(DEP_VERSION)/dep-linux-386 -o ./bin/dep
+	chmod +x ./bin/dep
+
+.PHONY: get-dep-osx
+get-dep-osx:
+	mkdir -p ./bin
+	curl -L https://github.com/golang/dep/releases/download/$(DEP_VERSION)/dep-darwin-386 -o ./bin/dep
```
nmeyerhans: Is the only difference between `get-dep-linux` and `get-dep-osx` the inclusion of the software platform name in the source URL? If so, I'd suggest figuring out the platform at runtime and combining these into a single rule. Make the computer figure out the right source path rather than the developer.

nmeyerhans: Consider something like:
```
PLATFORM:=$(shell uname -s)
ifeq (${PLATFORM},Linux)
	dep_arch=linux-386
else ifeq (${PLATFORM},Darwin)
	dep_arch=darwin-386
endif


.PHONY: get-dep
get-dep:
	mkdir -p ./bin
	curl -s -L https://github.com/golang/dep/releases/download/$(DEP_VERSION)/dep-${dep_arch} -o ./bin/dep
	chmod +x ./bin/dep
```

Are Linux and OSX the only software platforms we care about?

Also, the `clean` rule should probably clean up after this change. The idea is that `clean` should restore the repo to a pristine state.

This rule should also be idempotent. That is, running `make get-dep ; make get-dep` should only download the dep binary once. The ".stamp" stuff you see in some other rules is a common way to implement this.

nmeyerhans: Yes, sorry, I missed the update to `clean`

adnxn:
updated to use single command to for both linux/osx and manually tested on linux and osx.

Issue #1331, opened 2018-04-05
ecs-agent 1.17.2 stop connection to AWS and start reporting expired credential to task


CyrilDevOps:
### Summary
One of our ecs-agent stop connecting to ecs and start giving expired credential to tasks running in docker

### Description
After 7 days one ecs-agnet stop connecting to ECS, and start giving expired credential
to tasks running in docker.
after speaking with aws support, we did a docker stop <ecs agent container id>
and docker start <ecs agent container id>, and the ecs-agent start working correctly again.

AWS wasn't able to explain why it stopped connecting back.

```
/var/log/ecs/ecs-agent.log.2018-04-05-01:2018-04-05T01:32:07Z [INFO] ACS Websocket connection closed for a valid reason
/var/log/ecs/ecs-agent.log.2018-04-05-01:2018-04-05T01:57:38Z [INFO] ACS Websocket connection closed for a valid reason
/var/log/ecs/ecs-agent.log.2018-04-05-02:2018-04-05T02:13:13Z [INFO] ACS Websocket connection closed for a valid reason
<--- docker stop docker start
/var/log/ecs/ecs-agent.log.2018-04-05-20:2018-04-05T20:06:23Z [INFO] ACS Websocket connection closed for a valid reason
```

### Expected Behavior
We want it to continue to connect to ecs and keep giving correct credential.

### Environment Details
docker 17.09
on centos 7.3.7611
Version : Amazon ECS Agent - v1.17.2 (edc3e260)

# Log (sorry it was on a prod cluster, can't get too many log)
```
/var/log/ecs/ecs-agent.log.2018-04-05-01:2018-04-05T01:32:07Z [INFO] ACS Websocket connection closed for a valid reason
/var/log/ecs/ecs-agent.log.2018-04-05-01:2018-04-05T01:57:38Z [INFO] ACS Websocket connection closed for a valid reason
/var/log/ecs/ecs-agent.log.2018-04-05-02:2018-04-05T02:13:13Z [INFO] ACS Websocket connection closed for a valid reason
<--- docker stop docker start
/var/log/ecs/ecs-agent.log.2018-04-05-20:2018-04-05T20:06:23Z [INFO] ACS Websocket connection closed for a valid reason
```

aaithal:
[...]

PR #1363, opened 2018-04-25
Add timeout for Docker Version API

aaithal:
### Summary

Add timeout for Docker Version API

### Implementation details

* b1dae3d37eba6eab9c835648afa7f096af1fb7c9 dockerclient: add timeout for Version API
    This change adds a timeout for Docker client's Version API.
    Additionally, the Docker version returned in saved in the task engine.
    This helps with avoiding additional unnecessarily calls to the Docker
    Version API everytime the agent connects to the backend. This should
    help with issues seen in #1331

    Finally, the dockerclient package has been refactored to extract all
    timeout related constants to dockerclient package from the dockerapi
    package to avoid circular dependencies.

* 63f065004f5bae922d710e45086a2d817a86718b updated go-dickerclient for VersionWithContext
    Updated go-dockerclient to get the VersionWithContext() method, which
    helps with adding timeouts against the Docker Version() API, which can
    occasionally timeout.

### Testing

New tests cover the changes: 

### Licensing


agent/engine/docker_task_engine.go
```diff
@@ -1126,7 +1131,18 @@ func (engine *DockerTaskEngine) State() dockerstate.TaskEngineState {
 
 // Version returns the underlying docker version.
 func (engine *DockerTaskEngine) Version() (string, error) {
-	return engine.client.Version()
+	if engine.dockerVersion != "" {
```
richardpen: Do you think we need a lock for this method? Also, do we need a retry for this?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -949,12 +930,15 @@ func (dg *dockerGoClient) KnownVersions() []dockerclient.DockerVersion {
...
+func (dg *dockerGoClient) Version(ctx context.Context, timeout time.Duration) (string, error) {
+	derivedCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
 	client, err := dg.dockerClient()
 	if err != nil {
 		return "", err
 	}
-	info, err := client.Version()
+	info, err := client.VersionWithContext(derivedCtx)
```
richardpen: Why are we not listening to the `ctx.Done` channel like the other Docker APIs?

agent/dockerclient/clientfactory/dockerclientfactory.go
```diff
@@ -136,15 +138,18 @@ func (f *factory) getClient(version dockerclient.DockerVersion) (dockeriface.Cli
...
+func findDockerVersions(ctx context.Context, endpoint string) map[dockerclient.DockerVersion]dockeriface.Client {
 	// if the client version returns a MinAPIVersion and APIVersion, then use it to return
 	// all the Docker clients between MinAPIVersion and APIVersion, else try pinging
 	// the clients in getKnownAPIVersions
 	var minAPIVersion, apiVersion string
 	// get a Docker client with the default supported version
 	client, err := newVersionedClient(endpoint, string(minDockerAPIVersion))
 	if err == nil {
-		clientVersion, err := client.Version()
+		derivedCtx, cancel := context.WithTimeout(ctx, dockerclient.VersionTimeout)
```
richardpen: Can't just call `client.Version` here?

agent/dockerclient/dockerapi/docker_client.go
aaithal: Other APIs require special error handling on timeouts. They all require the output from Docker APIs to be massaged in some way and are invoked in an internal async method. `Version()` doesn't need any such semantics.

agent/dockerclient/clientfactory/dockerclientfactory.go
aaithal: Because `Version()` is defined on `dockerGoClient`. This one is `dockeriface.Client`

agent/engine/docker_task_engine.go
aaithal: I don't think this needs a retry. Looking at the usage, there's a remote possibility of a race condition. I'll add a scoped down mutex just for this for now. We can expand it's scope to the whole of `DockerTaskEngine` if needed.

agent/Gopkg.lock
```diff
@@ -292,7 +292,6 @@
   name = "golang.org/x/net"
   packages = [
     "context",
-    "context/ctxhttp",
```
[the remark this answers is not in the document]
aaithal: > Out of this  PR: probably we can get rid of `golang.org/x/` dependency?

Not sure, there are many sub-packages that we have a dependency on. Which one in particular you want to get rid of?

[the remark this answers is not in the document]
richardpen: I think the `golang.org/x/net` can be removed.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.17.4-dev
+* Bug - Fixed a bug where Docker Version() API never returns by adding a timeout [#1363](https://github.com/aws/amazon-ecs-agent/pull/1363)
```
sharanyad: Should we put Version() in `` ?

agent/dockerclient/clientfactory/dockerclientfactory.go
sharanyad: Why are we not calling `dockerGoClient`'s `Version` instead?

CHANGELOG.md
aaithal: I don't think it matters. I can modify it if you really insist (prefer not to 😃 ).

agent/dockerclient/clientfactory/dockerclientfactory.go
aaithal: Because it's not there. `dockerGoClient` belongs to the `engine` package and is only used by `DockerTaskEngine`

PR #1364, opened 2018-04-25
moving container related files from api/ to api/container/

yunhee-l:
### Summary
Refactoring container related code under api package to its own sub package. Will do the same with task related code after this. Container, Task, and volume are very tightly coupled under api package, so refactoring it in pieces.

### Implementation details
1. Moved MountPoint and VolumeFrom from api/volume to api/container, these are likely to be moved into taskresource/volume later
2. Took TaskStatus function in containerstatus.go and ContainerStatus function in taskstatus.go into statusmapping.go
3. Moved following files from api/ to api/container/:
container.go
container_test.go
container_unix.go
container_windows.go
containerevent.go
containeroverrides.go
containeroverrides_test.go
containerstatus.go
containerstatus_test.go
containertype.go
containertype_test.go
port_binding.go
port_binding_test.go
transitiondependency.go
transitiondependency_test.go
transport.go
transport_test.go

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler_test.go
+++ b/agent/acs/handler/acs_handler_test.go
@@ -29,6 +29,7 @@ import (
 	"context"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	"github.com/aws/amazon-ecs-agent/agent/api/mocks"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	rolecredentials "github.com/aws/amazon-ecs-agent/agent/credentials"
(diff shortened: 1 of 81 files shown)


agent/acs/handler/acs_handler_test.go
```diff
@@ -29,6 +29,7 @@ import (
 	"context"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
+	c "github.com/aws/amazon-ecs-agent/agent/api/container"
```
sharanyad: How about doing something like `. "github.com/aws/amazon-ecs-agent/agent/api/container"`, so that we can reference the types directly here without using package name?

yunhee-l: we need to do similar refactoring for task and eni under api, so I don't want to give priority to container with . import reference.

PR #1371, opened 2018-04-30
Refactoring task and eni under api package to their own subpackages

yunhee-l:
### Implementation details
1. task.ContainerStatus function made into util method "MapToContainerStatus" instead of task as a receiver (to decouple container and task, and for consistency with "MapToTaskStatus" method in statusmapping.go
2. Moved all eni related code under api/ to api/eni/
3. Move all task related code under api/ to api/eni/

### Testing

New tests cover the changes: 

### Licensing


agent/api/task/statusmapping.go
```diff
@@ -42,9 +42,9 @@ func MapToTaskStatus(knownState apicontainer.ContainerStatus, steadyState apicon
 	return TaskStatusNone
 }
 
-// ContainerStatus maps the task status to the corresponding container status
-func (ts *TaskStatus) ContainerStatus(steadyState apicontainer.ContainerStatus) apicontainer.ContainerStatus {
-	switch *ts {
+// MapToContainerStatus maps the task status to the corresponding container status
```
sharanyad: Why is this not called on `TaskStatus` anymore?

yunhee-l: ContainerStatus had TaskStatus function, and TaskStatus had ContainerStatus function, making them impossible to separate out into 2 packages, because it's immediately a cyclic import. So I decoupled these functions from ContainerStatus and TaskStatus receivers, and then updated the name to be more explicit, that we are mapping *Status param to the other *Status.

sharanyad: sounds good. how about naming it `MapTaskToContainerStatus` to make it more descriptive(since we also have resource status now)?

sharanyad:
looks like the author in the commit needs to be amended

PR #1372, opened 2018-05-01
Adding more randomness to the state change batches timer

petderek:
### Summary

Currently, we call the SubmitTaskStateChange API every 20 seconds. If a large
number of agents all start at the same time within a cluster, this
synchronization can cause a cluster to hit throttle limits. This can slow
the cluster scale up process.

By introducing jitter into the timer, we can further distribute the API calls
and make the throttling less likely.

### Implementation details

This is implemented by changing the static 20 second ticker to tick randomly between 10 and 30 seconds.

### Testing

New tests cover the changes: 

Enhancement - Improving SubmitTaskStateChange performance in large clusters

### Licensing

### DIFF
--- a/agent/utils/ticker.go
+++ b/agent/utils/ticker.go
@@ -1,0 +1,52 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package utils
+
+import (
+	"context"
+	"math/rand"
+	"time"
+)
+
+// NewJitteredTicker works like a time.Ticker except with randomly distributed ticks
+// between start and end duration.
+func NewJitteredTicker(ctx context.Context, start, end time.Duration) <-chan time.Time {
+	ticker := make(chan time.Time, 1)
+
+	go func() {
+		defer close(ticker)
+
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+				time.Sleep(randomDuration(start, end))
+				sendNow(ticker)
+			}
+		}
+	}()
+
+	return ticker
+}
+
+func randomDuration(start, end time.Duration) time.Duration {
+	return time.Duration(start.Nanoseconds()+rand.Int63n(end.Nanoseconds()-start.Nanoseconds())) * time.Nanosecond
+}
+
+func sendNow(ticker chan<- time.Time) {
+	select {
+	case ticker <- time.Now():
+	default:
+	}
+}
--- a/agent/utils/ticker_test.go
+++ b/agent/utils/ticker_test.go
@@ -1,0 +1,78 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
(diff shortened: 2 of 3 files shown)


agent/utils/ticker.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"context"
+	"math/rand"
+	"time"
+)
+
+// JitteredTicker works like a time.Ticker except with randomly distributed ticks
+// between start and end duration.
+func JitteredTicker(ctx context.Context, start, end time.Duration) <-chan time.Time {
```
adnxn: nit, naming: can we name this `NewJitteredTicker`? since stdlib uses `NewTicker`

agent/utils/ticker_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"context"
+	"math/rand"
+	"testing"
+	"time"
+)
+
+const (
+	TheBestNumber = 28
```
adnxn: nit: is it though?

petderek: yes

adnxn:
@petderek: as per offline conversation, +1 for getting rid of outer `loop` thing.

agent/utils/ticker.go
```diff
@@ -0,0 +1,41 @@
...
+				sendNow(ticker)
+			}
+		}
+	}()
+
+	return ticker
+}
+
+func randomDuration(start, end time.Duration) time.Duration {
+	return time.Duration(start.Nanoseconds()+rand.Int63n(end.Nanoseconds()-start.Nanoseconds())) * time.Nanosecond
```
richardpen: Why `Nanoseconds` instead of `Milliseconds`? I don't think Nanoseconds is enough for the jitter, as it means that there could still be up to 1e9 requests within one second.

agent/utils/ticker_test.go
```diff
@@ -0,0 +1,66 @@
...
+	times := 0
+	for {
+		_, ok := <-mTicker
+		times++
+		if !ok {
+			break
+		}
+	}
+
+	if 10 > times || times > 100  {
```
richardpen: nit: can you change it to `times<10 || times > 100`?

agent/utils/ticker_test.go
```diff
@@ -0,0 +1,66 @@
...
+func TestRandomDuration(t *testing.T) {
+	// The best randomness is deterministic
+	rand.Seed(TheBestNumber)
+
+	start := 10 * time.Second
+	end := 20 * time.Second
+
+	for i := 0; i < 10000; i++ {
+		c := randomDuration(start, end)
+		if c < start || end < c {
```
richardpen: nit: can you change it to `c <start || c > end`?

agent/utils/ticker_test.go
```diff
@@ -0,0 +1,66 @@
...
+	for {
+		select {
+		case <-c:
+			times++
+		default:
+			break loop
+		}
+	}
+
+	if times != 1 {
```
richardpen: Can you just use `assert` package?

agent/utils/ticker_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+	for i := 0; i < 10000; i++ {
+		c := randomDuration(start, end)
+		if c < start || end < c {
+			t.Errorf("Generated a bad time! seconds: %v nanoseconds: %d", c, c.Nanoseconds())
+		}
+	}
+}
+
+func TestSendNowOnlyPublishesOneEvent(t *testing.T) {
```
richardpen: I don't think the name is correct, if there are multiple goroutine listening to the channel, it could publish to each of these goroutines. If I understand correctly, you are testing the SendNow is not blocked even there is no listener.

agent/utils/ticker_test.go
petderek: Sure! I rewrote this a couple times because i couldn't decide what looked best.. :smile:

agent/utils/ticker_test.go
petderek: Yes!!! I'll fix that

agent/utils/ticker.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2014-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
richardpen: nit: this date should be 2018

PR #1377, opened 2018-05-07
Cgroup Resource sudo integration test

sharanyad:
### Summary
Integration test for task's cgroup resource creation and deletion as sudo 

### Implementation details
- added new make target `run-sudo-tests` for running integration tests as sudo
- added tags `!sudo` for other test files since `-tags sudo` runs all tests without sudo build tags too. 
- separated out some common functions needed for integration tests into `common_integ_test.go` and `common_unix_integ_test.go`.

### Testing

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ misc/certs/ca-certificates.crt:
 	docker run "amazon/amazon-ecs-agent-cert-source:make" cat /etc/ssl/certs/ca-certificates.crt > misc/certs/ca-certificates.crt
 
 test:
-	. ./scripts/shared_env && go test -race -timeout=25s -v -cover $(shell go list ./agent/... | grep -v /vendor/)
+	. ./scripts/shared_env && go test -race -tags unit -timeout=25s -v -cover $(shell go list ./agent/... | grep -v /vendor/)
 
 test-silent:
 	. ./scripts/shared_env && go test -timeout=25s -cover $(shell go list ./agent/... | grep -v /vendor/)
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,7 @@
 # language governing permissions and limitations under the License.
 
 USERID=$(shell id -u)
+GO_EXECUTABLE=$(shell command -v go 2> /dev/null)
 
 .PHONY: all gobuild static docker release certs test clean netkitten test-registry run-functional-tests benchmark-test gogenerate run-integ-tests pause-container get-cni-sources cni-plugins test-artifacts
 
@@ -220,8 +221,11 @@ cni-plugins: get-cni-sources .out-stamp
 		"amazon/amazon-ecs-build-cniplugins:make"
 	@echo "Built amazon-ecs-cni-plugins successfully."
 
-run-integ-tests: test-registry gremlin container-health-check-image
+run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tests
 	. ./scripts/shared_env && go test -race -tags integration -timeout=7m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
+
+run-sudo-tests:
+	. ./scripts/shared_env && sudo -E ${GO_EXECUTABLE} test -race -tags sudo -timeout=1m -v ./agent/engine/...
 
 .PHONY: codebuild
 codebuild: get-deps test-artifacts .out-stamp
--- a/agent/acs/client/acs_client_test.go
+++ b/agent/acs/client/acs_client_test.go
@@ -1,4 +1,6 @@
-// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// +build unit
+
+// Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"). You may
 // not use this file except in compliance with the License. A copy of the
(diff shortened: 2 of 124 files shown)


agent/app/agent_integ_test.go
```diff
@@ -1,4 +1,4 @@
-// +build integration
+// +build !sudo,integration
```
petderek: ex: if sudo implies integration, this file doesn't need to be changed.

Makefile
```diff
@@ -220,9 +221,12 @@ cni-plugins: get-cni-sources .out-stamp
 		"amazon/amazon-ecs-build-cniplugins:make"
 	@echo "Built amazon-ecs-cni-plugins successfully."
 
-run-integ-tests: test-registry gremlin container-health-check-image
+run-integ-tests: test-registry gremlin container-health-check-image run-sudo-integ-tests
 	. ./scripts/shared_env && go test -race -tags integration -timeout=7m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
 
+run-sudo-integ-tests:
+	. ./scripts/shared_env && sudo -E ${GO_EXECUTABLE} test -race -tags 'sudo integration' -timeout=5m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
```
petderek: I think it simplifies the build logic if you only use a 'sudo' built target that implies integration. You wouldn't need to add the !sudo to every additional integ test.

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -0,0 +1,80 @@
+// +build sudo,linux,integration
```
petderek: ex: this file would just have 'sudo,linux' as build tags

Makefile
sharanyad: My concern about using just `sudo` was there may be different types of tests in the future that would need to run as `sudo`. Hence kept it to two tags. What do you think?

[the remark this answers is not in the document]
sharanyad: `-tags` only accepts space separated lists.

petderek: I had considered that as well, but if we are interacting with host/docker resources its basically an integ test by definition.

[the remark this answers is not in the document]
petderek: I had thought this same thing! The syntax is weirder than I expected.

agent/app/agent_capability_test.go
```diff
@@ -1,3 +1,4 @@
+// +build !integration
```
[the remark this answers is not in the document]
sharanyad: I had to add `!sudo` and `!integration` since `-tags 'sudo integration'` would run tests in files without the tags `sudo` or `integration` too, which is weird.

Makefile
```diff
@@ -220,9 +221,12 @@ cni-plugins: get-cni-sources .out-stamp
 		"amazon/amazon-ecs-build-cniplugins:make"
 	@echo "Built amazon-ecs-cni-plugins successfully."
 
-run-integ-tests: test-registry gremlin container-health-check-image
+run-integ-tests: test-registry gremlin container-health-check-image run-sudo-integ-tests
 	. ./scripts/shared_env && go test -race -tags integration -timeout=7m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
 
+run-sudo-integ-tests:
+	. ./scripts/shared_env && sudo -E ${GO_EXECUTABLE} test -race -tags sudo -timeout=5m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
```
petderek: the only sudo test is in `./agent/engine`

You could get rid of the stats and app calls here, and then get rid of the !sudo tags in app and stats.

sharanyad: done

sharanyad:
> it will became more complicated if we are adding new flags in the future. So instead of using the negative tag !, can we just add tags to every test file? 

We would still need negative tags for the newly introduced tags. If the new tag is not present in a file, then go build assumes that it is present by default, which is why we would need the `!` flag. Does that make sense?

petderek:
> We don't run unit tests via integration tag today right?

We explicitly avoid that by using the `!integration` tag. `go test` would try to run them otherwise. If we use `unit`, we no longer need the `!integration` tag because its implied.

sharanyad:
Update: new changes include adding 'unit' build tag for unit test files, to avoid adding `!` tags when new ones are introduced. 
Changes: added 'unit' build tag, updated copyright year in the headers.

Makefile
```diff
@@ -220,9 +221,12 @@ cni-plugins: get-cni-sources .out-stamp
 		"amazon/amazon-ecs-build-cniplugins:make"
 	@echo "Built amazon-ecs-cni-plugins successfully."
 
-run-integ-tests: test-registry gremlin container-health-check-image
+run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tests
```
petderek: I'd prefer to keep the sudo business separated

agent/acs/client/acs_client_test.go
```diff
@@ -1,4 +1,5 @@
-// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// +build unit
```
petderek: > To distinguish build constraints from package documentation, a series of build constraints must be followed by a blank line.

I'm not sure if this is enforced? But we might want to change it.

https://golang.org/pkg/go/build/#hdr-Build_Constraints

Makefile
sharanyad: Since sudo tests are nothing but integ tests, I opted to keep it as part of `run-integ-tests`.

agent/acs/client/acs_client_test.go
sharanyad: Looks like this is not enforced. But better to change this.

PR #1384, opened 2018-05-14
api, dockerauth: move registryauth to api pkg, decouple from ecr

adnxn:
### Summary
This PR includes the following changes:
1. Vendor secretsmanager dependencies.
2. Consume acs model changes to include `ASMAuthData`, which is a new field in `RegistryAuthenticationData`. Ran gogenerate.
3. Move `RegistryAuthenticationData` out of the `ecr` pkg and into the `api` pkg.
4. Move auth abstraction in `dockerauth` up one level. This change uses the `RegistryAuthenticationData` abstraction since we will be handling multiple auth data backends (ECR, AWS Secrets Manager). Previously we were making the assumption that only ECR was going to be used for off instance registry authentication.

### Implementation details

n/a

### Testing

New tests cover the changes: 

### Licensing


agent/api/registryauth.go
```diff
@@ -11,7 +11,7 @@
 // express or implied. See the License for the specific language governing
 // permissions and limitations under the License.
 
-package ecr
+package api
```
samuelkarp: Can you keep this out of the main `api` package?  If you want it more in that path, I'd suggest something like `github.com/aws/amazon-ecs-agent/agent/api/auth` instead.

agent/dockerclient/dockerauth/ecr.go
```diff
@@ -65,9 +66,16 @@ func NewECRAuthProvider(ecrFactory ecr.ECRFactory, cache async.Cache) DockerAuth
 
 // GetAuthconfig retrieves the correct auth configuration for the given repository
 func (authProvider *ecrAuthProvider) GetAuthconfig(image string,
-	authData *ecr.ECRAuthData) (docker.AuthConfiguration, error) {
+	registryAuthData *api.RegistryAuthenticationData) (docker.AuthConfiguration, error) {
+
+	if registryAuthData == nil {
+		return docker.AuthConfiguration{}, fmt.Errorf("dockerauth: missing container's registry auth data")
```
samuelkarp: Should this be "container registry's"? Or just "missing container registry auth data" (no possessive)?

agent/api/registryauth.go
adnxn: discussed this with @yunhee-l, and since the task-resource branch does a heavy refactor of the api pkg and introduces subpackages, we decided it would be best to keep it in api for now.

agent/api/registryauth.go
```diff
@@ -36,6 +37,11 @@ type ECRAuthData struct {
 	lock             sync.RWMutex
 }
 
+type ASMAuthData struct {
+	CredentialsParameter string `json:"credentialsParameter"`
```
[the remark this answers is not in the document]
adnxn: hm so according to secrets manager the credentials the credentials parameter name is not sensitive data, we are to treat it as metadata like a tag or arn.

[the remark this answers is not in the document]
samuelkarp: Got it, I thought this was the actual credential values.  When you do implement it, you'll want to follow that pattern.

agent/api/registryauth.go
```diff
@@ -24,6 +24,7 @@ import (
 type RegistryAuthenticationData struct {
 	Type        string       `json:"type"`
 	ECRAuthData *ECRAuthData `json:"ecrAuthData"`
+	ASMAuthData *ASMAuthData `json:"asmAuthData"`
```
richardpen: Will this be unmarshalled from the payload message, can we modify the test to cover this?

adnxn: yes it will and completely missed this. will add test.

agent/api/registryauth.go
```diff
@@ -36,6 +37,11 @@ type ECRAuthData struct {
 	lock             sync.RWMutex
 }
 
+type ASMAuthData struct {
```
aaithal: lint: you're missing comments for a number of exported types and fields throughout. Please fix all of them

agent/resources/cgroup/factory/mock_factory/mock_cgroup_factory_linux.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2015-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sharanyad: `agent/resources` package is going away with PR https://github.com/aws/amazon-ecs-agent/pull/1382/. These changes can be excluded from this PR.

adnxn: i figured since these were generated, we can get rid of everything together during task-resources merge time.

PR #1403, opened 2018-06-03
unnnbreak my dev: return http request signing errors

adnxn:
### Summary
This change [unbreaks my dev](https://www.youtube.com/watch?v=wugSagirkBI). The aws sdk update now returns errors if `signer.Sign()` is not passed a `aws.ReadSeekCloser`, we used to pass in a `bytes.Buffer`, but since that doesn't implement a `Seek()` method, we were receiving an error from the sdk. Now we are instead passing in a `bytes.Reader`.

This change also returns errors thrown by the aws sdk when signing http requests and adds logic to pass the error up. We also make corresponding changes to tests to account for these new errors that are being surfaced. We had to swap out the AnonymousCredentials since they were passing in [empty strings](https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/credentials.go#L69) and causing the sdk's signing code paths to return errors about empty creds.

### Implementation details

🙃

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/tcs/client/client.go
+++ b/agent/tcs/client/client.go
@@ -114,19 +114,25 @@ func (cs *clientServer) MakeRequest(input interface{}) error {
 	}
 
 	seelog.Debugf("TCS client sending payload: %s", string(payload))
-	data := cs.signRequest(payload)
+	data, err := cs.signRequest(payload)
+	if err != nil {
+		return err
+	}
 
 	// Over the wire we send something like
 	// {"type":"AckRequest","message":{"messageId":"xyz"}}
 	return cs.WriteMessage(data)
 }
 
-func (cs *clientServer) signRequest(payload []byte) []byte {
-	reqBody := bytes.NewBuffer(payload)
+func (cs *clientServer) signRequest(payload []byte) ([]byte, error) {
+	reqBody := bytes.NewReader(payload)
 	// NewRequest never returns an error if the url parses and we just verified
 	// it did above
 	request, _ := http.NewRequest("GET", cs.URL, reqBody)
-	utils.SignHTTPRequest(request, cs.AgentConfig.AWSRegion, "ecs", cs.CredentialProvider, aws.ReadSeekCloser(reqBody))
+	err := utils.SignHTTPRequest(request, cs.AgentConfig.AWSRegion, "ecs", cs.CredentialProvider, aws.ReadSeekCloser(reqBody))
+	if err != nil {
+		return nil, err
+	}
 
 	request.Header.Add("Host", request.Host)
 	var dataBuffer bytes.Buffer
@@ -136,7 +142,7 @@ func (cs *clientServer) signRequest(payload []byte) []byte {
 	data := dataBuffer.Bytes()
 	data = append(data, payload...)
 
-	return data
+	return data, nil
 }
 
 // Close closes the underlying connection.
--- a/agent/tcs/client/client_test.go
+++ b/agent/tcs/client/client_test.go
@@ -46,6 +46,8 @@ const (
 	rwTimeout                  = time.Second
 )
 
+var testCreds = credentials.NewStaticCredentials("test-id", "test-secret", "test-token")
+
 type mockStatsEngine struct{}
 
 func (*mockStatsEngine) GetInstanceMetrics() (*ecstcs.MetricsMetadata, []*ecstcs.TaskMetric, error) {
@@ -167,12 +169,12 @@ func TestPublishMetricsRequest(t *testing.T) {
 
 	cs := testCS(conn)
 	defer cs.Close()
-
 	err := cs.MakeRequest(&ecstcs.PublishMetricsRequest{})
 	if err != nil {
 		t.Fatal(err)
 	}
 }
+
 func TestPublishMetricsOnceEmptyStatsError(t *testing.T) {
 	cs := clientServer{
 		statsEngine: &emptyStatsEngine{},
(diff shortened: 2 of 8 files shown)


agent/tcs/client/client_test.go
```diff
@@ -46,6 +46,8 @@ const (
 	rwTimeout                  = time.Second
 )
 
+var testCreds = credentials.NewStaticCredentials("test-id", "test-secret", "test-token")
```
fenxiong: Should we add a comment here like "Non-empty credentials should be used to avoid signing failure"? Otherwise it's not obvious that AnonymousCredentials can't be used in these tests..

adnxn: actually maybe ill just add it to the PR description. cause we cant add that comment to every test and if you try to use it in this context, the tests will fail for these code paths.

agent/tcs/client/client.go
```diff
@@ -114,19 +114,25 @@ func (cs *clientServer) MakeRequest(input interface{}) error {
 	}
 
 	seelog.Debugf("TCS client sending payload: %s", string(payload))
-	data := cs.signRequest(payload)
+	data, err := cs.signRequest(payload)
+	if err != nil {
+		return err
```
petderek: does this error get wrapped in the calling method?

agent/tcs/client/client.go
```diff
@@ -114,19 +114,25 @@ func (cs *clientServer) MakeRequest(input interface{}) error {
...
-	reqBody := bytes.NewBuffer(payload)
+func (cs *clientServer) signRequest(payload []byte) ([]byte, error) {
+	reqBody := bytes.NewReader(payload)
 	// NewRequest never returns an error if the url parses and we just verified
 	// it did above
 	request, _ := http.NewRequest("GET", cs.URL, reqBody)
-	utils.SignHTTPRequest(request, cs.AgentConfig.AWSRegion, "ecs", cs.CredentialProvider, aws.ReadSeekCloser(reqBody))
+	err := utils.SignHTTPRequest(request, cs.AgentConfig.AWSRegion, "ecs", cs.CredentialProvider, aws.ReadSeekCloser(reqBody))
+	if err != nil {
+		return nil, err
```
petderek: does this get wrapped in the calling method?

agent/tcs/client/client_test.go
```diff
@@ -237,12 +239,11 @@ func TestPublishOnceNonIdleStatsEngine(t *testing.T) {
 }
 
 func testCS(conn *mock_wsconn.MockWebsocketConn) wsclient.ClientServer {
-	testCreds := credentials.AnonymousCredentials
 	cfg := &config.Config{
 		AWSRegion:          "us-east-1",
 		AcceptInsecureCert: true,
 	}
-	cs := New("localhost:443", cfg, testCreds, &mockStatsEngine{},
+	cs := New("https://aws.amazon.com/ecs", cfg, testCreds, &mockStatsEngine{},
```
petderek: why?

agent/utils/sign_http_request.go
```diff
@@ -20,10 +20,16 @@ import (
...
 	"github.com/aws/aws-sdk-go/aws/signer/v4"
+	"github.com/cihub/seelog"
 )
 
 // SignHTTPRequest signs an http.Request struct with authv4 using the given region, service, and credentials.
-func SignHTTPRequest(req *http.Request, region, service string, creds *credentials.Credentials, body io.ReadSeeker) {
+func SignHTTPRequest(req *http.Request, region, service string, creds *credentials.Credentials, body io.ReadSeeker) error {
 	signer := v4.NewSigner(creds)
-	signer.Sign(req, body, service, region, time.Now())
+	_, err := signer.Sign(req, body, service, region, time.Now())
```
petderek: what does the signer return?

agent/wsclient/client.go
```diff
@@ -152,7 +152,10 @@ func (cs *ClientServerImpl) Connect() error {
 	request, _ := http.NewRequest("GET", parsedURL.String(), nil)
 
 	// Sign the request; we'll send its headers via the websocket client which includes the signature
-	utils.SignHTTPRequest(request, cs.AgentConfig.AWSRegion, ServiceName, cs.CredentialProvider, nil)
+	err = utils.SignHTTPRequest(request, cs.AgentConfig.AWSRegion, ServiceName, cs.CredentialProvider, nil)
+	if err != nil {
+		return err
```
petderek: does this get wrapped in the calling method?

agent/utils/sign_http_request.go
adnxn: it always returned the [header](https://github.com/aws/aws-sdk-go/blob/b2e866cc3d25372c6bde1ce18dbbf11862e954e9/aws/signer/v4/v4.go#L271)

agent/utils/sign_http_request.go
```diff
@@ -20,10 +20,16 @@ import (
...
+	if err != nil {
+		seelog.Warnf("Signing HTTP request failed: %v", err)
+		return err
```
aaithal: Echoing @petderek 's comments from above, can you please do `errors.Wrapf()` here?

adnxn:
@petderek, so i log a warn here at the [source of the sdk call](https://github.com/aws/amazon-ecs-agent/pull/1403/files#diff-fdf44a3eda83d2b41147433835c7675aR31). also adding errors.Wrapf() at the source of the call.

agent/tcs/client/client.go
adnxn: does now 😀

agent/utils/sign_http_request.go
```diff
@@ -20,10 +20,17 @@ import (
...
 
 // SignHTTPRequest signs an http.Request struct with authv4 using the given region, service, and credentials.
-func SignHTTPRequest(req *http.Request, region, service string, creds *credentials.Credentials, body io.ReadSeeker) {
+func SignHTTPRequest(req *http.Request, region, service string, creds *credentials.Credentials, body io.ReadSeeker) error {
 	signer := v4.NewSigner(creds)
-	signer.Sign(req, body, service, region, time.Now())
+	_, err := signer.Sign(req, body, service, region, time.Now())
+	if err != nil {
+		seelog.Warnf("Signing HTTP request failed: %v", err)
+		return errors.Wrapf(err, "aws sdk http signer: failed to sign http request")
```
aaithal: nit: `errors.Wrap()` should suffice here.

adnxn: oops.

agent/utils/sign_http_request.go
petderek: Why don't we need that?

haikuoliu:
Code lgtm!

agent/utils/sign_http_request.go
adnxn: looks like we pass in a pointer to the request and the signer does the `Request.Header.Set` things

PR #1406, opened 2018-06-05
Replace the empty container with a local volume

richardpen:
### Testing

New tests cover the changes: 
yes
### Licensing


agent/api/task/task.go
```diff
@@ -202,55 +206,53 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
-			Command:             []string{emptyvolume.Command}, // Command required, but this only gets created so N/A
-			MountPoints:         mountPoints,
-			Essential:           false,
-			Type:                apicontainer.ContainerEmptyHostVolume,
-			DesiredStatusUnsafe: apicontainer.ContainerRunning,
-		}
-		task.Containers = append(task.Containers, sourceContainer)
+	// if we have required local volumes, create one with default local drive
+	for _, volumeName := range requiredLocalVolumes {
+		vol, _ := task.HostVolumeByName(volumeName)
```
tehshwann: Im not sure how this code interacts with the Windows disk manager, so I dont have a good solution, but I am concerned that we arent using a unique identifier here.

What are we doing to ensure we are choosing the right volume and arent conflicting spaces between tasks?

agent/api/task/task.go
```diff
@@ -202,55 +206,53 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
+		// BUG(samuelkarp) On Windows, volumes with names that differ only by case will collide
+		scope := taskresourcevolume.TaskScope
+		autoprovision := false // must be false if scope = task
+		localDriver := "local"
+		localVolume := taskresourcevolume.NewVolumeResource(volumeName, vol.SourcePath(), scope, autoprovision, localDriver, make(map[string]string), make(map[string]string), dockerClient)
```
tehshwann: Does this driver work with Windows? If so, how are we excersizing this code path on Windows in testing?

agent/api/task/task.go
```diff
@@ -202,55 +206,53 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
+		task.AddResource(volumeName, localVolume)
 	}
 }
 
+func (task *Task) taskScopedVolumeName(name string) string {
```
tehshwann: does this need to be OS aware?

agent/api/task/task.go
```diff
@@ -347,7 +349,7 @@ func (task *Task) ContainerByName(name string) (*apicontainer.Container, bool) {
 
 // HostVolumeByName returns the task Volume for the given a volume name in that
 // task. The second return value indicates the presence of that volume
-func (task *Task) HostVolumeByName(name string) (HostVolume, bool) {
+func (task *Task) HostVolumeByName(name string) (taskresourcevolume.HostVolume, bool) {
```
tehshwann: Is the hostvolume something the user is volunteering to be used by the task? Or are we making that assumption for the user?

Concern:
It seems risky to make any assumptions related to disk space usage since there are limited tools, telemetry, and "rails" to manage disk utilization on the host to prevent disk space related issues.

agent/engine/engine_windows_integ_test.go
```diff
@@ -68,18 +72,14 @@ func createTestHostVolumeMountTask(tmpPath string) *apitask.Task {
...
 	testTask.Containers[0].Image = testVolumeImage
-	testTask.Containers[0].MountPoints = []apicontainer.MountPoint{{ContainerPath: "C:/empty", SourceVolume: "test-tmp"}}
-	testTask.Containers[0].Command = []string{`While($true){ if (Test-Path C:\empty\file) { exit 42 } }`}
-	testTask.Containers = append(testTask.Containers, createTestContainer())
-	testTask.Containers[1].Name = "test2"
-	testTask.Containers[1].Image = testVolumeImage
-	testTask.Containers[1].MountPoints = []apicontainer.MountPoint{{ContainerPath: "C:/alsoempty/", SourceVolume: "test-tmp"}}
-	testTask.Containers[1].Command = []string{`New-Item -Path C:\alsoempty\file`}
-	testTask.Containers[1].Essential = false
+	testTask.Containers[0].Command = []string{`echo -n empty-data-volume | Out-File -FilePath C:\host\tmp\hello-from-container -Encoding ascii`}
```
tehshwann: Im not sure what "echo -n" is... but echo is a cmd.exe that is aliased in PowerShell to "Write-Output" which doesnt have a "-n" arg.

This would be better powershell assuming you just want to print 'empty-data-volume':

`testTask.Containers[0].Command = []string{'Write-Output "empty-data-volume" | Out-File -FilePath C:\host\tmp\hello-from-container -Encoding ascii'}`

agent/engine/engine_windows_integ_test.go
```diff
@@ -68,18 +72,14 @@ func createTestHostVolumeMountTask(tmpPath string) *apitask.Task {
...
+	testTask.Containers[0].MountPoints = []apicontainer.MountPoint{{ContainerPath: "C:\\host\\tmp", SourceVolume: "test-tmp"}}
```
tehshwann: Also, "C:\host\tmp\" is fine for testing use. But a more traditional path for powershell/windows would be "$ENV:TEMP" for the current user's temp dir. I would consider using this so that running the tests dont create unnecessary and unconventional directories on the root volume

agent/engine/engine_windows_integ_test.go
```diff
@@ -127,6 +127,30 @@ func getContainerIP(client *docker.Client, id string) (string, error) {
...
+	verifyContainerStoppedStateChange(t, taskEngine)
+	verifyTaskIsStopped(stateChangeEvents, testTask)
+
+	assert.NotNil(t, testTask.Containers[0].GetKnownExitCode(), "No exit code found")
+	assert.Equal(t, 0, *testTask.Containers[0].GetKnownExitCode(), "Wrong exit code")
+
+	assert.Contains(t, testTask.Volumes[0].Volume.SourcePath(), "C:\\ProgramData\\docker\\volumes")
+	data, err := ioutil.ReadFile(filepath.Join(testTask.Volumes[0].Volume.SourcePath(), "hello-from-container"))
+	assert.Nil(t, err, "Unexpected error")
+	assert.Equal(t, "empty-data-volume", strings.TrimSpace(string(data)), "Incorrect file contents")
```
tehshwann: Im not sure how the Go libs work with ntfs here.. but will this opperate correctly if there is an EOF character?

agent/engine/task_manager_test.go
```diff
@@ -1622,6 +1623,255 @@ func TestApplyResourceStateFailures(t *testing.T) {
...
+				Task: &apitask.Task{},
+			}
+			transition := mtask.resourceNextState(&res)
+			assert.Equal(t, tc.NextState, transition.nextState)
+			assert.Equal(t, tc.ActionRequired, transition.actionRequired)
+		})
+	}
+}
+
+func TestStartVolumeResourceTransitionsHappyPath(t *testing.T) {
```
tehshwann: Should we be testing non-empty/non-happy path transitions? For example, if there are timeouts waiting for transition, or Docker has an issue handling the volume?

agent/taskresource/volume/dockervolume.go
```diff
@@ -53,15 +71,17 @@ type DockerVolumeConfig struct {
 
 // NewVolumeResource returns a docker volume wrapper object
 func NewVolumeResource(name string,
+	dockerVolumeName string,
```
tehshwann: Are these guaranteed to be unique? Does that matter to us?

agent/taskresource/volume/dockervolume.go
```diff
@@ -71,40 +91,124 @@ func NewVolumeResource(name string,
...
+// SteadyState returns the transition state of the resource defined as "ready"
+func (vol *VolumeResource) SteadyState() taskresource.ResourceStatus {
+	return taskresource.ResourceStatus(VolumeCreated)
+}
+
+// ApplyTransition calls the function required to move to the specified status
+func (vol *VolumeResource) ApplyTransition(nextState taskresource.ResourceStatus) error {
+	transitionFunc, ok := vol.resourceStatusToTransitionFunction[nextState]
+	if !ok {
+		seelog.Errorf("Volume Resource [%s]: unsupported desired state transition: %s",
```
tehshwann: Is this the right level of detail for failing to transition the state of a critical task resource?

tehshwann: How would failures here be debugged if, for example, docker was having an issue with the volume that bubbled up here and resulted in a timeout for a state transition?

agent/api/task/task.go
```diff
@@ -202,55 +206,53 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
 	task.initializeCredentialsEndpoint(credentialsManager)
 	task.addNetworkResourceProvisioningDependency(cfg)
 	return nil
 }
 
-func (task *Task) initializeEmptyVolumes() {
-	requiredEmptyVolumes := []string{}
+func (task *Task) initializeLocalVolumes(dockerClient dockerapi.DockerClient) {
+	requiredLocalVolumes := []string{}
+	reInvalidChars := regexp.MustCompile("[^A-Za-z0-9-]+")
```
aaithal: Why is this needed? Also, this should be a global var if needed.

agent/api/task/task.go
aaithal: Why are you ignoring the returned boolean here?

agent/api/task/task.go
```diff
@@ -202,55 +206,53 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
-			Type:                apicontainer.ContainerEmptyHostVolume,
-			DesiredStatusUnsafe: apicontainer.ContainerRunning,
-		}
-		task.Containers = append(task.Containers, sourceContainer)
+	// if we have required local volumes, create one with default local drive
+	for _, volumeName := range requiredLocalVolumes {
+		vol, _ := task.HostVolumeByName(volumeName)
+		// BUG(samuelkarp) On Windows, volumes with names that differ only by case will collide
+		scope := taskresourcevolume.TaskScope
+		autoprovision := false // must be false if scope = task
```
aaithal: %s/autoprovision/autoProvision/g

agent/api/task/task.go
```diff
@@ -202,55 +206,53 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
+		localDriver := "local"
```
aaithal: Please make this a const

agent/api/task/task.go
aaithal: please split this into multiple lines

agent/taskresource/volume/dockervolume.go
```diff
@@ -15,23 +15,41 @@ package volume
...
+	"github.com/pkg/errors"
 
 	"encoding/json"
 	"sync"
 	"time"
 )
 
-// VolumeResource represents volume resource
+const (
+	//TaskScope indicates that the volume is created and deleted with task
```
aaithal: there needs to be a space between '//' and the comment. Please fix this everywhere

agent/taskresource/volume/dockervolume.go
```diff
@@ -15,23 +15,41 @@ package volume
...
-	knownStatusUnsafe   VolumeStatus
-	client              dockerapi.DockerClient
+	desiredStatusUnsafe taskresource.ResourceStatus
+	knownStatusUnsafe   taskresource.ResourceStatus
+	// appliedStatus is the status that has been "applied" (e.g., we've called some
+	// operation such as 'Create' on the resource) but we don't yet know that the
+	// application was successful, which may then change the known status. This is
+	// used while progressing resource states in progressTask() of task manager
+	appliedStatus                      taskresource.ResourceStatus
+	resourceStatusToTransitionFunction map[taskresource.ResourceStatus]func() error
```
aaithal: `statusToTransitions` is a smaller and friendlier name for this.

agent/taskresource/volume/dockervolume.go
```diff
@@ -71,40 +91,124 @@ func NewVolumeResource(name string,
 		},
 		client: client,
 	}
+	v.initializeResourceStatusToTransitionFunction(scope)
+	return v
+}
+
+func (vol *VolumeResource) initializeResourceStatusToTransitionFunction(scope string) {
```
aaithal: Related to my previous comment, `initStatusToTransitions` is a smaller and friendlier name for this.

agent/taskresource/volume/dockervolume.go
aaithal: You don't need this log. It's already being logged in the engine

agent/api/task/task.go
```diff
@@ -202,55 +205,54 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
 		for _, mountPoint := range container.MountPoints {
 			vol, ok := task.HostVolumeByName(mountPoint.SourceVolume)
 			if !ok {
 				continue
 			}
-			if _, ok := vol.(*EmptyHostVolume); ok {
-				container.BuildContainerDependency(emptyHostVolumeName, apicontainer.ContainerRunning, apicontainer.ContainerCreated)
-				requiredEmptyVolumes = append(requiredEmptyVolumes, mountPoint.SourceVolume)
+			if localVolume, ok := vol.(*taskresourcevolume.LocalVolume); ok {
+				localVolume.HostPath = task.taskScopedVolumeName(mountPoint.SourceVolume)
```
aaithal: `task.taskScopedVolumeName` stutters. How about renaming it to just `VolumeName`?

agent/api/task/task.go
```diff
@@ -202,55 +205,54 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
-			Type:                apicontainer.ContainerEmptyHostVolume,
-			DesiredStatusUnsafe: apicontainer.ContainerRunning,
-		}
-		task.Containers = append(task.Containers, sourceContainer)
+	// if we have required local volumes, create one with default local drive
+	for _, volumeName := range requiredLocalVolumes {
+		vol, _ := task.HostVolumeByName(volumeName)
+		// BUG(samuelkarp) On Windows, volumes with names that differ only by case will collide
+		scope := taskresourcevolume.TaskScope
+		autoProvision := false // must be false if scope = task
```
aaithal: If this is a prereq, it should be enforced in the `NewVolumeResource`  method, where you need to check for that condition and return an error.

agent/taskresource/volume/dockervolume.go
```diff
@@ -15,22 +15,42 @@ package volume
...
+	// DockerLocalVolumeDriver is the name of the docker default volume driver
+	DockerLocalVolumeDriver = "local"
+)
+
+// VolumeResources represents volume resource
 type VolumeResource struct {
-	// Name is the name of docker volume
+	// Name is the name of the docker volume
 	Name string
+	// dockerVolumeName is internal docker name for this volume.
```
aaithal: nit: DockerVolumeName

agent/taskresource/volume/dockervolume.go
```diff
@@ -15,22 +15,42 @@ package volume
...
 	createdAtUnsafe     time.Time
-	desiredStatusUnsafe VolumeStatus
-	knownStatusUnsafe   VolumeStatus
+	desiredStatusUnsafe taskresource.ResourceStatus
+	knownStatusUnsafe   taskresource.ResourceStatus
+	// appliedStatus is the status that has been "applied" (e.g., we've called some
+	// operation such as 'Create' on the resource) but we don't yet know that the
+	// application was successful, which may then change the known status. This is
+	// used while progressing resource states in progressTask() of task manager
+	appliedStatus       taskresource.ResourceStatus
```
aaithal: Should this be `appliedStatusUnsafe` as well?

agent/taskresource/volume/dockervolume.go
```diff
@@ -71,40 +93,118 @@ func NewVolumeResource(name string,
...
+// SteadyState returns the transition state of the resource defined as "ready"
+func (vol *VolumeResource) SteadyState() taskresource.ResourceStatus {
+	return taskresource.ResourceStatus(VolumeCreated)
+}
+
+// ApplyTransition calls the function required to move to the specified status
+func (vol *VolumeResource) ApplyTransition(nextState taskresource.ResourceStatus) error {
+	transitionFunc, ok := vol.statusToTransitions[nextState]
+	if !ok {
+		return errors.Errorf("resource [%s]: transition to %s impossible", vol.Name,
```
aaithal: Why not begin the error string with 'volume' instead of 'resource' ?

agent/taskresource/volume/dockervolume.go
```diff
@@ -162,6 +267,12 @@ func (vol *VolumeResource) Create() error {
 
 // Cleanup performs resource cleanup
 func (vol *VolumeResource) Cleanup() error {
+	// Enable volume clean up if it's task scoped
+	if vol.VolumeConfig.Scope != TaskScope {
+		seelog.Debugf("Volume is shared, not removing", vol.Name)
```
aaithal: You're missing a %s here

agent/taskresource/volume/volumetypes.go
```diff
@@ -30,12 +30,14 @@ func (fs *FSHostVolume) SourcePath() string {
 	return fs.FSSourcePath
 }
 
-// EmptyHostVolume represents a volume without a specified host path
-type EmptyHostVolume struct {
+// LocalVolume represents a volume without a specified host path
+// This is essentially DockerVolume with only the name specified; however,
+// for backward compatibility we can't directly map to DockerVolume.
+type LocalVolume struct {
```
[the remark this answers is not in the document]
aaithal: This is still not addressed.

agent/taskresource/volume/dockervolume.go
```diff
@@ -15,22 +15,42 @@ package volume
...
 
 	"encoding/json"
 	"sync"
 	"time"
 )
 
-// VolumeResource represents volume resource
+const (
+	// TaskScope indicates that the volume is created and deleted with task
+	TaskScope = "task"
```
aaithal: It seems like both `TaskScope ` and `SharedScope` should be enums. wdyt?

richardpen: yes, I agree `enums` makes it looks better, but using constants here makes it simple to avoid conversion between string and enums, as this is what backend send to us. I think they works the same as enums is implemented based on constants, so I will keep this as it is unless you strongly want to change this.

agent/taskresource/volume/dockervolume.go
```diff
@@ -76,40 +102,124 @@ func NewVolumeResource(name string,
...
+	vol.client = resourceFields.DockerClient
+	vol.initStatusToTransitions()
+}
+
+func (vol *VolumeResource) initStatusToTransitions() {
+	statusToTransitions := map[resourcestatus.ResourceStatus]func() error{
+		resourcestatus.ResourceStatus(VolumeCreated): vol.Create,
+	}
+
+	statusToTransitions[resourcestatus.ResourceStatus(VolumeRemoved)] = vol.Cleanup
```
sharanyad: `Cleanup` need not be added to the status to transitions map since it is not called during `progressTask`. Resources cleanup happen here: 
https://github.com/aws/amazon-ecs-agent/blob/8dae73394282df16570d41c067fa364652db4e31/agent/engine/docker_task_engine.go#L464

agent/taskresource/volume/dockervolume.go
```diff
@@ -15,24 +15,44 @@ package volume
...
+	// TaskScope indicates that the volume is created and deleted with task
+	TaskScope = "task"
+	// SharedScope indicates that the volume's lifecycle is outside the scope of task
+	SharedScope = "shared"
+	// DockerLocalVolumeDriver is the name of the docker default volume driver
+	DockerLocalVolumeDriver = "local"
 )
 
-// VolumeResource represents volume resource
+// VolumeResources represents volume resource
```
sharanyad: nit: `VolumeResource`

agent/engine/docker_task_engine.go
```diff
@@ -323,11 +323,6 @@ func updateContainerMetadata(metadata *dockerapi.DockerContainerMetadata, contai
 		container.SetLabels(metadata.Labels)
 	}
 
-	// Update Volume
-	if metadata.Volumes != nil {
-		task.UpdateMountPoints(container, metadata.Volumes)
```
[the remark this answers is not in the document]
richardpen: This was previously used to update the mount point from the empty container, so removing this after the empty container was removed.

agent/taskresource/volume/dockervolume.go
richardpen: Yes, but it shouldn't hurt to add it here. As the `REMOVED` should match to the `Cleanup` in the transition function.

agent/api/task/task.go
yunhee-l: from moby issue #21829 (thought it might be relevant):
What is the volume size of the default volume?

Volumes created through the default ("local") volume driver don't have a size limit; they're basically just directories created inside /var/lib/docker/volumes. If you want to put a size constraint on volumes, you need to use a volume-plugin that supports this. Some plugins can be found in the documentation.

agent/engine/task_manager_test.go
```diff
@@ -1622,6 +1624,254 @@ func TestApplyResourceStateFailures(t *testing.T) {
 	}
 }
 
+///////
```
yunhee-l: nit: I had added this to mark on the file where new tests started, remove it?

agent/taskresource/volume/dockervolume.go
yunhee-l: Docker volume name for replacement "empty volume" are guaranteed to be unique (because we create it to force the scope to the running task). The Name for shared volume isn't, because it's used to refer to existing volume.

agent/api/task/task.go
```diff
@@ -202,53 +206,59 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
+	}
 	task.initializeCredentialsEndpoint(credentialsManager)
 	task.addNetworkResourceProvisioningDependency(cfg)
 	return nil
 }
 
-func (task *Task) initializeEmptyVolumes() {
-	requiredEmptyVolumes := []string{}
+func (task *Task) initializeDockerLocalVolumes(dockerClient dockerapi.DockerClient, ctx context.Context) error {
+	requiredLocalVolumes := []string{}
```
aaithal: nit: `var requiredLocalVolumes []string` is the idiomatic way of doing this i think

agent/taskresource/volume/dockervolume.go
```diff
@@ -56,16 +76,22 @@ type DockerVolumeConfig struct {
...
 func NewVolumeResource(name string,
+	dockerVolumeName string,
 	scope string,
 	autoprovision bool,
 	driver string,
 	driverOptions map[string]string,
 	labels map[string]string,
 	client dockerapi.DockerClient,
-	ctx context.Context) *VolumeResource {
+	ctx context.Context) (*VolumeResource, error) {
```
aaithal: `ctx` should be the first param to this method

agent/taskresource/volume/dockervolume.go
```diff
@@ -56,16 +76,22 @@ type DockerVolumeConfig struct {
...
 
-	return &VolumeResource{
-		Name: name,
+	if scope == TaskScope && autoprovision {
+		return nil, errors.New("Volume Resource creation: task scoped volume could not be autoprovisioned")
```
aaithal: error string doesn't follow conventions here.

agent/taskresource/volume/dockervolume_test.go
```diff
@@ -166,6 +221,46 @@ func TestUnmarshall(t *testing.T) {
...
+}
+
+func TestNewVolumeResource(t *testing.T) {
+	testCases := []struct {
+		scope         string
+		autoprovision bool
+		fail          bool
+	}{
+		{
+			"task",
```
aaithal: please add a line preceding each of these testcases with a description. you can also add a `name string` field to the struct and use that as well.

agent/taskresource/volume/dockervolume.go
sharanyad: Since it is not being used anywhere, it would be preferable to remove this.

agent/app/agent_unix.go
```diff
@@ -192,8 +192,10 @@ func contains(capabilities []string, capability string) bool {
 // object
 func (agent *ecsAgent) initializeResourceFields() {
 	agent.resourceFields = &taskresource.ResourceFields{
-		Control: cgroup.New(),
-		IOUtil:  ioutilwrapper.NewIOUtil(),
+		Control:      cgroup.New(),
```
sharanyad: Is it a concern that we initialize these fields irrespective of whether resources like cgroup/volumes are enabled or not?

richardpen: It's not a big concern to me, as it's just creating the object. This avoids an additional check, if you feel strongly about this, I can change it.

agent/taskresource/volume/dockervolume_test.go
```diff
@@ -113,17 +115,66 @@ func TestCleanupError(t *testing.T) {
...
+	volume.ApplyTransition(resourcestatus.ResourceStatus(VolumeCreated))
+}
+
+func TestApplyTransitionForSharedScopeVolume(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	mockClient := mock_dockerapi.NewMockDockerClient(ctrl)
+
+	name := "volumeName"
+	scope := "shared"
```
sharanyad: We do not cleanup shared volumes. Do we still expect `RemoveVolume` call here?

agent/taskresource/volume/dockervolume.go
```diff
@@ -146,12 +255,17 @@ func (vol *VolumeResource) GetMountPoint() string {
...
 }
 
+// SourcePath is fulfilling the HostVolume interface
+func (vol *VolumeResource) SourcePath() string {
+	return vol.GetMountPoint()
+}
+
 // Create performs resource creation
 func (vol *VolumeResource) Create() error {
 	seelog.Debugf("Creating volume with name %s using driver %s", vol.Name, vol.VolumeConfig.Driver)
```
sharanyad: nit: should this be `vol.DockerVolumeName`?

agent/taskresource/volume/dockervolume_test.go
richardpen: No, as you can see the times is zero. but I think we can just remove that `EXPECT`.

agent/taskresource/volume/dockervolume.go
richardpen: Yes, I will fix this in the other PR.

PR #1407, opened 2018-06-05
taskresource: add terminalReason msg string

adnxn:
### Summary

This change adds a terminalReason message to the task struct. The intended use is for the field in the task to be set when the agent explicitly transitions a task to stopped.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -153,6 +153,12 @@ type Task struct {
 	// platformFields consists of fields specific to linux/windows for a task
 	platformFields platformFields
 
+	// terminalReasonUnsafe should be used when we explicitly move a task to stopped.
+	// This ensures the task object carries some context for why it was explicitly
+	// stoppped.
+	terminalReasonUnsafe string
+	terminalReasonOnce   sync.Once
+
 	// lock is for protecting all fields in the task struct
 	lock sync.RWMutex
 }
@@ -1212,4 +1218,22 @@ func (task *Task) AddResource(resourceType string, resource taskresource.TaskRes
 	task.lock.Lock()
 	defer task.lock.Unlock()
 	task.ResourcesMapUnsafe[resourceType] = append(task.ResourcesMapUnsafe[resourceType], resource)
+}
+
+// SetTerminalReason sets the terminalReasonUnsafe string and this can only be set
+// once per the task's lifecycle. This field does not accept updates.
+func (task *Task) SetTerminalReason(reason string) {
+	seelog.Infof("Task [%s]: attempting to set terminal reason for task [%s]", task.Arn, reason)
+	task.terminalReasonOnce.Do(func() {
+		seelog.Infof("Task [%s]: setting terminal reason for task [%s]", task.Arn, reason)
+		task.terminalReasonUnsafe = reason
+	})
+}
+
+// GetTerminalReason retrieves the terminalReasonUnsafe string
+func (task *Task) GetTerminalReason() string {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	return task.terminalReasonUnsafe
 }
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -47,7 +47,6 @@ const (
 	stoppedSentWaitInterval               = 30 * time.Second
 	maxStoppedWaitTimes                   = 72 * time.Hour / stoppedSentWaitInterval
 	taskUnableToTransitionToStoppedReason = "TaskStateError: Agent could not progress task's state to stopped"
-	taskUnableToCreatePlatformResources   = "TaskStateError: Agent could not create task's platform resources"
 )
 
 var (
@@ -469,6 +468,7 @@ func (mtask *managedTask) handleResourceStateChange(resChange resourceStateChang
 		seelog.Errorf("Managed task [%s]: error while creating resource %s, setting the task's desired status to STOPPED",
 			mtask.Arn, res.GetName())
 		mtask.SetDesiredStatus(apitask.TaskStopped)
+		mtask.SetTerminalReason(res.GetTerminalReason())
 		mtask.engine.saver.Save()
 	}
 }
@@ -763,10 +763,20 @@ func (mtask *managedTask) progressTask() {
 	mtask.waitForTransition(transitions, transitionChange, transitionChangeEntity)
 	// update the task status
 	changed := mtask.UpdateStatus()
+
+	// If knownStatus changed, let it be known
+	mtask.emitEventIfStatusChanged(changed)
+}
+
+func (mtask *managedTask) emitEventIfStatusChanged(changed bool) {
 	if changed {
 		seelog.Debugf("Managed task [%s]: container or resource change also resulted in task change", mtask.Arn)
-		// If knownStatus changed, let it be known
-		mtask.emitTaskEvent(mtask.Task, "")
+
+		var taskStateChangeReason string
+		if mtask.GetKnownStatus().Terminal() {
+			taskStateChangeReason = mtask.GetTerminalReason()
+		}
+		mtask.emitTaskEvent(mtask.Task, taskStateChangeReason)
 	}
 }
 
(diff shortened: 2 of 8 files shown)


agent/api/task/task.go
```diff
@@ -153,6 +153,12 @@ type Task struct {
 	// platformFields consists of fields specific to linux/windows for a task
 	platformFields platformFields
 
+	// terminalReason should be used when we explicitly move a task to stopped.
+	// This ensures the task object carries some context for why it was explicitly
+	// stoppped.
+	terminalReason    string
+	termialReasonOnce sync.Once
```
sharanyad: nit: terminalReasonOnce

agent/api/task/task.go
```diff
@@ -1213,3 +1219,16 @@ func (task *Task) AddResource(resourceType string, resource taskresource.TaskRes
...
+// once per the task's lifecycle. This field does not accept updates.
+func (task *Task) SetTerminalReason(reason string) {
+	task.termialReasonOnce.Do(func() {
+		task.terminalReason = reason
+	})
+}
+
+// GetTerminalReason retrieves the terminalReason string
+func (task *Task) GetTerminalReason() string {
+	return task.terminalReason
```
sharanyad: does this require a mutex?

agent/engine/task_manager.go
```diff
@@ -765,8 +765,9 @@ func (mtask *managedTask) progressTask() {
 	changed := mtask.UpdateStatus()
 	if changed {
 		seelog.Debugf("Managed task [%s]: container or resource change also resulted in task change", mtask.Arn)
+
 		// If knownStatus changed, let it be known
-		mtask.emitTaskEvent(mtask.Task, "")
+		mtask.emitTaskEvent(mtask.Task, mtask.Task.GetTerminalReason())
```
sharanyad: Task's status change does not necessarily indicate terminal state. It would be good to check if the known status is `STOPPED` and if so, emit event with `terminalReason`.

adnxn: good catch.

agent/api/task/task.go
```diff
@@ -1213,3 +1219,19 @@ func (task *Task) AddResource(resourceType string, resource taskresource.TaskRes
 	defer task.lock.Unlock()
 	task.ResourcesMapUnsafe[resourceType] = append(task.ResourcesMapUnsafe[resourceType], resource)
 }
+
+// SetTerminalReason sets the terminalReason string and this can only be set
+// once per the task's lifecycle. This field does not accept updates.
+func (task *Task) SetTerminalReason(reason string) {
+	task.terminalReasonOnce.Do(func() {
```
aaithal: An Info log here stating that you're trying to set a task's terminal reason would be helpful in this first one wins scenario

agent/taskresource/cgroup/cgroup.go
```diff
@@ -32,10 +32,11 @@ import (
...
 const (
-	memorySubsystem         = "/memory"
-	memoryUseHierarchy      = "memory.use_hierarchy"
-	rootReadOnlyPermissions = os.FileMode(400)
-	resourceName            = "cgroup"
+	memorySubsystem           = "/memory"
+	memoryUseHierarchy        = "memory.use_hierarchy"
+	rootReadOnlyPermissions   = os.FileMode(400)
+	resourceName              = "cgroup"
+	resourceProvisioningError = "TaskResourceError: Agent could not create task's platform resources"
```
aaithal: :%s/TaskResourceError/CgroupError/ ?

agent/taskresource/volume/dockervolume.go
```diff
@@ -22,6 +22,8 @@ import (
 	"time"
 )
 
+const resourceProvisioningError = "TaskResourceError: Agent could not create task's volume resources"
```
aaithal: :%s/TaskResourceError/VolumeError/ ?

agent/api/task/task.go
adnxn: this is a good idea, adding.

agent/taskresource/cgroup/cgroup.go
adnxn: hm my thinking was it wouldn't be a customer resource, but bucketing by type is a good idea.

aaithal: "TaskResourceError" wouldn't make sense for a customer yet as they are not directly creating these resources. That is still an internal implementation detail, which is why I think "CgroupError" makes more sense in relaying what went wrong to customers in the `DescribeTasks` API response.

agent/engine/task_manager.go
```diff
@@ -765,8 +765,13 @@ func (mtask *managedTask) progressTask() {
 	changed := mtask.UpdateStatus()
 	if changed {
 		seelog.Debugf("Managed task [%s]: container or resource change also resulted in task change", mtask.Arn)
+
 		// If knownStatus changed, let it be known
-		mtask.emitTaskEvent(mtask.Task, "")
+		var taskStateChangeReason string
+		if mtask.GetKnownStatus().Terminal() {
+			taskStateChangeReason = mtask.Task.GetTerminalReason()
```
aaithal: Can you please add a unit test around this?

adnxn: right, gotcha

sharanyad: Can this just be `mtask.GetTerminalReason()`?

adnxn: yea you're right its an embedded type, we can just reach into the `Task` method

agent/api/task/task.go
```diff
@@ -1213,3 +1219,20 @@ func (task *Task) AddResource(resourceType string, resource taskresource.TaskRes
...
+		seelog.Infof("Task [%s]: setting terminal reason for task [%s]", task.Arn, reason)
```
aaithal: You also need one outside this block, which records that something is `trying` to set the reason for cases in which it might be ignored.

agent/api/task/task.go
```diff
@@ -153,6 +153,12 @@ type Task struct {
 	// platformFields consists of fields specific to linux/windows for a task
 	platformFields platformFields
 
+	// terminalReason should be used when we explicitly move a task to stopped.
+	// This ensures the task object carries some context for why it was explicitly
+	// stoppped.
+	terminalReason     string
```
aaithal: Please rename this as `terminalReasonUnsafe`

agent/engine/task_manager.go
```diff
@@ -765,8 +765,13 @@ func (mtask *managedTask) progressTask() {
 	changed := mtask.UpdateStatus()
 	if changed {
 		seelog.Debugf("Managed task [%s]: container or resource change also resulted in task change", mtask.Arn)
+
 		// If knownStatus changed, let it be known
-		mtask.emitTaskEvent(mtask.Task, "")
+		var taskStateChangeReason string
+		if mtask.GetKnownStatus().Terminal() {
```
aaithal: I still can't find a unit test for this. Am I missing something?

adnxn: no. still need to add!

agent/engine/task_manager.go
```diff
@@ -763,10 +763,20 @@ func (mtask *managedTask) progressTask() {
 	mtask.waitForTransition(transitions, transitionChange, transitionChangeEntity)
 	// update the task status
 	changed := mtask.UpdateStatus()
+
+	// If knownStatus changed, let it be known
+	mtask.emitEventIfStatusChanged(changed)
```
sharanyad: How about something like:
```
if changed {
 mtask.emitStatusChangeEvent()
}

func (mtask *managedTask) emitStatusChangeEvent() {
..
}
```

adnxn: there's already method for that exactly. i put the branching logic behind a method so we can test it.

sharanyad: The new branching logic that was added was the check for terminal status and what kind of reason was emitted, isn't it? Unit testing that should be fine I think.

agent/engine/task_manager_test.go
```diff
@@ -1564,6 +1564,32 @@ func TestWaitForResourceTransition(t *testing.T) {
 	task.waitForTransition(transitions, transition, transitionChangeResource)
 }
 
+func TestEmitEventIfStatusChanged(t *testing.T) {
```
sharanyad: Can you please add a test case for non terminal status change?

adnxn:
closing in favor of https://github.com/aws/amazon-ecs-agent/pull/1427
the commit is included as part of new pr.

PR #1424, opened 2018-06-27
Functional tests for task specific and shared Docker volumes

sharanyad:
### Summary
Functional tests for local and shared docker volumes for tasks

### Implementation details
Note: This PR cannot be merged until backend changes are in prod. 
TODO 
- [ ] Verify shared volumes not removed on task cleanup (only done manually for now)
- [ ] Shared, auto provisioned
- [ ] Windows tests

### Testing

### Licensing

### DIFF
--- a/agent/ecs_client/model/api/api-2.json
+++ b/agent/ecs_client/model/api/api-2.json
@@ -1036,6 +1036,16 @@
       "key":{"shape":"String"},
       "value":{"shape":"String"}
     },
+    "DockerVolumeConfiguration":{
+      "type":"structure",
+      "members":{
+        "scope":{"shape":"Scope"},
+        "autoprovision":{"shape":"BoxedBoolean"},
+        "driver":{"shape":"String"},
+        "driverOpts":{"shape":"StringMap"},
+        "labels":{"shape":"StringMap"}
+      }
+    },
     "Double":{"type":"double"},
     "EnvironmentVariables":{
       "type":"list",
@@ -1518,6 +1528,13 @@
         "failures":{"shape":"Failures"}
       }
     },
+    "Scope":{
+      "type":"string",
+      "enum":[
+        "task",
+        "shared"
+      ]
+    },
     "ServerException":{
       "type":"structure",
       "members":{
@@ -1641,6 +1658,11 @@
     "StringList":{
       "type":"list",
       "member":{"shape":"String"}
+    },
+    "StringMap":{
+      "type":"map",
+      "key":{"shape":"String"},
+      "value":{"shape":"String"}
     },
     "SubmitContainerStateChangeRequest":{
       "type":"structure",
@@ -1924,7 +1946,8 @@
       "type":"structure",
       "members":{
         "name":{"shape":"String"},
-        "host":{"shape":"HostVolumeProperties"}
+        "host":{"shape":"HostVolumeProperties"},
+        "dockerVolumeConfiguration":{"shape":"DockerVolumeConfiguration"}
       }
     },
     "VolumeFrom":{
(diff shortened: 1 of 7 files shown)


agent/ecs_client/model/api/api-2.json
```diff
@@ -679,6 +685,7 @@
       "members":{
         "name":{"shape":"String"},
         "image":{"shape":"String"},
+        "repositoryCredentials":{"shape":"RepositoryCredentials"},
```
yunhee-l: is this mixed with private registry work? if so, is this blocking and/or cause conflicts with any other work?

agent/ecs_client/model/api/api-2.json
```diff
@@ -807,7 +814,8 @@
         "placementConstraints":{"shape":"PlacementConstraints"},
         "placementStrategy":{"shape":"PlacementStrategies"},
         "networkConfiguration":{"shape":"NetworkConfiguration"},
-        "healthCheckGracePeriodSeconds":{"shape":"BoxedInteger"}
+        "healthCheckGracePeriodSeconds":{"shape":"BoxedInteger"},
+        "schedulingStrategy":{"shape":"SchedulingStrategy"}
```
yunhee-l: this doesn't seem related to volume plugin work

agent/ecs_client/model/api/api-2.json
adnxn: looks like this change is just bringing in the acs model changes. it shouldnt block and im expecting it to just resolve itself during merge.

agent/ecs_client/model/api/api-2.json
adnxn: are we expecting to bring these in already? specifically `schedulingStrategy`?

agent/ecs_client/model/api/api-2.json
```diff
@@ -1475,6 +1507,13 @@
         "taskDefinition":{"shape":"TaskDefinition"}
       }
     },
+    "RepositoryCredentials":{
+      "type":"structure",
+      "required":["credentialsParameter"],
+      "members":{
+        "credentialsParameter":{"shape":"String"}
```
tehshwann: Does this allow for more than one credential set? And is that something we should consider?

agent/functional_tests/tests/generated/simpletests_unix/simpletests_generated_unix_test.go
```diff
@@ -762,6 +762,86 @@ func TestSimpleExit(t *testing.T) {
...
+		if exit, ok := testTask.ContainerExitcode("exit"); !ok || exit != 42 {
+			t.Errorf("Expected exit to exit with 42; actually exited (%v) with %v", ok, exit)
+		}
+
+		defer agent.SweepTask(testTask)
+	}
+
+}
+
+// TestTaskSharedVolume Verify that shared Docker volume works as expected
```
tehshwann: Should we add some tests here that write real data to the volume rather than just "touching" it? For example... perhaps we should have something that fills the volume to make sure it can actually be used by a task?

richardpen: A better way to test the shared volume may be actually starting two tasks:
* task one to write to the volume
* task two read from the volume and verify the content
Any thoughts?

sharanyad: That's a good idea. I think we could test for volume cleanup through this too, where task 1 writes to the volume and stops executing (volume not removed here), and task 2 reads from the same volume.

sharanyad:
Closing this in favor of a different PR that @yunhee-l will put up.

PR #1435, opened 2018-07-06
[wip] Add AWSDomain in Config and compose awslogs endpoint

haikuoliu:
### Summary

Compose the awslogs endpoint and pass it to Docker, this prevents us from being blocked by two cases:
1. Docker doesn't update AWS SDK in time when there is a new AWS region.
2. The new AWS region is not recorded in AWS SDK.

I have submitted a [PR](https://github.com/moby/moby/pull/37374) to moby which introduces `awslogs-endpoint` log option.

Since Docker does check on log options, and current log options do not include `awslogs-endpoint`, this PR cannot be merged (but still needs review!) until we know the Docker version that supports `awslogs-endpoint` (both for Windows and Linux).

TODO:
1. Update `awslogsEndpointMinimumDockerVersion` when we know the Docker version that supports `awslogs-endpoint`.
2. Add entry in `CHANGELOG.md` when we are sure about the Agent version will support this feature.

### Implementation details

1. Get AWSDomain from EC2 metadata, add it to config.
2. Pass the config to `DockerHostConfig` method, construct a CloudWatch Logs endpoint using `config.AWSDomain`, `config.AWSRegion`/`awslogs-region` log option, add it to `awslogs-endpoint` option.
3. Add tests, refactor some tests, some manual tests with manually compiled Docker.

Reason that I don't use AWS SDK to construct the endpoint: 
The AWS SDK package doesn't expose the method to construct an endpoint without scheme, the endpoint you can get is like `https://logs.us-west-2.amazonaws.com`, but the endpoint we want is like `logs.us-west-2.amazonaws.com`. And some region is probably not in AWS SDK. 

Since we know from AWS SDK that the endpoint pattern is `{service}.{region}.{dnsSuffix}`, we can simply construct it by ourself.

### Testing

* Enhancement - Add AWSDomain in config, construct awslogs endpoint and pass it to Docker.

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -91,6 +91,21 @@ const (
 	ipcModeTask     = "task"
 	ipcModeSharable = "shareable"
 	ipcModeNone     = "none"
+
+	// awslogsType is the type of AWS logs specified in the log configuration
+	awslogsType = "awslogs"
+	// awslogsRegionOpt is the awslogs option to specify the AWS logs region
+	awslogsRegionOpt = "awslogs-region"
+	// awslogsEndpointOpt is the awslogs option to specify the AWS logs endpoint
+	awslogsEndpointOpt = "awslogs-endpoint"
+	// awslogsServiceName is the service name of AWS logs, it's used to
+	// compose the AWS logs endpoint
+	awslogsServiceName = "logs"
+
+	// The minimum docker CE version that supports "awslogs-endpoint" log option.
+	awslogsEndpointMinimumDockerCEVersion = "18.09-ce"
+	// The minimum docker EE version that supports "awslogs-endpoint" log option.
+	awslogsEndpointMinimumDockerEEVersion = "18.09-ee"
 )
 
 // TaskOverrides are the overrides applied to a task
@@ -255,7 +270,7 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	}
 
 	if task.requiresSSMSecret() {
-		task.initializeSSMSecretResource(credentialsManager, resourceFields)
+		task.initializeSSMSecretResource(cfg, credentialsManager, resourceFields)
 	}
 
 	if task.requiresASMSecret() {
@@ -606,9 +621,10 @@ func (task *Task) requiresSSMSecret() bool {
 }
 
 // initializeSSMSecretResource builds the resource dependency map for the SSM ssmsecret resource
-func (task *Task) initializeSSMSecretResource(credentialsManager credentials.Manager,
+func (task *Task) initializeSSMSecretResource(cfg *config.Config,
+	credentialsManager credentials.Manager,
 	resourceFields *taskresource.ResourceFields) {
-	ssmSecretResource := ssmsecret.NewSSMSecretResource(task.Arn, task.getAllSSMSecretRequirements(),
+	ssmSecretResource := ssmsecret.NewSSMSecretResource(cfg, task.Arn, task.getAllSSMSecretRequirements(),
 		task.ExecutionCredentialsID, credentialsManager, resourceFields.SSMClientCreator)
 	task.AddResource(ssmsecret.ResourceName, ssmSecretResource)
 
@@ -936,12 +952,15 @@ func (task *Task) dockerExposedPorts(container *apicontainer.Container) nat.Port
 }
 
 // DockerHostConfig construct the configuration recognized by docker
-func (task *Task) DockerHostConfig(container *apicontainer.Container, dockerContainerMap map[string]*apicontainer.DockerContainer, apiVersion dockerclient.DockerVersion) (*dockercontainer.HostConfig, *apierrors.HostConfigError) {
-	return task.dockerHostConfig(container, dockerContainerMap, apiVersion)
+func (task *Task) DockerHostConfig(cfg *config.Config,
+	container *apicontainer.Container,
+	dockerContainerMap map[string]*apicontainer.DockerContainer,
+	apiVersion dockerclient.DockerVersion, dockerVersion string) (*dockercontainer.HostConfig, *apierrors.HostConfigError) {
+	return task.dockerHostConfig(cfg, container, dockerContainerMap, apiVersion, dockerVersion)
 }
 
 // ApplyExecutionRoleLogsAuth will check whether the task has execution role
-// credentials, and add the genereated credentials endpoint to the associated HostConfig
+// credentials, and add the generated credentials endpoint to the associated HostConfig
 func (task *Task) ApplyExecutionRoleLogsAuth(hostConfig *dockercontainer.HostConfig, credentialsManager credentials.Manager) *apierrors.HostConfigError {
 	id := task.GetExecutionCredentialsID()
 	if id == "" {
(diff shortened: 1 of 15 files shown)


agent/config/config.go
```diff
@@ -174,9 +177,11 @@ func NewConfig(ec2client ec2.EC2MetadataClient) (*Config, error) {
 
 	if config.AWSRegion == "" {
 		// Get it from metadata only if we need to (network io)
-		config.Merge(ec2MetadataConfig(ec2client))
+		config.Merge(ec2MetadataRegionConfig(ec2client))
 	}
 
+	config.Merge(ec2MetadataDomainConfig(ec2client))
```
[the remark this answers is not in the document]
samuelkarp: `AWSDomain` can come from `fileConfig` as well, so you still need the check.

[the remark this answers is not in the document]
haikuoliu: OK, added.

adnxn:
@haikuoliu: does this need to be part of `1.19.0`?

haikuoliu:
@adnxn not sure about Agent version yet, it depends on when docker releases the `awslogs-endpoint` feature.

agent/api/task/task_linux.go
```diff
@@ -56,6 +56,10 @@ const (
 
 	minimumCPUPercent = 0
 	bytesPerMegabyte  = 1024 * 1024
+
+	// The minimum docker version that supports "awslogs-endpoint" log option.
+	// Will update this when we know the version.
+	awslogsEndpointMinimumDockerVersion = "19.03.1-ce"
```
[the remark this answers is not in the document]
haikuoliu: It's set both in `task_linux.go` and `task_windows.go`, because Linux and Windows use different docker versions.

I will update this version when the the endpoint option is released in Docker.

agent/api/task/task.go
```diff
@@ -648,6 +667,15 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
 		}
 	}
 
+	if hostConfig.LogConfig.Type == awslogsType {
+		match, err := utils.Version(dockerVersion).Matches(awslogsEndpointMinimumDockerVersion)
```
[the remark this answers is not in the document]
haikuoliu: oh yeah, the selector for `Matches` function should be `">=" + awslogsEndpointMinimumDockerVersion`

agent/api/task/task.go
```diff
@@ -69,9 +69,27 @@ const (
...
+	// awslogsRegionOpt is the awslogs option to specify the AWS logs region
+	awslogsRegionOpt = "awslogs-region"
+	// awslogsEndpointOpt is the awslogs option to specify the AWS logs endpoint
+	awslogsEndpointOpt = "awslogs-endpoint"
+	// awslogsServiceName is the service name of AWS logs, it's used to
+	// compose the AWS logs endpoint
+	awslogsServiceName = "logs"
+
+	// The minimum docker CE version that supports "awslogs-endpoint" log option.
+	// Will update this when we know the version.
```
samuelkarp: How will you remember to update this?

agent/api/task/task.go
```diff
@@ -846,6 +876,81 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
 	return hostConfig, nil
 }
 
+func (task *Task) awslogsEndpointOverride(hostConfig *docker.HostConfig, dockerVersion string) bool {
+	if hostConfig.LogConfig.Type != awslogsType {
+		return false
+	}
+
+	isVersionMatch := false
+	if task.isDockerCEVersion(dockerVersion) {
```
samuelkarp: You don't need to have this much logic; you can do it like this:

```go
match, err := utils.Version(">=19.03.1,>=19.03.1-ce,>=19.03.1-ee")
```

Or, alternately (and I think better!) would be to fix `utils.Version` (or maybe `utils.DockerVersion`) so that it can treat `-ce` and `-ee` as satisfying a bare number comparison (so `19.03.1-ce` would satisfy `>=19.03.1`).

agent/api/task/task.go
```diff
@@ -846,6 +876,81 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
...
+		}
+	} else if task.isDockerEEVersion(dockerVersion) {
+		match, err := utils.Version(dockerVersion).Matches(">=" + awslogsEndpointMinimumDockerEEVersion)
+		if err != nil {
+			seelog.Warnf("Unable to compare docker version: %v", err)
+		} else if match {
+			isVersionMatch = true
+		}
+	} else {
+		seelog.Warnf("Docker version %s is neither CE version or EE version", dockerVersion)
```
samuelkarp: This shouldn't be a warning.  Older versions of Docker won't be CE or EE, and customers that are building their own versions may be setting different things here too.

agent/api/task/task.go
```diff
@@ -846,6 +876,81 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
 	return hostConfig, nil
 }
 
+func (task *Task) awslogsEndpointOverride(hostConfig *docker.HostConfig, dockerVersion string) bool {
```
samuelkarp: Can you rename this to `shouldOverrideAWSLogsEndpoint` so that it conveys that it's just providing the logic for whether an override should be set?

agent/api/task/task.go
```diff
@@ -846,6 +876,81 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
...
+			isVersionMatch = true
+		}
+	} else {
+		seelog.Warnf("Docker version %s is neither CE version or EE version", dockerVersion)
+	}
+
+	if isVersionMatch {
+		return hostConfig.LogConfig.Config[awslogsEndpointOpt] == ""
+	} else {
+		// If the Docker version is not matched, make the awslogs endpoint option empty
```
samuelkarp: You're now overriding something that is set either in the task definition or by the ECS backend.  I'd caution against this as logic here would prevent us from forcing an override if we need to.

agent/api/task/task.go
```diff
@@ -846,6 +876,81 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
 	return hostConfig, nil
 }
 
+func (task *Task) awslogsEndpointOverride(hostConfig *docker.HostConfig, dockerVersion string) bool {
+	if hostConfig.LogConfig.Type != awslogsType {
+		return false
+	}
+
```
samuelkarp: You can skip all the version-matching logic if the log option is already set.  I'd suggest moving that check up here.

agent/api/task/task.go
```diff
@@ -846,6 +876,81 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
...
+		// as it's not supported
+		if hostConfig.LogConfig.Config[awslogsEndpointOpt] != "" {
+			hostConfig.LogConfig.Config[awslogsEndpointOpt] = ""
+		}
+	}
+
+	return false
+}
+
+func (task *Task) isDockerCEVersion(dockerVersion string) bool {
```
samuelkarp: I'd like you to remove these functions anyway (see above for an alternate suggestion), but for future reference: the Docker version is not a property of the task, so it shouldn't have a `(task *Task)` receiver.

agent/config/types.go
```diff
@@ -46,6 +46,9 @@ type Config struct {
 	// be inferred from the EC2 metadata service, but if it cannot be found this
 	// will be fatal.
 	AWSRegion string `missing:"fatal" trim:"true"`
+	// AWSDomain is the domain to run in (such as "amazonaws.com"). This value will
+	// be inferred from the EC2 metadata service.
+	AWSDomain string `trim:"true"`
```
samuelkarp: Is a missing domain fatal?

agent/api/task/task.go
haikuoliu: I won't merge this PR until I know the versions, and you don't need to approve it now, I will ask you (and someone else) to review this again when I am ready to merge.

agent/api/task/task.go
haikuoliu: Correct me if I am wrong, if the minimum docker ee and docker ce share same versions we can definitely do this, but when they have different versions the logic cannot be simplified this way.

Let's say our minimum docker ce version is `19.03.1-ce`, and minimum docker ee version is `20.0.0-ee`. When our current docker version is `19.05.1-ee`, case `">=19.03.1,>=19.03.1-ce,>=20.0.0-ee"` will be matched, because `Matches` method will return true as `19.05.1-ee` is greater than `19.03.1-ce`. But actually this case should not be matched.

I am unsure about those versions right now, so I'll leave it as it is until we know them.

samuelkarp: I'd rather have our `utils.Version` (or `utils.DockerVersion`) fixed to treat both `-ce` and `-ee` as satisfying a bare number comparison.  `utils.Version` was written to compare SemVer version numbers, but Docker releases don't follow SemVer; `-ce` and `-ee` are not prereleases even though our matcher currently treats them as such.

PR #1438, opened 2018-07-09
Adding interface package for Docker SDK clients

kavoor:
### Summary

Added a package, 'mobyiface' that provides an interface for moby clients to implement.
Generated mocks for this interface.
Vendored additional packages required for Client interface.

### Implementation details

Defined a new interface with necessary functions that map to current go-dockerclient functions.
Ran dep ensure to determine additional packages that needed vendoring.
Ran generate/mockgen script to generate mocks for this new interface.

### Testing

Tested by running 'make test' and Octokitten tests.

Feature - Create interface package for Docker SDK clients

### Licensing


sharanyad:
I would suggest opening this PR against a new branch.

agent/dockerclient/mobyiface/generate_mocks.go
```diff
@@ -0,0 +1,16 @@
+// Copyright 2014-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
```
haikuoliu: nit: `Copyright 2018 Amazon.com`

agent/dockerclient/mobyiface/interface.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"context"
+	"io"
+	"time"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/network"
+
```
haikuoliu: nit: redundant line.

yunhee-l:
do we need to let CP team know about supported moby version?

agent/dockerclient/mobyiface/interface.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/api/types/events"
+	"github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/volume"
+)
+
+// Client is an interface specifying the subset of
+// github.com/moby/moby/tree/master/client that the agent uses.
+type Client interface {
+	Events(ctx context.Context, options types.EventsOptions) (<-chan events.Message, <-chan error)
```
adnxn: lint: i *think* you tried to align the methods with the existing ordering in the old interface. what do you think about grouping the methods by what the act on? Container*, Image*, Volume*.

agent/dockerclient/mobyiface/interface.go
```diff
@@ -0,0 +1,56 @@
...
+// Package mobyiface contains an interface for moby matching the
+// subset used by the agent
+package mobyiface
+
+import (
+	"context"
+	"io"
+	"time"
+
+	"github.com/docker/docker/api/types"
```
richardpen: Can you change this to "github.com/moby/moby" instead, which means you need to modify the `Gopkg.lock` and few other files?

agent/dockerclient/mobyiface/interface.go
```diff
@@ -0,0 +1,56 @@
...
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+// Package mobyiface contains an interface for moby matching the
+// subset used by the agent
+package mobyiface
```
richardpen: Would `mobyclient` be a better name?

agent/dockerclient/mobyiface/interface.go
haikuoliu: I think "github.com/docker/docker", is the convention, if you take a look at moby's code, they all use "github.com/docker/docker".

kavoor: The examples in moby/moby client have [//import "github.com/docker/docker/client"]. In the documentation as well it says to: import "github.com/docker/docker/client". Is there a particular reason you think moby/moby is better?

samuelkarp: `// import "github.com/docker/docker/api/types/events"` is at the top of the files too so that Go uses that import path.  I think these should stay the way they are now.

agent/dockerclient/mobyiface/interface.go
kavoor: yes

PR #1448, opened 2018-07-13
Generate and inject v3 metadata url

fenxiong:
### Summary

This pr add functionality to generate and inject v3 metadata url as an environment variable for the container.
Manual tests are in progress.

### Implementation details

The agent generates an identifier for container when creating it, and injects the env. It saves the identifier mappings in engine state.

### Testing

New tests cover the changes: 
yes

### Licensing


agent/engine/dockerstate/dockerstate_test.go
```diff
@@ -361,3 +363,93 @@ func TestTaskIPAddress(t *testing.T) {
...
+			container,
+		},
+	}
+	mockUUID := mock_uuidwrapper.NewMockUUID(ctrl)
+
+	state.v3EndpointIDToTask["containerV3EndpointID"] = "taskArn"
+	state.v3EndpointIDToContainerName["containerV3EndpointID"] = "containerName"
+
+	for i := 0; i < maximumGenerateIDRetries; i++ {
+		mockUUID.EXPECT().New().Return("containerV3EndpointID")
```
petderek: A mock isn't necessary for this object. Just do something like this:

```golang
type StaticUUIDProvider struct {
    TestId string
}

func (t *StaticUUIDProvider) New() string {
    return t.TestId
}
```

It may also be possible to just use the uuid library directly in the test.

agent/utils/uuidwrapper/uuid.go
```diff
@@ -0,0 +1,35 @@
...
+// permissions and limitations under the License.
+
+package uuidwrapper
+
+import (
+	"github.com/pborman/uuid"
+)
+
+// UUID wraps 'uuid' methods for testing
+type UUID interface {
```
petderek: Call this UUIDProvider or something that implies that it generates a UUID.

Otherwise, it sounds like the object itself *is* a uuid. :)

agent/utils/uuidwrapper/uuid.go
```diff
@@ -0,0 +1,35 @@
...
+// License is located at
+//
+//     http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package uuidwrapper
```
petderek: nit: rename this package to be the same as whatever you call the interface.

agent/utils/uuidwrapper/mocks/uuidwrapper_mocks.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2015-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
petderek: (see other comment)

we don't need a mock for simple objects.

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -457,3 +470,70 @@ func (state *DockerTaskEngineState) GetTaskByIPAddress(addr string) (string, boo
...
+func (state *DockerTaskEngineState) ContainerByV3EndpointID(v3EndpointID string) (*apicontainer.DockerContainer, bool) {
+	state.lock.RLock()
+	defer state.lock.RUnlock()
+
+	taskArn, ok := state.v3EndpointIDToTask[v3EndpointID]
+	if !ok {
+		return nil, false
+	}
+
+	containerName, ok := state.v3EndpointIDToContainerName[v3EndpointID]
```
petderek: what does it mean if we fail here instead of later? are the failure conditions different at each step here?

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -457,3 +470,70 @@ func (state *DockerTaskEngineState) GetTaskByIPAddress(addr string) (string, boo
 	taskARN, ok := state.ipToTask[addr]
 	return taskARN, ok
 }
+
+// GenerateAndAddV3EndpointID generates a container identifier to be used in v3 metadata endpoint,
+// and update relevant maps. Return false if we fail to generate an unique identifier, otherwise
+// return true
+func (state *DockerTaskEngineState) GenerateAndAddV3EndpointID(task *apitask.Task, container *apicontainer.Container, uuid uuidwrapper.UUID) bool {
```
petderek: I'd prefer it if the uuid object were either bound to the DockerTaskEngineState struct or stored in a package-level variable.

```golang
// either do something like this...
var uuidFunc UUID

init() {
    uuidFunc = uuidwrapper.NewUUID()
}

// or set it up so you can do something like this (more consistent with other modules in agent):
state.uuid.New()
```

haikuoliu:
Should this to be merged into its own branch instead of `dev` branch?

agent/engine/docker_task_engine_test.go
```diff
@@ -744,6 +746,95 @@ func TestCreateContainerMergesLabels(t *testing.T) {
 	taskEngine.(*DockerTaskEngine).createContainer(testTask, testTask.Containers[0])
 }
 
+// TestCreateContainerV3EndpointIDGenerateFailure tests that when we successfully generate
+// v3 endpoint id for container, we will inject v3 endpoint to it
+func TestCreateContainerV3EndpointIDGenerateSucceed(t *testing.T) {
```
haikuoliu: Is it possible to put `TestCreateContainerV3EndpointIDGenerateSucceed` and `TestCreateContainerV3EndpointIDGenerateFailure` in the same function? They share the most of the logic.

agent/api/container/container.go
```diff
@@ -47,6 +47,13 @@ const (
 
 	// AuthTypeASM is to use image pull auth over AWS Secrets Manager
 	AuthTypeASM = "asm"
+
+	// ContainerMetadataURIEnvironmentVariableName defines the name of the environment
+	// variable in containers' config, which can be used by the containers to access the
+	// v3 metadata endpoint
+	ContainerMetadataURIEnvironmentVariableName = "AWS_CONTAINER_METADATA_URI"
```
haikuoliu: Will `ContainerMetadataURIEnvironmentVariableName` and `ContainerMetadataURIFormat` be used outside of this package?

agent/engine/dockerstate/dockerstate_test.go
```diff
@@ -361,3 +363,93 @@ func TestTaskIPAddress(t *testing.T) {
...
+	mockUUID := mock_uuidwrapper.NewMockUUID(ctrl)
+
+	assert.True(t, state.GenerateAndAddV3EndpointID(task, container, mockUUID))
+	assert.Equal(t, container.V3EndpointID, "containerV3EndpointID")
+}
+
+// TestGenerateAndAddV3EndpointIDFailure tests that when the container didn't
+// have a v3 endpoint id and we failed to generate a new uuid after maximum retries,
+// the function returns failure without setting the container's id
+func TestGenerateAndAddV3EndpointIDFailure(t *testing.T) {
```
haikuoliu: nit: `TestGenerateAndAddV3EndpointIDFails`.

agent/statemanager/state_manager.go
```diff
@@ -64,7 +64,10 @@ const (
 	//  b)Remove `AppliedStatus` field form 'apicontainer.Container'
 	// 12) Deprecate 'TransitionDependencySet' and add new 'TransitionDependenciesMap' in 'apicontainer.Container'
 	// 13) Add 'resources' field to 'api.task.task'
-	ECSDataVersion = 13
+	// 14)
+	//    a) Add 'V3EndpointID' field to 'Container' struct
```
haikuoliu: nit: 3 spaces before `a)`, please also do this for `11)`.

agent/api/container/container.go
fenxiong: yes it's used in TestCreateContainerV3EndpointIDGenerateSucceeds in task engine unit test

agent/engine/dockerstate/docker_task_engine_state.go
fenxiong: i think currently, if the first check succeeds, we won't fail any of the later checks. it's just a sanity check in case the updating logic related to those maps change in the future.

fenxiong:
@haikuoliu i changed the base branch to 'v3-metadata'

agent/engine/dockerstate/docker_task_engine_state.go
fenxiong: if it's a global var, then if i want to test it, i need to add the logic to override it when another uuid provider is specified, which seems tricky, because the logic is specially for testing.. i think passing a uuid provider around is fine since its structure is very simple (empty struct for dynamic provider, one string field for static provider)

agent/utils/uuidwrapper/uuid.go
fenxiong: i was mimicking other wrapper packages when i created this.. now that i don't use mock i think i will put it under utils package because it's pretty simple

agent/api/container/container.go
samuelkarp: Is `AWS_CONTAINER_METADATA_URI` intended to be generic across other AWS offerings, or is this an ECS-specific feature?  If it's ECS-specific, I'd recommend renaming this to `ECS_CONTAINER_METADATA_URI`.

agent/engine/docker_task_engine.go
```diff
@@ -930,6 +930,14 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		}
 	}
 
+	// Generate container id for v3 metadata endpoint, construct the endpoint and inject it as an environment variable in the container
+	generateSucceed := engine.state.GenerateAndAddV3EndpointID(task, container, utils.NewDynamicUUIDProvider())
```
samuelkarp: Here's a slightly more idiomatic approach:

```go
if ok := engine.state.GenerateAndAddV3EndpointID(task, container, utils.NewDynamicUUIDProvider()); ok {
	container.InjectV3MetadataEndpoint()
} else {
	seelog.Warnf(...)
}
```

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -457,3 +481,70 @@ func (state *DockerTaskEngineState) GetTaskByIPAddress(addr string) (string, boo
 	taskARN, ok := state.ipToTask[addr]
 	return taskARN, ok
 }
+
+// GenerateAndAddV3EndpointID generates a container identifier to be used in v3 metadata endpoint,
```
samuelkarp: I don't really think this is appropriately factored.  This function is doing several pretty distinct things, mixing abstraction layers, and is not really in the right place inside `DockerTaskEngineState`.  Here are the three things it does:

1. Generate an ID
2. Modify the container configuration
3. Update the mapping from ID to task and ID to container

I think 3 is the only part that's really appropriate to be inside `DockerTaskEngineState`.  Just looking at prior examples of us doing things like this, I think the most appropriate place (right now) to modify the container configuration (2 in my list above) is in `PostUnmarshalTask`; that's where we do things like injecting the credential endpoint.  1 can also likely be done there, and then you can read the `container.V3EndpointID` in this function (instead of writing to it).

fenxiong: i put them together because if i generate an ID for a container but not add it to state, there's nothing to prevent another container generating a duplicate ID. i can separate the function into two, but it doesn't seem that i can separate them in two packages

samuelkarp: In general, we can assume that the probability of a UUID being duplicated is negligible: https://en.wikipedia.org/wiki/Universally_unique_identifier.

haikuoliu: I agree that we can just ignore the case where UUID is duplicate, according the [doc](https://github.com/aws/amazon-ecs-agent/blob/c0defea9598061451cf5ffd452d2ffe3d91dbc5e/agent/vendor/github.com/pborman/uuid/version4.go#L16), the chance that we get a duplicate is already small enough, considering the uniqueness is only required within each Agent - that makes the duplicate almost impossible.

Agent has already use UUID without detecting the duplicate.

agent/engine/common_test.go
```diff
@@ -182,6 +191,45 @@ func validateContainerRunWorkflow(t *testing.T,
...
+	assert.True(t, reflect.DeepEqual(expectedConfig, config),
+		"Mismatch in container config; expected: %v, got: %v", expectedConfig, config)
+
+	expectedConfig.Env = expectedConfigEnvList
+	config.Env = configEnvList
+}
+
+// checkDockerConfigsEnv checks whether two docker configs have same list of environment
+// variables and each has same value.
+func checkDockerConfigsEnv(t *testing.T, expectedConfig *docker.Config, config *docker.Config) {
```
samuelkarp: Instead of writing this logic, you can use [`assert.ElementsMatch`](https://godoc.org/github.com/stretchr/testify/assert#ElementsMatch) to compare the unordered slice elements.

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/cihub/seelog"
 )
 
+const maximumGenerateIDRetries = 3
```
samuelkarp: This const is unused.

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -457,3 +480,55 @@ func (state *DockerTaskEngineState) GetTaskByIPAddress(addr string) (string, boo
...
 	return taskARN, ok
 }
+
+// AddV3EndpointID adds mappings from container's v3 endpoint id to corresponding container and task
+func (state *DockerTaskEngineState) AddV3EndpointID(task *apitask.Task, container *apicontainer.Container) {
+	state.lock.Lock()
+	defer state.lock.Unlock()
+
+	if container.V3EndpointID == "" {
+		seelog.Debugf("The v3 endpoint id of container %s has not been set, skip adding v3 endpoint id mappings to state", container.Name)
```
samuelkarp: Do you need this debug message?

agent/engine/common_test.go
```diff
@@ -182,6 +191,45 @@ func validateContainerRunWorkflow(t *testing.T,
...
+	expectedConfigEnvList := expectedConfig.Env
+	configEnvList := config.Env
+
+	expectedConfigEnvSet := make(map[string]bool)
+	for _, env := range expectedConfigEnvList {
+		expectedConfigEnvSet[env] = true
+	}
+
+	ok := true
+	for _, env := range configEnvList {
```
haikuoliu: What if `configEnvList` is a subset of `expectedConfigEnvSet`?

agent/api/task/task_windows_test.go
```diff
@@ -111,10 +111,25 @@ func TestPostUnmarshalWindowsCanonicalPaths(t *testing.T) {
...
+// removeV3EndpointConfig removes the v3 endpoint id and the injected env for a container
+// so that checking all other fields can be easier
+func removeV3EndpointConfig(container *apicontainer.Container) {
+	container.V3EndpointID = ""
+	if container.Environment != nil {
+		delete(container.Environment, apicontainer.ContainerMetadataURIEnvironmentVariableName)
+	}
+	if len(container.Environment) == 0 {
+		container.Environment = nil
+	}
```
haikuoliu: Is `container.Environment = nil` necessary?

haikuoliu:
Please make sure you include this [PR](https://github.com/aws/amazon-ecs-agent/pull/1444) before merge.

agent/api/task/task_windows_test.go
fenxiong: yes because empty map is different from nil. so this is needed for the test to pass

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -377,6 +387,11 @@ func (state *DockerTaskEngineState) RemoveTask(task *apitask.Task) {
 
 	for _, dockerContainer := range containerMap {
 		state.removeIDToContainerTaskUnsafe(dockerContainer)
+
+		v3EndpointID := dockerContainer.Container.V3EndpointID
```
richardpen: Can you add a getter in Container instead of referencing it directly? And you can combine this into the next line:
```
if dockerContainer.Container.GetV3EndpointID() == ""{
}
```

agent/statemanager/state_manager.go
```diff
@@ -60,11 +60,14 @@ const (
...
-	//  a) Add 'PrivateDNSName' field to 'api.ENI'
-	//  b)Remove `AppliedStatus` field form 'apicontainer.Container'
+	//   a) Add 'PrivateDNSName' field to 'api.ENI'
+	//   b)Remove `AppliedStatus` field form 'apicontainer.Container'
 	// 12) Deprecate 'TransitionDependencySet' and add new 'TransitionDependenciesMap' in 'apicontainer.Container'
 	// 13) Add 'resources' field to 'api.task.task'
-	ECSDataVersion = 13
+	// 14)
+	//   a) Add 'V3EndpointID' field to 'Container' struct
+	//   b) Add 'v3EndpointIDToTask' and 'v3EndpointIDToContainerName' maps to engine state
```
richardpen: Are you sure `v3EndpointIDToContainerName` and `v3EndpointIDToTask` are saved in the state file, can you verify that?

agent/api/container/container.go
```diff
@@ -77,6 +84,9 @@ type HealthStatus struct {
 type Container struct {
 	// Name is the name of the container specified in the task definition
 	Name string
+	// V3EndpointID is a container identifier used to construct v3 metadata endpoint; it's unique among
+	// all the containers managed by the agent
+	V3EndpointID string
```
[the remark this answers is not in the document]
fenxiong: discussed offline

agent/statemanager/state_manager.go
fenxiong: yes i had a manual test and verified that

agent/api/container/container.go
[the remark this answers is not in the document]
fenxiong: on second thought, i think once we persist this, we don't need to persist the other two maps, but can construct the maps during unmarshal, like the taskToID map. This way we don't persist redundant information

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -377,6 +387,11 @@ func (state *DockerTaskEngineState) RemoveTask(task *apitask.Task) {
 
 	for _, dockerContainer := range containerMap {
 		state.removeIDToContainerTaskUnsafe(dockerContainer)
+
+		v3EndpointID := dockerContainer.Container.GetV3EndpointID()
```
richardpen: Can you combine these lines into one: 
```
state.removeV3EndpointIDToTaskContainerUnsafe(dockerContainer.Container.GetV3EndpointID())
```

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -69,6 +69,12 @@ type TaskEngineState interface {
 	AddTaskIPAddress(addr string, taskARN string)
 	// GetTaskByIPAddress gets the task arn for an IP address
 	GetTaskByIPAddress(addr string) (string, bool)
+	// AddV3EndpointID adds mappings from container's v3 endpoint id to corresponding container and task
+	AddV3EndpointID(task *apitask.Task, container *apicontainer.Container)
```
richardpen: Can this be done in `AddContainer()`?

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -89,13 +95,15 @@ type TaskEngineState interface {
...
-	ipToTask       map[string]string // ip address -> task arn
+	tasks                       map[string]*apitask.Task                            // taskarn -> apitask.Task
+	idToTask                    map[string]string                                   // DockerId -> taskarn
+	taskToID                    map[string]map[string]*apicontainer.DockerContainer // taskarn -> (containername -> c.DockerContainer)
+	idToContainer               map[string]*apicontainer.DockerContainer            // DockerId -> c.DockerContainer
+	eniAttachments              map[string]*apieni.ENIAttachment                    // ENIMac -> apieni.ENIAttachment
+	imageStates                 map[string]*image.ImageState
+	ipToTask                    map[string]string // ip address -> task arn
+	v3EndpointIDToTask          map[string]string // container's v3 endpoint id -> taskarn
+	v3EndpointIDToContainerName map[string]string // container's v3 endpoint id -> container name
```
richardpen: Can this be map of id -> container Docker ID, as we already have a map of Docker ID to containers?

agent/utils/uuid.go
```diff
@@ -0,0 +1,50 @@
...
+
+// dynamicUUIDProvider is a uuid provider used in run time
+type dynamicUUIDProvider struct {
+}
+
+func NewDynamicUUIDProvider() UUIDProvider {
+	return &dynamicUUIDProvider{}
+}
+
+func (*dynamicUUIDProvider) New() string {
```
richardpen: Lint: please add comments to these methods.

agent/api/container/container.go
```diff
@@ -47,6 +47,13 @@ const (
 
 	// AuthTypeASM is to use image pull auth over AWS Secrets Manager
 	AuthTypeASM = "asm"
+
+	// MetadataURIEnvironmentVariableName defines the name of the environment
+	// variable in containers' config, which can be used by the containers to access the
+	// v3 metadata endpoint
+	MetadataURIEnvironmentVariableName = "ECS_CONTAINER_METADATA_URI"
+	// MetadataURIFormat defines the URI format for v3 metadata endpoint
+	MetadataURIFormat = "http://169.254.170.2/v3/%s"
```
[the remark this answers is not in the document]
fenxiong: discussed offline

PR #1459, opened 2018-07-19
dockerapi: Support for Docker SDK Client

kavoor:
### Summary

Adds support for Docker SDK Clients to exists side-by-side with go-dockerclient Clients.
This will allow the API calls to be migrated one-by-one until all of the API calls come from the Docker SDK Client. At this point, the go-dockerclient Clients will be removed.

### Implementation details

 - An **sdkClientFactory** field was added to the dockerGoClient struct to allow Docker SDK Clients to be generated. 
   - The DockerGoClient abstraction over both the go-dockerclient clientfactory and the Docker SDK clientfactory will make it easier to swap which client is making the API call at the function level.
   - This is because the dockerGoClients implement the DockerClient interface, the functions being migrated.
- To create these new dockerGoClients, a parameter for the new **sdkClientFactory** was added to **NewDockerGoClient()** to pass in when creating the dockerGoClient.
   - Code for NewDockerGoClient() was written with separation between go-dockerclient and Docker SDK calls to make it easier to remove go-dockerclient later on.
  - At every place "NewDockerGoClient" is called, the argument for sdkClientFactory was added, and the mocks for these packages were adjusted likewise.
   - Passing in the new arguments for sdkClientFactory was simple as both clientFactory and sdkClientFactory take in the same arguments.
- sdkDockerClient() was added as a getter method for the sdkClients
- A test for Ping() was added for the Docker SDK clients as it is called in NewDockerGoClient().

### Testing

 Feature - Support for Docker SDK Clients

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -32,6 +32,8 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/clientfactory"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerauth"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockeriface"
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient/sdkclient"
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient/sdkclientfactory"
 	"github.com/aws/amazon-ecs-agent/agent/ecr"
 	"github.com/aws/amazon-ecs-agent/agent/emptyvolume"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
@@ -198,8 +200,8 @@ type DockerClient interface {
 	LoadImage(context.Context, io.Reader, time.Duration) error
 }
 
-// DockerGoClient wraps the underlying go-dockerclient library.
-// It exists primarily for the following three purposes:
+// DockerGoClient wraps the underlying go-dockerclient and docker/docker library.
+// It exists primarily for the following four purposes:
 // 1) Provide an abstraction over inputs and outputs,
 //    a) Inputs: Trims them down to what we actually need (largely unchanged tbh)
 //    b) Outputs: Unifies error handling and the common 'start->inspect'
@@ -212,9 +214,13 @@ type DockerClient interface {
 //    pull-related issues in the Docker daemon.
 // 3) Versioning: It abstracts over multiple client versions to allow juggling
 //    appropriately there.
+// 4) Allows for both the go-dockerclient client and Docker SDK client to live
+//    side-by-side until migration to the Docker SDK is complete.
 // Implements DockerClient
+// TODO Remove clientfactory field once all API calls are migrated to sdkclientFactory
 type dockerGoClient struct {
 	clientFactory    clientfactory.Factory
+	sdkClientFactory sdkclientfactory.Factory
 	version          dockerclient.DockerVersion
 	ecrClientFactory ecr.ECRFactory
 	auth             dockerauth.DockerAuthProvider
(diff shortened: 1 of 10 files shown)


agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -75,13 +78,17 @@ func dockerClientSetupWithConfig(t *testing.T, conf config.Config) (
 	func()) {
 	ctrl := gomock.NewController(t)
 	mockDocker := mock_dockeriface.NewMockClient(ctrl)
+	mockDockerSDK := mock_sdkclient.NewMockClient(ctrl)
```
petderek: Its kind of weird alternating context on each line. I'd rather do one of these:

1) duplicate this file and have two sets of the same tests (so that we only ever test one thing per function)

2) use [subtests](https://blog.golang.org/subtests) and group the code into blocks instead of alternating each line?

```
func TestDockerFeature(t *testing.T) { {
    t.Run("sdk", func(t *testing.T) {
        // do sdk things
    })
    t.Run("docker", func(t *testing.T) {
        // old api things...
    })
}

Comment (kavoor) [reply agent/dockerclient/dockerapi/docker_client_test.go]: Idea behind this was that for most of these functions, a mock of NewDockerGoClient is created which requires both a clientfactory and an sdkClientFactory. Subtests looks like a great idea.

File: agent/dockerclient/dockerapi/docker_client_test.go [hunk 2]
```diff
@@ -798,14 +823,45 @@ func TestListContainersTimeout(t *testing.T) {
...
+	factory := mock_clientfactory.NewMockFactory(ctrl)
+	factory.EXPECT().GetDefaultClient().AnyTimes().Return(mockDocker, nil)
+	// Docker SDK tests
+	mockDockerSDK := mock_sdkclient.NewMockClient(ctrl)
+	mockDockerSDK.EXPECT().Ping(gomock.Any()).Return(types.Ping{}, nil).AnyTimes()
+	sdkFactory := mock_sdkclientfactory.NewMockFactory(ctrl)
+	sdkFactory.EXPECT().GetDefaultClient().AnyTimes().Return(mockDockerSDK, nil)
+
+	_, err := NewDockerGoClient(factory, sdkFactory, defaultTestConfig())
+	if err == nil {
```
sharanyad: you can use `assert.Error` here

agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -798,14 +823,45 @@ func TestListContainersTimeout(t *testing.T) {
...
-	_, err := NewDockerGoClient(factory, defaultTestConfig())
+	factory.EXPECT().GetDefaultClient().AnyTimes().Return(mockDocker, nil)
+	// Docker SDK tests
+	mockDockerSDK := mock_sdkclient.NewMockClient(ctrl)
+	mockDockerSDK.EXPECT().Ping(gomock.Any()).Return(types.Ping{}, errors.New("test error"))
+	sdkFactory := mock_sdkclientfactory.NewMockFactory(ctrl)
+	sdkFactory.EXPECT().GetDefaultClient().AnyTimes().Return(mockDockerSDK, nil)
+
+	_, err := NewDockerGoClient(factory, sdkFactory, defaultTestConfig())
 	if err == nil {
```
sharanyad: you can use `assert.Error` here

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -230,30 +236,49 @@ type dockerGoClient struct {
...
+	// Even if we have a DockerClient, the daemon might not be running. Ping from both clients
 	// to ensure it's up.
 	err = client.Ping()
 	if err != nil {
-		seelog.Errorf("DockerGoClient: unable to ping Docker daemon. Ensure Docker is running: %v", err)
+		seelog.Errorf("DockerGoClient: go-dockerclient unable to ping Docker daemon. " +
+			"Ensure Docker is running: %v", err)
+		return nil, err
+	}
+	_, err = sdkclient.Ping(context.TODO())
```
sharanyad: This should probably not be `context.TODO()`. We should be passing in context from parent.

PR #1467, opened 2018-07-25
acs: add an empty task in TaskStateChange for unrecognized tasks

haikuoliu:
### Summary

Add an empty task in TaskStateChange for unrecognized tasks

This fixes a bug where unrecognized task cannot be stopped

### Implementation details

Currently when Agent receives a task from ACS, if it's malformed and Agent cannot recognize it, it will be handled by [handleUnrecognizedTask](https://github.com/aws/amazon-ecs-agent/blob/3631cb9de33a569d16e7a19d77105beeef8160c2/agent/acs/handler/payload_handler.go#L326) function. The `handleUnrecognizedTask` intends to send a task change event to stop the task, but it fails to do so, because it returns a `TaskStateChange` without a `Task` in it, and this will be discarded by taskHandler, as it should not be sent according to [taskShouldBeSent](https://github.com/aws/amazon-ecs-agent/blob/3631cb9de33a569d16e7a19d77105beeef8160c2/agent/eventhandler/task_handler_types.go#L81)

So this fix is to add an empty task in the result of `handleUnrecognizedTask` so that the task change will not be discarded.

### Testing

Bug - Fixed a bug where unrecognized task cannot be stopped

### Licensing

### DIFF
--- a/agent/acs/handler/payload_handler.go
+++ b/agent/acs/handler/payload_handler.go
@@ -337,6 +337,10 @@ func (payloadHandler *payloadRequestHandler) handleUnrecognizedTask(task *ecsacs
 		TaskARN: *task.Arn,
 		Status:  apitaskstatus.TaskStopped,
 		Reason:  UnrecognizedTaskError{err}.Error(),
+		// The real task cannot be extracted from payload message, so we send an empty task.
+		// This is necessary because the task handler will not send an event whose
+		// Task is nil.
+		Task: &apitask.Task{},
 	}
 
 	payloadHandler.taskHandler.AddStateChangeEvent(taskEvent, payloadHandler.ecsClient)
--- a/agent/acs/handler/payload_handler_test.go
+++ b/agent/acs/handler/payload_handler_test.go
@@ -15,11 +15,12 @@
 package handler
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"reflect"
+	"sync"
 	"testing"
-
-	"context"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	"github.com/aws/amazon-ecs-agent/agent/api"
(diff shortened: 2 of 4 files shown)


adnxn:
@haikuoliu: was there a bug report for this? or how did we discover this gap? most just curious.

adnxn:
@haikuoliu: please add a unit test to cover this case

haikuoliu:
@adnxn I have added unit test for this.

agent/acs/handler/payload_handler_test.go
```diff
@@ -754,3 +755,30 @@ func TestPayloadHandlerAddedASMAuthData(t *testing.T) {
...
+	}
+
+	mockECSACSClient := mock_api.NewMockECSClient(tester.ctrl)
+	taskHandler := eventhandler.NewTaskHandler(tester.ctx, tester.payloadHandler.saver, nil, mockECSACSClient)
+	tester.payloadHandler.taskHandler = taskHandler
+
+	wait := &sync.WaitGroup{}
+	wait.Add(1)
+
+	mockECSACSClient.EXPECT().SubmitTaskStateChange(gomock.Any()).Do(func(x interface{}) {
```
adnxn: please check the Task field is not nil inside the mock call, probably [something like this](https://github.com/aws/amazon-ecs-agent/blob/master/agent/eventhandler/handler_test.go#L66).

agent/acs/handler/payload_handler.go
```diff
@@ -337,6 +337,10 @@ func (payloadHandler *payloadRequestHandler) handleUnrecognizedTask(task *ecsacs
 		TaskARN: *task.Arn,
 		Status:  apitaskstatus.TaskStopped,
 		Reason:  UnrecognizedTaskError{err}.Error(),
+		// The real task cannot be extracted from payload message, so we send an empty task.
+		// This is necessary because the task handler will not send an event whose
+		// Task is nil.
+		Task: &apitask.Task{},
```
richardpen: Can you do a nil check here, as the task isn't always nil here?

agent/acs/handler/payload_handler_test.go
```diff
@@ -754,3 +755,31 @@ func TestPayloadHandlerAddedASMAuthData(t *testing.T) {
...
+	taskHandler := eventhandler.NewTaskHandler(tester.ctx, tester.payloadHandler.saver, nil, mockECSACSClient)
+	tester.payloadHandler.taskHandler = taskHandler
+
+	wait := &sync.WaitGroup{}
+	wait.Add(1)
+
+	mockECSACSClient.EXPECT().SubmitTaskStateChange(gomock.Any()).Do(func(change api.TaskStateChange) {
+		assert.NotNil(t, change.Task)
+		wait.Done()
+		// Don't return, verify timeout happens.
```
tehshwann: If I understand this correctly, this test ensures that the handler deals with the empty task correctly. Are there other test cases we should be hitting? For example, does this handler code path get touched if the task is recognized but nill?

agent/acs/handler/payload_handler_test.go
```diff
@@ -754,3 +755,31 @@ func TestPayloadHandlerAddedASMAuthData(t *testing.T) {
...
+
+	mockECSACSClient := mock_api.NewMockECSClient(tester.ctrl)
+	taskHandler := eventhandler.NewTaskHandler(tester.ctx, tester.payloadHandler.saver, nil, mockECSACSClient)
+	tester.payloadHandler.taskHandler = taskHandler
+
+	wait := &sync.WaitGroup{}
+	wait.Add(1)
+
+	mockECSACSClient.EXPECT().SubmitTaskStateChange(gomock.Any()).Do(func(change api.TaskStateChange) {
+		assert.NotNil(t, change.Task)
```
tehshwann: Do we allow unrecognized tasks to run and complete? If so, how do we know that the task state changes after this are handled ok? 

i.e. Is the empty task passed on to another component? And if so, we we creating a new untested code path by creating the empty task?

agent/acs/handler/payload_handler_test.go
haikuoliu: This test is to ensure that the `SubmitTaskStateChange` will be hit when there are unrecognized tasks. I added a `not nil` check here is because we manually create a non-nil task and pass it along the code path.

The unrecognized task means that the task cannot be converted to apitask.Task or some of the fields are not in correct format, so if it's a nil task then it should be a unrecognized task.

agent/acs/handler/payload_handler.go
haikuoliu: Clarified offline.

PR #1468, opened 2018-07-25
dockerapi: Migrated VolumeRemove and APIVersion to Docker SDK

kavoor:
### Summary

Migration for VolumeRemove and APIVersion from go-dockerclient to the DockerSDK

### Implementation details

Changes were implemented by swapping API calls to new SDK Client. All changes could be done under the DockerClient interface making this migration fairly straightforward.

The VolumeRemove API call takes in a boolean for the force flag to remove a volume. Go-dockerclient did not support this flag, so 'false' is used as a placeholder.
### Testing

 Enhancement - VolumeRemove and APIVersion migrated to Docker SDK

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -1181,12 +1181,12 @@ func (dg *dockerGoClient) RemoveVolume(ctx context.Context, name string, timeout
 }
 
 func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
-	client, err := dg.dockerClient()
+	client, err := dg.sdkDockerClient()
 	if err != nil {
 		return &CannotGetDockerClientError{version: dg.version, err: err}
 	}
 
-	ok := client.RemoveVolume(name)
+	ok := client.VolumeRemove(ctx, name, false)
 	if ok != nil {
 		return &CannotRemoveVolumeError{err}
 	}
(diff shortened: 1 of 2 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1181,12 +1181,12 @@ func (dg *dockerGoClient) RemoveVolume(ctx context.Context, name string, timeout
...
 
 func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
-	client, err := dg.dockerClient()
+	client, err := dg.sdkDockerClient()
 	if err != nil {
 		return &CannotGetDockerClientError{version: dg.version, err: err}
 	}
 
-	ok := client.RemoveVolume(name)
+	ok := client.VolumeRemove(ctx, name, false)
```
sharanyad: What does the third argument indicate?

haikuoliu: It's means force remove: https://godoc.org/github.com/docker/docker/client#Client.VolumeRemove

The previous docker client that we use doesn't specify force remove: https://github.com/fsouza/go-dockerclient/blob/05828842f94da0aef6ad140aa3fe777d23d17139/volume.go#L123

The default is false from docker api doc:
https://docs.docker.com/engine/api/v1.36/#operation/VolumeDelete

So if the purpose is to achieve the same functionality, make it false is safe here.

kavoor: It is a placeholder for a force flag (-f) for removal of a volume. The boolean indicates whether the flag passed in or not, and with go-dockerclient, we did not pass it in.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1266,11 +1266,11 @@ func (dg *dockerGoClient) listPlugins(ctx context.Context) ListPluginsResponse {
 
 // APIVersion returns the client api version
 func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
-	client, err := dg.dockerClient()
+	client, err := dg.sdkDockerClient()
```
sharanyad: Since we have two different clients, how do we know if `APIVersion()` is returning the sdk client's version?

agent/dockerclient/dockerapi/docker_client.go
sharanyad: If removal fails for a task scoped volume, force removing it seems like the way to go. But this PR is not the right place for making this decision. We'll leave this as it is for now.

kavoor: I'll add this as a TODO to my development plan for this project.

PR #1437, opened 2018-07-09
handlers: refactor v1 and v2 handlers

haikuoliu:
### Summary

In Agent `handlers` folder, we have two handler versions: `v1` handlers defined in `handlers` package for Agent introspection endpoint, and `v2` handlers defined in `taskmetadata` package for endpoint that exposes task/container metadata, task/container stats, task IAM role credentials. 

Current handlers structure is quite confusing and makes it hard to add new handlers. So I reorganized the `handlers` folder, my goal is (in descending order of importance):

1. Do not affect handlers functionality, i.e. do not change tests logic, only make changes if it's necessary.
2. Make the structure clear, so that adding new handlers/endpoint is easy.
3. Make `v1` and `v2` handlers' code style consistent, reuse logic if possible.

### Implementation details

1. Create file for each type of handlers (except task and container metadata are in the same file). Move all handlers to `v1` and `v2` packages, define handler path in each handler file. Previously the credentials handler defines the `v1` and `v2` handlers together, so I split it, and make `v2.CredentialsHandler` use the major logic in `v1.CredentialsHandler`.
2. Move `PortResponse` from `v1` package to `v2` package. If we keep `PortResponse` in `v2` package, we will end up with cyclic import which is not supported by Golang.
3. Rename the file for server setup to `${version}_server_setup.go`, `${version}` is for server hostname version, i.e. `v2` server can contain `v2` and `v3` handlers.
4. Move the handler tests that use server setup to `${version}_server_setup_test.go` to avoid cyclic import.
5. Create the `utils` package, move `helper.go` and `types.go` to it.
6. Rename some variables that may be confusing, like `metadata` -> `agent_metadata`, `metadata` -> `task_container_metadata`.
7. Move new response logic to `response.go`.
8. Make some constants/method exported, as we need to use them across packages.
9. Some code style changes, and other minor enhancements.

### Testing

- [x] Builds on Linux (`make release`)
- [x] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [x] Unit tests on Linux (`make test`) pass
- [x] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [x] Integration tests on Linux (`make run-integ-tests`) pass
- [x] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [x] Functional tests on Linux (`make run-functional-tests`) pass
- [x] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: yes

[...]

PR #1471, opened 2018-07-25
handlers/v1: add json tags to response structs

adnxn:
### Testing

- [ ] Builds on Linux (`make release`)
- [ ] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [ ] Unit tests on Linux (`make test`) pass
- [ ] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [ ] Integration tests on Linux (`make run-integ-tests`) pass
- [ ] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [ ] Functional tests on Linux (`make run-functional-tests`) pass
- [ ] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: 

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1473, opened 2018-07-25
handlers/v1: add json tags to response structs

haikuoliu:
### Summary

This PR adds a json tag to `DockerID` in v1 handler response, so that when it's marshalled the json text of this `DockerID` attribute will be `DockerId`, this is to fix a bug (introduced in [this](https://github.com/aws/amazon-ecs-agent/pull/1437/commits/5970f46f68f332a8d8cbc5435d8b0c7f18c0e48f) PR) where we provide a different container response for v1 handler in Agent 1.19.0, and it breaks the customers who are using v1 handler.

Will add functional tests in a separate PR.

### Testing

### Licensing

### DIFF
--- a/agent/handlers/v1/response.go
+++ b/agent/handlers/v1/response.go
@@ -46,7 +46,7 @@ type TasksResponse struct {
 
 // ContainerResponse is the schema for the container response JSON object
 type ContainerResponse struct {
-	DockerID   string
+	DockerID   string `json:"DockerId"`
 	DockerName string
 	Name       string
 	Ports      []PortResponse              `json:",omitempty"`
--- a/agent/handlers/v1/response.go
+++ b/agent/handlers/v1/response.go
@@ -24,41 +24,41 @@ import (
 
 // MetadataResponse is the schema for the metadata response JSON object
 type MetadataResponse struct {
-	Cluster              string
-	ContainerInstanceArn *string
-	Version              string
+	Cluster              string  `json:"Cluster"`
+	ContainerInstanceArn *string `json:"ContainerInstanceArn"`
+	Version              string  `json:"Version"`
 }
 
 // TaskResponse is the schema for the task response JSON object
 type TaskResponse struct {
-	Arn           string
-	DesiredStatus string `json:",omitempty"`
-	KnownStatus   string
-	Family        string
-	Version       string
-	Containers    []ContainerResponse
+	Arn           string              `json:"Arn"`
+	DesiredStatus string              `json:"DesiredStatus,omitempty"`
+	KnownStatus   string              `json:"KnownStatus"`
+	Family        string              `json:"Family"`
+	Version       string              `json:"Version"`
+	Containers    []ContainerResponse `json:"Containers"`
 }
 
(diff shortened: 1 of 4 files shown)


agent/handlers/v1/response.go
```diff
@@ -46,7 +46,7 @@ type TasksResponse struct {
 
 // ContainerResponse is the schema for the container response JSON object
 type ContainerResponse struct {
-	DockerID   string
+	DockerID   string `json:"DockerId"`
```
aaithal: same comment as https://github.com/aws/amazon-ecs-agent/pull/1471#discussion_r205248209: 

> I'd also suggest adding unit tests that unmarshal raw strings and validate each of the fields in the unmarshalled object with the expected one

haikuoliu: Sure, there is a `TODO` in the summary that indicates I will add tests for it.

Will add it soon.

petderek: @aaithal I would rather use a functional test that consumes the endpoint metadata for this instead of a unit test on the unmarshaller. I think a functional test will raise more warning flags about things like this breaking in the future.

aaithal: I'd vote for both instead of one or the other. You'd know immediately (during development) if you're breaking contracts if you add unit tests.

adnxn: i think in this case either option would have helped because the tests were also changed with the field renames in the original PR. i cant think of any way to really help prevent this other than having more eyes reviewing. lmk if you have any other ideas though. 😕 

i do agree we should add both unit and functional tests though for this change.

adnxn:
@haikuoliu: please add json tags for the other fields as well.

agent/handlers/v1/response_test.go
```diff
@@ -0,0 +1,171 @@
...
+	taskARN        = "t1"
+	family         = "sleep"
+	version        = "1"
+	containerID    = "cid"
+	containerName  = "sleepy"
+	eniIPv4Address = "10.0.0.2"
+)
+
+func TestTaskResponse(t *testing.T) {
+	expectedTaskResponseJSONString :=
```
jahkeup: Instead of testing for string exactness, this should preferably test for the shape of the data: the expected keys and their values (so as to not test serialization/deserialization string format). This certainly handles the case at hand, but isn't very maintainable.

jahkeup: It's worth noting that adding fields or reordering (through any cause) will make these fields unaligned and test fail.

haikuoliu:
@aaithal @adnxn @petderek 
I have added the unit tests and json tags for all fields.

Instead of unmarshalling raw string, I choose to marshal the struct and then compare it with the raw string, because I found that when I unmarshalled the string, it'll be marshalled successfully when either json tag matches or struct attribute matches, and it's case insensitive, so it cannot prevent the `DockerId` -> `DockerID` case. 

I will add  functional test in a separate PR, so please review, thanks!

haikuoliu:
@jahkeup I have addressed your comments and added the changelog.

PR #1498, opened 2018-08-02
dockerapi: Migrated Events to Docker SDK

kavoor:
Migrated Events to Docker SDK
### Summary

 Migrated Events to Docker SDK
### Implementation details

- Migrated AddEventEventListener/RemoveEventListener from go-dockerclient to Events from the Docker SDK. 
- docker.APIEvents --> events.Message
- The new channels returned from Events contains values and not pointers, so in order to receive them in StartListening(), you must force an evaluation of the variable by adding it as a parameter to the function.
  - This is because the goroutine is accessing the variable 'event' via the closure, and this reference is evaluated when its reached the event variable in CopyEvents. The event variable is being updated by the loop, so there's a race between the loop updates and the goroutines that have started.

Link to branch: https://github.com/kavoor/amazon-ecs-agent/tree/containerEvent
### Testing

Enhancement - Migrated Events to Docker SDK
### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -900,23 +900,29 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
 	go func() {
 		for {
-			err := <-eventErr
-			if err == io.EOF {
-				// If error processed is EOF; return
-				seelog.Info("DockerGoClient: All events have been received.")
-				cancel()
+			select{
+				case err := <-eventErr:
+				if err == io.EOF {
+					// If error processed is EOF, ignore and continue listening
+					seelog.Info("DockerGoClient: All events have been received.")
+				} else {
+					// If error processed is not EOF, need to reopen channel
+					seelog.Errorf("DockerGoClient: error while listening to Docker Events : %v", err)
+					nextCtx, nextCancel := context.WithCancel(ctx)
+					seelog.Info("******* Reopen event stream")
+					dockerEvents, eventErr = client.Events(nextCtx, types.EventsOptions{})
+					// Cache the event from docker client.
+					seelog.Info("******* Start Listening to event stream")
+					go buffer.StartListening(nextCtx, dockerEvents)
+					// Close previous stream after starting to listen on new one
+					seelog.Info("******* Cancel previous event stream")
+					cancel()
+					// Reassign cancel variable next Cancel function to setup next iteration of loop.
+					cancel = nextCancel
+				}
+			case <-ctx.Done():
+				seelog.Info("******* Engine context cancelled event stream")
 				return
-			} else {
-				// If error processed is not EOF, need to reopen channel
-				seelog.Errorf("DockerGoClient: error while listening to Docker Events : %v", err)
-				nextCtx, nextCancel := context.WithCancel(ctx)
-				dockerEvents, eventErr = client.Events(nextCtx, types.EventsOptions{})
-				// Cache the event from docker client.
-				go buffer.StartListening(nextCtx, dockerEvents)
-				// Close previous stream after starting to listen on new one
-				cancel()
-				// Reassign cancel variable next Cancel function to setup next iteration of loop.
-				cancel = nextCancel
 			}
 		}
 	}()
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -42,6 +42,7 @@ import (
 
 	"github.com/cihub/seelog"
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/events"
 	"github.com/docker/docker/api/types/volume"
 	docker "github.com/fsouza/go-dockerclient"
 )
@@ -258,14 +259,14 @@ func NewDockerGoClient(clientFactory clientfactory.Factory, sdkclientFactory sdk
 	client, err := clientFactory.GetDefaultClient()
 
 	if err != nil {
-		seelog.Errorf("DockerGoClient: go-dockerclient unable to connect to Docker daemon. " +
+		seelog.Errorf("DockerGoClient: go-dockerclient unable to connect to Docker daemon. "+
 			"Ensure Docker is running: %v", err)
 		return nil, err
 	}
 	sdkclient, err := sdkclientFactory.GetDefaultClient()
 
 	if err != nil {
-		seelog.Errorf("DockerGoClient: Docker SDK client unable to connect to Docker daemon. " +
+		seelog.Errorf("DockerGoClient: Docker SDK client unable to connect to Docker daemon. "+
 			"Ensure Docker is running: %v", err)
 		return nil, err
 	}
@@ -274,13 +275,13 @@ func NewDockerGoClient(clientFactory clientfactory.Factory, sdkclientFactory sdk
 	// to ensure it's up.
 	err = client.Ping()
 	if err != nil {
-		seelog.Errorf("DockerGoClient: go-dockerclient unable to ping Docker daemon. " +
+		seelog.Errorf("DockerGoClient: go-dockerclient unable to ping Docker daemon. "+
 			"Ensure Docker is running: %v", err)
 		return nil, err
 	}
 	_, err = sdkclient.Ping(ctx)
 	if err != nil {
-		seelog.Errorf("DockerGoClient: Docker SDK client unable to ping Docker daemon. " +
+		seelog.Errorf("DockerGoClient: Docker SDK client unable to ping Docker daemon. "+
 			"Ensure Docker is running: %v", err)
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -882,36 +883,34 @@ func getMetadataHealthCheck(dockerContainer *docker.Container) apicontainer.Heal
...
 	buffer := NewInfiniteBuffer()
+	eventErr := make(<-chan error)
 
-	err = client.AddEventListener(dockerEvents)
-	if err != nil {
-		seelog.Errorf("DockerGoClient: unable to add a docker event listener: %v", err)
+	dockerEvents, eventErr = client.Events(ctx,  types.EventsOptions{})
+	select {
+	case err = <-eventErr:
+		seelog.Errorf("DockerGoClient: error while : %v", err)
```
haikuoliu: Do you mean `error while listening to Docker events`?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -882,36 +883,34 @@ func getMetadataHealthCheck(dockerContainer *docker.Container) apicontainer.Heal
...
 		return nil, err
```
haikuoliu: Correct me if I am wrong, but I think the `eventErr` is not used in this way according to the [SDK doc](https://github.com/moby/moby/blob/75fe41444009947d319f76e68b5e1f2124c79b4d/client/events.go#L19):

>Events returns a stream of events in the daemon. It's up to the caller to close the stream by cancelling the context. Once the stream has been completely read an io.EOF error will be sent over the error channel. If an error is sent all processing will be stopped. It's up to the caller to reopen the stream in the event of an error by reinvoking this method.

For your implementation, the err case will almost never be executed, because when the `eventErr` is returned, it's empty, and `select` will execute `default` case. But what if docker send back something via `eventErr` after you execute the `default` case? As SDK suggested, you probably need to:
1. Have a go routine listen to the `eventErr` channel.
2. Once we get error from `eventErr`, do something like logging and reopen the `dockerEvents` channel.

kavoor: Yes that makes sense. If, however, we get an error that is not EOF, does it make sense to reopen the channel and continue receiving events, or to propagate the error back up to return to ContainerEvents?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -882,38 +883,53 @@ func getMetadataHealthCheck(dockerContainer *docker.Container) apicontainer.Heal
...
+
+	// Cache the event from docker client. Channel closes when an error is passed to eventErr.
+	go buffer.StartListening(derivedCtx, dockerEvents)
+
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
+		go func(){
+			for {
+				err := <-eventErr
+				// Need to cancel context to close the stream
+				cancel()
```
sharanyad: In the case of not EOF, should we cancel previous derived ctx after `StartListening` call so that we don't lose any events just in case?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -882,38 +883,53 @@ func getMetadataHealthCheck(dockerContainer *docker.Container) apicontainer.Heal
...
-		client.RemoveEventListener(dockerEvents)
-	}()
+	derivedCtx, cancel := context.WithCancel(ctx)
+	dockerEvents, eventErr := client.Events(derivedCtx, types.EventsOptions{})
+
+	// Cache the event from docker client. Channel closes when an error is passed to eventErr.
+	go buffer.StartListening(derivedCtx, dockerEvents)
+
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
+		go func(){
```
petderek: go fmt this

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -882,38 +883,53 @@ func getMetadataHealthCheck(dockerContainer *docker.Container) apicontainer.Heal
...
+			for {
+				err := <-eventErr
+				// Need to cancel context to close the stream
+				cancel()
+				if err == io.EOF {
```
petderek: why is EOF okay?

kavoor: According to the Events() specification, once the stream has been completely read an io.EOF error will be sent over the error channel, and it is safe to close the channel by cancelling the context.
 This use case is currently not possible because it relates to a certain set of flags you pass to docker events, --until (stream events until this timestamp) and --since (show all events created since timestamp). Currently we do not pass those flags in to maintain go-dockerclient functionality. In order to use this functionality, we would just need to pass in the requisite flag to the ContainerEvents() function.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -900,23 +900,29 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
 	go func() {
 		for {
-			err := <-eventErr
-			if err == io.EOF {
-				// If error processed is EOF; return
-				seelog.Info("DockerGoClient: All events have been received.")
-				cancel()
+			select{
+				case err := <-eventErr:
```
haikuoliu: nit: the `case` statement should be aligned with `select`

agent/dockerclient/dockerapi/docker_client.go
haikuoliu: To keep the same functionality, when you get an EOF error, you can just return instead of reopening, like what fsouza/go-dockerclient does:
https://github.com/fsouza/go-dockerclient/blob/11e1e492616b3d3b7046fceb0f1e7e02f9eef0d8/event.go#L229

Although the EOF error will never happen theoretically.

agent/dockerclient/dockerapi/docker_events_buffer.go
```diff
@@ -51,15 +52,25 @@ func NewInfiniteBuffer() *InfiniteBuffer {
...
-		go buffer.CopyEvents(event)
+// When context is cancelled, stop listening
+func (buffer *InfiniteBuffer) StartListening(ctx context.Context, eventChan <-chan events.Message) {
+	for {
+		select {
+		// If context is cancelled, return
+		case <-ctx.Done():
+			return
+		default:
+			for event := range eventChan {
```
haikuoliu: You already have a for loop outside, so you don't need to do a for loop here, the go routine will be stuck here forever and you won't get a chance to return when context is done. 

Also, when the context is done, we need to make sure all the events in the channel are read so that we don't miss anything,  below is my suggestion:

```
func (buffer *InfiniteBuffer) StartListening(ctx context.Context, eventChan <-chan events.Message) {
   for {
      select {
      // If context is cancelled, return
      case <-ctx.Done():
         select {
         // Ensure all events are read in the channel before we return
         case event <- eventChan:
            go func(j events.Message) {
               go buffer.CopyEvents(&j)
            }(event)
         default:
            return
         }
      case event <- eventChan:
         go func(j events.Message) {
            go buffer.CopyEvents(&j)
         }(event)
      }
   }
}
```

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -882,38 +883,58 @@ func getMetadataHealthCheck(dockerContainer *docker.Container) apicontainer.Heal
...
+	// Cache the event from docker client. Channel closes when an error is passed to eventErr.
+	go buffer.StartListening(derivedCtx, dockerEvents)
+
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
 	go func() {
-		<-ctx.Done()
-		client.RemoveEventListener(dockerEvents)
+		for {
+			select{
+				case err := <-eventErr:
```
richardpen: Please gofmt this

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -882,38 +883,58 @@ func getMetadataHealthCheck(dockerContainer *docker.Container) apicontainer.Heal
...
+				if err == io.EOF {
+					// If error processed is EOF, ignore and continue listening
+					seelog.Info("DockerGoClient: All events have been received.")
+				} else {
+					// If error processed is not EOF, need to reopen channel
```
richardpen: Can you wrap this as a method?

agent/dockerclient/dockerapi/docker_events_buffer.go
```diff
@@ -51,15 +52,32 @@ func NewInfiniteBuffer() *InfiniteBuffer {
...
-		go buffer.CopyEvents(event)
+// When context is cancelled, stop listening
+func (buffer *InfiniteBuffer) StartListening(ctx context.Context, eventChan <-chan events.Message) {
+	for {
+		select {
+		// If context is cancelled, return
+		case <-ctx.Done():
+			select {
+			// Ensure all events are read in the channel before we return
+			case event := <-eventChan:
```
richardpen: I don't think this is necessary, as this only reads one more event from the channel, also in the previous code the new channel was opened first and then the context was canceled.

agent/dockerclient/dockerapi/docker_events_buffer.go
```diff
@@ -51,15 +52,32 @@ func NewInfiniteBuffer() *InfiniteBuffer {
...
+				go func(j events.Message) {
+					go buffer.CopyEvents(&j)
+				}(event)
+			default:
+				return
+			}
+			return
+		case event := <-eventChan:
+			go func(j events.Message) {
```
richardpen: Why not just `go buffer.CopyEvents(&event)`?

yunhee-l:
Closing, duplicate of https://github.com/aws/amazon-ecs-agent/pull/1560.

PR #1504, opened 2018-08-06
handlers: introduce v3 metadata handlers

haikuoliu:
### Summary

This PR introduces v3 metadata handlers that share the same format and endpoint as V2 metadata handlers, it will work for task with all network modes.

### Implementation details

1. To reuse the v2 handlers:
    + Expose some methods in v2 handlers.
    + To simplify code logic, modified `TaskEngineState` API. This can also be done by modifying existing methods in v2 handlers, but I don't want to make too many changes to v2 handlers because it brings risk of breaking existing API.

2. Move `limitReachedHandler` to its own file.

3. Use `gorilla/mux` in `v2SetupServer`:
    + Use `gorilla/mux` variable to get `Credential ID`/`Docker ID`/`v3 endpoint ID`.
    + Define paths in each handler file so that `v2SetupServer` can use them directly.
    + To keep the user experience for v2 handlers and provide similar user experience for v3 handlers, use regular expression to support the following cases: 
        + The `Docker ID`/`Task ARN`/`Credential ID` is empty in v2 handlers, example: `/v2/metadata/` -> http 400 error
        + The v3 endpoint ID is empty in v3 handlers: `/v3/`-> http 400 error, example: `/v3//task` -> 400 error

5. More tests: 
    + Tests for marshalling/unmarshalling v2 responses.
    + Tests for new handlers.
    + Tests for the paths where we get http 400 error.
    + Tests for the paths where we get http 404 error.
    + Tests for helper functions.
    + Manual tests for v3 handlers.
    + Will add functional tests in a separate PR.

### Testing

* Feature - Add support to an existing HTTP endpoint so that tasks with all network modes (awsvpc, bridge and host) can query metadata

### Licensing


agent/handlers/v2/task_container_stats_handler.go
```diff
@@ -22,13 +22,23 @@ import (
...
+	statsContainerIDMuxName = "statsContainerIDMuxName"
 
-// StatsHandler creates response for 'v2/stats' API.
-func StatsHandler(state dockerstate.TaskEngineState, statsEngine stats.Engine) func(http.ResponseWriter, *http.Request) {
+	// TaskStatsPath specifies the relative URI path for serving task stats.
+	TaskStatsPath = "/v2/stats"
+)
+
+// ContainerStatsPath specifies the relative URI path for serving container stats.
+var ContainerStatsPath = "/v2/stats/" + utils.ConstructMuxVar(statsContainerIDMuxName, utils.AnythingButSlashRegEx)
```
fenxiong: same: AnythingRegEx

agent/handlers/v2_server_setup_test.go
```diff
@@ -500,3 +504,203 @@ func TestTaskStats(t *testing.T) {
...
+	state := mock_dockerstate.NewMockTaskEngineState(ctrl)
+	auditLog := mock_audit.NewMockAuditLogger(ctrl)
+	statsEngine := mock_stats.NewMockEngine(ctrl)
+
+	server := v2SetupServer(credentials.NewManager(), auditLog, state, clusterName, statsEngine,
+		config.DefaultTaskMetadataSteadyStateRate, config.DefaultTaskMetadataBurstRate)
+
+	for _, testPath := range testPaths {
+		t.Run(fmt.Sprintf("Test path: %s", testPath), func(t *testing.T) {
+			// Make every possible call to state fail
```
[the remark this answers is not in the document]
haikuoliu: Done

agent/handlers/v2/task_container_metadata_handler.go
```diff
@@ -18,19 +18,27 @@ import (
...
+	// for container metadata.
+	metadataContainerIDMuxName = "metadataContainerIDMuxName"
 
-// TaskContainerMetadataHandler returns the handler method for handling task metadata requests.
+	// TaskMetadataPath specifies the relative URI path for serving task metadata.
+	TaskMetadataPath = "/v2/metadata"
+)
+
+// ContainerMetadataPath specifies the relative URI path for serving container metadata.
+var ContainerMetadataPath = "/v2/metadata/" + utils.ConstructMuxVar(metadataContainerIDMuxName, utils.AnythingButSlashRegEx)
```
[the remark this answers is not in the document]
haikuoliu: Good catch! I added `TaskMetadataPathWithSlash` and `TaskStatsPathWithSlash` to cover the edge cases, and replaced `AnythingButSlashRegEx` with `AnythingButEmptyRegEx` to ensure the same behavior as previous implementation.

Changed and added some test cases, and added all test cases for v2 handlers to a branch with previous implementation to make sure we don't break anything: https://github.com/haikuoliu/amazon-ecs-agent/commits/v3-metadata-task-test

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -494,34 +494,20 @@ func (state *DockerTaskEngineState) storeV3EndpointIDToDockerIDUnsafe(v3Endpoint
...
-		return nil, false
-	}
-
-	container, ok := state.idToContainer[dockerID]
-	if !ok {
-		return nil, false
-	}
-
-	return container, ok
+	return dockerID, ok
```
sharanyad: this can just be `return  state.v3EndpointIDToDockerID[v3EndpointID]`

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -494,34 +494,20 @@ func (state *DockerTaskEngineState) storeV3EndpointIDToDockerIDUnsafe(v3Endpoint
...
 	defer state.lock.RUnlock()
 
 	taskArn, ok := state.v3EndpointIDToTask[v3EndpointID]
-	if !ok {
-		return nil, false
-	}
-
-	task, ok := state.tasks[taskArn]
-	return task, ok
+	return taskArn, ok
```
sharanyad: this can just be `return state.v3EndpointIDToTask[v3EndpointID]`

agent/handlers/v2/response_test.go
```diff
@@ -188,3 +188,203 @@ func TestContainerResponse(t *testing.T) {
...
+		"Containers": []interface {}{
+			map[string]interface {}{
+				"DockerId": containerID,
+				"Name": containerName,
+				"DockerName": containerName,
+				"Image": imageName,
+				"ImageID": imageID,
+				"Ports": []interface {}{
+					map[string]interface {}{
+						"HostPort": float64(80),
```
petderek: why floats?

agent/handlers/utils/helpers_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestConstructMuxVar(t *testing.T) {
```
petderek: maybe include the patterns available as constants?

agent/handlers/limit_reached_handler.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"net/http"
+
+	"github.com/aws/amazon-ecs-agent/agent/logger/audit"
+	"github.com/aws/amazon-ecs-agent/agent/logger/audit/request"
+)
+
+// limitReachedHandler logs the throttled request in the credentials audit log
+func limitReachedHandler(auditLogger audit.AuditLogger) func(http.ResponseWriter, *http.Request) {
```
sharanyad: Can this method be added to `agent/handlers/utils/helpers.go`?

agent/handlers/v2/response_test.go
```diff
@@ -188,3 +188,203 @@ func TestContainerResponse(t *testing.T) {
 		})
 	}
 }
+
+func TestTaskResponseMarshal(t *testing.T) {
```
petderek: Thanks for adding tests over existing workflows! 👍

agent/handlers/v2_server_setup.go
```diff
@@ -21,78 +21,87 @@ import (
...
 	steadyStateRate int,
 	burstRate int) *http.Server {
-	serverMux := http.NewServeMux()
-	// Credentials handlers
-	serverMux.HandleFunc(credentials.V1CredentialsPath,
+	muxRouter := mux.NewRouter()
+
+	// Set this so that for request like "/v3//metadata/task", the Agent will pass
+	// it to task metadata handler instead of returning a 301 error.
+	muxRouter.SkipClean(true)
```
petderek: Is this a new feature or are we preserving existing behavior?

agent/handlers/v2_server_setup_test.go
```diff
@@ -381,7 +385,52 @@ func parseResponseBody(body *bytes.Buffer) (*credentials.IAMRoleCredentials, err
 	return &creds, nil
 }
 
-func TestTaskMetadata(t *testing.T) {
+func TestV2TaskMetadata(t *testing.T) {
```
petderek: Thanks for these tests as well!

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -494,34 +494,20 @@ func (state *DockerTaskEngineState) storeV3EndpointIDToDockerIDUnsafe(v3Endpoint
...
+func (state *DockerTaskEngineState) DockerIDByV3EndpointID(v3EndpointID string) (string, bool) {
 	state.lock.RLock()
 	defer state.lock.RUnlock()
 
 	dockerID, ok := state.v3EndpointIDToDockerID[v3EndpointID]
-	if !ok {
-		return nil, false
-	}
-
-	container, ok := state.idToContainer[dockerID]
```
petderek: this looks harmless but this change scares me. Is it ever different?

agent/handlers/v3/task_metadata_handler.go
```diff
@@ -0,0 +1,68 @@
...
+const v3EndpointIDMuxName = "v3EndpointIDMuxName"
+
+// TaskMetadataPath specifies the relative URI path for serving task metadata.
+var TaskMetadataPath = "/v3/" + utils.ConstructMuxVar(v3EndpointIDMuxName, utils.AnythingButSlashRegEx) + "/task"
+
+// TaskMetadataHandler returns the handler method for handling task metadata requests.
+func TaskMetadataHandler(state dockerstate.TaskEngineState, cluster string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		taskARN, err := getTaskARN(r, state)
+		if err != nil {
```
sharanyad: Are there unit tests for these negative cases as well?

agent/handlers/v2/response_test.go
haikuoliu: Because when we unmarshall something to `map[string]interface{}`, the number will be unmarshalled to `float64` type, make it float here so that we can compare the two maps. This was explained in my [previous PR](https://github.com/aws/amazon-ecs-agent/blob/02a6a9fff9bee5e0bbd2b936d4ceee50207595f8/agent/handlers/v1/response_test.go#L50), it would be redundant if I add comments for every case like this.

agent/handlers/v2_server_setup.go
haikuoliu: This is specifically for the V3 handlers, it's used to handle the case where the v3 endpoint ID is empty in the path (path like `/v3//metadata/task`), it sounds reasonable for us to return 400 bad request instead of 301 for empty v3 endpoint ID, I discussed this with @richardpen.

agent/engine/dockerstate/docker_task_engine_state.go
haikuoliu: When I do this, my IDE tells me `not enough arguments to return`, I think this is because 
you can use map in two ways: 
1. `dockerID := state.v3EndpointIDToDockerID[v3EndpointID]`
2. `dockerID, ok := state.v3EndpointIDToDockerID[v3EndpointID]`

It's unclear to the IDE (and programmers) when you just do `return state.v3EndpointIDToDockerID[v3EndpointID]`, so I prefer to keep it.

Same below.

agent/handlers/limit_reached_handler.go
haikuoliu: Yeah it could work in both ways, but I think it falls into the `handler` category, so keep it in the `handlers` package seems more reasonable to me.

agent/engine/dockerstate/docker_task_engine_state.go
haikuoliu: The purpose of this change is to reduce redundant code logic (because only `dockerID` is needed for V3 handler) and not make too many changes to the existing v2 handlers code. The V3 handler package is the only one that uses this method, so it's safe to change this.

This is also explained in the `Implementation details`.

agent/handlers/utils/helpers_test.go
haikuoliu: Added tests cases for all RegEx constants.

agent/handlers/limit_reached_handler.go
sharanyad: The `utils` package is also within `handlers`. `limitReachedHandler` seems more of like a helper method. Having a separate file just for it is not of much value imo.

adnxn:
> run gofmt to refactor the code

@haikuoliu: hrm we should get your editor setup to gofmt on file saves, maybe it already is 😕

agent/handlers/v3/container_metadata_handler.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+
+		seelog.Infof("V3 container metadata handler: handling request for container '%s'", containerID)
+
+		// v3 handler shares the same container metadata response format with v2 handler.
+		v2.WriteContainerMetadataResponse(w, containerID, state)
+	}
+}
+
+func getContainerID(r *http.Request, state dockerstate.TaskEngineState) (string, error) {
```
richardpen: Can you move this function, `getTaskArn` and `getV3EndpointID` into a separate file, also they all should be covered by unit test?

agent/handlers/v2_server_setup.go
```diff
@@ -21,78 +21,87 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
+	handlersutils "github.com/aws/amazon-ecs-agent/agent/handlers/utils"
```
richardpen: The name of the file is confusing, as both v3 and v2 server are set in this file, can you restructure this to the following:
```
-handlers
-----server_setup.go(call v1/v2/v3 server_setup to start all servers)
-----v1
---------v1_server_setup.go
-----v2
----------v2_server_setup.go
-----v3
----------v3_server_setup.go
```

agent/handlers/v2/task_container_stats_handler.go
```diff
@@ -37,16 +50,25 @@ func StatsHandler(state dockerstate.TaskEngineState, statsEngine stats.Engine) f
...
 
-		writeTaskStatsResponse(w, taskARN, state, statsEngine)
+		seelog.Infof("V2 task/container stats handler: handling request for task '%s'", taskARN)
+		WriteTaskStatsResponse(w, taskARN, state, statsEngine)
 	}
 }
 
-func writeTaskStatsResponse(w http.ResponseWriter,
+// getStatsContainerID gets the container ID for container stats
+func getStatsContainerID(r *http.Request) string {
```
richardpen: You can get rid of this function if you have a general method to retrieve from a request. See comments below.

agent/handlers/v2/task_container_stats_handler.go
```diff
@@ -22,13 +22,26 @@ import (
...
-func StatsHandler(state dockerstate.TaskEngineState, statsEngine stats.Engine) func(http.ResponseWriter, *http.Request) {
+	// TaskStatsPath specifies the relative URI path for serving task stats.
+	TaskStatsPath = "/v2/stats"
+
+	// TaskStatsPathWithSlash specifies the relative URI path for serving task stats.
+	TaskStatsPathWithSlash = "/v2/stats/"
+)
+
+// ContainerStatsPath specifies the relative URI path for serving container stats.
+var ContainerStatsPath = "/v2/stats/" + utils.ConstructMuxVar(statsContainerIDMuxName, utils.AnythingButEmptyRegEx)
```
richardpen: you can replace `/v2/stats/` with the constants.

agent/handlers/v2/task_container_metadata_handler.go
```diff
@@ -18,19 +18,30 @@ import (
...
-// TaskContainerMetadataHandler returns the handler method for handling task metadata requests.
+	// TaskMetadataPath specifies the relative URI path for serving task metadata.
+	TaskMetadataPath = "/v2/metadata"
+
+	// TaskMetadataPathWithSlash specifies the relative URI path for serving task metadata.
+	TaskMetadataPathWithSlash = "/v2/metadata/"
+)
+
+// ContainerMetadataPath specifies the relative URI path for serving container metadata.
+var ContainerMetadataPath = "/v2/metadata/" + utils.ConstructMuxVar(metadataContainerIDMuxName, utils.AnythingButEmptyRegEx)
```
richardpen: This can be replaced with the `TaskMetadataPathWithSlash`.

agent/handlers/v2/task_container_metadata_handler.go
```diff
@@ -40,21 +51,21 @@ func TaskContainerMetadataHandler(state dockerstate.TaskEngineState, cluster str
...
 }
 
-func getContainerID(reqURL *url.URL, prefix string) string {
-	if strings.HasPrefix(reqURL.Path, prefix+"/") {
-		return reqURL.String()[len(prefix+"/"):]
-	}
-
-	return ""
+// getMetadataContainerID gets container ID for container metadata.
+func getMetadataContainerID(r *http.Request) string {
```
richardpen: Same as below, this can be removed with a general function.

haikuoliu:
@richardpen I have addressed your comments, made the following changes as we have discussed offline:
1. Created handler setup methods that add handlers for each handler version.
2. Moved the license handler to v1 package as it's only used there.
3. Changed `getStatsContainerID`, `getMetadataContainerID`, etc. to generic method.
4. Created `helper.go` in v2 and v3 package, moved some helper methods into it.
5. Changed the v1 and v2 server to a more readable name: introspection server and task server.
6. Other minor refactor.

@sharanyad I have moved the `limitReachedHandler` to utils package.

PR #1509, opened 2018-08-08
Scripts to upload linux/windows images used for agent tests to ECR

yumex93:
### Summary
Add cloudformation template and related scripts that creates a AWS codebuild project which build images which are used by functional tests, upload images to ECR and replicate to regions in other partitions like cn-north-1.

### Implementation details
The cloudformation template codebuild-ecr-upload.yml is used to create the codebuild project.
buildspec-ecr-upload.yml is the file telling codebuild what commands to run when building.
Add some make targets which are used by codebuild to build images and upload to ECRs.

### Testing

New tests cover the changes: 

### Licensing


Makefile
```diff
@@ -183,6 +183,12 @@ test-in-docker:
 run-functional-tests: testnnp test-registry ecr-execution-role-image telemetry-test-image
 	. ./scripts/shared_env && go test -tags functional -timeout=30m -v ./agent/functional_tests/...
 
+build-image-for-ecr: netkitten volumes-test squid awscli image-cleanup-test-images fluentd taskmetadata-validator testnnp container-health-check-image telemetry-test-image ecr-execution-role-image-for-upload
```
petderek: add .PHONY tags to these

scripts/setup-ecr
```diff
@@ -0,0 +1,130 @@
...
+}
+
+mirror_local_image() {
+  docker tag $1 $2
+  docker push $2
+  docker rmi $2
+}
+
+# Push to ECR in Standard Region
+for image in "amazon-ecs-netkitten" "amazon-ecs-volumes-test" "squid" "awscli" "image-cleanup-test-image1" "image-cleanup-test-image2" "image-cleanup-test-image3" "fluentd" "amazon-ecs-taskmetadata-validator" "amazon-ecs-telemetry-test" "amazon-ecs-containerhealthcheck"; do
```
petderek: Thoughts on something like `$(ls misc/)` ?

Makefile
```diff
@@ -183,6 +183,12 @@ test-in-docker:
...
+	@./scripts/setup-ecr
+
+ecr-execution-role-image-for-upload:
```
petderek: Why do we need this?

scripts/setup-ecr
```diff
@@ -0,0 +1,130 @@
...
+# language governing permissions and limitations under the License.
+
+# Run a local registry on the 'well known' port 51670 if it is not running.
+# Also push images we will need to it.
+set -e
+
+# registry:2 from 2015-09-15
+REGISTRY_IMAGE="registry@sha256:b7de4f6226df56d18f83296efa77dedf9bb72e79838167be0484a3078836fab2"
+# nginx:latest from  2015-09-15
+NGINX_IMAGE="nginx@sha256:0324afc5c8191616576f7b23b297d001609726a2f1b6561c90e229e54ab701cf"
```
petderek: I dpn;'t think we need this?

scripts/setup-ecr
```diff
@@ -0,0 +1,130 @@
...
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+# Run a local registry on the 'well known' port 51670 if it is not running.
+# Also push images we will need to it.
+set -e
+
+# registry:2 from 2015-09-15
+REGISTRY_IMAGE="registry@sha256:b7de4f6226df56d18f83296efa77dedf9bb72e79838167be0484a3078836fab2"
```
petderek: I don't think we need this?

scripts/setup-ecr
```diff
@@ -0,0 +1,130 @@
...
+mirror_local_image ${PARALLEL_PULL_FTS_UBUNTU} "${REPLICATE_REPOSITORY}/ubuntu:parallel-pull-fts"
+mirror_local_image ${PARALLEL_PULL_FTS_CONSUL} "${REPLICATE_REPOSITORY}/consul:parallel-pull-fts"
+mirror_local_image ${PARALLEL_PULL_FTS_DEBIAN} "${REPLICATE_REPOSITORY}/debian:parallel-pull-fts"
+mirror_local_image ${PARALLEL_PULL_FTS_HTTPD} "${REPLICATE_REPOSITORY}/httpd:parallel-pull-fts"
+mirror_local_image ${PARALLEL_PULL_FTS_MONGO} "${REPLICATE_REPOSITORY}/mongo:parallel-pull-fts"
+mirror_local_image ${PARALLEL_PULL_FTS_NGINX} "${REPLICATE_REPOSITORY}/nginx:parallel-pull-fts"
+mirror_local_image ${PARALLEL_PULL_FTS_REDIS} "${REPLICATE_REPOSITORY}/redis:parallel-pull-fts"
+mirror_local_image ${PARALLEL_PULL_FTS_REGISTRY} "${REPLICATE_REPOSITORY}/registry:parallel-pull-fts"
+
+# Remove the tag so this image can be deleted successfully in the docker image cleanup integ tests
```
petderek: We don't explicitly need this here -- its for actually running the tests on the same host

buildspec-ecr-upload.yml
```diff
@@ -0,0 +1,12 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - echo Logging in to Amazon ECR...
+      - aws --version
+      - docker --version
```
samuelkarp: I'd recommend using `docker version` instead of `docker --version` so you get information about the server version as well.

codebuild-ecr-upload.yml
```diff
@@ -0,0 +1,112 @@
+Description: >
+    The template used to create resources that CodePipeline needs
```
samuelkarp: Can you write a more descriptive description here?

misc/ecr-execution-role-upload/dockerfile
```diff
@@ -0,0 +1,17 @@
+# Copyright 2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
samuelkarp: Can you change this to 2018?

misc/ecr-execution-role-upload/Makefile
```diff
@@ -0,0 +1,6 @@
+.PHONY: all clean
```
samuelkarp: Can you add the license header?

scripts/setup-ecr
```diff
@@ -0,0 +1,130 @@
+#!/bin/bash
+# Copyright 2014-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
samuelkarp: Can you change this to 2018?

Makefile
yumex93: You mean ecr-execution-role-image-for-upload?. Seems we need this image and I do not want to modify original files in misc/ecr folder which not only build the image, but also upload to the account ecr. While I only need the logic to build the image.

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+  echo $_.Exception.Message -logLevel "ERROR"
+}
+
+echo "-------------------------------------------------------------------------------"
+echo "Generating Windows Images..."
+echo "-------------------------------------------------------------------------------"
+
+$ErrorActionPreference="Stop"
+
+$application = Get-WmiObject Win32_Product -filter "Name='AWS Tools for Windows'"
```
petderek: do we need to do this on every run?

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+Import-Module AWSPowerShell
+
+echo "Pulling microsoft/windowsservercore:latest..."
+docker pull microsoft/windowsservercore:latest
+
+$env:ECS_WINDOWS_TEST_DIR="${BUILD_ROOT}"
+$env:ECS_FTEST_TMP="${WORK_ROOT}\ftest_temp"
+New-Item -Path "${env:ECS_FTEST_TMP}" -Type directory
+
+echo "${BUILD_ROOT}"
```
petderek: Add context:

echo "build root: ${BUILD_ROOT}"

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+Invoke-Expression "${BUILD_ROOT}\misc\stats-windows\build.ps1"
+Invoke-Expression "${BUILD_ROOT}\misc\netkitten\build.ps1"
+Invoke-Expression "${BUILD_ROOT}\misc\windows-listen80\Setup_Listen80.ps1"
+Invoke-Expression "${BUILD_ROOT}\misc\windows-telemetry\build.ps1"
+Invoke-Expression "${BUILD_ROOT}\misc\windows-python\build.ps1"
+Invoke-Expression "${BUILD_ROOT}\misc\container-health-windows\build.ps1"
+
+$sAccessKey=(Get-SSMParameter -Name SAccessKey -WithDecryption $TRUE).Value
+$sSecretKey=(Get-SSMParameter -Name SSecretKey -WithDecryption $TRUE).Value
+$accessKey=(Get-SSMParameter -Name AccessKey -WithDecryption $TRUE).Value
```
petderek: what do you think about relying on the instance role?

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+[Environment]::SetEnvironmentVariable("Path", $env:Path + ";C:\Go\bin;${WORK_ROOT}\go\bin", [EnvironmentVariableTarget]::Machine)
+[Environment]::SetEnvironmentVariable("GOPATH", "${WORK_ROOT}\go", [EnvironmentVariableTarget]::Machine)
+$env:Path=$env:Path + ";C:\Go\bin;${WORK_ROOT}\go\bin"
+$env:GOPATH="${WORK_ROOT}\go"
+
+echo "Checking out ${GIT_REPO} at ${COMMIT_SHA}"
+git clone "${GIT_REPO}" "${BUILD_ROOT}" 2>&1 | % {$_.ToString()}
+cd "${BUILD_ROOT}"
+git config --add remote.origin.fetch '+refs/pull/*/head:refs/remotes/origin/pr/*'
+git fetch 2>&1 | % {$_.ToString()}
```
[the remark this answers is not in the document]
yumex93: You mean " 2>&1 | % {$_.ToString()}"? Git is printing some weird thing to stderr. Like 
It prints the checking out files progress to stderr instead of stdout and also something like this:

"You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout."

I have logic to check stderr in my script to check the status of running. So I need to redirect it to stdout. The % {$_.ToString()} is just to format the output.

scripts/generate_image.ps1
petderek: I mean, relying on the instance role for standard partition. These are the keys for both partitions, right? If this is more straightforward, lets keep it.

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+}
+
+echo "==============================================================================="
+echo "Building windows images for commit ${COMMIT_SHA}"
+echo "==============================================================================="
+
+# Install Git
+[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+echo "Downloading Git..."
+Invoke-WebRequest "https://github.com/git-for-windows/git/releases/download/v${GitVersion}.windows.1/Git-${GitVersion}-64-bit.exe" -OutFile "${env:TEMP}\git-installer.exe" -UseBasicParsing
```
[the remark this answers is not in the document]
yumex93: When I change it to Start-Bitstransfer, it starts to throw error " 403: does not have access".

Makefile
tehshwann: Are we able to standardize aws account specific configurations like this to a naming standard that can be A) inferred if not provided, and B) overwrite if need?

misc/ecr-execution-role-upload/dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# License is located at
+#
+#	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+FROM busybox@sha256:5551dbdfc48d66734d0f01cafee0952cb6e8eeecd1e2492240bf2fd9640c2279
```
tehshwann: Is this sha ever going to change?

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+    [string]$S_REPOSITORY,
+    [string]$R_REPOSITORY,
+    [string]$S_REGION,
+    [string]$R_REGION
+)
+
+$ErrorActionPreference = 'Continue'
+
+# Constants
+$GitVersion="2.10.1"
```
tehshwann: Since the gitversion and go version are static here, is there any reason we hit github every time rather than pulling things down once for our entire test run?

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+
+echo "-------------------------------------------------------------------------------"
+echo "Generating Windows Images..."
+echo "-------------------------------------------------------------------------------"
+
+$ErrorActionPreference="Stop"
+
+$application = Get-WmiObject Win32_Product -filter "Name='AWS Tools for Windows'"
+if ($application) {
+    $application.Uninstall()
```
tehshwann: Why are we uninstalling the AWS Tools here? This operation has been known to fail in the past

scripts/generate_image.ps1
tehshwann: Have you tried force installing over the top?
If you have two versions installed, and the latter has the commandlet, the new one should be referenced.

scripts/generate_image.ps1
```diff
@@ -0,0 +1,175 @@
...
+    [string]$S_REGION,
+    [string]$R_REGION
+)
+
+$ErrorActionPreference = 'Continue'
+
+# Constants
+$GitVersion="2.10.1"
+$GoVersion="1.9"
+$WORK_ROOT="${env:TMP}\amazon-ecs-agent"
```
tehshwann: Can we update all the path construction to use: _$PathVariable = Join-Path $BasePath $ChildPath_

Couple of PS1 nits:
- Use ``'string'`` rather than ``"string"`` when possible
- Use ``$ENV:VARNAME`` in caps
- Use ``"$VarStr"-things`` rather than ``"${VarStr}-things"``'

Example:
``$WORK_ROOT = Join-Path $ENV:TMP 'amazon-ecs-agent'``

scripts/generate_image.ps1
yumex93: I tried to have two installed, it will only reference the old version, which is preinstalled with the AMI.

misc/ecr-execution-role-upload/dockerfile
yumex93: Seems it normally will not be changed. We do not keep updating it.

scripts/generate_image.ps1
yumex93: If the old version isn't been uninstalled, the new version does not get referenced.

misc/ecr-execution-role-upload/dockerfile
petderek: This locks it to a precise revision of busybox. IDK if its a practice we want to keep or not

scripts/generate_image.ps1
tehshwann: ok sounds good

Makefile
```diff
@@ -183,6 +183,19 @@ test-in-docker:
...
+
+build-image-for-ecr: netkitten volumes-test squid awscli image-cleanup-test-images fluentd taskmetadata-validator testnnp container-health-check-image telemetry-test-image ecr-execution-role-image-for-upload
+
+ecr-execution-role-image-for-upload:
+	$(MAKE) -C misc/ecr-execution-role-upload $(MFLAGS)
+
+upload-images: build-image-for-ecr
+	@./scripts/upload-images $(STANDARD_REGION) $(STANDARD_REPOSITORY)
+
+replicate-images: build-image-for-ecr
```
petderek: what do you think about pulling images from $STANDARD_REGION and pushing them to $REPLICATE_REGION?

That way, we have identical images.

PR #1521, opened 2018-08-15
test: add functional test for v3 task endpoint

haikuoliu:
### Summary
Add function test for v3 task endpoint that was added in [this](https://github.com/aws/amazon-ecs-agent/pull/1504) PR.

### Testing

### Licensing

### DIFF
--- a/agent/functional_tests/README.md
+++ b/agent/functional_tests/README.md
@@ -23,6 +23,11 @@ Before running these tests, you should build the ECS agent and tag its image as
 
 You should also run the registry the tests pull from. This can most easily be done via `make test-registry`.
 
+You should run following commands to set up the IP tables rules:
+* `sysctl -w net.ipv4.conf.all.route_localnet=1`
+* `iptables -t nat -A PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-destination 127.0.0.1:51679`
+* `iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679`
+
 #### Environment variables
 In order to run telemetry functional test in non Amazon Linux AMI environment
 with older versions of the ECS agent (pre-1.10.0), the following environment
@@ -44,8 +49,6 @@ execution behavior:
 #### Additional setup for IAM roles
 In order to run TaskIamRole functional tests, the following steps should be
 done first:
-* Run command: `sysctl -w net.ipv4.conf.all.route_localnet=1` and
-  `iptables -t nat -A PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-destination 127.0.0.1:51679`.
 * Set the environment variable of IAM roles the test will use:
   `export TASK_IAM_ROLE_ARN="iam role arn"`. The role should have the
   `ec2:DescribeRegions` permission and have a trust relationship with
@@ -54,8 +57,6 @@ done first:
   In this case, the IAM role should additionally have the
   `iam:GetInstanceProfile` permission.
 * To skip the test `TestTaskIAMRolesDefaultNetworkMode`, please set `TEST_DISABLE_TASK_IAM_ROLE` to be true.
-* Testing under net=host network mode requires additional commands:
-  `iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679`
 * To skip the test `TestTaskIAMRolesNetHostMode`, please set `TEST_DISABLE_TASK_IAM_ROLE_NET_HOST` to be true.
 
 
--- a/agent/functional_tests/testdata/taskdefinitions/v3-task-endpoint-validator-windows/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/v3-task-endpoint-validator-windows/task-definition.json
@@ -1,0 +1,26 @@
+{
+  "family": "ecsftest-v3-task-endpoint-validator-windows",
+  "containerDefinitions": [{
+    "image": "amazon/amazon-ecs-v3-task-endpoint-validator-windows",
+    "name": "v3-task-endpoint-validator-windows",
+    "memory": 512,
+    "cpu": 1024,
+    "healthCheck": {
+      "command": ["CMD-SHELL", "echo hello"],
+      "interval": 5,
+      "timeout": 5,
+      "retries": 2,
+      "startPeriod": 1
+    },
+    "entryPoint": ["powershell"],
+    "command": [".\\application.ps1; $env:AWS_REGION=\"$$$TEST_REGION$$$\";.\\v3-task-endpoint-validator-windows.exe; exit $LASTEXITCODE"],
+    "logConfiguration": {
+      "logDriver": "awslogs",
+      "options": {
+        "awslogs-group":"ecs-functional-tests",
+        "awslogs-region":"$$$TEST_REGION$$$",
+        "awslogs-stream-prefix":"$$$TEST_AWSLOGS_STREAM_PREFIX$$$"
+      }
+    }
+  }]
+}
--- a/agent/functional_tests/tests/functionaltests_unix_test.go
+++ b/agent/functional_tests/tests/functionaltests_unix_test.go
@@ -409,7 +409,7 @@ func TestTaskIAMRolesDefaultNetworkMode(t *testing.T) {
 }
 
 func taskIAMRoles(networkMode string, agent *TestAgent, t *testing.T) {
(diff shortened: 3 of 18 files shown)


misc/v3-task-endpoint-validator-windows/v3-task-endpoint-validator-windows.go
```diff
@@ -0,0 +1,387 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
richardpen: What's the difference of the file between windows and linux? Also is it possible to reuse the taskmetadata validator file?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -461,6 +461,79 @@ func taskIAMRoles(networkMode string, agent *TestAgent, t *testing.T) {
...
+		for _, env := range containerMetaData.Config.Env {
+			if strings.HasPrefix(env, "ECS_CONTAINER_METADATA_URI=") {
+				v3TaskEndpointEnabled = true
+				break
+			}
+		}
+	}
+	if !v3TaskEndpointEnabled {
+		task.Stop()
+		t.Fatalf("Could not found ECS_CONTAINER_METADATA_URI in the container environment variable")
```
richardpen: nit: `t.Fatal`, same below.

agent/functional_tests/tests/functionaltests_windows_test.go
```diff
@@ -178,7 +178,54 @@ func taskIamRolesTest(networkMode string, agent *TestAgent, t *testing.T) {
 	if containerMetaData.State.ExitCode != 42 {
 		t.Fatalf("Container exit code non-zero: %v", containerMetaData.State.ExitCode)
 	}
+}
+
+func TestV3TaskEndpointDefaultNetworkMode(t *testing.T) {
```
richardpen: This looks like duplicate code, can you extract them to be a function in the functional_test.go, so that they can be used in both windows and linux test.

misc/v3-task-endpoint-validator-windows/setup-v3-task-endpoint-validator.ps1
```diff
@@ -0,0 +1,37 @@
...
+$buildimage="golang:1.7-windowsservercore"
+docker pull $buildimage
+
+docker run `
+  --volume ${PSScriptRoot}:C:\ecs `
+  $buildimage `
+  powershell ${buildscript}
+
+Invoke-Expression "docker build -t amazon/amazon-ecs-v3-task-endpoint-validator-windows --file ${PSScriptRoot}\v3-task-endpoint-validator-windows.dockerfile ${PSScriptRoot}"
+$ErrorActionPreference = $oldPref
```
richardpen: Just for understanding, why this line is required?

misc/v3-task-endpoint-validator-windows/v3-task-endpoint-validator-windows.go
haikuoliu: The difference between windows and linux: we need to verify three network modes for linux, but we only need to verify one network mode for Windows, also windows has a different task name. 
This means the verification logic and some of the metadata fields are different. Also we may need to do some extra configuration to reuse the file in another folder.

is it possible to reuse the task metadata validator file: theoretically it's possible, but it will result in a very confusing configuration and verification logic.

In sum, I don't think it's worth to put them together.

agent/functional_tests/tests/functionaltests_windows_test.go
haikuoliu: It's doable, but it will be confusing as we have to pass in a lot of parameters, here are the differences between linux and windows (maybe we will have more in the future):
1. Linux has network setting, but windows doesn't, and start a awsvpc mode needs to call a different function.
2. Their aws logs prefix name is different.
3. The container names are different.

If you still feel strongly that we need to abstract them, I can do it, otherwise I'd prefer not.

agent/functional_tests/testdata/taskdefinitions/v3-task-endpoint-validator-windows/task-definition.json
```diff
@@ -0,0 +1,26 @@
+{
+  "family": "ecsftest-v3-task-endpoint-validator-windows",
+  "containerDefinitions": [{
+    "image": "amazon/amazon-ecs-v3-task-endpoint-validator-windows",
+    "name": "v3-task-endpoint-validator-windows",
+    "memory": 512,
+    "cpu": 1024,
+    "healthCheck": {
+      "command": ["CMD-SHELL", "echo hello"],
```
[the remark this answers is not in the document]
haikuoliu: Discussed offline, health check will be keeped.

misc/v3-task-endpoint-validator-windows/setup-v3-task-endpoint-validator.ps1
haikuoliu: Because we have set up the error action preference in previous lines: `$ErrorActionPreference = 'Stop'`, we set it back after we finish running this script, so that we won't change the behavior of the environment after testing.

misc/v3-task-endpoint-validator/v3-task-endpoint-validator.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	return body, nil
+}
+
+func notEmptyErrMsg(fieldName string) error {
+	return fmt.Errorf("field %s should not be empty", fieldName)
+}
+
+func isFieldEmpty(rawMsgMap map[string]json.RawMessage, fieldName string) error {
```
fenxiong: nit: isFieldEmpty and isFieldEqual should probably not start with 'is' since there are not returning bool. and 'fieldEmpty' should probably be 'fieldNotEmpty'

agent/functional_tests/README.md
```diff
@@ -23,6 +23,11 @@ Before running these tests, you should build the ECS agent and tag its image as
 
 You should also run the registry the tests pull from. This can most easily be done via `make test-registry`.
 
+You should run following commands to set up the IP tables rules:
```
sharanyad: Is this needed for all the tests now?

agent/functional_tests/testdata/taskdefinitions/v3-task-endpoint-validator-windows/task-definition.json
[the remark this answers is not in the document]
sharanyad: just curious, why is it needed?

agent/functional_tests/tests/functionaltests_windows_test.go
sharanyad: Instead of passing parameters to the common method, can we have them as constants in the respective files? 
I'd strongly recommend reusing test code as much as possible.

agent/functional_tests/README.md
haikuoliu: Yes I made it required now. 

If I make it optional, we need to have a setting like `TEST_DISABLE_V3_ENDPOINT_BRIDGE_TEST`, and we will result in having too much test environment variables. Since we our test environment has already set up these ip table rules, make it required seems more reasonable.

agent/functional_tests/testdata/taskdefinitions/v3-task-endpoint-validator-windows/task-definition.json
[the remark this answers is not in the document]
haikuoliu: @sharanyad I think Peng meant the `healthCheeck` is not needed, but to test the health check field in task response, we need to keep this.

misc/v3-task-endpoint-validator-windows/v3-task-endpoint-validator-windows.go
```diff
@@ -0,0 +1,387 @@
...
+
+	if err = fieldEqual(containerMetadataResponseMap, "DesiredStatus", "RUNNING"); err != nil {
+		return err
+	}
+
+	if err = fieldEqual(containerMetadataResponseMap, "KnownStatus", "RUNNING"); err != nil {
+		return err
+	}
+
+	if err = fieldNotEmpty(containerMetadataResponseMap, "Limits"); err != nil {
```
sharanyad: Instead of having the condition check repeated, you can simply have a list of strings with all the expected keys and do something like: 
```
for _, key := range keys {
    if err = fieldEqual(containerMetadataResponseMap, key); err != nil {
        return err
    }
}
```
This would make it easier to add more fields to these tests.

sharanyad: Similarly for `fieldEqual` checks, you can have a map of expected values too

Issue #1540, opened 2018-08-22
Add Windows Credentialspec Support


Cgettys:
### Summary

Add Windows Credentialspec support to enable Windows Auth in containers.

### Description

Although Docker for Windows supports Windows Auth through the use of credentialspec files, Amazon ECS does not support this capability.

Retaining the ability to use Windows Authentication within the docker ecosystem eliminates the need to refactor applications to use other forms of authentication. Given how ubiquitous Windows Auth is within many legacy dotnet applications, this feature would greatly streamline the migration process.

eisware:
It would be great if there were a way to just pass through the "credentialspec=file://gmsa.json" security-opt blindly and let me worry about whether the not the task is able to run successfully. I've already scripted the installation of the AD service account and creation of the credentialspec file I require when my ASG instances launch prior to initializing the ECS agent, this is literally the only thing preventing me from being able to use integrated windows authentication in my .NET Framework services. Thanks!

Cgettys:
@eisware I have used the Docker labels option and a modified version of the agent that parses the labels and injects the right label into the right place in the past. You can basically put credentialspec=file://gmsa.json into the labels, match that, and then put that where it belongs in the agent, where it's past the validation. I've done this successfully before with one or two lines of code, though unfortunately I don't have up to date source code to share, and it's not officially supported. I also recommend you reach out to your Amazon rep to get yourself added to the list of people who really want this feature, to increase the likelihood of it actually happening.
Disclaimer: I am not affiliated with Amazon. I cannot recommend you take this route. This is not officially supported and you're on your own if you do it.

eisware:
@Cgettys Thanks for the ideas :)

Cgettys:
@eisware No problem. I hit the exact same problem, went down the road of figuring out how to work around it, and thought I'd save a fellow developer the trouble since there's no downside to me doing so. Besides, the more people (especially enterprise customers) who tell them that they want this feature, the more likely they'll move it up their priority list, and that would benefit both of us.

Cgettys:
Hit the wrong button there, oops.

Cgettys:
Whoops, wrong button.

bitshop:
It seems like a few things don't have solutions discussed as @samuelkarp mentioned in #1543 

[...]

PR #1543, opened 2018-08-23
[WIP] Add credential spec support to agent

Cgettys:
### Summary

This pull request adds support for credentialspec files, addressing feature request #1540

### Implementation details

When a task definition defines a container with dockerSecurityOptions containing an entry of "credentialspec=somejsonblob", the amazon ecs agent writes that blob to {env:ProgramData}/Docker/credentialspecs/taskarn-containername.json, and replaces the entry with "credentialspec=file://taskarn-containername.json" so that Docker can consume the blob. This is necessary since at current, Docker only supports reading credentialspec files from the credentialspecs folder or from a particular registry key. 
Notes:
* task definition validation does not currently allow credentialspec to be specified as a dockerSecurityOption, so that will need to be tweaked before this is useful.
* The host must be domain joined and have the desired gMSAs installed, which this PR does not seek to address.

Work remaining: 
* Additional integration test
* Additional functional test
* Cleanup credentialspecs when task is done? (should be <10KB per file, but still)

### Testing

New tests cover the changes:  in progress (unit tests only)

* Feature - Add credentialspec support for Windows containers
### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -816,7 +816,7 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
 		}
 	}
 
-	err = task.platformHostConfigOverride(hostConfig)
+	err = task.platformHostConfigOverride(container, hostConfig)
 	if err != nil {
 		return nil, &apierrors.HostConfigError{err.Error()}
 	}
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -19,6 +19,7 @@ import (
 	"path/filepath"
 	"time"
 
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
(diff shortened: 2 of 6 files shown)


Cgettys:
@samuelkarp I took a crack at adding support for the feature I requested in #1540. I'd love feedback on whether I'm going about this the right way. Thanks so much!

samuelkarp:
Hi @Cgettys,

Thanks for sending this and apologies for the delay in response. Unfortunately, we are unable to merge your pull request right now, but we will keep it open so we can potentially merge it in the future.

As you noted, in order to accept a change like this we'll also need to change the task definition validation logic as well. Since this is introducing a new feature on the agent side that is not present with the current version of the agent, we'll need to introduce a new capability that the agent can advertise stating that it supports this feature. Because of this, we cannot merge your change until we've made the requisite changes elsewhere in ECS.

I think we'll want a different approach for the credential spec JSON blob than the one you've taken. We do not recommend using plaintext values in the task definition for sensitive information, such as credential data (see the [environment variable section](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_environment)); sensitive data in a task definition would be visible to any IAM user or role that can describe the task definition. We'll want to have an approach that keeps the sensitive data out of the task definition itself; other options include a file that already exists on disk, a Secure parameter in AWS Systems Manager Parameter Store, or in Amazon Secrets Manager.

Thanks,
Sam

adnxn:
> we cannot merge your change until we've made the requisite changes elsewhere in ECS.

Closing this PR for now since we require task definition changes. PR can be reopened in the future and the feature request is being tracked in the issue referenced in the PR description https://github.com/aws/amazon-ecs-agent/issues/1540.

PR #1558, opened 2018-09-04
Migrating Container stats to Docker SDK

yunhee-l:
### Summary
- Migrating Stats to Docker SDK
- Implemented inactivity timeout handling

### Implementation details
- Replaced docker.Stats with types.Stats
- Added unit tests for new stats function
- Added a function to waitForStats in docker_client_tests.go as the API call is made within a goroutine and we need to make sure it is called to avoid "missing call" error.
- updated mocks w/ mocked io.Reader to mock Engine Response.
- copied how go-dockerclient handled inactivity timeout handling into docker_client.go

### Testing

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
@@ -185,7 +186,7 @@ type DockerClient interface {
 
 	// Stats returns a channel of stat data for the specified container. A context should be provided so the request can
 	// be canceled.
-	Stats(string, context.Context) (<-chan *types.Stats, error)
+	Stats(context.Context, string, time.Duration) (<-chan *types.Stats, error)
 
 	// Version returns the version of the Docker daemon.
 	Version(context.Context, time.Duration) (string, error)
@@ -1222,42 +1223,56 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
 }
 
 // Stats returns a channel of *types.Stats entries for the container.
-func (dg *dockerGoClient) Stats(id string, ctx context.Context) (<-chan *types.Stats, error) {
+func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeout time.Duration) (<-chan *types.Stats, error) {
+	subCtx, cancelRequest := context.WithCancel(ctx)
+
 	client, err := dg.sdkDockerClient()
 	if err != nil {
+		cancelRequest()
 		return nil, err
 	}
 
 	// Create channel to hold the stats
-	stats := make(chan *types.Stats)
+	statsChnl := make(chan *types.Stats)
 	var resp types.ContainerStats
 
 	go func() {
+		defer cancelRequest()
+		defer close(statsChnl)
 		// ContainerStats outputs a io.ReadCloser and an OSType
-		resp, err = client.ContainerStats(ctx, id, true)
+		stream := true
+		resp, err = client.ContainerStats(subCtx, id, stream)
 		if err != nil {
-			seelog.Infof("DockerGoClient: Unable to retrieve stats for container %s: %v", id, err)
-			close(stats)
+			seelog.Warnf("DockerGoClient: Unable to retrieve stats for container %s: %v", id, err)
 			return
 		}
+
+		// handle inactivity timeout
+		var canceled uint32
+		var ch chan<- struct{}
+		resp.Body, ch = handleInactivityTimeout(resp.Body, inactivityTimeout, cancelRequest, &canceled)
+		defer resp.Body.Close()
+		defer close(ch)
+
 		// Returns a *Decoder and takes in a readCloser
 		decoder := json.NewDecoder(resp.Body)
 		data := new(types.Stats)
 		for err := decoder.Decode(data); err != io.EOF; err = decoder.Decode(data) {
 			if err != nil {
-				seelog.Infof("DockerGoClient: Unable to decode stats for container %s: %v", id, err)
-				close(stats)
+				seelog.Warnf("DockerGoClient: Unable to decode stats for container %s: %v", id, err)
 				return
 			}
-			stats <- data
+			if atomic.LoadUint32(&canceled) != 0 {
+				seelog.Warnf("DockerGoClient: inactivity time exceeded timeout while retrieving stats for container %s", id)
+				return
+			}
+
+			statsChnl <- data
 			data = new(types.Stats)
 		}
-		// Once EOF, closer io.ReadCloser and stream
-		resp.Body.Close()
-		close(stats)
 	}()
 
-	return stats, nil
+	return statsChnl, nil
 }
 
 // RemoveImage invokes github.com/fsouza/go-dockerclient.Client's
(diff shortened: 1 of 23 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1221,31 +1222,59 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
+	statsChnl := make(chan *types.Stats)
+	var resp types.ContainerStats
 
 	go func() {
-		statsErr := client.Stats(options)
-		if statsErr != nil {
-			seelog.Infof("DockerGoClient: Unable to retrieve stats for container %s: %v",
-				id, statsErr)
+		// ContainerStats outputs a io.ReadCloser and an OSType
+		resp, err = client.ContainerStats(ctx, id, true)
```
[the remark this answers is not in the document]
yunhee-l: It's a boolean flag for streaming. Refactored into a boolean variable named stream to provide more context within the code.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1298,3 +1328,35 @@ func (dg *dockerGoClient) loadImage(ctx context.Context, reader io.Reader) error
...
+
+type proxyReader struct {
+	io.ReadCloser
+	calls uint64
+}
+
+func (p *proxyReader) callCount() uint64 {
+	return atomic.LoadUint64(&p.calls)
+}
+
```
fenxiong: why is this method omitted: https://github.com/fsouza/go-dockerclient/blob/a7b85d61dcf384ef9fea13abaa9fefe5ef00ad25/client.go#L663 ? i think you need it in order for inactivity timeout to work

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1298,3 +1328,35 @@ func (dg *dockerGoClient) loadImage(ctx context.Context, reader io.Reader) error
...
+func handleInactivityTimeout(reader io.ReadCloser, timeout time.Duration, cancelRequest func(), canceled *uint32) (io.ReadCloser, chan<- struct{}) {
```
petderek: I'd like to see this pullled out into a different file with its own unit tests. Maybe add a todo?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1221,31 +1222,60 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
+		stream := true
+		resp, err = client.ContainerStats(ctx, id, stream)
+		if err != nil {
+			seelog.Warnf("DockerGoClient: Unable to retrieve stats for container %s: %v", id, err)
+			close(statsChnl)
+			return
 		}
+
+		// handle inactivity timeout
+		_, cancelRequest := context.WithCancel(ctx)
```
fenxiong: i think we need to pass the ctx returned here to client.ContainerStats(...) so the request can be cancelled when cancelRequest is invoked?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1221,31 +1222,60 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
+		defer cancelRequest()
+		var canceled uint32
+		var ch chan<- struct{}
+		resp.Body, ch = handleInactivityTimeout(resp.Body, StatsInactivityTimeout, cancelRequest, &canceled)
+		defer close(ch)
+
```
fenxiong: resp.Body is not always closed in the following code, probably should add defer res.Body.Close() here?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1221,31 +1222,61 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
+		var canceled uint32
+		var ch chan<- struct{}
+		resp.Body, ch = handleInactivityTimeout(resp.Body, StatsInactivityTimeout, cancelRequest, &canceled)
+		defer resp.Body.Close()
+		defer close(ch)
+
+		// Returns a *Decoder and takes in a readCloser
+		decoder := json.NewDecoder(resp.Body)
+		data := new(types.Stats)
+		for err := decoder.Decode(data); err != io.EOF; err = decoder.Decode(data) {
```
sharanyad: Should we add an unit test covering error `io.EOF`?

yunhee-l: io.EOF is covered in the NormalExit test case.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1221,31 +1222,61 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
+			if err != nil {
+				seelog.Warnf("DockerGoClient: Unable to decode stats for container %s: %v", id, err)
+				close(statsChnl)
```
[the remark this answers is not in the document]
yunhee-l: this caused error with race condition, provably with the order of closing other channels. I'll keep this as is.

agent/dockerclient/dockerapi/docker_client.go
yunhee-l: That's changing the behavior compared to before when go-dockerclient used to handle this timeout. I'll stick to what I think is the as-is translation for now, I don't want the sdk migration to contain any behavior changes.

fenxiong: i think go-dockerclient has this behavior, just under different level. In go-dockerclient, the inactivity timeout is handled in the [stream](https://github.com/aws/amazon-ecs-agent/blob/e8b5229a9ea8b748ebad4db4c098ccc8df2dcd5f/agent/vendor/github.com/fsouza/go-dockerclient/client.go#L521) function, in which it passes a subCtx when it actually calls docker engine via HTTPClient.Do [here](https://github.com/aws/amazon-ecs-agent/blob/e8b5229a9ea8b748ebad4db4c098ccc8df2dcd5f/agent/vendor/github.com/fsouza/go-dockerclient/client.go#L596), and the cancel function corresponds to the subCtx is called when the inactivity timeout happens so that the request will be canceled. In your implementation the inactivity timeout is handled in Stats function, and the request to docker is made via the docker api client.ContainerStats. In that case the client.ContainerStats call should be canceled when inactivity timeout happens, similar to what go-dockerclient does to HTTPClient.Do. 

Let me know if this doesn't make sense..

agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -908,104 +908,161 @@ func TestUnavailableVersionError(t *testing.T) {
...
-
-func checkStatRead(t *testing.T, stat *docker.Stats, read time.Time) {
-	assert.Equal(t, read, stat.Read, "Expected %v, but was %v", read, stat.Read)
+	newStat := <-stats
+	waitForStats(t, newStat)
+	assert.Equal(t, uint64(50), newStat.MemoryStats.Usage)
+	assert.Equal(t, uint64(100), newStat.CPUStats.SystemUsage)
 }
 
-func TestStatsClosed(t *testing.T) {
```
fenxiong: why is this test deleted?

yunhee-l: this as unit test doesn't make sense any more with SDK; with go-dockerclient it returned stream for us to manipulate, whereas with SDK it is returning reader wrapped in a response object.

agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -908,104 +908,161 @@ func TestUnavailableVersionError(t *testing.T) {
...
+			if stat == nil {
+				time.Sleep(time.Second)
+				continue
+			}
+			return
+		}
 	}
+}
+
+func TestStatsInactivityTyimeout(t *testing.T) {
```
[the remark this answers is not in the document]
yunhee-l: fixed

PR #1560, opened 2018-09-05
dockerapi: Migrated Events to Docker SDK

yunhee-l:
### Summary
Migrating Events to Docker SDK
Pulled from https://github.com/aws/amazon-ecs-agent/pull/1498 and fixed remaining PR comments. Benchmark test will be done after go-dockerclient to SDK migration completes.

### Implementation details
- Migrated AddEventEventListener/RemoveEventListener from go-dockerclient to Events from the Docker SDK.
- docker.APIEvents --> events.Message
- The new channels returned from Events contains values and not pointers, so in order to receive them in StartListening(), you must force an evaluation of the variable by adding it as a parameter to the function.
  - This is because the goroutine is accessing the variable 'event' via the closure, and this reference is evaluated when its reached the event variable in CopyEvents. The event variable is being updated by the loop, so there's a race between the loop updates and the goroutines that have started.

### Testing

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -43,6 +43,7 @@ import (
 	"github.com/cihub/seelog"
 	"github.com/docker/docker/api/types"
 	dockercontainer "github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/api/types/events"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/volume"
@@ -838,36 +839,57 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
 
 // Listen to the docker event stream for container changes and pass them up
 func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerContainerChangeEvent, error) {
-	client, err := dg.dockerClient()
+	client, err := dg.sdkDockerClient()
 	if err != nil {
 		return nil, err
 	}
-	dockerEvents := make(chan *docker.APIEvents, dockerEventBufferSize)
-	events := make(chan *docker.APIEvents)
+
+	events := make(chan *events.Message)
 	buffer := NewInfiniteBuffer()
 
-	err = client.AddEventListener(dockerEvents)
-	if err != nil {
-		seelog.Errorf("DockerGoClient: unable to add a docker event listener: %v", err)
-		return nil, err
-	}
+	derivedCtx, cancel := context.WithCancel(ctx)
+	dockerEvents, eventErr := client.Events(derivedCtx, types.EventsOptions{})
+
+	// Cache the event from docker client. Channel closes when an error is passed to eventErr.
+	go buffer.StartListening(derivedCtx, dockerEvents)
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
+	// TODO: move the error check into StartListening() to keep event streaming and error handling in one place.
 	go func() {
-		<-ctx.Done()
-		client.RemoveEventListener(dockerEvents)
+		for {
+			select {
+			case err := <-eventErr:
+				if err == io.EOF || err == io.ErrUnexpectedEOF {
+					seelog.Info("DockerGoClient: All events have been received")
+					cancel()
+					return
+				} else {
+					// If an error is returned, we need to reopen channel to continue listening
+					seelog.Errorf("DockerGoClient: error while listening to Docker Events : %v", err)
+					nextCtx, nextCancel := context.WithCancel(ctx)
+					dockerEvents, eventErr = client.Events(nextCtx, types.EventsOptions{})
+					// Cache the event from docker client.
+					go buffer.StartListening(nextCtx, dockerEvents)
+					// Close previous stream after starting to listen on new one
+					cancel()
+					// Reassign cancel variable next Cancel function to setup next iteration of loop.
+					cancel = nextCancel
+				}
+			case <-ctx.Done():
+				return
+			}
+		}
 	}()
 
-	// Cache the event from go docker client
-	go buffer.StartListening(dockerEvents)
 	// Read the buffered events and send to task engine
 	go buffer.Consume(events)
-
 	changedContainers := make(chan DockerContainerChangeEvent)
 	go dg.handleContainerEvents(ctx, events, changedContainers)
+
 	return changedContainers, nil
 }
 
 func (dg *dockerGoClient) handleContainerEvents(ctx context.Context,
-	events <-chan *docker.APIEvents,
+	events <-chan *events.Message,
 	changedContainers chan<- DockerContainerChangeEvent) {
 	for event := range events {
 		containerID := event.ID
--- a/agent/dockerclient/dockerapi/docker_events_buffer.go
+++ b/agent/dockerclient/dockerapi/docker_events_buffer.go
@@ -14,9 +14,10 @@
 package dockerapi
 
 import (
+	"context"
 	"sync"
 
-	docker "github.com/fsouza/go-dockerclient"
+	"github.com/docker/docker/api/types/events"
 )
 
 const (
@@ -38,7 +39,7 @@ var containerEvents = []string{
 // InfiniteBuffer defines an unlimited buffer, where it reads from
 // input channel and write to output channel.
 type InfiniteBuffer struct {
-	events       []*docker.APIEvents
+	events       []*events.Message
 	empty        bool
 	waitForEvent sync.WaitGroup
 	count        int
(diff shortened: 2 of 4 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -837,36 +838,55 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
...
+	derivedCtx, cancel := context.WithCancel(ctx)
+	dockerEvents, eventErr := client.Events(derivedCtx, types.EventsOptions{})
+
+	// Cache the event from docker client. Channel closes when an error is passed to eventErr.
+	go buffer.StartListening(derivedCtx, dockerEvents)
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
 	go func() {
-		<-ctx.Done()
-		client.RemoveEventListener(dockerEvents)
+		for {
```
petderek: I'm not the biggest fan of this. We have a loop checking for errors from another async function.

Can we make NewInfiniteBuffer.StartListening return an error so that we can handle it it one place?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -837,36 +838,55 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
...
+			select {
+			case err := <-eventErr:
+				if err == io.EOF {
+					// If error processed is EOF, ignore and continue listening
```
fenxiong: not sure why we ended up like this in previous pr, but from the implementation of client.Events(https://github.com/docker/go-docker/blob/4daae26030ad00e348edddff9767924ae57a3b82/events.go#L20), it seems that after an error comes out from the error channel (not matter it's EOF or not), there will be no more events coming out from event channel, and we need to call client.Events again to continue receiving event

agent/dockerclient/dockerapi/docker_client.go
yunhee-l: infiniteBuffer.StartListening is meant to filter and buffer docker events that agent is interested in. I don't want to introduce behavior change through this migration.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -838,36 +839,54 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
...
+	go buffer.StartListening(derivedCtx, dockerEvents)
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
 	go func() {
-		<-ctx.Done()
-		client.RemoveEventListener(dockerEvents)
+		for {
+			select {
+			case err := <-eventErr:
+				// If an error is returned, we need to reopen channel to continue listening
+				if err == io.EOF {
```
haikuoliu: For [fsouza client implementation](https://github.com/fsouza/go-dockerclient/blob/11e1e492616b3d3b7046fceb0f1e7e02f9eef0d8/event.go#L229), when client gets an error, it just returns, should we keep the same behavior?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -838,36 +839,54 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
...
+					seelog.Info("DockerGoClient: All events have been received.")
+				} else {
+					seelog.Errorf("DockerGoClient: error while listening to Docker Events : %v", err)
+				}
+				nextCtx, nextCancel := context.WithCancel(ctx)
+				dockerEvents, eventErr = client.Events(nextCtx, types.EventsOptions{})
+				// Cache the event from docker client.
+				go buffer.StartListening(nextCtx, dockerEvents)
```
haikuoliu: Is it possible to add a test for restarting listening channel case?

agent/dockerclient/dockerapi/docker_events_buffer.go
```diff
@@ -51,15 +52,21 @@ func NewInfiniteBuffer() *InfiniteBuffer {
...
-// TODO: wire in ctx to stop listening
-func (buffer *InfiniteBuffer) StartListening(events chan *docker.APIEvents) {
-	for event := range events {
-		go buffer.CopyEvents(event)
+// When context is cancelled, stop listening
+func (buffer *InfiniteBuffer) StartListening(ctx context.Context, eventChan <-chan events.Message) {
+	for {
+		select {
+		// If context is cancelled, return
+		case <-ctx.Done():
```
haikuoliu: What if the context is canceled but there is still event in the `eventChan` channel? We might miss some events in this case.

agent/dockerclient/dockerapi/docker_client.go
yunhee-l: See https://github.com/docker/go-docker/blob/4daae26030ad00e348edddff9767924ae57a3b82/events.go#L16L19
I think in fsouza implementation it's returning from monitoring the events, but it is not closing the event stream itself.

agent/dockerclient/dockerapi/docker_events_buffer.go
yunhee-l: I think Katy was implementing the "TODO: wire in ctx to stop listening". We stopped listening for events when context was cancelled: https://github.com/aws/amazon-ecs-agent/blob/dev/agent/dockerclient/dockerapi/docker_client.go#L787 
Once context is cancelled, I don't think we are interested in the events for it any more?

agent/dockerclient/dockerapi/docker_client.go
yunhee-l: looking into this

agent/dockerclient/dockerapi/docker_client.go
haikuoliu: For fsouza implementation, inside `monitorEvents`, I think `eventState.C` is the event event stream. And there are three cases to handle the events:
1. Receive an event, it will call `sendEvent` to handle it. 
2. Receive an EOF, it will call `eventState.disableEventMonitoring()` and then `return`, so that it doesn't receive events anymore - this is equivalent to close the event stream.
3. Receive error other than EOF: it will restart the `monitorEvents`.

See [docker events doc](https://docs.docker.com/engine/reference/commandline/events/), there are options like `since` and `until`, when you run the events command with these options, the call to docker dameon is not a blocking call, when you receive the last event, you will end up with an EOF error. But for the default options (Agent case), the `events` API call is a blocking call, which I guess will never end up with an EOF error.

yunhee-l: Yep you are totally right. Event with no chance of EOF error, I'll duplicate what we were doing with go-dockerclient.

agent/dockerclient/dockerapi/docker_events_buffer.go
yunhee-l: Added logic to drain the channel into buffer when the context is closed. Was worried about potentially introducing inherently racy condition but the eventChan is not shared across multiple contexts so I think it's fine.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -838,36 +839,56 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
...
+					nextCtx, nextCancel := context.WithCancel(ctx)
+					dockerEvents, eventErr = client.Events(nextCtx, types.EventsOptions{})
+					// Cache the event from docker client.
+					go buffer.StartListening(nextCtx, dockerEvents)
+					// Close previous stream after starting to listen on new one
+					cancel()
+					// Reassign cancel variable next Cancel function to setup next iteration of loop.
+					cancel = nextCancel
+				}
+			case <-ctx.Done():
```
sharanyad: Should this be `case <-derivedCtx.Done()`, since that's the one passed to `client.Events`?

haikuoliu: If this is `case <-derivedCtx.Done()`, this go routine will not exist anymore after we receive the first error and cancel the `derivedCtx` - and I think that's not what we want

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -838,36 +839,56 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
...
+	go buffer.StartListening(derivedCtx, dockerEvents)
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
 	go func() {
-		<-ctx.Done()
-		client.RemoveEventListener(dockerEvents)
+		for {
+			select {
+			case err := <-eventErr:
+				if err == io.EOF {
+					seelog.Info("DockerGoClient: All events have been received.")
```
haikuoliu: minor: I think Agent log usually doesn't end up with a dot 🤔

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -838,36 +839,57 @@ func getMetadataHealthCheck(dockerContainer *types.ContainerJSON) apicontainer.H
...
+	go buffer.StartListening(derivedCtx, dockerEvents)
+	// Receive errors from channels. If error thrown is not EOF, log and reopen channel.
+	// TODO: move the error check into StartListening() to keep event streaming and error handling in one place.
 	go func() {
-		<-ctx.Done()
-		client.RemoveEventListener(dockerEvents)
+		for {
+			select {
+			case err := <-eventErr:
+				if err == io.EOF {
```
fenxiong: io.ErrUnexpectedEOF is also treated as EOF event by fsouza's docker client: https://github.com/fsouza/go-dockerclient/blob/11e1e492616b3d3b7046fceb0f1e7e02f9eef0d8/event.go#L349

yunhee-l: will add ErrUnexpectedEOF here too.

PR #1571, opened 2018-09-12
fix windows drive volume mounting

fenxiong:
### Summary

For windows volume task, the agent fails to support a mounting from drive to drive, because when the container path is a bare drive like 'D:', the agent changes it to 'd:.' in the getCanonicalPath function, which is no longer supported by Docker because that is a directory, but container path can only be either a directory in C drive, or another drive. This pr is to fix this issue.

### Implementation details

Treat the bare drive path specially in getCanonicalPath.

### Testing

New tests cover the changes: 
unit test added; manual test is performed to verify that this fixes the issue; functional test is not added because it seems to be nontrivial to add another drive.

### Licensing

### DIFF
--- a/agent/api/task/task_windows.go
+++ b/agent/api/task/task_windows.go
@@ -74,7 +74,22 @@ func (task *Task) downcaseAllVolumePaths() {
 }
 
 func getCanonicalPath(path string) string {
-	return filepath.Clean(strings.ToLower(path))
+	lowercasedPath := strings.ToLower(path)
+	// if the path is a bare drive like "d:", don't filepath.Clean it because it will add a '.'.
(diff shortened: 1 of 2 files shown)


agent/api/task/task_windows.go
```diff
@@ -73,8 +73,23 @@ func (task *Task) downcaseAllVolumePaths() {
 	}
 }
 
+func isBareDrive(path string) bool {
```
samuelkarp: nit: I tend to prefer having the caller function above the callee function (so `getCanonicalPath` above `isBareDrive`) as I find it easier to read (I can understand why the function exists before reading the body of it).

fenxiong:
moved isBareDrive to be under getCanonicalPath and rebased.

PR #1584, opened 2018-09-27
Updating Task Model and implementing docker flags --pid and --ipc

linkar-ec2:
### Summary

Updates the Agent Task model to include PidMode and IpcMode for sharing PID namespace and IPC resource namespace, respectively. 
Adding unit tests for Pid and Ipc resource namespace sharing

### Implementation details

Agent Task struct is updated to support the PidMode and IpcMode that is in the Task struct received in the payload message from ACS. 
In PostUnmarshalTask(), the task is used to determine if a Pause container for namespace sharing needs to be provisioned. If so, it is added as an internal container to the task. 
In CreateContainer(), the DockerHostConfig() method determines if the current container's PidMode and IpcOption needs to be changed. 

### Testing

New tests cover the changes: 
Unit Tests cover the added Pause container provisioning methods and DockerHostConfig updates

Adding ECS Agent support for the --pid and --ipc Docker flags
### Licensing

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 # Changelog
+## 1.20.4-dev
+* Bug - Fixed a bug where Windows drive volume couldn't be mounted [#1571](https://github.com/aws/amazon-ecs-agent/pull/1571)
+* Bug - Fixed a bug where the Agent's Windows binaries didn't use consistent naming [#1573](https://github.com/aws/amazon-ecs-agent/pull/1573)
+* Bug - Fixed a bug where a port used by WinRM service was not reserved by the Agent by default [#1577](https://github.com/aws/amazon-ecs-agent/pull/1577)
+
 ## 1.20.3
 * Enhancement - Deprecate support for serial docker image pull [#1569](https://github.com/aws/amazon-ecs-agent/pull/1569)
 * Enhancement - Update the `amazon-ecs-cni-plugins` to `2018.08.0`
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -401,8 +401,8 @@ func (c *Container) GetNextKnownStateProgression() apicontainerstatus.ContainerS
 	return c.GetKnownStatus() + 1
 }
 
-// IsInternal returns true if the container type is either `ContainerEmptyHostVolume`
-// or `ContainerCNIPause`. It returns false otherwise
+// IsInternal returns true if the container type is `ContainerCNIPause`
+// or `ContainerNamespacePause`. It returns false otherwise
 func (c *Container) IsInternal() bool {
 	if c.Type == ContainerNormal {
 		return false
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -51,8 +51,12 @@ import (
 )
 
 const (
-	// PauseContainerName is the internal name for the pause container
-	PauseContainerName = "~internal~ecs~pause"
+	// NetworkPauseContainerName is the internal name for the pause container
+	NetworkPauseContainerName = "~internal~ecs~pause~network"
+
+	// NamespacePauseContainerName is the internal name for the IPC resource namespace and/or
+	// PID namespace sharing pause container
+	NamespacePauseContainerName = "~internal~ecs~pause~namespace"
 
 	emptyHostVolumeName = "~internal~ecs-emptyvolume-source"
 
@@ -65,13 +69,21 @@ const (
 	arnResourceDelimiter = "/"
 	// networkModeNone specifies the string used to define the `none` docker networking mode
 	networkModeNone = "none"
-	// networkModeContainerPrefix specifies the prefix string used for setting the
-	// container's network mode to be mapped to that of another existing container
-	networkModeContainerPrefix = "container:"
+	// dockerMappingContainerPrefix specifies the prefix string used for setting the
+	// container's option (network, ipc, or pid) to that of another existing container
+	dockerMappingContainerPrefix = "container:"
 
 	// awslogsCredsEndpointOpt is the awslogs option that is used to pass in an
 	// http endpoint for authentication
 	awslogsCredsEndpointOpt = "awslogs-credentials-endpoint"
+
+	// These contants identify the docker flag options
+	pidModeHost     = "host"
+	pidModeTask     = "task"
+	ipcModeHost     = "host"
+	ipcModeTask     = "task"
+	ipcModeSharable = "shareable"
+	ipcModeNone     = "none"
 )
 
 // TaskOverrides are the overrides applied to a task
@@ -167,6 +179,14 @@ type Task struct {
 	terminalReason     string
 	terminalReasonOnce sync.Once
 
+	// PIDMode is used to determine how PID namespaces are organized between
+	// containers of the Task
+	PIDMode string `json:"PidMode,omitempty"`
+
+	// IPCMode is used to determine how IPC resources should be shared among
+	// containers of the Task
+	IPCMode string `json:"IpcMode,omitempty"`
+
 	// lock is for protecting all fields in the task struct
 	lock sync.RWMutex
 }
@@ -233,6 +253,9 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 
 	task.initializeCredentialsEndpoint(credentialsManager)
 	task.addNetworkResourceProvisioningDependency(cfg)
+	// Adds necessary Pause containers for sharing PID or IPC namespaces
+	task.addNamespaceSharingProvisioningDependency(cfg)
+
 	return nil
 }
 
@@ -526,7 +549,7 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
 	}
 	pauseContainer := apicontainer.NewContainerWithSteadyState(apicontainerstatus.ContainerResourcesProvisioned)
 	pauseContainer.TransitionDependenciesMap = make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet)
-	pauseContainer.Name = PauseContainerName
+	pauseContainer.Name = NetworkPauseContainerName
 	pauseContainer.Image = fmt.Sprintf("%s:%s", cfg.PauseContainerImageName, cfg.PauseContainerTag)
 	pauseContainer.Essential = true
 	pauseContainer.Type = apicontainer.ContainerCNIPause
@@ -536,8 +559,30 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
 		if container.IsInternal() {
 			continue
 		}
-		container.BuildContainerDependency(PauseContainerName, apicontainerstatus.ContainerResourcesProvisioned, apicontainerstatus.ContainerPulled)
+		container.BuildContainerDependency(NetworkPauseContainerName, apicontainerstatus.ContainerResourcesProvisioned, apicontainerstatus.ContainerPulled)
 		pauseContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
+	}
+}
+
+func (task *Task) addNamespaceSharingProvisioningDependency(cfg *config.Config) {
+	// Pause container does not need to be created if no namespace sharing will be done at task level
+	if task.getIPCMode() != ipcModeTask && task.getPIDMode() != pidModeTask {
+		return
+	}
+	namespacePauseContainer := apicontainer.NewContainerWithSteadyState(apicontainerstatus.ContainerRunning)
+	namespacePauseContainer.TransitionDependenciesMap = make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet)
+	namespacePauseContainer.Name = NamespacePauseContainerName
+	namespacePauseContainer.Image = fmt.Sprintf("%s:%s", config.DefaultPauseContainerImageName, config.DefaultPauseContainerTag)
+	namespacePauseContainer.Essential = true
+	namespacePauseContainer.Type = apicontainer.ContainerNamespacePause
+	task.Containers = append(task.Containers, namespacePauseContainer)
+
+	for _, container := range task.Containers {
+		if container.IsInternal() {
+			continue
+		}
+		container.BuildContainerDependency(NamespacePauseContainerName, apicontainerstatus.ContainerRunning, apicontainerstatus.ContainerPulled)
+		namespacePauseContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
 	}
 }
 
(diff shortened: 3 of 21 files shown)


samuelkarp:
Is there any particular reason that you're not sharing the pause container between awsvpc mode and these new options?  They serve the same purpose: to provide root namespaces that can be shared.

linkar-ec2:
The purpose of the ENI Pause Container was to attach the individual containers within a task to an ENI. We did this by creating a common network namespace and then attaching the Pause container to the ENI.

Reusing this Pause container would tightly couple this ENI implementation with the namespace sharing for PID and IPC resources. With the resources consumed by the Pause container considered to be negligible, creating another container would allow single responsibility of the internal containers we create.  A concrete example of this would be the case when an ENI has custom DNS IPs and search list associated with it, in which case we will configure the HostConfig for the ENI Pause, but not for the PID/IPC namespace Pause container.

As regards to the possible question “why not create 2 Pause containers for the PID and IPC resources for this same reason of single responsibility?” this Namespace Pause container is simply provided as a parameter to the other containers within a task, for either PidMode or IpcMode, or both. There is no other HostConfig changes done to the Pause container itself (a difference from the ENI Pause).

Hope this sheds some light. Please don’t hesitate to ask questions or comments.

Best,

Kar

samuelkarp:
That's not exactly accurate.  The purpose of the existing pause container is to create a network namespace that can be shared among the containers in a task.  We don't attach the pause container to the ENI, but rather move the ENI into the network namespace that was created as a side-effect of running the pause container.  PID and IPC namespaces are similar in that they're just namespaces of the container.

agent/api/task/task.go
```diff
@@ -536,11 +551,37 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
...
+	if task.getIpcMode() == "" && task.getPidMode() == "" {
+		return
+	}
+
+	// Create a Pause container if either IPC resource or PID namespace needs to be shared
+	if task.getIpcMode() == "task" || task.getPidMode() == "task" {
+		namespacePauseContainer := apicontainer.NewContainerWithSteadyState(apicontainerstatus.ContainerResourcesProvisioned)
+		namespacePauseContainer.TransitionDependenciesMap = make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet)
+		namespacePauseContainer.Name = NamespacePauseContainerName
+		namespacePauseContainer.Image = fmt.Sprintf("%s:%s", cfg.PauseContainerImageName, cfg.PauseContainerTag)
```
haikuoliu: To avoid `PauseContainerImageName` being overridden by Fargate, one option is you can use `DefaultPauseContainerImageName` here. And you will also need to make change [here](https://github.com/aws/amazon-ecs-agent/blob/e8b5229a9ea8b748ebad4db4c098ccc8df2dcd5f/agent/app/agent_unix.go#L84) and `pause_linux.go` to make Agent always load *default* pause container tarball

linkar-ec2:
[After syncing with Sam and Derek] 
For closure, it is better to use separate Pause containers for the ENI/Network sharing and the PID/IPC resources namespace sharing. The image for the Pause container for the ENI attachment is specific to the requirements of the ENI. The image for the PID/IPC sharing Pause will be the default image.

agent/api/task/task.go
```diff
@@ -831,21 +872,30 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
...
+			if hosts := task.generateENIExtraHosts(); hosts != nil {
+				hostConfig.ExtraHosts = append(hostConfig.ExtraHosts, hosts...)
+			}
 
-		// apply ExtraHosts to HostConfig for pause container
-		if hosts := task.generateENIExtraHosts(); hosts != nil {
-			hostConfig.ExtraHosts = append(hostConfig.ExtraHosts, hosts...)
+			// Override the DNS settings for the pause container if ENI has custom
+			// DNS settings
+			return task.overrideDNS(hostConfig), nil
```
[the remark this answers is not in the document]
linkar-ec2: This return specifically applies to the CNIPauseContainer. Because we are using different Pause containers, we do not want to apply the Pid or Ipc settings to this specific Docker container. Therefore, our expected call flow would be to turn here.

agent/api/container/container.go
```diff
@@ -401,8 +401,8 @@ func (c *Container) GetNextKnownStateProgression() apicontainerstatus.ContainerS
 	return c.GetKnownStatus() + 1
 }
 
-// IsInternal returns true if the container type is either `ContainerEmptyHostVolume`
-// or `ContainerCNIPause`. It returns false otherwise
+// IsInternal returns true if the container type is `ContainerEmptyHostVolume`,
+// `ContainerCNIPause`, or `ContainerNamespacePause`. It returns false otherwise
```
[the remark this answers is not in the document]
linkar-ec2: Because the ContainerNamespacePause is used for both IPC resource namespace sharing and PID namespace sharing, it was named with 'Namespace' rather than 'Process'

agent/api/task/task.go
```diff
@@ -953,6 +1004,96 @@ func (task *Task) generateENIExtraHosts() []string {
...
+		return false, ""
+	}
+
+	// no need to override the PID mode. This means the container's PID namespace
+	// will be private
+	if task.getPIDMode() == "" || strings.EqualFold(task.getPIDMode(), "none") {
+		return false, ""
+	}
+
+	if strings.EqualFold(task.getPIDMode(), "host") {
```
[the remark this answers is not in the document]
linkar-ec2: @yunhee-l Functional tests need the Frontend changes to be complete before we can register task definitions. Integ tests are currently being made and will probably be in another PR

agent/api/task/task.go
```diff
@@ -1016,28 +1017,21 @@ func (task *Task) shouldOverridePidMode(container *apicontainer.Container, docke
...
-		for _, cont := range task.Containers {
-			if cont.Type == apicontainer.ContainerNamespacePause {
-				pidPauseContName = cont.Name
-				break
-			}
-		}
-		if pidPauseContName == "" {
-			seelog.Critical("Namespace Pause Container required, but not found in the task: %s", task.String())
-			return false, "" // cannot override the Pid without the Pause container
+	} else if strings.EqualFold(task.getPIDMode(), "task") {
```
haikuoliu: Any particular reason using `strings.EqualFold` instead of `==`? I think `==` should be sufficient here as we already know the result of `task.getPIDMode()`, i.e. it's `"task"` instead of `"Task"` or `"tAsk"`, this is more like a contract between Agent and ACS.

agent/app/agent_unix.go
```diff
@@ -81,17 +81,15 @@ func (agent *ecsAgent) initializeTaskENIDependencies(state dockerstate.TaskEngin
 		return err, true
 	}
 
-	if agent.cfg.ShouldLoadPauseContainerTarball() {
```
haikuoliu: Since we will load the default pause container anyway, `ShouldLoadPauseContainerTarball` loses its meaning, the only consumer of this function is `platformString`, can you please name it to something that matches its meaning? Or you can just delete this function and move the logic into `platformString`

agent/api/container/container.go
```diff
@@ -401,8 +401,8 @@ func (c *Container) GetNextKnownStateProgression() apicontainerstatus.ContainerS
 	return c.GetKnownStatus() + 1
 }
 
-// IsInternal returns true if the container type is either `ContainerEmptyHostVolume`
-// or `ContainerCNIPause`. It returns false otherwise
+// IsInternal returns true if the container type is `ContainerEmptyHostVolume`,
```
sharanyad: We don't use the `ContainerEmptyHostVolume` anymore, it can be removed from the comment.

agent/api/task/task.go
```diff
@@ -536,11 +561,33 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) {
...
 		pauseContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
 	}
 }
 
+func (task *Task) addNamespaceSharingProvisioningDependency(cfg *config.Config) {
+	// Pause container does not need to be created if no namespace sharing will be done at task level
+	if task.getIPCMode() != ipcModeTask && task.getPIDMode() != pidModeTask {
+		return
+	}
+	namespacePauseContainer := apicontainer.NewContainerWithSteadyState(apicontainerstatus.ContainerResourcesProvisioned)
```
sharanyad: What is the definition of `ContainerResourcesProvisioned` state for namespacePauseContainer?

agent/api/task/task.go
```diff
@@ -953,6 +1014,91 @@ func (task *Task) generateENIExtraHosts() []string {
...
+// false otherwise
+func (task *Task) shouldOverridePIDMode(container *apicontainer.Container, dockerContainerMap map[string]*apicontainer.DockerContainer) (bool, string) {
+	// If the container is an internal container (ContainerEmptyHostVolume,
+	// ContainerCNIPause, or ContainerNamespacePause), then PID namespace for
+	// the container itself should be private (default Docker option)
+	if container.IsInternal() {
+		return false, ""
+	}
+
+	if task.getPIDMode() == pidModeHost {
```
sharanyad: These can be switch statements instead of if, else. That would be easier to read.

agent/api/task/task.go
```diff
@@ -953,6 +1014,91 @@ func (task *Task) generateENIExtraHosts() []string {
...
+		return true, pidModeHost
+	} else if task.getPIDMode() == pidModeTask {
+		pauseContName, ok := task.findPauseContainerName(apicontainer.ContainerNamespacePause)
+		if !ok {
+			seelog.Critical("Namespace Pause container not found in the task: %s", task.String())
+			return false, ""
+		}
+		pauseToGiveDocker, ok := dockerContainerMap[pauseContName]
```
sharanyad: what does `pauseToGiveDocker` mean?

agent/api/task/task.go
```diff
@@ -953,6 +1014,91 @@ func (task *Task) generateENIExtraHosts() []string {
...
+		if !ok || pauseToGiveDocker == nil {
+			// Docker container shouldn't be nill or not exist if the Container definition within task exists; implies code-bug
```
sharanyad: nit: `nil`

agent/api/task/task.go
```diff
@@ -953,6 +1014,91 @@ func (task *Task) generateENIExtraHosts() []string {
...
+			seelog.Criticalf("Namespace Pause container required, but not found in container map for container: [%s] in task: %s",
```
sharanyad: This is a failure case, which should be handled in a way that the task is not run. If pidMode is task and pause container is not found, we should stop the task imo. same for ipcMode.

agent/api/task/task.go
linkar-ec2: This is the Docker container name that is contained in the map. We will 'give' this container to the Docker api for the --pid or --ipc flag

agent/api/task/task.go
linkar-ec2: The first if block tests if the container is internal. This is unrelated to the Task's pidMode, whose logic is an if and else-if block. For separation of logic, I think this if-else works better than the switch. What do you think?

agent/api/task/task.go
sharanyad: Got it. it's a little confusing. I would prefer just calling it pauseContainerID or something similar.

linkar-ec2: Changed!

agent/api/task/task.go
linkar-ec2: ContainerResourcesProvisioned seems to be the norm for internal containers. Even though this state checks to see if ENI resources need to be added (for our namespacePauseContainer, nothing additional will be done). Because we don't need anything in this method anyways, I set the Dependency to use Running. I don't think there should be difference in runtime though.

agent/api/task/task.go
```diff
@@ -583,7 +583,7 @@ func (task *Task) addNamespaceSharingProvisioningDependency(cfg *config.Config)
 		if container.IsInternal() {
 			continue
 		}
-		container.BuildContainerDependency(NamespacePauseContainerName, apicontainerstatus.ContainerResourcesProvisioned, apicontainerstatus.ContainerPulled)
+		container.BuildContainerDependency(NamespacePauseContainerName, apicontainerstatus.ContainerRunning, apicontainerstatus.ContainerPulled)
```
haikuoliu: You will also need to change 
```
namespacePauseContainer := apicontainer.NewContainerWithSteadyState(apicontainerstatus.ContainerResourcesProvisioned)
``` 
to 
```
namespacePauseContainer := apicontainer.NewContainerWithSteadyState(apicontainerstatus.ContainerRunning)
```
Otherwise the Agent will call `provisionContainerResources` to setup CNI for namespace pause container.

agent/api/task/task.go
```diff
@@ -76,6 +76,16 @@ const (
 	// awslogsCredsEndpointOpt is the awslogs option that is used to pass in an
 	// http endpoint for authentication
 	awslogsCredsEndpointOpt = "awslogs-credentials-endpoint"
+
+	// ipcModeNone specifies the string used to define the `none` docker IPC resource namespace sharing mode
```
haikuoliu: Any reason that `ipcModeNone` is not grouped with other pid/ipc constants?

agent/api/task/task.go
linkar-ec2: Good catch. Updated.

agent/api/task/task.go
linkar-ec2: Just to distinguish that None is an option for IPC but thinking again, it makes more sense to just group it and leave out the comment.

agent/api/task/task_test.go
```diff
@@ -635,6 +635,183 @@ func TestPostUnmarshalTaskWithLocalVolumes(t *testing.T) {
 
 }
 
+// Slice of structs for Table Driven testing for sharing PID and IPC resources
+var namespaceTests = []struct {
```
sharanyad: Please add tests for IPCMode None and Shareable as well.

agent/api/task/task_test.go
```diff
@@ -635,6 +635,183 @@ func TestPostUnmarshalTaskWithLocalVolumes(t *testing.T) {
...
+					case "host":
+						assert.Equal(t, "host", dockHostCfg.PidMode)
+					}
+				}
+			}
+		}
+	}
+}
+
+func TestAddNamespaceSharingProvisioningDependency(t *testing.T) {
```
sharanyad: I think this test is missing assertion of the fields of `TransitionDependenciesMap`, which is populated by the method `addNamespaceSharingProvisioningDependency`

agent/api/task/task_test.go
linkar-ec2: The customer will not be allowed to provide Shareable as an option. Shareable is internally set for the Pause container and is tested within a unit test.

agent/api/task/task_test.go
linkar-ec2: Added!

linkar-ec2:
Code Coverage: 
Task: 70.1%
Container: 57.3%

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
 # Changelog
+## 1.20.4-dev
```
sharanyad: Looks like this file should not be there in this PR.

linkar-ec2: I rebased and squashed all commits. Awaiting merge.

agent/api/task/task.go
```diff
@@ -953,6 +1006,95 @@ func (task *Task) generateENIExtraHosts() []string {
...
+		pauseCont, ok := task.ContainerByName(NamespacePauseContainerName)
+		if !ok {
+			seelog.Criticalf("Namespace Pause container not found in the task: %s; Setting Task's Desired Status to Stopped", task.Arn)
+			task.SetDesiredStatus(apitaskstatus.TaskStopped)
+			return false, ""
+		}
+		pauseDockerID, ok := dockerContainerMap[pauseCont.Name]
+		if !ok || pauseDockerID == nil {
+			// Docker container shouldn't be nil or not exist if the Container definition within task exists; implies code-bug
+			seelog.Criticalf("Namespace Pause container not found in the task: %s; Setting Task's Desired Status to Stopped", task.Arn)
```
sharanyad: nit: this error message is same as the previous condition. 
Please change it to something like not found in container map, so its easier while debugging.

PR #1385, opened 2018-05-14
[wip] config: parse config from user-data

aaithal:
Add support for parsing ECS agent configuration JSON from user-data. The
configuration blob parsed from userdata string will be merged after the
env config and JSON file configs.

### Summary

Parse config from user-data

### Implementation details

Add support for parsing ECS agent configuration JSON from user-data. The configuration blob parsed from userdata string will be merged after the env config and JSON file configs.

### Testing

- [ ] Builds on Linux (`make release`)
- [ ] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [ ] Unit tests on Linux (`make test`) pass
- [ ] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [ ] Integration tests on Linux (`make run-integ-tests`) pass
- [ ] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [ ] Functional tests on Linux (`make run-functional-tests`) pass
- [ ] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: Yes

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1599, opened 2018-10-05
config: parse config from user-data, in favor of #1385

adnxn:
### Summary
in favor of https://github.com/aws/amazon-ecs-agent/pull/1385

### Testing

New tests cover the changes: 

### Licensing


agent/config/config.go
```diff
@@ -359,6 +361,40 @@ func fileConfig() (Config, error) {
...
+		return parsedUserData.Config
+	}
+	// In the future, if we want to support base64 encoded config,
+	// we'd need to add logic to decode the string here.
+	err = json.Unmarshal([]byte(userData), &parsedUserData)
+	if err != nil {
+		seelog.Warnf("Unable to parse user data: %v", err)
+		// Unable to parse userdata as a valid JSON. Return the
+		// empty config
+		return parsedUserData.Config
```
sharanyad: Are we sure this would be empty `Config` as opposed to some junk value? 

Imo, it might be better to return empty Config object.

agent/config/config_test.go
```diff
@@ -532,6 +540,22 @@ func TestTaskMetadataRPSLimits(t *testing.T) {
 	}
 }
 
+func TestUserDataConfig(t *testing.T) {
```
sharanyad: We are missing negative test cases here.

agent/config/config.go
adnxn: agreed, making this change.

agent/config/config_test.go
adnxn: added failure cases.

agent/config/config_test.go
```diff
@@ -532,6 +540,58 @@ func TestTaskMetadataRPSLimits(t *testing.T) {
...
+	testcases := []struct {
+		name                      string
+		userDataResponse          string
+		userDataResponseError     error
+		expectedConfigCluster     string
+		expectedConfigAPIEndpoint string
+		shouldFail                bool
```
sharanyad: `shouldFail` can be removed.

adnxn: good catch

PR #1602, opened 2018-10-05
add secret model, agent capability for secret and secret task resource

yumex93:
### Summary
This pr includes code changes for updating acs model, adding secret model on container level, adding agent capability to advertise agent supporting of vending ssm secret as environment variables, adding secret from ssm as task resource and related initialization functions when starting task.

### Implementation details
1. The acs model is updated with secrets.
2. A secret model is added to container model.
3. A new agent capability for both linux/windows platforms to advertise agent support of vending secret from ssm(parameter store) as environment variables
4. A task resource is added to only secrets from ssm. Secrets from different providers are tackled separately, so we can take advantage of existing logic of initializing different task resources in parallel to speed up the process.

### Testing

There is no item for this pr, will be add together when ecs secret being released.
### Licensing


agent/acs/model/api/api-2.json
```diff
@@ -235,6 +236,20 @@
       "type":"list",
       "member":{"shape":"ElasticNetworkInterface"}
     },
+    "EncodedString":{
```
petderek: I don't see this used anywhere?

agent/ssm/generate_mocks.go
```diff
@@ -0,0 +1,16 @@
...
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package ssm
+
+//go:generate go run ../../scripts/generate/mockgen.go github.com/aws/aws-sdk-go/service/ssm/ssmiface SSMAPI mocks/ssmiface_mocks.go
```
petderek: We should scope this down to only the methods we use.

agent/ssm/factory/generate_mocks.go
```diff
@@ -0,0 +1,16 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package factory
```
petderek: I think we can avoid using this pattern for this use case.

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+	case err := <-errorEvents:
+		secret.setTerminalReason(err.Error())
+		return err
+	default:
+		break
+	}
+	return nil
+}
+
+func (secret *SSMSecretResource) getGoRoutineTotalNum() int {
```
adnxn: so this is for the total number of regions? should this be the same as the number of keys in your region:secrets map?

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+		secretKey := secretName + "_" + region
+		secret.secretData[secretKey] = secretValue
+	}
+
+	return
+}
+
+func (secret *SSMSecretResource) extractNameFromValueFrom(secretData apicontainer.Secret) string {
+	valueFrom := secretData.ValueFrom
+	match, _ := regexp.MatchString(ParameterARNRegex, valueFrom)
```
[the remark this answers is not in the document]
adnxn: definitely use the arn package please

agent/acs/model/api/api-2.json
adnxn: yea, it's not part of this feature. we should get rid of it.

agent/taskresource/ssmsecret/ssmsecret_test.go
```diff
@@ -0,0 +1,377 @@
+// +build unit
```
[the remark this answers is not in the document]
yumex93: Why you want me to add test cases for situation (1)? I have the test case that secrets in multiple regions are retrieved in multiple goroutines. In terms of reading, it uses more than one API call to retrieve values which is the same as (1). And in terms of writing, since we are locking on resource itself, it is still the same case as (1).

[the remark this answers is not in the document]
fenxiong: this blob is left untested if you don't have situation (1): https://github.com/aws/amazon-ecs-agent/blob/8096c6791257f3f096c870e90ae7f38005219bb9/agent/taskresource/ssmsecret/ssmsecret.go#L306-L312

agent/app/agent_capability.go
```diff
@@ -37,6 +37,7 @@ const (
 	capabilityDockerPluginInfix                 = "docker-plugin."
 	attributeSeparator                          = "."
 	capabilityPrivateRegistryAuthASM            = "private-registry-authentication.secretsmanager"
+	capabilitySecretEnvSSM                      = "secrets-ssm-environment-variables"
```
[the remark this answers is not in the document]
yumex93: This is how cp implemented in MTS so far. We need to discuss about what is the name we want. This is not really consistent with how we named private registry.

agent/taskresource/ssmsecret/ssmsecret.go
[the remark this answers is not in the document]
yumex93: I will change to arn package. However, after parsed with arn package, the resource name still have the prefix 'parameter/'

agent/api/task/task.go
```diff
@@ -497,6 +502,53 @@ func (task *Task) getAllASMAuthDataRequirements() []*apicontainer.ASMAuthData {
...
+
+	for _, container := range task.Containers {
+		for _, secret := range container.Secrets {
+			if secret.Provider == apicontainer.SecretProviderSSM {
+				if _, ok := reqs[secret.Region]; !ok {
+					reqs[secret.Region] = []apicontainer.Secret{}
+				}
+
+				allSecrets := reqs[secret.Region]
+				allSecrets = append(allSecrets, secret)
```
sharanyad: nit: this can just be `reqs[secret.Region] = append(reqs[secret.Region], secret)`?

agent/api/task/task_test.go
```diff
@@ -1713,3 +1715,138 @@ func TestPostUnmarshalTaskASMDockerAuth(t *testing.T) {
...
+	task := &Task{
+		Arn:                "test",
+		ResourcesMapUnsafe: make(map[string][]taskresource.TaskResource),
+		Containers:         []*apicontainer.Container{container},
+	}
+
+	reqs := task.getAllSSMSecretRequirements()
+	t.Log(reqs)
+	assert.Equal(t, secret1, reqs[regionWest][0])
+	assert.Equal(t, secret3, reqs[regionEast][0])
```
sharanyad: Please add an assert for secret2 as well.

agent/taskresource/ssmsecret/ssmsecretstatus.go
```diff
@@ -0,0 +1,78 @@
...
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+)
+
+type SSMSecretStatus resourcestatus.ResourceStatus
+
+const (
+	// is the zero state of a task resource
+	SSMSecretStatusNone SSMSecretStatus = iota
+	// represents a task resource which has been created
+	SSMSecretStatusCreated
```
sharanyad: This can just be `SSMSecretCreated` and `SSMSecretRemoved`

agent/api/task/task.go
```diff
@@ -497,6 +502,53 @@ func (task *Task) getAllASMAuthDataRequirements() []*apicontainer.ASMAuthData {
 	return reqs
 }
 
+// requiresSSMSecret returns true if at least one container in the task
+// needs to retrieve secret from SSM parameter
+func (task *Task) requiresSSMSecret() bool {
```
sharanyad: Please add unit tests for this method.

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+		// No need to log here. managedTask.applyResourceState already does that
+		return errors.New("ssm secret resource: unable to find execution role credentials")
+	}
+	iamCredentials := executionCredentials.GetIAMRoleCredentials()
+
+	var wg sync.WaitGroup
+	chanLen := secret.getGoRoutineTotalNum()
+	errorEvents := make(chan error, chanLen)
+	seelog.Infof("ssm secret resource: retrieving secrets for containers in task: [%s]", secret.taskARN)
+	if secret.secretData == nil {
```
sharanyad: Wouldn't this be nil here anyway? if yes, we can initialize this field in `NewSSMSecretResource`.

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+	default:
+		break
+	}
+	return nil
+}
+
+func (secret *SSMSecretResource) getGoRoutineTotalNum() int {
+	total := 0
+	for _, secrets := range secret.requiredSecrets {
+		total += len(secrets)/10 + 1
```
sharanyad: what is `10`?

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+	//To fail fast, check execution role first
+	executionCredentials, ok := secret.credentialsManager.GetTaskCredentials(secret.GetExecutionCredentialsID())
+	if !ok {
+		// No need to log here. managedTask.applyResourceState already does that
+		return errors.New("ssm secret resource: unable to find execution role credentials")
+	}
+	iamCredentials := executionCredentials.GetIAMRoleCredentials()
+
+	var wg sync.WaitGroup
+	chanLen := secret.getGoRoutineTotalNum()
```
sharanyad: Can you add comments explaining why we do this?

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+		index := strings.Index(valueFrom, ParameterARNPrefix)
+		return string(valueFrom[index+len(ParameterARNPrefix):])
+	} else {
+		return valueFrom
+	}
+
+}
+
+// GetRequiredSecrets returns the requiredSecrets field of ssmsecret task resource
+func (secret *SSMSecretResource) GetRequiredSecrets() map[string][]apicontainer.Secret {
```
sharanyad: This and `GetExecutionCredentialsID` need not be a public method?

agent/taskresource/ssmsecret/ssmsecret.go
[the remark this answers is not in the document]
petderek: So functionally your code will be the same either way. 

Could you consider doing a string.Split instead of a regex match?

agent/ssm/factory/generate_mocks.go
petderek: Hey -- lets revisit this later. Its established precedent that we can address in a different change.

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+	// the resource status needs to be reset to 'NONE' status so the secret value
+	// will be retrieved again
+
+	if taskKnownStatus < status.TaskCreated &&
+		taskDesiredStatus <= status.TaskRunning {
+		secret.SetKnownStatus(resourcestatus.ResourceStatusNone)
+	}
+}
+
+type SSMSecretResourceJSON struct {
```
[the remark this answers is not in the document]
yumex93: Bump to 16

agent/app/agent_capability.go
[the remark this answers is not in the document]
sharanyad: In that case, if there are environment variables and tmpfs support for secret provider `ssm` and only env var support for secret provider `xyz`, there will be 4 different capabilities as follows:
```
secrets-environment-variables
secrets-provider-ssm
secrets-tmpfs
secrets-provider-xyz
```

How will backend recognize if `xyz` supports tmpfs or not?

agent/ssm/ssm.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sharanyad: Any reason we don't have this inside `agent/taskresource` like `agent/taskresource/ssmsecret/ssm`? 
I don't see any use cases where agent would use `ssm` in a standalone way.

agent/taskresource/ssmsecret/ssmsecretstatus.go
```diff
@@ -0,0 +1,78 @@
...
+			return k
+		}
+	}
+	return "NONE"
+}
+
+// MarshalJSON overrides the logic for JSON-encoding the ResourceStatus type
+func (as *SSMSecretStatus) MarshalJSON() ([]byte, error) {
+	if as == nil {
+		return nil, errors.New("asm-auth resource status is nil")
```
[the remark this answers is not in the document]
yumex93: Updated

agent/ssm/ssm.go
```diff
@@ -0,0 +1,67 @@
...
+// retrieve secrets value in batches
+func GetSecretsFromSSM(names []string, client ssmiface.SSMAPI) (map[string]string, error) {
+	var secretNames []*string
+	for _, name := range names {
+		secretNames = append(secretNames, aws.String(name))
+	}
+
+	in := &ssm.GetParametersInput{
+		Names:          secretNames,
+		WithDecryption: aws.Bool(true),
```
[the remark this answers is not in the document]
yumex93: If it is unencrypted parameter, WithDecryption will be automatically ignored.

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: No. This should be the same number as goroutines we spin up to call GetParameters API. So it may not match the number of keys since one region can have multiple goroutines. This is the max number since we dedupe the secrets based on valueFrom + region.

agent/taskresource/ssmsecret/ssmsecret.go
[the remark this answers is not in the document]
yumex93: This is max number. I can refactor the name and add comments to describe the function.

agent/api/task/task_test.go
```diff
@@ -1713,3 +1715,138 @@ func TestPostUnmarshalTaskASMDockerAuth(t *testing.T) {
...
+	}
+
+	task := &Task{
+		Arn:                "test",
+		ResourcesMapUnsafe: make(map[string][]taskresource.TaskResource),
+		Containers:         []*apicontainer.Container{container},
+	}
+
+	reqs := task.getAllSSMSecretRequirements()
+	t.Log(reqs)
```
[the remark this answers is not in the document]
yumex93: Good catch. Removed

agent/app/agent_capability.go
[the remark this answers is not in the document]
petderek: > How will backend recognize if xyz supports tmpfs or not?

Does it matter? We have providers (ssm, xyz) and local mounts (env vars, tmpfs mounts). As far as I can tell, this should work for any combination.

My concern is that we add additional providers or mount points in the future and we have to extrapolate capabilities for *every combination* (providers * mounts) and not only the individual features (providers + mounts). 

Consider if we add support for S3 and Secrets Manager as providers and an http endpoint as a mount point.  That means that we need 9 capabilities instead of 6.

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+
+// Cleanup removes the secret value created for the task
+func (secret *SSMSecretResource) Cleanup() error {
+	secret.clearSSMSecretValue()
+	return nil
+}
+
+// clearSSMSecretValue cycles through the collection of secret value data and
+// removes them from the task
+func (secret *SSMSecretResource) clearSSMSecretValue() {
```
petderek: Why do we need this? Won't this get garbage collected?

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+// GetExecutionCredentialsID returns the execution role's credential ID
+func (secret *SSMSecretResource) GetExecutionCredentialsID() string {
+	secret.lock.RLock()
+	defer secret.lock.RUnlock()
+
+	return secret.executionCredentialsID
+}
+
+// Cleanup removes the secret value created for the task
+func (secret *SSMSecretResource) Cleanup() error {
```
petderek: Since we dont have a persisted resources, I think this can just return nil

agent/app/agent_capability.go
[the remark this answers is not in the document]
samuelkarp: I think if we leave the storage (environment variable and tmpfs) separate from the provider (ssm, xyz), then we're creating a requirement that we can use every combination of provider + storage.  If we ever break that, and say xyz can provide tmpfs but not environment variables, we then do not have a mechanism to indicate the capability in the future.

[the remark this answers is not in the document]
sharanyad: > Does it matter? We have providers (ssm, xyz) and local mounts (env vars, tmpfs mounts). As far as I can tell, this should work for any combination.

Yes, it does. 

Say, agent 1.22.0 supports the following: 
- environment vairables and tmpfs for `ssm` 
- only environment variables for `xyz`

So the capabilities would be - 
```
secrets-environment-variables
secrets-tmpfs
secrets-provider-ssm
secrets-provider-xyz
```

Now say, agent 1.23.0 supports the following: 
- tmpfs support for `xyz`
The capabilities would still be - 
```
secrets-environment-variables
secrets-tmpfs
secrets-provider-ssm
secrets-provider-xyz
```

In this case, if I want to run a task with tmpfs support for `xyz`, how will backend know (according to capabilities) if agent 1.22.0 supports it or agent 1.23.0?

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+	chanLen := secret.getGoRoutineTotalNum()
+	errorEvents := make(chan error, chanLen)
+	seelog.Infof("ssm secret resource: retrieving secrets for containers in task: [%s]", secret.taskARN)
+	if secret.secretData == nil {
+		secret.secretData = make(map[string]string)
+	}
+
+	for region, secrets := range secret.GetRequiredSecrets() {
+		wg.Add(1)
+		go secret.retrieveSSMSecretValuesByRegion(region, secrets, iamCredentials, &wg, errorEvents)
```
petderek: this is pretty neat

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+	}
+
+	wg.Wait()
+	//get the first error returned and set as terminal reason
+	select {
+	case err := <-errorEvents:
+		secret.setTerminalReason(err.Error())
+		return err
+	default:
+		break
```
petderek: I'd probably just `return nil` instead of break

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+
+func (secret *SSMSecretResource) retrieveSSMSecretValuesByRegion(region string, secrets []apicontainer.Secret, iamCredentials credentials.IAMRoleCredentials, wg *sync.WaitGroup, errorEvents chan error) {
+	seelog.Infof("ssm secret resource: retrieving secrets for region %s in task: [%s]", region, secret.taskARN)
+	defer wg.Done()
+
+	var wgPerRegion sync.WaitGroup
+	var secretNames []string
+
+	for _, s := range secrets {
+		secretName := secret.extractNameFromValueFrom(s)
```
petderek: `parseNameFromArn` would be a better hint at what this does.

agent/taskresource/ssmsecret/ssmsecret.go
petderek: I like this pattern. One thing I'd like to change is the WaitGroup behavior. I'd rather not pass it into the async function, and instead I'd like to keep it scoped to this function only.

We can do this by doing:

```
go func () {
    secret.retrieveSSMSecretValuesByRegion(region, secrets, iamCredentials, errorEvents)
    wg.Done()
}()
```

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+		secretKey := secretName + "_" + region
+		if _, ok := secret.GetSSMSecretValue(secretKey); ok {
```
petderek: GetCachedValue

agent/app/agent_capability.go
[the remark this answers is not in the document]
petderek: > I think if we leave the storage (environment variable and tmpfs) separate from the provider (ssm, xyz), then we're creating a requirement that we can use every combination of provider + storage.

I was hoping this was our plan. It seems like we will be making arbitrary decisions about which providers support which mount types. Especially if we are only taking a text value and exposing it somewhere else.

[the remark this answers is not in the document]
petderek: If we anticipate that this isn't the case, then you folks are right and this is the best way to do it.

agent/ssm/ssm.go
yumex93: What if later on, private registry or other features also need ssm?

agent/acs/model/api/api-2.json
```diff
@@ -419,6 +434,32 @@
...
+    },
+    "SecretProvider":{
+      "type":"string",
+      "enum":["SSM"]
+    },
+    "SecretType":{
+      "type":"string",
+      "enum":[
+        "ENVIRONMENT_VARIABLE",
+        "MOUNT_POINT"
```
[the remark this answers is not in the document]
yumex93: We assume it will be implemented in phase two.

agent/ssm/ssm.go
sharanyad: Yeah that could be a possibility. 
How about moving it under `agent/taskresource/` then? 

It does not feel right having it directly inside agent package now, since it's not the one using it.

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,476 @@
...
+
+func (secret *SSMSecretResource) getGoRoutineTotalNum() int {
+	total := 0
+	for _, secrets := range secret.requiredSecrets {
+		total += len(secrets)/10 + 1
+	}
+	return total
+}
+
+func (secret *SSMSecretResource) retrieveSSMSecretValuesByRegion(region string, secrets []apicontainer.Secret, iamCredentials credentials.IAMRoleCredentials, wg *sync.WaitGroup, errorEvents chan error) {
```
petderek: Another thought: Maybe move the client and retrieval logic into a different file. Just because this file gets super long as it is. You could make the API just take a list of arns and return secrets.

agent/api/task/task.go
```diff
@@ -497,6 +502,53 @@ func (task *Task) getAllASMAuthDataRequirements() []*apicontainer.ASMAuthData {
...
+		if container.ShouldCreateWithSSMSecret() {
+			container.BuildResourceDependency(ssmSecretResource.GetName(),
+				resourcestatus.ResourceStatus(ssmsecret.SSMSecretStatusCreated),
+				apicontainerstatus.ContainerCreated)
+		}
+	}
+}
+
+func (task *Task) getAllSSMSecretRequirements() map[string][]apicontainer.Secret {
+	reqs := make(map[string][]apicontainer.Secret)
```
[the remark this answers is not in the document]
yumex93: Comment added.

agent/taskresource/ssmsecret/ssmsecretstatus.go
yumex93: Why we do not need SSMSecretNone? I need this status when I initialize secret resource.

sharanyad: oh, what I meant was we can leave out `Status` in the enum value. yes, we would need `None` as well.

agent/api/task/task.go
yumex93: Updated

agent/api/container/container.go
```diff
@@ -25,7 +25,7 @@ import (
 	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
 	"github.com/aws/aws-sdk-go/aws"
 
-	docker "github.com/fsouza/go-dockerclient"
+	"github.com/fsouza/go-dockerclient"
```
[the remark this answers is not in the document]
yumex93: It is auto corrected by the IDE. Added it back

agent/api/task/task_test.go
yumex93: Added

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: I added a new constant to replace 10. It is the max parameter numbers that GetParameters API call can accept at a time. It is API constraint.

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: Removed this function, since after discussing with CP again, this logic moves to ACS.

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: Want to make sure, we remove all secret values.

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: I use public method here, so it can also used outside package if we need. It is one of those get/set functions to retrieve/set attributes outside packages. Even so far, it is only used within the package.

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: I tried the way you proposed here, seems does not work as expected.

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: This cannot be moved to NewSSMSecretResource. NewSSMSecretResource only be called in PostUnmashallTask, which will not be called when Agent restart and restore states from state file. Since we do not store secretData field in state file, under this situation, the secretData can also be nil, which also needs to be initialized. So this has to be kept in Create function.

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: Add comment for essential functions I added.

agent/taskresource/ssmsecret/ssmsecret.go
sharanyad: I believe the resource is accessed only using the interface methods outside the package. If it is not used outside the package, let's make them private.

agent/taskresource/ssmsecret/ssmsecret_test.go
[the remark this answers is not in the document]
yumex93: Test cases added

agent/taskresource/ssmsecret/ssmsecretstatus.go
```diff
@@ -0,0 +1,78 @@
...
+const (
+	// is the zero state of a task resource
+	SSMSecretStatusNone SSMSecretStatus = iota
+	// represents a task resource which has been created
+	SSMSecretStatusCreated
+	// represents a task resource which has been cleaned up
+	SSMSecretStatusRemoved
+)
+
+var SSMSecretStatusMap = map[string]SSMSecretStatus{
```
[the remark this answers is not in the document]
yumex93: Update it to be private.

agent/api/task/task_test.go
yumex93: I added another assertion for the length of secrets in regionWest to show the secret2 hasn't be added.

agent/api/task/task.go
yumex93: Test cases added.

agent/ssm/generate_mocks.go
yumex93: The mockgen.go seems does not take extra input like methods, how can we achieve this?

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: I see ur point, updated to private

agent/api/container/container_test.go
```diff
@@ -264,3 +264,42 @@ func TestInjectV3MetadataEndpoint(t *testing.T) {
 	assert.Equal(t, container.Environment[MetadataURIEnvironmentVariableName],
 		fmt.Sprintf(MetadataURIFormat, "myV3EndpointID"))
 }
+
+func TestShouldCreateWithSSMSecret(t *testing.T) {
```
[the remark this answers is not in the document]
yumex93: Updated

agent/ssm/ssm.go
adnxn: > It is a little bit weird for me that all other aws services packages are under agent folder 

you're right. historically we've kept service packages under the `agent`, i agree wth you and im inclined to leave it this way for now with `ssm` as well. 

maybe we can through everything under a `agent/awsservices` pkg? but that's refactoring work that can wait.

agent/taskresource/ssmsecret/ssmsecret.go
adnxn: > I tried the way you proposed here, seems does not work as expected.

🤔  really

agent/ssm/ssm.go
sharanyad: sometimes our existing code base is not the best example :) we can leave this as is and refactor later when needed.

agent/taskresource/ssmsecret/ssmsecret.go
sharanyad: I see. This brings a concern as to how we will retrieve the secrets when the resource is already created (status stored in state file) and agent restarts after that. In that case, don't we lose the retrieved secrets?

agent/taskresource/ssmsecret/ssmsecret.go
yumex93: Do you think it is reasonable that I add a todo here to refactor the code as the way you described here? I want to get this pr approved to unblock Adnan first and then if we have spare time, I can go back and refactor it.

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -0,0 +1,466 @@
...
+	secret.initStatusToTransition()
+	secret.credentialsManager = resourceFields.CredentialsManager
+	secret.ssmClientCreator = resourceFields.SSMClientCreator
+
+	// if task hasn't turn to 'created' status, and it's desire status is 'running'
+	// the resource status needs to be reset to 'NONE' status so the secret value
+	// will be retrieved again
+	if taskKnownStatus < status.TaskCreated &&
+		taskDesiredStatus <= status.TaskRunning {
+		secret.SetKnownStatus(resourcestatus.ResourceStatusNone)
```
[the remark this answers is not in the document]
yumex93: I think the desired status of resource under such situations should be its steady status 'created' already, should be no need to set it again.

agent/api/container/container_test.go
```diff
@@ -264,3 +296,10 @@ func TestInjectV3MetadataEndpoint(t *testing.T) {
...
+	for _, test := range shouldCreateWithSecretsTests {
+		container := test.in
+		assert.Equal(t, test.out, container.ShouldCreateWithSSMSecret())
+	}
+}
```
sharanyad: nit: empty line.

PR #1603, opened 2018-10-05
Cleaning Volumes after Integration Tests

linkar-ec2:
### Summary

TestSharedAutoprovisionVolume, TestSharedDoNotAutoprovisionVolume, and TestHostVolumeMount provision volumes but do not remove them after the test completes. Running Integration Tests repeatedly on an EC2 instance breaks because these volumes already exist. 

### Implementation details

Uses the Docker API to remove the volumes after the test completes. 

### Testing

Successfully ran integration tests repeatedly on EC2 instance.

New tests cover the changes: 
N/A

### Licensing

### DIFF
--- a/agent/engine/engine_integ_test.go
+++ b/agent/engine/engine_integ_test.go
@@ -48,6 +48,7 @@ const (
 	dialTimeout            = 200 * time.Millisecond
 	localhost              = "127.0.0.1"
 	waitForDockerDuration  = 50 * time.Millisecond
+	removeVolumeTimeout    = 5 * time.Second
 )
 
 func init() {
@@ -94,6 +95,13 @@ func removeImage(t *testing.T, img string) {
 	client, err := docker.NewClient(endpoint)
 	require.NoError(t, err, "create docker client failed")
 	client.RemoveImage(img)
+}
+
+func cleanVolumes(testTask *apitask.Task, taskEngine TaskEngine) {
+	client := taskEngine.(*DockerTaskEngine).client
+	for _, aVolume := range testTask.Volumes {
+		client.RemoveVolume(context.TODO(), aVolume.Name, removeVolumeTimeout)
+	}
 }
 
 // TestDockerStateToContainerState tests convert the container status from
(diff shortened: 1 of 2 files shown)


agent/engine/engine_integ_test.go
```diff
@@ -169,6 +169,11 @@ func TestHostVolumeMount(t *testing.T) {
 	data, err := ioutil.ReadFile(filepath.Join(tmpPath, "hello-from-container"))
 	assert.Nil(t, err, "Unexpected error")
 	assert.Equal(t, "hi", strings.TrimSpace(string(data)), "Incorrect file contents")
+
+	client := taskEngine.(*DockerTaskEngine).client
+	for _, aVolume := range testTask.Volumes {
+		client.RemoveVolume(context.TODO(), aVolume.Name, 5*time.Second)
```
haikuoliu: You can create a constant for those `5*time.Second`, or just use default remove volume timeout.

agent/engine/engine_integ_test.go
```diff
@@ -169,6 +170,11 @@ func TestHostVolumeMount(t *testing.T) {
 	data, err := ioutil.ReadFile(filepath.Join(tmpPath, "hello-from-container"))
 	assert.Nil(t, err, "Unexpected error")
 	assert.Equal(t, "hi", strings.TrimSpace(string(data)), "Incorrect file contents")
+
+	client := taskEngine.(*DockerTaskEngine).client
```
petderek: What do you think about putting the cleanup code in a `defer` block above the main code and verifications?

Same for the other Tests.

linkar-ec2: Good idea! Will incorporate

linkar-ec2: using a defer is not as appropriate in this situation because a defer is used to clean up the task engine. We can factor out the code and place it at the end of each method tho.

linkar-ec2:
Failed Functional Test is a flakey test. Continuing with merge.

PR #1610, opened 2018-10-10
amazon-ecs-cni-plugin: update to rev 2018.10.0@93f4377

adnxn:
### Summary

amazon-ecs-cni-plugin: update to rev 2018.10.0@93f4377 
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/ecscni/plugin_test.go
+++ b/agent/ecscni/plugin_test.go
@@ -1,4 +1,4 @@
-// +build unit
+// +build linux,unit
 
 // Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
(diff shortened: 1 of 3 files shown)


agent/ecscni/plugin_test.go
```diff
@@ -1,4 +1,4 @@
-// +build unit
+// +build linux,unit
```
shubham2892: Does this mean that these test will run only on Linux?

adnxn: yes, task networking is not supported on windows. i don't see why these code paths need to be tested on windows.

sharanyad: I think `agent/ecscni/plugin.go` itself is not linux specific, that's why we have the tests running in both. 
Can you add a note we need to change that as well and also the octokitten for windows that was updated and not needed anymore?

PR #1618, opened 2018-10-12
ECS Agent tagging support

haikuoliu:
### Summary

ECS local tags support

### Implementation details

1. Introduce `ECS_CONTAINER_INSTANCE_TAGS` env var to allow customer to configure tags on the instance.
2. Introduce `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS` to get the tags of the ec2 instance that Agent is running on using DescribeTags API. We can specify either `none` or `ec2_instance` for this env var, the reason that we don't use a boolean is because we may introduce other ways to propagate the tags (like ask backend to propagate tags instead of doing it on Agent side).
3. ECS model changes for tags, introduce a new API `ListTagsForResourceInput` for testing purpose.
4. ECS client API changes to allow tags passed from RCI call.
5. Introduce EC2 client API, `DescribeECSTagsForInstance` is used for RCI call, and `CreateTags` for func test.
6. Add `InstanceID` API to EC2 metadata client.
6. Functional test and manual test against Gamma endpoint.
7. Vendor packages update.
8. Joel reviewed my README.

TODO:
Make func test pass once the `ListTagsForResource` API is in prod.

### Testing

### Licensing


haikuoliu:
Func tests are expected to fail as the `ListTagsForResource` API is not in prod yet.

agent/config/config.go
```diff
@@ -125,6 +125,16 @@ const (
...
 )
 
+const (
+	// When ContainerInstancePropagateTagsEC2InstanceType is specified, no DescribeTags
+	// API call will be made.
+	ContainerInstancePropagateTagsNoneType ContainerInstancePropagateTagsType = iota
+
+	// When ContainerInstancePropagateTagsEC2InstanceType is specified, agent will
+	// make DescribeTags API call to get tags remotely.
+	ContainerInstancePropagateTagsEC2InstanceType
```
petderek: Does this get stored in the state file? if so, we should test that the values of these don't change. (IE, we don't accidentally reorder them in the future)

agent/config/config.go
```diff
@@ -406,6 +416,9 @@ func environmentConfig() (Config, error) {
 	var errs []error
 	instanceAttributes, errs = parseInstanceAttributes(errs)
 
+	var containerInstanceTags map[string]string
+	containerInstanceTags, errs = parseContainerInstanceTags(errs)
```
petderek: This might be a silly question, but why doesn't `:=` work?

agent/ec2/ec2_client.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/aws/aws-sdk-go/aws/session"
+	ec2sdk "github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/cihub/seelog"
+)
+
+const (
+	clientRetriesNum = 5
+
+	// InstanceMaxTagsNum is the maximum number of tags that an instance can have.
+	InstanceMaxTagsNum = 50
```
petderek: I wouldn't hardcode this expectation -- just in case they change in a future version.

agent/config/config_test.go
```diff
@@ -592,6 +601,49 @@ func TestUserDataConfig(t *testing.T) {
 	}
 }
```
petderek: Add test cases for the key:value pairs?

agent/config/config_test.go
```diff
@@ -592,6 +601,49 @@ func TestUserDataConfig(t *testing.T) {
...
+		{
+			name:      "invalid container instance propagate tags",
+			envVarVal: "none",
+			expectedContainerInstancePropagateTags: ContainerInstancePropagateTagsNoneType,
+		},
+	}
+
+	for _, tc := range testcases {
+		// Test the parse function only.
+		t.Run(tc.name, func(t *testing.T) {
```
petderek: I'd probably merge these two t.Run functions together and just make sure the "message" includes a note that the parsing didn't work as expected.

agent/utils/utils.go
```diff
@@ -208,3 +210,20 @@ func IsAWSErrorCodeEqual(err error, code string) bool {
 	awsErr, ok := err.(awserr.Error)
 	return ok && awsErr.Code() == code
 }
+
+// MapToTags converts a map to a slice of tags.
+func MapToTags(tagsMap map[string]string) []*ecs.Tag {
```
petderek: FFTI: I'd consider moving this and the other tagging-specific functions (mergeTags) into a tagging-specific file / package.

agent/ec2/ec2_client.go
haikuoliu: Well I don't want to hard code it as well, the API provides this parameter but it doesn't mention the default value, this number can be 5-1000, so we end up 3 options:
1. Use 50, as it's mentioned in EC2 public doc that an instance can have a maximum of 50 tags.
2. Use 1000, and assume that the max num of tags won't be changed to a number greater than 1000.
3. Don't provide this parameter or use a number like 50 (or even 1000), make multiple calls until we cannot get more tags.

Which one do you prefer? I'd say option 3 with the max tags number being 1000 is the safest option, looks like a bit wired tho as we already know 50 would be enough.

agent/config/config_test.go
haikuoliu: I think I've already added [here](https://github.com/aws/amazon-ecs-agent/pull/1618/commits/ed4f271d761ebb56d6c1836e1574ff93a7dfeaae#diff-540a525277ce057e5db7c092fb41f78fR90). LMK if you want me to include more test cases.

agent/api/ecsclient/client_test.go
```diff
@@ -823,3 +855,11 @@ func TestSubmitContainerStateChangeWhileTaskInPending(t *testing.T) {
 		})
 	}
 }
+
+func extractTagsMapFromRegisterContainerInstanceInput(req *ecs.RegisterContainerInstanceInput) map[string]string {
+	tagsMap := make(map[string]string, len(req.Tags))
+	for i := range req.Tags {
+		tagsMap[*req.Tags[i].Key] = aws.StringValue(req.Tags[i].Value)
```
yumex93: Why key is using pointer dereference while value is using aws.StringValue here?

agent/ec2/mocks/ec2_mocks.go
```diff
@@ -21,7 +21,9 @@ import (
 	reflect "reflect"
 
 	ec2 "github.com/aws/amazon-ecs-agent/agent/ec2"
+	ecs "github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	ec2metadata "github.com/aws/aws-sdk-go/aws/ec2metadata"
+	ec20 "github.com/aws/aws-sdk-go/service/ec2"
```
yumex93: This ec20 is a little bit weird as a name. Maybe change '/agent/ec2' to **agent_ec2** and keep '/service/ec2' as **ec2**.

haikuoliu: I don't know, this is something auto-generated...

README.md
```diff
@@ -131,6 +131,8 @@ additional details on each available environment variable.
 | `ECS_ENABLE_CPU_UNBOUNDED_WINDOWS_WORKAROUND` | `true` | When `true`, ECS will allow CPU unbounded(CPU=`0`) tasks to run along with CPU bounded tasks in Windows. | Not applicable | `false` |
 | `ECS_TASK_METADATA_RPS_LIMIT` | `100,150` | Comma separated integer values for steady state and burst throttle limits for task metadata endpoint | `40,60` | `40,60` |
 | `ECS_SHARED_VOLUME_MATCH_FULL_CONFIG` | `true` | When `true`, ECS Agent will compare name, driver options, and labels to make sure volumes are identical. When `false`, Agent will short circuit shared volume comparison if the names match. This is the default Docker behavior. If a volume is shared across instances, this should be set to `false`. | `false` | `false`|
+| `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS` | `"ec2_instance"` | If `"ec2_instance"` is specified, tags of the ec2 instance will be registered to ECS, make sure you allow `ec2:DescribeTags` action in your instance profile. | `"none"` | `"none"` |
```
sharanyad: Name `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS` can be a bit confusing and it sounds like a bool var.

how about something like - 
`ECS_INSTANCE_REGISTER_TAG_TYPES`  or `ECS_INSTANCE_OBTAIN_TAG_TYPES`?

agent/config/parse.go
```diff
@@ -206,6 +206,39 @@ func parseTaskMetadataThrottles() (int, int) {
...
+
+	for tagKey, tagValue := range containerInstanceTags {
+		seelog.Debugf("Setting instance tag %v: %v", tagKey, tagValue)
+	}
+
+	return containerInstanceTags, errs
+}
+
+func parseContainerInstancePropagateTags() ContainerInstancePropagateTagsType {
+	containerInstancePropagateTagsString := os.Getenv("ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS")
```
sharanyad: I feel like `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS` should be a list of strings rather than one, if we want to support multiple tag types in the future and not break backwards compatibility.

haikuoliu: I think customer only has one option - either ask Agent to propagate the tags or ask backend to propagate the tags, customer will get the same tags either way, so supporting both of them seems to do duplicate work.

README.md
haikuoliu: Confirmed with Akram, decided to go with `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS_FROM`.

agent/config/parse.go
haikuoliu: Yes, it can be other values, but even if there are other possible values, customer can only use one of them. For example, if we have `backend` as one of the values in the future, customer only needs `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS` to be either `ec2_instance` or `backend`, because we will get the same set of tags when we ask backend to propagate or we let Agent propagate tags.

haikuoliu:
@yumex93 @sharanyad I've refactored accordingly.

Also introduced PutAccountSetting API used in functional test, this is required for tags related API.

CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # Changelog
 
+## 1.22.0-dev
+* Feature - Introduce two environment variables to support ECS tagging [#1618](https://github.com/aws/amazon-ecs-agent/pull/1618)
```
sharanyad: Can we also specify what they are?

agent/ecs_client/model/api/api-2.json
```diff
@@ -310,6 +325,20 @@
         {"shape":"ServiceNotFoundException"}
       ]
     },
+    "PutAccountSetting":{
```
sharanyad: is this part of tagging?

agent/app/agent_test.go
```diff
@@ -922,6 +922,79 @@ func TestRegisterContainerInstanceInvalidParameterTerminalError(t *testing.T) {
...
+
+	assert.Equal(t, commonKey, aws.StringValue(mergedTags[0].Key))
+	assert.Equal(t, commonKeyLocalValue, aws.StringValue(mergedTags[0].Value))
+	assert.Equal(t, ec2Key, aws.StringValue(mergedTags[1].Key))
+	assert.Equal(t, ec2Value, aws.StringValue(mergedTags[1].Value))
+	assert.Equal(t, localKey, aws.StringValue(mergedTags[2].Key))
+	assert.Equal(t, localValue, aws.StringValue(mergedTags[2].Value))
+}
+
+func TestGetContainerInstanceTagsFromEC2API(t *testing.T) {
```
sharanyad: Please add negative test case as well

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -633,3 +634,79 @@ func testV3TaskEndpoint(t *testing.T, taskName, containerName, networkMode, awsl
...
+
+	// Set the env var for tags and start Agent.
+	agentOptions := &AgentOptions{
+		ExtraEnvironment: map[string]string{
+			"ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS_FROM": "ec2_instance",
+			"ECS_CONTAINER_INSTANCE_TAGS": fmt.Sprintf("{\"%s\": \"%s\", \"%s\": \"%s\"}",
+				localKey, localValue, commonKey, commonKeyLocalValue),
+		},
+	}
+	agent := RunAgent(t, agentOptions)
```
sharanyad: this test will require specific agent version.

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -633,3 +634,79 @@ func testV3TaskEndpoint(t *testing.T, taskName, containerName, networkMode, awsl
 	exitCode, _ := task.ContainerExitcode(containerName)
 	assert.Equal(t, 42, exitCode, fmt.Sprintf("Expected exit code of 42; got %d", exitCode))
 }
+
+func TestContainerInstanceTags(t *testing.T) {
+	// We need long container instance ARN for tagging APIs, PutAccountSettingInput
+	// will enable long container instance ARN.
+	putAccountSettingInput := ecsapi.PutAccountSettingInput{
```
sharanyad: Will customers have to manually enable this?

agent/ec2/ec2_client.go
sharanyad: Do we know if customers will ever have >50 tags?

agent/ecs_client/model/api/api-2.json
haikuoliu: Yes, used in func test.

agent/functional_tests/tests/functionaltests_test.go
haikuoliu: yes, it will be described in doc

agent/ec2/ec2_client.go
haikuoliu: Yes, docs will test customer that the max num of tags for each ECS resource is 50.

agent/config/config_test.go
```diff
@@ -88,6 +88,7 @@ func TestEnvironmentConfig(t *testing.T) {
 	defer setTestEnv("ECS_NUM_IMAGES_DELETE_PER_CYCLE", "2")()
 	defer setTestEnv("ECS_IMAGE_PULL_BEHAVIOR", "always")()
 	defer setTestEnv("ECS_INSTANCE_ATTRIBUTES", "{\"my_attribute\": \"testing\"}")()
+	defer setTestEnv("ECS_CONTAINER_INSTANCE_TAGS", "{\"my_tag\": \"testing\"}")()
```
petderek: nit: Use string literals instead of escaping:

```
defer setTestEnv("ECS_CONTAINER_INSTANCE_TAGS", `{"my_tag": "testing"}`)() 
```

agent/ec2/ec2_client.go
petderek: Makes sense!

agent/ec2/mocks/ec2_mocks.go
petderek: its the worst!

FLavalliere:
Hello is this : ECS_INSTANCE_ATTRIBUTES=  still supposed  to be functional with 1.22 ? It looks like my ECS Agent shutdown 
{code}
DEBUG] Added a new custom attribute xxxx
DEBUG] Added a new custom attribute yyyy
[ERROR] Error registering attributes: xxxx,yyyy
{code}

And prior 1.22 it seemed to be functional...

PR #1621, opened 2018-10-14
api/engine: pass ssm parameters to container via envvars

adnxn:
### Summary

This commit adds changes to take prepared ssmsecret resources and inject
them into the container's env struct so it is added to the docker
container config before create container is called.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -758,3 +758,19 @@ func (c *Container) ShouldCreateWithSSMSecret() bool {
 	}
 	return false
 }
+
+// MergeEnvironmentVariables appends additional envVarName:envVarValue pairs to
+// the the container's enviornment values structure
+func (c *Container) MergeEnvironmentVariables(secrets map[string]string) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// don't assume that the environment variable map has been initialized by others
+	if c.Environment == nil {
+		c.Environment = make(map[string]string)
+	}
+
+	for k, v := range secrets {
+		c.Environment[k] = v
+	}
+}
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -257,6 +257,12 @@ type Secret struct {
 	Provider      string `json:"provider"`
 }
 
+// GetSSMSecretResourceCacheKey returns the key required to access the secret
+// from the ssmsecret resource
+func (s *Secret) GetSSMSecretResourceCacheKey() string {
+	return s.ValueFrom + "_" + s.Region
+}
+
 // String returns a human readable string representation of DockerContainer
 func (dc *DockerContainer) String() string {
 	if dc == nil {
@@ -761,7 +767,7 @@ func (c *Container) ShouldCreateWithSSMSecret() bool {
 
 // MergeEnvironmentVariables appends additional envVarName:envVarValue pairs to
 // the the container's enviornment values structure
-func (c *Container) MergeEnvironmentVariables(secrets map[string]string) {
+func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
@@ -770,7 +776,7 @@ func (c *Container) MergeEnvironmentVariables(secrets map[string]string) {
 		c.Environment = make(map[string]string)
 	}
 
-	for k, v := range secrets {
+	for k, v := range envVars {
 		c.Environment[k] = v
 	}
 }
--- a/agent/api/container/container_test.go
+++ b/agent/api/container/container_test.go
@@ -17,6 +17,7 @@ package container
 
 import (
 	"fmt"
+	"reflect"
 	"testing"
 	"time"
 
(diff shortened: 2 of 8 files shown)


agent/api/container/container.go
```diff
@@ -758,3 +758,19 @@ func (c *Container) ShouldCreateWithSSMSecret() bool {
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	// don't assume that the environment variable map has been initialized by others
+	if c.Environment == nil {
+		c.Environment = make(map[string]string)
+	}
+
+	for k, v := range secrets {
+		c.Environment[k] = v
```
adnxn: @aws/aws-ecs-agent, @yumex93: how concerned should we be about user provided envvar names clobbering existing envvars? i'm not convinced we should be doing additional validation here.

agent/api/container/container.go
```diff
@@ -758,3 +758,19 @@ func (c *Container) ShouldCreateWithSSMSecret() bool {
 	}
 	return false
 }
+
+// MergeEnvironmentVariables appends additional envVarName:envVarValue pairs to
+// the the container's enviornment values structure
```
sharanyad: nit: extra the

agent/api/container/container.go
```diff
@@ -758,3 +758,19 @@ func (c *Container) ShouldCreateWithSSMSecret() bool {
...
+func (c *Container) MergeEnvironmentVariables(secrets map[string]string) {
```
sharanyad: I would name the method argument just `envVars` or something similar so this can be reused.

agent/api/task/task.go
```diff
@@ -1502,6 +1502,37 @@ func (task *Task) getASMAuthResource() ([]taskresource.TaskResource, bool) {
 	return res, ok
 }
 
+// PopulateSSMSecrets appends the container's env var map with ssm parameters
+func (task *Task) PopulateSSMSecrets(container *apicontainer.Container) *apierrors.DockerClientConfigError {
+	resource, ok := task.getSSMSecretsResource()
+	if !ok {
+		return &apierrors.DockerClientConfigError{"task secret data: unable to fetch SSM Secrets resource"}
```
sharanyad: why is this a `DockerClientConfigError`? This error happens when resource is not found in task resources map yeah?

agent/api/container/container.go
sharanyad: should do validation if there is an existing env var with the same key present?

agent/api/task/task.go
adnxn: yea i was wondering about this. thinking was `PopulateSSMSecrets` is acting on the docker config for this container. failure here halt the start container workflow because the config couldn't be constructed.

agent/api/container/container.go
adnxn: oh true, good catch.

agent/api/container/container.go
yumex93: I already talked with Feng when he started to implement MTS, this validation is done there, so in theory, agent does not need to worry about this.

adnxn: hm. how would we handle that case? hm maybe log.Warnf + not add to the map?

agent/taskresource/ssmsecret/ssmsecret.go
```diff
@@ -348,6 +348,15 @@ func (secret *SSMSecretResource) retrieveSSMSecretValues(region string, names []
 	}
 }
 
+// GetSecretValue returns the value retrieved from SSM
+func (secret *SSMSecretResource) GetSecretValue(key string) (string, bool) {
```
yumex93: There is another function in ssmsecret.go called getCachedSecretValue is doing the same thing as the function you added here. You may want to only keep one. Either remove that one or make that function public and remove this one. No matter what changes you made, do remember to update places calling getCachedSecretValue.

adnxn: wow how did i miss that.

agent/api/container/container_test.go
```diff
@@ -298,8 +267,109 @@ func TestInjectV3MetadataEndpoint(t *testing.T) {
...
+			OutEnvVarMap: map[string]string{
+				"SECRET1": "secret1",
+			},
+		},
+
+		{
+			Name: "merge zero items to existing container env var map",
+			InContainerEnvironment: map[string]string{
+				"CONFIG1": "config1"},
+			InEnvVarMap: map[string]string{},
```
yumex93: Just wonder in what situation the InEnvVarMap will be an empty map? Seems we check whether the container has secrets using **ShouldCreateWithSSMSecret** before calling **MergeEnvironmentVariables**?

agent/engine/docker_task_engine_test.go
```diff
@@ -1626,7 +1630,9 @@ func TestTaskUseExecutionRolePullPrivateRegistryImage(t *testing.T) {
 		ASMAuthData: asmAuthData,
 	}
 	requiredASMResources := []*apicontainer.ASMAuthData{asmAuthData}
-	asmClientCreator := mock_factory.NewMockClientCreator(ctrl)
+	asmClientCreator := mock_asm_factory.NewMockClientCreator(ctrl)
+	if false {
```
yumex93: For my own knowledge, what is this if false {} doing?

adnxn:
> code lgtm. are we adding an integ test for this later?

i didnt think it was a priority right now since it was a config (envvar) modification in the create container path. i'd like to move forward now and have functional tests exercise the whole workflow. we can revisit integ testing. lmk what you think.

PR #1623, opened 2018-10-15
Pullimage sdk migration

yunhee-l:
### Summary
Migrating pull image to SDK

### Implementation details
go-dockerclient used to handle a lot more including inactivity timeout and response "event" object from the image pull. These are handled in agent codebase now, reusing inactivity timeout from stats, and creating custom imagePullResponse object. There's an open issue for handling pull image response event on docker side (https://github.com/docker/engine-api/issues/89) so I didn't want to add premature dependency on it. 

### Testing

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -14,9 +14,11 @@
 package dockerapi
 
 import (
-	"bufio"
+	"bytes"
 	"context"
+	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"strconv"
@@ -47,7 +49,6 @@ import (
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/network"
 	"github.com/docker/docker/api/types/volume"
-	docker "github.com/fsouza/go-dockerclient"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -367,45 +382,79 @@ func wrapPullErrorAsNamedError(err error) apierrors.NamedError {
...
-
-	go dg.filterPullDebugOutput(pullDebugOut, pullBegan, image)
+	// pullBeganOnce ensures we only indicate it began once (since our channel will only be read 0 or 1 times)
+	pullBeganOnce := sync.Once{}
 
 	pullFinished := make(chan error, 1)
-	// TODO Migrate Pull Image once Inactivity Timeout is sorted out
+	subCtx, cancelRequest := context.WithCancel(ctx)
+
 	go func() {
```
linkar-ec2: Is this go routine along with the select statement necessary since we now implement the full go routine with the sdk change? It seems the select statement was only to catch a premature error (if nothing was being pulled), but now we can handle the error within the go routine without the select statement.

yunhee-l: given that images can be large and take long time to download, I think go routine here is necessary so that it is handled in a separate thread. I'm sure that this can be cleaned up, but I'd like to limit changes in behavior between go-dockerclient and SDK, and prefer to come back to this to make any improvements.

petderek: We're still using the goroutine for the pullBegan timeout. You could rework that, but I think its fine to only focus on translating the API for now.

PR #1522, opened 2018-08-15
Make docker health check configurable

richardpen:
### Summary

Fix #1235 Add an environment variable to disable the container health check.

### Testing

- [ ] Builds on Linux (`make release`)
- [ ] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [ ] Unit tests on Linux (`make test`) pass
- [ ] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [ ] Integration tests on Linux (`make run-integ-tests`) pass
- [ ] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [ ] Functional tests on Linux (`make run-functional-tests`) pass
- [ ] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: 

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1624, opened 2018-10-16
config: make docker health check configurable

adnxn:
### Summary

carried over from https://github.com/aws/amazon-ecs-agent/pull/1522

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/app/agent_capability_test.go
+++ b/agent/app/agent_capability_test.go
@@ -599,6 +599,42 @@ func TestCapabilitiesContainerHealth(t *testing.T) {
 	assert.True(t, ok, "Could not find container health check capability when expected; got capabilities %v", capabilities)
 }
 
+func TestCapabilitiesContainerHealthDisabled(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	client := mock_dockerapi.NewMockDockerClient(ctrl)
+	mockMobyPlugins := mock_mobypkgwrapper.NewMockPlugins(ctrl)
+
+	client.EXPECT().SupportedVersions().Return([]dockerclient.DockerVersion{
+		dockerclient.Version_1_24,
+	})
+	client.EXPECT().KnownVersions().Return(nil)
+	mockMobyPlugins.EXPECT().Scan().AnyTimes().Return([]string{}, nil)
+	client.EXPECT().ListPluginsWithFilters(gomock.Any(), gomock.Any(), gomock.Any(),
+		gomock.Any()).AnyTimes().Return([]string{}, nil)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	// Cancel the context to cancel async routines
+	defer cancel()
+	agent := &ecsAgent{
+		ctx:          ctx,
+		cfg:          &config.Config{DisableDockerHealthCheck: true},
+		dockerClient: client,
+		mobyPlugins:  mockMobyPlugins,
+	}
+
+	capabilities, err := agent.capabilities()
+	require.NoError(t, err)
+
+	capMap := make(map[string]bool)
+	for _, capability := range capabilities {
+		capMap[aws.StringValue(capability.Name)] = true
+	}
+
+	assert.NotContains(t, "ecs.capability.container-health-check", "Find container health check capability unexpected when it is disabled")
+}
+
 func TestCapabilitesListPluginsErrorCase(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
--- a/agent/tcs/handler/handler.go
+++ b/agent/tcs/handler/handler.go
@@ -47,7 +47,17 @@ const (
 // StartMetricsSession starts a metric session. It initializes the stats engine
 // and invokes StartSession.
 func StartMetricsSession(params TelemetrySessionParams) {
-	err := params.StatsEngine.MustInit(params.Ctx, params.TaskEngine, params.Cfg.Cluster,
+	ok, err := params.isContainerHealthMetricsDisabled()
+	if err != nil {
+		seelog.Warnf("Error starting metrics session: %v", err)
+		return
+	}
+	if ok {
+		seelog.Warnf("Metrics were disabled, not starting the telemetry session")
+		return
+	}
+
+	err = params.StatsEngine.MustInit(params.Ctx, params.TaskEngine, params.Cfg.Cluster,
 		params.ContainerInstanceArn)
 	if err != nil {
 		seelog.Warnf("Error initializing metrics engine: %v", err)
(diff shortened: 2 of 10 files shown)


agent/tcs/handler/types.go
```diff
@@ -50,3 +51,10 @@ func (params *TelemetrySessionParams) time() ttime.Time {
 	})
 	return params._time
 }
+
+func (params *TelemetrySessionParams) isMetricsDisabled() (bool, error) {
```
yumex93: The function name makes me a little bit confusing. It is called isMetricsDisabled, but when DisableMetrics flag is true, the function will not return true. Do you want to say MetricsSessionDisabled or something? Maybe change the function name and add some description as comments here.

agent/tcs/handler/types_test.go
```diff
@@ -0,0 +1,66 @@
...
+			result:      false,
+			err:         nil,
+			description: "Only telemetry metrics was disable should return false",
+		},
+		{
+			param:       &TelemetrySessionParams{Cfg: &config.Config{DisableMetrics: true, DisableDockerHealthCheck: true}},
+			result:      true,
+			err:         nil,
+			description: "both telemetry and health metrics were disable should return true",
+		},
```
yumex93: Is it in golang, for **&&** operator, if the first one is false, it will not proceed to check the value of second parameter? Since seems you miss a case, when **DisableMetrics** is false and **DisableDockerHealthCheck** is true.

agent/app/agent_capability_test.go
```diff
@@ -587,6 +587,43 @@ func TestCapabilitiesContainerHealth(t *testing.T) {
...
+	client.EXPECT().SupportedVersions().Return([]dockerclient.DockerVersion{
+		dockerclient.Version_1_24,
+	})
+	client.EXPECT().KnownVersions().Return(nil)
+	mockMobyPlugins.EXPECT().Scan().AnyTimes().Return([]string{}, nil)
+	client.EXPECT().ListPluginsWithFilters(gomock.Any(), gomock.Any(), gomock.Any(),
+		gomock.Any()).AnyTimes().Return([]string{}, nil)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	// Cancel the context to cancel async routines
```
fenxiong: why is this needed?

agent/app/agent_capability_test.go
```diff
@@ -587,6 +587,43 @@ func TestCapabilitiesContainerHealth(t *testing.T) {
...
+		ctx:          ctx,
+		cfg:          &config.Config{DisableDockerHealthCheck: true},
+		dockerClient: client,
+		mobyPlugins:  mockMobyPlugins,
+	}
+
+	capabilities, err := agent.capabilities()
+	require.NoError(t, err)
+
+	capMap := make(map[string]bool)
```
fenxiong: i think we can use [assert.NotContains](https://godoc.org/github.com/pgpst/pgpst/internal/github.com/stretchr/testify/assert#NotContains) instead of these

agent/tcs/handler/handler.go
```diff
@@ -47,7 +47,17 @@ const (
...
 // and invokes StartSession.
 func StartMetricsSession(params TelemetrySessionParams) {
-	err := params.StatsEngine.MustInit(params.Ctx, params.TaskEngine, params.Cfg.Cluster,
+	ok, err := params.isMetricsDisabled()
+	if err != nil {
+		seelog.Warnf("Error starting metrics session: %v", err)
+		return
+	}
+	if ok {
+		seelog.Warnf("Metrics were disabled, not start the telemetry session")
```
fenxiong: nit: start -> starting

agent/tcs/handler/handler_test.go
```diff
@@ -73,6 +73,19 @@ func (*mockStatsEngine) GetTaskHealthMetrics() (*ecstcs.HealthMetadata, []*ecstc
 	return nil, nil, nil
 }
 
+// TestDisableMetrics tests the StartMetricsSession will return immediately if
```
fenxiong: just wondering what is the result of breaking the logic we test here? it seems that in that case we will just not return immediately, but i'm not sure whether the test will fail?

agent/tcs/handler/types_test.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.description, func(t *testing.T) {
+			ok, err := tc.param.isMetricsDisabled()
+			if tc.err != nil {
+				assert.Error(t, err)
+			}
```
fenxiong: else we should assert no error?

agent/tcs/handler/types.go
```diff
@@ -50,3 +51,10 @@ func (params *TelemetrySessionParams) time() ttime.Time {
...
+	if params.Cfg != nil {
+		return params.Cfg.DisableMetrics && params.Cfg.DisableDockerHealthCheck, nil
```
haikuoliu: Just for understanding, is `DisableMetrics` something that we have introduced previously but never used? Would it be better to add a description in the PR summary for it?

adnxn:
thanks for taking a pass at this, will make changes and need to update the test for this actually as well.

agent/tcs/handler/types.go
adnxn: fair. changed to `isContainerHealthMetricsDisabled` since checking against both `DisabledMetrics && DisableDockerHealthCheck`

agent/app/agent_capability_test.go
adnxn: without it we end up with flakey tests that have run away go routines. maybe you've seen some errors similar to "goroutine ended after test" kind of thing. please correct me if i'm wrong though.

agent/tcs/handler/handler_test.go
adnxn: hrm yea this is a weird one, this looks like it needs a `StartMetricsSession` refactor to be able to have more sensible testing. we can track this as tech debt.

agent/tcs/handler/types.go
adnxn: `DisableMetrics` maps to `ECS_DISABLE_METRICS` and is already covered in the README. this PR is to add a toggle for the health checks, which relies on the underlying tcs/metrics subsystem.

agent/tcs/handler/handler_test.go
fenxiong: nit: add a TODO for this

adnxn:
windows ftest failing:

```
=== RUN   TestTwoTasksSharedLocalVolume
go : panic: test timed out after 30m0s
At C:\Users\Administrator\AppData\Local\Temp\amazon-ecs-agent\go\src\github.com\aws\amazon-ecs-agent\scripts\run-functi
onal-tests.ps1:26 char:3
+   go test -tags functional -timeout=30m -v ../agent/functional_tests/ ...
+   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (panic: test timed out after 30m0s:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
```

PR #1632, opened 2018-10-18
Make CGroups CPU period configurable

### LINKED ISSUE #1627: Make CFS quota period configurable
It seems that the CFS quota period is fixed in ecs agent for the parent cgroup that is created for each task. https://github.com/aws/amazon-ecs-agent/blob/d68e729f73e588982dc2189a1c618c18c47c931b/agent/api/task/task_linux.go#L39

Since the length of this period can have consequences on performance (see [this](https://github.com/kubernetes/kubernetes/issues/51135) and [this](https://gist.github.com/bobrik/2030ff040fad360327a5fab7a09c4ff1)) we would like it to be configurable, e.g. as an argument to ecs-agent.

boynux:
### Summary
Fixes #1627 
Make CFS CPU Period configurable

CFS keeps throttling tasks very excessively if the default period 100ms is specified, to reduce the impact we should be able to set lower CPU period when defining cgroup quotas.

### Implementation details
New environment parameter added to adjust CFS period.

### Testing
See here: https://gist.github.com/bobrik/2030ff040fad360327a5fab7a09c4ff1

-->

### Licensing


samuelkarp:
Hi @boynux, thanks for sending this!  I wanted to give some context on why the CPU period is hard-coded to 100ms today, and what needs to be considered in changing it.

CPU period and CPU quota are the mechanisms that we employ to power the [task size](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size) CPU hard limits.  When you specify vCPU in your task definition, we translate that into the CPU period and CPU quota settings that apply to the `cpu` cgroup subsystem.

CPU quota controls the amount of CPU time granted to a cgroup during a given CPU period.  Both settings are expressed in terms of microseconds; having a CPU quota that equals CPU period means that a cgroup can execute up to 100% on one vCPU (or 50% on each of two vCPUs, or any other fraction that totals to 100% of one).  The CPU quota has a maximum of 1000000us and CPU period has a minimum of 1ms, so it becomes a math problem to express limits that make sense for a given CPU count and the limits that you want to set.  Changing the CPU period without changing the CPU quota will cause you to have different effective limits than what you've specified in your task definition; these values need to co-vary with the vCPU task size.

The hard-coded 100ms period that we have today lets us express values for vCPUs ranging from 0.125 to 10, and this is reflected in the ECS task definition API validation (i.e., if you attempt to specify greater than 10 or less than 0.125 in your task definition, you'll get an error back).  Changing the quota changes the values that would be valid (from the Linux kernel's perspective) for controlling task size, and we would potentially need to adjust the ECS API to reflect that.

This is recorded in the comments of `agent/api/task/task_linux.go`:

```
	// TODO: DefaultCPUPeriod only permits 10VCPUs.
	// Adaptive calculation of CPUPeriod required for further support
	// (samuelkarp) The largest available EC2 instance in terms of CPU count is a x1.32xlarge,
	// with 128 vCPUs. If we assume a fixed evaluation period of 100ms (100000us),
	// we'd need a quota of 12800000us, which is longer than the maximum of 1000000.
	// For 128 vCPUs, we'd probably need something like a 1ms (1000us - the minimum)
	// evaluation period, an 128000us quota in order to stay within the min/max limits.
```

I don't think we'll be able to take this pull request in its current form, because of how it interacts with the task definition's task size feature.  In order to allow the CPU period to change, we'll need to both adjust the CPU quota in accordance with the task size and adjust the task definition validation logic.

boynux:
@samuelkarp Thanks for your reply, as it is mentioned in the description we are experiencing some issues with the current 100ms default period which seems to be related to the way CFS calculate the used quota. This makes task level quota very inefficient and unusable to some extend for us and most likely other users.
I'm happy to apply changes as you may suggest to make it more relevant to ECS use case. Unfortunately API and task validation settings are not something that I can help with as I believe they are not public code bases.
As an initial measure I can limit the parameter value ranging from 1ms to 100ms (the current limit), and I think that would cover all the valid cpu range from 0.128 - 10, am I correct here?

boynux:
@samuelkarp now it's limited to 1ms to 100ms only 👼 
Let me know what do you think. thank you.

adnxn:
> now it's limited to 1ms to 100ms only 

@boynux: do you mind updating you branch with the latest changes from dev and resolving the conflicts? though i'll double check with our team, the bounded range via the config seems like a reasonable compromise to me.

agent/config/config.go
```diff
@@ -130,6 +130,11 @@ const (
 	//Known cached image names
 	CachedImageNamePauseContainer = "amazon/amazon-ecs-pause:0.1.0"
 	CachedImageNameAgentContainer = "amazon/amazon-ecs-agent:latest"
+
+	// DefaultCPUPeriod is set to 100 ms to set CFS period and quota for task limits
+	defaultCGroupCPUPeriod = 100 * time.Millisecond
+	maximumCGroupCPUPeriod = 100 * time.Millisecond
+	minimumCGroupCPUPeriod = 1 * time.Millisecond
```
adnxn: seems like `minimumCGroupCPUPeriod` has to be at least `8ms` to keep the the calculated task quota value within [ranges considered valid by linux](https://www.kernel.org/doc/Documentation/scheduler/sched-bwc.txt).

adnxn:
@boynux: im closing this PR for now. but if feel free to address the comments and reopen if you'd like us to take a look again. sorry about the delay on our end! 😳

boynux:
@adnxn I changed the code according to your suggestions and opened a new PR  #1941

PR #1658, opened 2018-10-29
Fix tagging related things

haikuoliu:
### Summary

1. Don't pass tags with RCI call when empty
2. Remove tagging in README and CHANGALOG
3. Add missing change log for pid/ipc
4. Add a test for the case where tags are empty
5. Skip tagging func test

### Testing

### Licensing


agent/api/ecsclient/client.go
```diff
@@ -157,7 +157,9 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
 	// Add additional attributes such as the os type
 	registrationAttributes = append(registrationAttributes, client.getAdditionalAttributes()...)
 	registerRequest.Attributes = registrationAttributes
-	registerRequest.Tags = tags
+	if tagsEnabled && len(tags) > 0 {
```
adnxn: i don't think we need the `tagsEnabled` parameter. we can add an else block here for setting this field, lets try to keep the interfaces smaller if we can help it.

haikuoliu: Just removed the parameter

haikuoliu:
The failed func tests are v3 endpoint related tests, which are already fixed in #1660, and `TestAgentIntrospectionValidator` which is another known flaky test. Going to merge this PR.

PR #1660, opened 2018-10-30
Fix flaky v3 endpoint tests

haikuoliu:
### Summary

Modify logging for v3 endpoint tests. Previously we use `seelog`, but we didn't call `flush` after using seelog. Sometimes even flush won't work because in the test we just call `os.Exit()` to exit and the logs don't have enough time to be flushed. I ended up just using `fmt` which is a simpler solution.

The tests fail on prod but not fail on my local is because the revision of the task definition is 2 in prod but our code always checks the revision should be 1. To prevent this from happening, I will just remove the revision check.

### Testing

### Licensing

### DIFF
--- a/misc/v3-task-endpoint-validator/v3-task-endpoint-validator.go
+++ b/misc/v3-task-endpoint-validator/v3-task-endpoint-validator.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"time"
 
-	"github.com/cihub/seelog"
 	docker "github.com/fsouza/go-dockerclient"
 )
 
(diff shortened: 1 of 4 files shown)


misc/v3-task-endpoint-validator/v3-task-endpoint-validator.go
```diff
@@ -181,7 +180,6 @@ func verifyTaskMetadataResponse(taskMetadataRawMsg json.RawMessage) error {
 
 	taskExpectedFieldEqualMap := map[string]interface{}{
 		"Cluster":       "ecs-functional-tests",
-		"Revision":      "1",
```
fenxiong: i think we can move this one to non empty check instead of removing it?

haikuoliu:
The two functional tests are both `TestContainerInstanceTags`, they are failing because the backend change not in prod. They will be fixed in #1658. So I am going to merge this PR.

PR #1670, opened 2018-11-01
Get region from AWS SDK if NoIID = true

yunhee-l:
### Summary
if NoIID is set to true, we get region from AWS SDK ec2 metadata service.

### Implementation details
we already had EC2 MetadataService client, added Region() method wrapper.

### Testing

### Licensing

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -192,8 +192,17 @@ func NewConfig(ec2client ec2.EC2MetadataClient) (*Config, error) {
 	config.Merge(userDataConfig(ec2client))
 
 	if config.AWSRegion == "" {
-		// Get it from metadata only if we need to (network io)
-		config.Merge(ec2MetadataConfig(ec2client))
+		if config.NoIID {
+			// get it from AWS SDK if we don't have instance identity document
+			awsRegion, err := ec2client.Region()
+			if err != nil {
+				errs = append(errs, err)
+			}
+			config.AWSRegion = awsRegion
+		} else {
+			// Get it from metadata only if we need to (network io)
+			config.Merge(ec2MetadataConfig(ec2client))
+		}
 	}
 
 	return config, config.mergeDefaultConfig(errs)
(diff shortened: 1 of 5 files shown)


agent/config/config.go
```diff
@@ -192,8 +192,17 @@ func NewConfig(ec2client ec2.EC2MetadataClient) (*Config, error) {
 	config.Merge(userDataConfig(ec2client))
 
 	if config.AWSRegion == "" {
-		// Get it from metadata only if we need to (network io)
-		config.Merge(ec2MetadataConfig(ec2client))
+		if config.NoIID {
+			// get it from AWS SDK if we don't have instance identity document
+			awsRegion, err := ec2client.Region()
```
petderek: can we just do this all the time?

yunhee-l: we could, but I don't know if theres any reason behind why we haven't done this for all cases already.

agent/config/config_test.go
```diff
@@ -67,6 +67,25 @@ func TestBrokenEC2MetadataEndpoint(t *testing.T) {
 	assert.Zero(t, config.APIEndpoint, "Endpoint env variable not set; endpoint should be blank")
 }
 
+func TestGetRegionWithNoIID(t *testing.T) {
+	defer setTestEnv("AWS_DEFAULT_REGION", "")()
```
linkar-ec2: Can we change this to be setTestRegion() which sets region to us-west-2? Just to make this consistent with other tests.

yunhee-l: I changed it to use us-east-1 so that it is distinctively different from other tests, which all uses setTestRegion() convenience method that sets the region to use-west-2.

linkar-ec2: But since this is a defer, you are resetting the TestEnv var. Shouldn't it reset TestEnv in the same manner as other tests, regardless of how it is used in this one specific test?

yunhee-l: it sets the env var immediately and defers the unset call (see [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config_test.go#L623)).
What value we are using in the test shouldn't matter, as long as we unset it between tests and make sure one test's state is leaked into the subsequent tests.

PR #1690, opened 2018-11-13
Add Integration Tests For GPU

ubhattacharjya:
### Summary

This change adds integration tests to check if there is a Task with GPU Type association, then the containers requiring GPU has correct Environment variable attached to it. It also checks if the Docker HostConfig Runtime is being set as "nvidia" in that case.

### Testing

New tests cover the changes: 

### Licensing


agent/engine/engine_unix_integ_test.go
```diff
@@ -1331,3 +1369,41 @@ func TestTaskLevelVolume(t *testing.T) {
 	client := taskEngine.(*DockerTaskEngine).client
 	client.RemoveVolume(context.TODO(), "TestTaskLevelVolume", 5*time.Second)
 }
+
+func TestGPUAssociationTask(t *testing.T) {
+	if os.Getenv("ECS_ENABLE_GPU_SUPPORT") == "false" {
```
fenxiong: i think we should probably skip the task if ECS_ENABLE_GPU_SUPPORT is not explicitly set to true

agent/engine/engine_unix_integ_test.go
```diff
@@ -1331,3 +1369,40 @@ func TestTaskLevelVolume(t *testing.T) {
 	client := taskEngine.(*DockerTaskEngine).client
 	client.RemoveVolume(context.TODO(), "TestTaskLevelVolume", 5*time.Second)
 }
+
+func TestGPUAssociationTask(t *testing.T) {
+	gpuSupportEnabled := utils.ParseBool(os.Getenv("ECS_ENABLE_GPU_SUPPORT"), false)
```
petderek: nit: I'd rather use a specific env var for testing instead of reusing this one. The reason is that this could convey that this test will get run whenever agent is configured for GPU support.

agent/engine/engine_unix_integ_test.go
```diff
@@ -1331,3 +1369,40 @@ func TestTaskLevelVolume(t *testing.T) {
...
+	containerMap, _ := taskEngine.(*DockerTaskEngine).state.ContainerMapByArn(testTask.Arn)
+	cid := containerMap[testTask.Containers[0].Name].DockerID
+	state, _ := client.ContainerInspect(ctx, cid)
+	assert.Equal(t, apitask.NvidiaRuntime, state.HostConfig.Runtime)
+	assert.Contains(t, state.Config.Env, "NVIDIA_VISIBLE_DEVICES=1")
+
+	taskUpdate := *testTask
+	taskUpdate.SetDesiredStatus(apitaskstatus.TaskStopped)
+	go taskEngine.AddTask(&taskUpdate)
+	verifyTaskIsStopped(stateChangeEvents, testTask)
```
petderek: Can we go a layer deeper?  Is there a command we can run within the container to validate that the gpu is present?

agent/engine/engine_unix_integ_test.go
ubhattacharjya: Fixed this problem by adding a check if the testing instance is p2 or p3 and if the /var/lib/ecs/ecs.config file has ECS_ENABLE_GPU_SUPPORT is set to true or not.

agent/engine/engine_unix_integ_test.go
ubhattacharjya: Will definitely add this scenario in functional tests.

agent/engine/engine_unix_integ_test.go
```diff
@@ -70,6 +70,12 @@ const (
 	testIPCNamespaceCommand          = "if [ `ipcs -s | awk '$1+0 == 5' | wc -l` -gt 0 ]; then exit 1; else exit 2; fi;"
 	testIPCNamespaceResourceFound    = 1
 	testIPCNamespaceResourceNotFound = 2
+
+	testGPUImage	      = "nvidia/cuda:9.0-base"
+	testGPUContainerName  = "testGPUContainer"
+	testGPUP2InstanceType = "p2"
```
sharanyad: Can you instead make this an array of strings so it's easier to add more instances in the future?

ubhattacharjya: Fixed it.

agent/engine/engine_unix_integ_test.go
```diff
@@ -1331,3 +1394,55 @@ func TestTaskLevelVolume(t *testing.T) {
 	client := taskEngine.(*DockerTaskEngine).client
 	client.RemoveVolume(context.TODO(), "TestTaskLevelVolume", 5*time.Second)
 }
+
+func TestGPUAssociationTask(t *testing.T) {
+	gpuSupportEnabled := utils.ParseBool(getGPUEnvVar(gpuConfigFilePath)["ECS_ENABLE_GPU_SUPPORT"], false)
```
petderek: Grabbing this is the better approach here! I'd rather use the existing code to do it, though:

```golang
metadataClient := ec2.NewEC2MetadataClient(nil)
iid, _ := metadataClient.InstanceIdentityDocument()
// NewConfig already has the logic to parse the ecs.config file
fileCfg, err := config.NewConfig(metadataClient)
if err != nil {
    t.Skip("Couldn't load gpu config")
}
gpuSupportEnabled := fileCfg.GPUSupportEnabled
```

ubhattacharjya: ecs_config.go from the path /var/lib/ecs/ecs.config.go is getting read in init and then agent is started with environment variable ECS_ENABLE_GPU_SUPPORT= true/false . So that is why NewConfig wont be able to read from that path in integration test if we use NewConfig to set the gpuSupportEnabled.

agent/engine/engine_unix_integ_test.go
```diff
@@ -1331,3 +1394,55 @@ func TestTaskLevelVolume(t *testing.T) {
...
+
+	iid, _  := ec2.NewEC2MetadataClient(nil).InstanceIdentityDocument()
+
+	testGPUInstanceType := []string{"p2", "p3"}
```
sharanyad: This could be a constant, so we can reuse in functional tests

agent/engine/engine_unix_integ_test.go
```diff
@@ -1331,3 +1394,55 @@ func TestTaskLevelVolume(t *testing.T) {
...
+
+	for _, gpuInstanceType := range testGPUInstanceType {
+		if strings.HasPrefix(iid.InstanceType, gpuInstanceType) {
+			isGPUInstanceType = true
+			break
+		}
+	}
+
+	if !(isGPUInstanceType && gpuSupportEnabled) {
+		t.Skip("Skipped because either the ECS_ENABLE_GPU_SUPPORT is not set to true or the instance type is not a supported GPU instance Type")
```
sharanyad: nit: `either ECS_ENABLE_GPU_SUPPORT`

PR #1696, opened 2018-11-16
images include in user's imagesNotDeleteList won't be deleted automatically

cyastella:
### Summary

Users want to keep their own imagesNotDeleteList for future use. If image include in the list, these images will not be deleted automatically when the task finishes. 
### Implementation details

Create env called ECS_IMAGES_NOT_DELETE, after we parse it and agent can get  the users own the list of images that they want to keep for their  future use and will not be automatically cleaned up when task is finished.
### Testing

New tests cover the changes: 

### Licensing


agent/config/config_test.go
```diff
@@ -18,13 +18,12 @@ package config
...
 	"encoding/json"
 	"errors"
-	"os"
-	"testing"
-	"time"
-
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/ec2"
 	"github.com/aws/amazon-ecs-agent/agent/ec2/mocks"
+	"os"
```
samuelkarp: Can you undo this import change?

agent/config/config.go
```diff
@@ -474,6 +474,7 @@ func environmentConfig() (Config, error) {
 		ImageCleanupInterval:               parseEnvVariableDuration("ECS_IMAGE_CLEANUP_INTERVAL"),
 		NumImagesToDeletePerCycle:          parseNumImagesToDeletePerCycle(),
 		ImagePullBehavior:                  parseImagePullBehavior(),
+		ImageNotDeleteList:                 parseImageNotDeleteList("ECS_IMAGES_NOT_DELETE"),
```
samuelkarp: What about `ECS_IMAGE_CLEANUP_EXCLUDE` and `ImageCleanupExclusionList`?  I think that's more clear.

cyastella: yeah, will change it.

agent/config/config_test.go
haikuoliu: You can configure or run `goimport`

agent/engine/docker_image_manager.go
```diff
@@ -214,21 +216,42 @@ func (imageManager *dockerImageManager) removeImageState(imageStateToBeRemoved *
...
 	}
 	var imagesForDeletion []*image.ImageState
-	for _, imageState := range imageManager.imageStatesConsideredForDeletion {
+	for _, imageState := range candidateImagesConsiderForDeletion {
 		if imageManager.isImageOldEnough(imageState) && imageState.HasNoAssociatedContainers() {
 			seelog.Infof("Candidate image for deletion: [%s]", imageState.String())
+			// remove the images that cannot be added
+			if imageManager.includesImageCleanupExclusionList(imageState) {
+				//imageState that we want to keep
+				seelog.Infof("image excluded from deletion: [%s]", imageState.String())
```
haikuoliu: nit: `Image excluded`

And for the excluded images, I prefer to not print `Candidate image for deletion:`, it's kind of confusing

agent/engine/docker_image_manager.go
```diff
@@ -214,21 +216,42 @@ func (imageManager *dockerImageManager) removeImageState(imageStateToBeRemoved *
 }
 
 func (imageManager *dockerImageManager) getCandidateImagesForDeletion() []*image.ImageState {
-	if len(imageManager.imageStatesConsideredForDeletion) < 1 {
+	return imageManager.getCandidateImagesForDeletionHelper(imageManager.imageStatesConsideredForDeletion)
+}
+
+func (imageManager *dockerImageManager) getCandidateImagesForDeletionHelper(candidateImagesConsiderForDeletion map[string]*image.ImageState) []*image.ImageState {
```
haikuoliu: Why using a helper? The only thing that `getCandidateImagesForDeletion` does is to call `getCandidateImagesForDeletionHelper`

agent/engine/docker_image_manager.go
```diff
@@ -214,21 +216,42 @@ func (imageManager *dockerImageManager) removeImageState(imageStateToBeRemoved *
...
+				seelog.Infof("image excluded from deletion: [%s]", imageState.String())
+				continue
+			}
 			imagesForDeletion = append(imagesForDeletion, imageState)
 		}
 	}
 	return imagesForDeletion
 }
 
+func (ImageManager *dockerImageManager) includesImageCleanupExclusionList(imageState *image.ImageState) bool {
```
haikuoliu: Maybe better to call it `includedInImageCleanupExclusionList`?

agent/engine/docker_image_manager.go
```diff
@@ -214,21 +216,42 @@ func (imageManager *dockerImageManager) removeImageState(imageStateToBeRemoved *
...
 	var imagesForDeletion []*image.ImageState
-	for _, imageState := range imageManager.imageStatesConsideredForDeletion {
+	for _, imageState := range candidateImagesConsiderForDeletion {
 		if imageManager.isImageOldEnough(imageState) && imageState.HasNoAssociatedContainers() {
 			seelog.Infof("Candidate image for deletion: [%s]", imageState.String())
+			// remove the images that cannot be added
+			if imageManager.includesImageCleanupExclusionList(imageState) {
+				//imageState that we want to keep
+				seelog.Infof("image excluded from deletion: [%s]", imageState.String())
+				continue
```
haikuoliu: Optional: personally I prefer to use a `if else` because we either append it to the list or not append it to the list

agent/engine/docker_image_manager.go
```diff
@@ -214,21 +216,42 @@ func (imageManager *dockerImageManager) removeImageState(imageStateToBeRemoved *
...
+			}
 			imagesForDeletion = append(imagesForDeletion, imageState)
 		}
 	}
 	return imagesForDeletion
 }
 
+func (ImageManager *dockerImageManager) includesImageCleanupExclusionList(imageState *image.ImageState) bool {
+	for _, imageNotDelete := range ImageManager.imageCleanupExclusionList {
```
haikuoliu: Does it make more sense if you use a set (or create a set and use it everywhere) for `imageCleanupExclusionList` instead of a list? That way you just need to do one for loop here

agent/config/parse.go
```diff
@@ -280,3 +280,14 @@ func parseEnvVariableDuration(envVar string) time.Duration {
 	}
 	return duration
 }
+
+func parseImageCleanupExclusionList(envVar string) []string {
+	ImageEnv := os.Getenv(envVar)
```
sharanyad: this need not be a public variable. same below

agent/engine/docker_image_manager.go
```diff
@@ -214,21 +216,42 @@ func (imageManager *dockerImageManager) removeImageState(imageStateToBeRemoved *
...
 		// no image states present in image manager
 		return nil
 	}
 	var imagesForDeletion []*image.ImageState
-	for _, imageState := range imageManager.imageStatesConsideredForDeletion {
+	for _, imageState := range candidateImagesConsiderForDeletion {
 		if imageManager.isImageOldEnough(imageState) && imageState.HasNoAssociatedContainers() {
 			seelog.Infof("Candidate image for deletion: [%s]", imageState.String())
+			// remove the images that cannot be added
+			if imageManager.includesImageCleanupExclusionList(imageState) {
```
sharanyad: why don't we do this check before itself so we can avoid unnecessary criteria check for deletion for excluded images?

Issue #1596, opened 2018-10-04
Flaky test: TestAgentIntrospectionValidator


haikuoliu:
### Summary

It failed in my PR twice, but when I tried 3rd time it passed.

### Expected Behavior
Test pass

### Observed Behavior
Test fail

### Environment Details
Linux

### Supporting Log Snippets

Log:
```
--- FAIL: TestAgentIntrospectionValidator (21.87s)
	utils_unix.go:120: Created directory /tmp/tmp.npe4LKjcQI/ecs_integ_testdata698787712 to store test data in
	utils_unix.go:134: Launching agent with image: amazon/amazon-ecs-agent:latest
	utils_unix.go:218: Agent started as docker container: df93ec5130e35d416c6c34dd0338c6092a40e0eb0e715dd7756a863c84616c29
	utils.go:156: Found agent metadata: {Cluster:ecs-functional-tests ContainerInstanceArn:0xc420448a30 Version:Amazon ECS Agent - v1.20.3 (f8b6960b)}
	utils.go:177: Task definition: ecsftest-agent-introspection-validator-4c0f2e0d3d9e85fd9e5ca7fdf73a3628:1
	utils.go:197: Started task: arn:aws:ecs:us-west-2:535959970326:task/f450f364-3430-4e89-87f4-a38e9cba5c40
	functionaltests_unix_test.go:658: 
			Error Trace:	functionaltests_unix_test.go:658
			Error:      	Not equal: 
			            	expected: 42
			            	actual  : 1
			Test:       	TestAgentIntrospectionValidator
			Messages:   	Expected exit code of 42; got 1
	utils.go:164: Preserving test dir for failed test /tmp/tmp.npe4LKjcQI/ecs_integ_testdata698787712
```
from CW log:
```
Unable to get tasks metadata for 'http://localhost:51678/v1/tasks': tasks metadata: unable to verify response: field Volumes should not be empty
```

fenxiong:
Fixed in #1751, doesn't seem to be flakey anymore. closing

PR #1531, opened 2018-08-20
handlers: Add Volumes to v1 and v2 metadata

sharanyad:
### Summary
Expose volumes metadata as part of each container in task through v1 and v2 metadata endpoint. 
This addresses https://github.com/aws/amazon-ecs-agent/issues/1516 

### Implementation details

new task metadata(from functional tests): 
```
[Info] Received tasks metadata:
{
    "Tasks": [
        {
            "Arn": "arn:aws:ecs:us-west-2:074626758757:task/0353708a-6653-4bc2-9e84-ad2dcc8fe68c",
            "DesiredStatus": "RUNNING",
            "KnownStatus": "RUNNING",
            "Family": "ecsftest-agent-introspection-validator-4c0f2e0d3d9e85fd9e5ca7fdf73a3628",
            "Version": "1",
            "Containers": [
                {
                    "DockerId": "9c562697bcbeab2ce97491080d78dbeb38e64d21203417895edf1b5ce203ba24",
                    "DockerName": "ecs-ecsftest-agent-introspection-validator-4c0f2e0d3d9e85fd9e5ca7fdf73a3628-1-agent-introspection-validator-be84a5fd9680eac74b00",
                    "Name": "agent-introspection-validator",
                    "Volumes": [
                        {
                            "DockerName": "ecs-ecsftest-agent-introspection-validator-4c0f2e0d3d9e85fd9e5ca7fdf73a3628-1-task-local-bcb19ac3cc93fe8f8901",
                            "Source": "/var/lib/docker/volumes/ecs-ecsftest-agent-introspection-validator-4c0f2e0d3d9e85fd9e5ca7fdf73a3628-1-task-local-bcb19ac3cc93fe8f8901/_data",
                            "Destination": "/ecs"
                        }
                    ]
                }
            ]
        }
    ]
}
```

### Testing

- [x] Builds on Linux (`make release`)
- [x] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [x] Unit tests on Linux (`make test`) pass
- [x] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [x] Integration tests on Linux (`make run-integ-tests`) pass
- [x] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [x] Functional tests on Linux (`make run-functional-tests`) pass
- [x] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: yes

### Licensing

[...]

PR #1718, opened 2018-12-04
make TestAgentIntrospectionValidator less flakey

fenxiong:
### Summary

TestAgentIntrospectionValidator is a [known flakey test](https://github.com/aws/amazon-ecs-agent/issues/1596), and it seems that it's failing because the volumes field could be unavailable when a container just starts. This is because the volumes field is only populated after we start the container, inspect it and get the volumes from inspectContainer response (see https://github.com/aws/amazon-ecs-agent/pull/1531). Let the validator sleep for a while to make it less flakey, considering the fact that TestTaskMetadataValidator should suffer from the same issue but it's not identified as a flakey test because it sleeps for a while before checking (https://github.com/aws/amazon-ecs-agent/blob/master/misc/taskmetadata-validator/taskmetadata-validator.go#L214).

Also fix the task def for TestAgentIntrospectionValidator to use local test registry image. This mostly won't affect the test but will cause agent to keep trying to pull the image "amazon/amazon-ecs-agent-introspection-validator:make" because it won't be able to pull it.

### Implementation details

See above.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/agent-introspection-validator/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/agent-introspection-validator/task-definition.json
@@ -2,7 +2,7 @@
   "family": "ecsftest-agent-introspection-validator",
   "networkMode": "host",
   "containerDefinitions": [{
-    "image": "amazon/amazon-ecs-agent-introspection-validator:make",
+    "image": "127.0.0.1:51670/amazon/amazon-ecs-agent-introspection-validator:latest",
     "name": "agent-introspection-validator",
     "memory": 50,
     "logConfiguration": {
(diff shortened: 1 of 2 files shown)


sharanyad:
Could you please run this particular test with count of 1000 and check if it fails?

fenxiong:
@sharanyad ran it 1000 times:
```
GOCACHE=off go test -run TestAgentIntrospectionValidator -tags functional -count 500 -timeout 10h ./agent/functional_tests/tests/
ok      github.com/aws/amazon-ecs-agent/agent/functional_tests/tests    5924.191s

GOCACHE=off go test -run TestAgentIntrospectionValidator -tags functional -count 500 -timeout 10h ./agent/functional_tests/tests/
ok      github.com/aws/amazon-ecs-agent/agent/functional_tests/tests    5907.818s
```

PR #1721, opened 2018-12-05
ECR image pull retry to 5

shubham2892:
### Summary
Reducing the number of retry calls made by agent to pull a docker image. 

Docker has it's own getLayer retry in the daemon. Including that in the worst case, agent will make 10 * 5 * nLayers calls(nLayers is the number of layers in the image). This change will bring down the # of calls to 5 * 5 * nLayers.  

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -99,7 +99,7 @@ const (
 	StatsInactivityTimeout = 5 * time.Second
 
 	// retry settings for pulling images
-	maximumPullRetries        = 10
+	maximumPullRetries        = 5
 	minimumPullRetryDelay     = 250 * time.Millisecond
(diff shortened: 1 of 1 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -99,7 +99,7 @@ const (
 	StatsInactivityTimeout = 5 * time.Second
 
 	// retry settings for pulling images
-	maximumPullRetries        = 10
+	maximumPullRetries        = 5
```
sharanyad: Does it have any value to make this configurable?

shubham2892: As discussed offline, currently we don't see value in making it configurable but can be done in future if users see value in it.

PR #1726, opened 2018-12-06
set nvidia container runtime name

sharanyad:
### Summary
ECS will use the default runtime name "ecs-nvidia"
Making the runtime name configurable using environment variable `ECS_NVIDIA_RUNTIME` 

### Testing

New tests cover the changes: 

N/A

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,8 @@ additional details on each available environment variable.
 | `ECS_SHARED_VOLUME_MATCH_FULL_CONFIG` | `true` | When `true`, ECS Agent will compare name, driver options, and labels to make sure volumes are identical. When `false`, Agent will short circuit shared volume comparison if the names match. This is the default Docker behavior. If a volume is shared across instances, this should be set to `false`. | `false` | `false`|
 | `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS_FROM` | `ec2_instance` | If `ec2_instance` is specified, existing tags defined on the container instance will be registered to Amazon ECS and will be discoverable using the `ListTagsForResource` API. Using this requires that the IAM role associated with the container instance have the `ec2:DescribeTags` action allowed. | `none` | `none` |
 | `ECS_CONTAINER_INSTANCE_TAGS` | `{"tag_key": "tag_val"}` | The metadata that you apply to the container instance to help you categorize and organize them. Each tag consists of a key and an optional value, both of which you define. Tag keys can have a maximum character length of 128 characters, and tag values can have a maximum length of 256 characters. If tags also exist on your container instance that are propagated using the `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS_FROM` parameter, those tags will be overwritten by the tags specified using `ECS_CONTAINER_INSTANCE_TAGS`. | `{}` | `{}` |
+| `ECS_NVIDIA_RUNTIME` | ecs-nvidia | The Nvidia Runtime to be used to pass Nvidia GPU devices to containers. | ecs-nvidia | Not Applicable |
+
 
 ### Persistence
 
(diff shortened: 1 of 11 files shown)


agent/api/task/task.go
```diff
@@ -70,7 +70,7 @@ const (
 
 	NvidiaVisibleDevicesEnvVar = "NVIDIA_VISIBLE_DEVICES"
 	GPUAssociationType         = "gpu"
-	NvidiaRuntime              = "nvidia"
+	NvidiaRuntime              = "ecs-nvidia"
```
samuelkarp: You may need to make this configurable if we expect people to be able to use the normal Nvidia runtime on other Linux distributions like Ubuntu or Debian.

sharanyad: updated

sharanyad:
`README` update is pending

README.md
```diff
@@ -134,6 +134,8 @@ additional details on each available environment variable.
...
 | `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS_FROM` | `ec2_instance` | If `ec2_instance` is specified, existing tags defined on the container instance will be registered to Amazon ECS and will be discoverable using the `ListTagsForResource` API. Using this requires that the IAM role associated with the container instance have the `ec2:DescribeTags` action allowed. | `none` | `none` |
 | `ECS_CONTAINER_INSTANCE_TAGS` | `{"tag_key": "tag_val"}` | The metadata that you apply to the container instance to help you categorize and organize them. Each tag consists of a key and an optional value, both of which you define. Tag keys can have a maximum character length of 128 characters, and tag values can have a maximum length of 256 characters. If tags also exist on your container instance that are propagated using the `ECS_CONTAINER_INSTANCE_PROPAGATE_TAGS_FROM` parameter, those tags will be overwritten by the tags specified using `ECS_CONTAINER_INSTANCE_TAGS`. | `{}` | `{}` |
+| `ECS_NVIDIA_RUNTIME` | ecs-nvidia | The Nvidia Runtime to be used to pass Nvidia GPU devices to containers. | ecs-nvidia | Not Applicable |
```
samuelkarp: Is it worth calling this `ECS_GPU_RUNTIME` instead to be slightly more generic?

agent/config/parse.go
```diff
@@ -252,6 +252,14 @@ func parseContainerInstancePropagateTagsFrom() ContainerInstancePropagateTagsFro
 	}
 }
 
+func parseNvidiaRuntime() string {
```
samuelkarp: You shouldn't need this, but instead can set the default value in [here](https://github.com/aws/amazon-ecs-agent/blob/608808bb1a7eceed7e3bb390324c7894a75d2141/agent/config/config_unix.go#L44).

agent/config/config.go
```diff
@@ -489,6 +492,7 @@ func environmentConfig() (Config, error) {
 		ContainerInstanceTags:              containerInstanceTags,
 		ContainerInstancePropagateTagsFrom: parseContainerInstancePropagateTagsFrom(),
 		GPUSupportEnabled:                  utils.ParseBool(os.Getenv("ECS_ENABLE_GPU_SUPPORT"), false),
+		NvidiaRuntime:                      parseNvidiaRuntime(),
```
samuelkarp: You can just use `os.Getenv` here.

agent/config/parse.go
sharanyad: ah, missed this one. will modify.

README.md
samuelkarp: I don't have a strong opinion.  If you prefer to keep it tied to Nvidia, I'm comfortable with that.

sharanyad: left it as is, since gpu support is tied to Nvidia.

PR #1729, opened 2018-12-07
V1.23.0 stage

ubhattacharjya:
### Summary

ECS Agent Staging v1.23.0

### Testing

New tests cover the changes: 

### Licensing


VERSION
```diff
@@ -1 +1 @@
-1.22.0
+1.23.0
```
jhaynes: nit: missing a newline

haikuoliu: For this file, we used to not add new line at the end (i think it's a practice or something?), but it actually doesn't matter because it works fine when missed to remove the new line.

So I think we can start to always add a new line in the following release because it seems more natural, and it's easily to forget to remove the new line, and I think in ecs init VERSION file we have a new line at the end

jhaynes: Either way is ok with me. This was a change from what it was in the prior release; I just wanted to make sure it was intentional and we understood what it might affect.

Issue #1575, opened 2018-09-13
Feature request for container meta data file to include ECS host instance IP address


cloudpal:
### Summary

Customers are using instance meta data inside of the container to get IP address of the host ECS instance. This is a security risk as enabling instance meta data allows for querying IAM credentials as well.  

### Description

Instance meta data are typically disabled due to security risk of being able to query IAM roles/credentials. To avoid using instance meta data, containers use the container specific meta data file. However, currently the ECS host IP field is not present in container metadata file. Request to include host IP address field to the container meta data file.

adnxn:
closing issue, changes have been merged into `dev` with #1730

PR #1730, opened 2018-12-07
Passing Host IPv4 address to container metadata file

linkar-ec2:
### Summary

Adds Host EC2 instance Public IPv4 address to container metadata file.
Customer ask: [GitHub issue](https://github.com/aws/amazon-ecs-agent/issues/1575) 
### Implementation details

Added EC2 api call to describe instance and retrieve public ipv4 address at Agent startup

### Testing

Added unit tests for adding IP address to file. Manually tested as well.

New tests cover the changes: 

Saving Host IPv4 address in container metadata file
### Licensing


agent/ec2/ec2_client.go
```diff
@@ -105,3 +108,21 @@ func (c *ClientImpl) DescribeECSTagsForInstance(instanceID string) ([]*ecs.Tag,
...
+func (c *ClientImpl) GetHostPublicIPv4AddressForInstance(instanceID string) (string, error) {
+	describeInstancesInput := ec2sdk.DescribeInstancesInput{
+		InstanceIds: []*string{aws.String(instanceID)},
+	}
+	describeInstancesOutput, err := c.client.DescribeInstances(&describeInstancesInput)
+	if err != nil {
+		seelog.Criticalf("Error calling DescribeInstances API: %v", err)
+		return "", err
+	}
+	if len(describeInstancesOutput.Reservations) == 0 || len(describeInstancesOutput.Reservations[0].Instances) == 0 {
```
adnxn: `Only 1 address is expected.`, if only 1 is expected. should we explicitly check for 1 for each? am i missing something?

linkar-ec2: Because we only ever pass in a single instance ID, I can't imagine a scenario in which we get more than 1 instance back. This check is to ensure we don't run into a nil pointer exception

agent/app/agent.go
```diff
@@ -660,3 +661,19 @@ func mergeTags(localTags []*ecs.Tag, ec2Tags []*ecs.Tag) []*ecs.Tag {
...
+// getHostPublicIPv4AddressFromEC2API will retrieve the PublicIPAddress (IPv4) of this
+// instance through the EC2 API
+func (agent *ecsAgent) getHostPublicIPv4AddressFromEC2API() string {
+	// Get instance ID from ec2 metadata client.
+	instanceID, err := agent.ec2MetadataClient.InstanceID()
+	if err != nil {
+		return ""
+	}
+
+	hostPublicIPv4Address, err := agent.ec2Client.GetHostPublicIPv4AddressForInstance(instanceID)
```
samuelkarp: You can get this from the EC2 instance metadata, without making a call to the EC2 API (and requiring increased IAM permissions):

```bash
$ curl 169.254.169.254/latest/meta-data/public-ipv4
```

agent/ec2/ec2_client.go
```diff
@@ -105,3 +108,21 @@ func (c *ClientImpl) DescribeECSTagsForInstance(instanceID string) ([]*ecs.Tag,
...
 	return c.client.CreateTags(input)
 }
+
+// GetHostPublicIPv4AddressForInstance calls the DescribeInstances API to get the
+// PublicIPv4 address for a specific instance. Only 1 address is expected.
+func (c *ClientImpl) GetHostPublicIPv4AddressForInstance(instanceID string) (string, error) {
+	describeInstancesInput := ec2sdk.DescribeInstancesInput{
+		InstanceIds: []*string{aws.String(instanceID)},
+	}
+	describeInstancesOutput, err := c.client.DescribeInstances(&describeInstancesInput)
```
samuelkarp: See above; I don't think we should be making calls to DescribeInstances.

agent/app/agent.go
linkar-ec2: In the GitHub issue, the request was to put it in the metadata file as an alternative to using the EC2 instance metadata. I think the use case being when the EC2 instance metadata service is disabled, they'd still like to see the HostIP 

> "This is a security risk as enabling instance meta data allows for querying IAM credentials as well."

samuelkarp: It's not possible to disable the EC2 instance metadata service on the host, but it is possible to block container access to the IMDS using iptables rules.  The ECS agent has a facility for doing this when tasks use the `awsvpc` network mode with the `ECS_AWSVPC_BLOCK_IMDS` setting.

Note that the ECS agent relies on the IMDS being available for a variety of purposes, so you cannot have a functioning ECS agent without having access to IMDS.  (Your code above even has a dependency on it in this line: `instanceID, err := agent.ec2MetadataClient.InstanceID()`.  You should be able to change this to grab the public IP address instead of the instance ID and skip the rest of the logic you have.)

agent/app/agent.go
```diff
@@ -660,3 +661,15 @@ func mergeTags(localTags []*ecs.Tag, ec2Tags []*ecs.Tag) []*ecs.Tag {
...
 	return utils.MapToTags(tagsMap)
 }
+
+// getHostPublicIPv4AddressFromEC2Metadata will retrieve the PublicIPAddress (IPv4) of this
+// instance through the EC2 API
+func (agent *ecsAgent) getHostPublicIPv4AddressFromEC2Metadata() string {
+	// Get instance ID from ec2 metadata client.
+	hostPublicIPv4Address, err := agent.ec2MetadataClient.PublicIPv4Address()
+	if err != nil {
+		seelog.Criticalf("Unable to retrieve Host Instance PublicIPv4 Address: %v", err)
```
samuelkarp: I'd recommend having this as Error, not Critical.

linkar-ec2: Fixed!

linkar-ec2:
@sharanyad We aren't saving the Host IP to the state file, just metadata manager. I don't believe we need to bump it.

agent/functional_tests/testdata/taskdefinitions/container-metadata-file-validator-windows/task-definition.json
```diff
@@ -0,0 +1,27 @@
+{
+    "family": "ecsftest-container-metadata-file-validator-windows",
+    "containerDefinitions": [{
+      "image": "amazon/amazon-ecs-container-metadata-file-validator-windows",
+      "name": "container-metadata-file-validator",
+      "memory": 512,
+      "cpu": 1024,
+      "healthCheck": {
```
sharanyad: Do we need health check for this test?

misc/container-metadata-file-validator-windows/container-metadata-file-validator-windows.dockerfile
```diff
@@ -0,0 +1,16 @@
...
+# License is located at
+#
+#	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+FROM microsoft/windowsservercore:latest
```
sharanyad: Can we use microsoft/nanoserver for this?

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -635,6 +635,58 @@ func testV3TaskEndpoint(t *testing.T, taskName, containerName, networkMode, awsl
...
+	agentOptions := &AgentOptions{
+		ExtraEnvironment: map[string]string{
+			"ECS_ENABLE_CONTAINER_METADATA": "true",
+			"ECS_AVAILABLE_LOGGING_DRIVERS": `["awslogs"]`,
+		},
+	}
+
+	agent := RunAgent(t, agentOptions)
+	defer agent.Cleanup()
+	agent.RequireVersion(">=1.15.0")
```
sharanyad: since this test also checks for HostIPv4 field, shouldn't the agent version required be the most recent one?

misc/container-metadata-file-validator/container-metadata-file-validator.go
```diff
@@ -0,0 +1,225 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sharanyad: looks like this file is essentially the same as `misc/container-metadata-file-validator-windows/container-metadata-file-validator-windows.go` . is it possible to reuse?

linkar-ec2: There is different validation for network on linux, which is more detailed. The 2 files approach is taken from the v3-task-metadata-validator. The files are built/managed differently. In addition, the task definitions themselves are a bit different (with windows including "windows" identifiers)

agent/functional_tests/tests/functionaltests_test.go
linkar-ec2: Good catch. I think we should move away from requiring a version and represent these "requirements" as capabilities. For functional tests, all tests should run against the most up to date version of Agent.

misc/container-metadata-file-validator-windows/container-metadata-file-validator-windows.dockerfile
linkar-ec2: Yes! Decreases the image size from 11gb to 1.1gb!

agent/functional_tests/testdata/taskdefinitions/container-metadata-file-validator-windows/task-definition.json
linkar-ec2: Other tests that run a custom image use the healthCheck (like v3-task-endpoint-validator). Since this test runs for ~15 seconds, doesn't hurt to have a healthCheck every 5 secs.

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -636,6 +636,57 @@ func testV3TaskEndpoint(t *testing.T, taskName, containerName, networkMode, awsl
...
+// fields
+func testContainerMetadataFile(t *testing.T, taskName, awslogsPrefix string) {
+	agentOptions := &AgentOptions{
+		ExtraEnvironment: map[string]string{
+			"ECS_ENABLE_CONTAINER_METADATA": "true",
+			"ECS_AVAILABLE_LOGGING_DRIVERS": `["awslogs"]`,
+		},
+	}
+
+	agent := RunAgent(t, agentOptions)
```
sharanyad: I think we would still require the right agent version here, since functional tests can be run for different agent images.

agent/app/agent_test.go
```diff
@@ -1109,6 +1115,40 @@ func TestGetContainerInstanceTagsFromEC2APIFailToDescribeECSTagsForInstance(t *t
...
+	ec2Client := mock_ec2.NewMockClient(ctrl)
+
+	agent := &ecsAgent{
+		ec2MetadataClient: ec2MetadataClient,
+		ec2Client:         ec2Client,
+	}
+	ec2MetadataClient.EXPECT().PublicIPv4Address().Return(hostPublicIPv4Address, nil)
+
+	hostPublicIPv4AddressOutput := agent.getHostPublicIPv4AddressFromEC2Metadata()
+	assert.Equal(t, hostPublicIPv4Address, hostPublicIPv4AddressOutput)
```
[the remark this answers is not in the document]
linkar-ec2: Changed, along with the test below.

agent/functional_tests/tests/functionaltests_test.go
linkar-ec2: Gotcha. Fixed it to remain consistent with other tests.

PR #1747, opened 2018-12-17
Arm build script

shubham2892:
Arm build script

### Testing

New tests cover the changes: 

### Licensing


yunhee-l:
can you merge the two commits?

scripts/ec2_stage_arm.sh
```diff
@@ -0,0 +1,188 @@
...
+# License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+DRYRUN=true
+
+AWS_PROFILE=""
+AWS_REGION="us-east-1"
+AMI_ID="ami-053b2a8c2f3e87928" #amzn2-ami-hvm-2.0.20181020.0-aarch64-gp2
```
yumex93: So this is the only AMI_ID? Is there a possibility that this will be updated later or there will be more amis needed?

shubham2892: Ideally we won't need to update this AMI id in future. 

We do plan to move the build process to codebuild when they add the support for Arm. This script would then get redundant.

PR #1748, opened 2018-12-18
Updating a functional test

petderek:
Its possible to get an internal IP address that looks like this:

```
ip-10-0-1-1.ec2.internal
```

... which will completely fill the old regex check.

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/hostname-awsvpc/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/hostname-awsvpc/task-definition.json
@@ -6,7 +6,7 @@
         "name": "exit",
         "image": "127.0.0.1:51670/ubuntu:latest",
         "cpu": 100,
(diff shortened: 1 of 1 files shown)


petderek:
I tried using the github UI to make a pull request, but did it spectacularly wrong. I'll fix this up soon.

The only code that needs review is in the `taskdefinition.json` file

agent/functional_tests/testdata/taskdefinitions/hostname-awsvpc/task-definition.json
```diff
@@ -6,7 +6,7 @@
         "name": "exit",
         "image": "127.0.0.1:51670/ubuntu:latest",
         "cpu": 100,
-        "command":  [ "sh", "-c", "hostname | grep \"ip-[0-9]*-[0-9]*-[0-9]*-[0-9]*.*.compute.internal\"; if [ $? -eq 0 ]; then exit 0; else exit 1; fi" ],
+        "command":  [ "sh", "-c", "hostname | grep \"ip-[0-9]*-[0-9]*-[0-9]*-[0-9]*.*.internal\"; if [ $? -eq 0 ]; then exit 0; else exit 1; fi" ],
```
jahkeup: You could use single quotes here to avoid having to escape the double quote.

PR #1646, opened 2018-10-24
Added ability to poll for container stats instead of constant stream

yunhee-l:
### Summary
**Push blocked until next agent release**
Allows for container stat polling instead of forcing constant stream. This allows for a significant decrease in CPU usage when a lot of containers are deployed. Duplicate/fix of https://github.com/aws/amazon-ecs-agent/pull/1475

addresses issue https://github.com/aws/amazon-ecs-agent/issues/1422
### Implementation details
merging PR #1475 and fixing conflicts, and errors in unit tests.

### Testing

- [x] Builds on Linux (`make release`)
- [x] Builds on Windows (`go build -out amazon-ecs-agent.exe ./agent`)
- [x] Unit tests on Linux (`make test`) pass
- [x] Unit tests on Windows (`go test -timeout=25s ./agent/...`) pass
- [x] Integration tests on Linux (`make run-integ-tests`) pass
- [x] Integration tests on Windows (`.\scripts\run-integ-tests.ps1`) pass
- [x] Functional tests on Linux (`make run-functional-tests`) pass
- [x] Functional tests on Windows (`.\scripts\run-functional-tests.ps1`) pass

New tests cover the changes: yes

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1751, opened 2018-12-21
Fixing agent functional tests

fenxiong:
### Summary

Agent functional tests have a large failure rate when running on a large scale, with a variety of instance types and regions. This PR is to fix a couple of problematic functional tests so that we can have a better passing rate.

### Note
Apart from fixing the existing tests, this pr also bumps functional test timeout to 36m which seems to be a bare minimum if we want to have a decent passing rate when running across different instance types and regions. And this doesn't consider the fact that we are adding another telemetry test that takes at least 4 minutes in another pr (https://github.com/aws/amazon-ecs-agent/pull/1646). We will likely need to increase the timeout to 40m, but i'm publishing the pr as 36m first.

Update: now that #1646 is merged, i'm bumping timeout to 40m

### Implementation details

Changes to functional tests are outlined below:
#### TestContainerInstanceTags
Tagging feature requires the account to enable long arn, so what this test was doing before is to enable long arn for the account at the beginning of the test, and disable it at the end. However, this introduces a race condition when multiple test runs happen at the same time, because each of them can be enabling/disabling long arn for the same account which causes the test to fail. I fix it by removing the disabling long arn action at the end, so essentially this change makes all tests to be tested with long arn enabled now.

#### TestTelemetry
TestTelemetry was having two kinds of problem: (1) It requires 2GB of memory so we will fail to start the task on the instance types that don't have that much of memory; (2) It specifies the cpu shares to be 25% of total cpu resources on the instance, however there's a valid range for this parameter and we exceed the range on certain instance types because we always use 25% cpu.
This pr fixes (1) by enforcing a minimum memory requirement (skipping the test if there's not enough memory), and (2) by bounding the value of the cpu parameter.

#### TestParallelPull, TestShmSize, TestTmpfs
Those tests were failing because of memory constraint too. Fixed by enforcing minimum memory requirement.

#### TestAgentIntrospectionValidator

TestAgentIntrospectionValidator fails occasionally because of a known race condition. The volumes field in metadata response is not immediately available to the container when it just starts, but only sometimes after. The test runs a validator inside the container which checks the volumes field in metadata response, and when the volumes field is not available, the test will fail. I was trying to fix it before by letting the validator sleep for 3 seconds before checking (https://github.com/aws/amazon-ecs-agent/pull/1718), but that doesn't seem to be a good approach because we'd always spend 3 seconds on it. This pr removes the dummy sleep and refactors the verification logic into the existing retry loop so we can retry when volumes field is not presented and verification fails, which seems to be a better approach.

#### TestOOMContainer
This test fails occasionally on small instance types where we expect the container to fail with OOM reason, but instead it fails with general failure. The container was trying to consume 1GB of memory, which exceeds the total memory available on small instances. In that case it seems that it's possible for the container to use up all the memory on the instance and fails before it gets killed by OOM killer and gets a OOM failure reason. The pr fixes this by enforcing a minimum memory requirement, and also reduces the memory that the container tries to consume from 1GB to 0.5GB so that we can test this on micro instances.

#### TestOOMTask
This test suffers from same problem that TestOOMContainer has, so it has TestOOMContainer's fix. Additionally, the test sometimes fails where the container didn't get killed and finished successfully instead. This seems to cause by the fact that the container may use some amount of swap memory and didn't exceed memory usage enforced by cgroup, because i observe from the logs that when the container gets killed because of oom, it's already using some of the swap memory:
```
  Dec 20 20:27:50 ip-10-0-105-75 kernel: Task in /ecs/cd954ebc-c18d-4e50-bb87-d2610194b686/            6fde705f6715c5c398a2a714146c80f728cddeb6059190cb95b2bda3b0d76346 killed as a result of limit of /    ecs/cd954ebc-c18d-4e50-bb87-d2610194b686/                                                            6fde705f6715c5c398a2a714146c80f728cddeb6059190cb95b2bda3b0d76346
  Dec 20 20:27:50 ip-10-0-105-75 kernel: memory: usage 10240kB, limit 10240kB, failcnt 1686
  Dec 20 20:27:50 ip-10-0-105-75 kernel: memory+swap: usage 20376kB, limit 20480kB, failcnt 9
  Dec 20 20:27:50 ip-10-0-105-75 kernel: kmem: usage 840kB, limit 9007199254740988kB, failcnt 0
```
This pr fixes this by reducing the memory value in the task definition for this test, which seems to make it more stable now.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/oom-task/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/oom-task/task-definition.json
@@ -1,12 +1,12 @@
 {
   "family": "ecsftest-oom-task",
-  "memory": "64",
+  "memory": "10",
   "containerDefinitions": [{
     "essential": true,
     "name": "error",
     "cpu": 1,
     "image": "127.0.0.1:51670/python:2",
-    "command": ["python", "-c", "foo=' '*1024*1024*1024; import time; time.sleep(10)"]
+    "command": ["python", "-c", "foo=' '*1024*1024*512; import time; time.sleep(10)"]
   }]
 }
 
--- a/agent/functional_tests/util/utils.go
+++ b/agent/functional_tests/util/utils.go
@@ -43,14 +43,15 @@ import (
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/docker/docker/api/types"
 	docker "github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/system"
 	"github.com/docker/go-connections/nat"
-
 	"github.com/pkg/errors"
 )
 
 const (
 	arnResourceSections  = 2
 	arnResourceDelimiter = "/"
+	bytePerMegabyte      = 1024 * 1024
 )
 
 // GetTaskDefinition is a helper that provies the family:revision for the named
@@ -617,6 +618,18 @@ func RequireDockerVersion(t *testing.T, selector string) {
 
 	if !match {
 		t.Skipf("Skipping test; requires %v, but version is %v", selector, dockerVersion)
+	}
+}
+
+func RequireMinimumMemory(t *testing.T, minimumMemoryInMegaBytes int) {
+	memInfo, err := system.ReadMemInfo()
+	if err != nil {
+		t.Fatalf("Could not check system memory info before checking minimum memory requirement: %v", err)
+	}
+
+	totalMemory := int(memInfo.MemTotal / bytePerMegabyte)
+	if totalMemory < minimumMemoryInMegaBytes {
+		t.Skipf("Skipping the test since it requires %d MB of memory. Total memory on the instance: %d MB", minimumMemoryInMegaBytes, totalMemory)
 	}
 }
 
--- a/misc/agent-introspection-validator/agent-introspection-validator.go
+++ b/misc/agent-introspection-validator/agent-introspection-validator.go
@@ -86,11 +86,6 @@ func getTasksMetadata(client *http.Client, path string) (*TasksResponse, error) 
 
 	fmt.Printf("Received tasks metadata: %s \n", string(body))
 
-	err = verifyTasksMetadata(body)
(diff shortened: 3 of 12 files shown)


agent/functional_tests/testdata/taskdefinitions/oom-task/task-definition.json
```diff
@@ -1,12 +1,12 @@
 {
   "family": "ecsftest-oom-task",
-  "memory": "64",
+  "memory": "10",
```
yumex93: Why decreasing memory can help test pass? Is it because a lot of tests failure are caused by OOM?

misc/agent-introspection-validator/agent-introspection-validator.go
```diff
@@ -252,6 +242,23 @@ func metadataResponseOnce(client *http.Client, endpoint string, respType string)
 		return nil, fmt.Errorf("%s: unable to read response body: %v", respType, err)
 	}
 
+	fmt.Printf("Received %s metadata response: %s \n", respType, body)
+
+	// verify metadata response
+	if respType == tasksMetadataRespType {
+		err = verifyTasksMetadata(body)
+		if err != nil {
```
yumex93: Why we remove sleep to wait for volume code blob and add this part? Are they related?

agent/functional_tests/util/utils.go
```diff
@@ -620,6 +621,18 @@ func RequireDockerVersion(t *testing.T, selector string) {
 	}
 }
 
+func RequireMinimumMemory(t *testing.T, minimumMemoryInMegaBytes int) {
+	memInfo, err := system.ReadMemInfo()
+	if err != nil {
```
petderek: nit: I'd do require.NoError  and require.True

fenxiong:
Failing test TestV3TaskEndpointTags is unrelated to the pr. will merge this to unblock other changes and fix it separately.

PR #1754, opened 2018-12-22
Fetch AMI ID dynamically

shubham2892:
### Summary
Currently AMI Id used for building Arm artifact is hardcoded in the build script.  This AMI ID may change and is region dependent. 
 
This change fetches the Arm AMI Id dynamically such that the build script can be run from any account and region. 

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/scripts/ec2_stage_arm.sh
+++ b/scripts/ec2_stage_arm.sh
@@ -19,7 +19,6 @@ DRYRUN=true
 
 AWS_PROFILE=""
 AWS_REGION="us-east-1"
-AMI_ID="ami-053b2a8c2f3e87928" #amzn2-ami-hvm-2.0.20181020.0-aarch64-gp2
 ARTIFACT_BUCKET=""
 SOURCE_BUCKET=""
 KEY_NAME=""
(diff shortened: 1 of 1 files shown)


scripts/ec2_stage_arm.sh
```diff
@@ -162,6 +161,23 @@ echo "${userdata}"
 echo "======================================================"
 echo
 
+# Install jq
+add-apt-repository ppa:eugenesan/ppa
+apt-get update
+apt-get install jq
+
+# Fetch the Arm AMI Id
+AMI_ID=$(aws ec2 describe-images \
```
yunhee-l: is this guaranteed to return one result?

shubham2892: This does not guarantee one result but if the account has access to an Arm instance this should almost always return at least one result.

samuelkarp: I would not recommend doing this.  There are very few guarantees from the output here, including which specific account owns the image (a number of accounts are considered to be "amazon" when querying by owner).

The recommended way to retrieve AMI IDs for Amazon Linux is via SSM: https://aws.amazon.com/blogs/compute/query-for-the-latest-amazon-linux-ami-ids-using-aws-systems-manager-parameter-store/

yunhee-l: Can you make a todo note to use SSM when they are available?

yumex93: 1. If there is no ID return, will ec2 run-instances fail and error will be thrown so that people who run the script can know what happened? Just want to double check we are able to debug if error happens.
2. Maybe add a todo about changing to SSM parameter store once it is available?

samuelkarp: I would lean toward leaving this hard-coded until the SSM parameters are available.  The approach here does not give you guaranteed, predictable results.

scripts/ec2_stage_arm.sh
```diff
@@ -162,6 +161,45 @@ echo "${userdata}"
 echo "======================================================"
 echo
 
+# Install jq
```
petderek: Future reference: is We shouldn’t need to do this.

Just do —query and —output text

scripts/ec2_stage_arm.sh
petderek: We have had this pattern break in past situations. I don’t want us to do this.

shubham2892:
Closing the PR in favor of keeping the AMI ID constant.

PR #1762, opened 2018-12-26
gpu: better logging for debugging

sharanyad:
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/gpu/nvidia_gpu_manager_unix.go
+++ b/agent/gpu/nvidia_gpu_manager_unix.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/cihub/seelog"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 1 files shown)


agent/gpu/nvidia_gpu_manager_unix.go
```diff
@@ -78,6 +79,7 @@ func (n *NvidiaGPUManager) Initialize() error {
 		n.SetGPUIDs(gpuIDs)
 		n.SetDevices()
 	}
+	seelog.Info("Config for GPU support is enabled, but GPU information is not found; continuing without it")
```
yunhee-l: wondering if this should be handled as a failure, for example when customer is expecting to use GPU, shouldn't the customer be notified of the misconfiguration in a more explicit way?

sharanyad: This config is not recommended for customers to add. This is a GPU instance level config, but we also let customers override it to true/false. If for some reason, a customer sets it to true on non GPU instance, there is value is not stopping agent and continuing to register as a normal instance. I can change the log level to error.

sharanyad:
flakey windows functional test failure: 
`--- FAIL: TestOOMContainer (20.53s)`

`TestTaskLevelVolume` timed out for windows integ test, which seems unrelated. Will add a note and it could be investigated outside of this PR.

sharanyad:
merging PR since failures are not related to the changes

PR #1789, opened 2019-01-11
Update registry image to make it Arm compatible

shubham2892:
### Summary
- Update the logic to check for running and stoping containers in `setup-test-registry`.
- Update registry image version. This version has comaptibilty with arm and amd architecture.
- Update busybox image version in `ecr-execution-role-upload` to keep it consistent with busybox at other places in the code.
- Replace `runtime.GOOS` with `runtime.GOARCH`. `GOOS` is the running program's operating system target and `GOARCH`  is the running program's architecture target: one of 386, amd64, arm and so on. `GOARCH` is correct usage for the use case.
- Create a separate `executionrole` image for amd and arm.  
`misc/ecr/build.sh` uploads `executionrole` image to ecr and does not update the image if it already exists. 
If this image corresponding to amd, integ test on arm will not update this image and try to use the amd image in arm. Vice versa is also true. So integ tests will fail on arm or amd depending on the image that is present in ecr. Keeping separate image for amd and arm ensures that correct image is used for correct architecture. 

### Testing

Integ and functional tests are passing for Arm and x86 architecture. 

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/functional_tests/tests/functionaltests_unix_test.go
+++ b/agent/functional_tests/tests/functionaltests_unix_test.go
@@ -570,7 +570,8 @@ func TestFluentdTag(t *testing.T) {
 	// tag was added in docker 1.9.0
 	RequireDockerVersion(t, ">=1.9.0")
 
-	if runtime.GOOS == "arm64" {
+	// Skipping the test for arm as they do not have official support for Arm images
+	if runtime.GOARCH == "arm64" {
 		t.Skip()
 	}
 
@@ -584,7 +585,8 @@ func TestFluentdLogTag(t *testing.T) {
 	RequireDockerVersion(t, ">=1.8.0")
 	RequireDockerVersion(t, "<1.12.0")
 
-	if runtime.GOOS == "arm64" {
+	// Skipping the test for arm as they do not have official support for Arm images
+	if runtime.GOARCH == "arm64" {
 		t.Skip()
 	}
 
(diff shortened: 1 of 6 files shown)


agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -581,7 +582,8 @@ func TestFluentdLogTag(t *testing.T) {
 	RequireDockerVersion(t, ">=1.8.0")
 	RequireDockerVersion(t, "<1.12.0")
 
-	if runtime.GOOS == "arm64" {
+	// Skipping the test for arm as they do not have official support for Arm images
+	if runtime.GOARCH == "arm64" {
 		t.Skip()
```
jahkeup: Mind adding a string to the output with `t.Skip("unsupported, no image for arm64")` for posterity?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -567,7 +567,8 @@ func TestFluentdTag(t *testing.T) {
 	// tag was added in docker 1.9.0
 	RequireDockerVersion(t, ">=1.9.0")
 
-	if runtime.GOOS == "arm64" {
+	// Skipping the test for arm as they do not have official support for Arm images
+	if runtime.GOARCH == "arm64" {
 		t.Skip()
```
jahkeup: Mind adding a string to the output with `t.Skip("unsupported, no image for arm64")` for posterity?

PR #1806, opened 2019-01-22
use state.Reset() rather than copying new state

fierlion:
### Summary
`go vet` tells us that copying the value of a mutex (which we do with `*state = *clean`) is bad.  It's also arguably not so great to transfer 'ownership' like this line does.
This is a fix for the copy and the ownership transfer.  

### Implementation details
In order to ensure that everything flows smoothly once we call state.Reset(), I'm running a precheck to ensure that any error scenarios are shaken out.  Each of the docker_task_engine_state method calls will lock and unlock the state mutex as they make updates (ie `state.AddContainer(container, task)` etc.) so there's no need to lock following state.Reset()

Just For reference, here's the 'not so cautious' version: https://github.com/fierlion/amazon-ecs-agent/tree/unmarshal_state_fix

### Testing
Ran unit tests on ECS AMI EC2. 

Same tests cover the changes.

Update dockerstate unmarshal() to retain ownership

### Licensing

### DIFF
--- a/agent/engine/dockerstate/json.go
+++ b/agent/engine/dockerstate/json.go
@@ -55,47 +55,48 @@ func (state *DockerTaskEngineState) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
-	// reset it by just creating a new one and swapping shortly.
-	// This also means we don't have to lock for the remainder of this function
-	// because we are the only ones with a reference to clean
-	clean := newDockerTaskEngineState()
 
+	// run precheck to shake out all the errors before resetting state.
+	precheckState := newDockerTaskEngineState()
 	for _, task := range saved.Tasks {
-		clean.AddTask(task)
-	}
-	// add image states
-	for _, imageState := range saved.ImageStates {
-		clean.AddImageState(imageState)
+		precheckState.AddTask(task)
 	}
 	for id, container := range saved.IdToContainer {
 		taskArn, ok := saved.IdToTask[id]
 		if !ok {
 			return errors.New("Could not unmarshal state; incomplete save. There was no task for docker id " + id)
 		}
-		task, ok := clean.TaskByArn(taskArn)
+		task, ok := precheckState.TaskByArn(taskArn)
 		if !ok {
 			return errors.New("Could not unmarshal state; incomplete save. There was no task for arn " + taskArn)
 		}
-
-		// The container.Container pointers *must* match the task's container
-		// pointers for things to operate correctly; update them here
-		taskContainer, ok := task.ContainerByName(container.Container.Name)
+		_, ok = task.ContainerByName(container.Container.Name)
 		if !ok {
 			return errors.New("Could not resolve a container into a task based on name: " + task.String() + " -- " + container.String())
 		}
+	}
+
+	// reset state and safely populate with saved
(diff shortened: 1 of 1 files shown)


agent/engine/dockerstate/json.go
```diff
@@ -55,47 +55,50 @@ func (state *DockerTaskEngineState) UnmarshalJSON(data []byte) error {
...
-		container.Container = taskContainer
-		//pointer matching now; everyone happy
-		clean.AddContainer(container, task)
 	}
 
+	// reset state and safely populate with saved
+	state.Reset()
+	for _, task := range saved.Tasks {
+		state.AddTask(task)
+	}
```
sharanyad: Can state still be changed between the different calls i.e state could be changed between `AddTask` and `AddContainer`?

agent/engine/dockerstate/json.go
```diff
@@ -55,47 +55,50 @@ func (state *DockerTaskEngineState) UnmarshalJSON(data []byte) error {
 	if err != nil {
 		return err
 	}
-	// reset it by just creating a new one and swapping shortly.
-	// This also means we don't have to lock for the remainder of this function
-	// because we are the only ones with a reference to clean
-	clean := newDockerTaskEngineState()
 
+	// run precheck; shake out all the errors before resetting state.
```
petderek: why do we need to precheck?

agent/engine/dockerstate/json.go
fierlion: Yes, it can still be changed if its lock is acquired between these calls.  As far as I can tell,  unmarshalling json state into the docker state struct is only done when we start up agent from a backup.  Where are we worried about contention for this lock given the context of unmarshalling from json?  
I believe this issue was at least one big driver for copy (`*state = *clean`) in the first place.  The issue is that `go vet` is unequivocal: thou shalt not ever copy a mutex.
Any thoughts on how else to get around that would be great.

agent/engine/dockerstate/json.go
fierlion: don't want any of those errors to happen while we're setting up the state, leaving it in a nasty... well... state.

agent/engine/dockerstate/json.go
petderek: I was under the impression that this only happens at startup -- which means this should be safe.

agent/engine/dockerstate/json.go
```diff
@@ -55,47 +55,50 @@ func (state *DockerTaskEngineState) UnmarshalJSON(data []byte) error {
...
 			return errors.New("Could not unmarshal state; incomplete save. There was no task for docker id " + id)
 		}
-		task, ok := clean.TaskByArn(taskArn)
+		task, ok := precheckState.TaskByArn(taskArn)
 		if !ok {
 			return errors.New("Could not unmarshal state; incomplete save. There was no task for arn " + taskArn)
 		}
-
 		// The container.Container pointers *must* match the task's container
 		// pointers for things to operate correctly; update them here
```
fenxiong: nit: "update them here" can be removed

fierlion: made this update.  Will wait for tests to complete before merging to dev.

PR #1808, opened 2019-01-23
Modified pull image retry with 1100ms * 2^n strategy and rename Simpl…

suneyz:
### Summary

* Change retry interval for image pull from (250ms * 1.5^n) to (1.1s * 2^n) to make sure pull retry jumps over image pull throttle bucket for ECR images 
* Rename SimpleBackoff to ExponentialBackoff to align with what it actually does
* Refactor retry+backoff into its own package

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -1,4 +1,4 @@
-// Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2014-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"). You may
 // not use this file except in compliance with the License. A copy of the
@@ -41,6 +41,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/ecr"
 	"github.com/aws/amazon-ecs-agent/agent/metrics"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
 
 	"github.com/cihub/seelog"
@@ -84,9 +85,9 @@ const (
 
 	// retry settings for pulling images
 	maximumPullRetries        = 5
-	minimumPullRetryDelay     = 250 * time.Millisecond
-	maximumPullRetryDelay     = 1 * time.Second
-	pullRetryDelayMultiplier  = 1.5
+	minimumPullRetryDelay     = 1100 * time.Millisecond
+	maximumPullRetryDelay     = 5 * time.Second
+	pullRetryDelayMultiplier  = 2
 	pullRetryJitterMultiplier = 0.2
 )
 
@@ -205,6 +206,7 @@ type dockerGoClient struct {
 	ecrTokenCache    async.Cache
 	config           *config.Config
 	context          context.Context
+	imagePullBackoff retry.Backoff
 
 	_time     ttime.Time
 	_timeOnce sync.Once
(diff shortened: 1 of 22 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -84,9 +85,9 @@ const (
 
 	// retry settings for pulling images
 	maximumPullRetries        = 5
-	minimumPullRetryDelay     = 250 * time.Millisecond
-	maximumPullRetryDelay     = 1 * time.Second
-	pullRetryDelayMultiplier  = 1.5
+	minimumPullRetryDelay     = 1100 * time.Millisecond
+	maximumPullRetryDelay     = 5 * time.Second
+	pullRetryDelayMultiplier  = 2
```
shubham2892: Are these to make the tests which depend on pulling images from ECR less flaky?

agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -65,6 +66,15 @@ const xContainerShortTimeout = 1 * time.Millisecond
...
 
+const (
+	// retry settings for pulling images mock backoff
+	xMaximumPullRetries        = 5
+	xMinimumPullRetryDelay     = 25 * time.Millisecond
+	xMaximumPullRetryDelay     = 100 * time.Microsecond
+	xPullRetryDelayMultiplier  = 2
+	xPullRetryJitterMultiplier = 0.2
+)
+
```
shubham2892: Should these value match with the ones in `docker_client`? 

If no, how are these decided?

If yes, should we keep a constant to be reused at both the places.

agent/engine/docker_task_engine.go
```diff
@@ -229,8 +230,8 @@ func (engine *DockerTaskEngine) MustInit(ctx context.Context) {
 	defer engine.mustInitLock.Unlock()
 
 	errorOnce := sync.Once{}
-	taskEngineConnectBackoff := utils.NewSimpleBackoff(200*time.Millisecond, 2*time.Second, 0.20, 1.5)
-	utils.RetryWithBackoff(taskEngineConnectBackoff, func() error {
+	taskEngineConnectBackoff := retry.NewExponentialBackoff(200*time.Millisecond, 2*time.Second, 0.20, 1.5)
```
shubham2892: Can we move the values here to a constant as well, will give a better understanding of what they are.

appveyor.yml
```diff
@@ -13,4 +13,4 @@ install:
 build_script:
   - go build ./agent
 test_script:
-  - for /f "" %%G in ('go list github.com/aws/amazon-ecs-agent/agent/... ^| find /i /v "/vendor/"') do ( go test -race -tags unit -timeout 40s %%G & IF ERRORLEVEL == 1 EXIT 1) 
+  - for /f "" %%G in ('go list github.com/aws/amazon-ecs-agent/agent/... ^| find /i /v "/vendor/"') do ( go test -race -tags unit -timeout 40s %%G & IF ERRORLEVEL == 1 EXIT 1)
```
shubham2892: Should we increase the timeout to 60 here as well?

agent/dockerclient/dockerapi/docker_client.go
suneyz: Right, the longer interval will make sure that if user gets throttled by ECR. the next retry will jump over a throttle bucket (1s). With the current setup starting with 0.25s, the user will be throttled again because the next retry is still within the same throttle bucket.

agent/dockerclient/dockerapi/docker_client_test.go
suneyz: I was deciding wether to increase unit test timeout vs use a customized unit test backoff mechanism. I decided to use different backoff values for unit test is because that if we are using same values from docker client, it will make the unit test of this file to run ~60s. I noticed that our test already takes quite long to run, using customized value for unit might be a better option.

appveyor.yml
suneyz: Forgot to update summary after I decided to use a customized backoff to shorten runtime. I didn't go with the route to update longer unit test timeout. The summary should be updated now.

agent/engine/docker_task_engine.go
suneyz: Sure, make sense :)

agent/utils/retry/exponential_backoff.go
```diff
@@ -0,0 +1,48 @@
+package retry
```
fenxiong: i think you need to add copy right header for the new files in this package?

fenxiong: and update the header of the modified files to xxx-2019 i guess

suneyz: Good catch! updated

PR #1810, opened 2019-01-24
add govet target to Makefile

fierlion:
### Summary
This adds the `govet` target which will run `go vet` on all files in agent/ dir except mock files.  We are excluding mock files.  See implementation details below>>

### Implementation details
We have overloaded the canonical methods MarshalJSON and UnmarshalJSON and have mocks to test these overloaded methods.  
Problem: MarshalJSON() and UnmarshalJSON() are canonical methods, with hardcoded signature expectations: https://golang.org/src/cmd/vet/method.go 
while go mock requires two methods per mock: https://github.com/golang/mock/blob/master/sample/concurrent/mock/concurrent_mock.go (see the 2 sum method signatures at the bottom)
This requires either that we make changes to the method names themselves (risky) or exclude the mocks from `go vet` (less risky)

I've added a follow-up task to rename these methods and remove the exclusion from the `gomake` target.

### Testing
tested on ECS optimized EC2 Amazon Linux instance.

Also ran static-check target as part of build via travis ci:
output logs: https://api.travis-ci.org/v3/job/484041959/log.txt

Add govet target to Makefile

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ all: docker
 gobuild:
 	./scripts/build false
 
+
 # create output directories
 .out-stamp:
 	mkdir -p ./out/test-artifacts ./out/cni-plugins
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -24,6 +24,9 @@ all: docker
 gobuild:
 	./scripts/build false
 
+govet:
```
yunhee-l: can you move this down to line ~356 (https://github.com/aws/amazon-ecs-agent/pull/1810/files#diff-b67911656ef5d18c4ae36cb6741b7965R356) after gofmt and gocyclo? near where static-check related targets are?

yunhee-l: and add .PHONY

Makefile
```diff
@@ -24,6 +24,9 @@ all: docker
...
+	go vet $(shell go list ./agent/... | grep -v mocks)
```
haikuoliu: i tried to run this line one my mac, it shows `zsh: command not found: shell`, maybe you can remove the `shell`? shell should be used by default

samuelkarp: `shell` is a [make function](https://www.gnu.org/software/make/manual/html_node/Shell-Function.html).

fierlion: Can you try running `make govet` from your mac?   It works for me.

https://www.gnu.org/software/make/manual/html_node/Shell-Function.html

Makefile
fierlion: sure.  adding now.

Makefile
```diff
@@ -352,7 +358,7 @@ fmtcheck:
 
 #TODO, create and add go vet target
```
adnxn: this `TODO` can go

fierlion: done.

PR #1818, opened 2019-01-29
Add log driver secret to LogConfig and agent capability

tommyhahn:
### Summary
Add the secret value of Splunk logging driver into the Container's hostconfig section and add the agent capability for secret that supports Splunk

### Testing

New tests cover the changes: 

### Licensing


agent/api/task/task.go
```diff
@@ -900,6 +903,27 @@ func (task *Task) ApplyExecutionRoleLogsAuth(hostConfig *dockercontainer.HostCon
 	return nil
 }
 
+//ApplyLogDriverSecret will add the generated splunk secret into the associated HostConfig
```
samuelkarp: There's nothing specific to Splunk here.

agent/api/task/task.go
```diff
@@ -1699,23 +1723,43 @@ func (task *Task) PopulateSecretsAsEnv(container *apicontainer.Container) *apier
...
 			}
 		}
 	}
 
-	container.MergeEnvironmentVariables(envVars)
+	if len(envVars) != 0 {
+		container.MergeEnvironmentVariables(envVars)
+	}
+
+	//Check if the "splunk-token" has already been initialized and stored in the LogCofig map
```
samuelkarp: We should not include logic specific to Splunk or keys like "splunk-token".  If we need logic like this, it should be handled in the ECS backend so we do not need to keep updating the agent.

yunhee-l: I agree - this was discussed during design phase, to not cater implementation to one specific log driver.

agent/engine/docker_task_engine.go
```diff
@@ -859,9 +858,11 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		}
 	}
 
-	// apply secrets to container.Environment
-	if container.HasSecretAsEnv() {
-		err := task.PopulateSecretsAsEnv(container)
+	//Apply the log driver secret into container's LogConfig and Env secrets to container.Environment
+	if container.HasSecretAsEvnAndLogDriver() {
+		//splunkToken, ok := hostConfig.LogConfig.Config["splunk-token"]
```
suneyz: remove this comment code?

agent/api/task/task.go
tommyhahn: Figured out a way. Will discuss and update this pr shortly.

agent/engine/docker_task_engine.go
tommyhahn: Done 😀

agent/api/task/task.go
```diff
@@ -1699,23 +1723,43 @@ func (task *Task) PopulateSecretsAsEnv(container *apicontainer.Container) *apier
...
 	envVars := make(map[string]string)
 
+	LogDriverName := hostConfig.LogConfig.Type
+	LogDriverTokenSecretValue := ""
+
 	for _, secret := range container.Secrets {
-		if secret.Provider == apicontainer.SecretProviderSSM && secret.Type == apicontainer.SecretTypeEnv {
-			k := secret.GetSecretResourceCacheKey()
+		k := secret.GetSecretResourceCacheKey()
+		if secret.Provider == apicontainer.SecretProviderSSM {
```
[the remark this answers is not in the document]
tommyhahn: Combined 👍🏼

agent/api/container/container.go
```diff
@@ -807,7 +813,17 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
...
+func (c *Container) ApplyLogDriverSecret(hostConfig *dockercontainer.HostConfig, logDriverSecretName string, logDriverSecretValue string) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	hostConfig.LogConfig.Config[logDriverSecretName] = logDriverSecretValue
+}
+
+// MergeLogDrivers appends additional pairs to
+// the the container's Logging Drivers structure
+func (c *Container) HasSecretAsEvnAndLogDriver() bool {
```
sharanyad: Think it should be`HasSecretAsEnvOrLogDriver`. either way, just calling it `HasSecrets` or `HasSecretsAvailable` might be better. also the comment needs to be changed.

agent/api/container/container.go
```diff
@@ -807,7 +813,17 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
 	}
 }
 
-func (c *Container) HasSecretAsEnv() bool {
+//Append the logging driver using secret into the container's LogConfig
```
sharanyad: nit: space after `//`

agent/api/task/task.go
```diff
@@ -1737,8 +1737,8 @@ func (task *Task) getSSMSecretsResource() ([]taskresource.TaskResource, bool) {
 	return res, ok
 }
 
-// PopulateSecretsAsEnv appends the container's env var map with secrets
-func (task *Task) PopulateSecretsAsEnv(container *apicontainer.Container) *apierrors.DockerClientConfigError {
+// PopulateSecretsAsEnvAndLogDriver appends the container's env var map with secrets
+func (task *Task) PopulateSecretsAsEnvAndLogDriver(hostConfig *dockercontainer.HostConfig, container *apicontainer.Container) *apierrors.DockerClientConfigError {
```
sharanyad: Can we just call it `PopulateSecrets` so we that don't have to change the name again when new secret type is added?

agent/api/container/container.go
```diff
@@ -63,6 +65,9 @@ const (
 
 	// SecretTypeEnv is to show secret type being ENVIRONMENT_VARIABLE
 	SecretTypeEnv = "ENVIRONMENT_VARIABLE"
+
+	// TargetLogDriver is to show secret target being "LOG_DRIVER", the default will be "CONTAINER"
```
sharanyad: where is the default being set?

samuelkarp:
Since this code is no longer specific to Splunk, can we remove Splunk from the commit message and the PR title?

agent/api/container/container.go
tommyhahn: Will change it to HasSecretAsEnvOrLogDriver at this time. It is specifically for checking env and log driver secrets.

agent/api/container/container.go
```diff
@@ -807,7 +813,15 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
...
-func (c *Container) HasSecretAsEnv() bool {
+// Append the logging driver using secret into the container's LogConfig
+func (c *Container) ApplyLogDriverSecret(hostConfig *dockercontainer.HostConfig, logDriverSecretName string, logDriverSecretValue string) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	hostConfig.LogConfig.Config[logDriverSecretName] = logDriverSecretValue
+}
+
+func (c *Container) HasSecretAsEvnOrLogDriver() bool {
```
fenxiong: env

agent/api/container/container.go
```diff
@@ -807,7 +813,15 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
 	}
 }
 
-func (c *Container) HasSecretAsEnv() bool {
+// Append the logging driver using secret into the container's LogConfig
+func (c *Container) ApplyLogDriverSecret(hostConfig *dockercontainer.HostConfig, logDriverSecretName string, logDriverSecretValue string) {
```
fenxiong: i don't think this should be a method of container.. the container is not used. i think this can just be a helper function put below its caller

agent/api/task/task.go
```diff
@@ -1760,23 +1760,47 @@ func (task *Task) PopulateSecretsAsEnv(container *apicontainer.Container) *apier
...
+		if !validValue {
+			continue
 		}
+
+		if secret.Type == apicontainer.SecretTypeEnv {
+			envVars[secret.Name] = secretVal
+			continue
+		}
+		if secret.Target == apicontainer.SecretTargetLogDriver {
+			logDriverTokenName = secret.Name
```
fenxiong: i think currently you assume there's only one log driver secret, but i don't see why you need to have that assumption.. i think you can call ApplyLogDriverSecret here whenever you find a log driver secret?

agent/api/task/task.go
```diff
@@ -1760,23 +1760,47 @@ func (task *Task) PopulateSecretsAsEnv(container *apicontainer.Container) *apier
...
+		if secretValue, ok := ssmRes.GetCachedSecretValue(k); ok {
+			validValue = ok
+			secretVal = secretValue
+		}
+		if secretValue, ok := asmRes.GetCachedSecretValue(k); ok {
+			validValue = ok
+			secretVal = secretValue
 		}
 
-		if secret.Provider == apicontainer.SecretProviderASM && secret.Type == apicontainer.SecretTypeEnv {
```
fenxiong: I don't think you should the provider condition for env secret and always try to get secret value from both.. ssmRes and asmRes will not always be initialized

agent/app/agent_capability.go
```diff
@@ -138,6 +145,9 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	// ecs agent version 1.23.0 supports ecs secrets integrating with aws secrets manager
 	capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+capabilitySecretEnvASM)
 
+	// ecs agent version 1.25.0 supports ecs secrets for logging drivers
```
fenxiong: probably 1.26.0?

agent/api/container/container.go
```diff
@@ -269,6 +272,7 @@ type Secret struct {
 	ContainerPath string `json:"containerPath"`
 	Type          string `json:"type"`
 	Provider      string `json:"provider"`
+	Target        string `json:"target"`
```
fenxiong: sorry that i forgot.. i think you need to bump state file version for this

tommyhahn: Done!

tommyhahn:
#1856 
Flaky unit test TestStateManager issue

Issue #1814, opened 2019-01-26
[DOC] - Reserved Ports - Prometheus 51680


FLavalliere:
With the Addition of the Prometheus port the Doc / Default ECS Reserved ports should be revisited.

Add TCP Reserved Port of 51680

adnxn:
closing this issue since this port is only reserved if the feature is toggled by `ECS_ENABLE_PROMETHEUS_METRICS`, it is a not reserved by default. today this is not a public facing feature.

PR #1819, opened 2019-01-29
Add Prometheus port number to reserved ports

sharanyad:
### Summary
fixing README for https://github.com/aws/amazon-ecs-agent/issues/1814

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ additional details on each available environment variable.
 | Environment Key | Example Value(s)            | Description | Default value on Linux | Default value on Windows |
 |:----------------|:----------------------------|:------------|:-----------------------|:-------------------------|
 | `ECS_CLUSTER`       | clusterName             | The cluster this agent should check into. | default | default |
-| `ECS_RESERVED_PORTS` | `[22, 80, 5000, 8080]` | An array of ports that should be marked as unavailable for scheduling on this container instance. | `[22, 2375, 2376, 51678, 51679]` | `[53, 135, 139, 445, 2375, 2376, 3389, 5985, 5986, 51678, 51679]`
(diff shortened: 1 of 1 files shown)


yunhee-l:
does it matter that if ECS_ENABLE_PROMETHEUS_METRICS is set to false, we are technically not going to "reserve" port 51680?

sharanyad:
ah missed that flag. It's true that the port won't be reserved. But if ECS_ENABLE_PROMETHEUS_METRICS  is set to true, it will be reserved silently. What would you suggest doing here?

yunhee-l:
Prometheus metrics is probably a feature that majority of our customers won't be using for their day to day workload. Can we loop in doc writers and get their opinion on this?

petderek:
>reserved silently

 Does agent communicate this to the scheduler?

adnxn:
adding to @yunhee-l concerns - do we document what each of these ports are reserved for anywhere?

sharanyad:
> do we document what each of these ports are reserved for anywhere?

yes -- https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html

sharanyad:
> Does agent communicate this to the scheduler?

@petderek  no, agent does not communicate reserved ports to backend

samuelkarp:
> @petderek no, agent does not communicate reserved ports to backend

@sharanyad Reserved ports should be communicated as part of RegisterContainerInstance

yunhee-l:
we are not releasing prometheus as customer feature (see https://github.com/aws/amazon-ecs-agent/pull/1834), and we are not likely to enable this in our releases. Do we still need to reserve these ports?

samuelkarp:
> we are not releasing prometheus as customer feature (see #1834), and we are not likely to enable this in our releases. Do we still need to reserve these ports?

If the feature is enabled (including through the use of an environment variable) I would recommend reserving the ports.

sharanyad:
looks like it's already added 
see : https://github.com/aws/amazon-ecs-agent/issues/1843

hflamboauto1:
@sharanyad  I don't see this on master https://github.com/aws/amazon-ecs-agent/README.md
I see https://github.com/aws/amazon-ecs-agent/issues/1814

PR #1821, opened 2019-01-30
Add app mesh CNI plugin support

suneyz:
### Summary

Add app mesh cni plugin support. This will enable agent to invoke CNI plugin to configure ip rules to support app mesh during task start up. The code is work in progress and pending some action items

### TODO: 
Some of the model fields' name might change after a few of the followings are finalized
~~1. ACS model to be finalized and changed.~~
~~2. ACS app mesh model convert to app mesh model in agent.~~
~~3. CNI plugin to be implemented and finalized.~~
4. Build CNI binary (Will address this in a separate PR, functional tests failed because of this, cannot locate binary file)

### Implementation details

During agent start up, when verify CNI capability, agent will verify CNI plugin capability to support app mesh. Then pass the agent capability to control plane on container registration. Agent will also setup ip rule table through invoking app mesh CNI plugin during task start up. And clean it during stop task.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/acs/handler/payload_handler.go
+++ b/agent/acs/handler/payload_handler.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	"github.com/aws/amazon-ecs-agent/agent/api"
+	apiappmesh "github.com/aws/amazon-ecs-agent/agent/api/appmesh"
 	apieni "github.com/aws/amazon-ecs-agent/agent/api/eni"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
@@ -221,6 +222,16 @@ func (payloadHandler *payloadRequestHandler) addPayloadTasks(payload *ecsacs.Pay
 			}
 
 			apiTask.SetTaskENI(eni)
+		}
+		// Add the app mesh information to task struct
+		if task.ProxyConfiguration != nil {
+			appmesh, err := apiappmesh.AppMeshFromACS(task.ProxyConfiguration)
+			if err != nil {
+				payloadHandler.handleUnrecognizedTask(task, err, payload)
+				allTasksOK = false
+				continue
+			}
+			apiTask.SetAppMesh(appmesh)
 		}
 		if task.ExecutionRoleCredentials != nil {
 			// The payload message contains execution credentials for the task.
--- a/agent/api/appmesh/appmesh.go
+++ b/agent/api/appmesh/appmesh.go
@@ -1,0 +1,83 @@
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package appmesh
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+const (
+	appMesh            = "APPMESH"
+	splitter           = ","
+	ignoredUID         = "IgnoredUID"
+	ignoredGID         = "IgnoredGID"
+	proxyIngressPort   = "ProxyIngressPort"
+	proxyEgressPort    = "ProxyEgressPort"
+	appPorts           = "AppPorts"
+	egressIgnoredIPs   = "EgressIgnoredIPs"
+	egressIgnoredPorts = "EgressIgnoredPorts"
+)
+
+// AppMesh contains information of app mesh config
+type AppMesh struct {
+	// IgnoredUID is egress traffic from the processes owned by the UID will be ignored
+	IgnoredUID string
+	// IgnoredGID specifies egress traffic from the processes owned by the GID will be ignored
+	IgnoredGID string
+	// ProxyIngressPort is the ingress port number that proxy is listening on
+	ProxyIngressPort string
+	// ProxyEgressPort is the egress port number that proxy is listening on
+	ProxyEgressPort string
+	// AppPorts is the port number that application is listening on
+	AppPorts []string
+	// EgressIgnoredIPs is the list of ports for which egress traffic will be ignored
+	EgressIgnoredIPs []string
+	// EgressIgnoredPorts is the list of IPs for which egress traffic will be ignored
+	EgressIgnoredPorts []string
+}
+
+// AppMeshFromACS validates proxy config if it is app mesh type and creates AppMesh object
+func AppMeshFromACS(proxyConfig *ecsacs.ProxyConfiguration) (*AppMesh, error) {
+
+	if *proxyConfig.Type != appMesh {
+		return nil, fmt.Errorf("agent does not support proxy type other than app mesh")
+	}
+
+	var inputAppPorts []string
+	var inputEgressIgnoredIPs []string
+	var inputEgressIgnoredPorts []string
+
+	if proxyConfig.Properties[appPorts] != nil {
+		inputAppPorts = strings.Split(*proxyConfig.Properties[appPorts], splitter)
+	}
+	if proxyConfig.Properties[egressIgnoredIPs] != nil {
+		inputEgressIgnoredIPs = strings.Split(*proxyConfig.Properties[egressIgnoredIPs], splitter)
+	}
+	if proxyConfig.Properties[egressIgnoredPorts] != nil {
+		inputEgressIgnoredPorts = strings.Split(*proxyConfig.Properties[egressIgnoredPorts], splitter)
+	}
+
+	return &AppMesh{
+		IgnoredUID:         aws.StringValue(proxyConfig.Properties[ignoredUID]),
+		IgnoredGID:         aws.StringValue(proxyConfig.Properties[ignoredGID]),
+		ProxyIngressPort:   aws.StringValue(proxyConfig.Properties[proxyIngressPort]),
+		ProxyEgressPort:    aws.StringValue(proxyConfig.Properties[proxyEgressPort]),
+		AppPorts:           inputAppPorts,
+		EgressIgnoredIPs:   inputEgressIgnoredIPs,
+		EgressIgnoredPorts: inputEgressIgnoredPorts,
+	}, nil
+}
--- a/agent/api/appmesh/appmesh_test.go
+++ b/agent/api/appmesh/appmesh_test.go
@@ -1,0 +1,87 @@
+// +build unit
+
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package appmesh
+
+import (
+	"testing"
+
+	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	mockEgressIgnoredIP1   = "128.0.0.1"
+	mockEgressIgnoredIP2   = "171.1.3.24"
+	mockAppPort1           = "8000"
+	mockAppPort2           = "8001"
+	mockEgressIgnoredPort1 = "13000"
+	mockEgressIgnoredPort2 = "13001"
+	mockIgnoredUID         = "1337"
+	mockIgnoredGID         = "2339"
+	mockProxyIngressPort   = "9000"
+	mockProxyEgressPort    = "9001"
+	mockAppPorts           = mockAppPort1 + splitter + mockAppPort2
+	mockEgressIgnoredIPs   = mockEgressIgnoredIP1 + splitter + mockEgressIgnoredIP2
+	mockEgressIgnoredPorts = mockEgressIgnoredPort1 + splitter + mockEgressIgnoredPort2
+	mockContainerName      = "testEnvoyContainer"
+)
+
+func TestAppMeshFromACS(t *testing.T) {
+	testProxyConfig := prepareProxyConfig()
+
+	appMesh, err := AppMeshFromACS(&testProxyConfig)
+
+	assert.NoError(t, err)
+	assert.NotNil(t, appMesh)
+	assert.Equal(t, mockIgnoredUID, appMesh.IgnoredUID)
+	assert.Equal(t, mockIgnoredGID, appMesh.IgnoredGID)
+	assert.Equal(t, mockProxyEgressPort, appMesh.ProxyEgressPort)
+	assert.Equal(t, mockProxyIngressPort, appMesh.ProxyIngressPort)
+	assert.Equal(t, mockAppPort1, appMesh.AppPorts[0])
+	assert.Equal(t, mockAppPort2, appMesh.AppPorts[1])
+	assert.Equal(t, mockEgressIgnoredIP1, appMesh.EgressIgnoredIPs[0])
+	assert.Equal(t, mockEgressIgnoredIP2, appMesh.EgressIgnoredIPs[1])
+	assert.Equal(t, mockEgressIgnoredPort1, appMesh.EgressIgnoredPorts[0])
+	assert.Equal(t, mockEgressIgnoredPort2, appMesh.EgressIgnoredPorts[1])
+}
+
+func TestAppMeshFromACSNonAppMeshProxyInput(t *testing.T) {
+	someOtherProxyType := "fooProxy"
+	testProxyConfig := prepareProxyConfig()
+	testProxyConfig.Type = &someOtherProxyType
+
+	_, err := AppMeshFromACS(&testProxyConfig)
+
+	assert.Error(t, err)
+}
+
+func prepareProxyConfig() ecsacs.ProxyConfiguration {
+
+	return ecsacs.ProxyConfiguration{
+		Type:          aws.String(appMesh),
+		Properties:    map[string]*string{
+			ignoredUID:         aws.String(mockIgnoredUID),
+			ignoredGID:         aws.String(mockIgnoredGID),
+			proxyIngressPort:   aws.String(mockProxyIngressPort),
+			proxyEgressPort:    aws.String(mockProxyEgressPort),
+			appPorts:           aws.String(mockAppPorts),
+			egressIgnoredIPs:   aws.String(mockEgressIgnoredIPs),
+			egressIgnoredPorts: aws.String(mockEgressIgnoredPorts),
+		},
+		ContainerName: aws.String(mockContainerName),
+	}
+}
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	apiappmesh "github.com/aws/amazon-ecs-agent/agent/api/appmesh"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	apieni "github.com/aws/amazon-ecs-agent/agent/api/eni"
@@ -175,6 +176,9 @@ type Task struct {
 
 	// ENI is the elastic network interface specified by this task
 	ENI *apieni.ENI
+
+	// AppMesh is the service mesh specified by the task
+	AppMesh *apiappmesh.AppMesh
 
 	// MemoryCPULimitsEnabled to determine if task supports CPU, memory limits
 	MemoryCPULimitsEnabled bool `json:"MemoryCPULimitsEnabled,omitempty"`
@@ -693,8 +697,16 @@ func (task *Task) BuildCNIConfig() (*ecscni.Config, error) {
 	}
 
 	cfg := &ecscni.Config{}
-	eni := task.GetTaskENI()
+	convertENIToCNIConfig(task.GetTaskENI(), cfg)
+	if task.GetAppMesh() != nil {
+		convertAppMeshToCNIConfig(task.GetAppMesh(), cfg)
+	}
 
+	return cfg, nil
+}
+
+// convertENIToCNIConfig converts input eni config into cni config
+func convertENIToCNIConfig(eni *apieni.ENI, cfg *ecscni.Config) {
 	cfg.ENIID = eni.ID
 	cfg.ID = eni.MacAddress
 	cfg.ENIMACAddress = eni.MacAddress
@@ -705,13 +717,23 @@ func (task *Task) BuildCNIConfig() (*ecscni.Config, error) {
 			break
 		}
 	}
-
 	// If there is ipv6 assigned to eni then set it
 	if len(eni.IPV6Addresses) > 0 {
 		cfg.ENIIPV6Address = eni.IPV6Addresses[0].Address
 	}
+}
 
-	return cfg, nil
+// convertAppMeshToCNIConfig converts input app mesh config into cni config
+func convertAppMeshToCNIConfig(appMesh *apiappmesh.AppMesh, cfg *ecscni.Config) {
+	cfg.AppMeshCNIEnabled = true
+	cfg.IgnoredUID = appMesh.IgnoredUID
+	cfg.IgnoredGID = appMesh.IgnoredGID
+	cfg.ProxyIngressPort = appMesh.ProxyIngressPort
+	cfg.ProxyEgressPort = appMesh.ProxyEgressPort
+	cfg.AppPorts = appMesh.AppPorts
+	cfg.EgressIgnoredIPs = appMesh.EgressIgnoredIPs
+	cfg.EgressIgnoredPorts = appMesh.EgressIgnoredPorts
+
 }
 
 // isNetworkModeVPC checks if the task is configured to use task-networking feature
@@ -1520,6 +1542,22 @@ func (task *Task) GetTaskENI() *apieni.ENI {
 	defer task.lock.RUnlock()
 
 	return task.ENI
+}
+
+// SetAppMesh sets the app mesh config of the task
+func (task *Task) SetAppMesh(appMesh *apiappmesh.AppMesh) {
+	task.lock.Lock()
+	defer task.lock.Unlock()
+
+	task.AppMesh = appMesh
+}
+
+// GetAppMesh returns the app mesh config of the task
+func (task *Task) GetAppMesh() *apiappmesh.AppMesh {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	return task.AppMesh
 }
 
 // GetStopSequenceNumber returns the stop sequence number of a task
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -32,6 +32,7 @@ const (
 	capabilityTaskIAMRoleNetHost                = "task-iam-role-network-host"
 	taskENIAttributeSuffix                      = "task-eni"
 	taskENIBlockInstanceMetadataAttributeSuffix = "task-eni-block-instance-metadata"
+	appMeshAttributeSuffix                      = "aws-appmesh"
 	cniPluginVersionSuffix                      = "cni-plugin-version"
 	capabilityTaskCPUMemLimit                   = "task-cpu-mem-limit"
 	capabilityDockerPluginInfix                 = "docker-plugin."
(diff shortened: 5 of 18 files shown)


agent/acs/handler/payload_handler.go
```diff
@@ -222,6 +223,16 @@ func (payloadHandler *payloadRequestHandler) addPayloadTasks(payload *ecsacs.Pay
 
 			apiTask.SetTaskENI(eni)
 		}
+		// Add the app mesh information to task struct
+		if task.ProxyConfiguration != nil {
+			appmesh, err := apiappmesh.AppMeshFromACS(task.ProxyConfiguration)
```
fierlion: I'm still new to the idioms but would this work here, or is it less readable?
```
if appmesh, err := apiappmesh.AppMeshFromACS(task.ProxyConfiguration); err!= nil {
  //handle it
}
```

agent/api/appmesh/appmesh.go
```diff
@@ -0,0 +1,114 @@
...
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package appmesh
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
```
yumex93: The convention that we normally follows for import is 
import (
  "standard package"
  +empty line
  "third-party package"
)
e.g. import (
   "fmt"
  
   "github.com/****"
)

agent/api/appmesh/appmesh_test.go
```diff
@@ -0,0 +1,173 @@
...
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package appmesh
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/aws-sdk-go/aws"
```
yumex93: this import also needs to be updated

agent/api/appmesh/appmesh_test.go
```diff
@@ -0,0 +1,173 @@
...
+	delete(testProxyConfig.Properties, ignoredGID)
+	delete(testProxyConfig.Properties, ignoredUID)
+
+	appMesh, err := AppMeshFromACS(&testProxyConfig)
+
+	assert.Error(t, err)
+	assert.Nil(t, appMesh)
+}
+
+func TestAppMeshFromACSNpProxyIngressPort(t *testing.T) {
```
yumex93: **TestAppMeshFromACSNpProxyIngressPort** should be **TestAppMeshFromACSNoProxyIngressPort**?

agent/api/appmesh/appmesh_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	delete(testProxyConfig.Properties, proxyIngressPort)
+
+	appMesh, err := AppMeshFromACS(&testProxyConfig)
+
+	assert.Error(t, err)
+	assert.Nil(t, appMesh)
+}
+
+func TestAppMeshFromACSNpProxyEgressPort(t *testing.T) {
```
yumex93: same here, is the function name the name you want?

agent/api/appmesh/appmesh_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	delete(testProxyConfig.Properties, proxyEgressPort)
+
+	appMesh, err := AppMeshFromACS(&testProxyConfig)
+
+	assert.Error(t, err)
+	assert.Nil(t, appMesh)
+}
+
+func TestAppMeshFromACSNpAppPorts(t *testing.T) {
```
yumex93: **NoAppPorts**?

agent/api/appmesh/appmesh_test.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+func TestAppMeshFromACSNonAppMeshProxyInput(t *testing.T) {
+	someOtherProxyType := "fooProxy"
+	testProxyConfig := prepareProxyConfig()
+	testProxyConfig.Type = &someOtherProxyType
+
+	appMesh, err := AppMeshFromACS(&testProxyConfig)
+
+	assert.NoError(t, err)
```
yumex93: maybe use **assert.EqualError** to verify the error is the expected one?

agent/api/appmesh/appmesh_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+func TestAppMeshFromACS(t *testing.T) {
+	testProxyConfig := prepareProxyConfig()
+
+	appMesh, err := AppMeshFromACS(&testProxyConfig)
+
+	assert.NoError(t, err)
+	assert.NotNil(t, appMesh)
+	assert.True(t, appMesh.AppMeshEnabled)
+	assert.Equal(t, mockIgnoredUID, appMesh.IgnoredUID)
```
yumex93: Can u add test case to check AppMesh can be created successfully if optional parameters are missed?

agent/api/task/task.go
```diff
@@ -696,13 +707,33 @@ func (task *Task) BuildCNIConfig() (*ecscni.Config, error) {
...
+	cfg.AppMeshCNIEnabled = appMesh.AppMeshEnabled
+	cfg.IgnoredUID = appMesh.IgnoredUID
+	cfg.IgnoredGID = appMesh.IgnoredGID
+	cfg.ProxyIngressPort = appMesh.ProxyIngressPort
+	cfg.ProxyEgressPort = appMesh.ProxyEgressPort
+	cfg.AppPorts = appMesh.AppPorts
+
+	var egressIgnoredIPs []net.IP
+	for _, ip := range appMesh.EgressIgnoredIPs {
+		ip := net.ParseIP(ip)
```
yumex93: Do you know what validation that ECS service side is doing? Do they validate whether the ip address is a valid ipv4/ipv6 address and how do they handle heading or trailing spaces?

agent/api/task/task.go
```diff
@@ -696,13 +707,33 @@ func (task *Task) BuildCNIConfig() (*ecscni.Config, error) {
...
+		if ip == nil {
+			seelog.Warnf("app mesh config: malformed egress ignored ip")
```
yumex93: If I understand correctly, if the ip is invalid, we just ignore it and log as warning? If this is the case, I am not sure whether this is the right action since it is hard for customers to figure out what happened if only some of ip settings are missing. We need to discuss about this. Also, the log message needs to have the malformed ip printed out so we can know which one causes it.

agent/api/task/task.go
```diff
@@ -1513,6 +1544,22 @@ func (task *Task) GetTaskENI() *apieni.ENI {
 	return task.ENI
 }
 
+// SetAppMesh sets the app mesh config of the task
+func (task *Task) SetAppMesh(appMesh *apiappmesh.AppMesh) {
+	task.lock.RLock()
```
yumex93: Here should be task.lock.Lock()/task.lock.Unlock(). RLock is read lock.

agent/ecscni/types.go
```diff
@@ -41,13 +44,17 @@ const (
 	ECSBridgePluginName = "ecs-bridge"
 	// ECSENIPluginName is the binary of the eni plugin
 	ECSENIPluginName = "ecs-eni"
+	// ECSAppMeshPluginName is the binary of app mesh plugin
+	// TODO: Confirm this
+	ECSAppMeshPluginName = "ecs-app-mesh"
```
yumex93: the name will only be **app-mesh**

agent/ecscni/types.go
```diff
@@ -124,6 +131,28 @@ type ENIConfig struct {
...
 }
 
+// AppMeshConfig contains all the information needed to invoke the app mesh plugin
+type AppMeshConfig struct {
+	// Type is the cni plugin name
+	Type string `json:"type,omitempty"`
+	// CNIVersion is the cni spec version to use
+	CNIVersion string `json:"cniVersion,omitempty"`
+	// Egress traffic from the processes owned by the UID will be ignored
+	IgnoredUID string `json:"ignored-uid,omitempty"`
```
yumex93: For cni configuration, normally it would follow camel case. That is also what I implemented in cni plugin. These names need to be updated.

agent/config/config_unix.go
```diff
@@ -80,6 +80,7 @@ func DefaultConfig() Config {
 		PollMetrics:                         false,
 		PollingMetricsWaitDuration:          DefaultPollingMetricsWaitDuration,
 		NvidiaRuntime:                       DefaultNvidiaRuntime,
+		AppMeshEnabled:                      false,
```
yumex93: Do you also need to update the test case **TestConfigDefault** in config_unix_test.go?

agent/ecscni/plugin.go
```diff
@@ -180,11 +187,18 @@ func (client *cniClient) cleanupNS(cfg *Config) error {
 		return errors.Wrap(err, "cni cleanup: invoke bridge plugin failed")
 	}
 	seelog.Debugf("[ECSCNI] bridge cleanup done: %s", cfg.ContainerID)
-
+	// clean up eni network namespace
 	err = client.del(runtimeConfig, cfg, client.createENINetworkConfig)
 	if err != nil {
 		return errors.Wrap(err, "cni cleanup: invoke eni plugin failed")
 	}
+	if cfg.AppMeshCNIEnabled {
```
yumex93: Is it possible that cfg is nil when u call cfg.AppMeshCNIEnabled here?

agent/ecscni/types.go
```diff
@@ -30,6 +31,8 @@ const (
 	// defaultBridgeName is the default name of bridge created for container to
 	// communicate with ecs-agent
 	defaultBridgeName = "ecs-bridge"
+	// TODO: Confirm this
+	defaultAppMeshName = "ecs-app-mesh"
```
yumex93: the name will only be **app-mesh**

yumex93:
1. For commit message, I recommend to follow the conversion:
updated package name: what is changed.
e.g. acs, engine, wsclient: small spelling fixes

2. It is recommended that each commit is more granular. Do not commit all changes into one.

agent/api/appmesh/appmesh.go
suneyz: will do :)

agent/api/appmesh/appmesh_test.go
suneyz: updated

agent/api/appmesh/appmesh_test.go
suneyz: removed these tests, as we discussed that cni plugin will validate parameters and agent only pass through the config

agent/api/appmesh/appmesh_test.go
suneyz: good idea!

agent/api/task/task.go
suneyz: since we are not validating parameters, this will be removed

agent/api/task/task.go
suneyz: ditto

agent/api/task/task.go
suneyz: Didn't know the difference, changed to to regular lock

agent/config/config_unix.go
suneyz: updated

agent/ecscni/types.go
suneyz: updated plugin name

agent/ecscni/types.go
suneyz: sounds good, updated

agent/ecscni/plugin.go
suneyz: discussed offline, cfg is always constructed or will be error out before here.

agent/ecscni/types.go
suneyz: updated

agent/ecscni/types.go
```diff
@@ -156,4 +184,20 @@ type Config struct {
 	AdditionalLocalRoutes []cnitypes.IPNet
 	// SubnetGatewayIPV4Address is the address to the subnet gate for the eni
 	SubnetGatewayIPV4Address string
+	// App Mesh CNI enabled
```
aaithal: [multiple occurrences] please do read https://golang.org/doc/effective_go.html#commentary for how to comment in golang. This should read something like: `// AppMeshCNIEnabled is ...`

agent/api/appmesh/appmesh.go
```diff
@@ -0,0 +1,87 @@
...
+	// Port number that application is listening on
+	AppPorts []string
+	// List of ports for which egress traffic will be ignored
+	EgressIgnoredIPs []string
+	// List of IPs for which egress traffic will be ignored
+	EgressIgnoredPorts []string
+}
+
+// AppMeshFromACS validates proxy config if it is app mesh type and create AppMesh object
+func AppMeshFromACS(proxyConfig *ecsacs.ProxyConfiguration) (*AppMesh, error) {
```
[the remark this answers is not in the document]
suneyz: Good catch! i had some validation in first revision. We decided to do those validations in CNI plugin. To reduce duplicate validation and code. Will remove error from return type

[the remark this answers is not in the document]
suneyz: actually, we will do a proxy config type validation here. Will update with the validation

agent/api/appmesh/appmesh_test.go
suneyz: these tests were already included in this cr. but discussed offline that we don't need to validate anymore. will remove them

agent/api/task/task.go
```diff
@@ -176,6 +177,9 @@ type Task struct {
 	// ENI is the elastic network interface specified by this task
 	ENI *apieni.ENI
 
+	// App mesh is the service mesh specified by the task
```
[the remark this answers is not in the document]
suneyz: Updated

agent/api/task/task.go
```diff
@@ -1522,6 +1542,22 @@ func (task *Task) GetTaskENI() *apieni.ENI {
 	return task.ENI
 }
 
+// SetAppMesh sets the app mesh config of the task
+func (task *Task) SetAppMesh(appMesh *apiappmesh.AppMesh) {
+	task.lock.Lock()
+	defer task.lock.Unlock()
+
+	task.AppMesh = appMesh
```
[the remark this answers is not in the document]
suneyz: You are right, this shouldn't need a lock. I added it in originally to be consistent with all other fields. I can remove the lock if you prefer that way, let me know how do you think

agent/ecscni/types.go
```diff
@@ -30,6 +30,9 @@ const (
 	// defaultBridgeName is the default name of bridge created for container to
 	// communicate with ecs-agent
 	defaultBridgeName = "ecs-bridge"
+	// defaultBridgeName is the default name of app mesh to setup iptable rules
```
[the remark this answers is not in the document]
suneyz: updated :)

agent/ecscni/types.go
```diff
@@ -41,13 +44,16 @@ const (
...
+	// ECSAppMeshPluginName is the binary of app mesh plugin
+	ECSAppMeshPluginName = "app-mesh"
 	// TaskIAMRoleEndpoint is the endpoint of ecs-agent exposes credentials for
 	// task IAM role
 	TaskIAMRoleEndpoint = "169.254.170.2/32"
 	// CapabilityAWSVPCNetworkingMode is the capability string, which when
 	// present in the output of the '--capabilities' command of a CNI plugin
 	// indicates that the plugin can support the ECS "awsvpc" network mode
 	CapabilityAWSVPCNetworkingMode = "awsvpc-network-mode"
+	CapabilityAppMeshSupport       = "app-mesh-support"
```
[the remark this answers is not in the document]
suneyz: we decided to remove this as CNI plugin no longer pass capability. We will check if the cni plugin has appmesh binary during agent initialization.

agent/ecscni/types.go
suneyz: sounds good, updated

agent/api/task/task.go
```diff
@@ -1522,6 +1542,22 @@ func (task *Task) GetTaskENI() *apieni.ENI {
...
+}
+
+// GetAppMesh returns the app mesh config of the task
+func (task *Task) GetAppMesh() *apiappmesh.AppMesh {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	return task.AppMesh
```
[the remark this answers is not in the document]
suneyz: ditto

agent/ecscni/types.go
```diff
@@ -30,6 +30,9 @@ const (
 	// defaultBridgeName is the default name of bridge created for container to
 	// communicate with ecs-agent
 	defaultBridgeName = "ecs-bridge"
+	// defaultBridgeName is the default name of app mesh to setup iptable rules
+	// for app mesh container
+	defaultAppMeshName = "app-mesh"
```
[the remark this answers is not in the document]
suneyz: Good question. We are also discussing this within our team. Do you have suggestions on what should we pass into CNI plugin?  We don't use this field, but it is a required field to invoke CNI plugin. I followed the convention above to pass in the plugin name into it.

agent/config/types.go
```diff
@@ -159,6 +159,9 @@ type Config struct {
 	// defined EC2 networks
 	TaskENIEnabled bool
 
+	// AppMeshEnabled specifies if the Agent is capable of invoking app mesh through CNI plugin
+	AppMeshEnabled bool
```
yumex93: Why we need this in the agent config? Should not that new agent capability for app mesh is enough?

agent/app/agent_capability.go
```diff
@@ -32,6 +32,7 @@ const (
 	capabilityTaskIAMRoleNetHost                = "task-iam-role-network-host"
 	taskENIAttributeSuffix                      = "task-eni"
 	taskENIBlockInstanceMetadataAttributeSuffix = "task-eni-block-instance-metadata"
+	appMeshAttributeSuffix                      = "app-mesh"
```
yumex93: can you update this to aws-appmesh?

agent/app/agent_capability.go
```diff
@@ -69,6 +70,7 @@ const (
 //    ecs.capability.docker-volume-driver.${driverName}
 //    ecs.capability.task-eni
 //    ecs.capability.task-eni-block-instance-metadata
+//    ecs.capability.app-mesh
```
yumex93: same here. name should be 'aws-appmesh'

agent/ecscni/types.go
```diff
@@ -41,13 +44,16 @@ const (
 	ECSBridgePluginName = "ecs-bridge"
 	// ECSENIPluginName is the binary of the eni plugin
 	ECSENIPluginName = "ecs-eni"
+	// ECSAppMeshPluginName is the binary of app mesh plugin
+	ECSAppMeshPluginName = "app-mesh"
```
yumex93: The name should be 'aws-appmesh'

agent/api/appmesh/appmesh.go
```diff
@@ -0,0 +1,87 @@
...
+	proxyEgressPort    = "ProxyEgressPort"
+	appPorts           = "AppPorts"
+	egressIgnoredIPs   = "EgressIgnoredIPs"
+	egressIgnoredPorts = "EgressIgnoredPorts"
+)
+
+// AppMesh contains information of app mesh config
+type AppMesh struct {
+	// App Mesh CNI enabled
+	AppMeshEnabled bool
```
sharanyad: why exactly do we need this field? 

couldn't we find out if a task is appmesh enabled if AppMesh != nil?

agent/app/agent_unix.go
```diff
@@ -148,13 +149,25 @@ func isInstanceLaunchedInVPC(err error) bool {
...
 	// Check if we can get capabilities from each plugin
 	for _, plugin := range awsVPCCNIPlugins {
 		capabilities, err := agent.cniClient.Capabilities(plugin)
 		if err != nil {
 			return err
 		}
+		if plugin == ecscni.ECSAppMeshPluginName {
+			// check if cni plugin supports app mesh
+			if !contains(capabilities, ecscni.CapabilityAppMeshSupport) {
+				agent.cfg.AppMeshEnabled = false
```
sharanyad: why is `AppMeshEnabled` part of config? 

can it be part of agent struct itself?

agent/acs/handler/payload_handler.go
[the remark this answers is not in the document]
suneyz: I agree, I grabbed this naming convention from [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/eni/eni.go). I'd prefer to keep our consistent, we should follow up to unify them into maybe xxxConfig? How do you think?

agent/app/agent_capability.go
suneyz: updated

agent/app/agent_capability.go
suneyz: updated

agent/api/appmesh/appmesh.go
suneyz: make sense, i can check nil to determine app mesh is enabled

agent/ecscni/types.go
suneyz: updated

yumex93:
I think since there is model change on service side, you also need to update model files under agent/ecs_client folder

agent/app/agent_unix.go
suneyz: this is already removed since yumeng removed capability from cni plugin

agent/ecscni/plugin.go
```diff
@@ -125,6 +125,15 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "cni setup: invoke bridge plugin failed")
 	}
+	if cfg.AppMeshCNIEnabled {
+		// Invoke app mesh plugin ADD command
+		seelog.Debug("[APPMESH] Starting appmesh setup done")
```
aaithal: `"Starting appmesh setup done"` doesn't read right.

agent/ecscni/types.go
```diff
@@ -124,6 +129,28 @@ type ENIConfig struct {
...
 }
 
+// AppMeshConfig contains all the information needed to invoke the app mesh plugin
+type AppMeshConfig struct {
+	// Type is the cni plugin name
+	Type string `json:"type,omitempty"`
+	// CNIVersion is the cni spec version to use
+	CNIVersion string `json:"cniVersion,omitempty"`
+	// Egress traffic from the processes owned by the UID will be ignored
+	IgnoredUID string `json:"ignoredUID,omitempty"`
```
aaithal: Commentary for these fields is still not following golang convention.

agent/app/agent_unix.go
```diff
@@ -148,13 +149,18 @@ func isInstanceLaunchedInVPC(err error) bool {
...
 // c. ecs-ipam
+// d. aws-appmesh
 func (agent *ecsAgent) verifyCNIPluginsCapabilities() error {
 	// Check if we can get capabilities from each plugin
 	for _, plugin := range awsVPCCNIPlugins {
 		capabilities, err := agent.cniClient.Capabilities(plugin)
 		if err != nil {
 			return err
 		}
+		// appmesh plugin does not contain aws vpc network capability
```
aaithal: I think a more appropriate comment would be: `// appmesh plugin is not needed for awsvpc networking capability`

agent/api/appmesh/appmesh.go
```diff
@@ -0,0 +1,87 @@
...
+	ProxyEgressPort string
+	// AppPorts is the port number that application is listening on
+	AppPorts []string
+	// EgressIgnoredIPs is the list of ports for which egress traffic will be ignored
+	EgressIgnoredIPs []string
+	// EgressIgnoredPorts is the list of IPs for which egress traffic will be ignored
+	EgressIgnoredPorts []string
+}
+
+// AppMeshFromACS validates proxy config if it is app mesh type and create AppMesh object
```
yumex93: nit: creates AppMesh object

agent/api/appmesh/appmesh.go
```diff
@@ -0,0 +1,87 @@
...
+func AppMeshFromACS(proxyConfig *ecsacs.ProxyConfiguration) (*AppMesh, error) {
+
+	if proxyConfig == nil {
+		return nil, nil
+	}
+
+	if *proxyConfig.Type != appMesh {
+		return nil, fmt.Errorf("agent does not support proxy type other than app mesh")
```
yumex93: The function name is AppMeshFromACS, should we check the proxyConfig.Type outside this function? Like only when proxyConfig isn't nil and Type == appMesh, we call this AppMeshFromACS. Otherwise, we return the error. I don't feel it is the right place to have the validation of type here.

yumex93: If you decide to move the validation outside the function, remember to resolve aaithal@ comment and remove error in return

agent/app/agent_capability.go
```diff
@@ -138,6 +140,9 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	// ecs agent version 1.23.0 supports ecs secrets integrating with aws secrets manager
 	capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+capabilitySecretEnvASM)
 
+	// ecs agent version 1.26.0 supports appmesh cni plugin
```
yumex93: supports aws-appmesh cni plugin?

agent/ecscni/plugin.go
```diff
@@ -125,6 +125,15 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
...
+		result, err = client.add(runtimeConfig, cfg, client.createAppMeshConfig)
+		if err != nil {
+			return nil, errors.Wrap(err, "cni setup: invoke app mesh plugin failed")
+		}
+		seelog.Debugf("[APPMESH] appmesh setup done: %s", result.String())
```
yumex93: nit: [APPMESH] Set up aws-appmesh done.

agent/ecscni/plugin.go
```diff
@@ -125,6 +125,15 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "cni setup: invoke bridge plugin failed")
 	}
+	if cfg.AppMeshCNIEnabled {
+		// Invoke app mesh plugin ADD command
+		seelog.Debug("[APPMESH] Starting appmesh setup done")
+		result, err = client.add(runtimeConfig, cfg, client.createAppMeshConfig)
```
yumex93: how about change to result, err := client .... So we only keep the result and err within the if block, since it only matters within the block.

yumex93:
Why the functional tests and arm integ tests failed?

agent/acs/handler/payload_handler.go
suneyz: we discussed offline, this is referenced from other places in this file. we may want to turn them into this format in a separate CR

agent/api/appmesh/appmesh.go
suneyz: I agree if we have multiple proxy support we should not validate here(Prob not in payload handler, we should prob have some sort of proxy constructor file). But I intentionally put the appmesh type check here to make it clear in agent proxy configuration is strongly typed with appmesh struct . I am thinking if in foreseeable future we won't have any other proxy, appmesh struct is strongly typed with proxy configuration struct. we should make it clear in the code and in a centralized place when we construct appmesh struct, we should throw exceptions instead of same logic for appmesh spanned in multiple files. How do you think?

agent/app/agent_capability.go
suneyz: updated

agent/app/agent_unix.go
suneyz: updated

agent/ecscni/plugin.go
suneyz: remove "done"

agent/ecscni/plugin.go
suneyz: Can you confirm appmesh plugin will return a non nil result? It seems during our testing appmesh plugin result caused agent to exit 2

agent/ecscni/plugin.go
suneyz: updated

agent/ecscni/types.go
suneyz: My bad, updated few configs from earlier comment, missed this config. Updated.

agent/api/appmesh/appmesh.go
suneyz: updated

suneyz:
> Why the functional tests and arm integ tests failed?

I updated the summary, functional tests failed as expected since we are missing binaries. Integ test failed on this https://github.com/aws/amazon-ecs-agent/issues/1855

agent/ecscni/plugin.go
```diff
@@ -180,11 +189,18 @@ func (client *cniClient) cleanupNS(cfg *Config) error {
...
 	}
 	seelog.Debugf("[ECSCNI] bridge cleanup done: %s", cfg.ContainerID)
-
+	// clean up eni network namespace
 	err = client.del(runtimeConfig, cfg, client.createENINetworkConfig)
 	if err != nil {
 		return errors.Wrap(err, "cni cleanup: invoke eni plugin failed")
 	}
+	if cfg.AppMeshCNIEnabled {
+		// clean up app mesh network namespace
```
yumex93: Could you add a debug log for AppMesh so we can know whether it is called or not?

suneyz: added

agent/app/agent_unix.go
```diff
@@ -48,6 +48,7 @@ const initPID = 1
 var awsVPCCNIPlugins = []string{ecscni.ECSENIPluginName,
 	ecscni.ECSBridgePluginName,
 	ecscni.ECSIPAMPluginName,
+	ecscni.ECSAppMeshPluginName,
```
aaithal: `ECSAppMeshPluginName` shouldn't be a part of this list. It's not needed for the `awsvpc` mode. However, what you do need is a way to check if the plugin binary file exists or not. Here's an example: https://github.com/aws/amazon-ecs-agent/blob/9708e515dd932f7b8d9a0a1be9046c82b92d89a0/agent/app/agent_capability.go#L249

suneyz: Good point. I actually tried a bit and i noticed our currently testing logic in agent capabilities needs quite effort to refactor to make it work for both linux and windows. Discussed offline, I created a issue https://github.com/aws/amazon-ecs-agent/issues/1893 to work as follow up

suneyz:
pushing to app-mesh branch to combine with binary compilation. Will have another pr to merge to dev once done to make sure tests are passed

PR #1826, opened 2019-01-31
ENI Trunking - confirm Trunk ENI attachment

fenxiong:
### Summary

Add functionality in Agent to confirm Trunk ENI attachment after instance registration.

### Implementation details

A high level workflow to confirm trunk eni attachment is as follow:
1. An ACS handler, AttachInstanceENIHandler receives an AttachNetworkInterfacesMessage from ACS which contains the information of the trunk ENI. upon receiving this message the handler adds the trunk eni attachment to the attachment table in Agent's state;
2. Udev Watcher is periodically scanning through attached network devices on the instances and also listens to network device's attached event. Once it found that the trunk ENI is attached (it knows a device is the trunk ENI from information in attachment table added in step 1), it sends an AttachmentStateChange event to the global state change event channel;
3. An event handler, AttachmentEventHandler handles the AttachmentStateChange event from the event channel by sending it to backend.

Changes involve the following parts: 
#### Data Structure Changes
Added a new field "attachmentType" to existing ENI attachment struct to indicate the type of ENI attachment (regular or trunk).
A new type of state change, AttachmentStateChange is added to indicate the attach state change of the Trunk ENI.

#### ACS
Adds a new ACS handler AttachInstanceENIHandler in ACS client which handles a new type of ACS message, instance ENI attach message, from which backend tells Agent to start watching the attach status of the Trunk ENI.
The existing AttachENIHandler is renamed to AttachTaskENIHandler since it's only handling task's ENI in a regular awsvpc task. File name/tests adjusted accordingly. Common logic between AttachTaskENIHandler and AttachInstanceENIHandler is extracted to agent/acs/handler/attach_eni_handler_common.go to avoid duplicate code.

#### Event Handler
A new handler, AttachmentEventHandler is added to eventhandler package to handle the new AttachmentStateChange event. it contains a map of attachment handlers each responsible for handling a certain attachment.

#### Udev Watcher
Existing udev watcher is modified so that it sends different types of event for different types of ENI attachment. Regular ENI attachment gets attached results in a TaskStateChange and Trunk ENI attachment gets attached results in an AttachmentStateChange.

### Testing

Unit tests added.
Manually tested this by injecting [test code](https://github.com/fenxiong/amazon-ecs-agent/commit/63be2e0a304b33e2448bc6b58c51a311a84fdaa9) (not checked in to PR) in Agent that sends a test AttachInstancNetworkInterfacesMessage ACS message to AttachInstanceENIHandler that contains information of an ENI that's already attached to the instance, and verifies that the workflow is triggered as expected:
```
...
2019-02-13T01:19:46Z [INFO] Sending test acs message: {
  ClusterArn: "testClusterArn",
  ContainerInstanceArn: "testContainerInstanceArn",
  ElasticNetworkInterfaces: [{
      AttachmentArn: "testAttachmentArn",
      MacAddress: "06:45:e4:b0:db:ca"
    }],
  MessageId: "testMessageId",
  WaitTimeoutMs: 60000
}
2019-02-13T01:19:46Z [INFO] Sent test acs message
2019-02-13T01:19:46Z [INFO] Handling test acs message, skip sending ack
2019-02-13T01:19:46Z [INFO] Handling ENI attachment: testAttachmentArn
2019-02-13T01:19:46Z [INFO] Starting ENI ack timer with duration=59.999988789s, ENI Attachment: attachment=testAttachmentArn;attachmentType=trunk-eni;attachmentSent=false;mac=06:45:e4:b0:db:ca;status=NONE;expiresAt=2019-02-13 01:20:46.397439989 +0000 UTC m=+60.310384986
2019-02-13T01:19:46Z [INFO] Adding trunk eni attachment info to state, attachment=testAttachmentArn mac=06:45:e4:b0:db:ca
...
2019-02-13T01:20:16Z [INFO] Emitting instance ENI change event for: ENI Attachment: attachment=testAttachmentArn;attachmentType=trunk-eni;attachmentSent=false;mac=06:45:e4:b0:db:ca;status=ATTACHED;expiresAt=2019-02-13 01:20:46.397439989 +0000 UTC m=+60.310384986
2019-02-13T01:20:16Z [DEBUG] AttachmentHandler: acquiring attachment lock before sending attachment state change for attachment testAttachmentArn
2019-02-13T01:20:16Z [DEBUG] AttachmentHandler: acquired attachment lock for attachment testAttachmentArn
2019-02-13T01:20:16Z [INFO] AttachmentHandler: sending attachment state change: testAttachmentArn -> ATTACHED, ENI Attachment: attachment=testAttachmentArn;attachmentType=trunk-eni;attachmentSent=false;mac=06:45:e4:b0:db:ca;status=ATTACHED;expiresAt=2019-02-13 01:20:46.397439989 +0000 UTC m=+60.310384986
2019-02-13T01:20:16Z [INFO] SubmitAttachmentStateChange: handling attachment state change for test attachment, would have sent: {
  Attachment: {
    AttachmentArn: "testAttachmentArn",
    Status: "ATTACHED"
  },
  Cluster: "test-eni-trunking"
}
2019-02-13T01:20:16Z [INFO] Saving state! module="statemanager"
2019-02-13T01:20:16Z [DEBUG] AttachmentHandler: submitted attachment state change: testAttachmentArn -> ATTACHED, ENI Attachment: attachment=testAttachmentArn;attachmentType=trunk-eni;attachmentSent=true;mac=06:45:e4:b0:db:ca;status=ATTACHED;expiresAt=2019-02-13 01:20:46.397439989 +0000 UTC m=+60.310384986
```
(Note: in above manual test there are about 30 seconds delay before the watcher notices the attached status of the ENI because the ENI is already attached and will only be detected by the periodic reconciliation loop that happens every 30 seconds. In reality attaching the ENI should trigger the watcher to notice it immediately)

### Licensing


fenxiong:
Windows integ test timed out - tracked in https://github.com/aws/amazon-ecs-agent/issues/1835
Linux/Arm functional tests failed because of insufficient memory:
```
...error transitioning container [exit] to [RUNNING]: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "process_linux.go:402: container init caused \"process_linux.go:367: setting cgroup config for procHooks process caused \\\"failed to write 10485760 to memory.limit_in_bytes: write /sys/fs/cgroup/memory/ecs/00e11e24-1a4e-45c1-b7f5-3e17d27e0ee5/2e217d407c0dd27a5616e0fb9d94f22e35b306e16e0b3fc832b38aedee569d99/memory.limit_in_bytes: device or resource busy\\\"\"": unknown
```
seems like https://github.com/aws/amazon-ecs-agent/pull/1842 should fix this..

adnxn:
> seems like #1842 should fix this..

wait - did it fix your branch?

agent/acs/handler/attach_instance_eni_handler.go
```diff
@@ -0,0 +1,155 @@
...
+
+// stop is used to invoke a cancellation function
+func (handler *attachInstanceENIHandler) stop() {
+	handler.cancel()
+}
+
+// handleMessages handles each message one at a time
+func (handler *attachInstanceENIHandler) handleMessages() {
+	for {
+		select {
```
aaithal: You should also read `ctx.Done` channel here and exit the method if the context is done/cancelled. Not doing that leads to go routine leaks, which can cause problems in a long running agent.

agent/api/ecsclient/client.go
```diff
@@ -482,6 +482,31 @@ func (client *APIECSClient) SubmitContainerStateChange(change api.ContainerState
 	return nil
 }
 
+func (client *APIECSClient) SubmitAttachmentStateChange(change api.AttachmentStateChange) error {
+	if change.Attachment == nil { // this should not happen unless there's a bug
```
aaithal: If that's a bug, shouldn't you just panic and crash/stop the agent? Just logging the error message will hide the occurrence of that bug. If you do not want to invoke `panic` yourself, might be better to not do the check at all and have a NPE crashing the agent.

agent/acs/handler/attach_instance_eni_handler_test.go
```diff
@@ -0,0 +1,319 @@
...
+	ackSent.Add(1)
+	mockWSClient.EXPECT().MakeRequest(gomock.Any()).Do(func(ackRequest *ecsacs.AckRequest) {
+		assert.Equal(t, aws.StringValue(ackRequest.MessageId), eniMessageId)
+		ackSent.Done()
+	})
+	gomock.InOrder(
+		// Sending an attachment with ExpiresAt set in the past results in an
+		// error in starting the timer.
+		// Ensuring that statemanager.Save() is not invoked should be a strong
+		// enough check to ensure that the timer was started (since StartTimer would be
```
adnxn: 🤔 this sounds reasonable

agent/api/eni/eniattachment.go
```diff
@@ -23,8 +23,15 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	ENIAttachmentTypeENI      = "eni"
```
aaithal: lint will tell you that you need to document these global scoped const's

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,134 @@
...
+
+	go handler.submitAttachmentEvent(&event)
+	return nil
+}
+
+// submitAttachmentEvent submits an attachment state change to backend
+func (handler *AttachmentHandler) submitAttachmentEvent(attachmentChange *api.AttachmentStateChange) {
+	attachmentARN := attachmentChange.Attachment.AttachmentARN
+
+	var atmLock *sync.Mutex // the lock for the attachment
```
aaithal: This seems like a recipe for a future deadlock. I'd strongly suggest evaluating alternatives. One of which could be that instead of a single handler handling all attachments, you have a handler per attachment so that you don't have acquire 2 levels of locks. We've have multiple deadlocks in the agent in the past because of this pattern.

agent/acs/handler/attach_task_eni_handler_test.go
```diff
@@ -323,7 +323,7 @@ func TestENIAckHappyPath(t *testing.T) {
 	manager := mock_statemanager.NewMockStateManager(ctrl)
 
 	mockWSClient := mock_wsclient.NewMockClientServer(ctrl)
-	eniAttachHandler := newAttachENIHandler(ctx, clusterName, containerInstanceArn, mockWSClient, taskEngineState, manager)
+	eniAttachHandler := newAttachTaskENIHandler(ctx, clusterName, containerInstanceArn, mockWSClient, taskEngineState, manager)
```
adnxn: question: so for the tests you removed from this file - it seem like you're covering all the cases with the instance type eni yea?

aaithal:
A general comment would also be to make commit messages more meaningful. Someone looking at this code in their desktops wouldn't necessary open up the PR link. But, they'll definitely appreciate more meaningful, explanatory commit messages.

agent/api/eni/eniattachment.go
adnxn: missing doc lines

agent/eni/watcher/watcher_linux.go
```diff
@@ -214,17 +214,31 @@ func (udevWatcher *UdevWatcher) sendENIStateChange(mac string) error {
...
+		go udevWatcher.emitInstanceENIChangeEvent(eni)
+	} else {
+		go udevWatcher.emitTaskENIChangeEvent(eni)
+	}
 	return nil
 }
 
+// emitTaskENIChangeEvent sends a state change event for a task ENI attachment to the event channel
+func (udevWatcher *UdevWatcher) emitTaskENIChangeEvent(eni *apieni.ENIAttachment) {
+	eni.Status = apieni.ENIAttached
```
adnxn: so this method is only for submitting an `ENIAttached` state change message? seems that way- we should rename the method accordingly then.

agent/eventhandler/attachment_handler.go
adnxn: hm did you run `go vet`? does it complain about this `atmLock`?

agent/eventhandler/handler.go
```diff
@@ -31,11 +35,24 @@ func HandleEngineEvents(taskEngine engine.TaskEngine, client api.ECSClient, even
 					seelog.Error("Unable to handle state change event. The events channel is closed")
 					break
 				}
-				err := eventhandler.AddStateChangeEvent(event, client)
+
+				err := handleEngineEvent(event, client, taskHandler, attachmentHandler)
```
adnxn: nit: empty lines around this

agent/eventhandler/attachment_handler.go
fenxiong: refactored the attachment handler. now it has a map of attachment handlers each handling a certain attachment and has its own lock. there are still 2 level of locks but they won't be acquired together or affecting each other.

agent/api/ecsclient/client.go
fenxiong: removed this check

agent/acs/handler/attach_instance_eni_handler.go
fenxiong: not sure if i understand.. i already listen to handler.ctx.Done in the following lines

agent/api/eni/eniattachment.go
fenxiong: there is currently a large amount of output from golint. fixed the ones brought by my changes including this

agent/acs/handler/attach_eni_handler_common.go
```diff
@@ -0,0 +1,105 @@
...
+		AttachStatusSent: false,
+		MACAddress:       mac,
+		ExpiresAt:        expiresAt, // Stop tracking the eni attachment after timeout
+	}
+	eniAckTimeoutHandler := ackTimeoutHandler{mac: mac, state: state}
+	if err := eniAttachment.StartTimer(eniAckTimeoutHandler.handle); err != nil {
+		return err
+	}
+
+	if attachmentType == apieni.ENIAttachmentTypeENI {
```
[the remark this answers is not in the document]
fenxiong: updated

agent/acs/handler/attach_eni_handler_common_test.go
```diff
@@ -0,0 +1,116 @@
...
+func TestTaskENIAckTimeout(t *testing.T) {
+	testENIAckTimeout(t, apieni.ENIAttachmentTypeENI)
+}
+
+// TestTrunkENIAckTimeout tests acknowledge timeout for a trunk eni before submit the state change
+func TestTrunkENIAckTimeout(t *testing.T) {
+	testENIAckTimeout(t, apieni.ENIAttachmentTypeTrunkENI)
+}
+
+func testENIAckTimeout(t *testing.T, attachmentType string) {
```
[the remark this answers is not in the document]
fenxiong: not sure if i understand.. which part do you think that should be in a table?

agent/acs/handler/attach_instance_eni_handler.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// newAttachInstanceENIHandler returns an instance of the attachInstanceENIHandler struct
+func newAttachInstanceENIHandler(ctx context.Context,
+	cluster string,
+	containerInstanceArn string,
+	acsClient wsclient.ClientServer,
+	taskEngineState dockerstate.TaskEngineState,
+	saver statemanager.Saver) attachInstanceENIHandler {
+
```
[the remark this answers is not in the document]
fenxiong: fixed

agent/acs/handler/attach_instance_eni_handler.go
```diff
@@ -0,0 +1,155 @@
...
+	eni := enis[0]
+	if aws.StringValue(eni.MacAddress) == "" {
+		return errors.Errorf("MACAddress not provided")
+	}
+
+	timeout := aws.Int64Value(message.WaitTimeoutMs)
+	if timeout <= 0 {
+		return errors.Errorf("invalid timeout specified: %d", timeout)
+
```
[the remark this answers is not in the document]
fenxiong: fixed

agent/acs/handler/attach_instance_eni_handler_test.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
+	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate/mocks"
+	"github.com/aws/amazon-ecs-agent/agent/statemanager/mocks"
+	"github.com/aws/amazon-ecs-agent/agent/wsclient/mock"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestAttachInstanceENIMessageWithNoMessageId checks the validator against an
```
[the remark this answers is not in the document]
fenxiong: put them into a table

agent/acs/handler/attach_task_eni_handler_test.go
fenxiong: yes

agent/api/eni/eniattachment.go
fenxiong: fixed

agent/eni/watcher/watcher_linux.go
fenxiong: updated to emitTaskENIAttachedEvent, same for emitInstanceAttachedEvent

agent/eventhandler/handler.go
fenxiong: fixed

agent/eventhandler/attachment_handler.go
fenxiong: i think travis runs it and it passes? anyway removed this as per other comment

agent/acs/handler/attach_instance_eni_handler.go
aaithal: Sorry, I missed that. The convention is usually to have that as the first case statement. Probably why I missed that.

Having said that, even `sendAck` should handle the context being cancelled. Otherwise, that go routine can leak as well.

agent/acs/handler/attach_eni_handler_common.go
```diff
@@ -0,0 +1,109 @@
...
+		seelog.Warnf("Failed to ack request with messageId: %s, error: %v", aws.StringValue(messageId), err)
+	}
+}
+
+// handleENIAttachment handles an ENI attachment via the following:
+// 1. Check whether we already have this attachment in state, if so, start its ack timer and return
+// 2. Otherwise add the attachment to state, start its ack timer, and save the state
+// These are common tasks for handling a task ENI attachment and an instance ENI attachment, so they are put
+// into this function to be shared by both attachment handlers
+func handleENIAttachment(attachmentType, attachmentARN, taskARN, mac string, expiresAt time.Time, state dockerstate.TaskEngineState, saver statemanager.Saver) error {
```
aaithal: Can you please split this into multiple lines? Makes it easier to read.

agent/acs/handler/attach_eni_handler_common.go
```diff
@@ -0,0 +1,109 @@
...
+// ackTimeoutHandler remove ENI attachment from agent state after the ENI ack timeout
+type ackTimeoutHandler struct {
+	mac   string
+	state dockerstate.TaskEngineState
+}
+
+func (handler *ackTimeoutHandler) handle() {
+	eniAttachment, ok := handler.state.ENIByMac(handler.mac)
+	if !ok {
+		seelog.Warnf("Timed out waiting for ENI ack; ignoring unmanaged ENI attachment with MAC address: %s", handler.mac)
```
aaithal: Why does this mean a time out?

agent/acs/handler/attach_eni_handler_common.go
```diff
@@ -0,0 +1,109 @@
...
+		return
+	}
+	if !eniAttachment.IsSent() {
+		seelog.Infof("Timed out waiting for ENI ack; removing ENI attachment record with MAC address: %s", handler.mac)
```
aaithal: Why is this info where as the previous one was a warn?

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,150 @@
...
+
+		seelog.Infof("AttachmentHandler: sending attachment state change: %s", attachmentChange.String())
+		if err := handler.client.SubmitAttachmentStateChange(*attachmentChange); err != nil {
+			seelog.Errorf("AttachmentHandler: error submitting attachment state change [%s]: %v", attachmentChange.String(), err)
+			return err
+		}
+
+		attachmentChange.Attachment.SetSentStatus()
+		attachmentChange.Attachment.StopAckTimer()
+		handler.stateSaver.Save()
```
aaithal: Doesn't this method return an error? Shouldn't that be handled as well?

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// attachmentChangeShouldBeSent checks whether an attachment state change should be sent to backend
+func attachmentChangeShouldBeSent(attachmentChange *api.AttachmentStateChange) bool {
+	return !attachmentChange.Attachment.HasExpired() && !attachmentChange.Attachment.IsSent()
+}
+
+// GetSubmitAttachmentStateBackoffFunc is a func that returns a backoff object;
+// this is made as a function object so that we can use smaller backoff in unit test
+var GetSubmitAttachmentStateBackoffFunc = GetSubmitAttachmentStateBackoff
```
aaithal: a. does this need to be a global? 
b. can you please move all var's and const's to the top of the file?
c. it seems like you're using this to override in your tests. That's an anti-pattern as well. Having global variables that get overridden in tests leads to flaky tests. Can you make that a part of the `attachmentHandler` struct or something and override it in the test package when you construct the object?

agent/eventhandler/handler.go
```diff
@@ -31,11 +37,22 @@ func HandleEngineEvents(taskEngine engine.TaskEngine, client api.ECSClient, even
...
 					seelog.Errorf("Handler unable to add state change event %v: %v", event, err)
 				}
 			}
 		}
 	}
 }
+
+func handleEngineEvent(event statechange.Event, client api.ECSClient, taskHandler *TaskHandler,
+	attachmentEventHandler *AttachmentEventHandler) error {
+	if event.GetEventType() == statechange.TaskEvent || event.GetEventType() == statechange.ContainerEvent {
```
aaithal: switch-case might be a better alternative here to improve readability.

agent/eventhandler/attachment_handler.go
fenxiong: made the backoff as a field of attachmentHandler and overridden it in test

agent/eventhandler/attachment_handler.go
fenxiong: added logic to handle state saving error here and added a unit test for it

agent/eventhandler/handler.go
fenxiong: changed to switch-case

agent/acs/handler/attach_eni_handler_common.go
fenxiong: both should be warning. fixed

agent/acs/handler/attach_eni_handler_common.go
fenxiong: i don't think it should.. fixed

agent/acs/handler/attach_eni_handler_common.go
fenxiong: done

agent/acs/handler/attach_instance_eni_handler.go
fenxiong: Moved ctx.Done to first case statement. Regards to handling context in sendAck, I'm not sure how to do this since it just calls a function and returns rather than keeps doing something in a loop. Also the underlying vendor package gorilla/websocket that it uses does not handle context. Do you have a suggestion on how this can be done?

aaithal: I see. As far as `sendAck` itself is concerned, it's probably fine then. Note that we set read and write timeouts on the websocket channel itself to deal with `MakeRequest` taking too long.

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,158 @@
...
+// submitAttachmentEvent submits an attachment event to backend
+func (handler *attachmentHandler) submitAttachmentEvent(attachmentChange *api.AttachmentStateChange) {
+	seelog.Debugf("AttachmentHandler: acquiring attachment lock before sending attachment state change for attachment %s", handler.attachmentARN)
+	handler.lock.Lock()
+	seelog.Debugf("AttachmentHandler: acquired attachment lock for attachment %s", handler.attachmentARN)
+	defer handler.lock.Unlock()
+
+	// we need to continue retrying but skip submitting attachment state change when we successfully submitted the state
+	// change but failed to save state
+	skipSubmitting := false
```
aaithal: Just to clarify, `skipSubmitting` will be used to retry the `state.Save()` operation if there's a failure to save state the first time, yeah? If that is true, I think we should handle it in a different way. If there's an error saving state, more often than not, it means that we've run out of space on the disk. I can't think of many possibilities where saving state fails for the first time, but succeeds the second time. 

With that in mind, a better way to handle this might be retry it may be 3 times and then give up so that we don't hold `handler.lock` for too long here. 

A related question here is: should we attempt to save the state while holding this lock? Can that happen after the fact, without holding the lock?

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,158 @@
...
+// to backend
+type AttachmentEventHandler struct {
+	// stateSaver is a statemanager which may be used to save any
+	// changes to an attachment's SentStatus
+	stateSaver statemanager.Saver
+
+	// backoff is the backoff object used in submitting attachment state change
+	backoff retry.Backoff
+
+	// attachmentLock is a map from attachment ARN to the attachmentHandler that is
```
aaithal: outdated comment. Can you please update this?

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,158 @@
...
+		stateSaver:             stateSaver,
+		client:                 client,
+		attachmentARNToHandler: make(map[string]*attachmentHandler),
+		backoff: retry.NewExponentialBackoff(submitStateBackoffMin, submitStateBackoffMax,
+			submitStateBackoffJitterMultiple, submitStateBackoffMultiple),
+	}
+}
+
+// AddStateChangeEvent adds a state change event to AttachmentEventHandler for it to handle
+func (handler *AttachmentEventHandler) AddStateChangeEvent(change statechange.Event) error {
```
aaithal: can you rename this to be `eventHandler`? Makes it less confusing to read the code.

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,158 @@
...
+	attachmentHandler := handler.attachmentARNToHandler[attachmentARN]
+	go attachmentHandler.submitAttachmentEvent(&event)
+
+	return nil
+}
+
+// submitAttachmentEvent submits an attachment event to backend
+func (handler *attachmentHandler) submitAttachmentEvent(attachmentChange *api.AttachmentStateChange) {
+	seelog.Debugf("AttachmentHandler: acquiring attachment lock before sending attachment state change for attachment %s", handler.attachmentARN)
+	handler.lock.Lock()
```
aaithal: Just for my understanding, why should this lock be acquired here? It'd be awesome if you could add a comment regarding that as well.

agent/statemanager/state_manager.go
```diff
@@ -79,6 +79,7 @@ const (
 	//   b) Add 'GPUIDs' field to 'apicontainer.Container'
 	//   c) Add 'NvidiaRuntime' field to 'api.task.task'
 	// 20)  Add 'target' field to the Secret struct
+	// 21) Add 'attachmentType' field to 'api.ENIAttachment'
```
aaithal: Comment says `21`. The version is still `20`. What's the mismatch?

agent/eventhandler/attachment_handler.go
fenxiong: i think the udev watcher could send multiple state change events for a certain attachment, for example, one in reconciliation loop and one from udev event. in that case the lock for the attachment handler can prevent sending attachment state change for an attachment multiple times. added comment for this.

agent/statemanager/state_manager.go
fenxiong: missed this during rebase. updated data version to 21.

agent/eventhandler/attachment_handler.go
fenxiong: updated

agent/eventhandler/attachment_handler.go
fenxiong: updated

agent/eventhandler/attachment_handler.go
fenxiong: > Just to clarify, `skipSubmitting` will be used to retry the `state.Save()` operation if there's a failure to save state the first time, yeah?
Yes that's what i'm intended to do.

> A related question here is: should we attempt to save the state while holding this lock? Can that happen after the fact, without holding the lock?
I don't think we need to hold the lock when attempting to save state here. Updated to do this in a separate go routine and only retry 3 times for it

agent/api/statechange.go
```diff
@@ -141,6 +148,13 @@ func NewContainerStateChangeEvent(task *apitask.Task, cont *apicontainer.Contain
 	return event, nil
 }
 
+// NewAttachmentStateChangeEvent creates a new attachment state change event
+func NewAttachmentStateChangeEvent(eniAttachment *apieni.ENIAttachment) AttachmentStateChange {
```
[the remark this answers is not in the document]
fenxiong: because each state change (TaskStateChange, ContainerStateChange, AttachmentStateChange) implements the [Event](https://github.com/aws/amazon-ecs-agent/blob/master/agent/statechange/statechange.go#L31) interface and is treated as a event when it's sent to the event channel.. container state change and task state change are also constructed similarly in NewContainerStateChangeEvent and NewTaskStateChangeEvent

agent/acs/handler/attach_instance_eni_handler.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+}
+
+// handleSingleMessage acks the message received
+func (handler *attachInstanceENIHandler) handleSingleMessage(message *ecsacs.AttachInstanceNetworkInterfacesMessage) error {
+	receivedAt := time.Now()
+	// Validate fields in the message
+	if err := validateAttachInstanceNetworkInterfacesMessage(message); err != nil {
+		return errors.Wrapf(err,
+			"attach instance eni message handler: error validating AttachInstanceNetworkInterfacesMessage received from ACS")
```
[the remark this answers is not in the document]
fenxiong: removed "received from ACS" here

agent/eventhandler/attachment_handler.go
aaithal: I see. The `attachmentARNToHandler ` hashmap helps us with deduping that, yeah?

fenxiong: yes, with that map, events among different attachments are handled concurrently while events for same attachment is handled serially (therefore won't be sent multiple times)

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/aws/amazon-ecs-agent/agent/api"
+	"github.com/aws/amazon-ecs-agent/agent/statechange"
+	"github.com/aws/amazon-ecs-agent/agent/statemanager"
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
+	"github.com/cihub/seelog"
+)
+
+const (
+	// savingStateRetryNum is the number of retries we have when we successfully submitted an attachment
```
aaithal: nit: `saveAttachmentStateRetryNum is the ..`

agent/eventhandler/attachment_handler.go
```diff
@@ -0,0 +1,172 @@
...
+	// multiple times (this can happen when udev watch sends multiple attached events for a certain attachment,
+	// for example one from udev event and one from reconciliation loop)
+	seelog.Debugf("AttachmentHandler: acquiring attachment lock before sending attachment state change for attachment %s", handler.attachmentARN)
+	handler.lock.Lock()
+	seelog.Debugf("AttachmentHandler: acquired attachment lock for attachment %s", handler.attachmentARN)
+	defer handler.lock.Unlock()
+
+	// we need to continue retrying but skip submitting attachment state change when we successfully submitted the state
+	// change but failed to save state
+	retry.RetryWithBackoffCtx(handler.ctx, handler.backoff, func() error {
```
aaithal: There are far too many anonymous functions in this block. Please make it a point to come back and rectify that (with named methods) in the future (by creating an issue).

fenxiong:
failing windows unit test: TestPullImageInactivityTimeout tracked in https://github.com/aws/amazon-ecs-agent/issues/1864
failing arm functional test: TestServerEndpointValidator tracked in https://github.com/aws/amazon-ecs-agent/issues/1891
arm integ test failed and seemed like docker was down in the middle of the test:
```
2019-02-26T04:57:29Z [INFO] Unable to get Docker client for version 1.17: version detection check: failed to ping with Docker version: 1.17: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2019-02-26T04:57:29Z [INFO] Unable to get Docker client for version 1.18: version detection check: failed to ping with Docker version: 1.18: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
2019-02-26T04:57:29Z [INFO] Unable to get Docker client for version 1.19: version detection check: failed to ping with Docker version: 1.19: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
...
```

agent/eventhandler/attachment_handler.go
fenxiong: updated

agent/eventhandler/attachment_handler.go
fenxiong: added a TODO for this

aaithal: Can you please an issue and add a link to that as well, along with the `TODO` ?

fenxiong: created the issue [here](https://github.com/aws/amazon-ecs-agent/issues/1894) and added the link in comment

Issue #1837, opened 2019-02-11
Unable to start Service: failed to pull image amazon/amazon-ecs-pause:0.1.0


jessecollier:
### Summary
Trying to start a daemon service (datadog agent) on ec2 ecs. However task is being marked as "Stopped" with no information in the UI.

### Description
Following the setup guide here: https://docs.datadoghq.com/integrations/amazon_ecs/?tab=python#aws-cli

When launching the new daemon service task, the tasks are unable to start. Here's the relevant log line which marks the service as STOPPED

```
2019-02-11T17:24:36Z [ERROR] Managed task [arn:aws:ecs:us-east-1:<acctid>:task/ecs-cluster-staging/<taskid>]: error while pulling image amazon/amazon-ecs-pause:0.1.0 for container ~internal~ecs~pause~namespace , moving task to STOPPED: Error response from daemon: pull access denied for amazon/amazon-ecs-pause, repository does not exist or may require 'docker login'
```

### Expected Behavior
Service task starts

### Observed Behavior
Marked as STOPPED for task id, however complains about an amazon-ecs-pause unable to be PULLED

### Environment Details
Docker Info
```
Containers: 3
 Running: 3
 Paused: 0
 Stopped: 0
Images: 4
Server Version: 18.06.1-ce
Storage Driver: devicemapper
 Pool Name: docker-docker--pool
 Pool Blocksize: 524.3kB
 Base Device Size: 10.74GB
 Backing Filesystem: ext4
 Udev Sync Supported: true
 Data Space Used: 3.906GB
 Data Space Total: 530.4GB
 Data Space Available: 526.5GB
 Metadata Space Used: 18.22MB
 Metadata Space Total: 536.9MB
 Metadata Space Available: 518.7MB
 Thin Pool Minimum Free Space: 53.04GB
 Deferred Removal Enabled: true
 Deferred Deletion Enabled: true
 Deferred Deleted Device Count: 0
 Library Version: 1.02.135-RHEL7 (2016-11-16)
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge host macvlan null overlay
 Log: awslogs fluentd gcplogs gelf journald json-file logentries splunk syslog
Swarm: inactive
Runtimes: runc
Default Runtime: runc
Init Binary: docker-init
containerd version: 468a545b9edcd5932818eb9de8e72413e616e86e
[...]

PR #1840, opened 2019-02-12
avoid pull behavior check for namespace pause image

sharanyad:
### Summary
https://github.com/aws/amazon-ecs-agent/issues/1837 

when `ECS_IMAGE_PULL_BEHAVIOR` is set to `once` and pause images are needed at the same time, agent will try to download pause image once and fail. this leads to the task failing. 

### Implementation details
avoid pull behavior check for pause image used for container namespaces setup 
### Testing

New tests cover the changes: 

Manual testing: 
set`ECS_IMAGE_PULL_BEHAVIOR=once` and use `pidMode:"task"` in task definition. 
Task fails to run without this code change and passes otherwise. 

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -667,8 +667,8 @@ func (engine *DockerTaskEngine) GetTaskByArn(arn string) (*apitask.Task, bool) {
 
 func (engine *DockerTaskEngine) pullContainer(task *apitask.Task, container *apicontainer.Container) dockerapi.DockerContainerMetadata {
 	switch container.Type {
-	case apicontainer.ContainerCNIPause:
-		// ContainerCNIPause image are managed at startup
+	case apicontainer.ContainerCNIPause, apicontainer.ContainerNamespacePause:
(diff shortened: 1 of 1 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -670,6 +670,10 @@ func (engine *DockerTaskEngine) pullContainer(task *apitask.Task, container *api
 	case apicontainer.ContainerCNIPause:
 		// ContainerCNIPause image are managed at startup
 		return dockerapi.DockerContainerMetadata{}
+
+	case apicontainer.ContainerNamespacePause:
```
petderek: Wait, can you put this on one line?

`case api.ContainerCNIPause | api.ContainerNamespacePause:`

sharanyad: done

petderek: 🤦 oops

use ',' not '|' .... '|' does entirely the wrong thing

sharanyad:
functional tests are failing due to low memory limit

Issue #1839, opened 2019-02-11
ECS agent no longer redirects http://169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI


samuelkarp:
### Summary
The ECS agent makes IAM roles available to tasks at an endpoint defined by a well-known IP and the `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` environment variable.  In some version after 1.20.0, the ECS agent serving some requests.

### Description
The `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is an environment variable defining the relative path where credentials should be queried from 169.254.170.2.  The content of the environment variable always starts with `/` and it is expected that requests are made to `http://169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`.  In older versions of the ECS agent, requests to `http://169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` (which contains an extra `/`) would also be served by redirecting them to the correct URL.  This appears to have changed.

### Expected Behavior
Requests to `http://169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` are served or redirected to `http://169.254.170.2$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`

### Observed Behavior
Requests to `http://169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` result in a 404

yunhee-l:
Investigation by Sharanya: 
did a binary search and found out we broke it in https://github.com/aws/amazon-ecs-agent/releases/tag/v1.21.0 
suspect this “Add v3 task metadata support for awsvpc, host and bridge network mode”

bzemms:
https://stackoverflow.com/questions/54614616/got-404-from-http-169-254-170-2-aws-container-credentials-relative-uri/54618368?noredirect=1#comment96069973_54618368

sharanyad:
this is now fixed with https://github.com/aws/amazon-ecs-agent/releases/tag/v1.25.3

PR #1844, opened 2019-02-13
Set router skip clean to false 

suneyz:
### Summary

This enables agent to handle request like "//v2/credential" correctly. This brings back the support for //path that agent version 1.20.3 or before
https://github.com/aws/amazon-ecs-agent/issues/1839

### Implementation details

set router skip clean to false
doc for skip clean https://godoc.org/github.com/gorilla/mux#Router.SkipClean

### Testing

New tests cover the changes: 

yes,

unit test, functional test

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/task-server-endpoint-validator/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/task-server-endpoint-validator/task-definition.json
@@ -1,0 +1,11 @@
+{
+    "family": "task-server-endpoint-validator",
+    "taskRoleArn": "$$$TASK_ROLE$$$",
+    "networkMode": "host",
+    "containerDefinitions": [{
+        "image": "amazonlinux:2",
+        "name": "task_server_endpoint_validator_container",
+        "memory": 256,
+        "command": ["sh", "-c", "curl -L -o /dev/null -s -w \"%{http_code}\n\" http://169.254.170.2/$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI | grep \"200\" && exit 42 || exit 1"]
+    }]
+}
--- a/agent/functional_tests/tests/functionaltests_unix_test.go
+++ b/agent/functional_tests/tests/functionaltests_unix_test.go
@@ -1427,3 +1427,45 @@ func TestElasticInferenceValidator(t *testing.T) {
 	exitCode, _ := task.ContainerExitcode("container_1")
 	assert.Equal(t, 42, exitCode, fmt.Sprintf("Expected exit code of 42 for container; got %d", exitCode))
 }
+
+// TestServerEndpointValidator tests the workflow of task server endpoint
+func TestServerEndpointValidator(t *testing.T) {
+	// The test runs only when the environment TEST_IAM_ROLE was set
+	if os.Getenv("TEST_DISABLE_TASK_IAM_ROLE_NET_HOST") == "true" {
+		t.Skip("Skipping test TaskIamRole in host network mode, as TEST_DISABLE_TASK_IAM_ROLE_NET_HOST is set true")
+	}
+	agentOptions := &AgentOptions{
+		ExtraEnvironment: map[string]string{
+			"ECS_ENABLE_TASK_IAM_ROLE":              "true",
+			"ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST": "true",
+		},
+		PortBindings: map[nat.Port]map[string]string{
+			"51679/tcp": {
+				"HostIP":   "0.0.0.0",
+				"HostPort": "51679",
+			},
+		},
+	}
+
+	agent := RunAgent(t, agentOptions)
+	defer agent.Cleanup()
+
+	roleArn := os.Getenv("TASK_IAM_ROLE_ARN")
+	if utils.ZeroOrNil(roleArn) {
+		t.Logf("TASK_IAM_ROLE_ARN not set, will try to use the role attached to instance profile")
+		role, err := GetInstanceIAMRole()
+		require.NoError(t, err, "Error getting IAM Roles from instance profile")
+		roleArn = *role.Arn
+	}
+	tdOverride := make(map[string]string)
+	tdOverride["$$$TASK_ROLE$$$"] = roleArn
+
+	task, err := agent.StartTaskWithTaskDefinitionOverrides(t, "task-server-endpoint-validator", tdOverride)
+	require.NoError(t, err, "Error start task-server-endpoint-validator task")
+
+	err = task.WaitStopped(10 * time.Minute)
+	assert.NoError(t, err)
+	code, ok := task.ContainerExitcode("task_server_endpoint_validator_container")
+	assert.True(t, ok, "Get exit code failed")
+	assert.Equal(t, 42, code, "Wrong exit code")
+}
--- a/agent/handlers/task_server_setup.go
+++ b/agent/handlers/task_server_setup.go
@@ -56,9 +56,9 @@ func taskServerSetup(credentialsManager credentials.Manager,
 	containerInstanceArn string) *http.Server {
 	muxRouter := mux.NewRouter()
 
-	// Set this so that for request like "/v3//metadata/task", the Agent will pass
-	// it to task metadata handler instead of returning a 301 error.
-	muxRouter.SkipClean(true)
+	// Set this to false so that for request like "//v3//metadata/task"
+	// to permanently redirect(301) to "/v3/metadata/task" handler
+	muxRouter.SkipClean(false)
 
 	muxRouter.HandleFunc(v1.CredentialsPath,
 		v1.CredentialsHandler(credentialsManager, auditLogger))
@@ -79,7 +79,7 @@ func taskServerSetup(credentialsManager credentials.Manager,
 	loggingMuxRouter.Handle(rootPath, tollbooth.LimitHandler(
 		limiter, NewLoggingHandler(muxRouter)))
 
-	loggingMuxRouter.SkipClean(true)
+	loggingMuxRouter.SkipClean(false)
 
 	server := http.Server{
 		Addr:         ":" + strconv.Itoa(config.AgentCredentialsPort),
--- a/agent/handlers/task_server_setup_test.go
+++ b/agent/handlers/task_server_setup_test.go
@@ -1052,6 +1052,34 @@ func TestV3ContainerAssociation(t *testing.T) {
 	assert.Equal(t, expectedAssociationResponse, string(res))
 }
 
+func TestTaskHTTPEndpoint301Redirect(t *testing.T) {
+	testPathsMap := map[string]string{
+		"http://127.0.0.1/v3///task/": "http://127.0.0.1/v3/task/",
+		"http://127.0.0.1//v2/credentials/test": "http://127.0.0.1/v2/credentials/test",
+	}
+
+	ctrl := gomock.NewController(t)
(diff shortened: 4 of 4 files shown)


samuelkarp:
> New tests cover the changes:

I don't see any tests in the pull request.  Can you add a functional test covering this change?

yunhee-l:
v3 tests are failing, is this causing different regression?
TestTaskHTTPEndpointErrorCode404 (0.02s)
TestTaskHTTPEndpointErrorCode404/Test_path:_/v3///task/
TestTaskHTTPEndpointErrorCode400 
TestTaskHTTPEndpointErrorCode400/Test_path:_/v3//stats 
TestTaskHTTPEndpointErrorCode400/Test_path:_/v3//task 
TestTaskHTTPEndpointErrorCode400/Test_path:_/v3//task/stats

suneyz:
> > New tests cover the changes:
> 
> I don't see any tests in the pull request. Can you add a functional test covering this change?

Great call out. Added test file. There is actually unit test to test the router behavior.

suneyz:
> > New tests cover the changes:
> 
> I don't see any tests in the pull request. Can you add a functional test covering this change?

Great call out! Added tests, there is actually unit tests that covers router

samuelkarp:
Can you add a functional test as well?

suneyz:
> Can you add a functional test as well?

updated with functional tests

misc/task-server-endpoint-validator/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ubuntu:latest
```
shubham2892: Locking this to a version would be a good idea. We are using 16.04 at other places.

agent/handlers/task_server_setup_test.go
```diff
@@ -1101,11 +1128,11 @@ func TestTaskHTTPEndpointErrorCode400(t *testing.T) {
 		"/v3/wrong-v3-endpoint-id",
 		"/v3/",
 		"/v3/wrong-v3-endpoint-id/stats",
-		"/v3//stats",
+		"/v3/stats",
```
sharanyad: Why was this changed? Shouldn't `"/v3//stats"` work as well?

agent/functional_tests/testdata/taskdefinitions/task-server-endpoint-validator/task-definition.json
```diff
@@ -0,0 +1,10 @@
+{
+    "family": "task-server-endpoint-validator",
+    "taskRoleArn": "$$$TASK_ROLE$$$",
+    "networkMode": "host",
+    "containerDefinitions": [{
+        "image": "127.0.0.1:51670/amazon/task-server-endpoint-validator:latest",
+        "name": "container_1",
```
shubham2892: nit: rename this to a more descripritive one.

agent/handlers/task_server_setup_test.go
suneyz: we handle double/muti slash with a redirect prior to 1.20.3 release. at 1.21.0 release we changed the router and added the behavior to not clean double/multi slash. This is the root cause of this issue https://github.com/aws/amazon-ecs-agent/issues/1839.

agent/handlers/task_server_setup.go
```diff
@@ -56,9 +56,9 @@ func taskServerSetup(credentialsManager credentials.Manager,
 	containerInstanceArn string) *http.Server {
 	muxRouter := mux.NewRouter()
 
-	// Set this so that for request like "/v3//metadata/task", the Agent will pass
-	// it to task metadata handler instead of returning a 301 error.
-	muxRouter.SkipClean(true)
+	// Set this to false so that for request like "//v3//metadata/task"
+	// to permanently redirect(301) to "/v3/metadata/task" handler
+	muxRouter.SkipClean(false)
```
fierlion: Since this defaults to `false` (from the docs: "SkipClean defines the path cleaning behaviour for new routes. The initial value is false.") I initially wondered if we even need the explicit setting here.  
But I like it here.  It's deliberate.  
I would make the comment clearer (like the one in the docs: "When false, the path will be cleaned, so //v3//metadata/task will become /v3/metadata/task")

suneyz: I think a 301 redirect is more clear as it is not "becoming" the new url. it is a specific 301 redirect to the new url. let me know how you think

misc/task-server-endpoint-validator/Dockerfile
suneyz: make sense :)

agent/functional_tests/testdata/taskdefinitions/task-server-endpoint-validator/task-definition.json
suneyz: do you have a suggestion? i thought the task definition and image name is descriptive enough and i didn't want to introduce similar/same name to cause confusion as it is the only container in the test.

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1427,3 +1427,45 @@ func TestElasticInferenceValidator(t *testing.T) {
 	exitCode, _ := task.ContainerExitcode("container_1")
 	assert.Equal(t, 42, exitCode, fmt.Sprintf("Expected exit code of 42 for container; got %d", exitCode))
 }
+
+// TestServerEndpointValidator tests the workflow of task server endpoint
+func TestServerEndpointValidator(t *testing.T) {
```
samuelkarp: Is there a particular reason you hand-wrote this code instead of using the simpletests package?  I would recommend using that instead.

misc/task-server-endpoint-validator/Dockerfile
samuelkarp: If you use an image like `busybox` which already includes `wget` or an image like `amazonlinux:2` which already includes `curl`, you can avoid having to construct a new image and instead can just override the command.

agent/handlers/task_server_setup.go
fierlion: I see ```w.WriteHeader(http.StatusMovedPermanently)``` aka 301.  Cool.

agent/functional_tests/testdata/taskdefinitions/task-server-endpoint-validator/task-definition.json
shubham2892: Ah what about something like task_server_endpoint_validator_container? 🤔

misc/task-server-endpoint-validator/Dockerfile
suneyz: awesome! will change it to al2 image

scripts/setup-test-registry
```diff
@@ -45,7 +45,7 @@ mirror_local_image() {
   docker rmi $2
 }
 
-for image in "amazon/amazon-ecs-netkitten" "amazon/amazon-ecs-volumes-test" "amazon/amazon-ecs-pid-namespace-test" "amazon/amazon-ecs-ipc-namespace-test" "amazon/squid" "amazon/awscli" "amazon/image-cleanup-test-image1" "amazon/image-cleanup-test-image2" "amazon/image-cleanup-test-image3" "amazon/fluentd" "amazon/amazon-ecs-agent-introspection-validator" "amazon/amazon-ecs-taskmetadata-validator" "amazon/amazon-ecs-v3-task-endpoint-validator" "amazon/amazon-ecs-container-metadata-file-validator" "amazon/amazon-ecs-elastic-inference-validator"; do
+for image in "amazon/amazon-ecs-netkitten" "amazon/amazon-ecs-volumes-test" "amazon/amazon-ecs-pid-namespace-test" "amazon/amazon-ecs-ipc-namespace-test" "amazon/squid" "amazon/awscli" "amazon/image-cleanup-test-image1" "amazon/image-cleanup-test-image2" "amazon/image-cleanup-test-image3" "amazon/fluentd" "amazon/amazon-ecs-agent-introspection-validator" "amazon/amazon-ecs-taskmetadata-validator" "amazon/amazon-ecs-v3-task-endpoint-validator" "amazon/amazon-ecs-container-metadata-file-validator" "amazon/amazon-ecs-elastic-inference-validator" "amazon/task-server-endpoint-validator"; do
```
tommyhahn: Maybe it is trivial but I am wondering why this diff is the only line that specific changes are not highlighted in the deeper green for "amazon/task-server-endpoint-validator"? Or just it happens once in a while...

suneyz: not sure, maybe it is a github bug lol

agent/functional_tests/tests/functionaltests_unix_test.go
sharanyad: I believe it is for setting the task iam roles configuration.

sharanyad:
could you please shorten the commit message to 75 characters or less? 
You could a description in a different line below if needed. 
Example commit message: https://github.com/aws/amazon-ecs-agent/pull/993/commits/c140bd39db9fa89c39dbefbaae6cc090d4b152ab

agent/functional_tests/tests/functionaltests_unix_test.go
suneyz: Discussed with Sam offline, we might not need network mode setup for it to work. 

But I'm thinking for now, let's keep test file as hand-written to release it today to solve our customers' issue. I will keep troubleshooting to see if I can make functional test work with simpletests file

suneyz:
> could you please shorten the commit message to 75 characters or less?
> You could a description in a different line below if needed.
> Example commit message: [c140bd3](https://github.com/aws/amazon-ecs-agent/commit/c140bd39db9fa89c39dbefbaae6cc090d4b152ab)

Will do

suneyz:
failed on flaky test TestDockerStopTimeout for both linux and arm. merge the change

aaithal:
> ailed on flaky test TestDockerStopTimeout for both linux and arm. merge the change

Is there an issue tracking that? How long can we expect these tests to be flaky?

agent/functional_tests/tests/functionaltests_unix_test.go
samuelkarp: Task IAM roles are supported with "host", "bridge", and "awsvpc" network modes, and we have existing functional tests that validate this.  There should not be a dependency on using the "host" network mode.

suneyz:
> > failed on flaky test TestDockerStopTimeout for both linux and arm. merge the change
> 
> Is there an issue tracking that? How long can we expect these tests to be flaky?

It is tracked in our sprint board.

aaithal:
> It is tracked in our sprint board.

Can you please create an issue in this repo because that's where we track these things? That way, you can tie that into releases, track when it surfaced, when it was fixed etc. Also, as a customer, or as someone who doesn't have visibility into your sprint board, it's hard to track if there's nothing in the github repo related to that.

agent/functional_tests/tests/functionaltests_unix_test.go
suneyz: Read some document, agree on this should not depend on host mode. I will dive deep to see if my host needs some config to make it functional test on other modes, and I will follow up with another pr to address it once i get it figured out.

suneyz:
> > It is tracked in our sprint board.
> 
> Can you please create an issue in this repo because that's where we track these things? That way, you can tie that into releases, track when it surfaced, when it was fixed etc. Also, as a customer, or as someone who doesn't have visibility into your sprint board, it's hard to track if there's nothing in the github repo related to that.

Github issue created. https://github.com/aws/amazon-ecs-agent/issues/1855

PR #1862, opened 2019-02-15
Add the dependency conditions of "COMPLETE", "SUCCESS", "HEALTHY"

ubhattacharjya:
### Summary

Here I am adding the dependency conditions of "COMPLETE", "SUCCESS", "HEALTHY" for container ordering.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -25,6 +25,22 @@ import (
 	"strings"
 )
 
+const (
+	// StartCondition ensures that a container progresses to next state only when dependency container has started
+	startCondition = "START"
+	// RunningCondition ensures that a container progresses to next state only when dependency container is running
+	runningCondition = "RUNNING"
+	// SuccessCondition ensures that a container progresses to next state only when
+	// dependency container has successfully completed with exit code 0
+	successCondition = "SUCCESS"
+	// CompleteCondition ensures that a container progresses to next state only when dependency container has completed
+	completeCondition = "COMPLETE"
+	// StartCondition ensures that a container progresses to next state only when dependency container is healthy
+	healthyCondition = "HEALTHY"
+	// 0 is the standard exit code for success.
+	successExitCode = 0
+)
+
 var (
 	// CredentialsNotResolvedErr is the error where a container needs to wait for
 	// credentials before it can process by agent
@@ -256,26 +272,15 @@ func verifyResourceDependenciesResolved(target *apicontainer.Container, existing
 func containerOrderingDependenciesCanResolve(target *apicontainer.Container,
 	dependsOnContainer *apicontainer.Container,
 	dependsOnStatus string) bool {
+
 	targetDesiredStatus := target.GetDesiredStatus()
 	dependsOnContainerDesiredStatus := dependsOnContainer.GetDesiredStatus()
 
 	switch dependsOnStatus {
+	case startCondition:
+		return verifyContainerOrderingStatus(dependsOnContainer)
 
-	case "START":
-		// The 'target' container doesn't desire to move to either 'Created' or the 'steady' state,
-		// which is not allowed
-		if targetDesiredStatus != apicontainerstatus.ContainerCreated && targetDesiredStatus != target.GetSteadyStateStatus() {
-			return false
-		}
-		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
-		// Allow this only if the dependency container also desires to be started
-		// i.e it's status is any of 'Created', 'steady state' or 'Stopped'
-		return dependsOnContainerDesiredStatus == apicontainerstatus.ContainerCreated ||
-			dependsOnContainerDesiredStatus == apicontainerstatus.ContainerStopped ||
-			dependsOnContainerDesiredStatus == dependsOnContainer.GetSteadyStateStatus()
-
-	case "RUNNING":
-
+	case runningCondition:
 		if targetDesiredStatus == apicontainerstatus.ContainerCreated {
 			// The 'target' container desires to be moved to 'Created' state.
 			// Allow this only if the desired status of the dependency container is
@@ -287,9 +292,23 @@ func containerOrderingDependenciesCanResolve(target *apicontainer.Container,
 			// The 'target' container desires to be moved to its 'steady' state.
 			// Allow this only if the dependency container also desires to be in 'steady state'
 			return dependsOnContainerDesiredStatus == dependsOnContainer.GetSteadyStateStatus()
-
 		}
 		return false
+
+	case successCondition:
+		var dependencyStoppedSuccessfully bool
+
+		if dependsOnContainer.GetKnownExitCode() != nil {
+			dependencyStoppedSuccessfully = dependsOnContainer.GetKnownStatus() == apicontainerstatus.ContainerStopped &&
+				*dependsOnContainer.GetKnownExitCode() == successExitCode
+		}
+		return verifyContainerOrderingStatus(dependsOnContainer) || dependencyStoppedSuccessfully
+
+	case completeCondition:
+		return verifyContainerOrderingStatus(dependsOnContainer)
+
+	case healthyCondition:
+		return verifyContainerOrderingStatus(dependsOnContainer) && dependsOnContainer.HealthStatusShouldBeReported()
 
 	default:
 		return false
@@ -304,13 +323,7 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
 	dependsOnContainerKnownStatus := dependsOnContainer.GetKnownStatus()
 
 	switch dependsOnStatus {
-
-	case "START":
-		// The 'target' container doesn't desire to be moved to 'Created' or the 'steady' state.
-		// which is not allowed.
-		if targetDesiredStatus != apicontainerstatus.ContainerCreated && targetDesiredStatus != apicontainerstatus.ContainerRunning {
-			return false
-		}
+	case startCondition:
 		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
 		// Allow this only if the known status of the dependency container state is already started
 		// i.e it's state is any of 'Created', 'steady state' or 'Stopped'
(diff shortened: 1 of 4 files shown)


agent/engine/dependencygraph/graph.go
```diff
@@ -287,10 +276,22 @@ func containerOrderingDependenciesCanResolve(target *apicontainer.Container,
...
 		return false
 
+	case "SUCCESS":
+		isDependOnContainerStoppedSuccessfully := dependsOnContainer.GetKnownStatus() == apicontainerstatus.ContainerStopped &&
+			*dependsOnContainer.GetKnownExitCode() == 0
+		return verifyContainerOrderingStatus(target, dependsOnContainer) || isDependOnContainerStoppedSuccessfully
+
+	case "COMPLETE":
+		isDependOnContainerStopped := dependsOnContainer.GetKnownStatus() == apicontainerstatus.ContainerStopped &&
+			dependsOnContainer.GetKnownExitCode() != nil
```
petderek: do we need this, or is it implied by ContainerStopped?

agent/engine/dependencygraph/graph.go
```diff
@@ -319,26 +320,52 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
...
+		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
+		// Allow this only if the known status of the dependency container state is stopped with an exit code of 0
+		return dependsOnContainerKnownStatus == apicontainerstatus.ContainerStopped && *dependsOnContainer.GetKnownExitCode() == 0
+
+	case "COMPLETE":
+		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
+		// Allow this only if the known status of the dependency container state is stopped with any exit code
+		return dependsOnContainerKnownStatus == apicontainerstatus.ContainerStopped && dependsOnContainer.GetKnownExitCode() != nil
+
+	case "HEALTHY":
```
petderek: We could probably extract these into constants:

```
const (
    ContainerHealthy = "HEALTHY"
    // etc
)
```

agent/engine/dependencygraph/graph.go
ubhattacharjya: I think probably we can get rid of this. Initially I added this, to cover all corner cases but I thi nk we can get rid of this

agent/engine/dependencygraph/graph.go
```diff
@@ -25,6 +25,14 @@ import (
 	"strings"
 )
 
+const (
+	ContainerStartDependencyCondition    = "START"
```
petderek: we are already in dependency graph pacakges.. can we just call it StartCondition, etc

agent/engine/dependencygraph/graph.go
```diff
@@ -287,10 +284,20 @@ func containerOrderingDependenciesCanResolve(target *apicontainer.Container,
...
-
 		}
 		return false
 
+	case ContainerSuccessDependencyCondition:
+		isDependOnContainerStoppedSuccessfully := dependsOnContainer.GetKnownStatus() == apicontainerstatus.ContainerStopped &&
+			*dependsOnContainer.GetKnownExitCode() == 0
+		return verifyContainerOrderingStatus(dependsOnContainer) || isDependOnContainerStoppedSuccessfully
+
+	case ContainerCompleteDependencyCondition:
```
petderek: for this one you can do:

`case CompleteCondition,HealthyCondition:`

agent/engine/dependencygraph/graph.go
ubhattacharjya: makes sense. changed it.

agent/engine/dependencygraph/graph.go
```diff
@@ -256,26 +264,15 @@ func verifyResourceDependenciesResolved(target *apicontainer.Container, existing
...
 	dependsOnContainerDesiredStatus := dependsOnContainer.GetDesiredStatus()
 
 	switch dependsOnStatus {
+	case StartCondition:
+		return verifyContainerOrderingStatus(dependsOnContainer)
 
-	case "START":
-		// The 'target' container doesn't desire to move to either 'Created' or the 'steady' state,
-		// which is not allowed
-		if targetDesiredStatus != apicontainerstatus.ContainerCreated && targetDesiredStatus != target.GetSteadyStateStatus() {
```
sharanyad: for my understanding, why was this check removed?

agent/engine/dependencygraph/graph.go
```diff
@@ -287,10 +284,20 @@ func containerOrderingDependenciesCanResolve(target *apicontainer.Container,
 			// The 'target' container desires to be moved to its 'steady' state.
 			// Allow this only if the dependency container also desires to be in 'steady state'
 			return dependsOnContainerDesiredStatus == dependsOnContainer.GetSteadyStateStatus()
-
 		}
 		return false
 
+	case SuccessCondition:
+		isDependOnContainerStoppedSuccessfully := dependsOnContainer.GetKnownStatus() == apicontainerstatus.ContainerStopped &&
+			*dependsOnContainer.GetKnownExitCode() == 0
```
sharanyad: is `0` always and the only return success code?

agent/engine/dependencygraph/graph_test.go
```diff
@@ -640,25 +642,57 @@ func TestContainerOrderingCanResolve(t *testing.T) {
...
 }
 
 func TestContainerOrderingIsResolved(t *testing.T) {
 	testcases := []struct {
 		TargetDesired apicontainerstatus.ContainerStatus
-		DependencyKnown   apicontainerstatus.ContainerStatus
+		DependencyKnown apicontainerstatus.ContainerStatus
 		DependencyCondition string
-		Resolved      bool
+		Resolved  bool
```
sharanyad: needs `go fmt`

agent/engine/dependencygraph/graph.go
```diff
@@ -304,41 +311,55 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
...
+		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
+		// Allow this only if the known status of the dependency container state is stopped with an exit code of 0
+		return dependsOnContainerKnownStatus == apicontainerstatus.ContainerStopped && *dependsOnContainer.GetKnownExitCode() == 0
+
+	case CompleteCondition:
+		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
+		// Allow this only if the known status of the dependency container state is stopped with any exit code
+		return dependsOnContainerKnownStatus == apicontainerstatus.ContainerStopped && dependsOnContainer.GetKnownExitCode() != nil
+
+	case HealthyCondition:
```
sharanyad: are we verifying if health check is enabled for the `dependsOnContainer` if dependent status is `HEALTHY`?

aaithal:
Adding to what @sharanyad  said about the commit message, it says "what this commit is doing". Equally important is "why". If you could add details in the commit message body for that, it'd be awesome.

agent/engine/dependencygraph/graph.go
ubhattacharjya: I removed this because this function is getting called from verifyContainerOrderingStatusResolvable. It has a line which already checks this. https://github.com/ubhattacharjya/amazon-ecs-agent/blob/RefactorDependencyCode/agent/engine/dependencygraph/graph.go#L195

agent/engine/dependencygraph/graph.go
ubhattacharjya: Thanks Sharanya for pointing this out. I have included this check now.

agent/engine/dependencygraph/graph.go
ubhattacharjya: I think, 0 is the standard exit code for success. If the customer mentions some other exit code, then they will have to use the condition 'COMPLETE'.

petderek: [Yep!](https://www.tldp.org/LDP/abs/html/exit-status.html)

agent/engine/dependencygraph/graph.go
```diff
@@ -304,41 +311,56 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
...
 			// Allow this only if the dependency container is in 'steady state' as well
 			return dependsOnContainer.IsKnownSteadyState()
-
 		}
 		return false
 
+	case SuccessCondition:
+		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
+		// Allow this only if the known status of the dependency container state is stopped with an exit code of 0
+		return dependsOnContainerKnownStatus == apicontainerstatus.ContainerStopped && *dependsOnContainer.GetKnownExitCode() == 0
```
petderek: should we do a nil check before dereferencing?

agent/engine/dependencygraph/graph.go
sharanyad: let's check this on the service side as well

agent/engine/dependencygraph/graph.go
```diff
@@ -304,41 +314,59 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
...
 			// Allow this only if the dependency container is in 'steady state' as well
 			return dependsOnContainer.IsKnownSteadyState()
+		}
+		return false
 
+	case SuccessCondition:
+		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
+		// Allow this only if the known status of the dependency container state is stopped with an exit code of 0
+		if dependsOnContainer.GetKnownExitCode() != nil {
+			return dependsOnContainerKnownStatus == apicontainerstatus.ContainerStopped && *dependsOnContainer.GetKnownExitCode() == 0
```
sharanyad: nit: break this into multiple lines

sharanyad: lets make `0` a constant and call it `successExitCode`

agent/engine/dependencygraph/graph.go
```diff
@@ -25,6 +25,15 @@ import (
 	"strings"
 )
 
+const (
+	StartCondition    = "START"
```
aaithal: lint will tell you that you need to add comments for all these globally visible consts. Also, on that note, do they need to be globally visible? Adding comments would still be a good idea, with a brief description of what these consts are used for and why they have these values.

agent/engine/dependencygraph/graph.go
```diff
@@ -287,10 +285,23 @@ func containerOrderingDependenciesCanResolve(target *apicontainer.Container,
 			// The 'target' container desires to be moved to its 'steady' state.
 			// Allow this only if the dependency container also desires to be in 'steady state'
 			return dependsOnContainerDesiredStatus == dependsOnContainer.GetSteadyStateStatus()
-
 		}
 		return false
 
+	case SuccessCondition:
+		var isDependOnContainerStoppedSuccessfully bool
```
aaithal: You should consider renaming this variable. It's hard to grok what this means exactly.

agent/engine/dependencygraph/graph.go
ubhattacharjya: I may need global variables for the dependency conditions so that I can use them while writing functional tests. I have added comments to describe them. Thanks for the suggestion.

petderek: marking these as not-global is probably the best bet here.

aaithal: If the functional tests take a dependency on internal agent states, we're probably doing something wrong wrt how we're writing these tests. I'd prefer them to local scoped for now. You can always expand the scope when/if needed.

ubhattacharjya: I haved changed it to local.

PR #1865, opened 2019-02-18
Revert "Add log driver secret to LogConfig and agent capability" due to pending test process

tommyhahn:
Reverts aws/amazon-ecs-agent#1818 due to pending test process


adnxn:
mind adding context for why we're reverting in the commit msg please?

Issue #1835, opened 2019-02-06
testing: Windows Flakey integ tests 


sharanyad:
### Summary
As mentioned in https://github.com/aws/amazon-ecs-agent/pull/1830 and https://github.com/aws/amazon-ecs-agent/pull/1833, `TestVolumesFromRO`, `TestTaskLevelVolume`  are not passing consistently. 

### Expected Behavior
test passes

### Observed Behavior
test fails

adnxn:
are both of these timeout failures?

sharanyad:
```
goroutine 1654 [select, 4 minutes]:
github.com/aws/amazon-ecs-agent/agent/vendor/github.com/Microsoft/go-winio.(*win32File).asyncIo(0xc0424a2bb0, 0xc042871b30, 0xc0424a2be8, 0x0, 0x1344b00, 0x133f5b8, 0x1344b00, 0x133f5b8, 0x0)
	C:/Users/ADMINI~1/AppData/Local/Temp/amazon-ecs-agent/go/src/github.com/aws/amazon-ecs-agent/agent/vendor/github.com/Microsoft/go-winio/file.go:186 +0x229
github.com/aws/amazon-ecs-agent/agent/vendor/github.com/Microsoft/go-winio.(*win32File).Read(0xc0424a2bb0, 0xc042a9c000, 0x1000, 0x1000, 0x0, 0x0, 0x0)
	C:/Users/ADMINI~1/AppData/Local/Temp/amazon-ecs-agent/go/src/github.com/aws/amazon-ecs-agent/agent/vendor/github.com/Microsoft/go-winio/file.go:224 +0x19b
github.com/aws/amazon-ecs-agent/agent/vendor/github.com/Microsoft/go-winio.(*win32MessageBytePipe).Read(0xc042002a20, 0xc042a9c000, 0x1000, 0x1000, 0xc04239ff00, 0xc042843808, 0x477657)
	C:/Users/ADMINI~1/AppData/Local/Temp/amazon-ecs-agent/go/src/github.com/aws/amazon-ecs-agent/agent/vendor/github.com/Microsoft/go-winio/pipe.go:116 +0xa5
net/http.(*persistConn).Read(0xc042958900, 0xc042a9c000, 0x1000, 0x1000, 0x0, 0xe1f, 0x1)
	C:/Go/src/net/http/transport.go:1391 +0x232
bufio.(*Reader).fill(0xc0428f27e0)
[...]

PR #1866, opened 2019-02-18
dependencygraph: Enforce shutdown order

petderek:
### Summary

Enforce shutdown ordering within a task.

### Implementation details

We now apply shutdown order in any dependency case, including
dependsOn directives, links, or volumes. What this means is that
agent will now make a best effort attempt to shutdown containers
in the inverse order they were created.

For example, a container using a link for communication will wait
until the linked container has terminated before terminating
itself. Likewise, a container named in another container's
dependsOn directive will wait until that dependent container
terminates.

One note about the current implementation is that the dependencies
aren't assumed to be transitive, so if a chain exists such that:

A -> B -> C

Container "C" will shutdown before "B", but it won't check status
against container "A" explicity. If A depends on C, we expect:

A -> B -> C
     A -> C

The lack of transitive dependency logic applies is consistent with
startup order as of this commit.
 
### Testing
Added additional tests to cover the shutdown use cases.

### Licensing

### DIFF
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -328,9 +328,7 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
 		// The 'target' container desires to be moved to 'Created' or the 'steady' state.
 		// Allow this only if the known status of the dependency container state is already started
 		// i.e it's state is any of 'Created', 'steady state' or 'Stopped'
-		return dependsOnContainerKnownStatus == apicontainerstatus.ContainerCreated ||
-			dependsOnContainerKnownStatus == apicontainerstatus.ContainerStopped ||
-			dependsOnContainerKnownStatus == dependsOnContainer.GetSteadyStateStatus()
+		return dependsOnContainerKnownStatus >= apicontainerstatus.ContainerCreated
 
 	case runningCondition:
 		if targetDesiredStatus == apicontainerstatus.ContainerCreated {
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -57,16 +57,9 @@ var (
 	ErrResourceDependencyNotResolved = errors.New("dependency graph: dependency on resources not resolved")
 )
 
-// Because a container may depend on another container being created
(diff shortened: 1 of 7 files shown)


agent/engine/dependencygraph/graph.go
```diff
@@ -339,6 +336,46 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
...
+
+	if len(missingShutdownDependencies) == 0 {
+		return nil
+	}
+
+	return fmt.Errorf("dependency graph: target %s needs other containers stopped before it can stop: [%s]",
+		target.Name, strings.Join(missingShutdownDependencies, "], ["))
+}
+
+func verifyContainerOrderingStatus(dependsOnContainer *apicontainer.Container) bool {
```
yunhee-l: where is this used?

ubhattacharjya: Is this function part of your commit? A bit confused

petderek: I did a bad rebase -- will fix

agent/engine/dependencygraph/graph.go
```diff
@@ -377,6 +372,36 @@ func verifyContainerOrderingStatus(dependsOnContainer *apicontainer.Container) b
 		dependsOnContainerDesiredStatus == dependsOnContainer.GetSteadyStateStatus()
 }
 
+func verifyShutdownOrder(target *apicontainer.Container, existingContainers map[string]*apicontainer.Container) error {
+	// If the target is already terminal or if it doesn't need to be terminal, shutdown order doesn't apply
+	if !target.DesiredTerminal() || target.KnownTerminal() {
```
sharanyad: should we move this check outside of this method? it's more logical to verify shutdown order only if desired status is terminal

adnxn:
are the ft failures due to timeouts?

petderek:
Great question. They were passing before i rebased 😖

petderek:
Unit tests are both passing in travis. The linux test isn't failing; CI is just reporting failed for some reason?

petderek:
Logdriver test is failing suddenly. I've made a note to fix the test in #1835

PR #1867, opened 2019-02-18
ENI trunking - Running trunk/branch AWSVPC task 

shubham2892:
### Summary

This PR is WIP for ENI trunking:
- Model changes for ACS and CNI plugin.
- Invoke the correct CNI plugin based on ENI received.

#### High level workflow

For an AWSVPC task using trunk ENI "vpc-branch-eni" CNI plugin will be used(for ADD and DEL) instead of "ecs-cni" plugin when setting up pause container namespace.
 
#### Model changes
- "InterfaceVlanProperties" struct added. It will contain mac address and ID of the trunk ENI.
- "ENI" struct extended to add "InterfaceAssociationProtocol" field and "InterfaceVlanProperties" field. InterfaceAssociationProtocol type indicates the type of ENI. It could be either "default" or "vlan". "InterfaceVlanProperties" field points to  InterfaceVlanProperties struct which will contain trunk ENI information. 
- "BranchENIConfig" struct added. This will be used to invoke the "vpc-branch-eni" plugin in the similar fashion as "ENIConfig" is used for invoking "ecs-cni" plugin. 
- A "Config" object contains all the information required for invoking the CNI plugin which sets up the pause container's network namespace. The config object is extended to include trunk ENI information which will be needed to construct "BranchENIConfig" object.
 
### Testing

- Unit tests added.
- Manual testing by building a modified version on agent on an instance and running a task.

New tests cover the changes:yes
### Licensing

### DIFF
--- a/agent/acs/handler/payload_handler.go
+++ b/agent/acs/handler/payload_handler.go
@@ -13,9 +13,8 @@
 package handler
 
 import (
-	"fmt"
-
 	"context"
+	"fmt"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	"github.com/aws/amazon-ecs-agent/agent/api"
@@ -220,7 +219,6 @@ func (payloadHandler *payloadRequestHandler) addPayloadTasks(payload *ecsacs.Pay
 				allTasksOK = false
 				continue
 			}
-
 			apiTask.SetTaskENI(eni)
 		}
 		// Add the app mesh information to task struct
@@ -322,12 +320,12 @@ func (payloadHandler *payloadRequestHandler) ackCredentials(messageID *string, c
 // and returns the boolean comparison result
 type skipAddTaskComparatorFunc func(apitaskstatus.TaskStatus) bool
 
-// isTaskStatusStopped returns true if the task status == STOPPTED
+// isTaskStatusStopped returns true if the task status == STOPPED
 func isTaskStatusStopped(status apitaskstatus.TaskStatus) bool {
 	return status == apitaskstatus.TaskStopped
 }
 
-// isTaskStatusNotStopped returns true if the task status != STOPPTED
+// isTaskStatusNotStopped returns true if the task status != STOPPED
 func isTaskStatusNotStopped(status apitaskstatus.TaskStatus) bool {
 	return status != apitaskstatus.TaskStopped
 }
--- a/agent/api/eni/eni.go
+++ b/agent/api/eni/eni.go
@@ -200,9 +200,11 @@ func ValidateTaskENI(acsenis []*ecsacs.ElasticNetworkInterface) error {
 	if len(acsenis) != 1 {
 		return errors.Errorf("eni message validation: more than one ENIs in the message(%d)", len(acsenis))
 	} else if len(acsenis[0].Ipv4Addresses) != 1 {
-		return errors.Errorf("eni message validation: more than one ipv4 addresses in the message(%d)", len(acsenis[0].Ipv4Addresses))
+		return errors.Errorf("eni message validation: more than one ipv4 addresses in the message(%d)",
+			len(acsenis[0].Ipv4Addresses))
 	} else if len(acsenis[0].Ipv6Addresses) > 1 {
-		return errors.Errorf("eni message validation: more than one ipv6 addresses in the message(%d)", len(acsenis[0].Ipv6Addresses))
+		return errors.Errorf("eni message validation: more than one ipv6 addresses in the message(%d)",
+			len(acsenis[0].Ipv6Addresses))
 	}
 
 	if acsenis[0].MacAddress == nil {
@@ -214,14 +216,18 @@ func ValidateTaskENI(acsenis []*ecsacs.ElasticNetworkInterface) error {
 	}
 
 	if (acsenis[0].InterfaceAssociationProtocol != nil) && (aws.StringValue(acsenis[0].InterfaceAssociationProtocol) !=
-		VLANInterfaceAssociationProtocol) && (aws.StringValue(acsenis[0].InterfaceAssociationProtocol) != DefaultInterfaceAssociationProtocol) {
-		return errors.Errorf("Invalid ENI interface type: %s", aws.StringValue(acsenis[0].InterfaceAssociationProtocol))
+		VLANInterfaceAssociationProtocol) && (aws.StringValue(acsenis[0].InterfaceAssociationProtocol) !=
+		DefaultInterfaceAssociationProtocol) {
+		return errors.Errorf("invalid ENI interface type: %s",
+			aws.StringValue(acsenis[0].InterfaceAssociationProtocol))
 	}
 
 	// If ENI type is vlan, InterfaceVlanProperties must be not nil.
 	if aws.StringValue(acsenis[0].InterfaceAssociationProtocol) == VLANInterfaceAssociationProtocol {
-		if acsenis[0].InterfaceVlanProperties == nil || len(aws.StringValue(acsenis[0].InterfaceVlanProperties.VlanId)) == 0 || len(aws.StringValue(acsenis[0].InterfaceVlanProperties.TrunkInterfaceMacAddress)) == 0 {
-			return errors.Errorf("vlan interface properties missing.")
+		if acsenis[0].InterfaceVlanProperties == nil ||
+			len(aws.StringValue(acsenis[0].InterfaceVlanProperties.VlanId)) == 0 ||
+			len(aws.StringValue(acsenis[0].InterfaceVlanProperties.TrunkInterfaceMacAddress)) == 0 {
+			return errors.New("vlan interface properties missing.")
 		}
 	}
 
--- a/agent/api/eni/eni.go
+++ b/agent/api/eni/eni.go
@@ -26,8 +26,8 @@ import (
 type ENI struct {
 	// ID is the id of eni
 	ID string `json:"ec2Id"`
-	// ENIType is the type of ENI, valid value: "default", "vlan"
-	ENIType string `json:",omitempty"`
+	// InterfaceAssociationProtocol is the type of ENI, valid value: "default", "vlan"
+	InterfaceAssociationProtocol string `json:",omitempty"`
 	// IPV4Addresses is the ipv4 address associated with the eni
 	IPV4Addresses []*ENIIPV4Address
 	// IPV6Addresses is the ipv6 address associated with the eni
@@ -58,11 +58,11 @@ type InterfaceVlanProperties struct {
 }
 
 const (
-	// RegularENIType represents the standard ENI type.
-	RegularENIType = "default"
+	// DefaultInterfaceAssociationProtocol represents the standard ENI type.
+	DefaultInterfaceAssociationProtocol = "default"
 
-	// BranchENIType represents the ENI with trunking enabled.
-	BranchENIType = "vlan"
+	// VLANInterfaceAssociationProtocol represents the ENI with trunking enabled.
+	VLANInterfaceAssociationProtocol = "vlan"
 )
 
 // GetIPV4Addresses returns a list of ipv4 addresses allocated to the ENI
@@ -109,8 +109,8 @@ func (eni *ENI) String() string {
 
 	eniString := ""
 
-	if len(eni.ENIType) == 0 {
-		eniString += fmt.Sprintf(" ,ENI type: [%s]", eni.ENIType)
+	if len(eni.InterfaceAssociationProtocol) == 0 {
+		eniString += fmt.Sprintf(" ,ENI type: [%s]", eni.InterfaceAssociationProtocol)
 	}
 
 	if eni.InterfaceVlanProperties != nil {
@@ -166,20 +166,20 @@ func ENIFromACS(acsenis []*ecsacs.ElasticNetworkInterface) (*ENI, error) {
 
 	var interfaceVlanProperties InterfaceVlanProperties
 
-	if aws.StringValue(acsenis[0].InterfaceAssociationProtocol) == BranchENIType {
+	if aws.StringValue(acsenis[0].InterfaceAssociationProtocol) == VLANInterfaceAssociationProtocol {
 		interfaceVlanProperties.TrunkInterfaceMacAddress = aws.StringValue(acsenis[0].InterfaceVlanProperties.TrunkInterfaceMacAddress)
 		interfaceVlanProperties.VlanID = aws.StringValue(acsenis[0].InterfaceVlanProperties.VlanId)
 	}
 
 	eni := &ENI{
-		ID:                       aws.StringValue(acsenis[0].Ec2Id),
-		IPV4Addresses:            ipv4,
-		IPV6Addresses:            ipv6,
-		MacAddress:               aws.StringValue(acsenis[0].MacAddress),
-		PrivateDNSName:           aws.StringValue(acsenis[0].PrivateDnsName),
-		SubnetGatewayIPV4Address: aws.StringValue(acsenis[0].SubnetGatewayIpv4Address),
-		ENIType:                  aws.StringValue(acsenis[0].InterfaceAssociationProtocol),
-		InterfaceVlanProperties:  &interfaceVlanProperties,
+		ID:                           aws.StringValue(acsenis[0].Ec2Id),
+		IPV4Addresses:                ipv4,
+		IPV6Addresses:                ipv6,
+		MacAddress:                   aws.StringValue(acsenis[0].MacAddress),
+		PrivateDNSName:               aws.StringValue(acsenis[0].PrivateDnsName),
+		SubnetGatewayIPV4Address:     aws.StringValue(acsenis[0].SubnetGatewayIpv4Address),
+		InterfaceAssociationProtocol: aws.StringValue(acsenis[0].InterfaceAssociationProtocol),
+		InterfaceVlanProperties:      &interfaceVlanProperties,
 	}
 
 	for _, nameserverIP := range acsenis[0].DomainNameServers {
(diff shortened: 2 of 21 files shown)


agent/acs/client/acs_client_test.go
```diff
@@ -18,6 +18,7 @@ package acsclient
 import (
 	"encoding/json"
 	"errors"
+	"github.com/aws/amazon-ecs-agent/agent/api/eni"
```
fenxiong: can you organize the import following the guideline [here](https://github.com/golang/go/wiki/CodeReviewComments#imports)? same for other imports you added

agent/api/eni/eni.go
```diff
@@ -171,5 +197,15 @@ func ValidateTaskENI(acsenis []*ecsacs.ElasticNetworkInterface) error {
 		return errors.Errorf("eni message validation: empty eni id in the message")
 	}
 
+	if (aws.StringValue(acsenis[0].InterfaceType) != BranchENIName) && (aws.StringValue(acsenis[0].InterfaceType) != RegularENIName) {
```
fenxiong: you don't need this check i think. existing eni will not have a type and can be assumed to be standard type

agent/api/eni/eni.go
```diff
@@ -171,5 +197,15 @@ func ValidateTaskENI(acsenis []*ecsacs.ElasticNetworkInterface) error {
...
+		return errors.Errorf("Invalid ENI interface type: %s", aws.StringValue(acsenis[0].InterfaceType))
+	}
+
+	if aws.StringValue(acsenis[0].InterfaceType) == BranchENIName {
+		if acsenis[0].InterfaceVlanProperties == nil {
+			return errors.Errorf("vlan interface properties missing.")
```
fenxiong: can you also check that vlan id and trunk mac address fields are not empty here?

agent/api/eni/eni.go
```diff
@@ -38,14 +40,28 @@ type ENI struct {
 	// DomainNameSearchList specifies the search list for the domain
 	// name lookup, for the eni
 	DomainNameSearchList []string `json:",omitempty"`
-
+	// InterfaceVlanProperties contains information for an interface
+	// that is supposed to be used as a VLAN device
+	InterfaceVlanProperties *InterfaceVlanProperties
```
fenxiong: can you modify the String() method in this file to include the information of this field for branch eni? also this field should probably be omitted if empty

agent/ecscni/types.go
```diff
@@ -124,6 +126,36 @@ type ENIConfig struct {
...
+	// BranchGatewayIPAddress is the IP address of the branch ENI's default gateway.
+	BranchGatewayIPAddress string `json:"branchGatewayIPAddress"`
+	// InterfaceType is the type of the interface to connect the branch ENI to
+	InterfaceType string `json:"interfaceType,omitempty"`
+	// BlockInstanceMetdata specifies if InstanceMetadata endpoint should be
+	// blocked
+	BlockInstanceMetdata bool `json:"block-instance-metadata"`
+	// SubnetGatewayIPV4Address specifies the ipv4 address of the subnet gateway
+	// for the ENI
+	SubnetGatewayIPV4Address string `json:"subnetgateway-ipv4-address"`
```
fenxiong: this field is not needed

agent/ecscni/types.go
```diff
@@ -124,6 +126,36 @@ type ENIConfig struct {
...
+	BranchMACAddress string `json:"branchMACAddress"`
+	// BranchIPAddress is the IP address of the branch ENI
+	BranchIPAddress string `json:"branchIPAddress"`
+	// BranchGatewayIPAddress is the IP address of the branch ENI's default gateway.
+	BranchGatewayIPAddress string `json:"branchGatewayIPAddress"`
+	// InterfaceType is the type of the interface to connect the branch ENI to
+	InterfaceType string `json:"interfaceType,omitempty"`
+	// BlockInstanceMetdata specifies if InstanceMetadata endpoint should be
+	// blocked
+	BlockInstanceMetdata bool `json:"block-instance-metadata"`
```
fenxiong: block-instance-metadata should be blockInstanceMetadata

agent/ecscni/plugin.go
```diff
@@ -311,6 +329,24 @@ func (client *cniClient) createENINetworkConfig(cfg *Config) (string, *libcni.Ne
...
+func (client *cniClient) createBranchENINetworkConfig(cfg *Config) (string, *libcni.NetworkConfig, error) {
+	eniConf := BranchENIConfig{
+		TrunkMACAddress:          cfg.TrunkMACAddress,
+		Type:                     ECSBranchENIPluginName,
+		CNIVersion:               client.cniVersion,
+		BranchVlanID:             cfg.BranchVlanID,
+		BranchIPAddress:          cfg.ENIIPV4Address,
+		BranchMACAddress:         cfg.ENIMACAddress,
+		BlockInstanceMetdata:     cfg.BlockInstanceMetdata,
+		SubnetGatewayIPV4Address: cfg.SubnetGatewayIPV4Address,
```
fenxiong: this is not a field for the new plugin. instead this should be BranchGatewayIPAddress

fenxiong:
are the functional test failures related to your changes?

shubham2892:
> are the functional test failures related to your changes?

Yes the functional test failure are related to changes I made, working on fixing this. Thanks for the review. 😄

agent/ecscni/plugin.go
```diff
@@ -113,15 +115,24 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
 	os.Setenv("ECS_CNI_LOGLEVEL", logger.GetLevel())
```
fenxiong: when vpc-branch-eni plugin is used, similar to here, you will need to specify the log level of vpc-branch-eni via "VPC_CNI_LOG_LEVEL"; you also need to specify "VPC_CNI_LOG_FILE" to be "/log/vpc-branch-eni.log" since the default log path used by the plugin is "/var/log/vpc-branch-eni.log" which is not under the log path that's mounted to the agent (which is [this](https://github.com/aws/amazon-ecs-init/blob/master/ecs-init/docker/docker.go#L32))

fenxiong: same in cleanup

agent/ecscni/plugin.go
```diff
@@ -37,6 +39,7 @@ const (
 	currentECSCNIVersion = "2018.10.0"
 	currentECSCNIGitHash = "93f4377604504bff92e7555da73b0cba732a4fbb"
 	currentVPCCNIGitHash = "76973f587f9dfb0b40092a78fb5623c9547bc647"
```
suneyz: can you check why you have update on amazon-vpc-plugins submodule without changing this git hash?

agent/ecscni/plugin.go
```diff
@@ -111,18 +114,34 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
...
-	os.Setenv("ECS_CNI_LOGLEVEL", logger.GetLevel())
-	defer os.Unsetenv("ECS_CNI_LOGLEVEL")
 
-	// Invoke eni plugin ADD command
-	result, err := client.add(runtimeConfig, cfg, client.createENINetworkConfig)
-	if err != nil {
-		return nil, errors.Wrap(err, "cni setup: invoke eni plugin failed")
+	// Invoke eni plugin ADD command based on the type of eni plugin
+	if cfg.ENIType == apieni.BranchENIType {
+		os.Setenv("VPC_CNI_LOG_LEVEL", logger.GetLevel())
```
suneyz: curious, why do we need this log level variable to be set?

agent/ecscni/plugin.go
```diff
@@ -111,18 +114,34 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
...
+		defer os.Unsetenv("VPC_CNI_LOG_LEVEL")
+		defer os.Unsetenv("VPC_CNI_LOG_FILE")
+
+		result, err := client.add(runtimeConfig, cfg, client.createBranchENINetworkConfig)
+		if err != nil {
+			return nil, errors.Wrap(err, "cni setup: invoke branch eni plugin failed")
+		}
+		seelog.Debugf("[ECSVPCCNI] Branch ENI setup done: %s", result.String())
+	} else {
+		os.Setenv("VPC_CNI_LOG_LEVEL", logger.GetLevel())
```
suneyz: is client.createENINetworkConfig invoking previous ecs-cni plugin? why not set ECS_CNI_LOGLEVEL?

agent/ecscni/plugin.go
```diff
@@ -111,18 +114,34 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
...
 	seelog.Debugf("[ECSCNI] Starting ENI (%s) setup in the the container namespace: %s", cfg.ENIID, cfg.ContainerID)
-	os.Setenv("ECS_CNI_LOGLEVEL", logger.GetLevel())
-	defer os.Unsetenv("ECS_CNI_LOGLEVEL")
 
-	// Invoke eni plugin ADD command
-	result, err := client.add(runtimeConfig, cfg, client.createENINetworkConfig)
-	if err != nil {
-		return nil, errors.Wrap(err, "cni setup: invoke eni plugin failed")
+	// Invoke eni plugin ADD command based on the type of eni plugin
+	if cfg.ENIType == apieni.BranchENIType {
```
suneyz: how do you think about the following instead to reduce code duplication? same with delete logic

```golang
var eniNetworkConfig
if eni-trunking {
  eniNetworkConfig = client.createBranchENINetworkConfig
// maybe also set env variable here
} else {
  eniNetworkConfig = client.createENINetworkConfig
}
// setup ENI
os.setenv(xxxx)
result, err := client.add(runtimeConfig, cfg, eniNetworkConfig)
if err != nil {
  return nil, errors.Wrap(err, "cni setup: invoke %v plugin failed", eniNetworkConfig.Type)
}
seelog.Debugf(xxx)
```

agent/ecscni/plugin.go
```diff
@@ -181,20 +200,41 @@ func (client *cniClient) cleanupNS(cfg *Config) error {
...
+
+		err = client.del(runtimeConfig, cfg, client.createENINetworkConfig)
+		if err != nil {
+			return errors.Wrap(err, "cni cleanup: invoke eni plugin failed")
+		}
+		seelog.Debugf("[ECSCNI] container namespace cleanup done: %s", cfg.ContainerID)
+
 	}
+
+
```
suneyz: nit: extra space

agent/api/ecsclient/client.go
```diff
@@ -558,4 +558,4 @@ func (client *APIECSClient) GetResourceTags(resourceArn string) ([]*ecs.Tag, err
 		return nil, err
 	}
 	return output.Tags, nil
-}
+}
```
suneyz: nit: you removed the [extra line](https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file)

agent/acs/handler/payload_handler.go
```diff
@@ -322,12 +320,12 @@ func (payloadHandler *payloadRequestHandler) ackCredentials(messageID *string, c
...
 type skipAddTaskComparatorFunc func(apitaskstatus.TaskStatus) bool
 
-// isTaskStatusStopped returns true if the task status == STOPPTED
+// isTaskStatusStopped returns true if the task status == STOPPED
 func isTaskStatusStopped(status apitaskstatus.TaskStatus) bool {
 	return status == apitaskstatus.TaskStopped
 }
 
-// isTaskStatusNotStopped returns true if the task status != STOPPTED
+// isTaskStatusNotStopped returns true if the task status != STOPPED
```
suneyz: good catch on these 👍

agent/ecscni/plugin.go
```diff
@@ -111,18 +114,34 @@ func (client *cniClient) setupNS(cfg *Config) (*current.Result, error) {
...
-	defer os.Unsetenv("ECS_CNI_LOGLEVEL")
 
-	// Invoke eni plugin ADD command
-	result, err := client.add(runtimeConfig, cfg, client.createENINetworkConfig)
-	if err != nil {
-		return nil, errors.Wrap(err, "cni setup: invoke eni plugin failed")
+	// Invoke eni plugin ADD command based on the type of eni plugin
+	if cfg.ENIType == apieni.BranchENIType {
+		os.Setenv("VPC_CNI_LOG_LEVEL", logger.GetLevel())
+		os.Setenv("VPC_CNI_LOG_FILE", vpcCNIPluginPath)
```
suneyz: appmesh plugin got around this environment variable by [here](https://github.com/aws/amazon-vpc-cni-plugins/blob/master/plugins/aws-appmesh/plugin/plugin.go#L27)

agent/ecscni/plugin.go
shubham2892: Fixed it.

agent/ecscni/plugin.go
shubham2892: VPC CNI plugin uses this environment variable [here](https://github.com/aws/amazon-vpc-cni-plugins/blob/master/logger/logger.go#L24-L26).

agent/ecscni/plugin.go
shubham2892: Fixed.

agent/ecscni/plugin.go
shubham2892: This might not work because we are setting different environment variable and setting different error message in the IF blocks. We definitely can save a line when we are setting result, err but I think there is value in keeping it that way in favor of readability.

agent/ecscni/plugin.go
shubham2892: Discussed offline, going forward with the environment variable way of setting the log path for now.

agent/statemanager/state_manager.go
```diff
@@ -83,7 +83,13 @@ const (
 	//   b) Add 'StartTime' field to 'api.container.Container'
 	//   c) Add 'StopTime' field to 'api.container.Container'
 	// 21) Add 'target' field to the Secret struct
-	// 22) Add 'attachmentType' field to 'api.ENIAttachment'
+	// 22)
+	// 	 a) Add 'attachmentType' field to 'api.ENIAttachment'
+	//	 b) Add 'InterfaceAssociationProtocol' field to 'api.ENI'
+	//	 c) Add 'InterfaceVlanProperties' field to 'api.ENI'
+	//	 d) Add 'NetworkInterfaceVlanProperties' struct
```
fenxiong: d) and e) are not needed since they are structs for acs model and won't be saved in state file

agent/api/eni/eni.go
```diff
@@ -171,5 +213,17 @@ func ValidateTaskENI(acsenis []*ecsacs.ElasticNetworkInterface) error {
 		return errors.Errorf("eni message validation: empty eni id in the message")
 	}
 
+	if (acsenis[0].InterfaceAssociationProtocol != nil) && (aws.StringValue(acsenis[0].InterfaceAssociationProtocol) !=
+		VLANInterfaceAssociationProtocol) && (aws.StringValue(acsenis[0].InterfaceAssociationProtocol) != DefaultInterfaceAssociationProtocol) {
+		return errors.Errorf("Invalid ENI interface type: %s", aws.StringValue(acsenis[0].InterfaceAssociationProtocol))
```
fenxiong: error message shouldn't be capitalized

agent/engine/docker_task_engine_linux_test.go
```diff
@@ -1,4 +1,4 @@
-// +build linux,unit
```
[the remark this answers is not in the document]
shubham2892: This is a mistake, should not have been removed.

agent/ecscni/plugin.go
```diff
@@ -34,9 +36,11 @@ import (
 const (
 	currentCNISpec = "0.3.1"
 	// ECSCNIVersion, ECSCNIGitHash, VPCCNIGitHash needs to be updated every time CNI plugin is updated
-	currentECSCNIVersion = "2018.10.0"
-	currentECSCNIGitHash = "93f4377604504bff92e7555da73b0cba732a4fbb"
-	currentVPCCNIGitHash = "76973f587f9dfb0b40092a78fb5623c9547bc647"
+	currentECSCNIVersion      = "2018.10.0"
+	currentECSCNIGitHash      = "93f4377604504bff92e7555da73b0cba732a4fbb"
+	currentVPCCNIGitHash      = "8b5e552209b0009aecf3c60568b5a5041c6342c0"
```
suneyz: Non-blocker, but will save you future effort to merge to dev. Can you merge the change from mainline instead of updating here to avoid merge conflict from your branch to dev? 

This is updated from this [commit](https://github.com/aws/amazon-ecs-agent/commit/6526bdfab6a52e490b39b2931c0b665a53ab99f1)

shubham2892: Merging this to `eni-trunking` branch, will change this directly in that branch after all the trunking related changes are merged to that branch so that our unit tests keep passing during development. Thanks for the heads up. 😄

agent/ecscni/plugin.go
ofiliz: We can change the log config behavior to match the ecs-cni plugin if that is going to result in simpler agent code.

agent/ecscni/plugin.go
```diff
@@ -349,6 +389,32 @@ func (client *cniClient) createAppMeshConfig(cfg *Config) (string, *libcni.Netwo
 	return defaultAppMeshIfName, networkConfig, nil
 }
 
+func (client *cniClient) createBranchENINetworkConfig(cfg *Config) (string, *libcni.NetworkConfig, error) {
+
+	// cfg.ENIIPV4Address does not have a prefix length while BranchIPAddress expects a prefix length
```
ofiliz: I believe I had to write this prefix length computation code when implementing the proof of concept. But I was hoping that we could have the control plane send the IP addresses in the correct format in production version. I'm sure there was a good reason why CP sends these addresses in the format it does today, but technically the prefix length is an attribute of the task IP address. On the other hand, we don't care what the prefix length of the gateway is; it is just the next hop of the default route.

Is this a good opportunity to iron out these issues?

fenxiong: if we are going to change the format, backend will need to have some logic like "if this is not eni trunking task, use old format, else use new format". not sure if they will be happy with such logic. we can discuss with them about this though

PR #1876, opened 2019-02-21
engine: adding poll function during progressTask

petderek:
### Summary

This ensures that SUCCESS and HEALTHY conditions are reachable
when progressTask sees no activity.

### Implementation details

Prior to this commit, we only tracked state we explicitly tried to
change when the task was starting. We did not respond to the event
stream or any other source of information from Docker. This means
that when we are waiting for certain dependency conditions
("SUCCESS", "COMPLETE", or "HEALTHY") the task progression logic
does not update the agent-internal model of container state.

Since we rely on that state for determining when conditions are
met, tasks would get stuck in infinite startup loops. This change
adds a call to engine.checkTaskState(), which explicity updates
any changed container state. We only call this function if we know
that we are waiting on the aforementioned subset of dependency
conditions.

### Testing
We manually tested use cases involving "success" and "healthy".

Integ tests are on the way.

New tests cover the changes: **WIP**, will be added to this PR or another one depending on how lengthly they get.

### Licensing

### DIFF
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -105,7 +105,7 @@ func dependenciesCanBeResolved(target *apicontainer.Container, by []*apicontaine
 		nameMap[cont.Name] = cont
 	}
 
-	if err := verifyContainerOrderingStatusResolvable(target, nameMap, containerOrderingDependenciesCanResolve); err != nil {
+	if _, err := verifyContainerOrderingStatusResolvable(target, nameMap, containerOrderingDependenciesCanResolve); err != nil {
 		return false
 	}
 	return verifyStatusResolvable(target, nameMap, target.SteadyStateDependencies, onSteadyStateCanResolve)
@@ -122,9 +122,9 @@ func DependenciesAreResolved(target *apicontainer.Container,
 	by []*apicontainer.Container,
 	id string,
 	manager credentials.Manager,
-	resources []taskresource.TaskResource) error {
+	resources []taskresource.TaskResource) (*apicontainer.DependsOn, error) {
 	if !executionCredentialsResolved(target, id, manager) {
-		return CredentialsNotResolvedErr
+		return nil, CredentialsNotResolvedErr
 	}
 
 	nameMap := make(map[string]*apicontainer.Container)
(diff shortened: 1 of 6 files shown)


agent/engine/dependencygraph/graph.go
```diff
@@ -205,25 +209,25 @@ func verifyStatusResolvable(target *apicontainer.Container, existingContainers m
...
 		// wants regardless of what dependencies it has
-		return nil
+		return nil, nil
 	}
 
 	for _, dependency := range target.DependsOn {
 		dependencyContainer, ok := existingContainers[dependency.Container]
 		if !ok {
-			return fmt.Errorf("dependency graph: container ordering dependency [%v] for target [%v] does not exist.", dependencyContainer, target)
+			return nil,fmt.Errorf("dependency graph: container ordering dependency [%v] for target [%v] does not exist.", dependencyContainer, target)
```
ubhattacharjya: Not sure but do you want to return the &dependency in here?

agent/engine/dependencygraph/graph.go
```diff
@@ -105,7 +105,7 @@ func dependenciesCanBeResolved(target *apicontainer.Container, by []*apicontaine
 		nameMap[cont.Name] = cont
 	}
 
-	if err := verifyContainerOrderingStatusResolvable(target, nameMap, containerOrderingDependenciesCanResolve); err != nil {
+	if _,err := verifyContainerOrderingStatusResolvable(target, nameMap, containerOrderingDependenciesCanResolve); err != nil {
```
ubhattacharjya: nit: I think there will be a space after ,

agent/engine/dependencygraph/graph.go
petderek: I don't think so, since this function is only validating that its possible to resolve.

agent/engine/task_manager.go
```diff
@@ -740,16 +742,31 @@ func (mtask *managedTask) progressTask() {
...
 			transitionChange <- struct{}{}
 			transitionChangeEntity <- container.Name
 		})
 
-	if !anyContainerTransition && !anyResourceTransition {
-		if !mtask.waitForExecutionCredentialsFromACS(reasons) {
-			mtask.onContainersUnableToTransitionState()
+	atLeastOneTransitionStarted := anyResourceTransition || anyContainerTransition
+
+	if !atLeastOneTransitionStarted && len(blockedTransitions) == 0 {
```
aaithal: Can you please add comments around this. What are we checking? And why?

agent/engine/task_manager.go
```diff
@@ -740,16 +742,31 @@ func (mtask *managedTask) progressTask() {
...
+		}
+		return
+	}
+
+	// If no containers are starting and we are blocked on dependency conditions, we should watch for the task to change
+	// over time. This will update the containers if they become healthy or stop, which makes it possible for the
+	// conditions "HEALTHY" and "SUCCESS" to succeed.
+	if !atLeastOneTransitionStarted && len(blockedTransitions) > 0 {
+		go mtask.engine.checkTaskState(mtask.Task)
+		ctx, cl := context.WithTimeout(context.Background(), transitionPollTime)
```
aaithal: Please use `cancel`. I don't think we're saving much in 3 letters.

petderek:
Merging this, will do integ tests on a different PR before we merge this to dev

PR #1878, opened 2019-02-22
Validating that the dependency container has not timed out

ubhattacharjya:
Here, the time duration(StartTimeout) mentioned by the user for a container is expired or not is checked before resolving the dependency for target container.

For example,
* if a target container 'A' has dependency on container 'B' and the dependency condiiton is 'SUCCESS', then the dependency will not be resolved if B times out before exitting successfully with exit code 0.
* if a target container 'A' has dependency on container 'B' and the dependency condiiton is 'COMPLETE', then the dependency will not be resolved if B times out before exitting.
* if a target container 'A' has dependency on container 'B' and the dependency condiiton is 'HEALTHY', then the dependency will not be resolved if B times out before emtting 'Healthy' status.

The advantage of this is that the user will get to know that something is wrong with the task if the task is stuck in pending.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -23,6 +23,7 @@ import (
 	log "github.com/cihub/seelog"
 	"github.com/pkg/errors"
 	"strings"
+	"time"
 )
 
 const (
@@ -208,6 +209,7 @@ func verifyContainerOrderingStatusResolvable(target *apicontainer.Container, exi
 	resolves func(*apicontainer.Container, *apicontainer.Container, string) bool) (*apicontainer.DependsOn, error) {
 
 	targetGoal := target.GetDesiredStatus()
+	targetKnown := target.GetKnownStatus()
 	if targetGoal != target.GetSteadyStateStatus() && targetGoal != apicontainerstatus.ContainerCreated {
 		// A container can always stop, die, or reach whatever other state it
 		// wants regardless of what dependencies it has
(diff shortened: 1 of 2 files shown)


agent/engine/dependencygraph/graph.go
```diff
@@ -367,6 +372,15 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
 	}
 }
 
+func hasDependencyTimedOut(dependOnContainer *apicontainer.Container, dependencyCondition string) bool {
+	if dependencyCondition == successCondition || dependencyCondition == completeCondition || dependencyCondition == healthyCondition {
```
petderek: Couple things:

* I feel like this should return "true" when there is a timeout.

* I'd rather use a switch:
```
switch  dependencyCondition {
case success,complete,healthy:
   // it _has_ timed out if this returns true
    return time.Now().Af....
default:
   // it has not timed out
    return false
}
```

agent/engine/dependencygraph/graph.go
```diff
@@ -219,6 +220,10 @@ func verifyContainerOrderingStatusResolvable(target *apicontainer.Container, exi
 		if !ok {
 			return fmt.Errorf("dependency graph: container ordering dependency [%v] for target [%v] does not exist.", dependencyContainer, target)
 		}
+		if !hasDependencyTimedOut(dependencyContainer, dependency.Condition) {
```
petderek: If you take my suggestion to make the function return true when there is a time out, you can get rid of the `!`

agent/engine/dependencygraph/graph.go
```diff
@@ -367,6 +372,15 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
 	}
 }
 
+func hasDependencyTimedOut(dependOnContainer *apicontainer.Container, dependencyCondition string) bool {
+	switch dependencyCondition {
+	case successCondition, completeCondition, healthyCondition:
+		return time.Now().After(dependOnContainer.GetStartedAt().Add(dependOnContainer.GetStartTimeout()))
```
sharanyad: what happens when `GetStartedAt` is not set?

petderek: If we call this after "resolves" then this shouldn't happen. But more defensive programming is a good idea too.

ubhattacharjya: Thanks Sharanya and Derek for the useful inputs. I have added both the checks.

agent/engine/dependencygraph/graph.go
```diff
@@ -367,6 +379,18 @@ func containerOrderingDependenciesIsResolved(target *apicontainer.Container,
 	}
 }
 
+func hasDependencyTimedOut(dependOnContainer *apicontainer.Container, dependencyCondition string) bool {
+	switch dependencyCondition {
+	case successCondition, completeCondition, healthyCondition:
+		if !dependOnContainer.GetStartedAt().IsZero() && dependOnContainer.GetStartTimeout() > 0 {
```
petderek: Can you have both of these conditions follow the same idiom? Either IsZero or comparison is fine.

You can reduce indentation by putting this above the switch:

```
if dependOnContainer.GetStartedAt().IsZero() || dependOnContainer.GetStartTimeout().IsZero()
   return false
}

switch dependencyCondition {
```

ubhattacharjya: isZero() checks if the time.Time value is the 0th time for golang. and on the other hand, GetStartTimeout () returns a time duration. That is why, isZero() is not valid for GetStartTimeout (as it returns time.Duration) and >0 cannot be used for GetStartedAt (as it returns time.Time).

PR #1880, opened 2019-02-22
Remove the functionality of StartTimeout as Docker API Start Timeout

ubhattacharjya:
The ‘StartTimeout’ now will only serve as the the time duration after which if a container has a dependency on another container and the conditions are ‘SUCCESS’, ‘HEALTHY’ and ‘COMPLETE’, then the dependency will not be resolved. For example:

•	If a container A has a dependency on container B and the condition is ‘START’, the StartTimeout for container B will roughly be the time required for it to exit successfully with exit code 0
•	If a container A has a dependency on container B and the condition is ‘COMPLETE’, the StartTimeout for container B will roughly be the time required for it to exit.
•	If a container A has a dependency on container B and the condition is ‘HEALTHY’, the StartTimeout for container B will roughly be the time required for it to emit a ‘HEALTHY’ status.

If the StartTimeout exceeds in any of the above cases, container A will not be able to transition to ‘CREATED’ status.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -955,13 +955,7 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
 		}
 	}
 	startContainerBegin := time.Now()
-
-	ctxTimeoutStartContainer := container.GetStartTimeout()
-	if ctxTimeoutStartContainer <= 0 {
-		ctxTimeoutStartContainer = engine.cfg.ContainerStartTimeout
-	}
-
-	dockerContainerMD := client.StartContainer(engine.ctx, dockerContainer.DockerID, ctxTimeoutStartContainer)
+	dockerContainerMD := client.StartContainer(engine.ctx, dockerContainer.DockerID, engine.cfg.ContainerStartTimeout)
 
 	// Get metadata through container inspection and available task information then write this to the metadata file
 	// Performs this in the background to avoid delaying container start
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -140,8 +140,9 @@ type Container struct {
 	// LogsAuthStrategy specifies how the logs driver for the container will be
 	// authenticated
 	LogsAuthStrategy string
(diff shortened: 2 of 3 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -955,13 +955,7 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
...
 	}
 	startContainerBegin := time.Now()
-
-	ctxTimeoutStartContainer := container.GetStartTimeout()
-	if ctxTimeoutStartContainer <= 0 {
-		ctxTimeoutStartContainer = engine.cfg.ContainerStartTimeout
-	}
-
-	dockerContainerMD := client.StartContainer(engine.ctx, dockerContainer.DockerID, ctxTimeoutStartContainer)
+	dockerContainerMD := client.StartContainer(engine.ctx, dockerContainer.DockerID, engine.cfg.ContainerStartTimeout)
```
yunhee-l: is my understanding correct?
container.GetStartTimeout() (returns StartTimeout): for container dependency timing out container start
engine.cfg.ContainerStartTimeout: configured value for docker API timeout; this was used as default for above prior to this fix - if so, what about it was the problem?

ubhattacharjya: Yes, initially we thought that we will utilize this timeout as timeout for Docker API startContainer timeout as well to check whether the dependency container has timed out or not for container ordering. But now that we think about it, it might become a problem. For example,

If container A depends on container B and the condition is 'SUCCESS', the container A cannot start before B exits successfully. If B takes 2 days to exit, then StartTimeout will roughly be 2 days. Now suppose, Docker API StartContainer call fails, the customer will get to know about it after 2 days. That is why, this change is made in order to fail fast if startContainer Docker API call fails.

Related PR: https://github.com/aws/amazon-ecs-agent/pull/1878

petderek:
Arm tests are passing manually. We will investigate this before we merge to dev.

Issue #1856, opened 2019-02-14
testing: TestStateManager failing intermittently with timeouts


suneyz:
Flaky unit test. Sometimes it will timeout.

### Supporting Log Snippets

````
=== RUN   TestStateManager
2019-01-29T23:30:10Z [INFO] Loading state! module="statemanager"
2019-01-29T23:30:10Z [INFO] Saving state! module="statemanager"
panic: test timed out after 30s

goroutine 69 [running]:
testing.startAlarm.func1()
    /usr/local/go/src/testing/testing.go:1145 +0x139
created by time.goFunc
    /usr/local/go/src/time/sleep.go:170 +0x52

goroutine 1 [chan receive]:
testing.(*T).Run(0xc420374000, 0xd5d0d4, 0x10, 0xd7db28, 0xc42024db01)
    /usr/local/go/src/testing/testing.go:790 +0x59b
testing.runTests.func1(0xc420374000)
    /usr/local/go/src/testing/testing.go:1004 +0xa8
testing.tRunner(0xc420374000, 0xc42024dd28)
    /usr/local/go/src/testing/testing.go:746 +0x16d
testing.runTests(0xc4202a10c0, 0x11da120, 0xb, 0xb, 0xc4200a2c48)
    /usr/local/go/src/testing/testing.go:1002 +0x522
testing.(*M).Run(0xc4204d1f20, 0x30)
    /usr/local/go/src/testing/testing.go:921 +0x207
main.main()
    github.com/aws/amazon-ecs-agent/agent/statemanager/_test/_testmain.go:118 +0x2f1

goroutine 5 [semacquire]:
sync.runtime_notifyListWait(0xc420057250, 0x0)
    /usr/local/go/src/runtime/sema.go:507 +0x11f
sync.(*Cond).Wait(0xc420057240)
    /usr/local/go/src/sync/cond.go:56 +0x8e
github.com/aws/amazon-ecs-agent/agent/vendor/github.com/cihub/seelog.(*asyncLoopLogger).processItem(0xc420162900, 0x0)
    /opt/amazon-ecs-agent/go/src/github.com/aws/amazon-ecs-agent/agent/vendor/github.com/cihub/seelog/behavior_asynclooplogger.go:50 +0x108
github.com/aws/amazon-ecs-agent/agent/vendor/github.com/cihub/seelog.(*asyncLoopLogger).processQueue(0xc420162900)
[...]

PR #1884, opened 2019-02-24
Checking dependency resolution after timeout and successful exit check

ubhattacharjya:
Initially, we were first checking if the dependency is resolved. If resolved, then we checked if the dependency timed out or exitted unsuccessfully. But later we realized that, in doing so we are not checking the timeout and succssfully exitted or not if dependency not resolved. That is why, moved the timeout check and successfully completed check before checking whether the dependency has resolved or not.

### Testing

New tests cover the changes: 

### Licensing


petderek:
Could you rework your commit message?

ubhattacharjya:
Flaky test TestStateManager failing https://github.com/aws/amazon-ecs-agent/issues/1856

ubhattacharjya:
linux TestInitOOMEvent
windows integ TestSharedAutoprovisionVolume
windows functional  TestOOMContainer

Will adress these tests before merging into dev...

PR #1889, opened 2019-02-26
Adding functional tests for container ordering

ubhattacharjya:
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/container-ordering/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/container-ordering/task-definition.json
@@ -1,0 +1,77 @@
+{
+  "family": "ecsftest-container-ordering",
+  "containerDefinitions": [
+    {
+      "name": "success",
+      "image": "127.0.0.1:51670/busybox:latest",
+      "cpu": 0,
+      "memory": 64,
+      "essential": true,
+      "command": ["sh", "-c", "sleep 50"],
+      "dependsOn": [
+        {
+          "containerName": "success-dependency",
+          "condition": "SUCCESS"
+        }
+      ]
+    },
+    {
+      "name": "success-dependency",
+      "image": "127.0.0.1:51670/busybox:latest",
+      "cpu": 0,
+      "memory": 64,
+      "essential": false,
+      "command": ["sh", "-c", "sleep 1"]
+    },
+    {
+      "name": "complete",
+      "image": "127.0.0.1:51670/busybox:latest",
+      "cpu": 0,
+      "memory": 64,
+      "essential": false,
+      "command": ["sh", "-c", "sleep 1"],
+      "dependsOn": [
+        {
+          "containerName": "complete-dependency",
+          "condition": "COMPLETE"
+        }
+      ]
+    },
+    {
+      "name": "complete-dependency",
+      "image": "127.0.0.1:51670/busybox:latest",
+      "cpu": 0,
+      "memory": 64,
+      "essential": false,
+      "command": ["sh", "-c", "sleep 1"]
+    },
+    {
+      "name": "healthy",
+      "image": "127.0.0.1:51670/busybox:latest",
+      "cpu": 0,
+      "memory": 64,
+      "essential": false,
+      "command": ["sh", "-c", "sleep 1"],
+      "dependsOn": [
+        {
+          "containerName": "healthy-dependency",
+          "condition": "HEALTHY"
+        }
+      ]
+    },
+    {
+      "name": "healthy-dependency",
+      "image": "127.0.0.1:51670/busybox:latest",
+      "healthCheck": {
+        "command": ["CMD-SHELL", "echo hello"],
+        "interval": 5,
+        "timeout": 50,
+        "retries": 3
+      },
+      "cpu": 0,
+      "memory": 64,
+      "essential": false,
+      "command": ["sh", "-c", "sleep 15"]
+    }
+  ]
+}
--- a/agent/functional_tests/tests/functionaltests_test.go
+++ b/agent/functional_tests/tests/functionaltests_test.go
@@ -566,7 +566,7 @@ func telemetryTest(t *testing.T, taskDefinition string) {
 func telemetryTestWithStatsPolling(t *testing.T, taskDefinition string) {
 	// telemetry task requires 2GB of memory (for either linux or windows); requires a bit more to be stable
 	RequireMinimumMemory(t, 2200)
-	
+
 	// Try to let the container use 25% cpu, but bound it within valid range
 	cpuShare, expectedCPUPercentage := calculateCpuLimits(0.25)
 
--- a/agent/engine/engine_integ_test.go
+++ b/agent/engine/engine_integ_test.go
@@ -42,7 +42,7 @@ import (
 )
 
 const (
-	testDockerStopTimeout  = 2 * time.Second
+	testDockerStopTimeout  = 5 * time.Second
 	credentialsIDIntegTest = "credsid"
 	containerPortOne       = 24751
 	containerPortTwo       = 24752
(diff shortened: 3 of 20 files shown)


agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -948,3 +949,69 @@ func testV3TaskEndpointTags(t *testing.T, taskName, containerName, networkMode s
...
+		t.Fatalf("Could not register task definition: %v", err)
+	}
+
+	var testTasks []*TestTask
+	testTasks, err = agent.StartMultipleTasks(t, td, 1)
+	if err != nil {
+		t.Fatalf("Could not start task: %v", err)
+	}
+
+	timeout, err := time.ParseDuration("5m")
```
suneyz: would 5m be too long for functional test since our overall timeout is 40m from [here](https://github.com/aws/amazon-ecs-agent/blob/master/Makefile#L202)

ubhattacharjya: Cool. I have made it 3m. I have 6 containers in my task def, that is why increased the timeout a bit

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -948,3 +949,69 @@ func testV3TaskEndpointTags(t *testing.T, taskName, containerName, networkMode s
 	exitCode, _ := task.ContainerExitcode(containerName)
 	assert.Equal(t, 42, exitCode, fmt.Sprintf("Expected exit code of 42; got %d", exitCode))
 }
+
+func TestContainerOrdering(t *testing.T) {
+	agent := RunAgent(t, nil)
+	defer agent.Cleanup()
+	agent.RequireVersion(">=1.26.0")
```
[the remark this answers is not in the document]
ubhattacharjya: Thanks Eric. I have made it >=1.25.0

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -948,3 +949,69 @@ func testV3TaskEndpointTags(t *testing.T, taskName, containerName, networkMode s
...
+		t.Fatalf("Could not parse timeout: %#v", err)
+	}
+
+	for _, testTask := range testTasks {
+		err = testTask.WaitStopped(timeout)
+		if err != nil {
+			t.Fatalf("Timed out waiting for task to reach stopped. Error %#v, task %#v", err, testTask)
+		}
+
+		exitCode, _ := testTask.ContainerExitcode("success-timeout-dependency")
```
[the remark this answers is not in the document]
ubhattacharjya: I ignored this error, because I was not using it.

suneyz:
overall lgtm. tests are failing, make sure to have them pass or point failed tests out if they are flaky

agent/functional_tests/tests/functionaltests_test.go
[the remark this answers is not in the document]
petderek: Please keep this at 1.26.0. This won't run on 1.25.0 (since we are at 1.25.3 + changes).  Its true this won't run if you build without updating the version file -- but we will do just that when we go to do the next release.

agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -948,3 +948,69 @@ func testV3TaskEndpointTags(t *testing.T, taskName, containerName, networkMode s
 	exitCode, _ := task.ContainerExitcode(containerName)
 	assert.Equal(t, 42, exitCode, fmt.Sprintf("Expected exit code of 42; got %d", exitCode))
 }
+
+func TestContainerOrdering(t *testing.T) {
```
petderek: This should be included in the simple tests  (generated) since it isn't doing any complex validations here.

petderek:
--- FAIL: TestContainerOrderingTimedout (35.58s)

petderek:
--- FAIL: TestContainerOrdering (0.35s)

petderek:
Rebased and fixed a typo in the json file

petderek:
Checked the failing arm test:
TestPerContainerStopTimeout

Its working when i validate it locally. we will increase the timeout in a separate pr

ubhattacharjya:
Linux Integ test failed because of flaky test TestDockerStateToContainerState Reference: https://github.com/aws/amazon-ecs-agent/issues/1899

ubhattacharjya:
Windows integ tests are failing because of flaky test TestStatsEngineWithNewContainersWithPolling : https://github.com/aws/amazon-ecs-agent/issues/1912

agent/functional_tests/testdata/taskdefinitions/container-ordering/task-definition.json
```diff
@@ -0,0 +1,77 @@
+{
```
aaithal: So, I understand that this task def uses many of the fields necessary for ordering containers. However, I'm unable to understand how running this test ensures that the feature is working correctly. As in, how are we validating that `"success"` is only running when `"success-dependency"` has reached the `"SUCCESS"` state? For all we know, ECS agent could be just ignoring those fields. Shouldn't that functionality testing be part of a functional test?

If you're testing that somewhere and I'm missing that/misreading this test, please point me to the place where such validation is happening.

ubhattacharjya: We are validating that the feature is working fine as part of integ tests in ordering-integ-tests.go. Here we are just validating the happy case.

aaithal: Right, I get that. But, functional tests by definition should test the functionality of the feature that they were written for. I think the task definition should do more than just setting a field and ensuring that the task runs. For example, you could have a `"producer"` container, which creates a file in a shared task volume. The `"consumer"` container, which is ordered to start after the `"producer"` would ensure the existence of the file and that would be a better test than just doing nothing in the task def. If we're not actually testing the functionality of the feature, then there's very little value in adding these tests.

petderek: We used similar hard-coded overrides while testing to ensure that tests wouldn't get stuck in an infinite startup cycle.

Explicitly checking order is something we have been doing in our integration tests.

petderek: I get what you are saying. We'll make an action to add additional checks in our functional workflow before the next release. 

We will be validating this behavior manually over the next few days, too.

aaithal: > We used similar hard-coded overrides while testing to ensure that tests wouldn't get stuck in an infinite startup cycle.
> 
> Explicitly checking order is something we have been doing in our integration tests.

I'm glad that we have integration test coverage for this. But, I'm questioning functional test coverage here. Given that this (functional tests) are considered black-box tests used for testing the acceptance criteria, we can sure strive for a higher bar by having more meaningful task defs? For example, there's a reason this `"awslogs"` test tests that a log entry is available in cw logs endpoint: https://github.com/aws/amazon-ecs-agent/blob/a906f1509da9712b728b0a67f5808fc39a571228/agent/functional_tests/tests/functionaltests_unix_test.go#L318

aaithal: > I get what you are saying. We'll make an action to add additional checks in our functional workflow before the next release.

Why not this release? Why is it ok for the bar to be lower in this release?

> We will be validating this behavior manually over the next few days, too.

Again, glad that you're manually testing this. But, I think it'd be equally good/better use of resources/time to automate the same.

samuelkarp:
The following checks failed, and this was still merged:

* ecs/linux/integ_test Failed: Linux integration test suite
* ecs/windows/functional_test Failed: Windows functional test suite
* ecs/windows/integ_test Failed: Windows integration test suite 

Considering the conversation in https://github.com/aws/amazon-ecs-agent/pull/1889#pullrequestreview-209297654 and the insistence that this is covered in integration tests, the failure of the integration tests is pretty concerning.

ubhattacharjya:
Merging this because all tests involving container ordering tests have passed and the ones failing are because of flaky tests.
Reference:
Linux integ test: https://github.com/aws/amazon-ecs-agent/issues/1899
Windows Integ tes: https://github.com/aws/amazon-ecs-agent/issues/1912
Windows Functional test: https://github.com/aws/amazon-ecs-agent/issues/1903

PR #1897, opened 2019-02-27
Fix error detection case when image that is being deleted does not exist

### LINKED ISSUE #1871: ECS-Agent image deletion stops if non-existant image encountered
### Summary

The ECS-Agent image deletion / clean-up stops with an error if it encounters a reference to an image that was removed manually via Docker. It processes no other candidate images following this error, nor does it ignore the absent image during future clean-up cycles.

### Description

The following command was issued on an EC2 container instance in response to low disk space:
`docker rmi $(docker images -q)`
This was intended to remove all images not currently associated with running or stopped containers.

However, the ECS-Agent now repeatedly looks for these absent images and will error out after not being able to locate the first one. Ex:
```
2019-02-19T20:08:34Z [INFO] Found 52 eligible images for deletion
2019-02-19T20:08:34Z [INFO] Removing Image: mycompany/goms-rest-staging:latest
2019-02-19T20:08:34Z [ERROR] Error removing Image mycompany/goms-rest-staging:latest - Error: No such image: mycompany/goms-rest-staging:latest
```
At this point, the clean-up stops, and no further image deletion is attempted. The result is a steadily increasing number of images that are not used by running containers.

### Expected Behavior
Ideally, the ECS-Agent should recognize that the image is no longer available and continue to the next candidate image. In addition, it should remove the image from its tracking list such that it is not referenced again as a "candidate".

Another acceptable alternative would be to reset the agent's image tracking list such that it only tracked images related to managed containers on start-up

shubham2892:
### Summary
Fixes #1871 

When the agent tries to remove an image that does not exist on the instance, the error message received back by the agent is `no such image: <image name>`.

Agent tries to match the error message with `no such image` instead of `no such image: <image name>`. As a result, agent is not able to correctly detect that the error thrown back is because the image does not exist.

### Testing

Test `TestDeleteImageNotFoundError` modified to mock correct error message and fails without this change in place. 

### Licensing

### DIFF
--- a/agent/engine/docker_image_manager_test.go
+++ b/agent/engine/docker_image_manager_test.go
@@ -1064,7 +1064,9 @@ func TestDeleteImage(t *testing.T) {
 	}
 }
 
-func TestDeleteImageNotFoundError(t *testing.T) {
+// This test tests that we detect correctly in agent when the agent is trying to delete image that
+// does not exist for older version of docker. 
+func TestDeleteImageNotFoundOldDockerMessageError(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 	client := mock_dockerapi.NewMockDockerClient(ctrl)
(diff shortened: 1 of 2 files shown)


agent/engine/docker_image_manager_test.go
```diff
@@ -1084,7 +1084,7 @@ func TestDeleteImageNotFoundError(t *testing.T) {
 	}
 	imageState, _ := imageManager.getImageState(imageInspected.ID)
 	client.EXPECT().RemoveImage(gomock.Any(), container.Image, dockerclient.RemoveImageTimeout).Return(
-		errors.New("no such image"))
+		errors.New("no such image: " + container.Image))
```
samuelkarp: Can you test the old behavior as well?  We want to ensure that we don't have a regression against older versions of Docker that may return "no such image" instead of including the suffix.

shubham2892:
ARM tests are failing because of this https://github.com/nginxinc/docker-nginx/issues/313

PR #1905, opened 2019-02-28
Fix windows validator test

petderek:
### Summary
We switched the pull behavior to prefer-cached on windows tests. However, the metadata validator explicitly validates that PullStartedAt is set in all cases. This will prevent the test from failing, but we need a better plan for the long term.

### Licensing


petderek:
Arm tests are expected to fail pending Yumeng's fix.

Windows unit tests are passing in travis.

sharanyad:
can you add a TODO in the code that this needs to be tested for other pull behavior types too? this might be lost in the PR

sharanyad:
nvm, found it :)

PR #1886, opened 2019-02-24
Faster windows test

petderek:
### Summary
Reduces windows integ test run to take ~45 minutes.
Functional test run still takes a significant amount of time.

### Implementation details
* Removed dependency on python, golang, golang:nanoserver and nanoserver docker containers
* Removed need to pull windowsservercore:latest so that docker will prefer the cached 

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #1869, opened 2019-02-19
testing: Windows functional tests failing intermittently


ubhattacharjya:
### Summary

The failing tests on windows are the following:

TestV3TaskEndpointDefaultNetworkMode
TestV3TaskEndpointTags
TestTaskIamRolesDefaultNetworkMode
TestOOMContainer

### Supporting Log Snippets

````
=== RUN   TestTaskIamRolesDefaultNetworkMode
--- FAIL: TestTaskIamRolesDefaultNetworkMode (71.65s)
	utils_windows.go:87: Registry location test_path_1550543200333825500
	utils_windows.go:97: datadir C:\Users\ADMINI~1\AppData\Local\Temp\amazon-ecs-agent\ftest_temp\ecs_integ_testdata961815368\data
	utils_windows.go:105: Created directory C:\Users\ADMINI~1\AppData\Local\Temp\amazon-ecs-agent\ftest_temp\ecs_integ_testdata961815368 to store test data in
	utils.go:175: Found agent metadata: {Cluster:ecs-functional-tests ContainerInstanceArn:0xc0424ef1e0 Version:Amazon ECS Agent - v1.25.2 (*UNKNOWN)}
	utils.go:196: Task definition: ecsftest-iamrole-test-86b074466104c867ca8c9f87c5759030:1
	utils.go:216: Started task: arn:aws:ecs:us-west-2:535959970326:task/5f579266-56b4-427f-898a-9912265f9267
	functionaltests_windows_test.go:181: Container exit code non-zero: 1
	utils.go:183: Preserving test dir for failed test C:\Users\ADMINI~1\AppData\Local\Temp\amazon-ecs-agent\ftest_temp\ecs_integ_testdata961815368
=== RUN   TestV3TaskEndpointDefaultNetworkMode
--- FAIL: TestV3TaskEndpointDefaultNetworkMode (76.86s)
	utils_windows.go:87: Registry location test_path_1550543272213645900
	utils_windows.go:97: datadir C:\Users\ADMINI~1\AppData\Local\Temp\amazon-ecs-agent\ftest_temp\ecs_integ_testdata168872455\data
	utils_windows.go:105: Created directory C:\Users\ADMINI~1\AppData\Local\Temp\amazon-ecs-agent\ftest_temp\ecs_integ_testdata168872455 to store test data in
[...]

PR #1919, opened 2019-03-01
Fix Windows Functional Tests

fenxiong:
### Summary

Fix TestOOMContainer, TestTaskIamRolesDefaultNetworkMode, TestV3TaskEndpointDefaultNetworkMode and TestV3TaskEndpointTags. As per https://github.com/aws/amazon-ecs-agent/issues/1869.

Also bring in one of derek's windows integ test fix into dev (https://github.com/aws/amazon-ecs-agent/pull/1919/commits/e80ccb86c90aa5fefbd8bf175c57aabc9119d4ce).

#### TestOOMContainer
TestOOMContainer runs a task that's expected to fail fast because the container tries to use too much memory. The test had a check where it was waiting for the task to reach running, but since the task stops quickly, it's possible that when the test describes the task, the task has already stopped and the test fails with the following:
```
Error Trace:	functionaltests_windows_test.go:214
Error:      	Received unexpected error:
        	Task terminal; will never reach RUNNING
```

 Fixing by removing the wait running check.

#### TestTaskIamRolesDefaultNetworkMode / TestV3TaskEndpointDefaultNetworkMode / TestV3TaskEndpointTags
There was a race condition in setting up credential port forwarding in which the adapter has multiple addresses before the new ip is assigned. Allowing this race condition can cause the netsh interface portproxy setup to result in the network adapter in an unrecoverable bad state where the proxy is unable to listen on port 80 because the port is in use by a non-functional proxy rule.

This race condition happened rarely before, but started becoming a problem when changes were made to let windows test run faster (#1886), which let TestV3TaskEndpointDefaultNetworkMode, TestV3TaskEndpointTags and TestV3TaskEndpointDefaultNetworkMode have a high failure rate because they need the credential port forwarding.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/ordering_integ_test.go
+++ b/agent/engine/ordering_integ_test.go
@@ -43,16 +43,17 @@ func TestDependencyHealthCheck(t *testing.T) {
 	dependency := createTestContainerWithImageAndName(baseImageForOS, "dependency")
 
 	parent.EntryPoint = &entryPointForOS
-	parent.Command = []string{"sleep 5 && exit 1"}
+	parent.Command = []string{"exit 0"}
 	parent.DependsOn = []apicontainer.DependsOn{
 		{
 			ContainerName: "dependency",
 			Condition:     "HEALTHY",
 		},
 	}
+	parent.Essential = true
 
 	dependency.EntryPoint = &entryPointForOS
-	dependency.Command = []string{"sleep 60 && exit 0"}
+	dependency.Command = []string{"sleep 90"}
 	dependency.HealthCheckType = apicontainer.DockerHealthCheckType
 	dependency.DockerConfig.Config = aws.String(alwaysHealthyHealthCheckConfig)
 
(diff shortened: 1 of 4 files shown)


fenxiong:
Failing windows integ test: TestDependencyHealthCheck
Failing windows functional tests: TestTaskIamRolesDefaultNetworkMode, TestV3TaskEndpointDefaultNetworkMode, TestV3TaskEndpointTags
Failing arm functional test: TestAWSLogsDriverMultilinePattern

yunhee-l:
TestTaskIamRolesDefaultNetworkMode: https://github.com/aws/amazon-ecs-agent/issues/1869 
TestV3TaskEndpointDefaultNetworkMode, TestV3TaskEndpointTags, TestTaskIamRolesDefaultNetworkMode: https://github.com/aws/amazon-ecs-agent/issues/1869

petderek:
Good find 👍

PR #1923, opened 2019-03-05
add more forgiving idle percent

fierlion:
### Summary
When checking cloudwatch metrics for a manual run of telemetry test, I found that there are intermittent bursts in idle traffic from 0.0 to 0.062 and 0.011.  
When we check for idle cpu/memory elsewhere we have a range of 0-5%.  This change will make the failure case as forgiving as our other checks.  

### Implementation details
This changes the range for idle and non-idle to include any percentages up to 5% rather than a hard 0.0%

Functional test fix for intermittent failure case.

### Licensing

### DIFF
--- a/agent/functional_tests/util/utils.go
+++ b/agent/functional_tests/util/utils.go
@@ -353,13 +353,14 @@ func VerifyMetrics(cwclient *cloudwatch.CloudWatch, params *cloudwatch.GetMetric
 		return nil, fmt.Errorf("Incorrect SampleCount %f, expected 1", *datapoint.SampleCount)
 	}
 
+	// idleCluster cloudWatch average may have intermittent jumps above 0,
 	if idleCluster {
-		if *datapoint.Average != 0.0 {
-			return nil, fmt.Errorf("non-zero utilization for idle cluster")
+		if *datapoint.Average >= 5.0 {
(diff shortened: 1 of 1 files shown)


sharanyad:
> there are intermittent bursts in idle traffic from 0.0 to 0.062 and 0.011 

Should we then add stricter checks? like fail when `>=2.0`?

suneyz:
> > there are intermittent bursts in idle traffic from 0.0 to 0.062 and 0.011
> 
> Should we then add stricter checks? like fail when `>=2.0`?

Maybe we should also do some experiment(or if it is already done) to justify whatever threshold we are changing. The tests will still fail if the percentage above the one we are changing. 

I'm confident that by changing threshold we will be able to relieve some pain for flaky tests. But if we are changing to a more forgiving threshold without confidence that it will solve flaky test completely or justification, I'd worry that this change will make flaky tests even harder to reproduce to harder to troubleshoot in the future.

fierlion:
// Also verify the cpu usage is around expectedCPUPercentage +/- 5%
	assert.InDelta(t, expectedCPUPercentage*100.0, *metrics.Average, 5)

we use a delta of 5% for our other checks.  This is why I went with 5%.  I agree that this isn't the perfect fix for this issue.

I based this change on multiple manual runs

Issue #1899, opened 2019-02-27
Testing - TestDockerStateToContainerState flaky


suneyz:
### Summary

TestDockerStateToContainerState failed intermittently. 

### Description

````
=== RUN   TestDockerStateToContainerState
2019-02-27T02:29:04Z [INFO] Image excluded from cleanup: amazon/amazon-ecs-agent:latest
2019-02-27T02:29:04Z [INFO] Image excluded from cleanup: amazon/amazon-ecs-pause:0.1.0
2019-02-27T02:29:04Z [WARN] Unable to fetch user data: blackholed
2019-02-27T02:29:04Z [CRITICAL] Unable to communicate with EC2 Metadata service to infer region: blackholed
2019-02-27T02:29:04Z [CRITICAL] Configuration key not set, key: AWSRegion
2019-02-27T02:29:04Z [INFO] Task engine [test_task]: pulling image 127.0.0.1:51670/amazon/amazon-ecs-netkitten:latest for container netcat concurrently
2019-02-27T02:29:04Z [INFO] Task engine [test_task]: recording timestamp for starting image pulltime: 2019-02-27 02:29:04.891014276 +0000 UTC m=+204.493824478
2019-02-27T02:29:04Z [INFO] Adding image name- 127.0.0.1:51670/amazon/amazon-ecs-netkitten:latest to Image state- sha256:3fa2bb1695db7989ffe173baaa9bc6efebf5cc80a353f6891619c50bd40b6ea0
2019-02-27T02:29:04Z [INFO] Updating container reference netcat in Image State - sha256:3fa2bb1695db7989ffe173baaa9bc6efebf5cc80a353f6891619c50bd40b6ea0
2019-02-27T02:29:04Z [INFO] Task engine [test_task]: finished pulling image 127.0.0.1:51670/amazon/amazon-ecs-netkitten:latest for container netcat in 9.630511ms
2019-02-27T02:29:04Z [INFO] Task engine [test_task]: creating container: netcat
2019-02-27T02:29:04Z [INFO] Task engine [test_task]: created container name mapping for task:  netcat -> ecs-family-1-netcat-cc8480a9a1c1f6acf501
2019-02-27T02:29:05Z [INFO] Task engine [test_task]: created docker container for task: netcat -> faea220e1f818a573da361eb57b37a662934c68edc2bb40e4f30a6a3e8f8700c
[...]

Issue #1864, opened 2019-02-18
testing: TestPullImageInactivityTimeout failing intermittently on windows


adnxn:
`TestPullImageInactivityTimeout` appears to be failing intermittently on windows. see below for logs, though they appear to be truncated.

```
2019-02-18T21:14:12Z [WARN] DockerGoClient: failed to pull image image: Could not transition to pullBegin; timed out after waiting 5m0s
2019-02-18T21:14:12Z [WARN] DockerGoClient: failed to pull image image: Could not transition to pullBegin; timed out after waiting 5m0s
2019-02-18T21:14:12Z [WARN] DockerGoClient: failed to pull image image: Could not transition to pullBegin; timed out after waiting 5m0s
2019-02-18T21:14:12Z [WARN] DockerGoClient: failed to pull image image: Could not transition to pullBegin; timed out after waiting 5m0s
2019-02-18T21:14:12Z [WARN] DockerGoClient: failed to pull image image: Could not transition to pullBegin; timed out after waiting 5m0s
2019-02-18T21:14:13Z [WARN] DockerGoClient: inactivity time exceeded timeout while pulling image image
2019-02-18T21:14:13Z [WARN] DockerGoClient: failed to pull image image: inactivity time exceeded timeout while pulling image
2019-02-18T21:14:13Z [WARN] DockerGoClient: inactivity time exceeded timeout while pulling image image
2019-02-18T21:14:13Z [WARN] DockerGoClient: failed to pull image image: inactivity time exceeded timeout while pulling image
2019-02-18T21:14:14Z [WARN] DockerGoClient: inactivity time exceeded timeout while pulling image image
2019-02-18T21:14:14Z [WARN] DockerGoClient: failed to pull image image: inactivity time exceeded timeout while pulling image
--- FAIL: TestPullImageInactivityTimeout (1.72s)
	docker_client_test.go:188: 
			Error Trace:	docker_client_test.go:188
			Error:      	An error is expected but got nil.
			Test:       	TestPullImageInactivityTimeout
			Messages:   	Expected error for pull inactivity timeout
	docker_client_test.go:119: missing call(s) to *mock_sdkclient.MockClient.ImagePull(is anything, is equal to image:latest, is anything) C:/Users/ADMINI~1/AppData/Local/Temp/amazon-ecs-agent/go/src/github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi/docker_client_test.go:175
[...]

PR #1926, opened 2019-03-06
Fix flakey test - TestDockerStateToContainerState

fenxiong:
### Summary

Fix flakey integ test TestDockerStateToContainerState. Issue: https://github.com/aws/amazon-ecs-agent/issues/1899.

### Implementation details

During the workflow of TestDockerStateToContainerState, it starts a container and checks whether it's running by inspecting it. However, the container it starts exits immediately, so it's possible that when we check whether the container is running it's already stopped. This leads to flakey test as per https://github.com/aws/amazon-ecs-agent/issues/1899. Fixing by letting the container keep running rather than exit itself.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/engine_windows_integ_test.go
+++ b/agent/engine/engine_windows_integ_test.go
@@ -66,6 +66,12 @@ func createTestContainer() *apicontainer.Container {
 	}
 }
 
+// getLongRunningCommand returns the command that keeps the container running for the container
+// that uses the default integ test image (microsoft/windowsservercore for windows)
+func getLongRunningCommand() []string {
+	return []string{"ping", "-t", "localhost"}
+}
+
 func createTestHostVolumeMountTask(tmpPath string) *apitask.Task {
 	testTask := createTestTask("testHostVolumeMount")
 	testTask.Volumes = []apitask.TaskVolume{{Name: "test-tmp", Volume: &taskresourcevolume.FSHostVolume{FSSourcePath: tmpPath}}}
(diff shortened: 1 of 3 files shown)


fenxiong:
Looks like TestPullImageInactivityTimeout is still flakey... (https://github.com/aws/amazon-ecs-agent/issues/1864)
```
=== RUN   TestPullImageInactivityTimeout
--- FAIL: TestPullImageInactivityTimeout (0.70s)
    <autogenerated>:1: 
        	Error Trace:	docker_client_test.go:188
        	Error:      	An error is expected but got nil.
        	Test:       	TestPullImageInactivityTimeout
        	Messages:   	Expected error for pull inactivity timeout
    <autogenerated>:1: missing call(s) to *mock_sdkclient.MockClient.ImagePull(is anything, is equal to image:latest, is anything) C:/Users/travis/gopath/src/github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi/docker_client_test.go:175
    <autogenerated>:1: aborting test due to missing call(s)
```

agent/engine/engine_windows_integ_test.go
```diff
@@ -66,6 +66,12 @@ func createTestContainer() *apicontainer.Container {
 	}
 }
 
+// getLongRunningCommand returns the command that keeps the container running for the container
+// that uses the default integ test image (microsoft/windowsservercore for windows)
+func getLongRunningCommand() []string {
+	return []string{"ping", "-t", "localhost"}
```
yunhee-l: Is ping a long running command on Windows?

fenxiong: From the doc [here](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/ping), "-t" specifies that ping continue sending echo Request messages to the destination until interrupted. Without the flag it only sends 4 packets.

Issue #1927, opened 2019-03-07
Can not acquire network metric in EC 2/Bridge mode


hayajo:
### Summary

With #1743 merged in v1.24.0, network metrics can not be acquired in EC2/Bridge mode.

### Description

#1743 changed Stats struct from github.com/fsouza/go-dockerclient to github.com/docker/docker/api/types, but in EC2/Bridge mode network metrics I can not acquire it.

Before:

```
root@ebb5c8c90634:/# curl -s ${ECS_CONTAINER_METADATA_URI}/stats | jq .networks
{
  "eth0": {
    "rx_bytes": 17331985,
    "tx_packets": 932,
    "rx_packets": 1353,
    "tx_bytes": 77755
  }
}
```

After:

```
root@5291861e33b5:/# curl -s ${ECS_CONTAINER_METADATA_URI} | jq .Networks
[
  {
    "NetworkMode": "bridge",
    "IPv4Addresses": [
      "172.17.0.8"
    ]
  }
]

root@5291861e33b5:/# curl -s ${ECS_CONTAINER_METADATA_URI}/stats | jq .networks
null
```

If you change struct to StatsJSON of github.com/docker/docker/api/types, it seems to solve the problem.

How about that?

fenxiong:
Hi @hayajo,
Thanks for reporting the issue. I can reproduce this on my end same as what you described. I will mark this as a bug and we will work on implementing a fix for this.

adnxn:
closing - fixed in [v1.26.1 release](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.26.1)

hayajo:
Thank you! I confirmed that.

PR #1932, opened 2019-03-11
Fix a bug where network stats is not presented in container stats

fenxiong:
### Summary

Fix a bug where network stats are not presented in container stats.

### Implementation details

When switching to use Docker Go SDK, we used [types.Stats](https://github.com/docker/go-docker/blob/master/api/types/stats.go#L153) to save the stats response from corresponding Docker Engine API. This struct doesn't contain all fields in the API response, notably the "networks" field is missing and therefore the field isn't presented in container stats response. This field was populated when we were using fsouza/go-dockerclient prior to v1.24.0. Fixing it by changing to use [types.StatsJSON](https://github.com/docker/go-docker/blob/master/api/types/stats.go#L173) struct which contains the "networks" field along with fields in types.Stats. 

Related issue: https://github.com/aws/amazon-ecs-agent/issues/1927

#### Backward compatibility
Since types.StatsJSON contains strictly more fields than types.Stats, this change should be backward compatible.

#### Docker engine api version
The "networks" field is added in Docker Engine API version 1.21, corresponding to Docker version 1.9.x. Since we already deprecated support for Docker version older than 1.9 we should always expect the field to be presented.

### Testing

The "networks" field will be populated by Docker in bridge network mode on Linux, and nat network mode on Windows. Related v3 metadata functional tests are updated to check that the field is populated in those modes (v2 metadata is not affected since it only works for awsvpc network mode). Also did manual tests to verify that it works.

Bug - Fixed a bug where network stats are not presented in container stats

### Licensing

### DIFF
--- a/misc/v3-task-endpoint-validator/v3-task-endpoint-validator.go
+++ b/misc/v3-task-endpoint-validator/v3-task-endpoint-validator.go
@@ -32,6 +32,7 @@ const (
 )
 
 var isAWSVPCNetworkMode bool
+var isBridgeNetworkMode bool
 var checkContainerInstanceTags bool
 var networkModes map[string]bool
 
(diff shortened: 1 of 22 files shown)


misc/v3-task-endpoint-validator/v3-task-endpoint-validator.go
```diff
@@ -168,12 +176,21 @@ func verifyTaskStats(client *http.Client, taskStatsEndpoint string) error {
...
 	fmt.Printf("Received task stats: %s \n", string(body))
 
-	var taskStats map[string]*types.Stats
+	var taskStats map[string]*types.StatsJSON
 	err = json.Unmarshal(body, &taskStats)
 	if err != nil {
 		return fmt.Errorf("task stats: unable to parse response body: %v", err)
 	}
 
+	if isBridgeNetworkMode {
```
suneyz: curious, do we also want to validate for other network mode or the containerStats.Networks will be blank other than bridge mode?

fenxiong: it's blank in other modes

PR #1933, opened 2019-03-11
add functional tests for appmesh cni plugin

yumex93:
### Summary
Add functional test to test AppMesh CNI plugin to make sure the ip rules can be successfully configured as user defined.

### Testing

New tests cover the changes: 

### Licensing


suneyz:
Instead of bringing in extra vendor files for functional test only. Why not simplify the test logic to build a test image from ubuntu/amazonlinux and install iptables through yum install. Then use linux command to test? This way you might only need ~100 lines of code vs over 1000 lines

misc/appmesh-plugin-validator/appmesh-plugin-validator.go
```diff
@@ -0,0 +1,99 @@
...
+// validateIPRules validates IP rules created in the container network namespace.
+func validateIPRules() error {
+	var iptable *iptables.IPTables
+	var err error
+
+	if iptable, err = iptables.NewWithProtocol(iptables.ProtocolIPv4); err != nil {
+		return err
+	}
+
+	// Print detailed info for debugging.
```
fenxiong: were you planning to use awslogs to surface the output? i didn't see you use it in functional test

petderek:
I’m with Eric on this, unless the validation gets tricky to read.

Another approach is that you could also use go modules to avoid checking in the libraries.

misc/appmesh-plugin-validator/appmesh-plugin-validator.go
yumex93: I haven't thought about this. So storing log on the awslog is for easier to debug?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1456,3 +1456,19 @@ func TestServerEndpointValidator(t *testing.T) {
 	assert.True(t, ok, "Get exit code failed")
 	assert.Equal(t, 42, code, "Wrong exit code")
 }
+
+// TestAppMeshCNIPlugin validates the functionality of appmesh cni plugin.
+func TestAppMeshCNIPlugin(t *testing.T) {
```
suneyz: You could also simplify here through test [generator](https://github.com/aws/amazon-ecs-agent/blob/master/agent/functional_tests/tests/generated/generate_tests.go) to generate this test in this [file](https://github.com/aws/amazon-ecs-agent/blob/master/agent/functional_tests/tests/generated/simpletests_unix/simpletests_generated_unix_test.go)

yumex93:
> I’m with Eric on this, unless the validation gets tricky to read.
> 
> Another approach is that you could also use go modules to avoid checking in the libraries.

I prefer to use go modules. Since for app mesh functionality validation, it is not just one line bash command. I wanna check iptables submodules also. So instead of writing hundred lines of bash script or writing my own iptables go lib. I am leaning towards using current iptables lib.

agent/functional_tests/tests/functionaltests_unix_test.go
yumex93: Good point. Considering I will need to set up a awslog to store container log, this does not apply to my test case anymore.

suneyz: Can you elaborate why do we need awslog here instead of troubleshooting like other tests through the functional output?

suneyz:
Can you squash your commits?

misc/appmesh-plugin-validator/appmesh-plugin-validator.go
petderek: substantially.

yumex93:
> Can you squash your commits?
It is just for testing by myself. Hasn't ready for pr review again. I will add a WIP here. After I finish testing I will squash them.

agent/functional_tests/tests/functionaltests_unix_test.go
yumex93: So the functional output does not give us the insight of what happened within the container. So in order to know what happened in the container based on the container log, I need to upload the container log to CW.

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1456,3 +1456,48 @@ func TestServerEndpointValidator(t *testing.T) {
...
+	}
+
+	agent := RunAgent(t, &AgentOptions{
+		EnableTaskENI: true,
+		ExtraEnvironment: map[string]string{
+			"ECS_AVAILABLE_LOGGING_DRIVERS": `["awslogs"]`,
+		},
+	})
+	defer agent.Cleanup()
+	agent.RequireVersion(">=1.25.3")
```
haikuoliu: Is this feature added in 1.25.3? if not, add a TODO to update the  version.

misc/appmesh-plugin-validator/appmesh-plugin-validator.go
```diff
@@ -0,0 +1,99 @@
...
+	for _, str = range strs {
+		fmt.Printf("Ingress chain stats:%v", str)
+	}
+	strs, _ = iptable.Stats("nat", egressChain)
+	for _, str = range strs {
+		fmt.Printf("Egress chain stats:%v", str)
+	}
+
+	if exist, err := iptable.Exists("nat", ingressChain, "-p", "tcp", "-m", "multiport", "--dports",
+		appPorts, "-j", "REDIRECT", "--to-port", proxyIngressPort); !exist {
```
haikuoliu: If the `exist` is false, do you guarantee to have an non-nil error? You can probably separate the case where `err` is non-nil and the case where `exist` is false.

agent/functional_tests/tests/functionaltests_unix_test.go
yumex93: In 1.26.0, I will add a todo there.

misc/appmesh-plugin-validator/appmesh-plugin-validator.go
yumex93: So based on the logic implemented in the library I used, for this method, only when exist is false, there will be an error(nil/non-nil). When exist is true, the error will always be nil. That's why I wrote it like this way.

yumex93:
testShutdownOrder times out on linux integ test.

yumex93:
Unit test failed on arm: TestHandlerReconnectsOnDiscoverPollEndpointError

PR #1949, opened 2019-03-20
api: set user for pause container when app mesh is enabled

yumex93:
### Summary

If App Mesh is enabled, set the pauseContainer user as the same as proxy container's user, so the traffic coming from pause container can be ignored by proxy.

### Testing

New tests cover the changes: 
yes

### Licensing


agent/api/task/task.go
```diff
@@ -760,16 +764,51 @@ func (task *Task) isNetworkModeVPC() bool {
...
 	pauseContainer.Name = NetworkPauseContainerName
 	pauseContainer.Image = fmt.Sprintf("%s:%s", cfg.PauseContainerImageName, cfg.PauseContainerTag)
 	pauseContainer.Essential = true
 	pauseContainer.Type = apicontainer.ContainerCNIPause
+
+	// Set pauseContainer user the same as proxy container user
+	if task.GetAppMesh() != nil {
+		appMeshConfig := task.GetAppMesh()
+		for _, container := range task.Containers {
+			if container.Name == appMeshConfig.ContainerName {
```
petderek: you can get rid of a layer of indent here if you do:

```
if containerName != appmesh... {
    continue
}

yumex93: Done.

samuelkarp:
As discussed in person yesterday, @nmeyerhans and I both strongly recommend against this approach.  In the absence of user namespaces and UID remapping (which we do not use by default), nothing prevents a given UID/GID from being reused by another container.

samuelkarp:
@petderek, @nmeyerhans and I discussed this offline more today, and we're more comfortable with this approach as it no longer appears to be a one-way door or creating new unexpected behavior (i.e., it fits with existing expectations elsewhere).

yumex93:
TestContainerMetadataFile are failed on linux/arm/windows with known issue.
2019-03-20T17:53:03Z [ERROR] Unable to retrieve Host Instance PublicIPv4 Address: EC2MetadataError: failed to make EC2Metadata request
caused by: <?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>

PR #1950, opened 2019-03-20
Skipping ftest requiring public IP for now

yunhee-l:
We've changed to run tests in private vpc, this will be reviewed later.

### Summary
temporarily skipping testContainerMetadataFile, this gets run in public vpc during ami build

### Implementation details
skip if public IPv4 address is not available

### Testing

New tests cover the changes: NA

### Licensing

### DIFF
--- a/agent/functional_tests/tests/functionaltests_test.go
+++ b/agent/functional_tests/tests/functionaltests_test.go
@@ -759,6 +759,11 @@ func testContainerMetadataFile(t *testing.T, taskName, awslogsPrefix string) {
 	// TODO: Bump version to 1.24.0 (or next release after 1.23.0)
 	agent.RequireVersion(">=1.22.0")
 
+	ec2MetadataClient := ec2.NewEC2MetadataClient(nil)
+	if ec2MetadataClient.PublicIPv4Address() == "" {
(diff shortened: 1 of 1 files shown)


agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -758,6 +758,9 @@ func testContainerMetadataFile(t *testing.T, taskName, awslogsPrefix string) {
 	defer agent.Cleanup()
 	// TODO: Bump version to 1.24.0 (or next release after 1.23.0)
 	agent.RequireVersion(">=1.22.0")
+	if agent.getHostPublicIPv4AddressFromEC2Metadata() == "" {
```
fenxiong: sorry that i missed this.. i don't think the TestAgent struct has this method?

fenxiong: maybe just do something like this - https://github.com/aws/amazon-ecs-agent/blob/ebac2200d4952ff1bb08319fdabbd0e0eb1cebfb/agent/functional_tests/tests/functionaltests_test.go#L817-L819

fenxiong:
closing in favor of https://github.com/aws/amazon-ecs-agent/pull/1958

PR #1954, opened 2019-03-20
Fix for TestStatsEngineWithNewContainersWithPolling

ubhattacharjya:
Stats collection takes some more time to start and that is why
increasing the wait time. After making this change, I ran the test 50 times in both windows and linux and the test passes each time.

### Testing

New tests cover the changes: 

### Licensing


fenxiong:
I read through the test and i think the test itself has a bug.. the configs set here are not being used anywhere: https://github.com/aws/amazon-ecs-agent/blob/8b5e18630375fc2ff656704d6228edfe4d412048/agent/stats/engine_integ_test.go#L345-L346

i think those configs need to be used to configure the dockerClient, which is currently configured via default config. Can you fix the bug for the test?

PR #1955, opened 2019-03-21
add noiseDelta to telemetry test

fierlion:
### Summary
The docker stats reported to CloudWatch have little fluctuations.  Where we expect absolute 0.0% CPU usage now, the stats actually fluctuate.  See testing done below >>
This unifies the delta that we use to account for noise across the entire telemetry test.  

### Implementation details
added a noise constant (5.0).  Updated VerifyMetrics to take this same constant, so we can use it everywhere, and change it only at the const.

### Testing
In my testing, I saw idle cloudwatch metrics for CPU reach anywhere between 0.0 and 2.3% max when I ran the telemetry task.  This means that we will see failures if we query cloudwatch and expect an absolute 0.0% average for an idleCluster.

Add functional test fix

### Licensing

### DIFF
--- a/agent/functional_tests/tests/functionaltests_test.go
+++ b/agent/functional_tests/tests/functionaltests_test.go
@@ -474,6 +474,9 @@ func telemetryTest(t *testing.T, taskDefinition string) {
 	// Try to let the container use 25% cpu, but bound it within valid range
 	cpuShare, expectedCPUPercentage := calculateCpuLimits(0.25)
 
+	// account for docker stats / CloudWatch noise
+	statsNoiseDelta := 5.0
+
 	// Try to use a new cluster for this test, ensure no other task metrics for this cluster
 	newClusterName := "ecstest-telemetry-" + uuid.New()
 	_, err := ECS.CreateCluster(&ecsapi.CreateClusterInput{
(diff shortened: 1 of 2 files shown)


agent/functional_tests/tests/functionaltests_test.go
```diff
@@ -621,13 +626,13 @@ func telemetryTestWithStatsPolling(t *testing.T, taskDefinition string) {
...
-	metrics, err := VerifyMetrics(cwclient, params, false)
+	metrics, err := VerifyMetrics(cwclient, params, false, 0.0)
 	assert.NoError(t, err, "Task is running, verify metrics for CPU utilization failed")
 	// Also verify the cpu usage is around expectedCPUPercentage +/- 5%
-	assert.InDelta(t, expectedCPUPercentage*100.0, *metrics.Average, 5)
+	assert.InDelta(t, expectedCPUPercentage*100.0, *metrics.Average, statsNoiseDelta)
 
 	params.MetricName = aws.String("MemoryUtilization")
-	metrics, err = VerifyMetrics(cwclient, params, false)
+	metrics, err = VerifyMetrics(cwclient, params, false, 0.0)
```
petderek: why is this one zero instead of statsNoiseDelta?

fierlion: golang won't let me overload a function/use variadic args.

The existing `VerifyMetrics` signature uses the `idleCluster` bool:
```
func VerifyMetrics(cwclient *cloudwatch.CloudWatch, params *cloudwatch.GetMetricStatisticsInput, idleCluster bool)
```
So now if `idleCluster = true`, we need a corresponding noiseDelta float.  There are no checks for `idleCluster = false` so the 0.0 is effectively a nil placeholder wherever we have `false`.  If there's a better way of doing this, I'm open to it.

petderek: Can we use statsNoiseDelta as a constant within the function? I don't think we gain any flexibility by passing it around.

fierlion: updated.

tommyhahn:
This PR is created against the master branch, not dev...two secrets tests failed...

fierlion:
> This PR is created against the master branch, not dev...two secrets tests failed...

Updated to point to dev.

fierlion:
After re-running tests, checks are all green.  See commit status (https://github.com/fierlion/amazon-ecs-agent/branches), not status above. 
A couple of flaky tests failed.  None related to telemetry.

PR #1957, opened 2019-03-22
Export SHAs docker hub image

shubham2892:
### Summary
publish-docker-images now export SHAs of the uploaded image that can be used by other scripts.

### Testing
The script is exporting the right SHAs.

Here is a sample run of the script(changed to use my personal docker hub account), I have added "echo" after export command to see the values of variables being exported:

```bash

88e9fe5231bf:amazon-ecs-agent shugy$ ./scripts/publish-docker-images -s codebuild-agent-source -d false

======================================================
Pushing linux amd images to docker hub
======================================================

Copying s3://codebuild-agent-source/ecs-agent-30c0caea.tar to /var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.qmVTdKvB
download: s3://codebuild-agent-source/ecs-agent-30c0caea.tar to ../../../../../../../var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.qmVTdKvB
Copying s3://codebuild-agent-source/ecs-agent-30c0caea.tar.md5 to /var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.otBbFnGn
download: s3://codebuild-agent-source/ecs-agent-30c0caea.tar.md5 to ../../../../../../../var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.otBbFnGn
Checking ecs-agent-30c0caea.tar md5 sum against ecs-agent-30c0caea.tar.md5
md5sum Check Successful
Loaded image: amazon/amazon-ecs-agent:latest
RUNNING: docker push shugy/amazon-ecs-agent:amd64-v1.26.0
RUNNING: docker push shugy/amazon-ecs-agent:amd64-30c0caea
RUNNING: docker push shugy/amazon-ecs-agent:amd64-latest
sha256:a10696b0fa9340ae408c6d5f5b5130e9afc5fe0916f8b65e56dfd92916edbf66 <==AMD_IMAGE_SHA

=======================================================
Pushing linux arm images to docker hub
=======================================================

Copying s3://codebuild-agent-source/ecs-agent-arm64-30c0caea.tar to /var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.QDSAKROr
download: s3://codebuild-agent-source/ecs-agent-arm64-30c0caea.tar to ../../../../../../../var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.QDSAKROr
Copying s3://codebuild-agent-source/ecs-agent-arm64-30c0caea.tar.md5 to /var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.zJxcECcM
download: s3://codebuild-agent-source/ecs-agent-arm64-30c0caea.tar.md5 to ../../../../../../../var/folders/8f/f7l_y7l96z53x4sm1zm2qm3s1srydb/T/tmp.zJxcECcM
Checking ecs-agent-arm64-30c0caea.tar md5 sum against ecs-agent-arm64-30c0caea.tar.md5
md5sum Check Successful
Loaded image: amazon/amazon-ecs-agent:latest
RUNNING: docker push shugy/amazon-ecs-agent:arm64-v1.26.0
RUNNING: docker push shugy/amazon-ecs-agent:arm64-30c0caea
RUNNING: docker push shugy/amazon-ecs-agent:arm64-latest
sha256:a10696b0fa9340ae408c6d5f5b5130e9afc5fe0916f8b65e56dfd92916edbf66 <==ARM_IMAGE_SHA


==========================================================
Pushing multi-arch manifest list to docker hub
==========================================================

Cloning into 'manifest-tool'...
remote: Enumerating objects: 4603, done.
remote: Total 4603 (delta 0), reused 0 (delta 0), pack-reused 4603
Receiving objects: 100% (4603/4603), 3.70 MiB | 9.51 MiB/s, done.
Resolving deltas: 100% (1919/1919), done.
Note: checking out '6bcfccea0827b4ff930f2798440cfce6cd4f1318'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 6bcfcce Merge pull request #62 from estesp/prepare-v0.9.0
go build -ldflags "-X main.gitCommit="6bcfccea0827b4ff930f2798440cfce6cd4f1318"" -o manifest-tool github.com/estesp/manifest-tool
RUNNING: ./manifest-tool push from-spec multi-arch-temp.yaml
RUNNING: ./manifest-tool push from-spec multi-arch-temp.yaml
RUNNING: ./manifest-tool push from-spec multi-arch-temp.yaml
sha256:ca6b2f6e04f25e8c614a455cb72b544c5dff5a5cb9653a01900028dee6491388 <==MULTIARCH_IMAGE_SHA

```

New tests cover the changes: 

### Licensing

### DIFF
--- a/scripts/publish-docker-images
+++ b/scripts/publish-docker-images
@@ -35,11 +35,11 @@ usage() {
 	echo
 	echo "Options"
 	echo "  -d  true|false  Dryrun (default is true)"
-	echo "  -p  PROFILE	 AWS CLI Profile (default is none)"
-	echo "  -s  BUCKET	  AWS S3 Bucket for staging"
-	echo "  -i  IMAGE	   Docker image name"
-	echo "	-f  FILE PATH	Path to file where manifest JSON output is stored"
-	echo "  -h			  Display this help message"
+	echo "  -p  PROFILE	 	AWS CLI Profile (default is none)"
+	echo "  -s  BUCKET	  	AWS S3 Bucket for staging"
+	echo "  -i  IMAGE	   	Docker image name"
+	echo "	-f  FILE_PATH	Path to file where manifest JSON output is stored"
+	echo "  -h			  	Display this help message"
 }
 
 verify_md5() {
@@ -63,17 +63,17 @@ push_linux_amd_agent_docker_hub() {
 	s3_cp "s3://${STAGE_S3_BUCKET}/ecs-agent-${IMAGE_TAG_SHA}.tar.md5" "${tarball_md5}"
 	echo "Checking ecs-agent-${IMAGE_TAG_SHA}.tar md5 sum against ecs-agent-${IMAGE_TAG_SHA}.tar.md5"
 
-	verify_md5 ${tarball} ${tarball_md5}
+	verify_md5 "${tarball}" "${tarball_md5}"
 	echo "md5sum Check Successful"
 
 	# Pushing to docker hub
 	docker load < "${tarball}"
-	for image_tag in ${IMAGE_TAG_VERSION_AMD} ${IMAGE_TAG_SHA_AMD} ${IMAGE_TAG_LATEST_AMD}; do
-		tag_and_push_docker ${IMAGE_NAME} ${image_tag}
+	for image_tag in "${IMAGE_TAG_VERSION_AMD}" "${IMAGE_TAG_SHA_AMD}" "${IMAGE_TAG_LATEST_AMD}"; do
+		tag_and_push_docker "${IMAGE_NAME}" "${image_tag}"
 	done
 
-	rm ${tarball}
-	rm ${tarball_md5}
+	rm "${tarball}"
+	rm "${tarball_md5}"
 }
 
 push_linux_arm_agent_docker_hub() {
@@ -89,17 +89,17 @@ push_linux_arm_agent_docker_hub() {
 	s3_cp "s3://${STAGE_S3_BUCKET}/ecs-agent-arm64-${IMAGE_TAG_SHA}.tar.md5" "${tarball_md5_arm}"
 	echo "Checking ecs-agent-arm64-${IMAGE_TAG_SHA}.tar md5 sum against ecs-agent-arm64-${IMAGE_TAG_SHA}.tar.md5"
 
-	verify_md5 ${tarball_arm} ${tarball_md5_arm}
+	verify_md5 "${tarball_arm}" "${tarball_md5_arm}"
 	echo "md5sum Check Successful"
 
 	# Pushing to docker hub
 	docker load < "${tarball_arm}"
-	for image_tag in ${IMAGE_TAG_VERSION_ARM} ${IMAGE_TAG_SHA_ARM} ${IMAGE_TAG_LATEST_ARM}; do
-		tag_and_push_docker ${IMAGE_NAME} ${image_tag}
+	for image_tag in "${IMAGE_TAG_VERSION_ARM}" "${IMAGE_TAG_SHA_ARM}" "${IMAGE_TAG_LATEST_ARM}"; do
+		tag_and_push_docker "${IMAGE_NAME}" "${image_tag}"
 	done
 
-	rm ${tarball_arm}
-	rm ${tarball_md5_arm}
+	rm "${tarball_arm}"
+	rm "${tarball_md5_arm}"
 }
 
 push_multi-arch_manifest_docker_hub() {
@@ -111,17 +111,15 @@ push_multi-arch_manifest_docker_hub() {
 
 	manifest_tool_dir="manifest-tool"
 	# Remove manifest-tool folder if it already exists
-	rm -rf ${manifest_tool_dir}
+	rm -rf "${manifest_tool_dir}"
 
-	manifest_tool_commit_id="6bcfccea0827b4ff930f2798440cfce6cd4f1318" # v0.9.0
 	# Clone the manifest-tool from github
-	git clone https://github.com/estesp/manifest-tool ${manifest_tool_dir}
-	cd ${manifest_tool_dir}
-	git checkout ${manifest_tool_commit_id}
+	git clone --depth 1 --branch v0.9.0 https://github.com/estesp/manifest-tool "${manifest_tool_dir}"
+
+	pushd "${manifest_tool_dir}"
 
 	make build
-
-	for image_tag in ${IMAGE_TAG_VERSION} ${IMAGE_TAG_SHA} ${IMAGE_TAG_LATEST}; do
+	for image_tag in "${IMAGE_TAG_VERSION}" "${IMAGE_TAG_SHA}" "${IMAGE_TAG_LATEST}"; do
 		# Replace the tags in multi-arch.yaml
 		sed -e "s/\${docker-image-tag}/${image_tag}/"  ../scripts/multi-arch.yaml > multi-arch-temp.yaml
 		# Generate the manifest list and push to docker hub
@@ -129,27 +127,31 @@ push_multi-arch_manifest_docker_hub() {
 		rm multi-arch-temp.yaml
 	done
 
+
 	# Create digest files
-	if [[ ! -z "${FILE_PATH}" ]]; then
-		if [[ -d "${FILE_PATH}" ]]; then
+	if [[ -n "${FILE_PATH}" ]]; then
 			echo "Creating digest files"
+			# Generate a combined JSON object for each image name with its
+			# manifest-tool output of the form:
+			#
+			# {
+			#    "amazon/amazon-ecs-agent:sha": [ ...inspected-manifest ],
+			# 	 "amazon/amazon-ecs-agent:arm64-sha": [ ...inspected-manifest ],
+			# 	 "amazon/amazon-ecs-agent:amd64-sha": [ ...inspected-manifest ]
+			# }
+			# The JSON is stored in a file passed in -f parameter
 
-			./manifest-tool inspect --raw "${IMAGE_NAME}:${IMAGE_TAG_LATEST}" \
-				> "${FILE_PATH}/${IMAGE_TAG_SHA}-digest.json"
-
-			./manifest-tool inspect --raw "${IMAGE_NAME}:${IMAGE_TAG_LATEST_ARM}" \
-				> "${FILE_PATH}/${IMAGE_TAG_SHA_ARM}-digest.json"
-
-			./manifest-tool inspect --raw "${IMAGE_NAME}:${IMAGE_TAG_LATEST_AMD}" \
-				> "${FILE_PATH}/${IMAGE_TAG_SHA_AMD}-digest.json"
+			{
+				for v in "${IMAGE_TAG_SHA}" "${IMAGE_TAG_SHA_ARM}" "${IMAGE_TAG_SHA_AMD}"; do
+					./manifest-tool inspect --raw "${IMAGE_NAME}:${v}" | jq "{\"${IMAGE_NAME}:${v}\":.}"
+				done
+			} | jq -s add > "${FILE_PATH}"
 
 			echo "Digest file created"
-		else
-			echo "File path is not valid: ${FILE_PATH}"
-		fi
 	fi
 
-	rm -rf ${manifest_tool_dir}
+	popd
+	rm -rf "${manifest_tool_dir}"
 
 }
 
--- a/scripts/publish-docker-images
+++ b/scripts/publish-docker-images
@@ -21,8 +21,8 @@ IMAGE_NAME="amazon/amazon-ecs-agent"
 
 AWS_PROFILE=""
 STAGE_S3_BUCKET=""
-
-source $(dirname "${0}")/publishing-common.sh
+S3_ACL_OVERRIDE=""
+source scripts/publishing-common.sh
 
 usage() {
 	echo "Usage: ${0} -s BUCKET [OPTIONS]"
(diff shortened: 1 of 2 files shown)


scripts/publish-docker-images
```diff
@@ -123,9 +127,10 @@ push_multi-arch_manifest_docker_hub() {
 		# Replace the tags in multi-arch.yaml
 		sed -e "s/\${docker-image-tag}/${image_tag}/"  ../scripts/multi-arch.yaml > multi-arch-temp.yaml
 		# Generate the manifest list and push to docker hub
-		dryval ./manifest-tool push from-spec multi-arch-temp.yaml
+		MULTIARCH_IMAGE_SHA=$(dryval ./manifest-tool push from-spec multi-arch-temp.yaml)
```
yumex93: The return value is just a sha?
Based on this: https://github.com/estesp/manifest-tool/blob/6bcfccea0827b4ff930f2798440cfce6cd4f1318/push.go. Seems it has sha and length as output?

shubham2892: You are right, fixed it.

shubham2892:
> Can you share a sample test run output?

Added to description. 😄

scripts/publish-docker-images
```diff
@@ -21,8 +21,8 @@ IMAGE_NAME="amazon/amazon-ecs-agent"
 
 AWS_PROFILE=""
 STAGE_S3_BUCKET=""
-
-source $(dirname "${0}")/publishing-common.sh
+S3_ACL_OVERRIDE=""
```
yumex93: What is this used for?

scripts/publish-docker-images
```diff
@@ -21,8 +21,8 @@ IMAGE_NAME="amazon/amazon-ecs-agent"
...
+source scripts/publishing-common.sh
```
jahkeup: Depending on why this handling is being updated, you may want to consider an alternative to the prior path expansion. In some situations (and some versions of bash) such as sourcing a script: the existing `$(dirname "$0")` will expand to the sourcing script's `$0` and return the wrong path for sourcing `publishing-common.sh`.

Currently, this makes the assumption about where the script is being called or sourced from and should not be merged as is.

jahkeup: Here is a fairly well accepted and robust means to resolve the path for use when you can guarantee a script will be using `bash`:

```
DIR="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
```

---

For example of why this is needed:

This is what happens when you source a script:

```
$ source /tmp/mydir.sh 
# #!/usr/bin/env bash
# from_bash_source="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
# from_dollar_0="$(cd "$(dirname "$0")"; pwd -P )"
# echo using bash_source=$from_bash_source
# echo using \$0=$from_dollar_0
I'm working out of: /tmp/myworkdir
using bash_source=/tmp
using $0=/tmp/myworkdir
```

And this is what happens when you run a script directly:

```
$ /tmp/mydir.sh 
# #!/usr/bin/env bash
# from_bash_source="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
# from_dollar_0="$(cd "$(dirname "$0")"; pwd -P )"
# echo using bash_source=$from_bash_source
# echo using \$0=$from_dollar_0
I'm working out of: /tmp/myworkdir
using bash_source=/tmp
using $0=/tmp
```

scripts/publish-docker-images
shubham2892: This variable is used in `publishing-common.sh->s3_cp`. We need `acl` to be `public_read` in that function.

scripts/publish-docker-images
shubham2892: Got it. Thanks a lot. 👍

scripts/publish-docker-images
```diff
@@ -21,8 +21,9 @@ IMAGE_NAME="amazon/amazon-ecs-agent"
 
 AWS_PROFILE=""
 STAGE_S3_BUCKET=""
-
-source $(dirname "${0}")/publishing-common.sh
+S3_ACL_OVERRIDE=""
+DIR="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
+source ${DIR}/publishing-common.sh
```
jahkeup: This should be quoted to protect against paths that would otherwise be misinterpreted or split here.

```suggestion
source "${DIR}/publishing-common.sh"
```

scripts/publish-docker-images
```diff
@@ -68,9 +69,11 @@ push_linux_amd_agent_docker_hub() {
 	# Pushing to docker hub
 	docker load < "${tarball}"
 	for image_tag in ${IMAGE_TAG_VERSION_AMD} ${IMAGE_TAG_SHA_AMD} ${IMAGE_TAG_LATEST_AMD}; do
-		tag_and_push_docker ${IMAGE_NAME} ${image_tag}
+		AMD_IMAGE_SHA=$(tag_and_push_docker ${IMAGE_NAME} ${image_tag} | tail -1 | cut -d" " -f3)
 	done
 
+	export AMD_IMAGE_SHA
```
jahkeup: I'm still not a huge fan of extending this script for `source`-ing. Exporting this variable here isn't entirely necessary can be be declared in the script's top level scope - I'd prefer it be there if this going to be part of the "API".

Also, if you haven't done so already, please take a read through the script knowing that the functions and variables provided here are going to be `source`'d into the sourcing script's scope.

Things to keep in mind when sourcing:

1. Are there conflicts to behavior here that could be inherited from the caller?
 
   * Are these things protected against here, should they be? (yes, they should)

2. Are there variables that would further influence the caller? 

   * Are the callers equipped to or protected against changes here that they may or may not be aware of?

3. Are you clobbering data or functions the caller already has?

   * Default settings in this scope may be applied in such a way that overwrites what the caller has.

scripts/publishing-common.sh
```diff
@@ -13,7 +13,8 @@
 # License for the specific language governing permissions and
 # limitations under the License.
 
-export VERSION=$(cat $(dirname "${0}")/../VERSION)
+DIR="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
+export VERSION=$(cat ${DIR}/../VERSION)
```
jahkeup: Please "quote" the path given to cat.
```suggestion
export VERSION=$(cat "${DIR}/../VERSION")
```

scripts/publish-docker-images
```diff
@@ -38,6 +38,7 @@ usage() {
 	echo "  -p  PROFILE	 AWS CLI Profile (default is none)"
 	echo "  -s  BUCKET	  AWS S3 Bucket for staging"
 	echo "  -i  IMAGE	   Docker image name"
+	echo "	-f  FILE PATH	Path to file where manifest JSON output is stored"
```
jahkeup: nit: conventionally the flag's option is a single token like `FILE_PATH` when the expectation that its a single arg.

scripts/publish-docker-images
```diff
@@ -127,12 +129,31 @@ push_multi-arch_manifest_docker_hub() {
 		rm multi-arch-temp.yaml
 	done
 
+	# Create digest files
+	if [[ ! -z "${FILE_PATH}" ]]; then
+		if [[ -d "${FILE_PATH}" ]]; then
```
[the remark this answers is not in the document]
shubham2892: > we'd be able to have them in a single file and extract as needed. 

We have them in single file. The output of `manifest-tool inspect --raw "${IMAGE_NAME}:${IMAGE_TAG_LATEST}"` gives us all the required digest. Output JSON is in the form -> [{},{},{}] . 

The first entry is always corresponds to the multi arch image. For the second and third entry in JSON we have to check the _architecture_ flag to check which entry is which(ARM or AMD).
By having three separate files we can avoid doing that check and always pick up the digest of the first entry in the JSON 

> let's rename this variable along with its flag usage

Agree with this. Will do.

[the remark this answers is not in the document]
jahkeup: The code currently writes to *multiple files* in the path that user provides where each has the image's inspected manifest. The merged json with key structure I proposed is instead results in a *single file* that the user specified - which, again, is what I had in mind when we had discussed ways to pass this data along.  

So that the user doesn't have a whole directory to enumerate and transfer if needed, they could instead have a single file with the structured data in the json, this doesn't change that the first element in each value is the "magic" manifest element. 

The "magic" `manifest.list` entry is otherwise lacking any other identifier than its "MediaType"; the keying I suggest doesn't change this magical state of affairs (you'd use something like `.["amazon/amazon-ecs-agent:latest"][0].Digest` to get to that "magic" element). This also eliminates the metadata maintained in the filename for this use case.

scripts/publish-docker-images
```diff
@@ -127,12 +128,22 @@ push_multi-arch_manifest_docker_hub() {
 		rm multi-arch-temp.yaml
 	done
```
jahkeup: nit: I spotted a few variables for paths used this function that are without `"$quotes"`, if you have cycles while updating for review feedback can you fix this in this function's scope?

scripts/publish-docker-images
```diff
@@ -127,12 +128,22 @@ push_multi-arch_manifest_docker_hub() {
...
+			echo "Creating digest files"
+			{
+				for v in ${IMAGE_TAG_SHA} ${IMAGE_TAG_SHA_ARM} ${IMAGE_TAG_SHA_AMD}; do
+					./manifest-tool inspect --raw "${IMAGE_NAME}:${v}" | jq "{\"${IMAGE_NAME}:${v}\":.}";
+				done;
+			} | jq -s add > "${FILE_PATH}"
+			echo "Digest file created"
+	fi
+
 	rm -rf ${manifest_tool_dir}
```
jahkeup: Wait a sec, I don't see a `cd` or a `popd` prior to this `rm`, so the dir removed is not actually where the `manifest-tool` is cloned to. 

Are we running into issues with this script as used today? I have a feeling we're not as we tend to use `mktemp` liberally. The existing management of the directory could become an issue if the caller's usage were to change.

This could be fixed by either using a `mktemp -d` where `manifest-tool` is cloned into, built, and used from (necessitating something like `make -C "$manifest_tool_dir"`, `"manifest_tool="$manifest_tool_dir/manifest-tool"`, `"$manifest_tool" inspect --raw ...`). As an added bonus, the `git clone` could be simplified to: `git clone --depth 1 --branch v0.9.0 https://github.com/estesp/manifest-tool "$manifest_tool_dir"` which avoids unnecessary network traffic, additional commands to checkout the rev, and leaves the repo ready to rumble!

If we don't want to go that route, the `manifest-tool` directory /could/ remain in its current place and simply use `pushd "$manifest_tool_dir"; $do_work; popd`, but I tend to prefer temp files :smile: 

Either way, take another read over this function to see if you see other broken assumptions or issues with using a temp dir - nicely enough, `manifest-tool` is fairly well scoped to this function so its pretty quick work.

shubham2892: Added a `popd` before `rm -rf "${manifest_tool_dir}"`.

scripts/publish-docker-images
```diff
@@ -131,22 +130,13 @@ push_multi-arch_manifest_docker_hub() {
 
 	# Create digest files
 	if [[ ! -z "${FILE_PATH}" ]]; then
```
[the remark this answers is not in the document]
jahkeup: This still look unresolved, are there more commits coming?

scripts/publish-docker-images
jahkeup: The variables here are still unquoted, are there pending changes?

scripts/publish-docker-images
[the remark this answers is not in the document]
shubham2892: Yup

scripts/publishing-common.sh
```diff
@@ -13,18 +13,20 @@
 # License for the specific language governing permissions and
 # limitations under the License.
 
-export VERSION=$(cat $(dirname "${0}")/../VERSION)
+DIR="$( cd "$(dirname "${BASH_SOURCE[0]}")" ; pwd -P )"
```
jahkeup: This path resolution method should be used at [the beginning of the script as it sources the helpers](https://github.com/aws/amazon-ecs-agent/pull/1957/files#diff-4ccd150869ada089e3677e1655559c73R25), though the use of this script by `source`-ing is a bit less tempting with these latest changes so its not as critical to block on. As with our earlier conversation about sourcing and considerations - keep in mind this could be clobbering a consumer's variable.

If its not going to get fixed now, let's jot down some notes (can you make an issue?) on what to change so that it gets fixed up at some point.

scripts/publish-docker-images
```diff
@@ -150,6 +173,9 @@ while getopts ":d:p:s:i" opt; do
 		i)
 			IMAGE_NAME="${OPTARG}"
 			;;
+		f)
+			FILE_PATH="${OPTARG}"
```
jahkeup: One of the things we kinda dropped as we moved away from the directory approach was the validation of the provided path. It's /not critical and shouldn't block this merge/, but its something that should generally be applied (to all user input!). Maybe this could be captured for a next iteration improvement?

---

Any and all checks should be performed prior to running so as to error early and not bail mid-way through the script's run (because it could interact poorly with other release processes):

* can we write to the path provided?
* can/should we continue execution if we can't write this file?
* do all the directories in the path exist? do we create them if missing?
* does this file already have content? do we overwrite it?

Again, its not critical to fix these now and I think we are circumventing the need for these checks where its used (yay `mktemp`!).

shubham2892: Will create an issue to track the next iteration improvement from this comment and the comment above.

shubham2892:
Flakey Windows functional test failing as tracked in #1991.

PR #1961, opened 2019-03-25
Add func test for trunk eni attach/detach workflow

fenxiong:
### Summary

Add func test for trunk eni attach/detach workflow.

### Implementation details

The test verifies that when ENI trunking is enabled, the Trunk ENI is attached when container instance reaches ACTIVE status, and it's detached when the container instance is deregistered.

### Testing

Test passes with gamma. cannot pass with prod yet.
Update: test passes with prod now.

```
=== RUN   TestTrunkENIAttachDetachWorkflow
--- PASS: TestTrunkENIAttachDetachWorkflow (31.70s)
	utils_unix.go:130: Created directory /tmp/ecs_integ_testdata042066345 to store test data in
	utils_unix.go:146: Launching agent with image: amazon/amazon-ecs-agent:latest
	utils_unix.go:237: Agent started as docker container: 46652bd4cb18e2f6293f71c966cf1321068faec9a5b88c17b730226af5120d1c
	utils.go:177: Found agent metadata: {Cluster:ecs-functional-tests ContainerInstanceArn:0xc420298410 Version:Amazon ECS Agent - v1.26.1 (d030fc86)}
	utils.go:202: Removing test dir for passed test /tmp/ecs_integ_testdata042066345
PASS
ok  	github.com/aws/amazon-ecs-agent/agent/functional_tests/tests	31.799s
```

### Licensing

### DIFF
--- a/agent/ec2/ec2_metadata_client.go
+++ b/agent/ec2/ec2_metadata_client.go
@@ -30,6 +30,7 @@ const (
 	InstanceIdentityDocumentResource          = "instance-identity/document"
 	InstanceIdentityDocumentSignatureResource = "instance-identity/signature"
 	MacResource                               = "mac"
+	AllMacResource                            = "network/interfaces/macs"
 	VPCIDResourceFormat                       = "network/interfaces/macs/%s/vpc-id"
 	SubnetIDResourceFormat                    = "network/interfaces/macs/%s/subnet-id"
 	InstanceIDResource                        = "instance-id"
@@ -68,6 +69,7 @@ type EC2MetadataClient interface {
 	VPCID(mac string) (string, error)
 	SubnetID(mac string) (string, error)
 	PrimaryENIMAC() (string, error)
+	AllENIMacs() (string, error)
 	InstanceID() (string, error)
 	GetUserData() (string, error)
 	Region() (string, error)
@@ -136,6 +138,11 @@ func (c *ec2MetadataClientImpl) GetMetadata(path string) (string, error) {
 // network interface of the instance
 func (c *ec2MetadataClientImpl) PrimaryENIMAC() (string, error) {
 	return c.client.GetMetadata(MacResource)
+}
+
+// AllENIMacs returns the mac addresses for all the network interfaces attached to the instance
+func (c *ec2MetadataClientImpl) AllENIMacs() (string, error) {
+	return c.client.GetMetadata(AllMacResource)
 }
 
 // VPCID returns the VPC id for the network interface, given
--- a/agent/functional_tests/tests/functionaltests_unix_test.go
+++ b/agent/functional_tests/tests/functionaltests_unix_test.go
@@ -799,8 +799,7 @@ func TestRunAWSVPCTaskWithENITrunkingEndPointValidation(t *testing.T) {
 
 	defer agent.Cleanup()
 
-	// TODO: change back to 1.27.1 when merging the pr
-	agent.RequireVersion(">=1.26.0")
+	agent.RequireVersion(">=1.27.1")
 
 	roleArn := os.Getenv("TASK_IAM_ROLE_ARN")
 	if utils.ZeroOrNil(roleArn) {
@@ -1717,8 +1716,8 @@ func TestTrunkENIAttachDetachWorkflow(t *testing.T) {
 		err = WaitNetworkInterfaceCount(existingInterfaceCount, asyncWaitDuration)
 		assert.NoError(t, err)
 	}()
-	// TODO: when merging the changes, change to 1.27.1
-	agent.RequireVersion(">=1.26.0")
+
+	agent.RequireVersion(">=1.27.1")
 
 	// Expect one more interface to be attached (i.e. the Trunk)
 	macs, err := GetNetworkInterfaceMacs()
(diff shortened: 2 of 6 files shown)


agent/functional_tests/util/utils.go
```diff
@@ -719,6 +730,20 @@ func RequireRegions(t *testing.T, supportedRegions []string, region string) {
 	}
 }
 
+// RequireInstanceTypes skips the test if current instance type is not a supported instance type
+func RequireInstanceTypes(t *testing.T, supportedTypePrefixes []string) {
+	iid, _ := ec2.NewEC2MetadataClient(nil).InstanceIdentityDocument()
```
yunhee-l: are we ok with adding this dependency? we were heading towards removing these, weren't we?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1607,3 +1607,58 @@ func TestAppMeshCNIPlugin(t *testing.T) {
...
+	agent.RequireVersion(">=1.26.0")
+
+	// Wait for container instance to become active
+	err = agent.WaitContainerInstanceActive(asyncWaitDuration)
+	require.NoError(t, err)
+
+	// Expect one more interface to be attached (i.e. the Trunk)
+	interfaceCount, err := GetNetworkInterfaceCount()
+	require.NoError(t, err)
+	assert.Equal(t, existingInterfaceCount + 1, interfaceCount)
```
sharanyad: is there any other way to verify this rather than checking count +1 ? like expected trunk interface name, querying etc.? 
(+1 doesn't always guarantee trunk attached right?)

agent/functional_tests/util/utils_unix.go
```diff
@@ -38,6 +37,7 @@ import (
 	dockercontainer "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/network"
 	docker "github.com/docker/docker/client"
+	"github.com/pkg/errors"
```
sparrc: IDE mistake?

agent/functional_tests/util/utils.go
fenxiong: since right now we are running functional tests across instance types, and the test can only pass on some instance types, this is the only way i know to avoid test failing. we can revisit this when our test facility support us to remove iid i think?

agent/functional_tests/util/utils_unix.go
fenxiong: probably autogenerated when i changed something and removed it in this file.. will undo the change

agent/functional_tests/tests/functionaltests_unix_test.go
fenxiong: i guess i can check this with DescribeContainerInstances response.. will make the change

fenxiong: fixed

agent/functional_tests/util/utils_unix.go
fenxiong: fixed

agent/ec2/ec2_metadata_client.go
```diff
@@ -138,6 +140,11 @@ func (c *ec2MetadataClientImpl) PrimaryENIMAC() (string, error) {
 	return c.client.GetMetadata(MacResource)
 }
 
+// AllENIMacs returns the mac addresses for all the network interfaces attached to the instance
```
yunhee-l: just curious, why not a list of strings? Plural method name, returning a delimited single string struck me odd.

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1672,3 +1673,75 @@ func TestAppMeshCNIPlugin(t *testing.T) {
...
+			Force:             aws.Bool(true),
+		})
+		require.NoError(t, err)
+
+		// Wait and verify that the Trunk ENI is detached after deregistration
+		err = WaitNetworkInterfaceCount(existingInterfaceCount, asyncWaitDuration)
+		assert.NoError(t, err)
+	}()
+	// TODO: when merging the changes, change to 1.27.1
+	agent.RequireVersion(">=1.26.0")
```
fierlion: I see this same `agent.RequireVersion()` up on line 803.  could this be declared once up top as a const?

agent/ec2/ec2_metadata_client.go
fenxiong: this is a wrapper method that simply uses ec2's [GetMetadata](https://docs.aws.amazon.com/sdk-for-go/api/aws/ec2metadata/#EC2Metadata.GetMetadata) API and returns the result as it. the result is converted to a list of strings in the [GetNetworkInterfaceMacs](https://github.com/fenxiong/amazon-ecs-agent/blob/eni-trunking-func/agent/functional_tests/util/utils.go#L975-L980) method

agent/functional_tests/tests/functionaltests_unix_test.go
fenxiong: i think different functional tests may require different version of agents, so this probably shouldn't be declared globally. the agent version check enables us to run functional tests across different version of agents without changing the test code

PR #1964, opened 2019-03-28
Increase timeout for integ tests

ubhattacharjya:
### Summary

TestShutDownOrder needs some more time to complete as there are no errors when it times out. That is why increasing the timeout for integ tests.
### Testing

New tests cover the changes: 

### Licensing


Makefile
```diff
@@ -276,10 +276,10 @@ cni-plugins: get-cni-sources .out-stamp build-ecs-cni-plugins build-vpc-cni-plug
 
 ifeq (${BUILD_PLATFORM},aarch64)
 run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tests
-	. ./scripts/shared_env && go test -tags integration -timeout=20m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
+	. ./scripts/shared_env && go test -tags integration -timeout=30m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
```
sharanyad: this seems like a big jump. why are we doing this?

ubhattacharjya: reduced the timeout from 30 to 25

PR #1971, opened 2019-04-01
Functional Tests: pull fewer images

petderek:
### Summary
This removes the need to pull python, consul, debian, httpd, crux, and redis containers as part of the functional test run.

### Implementation details
Many of the containers were only used for the parallel pull test. However, this was kind of slow and caused occasional failures (particularly in Arm testing).   Reducing the number of pulls should make testing more consistent. 

One way of doing this is to generate the containers on disk instead. For parallel pull, the content isn't important: it only matters that we are pulling lots of containers at once. 

For python, we only required it for a one-liner to allocate memory. The telemetry container already does this, so I'm reusing that instead of pulling python.

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/oom-container/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/oom-container/task-definition.json
@@ -5,8 +5,8 @@
     "memory": 64,
     "name": "error",
     "cpu": 1,
-    "image": "127.0.0.1:51670/python:2",
-    "command": ["python", "-c", "foo=' '*1024*1024*512; import time; time.sleep(10)"]
+    "image": "amazon/amazon-ecs-stress:make",
+    "command": ["-concurrency", "10", "-memory", "100"]
   }]
 }
 
--- a/agent/functional_tests/testdata/taskdefinitions/oom-task/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/oom-task/task-definition.json
@@ -5,8 +5,8 @@
     "essential": true,
     "name": "error",
     "cpu": 1,
-    "image": "127.0.0.1:51670/python:2",
-    "command": ["python", "-c", "foo=' '*1024*1024*512; import time; time.sleep(10)"]
+    "image": "amazon/amazon-ecs-stress:make",
+    "command": ["-concurrency", "10", "-memory", "100"]
   }]
 }
 
--- a/agent/functional_tests/testdata/taskdefinitions/parallel-pull/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/parallel-pull/task-definition.json
@@ -5,7 +5,6 @@
         "name": "busybox",
         "image": "127.0.0.1:51670/busybox:parallel-pull-fts",
         "cpu": 10,
-        "command":  [ "sh", "-c", "while [ true ]; do sleep 1s; date +%T; done"],
         "memory": 64,
         "essential": true
     },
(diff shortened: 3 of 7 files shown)


agent/functional_tests/testdata/taskdefinitions/oom-task/task-definition.json
```diff
@@ -5,8 +5,8 @@
     "essential": true,
     "name": "error",
     "cpu": 1,
-    "image": "127.0.0.1:51670/python:2",
-    "command": ["python", "-c", "foo=' '*1024*1024*512; import time; time.sleep(10)"]
+    "image": "amazon/amazon-ecs-stress:make",
+    "command": ["-concurrency", "10", "-memory", "100"]
```
adnxn: "neat"

scripts/setup-test-registry
```diff
@@ -45,52 +44,47 @@ mirror_local_image() {
   docker rmi $2
 }
 
+setup_parallel_pull_image() {
+  local image_name="127.0.0.1:51670/$1:parallel-pull-fts"
+  echo "Generating parallel pull image $image_name"
+ 
+  # Get new random bits every time we build 
+  head -c 25m < /dev/urandom > docker-context/random-bits
```
sharanyad: could you also add comments on why we do this?

scripts/setup-test-registry
```diff
@@ -45,52 +44,47 @@ mirror_local_image() {
...
+mirror_image busybox:1.29.3 "127.0.0.1:51670/busybox:latest"
+mirror_image "$NGINX_IMAGE" "127.0.0.1:51670/nginx:latest"
+mirror_image ubuntu:16.04 "127.0.0.1:51670/ubuntu:latest"
+
+# create a context folder used by docker build. It will only have a file
+# full of random bits so that the parallel pull images are different.
+mkdir -p docker-context
+cat << EOF > docker-context/Dockerfile
+FROM amazon/amazon-ecs-pause:0.1.0 
+ADD random-bits /random-bits
```
yumex93: Why we are doing this?

adnxn: i think so we get distinct image layers aka distinct images for the pulls 🤔

agent/functional_tests/testdata/taskdefinitions/oom-container/task-definition.json
```diff
@@ -5,8 +5,8 @@
     "memory": 64,
     "name": "error",
     "cpu": 1,
-    "image": "127.0.0.1:51670/python:2",
-    "command": ["python", "-c", "foo=' '*1024*1024*512; import time; time.sleep(10)"]
+    "image": "amazon/amazon-ecs-stress:make",
+    "command": ["-concurrency", "10", "-memory", "100"]
```
yumex93: Have you run the test multiple times to see whether all tests passed? I think feng made some changes to increase the memory we allocated here in order to make the test less flaky. You may wanna to double check the new changes will not break it.

yumex93:
Can we run maybe 100 times **TestParallelPull** to see how it is going? This is a flaky test. I would like to see whether new changes help with it.

agent/functional_tests/testdata/taskdefinitions/parallel-pull/task-definition.json
```diff
@@ -5,7 +5,6 @@
         "name": "busybox",
         "image": "127.0.0.1:51670/busybox:parallel-pull-fts",
         "cpu": 10,
-        "command":  [ "sh", "-c", "while [ true ]; do sleep 1s; date +%T; done"],
```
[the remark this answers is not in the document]
petderek: Ubuntu / Debian / busybox  just 'exit' if they aren't given something to do.

scripts/setup-test-registry
petderek: Its so we get distinct image layers aka distinct images for the pulls. :)

PR #1972, opened 2019-04-02
engine: update container metadata when container status stays the same

yumex93:
Agent discards container metadata update for redundant container status changing, however, for some cases, even the status is not changed, the metadata may change. Using container health status as an example, there is situation that when container status just transits to RUNNING, health status is still UNKNOWN and after some time, it changes to HEALTHY.

### Testing

New tests cover the changes: 
yes

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -407,6 +407,11 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	if event.Status <= containerKnownStatus {
 		seelog.Infof("Managed task [%s]: redundant container state change. %s to %s, but already %s",
 			mtask.Arn, container.Name, event.Status.String(), containerKnownStatus.String())
+
+		// Only update container metadata when status stays RUNNING
+		if event.Status == containerKnownStatus && event.Status == apicontainerstatus.ContainerRunning {
+			updateContainerMetadata(&event.DockerContainerMetadata, container, mtask.Task)
+		}
 		return
 	}
 
@@ -420,11 +425,6 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 		if !proceedAnyway {
 			return
 		}
-	}
-
-	// Update the container health status
-	if container.HealthStatusShouldBeReported() {
-		container.SetHealthStatus(event.Health)
 	}
 
 	mtask.RecordExecutionStoppedAt(container)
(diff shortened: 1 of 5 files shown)


agent/engine/task_manager.go
```diff
@@ -422,11 +428,6 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 		}
 	}
 
-	// Update the container health status
-	if container.HealthStatusShouldBeReported() {
```
tommyhahn: Could you explain a bit why this block of code is deleted?

fenxiong: +1. i don't think this should be deleted?

agent/engine/task_manager.go
```diff
@@ -407,6 +407,12 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	if event.Status <= containerKnownStatus {
 		seelog.Infof("Managed task [%s]: redundant container state change. %s to %s, but already %s",
 			mtask.Arn, container.Name, event.Status.String(), containerKnownStatus.String())
+
+		// Only update container metadata when status keeps the same, ignore updating metadata if there
+		// is backward transition
+		if event.Status == containerKnownStatus {
```
sharanyad: should we explicitly check for RUNNING status if the use case covers only that?

agent/engine/task_manager.go
yumex93: **updateContainerMetadata** already has this part of logic.

agent/engine/task_manager.go
yumex93: I will update it to start from only covering RUNNING since so far I haven't found out other use cases.

agent/engine/task_manager.go
```diff
@@ -407,6 +407,11 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	if event.Status <= containerKnownStatus {
 		seelog.Infof("Managed task [%s]: redundant container state change. %s to %s, but already %s",
 			mtask.Arn, container.Name, event.Status.String(), containerKnownStatus.String())
+
+		// Only update container metadata when status stays RUNNING
+		if event.Status == containerKnownStatus && event.Status == apicontainerstatus.ContainerRunning {
```
tommyhahn: I wonder if we are updating the container metadata without the container state change, should we do this checking "if event.Status == containerKnownStatus" independently outside this "if" block and update the log with this change? Also if we only update container metadata when status stays RUNNING, maybe checking "event.Status == apicontainerstatus.ContainerRunning" before "event.Status == containerKnownStatus" makes more sense?

yumex93: I don't really wanna have repetitive code for logging, that's why the if is inside the block. And what's the reason behind checking "event.Status == apicontainerstatus.ContainerRunning" first? I think it does not really make a big difference.

PR #1973, opened 2019-04-02
Fix for TestOOMTask

ubhattacharjya:
### Summary

ARM and AL2 instances have cgroupPath in a different path than for AL1 instances. That is why TestOOMTask was failing on ARM instances but passing for the linux functional test suite.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/functional_tests/util/utils_unix.go
+++ b/agent/functional_tests/util/utils_unix.go
@@ -38,6 +38,7 @@ import (
 	dockercontainer "github.com/docker/docker/api/types/container"
 	"github.com/docker/docker/api/types/network"
 	docker "github.com/docker/docker/client"
+	"runtime"
 )
 
 const (
@@ -260,7 +261,14 @@ func (agent *TestAgent) StartAgent() error {
 // * EXECDRIVER_PATH: the path of metrics
 func (agent *TestAgent) getBindMounts() []string {
 	var binds []string
-	cgroupPath := utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPath)
+	var cgroupPath string
+
+	if runtime.GOARCH == "arm64" {
+		cgroupPath = utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPathAgentMount)
+	} else {
+		cgroupPath = utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPath)
+	}
+
 	cgroupBind := cgroupPath + ":" + defaultCgroupPathAgentMount
 	binds = append(binds, cgroupBind)
 


agent/functional_tests/util/utils_unix.go
```diff
@@ -260,7 +261,14 @@ func (agent *TestAgent) StartAgent() error {
 // * EXECDRIVER_PATH: the path of metrics
 func (agent *TestAgent) getBindMounts() []string {
 	var binds []string
-	cgroupPath := utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPath)
+	var cgroupPath string
+
+	if runtime.GOARCH == "arm64" {
+		cgroupPath := utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPathAgentMount)
+	} else {
+		cgroupPath := utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPath)
```
sharanyad: should we add a condition for al2 as well?

yumex93:
Could you link the error message here and the test result on AL2 so we are able to know this pr fix the test?

agent/functional_tests/util/utils_unix.go
```diff
@@ -260,7 +261,14 @@ func (agent *TestAgent) StartAgent() error {
 // * EXECDRIVER_PATH: the path of metrics
 func (agent *TestAgent) getBindMounts() []string {
 	var binds []string
-	cgroupPath := utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPath)
+	var cgroupPath string
+
+	if runtime.GOARCH == "arm64" {
+		cgroupPath = utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPathAgentMount)
+	} else {
+		cgroupPath = utils.DefaultIfBlank(os.Getenv("CGROUP_PATH"), defaultCgroupPath)
```
yumex93: What is the right cgroup path for AL2? Seems here only arm64 gets updated to a different path?

agent/functional_tests/util/utils_unix.go
ubhattacharjya: 2019-03-29T23:25:33Z [WARN] Disabling TaskCPUMemLimit because agent is unabled to setup '/ecs' cgroup: cgroup create: unable to create controller: cgroups: cgroup mountpoint does not exist

agent/functional_tests/util/utils_unix.go
ubhattacharjya: > Could you link the error message here and the test result on AL2 so we are able to know this pr fix the test?

2019-03-29T23:25:33Z [WARN] Disabling TaskCPUMemLimit because agent is unabled to setup '/ecs' cgroup: cgroup create: unable to create controller: cgroups: cgroup mountpoint does not exist

ubhattacharjya: > What is the right cgroup path for AL2? Seems here only arm64 gets updated to a different path?

/sys/fs/cgroup

Issue #1918, opened 2019-03-01
Flakey windows unit test - TestStatsInactivityTimeout


fenxiong:
### Summary

TestStatsInactivityTimeout failed in travis.

```
=== RUN   TestStatsInactivityTimeout
--- FAIL: TestStatsInactivityTimeout (0.30s)
    <autogenerated>:1: 
        	Error Trace:	docker_client_test.go:1086
[...]

PR #1977, opened 2019-04-04
Fix inactivity timeout unit tests

fenxiong:
### Summary

Fix inactivity timeout unit tests - https://github.com/aws/amazon-ecs-agent/issues/1918 https://github.com/aws/amazon-ecs-agent/issues/1864

### Implementation details

In TestPullImageInactivityTimeout and TestStatsInactivityTimeout, on rare case, the inactivity handling goroutine has some delay and isn't able to trigger inactivity timeout before the reader finishes reading. This only happens in travis windows unit test, and when i reproduce (by triggering travis in my own fork), TestPullImageInactivityTimeout fails about 1 in 1000 times and TestStatsInactivityTimeout fails once in 3000 times. After reducing the timeout to 1 the tests don't fail in 5000 times.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client_test.go
+++ b/agent/dockerclient/dockerapi/docker_client_test.go
@@ -169,7 +169,7 @@ func TestPullImageInactivityTimeout(t *testing.T) {
 	mockDockerSDK, client, testTime, _, _, done := dockerClientSetup(t)
 	defer done()
 
-	client.config.ImagePullInactivityTimeout = 100 * time.Millisecond
+	client.config.ImagePullInactivityTimeout = 1 * time.Millisecond
 
 	testTime.EXPECT().After(gomock.Any()).AnyTimes()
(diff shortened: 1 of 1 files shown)


sharanyad:
> the inactivity handling goroutine has some delay

do we know whats causing this delay?

fenxiong:
> > the inactivity handling goroutine has some delay
> 
> do we know whats causing this delay?

i think the delay comes from either goroutine scheduling or the atomic load/save operations used by the inactivity handler. i wrote a [test](https://github.com/fenxiong/amazon-ecs-agent/blob/c388f4250f0d6019e19981cb47bde19b69d36ef6/agent/dockerclient/dockerapi/docker_client_test.go#L1090-L1131) that is a simplified version of TestStatsInactivityTimeout, which contains same goroutine generating logic and atomic operations as TestStatsInactivityTimeout but nothing else. When i tried running that test in travis, it has similar failure rate as TestStatsInactivityTimeout (about 1 in every 2k-3k times).

PR #1985, opened 2019-04-14
Fix broken make target

fenxiong:
### Summary

make release is failing because Go 1.12 [enforces using cache](https://golang.org/doc/go1.12#gocache) and by default some of the build scripts will try to use /.cache as the cache directory which is not writable (for user other than root), and the scripts fail with the following:
```
failed to initialize build cache at /.cache/go-build: mkdir /.cache: permission denied
```
See here for more details - https://github.com/docker-library/golang/issues/225#issuecomment-403170792. 
Fixing by specifying Golang's cache directory to be under /tmp.

### Testing

"make release" fails without the fix. passes with.

### Licensing

### DIFF
--- a/scripts/dockerfiles/Dockerfile.buildECSCNIPlugins
+++ b/scripts/dockerfiles/Dockerfile.buildECSCNIPlugins
@@ -14,6 +14,8 @@
 FROM golang:1.12
 MAINTAINER Amazon Web Services, Inc.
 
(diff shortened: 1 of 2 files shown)


fenxiong:
> I’m assuming this only happens in our dev environments and not in the CI?

yeah the CI tool isn't affected by this. tests are currently failing on arm for other reason though 😅  (unable to pull golang image) not sure if it's temporary

PR #1986, opened 2019-04-15
Make acs handler tests more predictable

shubham2892:
### Summary

Currently in the [select statement here](https://github.com/aws/amazon-ecs-agent/blob/aabe65ee15a013a6bc594da52b94d817419ee74a/agent/acs/handler/acs_handler.go#L335), both the conditions `<-acsSession.ctx.Done()` and `err := <-serveErr` will be true for the tests `TestHandlerReconnectsOnConnectErrors`, `TestHandlerReconnectsOnDiscoverPollEndpointError` towards the end of the test, because of how the mocks are structured in the test. Specifically:

```Go
	mockWsClient.EXPECT().Serve().Do(func() {
		// Serve() cancels the context
		cancel()
	}).Return(io.EOF).MinTimes(1)
```
`Cancel()` triggers the `<-acsSession.ctx.Done()` part and `.Return(io.EOF)` triggers the `err := <-serveErr`. 

This makes the behavior of the function unpredictable as either of the select condition could be selected and each it leads to different code path when it returns to `Start()` function.

By changing the mock to 

```Go
mockWsClient.EXPECT().Connect().Do(func() {
		// Serve() cancels the context
		cancel()
	}).Return(nil).MinTimes(1)
```
We have a more predictable path for the test where it will always trigger `<-acsSession.ctx.Done()` in the select case. 

Also adding some logs which will help in tracing out what code path test is taking. 

### Testing
Tested by adding `close(connectToACS)` [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/acs_handler.go#L231) and running `TestHandlerReconnectsOnConnectErrors`, `TestHandlerReconnectsOnDiscoverPollEndpointError` 100 times each. The tests passed everytime after this change, failed intermittently before. 

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -206,7 +206,7 @@ func (acsSession *session) Start() error {
 			if shouldReconnectWithoutBackoff(acsError) {
 				// If ACS closed the connection, there's no need to backoff,
 				// reconnect immediately
-				seelog.Info("ACS Websocket connection closed for a valid reason")
+				seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
 				acsSession.backoff.Reset()
 				sendEmptyMessageOnChannel(connectToACS)
 			} else {
@@ -219,6 +219,7 @@ func (acsSession *session) Start() error {
 					// If the context was not cancelled and we've waited for the
 					// wait duration without any errors, send the message to the channel
 					// to reconnect to ACS
+					seelog.Info("Done waiting; reconnecting to ACS")
 					sendEmptyMessageOnChannel(connectToACS)
 				} else {
 					// Wait was interrupted. We expect the session to close as canceling
@@ -308,6 +309,7 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 		seelog.Errorf("Error connecting to ACS: %v", err)
 		return err
 	}
+
 	seelog.Info("Connected to ACS endpoint")
 	// Start inactivity timer for closing the connection
 	timer := newDisconnectionTimer(client, acsSession.heartbeatTimeout(), acsSession.heartbeatJitter())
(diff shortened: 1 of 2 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -206,16 +206,16 @@ func (acsSession *session) Start() error {
...
 				// reconnect immediately
-				seelog.Info("ACS Websocket connection closed for a valid reason")
+				seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
 				acsSession.backoff.Reset()
 				sendEmptyMessageOnChannel(connectToACS)
 			} else {
 				// Disconnected unexpectedly from ACS, compute backoff duration to
 				// reconnect
 				reconnectDelay := acsSession.computeReconnectDelay(isInactiveInstance)
-				seelog.Infof("Reconnecting to ACS in: %s", reconnectDelay.String())
```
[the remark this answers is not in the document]
shubham2892: I will add another line.

agent/acs/handler/acs_handler_test.go
```diff
@@ -711,12 +712,12 @@ func TestHandlerReconnectsOnDiscoverPollEndpointError(t *testing.T) {
...
 	mockWsClient.EXPECT().AddRequestHandler(gomock.Any()).AnyTimes()
-	mockWsClient.EXPECT().Connect().Return(nil).AnyTimes()
+	mockWsClient.EXPECT().Serve().AnyTimes()
 	mockWsClient.EXPECT().Close().Return(nil).AnyTimes()
-	mockWsClient.EXPECT().Serve().Do(func() {
+	mockWsClient.EXPECT().Connect().Do(func() {
 		// Serve() cancels the context
 		cancel()
-	}).Return(io.EOF).MinTimes(1)
+	}).Return(nil).MinTimes(1)
```
petderek: This feels like a pretty fundamental change to how we expect the server to behave.

shubham2892: This test tests if handler retries to re-establish the session with ACS when DiscoverPollEndpoint throws out an error. 

Changing what is returned from `serve()` will change how the acs connections exits after the connection is established.

PR #2001, opened 2019-04-24
eni/watcher: skip logging eni status sent error

fenxiong:
### Summary

Currently when eni trunking is turned on, the following warning log repeatedly appears every 30 seconds:
```
2019-04-18T02:43:50Z [WARN] Udev watcher reconciliation: unable to send state change: udev watcher send ENI state change: eni status already sent: ENI Attachment: attachment=arn:aws:ecs:us-west-2:xxxxx:attachment/31ed60cc-4aa9-4d69-9281-b0fa9ee50b0a;attachmentType=instance-eni;attachmentSent=true;mac=06:49:fd:79:f4:10;status=ATTACHED;expiresAt=2019-04-18 02:45:53.693192688 +0000 UTC m=+183.540686203
```
This does not really indicate a problem since we did not actually call the api to send something we already sent. it's just that the watcher will log this whenever it finds the eni. In eni trunking we attach an ENI upon startup, so the log appears all the time. Since we already have log when submitting eni state change, this log is not needed for debug and is confusing, so removing it.

Note that the log is also removed when eni trunking is disabled, as i don't think it has any value in that case either so didn't differentiate between the two.

### Testing

Manually tested this by launching an instance with eni trunking enabled.

### Licensing

### DIFF
--- a/agent/eni/watcher/watcher_linux.go
+++ b/agent/eni/watcher/watcher_linux.go
@@ -18,6 +18,7 @@ package watcher
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	log "github.com/cihub/seelog"
@@ -73,6 +74,10 @@ const (
 	// for the host to learn about an ENIs mac address from netlink.LinkList().
 	// We are capping off this duration to 1s assuming worst-case behavior
 	macAddressRetryTimeout = 2 * time.Second
+
+	// eniStatusSentMsg is the error message to use when trying to send an eni status that's
+	// already been sent
+	eniStatusSentMsg = "eni status already sent"
 )
 
 // UdevWatcher maintains the state of attached ENIs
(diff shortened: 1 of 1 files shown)


agent/eni/watcher/watcher_linux.go
```diff
@@ -184,6 +189,11 @@ func (udevWatcher *UdevWatcher) reconcileOnce() error {
 	// Add new interfaces next
 	for mac := range currentState {
 		if err := udevWatcher.sendENIStateChange(mac); err != nil {
+			// skip logging status sent error as it's redundant and doesn't really indicate a problem
+			if strings.Contains(err.Error(), eniStatusSentMsg) {
+				continue
+			}
```
shubham2892: Will this show the log message with ENI trunking disabled?

fenxiong: no. but i don't think it needs to show in that case either?

shubham2892: Sounds good.

fenxiong:
Rebased to bing in test fix. no conflicts.

PR #2004, opened 2019-04-25
taskresources: set terminal reason msg for all errors

### LINKED ISSUE #1631: taskresources: ensure terminal reason messages are applied in all known failure paths
### Summary
For task resource cgroup and volume, the terminal reason attributes need to to added to the model and when it has errors, the error message needs to be populated into the terminal reason. So the terminal reason can be communicated to service side and show to customers. 

### Expected Behavior
As what ASMAuthResource is doing nowadays, there is a terminal reason attribute in the model, and every time when resource logic has errors, this attribute will be populated with the error message. Later on, it will be used as task stop reason and customers are able to see it.

### Observed Behavior
So far, cgroup and volume do not have this attribute, so the error message cannot be communicated back to customers.

sparrc:
closes #1631

### Summary

This adds a terminal reason to all relevant errors in taskresources

### Licensing

### DIFF
--- a/agent/taskresource/volume/dockervolume.go
+++ b/agent/taskresource/volume/dockervolume.go
@@ -34,10 +34,9 @@ const (
 	// SharedScope indicates that the volume's lifecycle is outside the scope of task
 	SharedScope = "shared"
 	// DockerLocalVolumeDriver is the name of the docker default volume driver
-	DockerLocalVolumeDriver = "local"
+	DockerLocalVolumeDriver   = "local"
+	resourceProvisioningError = "VolumeError: Agent could not create task's volume resources"
 )
-
-const resourceProvisioningError = "VolumeError: Agent could not create task's volume resources"
 
 // VolumeResource represents volume resource
 type VolumeResource struct {
@@ -56,6 +55,12 @@ type VolumeResource struct {
 	statusToTransitions map[resourcestatus.ResourceStatus]func() error
 	client              dockerapi.DockerClient
 	ctx                 context.Context
+
+	// terminalReason should be set for resource creation failures. This ensures
+	// the resource object carries some context for why provisoning failed.
+	terminalReason     string
+	terminalReasonOnce sync.Once
+
 	// lock is used for fields that are accessed and updated concurrently
 	lock sync.RWMutex
 }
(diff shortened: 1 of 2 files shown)


agent/taskresource/volume/dockervolume.go
```diff
@@ -56,6 +55,12 @@ type VolumeResource struct {
 	statusToTransitions map[resourcestatus.ResourceStatus]func() error
 	client              dockerapi.DockerClient
 	ctx                 context.Context
+
+	// terminalReason should be set for resource creation failures. This ensures
+	// the resource object carries some context for why provisoning failed.
+	terminalReason     string
+	terminalReasonOnce sync.Once
```
petderek: Whats the rationale for including the sync.Once?

adnxn: hm it was in the existing codepaths that populated the terminalReason string. need to dig into the old ones to figure out why we added that in the first place 🤔

adnxn: sigh, noting outside of `This field does not accept updates.` [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1822)

im thinking since its a public setter for terminal reason, we didn't want the field clobbered once it's been set

agent/taskresource/cgroup/cgroup.go
```diff
@@ -88,9 +94,18 @@ func NewCgroupResource(taskARN string,
 // GetTerminalReason returns an error string to propagate up through to task
 // state change messages
 func (cgroup *CgroupResource) GetTerminalReason() string {
-	// for cgroups we can send up a static string because this is an
-	// implementation detail and unrelated to customer resources
-	return resourceProvisioningError
+	if cgroup.terminalReason == "" {
```
sharanyad: Details about cgroup creation failure was not surfaced for the same reason as mentioned in the comment. It is mostly an implementation within agent and customers are unaware of it. I don't believe we need to surface it :)

sparrc: Does that apply to volume errors as well? #1631 mentions volume and cgroup.

sharanyad: Volume is still a resource that customer configures through task definition, so it may be useful.

sparrc: okay, I will change it to only volumes

sparrc:
> do you think we should amend the related task resource tests? so we're checking for terminal string in tests that would have the `terminalReason` populated.

yes I will add unit tests 👍

PR #2008, opened 2019-04-29
Fix error detection case when image that is being deleted does not exist - continued.

bendavies:
### Summary
Corrects the fix provided in https://github.com/aws/amazon-ecs-agent/pull/1897.

We were experiencing images not being removed and upgraded `ecs-agent` to fix the issue, after seeing the above PR.
It did not fix the issue. 

Investigation revealed that the above PR does not fix the issue (at least for more recent docker versions because the actual error return from docker is `Error: No such image` - capital `N`.

### Implementation details
Make the string contains check for `no such image` case insensitive.

### Testing

 - update test `TestDeleteImageNotFoundError` to make the error the correct case.
 - left test `TestDeleteImageNotFoundOldDockerMessageError` alone, using existing lower case error message. I am not sure if that is the correct case for old docker versions, but now we have 2 tests covering both cases.

New tests cover the changes: yes, corrected existing.

Bug - Fix cleaning up images when some images are not found. Continuation of #1897

### Licensing

### DIFF
--- a/agent/engine/docker_image_manager.go
+++ b/agent/engine/docker_image_manager.go
@@ -547,7 +547,7 @@ func (imageManager *dockerImageManager) deleteImage(ctx context.Context, imageID
 	seelog.Infof("Removing Image: %s", imageID)
 	err := imageManager.client.RemoveImage(ctx, imageID, dockerclient.RemoveImageTimeout)
 	if err != nil {
-		if strings.Contains(err.Error(), imageNotFoundForDeletionError) {
+		if strings.Contains(strings.ToLower(err.Error()), imageNotFoundForDeletionError) {
 			seelog.Errorf("Image already removed from the instance: %v", err)
 		} else {
 			seelog.Errorf("Error removing Image %v - %v", imageID, err)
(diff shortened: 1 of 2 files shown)


bendavies:
By the way. can someone explain to me how the original report https://github.com/aws/amazon-ecs-agent/issues/1871 was fixed?
The original report was that processing of image removal stops when a `not found` image was encountered, but I can't follow or see how that occurs at all. I'm not a `Go` person however, so that might explain it :)

yunhee-l:
Could you fix the error reported by travis and update this PR please?
agent\engine\docker_image_manager.go:541:23: cannot refer to unexported name strings.toLower
agent\engine\docker_image_manager.go:541:23: undefined: strings.toLower

Also, PR https://github.com/aws/amazon-ecs-agent/pull/1897 is the PR that's fixing issue https://github.com/aws/amazon-ecs-agent/issues/1871.

bendavies:
hi @yunhee-l thanks - fixed.
Yes, I mention #1897 in my OP, but I can't see how it fixes the symptom of image removal being halted after encountering a `not found` image, as described in #1871. Maybe i'm missing something obvious.

bendavies:
ping @shubham2892 as the original fixer

shubham2892:
> By the way, can someone explain to me how the original report #1871 was fixed?

@bendavies Thanks for looking into this. 

As far as I remember, it's related to the cleaning that happens after the `If` [statement](https://github.com/shubham2892/amazon-ecs-agent/blob/1d80631d77c38fc3b05f6f6d4f32f560e910bf87/agent/engine/docker_image_manager.go#L549), feel free to explore this case more, it's possible I might have missed something.

shubham2892:
@bendavies Can you rebase the branch against dev please. Thanks.

bendavies:
@shubham2892 done

bendavies:
will this make it into V1.28.0?

yunhee-l:
It should be in V1.29.0.

PR #2010, opened 2019-04-30
Wait status active/inactive when task networking is enabled

fenxiong:
### Summary

When task networking is enabled and instance type supports eni trunking, eni trunking will be turned on. In that case, after agent registers, there will be a "REGISTERING" status for container instance where we attach the trunk ENI and no task can be placed on it during that time, and after container instance is deregistered, there will be a "DEREGISTERING" status where the trunk is detached. We need to wait for container instance to reach ACTIVE to avoid start task failure, and wait for container instance to reach INACTIVE to ensure the trunk is detached.

### Implementation details

Add waiting in start/cleanup util function.

### Testing

All tests passed in gamma. When backend trunking changes are ready, this can be tested by our CI.
update: backend changes are ready now, and tests pass in prod.

### Licensing

### DIFF
--- a/agent/functional_tests/util/utils.go
+++ b/agent/functional_tests/util/utils.go
@@ -51,9 +51,10 @@ import (
 )
 
 const (
-	arnResourceSections  = 2
-	arnResourceDelimiter = "/"
-	bytePerMegabyte      = 1024 * 1024
+	arnResourceSections                 = 2
+	arnResourceDelimiter                = "/"
+	bytePerMegabyte                     = 1024 * 1024
+	waitContainerInstanceStatusDuration = time.Minute
 )
 
 // GetTaskDefinition is a helper that provies the family:revision for the named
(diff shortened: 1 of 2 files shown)


agent/functional_tests/util/utils.go
```diff
@@ -886,3 +887,60 @@ func GetTaskID(taskARN string) (string, error) {
...
+		errChan <- nil
+	}()
+
+	select {
+	case err := <-errChan:
+		return err
+	case <-timer.C:
+		cancelled = true
+		return errors.Errorf("timed out waiting for container instance '%s' to reach 'ACTIVE', status is '%s'",
+			agent.ContainerInstanceArn, containerInstanceStatus)
```
tommyhahn: What happens if the desiredStatus is "INACTIVE" but something goes wrong? Any error message or something else to indicate that the trunk is not detached successfully?

fenxiong: here i'm assuming the trunk is detached when instance goes to INACTIVE, which is the expected behavior. I'm adding a test [here](https://github.com/aws/amazon-ecs-agent/pull/1961) to check whether that's true. i don't think we need to check this in many tests, since if somehow the logic breaks there's already a failing test that can inform us of the issue.

fenxiong: maybe i misunderstood your concern. if backend failed to detach the trunk, the instance will still go to INACTIVE but not very soon. we will time out here.

tommyhahn: Then in that case, we might need to output an error message like "timed out waiting for container instance to reach "INACTIVE" because of the failure of trunk detachment" separately?

fenxiong: in case of timeout, the agent can't tell whether backend has tried something and failed, or it's slow and hasn't tried anything.

agent/functional_tests/util/utils_unix.go
```diff
@@ -133,6 +134,12 @@ func RunAgent(t *testing.T, options *AgentOptions) *TestAgent {
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	if options != nil && options.EnableTaskENI {
+		// if task networking is enabled, needs to wait for container instance to become active
+		err = agent.WaitContainerInstanceStatus("ACTIVE")
```
shubham2892: Can this be skipped if the instance type does not support trunking?

fenxiong: we can do that. but i'm trying to avoid more instance identity document dependency as per [this comment](https://github.com/aws/amazon-ecs-agent/pull/1961#discussion_r279854839). i think it's probably fine to add two more DescribeContainerInstances call for a few tests (currently around 5?) to avoid that dependency? the test can pass either way..

shubham2892: Umm, your call, totally.

PR #2012, opened 2019-04-30
adds StorageReadBytes and StorageWriteBytes to rawstats collection

fierlion:
adds ULongStatsSet with int64 truncation to queue

### Summary
This adds StorageReadBytes and StorageWriteBytes to the rawstats collection as well as adding ULongStatsSet and int64 truncation to the queue.  

### Implementation details
changes follow existing patterns across multiple files.  Where necessary I changed names to be more general for re-use.

### Testing
For most of this, I followed TDD and wrote my tests first, then updated the stats code.  I added multiple tests to cover the new stats, and the overflow uint64 --> int64 truncation. 

add block stats collection

### Licensing

### DIFF
--- a/agent/stats/queue.go
+++ b/agent/stats/queue.go
@@ -82,6 +82,8 @@ func (queue *Queue) add(rawStat *ContainerStats) {
 	stat := UsageStats{
 		CPUUsagePerc:      float32(nan32()),
 		MemoryUsageInMegs: uint32(rawStat.memoryUsage / BytesInMiB),
+		StorageReadBytes:  rawStat.storageReadBytes,
+		StorageWriteBytes: rawStat.storageWriteBytes,
 		Timestamp:         rawStat.timestamp,
 		cpuUsage:          rawStat.cpuUsage,
 	}
@@ -125,6 +127,16 @@ func (queue *Queue) GetMemoryStatsSet() (*ecstcs.CWStatsSet, error) {
 	return queue.getCWStatsSet(getMemoryUsagePerc)
 }
 
+// GetStorageReadStatsSet gets the stats set for aggregate storage read
+func (queue *Queue) GetStorageReadStatsSet() (*ecstcs.ULongStatsSet, error) {
+	return queue.getULongStatsSet(getStorageReadBytes)
+}
+
+// GetStorageWriteStatsSet gets the stats set for aggregate storage written
+func (queue *Queue) GetStorageWriteStatsSet() (*ecstcs.ULongStatsSet, error) {
+	return queue.getULongStatsSet(getStorageWriteBytes)
+}
+
 // GetRawUsageStats gets the array of most recent raw UsageStats, in descending
 // order of timestamps.
 func (queue *Queue) GetRawUsageStats(numStats int) ([]UsageStats, error) {
@@ -147,6 +159,8 @@ func (queue *Queue) GetRawUsageStats(numStats int) ([]UsageStats, error) {
 		usageStats[i] = UsageStats{
 			CPUUsagePerc:      rawUsageStat.CPUUsagePerc,
 			MemoryUsageInMegs: rawUsageStat.MemoryUsageInMegs,
+			StorageReadBytes:  rawUsageStat.StorageReadBytes,
+			StorageWriteBytes: rawUsageStat.StorageWriteBytes,
 			Timestamp:         rawUsageStat.Timestamp,
 		}
 	}
@@ -162,7 +176,26 @@ func getMemoryUsagePerc(s *UsageStats) float64 {
 	return float64(s.MemoryUsageInMegs)
 }
 
-type getUsageFunc func(*UsageStats) float64
+func getStorageReadBytes(s *UsageStats) uint64 {
+	return s.StorageReadBytes
+}
+
+func getStorageWriteBytes(s *UsageStats) uint64 {
+	return s.StorageWriteBytes
+}
+
+// getInt64WithOverflow truncates a uint64 to fit an int64
+// it returns overflow as a second int64
+func getInt64WithOverflow(uintStat uint64) (int64, int64) {
+	if uintStat > math.MaxInt64 {
+		overflow := int64(uintStat % uint64(math.MaxInt64))
+		return math.MaxInt64, overflow
+	}
+	return int64(uintStat), int64(0)
+}
+
+type getUsageFloatFunc func(*UsageStats) float64
+type getUsageIntFunc func(*UsageStats) uint64
 
 func (queue *Queue) resetThresholdElapsed(timeout time.Duration) bool {
 	queue.lock.RLock()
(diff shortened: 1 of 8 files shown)


fierlion:
NOTE: TestShutdownOrder integ test is failing intermittently.  It succeeded for arm but not linux(AL2).  I opened a ticket for it.  This failure is not related to the above changes.

agent/stats/utils_unix.go
```diff
@@ -31,9 +31,25 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 
 	cpuUsage := dockerStats.CPUStats.CPUUsage.TotalUsage / numCores
 	memoryUsage := dockerStats.MemoryStats.Usage - dockerStats.MemoryStats.Stats["cache"]
+	// initialize block io and loop over stats to aggregate
+	storageReadBytes := uint64(0)
```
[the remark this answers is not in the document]
fierlion: there are several places where just declaring `expectUint64Here := 0` would initialize as an int type.  I could also have declared this as `var storageReadBytes uint64 = 0`.  Either is fine.  Just need a way to tell the compiler which type to initialize.

agent/stats/queue.go
```diff
@@ -162,7 +176,26 @@ func getMemoryUsagePerc(s *UsageStats) float64 {
...
+	return s.StorageReadBytes
+}
+
+func getStorageWriteBytes(s *UsageStats) uint64 {
+	return s.StorageWriteBytes
+}
+
+// getOverflow truncates a uint64 to fit an int64
+// it returns overflow as a second int64
+func getOverflow(uintStat uint64) (int64, int64) {
```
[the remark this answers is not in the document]
fierlion: How does `getInt64WithOverflow(uintStat uint64)` sound?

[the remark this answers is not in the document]
sharanyad: sounds good!

agent/stats/queue_test.go
```diff
@@ -30,6 +30,9 @@ const (
 	// predictableHighMemoryUtilizationInMiB is the expected Memory usage in MiB for
 	// the "predictableHighMemoryUtilizationInBytes" value (7377772544 / (1024 * 1024))
 	predictableHighMemoryUtilizationInMiB = 7035
+	// the "predictableInt64Overflow" requires 3 metrics to guarantee overflow
+	// note math.MaxInt64 is odd, so integer division will trim off 1
+	predictableInt64Overflow = math.MaxInt64 / int64(2)
```
sharanyad: nit: this can just be `math.MaxInt64 / 2`

agent/stats/queue_test.go
```diff
@@ -122,26 +125,43 @@ func getPredictableHighMemoryUtilizationInBytes(size int) []uint64 {
 	return memBytes
 }
 
-func createQueue(size int, predictableHighMemoryUtilization bool) *Queue {
+func getPredictableInt64Overflow(size int) []uint64 {
```
sharanyad: this method name is a bit misleading.. how about calling it getLargeInt64Stats ?

agent/stats/queue_test.go
```diff
@@ -189,9 +209,45 @@ func TestQueueAddRemove(t *testing.T) {
 		t.Error("Sum value incorrectly set: ", *memStatsSet.Sum)
 	}
 
+	storageReadStatsSet, err := queue.GetStorageReadStatsSet()
+	if err != nil {
+		t.Error("Error getting storage read stats set:", err)
+	}
+	// assuming min is initialized to math.MaxUint64 then truncated
+	if *storageReadStatsSet.Min == int64(math.MaxInt64) &&
```
fierlion: I'll remove the int64 cast here and in line 217.  These are the result of an overabundance of caution.

agent/stats/queue_test.go
fierlion: when I do this I get:
`constant 4611686018427387903 overflows int`
the cast is necessary here

agent/stats/queue_test.go
fierlion: sure.  I'll make that change.

agent/stats/queue.go
```diff
@@ -215,3 +249,51 @@ func (queue *Queue) getCWStatsSet(f getUsageFunc) (*ecstcs.CWStatsSet, error) {
...
+	for _, stat := range queue.buffer {
+		thisStat := f(&stat)
+		if thisStat < min {
+			min = thisStat
+		}
+		if thisStat > max {
+			max = thisStat
+		}
+		sum += thisStat
+		sampleCount++
```
tommyhahn: Do we need to calculate this sampleCount? Could we use queueLength?

agent/stats/queue.go
```diff
@@ -215,3 +249,51 @@ func (queue *Queue) getCWStatsSet(f getUsageFunc) (*ecstcs.CWStatsSet, error) {
...
+// where there is overflow (math.MaxInt64 + 1 or greater)
+// we capture the excess in optional overflow fields.
+func (queue *Queue) getULongStatsSet(f getUsageIntFunc) (*ecstcs.ULongStatsSet, error) {
+	queue.lock.Lock()
+	defer queue.lock.Unlock()
+
+	queueLength := len(queue.buffer)
+	if queueLength < 2 {
+		// Need at least 2 data points to calculate this.
+		return nil, fmt.Errorf("No data in the queue")
```
tommyhahn: Error message: "We need at least 2 data points to calculate this" might be better?

agent/stats/queue_test.go
```diff
@@ -257,6 +339,26 @@ func TestQueueAddPredictableHighMemoryUtilization(t *testing.T) {
 	}
 }
 
+func TestUintOverflow(t *testing.T) {
```
petderek: What are your thoughts on rewriting this as a table-driven test?

Another thing I'd like to see here is uint64(math.MaxInt64) + 1

fierlion: In my eyes, the `baseOverUint, overflowOverUint := getInt64WithOverflow(overUint)` check is checking the same thing as `uint64(math.MaxInt64) + 1`

`overUint = uint64(maxInt) + uint64(overInt)` is essentially `uint64(math.MaxInt64) + 20`
Would you prefer I change the 20 to a 1?

fierlion: and as for the table, I think that only 3 checks are necessary: `<`, `=` and `>` so it doesn't really need to get fancier.  I could narrow the `<` and `>` range to 1 if you think it would make a difference.

agent/stats/queue.go
fierlion: as I see it, either is fine.  I wanted to duplicate what we do in cwstats function above.

agent/stats/queue.go
fierlion: I'll make this change.

agent/stats/queue.go
```diff
@@ -179,7 +212,8 @@ func (queue *Queue) enoughDatapointsInBuffer() bool {
 
 // getCWStatsSet gets the stats set for either CPU or Memory based on the
 // function pointer.
-func (queue *Queue) getCWStatsSet(f getUsageFunc) (*ecstcs.CWStatsSet, error) {
+// stats are float64 type
```
yumex93: nit: add the period at the end of sentence?

fierlion: `stats are float64 type.` ? sure I'll make this change.

yumex93: nit: I think there are other places also missing the period for the comments before function. Since it can be used to generate go doc. (I am not sure whether people will do this). It will be nice if our comments are well formatted. Other than this, LGTM.

fierlion: I ended up removing this particular comment.  When we have time, it wouldn't be a terrible thing to audit our comments agent-wide for style issues.

agent/stats/queue_test.go
petderek: The QA in me likes thinking about "arbitrarily larger" and "maxint + 1" as different cases, since "maxint+1" is where the overflow would actually occur.

fierlion: made this update.

petderek: ^ nvm, we talked offline and what you have makes sense.

n-1 and n cover the overflow case. (The plot twist is that *I* was the off by one error the whole time)

Issue #1684, opened 2018-11-09
Clean up dead containers after ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION


yumex93:
### Summary
Based on what I got from customers, so far after ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION, agent cleans up only the stopped tasks and docker images that are not being used by any tasks on your container instances. They also want agent to clean up containers in 'dead' status. 

Currently, customers' workaround is to run this command "#docker rm $(docker ps --all -q -f status=dead)" to manually remove the dead containers.

sparrc:
While working on the solution to this I also noticed that we had a problem with cleaning up "dangling" docker images (`docker images -f dangling=true`). This was due to us trying to inspect images based on the name rather than the ID (which in the case of dangling images is `<none>`).

I have included a fix for this as well as part of #2015

sparrc:
closed by #2015

PR #2015, opened 2019-05-02
image manager: cleanup 'dead' and 'created' containers

sparrc:
also cleanup of 'dangling' images that have no tags or names associated
with them (ie, they show as <none> in 'docker images')

see #1684

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/docker_image_manager.go
+++ b/agent/engine/docker_image_manager.go
@@ -312,6 +312,7 @@ func (imageManager *dockerImageManager) removeUnusedImages(ctx context.Context) 
 
 	var numECSImagesDeleted int
 	imageManager.imageStatesConsideredForDeletion = imageManager.imagesConsiderForDeletion(imageManager.getAllImageStates())
+
 	for i := 0; i < imageManager.numImagesToDelete; i++ {
 		err := imageManager.removeLeastRecentlyUsedImage(ctx)
 		numECSImagesDeleted = i
@@ -342,9 +343,17 @@ func (imageManager *dockerImageManager) removeNonECSContainers(ctx context.Conte
 			continue
 		}
 
-		finishedTime, _ := time.Parse(time.Now().String(), response.State.FinishedAt)
+		seelog.Debugf("Inspecting Non-ECS Container ID [%s] for removal, Finished [%s] Status [%s]", id, response.State.FinishedAt, response.State.Status)
+		finishedTime, err := time.Parse(time.RFC3339Nano, response.State.FinishedAt)
+		if err != nil {
+			seelog.Errorf("Error parsing time string for container. id: %s, time: %s err: %s", id, response.State.FinishedAt, err)
+			continue
+		}
 
-		if response.State.Status == "exited" && time.Now().Sub(finishedTime) > imageManager.nonECSContainerCleanupWaitDuration {
+		if (response.State.Status == "exited" ||
+			response.State.Status == "dead" ||
+			response.State.Status == "created") &&
+			time.Now().Sub(finishedTime) > imageManager.nonECSContainerCleanupWaitDuration {
 			nonECSContainerRemoveAvailableIDs = append(nonECSContainerRemoveAvailableIDs, id)
 		}
 	}
@@ -353,13 +362,13 @@ func (imageManager *dockerImageManager) removeNonECSContainers(ctx context.Conte
 		if numNonECSContainerDeleted == imageManager.numNonECSContainersToDelete {
 			break
 		}
-		seelog.Infof("Removing non-ECS container id: %s", id)
+		seelog.Debugf("Removing non-ECS Container ID %s", id)
 		err := imageManager.client.RemoveContainer(ctx, id, dockerclient.RemoveContainerTimeout)
 		if err == nil {
-			seelog.Infof("Image removed: %s", id)
+			seelog.Infof("Removed Container ID: %s", id)
 			numNonECSContainerDeleted++
 		} else {
-			seelog.Errorf("Error removing Image %s - %v", id, err)
+			seelog.Errorf("Error Removing Container ID %s - %s", id, err)
 			continue
 		}
 	}
--- a/agent/engine/docker_image_manager_test.go
+++ b/agent/engine/docker_image_manager_test.go
@@ -27,7 +27,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
-	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi/mocks"
+	mock_dockerapi "github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi/mocks"
 	"github.com/aws/amazon-ecs-agent/agent/ec2"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/engine/image"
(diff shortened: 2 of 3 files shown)


agent/engine/docker_image_manager.go
```diff
@@ -342,9 +343,13 @@ func (imageManager *dockerImageManager) removeNonECSContainers(ctx context.Conte
 			continue
 		}
 
-		finishedTime, _ := time.Parse(time.Now().String(), response.State.FinishedAt)
+		seelog.Infof("Inspecting Non-ECS Container ID [%s] for removal, Finished [%s] Status [%s]", id, response.State.FinishedAt, response.State.Status)
```
sharanyad: this log can be at debug level

agent/engine/docker_image_manager.go
```diff
@@ -342,9 +343,13 @@ func (imageManager *dockerImageManager) removeNonECSContainers(ctx context.Conte
...
+		finishedTime, err := time.Parse(time.RFC3339Nano, response.State.FinishedAt)
+		if err != nil {
+			seelog.Errorf("Error parsing time string for container. id: %s, time: %s err: %s", id, response.State.FinishedAt, err)
+		}
 
-		if response.State.Status == "exited" && time.Now().Sub(finishedTime) > imageManager.nonECSContainerCleanupWaitDuration {
+		if (response.State.Status == "exited" || response.State.Status == "dead" || response.State.Status == "created") && time.Now().Sub(finishedTime) > imageManager.nonECSContainerCleanupWaitDuration {
```
sharanyad: can you please break this into multiple lines, so it's readable?

agent/engine/docker_image_manager.go
```diff
@@ -353,13 +358,13 @@ func (imageManager *dockerImageManager) removeNonECSContainers(ctx context.Conte
 		if numNonECSContainerDeleted == imageManager.numNonECSContainersToDelete {
 			break
 		}
-		seelog.Infof("Removing non-ECS container id: %s", id)
+		seelog.Debugf("Removing non-ECS container id: %s", id)
 		err := imageManager.client.RemoveContainer(ctx, id, dockerclient.RemoveContainerTimeout)
 		if err == nil {
 			seelog.Infof("Image removed: %s", id)
```
sharanyad: should this be `Container removed?`

agent/engine/docker_image_manager.go
```diff
@@ -379,70 +384,75 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
 	if nonECSImagesNumToDelete == 0 {
 		return
 	}
-	var nonECSImageNames = imageManager.getNonECSImageNames(ctx)
-	var nonECSImageNamesRemoveEligible []string
-	for _, nonECSImage := range nonECSImageNames {
-		if !isInExclusionList(nonECSImage, imageManager.imageCleanupExclusionList) {
-			nonECSImageNamesRemoveEligible = append(nonECSImageNamesRemoveEligible, nonECSImage)
+	nonECSImages := imageManager.getNonECSImages(ctx)
+	var nonECSImagesRmEligible []ImageWithSizeID
```
sharanyad: nit: don't think we need the word 'remove' or 'rm' in it since its inside `removeNonECSImages` method.

agent/engine/docker_image_manager.go
```diff
@@ -379,70 +384,75 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
 	}
 
-	var imageWithSizeList []ImageWithSize
-	for _, imageName := range nonECSImageNamesRemoveEligible {
-		resp, iiErr := imageManager.client.InspectImage(imageName)
-		if iiErr != nil {
-			seelog.Errorf("Error inspecting non-ECS image name: %s - %v", imageName, iiErr)
+	// Get the all image sizes
+	for _, image := range nonECSImagesRmEligible {
+		resp, err := imageManager.client.InspectImage(image.ImageID)
```
sharanyad: any reason why this was change from imageName to imageID? 
Sometimes, two images can have same ID but different names. just making sure this doesn't break anything.

agent/engine/docker_image_manager.go
sparrc: 👍

agent/engine/docker_image_manager.go
sparrc: yes, I will fix that 👍

agent/engine/docker_image_manager.go
sparrc: The reason I changed this to image ID is because with image name we don't properly cleanup 'dangling' images. I made a comment about this here: https://github.com/aws/amazon-ecs-agent/issues/1684#issuecomment-488866388

But that is a good catch, I did not realize two images can have the same ID but a different name. Maybe changing this behavior should be broken out into a separate issue and PR, if we want to do it at all.

agent/engine/docker_image_manager.go
sparrc: done

agent/engine/docker_image_manager.go
sparrc: done

agent/engine/docker_image_manager.go
sparrc: okay, I've removed the parts that removed 'dangling images'

agent/engine/docker_image_manager.go
```diff
@@ -342,9 +343,16 @@ func (imageManager *dockerImageManager) removeNonECSContainers(ctx context.Conte
 			continue
 		}
 
-		finishedTime, _ := time.Parse(time.Now().String(), response.State.FinishedAt)
+		seelog.Debugf("Inspecting Non-ECS Container ID [%s] for removal, Finished [%s] Status [%s]", id, response.State.FinishedAt, response.State.Status)
+		finishedTime, err := time.Parse(time.RFC3339Nano, response.State.FinishedAt)
+		if err != nil {
```
sharanyad: adding as note as discussed offline: do not remove containers that do not have `finishedTime` and in `created` state

sparrc: fixed

agent/engine/docker_image_manager_test.go
```diff
@@ -1002,20 +1004,215 @@ func TestNonECSImageAndContainersCleanupRemoveImage(t *testing.T) {
...
+	client.EXPECT().InspectImage(listImagesResponse.RepoTags[0]).Return(inspectImageResponse, nil).Times(1)
+	client.EXPECT().RemoveImage(gomock.Any(), listImagesResponse.RepoTags[0], dockerclient.RemoveImageTimeout).Return(nil).Times(1)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	imageManager.removeUnusedImages(ctx)
+
+	if len(listContainersResponse.DockerIDs) != 1 {
+		t.Error("Error removing containers ids")
```
sharanyad: you can use `assert.Len` https://godoc.org/github.com/stretchr/testify/assert#Assertions.Len / `assert.Equal` for this and below ones https://godoc.org/github.com/stretchr/testify/assert#Equal

agent/engine/docker_image_manager.go
```diff
@@ -342,9 +343,17 @@ func (imageManager *dockerImageManager) removeNonECSContainers(ctx context.Conte
...
+			seelog.Errorf("Error parsing time string for container. id: %s, time: %s err: %s", id, response.State.FinishedAt, err)
+			continue
```
sharanyad: Could you please add a test case for this condition as well?

sparrc: yes, done

agent/engine/docker_image_manager_test.go
sparrc: 👍

PR #2016, opened 2019-05-02
add network stats to container metrics

sharanyad:
### Summary
Adds new network stats to task metrics. 
Piggybacks on https://github.com/aws/amazon-ecs-agent/pull/2012 with minor refactoring, windows network metrics and adding to CW metrics. 

### Testing

### Licensing

### DIFF
--- a/agent/stats/common_test.go
+++ b/agent/stats/common_test.go
@@ -154,9 +154,6 @@ func validateContainerMetrics(containerMetrics []*ecstcs.ContainerMetric, expect
 		if containerMetric.MemoryStatsSet == nil {
 			return fmt.Errorf("MemoryStatsSet is nil")
 		}
-		if containerMetric.NetworkStatsSet == nil {
-			return fmt.Errorf("NetworkStatsSet is nil")
-		}
 	}
 	return nil
 }
@@ -251,14 +248,14 @@ func validateEmptyTaskHealthMetrics(t *testing.T, engine *DockerStatsEngine) {
 
 func createFakeContainerStats() []*ContainerStats {
 	netStats := &NetworkStats{
-		rxBytes:   796,
-		rxDropped: 6,
-		rxErrors:  0,
-		rxPackets: 10,
-		txBytes:   8192,
-		txDropped: 5,
-		txErrors:  0,
-		txPackets: 60,
+		RxBytes:   796,
+		RxDropped: 6,
+		RxErrors:  0,
+		RxPackets: 10,
+		TxBytes:   8192,
+		TxDropped: 5,
+		TxErrors:  0,
+		TxPackets: 60,
 	}
 	return []*ContainerStats{
 		{22400432, 1839104, uint64(0), uint64(0), netStats, parseNanoTime("2015-02-12T21:22:05.131117533Z")},
--- a/agent/stats/common_test.go
+++ b/agent/stats/common_test.go
@@ -154,6 +154,9 @@ func validateContainerMetrics(containerMetrics []*ecstcs.ContainerMetric, expect
 		if containerMetric.MemoryStatsSet == nil {
 			return fmt.Errorf("MemoryStatsSet is nil")
 		}
+		if containerMetric.NetworkStatsSet == nil {
+			return fmt.Errorf("NetworkStatsSet is nil")
+		}
 	}
 	return nil
 }
@@ -247,9 +250,19 @@ func validateEmptyTaskHealthMetrics(t *testing.T, engine *DockerStatsEngine) {
 }
 
 func createFakeContainerStats() []*ContainerStats {
+	netStats := &NetworkStats{
+		rxBytes:   796,
+		rxDropped: 6,
+		rxErrors:  0,
+		rxPackets: 10,
+		txBytes:   8192,
+		txDropped: 5,
+		txErrors:  0,
+		txPackets: 60,
+	}
 	return []*ContainerStats{
-		{22400432, 1839104, uint64(0), uint64(0), parseNanoTime("2015-02-12T21:22:05.131117533Z")},
-		{116499979, 3649536, uint64(0), uint64(0), parseNanoTime("2015-02-12T21:22:05.232291187Z")},
+		{22400432, 1839104, uint64(0), uint64(0), netStats, parseNanoTime("2015-02-12T21:22:05.131117533Z")},
+		{116499979, 3649536, uint64(0), uint64(0), netStats, parseNanoTime("2015-02-12T21:22:05.232291187Z")},
 	}
 }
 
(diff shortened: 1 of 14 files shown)


agent/stats/common_test.go
```diff
@@ -247,9 +250,19 @@ func validateEmptyTaskHealthMetrics(t *testing.T, engine *DockerStatsEngine) {
...
+		txBytes:   8192,
+		txDropped: 5,
+		txErrors:  0,
+		txPackets: 60,
+	}
 	return []*ContainerStats{
-		{22400432, 1839104, uint64(0), uint64(0), parseNanoTime("2015-02-12T21:22:05.131117533Z")},
-		{116499979, 3649536, uint64(0), uint64(0), parseNanoTime("2015-02-12T21:22:05.232291187Z")},
+		{22400432, 1839104, uint64(0), uint64(0), netStats, parseNanoTime("2015-02-12T21:22:05.131117533Z")},
+		{116499979, 3649536, uint64(0), uint64(0), netStats, parseNanoTime("2015-02-12T21:22:05.232291187Z")},
```
fierlion: good change.  I'll actually need to change the two `uint64(0)` to be `storageReadStats` and `storageWriteStats`.

agent/stats/common_test.go
```diff
@@ -154,6 +154,9 @@ func validateContainerMetrics(containerMetrics []*ecstcs.ContainerMetric, expect
 		if containerMetric.MemoryStatsSet == nil {
 			return fmt.Errorf("MemoryStatsSet is nil")
 		}
+		if containerMetric.NetworkStatsSet == nil {
+			return fmt.Errorf("NetworkStatsSet is nil")
+		}
```
fierlion: Should we also check the type of the stats set here or is that overkill?
In my additions for `storageReadStats` and storageWriteStats` in the `createFakeContainerStats()` function below, I just dropped in a uint64(0) for each; this will pass the null checks above.

agent/stats/engine_test.go
```diff
@@ -190,11 +190,21 @@ func TestStatsEngineMetadataInStatsSets(t *testing.T) {
...
+		txPackets: 60,
+	}
 	containerStats := []*ContainerStats{
-		{22400432, 1839104, uint64(0), uint64(0), ts1},
-		{116499979, 3649536, uint64(0), uint64(0), ts2},
+		{22400432, 1839104, uint64(0), uint64(0), netStats, ts1},
+		{116499979, 3649536, uint64(0), uint64(0), netStats, ts2},
 	}
-	dockerStats := []*types.StatsJSON{{}, {},}
+	dockerStats := []*types.StatsJSON{{}, {}}
```
fierlion: assuming `go fmt` removed this comma.  It did on my branch.

agent/stats/queue.go
```diff
@@ -137,6 +138,101 @@ func (queue *Queue) GetStorageWriteStatsSet() (*ecstcs.ULongStatsSet, error) {
 	return queue.getULongStatsSet(getStorageWriteBytes)
 }
 
+// GetNetworkStatsSet gets the stats set for network metrics.
+func (queue *Queue) GetNetworkStatsSet() (*ecstcs.NetworkStatsSet, error) {
```
fierlion: do we need to lock the queue for this and like methods?  
I see that the `GetRawUsageStats()` Queue method locks during its execution

agent/stats/queue.go
```diff
@@ -137,6 +138,101 @@ func (queue *Queue) GetStorageWriteStatsSet() (*ecstcs.ULongStatsSet, error) {
...
+		seelog.Warnf("Error getting network tx packets: %v", err)
+	}
+	return networkStatsSet, err
+}
+
+func getNetworkRxBytes(s *UsageStats) uint64 {
+	if s.NetworkStats != nil {
+		return s.NetworkStats.rxBytes
+	}
+	return uint64(0)
```
fierlion: for this and subsequent stats where do we surface that our s.<stat> is nil?  
This looks like a silent failure.

agent/stats/queue_test.go
```diff
@@ -20,6 +20,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/tcs/model/ecstcs"
+
```
fierlion: nit extra line

agent/stats/common_test.go
sharanyad: > Should we also check the type of the NetworkStatsSet here? 

what does this mean?

agent/stats/engine_test.go
sharanyad: yes :)

agent/stats/queue.go
sharanyad: this one ultimately calls `getULongStatsSet` , which has lock on the queue.

agent/stats/queue.go
sharanyad: this is not exactly a silent failure, since some network modes might have this as nil. hence just sending 0 . what do you think?

agent/stats/engine_test.go
```diff
@@ -190,11 +190,21 @@ func TestStatsEngineMetadataInStatsSets(t *testing.T) {
 	engine.addAndStartStatsContainer("c1")
 	ts1 := parseNanoTime("2015-02-12T21:22:05.131117533Z")
 	ts2 := parseNanoTime("2015-02-12T21:22:05.232291187Z")
+	netStats := &NetworkStats{
```
yunhee-l: why not use createFakeContainerStats() for this test data?

agent/stats/common_test.go
fierlion: a type check like:
```
networkType, ok := containerMetric.NetworkStatsSet.(ecstcs.NetworkStatsSet)
if !ok {
    return fmt.Errorf("NetworkStatsSet is invalid type")
}
```
Like I said... might be overkill.

agent/stats/queue.go
fierlion: This is fine, as long as we are capturing true stats failures elsewhere.  We will need to clearly define what our true failure cases are -- for example, for storage stats, I'd assume any null stats represent a true failure and I need to surface that somehow.  I might return a uint64(0) here for storage stats, but I need to be sure that I've got a debug statement linked to this failure that will help me troubleshoot.

fierlion: If you agree that we need this, can we at least add a TODO here to be sure we don't lose the thought?

agent/stats/engine_test.go
sharanyad: good point, not sure why this was not used before. will change it

agent/stats/queue.go
sharanyad: network stats being nil is a valid case for host/awsvpc mode. 
instead of throwing an error, I have added a debug statement here https://github.com/aws/amazon-ecs-agent/pull/2016/files#diff-598be0dc5499d16fd579d5ce5dee96abR57

fierlion:
Functional test failures are expected.  Once all other tests are green, this is good to go

PR #2018, opened 2019-05-03
add StorageStatsSet and windows stats collection

fierlion:
### Summary
Adds the StorageStatsSet which we will be using in engine.go.  Adds windows storage stats read from docker stats.

### Implementation details
The `GetStorageStatsSet()` queue method is a combination of the logic previously in `GetStorageReadStatsSet()` and `GetStorageWriteStatsSet()`.  Based on the model, these will not be used individually for logging to cloudwatch.
### Testing
Unit tests added and modified to check new functionality.
Added windows_test_stats.json which mimics the response from windows docker stats output.

updates storageStatsSet

### Licensing

### DIFF
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -594,7 +594,6 @@ func (engine *DockerStatsEngine) taskContainerMetricsUnsafe(taskArn string) ([]*
 			continue
 		}
 
-		// Get memory stats set
 		memoryStatsSet, err := container.statsQueue.GetMemoryStatsSet()
 		if err != nil {
 			seelog.Warnf("Error getting memory stats, err: %v, container: %v", err, dockerID)
--- a/agent/stats/queue_test.go
+++ b/agent/stats/queue_test.go
@@ -219,11 +219,12 @@ func TestQueueAddRemove(t *testing.T) {
 		t.Error("Sum value incorrectly set: ", *memStatsSet.Sum)
 	}
 
-	storageReadStatsSet, err := queue.GetStorageReadStatsSet()
+	storageStatsSet, err := queue.GetStorageStatsSet()
 	if err != nil {
-		t.Error("Error getting storage read stats set:", err)
+		t.Error("Error getting storage stats set:", err)
 	}
 	// assuming min is initialized to math.MaxUint64 then truncated
+	storageReadStatsSet := storageStatsSet.ReadSizeBytes
 	if *storageReadStatsSet.Min == int64(math.MaxInt64) &&
 		*storageReadStatsSet.OverflowMin == int64(math.MaxInt64) {
 		t.Error("Min value incorrectly set: ", *storageReadStatsSet.Min)
@@ -238,10 +239,7 @@ func TestQueueAddRemove(t *testing.T) {
 		t.Error("Sum value incorrectly set: ", *storageReadStatsSet.Sum)
 	}
 
-	storageWriteStatsSet, err := queue.GetStorageWriteStatsSet()
-	if err != nil {
-		t.Error("Error getting storage read stats set:", err)
-	}
+	storageWriteStatsSet := storageStatsSet.WriteSizeBytes
 	if *storageWriteStatsSet.Min == int64(math.MaxInt64) {
 		t.Error("Min value incorrectly set: ", *storageWriteStatsSet.Min)
 	}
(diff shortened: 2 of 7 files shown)


agent/stats/queue_test.go
```diff
@@ -238,7 +239,7 @@ func TestQueueAddRemove(t *testing.T) {
 		t.Error("Sum value incorrectly set: ", *storageReadStatsSet.Sum)
 	}
 
-	storageWriteStatsSet, err := queue.GetStorageWriteStatsSet()
+	storageWriteStatsSet := storageStatsSet.WriteSizeBytes
```
sharanyad: why is err ignored here? also below error message should be `Error getting storage write stats set`

agent/stats/utils_windows_test.go
```diff
@@ -28,41 +27,27 @@ import (
...
 	numCores = uint64(0)
-	jsonStat := fmt.Sprintf(`
-		{
-			"cpu_stats":{
-				"cpu_usage":{
-					"total_usage":%d
-				}
-			}
-		}`, 100)
+	inputJsonFile, _ := filepath.Abs("./windows_test_stats.json")
```
sharanyad: the reason why I didn't change this was - it would be costly to open/read a file on disk for each test. Do you think it's possible to combine below test with this one?

yunhee-l: +1, windows anything is already slow

agent/stats/queue_test.go
fierlion: actually the `err` is logged only once, above on line 222 where we get `storageStatsSet`.  I'm removing this unnecessary err condition altogether.

agent/stats/utils_windows_test.go
fierlion: I combined all tests requiring stats.json into a single test so we only read the file once.

agent/stats/engine.go
```diff
@@ -594,7 +594,6 @@ func (engine *DockerStatsEngine) taskContainerMetricsUnsafe(taskArn string) ([]*
 			continue
 		}
 
-		// Get memory stats set
```
sparrc: 👍

agent/stats/utils_windows.go
```diff
@@ -31,10 +31,14 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 	cpuUsage := (dockerStats.CPUStats.CPUUsage.TotalUsage * 100) / numCores
 	memoryUsage := dockerStats.MemoryStats.PrivateWorkingSet
 	networkStats := getNetworkStats(dockerStats)
+	storageReadBytes := dockerStats.StorageStats.ReadSizeBytes
```
sharanyad: It would be better to check `if dockerStats.StorageStats != nil` , else this could break. 
same for linux (just as a safeguard - sorry missed this earlier)

fierlion: We will need to add this all over the place.  Just in this file we also have `dockerStats.CPUStats.CPUUsage.TotalUsage`, `dockerStats.MemoryStats.PrivateWorkingSet`.  These have been in place for a while now with no checks.  Should we expect that the StorageStats object potentially will not exist in the docker stats?

sharanyad: I'm not sure. I have seen networkStats to be nil in some cases. If you're sure storage stats will never be empty we can leave this as is.

fierlion: Will need to research which stats fields can be nil (which aren't required).

sharanyad: we can revisit it outside the PR

agent/stats/utils_windows_test.go
```diff
@@ -42,39 +42,23 @@ func TestDockerStatsToContainerStatsZeroCoresGeneratesError(t *testing.T) {
...
-
 func TestDockerStatsToContainerStats(t *testing.T) {
+	numCores = 4
 	inputJsonFile, _ := filepath.Abs("./windows_test_stats.json")
 	jsonBytes, _ := ioutil.ReadFile(inputJsonFile)
 	dockerStat := &types.StatsJSON{}
 	json.Unmarshal([]byte(jsonBytes), dockerStat)
 	containerStats, err := dockerStatsToContainerStats(dockerStat)
 	assert.NoError(t, err, "converting container stats failed")
-	require.NotNil(t, containerStats, "containerStats should not be nil")
```
fenxiong: i think this should be require rather than assert. if containerStats is nil, there's no point to check what's inside it

fierlion: I'll make that change.

PR #2021, opened 2019-05-07
adds nil check and logging for windows and unix stats

fierlion:
### Summary
Adds nil check and logging for disk metrics collection from docker stats.

### Implementation details
Initialize all returns to `uint64(0)` so we will never have a nil stats return value.  Add log statements when nil fields are encountered in Docker stats for expected disk metrics.

### Testing
Tested manually on linux and windows ec2.

adds safety checks to disk metrics collection

### Licensing

### DIFF
--- a/agent/stats/utils_unix.go
+++ b/agent/stats/utils_unix.go
@@ -43,7 +43,11 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 	}, nil
 }
 
+// Safe return of storage stats based on Docker Stats (Unix)
 func getStorageStats(dockerStats *types.StatsJSON) (uint64, uint64) {
+	if dockerStats.BlkioStats.IoServiceBytesRecursive == nil {
+		seelog.Debug("storage stats (Unix) not reported for container")
+	}
 	// initialize block io and loop over stats to aggregate
 	storageReadBytes := uint64(0)
 	storageWriteBytes := uint64(0)
--- a/agent/stats/utils_windows.go
+++ b/agent/stats/utils_windows.go
@@ -31,8 +31,7 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 	cpuUsage := (dockerStats.CPUStats.CPUUsage.TotalUsage * 100) / numCores
 	memoryUsage := dockerStats.MemoryStats.PrivateWorkingSet
 	networkStats := getNetworkStats(dockerStats)
-	storageReadBytes := dockerStats.StorageStats.ReadSizeBytes
-	storageWriteBytes := dockerStats.StorageStats.WriteSizeBytes
+	storageReadBytes, storageWriteBytes := getStorageStats(dockerStats)
 	return &ContainerStats{
 		cpuUsage:          cpuUsage,
 		memoryUsage:       memoryUsage,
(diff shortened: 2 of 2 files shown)


agent/stats/utils_unix.go
```diff
@@ -43,7 +43,11 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 	}, nil
 }
 
+// Safe return of storage stats based on Docker Stats (Unix)
 func getStorageStats(dockerStats *types.StatsJSON) (uint64, uint64) {
+	if dockerStats.BlkioStats.IoServiceBytesRecursive == nil {
+		seelog.Debug("storage stats (Unix) not reported for container")
+	}
```
yunhee-l: should you return here? if there's no storage stats to report, why are we proceeding to the storage stats logic below?

fierlion: The point of this change is to ensure that we have uint64(0) as our null value rather than breaking for read and write block stats.  The rest of the function exists to initialize the return values, then update them (unless they aren't available -- in which case we return 0).
This is in line with what we are doing for network stats.  See the `getNetworkStats() ` in `agent/stats/utils.go`.

fierlion:
NOTE: TestHandlerReconnectsOnConnectErrors is failing in travis CI.

agent/stats/utils_unix.go
fierlion: Updated based on https://github.com/moby/moby/blob/master/api/types/stats.go#L111-L114 and the zero-initialization of the StorageStats struct.  We are guaranteed zero uint64 values if the stats object exists.  The windows StorageStats struct is initialized here:
https://github.com/moby/moby/blob/master/daemon/daemon_windows.go#L579-L582

agent/stats/utils_unix.go
```diff
@@ -43,7 +43,11 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 	}, nil
 }
 
+// Safe return of storage stats based on Docker Stats (Unix)
 func getStorageStats(dockerStats *types.StatsJSON) (uint64, uint64) {
+	if dockerStats.BlkioStats.IoServiceBytesRecursive == nil {
```
sharanyad: I believe we should also do the following check: 
`if dockerStats.BlkioStats == nil`

agent/stats/utils_windows.go
```diff
@@ -42,3 +41,18 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 		networkStats:      networkStats,
 	}, nil
 }
+
+// Safe return of storage stats based on Docker Stats (Windows)
+func getStorageStats(dockerStats *types.StatsJSON) (uint64, uint64) {
+	if dockerStats == nil {
+		return uint64(0), uint64(0)
+	}
+	if dockerStats.StorageStats.ReadSizeBytes == 0 && dockerStats.StorageStats.WriteSizeBytes == 0 {
```
sharanyad: we should also check 
`if dockerStats.StorageStats == nil`

agent/stats/utils_unix.go
fierlion: I'll add that.

agent/stats/utils_windows.go
fierlion: see the above links to moby's implementation of the windows daemon.  It looks like they initialize all the structs or return nil for dockerStats.  So it's impossible to have a dockerstats without a StorageStats.

PR #2023, opened 2019-05-07
non-ECS image cleanup: cleanup dangling images

### LINKED ISSUE #2017: Agent errors and does not clean-up 'dangling' images
### Summary

The Agent is not able to cleanup 'dangling' images (`docker images -f dangling=true`). These are image IDs that do not have any tags. They will show up in `docker images` with name `<none>:<none>`

The issue is that the image manager lists all images by name, then attempts to inspect each image based on the name. Dangling images with name `none` throw an error on image inspection and will never get cleaned up.

Solution could be to list images by ID, then determine exclusions after inspecting the image.

### Supporting Log Snippets

When a customer has a dangling image, they will see error messages like this in the logs: 

```
2019-05-03T17:50:11Z [ERROR] Error inspecting non-ECS image name: <none>:<none> - Error response from daemon: no such image: <none>:<none>: invalid reference format
```

sparrc:
Use ImageID to cleanup images rather than tag name, to guarantee that we
    are passing a legitimate identifier to InspectImage and RemoveImage.

closes #2017

### Licensing

### DIFF
--- a/agent/engine/docker_image_manager.go
+++ b/agent/engine/docker_image_manager.go
@@ -388,75 +388,106 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
 	return nonECSContainersIDs, nil
 }
 
-type ImageWithSize struct {
-	ImageName string
-	Size      int64
+type ImageWithSizeID struct {
+	RepoTags []string
+	ImageID  string
+	Size     int64
 }
 
 func (imageManager *dockerImageManager) removeNonECSImages(ctx context.Context, nonECSImagesNumToDelete int) {
 	if nonECSImagesNumToDelete == 0 {
 		return
 	}
-	var nonECSImageNames = imageManager.getNonECSImageNames(ctx)
-	var nonECSImageNamesRemoveEligible []string
-	for _, nonECSImage := range nonECSImageNames {
-		if !isInExclusionList(nonECSImage, imageManager.imageCleanupExclusionList) {
-			nonECSImageNamesRemoveEligible = append(nonECSImageNamesRemoveEligible, nonECSImage)
-		}
-	}
+	nonECSImages := imageManager.getNonECSImages(ctx)
 
-	var imageWithSizeList []ImageWithSize
-	for _, imageName := range nonECSImageNamesRemoveEligible {
-		resp, iiErr := imageManager.client.InspectImage(imageName)
-		if iiErr != nil {
-			seelog.Errorf("Error inspecting non-ECS image name: %s - %v", imageName, iiErr)
-			continue
-		}
-		imageWithSizeList = append(imageWithSizeList, ImageWithSize{imageName, resp.Size})
-	}
 	// we want to sort images with size ascending
-	sort.Slice(imageWithSizeList, func(i, j int) bool {
-		return imageWithSizeList[i].Size < imageWithSizeList[j].Size
+	sort.Slice(nonECSImages, func(i, j int) bool {
+		return nonECSImages[i].Size < nonECSImages[j].Size
 	})
 
 	// we will remove the remaining nonECSImages in each performPeriodicImageCleanup call()
 	var numImagesAlreadyDeleted = 0
-	for _, kv := range imageWithSizeList {
-		if numImagesAlreadyDeleted == nonECSImagesNumToDelete {
+	for _, image := range nonECSImages {
+		if numImagesAlreadyDeleted >= nonECSImagesNumToDelete {
 			break
 		}
-		seelog.Infof("Removing non-ECS Image: %s", kv.ImageName)
-		err := imageManager.client.RemoveImage(ctx, kv.ImageName, dockerclient.RemoveImageTimeout)
-		if err != nil {
-			seelog.Errorf("Error removing Image %s - %v", kv.ImageName, err)
-			continue
+		if len(image.RepoTags) > 1 {
+			seelog.Debugf("Non-ECS image has more than one tag Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
+			for _, tag := range image.RepoTags {
+				err := imageManager.client.RemoveImage(ctx, tag, dockerclient.RemoveImageTimeout)
+				if err != nil {
+					seelog.Errorf("Error removing RepoTag (ImageID: %s, Tag: %s) %v", image.ImageID, tag, err)
+				} else {
+					seelog.Infof("Image Tag Removed: %s (ImageID: %s)", tag, image.ImageID)
+					numImagesAlreadyDeleted++
+				}
+			}
 		} else {
-			seelog.Infof("Image removed: %s", kv.ImageName)
-			numImagesAlreadyDeleted++
+			seelog.Debugf("Removing non-ECS Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
+			err := imageManager.client.RemoveImage(ctx, image.ImageID, dockerclient.RemoveImageTimeout)
+			if err != nil {
+				seelog.Errorf("Error removing Image %s (Tags: %s) - %v", image.ImageID, image.RepoTags, err)
+			} else {
+				seelog.Infof("Image removed: %s (Tags: %s)", image.ImageID, image.RepoTags)
+				numImagesAlreadyDeleted++
+			}
 		}
 	}
 }
 
-func (imageManager *dockerImageManager) getNonECSImageNames(ctx context.Context) []string {
-	response := imageManager.client.ListImages(ctx, dockerclient.ListImagesTimeout)
-	var allImagesNames []string
-	for _, name := range response.RepoTags {
-		allImagesNames = append(allImagesNames, name)
-	}
-	var ecsImageNames []string
-	for _, imageState := range imageManager.getAllImageStates() {
-		for _, imageName := range imageState.Image.Names {
-			ecsImageNames = append(ecsImageNames, imageName)
+// getNonECSImages returns type ImageWithSizeID with all fields populated.
+func (imageManager *dockerImageManager) getNonECSImages(ctx context.Context) []ImageWithSizeID {
+	r := imageManager.client.ListImages(ctx, dockerclient.ListImagesTimeout)
+	var allImages []ImageWithSizeID
+	// inspect all images
+	for _, imageID := range r.ImageIDs {
+		resp, err := imageManager.client.InspectImage(imageID)
+		if err != nil {
+			seelog.Errorf("Error inspecting non-ECS image: (ImageID: %s), %s", imageID, err)
+			continue
 		}
+		allImages = append(allImages,
+			ImageWithSizeID{
+				ImageID:  imageID,
+				Size:     resp.Size,
+				RepoTags: resp.RepoTags,
+			})
 	}
 
-	var nonECSImageNames = exclude(allImagesNames, ecsImageNames)
-	return nonECSImageNames
+	// get all 'ecs' image IDs
+	var ecsImageIDs []string
+	for _, imageState := range imageManager.getAllImageStates() {
+		ecsImageIDs = append(ecsImageIDs, imageState.Image.ImageID)
+	}
+
+	// exclude 'ecs' image IDs and image IDs with an explicitly excluded tag
+	var nonECSImages []ImageWithSizeID
+	for _, image := range allImages {
+		// check image ID is not excluded
+		if isInExclusionList(image.ImageID, ecsImageIDs) {
+			continue
+		}
+		// check image TAG(s) is not excluded
+		if !anyIsInExclusionList(image.RepoTags, imageManager.imageCleanupExclusionList) {
+			nonECSImages = append(nonECSImages, image)
+		}
+	}
+	return nonECSImages
 }
 
 func isInExclusionList(imageName string, imageExclusionList []string) bool {
 	for _, exclusionName := range imageExclusionList {
 		if imageName == exclusionName {
+			return true
+		}
+	}
+	return false
+}
+
+// anyIsInExclusionList returns true if any name is in the exclusion list.
+func anyIsInExclusionList(imageNames []string, nameExclusionList []string) bool {
+	for _, name := range imageNames {
+		if isInExclusionList(name, nameExclusionList) {
 			return true
 		}
 	}
@@ -486,8 +517,9 @@ func (imageManager *dockerImageManager) imagesConsiderForDeletion(allImageStates
 	for _, imageState := range allImageStates {
 		if imageManager.isExcludedFromCleanup(imageState) {
 			//imageState that we want to keep
-			seelog.Infof("Image excluded from deletion: [%s]", imageState.String())
+			seelog.Debugf("Image excluded from deletion: [%s]", imageState.String())
 		} else {
+			seelog.Debugf("Image going to be considered for deletion: [%s]", imageState.String())
 			imagesConsiderForDeletionMap[imageState.Image.ImageID] = imageState
 		}
 	}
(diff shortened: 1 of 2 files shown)


agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,78 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
 }
 
 func (imageManager *dockerImageManager) removeNonECSImages(ctx context.Context, nonECSImagesNumToDelete int) {
 	if nonECSImagesNumToDelete == 0 {
 		return
 	}
-	var nonECSImageNames = imageManager.getNonECSImageNames(ctx)
-	var nonECSImageNamesRemoveEligible []string
-	for _, nonECSImage := range nonECSImageNames {
-		if !isInExclusionList(nonECSImage, imageManager.imageCleanupExclusionList) {
```
sparrc: all exclusion lists were centralized in getNonECSImages below

agent/engine/docker_image_manager_test.go
```diff
@@ -1079,8 +1200,8 @@ func TestNonECSImageAndContainers_RemoveDeadContainer(t *testing.T) {
 	client.EXPECT().InspectContainer(gomock.Any(), gomock.Any(), dockerclient.InspectContainerTimeout).Return(inspectContainerResponse, nil).AnyTimes()
 	client.EXPECT().RemoveContainer(gomock.Any(), gomock.Any(), dockerclient.RemoveContainerTimeout).Return(nil).Times(1)
 	client.EXPECT().ListImages(gomock.Any(), dockerclient.ListImagesTimeout).Return(listImagesResponse).AnyTimes()
-	client.EXPECT().InspectImage(listImagesResponse.RepoTags[0]).Return(inspectImageResponse, nil).Times(1)
-	client.EXPECT().RemoveImage(gomock.Any(), listImagesResponse.RepoTags[0], dockerclient.RemoveImageTimeout).Return(nil).Times(1)
+	client.EXPECT().InspectImage(listImagesResponse.ImageIDs[0]).Return(inspectImageResponse, nil).Times(1)
```
sparrc: we are now deleting images on their ID rather than their name (tag)

adnxn:
oh also this `This branch is out-of-date with the base branch`

agent/engine/docker_image_manager_test.go
```diff
@@ -1035,7 +1035,128 @@ func TestNonECSImageAndContainersCleanupRemoveImage(t *testing.T) {
...
+	client.EXPECT().RemoveImage(gomock.Any(), listImagesResponse.ImageIDs[0], dockerclient.RemoveImageTimeout).Return(nil).Times(0)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	imageManager.removeUnusedImages(ctx)
+
+	assert.Len(t, listContainersResponse.DockerIDs, 1, "error removing container IDs")
+	assert.Len(t, inspectContainerResponse.ID, 1, "error inspecting containers ids")
```
shubham2892: NIT: upper case for "error".

agent/engine/docker_image_manager_test.go
```diff
@@ -1035,7 +1035,128 @@ func TestNonECSImageAndContainersCleanupRemoveImage(t *testing.T) {
...
+	assert.Len(t, imageManager.imageStates, 1, "there should still be an image state")
```
shubham2892: NIT: upper case for "error".

agent/engine/docker_image_manager_test.go
```diff
@@ -1092,7 +1213,7 @@ func TestNonECSImageAndContainers_RemoveDeadContainer(t *testing.T) {
 	assert.Len(t, imageManager.imageStates, 0, "Error removing image state after the image is removed")
 }
 
-// Old 'Created' Images should be cleaned up
+// Old 'Created' cotainers should be cleaned up
```
shubham2892: NIT: typo in "cotainers'

sharanyad:
just a general question: 
is the image ID used to clean up dangling images or all ECS/other non ECS images?

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,82 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
+		ecsImageIDs = append(ecsImageIDs, imageState.Image.ImageID)
+	}
+
+	// exclude 'ecs' image IDs and image IDs with an explicitly excluded tag
+	var nonECSImages []ImageWithSizeID
+	for _, image := range allImages {
+		if !isInExclusionList(image.ImageID, ecsImageIDs) {
+			var exclude bool
+			for _, repoTag := range image.RepoTags {
+				if isInExclusionList(repoTag, imageManager.imageCleanupExclusionList) {
```
sharanyad: this can be refactored as - 
```
if !isInExclusionList(repoTag, imageManager.imageCleanupExclusionList) {
  nonECSImages = append(nonECSImages, image)
}
```
avoids the use of local var

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,82 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
 	var numImagesAlreadyDeleted = 0
-	for _, kv := range imageWithSizeList {
+	for _, image := range nonECSImages {
 		if numImagesAlreadyDeleted == nonECSImagesNumToDelete {
 			break
 		}
-		seelog.Infof("Removing non-ECS Image: %s", kv.ImageName)
-		err := imageManager.client.RemoveImage(ctx, kv.ImageName, dockerclient.RemoveImageTimeout)
+		seelog.Debugf("Removing non-ECS Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
+		err := imageManager.client.RemoveImage(ctx, image.ImageID, dockerclient.RemoveImageTimeout)
```
sharanyad: has the case been tested where there can be two images with same image ID, but different names? 
Does this remove both the images?

agent/engine/docker_image_manager.go
sparrc: The above if-statement would add the image ID to "nonECSImages" if _any_ tag is not in the exclusion list. The local variable is necessary because we want to skip Image IDs that have any tag in the exclusion list.

ie, an ImageID with 1 out of 10 tags in the exclusion list should be excluded.

agent/engine/docker_image_manager.go
sparrc: They should both be removed, though I haven't tested this case explicitly. I can add a unit test for this.

sparrc:
> just a general question:
> is the image ID used to clean up dangling images or all ECS/other non ECS images?

With this change, the ImageID will always be used for both cleanup and inspection of images.

agent/engine/docker_image_manager.go
sharanyad: please tell manually once. on a quick testing, it does not remove the image
```
$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
ubuntu              latest              d131e0fa2585        2 weeks ago         102MB
ubuntu              mirror              d131e0fa2585        2 weeks ago         102MB
$ docker rmi d131e0fa2585
Error response from daemon: conflict: unable to delete d131e0fa2585 (must be forced) - image is referenced in multiple repositories
$ docker rmi ubuntu:mirror
Untagged: ubuntu:mirror
$ docker rmi d131e0fa2585
Untagged: ubuntu:latest
Untagged: ubuntu@sha256:70fc21e832af32eeec9b0161a805c08f6dddf64d341748379de9a527c01b6ca1
Deleted: sha256:d131e0fa2585a7efbfb187f70d648aa50e251d9d3b7031edf4730ca6154e221e
Deleted: sha256:c59a62c2bba8db73ca8b8847baa7fa77e4d573f38e57d69592a9716f9fa075c4
Deleted: sha256:a4eb6208f601c2f32e043972b9fa8e813767aef2e60676e91796cabf8a0afdc4
Deleted: sha256:a14c708b62677e8acfb75ac873147e71dd26aa7bf75a8b63b3408e0826b174dc
Deleted: sha256:604cbde1a4c8fee1b102f8b64d4f41e62d770b5f8a6b06fb809cfd873a2643c3
```

sparrc: You're right, I've added a change now to detect if an ImageID has multiple tags. If there are multiple tags on an ImageID, the tags will be removed first before removing the image.

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,87 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
-	var nonECSImageNames = exclude(allImagesNames, ecsImageNames)
-	return nonECSImageNames
+	// exclude 'ecs' image IDs and image IDs with an explicitly excluded tag
+	var nonECSImages []ImageWithSizeID
+	for _, image := range allImages {
+		// check image ID is not excluded
+		if !isInExclusionList(image.ImageID, ecsImageIDs) {
+			// check image TAG(s) is not excluded
+			if !anyIsInExclusionList(image.RepoTags, imageManager.imageCleanupExclusionList) {
+				nonECSImages = append(nonECSImages, image)
```
yumex93: We can get rid of this nested if statement by using:
if isInExclustionList(){
   continue
}
if !anyIsInExclusionList(){
}
The nested if statement plus negative condition makes a little bit hard to understand what we are trying to do here.  However, this is not a blocker for you to merge the change.

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,87 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
+			seelog.Debugf("Non-ECS image has more than one tag Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
+			for _, tag := range image.RepoTags {
+				err := imageManager.client.RemoveImage(ctx, tag, dockerclient.RemoveImageTimeout)
+				if err != nil {
+					seelog.Errorf("Error removing RepoTag (ImageID: %s, Tag: %s) %v", image.ImageID, tag, err)
+				}
+			}
+		}
+		seelog.Debugf("Removing non-ECS Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
+		err := imageManager.client.RemoveImage(ctx, image.ImageID, dockerclient.RemoveImageTimeout)
```
sharanyad: The image layers are all removed when the last tag/reference is removed. So this statement needs to be inside an else. (else image deletion will fail and falsely log as image cannot be removed)

agent/engine/docker_image_manager.go
sharanyad: thanks, do we have unit tests for this case as well?

agent/engine/docker_image_manager.go
sparrc: will do

agent/engine/docker_image_manager.go
sparrc: yes, this is tested here: https://github.com/aws/amazon-ecs-agent/pull/2023/files#diff-47eca467a5b10ea4396074d535ccea4cR1038

agent/engine/docker_image_manager.go
sparrc: fixed

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,89 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
-			seelog.Infof("Image removed: %s", kv.ImageName)
-			numImagesAlreadyDeleted++
+			seelog.Debugf("Removing non-ECS Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
+			err := imageManager.client.RemoveImage(ctx, image.ImageID, dockerclient.RemoveImageTimeout)
+			if err != nil {
+				seelog.Errorf("Error removing Image %s (Tags: %s) - %v", image.ImageID, image.RepoTags, err)
+				continue
+			} else {
+				seelog.Infof("Image removed: %s (Tags: %s)", image.ImageID, image.RepoTags)
+				numImagesAlreadyDeleted++
```
sharanyad: just noticed, we should increment `numImagesAlreadyDeleted` while removing by tags as well..

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,89 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
 		if numImagesAlreadyDeleted == nonECSImagesNumToDelete {
 			break
 		}
-		seelog.Infof("Removing non-ECS Image: %s", kv.ImageName)
-		err := imageManager.client.RemoveImage(ctx, kv.ImageName, dockerclient.RemoveImageTimeout)
-		if err != nil {
-			seelog.Errorf("Error removing Image %s - %v", kv.ImageName, err)
-			continue
+		if len(image.RepoTags) > 1 {
+			seelog.Debugf("Non-ECS image has more than one tag Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
```
sharanyad: this should probably be `(Tags: %v)` , since repoTags is a list

sparrc: `%s` will print the list the same as `%v`: https://play.golang.org/p/tTLHNpF3DrG

agent/engine/docker_image_manager.go
sparrc: good catch, will do

sparrc:
> Could you please edit the commit message as well to mention this change is only for non-ECS images?

yes, done 👍

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,90 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
+			for _, tag := range image.RepoTags {
+				err := imageManager.client.RemoveImage(ctx, tag, dockerclient.RemoveImageTimeout)
+				if err != nil {
+					seelog.Errorf("Error removing RepoTag (ImageID: %s, Tag: %s) %v", image.ImageID, tag, err)
+				}
+				numImagesAlreadyDeleted++
```
sharanyad: this should be inside else or we can add `continue` in the above if

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,90 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
-		return imageWithSizeList[i].Size < imageWithSizeList[j].Size
+	sort.Slice(nonECSImages, func(i, j int) bool {
+		return nonECSImages[i].Size < nonECSImages[j].Size
 	})
 
 	// we will remove the remaining nonECSImages in each performPeriodicImageCleanup call()
 	var numImagesAlreadyDeleted = 0
-	for _, kv := range imageWithSizeList {
+	for _, image := range nonECSImages {
 		if numImagesAlreadyDeleted == nonECSImagesNumToDelete {
```
sharanyad: if we are incrementing `numImagesAlreadyDeleted` for each repo tag of the image in a for loop, we should be checking if `numImagesAlreadyDeleted >= nonECSImagesNumToDelete` here

sparrc: done

agent/engine/docker_image_manager.go
sparrc: true, I have put it in an `else`

agent/engine/docker_image_manager.go
```diff
@@ -388,70 +388,90 @@ func (imageManager *dockerImageManager) getNonECSContainerIDs(ctx context.Contex
...
-			seelog.Errorf("Error removing Image %s - %v", kv.ImageName, err)
-			continue
+		if len(image.RepoTags) > 1 {
+			seelog.Debugf("Non-ECS image has more than one tag Image: %s (Tags: %s)", image.ImageID, image.RepoTags)
+			for _, tag := range image.RepoTags {
+				err := imageManager.client.RemoveImage(ctx, tag, dockerclient.RemoveImageTimeout)
+				if err != nil {
+					seelog.Errorf("Error removing RepoTag (ImageID: %s, Tag: %s) %v", image.ImageID, tag, err)
+				} else {
+					numImagesAlreadyDeleted++
```
sharanyad: nit: we can add a log statement here that image with the repo tag was removed

sparrc: done

PR #2024, opened 2019-05-07
add nil check for block stats

fierlion:
### Summary
Adds a nil check for `BlkioStats.IoServiceBytesRecursive`.  All other stats are zero initialized so we are safe with the implementation as-is.

### Testing
manually on linux and windows instances

### Licensing


agent/stats/utils_unix.go
```diff
@@ -45,6 +45,10 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 
 func getStorageStats(dockerStats *types.StatsJSON) (uint64, uint64) {
 	// initialize block io and loop over stats to aggregate
+	if dockerStats.BlkioStats.IoServiceBytesRecursive == nil {
+		seelog.Debug("IoServiceBytes nil for container")
```
sharanyad: nit: I liked `storage stats not reported for container`:)

PR #2041, opened 2019-05-17
empty network stats for host and awsvpc mode

sharanyad:
### Summary
Currently, Docker does not return Network stats for ECS network
modes - awsvpc, host. Send empty network metrics to TCS for
those network modes. Earlier, we were sending zero value network stats.

### Testing

### Licensing

### DIFF
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -42,6 +42,8 @@ import (
 const (
 	containerChangeHandler = "DockerStatsEngineDockerEventsHandler"
 	queueResetThreshold    = 2 * dockerclient.StatsInactivityTimeout
+	hostNetworkMode        = "host"
+	noneNetworkMode        = "none"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


agent/stats/engine_unix_test.go
```diff
@@ -0,0 +1,99 @@
+//+build unit
```
fenxiong: why is the test unix specific? the "default" mode is for windows right

agent/stats/engine.go
```diff
@@ -604,26 +604,34 @@ func (engine *DockerStatsEngine) taskContainerMetricsUnsafe(taskArn string) ([]*
...
+		}
+
+		task, err := engine.resolver.ResolveTask(dockerID)
 		if err != nil {
-			// we log the error and still continue to publish cpu, memory stats
-			seelog.Warnf("Error getting network stats: %v, container: %v", err, dockerID)
+			seelog.Warnf("Task not found for container ID: %s", dockerID)
+		} else {
+			// send network stats for default/bridge/nat network modes
+			if task.ENI == nil && container.containerMetadata.NetworkMode != "host" {
```
fenxiong: just wondering, do you need to deal with linux's "none" network mode?

agent/stats/engine_unix_test.go
sharanyad: default is for both linux, windows. but good catch, will add this mode to non os specific file.

agent/stats/engine.go
sharanyad: yes, will add it

agent/stats/engine.go
```diff
@@ -604,26 +606,34 @@ func (engine *DockerStatsEngine) taskContainerMetricsUnsafe(taskArn string) ([]*
...
+			MemoryStatsSet: memoryStatsSet,
+		}
+
+		task, err := engine.resolver.ResolveTask(dockerID)
 		if err != nil {
-			// we log the error and still continue to publish cpu, memory stats
-			seelog.Warnf("Error getting network stats: %v, container: %v", err, dockerID)
+			seelog.Warnf("Task not found for container ID: %s", dockerID)
+		} else {
+			// send network stats for default/bridge/nat network modes
```
fierlion: Do we need to log when the customer is using `host`/`none` network modes, (if it's not in the logs already)?  We might need this to troubleshoot logs out of context.

sharanyad: We already log as debug here - https://github.com/aws/amazon-ecs-agent/blob/network-disk-metrics/agent/stats/utils.go#L57 
Adding another log statement may make the logs too messy since these will be logged once every 20 seconds.

sharanyad:
non-os specific test has been added already - https://github.com/aws/amazon-ecs-agent/pull/2041/files#diff-f800031aeb5d09258c17bc856f46f8ecR448

sharanyad:
```
=== RUN   TestPrivateRegistryAuthOverASM
--- FAIL: TestPrivateRegistryAuthOverASM (11.03s)
	functionaltests_unix_test.go:1123: AWS Secrets Manager resource already exists
	utils_unix.go:131: Created directory /tmp/tmp.YoHTbu3uJL/ecs_integ_testdata646515998 to store test data in
	utils_unix.go:153: Launching agent with image: amazon/amazon-ecs-agent:latest
	utils_unix.go:244: Agent started as docker container: 490cf82fdff2fe3f5bbeb10c8432310e44ae7235335d7b8b927c78b982ac919f
	utils_unix.go:135: Could not get agent metadata after launching it
```

sharanyad:
all tests passed, merging.

PR #2049, opened 2019-05-23
updated with block stats check

fierlion:
### Summary
Add integ test for StorageStatsSet aka BlockIO stats

### Implementation details
Follows pattern established to check that NetworkStats are present.
Gremlin container creates noisy stats; all stats are expected to be present.  

### Testing
Tested manually on ec2 linux.

New tests cover the changes: 

### Licensing


agent/stats/engine_integ_test.go
```diff
@@ -659,3 +659,79 @@ func testNetworkModeStats(t *testing.T, networkMode string, statsEmpty bool) {
 	validateIdleContainerMetrics(t, engine)
 	validateEmptyTaskHealthMetrics(t, engine)
 }
+
+func TestBlockStats(t *testing.T) {
```
sharanyad: how about renaming it to StorageStats and in the event stream below?

fierlion: I'll do that.

PR #2054, opened 2019-05-29
Support reserving CPU units for the system

patbos:
### Summary
Support reserving CPU for system

### Implementation details

Will remove CPU units reported to ECS 

### Testing

Tested in our environment

New tests cover the changes: 

### Licensing


samuelkarp:
Hi @patbos! Can you explain a bit more about your desire to reserve CPU shares?  There's previous discussion and explanation in https://github.com/aws/amazon-ecs-agent/issues/369#issuecomment-209108757.  The only relevant difference from the comment at that time is that ECS does now support hard CPU quotas through the [task size](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#task_size).

patbos:
Hi @samuelkarp ,

Setting a hard CPU limit in the task definition is exactly what we are preaching to our development teams. Unfortunately not everyone are doing that (yet) and we have not (yet) started to enforce it. Thing is we want to be (over)protective and make sure that the ECS Agent is running no matter what the services are doing. There is also the case of not optimal scaling algorithms together with CPU over-allocation that might also cause this issue during deploy/startup.

adnxn:
> Setting a hard CPU limit in the task definition is exactly what we are preaching to our development teams.

we recommend keeping on this path since we have no immediate plans to support reserving cpu shares for the agent. however you could open an ECS feature request on the [aws/containers-roadmap](https://github.com/aws/containers-roadmap) to continue this discussion.

closing this pr.

PR #2059, opened 2019-05-31
Using correct error to construct CannotRemoveVolumeError

yunhee-l:
### Summary
We are constructing CannotRemoveVolumeError with wrong docker error, which can result in nullpointer

### Implementation details
Using the correct docker error object

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -945,8 +945,8 @@ func (dg *dockerGoClient) handleContainerEvents(ctx context.Context,
 		metadata := dg.containerMetadata(ctx, containerID)
 
 		changedContainers <- DockerContainerChangeEvent{
-			Status:                  status,
-			Type:                    eventType,
+			Status: status,
+			Type:   eventType,
 			DockerContainerMetadata: metadata,
 		}
 	}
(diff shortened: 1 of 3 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1209,9 +1209,9 @@ func (dg *dockerGoClient) removeVolume(ctx context.Context, name string) error {
 		return &CannotGetDockerClientError{version: dg.version, err: err}
 	}
 
-	ok := client.VolumeRemove(ctx, name, false)
-	if ok != nil {
-		return &CannotRemoveVolumeError{err}
+	removeErr := client.VolumeRemove(ctx, name, false)
```
sharanyad: nit: this can just `err`

yunhee-l:
> Could you also add an assert here for `err.Error()` here https://github.com/aws/amazon-ecs-agent/blob/master/agent/dockerclient/dockerapi/docker_client_test.go#L1626?

This is a unit test, so yes we can add it, but it won't actually be testing other than our unit test setup?

agent/dockerclient/dockerapi/docker_client.go
yunhee-l: i wanted to make it very distinctively separate from the err above this, but I don't feel strongly about it, I'll update it to use err for consistency.

PR #2081, opened 2019-06-12
Adding a goimports check

petderek:
### Summary

Adds the following targets to make:

* `importcheck` - used to validate that import order is preserved
* `goimports` - runs 'goimports -w' on all relevant files
* `gofmt` - runs 'go fmt' on all relevant files

### Implementation details

* `importcheck` is added as a dependency to static-check, so this will get invoked on every travis run.
* Removed lines like `grep -v vendor` -- go list, test, etc now automatically ignore the vendor directory.
* made `fmtcheck` and `importcheck` also apply to test / mock files.

### Testing

`make static-check` on Ubuntu 

### Licensing


yunhee-l:
ARM unit test failed on TestHandlerReconnectsOnDiscoverPollEndpointError, which is a known flakey test (https://github.com/aws/amazon-ecs-agent/pull/1986). Retrying.

PR #2082, opened 2019-06-12
Simplify mock generation logic

petderek:
### Summary

This gets rid of a small hack we used to inject the copyright headers into gomock
generated code.
### Implementation details

Mockgen now supports a flag for adding a copyright header. We can now get rid
of the custom script we had in place to add copyright headers.

The previous call chain...

makefile -> ./scripts/gogenerate -> go:generate directives -> mockgen.go -> mockgen

...is now simplified to:

makefile -> go:generate directives -> mockgen

This also obviates the need for the shared_env file, which was only used to put
./scripts/generate in the PATH. This file has been removed and its consumers
have been updated.
### Testing

ran `make gogenerate` to verify changes.

### Licensing


adnxn:
"neet"

yunhee-l:
can you clean up the commits please? maybe the last two can be squashed.

PR #2084, opened 2019-06-13
Update task cleanup wait logic to clean task resources immediately in…

mssrivas:
Update task cleanup wait logic to clean task resources immediately instead of waiting 3 hours

### Summary
Currently if the clean up duration for tasks are not defined, It defaults to 3hr. The current logic in cleanupTask is once it passed the cleanup time, agent will wait for 3 more hours (DefaultTaskCleanupWaitDuration) instead of cleaning up the task immediately. Instead with below changes, it will begin cleanup immediately .

### Implementation details
Once it passed the cleanup time, it waits a nanosecond and begins the cleanup task.

### Testing
Created a task definition that uses a volume called test_vol. The check for successful clean being that docker volume ls does not return test_vol in the list.
Five cases tested.
1. No cleanup is specified - The cleanup duration is set to 3hrs
2. Cleanup specified is negative - The cleanup is set to 3hrs
3. Cleanup specified is less than minimum value - The cleanup is set to 3hrs
4. Cleanup specified is 10 mins - Clean up begins after 5min after container goes to stop state 
5. Cleanup specified is 5mins - Clean up begins after 5min after container goes to stop state

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -1092,18 +1092,16 @@ func (mtask *managedTask) time() ttime.Time {
 
 func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 	cleanupTimeDuration := mtask.GetKnownStatusTime().Add(taskStoppedDuration).Sub(ttime.Now())
-	// There is a potential deadlock here if cleanupTime is negative. Ignore the computed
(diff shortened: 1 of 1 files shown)


ubhattacharjya:
Can you describe how did you verify that the task resources got cleaned up in the testing done section?

PR #2036, opened 2019-05-10
Update libcni to the latest version

richardpen:
### Summary

Update the containernetworking/cni to the latest version that take a context in the API.

### Implementation details

Update the dependency package and the API in ecscni package.

### Testing

New tests cover the changes: 

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2085, opened 2019-06-14
Fill in Type field when invoking ipam DEL command

fenxiong:
### Summary

After updating libcni to 0.7.0 (https://github.com/aws/amazon-ecs-agent/pull/2036), agent fails to invoke IPAM DEL command when cleaning up awsvpc task:
```
2019-06-14T17:17:30Z [WARN] Managed task [...]: failed to release ip; IPAM error: error parsing configuration: missing 'type'
```
This results in IPAM record not being deleted on the instance, which can ultimately lead to exhausting ip address, and when that happens, subsequent awsvpc task will fail to launch:
```
...error transitioning container [~internal~ecs~pause] to [RESOURCES_PROVISIONED]: container resource provisioning: failed to setup network namespace: cni setup: invoke bridge plugin failed: bridge ipam ADD: failed to execute plugin: ecs-ipam: getIPV4AddressFromDB commands: failed to get available ip from the db: getAvailableIP ipstore: failed to find available ip addresses in the subnet
```

It appears that starting from libcni 0.7.0, the package mandates the json blob in [NetworkConfig.Bytes](https://github.com/containernetworking/cni/blob/master/libcni/api.go#L57) to have a non-empty Type field, and we are not filling in this field when invoking IPAM DEL command, so the invocation fails. This PR fill in the field to conform with what the package expects.

### Implementation details

Fill in the Type field with plugin name.

### Testing

Added unit test. Manually tested that the ip record is deleted in /var/lib/ecs/data/eni-ipam.db with the fix. I checked whether i can add a functional test for this, but seems like the boltdb file written by the ipam plugin can't be opened by non-root user which makes this tricky.

Bug - Fixed a bug where IPAM record was not deleted when cleaning up awsvpc task.

### Licensing

### DIFF
--- a/agent/ecscni/plugin.go
+++ b/agent/ecscni/plugin.go
@@ -404,6 +404,7 @@ func (client *cniClient) createIPAMNetworkConfig(cfg *Config) (string, *libcni.N
 
 	ipamNetworkConfig := IPAMNetworkConfig{
 		Name:       ECSIPAMPluginName,
+		Type:       ECSIPAMPluginName,
 		CNIVersion: client.cniVersion,
 		IPAM:       ipamConfig,
 	}
--- a/agent/ecscni/plugin_test.go
+++ b/agent/ecscni/plugin_test.go
@@ -389,6 +389,36 @@ func TestConstructAppMeshNetworkConfig(t *testing.T) {
 	assert.Equal(t, config.EgressIgnoredIPs[0], appMeshConfig.EgressIgnoredIPs[0])
 }
 
+func TestConstructIPAMNetworkConfig(t *testing.T) {
+	ecscniClient := NewClient(&Config{})
+
+	config := &Config{
+		ID:                       "02:7b:64:49:b1:40",
+		ENIID:                    "eni-12345678",
+		ContainerID:              "containerid12",
+		ContainerPID:             "pid",
+		ENIIPV4Address:           "172.31.21.40",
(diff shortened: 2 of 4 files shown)


agent/ecscni/plugin.go
```diff
@@ -404,6 +404,7 @@ func (client *cniClient) createIPAMNetworkConfig(cfg *Config) (string, *libcni.N
 
 	ipamNetworkConfig := IPAMNetworkConfig{
 		Name:       ECSIPAMPluginName,
+		Type:       ECSIPAMPluginName,
```
yunhee-l: is type a free string too?

fenxiong: what do you mean by "free string"?

fenxiong: this field is used to find the plugin binary (as the binary's name)

yumex93:
Just wanna know are we able to have functional tests to catch such kind of errors before releasing the change? Otherwise, this thing may happen again in the feature.

fenxiong:
> Just wanna know are we able to have functional tests to catch such kind of errors before releasing the change? Otherwise, this thing may happen again in the feature.

"I checked whether i can add a functional test for this, but seems like the boltdb file written by the ipam plugin can't be opened by non-root user which makes this tricky."

so i'm not sure whether there's a way to check this without having to run the test as root.

PR #2095, opened 2019-06-24
ACS/Agent Model changes for log router

mssrivas:
### Summary
Changing the ACCS model and adding the generated model from it for log router

### Implementation details
Changes were made following - ECS-Agent-model-update runbook

### Testing
make test
make run-integ-tests

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/acs/handler/payload_handler_test.go
+++ b/agent/acs/handler/payload_handler_test.go
@@ -907,3 +907,50 @@ func TestHandleUnrecognizedTask(t *testing.T) {
 	tester.payloadHandler.handleUnrecognizedTask(ecsacsTask, errors.New("test error"), payloadMessage)
 	wait.Wait()
 }
+
+func TestPayloadHandlerAddedLogRouterData(t *testing.T) {
+	tester := setup(t)
+	defer tester.ctrl.Finish()
+
+	var addedTask *apitask.Task
+	tester.mockTaskEngine.EXPECT().AddTask(gomock.Any()).Do(
+		func(task *apitask.Task) {
+			addedTask = task
+		})
+
+	payloadMessage := &ecsacs.PayloadMessage{
+		Tasks: []*ecsacs.Task{
+			{
+				Arn: aws.String("arn"),
+				Containers: []*ecsacs.Container{
+					{
+						LogRouter: &ecsacs.LogRouter{
+							Config: &ecsacs.LogRouterConfig{
+								Type:  aws.String("arn"),
+								Value: aws.String("s3-arn"),
+							},
+							EnableECSLogMetaData: aws.Bool(true),
+							Type:                 aws.String("fluentd"),
+						},
+					},
+				},
+			},
+		},
+		MessageId: aws.String(payloadMessageId),
+	}
+
+	err := tester.payloadHandler.handleSingleMessage(payloadMessage)
+	assert.NoError(t, err)
+
+	// Validate the pieces of the LogRouter
+	expected := payloadMessage.Tasks[0].Containers[0].LogRouter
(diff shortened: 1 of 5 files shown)


agent/api/container/container.go
```diff
@@ -115,6 +115,8 @@ type Container struct {
 	Memory uint
 	// Links contains a list of containers to link, corresponding to docker option: --link
 	Links []string
+	//
+	LogRouter *LogRouter `json:"logRouter"`
```
yunhee-l: does this change need our state file version to be changed?

fenxiong: can you add a unit test in https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/payload_handler_test.go to make sure the new field can be populated from acs payload? also this field is missing comment here

agent/api/container/container.go
```diff
@@ -272,6 +274,19 @@ type MountPoint struct {
 	ReadOnly      bool   `json:"readOnly"`
 }
 
+// Log Router things
```
fenxiong: See https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences, comments should begin with the name of the thing being described and end in a period. Maybe "// LogRouter contains information needed by a log router container."

agent/api/container/container.go
```diff
@@ -272,6 +274,19 @@ type MountPoint struct {
...
+type LogRouter struct {
+	Config               *LogRouterConfig `json:"config"`
+	EnableECSLogMetaData bool             `json:"enableECSLogMetaData"`
+	Type                 string           `json:"type"`
+}
+
+// Log Router things
```
fenxiong: same as above. maybe something like "// LogRouterConfig specifies the config for the log router."

agent/api/container/container.go
mssrivas: Done. Changed the version to 23 and added a comment about the new change.

mssrivas: added.

agent/acs/handler/payload_handler_test.go
```diff
@@ -907,3 +907,50 @@ func TestHandleUnrecognizedTask(t *testing.T) {
...
+				Arn: aws.String("arn"),
+				Containers: []*ecsacs.Container{
+					{
+						LogRouter: &ecsacs.LogRouter {
+							Config: &ecsacs.LogRouterConfig{
+								Type: aws.String("arn"),
+								Value :  aws.String("asm-arn"),
+							},
+							EnableECSLogMetaData:aws.Bool(true),
+							Type: aws.String("fluentd"),
```
PettitWesley: nits: The spacing here is a little bit weird, and I'd do `s3-arn` (or a valid looking arn), instead of `asm-arn`.

agent/acs/model/api/api-2.json
```diff
@@ -551,6 +582,13 @@
         "asm"
       ]
     },
+    "SecretTarget":{
```
PettitWesley: (May be I'm clueless because I don't know the agent that well) Why are things related to secrets added in this?

fenxiong: i think last time this field was manually added to api.go.. so if we don't add this field to api-2.json, generating api.go from api-2.json will cause this field be deleted from api.go..

PettitWesley: Oh I see, I should've guessed this file was auto-generated

agent/acs/handler/payload_handler_test.go
mssrivas: Fixed

PR #2097, opened 2019-06-25
Implement log router task resource

fenxiong:
### Summary

Implement log router related resource as a task resource. Log router task resource models log router related resources as a task resource. Creating this resource prepares everything a log router container needs, which includes creating a config directory, a socket directory, and a config file generated under the config directory. Cleaning up this resource means removing all those directories and config file. 

This PR only covers implementing this task resource, but not using it. The resource will later be added as a dependency for a log router container in a separate PR.

### Implementation details

Followed similar pattern as other task resource. Log router specific logic is mostly in resource's Create() and Cleanup() method.

For readability purpose I separate the changes in several commits. 
Commit 1: Add fluent config generator vendor package. (https://github.com/awslabs/go-config-generator-for-fluentd-and-fluentbit), which is used to generate the config file. License will be added in a separate PR.
Commit 2: Implementation.
Commit 3: Unit tests.

### Testing

Unit test added. Manually test this by injecting [test code](https://github.com/fenxiong/amazon-ecs-agent/commit/8414ac855b9ba7de33f4bdaffc25123e366da626) in task.go and docker_task_engine.go and verified that the generated config file and socket directory can be used by the log router to send logs for both fluentd and fluentbit.

### Licensing

### DIFF
--- a/agent/taskresource/logrouter/json_unix.go
+++ b/agent/taskresource/logrouter/json_unix.go
@@ -1,0 +1,110 @@
+// +build linux
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package logrouter
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/pkg/errors"
+
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+)
+
+type logRouterResourceJSON struct {
+	Cluster                string
+	TaskARN                string
+	TaskDefinition         string
+	EC2InstanceID          string
+	ResourceDir            string
+	LogRouterType          string
+	ECSMetadataEnabled     bool
+	ContainersToLogOptions map[string]map[string]string
+	TerminalReason         string
+
+	CreatedAt     time.Time
+	DesiredStatus *LogRouterStatus
+	KnownStatus   *LogRouterStatus
+	AppliedStatus *LogRouterStatus
+}
+
+// MarshalJSON marshals logRouter object into bytes of json.
+func (logRouter *LogRouterResource) MarshalJSON() ([]byte, error) {
+	if logRouter == nil {
+		return nil, errors.New("log router resource is nil")
+	}
+
+	logRouter.lock.RLock()
+	defer logRouter.lock.RUnlock()
+
+	return json.Marshal(logRouterResourceJSON{
+		Cluster:                logRouter.cluster,
+		TaskARN:                logRouter.taskARN,
+		TaskDefinition:         logRouter.taskDefinition,
+		EC2InstanceID:          logRouter.ec2InstanceID,
+		ResourceDir:            logRouter.resourceDir,
+		LogRouterType:          logRouter.logRouterType,
+		ECSMetadataEnabled:     logRouter.ecsMetadataEnabled,
+		ContainersToLogOptions: logRouter.containerToLogOptions,
+		TerminalReason:         logRouter.terminalReason,
+		CreatedAt:              logRouter.createdAtUnsafe,
+		DesiredStatus: func() *LogRouterStatus {
+			desiredStatus := logRouter.desiredStatusUnsafe
+			s := LogRouterStatus(desiredStatus)
+			return &s
+		}(),
+		KnownStatus: func() *LogRouterStatus {
+			knownStatus := logRouter.knownStatusUnsafe
+			s := LogRouterStatus(knownStatus)
+			return &s
+		}(),
+		AppliedStatus: func() *LogRouterStatus {
+			appliedStatus := logRouter.appliedStatusUnsafe
+			s := LogRouterStatus(appliedStatus)
+			return &s
+		}(),
+	})
+}
+
+// UnmarshalJSON unmarshals bytes of json into logRouter object.
+func (logRouter *LogRouterResource) UnmarshalJSON(b []byte) error {
+	if logRouter == nil {
+		return errors.New("log router resource is nil")
+	}
+
+	temp := logRouterResourceJSON{}
+	if err := json.Unmarshal(b, &temp); err != nil {
+		return err
+	}
+
+	logRouter.lock.Lock()
+	defer logRouter.lock.Unlock()
+
+	logRouter.cluster = temp.Cluster
+	logRouter.taskARN = temp.TaskARN
+	logRouter.taskDefinition = temp.TaskDefinition
+	logRouter.ec2InstanceID = temp.EC2InstanceID
+	logRouter.resourceDir = temp.ResourceDir
+	logRouter.logRouterType = temp.LogRouterType
+	logRouter.ecsMetadataEnabled = temp.ECSMetadataEnabled
+	logRouter.containerToLogOptions = temp.ContainersToLogOptions
+	logRouter.terminalReason = temp.TerminalReason
+	logRouter.createdAtUnsafe = temp.CreatedAt
+	logRouter.desiredStatusUnsafe = resourcestatus.ResourceStatus(*temp.DesiredStatus)
+	logRouter.knownStatusUnsafe = resourcestatus.ResourceStatus(*temp.KnownStatus)
+	logRouter.appliedStatusUnsafe = resourcestatus.ResourceStatus(*temp.AppliedStatus)
+
+	return nil
+}
--- a/agent/taskresource/logrouter/logconfig_unix.go
+++ b/agent/taskresource/logrouter/logconfig_unix.go
@@ -1,0 +1,145 @@
+// +build linux
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package logrouter
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+
+	generator "github.com/awslabs/go-config-generator-for-fluentd-and-fluentbit"
+)
+
+const (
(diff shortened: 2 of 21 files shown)


agent/taskresource/logrouter/logconfig_unix.go
```diff
@@ -0,0 +1,144 @@
...
+	socketInputPathOptionFluentd = "path"
+
+	// socketInputPathOptionFluentbit is the key for specifying socket path for fluentbit.
+	socketInputPathOptionFluentbit = "unix_path"
+
+	// outputLogOptionKeyFluentd is the key for the log option that specifies output plugin name for fluentd.
+	outputLogOptionKeyFluentd = "@type"
+
+	// outputLogOptionKeyFluentbit is the key for the log option that specifies output plugin name for fluentbit.
+	outputLogOptionKeyFluentbit = "Name"
```
PettitWesley: How about `socketLogOptionKeyFluentbit` and `socketLogOptionKeyFluentd` to keep with the pattern of the above.

Or, even better:

`outputTypeLogOptionKeyFluentbit` and `outputTypeLogOptionKeyFluentbit ` to make it clear this is the type of output plugin, not just some option

agent/taskresource/logrouter/logconfig_unix.go
```diff
@@ -0,0 +1,144 @@
...
+	var inputName, inputPathOption string
+	if logRouter.logRouterType == LogRouterTypeFluentd {
+		inputName = socketInputNameFluentd
+		inputPathOption = socketInputPathOptionFluentd
+	} else {
+		inputName = socketInputNameFluentbit
+		inputPathOption = socketInputPathOptionFluentbit
+	}
+	config.AddInput(inputName, "", map[string]string{
+		inputPathOption: fmt.Sprintf(socketPathFormat, logRouter.logRouterType),
```
PettitWesley: This is fine, but the unix socket path could be the same in the container for both. That will make it easier on the host path/docker connection side of things too.

So I'd change your socketPathFormat to just a constant, with the value `/var/run/fluent.sock`.

agent/taskresource/logrouter/logconfig_unix.go
```diff
@@ -0,0 +1,144 @@
...
+			AddFieldToRecord("ecs_task_arn", logRouter.taskARN, "*").
+			AddFieldToRecord("ecs_task_definition", logRouter.taskDefinition, "*")
+		if logRouter.ec2InstanceID != "" {
+			config.AddFieldToRecord("ec2_instance_id", logRouter.ec2InstanceID, "*")
+		}
+	}
+
+	// Specify log stream output. Each container that uses the log router container to stream logs
+	// will have its own output section, with its own log options.
+	fields := strings.Split(logRouter.taskARN, "/")
```
PettitWesley: nit: the SDK go has an [arn package](https://docs.aws.amazon.com/sdk-for-go/api/aws/arn/)

(I'm fine with this though)

agent/taskresource/logrouter/logrouter_unimplemented.go
```diff
@@ -0,0 +1,135 @@
...
+	return nil, nil
+}
+
+// UnmarshalJSON unmarshals LogRouterResource object.
+func (c *LogRouterResource) UnmarshalJSON(b []byte) error {
+	return nil
+}
+
+// Initialize fills in the resource fields.
+func (cgroup *LogRouterResource) Initialize(resourceFields *taskresource.ResourceFields,
```
PettitWesley: nit: Calling the LogRouterResource a `cgroup` or `c` looks like code copy pasta :P

agent/taskresource/logrouter/logrouter_unix.go
```diff
@@ -0,0 +1,355 @@
...
+		return errors.Wrap(err, "unable to generate log router config")
+	}
+
+	temp, err := logRouter.ioutil.TempFile(logRouter.resourceDir, tempFile)
+	if err != nil {
+		return err
+	}
+	defer temp.Close()
+	if logRouter.logRouterType == LogRouterTypeFluentd {
+		err = config.WriteFluentdConfig(temp)
```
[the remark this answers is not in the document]
PettitWesley: This comes from the little library I wrote: https://github.com/awslabs/go-config-generator-for-fluentd-and-fluentbit

That's a good idea; I can/should change it to that

agent/taskresource/logrouter/logrouter_unimplemented.go
```diff
@@ -0,0 +1,135 @@
...
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+)
+
+const (
+	// ResourceName is the name of the log router resource.
+	ResourceName = "logrouter"
+)
+
+// LogRouterResource represents the log router resource.
+type LogRouterResource struct{}
```
PettitWesley: Just for my own understanding... what's the point of this "unimplemented" LogRouterResource?

agent/taskresource/logrouter/logrouter_unix.go
```diff
@@ -0,0 +1,355 @@
...
+		return err
+	}
+
+	err = temp.Chmod(os.FileMode(configFilePerm))
+	if err != nil {
+		return err
+	}
+
+	// Persist the config file to disk.
+	err = temp.Sync()
```
PettitWesley: I'm curious why you create a temp file, and then rename it to the desired path? What's the purpose of that?

agent/taskresource/logrouter/logrouter_unix.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		return err
+	}
+
+	confFilePath := filepath.Join(filepath.Join(logRouter.resourceDir, "config"), "fluent.conf")
```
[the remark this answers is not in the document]
fenxiong: good point. updated

agent/taskresource/logrouter/logconfig_unix.go
```diff
@@ -0,0 +1,144 @@
...
+	output, ok := logOptions[outputKey]
+	if !ok {
+		return config, errors.New(
+			fmt.Sprintf("missing output option %s which is required for log router type %s",
+				outputKey, logRouterType))
+	}
+
+	outputOptions := make(map[string]string)
+	for key, value := range logOptions {
+		if key == outputKey {
```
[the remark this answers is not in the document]
fenxiong: updated

agent/taskresource/logrouter/logconfig_unix.go
fenxiong: sgtm, updated to outputTypeLogOptionKeyxxx

agent/taskresource/logrouter/logconfig_unix.go
fenxiong: yeah makes sense, updated to /var/run/fluent.sock

agent/taskresource/logrouter/logrouter_unimplemented.go
fenxiong: good catch. fixed

agent/taskresource/logrouter/logrouter_unimplemented.go
fenxiong: the current implementation of LogRouterResource is very specific to Linux so it's done in logrouter_unix.go with a linux build tag (at the first line). This file has a "!linux" build tag so the code can compile in all platforms without having to restrict the reference of LogRouterResource in linux specific place as well (which probably isn't possible without refactoring existing agent code)

agent/taskresource/logrouter/logrouter_unix.go
fenxiong: i think this ensures we never end up using a partially written config file (when write either hasn't finished or fails in the middle). Though we can trust resource dependency management to prevent that as well, i think there's no downside of having this to keep it safe. and it's the pattern being used to write agent state file and container metadata file anyway..

agent/taskresource/logrouter/logconfig_unix.go
fenxiong: i think even if we use that package we'd still have to split the "Resource" field in the same way.. so keeping this

agent/taskresource/logrouter/logrouter_unix.go
yunhee-l: we do this in other places of agent too https://github.com/aws/amazon-ecs-agent/pull/1765

agent/taskresource/logrouter/logconfig_unix.go
```diff
@@ -0,0 +1,145 @@
...
+func (logRouter *LogRouterResource) generateConfig() (generator.FluentConfig, error) {
+	config := generator.New()
+
+	// Specify log stream input, which is a unix socket that will be used for communication between log router container
+	// and other containers.
+	var inputName, inputPathOption string
+	if logRouter.logRouterType == LogRouterTypeFluentd {
+		inputName = socketInputNameFluentd
+		inputPathOption = socketInputPathOptionFluentd
+	} else {
```
sharanyad: do we anticipate only these two router types or can there be third in the future?

agent/taskresource/logrouter/logrouter_unix.go
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	seelog.Infof("Generated log router config file at: %s", confFilePath)
+	return nil
+}
+
+// Cleanup performs resource cleanup.
+func (logRouter *LogRouterResource) Cleanup() error {
+	err := logRouter.os.RemoveAll(logRouter.resourceDir)
+	if err != nil {
```
sharanyad: is there a way we can force remove them?

agent/taskresource/logrouter/logconfig_unix.go
PettitWesley: There could theoretically be a third, but realistically for the near future (like the next year or more) there will only be these 2

agent/taskresource/logrouter/json_unix.go
```diff
@@ -0,0 +1,110 @@
...
+		ContainersToLogOptions: logRouter.containerToLogOptions,
+		TerminalReason:         logRouter.terminalReason,
+		CreatedAt:              logRouter.createdAtUnsafe,
+		DesiredStatus: func() *LogRouterStatus {
+			desiredStatus := logRouter.desiredStatusUnsafe
+			s := LogRouterStatus(desiredStatus)
+			return &s
+		}(),
+		KnownStatus: func() *LogRouterStatus {
+			desiredStatus := logRouter.knownStatusUnsafe
```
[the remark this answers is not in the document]
fenxiong: good catch. fixed the name

agent/taskresource/logrouter/logrouter_unix.go
fenxiong: hmm if we fail to remove the directory, i guess we just can't remove it.. i can't think of why that can happen though

agent/taskresource/logrouter/logconfig_unix.go
sharanyad: in that case, let's explicitly check if the router type is fluentbit in the `else`, instead of assuming it.

fenxiong: i already checked that there can only be two types here: https://github.com/aws/amazon-ecs-agent/pull/2097/commits/c643d889f318f738f9b965aed7a4e6e2d1e9392a#diff-27d3a72427896263cf317001fda17336R258. so i don't think we need to check again..

PR #2098, opened 2019-06-25
Fix task resource unmarshal error case

fenxiong:
### Summary

Previously task resource unmarshal error was ignored rather than returned. This PR fixes it.

Also fixes unit tests that start failing on mac after exposing the error. I had to fix a lot of state files used in state manager unit tests because they were using cgroup, but cgroup is not supported on non-linux platform.

### Testing

New tests cover the changes: 

### Licensing


agent/taskresource/types/types.go
```diff
@@ -51,35 +51,43 @@ func (rm *ResourcesMap) UnmarshalJSON(data []byte) error {
...
 		}
 	}
 	*rm = result
 	return nil
 }
 
+func unmarshalResource(key string, value json.RawMessage, result map[string][]taskresource.TaskResource) error {
+	switch key {
+	case CgroupKey:
+		if err := unmarshlCgroup(key, value, result); err != nil {
```
petderek: these could all be simplified to:

return umarshalThing(key,value,result)

```
// this...
if err != nil {
    return err
} else {
    return nil
}

// ...is the same as
return err

File: agent/taskresource/cgroup/cgroup_unsupported.go [hunk 2]
```diff
@@ -113,12 +113,12 @@ func (c *CgroupResource) GetTerminalReason() string {
 
 // MarshalJSON marshals CgroupResource object
 func (c *CgroupResource) MarshalJSON() ([]byte, error) {
-	return nil, errors.New("unsupported platform")
```
petderek: We'd want this to fail loudly on windows if this path got crossed, right?

fenxiong: yeah. keeping the error and fixes the state file instead

agent/taskresource/types/types.go
fenxiong: updated

PR #2099, opened 2019-06-26
Support for namespace-aware mounts

petderek:
### Summary

Adds a library that makes it possible to mount given a network namespace. This will eventually be used to send NFS traffic over different ENIs.

### Implementation details

This makes use of the mount syscall, so we don't need to directly depend on nfs utilities existing on the host operating system. 

### Testing

Testing includes both integ and unit tests. 

The unit tests mostly assure us that errors are handled in correctly, and that the mount options are set correctly for the happy case. 

The integration tests are a bit more complicated. We setup a test nfs server inside a docker container and setup mount connections to it. Definitely take a look at the misc/nfs directory contents to see how the test mount is created and how it runs. If the explanations on the test don't make much sense, let me know and I'll try to clarify. I'd like to reuse this test container as we build out additional support.

(none yet -- just library code)

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -241,7 +241,7 @@ test-registry: netkitten volumes-test namespace-tests pause-container squid awsc
 
 # TODO, replace this with a build on dockerhub or a mechanism for the
 # functional tests themselves to build this
-.PHONY: squid awscli fluentd gremlin agent-introspection-validator taskmetadata-validator v3-task-endpoint-validator container-metadata-file-validator elastic-inference-validator image-cleanup-test-images ecr-execution-role-image container-health-check-image telemetry-test-image storage-stats-test-image
+.PHONY: squid awscli fluentd gremlin agent-introspection-validator taskmetadata-validator v3-task-endpoint-validator container-metadata-file-validator elastic-inference-validator image-cleanup-test-images ecr-execution-role-image container-health-check-image telemetry-test-image storage-stats-test-image nfs-server-image
 
 squid:
 	$(MAKE) -C misc/squid $(MFLAGS)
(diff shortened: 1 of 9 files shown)


agent/taskresource/efs/efs_linux.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Mount creates the nfs mount, placing it at TargetDirectory on the host
+func (nm *NFSMount) Mount() error {
+	if nm.IPAddress == "" {
+		return errors.New("no IP Address set for nfs mount")
+	}
+
+	if nm.NamespacePath != "" {
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
```
fenxiong: do we need to set back the network namespace before unlocking the thread?

Makefile
```diff
@@ -293,12 +293,13 @@ run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tes
 	ECS_LOGLEVEL=debug go test -race -tags integration -timeout=25m -v ./agent/engine/... ./agent/stats/... ./agent/app/...
 endif
 
+NAMESPACE_PID = $(shell docker inspect -f "{{ .State.Pid }}" nfs-test)
 ifeq (${BUILD_PLATFORM},aarch64)
-run-sudo-tests::
-	sudo -E ${GO_EXECUTABLE} test -tags sudo -timeout=10m -v ./agent/engine/...
+run-sudo-tests:: nfs-server-image
+	ECS_FTEST_EFS_NETWORK_NAMESPACE=/proc/$(NAMESPACE_PID)/ns/net ECS_FTEST_EFS_IP_ADDRESS=127.0.0.1 sudo -E ${GO_EXECUTABLE} test -tags sudo -timeout=10m -v ./agent/engine/...
```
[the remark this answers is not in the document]
petderek: yup. I merged this wrong. will fix.

agent/taskresource/efs/efs_linux.go
petderek: Good catch! I've been testing this only in short lived binaries.

PR #2106, opened 2019-07-01
Set up log router resource and log router container config

fenxiong:
### Summary

Set up log router resource and log router container config:
1. When there's a log router container in a task, initialize a log router task resource and add it as a dependency of the log router container.
2. When creating the log router container, add config file bind mount and socket directory bind mount to its host config.

### Implementation details

1 is done in task's PostUnmarshalTask method similar to what was done for initializing other task resource.
2 is done in engine's createContainer method.

### Testing

Unit tests added. Manually tested this by injecting [test code](https://github.com/fenxiong/amazon-ecs-agent/commit/8df9c520ce342f7b64b7c8d6da34f2298958d630) that hardcodes the missing pieces.

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -39,6 +39,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/asmauth"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/asmsecret"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/logrouter"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/ssmsecret"
 	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
 	resourcetype "github.com/aws/amazon-ecs-agent/agent/taskresource/types"
@@ -94,6 +95,22 @@ const (
 	ipcModeTask     = "task"
 	ipcModeSharable = "shareable"
 	ipcModeNone     = "none"
+
+	// logRouterConfigBindFormatFluentd and logRouterConfigBindFormatFluentbit specifies the format of log router
+	// config file bind mount for fluentd and fluentbit log router respectively.
+	// First placeholder is host data dir, second placeholder is taskID.
+	logRouterConfigBindFormatFluentd   = "%s/data/logrouter/%s/config/fluent.conf:/fluentd/etc/fluent.conf"
+	logRouterConfigBindFormatFluentbit = "%s/data/logrouter/%s/config/fluent.conf:/fluent-bit/etc/fluent-bit.conf"
+	// logRouterSocketBindFormat specifies the format for log router socket directory bind mount.
+	// First placeholder is host data dir, second placeholder is taskID.
+	logRouterSocketBindFormat = "%s/data/logrouter/%s/socket/:/var/run/"
+	// logRouterDriverName is the log driver name for containers that want to use the log router.
+	logRouterDriverName = "awsrouter"
+
+	// awsExecutionEnvKey is the key of the env specifying the execution environment.
+	awsExecutionEnvKey = "AWS_EXECUTION_ENV"
+	// ec2ExecutionEnv specifies the ec2 execution environment.
+	ec2ExecutionEnv = "AWS_ECS_EC2"
 )
 
 // TaskOverrides are the overrides applied to a task
@@ -306,6 +323,13 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	}
 	// Adds necessary Pause containers for sharing PID or IPC namespaces
 	task.addNamespaceSharingProvisioningDependency(cfg)
+
+	if task.hasLogRouter() {
+		err = task.initializeLogRouterResource(cfg, resourceFields)
+		if err != nil {
+			return apierrors.NewResourceInitError(task.Arn, err)
+		}
+	}
 
 	return nil
 }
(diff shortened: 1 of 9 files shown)


agent/api/task/task.go
```diff
@@ -709,6 +734,92 @@ func (task *Task) getAllASMSecretRequirements() map[string]apicontainer.Secret {
 	return reqs
 }
 
+// hasLogRouter returns true if the task has a log router container.
+func (task *Task) hasLogRouter() bool {
+	for _, container := range task.Containers {
+		if container.LogRouter != nil { // This is a log router container.
```
sharanyad: where is this set?

agent/api/task/task.go
```diff
@@ -709,6 +734,92 @@ func (task *Task) getAllASMSecretRequirements() map[string]apicontainer.Secret {
...
+			continue
+		}
+
+		hostConfig := &dockercontainer.HostConfig{}
+		err := json.Unmarshal([]byte(*container.DockerConfig.HostConfig), hostConfig)
+		if err != nil {
+			return nil, errors.Wrapf(err, "unable to decode host config of container %s", container.Name)
+		}
+
+		if hostConfig.LogConfig.Type == logRouterDriverName {
```
sharanyad: where is this set?

agent/api/task/task.go
```diff
@@ -709,6 +734,92 @@ func (task *Task) getAllASMSecretRequirements() map[string]apicontainer.Secret {
...
+	if err != nil {
+		return errors.Wrap(err, "unable to initialize log router resource")
+	}
+
+	var logRouterResource *logrouter.LogRouterResource
+	for _, container := range task.Containers {
+		logRouter := container.LogRouter
+		if logRouter != nil {
+			var ec2InstanceID string
+			if container.Environment != nil && container.Environment[awsExecutionEnvKey] == ec2ExecutionEnv {
```
sharanyad: why is this check needed?

agent/api/task/task.go
fenxiong: customer will need to specify this in task definition, which will then goes into acs task payload

agent/api/task/task.go
fenxiong: same as above, this comes from task definition

sharanyad: customer don't set the hostConfig right? does ACS set this?

fenxiong: yes. ACS sets this field to be the value of the logDriver field specified in LogConfiguration object in task definition (https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_LogConfiguration.html).

agent/api/task/task.go
```diff
@@ -709,6 +734,92 @@ func (task *Task) getAllASMSecretRequirements() map[string]apicontainer.Secret {
 	return reqs
 }
 
+// hasLogRouter returns true if the task has a log router container.
+func (task *Task) hasLogRouter() bool {
+	for _, container := range task.Containers {
+		if container.LogRouter != nil { // This is a log router container.
+			return true
```
sharanyad: it might be better to check the log router type here. But this can be refactored later as we add more log routers

agent/api/task/task.go
fenxiong: discussed offline. we don't want to show the instance id for all the cases

agent/taskresource/logrouter/logrouter_unix.go
```diff
@@ -89,6 +89,41 @@ func NewLogRouterResource(cluster, taskARN, taskDefinition, ec2InstanceID, dataD
 	return logRouterResource
 }
 
+// GetCluster returns the cluster
+func (logRouter *LogRouterResource) GetCluster() string {
```
sharanyad: where are these getters used? I see them only in tests

fenxiong: yeah, they are only used in tests in the api/task package. None of the existing task resources exposes public fields, so i keep these fields private and use getter.

agent/api/task/task.go
fenxiong: not sure if i understand correctly, but i think regardless of the log router type, hasLogRouter should return true if container.LogRouter != nil. If there are more log router types later, initializeLogRouterResource method or the log router resource itself probably needs to change to deal with new type.

agent/api/task/task.go
```diff
@@ -94,6 +95,23 @@ const (
...
+	// logRouterSocketBindFormat specifies the format for log router socket directory bind mount.
+	// First placeholder is host data dir, second placeholder is taskID.
+	logRouterSocketBindFormat = "%s/data/logrouter/%s/socket/:/var/run/"
+	// logRouterDriverName is the log driver name for containers that want to use the log router.
+	// TODO: name is still being finalized.
+	logRouterDriverName = "awslogrouter"
+
+	// awsExecutionEnvKey is the key of the env specifying the execution environment.
+	awsExecutionEnvKey = "AWS_EXECUTION_ENV"
+	// ec2ExecutionEnv specifies the ec2 execution environment.
```
yumex93: just wanna know what can be the value of ec2 execution environment?

agent/api/task/task.go
```diff
@@ -709,6 +734,92 @@ func (task *Task) getAllASMSecretRequirements() map[string]apicontainer.Secret {
...
+			if container.Environment != nil && container.Environment[awsExecutionEnvKey] == ec2ExecutionEnv {
+				ec2InstanceID = resourceFields.EC2InstanceID
+			}
+
+			logRouterResource = logrouter.NewLogRouterResource(config.Cluster, task.Arn, task.Family+":"+task.Version,
+				ec2InstanceID, config.DataDir, logRouter.Type, logRouter.EnableECSLogMetaData, containerToLogOptions)
+			task.AddResource(logrouter.ResourceName, logRouterResource)
+			container.BuildResourceDependency(logRouterResource.GetName(), resourcestatus.ResourceCreated,
+				apicontainerstatus.ContainerCreated)
+			return nil
```
yumex93: It directly returns nil here. Does this mean that it only has one log router container per task?

fenxiong: yes, there can only be one.

agent/api/task/task.go
```diff
@@ -709,6 +734,92 @@ func (task *Task) getAllASMSecretRequirements() map[string]apicontainer.Secret {
...
+		}
+	}
+
+	return errors.New("unable to initialize log router resource because there's no log router container")
```
yumex93: In order for log router set up correctly, do other containers need to wait for log router container running first before they transit to running status?

fenxiong: yeah, other containers can only send logs after the log router container starts running. The design is to not add this ordering internally, but rely on customer to specify the dependency in task definition.

yumex93: Will the backend side have the logic to verify whether customers specify the dependency or not?

fenxiong: no (though we probably can.. i don't have a preference on this). we added an option "fluentd-async-connect" to containers using log router (https://github.com/aws/amazon-ecs-agent/pull/2120) so that they will not fail to start if they start before the log router starts (and message will be buffered until the log router starts). If customers don't want to buffer the messages, they need to specify the container ordering. Otherwise the message will be buffered up to a limit of log sender container's memory limit before sending to log router container when it starts, if it starts later than the log sender container (see https://docs.docker.com/config/containers/logging/fluentd/, fluentd-async-connect and fluentd-buffer-limit options)

agent/taskresource/logrouter/logrouter_unix.go
sharanyad: ok, could you add a comment mentioning these are used only for tests for now? I don't want us to accidentally use them without locks.

fenxiong: sure. added comments for each of these methods

fenxiong:
TestContainerHealthCheck failed in feature branch but has been fixed in dev branch https://github.com/aws/amazon-ecs-agent/pull/2127. Merging

PR #2107, opened 2019-07-01
Cleanup makefile

petderek:
### Summary

Cleans up the makefile a bit.

    1. Removes the huge list of test artifacts (they aren't used and likely won't
    be).
    
    2. Simplifies a lot of the branching logic in 'go test'
    
    3. Use -count=1 when running any of the tests to prevent caching from causing
    goofy test results. 'go help test' will suggest that this is the canonical way
    to prevent caching.

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,6 @@
 # language governing permissions and limitations under the License.
 
 USERID=$(shell id -u)
-GO_EXECUTABLE=$(shell command -v go 2> /dev/null)
 
 .PHONY: all gobuild static xplatform-build docker release certs test clean netkitten test-registry namespace-tests run-functional-tests benchmark-test gogenerate run-integ-tests pause-container get-cni-sources cni-plugins test-artifacts
 BUILD_PLATFORM:=$(shell uname -m)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -93,129 +92,49 @@ release: certs docker-release
...
-	go test -tags unit -timeout=30s -cover $(shell go list ./agent/... | grep -v /vendor/)
-else
-test-silent::
-	go test -race -tags unit -timeout=30s -cover $(shell go list ./agent/... | grep -v /vendor/)
-endif
-
-benchmark-test:
-	go test -run=XX -bench=. $(shell go list ./agent/... | grep -v /vendor/)
+gogenerate:
+	go generate -x ./agent/...
```
fierlion: I can't find documentation about `$(shell go list ./agent/...)` vs `./agent/...`
Can you explain this change?  is there an update to the generate that means we no longer need to list, or was this just un-necessary in the first place?

Makefile
```diff
@@ -93,129 +92,49 @@ release: certs docker-release
...
-		--rm \
-		$(BUILDER_IMAGE) \
-		$(1)
-endef
+# VERBOSE includes the options that make the test opt loud
+VERBOSE=-v -cover
 
-# TODO: use `go list -f` to target the test files more directly
-ALL_GO_FILES = $(shell find . -name "*.go" -print | tr "\n" " ")
+# -count=1 prevents the build cache from breaking integ tests
```
fierlion: This wording is less clear than what you said in the message above `Use -count=1 when running any of the tests to prevent caching from causing goofy test results`.  Can you tweak this to be super-clear that count=1 prevents caching.

Makefile
petderek: I mentioned it in another PR (I don't think its worth putting references to old code in current code). Basically,

`go test ./my-package/...` used to also do things to ./my-package/vendor/. They eventually made it so that go test / list didn't expand to include the vendored directory by default.

We were using `go list ./agent/... | grep -v vendor` to exclude manually, but thats no longer needed.

Makefile
petderek: Yep, good point

Makefile
```diff
@@ -93,129 +92,49 @@ release: certs docker-release
...
-
-LINUX_ARTIFACTS_TARGETS := out/test-artifacts/linux-engine-tests out/test-artifacts/linux-stats-tests
-LINUX_ARTIFACTS_TARGETS += out/test-artifacts/linux-app-tests out/test-artifacts/linux-simple-tests
-LINUX_ARTIFACTS_TARGETS += out/test-artifacts/linux-handwritten-tests
-
-test-artifacts-windows: $(WINDOWS_ARTIFACTS_TARGETS)
-
-test-artifacts-linux: $(LINUX_ARTIFACTS_TARGETS)
+run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tests
+	ECS_LOGLEVEL=debug ${GOTEST} -tags integration -timeout=25m ./agent/...
```
shubham2892: Are we skipping the `-race` flag? 🤔

Makefile
```diff
@@ -93,129 +92,49 @@ release: certs docker-release
...
-		--net none \
-		--user "$(USERID)" \
-		--volume "$(PWD)/out:/out" \
-		--volume "$(PWD):/go/src/github.com/aws/amazon-ecs-agent" \
-		--rm \
-		$(BUILDER_IMAGE) \
-		$(1)
-endef
+# 'go' may not be on the $PATH for sudo tests
+GO_EXECUTABLE=$(shell command -v go 2> /dev/null)
```
sharanyad: does this work for non sudo as well?

Makefile
petderek: Nope. It gets added conditionally to $GOTEST

Makefile
petderek: Yep. This just grabs the absolute path to the go binary (/usr/local/go/bin/go usually). 

If you install this via yum/apt, the root user will have go in its path. If you install it manually, root has no way to find the binary (and therefore needs the absolute path). 

Path considerations aside,  'go test' and '/usr/local/go/bin/go test' are equivalent.

PR #2108, opened 2019-07-01
 Specify fluentd log driver options for containers using log router

mssrivas:
### Implementation details
If LogRouter values are present, check to see if they are the ones expected and then specify the log driver options

### Testing
Added unit test
Test ran on local and linux

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -16,8 +16,10 @@ package engine
 
 import (
 	"context"
+	"path/filepath"
 	"regexp"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -45,6 +47,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
 	utilsync "github.com/aws/amazon-ecs-agent/agent/utils/sync"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
+	dockercontainer "github.com/docker/docker/api/types/container"
 
 	"github.com/cihub/seelog"
 	"github.com/docker/docker/api/types"
@@ -69,6 +72,13 @@ const (
 	maxEngineConnectRetryDelay         = 2 * time.Second
 	engineConnectRetryJitterMultiplier = 0.20
 	engineConnectRetryDelayMultiplier  = 1.5
+	logConfigType                      = "awslogrouter"
+	logDriverType                      = "fluentd"
+	logDriverTag                       = "tag"
+	logDriverFluentdAddress            = "fluentd-address"
+	dataLogDriverPath                  = "/data/logrouter/"
+	dataLogDriverSocketPath            = "/socket/fluent.sock"
+	socketPathPrefix                   = "unix://"
 )
 
 // DockerTaskEngine is a state machine for managing a task and its containers
@@ -868,6 +878,14 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(hcerr)}
 	}
 
+	// If the container is using a special log driver type "awslogrouter", it means the container wants to use
+	// log router to send logs. In this case, override the log driver type to be fluentd
+	// and specify appropriate tag and fluentd-address, so that the logs are sent to and routed by the log router
+	// For reference - https://docs.docker.com/config/containers/logging/fluentd/
+	if hostConfig.LogConfig.Type == logConfigType {
+		hostConfig.LogConfig = getLogRouterConfig(task, container, hostConfig, engine.cfg)
+	}
+
 	if container.AWSLogAuthExecutionRole() {
 		err := task.ApplyExecutionRoleLogsAuth(hostConfig, engine.credentialsManager)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -868,6 +877,11 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(hcerr)}
 	}
 
+	// If LogRouter values are present, check to see if they are the ones expected and then specify the log driver options
+	if container.LogRouter != nil && container.LogRouter.Type != "" && container.LogRouter.Type == logConfigType {
```
petderek: you can omit the middle comparison (redundant).

agent/engine/docker_task_engine.go
```diff
@@ -945,6 +959,24 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 	return metadata
 }
 
+// Log driver options type, tag and fluentd-address are added
+// For reference - https://docs.docker.com/config/containers/logging/fluentd/
+func getLogRouterConfig(task *apitask.Task, container *apicontainer.Container, hostConfig *container.HostConfig, engine *DockerTaskEngine) container.LogConfig {
+	seelog.Infof("Since the value for logrouter in container is %s, we modify LogConfig values in hostconfig to support log router driver", logConfigType)
```
petderek: I wouldn't put this on the getter. I'd move it to where it gets applied. Consider setting it as a debug level log.

Also, this is super friendly and helpful, but for the logging we should be more terse. Something like:
`"applying logrouter config for container %s"`

Remember that the consumers of the log will be us or other devs - folks that'll be able to search the for the error string to find whats going on if they suspect problems.

agent/engine/docker_task_engine.go
```diff
@@ -945,6 +959,24 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+	fields := strings.Split(task.Arn, "/")
+	taskID := fields[len(fields)-1]
+	tag := container.Name + "-" + taskID
+	fluentd := engine.cfg.DataDirOnHost + dataLogDriverPath + taskID + dataLogDriverSocketPath
```
petderek: nit: use filepath.Join()

agent/engine/docker_task_engine.go
```diff
@@ -945,6 +959,24 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+	logConfig := hostConfig.LogConfig
```
petderek: I think what you want to do here is something like:

logConfig := LogConfig{} // new instance of log config object

But correct me if I'm mistaken :)

agent/engine/docker_task_engine.go
```diff
@@ -868,6 +878,13 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(hcerr)}
 	}
 
+	// If LogRouter values are present, check to see if they are the ones expected and
+	// then specify the log driver options:  tag and fluentd-address are added
+	// For reference - https://docs.docker.com/config/containers/logging/fluentd/
+	if container.LogRouter != nil && container.LogRouter.Type == logConfigType {
```
fenxiong: For a task using log router, there is one "log router container", and 1 or many "container that uses log router". The "log router container" has container.LogRouter != nil that contains definition for the log router, while a "container that uses log router" has a special log driver name "logrouter" (or whatever the name we end up using). The log driver name for a container is saved in hostConfig.LogConfig.Type. So you should check hostConfig.LogConfig.Type here, rather than container.LogRouter.Type.

agent/engine/docker_task_engine.go
```diff
@@ -945,6 +961,22 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 	return metadata
 }
 
+func getLogRouterConfig(task *apitask.Task, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig, engine *DockerTaskEngine) dockercontainer.LogConfig {
+	seelog.Debugf("Applying logrouter config for container %s", container.Name)
+	fields := strings.Split(task.Arn, "/")
+	taskID := fields[len(fields)-1]
+	tag := container.Name + "-" + taskID
+	fluentd := filepath.Join(engine.cfg.DataDirOnHost, dataLogDriverPath, taskID, dataLogDriverSocketPath)
```
fenxiong: this needs to have a prefix "unix://". when the driver uses a socket it needs to have that prefix

agent/engine/docker_task_engine.go
```diff
@@ -945,6 +961,22 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 	return metadata
 }
 
+func getLogRouterConfig(task *apitask.Task, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig, engine *DockerTaskEngine) dockercontainer.LogConfig {
```
fenxiong: can you add a unit test for this method? 
nit: "engine *DockerTaskEngine" argument can just be "cfg *config.Config" since we only use the config rather than other part of the engine.

agent/engine/docker_task_engine_test.go
```diff
@@ -2694,3 +2695,83 @@ func TestTaskSecretsEnvironmentVariables(t *testing.T) {
...
+			name:                           "test create fluentd log router container",
+			task:                           getTask(),
+			expectedLogConfigType:          "fluentd",
+			expectedLogConfigTag:           taskName + "-" + taskARN,
+			expectedLogConfigFluentAddress: filepath.Join(defaultConfig.DataDirOnHost, dataLogDriverPath, taskARN, dataLogDriverSocketPath),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
```
fenxiong: since there's only one test case, it seems weird to have a for loop going through a list of test cases. Can you just write the test directly (without a loop and testCase struct)?

agent/engine/docker_task_engine.go
```diff
@@ -868,6 +877,13 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(hcerr)}
 	}
 
+	// If LogRouter values are present, check to see if they are the ones expected and
```
fenxiong: This should probably be something like: If the container is using a special log driver type "awslogrouter", it means the container wants to use log router to send logs. In this case, override the log driver type to be fluentd and specify appropriate tag and fluentd-address, so that the logs are sent to and routed by the log router.

agent/engine/docker_task_engine_test.go
mssrivas: Added a new one to test it doesn't populate logdriver in case the logconfig is not whats expected

agent/engine/docker_task_engine.go
mssrivas: Talked about this offline.

agent/engine/docker_task_engine_test.go
```diff
@@ -2694,3 +2695,94 @@ func TestTaskSecretsEnvironmentVariables(t *testing.T) {
 		})
 	}
 }
+
+// TestCreateContainerAddLogDriverConfig tests that in createContainer, when the
+// container's logConfig is set, the driver values are as expected
+func TestCreateContainerAddLogDriverConfig(t *testing.T) {
+	taskName := "logSenderTask"
+	taskARN := "arn:aws:ecs:region:account-id:task/task-id"
+	taskid := "task-id"
```
fenxiong: nit: taskID rather than taskid

PR #2115, opened 2019-07-08
Update Agent to 1.29.1

shubham2892:
### Summary

Update Agent to 1.29.1

### Testing

New tests cover the changes: 

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2116, opened 2019-07-08
Add efs task resource for efs support

yumex93:
### Summary

1. Add efs resource. For create function, it will create mount target directory and call mount system call to mount the EFS file system. For cleanup function, it will unmount the EFS and delete the directory.
2. Add four new functions for TaskResource interface to support new task resource transition logic, so that task resource transition can be the same as container, which depends on the transition map.
The functions are: GetAppliedStatus(), DependOnTaskNetwork(), GetContainerDependencies(), BuildContainerDependency()
3. Add the implementation of the functions mentioned above to other task resources, so that they can still implement TaskResource interface. It is included in the same commit so that all tests can pass.

### Testing

New tests cover the changes: 

### Licensing


yumex93:
The fail test is TestContainerHealthCheck, which is marked as flaky test so far.

agent/taskresource/mocks/taskresource_mocks.go
```diff
@@ -52,160 +54,185 @@ func (m *MockTaskResource) EXPECT() *MockTaskResourceMockRecorder {
 }
 
 // ApplyTransition mocks base method
-func (m *MockTaskResource) ApplyTransition(arg0 status0.ResourceStatus) error {
-	m.ctrl.T.Helper()
```
fenxiong: what version of gomock are you using? seems like you are reverting the changes [here](https://github.com/aws/amazon-ecs-agent/pull/2115/commits/e0bde51c8934623932faaa1ea6d257a498478991)

PR #2119, opened 2019-07-09
Add agent capability for efs support

yumex93:
### Summary

1. Add agent capability for efs support. Since the efs will only be supported on linux, the capability logic is only added to linux go file.
2. Remove some redundant test cases, use one test case to just verify all capabilities for the latest agent version. 

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/app/agent_capability_unix.go
+++ b/agent/app/agent_capability_unix.go
@@ -114,3 +114,7 @@ func (agent *ecsAgent) appendAppMeshCapabilities(capabilities []*ecs.Attribute) 
 func (agent *ecsAgent) appendTaskEIACapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
 	return appendNameOnlyAttribute(capabilities, attributePrefix+taskEIAAttributeSuffix)
 }
+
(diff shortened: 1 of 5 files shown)


agent/app/agent_capability_unix.go
```diff
@@ -114,3 +114,7 @@ func (agent *ecsAgent) appendAppMeshCapabilities(capabilities []*ecs.Attribute)
 func (agent *ecsAgent) appendTaskEIACapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
 	return appendNameOnlyAttribute(capabilities, attributePrefix+taskEIAAttributeSuffix)
 }
+
+func (agent *ecsAgent) appendEFSCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
```
petderek: can you add a TODO here to check for the correct bind mount?

sharanyad:
is efs capability the only way backend could know if an agent supports it? 
or do we pass some info in RCI (like GPU) that could help identify efs support?

yumex93:
Yeah. This will be the only place. What the reason behind GPU passing info through RCI?

adnxn:
closing in favor of https://github.com/aws/amazon-ecs-agent/pull/2248

PR #2121, opened 2019-07-09
Edge case fixes for log router resource

fenxiong:
### Summary

Fix some edge cases for log router resource missed in previous PR.  This includes:
1. Initialize os and ioutil fields to avoid NPE upon agent restarts.
2. Update applied status back to none when setting resource known status.

### Testing

Added unit tests for the change.

### Licensing

### DIFF
--- a/agent/taskresource/logrouter/logrouter_unix_test.go
+++ b/agent/taskresource/logrouter/logrouter_unix_test.go
@@ -22,6 +22,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 
+	"github.com/aws/amazon-ecs-agent/agent/api/task/status"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
 	mock_ioutilwrapper "github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper/mocks"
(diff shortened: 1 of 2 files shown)


agent/taskresource/logrouter/logrouter_unix_test.go
```diff
@@ -260,3 +263,22 @@ func TestCleanupLogRouterResourceError(t *testing.T) {
...
+	logRouterResource := newMockLogRouterResource(LogRouterTypeFluentd, testFluentdOptions, nil, nil)
+	logRouterResource.Initialize(&taskresource.ResourceFields{}, status.TaskRunning, status.TaskRunning)
+
+	assert.NotNil(t, logRouterResource.statusToTransitions)
+	assert.Equal(t, 1, len(logRouterResource.statusToTransitions))
+	assert.NotNil(t, logRouterResource.os)
+	assert.NotNil(t, logRouterResource.ioutil)
+}
+
+func TestSetKnownStatus(t *testing.T) {
```
yumex93: nit: would be better to add a test case to make sure when appliedStatusUnsafe > knownState, the applied status not get reset to none.

fenxiong: added the test

fenxiong:
> did you catch this during release testing?

This was spotted by @yumex93 by eyes :joy: i have manually verified that without the fix there will be NPE, and this PR fixes it.

PR #2129, opened 2019-07-11
Add opts

petderek:
### Summary

Adds options for 'readonly' and 'additional options', meaning we can now specify read only mounts and supported nfs options.

Fixes a bug with "source directory" where a double slash would be present (ie, `//foo/bar`).

### Implementation Details
We maintain a map of supported nfs options and compare client input to the supported option list. Options provided by clients will override default values.

### Testing
Additional unit tests

### Licensing

### DIFF
--- a/agent/taskresource/efs/efs_linux.go
+++ b/agent/taskresource/efs/efs_linux.go
@@ -23,13 +23,13 @@ import (
 )
 
 const (
-	// TODO: break this out into a better options stream (IE, we will need overrides when specified in the service)
-	defaultOptsForEFS = "rsize=1048576,wsize=1048576,timeo=10,hard,retrans=2,noresvport,vers=4"
-	defaultSource     = ":/"
-	fsTypeNFS         = "nfs"
-	mountFlags        = 0
-	unmountFlags      = 0
-	mountTimeout      = time.Minute
+	defaultOptsForEFS  = "rsize=1048576,wsize=1048576,timeo=10,hard,retrans=2,noresvport"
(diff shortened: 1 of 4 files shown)


agent/taskresource/efs/options_linux.go
```diff
@@ -0,0 +1,105 @@
...
+	"cto":      "nocto",
+	"rsize":    "",
+	"wsize":    "",
+	"timeo":    "",
+	"retrans":  "",
+}
+
+// checkOptionSet validates customer input against the supported NFS options
+// TODO: move to service layer exclusively?
+func checkOptionSet(optionSet string) error {
```
sparrc: could you add unit tests for this?

agent/taskresource/efs/efs_linux.go
```diff
@@ -23,13 +24,12 @@ import (
...
 const (
-	// TODO: break this out into a better options stream (IE, we will need overrides when specified in the service)
-	defaultOptsForEFS = "rsize=1048576,wsize=1048576,timeo=10,hard,retrans=2,noresvport,vers=4"
-	defaultSource     = ":/"
-	fsTypeNFS         = "nfs"
-	mountFlags        = 0
-	unmountFlags      = 0
-	mountTimeout      = time.Minute
+	defaultOptsForEFS  = "rsize=1048576,wsize=1048576,timeo=10,hard,retrans=2,noresvport"
+	enforcedOptsForEFS = "fg,vers=4"
```
sparrc: why are these enforced? are they documented somewhere?

PR #2139, opened 2019-07-24
Integration test for fluentbit and other Bug Fixes

mssrivas:
Change EnableECSLogMetaData to EnableECSLogMetadata, file path for socket and Integration test for fluentbit. The fluentd log setting are being tested as part of functional test in here - https://github.com/aws/amazon-ecs-agent/pull/2134/files

### Summary
1. EnableECSLogMetaData to EnableECSLogMetadata - to match the data model being generated.
2. File path bug found during integ test.
3. Integ test for fluentbit.
4. Adding command in make file to test sudo integ tests

### Testing
sudo -E /usr/local/go/bin/go test -run "TestLogSenderAndLogRouterForFluentbit" -tags 'sudo integration' -v ./agent/engine/... -timeout=3m

New tests cover the changes: 

### Licensing


agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -127,3 +153,166 @@ func createTestLocalVolumeMountTask() *apitask.Task {
...
+
+	time.Sleep(cfg.TaskCleanupWaitDuration)
+	for i := 0; i < 60; i++ {
+		_, ok = taskEngine.(*DockerTaskEngine).State().TaskByArn(testTask.Arn)
+		if !ok {
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+	// Make sure all the resource is cleaned up
```
fenxiong: can you check that the directory /var/lib/ecs/data/logrouter/<task-id> is removed as well?

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -127,3 +153,166 @@ func createTestLocalVolumeMountTask() *apitask.Task {
...
+			awsError, ok := err.(awserr.Error)
+			if !ok || awsError.Code() != "ResourceNotFoundException" {
+				return nil, err
+			}
+		} else if len(resp.Events) > 0 {
+			return resp, nil
+		}
+		time.Sleep(time.Second)
+	}
+	return nil, fmt.Errorf("Timeout waiting for the logs to be sent to cloud watch logs")
```
fenxiong: Timeout -> timeout

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -127,3 +153,166 @@ func createTestLocalVolumeMountTask() *apitask.Task {
...
+			break
+		}
+		time.Sleep(1 * time.Second)
+	}
+	// Make sure all the resource is cleaned up
+	assert.True(t, ok)
+}
+
+func createLogDriverAndContainersUsingLogDriverTask(t *testing.T) *apitask.Task {
+	testTask := createTestTask(validTaskArn + randSeq())
```
fenxiong: i think you can just use "arn:aws:ecs:region:account-id:task/" + uuid.New()

Makefile
```diff
@@ -125,12 +125,15 @@ test-silent:
...
 
-run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tests
+run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tests run-sudo-integ-tests
 	ECS_LOGLEVEL=debug ${GOTEST} -tags integration -timeout=25m ./agent/...
 
 run-sudo-tests:
 	sudo -E ${GOTEST} -tags sudo -timeout=10m ./agent/...
 
+run-sudo-integ-tests:
+	sudo -E ${GOTEST} -tags 'sudo integration' -timeout=10m ./agent/...
```
fenxiong: i think the new test takes more than 1 min to finish, so you should probably bump timeout here.. e.g. 12m

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -16,12 +16,24 @@
 package engine
 
 import (
+	"encoding/json"
```
fenxiong: nit: can you group the imports to be similar to what they were before: misc packages, empty line, "github.com/aws/amazon-ecs-agent/xxx" packages, empty line, "github.com/other/xxx" packages

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -32,11 +44,25 @@ import (
...
 	"github.com/stretchr/testify/assert"
 )
 
 const (
-	testVolumeImage = "127.0.0.1:51670/amazon/amazon-ecs-volumes-test:latest"
+	testLogSenderImage  = "amazonlinux:2"
+	testFluentbitImage  = "amazon/aws-for-fluent-bit:latest"
+	validTaskArn        = "arn:aws:ecs:region:account-id:task/"
+	testCluster         = "testCluster"
+	testDataDir         = "testDataDir"
```
fenxiong: testCluster, testDataDir and testDataDirOnHost are not used

agent/engine/engine_sudo_linux_integ_test.go
mssrivas: I did a goimport command and this is how it did it

Makefile
mssrivas: Okay.

fenxiong: nvm seems like the test was hanging on task cleanup, otherwise it should just take a few seconds. please fix cleanup instead of bumping timeout

agent/engine/docker_task_engine.go
```diff
@@ -975,12 +975,10 @@ func getLogRouterConfig(task *apitask.Task, container *apicontainer.Container, h
 	fields := strings.Split(task.Arn, "/")
 	taskID := fields[len(fields)-1]
 	tag := container.Name + "-" + taskID
-	fluentd := filepath.Join(socketPathPrefix, cfg.DataDirOnHost, dataLogDriverPath, taskID, dataLogDriverSocketPath)
+	fluentd := socketPathPrefix + filepath.Join(cfg.DataDirOnHost, dataLogDriverPath, taskID, dataLogDriverSocketPath)
```
yumex93: what's the difference between before/after change?

mssrivas: We want to keep the Unix:/// , the filepath join removes the extra //

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -127,3 +151,166 @@ func createTestLocalVolumeMountTask() *apitask.Task {
...
+		resp, err := client.GetLogEvents(params)
+		if err != nil {
+			awsError, ok := err.(awserr.Error)
+			if !ok || awsError.Code() != "ResourceNotFoundException" {
+				return nil, err
+			}
+		} else if len(resp.Events) > 0 {
+			return resp, nil
+		}
+		time.Sleep(time.Second)
```
yumex93: Would this be better to be exponential backoff? Will this meet throttling issue?

fenxiong: it's probably better, but i don't think it's needed for now as many of our functional tests are already calling APIs like this and are not throttled

PR #2144, opened 2019-07-29
Change logrouter names to firelens and fix bug

fenxiong:
### Summary

Change logrouter names to firelens and fix bug.

### Implementation details

Commit 1: change all log router related names to firelens, including acs/agent model, variable names, comments etc.
For reference, a firelens container with previous model looks like this:
```
Container {
  Name: ...
  ...
  LogRouter: {
    Type: "fluentd|fluentbit"
    EnableECSLogMetadata: true|false
  }
}
```

A firelens container with new model looks like this:
```
Container {
  Name: ...
  ...
  FirelensConfig: {
    Type: "fluentd|fluentbit"
    Options: map[string]string{
      "enable-ecs-log-metadata": "true|false",
    }
  }
}
```

Commit 2: fix a bug found in manual test, explained in code comment.

### Testing

The firelens integ test (TestFirelensFluentbit) passed. 

The firelens functional tests have been updated with the new fields in https://github.com/aws/amazon-ecs-agent/pull/2134 and tests pass in gamma.

Manually run some tasks as well.

### Licensing


agent/acs/model/api/api-2.json
```diff
@@ -228,7 +228,7 @@
         "dependsOn":{"shape":"ContainerDependencies"},
         "startTimeout":{"shape":"Integer"},
         "stopTimeout":{"shape":"Integer"},
-        "logRouter":{"shape":"LogRouter"}
+        "firelensConfiguration":{"shape":"FirelensConfiguration"}
```
yunhee-l: Do we not need to update the state file version for this? don't think I see that change in this cr

fenxiong: there are some changes in agent/statemanager/state_manager.go

yunhee-l: oh yep i see it now 👍

PR #2145, opened 2019-07-29
 Adjust firelens default behaviors

fenxiong:
### Summary

Adjust firelens default behaviors. This includes:
1. Add a START dependency between each log sender container and the firelens container if there's no dependency between them.
2. Enable appending ecs metadata to log stream by default.

### Implementation details

Implemented in PostUnmarshalTask and its callees.

### Testing

Added unit tests. Manually ran tasks against gamma and verified that the firelens container starts first and ecs metadata is enabled.

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -325,10 +325,19 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	task.addNamespaceSharingProvisioningDependency(cfg)
 
 	if task.hasFirelensContainer() {
-		err = task.initializeFirelensResource(cfg, resourceFields)
-		if err != nil {
-			return apierrors.NewResourceInitError(task.Arn, err)
-		}
+		return task.applyFirelensSetup(cfg, resourceFields)
+	}
+	return nil
+}
+
+func (task *Task) applyFirelensSetup(cfg *config.Config, resourceFields *taskresource.ResourceFields) error {
+	err := task.initializeFirelensResource(cfg, resourceFields)
+	if err != nil {
+		return apierrors.NewResourceInitError(task.Arn, err)
+	}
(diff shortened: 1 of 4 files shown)


agent/api/task/task.go
```diff
@@ -777,6 +786,43 @@ func (task *Task) initializeFirelensResource(config *config.Config, resourceFiel
...
+		hostConfig := &dockercontainer.HostConfig{}
+		err := json.Unmarshal([]byte(*container.DockerConfig.HostConfig), hostConfig)
+		if err != nil {
+			return errors.Wrapf(err, "unable to decode host config of container %s", container.Name)
+		}
+
+		if hostConfig.LogConfig.Type == firelensDriverName {
+			// If there's no dependency between the app container and the firelens container, make firelens container
+			// start first to be the default behavior by adding a START container depdendency.
+			if !container.DependsOnContainer(firelensContainer.Name) && !firelensContainer.DependsOnContainer(container.Name) {
```
yumex93: Are we going to have this documented in the public doc that recommend people to add this dependency for their task def?

fenxiong: i think we will document that if no dependency is specified, we will add a firelens container dependency internally by default. not sure why you think we should recommend people to add this dependency themselves?

yumex93: Cause we have a check that if customer defines a dependency already, we will not add the dependency. So if customers define the wrong dependency, it will still lead to the unexpected behavior. That's why I think in doc we should tell customers what is the right dependency look like or they can just leave it not specified.

fenxiong: > That's why I think in doc we should tell customers what is the right dependency look like or they can just leave it not specified.

that sounds good to me. i think we can mention these in the doc

PR #2147, opened 2019-07-31
Firelens: fix potential circular dependency

fenxiong:
### Summary

https://github.com/aws/amazon-ecs-agent/pull/2145 causes circular dependency when: (1) Firelens container indirectly depends on a container that uses awsfirelens log driver; (2) Firelens container itself is using awsfirelens log driver. This PR is fixing these.

### Implementation details

To not make things too complicated, we will simply skip adding any dependency when firelens container itself has dependency, in order to fix (1). Avoid adding firelens container dependency onto itself to fix (2).

### Testing

Unit tests added. Manually run tasks against gamma to verify these two situations are fixed.

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -786,6 +786,8 @@ func (task *Task) initializeFirelensResource(config *config.Config, resourceFiel
 	return errors.New("unable to initialize firelens resource because there's no firelens container")
 }
 
+// addFirelensContainerDependency adds a START dependency between each container using awsfirelens log driver
+// and the firelens container.
 func (task *Task) addFirelensContainerDependency() error {
 	var firelensContainer *apicontainer.Container
 	for _, container := range task.Containers {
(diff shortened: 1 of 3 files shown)


sharanyad:
We should let customers explicitly know that we will be dropping the dependencies through agent logs/console. Otherwise things could be failing silently.

fenxiong:
> We should let customers explicitly know that we will be dropping the dependencies through agent logs/console. Otherwise things could be failing silently.

I guess it's not "we will be dropping", but just we are not adding it for customer. For the case where agent won't add the dependency - i.e. firelens container itself has container dependency, backend will require customers themselves to specify the START dependency if it's bridge mode (where this dependency is required) in task definition, so that they won't end up using a task definition that's not working.

sharanyad:
> I guess it's not "we will be dropping", but just we are not adding it for customer

In network modes other than bridge mode, if customer has this circular dependency in their task def, agent doesn't add the dependency, yeah? So we should let the customer know that agent is not adding it(instead of assuming it is ok to).

fenxiong:
> In network modes other than bridge mode, if customer has this circular dependency in their task def, agent doesn't add the dependency, yeah? So we should let the customer know that agent is not adding it(instead of assuming it is ok to).

Yeah that makes sense. added a warning log for this.

agent/api/task/task.go
```diff
@@ -798,11 +800,23 @@ func (task *Task) addFirelensContainerDependency() error {
...
 		}
 
+		// Firelens container itself could be using awsfirelens log driver. Don't add container dependency in this case.
+		if container.Name == firelensContainer.Name {
+			continue
+		}
+
 		hostConfig := &dockercontainer.HostConfig{}
 		err := json.Unmarshal([]byte(*container.DockerConfig.HostConfig), hostConfig)
 		if err != nil {
```
yumex93: I think we don't need the condition "!firelensContainer.DependsOnContainer(container.Name)" anymore, since you already check whether firelensContainer has any dependency beforehand. So if it already comes to this point, it should already be true.

fenxiong: good catch, fixed.

PR #2150, opened 2019-08-02
Support secret log options for firelens

fenxiong:
### Summary

Since log options for awsfirelens log driver are written to config file rather than passed to Docker, secret log options for awsfirelens log driver need to be implemented differently as well. This PR implements it.

### Implementation details

The workflow is as follow:
1. Each secret log option for a container using awsfirelens driver has a value of a config file variable with name option-key_container-name when generating the config file.
2. When creating the firelens container, retrieves the secret log option value from secret  task resource and set the value of each config file variable to the secret value as the container's env.
3. Firelens container will then resolve the config file variable to the secret value when loading the config file base on the env.

Besides the above workflow, logic is added to let firelens container depends on secret resource when needed.

### Example
As an example, suppose we are using cloudwatch plugin for fluentbit firelens router, with a plain text log option "region" with value "us-west-2", and a secret log option "log_group_name", with value stored in ssm parameter "log_group_key", the task definition will be something like the following:
```
{
	"family": "xxx",
	"executionRoleArn": "xxx",
	"containerDefinitions": [
		{
			...
			"name": "log_router", // firelens container
			"firelensConfiguration": {
				"type": "fluentbit"
			}
		 },
		 {
			 ...
			 "name": "app", // application container that sends log using firelens
			 "logConfiguration": {
				 "logDriver":"awsfirelens",
				 "options": {
					"Name": "cloudwatch",
					"region": "us-west-2",
					"...
				},
				"secretOptions": [{
					"name": "log_group_name",
					"valueFrom": "log_group_key"
				}]
			}
		}
	]
}
```
When running the above task definition, container "app" will have a Secret looks like the following:
```
{
    Name: "log_group_name",
    ValueFrom: "log_group_key",
    Target: "log_driver",
    Provider: "ssm",
}
```

When initializing the firelens task resource, we initialize it with the following log options:
```
option 1 (plain text log option): key - region, value - us-west-2
option 2 (secret log option): key - log_group_name, value - ${log_group_name_app}

containerToLogOptions parameter for initializeFirelensResource will look like the following
{
"app": {
    "region": "us-west-2",
    "log_group_name": "${log_group_name_app}"
}
}
```
The options got written to firelens config file like the following (xxx will be task id):
```
...
<match app-xxx>
    Name cloudwatch
    region us-west-2
    log_group_name ${log_group_name_app}
</match>
...
```
When creating the firelens container, suppose the log group name stored in the ssm parameter is "testing", we inject an env with key "log_group_name_app" and value "testing" to firelens container's env. When firelens container starts and loads its config file, it will attempt to resolve the config variable ${log_group_name_app} with its env, and the config will be resolved  to:
```
...
<match app-xxx>
    Name cloudwatch
    region us-west-2
    log_group_name testing
</match>
...
```
And then the plugin will be properly initialize with the above config.

### Testing

Unit tests added. Manually run tasks against gamma to verify it works. For functional tests I will modify the ones in https://github.com/aws/amazon-ecs-agent/pull/2134 to cover secret log options.

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -14,6 +14,7 @@
 package container
 
 import (
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"sync"
@@ -23,9 +24,11 @@ import (
 	apierrors "github.com/aws/amazon-ecs-agent/agent/api/errors"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
-	"github.com/aws/aws-sdk-go/aws"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/cihub/seelog"
 	"github.com/docker/docker/api/types"
+	dockercontainer "github.com/docker/docker/api/types/container"
 )
 
 const (
@@ -871,19 +874,21 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
 	}
 }
 
-func (c *Container) HasSecretAsEnvOrLogDriver() bool {
+// HasSecret returns whether a container has secret based on a certain condition.
+func (c *Container) HasSecret(f func(s Secret) bool) bool {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 
-	// Secrets field will be nil if there is no secrets for container
 	if c.Secrets == nil {
 		return false
 	}
+
 	for _, secret := range c.Secrets {
-		if secret.Type == SecretTypeEnv || secret.Target == SecretTargetLogDriver {
+		if f(secret) {
 			return true
 		}
 	}
+
 	return false
 }
 
@@ -924,4 +929,39 @@ func (c *Container) AddContainerDependency(name string, condition string) {
 		ContainerName: name,
 		Condition:     condition,
 	})
+}
+
+// GetLogDriver returns the log driver used by the container.
+func (c *Container) GetLogDriver() string {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return ""
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encountered error when trying to get log driver for container %s: %v", err)
+		return ""
+	}
+
+	return hostConfig.LogConfig.Type
+}
+
+// GetHostConfig returns the container's host config.
+func (c *Container) GetHostConfig() *string {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.DockerConfig.HostConfig
+}
+
+// GetFirelensConfig returns the container's firelens config.
+func (c *Container) GetFirelensConfig() *FirelensConfig {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.FirelensConfig
 }
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -107,6 +107,14 @@ const (
 	// firelensDriverName is the log driver name for containers that want to use the firelens container to send logs.
 	firelensDriverName = "awsfirelens"
 
+	// firelensConfigVarFmt specifies the format for firelens config variable name. The first placeholder
+	// is option name. The second placeholder is container name.
+	firelensConfigVarFmt = "%s_%s"
+	// firelensConfigVarPlaceholderFmtFluentd and firelensConfigVarPlaceholderFmtFluentbit specify the config var
+	// placeholder format expected by fluentd and fluentbit respectively.
+	firelensConfigVarPlaceholderFmtFluentd   = "\"#{ENV['%s']}\""
+	firelensConfigVarPlaceholderFmtFluentbit = "${%s}"
+
 	// awsExecutionEnvKey is the key of the env specifying the execution environment.
 	awsExecutionEnvKey = "AWS_EXECUTION_ENV"
 	// ec2ExecutionEnv specifies the ec2 execution environment.
@@ -324,14 +332,19 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	// Adds necessary Pause containers for sharing PID or IPC namespaces
 	task.addNamespaceSharingProvisioningDependency(cfg)
 
-	if task.hasFirelensContainer() {
-		return task.applyFirelensSetup(cfg, resourceFields)
+	firelensContainer := task.getFirelensContainer()
+	if firelensContainer != nil {
+		err = task.applyFirelensSetup(cfg, resourceFields, firelensContainer)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
 
-func (task *Task) applyFirelensSetup(cfg *config.Config, resourceFields *taskresource.ResourceFields) error {
-	err := task.initializeFirelensResource(cfg, resourceFields)
+func (task *Task) applyFirelensSetup(cfg *config.Config, resourceFields *taskresource.ResourceFields,
+	firelensContainer *apicontainer.Container) error {
+	err := task.initializeFirelensResource(cfg, resourceFields, firelensContainer)
 	if err != nil {
 		return apierrors.NewResourceInitError(task.Arn, err)
 	}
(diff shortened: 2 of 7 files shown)


agent/api/container/container.go
```diff
@@ -925,3 +928,40 @@ func (c *Container) AddContainerDependency(name string, condition string) {
...
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encounter error when trying to get log driver for container %s: %v", err)
+		return ""
+	}
+
+	return hostConfig.LogConfig.Type
+}
+
+// HasSecret returns whether a container has a secret with a certain provider and target type.
```
yumex93: We already have a helper function HasSecretAsEnvOrLogDriver, it would be better to combine these two functions since both of them are used for checking secret type or target.

fenxiong: but these two are doing different things.. i'm not sure how to combine them?

fenxiong:
deleted by mistake.. will address comment

agent/api/task/task.go
```diff
@@ -823,10 +872,12 @@ func (task *Task) addFirelensContainerDependency() error {
 	return nil
 }
 
-// collectLogOptions collects the log options for all the containers that use the firelens container
+// collectFirelensLogOptions collects the log options for all the containers that use the firelens container
 // as the log driver.
-func (task *Task) collectLogOptions() (map[string]map[string]string, error) {
-	containerToLogOptions := make(map[string]map[string]string)
+func (task *Task) collectFirelensLogOptions(containerToLogOptions map[string]map[string]string) error {
```
fierlion: can you add a simple example of this nested map return value -- as I'm reading over this code I get tangled up... actually I get tangled above where this is used `containerToLogOptions := make(map[string]map[string]string)` but I think the comment would be nice here.

agent/api/task/task.go
```diff
@@ -836,15 +887,55 @@ func (task *Task) collectLogOptions() (map[string]map[string]string, error) {
...
-	return containerToLogOptions, nil
+	return nil
+}
+
+// collectFirelensLogEnvOptions collects all the log secret options. Each secret log option will have a value
+// of a config file variable (e.g. "${config_var_name}") and we will pass the secret value as env to the firelens
+// container and it will resolve the config file variable from the env.
+// Each config variable name has a format of log-option-key_container-name. We need the container name because options
+// from different containers using awsfirelens log driver in a task will be presented in the same firelens config file.
+func (task *Task) collectFirelensLogEnvOptions(containerToLogOptions map[string]map[string]string, firelensConfigType string) error {
```
fierlion: we have the function above `initializeFirelensResource()` which initializes the log options map and then populates it with`task.collectFirelensLogOptions(containerToLogOptions)`.

Is this doing some of that same work? 

I'd like to see these two functions more clearly differentiated, with any common functionality pulled out.  This will help me as a reader understand how initialize and collect work together.

agent/api/task/task.go
```diff
@@ -2067,6 +2158,12 @@ func (task *Task) PopulateSecrets(hostConfig *dockercontainer.HostConfig, contai
 		asmRes = resource[0].(*asmsecret.ASMSecretResource)
 	}
 
+	populateContainerSecrets(hostConfig, container, ssmRes, asmRes)
+	return nil
+}
+
+func populateContainerSecrets(hostConfig *dockercontainer.HostConfig, container *apicontainer.Container,
```
yumex93: why this is separated to another function. I don't feel it is necessary here.

agent/api/task/task.go
fenxiong: not sure what's your concern here.. initializeFirelensResource is the caller of collectFirelensLogOptions. i will remove the map initialization in the callee since that seems to make you confused..

agent/api/task/task.go
fenxiong: it exceeds the complexity limit of the gocyclo target if i don't do this separation

agent/api/task/task.go
fierlion: I see this in the PR detail now.  That's good enough.

agent/api/task/task.go
fenxiong: updated

agent/api/task/task.go
```diff
@@ -675,7 +685,28 @@ func (task *Task) initializeSSMSecretResource(credentialsManager credentials.Man
 				resourcestatus.ResourceStatus(ssmsecret.SSMSecretCreated),
 				apicontainerstatus.ContainerCreated)
 		}
+
+		// Firelens container needs to depends on secret if other containers use secret log options.
+		if container.FirelensConfig != nil && task.firelensDependsOnSecretResource(apicontainer.SecretProviderSSM) {
```
sharanyad: `if container.FirelensConfig != nil` would need a lock. Instead why don't we add a method to Container struct called `isFirelensContainer` and do this check there?

agent/api/task/task.go
```diff
@@ -675,7 +685,28 @@ func (task *Task) initializeSSMSecretResource(credentialsManager credentials.Man
...
+				resourcestatus.ResourceStatus(ssmsecret.SSMSecretCreated),
+				apicontainerstatus.ContainerCreated)
+		}
+	}
+}
+
+// firelensDependsOnSecret checks whether the firelens container needs to depends on a secret resource of
+// a certain provider type.
+func (task *Task) firelensDependsOnSecretResource(secretProvider string) bool {
+	f := func(s apicontainer.Secret) bool {
```
sharanyad: can you name this something meaningful instead of `f`?

fenxiong: updated

agent/api/container/container.go
```diff
@@ -925,3 +928,40 @@ func (c *Container) AddContainerDependency(name string, condition string) {
...
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return ""
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encounter error when trying to get log driver for container %s: %v", err)
```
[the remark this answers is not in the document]
fenxiong: updated

agent/api/task/task.go
```diff
@@ -324,14 +332,16 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	// Adds necessary Pause containers for sharing PID or IPC namespaces
 	task.addNamespaceSharingProvisioningDependency(cfg)
 
-	if task.hasFirelensContainer() {
-		return task.applyFirelensSetup(cfg, resourceFields)
+	firelensContainer := task.getFirelensContainer()
+	if firelensContainer != nil {
+		return task.applyFirelensSetup(cfg, resourceFields, firelensContainer)
```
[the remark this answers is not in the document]
fenxiong: updated

agent/api/task/task.go
fenxiong: i ended up adding GetFirelensConfig to protect it with lock since there's some place i want to look at the config

agent/api/container/container.go
```diff
@@ -871,19 +874,20 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
 	}
 }
 
-func (c *Container) HasSecretAsEnvOrLogDriver() bool {
+func (c *Container) HasSecret(f func(s Secret) bool) bool {
```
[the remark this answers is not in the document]
fenxiong: added

agent/api/container/container.go
```diff
@@ -925,3 +929,36 @@ func (c *Container) AddContainerDependency(name string, condition string) {
...
+}
+
+func (c *Container) GetHostConfig() *string {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.DockerConfig.HostConfig
+}
+
+func (c *Container) GetFirelensConfig() *FirelensConfig {
```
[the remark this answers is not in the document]
fenxiong: fixed

agent/api/task/task.go
```diff
@@ -2106,9 +2210,84 @@ func (task *Task) PopulateSecrets(hostConfig *dockercontainer.HostConfig, contai
...
+// PopulateSecretLogOptionsToFirelensContainer collects secret log option values for awsfirelens log driver from task
+// resource and specified then as envs of firelens container. Firelens container will use the envs to resolve config
+// file variables constructed for secret log options when loading the config file.
+func (task *Task) PopulateSecretLogOptionsToFirelensContainer(firelensContainer *apicontainer.Container) *apierrors.DockerClientConfigError {
+	firelensENVs := make(map[string]string)
+
+	var ssmRes *ssmsecret.SSMSecretResource
+	var asmRes *asmsecret.ASMSecretResource
+
+	resource, ok := task.getSSMSecretsResource()
```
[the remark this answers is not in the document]
fenxiong: i think we always have 1 resource per type, but task.ResourcesMapUnsafe stores each type of resource as a list so this is a list..

agent/api/task/task.go
```diff
@@ -836,15 +891,51 @@ func (task *Task) collectLogOptions() (map[string]map[string]string, error) {
...
+	switch firelensConfigType {
+	case firelens.FirelensConfigTypeFluentd:
+		placeholderFmt = firelensConfigVarPlaceholderFmtFluentd
+	case firelens.FirelensConfigTypeFluentbit:
+		placeholderFmt = firelensConfigVarPlaceholderFmtFluentbit
+	default:
+		return errors.Errorf("unsupported firelens config type %s", firelensConfigType)
+	}
+
+	for _, container := range task.Containers {
```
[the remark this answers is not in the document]
fenxiong: ok, thanks for catching this. i will do it in a later pr

agent/api/task/task_linux_test.go
```diff
@@ -854,6 +909,90 @@ func TestAddFirelensContainerBindMounts(t *testing.T) {
...
+}
+
+func TestPopulateSecretLogOptionsToFirelensContainer(t *testing.T) {
+	task := getFirelensTask(t)
+	ssmRes := &ssmsecret.SSMSecretResource{}
+	ssmRes.SetCachedSecretValue("secret-value-from_us-west-2", "secret-val")
+	task.AddResource(ssmsecret.ResourceName, ssmRes)
+
+	assert.Nil(t, task.PopulateSecretLogOptionsToFirelensContainer(task.Containers[1]))
+	assert.Equal(t, 2, len(task.Containers[1].Environment))
```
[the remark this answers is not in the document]
fenxiong: didn't know about that.. changed to assert.Len for the tests touched by this PR

PR #2153, opened 2019-08-06
Add new EI capability for checking cpu AVX/SSE support

yumex93:
### Summary

In order to not impact performance of DL job, it requires CPU has support for AVX, AVX2, SSE4.1 and SSE4.2 instruction sets. Agent will check for this and advertise as a new capability so that Backend can place tasks accordingly. 

### Implementation details

1. Whether each processor supports the required instruction set is reflected in /proc/cpuinfo file as 'flags' or 'Features'. 
2. Normally for all processors on the same host, they should have same values for 'flags' or 'Features'. However, there is possibility that customers may change some configuration which make it different. Here, I merge the values from all processors. (There are some examples in the added testdata folder).

### Testing

1. Add new unit tests
2. Build agent and manually test on multiple platforms: AL1, AL2, ARM

New tests cover the changes: 
yes
### Licensing

### DIFF
--- a/agent/app/agent_capability_unix.go
+++ b/agent/app/agent_capability_unix.go
@@ -1,6 +1,6 @@
 // +build linux
 
-// Copyright 2014-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2014-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License"). You may
 // not use this file except in compliance with the License. A copy of the
@@ -24,8 +24,17 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	"github.com/aws/amazon-ecs-agent/agent/ecscni"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
+)
+
+const (
+	AVX         = "avx"
+	AVX2        = "avx2"
+	SSE41       = "sse4_1"
+	SSE42       = "sse4_2"
+	CpuInfoPath = "/proc/cpuinfo"
 )
 
 func (agent *ecsAgent) appendVolumeDriverCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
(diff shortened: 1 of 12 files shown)


agent/utils/cpuinfo_test.go
```diff
@@ -0,0 +1,61 @@
+// +build linux,unit
```
petderek: why is there a linux build tag on this but not on the actual file?

agent/utils/cpuinfo.go
```diff
@@ -0,0 +1,70 @@
...
+
+type CPUInfo struct {
+	Processors []Processor `json:"processors"`
+}
+
+type Processor struct {
+	Flags []string `json:"flags"`
+}
+
+var infoRegExp = regexp.MustCompile("([^:]*?)\\s*:\\s*(.*)$")
```
petderek: Can you add a sample line and some explanatory text

agent/app/agent_capability_unix_test.go
```diff
@@ -702,6 +730,104 @@ func TestTaskEIACapabilitiesUnix(t *testing.T) {
...
+	expectedCapabilityNames := []string{
+		"com.amazonaws.ecs.capability.docker-remote-api.1.17",
+	}
+
+	var expectedCapabilities []*ecs.Attribute
+	for _, name := range expectedCapabilityNames {
+		expectedCapabilities = append(expectedCapabilities,
+			&ecs.Attribute{Name: aws.String(name)})
+	}
+	expectedCapabilities = append(expectedCapabilities,
```
petderek: nit: for this test, can we just do the relevant capabilities? And do an assert.Contains? otherwise we need to update this list in the future.

agent/app/agent_capability_unix.go
```diff
@@ -112,5 +121,35 @@ func (agent *ecsAgent) appendAppMeshCapabilities(capabilities []*ecs.Attribute)
...
+	capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+taskEIAAttributeSuffix)
+
+	eiaRequiredFlags := []string{AVX, AVX2, SSE41, SSE42}
+	cpuInfo, err := utils.ReadCPUInfo(CpuInfoPath, agent.ioutil)
+	if err != nil {
+		seelog.Warnf("Unable to read cpuinfo: %v", err)
+		return capabilities
+	}
+
+	// Merge all processors' flags
```
petderek: nit: should the utility package return a merged set of flags? Just a thought.

agent/app/agent_capability_unix.go
```diff
@@ -112,5 +121,35 @@ func (agent *ecsAgent) appendAppMeshCapabilities(capabilities []*ecs.Attribute)
...
+	flagSet := map[string]bool{}
+	for _, proc := range cpuInfo.Processors {
+		for _, flag := range proc.Flags {
+			flagSet[flag] = true
+		}
+	}
+
+	missRequiredFlags := []string{}
```
petderek: 'missingFlags'

agent/app/agent_capability_unix.go
```diff
@@ -112,5 +121,35 @@ func (agent *ecsAgent) appendAppMeshCapabilities(capabilities []*ecs.Attribute)
...
+	for _, requiredFlag := range eiaRequiredFlags {
+		if _, ok := flagSet[requiredFlag]; !ok {
+			missRequiredFlags = append(missRequiredFlags, requiredFlag)
+		}
+	}
+
+	if len(missRequiredFlags) > 0 {
+		seelog.Warnf("Missing cpu flags for eia support: %v", strings.Join(missRequiredFlags, ","))
```
petderek: `Infof`

For many customers this won't need to be loud

petderek:
Actually. Scratch that:

https://godoc.org/golang.org/x/sys/cpu

agent/utils/cpuinfo.go
yumex93: I will get rid of regex to make it easy to read and maintain

agent/app/agent_capability_unix_test.go
yumex93: Why doing contains instead of equal?

agent/app/agent_capability_unix.go
yumex93: I will add another helper function to do this.

agent/app/agent_capability_unix_test.go
petderek: Otherwise you need to modify this list every time we add new capabilities. We can instead only focus on the relevant capabilities

agent/utils/cpuinfo.go
```diff
@@ -0,0 +1,95 @@
...
+	var lines []string
+
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+
+	var cpuinfo = CPUInfo{}
+	var processor = &Processor{}
+
+	for i, line := range lines {
```
petderek: You should be able to do this with only one scan loop without using the “lines” variable.

yumex93: Discussed offline. Scanner does not have an elegant way to meet my requirements. Will still keep lines variable here.

PR #2161, opened 2019-08-08
phase 1: auto restart non-essential contianer & integ tests

jiapent:
### Summary

Add auto restart non-essential container feature and integration tests.
### Implementation details

Add a new container status: "ContainerRestarting" which represents a container is waiting for restarting. 
### Testing

Integration tests for different restart policies are implemented. 

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -16,9 +16,12 @@ package container
 import (
 	"encoding/json"
 	"fmt"
+	"math"
 	"strconv"
 	"sync"
 	"time"
+
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
 
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	apierrors "github.com/aws/amazon-ecs-agent/agent/api/errors"
@@ -69,7 +72,40 @@ const (
 
 	// TargetLogDriver is to show secret target being "LOG_DRIVER", the default will be "CONTAINER"
 	SecretTargetLogDriver = "LOG_DRIVER"
+
+	// Default RestartMaxAttempts OnFailure
+	DefaultRestartMaxAttemptsOnFailure = math.MaxInt32
 )
+
+// containerNextKnownStateProgressionMap represents the map from container state to next progression state
+var containerNextKnownStateProgressionMap = map[apicontainerstatus.ContainerStatus]apicontainerstatus.ContainerStatus{
+	apicontainerstatus.ContainerStatusNone: apicontainerstatus.ContainerPulled,
+	apicontainerstatus.ContainerPulled:     apicontainerstatus.ContainerCreated,
+	apicontainerstatus.ContainerCreated:    apicontainerstatus.ContainerRunning,
+	apicontainerstatus.ContainerRestarting: apicontainerstatus.ContainerRunning,
+	apicontainerstatus.ContainerRunning:    apicontainerstatus.ContainerResourcesProvisioned,
+	apicontainerstatus.ContainerStopped:    apicontainerstatus.ContainerZombie,
+}
+
+// RestartPolicyMap represents the map from restart policy string to internal constant
+var RestartPolicyMap = map[string]RestartPolicy{
+	"NEVER":               Never,
+	"UNLESS_TASK_STOPPED": UnlessTaskStopped,
+	"ON_FAILURE":          OnFailure,
+}
+
+const (
+	// not restarting at all
+	Never RestartPolicy = iota
+	// always restart container regardless of exit code and max retries unless the task is stopped
+	UnlessTaskStopped
+	// restarts a container if exit code is non-zero
+	OnFailure
+)
+
+type RestartPolicy int32
+
+type RestartCount int32
 
 // DockerConfig represents additional metadata about a container to run. It's
 // remodeled from the `ecsacs` api model file. Eventually it should not exist
@@ -93,6 +129,31 @@ type HealthStatus struct {
 	ExitCode int `json:"exitCode,omitempty"`
 	// Output is the output of health check
 	Output string `json:"output,omitempty"`
+}
+
+type RestartInfo struct {
+	// RestartPolicy define in what condition will container be restarted
+	RestartPolicy RestartPolicy
+	// max restart retries if restarts 'On-Failure'
+	RestartMaxAttempts RestartCount
+	// current retries used
+	RestartAttempts RestartCount
+	// auto restart exponential backoff
+	RestartBackoff *retry.ExponentialBackoff
+
+	// ForceRestart is true if we need to restart container due to api call
+	// (`Docker container start`) error. This is  used when ever "Docker Start" has an error,
+	// in this situation we should restart the container not considering the exit code.
+	ForceRestart bool
+
+	// ForceStop is true if we are forcing a container to stop due to error,
+	// so won't try to restart. This field is used in 2 situations:
+	// 1. Other Docker api calls have error, e.g. "Docker Create" has an error and we are not trying to restart it.
+	// 2. The task is stopped and every restarting container should not try to restart but drive to stop.
+	//
+	// Typically, `ForceRestart` and `ForceStop` are
+	// both false if no error occurs calling Docker apis and the task is not desired to stop.
+	ForceStop bool
 }
 
 // Container is the internal representation of a container in the ECS agent
@@ -132,6 +193,9 @@ type Container struct {
 	Secrets []Secret `json:"secrets"`
 	// Essential denotes whether the container is essential or not
 	Essential bool
+	// Restart information for container
+	RestartInfo *RestartInfo
+
 	// EntryPoint is entrypoint of the container, corresponding to docker option: --entrypoint
 	EntryPoint *[]string
 	// Environment is the environment variable set in the container
@@ -454,8 +518,9 @@ func (c *Container) IsKnownSteadyState() bool {
 
 // GetNextKnownStateProgression returns the state that the container should
 // progress to based on its `KnownState`. The progression is
-// incremental until the container reaches its steady state. From then on,
-// it transitions to `ContainerStopped`.
+// incremental until the container reaches its steady state. (The next state of
+// `ContainerCreated` and `ContainerRestarting` will both be 'ContainerRunning`.)
+// From then on, it transitions to `ContainerStopped`.
 //
 // For example:
 // a. if the steady state of the container is defined as `ContainerRunning`,
@@ -474,7 +539,7 @@ func (c *Container) GetNextKnownStateProgression() apicontainerstatus.ContainerS
 		return apicontainerstatus.ContainerStopped
 	}
 
-	return c.GetKnownStatus() + 1
+	return containerNextKnownStateProgressionMap[c.GetKnownStatus()]
 }
 
 // IsInternal returns true if the container type is `ContainerCNIPause`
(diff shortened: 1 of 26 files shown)


agent/api/task/task.go
```diff
@@ -1505,6 +1545,10 @@ func (task *Task) updateContainerDesiredStatusUnsafe(taskDesiredStatus apitaskst
 	for _, container := range task.Containers {
 		taskDesiredStatusToContainerStatus := apitaskstatus.MapTaskToContainerStatus(taskDesiredStatus, container.GetSteadyStateStatus())
 		if container.GetDesiredStatus() < taskDesiredStatusToContainerStatus {
+			if taskDesiredStatusToContainerStatus == apicontainerstatus.ContainerStopped &&
+				container.IsAutoRestartNonEssentialContainer() {
+				container.DesiredToFullyStopWhenReceivingStopped = true
```
fenxiong: needs lock? same for all other places that are setting this..

agent/engine/task_manager.go
```diff
@@ -53,6 +53,18 @@ const (
...
 )
 
+const (
+	// source for dockerContainerChange
+
+	// container transitions without any Docker actions, e.g. a `Created` container's desired status is set as `Stopped`
+	none ContainerChangeSource = iota
+	fromDockerEvent
+	fromInspect
+	fromDockerApi // from 'Docker start container' or 'Docker stop container'
```
fenxiong: i think we inspect container after start container and stop container.. so i don't think you need a separate one for this?

agent/engine/task_manager.go
```diff
@@ -449,6 +508,30 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	}
 }
 
+// check if we need to restart the container
+func shouldRestartContainer(container *apicontainer.Container, containerChange dockerContainerChange) bool {
```
fenxiong: i think this should be a method of container

agent/api/container/container.go
```diff
@@ -127,6 +145,18 @@ type Container struct {
 	Secrets []Secret `json:"secrets"`
 	// Essential denotes whether the container is essential or not
 	Essential bool
+	// RestartPolicy define in what condition will container be restarted
+	RestartPolicy RestartPolicy
+	// max restart retries if restarts 'On-Failure'
+	RestartMaxAttempts RestartCount
```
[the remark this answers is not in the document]
jiapent: Cause it can never been negative, so I used uint.

yunhee-l:
Travis is failing with make: 
*** [gocyclo] Error 1
The command "make static-check" exited with 2.

Can you run make static-check and fix the issue?

agent/api/container/container.go
```diff
@@ -66,8 +68,24 @@ const (
 
 	// TargetLogDriver is to show secret target being "LOG_DRIVER", the default will be "CONTAINER"
 	SecretTargetLogDriver = "LOG_DRIVER"
+
+	// Default RestartMaxAttempts OnFailure
+	DefaultRestartMaxAttemptsOnFailure = math.MaxUint32
```
yunhee-l: is this a sensible default value? 1<<32 - 1 is a big number, practically attempting to restart forever?

agent/api/container/container.go
```diff
@@ -127,6 +145,18 @@ type Container struct {
...
+	// RestartPolicy define in what condition will container be restarted
+	RestartPolicy RestartPolicy
+	// max restart retries if restarts 'On-Failure'
+	RestartMaxAttempts RestartCount
+	// current retries used
+	RestartAttempts RestartCount
+	// auto restart exponential backoff
+	RestartBackoff *retry.ExponentialBackoff
+	// DesiredToFullyStop is true if we are forcing a container to stop due to error, so won't try to restart
+	DesiredToFullyStopWhenReceivingStopped bool
```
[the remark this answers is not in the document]
jiapent: `DesiredToRestartWhenReceivingStopped` is used when ever "Docker Start" has an error, in this situation we should restart the container not considering the exit code. `DesiredToFullyStopWhenReceivingStopped` is used in 2 situations: 1. Other Docker api calls have error, e.g. "Docker Create" has an error and we are not trying to restart it. 2. The task is stopped and every restarting container should not try to restart but drive to stop.  

Typically, these to fields are both false if no error occurs calling Docker apis and the task is not desired to stop.

Another example of these two fields both being true: A container has error starting it, and we set `DesiredToRestartWhenReceivingStopped` to be true, and call Docker api to stop it. Before we receiver the "Stopped" response, an essential of the same task may be stopped, and this container's `DesiredToFullyStopWhenReceivingStopped` is set to be true. When the "Stopped" message come back, we are not trying to restart it.

agent/api/container/container.go
```diff
@@ -910,3 +945,56 @@ func (c *Container) GetStopTimeout() time.Duration {
...
+
+func (c *Container) IncrementRestartAttempts() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.RestartAttempts += 1
+}
+
+func (c *Container) CanMakeRestartAttempt() bool {
+	c.lock.Lock()
+	defer c.lock.Unlock()
```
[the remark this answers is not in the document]
jiapent: K

agent/api/container/container.go
```diff
@@ -910,3 +945,56 @@ func (c *Container) GetStopTimeout() time.Duration {
...
+}
+
+func (c *Container) IsAutoRestartNonEssentialContainer() bool {
+	return !c.IsEssential() && c.RestartPolicy != NEVER
+}
+
+func (c *Container) IsDesiredToRestartWhenReceivingStopped() bool {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	defer func() {
```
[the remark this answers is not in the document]
jiapent: This is used for checking if `DesiredToRestartWhenReceivingStopped` is true, and if it's true set it as false. May be I can remove the lock, since we are not calling `handleContainerChange()` asynchronously and the flag is only set in this function?

agent/api/container/container.go
```diff
@@ -127,6 +145,18 @@ type Container struct {
 	Secrets []Secret `json:"secrets"`
 	// Essential denotes whether the container is essential or not
 	Essential bool
+	// RestartPolicy define in what condition will container be restarted
```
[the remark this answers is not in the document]
jiapent: You're right, will refactor this.

agent/api/container/container.go
```diff
@@ -463,6 +494,10 @@ func (c *Container) GetNextKnownStateProgression() apicontainerstatus.ContainerS
 		return apicontainerstatus.ContainerStopped
 	}
 
+	if c.GetKnownStatus() == apicontainerstatus.ContainerCreated {
```
sharanyad: this doesn't seem to be the best solution, when more non linear states are added to the list. 

One way to handle this would be to maintain map of current state to next state information, and return that here. then we do not have to maintain logic specific to different states here.

agent/api/task/task.go
jiapent: Yeah, will use a Set function with lock instead.

agent/api/container/container.go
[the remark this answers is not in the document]
sharanyad: could you please add these as comments in the code as well?

agent/engine/task_manager.go
jiapent: We do inspect for every start and stop calls, and also Docker events to get Metadata. Actually I'm handling `fromInspect` or `fromDockerAPI` the same way. But is it better to keep both so it's more clear what the source of the ContainerChange is?

agent/engine/task_manager.go
jiapent: Will refactor this.

agent/api/container/container.go
jiapent: `RestartMaxAttempts` is only valid for policy: `OnFailure`. If customer doesn't specify `RestartMaxAttempts`, we'll keep restarting it forever unless exit successfully.

agent/api/container/container.go
jiapent: Cool, will refactor this.

agent/api/container/container.go
```diff
@@ -910,3 +945,56 @@ func (c *Container) GetStopTimeout() time.Duration {
...
+	return c.RestartAttempts
+}
+
+func (c *Container) IncrementRestartAttempts() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.RestartAttempts += 1
+}
+
+func (c *Container) CanMakeRestartAttempt() bool {
```
shubham2892: Can rename this `CanRestart`?

agent/api/container/container.go
```diff
@@ -66,8 +68,24 @@ const (
...
 	// TargetLogDriver is to show secret target being "LOG_DRIVER", the default will be "CONTAINER"
 	SecretTargetLogDriver = "LOG_DRIVER"
+
+	// Default RestartMaxAttempts OnFailure
+	DefaultRestartMaxAttemptsOnFailure = math.MaxUint32
 )
 
+const (
+	// not restarting at all
+	NEVER RestartPolicy = iota
```
shubham2892: Any specific reason NEVER is all caps?

agent/api/container/container.go
```diff
@@ -910,3 +945,56 @@ func (c *Container) GetStopTimeout() time.Duration {
...
+func (c *Container) SetDesiredToRestartWhenReceivingStopped() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.DesiredToRestartWhenReceivingStopped = true
+}
+
+func (c *Container) SetDefaultRestartMaxAttemptsOnFailure() {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	if c.RestartPolicy == OnFailure && c.RestartMaxAttempts == 0 {
```
shubham2892: Why are we setting this to `RestartMaxAttempts` to `DefaultRestartMaxAttemptsOnFailure` when it 0?

agent/api/task/task.go
```diff
@@ -955,6 +981,9 @@ func (task *Task) getEarliestKnownTaskStatusForContainers() apitaskstatus.TaskSt
 	// Set earliest container status to an impossible to reach 'high' task status
 	earliest := apitaskstatus.TaskZombie
 	for _, container := range task.Containers {
+		if !container.IsEssential() && container.GetRestartAttempts() > 0 {
+			continue
```
shubham2892: Can you add debug log here too.

agent/engine/docker_task_engine.go
```diff
@@ -449,11 +450,45 @@ func (engine *DockerTaskEngine) checkTaskState(task *apitask.Task) {
...
 					DockerContainerMetadata: metadata,
 				},
+				source: fromInspect,
+				containerPrevRestartAttempts: restartAttempts,
 			})
 		}
 	}
 }
 
+// checkTaskState inspects the state of input container and writes
```
shubham2892: checkContainerState

agent/engine/docker_task_engine.go
```diff
@@ -449,11 +450,45 @@ func (engine *DockerTaskEngine) checkTaskState(task *apitask.Task) {
...
+	if !ok {
+		return
+	}
+	restartAttempts := container.GetRestartAttempts()
+	status, metadata := engine.client.DescribeContainer(engine.ctx, dockerContainer.DockerID)
+	engine.tasksLock.RLock()
+	managedTask, ok := engine.managedTasks[task.Arn]
+	engine.tasksLock.RUnlock()
+
+	if ok {
```
shubham2892: if !ok, nothing should happen?

Can we throw an error if !ok?

agent/engine/task_manager_test.go
```diff
@@ -373,6 +557,18 @@ func TestContainerNextStateWithTransitionDependencies(t *testing.T) {
 			expectedTransitionActionable: false,
 			reason:                       dependencygraph.ErrContainerDependencyNotResolved,
 		},
+		// NONE -> RUNNING transition is not allowed and not actionable, when desired is Running and dependency is Restarting
```
shubham2892: is NONE -> RESTARTING possible?

agent/api/container/container.go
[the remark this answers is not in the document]
jiapent: Yes

agent/api/container/container.go
jiapent: Yes

agent/api/container/container.go
jiapent: Will keep this consistent with the other two, also as suggested, `Always` will be renamed as `UnlessTaskStopped`.

agent/api/container/container.go
jiapent: By default, if user doesn't specify `RestartMaxAttempts` for policy: `OnFailure', we'll keep restarting the container if its exit code is not 0.

agent/api/task/task.go
jiapent: Yes.

agent/engine/docker_task_engine.go
jiapent: If !ok, the task of this container is not tracked by Agent. If this happens, I think Agent couldn't do much about this and can just ignore it.

agent/engine/task_manager_test.go
jiapent: This function test the next state of containers which are not in steady states. If `None` is the current status, and `Running` is the desired states, the next state of the container should be `Pulled` and then Agent will call Docker to pull image. Agent will never set a container desired status to be `Restarting`, so `None`-> `Restarting` will not be possible.

agent/engine/task_manager.go
jiapent: Actually this file imports `container.go`, we'll have circular import error by changing this to a method of Container.

fenxiong: why do you need to import task engine in container.go? i think you can just pass some strings as arguments and return a bool

PR #2164, opened 2019-08-12
Fix bug in secret log options for firelens

fenxiong:
### Summary

Previously, container name is appended to firelens config variable name in order to differentiate log options from different containers (see description in https://github.com/aws/amazon-ecs-agent/pull/2150 for more details). For fluentd this doesn't work when container name has hyphen in it, potentially because ruby (fluentd source) doesn't allow hyphen in env name.. Fix by using the index of the container instead.

### Implementation details

Change firelensConfigVarFmt and fill in index. The index doesn't count internal container so that we don't need to worry about potentially getting different index because an internal container is injected (internal container doesn't use firelens anyway).

### Testing

Existing unit tests modified for the change. Manually ran task and verified the fix.

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -108,8 +108,12 @@ const (
 	firelensDriverName = "awsfirelens"
 
 	// firelensConfigVarFmt specifies the format for firelens config variable name. The first placeholder
-	// is option name. The second placeholder is container name.
-	firelensConfigVarFmt = "%s_%s"
+	// is option name. The second placeholder is the index of the container in the task's container list, appended
+	// for the purpose of avoiding config vars from different containers within a task collide (note: can't append
+	// container name here because it may contain hyphen which will break the config var resolution (see PR 2164 for
+	// details), and can't append container ID either because we need the config var in PostUnmarshalTask, which is
+	// before all the containers being created).
+	firelensConfigVarFmt = "%s_%d"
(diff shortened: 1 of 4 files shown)


fenxiong:
> Container names are not unique?

They are unique. it's just that they may have hyphen so can't be appended to config variable name.

agent/api/task/task.go
```diff
@@ -2374,3 +2385,21 @@ func (task *Task) AssociationByTypeAndName(associationType, associationName stri
...
 	return nil, false
 }
+
+// GetContainerIndex returns the index of the container in the container list. This doesn't count internal container.
+func (task *Task) GetContainerIndex(containerName string) int {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	idx := 0
+	for _, container := range task.Containers {
```
petderek: you can do the same thing by doing:
for idx, contianer

fenxiong: i'm trying to skip counting internal container..

sharanyad: if the index actually doesn't matter (since it just needs to be unique), for loop's index can just be used.

fenxiong: GetContainerIndex is invoked two times, one is during PostUnmarshalTask, the other is createContainer, and the index needs to remain the same for those two. i'm trying to avoid an assumption that no internal container is injected between these two. lmk if that doesn't make sense..

PR #2168, opened 2019-08-14
Change fluent tag format for Firelens

PettitWesley:
### Summary
Change the fluent tag format for FireLens.

Container logs from Docker will have a tag in the format of `container_name-firelens-taskID`. This is because the tag will become part of the log stream name in the CloudWatch plugin, so the Task ID must be in there so that each task has a unique stream.

The filters/outputs will match with the expression `container_name-firelens*`. This is done so that customers can use the Fluent Logger SDKs (in the future, once we enable that) and tag their logs with `container_name-firelens*` if they want to use the outputs configured in the task definition.
### Testing

N/A
### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -16,6 +16,7 @@ package engine
 
 import (
 	"context"
+	"fmt"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -82,6 +83,9 @@ const (
 	logDriverAsyncConnect   = "fluentd-async-connect"
 	dataLogDriverSocketPath = "/socket/fluent.sock"
 	socketPathPrefix        = "unix://"
+
+	// fluentTagDockerFormat is the format for the log tag, which is "containerName-firelens-taskID"
+	fluentTagDockerFormat = "%s-firelens-%s"
 )
 
 // DockerTaskEngine is a state machine for managing a task and its containers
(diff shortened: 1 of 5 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -82,6 +83,9 @@ const (
 	logDriverAsyncConnect   = "fluentd-async-connect"
 	dataLogDriverSocketPath = "/socket/fluent.sock"
 	socketPathPrefix        = "unix://"
+
+	// fluentTagDockerFormat is the format for the log tag, which is container name with "-firelens-taskID" appended
```
petderek: "containerName-firelens-taskID"

agent/taskresource/firelens/firelensconfig_unix.go
```diff
@@ -56,6 +55,9 @@ const (
 
 	// socketPath is the path for socket file.
 	socketPath = "/var/run/fluent.sock"
+
+	// fluentTagOutputFormat is the format for the log tag, which is container name with "-firelens" appended
+	fluentTagOutputFormat = "%s-firelens*"
```
petderek: why did we make this less specific

PettitWesley: See the description of the PR- it makes it so that later on customers can send logs to these task def configured outputs using the fluent logger SDKs. If the output matches the task ID, then their code would have to know the task ID (which is possible but annoying).

PR #2172, opened 2019-08-16
add agent and docker version to tcs ws URL

sharanyad:
### Summary
adds agent hash, agent and docker version to tcs ws URL
refactor tests to use assert 

### Testing

### Licensing

### DIFF
--- a/agent/tcs/handler/handler.go
+++ b/agent/tcs/handler/handler.go
@@ -25,6 +25,7 @@ import (
 	tcsclient "github.com/aws/amazon-ecs-agent/agent/tcs/client"
 	"github.com/aws/amazon-ecs-agent/agent/tcs/model/ecstcs"
 	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
+	"github.com/aws/amazon-ecs-agent/agent/version"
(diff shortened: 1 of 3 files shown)


agent/tcs/handler/handler.go
```diff
@@ -179,13 +181,18 @@ func anyMessageHandler(client wsclient.ClientServer) func(interface{}) {
...
 	tcsURL := endpoint
 	if !strings.HasSuffix(tcsURL, "/") {
 		tcsURL += "/"
 	}
 	query := url.Values{}
 	query.Set("cluster", cluster)
 	query.Set("containerInstance", containerInstance)
+	query.Set("agentVersion", version.Version)
+	query.Set("agentHash", version.GitHashString())
+	if dockerVersion, err := taskEngine.Version(); err == nil {
```
yumex93: If there is an error and we don't set dockerVersion, will the dockerVersion just be empty string in TACS or actually there will be no this attribute?

sharanyad: no attribute. this is inspired from ACS -
https://github.com/aws/amazon-ecs-agent/blob/f95f731bce74e6276f9fd8bfbb87a62d3a790ce8/agent/acs/handler/acs_handler.go#L434-L436

PR #2176, opened 2019-08-21
Propagate task and container instance tags to the introspection endpoint

xlucas:
### Summary

It propagates task and container instance tags to the introspection endpoint as proposed in #2122 .

### Implementation details

A new route `/v1/tasksWithTags` is added, similarly to what has been done for v2 and v3 metadata endpoints.

### Testing

I ran both unit and integration tests on Linux.
I also [built an image](https://hub.docker.com/r/xlucas/amazon-ecs-agent/tags) with these changes and performed tests in my ECS cluster.

(_Note_: an unrelated unit test, this [one](https://github.com/aws/amazon-ecs-agent/blob/f95f731bce74e6276f9fd8bfbb87a62d3a790ce8/agent/ecscni/plugin_test.go#L488) is failing due to a wrong parsing logic of `git submodule`. On my machine this outputs 

```
 91ccefc8864ec14a32bd2b9d7e7de3060b685383 amazon-ecs-cni-plugins (remotes/origin/HEAD)
-fa071b269ea57644ab566cc915698c61369e158f amazon-vpc-cni-plugins
```

So [this line](https://github.com/aws/amazon-ecs-agent/blob/f95f731bce74e6276f9fd8bfbb87a62d3a790ce8/agent/ecscni/plugin_test.go#L498) fails as it compares `amazon-vpc-cni-plugins` against the hash, whilst for the first submodule there's an extra leading space so the hash comparison is correct in this case.)

Feature - Propagate task and container instance tags to the Introspection endpoint

### Licensing


petderek:
Thanks for opening this. Unfortunately this isn't something we will want to be generally available in agent for two main reasons:
 
1. We don't want a long-term dependency on this API
2. We don't want to modify the managed role to include ListTagsForResource by default
 
The permissions issue is a bit more of the hard blocker here, since adding generalized access to the ListTagsForResource api would allow agent to retrieve the tags for any resource in a user's account. We have some of the code in v2/v3 already, but don't want to encourage its use yet since it requires adding additional permissions to the instance role.
 
We had discussed this internally and intend to stream the tags down to agent as part of the payload that gets sent at run-task time.

xlucas:
@petderek Thanks for the explanation.

Could you just clarify if:

- current (undocumented) v2/v3 URIs returning tags alongside payloads would still be those used in the future even if the internal implementation to gather tags from resources would be subject to change
- having one URI in the introspection endpoint (as proposed in this PR but implemented differently in that case) is still something you consider

Thanks!

irabinovitch:
@petderek Did you have any thoughts on @xlucas 's questions?

petderek:
Hey, apologies for the late response on this:

> current (undocumented) v2/v3 URIs returning tags alongside payloads would still be those used in the future even if the internal implementation to gather tags from resources would be subject to change

We won't change the current implementation even though it isn't document. Part of the reason it is not yet documented is because of the problems we described above.

> having one URI in the introspection endpoint (as proposed in this PR but implemented differently in that case) is still something you consider

What do you mean by one URI? What is the use case you're addressing with that?

xlucas:
@petderek 

> What do you mean by one URI? What is the use case you're addressing with that?

I mean by that what this PR was trying to do, i.e. introducing `/v1/tasksWithTags` or something of the sort at some point. The use case here is that the v1 API is useful to pull information about all tasks running on an instance but it's missing tag propagation in the payload, whereas v2 & v3 give this information but by container.

PR #2179, opened 2019-08-22
 Support external firelens config from local and S3

fenxiong:
### Summary

Support two types of external firelens config - "file" and "s3". For external config type "file", the firelens container config file includes a config file inside the container at a given path. For external config type "s3", agent downloads a config file from S3 and bind mounts it to the firelens container at a given path, and the firelens container config file includes the config file at that path.

#### Packages changed
(1) `agent/s3`: add an s3 interface package that interacts with s3. very similar to agent/ssm and agent/asm.
(2) `agent/taskresource/firelens`: update firelens task resource to support including external firelens config.
(3) `agent/api/task`: Bind mount downloaded s3 config file to firelens container.
(4) miscellaneous: `agent/statemanager` - update state file version; `agent/app` - add new capabilities "ecs.capability.firelens.options.config.file" and "ecs.capability.firelens.options.config.arn" that correspond to the added functionalities.

#### Examples

For local config file, an example is as follow:
(1) Customer specifies task definition like the following:
```
{
       "family": "...",
	...
	"containerDefinitions": [
		{
			...
			"name": "log_router", // firelens container
                        "image": "xx", // an image that has /tmp/dummy.conf inside; or alternative, has the file at this path in a volume
			"firelensConfiguration": {
				"type": "fluentd",
				"options": {
					"config-file-type": "file",
					"config-file-value": "/tmp/dummy.conf"
				}
			},
			...
		 },
		 ...
	]
```
(2) A firelens task resource is initialized as the following:
```
firelensResource {
  externalConfigType: "file",
  externalConfigValue: "/tmp/dummy.conf"
}
```

(3) When creating the firelens task resource, the following section is added to firelens config file:
```
...
@include /tmp/dummy.conf
...
```

For external config file, an example is as follow:
(1) Customer specifies task definition like the following:
```
{
       "family": "...",
	...
	"containerDefinitions": [
		{
			...
			"name": "log_router", // firelens container
                        "image": "xx",
			"firelensConfiguration": {
				"type": "fluentd",
				"options": {
					"config-file-type": "s3",
					"config-file-value": "arn:aws:s3:::bucket/key"
				}
			},
			...
		 },
		 ...
	]
```
(2) A firelens task resource is initialized as the following:
```
firelensResource {
  externalConfigType: "s3",
  externalConfigValue: "arn:aws:s3:::bucket/key"
}
```

(3) When creating the firelens task resource, the agent downloads the config from "arn:aws:s3:::bucket/key", saves it at $(ECS_HOST_DATA_DIR)/data/firelens/config/task-id/external.conf. and the following section is added to firelens config file:
```
...
@include /fluentd/etc/external.conf
...
```
(4) When creating the firelens container, an additional bind mount is specified for container: "$(ECS_HOST_DATA_DIR)/data/firelens/config/task-id/external.conf:/fluentd/etc/external.conf".

#### Note about S3
Agent calls S3 with task execution role credentials. For the region, it infers the region from the bucket that the config file is in, and then uses a client in that region. This assumes the config file is in the same partition. It requires customer's task execution role to have s3:GetBucketLocation permission on the bucket, and s3:GetObject permission on the config file.

### Testing

Manually ran tasks and verified that both types of configs are working for both fluentd and fluentbit. Will add functional test in later PR.

### Licensing


agent/api/task/task.go
```diff
@@ -96,11 +96,18 @@ const (
...
 	// config file bind mount for fluentd and fluentbit firelens container respectively.
 	// First placeholder is host data dir, second placeholder is taskID.
 	firelensConfigBindFormatFluentd   = "%s/data/firelens/%s/config/fluent.conf:/fluentd/etc/fluent.conf"
 	firelensConfigBindFormatFluentbit = "%s/data/firelens/%s/config/fluent.conf:/fluent-bit/etc/fluent-bit.conf"
+
+	// firelensS3ConfigBindFormatFluentd and firelensS3ConfigBindFormatFluentbit specify the format of the bind mount
+	// for the firelens config file downloaded from S3. First placeholder is host data dir, second placeholder is taskID,
+	// third placeholder is the s3 config path inside the firelens container.
+	firelensS3ConfigBindFormatFluentd   = "%s/data/firelens/%s/config/s3.conf:%s"
+	firelensS3ConfigBindFormatFluentbit = "%s/data/firelens/%s/config/s3.conf:%s"
```
PettitWesley: How about `external.conf` or `include.conf` or something besides `s3.conf`. Since we talked about adding support for pulling from SSM in the future.

agent/taskresource/firelens/firelens_unix.go
```diff
@@ -406,6 +489,54 @@ func (firelens *FirelensResource) generateConfigFile() error {
 	return nil
 }
 
+func (firelens *FirelensResource) downloadConfigFromS3() error {
```
PettitWesley: Just for my own understanding: This call comes in the FireLens task resource, which means it blocks the firelens container and any dependent containers from launching, but otherwise its asynchronous, so the agent can launch other containers/etc while the download is happening. Is that correct?

fenxiong: yes, it only blocks the firelens container and containers that use awsfirelens driver. other containers can start running before/during the download.

agent/api/task/task.go
fenxiong: sure. changed to external.conf

agent/taskresource/firelens/firelens_unix.go
```diff
@@ -378,16 +461,62 @@ func (firelens *FirelensResource) generateConfigFile() error {
...
+	confFilePath := filepath.Join(firelens.resourceDir, "config", "external.conf")
+	err = firelens.writeConfigFile(func(file oswrapper.File) error {
+		return s3.DownloadFile(bucket, key, s3DownloadTimeout, file, s3Client)
+	}, confFilePath)
+
+	if err != nil {
+		return errors.Wrapf(err, "unable to download s3 config %s from bucket %s", key, bucket)
+	}
+
+	seelog.Infof("Downloaded firelens config file from s3 and saved to: %s", confFilePath)
```
sharanyad: what's the value of this in info level?

fenxiong: hmm can probably just be debug here.. will change

PR #2180, opened 2019-08-22
Add ECS_SPOT_INSTANCE_DRAINING_ENABLED configuration variable

sparrc:
# Summary

First step implementing https://github.com/aws/containers-roadmap/issues/190

### Testing

make test

Added ECS_SPOT_INSTANCE_DRAINING_ENABLED

### Licensing

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -545,7 +545,7 @@ func environmentConfig() (Config, error) {
 		NvidiaRuntime:                       os.Getenv("ECS_NVIDIA_RUNTIME"),
 		TaskMetadataAZDisabled:              utils.ParseBool(os.Getenv("ECS_DISABLE_TASK_METADATA_AZ"), false),
 		CgroupCPUPeriod:                     parseCgroupCPUPeriod(),
-		SpotInstanceDrainingEnabled:         utils.ParseBool(os.Getenv("ECS_SPOT_INSTANCE_DRAINING_ENABLED"), false),
+		SpotInstanceDrainingEnabled:         utils.ParseBool(os.Getenv("ECS_ENABLE_SPOT_INSTANCE_DRAINING"), false),
 	}, err
 }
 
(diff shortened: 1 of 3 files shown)


agent/config/config.go
```diff
@@ -545,6 +545,7 @@ func environmentConfig() (Config, error) {
 		NvidiaRuntime:                       os.Getenv("ECS_NVIDIA_RUNTIME"),
 		TaskMetadataAZDisabled:              utils.ParseBool(os.Getenv("ECS_DISABLE_TASK_METADATA_AZ"), false),
 		CgroupCPUPeriod:                     parseCgroupCPUPeriod(),
+		SpotInstanceDrainingEnabled:         utils.ParseBool(os.Getenv("ECS_SPOT_INSTANCE_DRAINING_ENABLED"), false),
```
sharanyad: isn't this going to be true by default?
if yes, then let's have the env var named `ECS_DISABLE_SPOT_INSTANCE_DRAINING` and have default as false.

sparrc: PMs actually want this to be 'false' by default because it's a change in the behavior. In documentation it will be recommended to set to true when using spot instances with ECS.

sharanyad: ah ok, I thought it was decided to enable this by default. 
either way, please rename this to `ECS_ENABLE_SPOT_INSTANCE_DRAINING` instead of `ENABLED` as suffix, following the convention - https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html

sparrc: 👍

agent/config/types.go
```diff
@@ -294,4 +294,12 @@ type Config struct {
 
 	// CgroupCPUPeriod is config option to set different CFS quota and period values in microsecond, defaults to 100 ms
 	CgroupCPUPeriod time.Duration
+
+	// SpotInstanceDrainingEnabled, if true, agent will poll the container instance's metadata endpoint for an ec2 spot
+	//   instance termination notice. If EC2 sends a spot termination notice, then agent will set the instance's state
```
yumex93: nit: agent is not actually the one setting the instance state as DRAINING. Should we mention backend side is the one transiting the state to DRAINING here?

sparrc: Agent calls the API that sets the state to DRAINING though

PR #2183, opened 2019-08-25
Parallel pull

yumex93:
### Summary

Based on PR [1971](https://github.com/aws/amazon-ecs-agent/pull/1971), make the following minor changes:
1. Update the stopping waiting timeout to be consistent as other functional tests. Originally it is 1m, during my testing, it sometimes takes longer that 1m before the task being stopped completely. 
2. Decrease the value of memory of each container, since during AMI release, we will run agent tests on instances types with small memory like t2.small. Before changing the value, it fails on this type every time. 
### Testing

Run TestParallelPull, TestOOMContainer and TestOOMTask three test cases across different regions, different instance types on each platform around 200 times. All tests passed. 

New tests cover the changes: 

### Licensing


agent/functional_tests/tests/generated/simpletests_unix/simpletests_generated_unix_test.go
```diff
@@ -1149,7 +1149,7 @@ func TestParallelPull(t *testing.T) {
 		}
 	}
 
-	timeout, err := time.ParseDuration("1m")
+	timeout, err := time.ParseDuration("2m")
```
petderek: this is generated -- you also need to update the config file

PR #2181, opened 2019-08-23
Add network support for firelens

mssrivas:
Summary
Support two types of network mode - bridge and AWSVPC. 
To enable networking we need to add teo env variables FLUENT_HOST and FLUENT_PORT. The value of FLUWNT_PORT is 24224 and the value of host will be 127.0.0.1 if network is awsvpc and will be value of bridge IP for bridge mode. Also, added the http specific values as input to config setting for Fluentd and Fluentbit respectively.
For GA launch, Host Mode is not supported.

### Testing

New tests cover the changes:yes.

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2189, opened 2019-08-29
Add firelens container health check support

fenxiong:
### Note
This depends on https://github.com/aws/amazon-ecs-agent/pull/2181. Changes are in second commit of this pr.

### Summary

 Add firelens container health check support for bridge and awsvpc network mode.

### Implementation details

Add an input section in firelens config file so that it listens to a port for health check messages, and an output section that consumes the health check messages - 
```
[INPUT]
    Name tcp
    Tag firelens-healthcheck
    Listen 127.0.0.1
    Port 8877

[OUTPUT]
    Name null
    Match firelens-healthcheck
```
With these sections customer can use a container health check command to do health check on firelens container to verify the server is up and running - 
```
echo '{"health": "check"}' | nc 127.0.0.1 8877 || exit 1
```
Health check is only supported for fluentbit but not fluentd because ECS only maintains image for fluentbit and we can ensure `nc` is installed on the image.

https://github.com/awslabs/go-config-generator-for-fluentd-and-fluentbit dependency is updated to fix a bug in template that affects the health check sections.

### Testing

Modified existing unit tests to cover the changes; manually verified that the health check command works. functional test is waiting on @PettitWesley to add `nc` to amazon/aws-for-fluent-bit image.

### Licensing

### DIFF
--- a/agent/taskresource/firelens/firelensconfig_unix.go
+++ b/agent/taskresource/firelens/firelensconfig_unix.go
@@ -74,6 +74,18 @@ const (
 	// inputPortValue is the value for specifying port for fluentd for tcp.
 	inputPortValue = "24224"
 
+	// healthcheckInputNameFluentbit in the input plugin used to receive health check message for fluentbit.
+	healthcheckInputNameFluentbit = "tcp"
+	// healthcheckInputBindValue is the source where health check message comes from.
+	healthcheckInputBindValue = "127.0.0.1"
+	// healthcheckInputPortValue is the port for healthcheck.
+	healthcheckInputPortValue = "8877"
+	// healthcheckTag is the tag for health check message.
+	healthcheckTag = "firelens-healthcheck"
+	// healthcheckOutputName is the output plugin that health check message goes to. It's a black hole so that the health
+	// check messages don't go into logs.
+	healthcheckOutputName = "null"
+
 	// inputListenOptionFluentbit is the key for the log option that specifies host for fluentbit.
 	inputListenOptionFluentbit = "Listen"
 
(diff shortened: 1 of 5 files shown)


agent/api/task/task.go
```diff
@@ -834,9 +841,16 @@ func (task *Task) initializeFirelensResource(config *config.Config, resourceFiel
...
 				enableECSLogMetadata = false
 			}
-
+			var networkMode string
+			if task.IsNetworkModeAWSVPC() {
+				networkMode = AWSVPCNetworkMode
+			} else if container.GetNetworkMode() == "" {
+				networkMode = BridgeNetworkMode
+			} else {
+				networkMode = container.GetNetworkMode()
```
petderek: why won't this line work for all three cases?

agent/taskresource/firelens/firelensconfig_unix.go
```diff
@@ -155,6 +170,25 @@ func (firelens *FirelensResource) generateConfig() (generator.FluentConfig, erro
 	return config, nil
 }
 
+// addHealthcheckSections adds a health check input section and a health check output section to the config.
+func (firelens *FirelensResource) addHealthcheckSections(config generator.FluentConfig) {
+	// Health check supported is only added for fluentbit.
+	if firelens.firelensConfigType != FirelensConfigTypeFluentbit {
```
petderek: I don't think this restriction makes sense. We're blocking customers from using this feature even if they bring their own container.

PettitWesley: Customer's can add a health check for Fluentd, it just has to be done manually.

We add the health check set up because it will work with our `amazon/aws-for-fluent-bit` image. 

For Fluentd, customers would need to modify the public fluentd docker images to support a health check. We will probably provide suggestions on how to do this. But since we can not provide a nice end-to-end experience for fluentd, we do not add a health check configuration by default.

agent/api/task/task.go
fenxiong: both awsvpc mode and default network mode has network mode as empty string. so we need to specially treat those two (fyi this change belongs to https://github.com/aws/amazon-ecs-agent/pull/2181)

agent/taskresource/firelens/firelensconfig_unix.go
fenxiong: >  We're blocking customers from using this feature even if they bring their own container.

if customers bring their own firelens container, they can have an external config file in the image that has the health check input section (as supported in https://github.com/aws/amazon-ecs-agent/pull/2179), and then they can use a health check command base on that. so i guess it's not blocking them, just that we are not doing things for them in that case.

petderek: I missed that they could add their own configuration too and bypass ours entirely.

PR #2191, opened 2019-09-03
Full task sync

shubham2892:
### Summary
A new capability `full-sync` is added which advertises that Agent can handle this new type of messages. 

#### Workflow

When the Agent connects to the ACS , ACS with send a `TaskManifest` type of message. Agent is expected to:

1. Check for staleness: The agent will keep track of the highest *timeline* number received in both PayloadMessage and TaskManifestMessage. If the *timeline* number on TaskManifestMessage is lower to the latest known one, the TaskManifestMessage must be discarded. 

2. If the Manifest is not stale, compare the list of tasks running on the instance to the list of *RUNNING* tasks in the Manifest. Tasks not present in the manifest become candidates to be stopped.  

3. If there are stop candidates, Agent will send a TaskStopVerificationMessage to ACS, containing all the tasks it determines need to stop (stop candidates) with a desired status of *STOPPED*. ACS will then verify if the *STOPPED* tasks in the message should be terminated. ACS will reply with TaskStopVerificationAck, containing all the tasks that must be terminated with a desired status *STOPPED*.
 
4. The agent must terminate all tasks listed in TaskStopVerificationAck. Exceptions saying the task being terminated does not exist should be ignored. Tasks in the Manifest but NOT on the instance are simply ignored.

image

### Implementation details
- `seqNumTaskManifest` is added to Agent struct to keep track of latest timeline number Agent has processed.
- `seqNumTaskManifest` is stored in the state file so that it can be restored when Agent restarts.
- `seqNumTaskManifest` is updated in the state file and Agent struct, on every payload message and TaskManifest message.
- TaskManifestHandler is added to process new type of messages.

### Testing

New tests cover the changes:yes

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -90,6 +90,7 @@ type session struct {
 	cancel                          context.CancelFunc
 	backoff                         retry.Backoff
 	resources                       sessionResources
+	latestSeqNumTaskManifest        *int64
 	_heartbeatTimeout               time.Duration
 	_heartbeatJitter                time.Duration
 	_inactiveInstanceReconnectDelay time.Duration
@@ -143,7 +144,7 @@ func NewSession(ctx context.Context,
 	stateManager statemanager.StateManager,
 	taskEngine engine.TaskEngine,
 	credentialsManager rolecredentials.Manager,
-	taskHandler *eventhandler.TaskHandler) Session {
+	taskHandler *eventhandler.TaskHandler, latestSeqNumTaskManifest *int64) Session {
 	resources := newSessionResources(credentialsProvider)
 	backoff := retry.NewExponentialBackoff(connectionBackoffMin, connectionBackoffMax,
 		connectionBackoffJitter, connectionBackoffMultiplier)
@@ -164,6 +165,7 @@ func NewSession(ctx context.Context,
 		cancel:                          cancel,
 		backoff:                         backoff,
 		resources:                       resources,
+		latestSeqNumTaskManifest:        latestSeqNumTaskManifest,
 		_heartbeatTimeout:               heartbeatTimeout,
 		_heartbeatJitter:                heartbeatJitter,
 		_inactiveInstanceReconnectDelay: inactiveInstanceReconnectDelay,
@@ -294,6 +296,17 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 
 	client.AddRequestHandler(instanceENIAttachHandler.handlerFunc())
 
+	// Add TaskManifestHandler
+	taskManifestHandler := newTaskManifestHandler(acsSession.ctx, cfg.Cluster, acsSession.containerInstanceARN,
+		client, acsSession.stateManager, acsSession.taskEngine, acsSession.latestSeqNumTaskManifest)
+
+	defer taskManifestHandler.clearAcks()
+	taskManifestHandler.start()
+	defer taskManifestHandler.stop()
+
+	client.AddRequestHandler(taskManifestHandler.handlerFuncTaskManifestMessage())
+	client.AddRequestHandler(taskManifestHandler.handlerFuncTaskStopVerificationMessage())
+
 	// Add request handler for handling payload messages from ACS
 	payloadHandler := newPayloadRequestHandler(
 		acsSession.ctx,
@@ -305,7 +318,7 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 		acsSession.stateManager,
 		refreshCredsHandler,
 		acsSession.credentialsManager,
-		acsSession.taskHandler)
+		acsSession.taskHandler, acsSession.latestSeqNumTaskManifest)
 	// Clear the acks channel on return because acks of messageids don't have any value across sessions
 	defer payloadHandler.clearAcks()
 	payloadHandler.start()
--- a/agent/acs/handler/acs_handler_test.go
+++ b/agent/acs/handler/acs_handler_test.go
@@ -47,6 +47,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/version"
 	"github.com/aws/amazon-ecs-agent/agent/wsclient"
 	mock_wsclient "github.com/aws/amazon-ecs-agent/agent/wsclient/mock"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/gorilla/websocket"
 	"github.com/pkg/errors"
@@ -365,6 +366,7 @@ func TestHandlerReconnectsWithoutBackoffOnEOFError(t *testing.T) {
 		ctx:                             ctx,
 		cancel:                          cancel,
 		resources:                       &mockSessionResources{mockWsClient},
+		latestSeqNumTaskManifest:        aws.Int64(10),
 		_heartbeatTimeout:               20 * time.Millisecond,
 		_heartbeatJitter:                10 * time.Millisecond,
 		_inactiveInstanceReconnectDelay: inactiveInstanceReconnectDelay,
@@ -833,18 +835,19 @@ func TestHandlerDoesntLeakGoroutines(t *testing.T) {
 	go func() {
 
 		acsSession := session{
-			containerInstanceARN: "myArn",
-			credentialsProvider:  testCreds,
-			agentConfig:          testConfig,
-			taskEngine:           taskEngine,
-			ecsClient:            ecsClient,
-			stateManager:         stateManager,
-			taskHandler:          taskHandler,
-			ctx:                  ctx,
-			_heartbeatTimeout:    1 * time.Second,
-			backoff:              retry.NewExponentialBackoff(connectionBackoffMin, connectionBackoffMax, connectionBackoffJitter, connectionBackoffMultiplier),
-			resources:            newSessionResources(testCreds),
-			credentialsManager:   rolecredentials.NewManager(),
+			containerInstanceARN:     "myArn",
+			credentialsProvider:      testCreds,
+			agentConfig:              testConfig,
+			taskEngine:               taskEngine,
+			ecsClient:                ecsClient,
+			stateManager:             stateManager,
+			taskHandler:              taskHandler,
+			ctx:                      ctx,
+			_heartbeatTimeout:        1 * time.Second,
+			backoff:                  retry.NewExponentialBackoff(connectionBackoffMin, connectionBackoffMax, connectionBackoffJitter, connectionBackoffMultiplier),
+			resources:                newSessionResources(testCreds),
+			credentialsManager:       rolecredentials.NewManager(),
+			latestSeqNumTaskManifest: aws.Int64(12),
 		}
 		acsSession.Start()
 		ended <- true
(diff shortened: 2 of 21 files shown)


agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func (taskManifestHandler *taskManifestHandler) handlerFuncTaskVerificationMessage() func(
+	message *ecsacs.TaskStopVerificationAck) {
+	return func(message *ecsacs.TaskStopVerificationAck) {
+		taskManifestHandler.messageBufferTaskStopVerificationAck <- message
+	}
+}
+
+func (taskManifestHandler *taskManifestHandler) start() {
```
petderek: should we split this out into multiple (probably 2) handlers?

shubham2892: Can be done, but they are a part of one operation(they are incomplete without each other), so will prefer to keep them together, thoughts?

agent/app/agent.go
```diff
@@ -344,7 +347,7 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 	// previousStateManager is used to verify that our current runtime configuration is
 	// compatible with our past configuration as reflected by our state-file
 	previousStateManager, err := agent.newStateManager(previousTaskEngine, &previousCluster,
-		&previousContainerInstanceArn, &previousEC2InstanceID, &previousAZ)
+		&previousContainerInstanceArn, &previousEC2InstanceID, &previousAZ, agent.seqNumTaskManifest)
```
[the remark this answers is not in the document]
shubham2892: Synced offline about this.

agent/acs/handler/acs_handler.go
```diff
@@ -90,6 +90,7 @@ type session struct {
 	cancel                          context.CancelFunc
 	backoff                         retry.Backoff
 	resources                       sessionResources
+	seqNumTaskManifest              *int64
```
sharanyad: how about `lastTaskManifestSeqNum` since it's the latest number ?

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+	// Task stop verification message and it's ack
+	go taskManifestHandler.sendTaskStopVerificationMessage()
+	go taskManifestHandler.handleTaskStopVerificationAck()
+
+}
+
+func (taskManifestHandler *taskManifestHandler) sendTaskManifestMessageAck() {
+	for {
+		select {
+		case mid := <-taskManifestHandler.messageBufferTaskManifestAck:
```
sharanyad: what's mid?

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+	}
+}
+
+func (taskManifestHandler *taskManifestHandler) sendTaskStopVerificationMessage() {
+	for {
+		select {
+		case message := <-taskManifestHandler.messageBufferTaskStopVerificationMessage:
+			if err := taskManifestHandler.sendTaskTerminationVerificationMessage(message); err != nil {
+				seelog.Warnf("Unable to handle taskStopVerification message [%s]: %v", message.String(), err)
```
sharanyad: i think this should be unable to send

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+			for _, receivedTask := range receivedTaskList {
+				if *receivedTask.TaskArn == runningTask.Arn && *receivedTask.DesiredStatus == "RUNNING" {
+					// Task present, does not need to be stopped
+					taskPresent = true
+					break
+				}
+			}
+			if !taskPresent {
+				tasksToBeKilled = append(tasksToBeKilled, &ecsacs.TaskIdentifier{
+					DesiredStatus: aws.String("STOPPED"),
```
sharanyad: why is the string used here instead of constant(taskstatus.Stopped?)?

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+func (taskManifestHandler *taskManifestHandler) handleSingleMessageVerificationAck(
+	message *ecsacs.TaskStopVerificationAck) error {
+	seelog.Debugf("Message received from sending back task verification ack: ", message)
+
+	// Ensure that we have received a corresponding task manifest message before
+	if *message.MessageId == taskManifestHandler.messageId {
+		// Reset the message id
+		taskManifestHandler.messageId = ""
```
sharanyad: why are we resetting the ID? can you add comments for it?

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+		for _, taskToKill := range message.StopTasks {
+			if *taskToKill.DesiredStatus == "STOPPED" {
```
sharanyad: lets not use string here

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+				task, isPresent := taskManifestHandler.taskEngine.GetTaskByArn(*taskToKill.TaskArn)
+				if isPresent {
+					seelog.Info("Killing task: %s", task.Arn)
```
sharanyad: probably saying stopping task might make more sense. also could you please add more context that this is from task manifest handler?

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+	}
+	return nil
+}
+
+func (taskManifestHandler *taskManifestHandler) handleTaskManifestSingleMessage(
+	message *ecsacs.TaskManifestMessage) error {
+
+	taskListManifestHandler := message.Tasks
+	seqNumberFromMessage := *message.Timeline
+	agentSequenceNumber := *taskManifestHandler.seqNumTaskManifest
```
sharanyad: this sounds like sequence number of the agent.. can you rename this to something like last or latest sequence number?

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,280 @@
...
+		// Update state file
+		*taskManifestHandler.seqNumTaskManifest = *message.Timeline
+		err = taskManifestHandler.saver.Save()
+		if err != nil {
+			return seelog.Errorf("Error saving state for task manifest message: %v", err)
+		}
+
+		tasksToKill := compareTasks(taskListManifestHandler, runningTasksOnInstance)
+
+		taskStopVerificationMessage := ecsacs.TaskStopVerificationMessage{
```
sharanyad: let's skip constructing this message if there are no tasks to kill.

agent/acs/handler/task_manifest_handler.go
shubham2892: It's just a placeholder variable, will change it to a more meaningful name.

agent/acs/handler/task_manifest_handler.go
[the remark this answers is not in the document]
shubham2892: Good catch, added a getter and setter for messageId with locks.

agent/acs/handler/task_manifest_handler.go
shubham2892: Added a string constant.

agent/acs/handler/acs_handler_test.go
```diff
@@ -28,6 +28,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
```
fenxiong: nit: group import

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,293 @@
...
+	go taskManifestHandler.sendTaskManifestMessageAck()
+
+	// Task stop verification message and it's ack
+	go taskManifestHandler.sendTaskStopVerificationMessage()
+	go taskManifestHandler.handleTaskStopVerificationAck()
+
+}
+
+func (taskManifestHandler *taskManifestHandler) getMessageId() string {
+	taskManifestHandler.lock.Lock()
```
fenxiong: RLock/RUnlock

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+func (taskManifestHandler *taskManifestHandler) sendTaskTerminationVerificationMessage(
+	message *ecsacs.TaskStopVerificationMessage) error {
+
+	seelog.Debugf("Sending Task Termination Verification Message: %s", message.String())
+	err := taskManifestHandler.acsClient.MakeRequest(message)
+	if err != nil {
+		seelog.Warnf("Error 'ack'ing request with messageID: %s, error: %v", message.MessageId, err)
```
fenxiong: You already logged this error iin sendTaskStopVerificationMessage so i don't think you need to log it here?

agent/api/task/status/taskstatus.go
```diff
@@ -31,16 +31,24 @@ const (
 	TaskStopped
 	// TaskZombie is an "impossible" state that is used as the maximum
 	TaskZombie
+	// Task Stopped Status String
```
fenxiong: (golint) should start with // TaskStoppedString... same below

agent/acs/handler/task_manifest_handler.go
shubham2892: Yeah makes sense.

agent/acs/handler/task_manifest_handler.go
```diff
@@ -0,0 +1,287 @@
...
+			}
+		case <-taskManifestHandler.ctx.Done():
+			return
+		}
+	}
+}
+
+// compares the list of tasks received in the task manifest message and tasks running on the the instance
+// It returns all the task that are running on the instance but not present in task manifest message task list
+func compareTasks(receivedTaskList []*ecsacs.TaskIdentifier, runningTaskList []*apitask.Task) []*ecsacs.TaskIdentifier {
```
sharanyad: missing unit tests for this method.

agent/api/task/status/taskstatus.go
```diff
@@ -31,16 +31,24 @@ const (
...
 
 // TaskStatus is an enumeration of valid states in the task lifecycle
 type TaskStatus int32
 
 var taskStatusMap = map[string]TaskStatus{
-	"NONE":    TaskStatusNone,
-	"CREATED": TaskCreated,
-	"RUNNING": TaskRunning,
-	"STOPPED": TaskStopped,
+	TaskNoneString:    TaskStatusNone,
```
sharanyad: thanks for changing these.

agent/app/agent_capability.go
```diff
@@ -55,6 +55,7 @@ const (
 	capabilityFirelensLoggingDriver             = "logging-driver.awsfirelens"
 	capabilityFirelensConfigFile                = "firelens.options.config.file"
 	capabilityFirelensConfigS3                  = "firelens.options.config.s3"
+	capabilityFullTaskSync                      = "full-sync"
```
sharanyad: why not `full-task-sync` ?

shubham2892: Backend named this before us, they decided on full-sync. 🙂

PR #2198, opened 2019-09-11
BugFix for hostmode support for Firelens

mssrivas:
### Summary
BugFix for hostmode support for Firelens. 

### Implementation details
Instead of getting network mode from the container itself, I changed it to get it hostconfig instead. 

### Testing
Checked to see the correct network mode is identified. 

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -990,6 +990,26 @@ func (c *Container) GetLogDriver() string {
 	return hostConfig.LogConfig.Type
 }
 
+// GetNetworkModeFromHostConfig returns the network mode used by the container from the host config .
+func (c *Container) GetNetworkModeFromHostConfig() string {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return ""
+	}
+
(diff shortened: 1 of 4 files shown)


agent/api/container/container.go
```diff
@@ -990,6 +990,25 @@ func (c *Container) GetLogDriver() string {
...
+
+	if c.DockerConfig.HostConfig == nil {
+		return ""
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encountered error when trying to get network mode for container %s: %v", err)
+		return ""
```
sharanyad: this assumes that it is default network mode, which is ok for now. can you add a TODO to return err, string from this method to distinguish between error and default mode?

mssrivas: I added the todo.

sharanyad:
has this been manually tested for different network modes?

mssrivas:
> has this been manually tested for different network modes?

Yes, Tested for bridge, none, blank

PR #2182, opened 2019-08-23
Added support for automatic spot instance draining.

sparrc:
### Summary

This change adds a polling go-routine to the agent that checks if the instance has been scheduled for a [spot termination.](https://aws.amazon.com/blogs/aws/new-ec2-spot-instance-termination-notices/) If it has, agent will set the [instance's state to DRAINING](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container-instance-draining.html).

see also https://github.com/aws/containers-roadmap/issues/190

### Implementation details

1. Added a new function to the internal ECS client: UpdateContainerInstancesState
2. Added a new function to the internal ec2 metadata client: SpotTerminationTime, checks metadata endpoint at spot/termination-time for a termination notice. If the notice has not been set this endpoint will 404.
3. Added a polling go-routine to check if the spot termination time has been set every 1s. If it has been set, the agent updates the container instance state to DRAINING.

### Testing

Unit tests added with full coverage. All integration tests run. Manually tested to verify functionality.

New tests cover the changes: yes

### Description for the changelog

Added support for automatic spot instance draining.

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2199, opened 2019-09-11
Change termination-time query to instance-action

sparrc:
### Summary

This is an update to https://github.com/aws/amazon-ecs-agent/pull/2182

The previous PR only handled the most common type of interruption notice: termination.

But it's possible for users to configure their spot instances to terminate via "stop" or "hibernate", so this PR will handle these types of notices as well.

### Implementation details

As explained here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html#using-spot-instances-managing-interruptions, it's possible for spot instances to be interrupted with hibernate or stop actions.

In our implementation, we don't care about which action happened or what time the instance is scheduled to be interrupted. In the same way the termination-time endpoint worked, the instance-action endpoint will 404 until an instance-action has been scheduled (stop, hibernate, or terminate). So once we know that one of these interruptions has been scheduled, we simply set the instance status to DRAINING ASAP.

### Testing

unit, integration, manual

Added support for automatic spot instance draining.

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -15,6 +15,7 @@ package app
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"time"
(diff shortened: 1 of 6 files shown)


agent/app/agent.go
```diff
@@ -626,13 +626,13 @@ func (agent *ecsAgent) startSpotInstanceDrainingPoller(client api.ECSClient) {
...
 }
 
-// spotInstanceDrainingPoller returns true if spot instance termination time has been
+// spotInstanceDrainingPoller returns true if spot instance interruption has been
 // set AND the container instance state is successfully updated to DRAINING.
 func (agent *ecsAgent) spotInstanceDrainingPoller(client api.ECSClient) bool {
-	// this endpoint 404s unless a termination time has been set, so expect failure in most cases.
-	termtime, err := agent.ec2MetadataClient.SpotTerminationTime()
+	// this endpoint 404s unless a interruption has been set, so expect failure in most cases.
+	termtime, err := agent.ec2MetadataClient.SpotInstanceAction()
```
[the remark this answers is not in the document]
sparrc: json validation now added

agent/app/agent.go
```diff
@@ -626,13 +627,32 @@ func (agent *ecsAgent) startSpotInstanceDrainingPoller(client api.ECSClient) {
...
+		ia := InstanceAction{}
+
+		err := json.Unmarshal([]byte(resp), &ia)
+		if err != nil {
+			seelog.Errorf("Invalid response from /spot/instance-action endpoint: %s Error: %s", resp, err)
+			return false
+		}
+
+		switch ia.Action {
+		case "hibernate", "terminate", "stop":
```
sharanyad: is it good enough to check just action? is it guaranteed that time will be present?

agent/app/agent_test.go
```diff
@@ -1182,7 +1182,7 @@ func TestGetHostPublicIPv4AddressFromEC2MetadataFailWithError(t *testing.T) {
 	assert.Empty(t, agent.getHostPublicIPv4AddressFromEC2Metadata())
 }
 
-func TestSpotTerminationTimeCheck_Yes(t *testing.T) {
+func TestSpotInstanceActionCheck_Terminate(t *testing.T) {
```
sharanyad: all these unit tests could be made table-driven.

agent/app/agent.go
sparrc: yep, there will not be an instance-action present unless it's already been scheduled.

agent/app/agent_test.go
sparrc: refactored to be table-driven now

PR #2208, opened 2019-09-16
Functional test to test fluentd and fluentbit with fluent-logger

mssrivas:
### Summary
Functional test to test fluentd and fluentbit with fluent-logger

### Implementation details
Added image for fluent-logger, for here - https://github.com/fluent/fluent-logger-golang modified to work for the test case.
Added two functional test using the fluent-logger for fluentd and fluentbit

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/functional_tests/testdata/taskdefinitions/firelens-fluentd-fluent-logger/task-definition.json
+++ b/agent/functional_tests/testdata/taskdefinitions/firelens-fluentd-fluent-logger/task-definition.json
@@ -1,0 +1,53 @@
+{
+  "family": "ecsftest-firelens-fluentd-fluent-logger",
+  "networkMode": "bridge",
+  "taskRoleArn": "$$$TASK_ROLE$$$",
+  "executionRoleArn": "$$$EXECUTION_ROLE$$$",
+  "containerDefinitions": [
+    {
+      "name": "firelens",
+      "image": "127.0.0.1:51670/amazon/firelens-fluentd:latest",
+      "essential": true,
+      "memory": 256,
+      "firelensConfiguration": {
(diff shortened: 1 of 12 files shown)


agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1919,20 +1934,24 @@ func testFirelens(t *testing.T, firelensConfigType, secretLogOptionKey, secretLo
 	agent := RunAgent(t, agentOptions)
 	defer agent.Cleanup()
 
-	agent.RequireVersion(">=1.31.0")
-
+	agent.RequireVersion(">=1.30.0")
```
mssrivas: Will update this before pushing

misc/fluent-logger/fluent-logger.go
```diff
@@ -0,0 +1,27 @@
+package main
```
fenxiong: you should remove the binary file "misc/fluent-logger/fluent-logger" from the pr (can't add comment on the binary so commented here)

fenxiong: missing license header for this file

agent/functional_tests/testdata/taskdefinitions/firelens-fluentd-fluent-logger/task-definition.json
```diff
@@ -0,0 +1,53 @@
+{
+  "family": "ecsftest-firelens-fluentd-fluent-logger",
+  "networkMode": "bridge",
```
fenxiong: i think you should add tests for both bridge and awsvpc

PR #2212, opened 2019-09-18
Wait for agent to exit before exiting test case

sparrc:
go's test framework can panic if there are logs within a test-case after
the test already exits. This makes sure that the agent exits before the
test exits so that we don't get a false test failure.

### Licensing

### DIFF
--- a/agent/app/agent_unix_test.go
+++ b/agent/app/agent_unix_test.go
@@ -105,7 +105,6 @@ func TestDoStartHappyPath(t *testing.T) {
 	cfg := getTestConfig()
 	ctx, cancel := context.WithCancel(context.TODO())
 	// Cancel the context to cancel async routines
-	defer cancel()
 	agent := &ecsAgent{
 		ctx:                ctx,
 		cfg:                &cfg,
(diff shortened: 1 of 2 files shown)


yunhee-l:
given the flakiness of this, have you tested running in bulk?

sparrc:
> given the flakiness of this, have you tested running in bulk?

Unfortunately I haven't been able to get this to reproduce, before or after these changes, but we have a traceback that indicates this was the issue in at least one case.

agent/app/agent_unix_test.go
```diff
@@ -115,13 +114,20 @@ func TestDoStartHappyPath(t *testing.T) {
 		mobyPlugins:        mockMobyPlugins,
 	}
 
-	go agent.doStart(eventstream.NewEventStream("events", ctx),
-		credentialsManager, state, imageManager, client)
+	var agentW sync.WaitGroup
+	agentW.Add(1)
+	go func() {
+		agent.doStart(eventstream.NewEventStream("events", ctx),
```
fenxiong: there are some more tests running agent.doStart in agent_test.go and agent_windows_test.go. it's better to fix them together?

sparrc: 👍 , I can fix those too

fenxiong: actually just one, TestDoStartRegisterAvailabilityZone

sparrc: There was only one other instance in agent_test.go and it's fixed now too, the other  instances wait to check the status code so they are OK.

PR #2223, opened 2019-09-24
Flaky unit test: metrics collection

sparrc:
Few things to make this test less flaky:
- reduce the total number of goroutines
- reduce the length of the sleeps
- sleeping before waitgroup wasn't doing anything, sleep after

### Licensing


shubham2892:
How is it tested to ensure that it is less flaky?

sparrc:
> How is it tested to ensure that it is less flaky?

It's not been tested because I can't reproduce on macos or linux (it's also hard to get it to happen on ARM anyways. The test depends on lots of timing and goroutines to line-up. I'm assuming that by reducing the timing and the number of goroutines we'll reduce the chance of a false failure.

PR #2228, opened 2019-09-30
Change DockerTestAuth from functional to integ

shubham2892:
### Summary

There is already an integration test `TestDockerCfgAuth` which tests the same functionality as `TestDockerAuth`. `TestDockerAuth` additionally scrapes through the log to check it does not contain any authentication detail.

This PR adds a function which when added to an integration tests stores the log in the same way as stored by functional tests and adds the log check to `TestDockerCfgAuth`.

### Testing

### Licensing


agent/engine/common_integ_test.go
```diff
@@ -56,6 +60,43 @@ func createTestTask(arn string) *apitask.Task {
...
+
+func loggerConfigIntegrationTest(logfile string) string {
+	config := `
+	<seelog type="asyncloop" minlevel="debug">
+		<outputs formatid="main">
+			<console />`
+	if logfile != "" {
+		config += `<rollingfile filename="` + logfile + `/ecs-agent-log.log" type="date"
+			 datepattern="2006-01-02-15" archivetype="none" maxrolls="24" />`
+	}
```
yunhee-l: nit: would using sprintf make this a bit neater?

shubham2892: Yes will use that.

agent/engine/common_integ_test.go
```diff
@@ -56,6 +61,40 @@ func createTestTask(arn string) *apitask.Task {
 	}
 }
 
+func setupIntegTestLogs(t *testing.T) string {
+	// Create a directory for storing test logs.
+	testLogDir, err := ioutil.TempDir("", "ecs-integ-test")
+	require.NoError(t, err, "Unable to create directory for storing test logs")
+
+	logger, err := log.LoggerFromConfigAsString(loggerConfigIntegrationTest(testLogDir))
+	if err != nil {
```
sharanyad: can use `assert` here and below

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -156,56 +156,6 @@ func TestCommandOverrides(t *testing.T) {
 	assert.Equal(t, 21, exitCode, fmt.Sprintf("Expected exit code of 21; got %d", exitCode))
 }
 
-func TestDockerAuth(t *testing.T) {
-	agent := RunAgent(t, &AgentOptions{
-		ExtraEnvironment: map[string]string{
-			"ECS_ENGINE_AUTH_TYPE": "dockercfg",
```
sharanyad: with this migration, we essentially lose testing of these env vars?

shubham2892: We are still setting the same flags with these values but in task definition instead os environment variables.

PR #2229, opened 2019-10-02
Migrate TestFluentdTag from functional test to integration test

shubham2892:
### Summary
- Migrate `TestFluentdTag` from functional test to integration test. 
- Remove `TestFluentdlogTag` functional test as it uses the older version of Docker which we don't use anymore in our test framework. 
### Testing

### Licensing


sharanyad:
looks like we still support docker versions <1.12.0? https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html

shubham2892:
> looks like we still support docker versions <1.12.0? https://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html

Synced offline, will add a integ test in separate PR for customers who might be still using old Docker(<1.12.0)

PR #2230, opened 2019-10-07
Add empty string validation for app mesh

yumex93:
### Summary

1. Based what documented in current public [doc](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ProxyConfiguration.html), people can leave empty list for EgressIgnoredIPs and EgressIgnoredPorts as empty list, which will be an empty string in task def. However, in current agent logic, if an empty string is specified, it will not return an empty array, instead, it returns an array including an empty string which will cause task fail to start since it is considered as an invalid ip address/port in app-mesh plugin. 

The bug is found internally. Seems due to on [integrating with App Mesh instruction page](https://docs.aws.amazon.com/app-mesh/latest/userguide/mesh-getting-started-ecs.html), the example shows to not specify these two attributes when not needed, customers may not really use empty list under such cases.

2. Update ECS CNI version to 2019.10.0.
3. Increase integration test timeout time. 

### Testing

New tests cover the changes: 

### Licensing


petderek:
Can we defer this validation to the plugin?

fenxiong:
> Can we defer this validation to the plugin?

I think this is to avoid an invalid input, rather than adding a validation?

petderek:
> > Can we defer this validation to the plugin?
> 
> I think this is to avoid an invalid input, rather than adding a validation?

Makes sense to me!

PR #2232, opened 2019-10-08
Replace squid proxy functional test with unit test

shubham2892:
### Summary
~~Replace `TestSquidProxy` functional test with~~ Add Unit tests to test that agent is using the proxies when flags(_HTTP_PROXY_, _HTTPS_PROXY_) are set. 

~~#### Why is unit tests sufficient to test what a functional test was doing?~~

~~`TestSquidProxy` functional test does the following~~
~~1. Start a squid proxy container.~~
~~2. Start agent container with _HTTP_PROXY_ set so that all the requests made by agent goes through the proxy.~~
~~3. Scrape the logs in squid proxy container to verify that the requests in agent container actually went through the proxy.~~

~~By replacing the functional test with unit tests, we are testing part -2, that is,  Agent is able to pick up the proxy address that is set via _HTTP_PROXY_ flag and makes request using the proxy address. We are skipping the part -1 and part - 3 which is not required to ensure that agent is able to use HTTP proxy for making requests.~~
### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/httpclient/httpclient_test.go
+++ b/agent/httpclient/httpclient_test.go
@@ -17,6 +17,8 @@ package httpclient
 
 import (
 	"net/http"
+	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


samuelkarp:
> We are skipping the part -1 and part - 3 which is not required to ensure that agent is able to use HTTP proxy for making requests.

Why do you believe this is true?  The `HTTP_PROXY` environment variable is respected by default with Go's HTTP client, but is not forced to be respected.  If we modify the HTTP client, we no longer have a guarantee that the proxy will be used.

The point of these functional tests is to verify that the agent functions as expected, especially for critical functionality like this.  Removing this test makes me concerned about future breakage.

shubham2892:
> > We are skipping the part -1 and part - 3 which is not required to ensure that agent is able to use HTTP proxy for making requests.
> 
> Why do you believe this is true? The `HTTP_PROXY` environment variable is respected by default with Go's HTTP client, but is not forced to be respected. If we modify the HTTP client, we no longer have a guarantee that the proxy will be used.

@samuelkarp With the unit tests, I am trying to test if the HTTP client is modified in a way `HTTP_PROXY` environment variable is not respected, it will fail.

petderek:
Theres still value in adding the unit tests. I think the way to do this would be to use the [httptest](https://golang.org/pkg/net/http/httptest/) utility, which will let you validate behavior without attempting to make a real network connection. You don't want to do that in the context of a unit test.

I'd like to see a simple positive e2e case where we connect via proxy -- and we're losing that validation here.

samuelkarp:
> I am trying to test, if the HTTP client is modified in a way HTTP_PROXY environment variable is not respected, it will fail.

That's only a failure case.  There needs to be a success case here for validating that the functionality actually works.

I much prefer the existing functional test, as it runs a real agent and validates that it functions.

shubham2892:
Synced offline, right now we are adding unit tests and not removing the functional test as a step towards deprecating `TestSquidProxy` functional test.

petderek:
To be clear -- we aren't deprecating it. We're finding another way to replace it.

PR #2234, opened 2019-10-08
Added EFSVolumeConfiguration

sparrc:
### Summary

This adds support for a new field in the task definition: EFSVolumeConfiguration

For now we only support a minimal version of EFS support that essentially just translates the EFS volume configuration into an NFS docker volume.

In the future we expect to roll out support for EFS-specific features by utilizing the efs-agent to mount the volumes.

### Implementation details

ECS already supports a field `dockerVolumeConfiguration` in the task definition that can be used to mount EFS volumes via NFS (see [here](https://aws.amazon.com/premiumsupport/knowledge-center/ecs-create-docker-volume-efs/)). By utilizing this, we can translate `EFSVolumeConfiguration` into a dockerVolumeConfiguration for basic support of mounting EFS volumes (although without support for EFS-specific features).

The plan is to maintain this "simple mode" mount whenever customers want to mount an EFS volume without utilizing any EFS-specific features. Simple EFS configuration would look something like this:

```json
"EFSVolumeConfiguration": {
    "filesystem": "fs-abcde",
    "rootDirectory": "/mydata",
    "readonly": true
}
```

### Testing

Added support for EFSVolumeConfiguration in task definition

### Licensing

### DIFF
--- a/agent/acs/model/api/api-2.json
+++ b/agent/acs/model/api/api-2.json
@@ -304,8 +304,7 @@
       "type":"structure",
       "members":{
         "filesystem":{"shape":"String"},
-        "rootDirectory":{"shape":"String"},
-        "readonly":{"shape":"Boolean"}
+        "rootDirectory":{"shape":"String"}
       }
     },
     "ElasticNetworkInterface":{
--- a/agent/acs/model/api/api-2.json
+++ b/agent/acs/model/api/api-2.json
@@ -303,7 +303,7 @@
     "EFSVolumeConfiguration": {
       "type":"structure",
       "members":{
-        "filesystem":{"shape":"String"},
+        "filesystemId":{"shape":"String"},
         "rootDirectory":{"shape":"String"}
       }
     },
@@ -699,7 +699,8 @@
       "type":"string",
       "enum":[
         "host",
-        "docker"
+        "docker",
+        "efs"
       ]
     }
   }
--- a/agent/acs/model/api/api-2.json
+++ b/agent/acs/model/api/api-2.json
@@ -303,7 +303,7 @@
     "EFSVolumeConfiguration": {
       "type":"structure",
       "members":{
-        "filesystemId":{"shape":"String"},
+        "fileSystemId":{"shape":"String"},
         "rootDirectory":{"shape":"String"}
       }
     },
--- a/agent/acs/model/api/api-2.json
+++ b/agent/acs/model/api/api-2.json
@@ -300,6 +300,14 @@
       }
     },
 
+    "EFSVolumeConfiguration": {
+      "type":"structure",
+      "members":{
+        "filesystem":{"shape":"String"},
+        "rootDirectory":{"shape":"String"},
+        "readonly":{"shape":"Boolean"}
+      }
+    },
     "ElasticNetworkInterface":{
       "type":"structure",
       "members":{
@@ -669,7 +677,8 @@
         "name":{"shape":"String"},
         "type":{"shape":"VolumeType"},
         "host":{"shape":"HostVolumeProperties"},
-        "dockerVolumeConfiguration":{"shape":"DockerVolumeConfiguration"}
+        "dockerVolumeConfiguration":{"shape":"DockerVolumeConfiguration"},
+        "EFSVolumeConfiguration":{"shape":"EFSVolumeConfiguration"}
       }
     },
     "VolumeFrom":{
--- a/agent/acs/model/ecsacs/api.go
+++ b/agent/acs/model/ecsacs/api.go
@@ -413,6 +413,26 @@ func (s ECRAuthData) GoString() string {
 	return s.String()
 }
 
+type EFSVolumeConfiguration struct {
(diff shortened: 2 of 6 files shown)


agent/api/task/task.go
```diff
@@ -285,6 +285,22 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
 	return task, nil
 }
 
+func (task *Task) initializeVolumes(cfg *config.Config, dockerClient dockerapi.DockerClient, ctx context.Context) error {
```
sparrc: this had to be factored out to make gocyclo happy

agent/api/task/task.go
```diff
@@ -500,6 +512,65 @@ func (task *Task) initializeDockerVolumes(sharedVolumeMatchFullConfig bool, dock
...
+// addEFSVolumes converts the EFS task definition into an internal docker 'local' volume
+// mounted with NFS struct and updates container dependency
+func (task *Task) addEFSVolumes(
+	ctx context.Context,
+	cfg *config.Config,
+	dockerClient dockerapi.DockerClient,
+	vol *TaskVolume,
+	efsvol *taskresourcevolume.EFSVolumeConfig,
+) error {
+	ostr := fmt.Sprintf("addr=%s.efs.%s.amazonaws.com,nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport,ro", efsvol.Filesystem, cfg.AWSRegion)
```
petderek: you can remove 'ro'

Can you add a todo for partitional regarding the DNS suffix?

agent/api/task/task.go
```diff
@@ -500,6 +512,66 @@ func (task *Task) initializeDockerVolumes(sharedVolumeMatchFullConfig bool, dock
...
+// mounted with NFS struct and updates container dependency
+func (task *Task) addEFSVolumes(
+	ctx context.Context,
+	cfg *config.Config,
+	dockerClient dockerapi.DockerClient,
+	vol *TaskVolume,
+	efsvol *taskresourcevolume.EFSVolumeConfig,
+) error {
+	// TODO CN and gov partition logic
+	ostr := fmt.Sprintf("addr=%s.efs.%s.amazonaws.com,nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport", efsvol.Filesystem, cfg.AWSRegion)
```
sharanyad: might be worth it to make these values constant

agent/acs/model/api/api-2.json
```diff
@@ -669,7 +676,8 @@
         "name":{"shape":"String"},
         "type":{"shape":"VolumeType"},
```
sharanyad: `VolumeType` should change to add 'efs' option.

agent/acs/model/ecsacs/api.go
```diff
@@ -413,6 +413,24 @@ func (s ECRAuthData) GoString() string {
 	return s.String()
 }
 
+type EFSVolumeConfiguration struct {
+	_ struct{} `type:"structure"`
+
+	Filesystem *string `locationName:"filesystem" type:"string"`
```
sharanyad: `FilesystemID` makes more sense

agent/api/task/task.go
sparrc: I'd prefer not to since it's only used in one place as of now

agent/acs/model/api/api-2.json
sparrc: good catch!!

agent/acs/model/ecsacs/api.go
sparrc: 👍

agent/api/task/task.go
sharanyad: alright, let's at least add comments on what these are and what the options mean.

petderek: It'd make sense to link out to the efs docs about these options:

https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-general.html

agent/acs/model/ecsacs/api.go
sparrc: I've changed it to `FilesystemId` as this is auto-generated code that depends on the ACS model, which will have the `d` lowercase.

agent/acs/model/api/api-2.json
```diff
@@ -300,6 +300,13 @@
       }
     },
 
+    "EFSVolumeConfiguration": {
+      "type":"structure",
+      "members":{
+        "filesystemId":{"shape":"String"},
```
yumex93: In our PRFAQ, this attribute is defined as **filesystem**, ACS will keep the consistency and use **filesystem**. Just a reminder that we may want to wait for ACS model finalized and regenerate the file again.

sparrc: yep, the exact shape of this is subject to change, we'll keep this code in a feature branch until the ACS model is finalized

PR #2237, opened 2019-10-11
Skip firelens s3 test when lacking s3 access

fenxiong:
### Summary

Skip firelens s3 tests when lacking s3 access. Our normal CI/qual process will still be able to run the tests, and this is to unblock individual test runner that lacks s3 access to the specified bucket

### Testing

tested with a bucket that the test doesn't have permission to and verified that it skips the test.

### Licensing

### DIFF
--- a/agent/functional_tests/tests/functionaltests_unix_test.go
+++ b/agent/functional_tests/tests/functionaltests_unix_test.go
@@ -54,6 +54,7 @@ const (
 	savedStateTaskDefinition        = "nginx"
 	portResContentionTaskDefinition = "busybox-port-5180"
 	labelsTaskDefinition            = "labels"
+	errCodeAccessDenied             = "AccessDenied"
 )
 
 var trunkingInstancePrefixes = []string{"c5.", "m5."}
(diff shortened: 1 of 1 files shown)


agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1642,7 +1642,9 @@ func TestFirelensWithS3ConfigFluentd(t *testing.T) {
 		Key:    aws.String("testfiles/fluentd.conf"),
 		Body:   strings.NewReader(content),
 	})
-	require.NoError(t, err, "unable to upload config file to s3 which is needed for the test")
+	if err != nil {
+		t.Skipf("Skipping the test since unable to upload config file to s3: %v", err)
+	}
```
shubham2892: Is it possible to check permission before attemping to upload?

Failure to upload will always mean that the permission is missing? Could it be an indicator of something else going wrong, can we check the error message and based on that skip the test?

fenxiong: yeah, updated to only skip test on access denied

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1642,7 +1642,13 @@ func TestFirelensWithS3ConfigFluentd(t *testing.T) {
 		Key:    aws.String("testfiles/fluentd.conf"),
 		Body:   strings.NewReader(content),
 	})
-	require.NoError(t, err, "unable to upload config file to s3 which is needed for the test")
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == "AccessDenied" {
```
sharanyad: having this as a string is fine now, but if we could have it reference an actual error such as `s3.ErrAccessDenied`, it would be better. 
also, this can be made a constant.

fenxiong: i looked at aws go sdk, but it doesn't seem like s3 is throwing a specific error for the api, or even defining the error code as a constant anywhere. i updated the error code to be a constant

PR #2238, opened 2019-10-11
Add Outpost ID to RCI call

shubham2892:
### Summary

Pass outpost Arn in RegistrationContainerInstance call. Outpost Arn is fetched using instance metadata. 

Outpost Arn is only passed for outpost instance, for non-outpost instance no change in RCI call. 

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -110,8 +110,10 @@ func (client *APIECSClient) CreateCluster(clusterName string) (string, error) {
 // ContainerInstanceARN if successful. Supplying a non-empty container
 // instance ARN allows a container instance to update its registered
 // resources.
-func (client *APIECSClient) RegisterContainerInstance(containerInstanceArn string,
-	attributes []*ecs.Attribute, tags []*ecs.Tag, registrationToken string, platformDevices []*ecs.PlatformDevice) (string, string, error) {
+func (client *APIECSClient) RegisterContainerInstance(containerInstanceArn string, attributes []*ecs.Attribute,
+	tags []*ecs.Tag, registrationToken string, platformDevices []*ecs.PlatformDevice,
+	outpostARN string) (string, string, error) {
+
 	clusterRef := client.config.Cluster
 	// If our clusterRef is empty, we should try to create the default
 	if clusterRef == "" {
@@ -122,7 +124,8 @@ func (client *APIECSClient) RegisterContainerInstance(containerInstanceArn strin
 		}()
 		// Attempt to register without checking existence of the cluster so we don't require
 		// excess permissions in the case where the cluster already exists and is active
-		containerInstanceArn, availabilityzone, err := client.registerContainerInstance(clusterRef, containerInstanceArn, attributes, tags, registrationToken, platformDevices)
+		containerInstanceArn, availabilityzone, err := client.registerContainerInstance(clusterRef,
+			containerInstanceArn, attributes, tags, registrationToken, platformDevices, outpostARN)
 		if err == nil {
 			return containerInstanceArn, availabilityzone, nil
 		}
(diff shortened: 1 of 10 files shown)


agent/app/agent.go
```diff
@@ -435,6 +435,17 @@ func (agent *ecsAgent) getEC2InstanceID() string {
 	return instanceID
 }
 
+// getoutpostARN gets the Outpost ID from the metadata service
```
petderek: > getoutpostARN gets the Outpost ID

Is it outpost id? outpost arn? aws id (from your commit message)?  Please make everything match.

shubham2892: Fixed.

agent/api/ecsclient/client.go
```diff
@@ -177,6 +185,7 @@ func (client *APIECSClient) registerContainerInstance(clusterRef string, contain
 	registerRequest.TotalResources = resources
 
 	registerRequest.ClientToken = &registrationToken
+	seelog.Info(registerRequest)
```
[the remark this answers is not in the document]
shubham2892: Yes removed.

agent/app/agent.go
```diff
@@ -435,6 +435,17 @@ func (agent *ecsAgent) getEC2InstanceID() string {
 	return instanceID
 }
 
+// getoutpostARN gets the Outpost ARN from the metadata service
+func (agent *ecsAgent) getoutpostARN() string {
+	outpostARN, err := agent.ec2MetadataClient.OutpostARN()
+	if err != nil {
+		seelog.Warnf(
+			"Unable to access EC2 Metadata service to determine Outpost ID: %v", err)
+		return ""
```
sharanyad: 1.would this also be the case for a non-outpost instance?

2. what happens when an actual outpost does not have an ARN in the backend?

shubham2892: > 1.would this also be the case for a non-outpost instance?
Yes, this will happen for the non-outpost instance. There is another check which ensures that if outpost ARN is empty it will not be sent in RCI call. 
> 
>     1. what happens when an actual outpost does not have an ARN in the backend?
In that case, it will be treated like a non-outpost instance. Outpost ARN is how backend will know it's running on outpost.

agent/api/ecsclient/client_test.go
```diff
@@ -356,7 +358,7 @@ func TestReRegisterContainerInstance(t *testing.T) {
 			assert.True(t, ok, `Could not find resource "PORTS_UDP"`)
 			assert.Equal(t, "STRINGSET", *resource.Type, `Wrong type for resource "PORTS_UDP"`)
 			// "ecs.os-type" and the 2 that we specified as additionalAttributes
```
yumex93: nit: The comment needs to be updated.

agent/api/ecsclient/client_test.go
```diff
@@ -435,7 +439,7 @@ func TestRegisterContainerInstance(t *testing.T) {
 			assert.True(t, ok, `Could not find resource "PORTS_UDP"`)
 			assert.Equal(t, "STRINGSET", *resource.Type, `Wrong type for resource "PORTS_UDP"`)
 			// 3 from expectedAttributes and 2 from additionalAttributes
```
yumex93: same here. The comments are not correct anymore

agent/app/agent.go
```diff
@@ -435,6 +435,17 @@ func (agent *ecsAgent) getEC2InstanceID() string {
 	return instanceID
 }
 
+// getoutpostARN gets the Outpost ARN from the metadata service
+func (agent *ecsAgent) getoutpostARN() string {
+	outpostARN, err := agent.ec2MetadataClient.OutpostARN()
+	if err != nil {
+		seelog.Warnf(
+			"Unable to access EC2 Metadata service to determine Outpost ID: %v", err)
```
yumex93: nit: Can you update the message to be Outpost ARN?

agent/app/agent.go
sharanyad: can we edit the warn message to something like "Unable to obtain Outpost ARN from EC2 Metadata"?  Current one makes it looks like EC2 metadata is down.

shubham2892: Yes will do.

agent/api/ecsclient/client_test.go
```diff
@@ -499,7 +505,7 @@ func TestRegisterContainerInstanceNoIID(t *testing.T) {
...
```
yumex93: Can you update this comment before you merge the pr?

shubham2892: Yes.

agent/app/agent.go
```diff
@@ -435,6 +435,17 @@ func (agent *ecsAgent) getEC2InstanceID() string {
 	return instanceID
 }
 
+// getoutpostARN gets the Outpost ARN from the metadata service
+func (agent *ecsAgent) getoutpostARN() string {
```
sharanyad: synced offline. let's move this to `agent_unix.go` to make the IMDS call, since this is not supported in Windows.

Issue #591, opened 2016-11-07
ecs-agent needs to reopen docker.sock on docker restart


rhuddleston:
We have ""live-restore": true" in our docker daemon.json configuration file. This allows us to restart docker without restarting all the containers it's running to change docker configuration etc.

With ecs-agent though it seems  it must keep the old docker.sock file open as after the docker restart it no longer recognized when a service exits until I restart ecs-agent and it reopens the docker.sock.

Steps to reproduce:

• ensure ecs-agent is working correctly, try to kill off running tasks and notice ecs immediately rescedules the tasks to run again on an ecs node. 
• add live-restore to /etc/docker/daemon.json:
{
  "live-restore": true
}
• sudo /bin/kill -SIGHUP $(pidof dockerd)
• sudo service docker restart
• try again to  kill off running tasks. Notice ecs-agent does not recogzixe that task/container exited and ecs still thinks the process is executing. It will stay in this state until ecs-agent is restarts.

Please ensure ecs-agent recognizes changed to docker.sock and reopens it when it changes. I tested this with two other programs that also use docker.sock file (e.g. registrator) and verified it works correctly through a docker restart without issue

richardpen:
@rhuddleston Thanks for your reporting. Currently ecs agent doesn't have a mechnism to reopen docker socket if the previous connection is closed which happened when docker restart.

The docker live-restore feature is supported in docker 1.12.0, while ECS currently only support docker 1.11.2. I'll mark this as a feature request. Currently you can get around this by restart the ecs-agent container.

Thanks,
Peng

markuman:
When talking about high-availability, than this is a must have feature.

adnxn:
closing, released with version [v1.32.1](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.32.1)

PR #2240, opened 2019-10-16
Reopen Docker event stream when getting EOF error

fenxiong:
### Summary

Fix https://github.com/aws/amazon-ecs-agent/issues/591.

Previously, agent did not reopen docker event stream when it gets EOF/UnexpectedEOF error. When fsouza/go-dockerclient was used, we had this behavior because fsouza/go-dockerclient can't reopen the stream upon EOF/UnexpectedEOF due to some of its own implementation details (see https://github.com/fsouza/go-dockerclient/pull/171) and we did not retry in our layer. When migrated to docker/docker/client, this behavior was maintained.

In the case that Docker is restarted (and container live-restore is enabled), agent's event stream listening goroutine gets an io.UnexpectedEOF and finishes, and we never open the stream again and as a result loss all Docker events until the agent is restarted. So making this changes to reopen Docker event stream when getting EOF error.

### Implementation details

Few changes in ContainerEvents.

### Testing

Updated the TestContainerEventsEOFError to conform with new behavior.

Manually tested with the steps to reproduce described in https://github.com/aws/amazon-ecs-agent/issues/591 and verified that it's fixed.

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -863,22 +863,26 @@ func (dg *dockerGoClient) ContainerEvents(ctx context.Context) (<-chan DockerCon
 		for {
 			select {
 			case err := <-eventErr:
-				if err == io.EOF || err == io.ErrUnexpectedEOF {
-					seelog.Info("DockerGoClient: All events have been received")
-					cancel()
+				// If parent ctx has been canceled, stop listening and return. Otherwise reopen the stream.
+				if ctx.Err() != nil {
 					return
-				} else {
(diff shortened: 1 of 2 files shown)


fenxiong:
> Is error `io.UnexpectedEOF` only expected when docker restarts, are there any other cases when we get the same error, that we need to handle differently?

It can happen as long as the connection to docker socket is closed. IMO the agent should always attempt to reopen Docker event stream, since we rely on that to detect changes from customer's container. I can't think of a situation where we want the agent to stop listening to Docker events other than the case that the task engine is shut down - in that case the context will be canceled and we will stop listening.

fierlion:
These changes look good to me.

Issue #1708, opened 2018-11-29
StartContainer edge case


petderek:
### Summary
We've observed that occasionally, go-dockerclient will return an EOF instead of a complete response from Docker. Our StartContainer logic does not account for this, and the task gets marked as 'failed to start' when in reality the containers may be running. Agent should be modified to check for this failure state.

### Expected Behavior
Either:

1) Containers do not start, task gets marked as failed.

or

2) Containers start, agent checks if they actually started, task marks as success.

### Observed Behavior

* Docker API returns invalid response
* Docker daemon starts container
* Agent marks task as failed to start
* Container runs anyway, agent does not track or stop it

### Supporting Log Snippets

Logs will contain messaging similar to:

```
error transitioning container [<container_name>] to [RUNNING]: Post http://unix.sock/v1.19/containers/<container_id>/start EOF 
```

yunhee-l:
Latest agent release ([v1.24.0](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.24.0)) has replaced go-dockerclient with Docker's official SDK. @mfortin is upgrading to the latest release an option for you?

mfortin:
@yunhee-l It is. I will be testing it out starting Monday. Thanks!

mfortin:
I did. Now this is new:
CannotStartContainerError: error during connect: Post http://%2Fvar%2Frun%2Fdocker.sock/v1.19/containers/eaeca3d1c68c067c0963a2a48785a65b1b5961acd9c8c6f72c7e316afef11442/start: EOF

adnxn:
> I have grabbed the logs using ecs-logs-collector if you want to have a look.

@mfortin: mind sending the logs to adnkha at amazon dot com?

adnxn:
closing, released with version [v1.32.1](https://github.com/aws/amazon-ecs-agent/releases/tag/v1.32.1)

PR #2245, opened 2019-10-22
Issue a stop when start container failed with EOF error

fenxiong:
### Summary

Fix https://github.com/aws/amazon-ecs-agent/issues/1708. When start container failed with EOF, there's a chance that the container is started anyway, so issue a stop for the container in that case.

### Implementation details

The stop is issued in `handleEventErroor`, in the same way as how we issue a stop when start container times out. I can't find a typed error from Docker or golang for this, so check is based on the error string.

### Testing

Unit test added. Manually tested by injecting the error in docker client and verified that the container is forced to stop.

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -15,6 +15,8 @@ package engine
 
 import (
 	"context"
+	"io"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


agent/engine/task_manager.go
```diff
@@ -57,6 +58,9 @@ var (
 	_stoppedSentWaitInterval       = stoppedSentWaitInterval
 	_maxStoppedWaitTimes           = int(maxStoppedWaitTimes)
 	taskNotWaitForSteadyStateError = errors.New("managed task: steady state check context is nil")
+	// Based on https://github.com/aws/amazon-ecs-agent/blob/master/agent/vendor/github.com/docker/docker/client/request.go#L182
+	startContainerEOFErrorRegex = "error during connect: Post .+: EOF"
```
adnxn: should we catch all EOFs responses for the startContainer workflow and handle them similarly? vs matching on that  first part of the error string as well. wyt?

fenxiong: yeah make sense. updated to check that error ends in EOF, instead of hardcoding the error as a regex

agent/engine/task_manager_test.go
```diff
@@ -139,6 +139,18 @@ func TestHandleEventError(t *testing.T) {
 			ExpectedContainerDesiredStatusStopped: true,
 			ExpectedOK:                            false,
 		},
+		{
+			Name:                        "Start EOF",
```
adnxn: nit: more descriptive naming for these test cases please

fenxiong: sure, updated

agent/engine/task_manager.go
```diff
@@ -666,10 +668,20 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
...
+		errorStr := event.Error.Error()
+		shouldForceStop := false
 		if errorName == dockerapi.DockerTimeoutErrorName || errorName == dockerapi.CannotInspectContainerErrorName {
 			// If there's an error with inspecting the container or in case of timeout error,
-			// we'll also assume that the container has transitioned to RUNNING and issue
+			// we'll assume that the container has transitioned to RUNNING and issue
 			// a stop. See #1043 for details
+			shouldForceStop = true
+		} else if errorName == dockerapi.CannotStartContainerErrorName && strings.HasSuffix(errorStr, io.EOF.Error()) {
+			// If we get an EOF error from Docker when starting the container, we don't really know whether the
```
sharanyad: do we know when does an EOF error occur?

fenxiong: From my understanding, an EOF means the client didn't read anything from the response, which can happen if the connection closes before server writes the response.

Issue #1873, opened 2019-02-20
ECS agent should start with `--net=host` for all functional tests


aaithal:
### Summary/Description

Most of the functional tests start the agent with `bridge` networking mode. This runs contrary to how the agent is actually started in production. We should also get rid of the `ECS_FTEST_FORCE_NET_HOST` env var while we're addressing this.

petderek:
We have no real reason to test on bridge mode anymore, correct?

aaithal:
> We have no real reason to test on bridge mode anymore, correct?

Yeah, I think so (unless there are some conditional tests that require old agent versions).

fenxiong:
done in https://github.com/aws/amazon-ecs-agent/pull/2253

PR #2247, opened 2019-10-23
EFS functional testing

sparrc:
### Summary

Adds a functional test testing writing/reading from EFS volumes mounted via Docker's NFS integration.

### Implementation details

In this test we will setup an EFS volume with mount target. We will then mount this into two tasks: first a task that writes to the EFS volume, then a task that reads from the EFS volume.

Both tasks exit with code `42` if they are successful, so we check for this exit code to verify that the test passed.

Currently this is implemented using the existing `dockerVolumeConfiguration` model. When the feature is ready this will be changed to the `EFSVolumeConfiguration` model.

EFS volume configuration functional testing

### Licensing


agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1359,6 +1360,106 @@ func TestASMSecretsARN(t *testing.T) {
...
+//   1. creates an EFS filesystem with a mount target.
+//   2. spins up a task to mount and write to the filesystem.
+//   3. spins up a task to mount and read from the filesystem.
+//   4. TODO: do this via EFSVolumeConfiguration instead of via NFS/Docker.
+func TestRunEFSVolumeTask(t *testing.T) {
+	if os.Getenv("TEST_DISABLE_EXECUTION_ROLE") == "true" {
+		t.Skip("TEST_DISABLE_EXECUTION_ROLE was set to true")
+	}
+
+	if IsCNPartition() {
```
sharanyad: why is this needed?

sparrc: this URL will not work in the CN partition: `addr=FILESYSTEM_ID.efs.TEST_REGION.amazonaws.com`, there is currently a TODO in the feature branch (and design doc) to deal with this in EFSVolumeConfiguration (see https://github.com/aws/amazon-ecs-agent/pull/2234/files#diff-a617e6edf02b5af84b984ccb06ce315fR547)

agent/Gopkg.toml
```diff
@@ -19,7 +19,7 @@
 #  name = "github.com/x/y"
 #  version = "2.4.0"
 
-required = ["github.com/golang/mock/mockgen/model"]
+required = ["github.com/golang/mock/mockgen/model", "github.com/aws/aws-sdk-go/service/efs"]
```
adnxn: why is this separate from the regular aws-sdk-go constraint?

sparrc: Yes I'm going to figure out how to resolve this. The reason was that we are only using the efs client in a linux-specific unit test, so I wasn't able to get dep to recognize the dependency from my work laptop.

sparrc: this is resolved now, deleting the entry and running "dep ensure" again seems to have fixed it

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1359,6 +1360,126 @@ func TestASMSecretsARN(t *testing.T) {
...
+	creationToken := "efs-func-tests"
+	_, err := efsClient.CreateFileSystem(&efs.CreateFileSystemInput{
+		CreationToken: aws.String(creationToken),
+	})
+
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case efs.ErrCodeFileSystemAlreadyExists:
+				t.Logf("EFS filesystem already exists")
```
sharanyad: Do we expect to reuse the same file system created once for the tests? I don't see any cleanup logic.

sparrc: yes, we expect to reuse the filesystem, that's why we check for ErrCodeFileSystemAlreadyExists (filesystem already exists) and ErrCodeMountTargetConflict (mount target already exists)

The main reason for this is speed, and the EFS api makes it easy with the CreationToken: https://docs.aws.amazon.com/efs/latest/ug/API_CreateFileSystem.html#API_CreateFileSystem_RequestSyntax

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1359,6 +1360,126 @@ func TestASMSecretsARN(t *testing.T) {
...
+//   2. spins up a task to mount and write to the filesystem.
+//   3. spins up a task to mount and read from the filesystem.
+//   4. TODO: do this via EFSVolumeConfiguration instead of via NFS/Docker.
+func TestRunEFSVolumeTask(t *testing.T) {
+	if os.Getenv("TEST_DISABLE_EXECUTION_ROLE") == "true" {
+		t.Skip("TEST_DISABLE_EXECUTION_ROLE was set to true")
+	}
+
+	if IsCNPartition() {
+		t.Skip("Skip TestSSMSecretsEncryptedParameter in China partition")
```
fenxiong: TestSSMSecretsEncryptedParameter -> TestRunEFSVolumeTask

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1359,6 +1360,126 @@ func TestASMSecretsARN(t *testing.T) {
 	assert.Equal(t, 42, exitCode, fmt.Sprintf("Expected exit code of 42; got %d", exitCode))
 }
 
+// TestRunEFSVolumeTask does the following:
+//   1. creates an EFS filesystem with a mount target.
+//   2. spins up a task to mount and write to the filesystem.
+//   3. spins up a task to mount and read from the filesystem.
+//   4. TODO: do this via EFSVolumeConfiguration instead of via NFS/Docker.
+func TestRunEFSVolumeTask(t *testing.T) {
+	if os.Getenv("TEST_DISABLE_EXECUTION_ROLE") == "true" {
```
fenxiong: why is this check needed?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1359,6 +1360,126 @@ func TestASMSecretsARN(t *testing.T) {
...
+		t.Skip("TEST_DISABLE_EXECUTION_ROLE was set to true")
+	}
+
+	if IsCNPartition() {
+		t.Skip("Skip TestSSMSecretsEncryptedParameter in China partition")
+	}
+
```
fenxiong: i think you should have the following to let agent runs as host network mode (see https://github.com/aws/amazon-ecs-agent/issues/1873)
```
os.Setenv("ECS_FTEST_FORCE_NET_HOST", "true")
defer os.Unsetenv("ECS_FTEST_FORCE_NET_HOST")
```

agent/functional_tests/tests/functionaltests_unix_test.go
petderek: update this to say EFS :)

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1359,6 +1360,126 @@ func TestASMSecretsARN(t *testing.T) {
...
+		"TEST_REGION":   *ECS.Config.Region,
+	}
+	wTask, err := agent.StartTaskWithTaskDefinitionOverrides(t, "task-efs-vol-write", overrides)
+	require.NoError(t, err, "Register task definition failed")
+	// Wait for the first task to create the volume
+	wErr := wTask.WaitStopped(waitTaskStateChangeDuration)
+	require.NoError(t, wErr, "Error waiting for task to transition to STOPPED")
+	wExitCode, ok := wTask.ContainerExitcode("task-efs-vol-write")
+	require.True(t, ok, "Error code for container [task-efs-vol-write] not found, check the logs")
+	assert.Equal(t, 42, wExitCode, fmt.Sprintf("Expected exit code of 42; got %d", wExitCode))
```
fenxiong: nit: i think this can just be require since there's no need to continue if this fails?

agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -1359,6 +1360,126 @@ func TestASMSecretsARN(t *testing.T) {
...
+				require.NoError(t, err, "Unexpected error from DescribeFileSystems: %s", err)
+				return *out.FileSystems[0].FileSystemId
+			default:
+				require.NoError(t, err, "Error creating EFS Filesystem")
+			}
+		}
+	}
+
+	// Wait for filesystem to be "available"
+	for true {
```
petderek: I'd prefer if this had a finite timeout -- like try 10 times or so.

agent/functional_tests/util/utils.go
```diff
@@ -809,8 +809,8 @@ func AttributesToMap(attributes []*ecs.Attribute) map[string]string {
 	return attributeMap
 }
 
-// getSubnetID gets the subnet id for the instance from ec2 instance metadata
-func getSubnetID() (string, error) {
+// GetSubnetID gets the subnet id for the instance from ec2 instance metadata
+func GetSubnetID() (string, error) {
```
petderek: nice

agent/functional_tests/tests/functionaltests_unix_test.go
sparrc: will do

agent/functional_tests/tests/functionaltests_unix_test.go
sparrc: okay, will fix that

agent/functional_tests/tests/functionaltests_unix_test.go
sparrc: 👍

agent/functional_tests/tests/functionaltests_unix_test.go
sparrc: 👍

agent/functional_tests/tests/functionaltests_unix_test.go
sparrc: That's probably a c&p artifact, I'll remove it

yumex93:
LGTM. Don't forget to squash your commits before merging to dev.

PR #2248, opened 2019-10-24
Add efs capability

sparrc:
just adding the "efs" capability to the feature branch.

### DIFF
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -55,6 +55,7 @@ const (
 	capabilityFirelensLoggingDriver             = "logging-driver.awsfirelens"
 	capabilityFirelensConfigFile                = "firelens.options.config.file"
 	capabilityFirelensConfigS3                  = "firelens.options.config.s3"
+	capabilityEFS                               = "efs"
 )
 
 // capabilities returns the supported capabilities of this agent / docker-client pair.
@@ -95,6 +96,7 @@ const (
 //    ecs.capability.task-eia.optimized-cpu
 //    ecs.capability.firelens.fluentd
 //    ecs.capability.firelens.fluentbit
+//    ecs.capability.efs
 //    com.amazonaws.ecs.capability.logging-driver.awsfirelens
 //    ecs.capability.firelens.options.config.file
 //    ecs.capability.firelens.options.config.s3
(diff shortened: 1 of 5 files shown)


agent/app/agent_capability_unix.go
```diff
@@ -153,6 +153,10 @@ func (agent *ecsAgent) appendFirelensFluentbitCapabilities(capabilities []*ecs.A
 	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityFirelensFluentbit)
 }
 
+func (agent *ecsAgent) appendEFSCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
```
sharanyad: missing calling of this method

sharanyad: needs to be there in `agent_capability_windows.go` as well

sparrc: I thought we are only supporting linux for now?

sharanyad: oh yes! just a dummy method returning attributes.

sparrc: I see now, pushed

adnxn:
same changes as this? https://github.com/aws/amazon-ecs-agent/pull/2119

nit: mb remove those redundant test cases as well?

sparrc:
@adnxn yes that looks related but this is the most up-to-date capability, confirmed with CP.

AFAICT that separated EIA unit test has been removed already in some other PR

agent/app/agent_capability.go
```diff
@@ -185,6 +186,9 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	// support aws router capabilities for log driver router
 	capabilities = agent.appendFirelensLoggingDriverCapabilities(capabilities)
 
+	// support efs on ecs capabilities
+	capabilities = agent.appendEFSCapabilities(capabilities)
```
fenxiong: nit: update comment above capabilities() to include new one

sparrc: to include the new one of what?

fenxiong: new capabiliity "ecs.capability.efs"

sparrc: got it, pushed

agent/app/agent_capability_unix.go
sparrc: done

PR #2249, opened 2019-10-24
api/task: initialize config in hostConfig when it is nil

yumex93:
Fix the bug that when Config is nil, agent will panic with error **panic: assignment to entry in nil map**.
The Config attribute in hostConfig can be nil under some situations. For example, if in task definition, for log driver, only secretOptions are specified while options is null. 

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/task/task_test.go
+++ b/agent/api/task/task_test.go
@@ -1602,9 +1602,56 @@ func TestApplyExecutionRoleLogsAuthSet(t *testing.T) {
 	}
 
 	rawHostConfig, err := json.Marshal(&rawHostConfigInput)
-	if err != nil {
-		t.Fatal(err)
+	require.NoError(t, err)
+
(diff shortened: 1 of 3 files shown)


agent/api/task/task_test.go
```diff
@@ -1638,6 +1638,57 @@ func TestApplyExecutionRoleLogsAuthSet(t *testing.T) {
...
+	expectedEndpoint := "/v2/credentials/" + credentialsIDInTask
+
+	rawHostConfigInput := dockercontainer.HostConfig{
+		LogConfig: dockercontainer.LogConfig{
+			Type:   "foo",
+		},
+	}
+
+	rawHostConfig, err := json.Marshal(&rawHostConfigInput)
+	if err != nil {
```
fenxiong: require.NoError?

Issue #2087, opened 2019-06-17
ECS_ENABLE_UNTRACKED_IMAGE_CLEANUP does not respect the ECS_IMAGE_MINIMUM_CLEANUP_AGE setting


AaronKalair:
### Summary

When the ECS Agent runs it will delete images that the ECS Agent did not create that are younger than the `ECS_IMAGE_MINIMUM_CLEANUP_AGE` setting.

### Description

If you set
*  `ECS_ENABLED_UNTRACKED_IMAGE_CLEANUP` to `true`
* `ECS_IMAGE_MINIMUM_CLEANUP_AGE` to anything lets say `8h` for example

When the ECS Agent runs it will delete images that the ECS Agent did not create that are younger than the `ECS_IMAGE_MINIMUM_CLEANUP_AGE` setting.

Looking through the code, it appears that the check for image age:

https://github.com/aws/amazon-ecs-agent/blob/7e013ce5735a310315278ad8d3027ebf451574b7/agent/engine/docker_image_manager.go#L241

Is only called for checking images that the ECS agent tracks are older than the specified threshold.

Is this intentional? Is is possible to specify the `MINIMUM_CLEANUP_AGE` for `UNTRACKED_IMAGES` ?

We're using ECS to run Jenkins agents and in our Jenkins job are building Docker images, which are then referenced in a later step, however if the ECS Agent runs between us creating the image and then using it, the image will no longer exist as it has been cleaned up.

### Expected Behavior
* ECS Agent respects the `ECS_IMAGE_MINIMUM_CLEANUP_AGE` setting for Non ECS Tracked Images.

### Observed Behavior
* ECS Agent does not respect the `ECS_IMAGE_MINIMUM_CLEANUP_AGE` setting for Non ECS Tracked Images

### Environment Details
* ECS Agent 1.28.0 on AWS Linux 2

petderek:
Hey,

Thanks for reaching out to us. The ECS agent will track when an image was last used, and it uses that information to determine when a managed image is ready to be cleaned up.

The challenge here is that Docker does not track the last time an image was used. We'd definitely like to improve this feature for your use case, though.

AaronKalair:
Thanks for replying, I see why it doesn't work for us at the moment.

For our use case, clearing up the image based on how long ago it was created would be fine, I believe Docker does expose that information?

cyastella:
[...]

PR #2251, opened 2019-10-28
NonECSImageOldEnough removal

cyastella:
### Summary

ECS Agent respects the NON_ECS_IMAGE_MINIMUM_CLEANUP_AGE setting for Non ECS Tracked Images removal.

Currently, customer cannot set the certain age when the nonECS images are deleted. And non ecs images are deleted periodically. 

This enhancement gives the customer an option to set NON_ECS_IMAGE_MINIMUM_CLEANUP_AGE env to let customers cleanup the NonECS Images which is older than the certain age (the wait duration between current time and image creation time) .

 The other nonECS images will remain undeleted until they are old enough to be deleted since the image is created.   

This is related to the github issue https://github.com/aws/amazon-ecs-agent/issues/2087
### Testing

New tests cover the changes: 

### Licensing


agent/engine/docker_image_manager.go
```diff
@@ -273,6 +275,11 @@ func (imageManager *dockerImageManager) isImageOldEnough(imageState *image.Image
 	return ageOfImage > imageManager.minimumAgeBeforeDeletion
 }
 
+func (imageManager *dockerImageManager) nonECSImageOldEnough(NonECSImage ImageWithSizeID) bool {
+	ageOfImage := time.Now().Sub(NonECSImage.createdTime)
```
sparrc: why is this using "createdTime" instead of "PulledAt" time (that ecs images are using)? is there much difference between these?

sharanyad: it should be `PulledAt`

agent/config/config.go
```diff
@@ -83,6 +83,10 @@ const (
 	// has been pulled before it can be deleted.
 	DefaultImageDeletionAge = 1 * time.Hour
 
+	// DefaultNonECSImageDeletionAge specifies the default value for minimum amount of elapsed time after an image
+	// has been created before it can be deleted
+	DefaultNonECSImageDeletionAge = 1 * time.Hour
```
petderek: do we need a new constant for this?

agent/engine/docker_image_manager.go
petderek: PulledAt may be null if the image was created locally -- I think thats the idea here.

petderek: However, you may need to take both into consideration. if you go onto the host and 'docker pull busybox' its 'create' date may be days ago.

sharanyad: hmm, this kinda breaks the logic of "old enough". an image could have been created a long time ago, but could have been used very recently.

cyastella: Do we want to track the created time or last used time of the image?

agent/config/config.go
cyastella: I will combine this with the constant above together.

sharanyad: imo, it makes sense to have them separate, since they have different semantics (pulledAt/ createdAt)

agent/engine/docker_image_manager.go
cyastella: Since from api response, we only know the the created time. Will change the env var to make more clear to let customer set the variable.

agent/config/config.go
cyastella: will make this separate since the nonECS image removal track the created time.

agent/engine/docker_image_manager.go
petderek: I think the problem is right now it will remove regardless of any age.

agent/config/config.go
```diff
@@ -520,6 +524,7 @@ func environmentConfig() (Config, error) {
 		TaskIAMRoleEnabledForNetworkHost:    utils.ParseBool(os.Getenv("ECS_ENABLE_TASK_IAM_ROLE_NETWORK_HOST"), false),
 		ImageCleanupDisabled:                utils.ParseBool(os.Getenv("ECS_DISABLE_IMAGE_CLEANUP"), false),
 		MinimumImageDeletionAge:             parseEnvVariableDuration("ECS_IMAGE_MINIMUM_CLEANUP_AGE"),
+		NonECSMinimumImageDeletionAge:       parseEnvVariableDuration("NON_ECS_IMAGE_MINIMUM_CLEANUP_WAIT_DURATION_SINCE_CREATED"),
```
yumex93: Can we just name it as **NON_ECS_IMAGE_MINIMUM_CLEANUP_AGE** since it serves the same purpose of **ECS_IMAGE_MINIMUM_CLEANUP_AGE**? And for more implementation details, we just keep it in README and code comments.

agent/engine/docker_image_manager.go
yumex93: Can we check **PulledAt** first and if it is null, check the **createdTime**? Since docker does not rly have the feature to tell us last used time of image, we just assume **PulledAt** time for images that pulled and **createdTime** for images created on the host as last used time.

petderek: @cyastella can you add some notes to the git commit message and PR summary explaining what problem we have now and what problem we are solving?

agent/config/config.go
sharanyad: how about `NON_ECS_IMAGE_CLEANUP_MINIMUM_CREATION_AGE` . I suggest we add some context to the env var name about how we are using creation time for cleanup logic

agent/engine/docker_image_manager.go
cyastella: yeah, will add it

yumex93:
Since we also need to update README, and this is just a small feature, let's keep README change in the same PR so easier for people to trace back.

cyastella:
> Since we also need to update README, and this is just a small feature, let's keep README change in the same PR so easier for people to trace back.

Yeah, will add it to this commit.

README.md
```diff
@@ -155,6 +155,7 @@ additional details on each available environment variable.
 | `ECS_DISABLE_IMAGE_CLEANUP` | `true` | Whether to disable automated image cleanup for the ECS Agent. | `false` | `false` |
 | `ECS_IMAGE_CLEANUP_INTERVAL` | 30m | The time interval between automated image cleanup cycles. If set to less than 10 minutes, the value is ignored. | 30m | 30m |
 | `ECS_IMAGE_MINIMUM_CLEANUP_AGE` | 30m | The minimum time interval between when an image is pulled and when it can be considered for automated image cleanup. | 1h | 1h |
+| `NON_ECS_IMAGE_MINIMUM_CLEANUP_AGE` | 30m | The minimum time interval between when an non ECS image is created and when it can be considered for automated image cleanup. | 1h | 1h |
```
sharanyad: nit: `when a non`

agent/engine/docker_image_manager_test.go
```diff
@@ -1253,6 +1253,124 @@ func TestNonECSImageAndContainersCleanupRemoveImage_DontDeleteExcludedImage(t *t
...
+	client.EXPECT().ListImages(gomock.Any(), dockerclient.ListImagesTimeout).Return(listImagesResponse).AnyTimes()
+	client.EXPECT().InspectImage(listImagesResponse.ImageIDs[0]).Return(inspectImageResponse, nil).Times(1)
+	client.EXPECT().RemoveImage(gomock.Any(), listImagesResponse.ImageIDs[0], dockerclient.RemoveImageTimeout).Return(nil).Times(0)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	imageManager.removeUnusedImages(ctx)
+
+	assert.Len(t, listContainersResponse.DockerIDs, 1, "Error removing container IDs")
```
sharanyad: since `listContainersResponse` , `inspectContainerResponse` are defined already, what exactly are we checking here?

agent/engine/docker_image_manager_test.go
```diff
@@ -1253,6 +1253,124 @@ func TestNonECSImageAndContainersCleanupRemoveImage_DontDeleteExcludedImage(t *t
...
+		ImageIDs: []string{"sha256:qwerty1"},
+		RepoTags: []string{"tester"},
+	}
+
+	client.EXPECT().ListContainers(gomock.Any(), gomock.Any(), dockerclient.ListImagesTimeout).Return(listContainersResponse).AnyTimes()
+	client.EXPECT().InspectContainer(gomock.Any(), gomock.Any(), dockerclient.InspectContainerTimeout).Return(inspectContainerResponse, nil).AnyTimes()
+	client.EXPECT().RemoveContainer(gomock.Any(), gomock.Any(), dockerclient.RemoveContainerTimeout).Return(nil).Times(1)
+	client.EXPECT().ListImages(gomock.Any(), dockerclient.ListImagesTimeout).Return(listImagesResponse).AnyTimes()
+	client.EXPECT().InspectImage(listImagesResponse.ImageIDs[0]).Return(inspectImageResponse, nil).Times(1)
+	client.EXPECT().RemoveImage(gomock.Any(), listImagesResponse.ImageIDs[0], dockerclient.RemoveImageTimeout).Return(nil).Times(0)
```
sharanyad: what is the removeImage for, when we don't expect the image to be deleted?

agent/engine/docker_image_manager_test.go
cyastella: will remove it

agent/engine/docker_image_manager_test.go
cyastella: should remove it.

agent/engine/docker_image_manager.go
```diff
@@ -273,6 +275,12 @@ func (imageManager *dockerImageManager) isImageOldEnough(imageState *image.Image
 	return ageOfImage > imageManager.minimumAgeBeforeDeletion
 }
 
+//TODO: change createdTime to pullAtTime when docker support it in the future
```
yumex93: I think it would be more reasonable in **TODO** that change createdTime to image last used time since this is what we really want to keep track of.

agent/engine/docker_image_manager_test.go
```diff
@@ -1253,6 +1253,119 @@ func TestNonECSImageAndContainersCleanupRemoveImage_DontDeleteExcludedImage(t *t
...
+	client.EXPECT().ListImages(gomock.Any(), dockerclient.ListImagesTimeout).Return(listImagesResponse).AnyTimes()
+	client.EXPECT().InspectImage(listImagesResponse.ImageIDs[0]).Return(inspectImageResponse, nil).Times(1)
+	client.EXPECT().RemoveImage(gomock.Any(), listImagesResponse.ImageIDs[0], dockerclient.RemoveImageTimeout).Return(nil).Times(1)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+
+	imageManager.removeUnusedImages(ctx)
+
+	assert.Len(t, imageManager.imageStates, 0, "Error removing image state after the image is removed")
```
sharanyad: i think this assert is not needed as well, since we do not add non ecs image state to image manager. same for above

agent/engine/docker_image_manager.go
```diff
@@ -273,6 +275,12 @@ func (imageManager *dockerImageManager) isImageOldEnough(imageState *image.Image
...
+func (imageManager *dockerImageManager) nonECSImageOldEnough(NonECSImage ImageWithSizeID) bool {
+	ageOfImage := time.Now().Sub(NonECSImage.createdTime)
+	return ageOfImage >= imageManager.nonECSMinimumAgeBeforeDeletion
```
sharanyad: It does not matter very much if its `>=` or `>`. just to be consistent with `isImageOldEnough`, we can make it just `>`

agent/engine/docker_image_manager.go
```diff
@@ -476,11 +489,17 @@ func (imageManager *dockerImageManager) getNonECSImages(ctx context.Context) []I
 			seelog.Errorf("Error inspecting non-ECS image: (ImageID: %s), %s", imageID, err)
 			continue
 		}
+		createTime := time.Time{}
+		createTime, err = time.Parse(time.RFC3339, resp.Created)
+		if err != nil {
+			seelog.Errorf("Error parse the inspected non-ECS image created time (ImageID: %s), %s", imageID, err)
```
yumex93: This log level can just be WARN, considering this will not impact real functionality and we don't want to make customer panic when they see this as an error.

PR #2253, opened 2019-10-30
Start agent with host network mode in all linux functional tests

fenxiong:
### Summary

Fix https://github.com/aws/amazon-ecs-agent/issues/1873.

### Implementation details

Remove `ECS_FTEST_FORCE_NET_HOST ` and start agent with host network mode for all linux functional tests.

Few updates to make it work:
1. Since each test starts an agent, functional test can not be started in parallel at all. Adding "-p 1" option to enforce that (otherwise tests in simpletests package run in parallel with the ones in functional_tests package)
2. TestSquidProxy was using container linking and required agent to start with bridge mode. Update the test to make it to work in host mode.
3. Some fixes to address temporary DescribeContainerInstances API failures.

### Testing

Run existing tests.

### Licensing


agent/functional_tests/tests/functionaltests_unix_test.go
```diff
@@ -192,16 +192,21 @@ func TestSquidProxy(t *testing.T) {
...
-	t.Logf("Started squid container: %v", squidContainerJSON.Name)
+	t.Logf("Started squid container: %s", squidContainerJSON.Name)
 
 	agent := RunAgent(t, &AgentOptions{
 		ExtraEnvironment: map[string]string{
-			"HTTP_PROXY": "squid:3128",
+			"HTTP_PROXY": fmt.Sprintf("http://%s:3128", squidIP),
 			"NO_PROXY":   "169.254.169.254,/var/run/docker.sock",
 		},
-		ContainerLinks: []string{squidContainerJSON.Name + ":squid"},
```
fierlion: not totally clear on why this is going away.  What's the `ContainerLinks` field doing and why can we get rid of it?  (...same question applies below where this is used in the hostConfig.Links.)

fenxiong: Container link enables containers launched in bridge mode to discover each other without ip (see [here](https://docs.docker.com/network/links/#communication-across-links)). So when the agent starts with bridge mode and specify squid container as the link, it can set HTTP_PROXY to "squid_container_name:port" without the need to know the squid container's ip (Docker will resolve that). Since we now want to start agent in host mode, we can't use that anymore (container link is only supported for bridge mode), and instead we get the squid container's ip from container inspect result and use it in HTTP_PROXY

shubham2892: As far as I understand, containerLink made it possible to set HTTP_PROXY using _squid_ word, as the name of the proxy squid container is _squid_, instead we can just pass the ip address on which squid proxy is listening to.

Issue #800, opened 2017-05-11
Support Dockerfile HEALTHCHECK of agent


jhmartin:
Docker 1.12 and newer supports a HEALTHCHECK directive that allows the docker daemon to detect the health of an application inside a container. Please consider adding one to the agent Dockerfile.

Docs: https://docs.docker.com/engine/reference/builder/#healthcheck

samuelkarp:
Hi @jhmartin, thanks for reaching out.  This issue is a duplicate of https://github.com/aws/amazon-ecs-agent/issues/534, and discussion is happening there.  If you have any comments to add to the discussion, please add it on https://github.com/aws/amazon-ecs-agent/issues/534.  Thanks!

jhmartin:
@samuelkarp It is different. #534 suggests allowing the ecs-agent to use the HEALTHCHECK output to detect the state of containers managed by ECS. This issue is for the ECS agent *itself* to specify a HEALTHCHECK.

samuelkarp:
Ah, I think I missed that in your description.  My apologies for the premature close.

The agent's Dockerfile is [here](https://github.com/aws/amazon-ecs-agent/blob/dev/scripts/dockerfiles/Dockerfile.release) if you're interested in making a contribution.

adnxn:
@jhmartin, We've added task def support for the [health check command](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_healthcheck) and associated configuration parameters for the container. This parameter maps to HealthCheck in the Create a container section of the Docker Remote API and the HEALTHCHECK parameter of docker run.

cyastella:
closing issue since a new set of amis have been released and can be tracked [here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-ami-versions.html)

PR #2260, opened 2019-11-04
Support Dockerfile HEALTHCHECK of agent

cyastella:
### Summary

Address #800. Previously, ECS does not have a Healthcheck directive for itself. This PR add the healthcheck in the dockerfile to enable detect the health status of ECS agent.
### Implementation details

Enable ECS agent healthcheck flag 
Add the healthcheck command in the dockerfile. 
### Licensing

### DIFF
--- a/agent/app/args/flag.go
+++ b/agent/app/args/flag.go
@@ -23,6 +23,7 @@ const (
 	licenseUsage             = "Print the LICENSE and NOTICE files and exit"
 	blacholeEC2MetadataUsage = "Blackhole the EC2 Metadata requests. Setting this option can cause the ECS Agent to fail to work properly.  We do not recommend setting this option"
 	windowsServiceUsage      = "Run the ECS agent as a Windows Service"
+	healthcheckServiceUsage  = "Run the agent healthcheck"
 
 	versionFlagName              = "version"
 	logLevelFlagName             = "loglevel"
@@ -31,6 +32,7 @@ const (
 	licenseFlagName              = "license"
 	blackholeEC2MetadataFlagName = "blackhole-ec2-metadata"
 	windowsServiceFlagName       = "windows-service"
+	healthCheckFlagName          = "healthcheck"
 )
 
 // Args wraps various ECS Agent arguments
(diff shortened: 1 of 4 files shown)


samuelkarp:
Instead of mounting in another binary, can you extend the existing `agent` binary inside the container to be able to report health of the container?

petderek:
`/agent --healthcheck` would be appropriate, and makes it a bit more extensible.

agent/app/agent.go
```diff
@@ -196,6 +199,15 @@ func (agent *ecsAgent) printECSAttributes() int {
 	return exitcodes.ExitSuccess
 }
 
+func (agent *ecsAgent) runHealthcheck() int {
+	_, err := http.Get("http://localhost:51678/v1/metadata")
+	if err != nil {
+		seelog.Warnf("Unable to obtain healthcheck: %v", err)
```
fenxiong: How about something like "Health check failed with error: ..."

agent/app/args/flag.go
```diff
@@ -23,6 +23,7 @@ const (
 	licenseUsage             = "Print the LICENSE and NOTICE files and exit"
 	blacholeEC2MetadataUsage = "Blackhole the EC2 Metadata requests. Setting this option can cause the ECS Agent to fail to work properly.  We do not recommend setting this option"
 	windowsServiceUsage      = "Run the ECS agent as a Windows Service"
+	healthcheckServiceUsage  = "Run the agent healthcheck info"
```
fenxiong: maybe just "Run the agent healthcheck"..

agent/app/agent.go
cyastella: yeah, will change it

agent/app/healthcheck.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2014-2017 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
fenxiong: nit: -2019

suneyz: `Copyright 2019 Amazon.com` since it's a new file?

agent/app/healthcheck.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2014-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
yumex93: nit: the year can be just 2019

PR #2262, opened 2019-11-05
GMSA Integration

vsiddharth:
This patch introduces a new task level resource called credentialspec.
During task unmarshal, a conditional task level dependency is established.

The credentialspec option is injected into the containers' hostConfig
prior to creation using the runtime interface.

### Summary

This change enabled gMSA integration for ECS. 

### Implementation details

A new task level resource abstraction called credentialspec is introduced. 
Currently, the supported options for credentialspecs are:
* File on the local filesystem
* S3 file 
* SSM parameter

### TODO
* Functional Tests

### Testing

New tests cover the changes: 
yes

* Feature - Add gMSA integration for ECS tasks

### Licensing

### DIFF
--- a/agent/api/container/container_windows.go
+++ b/agent/api/container/container_windows.go
@@ -15,8 +15,75 @@
 
 package container
 
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/cihub/seelog"
+	dockercontainer "github.com/docker/docker/api/types/container"
+	"github.com/pkg/errors"
+)
+
 const (
 	// DockerContainerMinimumMemoryInBytes is the minimum amount of
 	// memory to be allocated to a docker container
 	DockerContainerMinimumMemoryInBytes = 256 * 1024 * 1024 // 256MB
 )
+
+// RequiresCredentialSpec checks if container needs a credentialspec resource
+func (c *Container) RequiresCredentialSpec() bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return false
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", err)
+		return false
+	}
+
+	if len(hostConfig.SecurityOpt) == 0 {
+		return false
+	}
+
+	for _, opt := range hostConfig.SecurityOpt {
+		if strings.HasPrefix(opt, "credentialspec") {
+			return true
+		}
+	}
+
+	return false
+}
+
+// GetCredentialSpec is used to retrieve the current credentialspec resource
+func (c *Container) GetCredentialSpec() (string, error) {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return "", errors.New("empty container hostConfig")
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", err)
+		return "", errors.New("unable to unmarshal container hostConfig")
+	}
+
+	if len(hostConfig.SecurityOpt) == 0 {
+		return "", errors.New("unable to find container security options")
+	}
+
+	for _, opt := range hostConfig.SecurityOpt {
+		if strings.HasPrefix(opt, "credentialspec") {
+			return opt, nil
+		}
+	}
+
+	return "", errors.New("unable to obtain credentialspec")
+}
--- a/agent/api/container/container_windows.go
+++ b/agent/api/container/container_windows.go
@@ -42,7 +42,7 @@ func (c *Container) RequiresCredentialSpec() bool {
 	hostConfig := &dockercontainer.HostConfig{}
 	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
 	if err != nil {
-		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", err)
+		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", c.Name, err)
 		return false
 	}
 
@@ -71,7 +71,7 @@ func (c *Container) GetCredentialSpec() (string, error) {
 	hostConfig := &dockercontainer.HostConfig{}
 	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
 	if err != nil {
-		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", err)
+		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", c.Name, err)
 		return "", errors.New("unable to unmarshal container hostConfig")
 	}
 
--- a/agent/api/container/container_windows.go
+++ b/agent/api/container/container_windows.go
@@ -19,7 +19,6 @@ import (
 	"encoding/json"
 	"strings"
 
-	"github.com/cihub/seelog"
 	dockercontainer "github.com/docker/docker/api/types/container"
 	"github.com/pkg/errors"
 )
@@ -32,35 +31,20 @@ const (
 
 // RequiresCredentialSpec checks if container needs a credentialspec resource
 func (c *Container) RequiresCredentialSpec() bool {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
-
-	if c.DockerConfig.HostConfig == nil {
+	credSpec, err := c.getCredentialSpec()
+	if err != nil || credSpec == "" {
 		return false
 	}
 
-	hostConfig := &dockercontainer.HostConfig{}
-	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
-	if err != nil {
-		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", c.Name, err)
-		return false
-	}
-
-	if len(hostConfig.SecurityOpt) == 0 {
-		return false
-	}
-
-	for _, opt := range hostConfig.SecurityOpt {
-		if strings.HasPrefix(opt, "credentialspec") {
-			return true
-		}
-	}
-
-	return false
+	return true
 }
 
 // GetCredentialSpec is used to retrieve the current credentialspec resource
 func (c *Container) GetCredentialSpec() (string, error) {
+	return c.getCredentialSpec()
+}
+
+func (c *Container) getCredentialSpec() (string, error) {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 
@@ -70,13 +54,8 @@ func (c *Container) GetCredentialSpec() (string, error) {
 
 	hostConfig := &dockercontainer.HostConfig{}
 	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
-	if err != nil {
-		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", c.Name, err)
-		return "", errors.New("unable to unmarshal container hostConfig")
-	}
-
-	if len(hostConfig.SecurityOpt) == 0 {
-		return "", errors.New("unable to find container security options")
+	if err != nil || len(hostConfig.SecurityOpt) == 0 {
+		return "", errors.New("unable to obtain security options from container hostConfig")
 	}
 
 	for _, opt := range hostConfig.SecurityOpt {
--- a/agent/api/container/container_windows_test.go
+++ b/agent/api/container/container_windows_test.go
@@ -1,0 +1,138 @@
+// +build windows,unit
+
+// Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package container
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRequiresCredentialSpec(t *testing.T) {
+	getContainer := func(hostConfig string) *Container {
+		c := &Container{
+			Name: "c",
+		}
+		c.DockerConfig.HostConfig = &hostConfig
+		return c
+	}
+
+	testCases := []struct {
+		name           string
+		container      *Container
+		expectedOutput bool
+	}{
+		{
+			name:           "hostconfig_nil",
+			container:      &Container{},
+			expectedOutput: false,
+		},
+		{
+			name:           "invalid_case",
+			container:      getContainer("invalid"),
+			expectedOutput: false,
+		},
+		{
+			name:           "empty_sec_opt",
+			container:      getContainer("{\"NetworkMode\":\"bridge\"}"),
+			expectedOutput: false,
+		},
+		{
+			name:           "missing_credentialspec",
+			container:      getContainer("{\"SecurityOpt\": [\"invalid-sec-opt\"]}"),
+			expectedOutput: false,
+		},
+		{
+			name:           "valid_credentialspec",
+			container:      getContainer("{\"SecurityOpt\": [\"credentialspec:file://gmsa_gmsa-acct.json\"]}"),
+			expectedOutput: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expectedOutput, tc.container.RequiresCredentialSpec())
+		})
+	}
+}
+
+func TestGetCredentialSpecErr(t *testing.T) {
+	getContainer := func(hostConfig string) *Container {
+		c := &Container{
+			Name: "c",
+		}
+		c.DockerConfig.HostConfig = &hostConfig
+		return c
+	}
+
+	testCases := []struct {
+		name                 string
+		container            *Container
+		expectedOutputString string
+		expectedErrorString  string
+	}{
+		{
+			name:                 "hostconfig_nil",
+			container:            &Container{},
+			expectedOutputString: "",
+			expectedErrorString:  "empty container hostConfig",
+		},
+		{
+			name:                 "invalid_case",
+			container:            getContainer("invalid"),
+			expectedOutputString: "",
+			expectedErrorString:  "unable to unmarshal container hostConfig",
+		},
+		{
+			name:                 "empty_sec_opt",
+			container:            getContainer("{\"NetworkMode\":\"bridge\"}"),
+			expectedOutputString: "",
+			expectedErrorString:  "unable to find container security options",
+		},
+		{
+			name:                 "missing_credentialspec",
+			container:            getContainer("{\"SecurityOpt\": [\"invalid-sec-opt\"]}"),
+			expectedOutputString: "",
+			expectedErrorString:  "unable to obtain credentialspec",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			expectedOutputStr, err := tc.container.GetCredentialSpec()
+			assert.Equal(t, tc.expectedOutputString, expectedOutputStr)
+			assert.EqualError(t, err, tc.expectedErrorString)
+		})
+	}
+}
+
+func TestGetCredentialSpecHappyPath(t *testing.T) {
+	getContainer := func(hostConfig string) *Container {
+		c := &Container{
+			Name: "c",
+		}
+		c.DockerConfig.HostConfig = &hostConfig
+		return c
+	}
+
+	c := getContainer("{\"SecurityOpt\": [\"credentialspec:file://gmsa_gmsa-acct.json\"]}")
+
+	expectedCredentialSpec := "credentialspec:file://gmsa_gmsa-acct.json"
+
+	credentialspec, err := c.GetCredentialSpec()
+	assert.NoError(t, err)
+	assert.EqualValues(t, expectedCredentialSpec, credentialspec)
+}
--- a/agent/api/container/container_windows_test.go
+++ b/agent/api/container/container_windows_test.go
@@ -22,14 +22,6 @@ import (
 )
 
 func TestRequiresCredentialSpec(t *testing.T) {
-	getContainer := func(hostConfig string) *Container {
-		c := &Container{
-			Name: "c",
-		}
-		c.DockerConfig.HostConfig = &hostConfig
-		return c
-	}
-
 	testCases := []struct {
 		name           string
 		container      *Container
@@ -80,14 +72,6 @@ func TestRequiresCredentialSpec(t *testing.T) {
 }
 
 func TestGetCredentialSpecErr(t *testing.T) {
-	getContainer := func(hostConfig string) *Container {
-		c := &Container{
-			Name: "c",
-		}
-		c.DockerConfig.HostConfig = &hostConfig
-		return c
-	}
-
 	testCases := []struct {
 		name                 string
 		container            *Container
@@ -130,14 +114,6 @@ func TestGetCredentialSpecErr(t *testing.T) {
 }
 
 func TestGetCredentialSpecHappyPath(t *testing.T) {
-	getContainer := func(hostConfig string) *Container {
-		c := &Container{
-			Name: "c",
-		}
-		c.DockerConfig.HostConfig = &hostConfig
-		return c
-	}
-
 	c := getContainer("{\"SecurityOpt\": [\"credentialspec:file://gmsa_gmsa-acct.json\"]}")
 
 	expectedCredentialSpec := "credentialspec:file://gmsa_gmsa-acct.json"
@@ -145,4 +121,12 @@ func TestGetCredentialSpecHappyPath(t *testing.T) {
 	credentialspec, err := c.GetCredentialSpec()
 	assert.NoError(t, err)
 	assert.EqualValues(t, expectedCredentialSpec, credentialspec)
+}
+
+func getContainer(hostConfig string) *Container {
+	c := &Container{
+		Name: "c",
+	}
+	c.DockerConfig.HostConfig = &hostConfig
+	return c
 }
--- a/agent/api/container/container_windows_test.go
+++ b/agent/api/container/container_windows_test.go
@@ -88,13 +88,13 @@ func TestGetCredentialSpecErr(t *testing.T) {
 			name:                 "invalid_case",
 			container:            getContainer("invalid"),
 			expectedOutputString: "",
-			expectedErrorString:  "unable to unmarshal container hostConfig",
+			expectedErrorString:  "unable to obtain security options from container hostConfig",
 		},
 		{
 			name:                 "empty_sec_opt",
 			container:            getContainer("{\"NetworkMode\":\"bridge\"}"),
 			expectedOutputString: "",
-			expectedErrorString:  "unable to find container security options",
+			expectedErrorString:  "unable to obtain security options from container hostConfig",
 		},
 		{
 			name:                 "missing_credentialspec",
--- a/agent/api/container/container_windows_test.go
+++ b/agent/api/container/container_windows_test.go
@@ -56,8 +56,18 @@ func TestRequiresCredentialSpec(t *testing.T) {
 			expectedOutput: false,
 		},
 		{
-			name:           "valid_credentialspec",
+			name:           "valid_credentialspec_file",
 			container:      getContainer("{\"SecurityOpt\": [\"credentialspec:file://gmsa_gmsa-acct.json\"]}"),
+			expectedOutput: true,
+		},
+		{
+			name:           "valid_credentialspec_s3",
+			container:      getContainer("{\"SecurityOpt\": [\"credentialspec:arn:aws:s3:::${BucketName}/${ObjectName}\"]}"),
+			expectedOutput: true,
+		},
+		{
+			name:           "valid_credentialspec_ssm",
+			container:      getContainer("{\"SecurityOpt\": [\"credentialspec:arn:aws:ssm:region:aws_account_id:parameter/parameter_name\"]}"),
 			expectedOutput: true,
 		},
 	}
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -19,7 +19,6 @@ import (
 	"path/filepath"
 	"time"
 
-	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
@@ -233,7 +232,7 @@ func (task *Task) initializeCredentialSpecResource(config *config.Config, creden
 }
 
 // getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
-func (task *Task) getAllCredentialSpecRequirements() map[string][]*apicontainer.Container {
+func (task *Task) getAllCredentialSpecRequirements() []string {
 	return nil
 }
 
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -19,8 +19,10 @@ import (
 	"path/filepath"
 	"time"
 
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/cgroup"
 	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
@@ -216,4 +218,26 @@ func (task *Task) dockerCPUShares(containerCPU uint) int64 {
 		return 2
 	}
 	return int64(containerCPU)
+}
+
+// requiresCredentialSpecResource returns true if at least one container in the task
+// needs a valid credentialspec resource
+func (task *Task) requiresCredentialSpecResource() bool {
+	return false
+}
+
+// initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
+func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
+	resourceFields *taskresource.ResourceFields) {
+	return
+}
+
+// getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
+func (task *Task) getAllCredentialSpecRequirements() map[string][]*apicontainer.Container {
+	return nil
+}
+
+// GetCredentialSpecResource retrieves credentialspec resource from resource map
+func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool) {
+	return []taskresource.TaskResource{}, false
 }
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -228,8 +228,8 @@ func (task *Task) requiresCredentialSpecResource() bool {
 
 // initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
 func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
-	resourceFields *taskresource.ResourceFields) {
-	return
+	resourceFields *taskresource.ResourceFields) error {
+	return errors.New("task credentialspec only supported on windows")
 }
 
 // getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -229,7 +229,7 @@ func (task *Task) requiresCredentialSpecResource() bool {
 // initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
 func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
 	resourceFields *taskresource.ResourceFields) error {
-	return errors.New("task credentialspec only supported on windows")
+	return errors.New("task credentialspec is only supported on windows")
 }
 
 // getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
--- a/agent/api/task/task_unsupported.go
+++ b/agent/api/task/task_unsupported.go
@@ -18,7 +18,9 @@ package task
 import (
 	"time"
 
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	"github.com/cihub/seelog"
 	dockercontainer "github.com/docker/docker/api/types/container"
@@ -64,4 +66,26 @@ func (task *Task) dockerCPUShares(containerCPU uint) int64 {
 		return 2
 	}
 	return int64(containerCPU)
+}
+
+// requiresCredentialSpecResource returns true if at least one container in the task
+// needs a valid credentialspec resource
+func (task *Task) requiresCredentialSpecResource() bool {
+	return false
+}
+
+// initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
+func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
+	resourceFields *taskresource.ResourceFields) error {
+	return nil
+}
+
+// getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
+func (task *Task) getAllCredentialSpecRequirements() map[string][]*apicontainer.Container {
+	return nil
+}
+
+// GetCredentialSpecResource retrieves credentialspec resource from resource map
+func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool) {
+	return []taskresource.TaskResource{}, false
 }
--- a/agent/api/task/task_unsupported.go
+++ b/agent/api/task/task_unsupported.go
@@ -18,7 +18,6 @@ package task
 import (
 	"time"
 
-	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
@@ -82,7 +81,7 @@ func (task *Task) initializeCredentialSpecResource(config *config.Config, creden
 }
 
 // getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
-func (task *Task) getAllCredentialSpecRequirements() map[string][]*apicontainer.Container {
+func (task *Task) getAllCredentialSpecRequirements() []string {
 	return nil
 }
 
--- a/agent/api/task/task_unsupported.go
+++ b/agent/api/task/task_unsupported.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	"github.com/cihub/seelog"
 	dockercontainer "github.com/docker/docker/api/types/container"
+	"github.com/pkg/errors"
 )
 
 const (
@@ -77,7 +78,7 @@ func (task *Task) requiresCredentialSpecResource() bool {
 // initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
 func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
 	resourceFields *taskresource.ResourceFields) error {
-	return nil
+	return errors.New("task credentialspec only supported on windows")
 }
 
 // getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
--- a/agent/api/task/task_unsupported.go
+++ b/agent/api/task/task_unsupported.go
@@ -78,7 +78,7 @@ func (task *Task) requiresCredentialSpecResource() bool {
 // initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
 func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
 	resourceFields *taskresource.ResourceFields) error {
-	return errors.New("task credentialspec only supported on windows")
+	return errors.New("task credentialspec is only supported on windows")
 }
 
 // getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
--- a/agent/api/task/task_windows.go
+++ b/agent/api/task/task_windows.go
@@ -23,8 +23,13 @@ import (
 	"strings"
 	"time"
 
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/credentialspec"
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
 	taskresourcevolume "github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
 	"github.com/cihub/seelog"
 	dockercontainer "github.com/docker/docker/api/types/container"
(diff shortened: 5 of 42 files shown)


agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+	cs.taskARN = temp.TaskARN
+	cs.executionCredentialsID = temp.ExecutionCredentialsID
+
+	return nil
+}
+
+func (cs *CredentialSpecResource) setCredentialSpecResourceLocation() error {
+	// TODO: Use registry
+	// This should always be available on Windows instances
+	appDataDir := os.Getenv("APPDATA")
```
sharanyad: could you make APPDATA and TEMP constants?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+		cs.requiredCredentialSpecs = temp.RequiredCredentialSpecs
+	}
+	cs.taskARN = temp.TaskARN
+	cs.executionCredentialsID = temp.ExecutionCredentialsID
+
+	return nil
+}
+
+func (cs *CredentialSpecResource) setCredentialSpecResourceLocation() error {
+	// TODO: Use registry
```
sharanyad: could you add more context to this TODO?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+}
+
+// clearCredentialSpec cycles through the collection of credentialspec data and
+// removes them from the task
+func (cs *CredentialSpecResource) clearCredentialSpec() {
+	cs.lock.Lock()
+	defer cs.lock.Unlock()
+
+	for key := range cs.credSpecMap {
+		// TODO: Cleanup file on container instance (optional)
```
sharanyad: why is this optional? I think we should clean it, if task is stopped.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+	// s3ClientCreator is a factory interface that creates new S3 clients. This is
+	// needed mostly for testing.
+	s3ClientCreator s3factory.S3ClientCreator
+
+	// credentialSpecResourceLocation is the location for all the tasks' credentialspec artifacts
+	credentialSpecResourceLocation string
+
+	// required for processing credentialspecs, key is input credentialspec
+	// Example key := credentialspec:file://credentialspec.json
+	requiredCredentialSpecs map[string][]*apicontainer.Container
```
sharanyad: why is this a map with the container object? do we use it anywhere? if yes, could it just be the container name as val?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+	taskDesiredStatus status.TaskStatus) {
+	cs.initStatusToTransition()
+	cs.credentialsManager = resourceFields.CredentialsManager
+	cs.ssmClientCreator = resourceFields.SSMClientCreator
+	cs.s3ClientCreator = resourceFields.S3ClientCreator
+
+	// if task hasn't turn to 'created' status, and it's desire status is 'running'
+	// the resource status needs to be reset to 'NONE' status so the cs value
+	// will be retrieved again
+	if taskKnownStatus < status.TaskCreated &&
```
sharanyad: If the resource has already moved to created, that means the s3 files would have been downloaded. in this case, they will be downloaded again. how are we planning to handle this? ideally, the files would have to be cleaned up in some way.

agent/api/container/container_windows.go
```diff
@@ -15,8 +15,75 @@
...
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return "", errors.New("empty container hostConfig")
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encountered error when trying to get hostConfig for container %s: %v", err)
```
somujay: Same as above. First arg is missing.

agent/api/container/container_windows_test.go
```diff
@@ -0,0 +1,138 @@
...
+		{
+			name:           "missing_credentialspec",
+			container:      getContainer("{\"SecurityOpt\": [\"invalid-sec-opt\"]}"),
+			expectedOutput: false,
+		},
+		{
+			name:           "valid_credentialspec",
+			container:      getContainer("{\"SecurityOpt\": [\"credentialspec:file://gmsa_gmsa-acct.json\"]}"),
+			expectedOutput: true,
+		},
```
somujay: You should add additional testcases: 
1) Multiple credentialspec : fail
2) Other formats: SSM & S3

agent/api/task/task_windows.go
```diff
@@ -150,3 +155,62 @@ func (task *Task) dockerCPUShares(containerCPU uint) int64 {
...
+
+// getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
+func (task *Task) getAllCredentialSpecRequirements() map[string][]*apicontainer.Container {
+	reqs := map[string][]*apicontainer.Container{}
+
+	for _, container := range task.Containers {
+		if container.RequiresCredentialSpec() {
+			credentialSpec, err := container.GetCredentialSpec()
+			if err == nil && credentialSpec != "" {
+				reqs[credentialSpec] = append(reqs[credentialSpec], container)
```
somujay: What happens if the same credentialspec is referenced in two different containers?

agent/config/parse_windows.go
```diff
@@ -0,0 +1,55 @@
...
+	"unsafe"
+
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+)
+
+// parseGMSACapability is used to determine if gMSA support can be enabled
+func parseGMSACapability() bool {
+	envStatus := utils.ParseBool(os.Getenv("ECS_GMSA_SUPPORTED"), true)
+	if envStatus {
+		// If gMSA feature is explicitly enabled, check if container instance is domain joined.
```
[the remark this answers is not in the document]
somujay: Still customer can override the environment variable and change the default behavior to be not supported. In those cases , we don't have the check the domain join.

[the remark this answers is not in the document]
somujay: > isn't this check being done irrespective of the flag being explicitly enabled or not(true by default)?

Still customer can override the environment variable and change the default behavior to be not supported. In those cases , we don't have the check the domain join.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+		// No need to log here. managedTask.applyResourceState already does that
+		err := errors.New("credentialspec resource: unable to find execution role credentials")
+		cs.setTerminalReason(err.Error())
+		return err
+	}
+	iamCredentials := executionCredentials.GetIAMRoleCredentials()
+
+	for credSpecStr, _ := range cs.requiredCredentialSpecs {
+		credSpecSplit := strings.SplitAfterN(credSpecStr, "credentialspec:", 2)
+		credSpecValue := credSpecSplit[1]
```
suneyz: Can you check the bound for this array before using a coded indix

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+
+func (cs *CredentialSpecResource) handleCredentialspecFile(credentialspec string) error {
+	credSpecSplit := strings.SplitAfterN(credentialspec, "credentialspec:", 2)
+	credSpecFile := credSpecSplit[1]
+
+	if !strings.HasPrefix(credSpecFile, "file://") {
+		return errors.New("invalid credentialspec file specification")
+	}
+
+	dockerHostconfigSecOptCredSpec := strings.Replace(credentialspec, "credentialspec:", "credentialspec=", 1)
```
suneyz: It seems like you are expecting the format of file to be an exact format. Do you want to use maybe regex to check the format before or after you execute this to avoid silent failure?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+}
+
+func (cs *CredentialSpecResource) setCredentialSpecResourceLocation() error {
+	// TODO: Use registry
+	// This should always be available on Windows instances
+	appDataDir := os.Getenv("APPDATA")
+	if appDataDir != "" {
+		cs.credentialSpecResourceLocation = appDataDir
+	} else {
+		tempDir := os.Getenv("TEMP")
```
suneyz: These seems to be a very generic environment name, it might be easily conflicting with some other user defined environment variables. Do you want to revisit this? 

Also, I'm not sure if it is better to define this [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/types.go#L31) also and pass it down here also, so that agent will load all of the required information when it starts up. @sharanyad might have more context here.

agent/taskresource/credentialspec/credentialspec_windows_test.go
```diff
@@ -0,0 +1,110 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package credentialspec
```
suneyz: I see more public accessible function in the tested file than those 3. Can you ensure they are properly covered by unit test?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,601 @@
...
+
+// clearCredentialSpec cycles through the collection of credentialspec data and
+// removes them from the task
+func (cs *CredentialSpecResource) clearCredentialSpec() {
+	cs.lock.Lock()
+	defer cs.lock.Unlock()
+
+	for key, value := range cs.CredSpecMap {
+		credSpecSplit := strings.SplitAfterN(value, "credentialspec=", 2)
+		localCredentialSpecFile := credSpecSplit[1]
```
somujay: Shouldn't delete the Credentialspec with file option

agent/config/parse_windows.go
```diff
@@ -0,0 +1,55 @@
...
+	"os"
+	"syscall"
+	"unsafe"
+
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+)
+
+// parseGMSACapability is used to determine if gMSA support can be enabled
+func parseGMSACapability() bool {
+	envStatus := utils.ParseBool(os.Getenv("ECS_GMSA_SUPPORTED"), true)
```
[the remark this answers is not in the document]
vsiddharth: yes

agent/ssm/ssm.go
```diff
@@ -43,6 +43,28 @@ func GetSecretsFromSSM(names []string, client SSMClient) (map[string]string, err
 	return extractSSMValues(out)
 }
 
+// TODO: Refactor GetSecretsFromSSM to parameterize `WithDecryption`
```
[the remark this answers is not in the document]
vsiddharth: This is being tracked by @yumex93

Makefile
```diff
@@ -313,7 +313,7 @@ GOFILES:=$(shell go list -f '{{$$p := .}}{{range $$f := .GoFiles}}{{$$p.Dir}}/{{
 .PHONY: gocyclo
 gocyclo:
 	# Run gocyclo over all .go files
-	gocyclo -over 15 ${GOFILES}
+	gocyclo -over 17 ${GOFILES}
```
yumex93: How did you rebase the change? This shouldn't show up as a part of your pr.

agent/api/container/container_windows_test.go
```diff
@@ -0,0 +1,148 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+			expectedOutputStr, err := tc.container.GetCredentialSpec()
+			assert.Equal(t, tc.expectedOutputString, expectedOutputStr)
+			assert.EqualError(t, err, tc.expectedErrorString)
+		})
+	}
+}
+
+func TestGetCredentialSpecHappyPath(t *testing.T) {
+	getContainer := func(hostConfig string) *Container {
```
yumex93: You are using getContainer for both of test functions, why not just make it as a helper function?

agent/api/task/task_linux.go
```diff
@@ -217,3 +219,25 @@ func (task *Task) dockerCPUShares(containerCPU uint) int64 {
...
+// requiresCredentialSpecResource returns true if at least one container in the task
+// needs a valid credentialspec resource
+func (task *Task) requiresCredentialSpecResource() bool {
+	return false
+}
+
+// initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
+func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
+	resourceFields *taskresource.ResourceFields) error {
+	return errors.New("task credentialspec only supported on windows")
```
yumex93: nit: task credentialspec is only supported on windows

agent/api/task/task_unsupported.go
```diff
@@ -65,3 +68,25 @@ func (task *Task) dockerCPUShares(containerCPU uint) int64 {
...
+// requiresCredentialSpecResource returns true if at least one container in the task
+// needs a valid credentialspec resource
+func (task *Task) requiresCredentialSpecResource() bool {
+	return false
+}
+
+// initializeCredentialSpecResource builds the resource dependency map for the credentialspec resource
+func (task *Task) initializeCredentialSpecResource(config *config.Config, credentialsManager credentials.Manager,
+	resourceFields *taskresource.ResourceFields) error {
+	return errors.New("task credentialspec only supported on windows")
```
yumex93: nit: task credentialspec is only supported on windows

agent/api/task/task_windows.go
```diff
@@ -150,3 +155,62 @@ func (task *Task) dockerCPUShares(containerCPU uint) int64 {
...
+
+	return nil
+}
+
+// getAllCredentialSpecRequirements is used to build all the credential spec requirements for the task
+func (task *Task) getAllCredentialSpecRequirements() map[string][]*apicontainer.Container {
+	reqs := map[string][]*apicontainer.Container{}
+
+	for _, container := range task.Containers {
+		if container.RequiresCredentialSpec() {
```
yumex93: You don't rly need to call **RequiresCredentialSpec()** here. It doesn't matter whether the container needs credential spec or not. If the container does not require condential spec, your **GetCredentialSpec()** will just fall to the case **credentialSpec == ""**.

agent/app/agent_capability.go
```diff
@@ -193,6 +195,10 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	// support external firelens config
 	capabilities = agent.appendFirelensConfigCapabilities(capabilities)
 
+	if agent.cfg.GMSACapable {
```
yumex93: This if condition can be just moved to appendGMSACapabilities and in linux, we don't rly need to check the **agent.cfg.GMSACapable**

agent/app/agent_capability_windows_test.go
```diff
@@ -234,3 +234,24 @@ func TestSupportedCapabilitiesWindows(t *testing.T) {
 		assert.Equal(t, aws.StringValue(expected.Value), aws.StringValue(capabilities[i].Value))
 	}
 }
+
+func TestAppendGMSACapabilities(t *testing.T) {
```
yumex93: We also need unit test in linux to make sure we don't advertise the capability for gMSA in linux.

agent/config/parse_windows.go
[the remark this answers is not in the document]
yumex93: Is there any use case that customers want to specifically disable it? Like if customer enables this feature, it will cause problems for their application?

From my understanding, even customers are not using this feature, we can still set this flag as true and advertise the capability. It will not affect anything.

Makefile
vsiddharth: Your commit was cherry-picked for the feature branch.

agent/engine/docker_task_engine.go
```diff
@@ -979,6 +980,50 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+		if !ok {
+			okErr := &apierrors.DockerClientConfigError{Msg: "unable to fetch task credentialspec resource"}
+			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(okErr)}
+		}
+
+		// Setup credentialSpec resource object for injection
+		if len(resource) <= 0 {
+			resMissingErr := &apierrors.DockerClientConfigError{Msg: "unable to fetch task resource credentialspec"}
+			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(resMissingErr)}
+		}
```
yumex93: You can just combine them together and do **!ok || len(resource) <=0**

agent/api/container/container_windows.go
```diff
@@ -15,8 +15,75 @@
...
+		return "", errors.New("unable to find container security options")
+	}
+
+	for _, opt := range hostConfig.SecurityOpt {
+		if strings.HasPrefix(opt, "credentialspec") {
+			return opt, nil
+		}
+	}
+
+	return "", errors.New("unable to obtain credentialspec")
```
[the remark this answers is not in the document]
vsiddharth: Given the time crunch, I did not write a lot of helpers. Duplication expedited delivery schedule. 
Also, these two methods have different usage patterns.

agent/app/agent_capability_windows_test.go
vsiddharth: Done

agent/app/agent_unix.go
```diff
@@ -211,6 +213,7 @@ func (agent *ecsAgent) initializeResourceFields(credentialsManager credentials.M
 			IOUtil:             ioutilwrapper.NewIOUtil(),
 			ASMClientCreator:   asmfactory.NewClientCreator(),
 			SSMClientCreator:   ssmfactory.NewSSMClientCreator(),
+			S3ClientCreator:    s3factory.NewS3ClientCreator(),
```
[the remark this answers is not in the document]
vsiddharth: I did not want to distinguish the two platforms but setting these creators to nil on one and not on the other. 
Also gMSA could be enabled for non-windows in the near future

agent/config/parse_windows.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"os"
+	"syscall"
+	"unsafe"
+
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+)
+
+// parseGMSACapability is used to determine if gMSA support can be enabled
+func parseGMSACapability() bool {
```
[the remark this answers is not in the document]
vsiddharth: Done

agent/config/parse_windows.go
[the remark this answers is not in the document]
vsiddharth: We already have a field there.

agent/config/parse_windows.go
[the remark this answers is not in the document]
vsiddharth: The capability should not be advertised when the environment variable is set but the container instance is not part of a valid active-directory.

agent/config/parse_windows.go
```diff
@@ -0,0 +1,55 @@
...
+func isDomainJoined() (bool, error) {
+	var domain *uint16
+	var status uint32
+
+	err := syscall.NetGetJoinInformation(nil, &domain, &status)
+	if err != nil {
+		return false, err
+	}
+
+	syscall.NetApiBufferFree((*byte)(unsafe.Pointer(domain)))
```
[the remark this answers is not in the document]
vsiddharth: Good question, but this method is an exact copy from the linked reference.

agent/engine/docker_task_engine.go
```diff
@@ -979,6 +980,41 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		}
 	}
 
+	// Populate credentialspec resource
+	if container.RequiresCredentialSpec() {
```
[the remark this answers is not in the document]
vsiddharth: done

agent/engine/docker_task_engine.go
```diff
@@ -979,6 +980,42 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+		seelog.Debugf("Obtained container %s with credentialspec resource requirement for task %s.", container.Name, task.Arn)
+		var credSpecResource *credentialspec.CredentialSpecResource
+		resource, ok := task.GetCredentialSpecResource()
+		if !ok {
+			okErr := &apierrors.DockerClientConfigError{Msg: "unable to fetch task credentialspec resource"}
+			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(okErr)}
+		}
+
+		// Setup credentialSpec resource object for injection
+		credSpecResource = resource[0].(*credentialspec.CredentialSpecResource)
```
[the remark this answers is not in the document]
vsiddharth: done.

agent/engine/docker_task_engine.go
```diff
@@ -979,6 +980,41 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+
+		containerCredSpec, err := container.GetCredentialSpec()
+		if err != nil && containerCredSpec != "" {
+			desiredCredSpecInjection, err := credSpecResource.GetTargetMapping(containerCredSpec)
+			if err != nil || desiredCredSpecInjection == "" {
+				missingErr := &apierrors.DockerClientConfigError{Msg: "unable to fetch valid credentialspec mapping"}
```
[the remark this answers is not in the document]
vsiddharth: Added comment to reduce ambiguity.

agent/engine/docker_task_engine.go
```diff
@@ -979,6 +980,41 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+				return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(missingErr)}
+			}
+
+			// Inject containers' hostConfig.SecurityOpt with the credentialspec resource
+			seelog.Infof("Injecting container %s with credentialspec %s.", container.Name, desiredCredSpecInjection)
+			if len(hostConfig.SecurityOpt) == 0 {
```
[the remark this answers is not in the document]
vsiddharth: hostConfig is an object created just a few lines above. By this point the nil check would have already been done.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,600 @@
...
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+// CredentialSpecResource is the abstraction for credentialspec resources
+type CredentialSpecResource struct {
+	taskARN string
+	region  string
+
+	credentialsManager     credentials.Manager
```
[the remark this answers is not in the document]
vsiddharth: go fmt has already been done on this file.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: We only have a pointer here and it is for safe keeping and future extensions.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: As per the design, it is alright to download the same file multiple times.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Done

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: The exact format is expected by design. We do not want to use regex for handling these artifacts.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Done.

agent/taskresource/credentialspec/credentialspec_windows_test.go
vsiddharth: Add more tests to cover the exported and critical methods. Boilerplate code is not covered by unit tests.

agent/taskresource/credentialspec/credentialspec_windows.go
suneyz: I'm not challenging how the format is designed. My question is how do you guarantee that it is valid before/after you perform the string replacement.

agent/api/container/container_windows.go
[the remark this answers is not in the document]
suneyz: I'm not following your response. L35-59 and L64-88 are pretty much the same logic besides return type. What's blocking you from refactor it to reduce code duplication?

[more information about code duplication](https://www.google.com/search?q=code+duplication&oq=code+duplication&aqs=chrome..69i57j0l5.8622j0j7&sourceid=chrome&ie=UTF-8)

agent/config/parse_windows.go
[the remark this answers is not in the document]
suneyz: if it is a direct copy. what's blocking you from importing it and using it directly?

agent/engine/docker_task_engine.go
```diff
@@ -979,6 +980,44 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+					if strings.HasPrefix(opt, "credentialspec:") {
+						hostConfig.SecurityOpt[idx] = desiredCredSpecInjection
+					}
+				}
+			}
+
+		} else {
+			emptyErr := &apierrors.DockerClientConfigError{Msg: "unable to fetch valid credentialspec"}
+			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(emptyErr)}
+		}
```
suneyz: If `err != nil`, can we surface err back as well?

agent/engine/docker_task_engine_windows_test.go
```diff
@@ -48,3 +54,85 @@ func TestDeleteTask(t *testing.T) {
 
 	taskEngine.deleteTask(task)
 }
+
+func TestCredentialSpecResourceTaskFile(t *testing.T) {
```
suneyz: Can you also add negative test case?

agent/ssm/ssm.go
```diff
@@ -43,6 +43,28 @@ func GetSecretsFromSSM(names []string, client SSMClient) (map[string]string, err
...
+// GetParametersFromSSM makes the api call to the AWS SSM parameter store to
+// retrieve parameter value in batches
+func GetParametersFromSSM(names []string, client SSMClient) (map[string]string, error) {
```
suneyz: Can we add a unit test for this?

agent/config/parse_windows.go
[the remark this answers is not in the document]
vsiddharth: It is not an exported method

agent/ssm/ssm.go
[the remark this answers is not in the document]
suneyz: I don't see a blocker using the following, yeah?

@yumex93 

```golang
func getParameter(names []string, client SSMClient, withDecryption bool) (map[string]string, error) {
	var params []*string
	for _, name := range names {
		params = append(params, aws.String(name))
	}

	in := &ssm.GetParametersInput{
		Names:          params,
		WithDecryption: aws.Bool(withDecryption),
	}

	out, err := client.GetParameters(in)
	if err != nil {
		return nil, err
	}

	return extractSSMValues(out)
}
```

agent/api/container/container_windows.go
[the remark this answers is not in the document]
yumex93: I don't think time crunch is the right reason to not reduce the duplicated code.  You can just write a helper function to return **string, error**, and for **RequiresCredentialSpec**, you can check the helper function result and return the true/false accordingly.

agent/taskresource/credentialspec/credentialspec_windows.go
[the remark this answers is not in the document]
suneyz: I think the root cause of this weird indentation  is that you might want to group them together. in this way go fmt will align `<name> <type>` for you

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,609 @@
...
+
+// GetTerminalReason returns an error string to propagate up through to task
+// state change messages
+func (cs *CredentialSpecResource) GetTerminalReason() string {
+	return cs.terminalReason
+}
+
+func (cs *CredentialSpecResource) setTerminalReason(reason string) {
+	cs.terminalReasonOnce.Do(func() {
+		seelog.Infof("credentialspec resource: setting terminal reason for credentialspec resource in task: [%s]", cs.taskARN)
```
suneyz: not sure if this will cause lots of unnecessary logs in agent log, maybe log as debug?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,609 @@
...
+
+// TerminalStatus returns the last transition state of credentialspec
+func (cs *CredentialSpecResource) TerminalStatus() resourcestatus.ResourceStatus {
+	return resourcestatus.ResourceStatus(CredentialSpecRemoved)
+}
+
+// NextKnownState returns the state that the resource should
+// progress to based on its `KnownState`.
+func (cs *CredentialSpecResource) NextKnownState() resourcestatus.ResourceStatus {
+	return cs.GetKnownStatus() + 1
```
suneyz: Can we not use magic number?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,609 @@
...
+
+		parsedARN, err := arn.Parse(credSpecValue)
+		if err != nil {
+			cs.setTerminalReason(err.Error())
+			return err
+		}
+
+		parsedARNService := parsedARN.Service
+		if parsedARNService == "s3" {
+			return cs.handleS3CredentialspecFile(credSpecStr, credSpecValue, iamCredentials)
```
suneyz: It looks like we are breaking out of the loop even if we have more requiredCredentialSpecs to process right?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,609 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (cs *CredentialSpecResource) handleCredentialspecFile(credentialspec string) error {
+	credSpecSplit := strings.SplitAfterN(credentialspec, "credentialspec:", 2)
+	credSpecFile := credSpecSplit[1]
```
suneyz: same as previous comment, please check length before you perform array slicing. Can you please check for other occurrences

agent/api/container/container_windows.go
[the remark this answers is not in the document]
vsiddharth: I have added the helper and refactored the code.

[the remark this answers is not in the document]
vsiddharth: I have added the helper and refactored the code.

vsiddharth:
@suneyz @yumex93 Could you PTAL?

agent/api/container/container_windows_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+		{
+			name:                 "invalid_case",
+			container:            getContainer("invalid"),
+			expectedOutputString: "",
+			expectedErrorString:  "unable to unmarshal container hostConfig",
+		},
+		{
+			name:                 "empty_sec_opt",
+			container:            getContainer("{\"NetworkMode\":\"bridge\"}"),
```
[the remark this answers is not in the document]
vsiddharth: This is just a test string

agent/ssm/ssm.go
vsiddharth: Done

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: This is part of the same boilerplate used in multiple places within the repo
Refer:
* https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/firelens/firelens_unix.go#L323
* https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/ssmsecret/ssmsecret.go#L162

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Currently, there is no strict validation. We obtain the source option and generate it to a known valid option. 
Example: 
```
key: "credentialspec:file://test.json"
value: "credentialspec=file://test.json"
``` 
Both the key and value are maintained in a map.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: done

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: We have moved away from using these.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: The initial design did not account for cleanup. However cleanup is now being handled.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: A different approach has been taken to resolve this issue. Please take a look at the recent changes.

agent/api/container/container_windows_test.go
vsiddharth: Done

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,617 @@
...
+		cs.requiredCredentialSpecs = temp.RequiredCredentialSpecs
+	}
+	cs.taskARN = temp.TaskARN
+	cs.executionCredentialsID = temp.ExecutionCredentialsID
+
+	return nil
+}
+
+func (cs *CredentialSpecResource) setCredentialSpecResourceLocation() error {
+	// TODO: Use registry to setup credentialspec resource location
```
yumex93: Do you have anywhere to track this todo? And what is the plan for you guys to switch to registry? Otherwise, this TODO in our code base may end up nowhere.

agent/taskresource/credentialspec/credentialspec_windows.go
yumex93: This is not a big concern for me for now. I will resolve it.

agent/taskresource/credentialspec/credentialspec_windows.go
yumex93: My question here is do we need to download it again? So for secret, we need to call ASM/SSM API to retrieve secrets again is bcuz they are all sensitive data and we don't want to preserve the data in our data file. However, based on your case, it is just a file name. So you should be able to directly save it to data file and restore it when agent restarts. It is not rly necessary to download it again.

agent/taskresource/credentialspec/credentialspec_windows.go
suneyz: My question is that if you don't validate `credentialspec:file://test.json ` format.  let's say the input is `credentialspecXYZ:file://test.json `. How would your code react to this case?

agent/taskresource/credentialspec/credentialspec_windows.go
suneyz: I see that you already addressed this, if early termination is not a case that you are expecting. Can you also add a unit test to make sure this logic won't be changed?

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: This validation would be performed during task registration.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Yes, we do. This is on the list of items that will be a fast follow up.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: I have already added unit tests to cover this change.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,617 @@
...
+	for key, value := range cs.CredSpecMap {
+		if strings.HasPrefix(key, "credentialspec:file://") {
+			seelog.Debugf("Skipping cleanup of local credentialspec file option: %s", key)
+			continue
+		}
+		credSpecSplit := strings.SplitAfterN(value, "credentialspec=", 2)
+		localCredentialSpecFile := credSpecSplit[1]
+		err := cs.os.Remove(localCredentialSpecFile)
+		if err != nil {
+			seelog.Debugf("Unable to clear local credential spec file %s for task %s", localCredentialSpecFile, cs.taskARN)
```
yumex93: Can you update this to **Warnf**, customers should still be able to see it under INFO log level, since potentially it may cause them some issues.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,617 @@
...
+			return &s
+		}(),
+		RequiredCredentialSpecs: cs.getRequiredCredentialSpecs(),
+		CredSpecMap:             cs.getCredSpecMap(),
+		ExecutionCredentialsID:  cs.getExecutionCredentialsID(),
+	})
+}
+
+// UnmarshalJSON deserialises the raw JSON to a CredentialSpecResourceJSON struct
+func (cs *CredentialSpecResource) UnmarshalJSON(b []byte) error {
```
yumex93: Why unmarshalJSON does not have CredSpecMap field?

agent/ssm/ssm.go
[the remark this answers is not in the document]
yumex93: Yeah. It works. @vsiddharth I would appreciate it a lot if you can refactor your function to use this helper function.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Done

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Good catch, thanks! Fixed it now.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Already updated. Please take a look at the latest changes.

agent/engine/docker_task_engine_windows_test.go
suneyz: Did you update the negative test case before resolving this?

agent/ssm/ssm.go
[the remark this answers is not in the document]
vsiddharth: Done.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Done, thanks!

agent/taskresource/credentialspec/credentialspec_windows.go
suneyz: To elaborate on my first comment here, if a case where a have 2 `credSpecStr ` (s3 and ssm or 2 s3) in `cs.requiredCredentialSpecs`, this (cs.handleS3CredentialspecFile/handleSSMCredentialspecFile) function will only be invoked once leading to early termination of the loop based on your current logic. Is this what expected?

If this is expected, please add a case where you have multiple `credSpecStr` and clarify with maybe a comment.

agent/config/parse_windows.go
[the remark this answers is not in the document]
yumex93: so Sid mentioned that the domainJoin is not enough to check whether it is joining the right domain. Like it can join the domain/AD which is not the right one for credentialSpec, if this is the case, is this check still adding value?

agent/taskresource/credentialspec/credentialspec_windows.go
sharanyad: I don't see a point in keeping this if we don't anticipate using this now. What future extensions are you thinking about?

agent/taskresource/credentialspec/credentialspec_windows.go
sharanyad: it is possible that someone removes the files in between.. we could either check if the required files are present and then try to download it again. 
What's concerning is multiple downloads.

sharanyad: >  it is alright to download the same file multiple times.

why do you think so?

agent/app/agent_unix.go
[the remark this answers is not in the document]
sharanyad: this seems to be currently only used for windows, so lets move this to https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/types_unix.go . `ResourceFieldsCommon` is meant for resources that's used in all the supported platforms.

agent/config/parse_windows.go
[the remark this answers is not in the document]
sharanyad: can you add comments in the code about what this is doing? someone reading the code here will not have any context.

agent/taskresource/credentialspec/credentialspec_windows.go
sharanyad: discussed offline: to avoid inconsistencies, we are not downloading resources that are already.

agent/taskresource/credentialspec/credentialspec_windows.go
yumex93: This is not a one way door solution. We can remove it and add it back if we have real use cases to use it.

agent/config/parse_windows.go
```diff
@@ -0,0 +1,55 @@
...
```
yumex93: This system call can also return error. Can you have error check here?

agent/config/parse_windows.go
```diff
@@ -0,0 +1,55 @@
...
+	envStatus := utils.ParseBool(os.Getenv("ECS_GMSA_SUPPORTED"), true)
+	if envStatus {
+		// If gMSA feature is explicitly enabled, check if container instance is domain joined.
+		// If container instance is not domain joined, explicitly disable feature configuration.
+		status, err := isDomainJoined()
+		if err == nil && status == true {
+			return true
+		}
+	}
+
```
yumex93: Can you log the error message as warning here? So if there are any unexpected issues that causing isDomainJoined return err, customers can use the information to debug

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Fixed.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: Changed as discussed.

agent/engine/docker_task_engine.go
vsiddharth: Done!

agent/app/agent_capability_windows.go
```diff
@@ -60,3 +60,11 @@ func (agent *ecsAgent) appendFirelensLoggingDriverCapabilities(capabilities []*e
 func (agent *ecsAgent) appendFirelensConfigCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
 	return capabilities
 }
+
+func (agent *ecsAgent) appendGMSACapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	if agent.cfg.GMSACapable {
+		return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityGMSA)
```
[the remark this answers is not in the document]
vsiddharth: Adding the space breaks `gocyclo`. 
This file has been formatted using `gofmt`

agent/config/parse_windows_test.go
```diff
@@ -0,0 +1,30 @@
...
+package config
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseGMSACapability(t *testing.T) {
```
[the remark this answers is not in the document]
vsiddharth: Adding that test requires mocking syscalls to windows. Are there any interfaces setup for that purpose?

agent/config/parse_windows.go
vsiddharth: Done

agent/config/parse_windows.go
vsiddharth: Done!

vsiddharth:
> Can you add a note in the description testing section that you are planning to have the functional/integ tests reviewed in a separate PR? And this PR only contains unit tests

A todo has been added.

agent/app/agent_capability_windows_test.go
```diff
@@ -234,3 +234,29 @@ func TestSupportedCapabilitiesWindows(t *testing.T) {
...
```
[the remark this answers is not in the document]
vsiddharth: Done.

agent/engine/docker_task_engine_windows_test.go
vsiddharth: A negative exists for the same.

agent/config/parse_windows.go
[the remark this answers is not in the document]
vsiddharth: @yumex93 I did not mention this. As discussed offline, the plan is to still have this check.

agent/config/parse_windows.go
```diff
@@ -0,0 +1,61 @@
...
+// parseGMSACapability is used to determine if gMSA support can be enabled
+func parseGMSACapability() bool {
+	envStatus := utils.ParseBool(os.Getenv("ECS_GMSA_SUPPORTED"), true)
+	if envStatus {
+		// If gMSA feature is explicitly enabled, check if container instance is domain joined.
+		// If container instance is not domain joined, explicitly disable feature configuration.
+		status, err := isDomainJoined()
+		if err == nil && status == true {
+			return true
+		} else {
```
yumex93: You don't need an else here, considering the happy path will directly return before logging.

agent/api/container/container_windows.go
```diff
@@ -15,8 +15,54 @@
...
+
+	return true
+}
+
+// GetCredentialSpec is used to retrieve the current credentialspec resource
+func (c *Container) GetCredentialSpec() (string, error) {
+	return c.getCredentialSpec()
+}
+
+func (c *Container) getCredentialSpec() (string, error) {
```
yumex93: Do you know whether we have validation on backend side to check customers only specify one credential spec per container?

vsiddharth: Yes, the plan is to only support one `credentialspec` per container and the backend should have this check.

agent/config/parse_windows_test.go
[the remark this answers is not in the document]
suneyz: You can mock your helper function instead of mocking sys calls.
in agent/config/parse_windows.go
```golang
var (
    isDomainJoinedFunction = isDomainJoined
}
```

in test file
```golang
isDomainJoinedFunction = func (bool, error) {
    // your mock logic goes here
}
```

agent/app/agent_unix.go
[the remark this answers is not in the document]
vsiddharth: Done.

agent/config/parse_windows_test.go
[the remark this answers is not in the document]
vsiddharth: I do not like this pattern at all. 
However, if you think this is strong requirement,  I will use the proposed idea.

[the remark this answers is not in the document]
suneyz: i don't have strong opinion on the test approach. but i do think you should test the true case. If your planned functional tests will cover the case. That's ok with me.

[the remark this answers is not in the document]
vsiddharth: The functional tests will address these cases.

agent/config/parse_windows.go
vsiddharth: Done, thanks!

agent/taskresource/types_windows_mac.go
```diff
@@ -19,11 +19,13 @@ import (
...
 )
 
 // ResourceFields is the list of fields required for creation of task resources
 type ResourceFields struct {
 	*ResourceFieldsCommon
-	Ctx          context.Context
-	DockerClient dockerapi.DockerClient
+	Ctx             context.Context
+	DockerClient    dockerapi.DockerClient
+	S3ClientCreator s3factory.S3ClientCreator
```
yumex93: This is for windows_mac. The right move should be have a types_windows.go and types_unsupported.go and only add S3ClientCreator in windows.go.

Makefile
yumex93: Can you remove this cherry pick commit? It will cause conflicts when we merge this pr back to dev. I already merge the gocyclo change commit to the feature branch, your tests should pass even without this commit.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -0,0 +1,612 @@
...
+	// needed mostly for testing.
+	ssmClientCreator ssmfactory.SSMClientCreator
+	// s3ClientCreator is a factory interface that creates new S3 clients. This is
+	// needed mostly for testing.
+	s3ClientCreator s3factory.S3ClientCreator
+	// credentialSpecResourceLocation is the location for all the tasks' credentialspec artifacts
+	credentialSpecResourceLocation string
+	// required for processing credentialspecs
+	// Example item := credentialspec:file://credentialspec.json
+	requiredCredentialSpecs []string
```
yumex93: I think we were talking about removing this attribute, what is the reason that we still keep here since we don't rly use it.

agent/config/parse_windows.go
```diff
@@ -0,0 +1,60 @@
...
+func isDomainJoined() (bool, error) {
+	var domain *uint16
+	var status uint32
+
+	err := syscall.NetGetJoinInformation(nil, &domain, &status)
+	if err != nil {
+		return false, err
+	}
+
+	err = syscall.NetApiBufferFree((*byte)(unsafe.Pointer(domain)))
```
yumex93: Can you add a comment to explain why we need this system call here? The reference does not have any information on this.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: This one is used extensively within the credentialspec task resource.

agent/config/parse_windows.go
vsiddharth: There is no clear publicly available documentation around this. 
We are just re-using the method from the golang library.

agent/taskresource/types_windows_mac.go
vsiddharth: Done!

Makefile
vsiddharth: Done!

PR #2278, opened 2019-11-14
Fix gMSA locks

vsiddharth:

### Summary

This patch decouples the credentialspec status and data locks.

### Implementation details

The datalock would protect access to the credential spec resources.
The other lock would be used for all other transitions.

### Testing

New tests cover the changes: 
No 

### Licensing

### DIFF
--- a/agent/taskresource/credentialspec/credentialspec_windows.go
+++ b/agent/taskresource/credentialspec/credentialspec_windows.go
@@ -77,7 +77,10 @@ type CredentialSpecResource struct {
 	// * key := credentialspec:s3ARN, value := credentialspec=file://CredentialSpecResourceLocation/s3_taskARN_fileName.json
 	// * key := credentialspec:ssmARN, value := credentialspec=file://CredentialSpecResourceLocation/ssm_taskARN_param.json
 	CredSpecMap map[string]string
-	// datalock should protect the CredSpecMap
+	// datalock should protect the credentialspec data
+	// updateCredSpecMapping() and GetDesiredStatus() could potentially deadlock
+	// if they depend on the same mutex. However, the credential spec data can be
+	// updated independently to prevent lock contention.
 	datalock sync.RWMutex
 
 	// lock is used for fields that are accessed and updated concurrently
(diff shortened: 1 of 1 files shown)


agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -98,6 +101,8 @@ func NewCredentialSpecResource(taskARN, region string,
 		ssmClientCreator:        ssmClientCreator,
 		s3ClientCreator:         s3ClientCreator,
 		CredSpecMap:             make(map[string]string),
+		os:                      oswrapper.NewOS(),
```
yumex93: Where are these used?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -77,6 +77,9 @@ type CredentialSpecResource struct {
 	// * key := credentialspec:s3ARN, value := credentialspec=file://CredentialSpecResourceLocation/s3_taskARN_fileName.json
 	// * key := credentialspec:ssmARN, value := credentialspec=file://CredentialSpecResourceLocation/ssm_taskARN_param.json
 	CredSpecMap map[string]string
+	// datalock should protect the CredSpecMap
+	datalock sync.RWMutex
```
sharanyad: could you also add comments on why this separate lock is needed?

vsiddharth: I added information to the commit message. 
Do you want it replicated here?

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: There are used in the handlers within the package.

vsiddharth: * https://github.com/aws/amazon-ecs-agent/blob/feature/gMSA/agent/taskresource/credentialspec/credentialspec_windows.go#L525
* https://github.com/aws/amazon-ecs-agent/blob/feature/gMSA/agent/taskresource/credentialspec/credentialspec_windows.go#L468

agent/taskresource/credentialspec/credentialspec_windows.go
sharanyad: that explains 'what' the locks are for. I'm looking for the answer for 'why'.

vsiddharth: Added more details.

fenxiong:
I think the deadlock is because you get the lock in `ApplyTransition` https://github.com/aws/amazon-ecs-agent/blob/dd6a824a05f9e9fbfd980b32a06f3ae59065b619/agent/taskresource/credentialspec/credentialspec_windows.go#L195-L196 which will then call `Create`, and `Create` will stuck at getting the same lock (might be because Firelens resource does that but Firelens doesn't get lock during `Create`...). I think if you remove the lock in `ApplyTransition` you shouldn't have this issue, and wouldn't need another lock.

vsiddharth:
Superseded by https://github.com/aws/amazon-ecs-agent/pull/2279

PR #2279, opened 2019-11-15
GMSA Improvements

vsiddharth:
### Summary

* Fix the locking issue with taskresource credentialspec
* Update statefile version
* Add test to verify state-file 
* Add integration test 

### Testing

New tests cover the changes: 
Yes

### Licensing


agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -298,15 +297,14 @@ func (cs *CredentialSpecResource) GetName() string {
...
 	executionCredentials, ok := cs.credentialsManager.GetTaskCredentials(cs.getExecutionCredentialsID())
-	if !ok {
-		// No need to log here. managedTask.applyResourceState already does that
-		err = errors.New("credentialspec resource: unable to find execution role credentials")
-		cs.setTerminalReason(err.Error())
-		return err
+	if ok {
+		iamCredentials = executionCredentials.GetIAMRoleCredentials()
+	} else {
+		seelog.Debugf("Unable to obtain execution credentials: %v", ok)
```
yumex93: This change will make it not return error when cannot find execution role, you should not change this.

vsiddharth: We do not need the credentials for using the file option. This should only be required for the other options. We also want to fail later in this situation.

agent/statemanager/state_manager_test.go
```diff
@@ -486,3 +487,43 @@ func TestLoadsDataSeqTaskManifest(t *testing.T) {
 	assert.EqualValues(t, 0, sequenceNumber)
 	assert.EqualValues(t, 7, seqNumTaskManifest)
 }
+
+func TestLoadsDataForGMSATask(t *testing.T) {
```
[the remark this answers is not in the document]
vsiddharth: That file is part of a different package. 
Additionally adding the test there would result in a circular dependency as statemanger -> engine -> statemanager.

agent/taskresource/credentialspec/credentialspec_windows.go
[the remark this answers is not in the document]
vsiddharth: We do not want to return an error here as the credentials are only required for S3/SSM and not the file option.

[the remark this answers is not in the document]
yumex93: Then I think you should skip this check when it is for file on host and keep this check before you require the S3/SSM. We want customer to get clear error message when task fails instead that they need to turn on debug log for agent in order to know what is happening.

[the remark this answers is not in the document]
suneyz: Fail fast is definitely preferred over a debug log which customers would be struggling to figure out what went wrong

agent/statemanager/state_manager_test.go
```diff
@@ -486,3 +487,43 @@ func TestLoadsDataSeqTaskManifest(t *testing.T) {
...
+	assert.Equal(t, "arn:aws:ecs:ap-northeast-1:1234567890:task/b8e2bd3c-c82a-4b43-9bde-199ae05b49a5", task.Arn)
+	assert.Equal(t, "gmsa-test", task.Family)
+	assert.Equal(t, 1, len(task.Containers))
+	container := task.Containers[0]
+	assert.Equal(t, "windows_sample_app", container.Name)
+	assert.NotNil(t, container.DockerConfig.HostConfig)
+
+	resource, ok := task.GetCredentialSpecResource()
+	assert.True(t, ok)
+	assert.NotEmpty(t, resource)
```
yumex93: Can you also check the attributes of the task resource? You need to depend on the state file value when agent restarts in order for container created correctly. So you need make sure all attributes you depend on have correct values there.

agent/taskresource/credentialspec/credentialspec_windows.go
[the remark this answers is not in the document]
vsiddharth: We cannot fail fast in this scenario. The credentials are required only for S3/SSM option. 
Also we do not want to obtain the credentials again and again for the same task across multiple containers. CP already enforces the credentials when S3/SSM option is used.

agent/statemanager/state_manager_test.go
vsiddharth: That would make the test depend on the platform as the CredSpecMap is only available on windows. The test would fail on other platforms.

agent/taskresource/credentialspec/credentialspec_windows.go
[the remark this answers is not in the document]
yumex93: I think you can remove the debug log here and check whether iamCredentials are nil or not in **handleS3CredentialspecFile/handleSSMCredentialspecFile** and return error inside it. It will not cause confusion for customers when they are using credentialSpec on the host, besides, it can also error out when it is execution role error.  When agent restarts, the execution credentials may not exist, we want to explicitly catch this situation and don't treat them same as other runtime error.

agent/statemanager/state_manager_test.go
yumex93: you have the check that this test will not be run on platforms other than windows.

agent/taskresource/credentialspec/credentialspec_windows.go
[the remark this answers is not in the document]
sharanyad: what happens for some reason role is not present for s3/ssm option. 
We cannot always assume this will be the case only for local file.

agent/engine/engine_windows_integ_test.go
```diff
@@ -611,3 +622,152 @@ func TestTaskLevelVolume(t *testing.T) {
...
+}
+
+func verifyContainerCredentialSpec(client *sdkClient.Client, id string) error {
+	dockerContainer, err := client.ContainerInspect(context.TODO(), id)
+	if err != nil {
+		return err
+	}
+
+	for _, opt := range dockerContainer.HostConfig.SecurityOpt {
+		if strings.HasPrefix(opt, "credentialspec=") {
```
yumex93: why we don't directly check the credentialspec value here?

agent/engine/engine_windows_integ_test.go
```diff
@@ -611,3 +622,152 @@ func TestTaskLevelVolume(t *testing.T) {
...
+
+	taskEngine := NewDockerTaskEngine(cfg, dockerClient, credentialsManager,
+		eventstream.NewEventStream("ENGINEINTEGTEST", context.Background()), imageManager, state, metadataManager, resourceFields)
+	taskEngine.MustInit(context.TODO())
+	return taskEngine, func() {
+		taskEngine.Shutdown()
+	}, credentialsManager
+}
+
+func setupWithGMSA(t *testing.T) (TaskEngine, func(), credentials.Manager) {
```
yumex93: We don't rly need this helper function? Can directly call setupGMSA().

agent/taskresource/credentialspec/credentialspec_windows.go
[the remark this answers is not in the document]
vsiddharth: CP has validation to ensure that for S3/SSM we will always have valid credentials. 
However, if we somehow end up with invalid creds, it should be a runtime failure. 

I have updated the PR with new changes.

agent/statemanager/state_manager_test.go
vsiddharth: It will still result in a failure on other platforms when you build a test binary for execution.

agent/engine/engine_windows_integ_test.go
vsiddharth: Updated helper.

agent/engine/engine_windows_integ_test.go
vsiddharth: Removed

agent/taskresource/credentialspec/credentialspec_windows.go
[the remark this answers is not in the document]
sharanyad: there are cases where agent could restart in the middle of resource creation. in that case, agent might end up not having the required credentials. So this check will be needed. I suggest you test this use case as well.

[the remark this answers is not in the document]
sharanyad: please follow @yumex93 's suggestion for fixing this.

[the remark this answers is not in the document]
sharanyad: nvm, just saw its fixed

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -298,15 +297,12 @@ func (cs *CredentialSpecResource) GetName() string {
 // Create is used to create all the credentialspec resources for a given task
 func (cs *CredentialSpecResource) Create() error {
 	var err error
-	// To fail fast, check execution role first
+	var iamCredentials credentials.IAMRoleCredentials
```
yumex93: Sorry. I should say it more clearly. Since here is not a pointer, you should check whether it is an empty struct or not. You don't need to declare the variable at the beginning. 
Just do 
```
executionCredentials, ok := cs.credentialsManager.GetTaskCredentials(cs.getExecutionCredentialsID())
```
and in the handleS3CredentialspecFile/handleSSMCredentialspecFile,  check whether it is an empty struct, something like
```
iamCredentials ==  credentials.IAMRoleCredentials{}
```

vsiddharth: Instead of copying the credentials, I'm passing the credentials around via reference.

vsiddharth: Thanks for explaining. Fixed.

agent/statemanager/state_manager_test.go
yumex93: This cannot be the reason that we don't check the value here. As a temporary workaround, considering windows tests fix needs some time, you can add some essential attributes/functions in credentialspec_unsupported.go file to make it work. Otherwise, you will need to refactor the windows statemanager test or wait for us to fix it.

vsiddharth: Changing the functional code to accommodate for test infrastructure issues is never a good idea.

vsiddharth: I have updated the test.

PR #2281, opened 2019-11-19
Bugfixes for gMSA Windows

sandeepindraganti:
### Summary
This pull request addresses couple of bug fixes for gMSA windows feature . The following bugs are fixed with this commit.
1. Issue with chmod on windows
2. Authorization exception when one process holds the handle for the file and other process tries to move or rename it
3. Filenaming issues on windows (Fixed issue of using taskid as part of fileName)
4. Fix issue to avoid creating credentialspec files with .json extension

### Implementation details
These code changes addresses the issues  we found during our internal testing and during bugbash. More details about the issues 
1) Chmod on windows: The Go library has some bugs on windows and due to this its not executing the right ACL that it has to execute on windows and fails.  I realized that we don't need it here , not sure why the previous developer had it but, i tested and verified it is not needed and so removed it.

2) Current current code uses fullArn format to name the file. Windows doesn't support that, to fix this issue we get the taskId and use it to name the file instead of using fullARn

3) Current implementation creates file with .json extension. Due to this if customer uploads file to S3 as foo.json then we will append one more .json to it which is not good. Since docker doesn't mandate the file with extension we don't need to have file with .json extension . So trimming it

### Testing
Intergration tests passed.
The agent is manually built and deployed to an windows instance and all  the test scenarios are manually run.


agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -436,8 +443,12 @@ func (cs *CredentialSpecResource) handleSSMCredentialspecFile(originalCredential
 
 	ssmParamData := ssmParamMap[ssmParam]
 
-	localCredSpecFilePath := fmt.Sprintf("%s/ssm_%s_%s.json", cs.credentialSpecResourceLocation, cs.taskARN, ssmParam)
-
+	taskArnSplit := strings.SplitN(cs.taskARN, "/", 2)
```
petderek: Neat! You will need to also handle the case for long task arns:

```
// your code covers this one
arn:aws:ecs:us-west-2:123456789012:task/12345-678901234-56789

// this is also possible if customers opt in
arn:aws:ecs:us-west-2:123456789012:task/MyCluster/1234567890123456789
```

At best, your code works but creates a directory called "MyCluster" which may be inconsistent with what you expect. It's also possible that it fails in the long arn case.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -18,6 +18,8 @@ package credentialspec
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/hectane/go-acl"
```
petderek: Talk with @yumex93 or @suneyz about vendoring this -- thats why the travis build is failing.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -455,26 +466,42 @@ func (cs *CredentialSpecResource) writeS3File(writeFunc func(file oswrapper.File
...
+	}
+
+	err = temp.Close()
 	if err != nil {
+		seelog.Errorf("Error while closing the handle to file:%v", temp.Name())
 		return err
 	}
 
-	err = cs.os.Rename(temp.Name(), filePath)
+	err = os.Remove(temp.Name())
```
petderek: Looking at this, it seems like the order in which the chmod / rename happens is the important factor here.

If you move the chmod after the old Rename function, wouldn't that accomplish the same thing? Or is there something else happening here?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -455,26 +466,42 @@ func (cs *CredentialSpecResource) writeS3File(writeFunc func(file oswrapper.File
...
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(filePath, input, 0644)
+	if err != nil {
+		return err
+	}
+
+	err = acl.Chmod(filePath, filePerm);
```
petderek: why won't the built in `os.Chmod()` work for our use case?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -18,6 +18,8 @@ package credentialspec
 import (
```
yumex93: Can you add description to the pr to describe why these code changes are needed? This can help our team members who don't have context to understand your pr.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -455,26 +466,42 @@ func (cs *CredentialSpecResource) writeS3File(writeFunc func(file oswrapper.File
...
-	err = temp.Chmod(os.FileMode(filePerm))
+	// Persist the file to disk.
+	err = temp.Sync()
 	if err != nil {
 		return err
 	}
 
-	// Persist the file to disk.
-	err = temp.Sync()
+	input, err := ioutil.ReadFile(temp.Name())
```
yumex93: Can you explain why this needs to be read/write again?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -436,8 +443,12 @@ func (cs *CredentialSpecResource) handleSSMCredentialspecFile(originalCredential
 
 	ssmParamData := ssmParamMap[ssmParam]
 
-	localCredSpecFilePath := fmt.Sprintf("%s/ssm_%s_%s.json", cs.credentialSpecResourceLocation, cs.taskARN, ssmParam)
-
+	taskArnSplit := strings.SplitN(cs.taskARN, "/", 2)
+	if (len(taskArnSplit) != 2) {
+		seelog.Errorf("Failed to retrieve taskId from taskArn.")
```
yumex93: Can you add seelog.Errorf() to the code places that calling **handleSSMCredentialspecFile/handleS3CredentialspecFile/handleCredentialspecFile**, which is in the create function, instead of just adding a single error log printing here.

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: How about using `filepath.Base` here?

vsiddharth: You could also try using the aws sdk's `arn.Parse`

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -455,26 +466,42 @@ func (cs *CredentialSpecResource) writeS3File(writeFunc func(file oswrapper.File
...
 	}
 
-	// Persist the file to disk.
-	err = temp.Sync()
+	input, err := ioutil.ReadFile(temp.Name())
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(filePath, input, 0644)
```
yumex93: Why the permission is 0644 here instead of **filePerm**?

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: Good catch! I addressed it.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: The problem is not with file permissions. It was failing because some other process is still accessing the memory .  I am guessing it might be S3 library . I have seen similar issues in the past . Moving chmod after rename will not fix the issue.

petderek: I see two changes here: (1) chmod now happens after the file operation (2) the file operation is changed from os.Rename() to explicitly reading the file into memory and writing it out.

To me, (1) makes the most since for what is solving the issue. I don't understand how memory  access, other processes, or S3's library come into play here, could you explain?

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: i removed this external library dependency .

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: os.chmod() has issues on windows . That's the reason i used an external library built for windows and replaced the existing os.chmod call . But i just realized that we don't need to change the permissions explicitly. I tested it and it works. Not sure why the previous developer had this call but i feel its not needed. I removed this.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: The memory is being accessed by other thread and you get handle to it . i have seen similar issues on windows , not sure if that was because of s3 client though.  Having this way is safe and will ensure that we will not have any exceptions in future.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: Done

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: Sure thing.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: Why do we need 755 permissions. 0644 should be enough and it has to be the same the other place as well. I changed the filePerm to 0644

agent/taskresource/credentialspec/credentialspec_windows.go
petderek: >  I have seen similar issues in the past 

Alternatively, is there a best practice for this situation if its common to windows? I don't have confidence that we understand the underlying issue here.

Recall that agent is the only thing that has touched the file at this point in time, so if something is mismanaging the file handle, its our problem.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -394,8 +398,14 @@ func (cs *CredentialSpecResource) handleS3CredentialspecFile(originalCredentials
 	}
 
 	resourceBase := filepath.Base(parsedARN.Resource)
-	localCredSpecFilePath := fmt.Sprintf("%s/s3_%s_%s.json", cs.credentialSpecResourceLocation, cs.taskARN, resourceBase)
+	taskArnSplit := strings.SplitN(cs.taskARN, "/", -1)
```
yumex93: You can just use Split(cs.taskARN, "/"), it works as the same way.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -455,26 +469,38 @@ func (cs *CredentialSpecResource) writeS3File(writeFunc func(file oswrapper.File
...
 
-	// Persist the file to disk.
-	err = temp.Sync()
+	//Read the file to a new chunk to avoid access violation issues.
+	input, err := ioutil.ReadFile(temp.Name())
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(filePath, input, filePerm)
```
yumex93: We already have ioutil passed into credentialspec object. Can you use cs.ioutil instead of ioutil?

agent/taskresource/credentialspec/credentialspec_windows.go
yumex93: same here, if you are going to use os, plz use the one passed into the credentialspec object

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -394,8 +398,14 @@ func (cs *CredentialSpecResource) handleS3CredentialspecFile(originalCredentials
...
 	resourceBase := filepath.Base(parsedARN.Resource)
-	localCredSpecFilePath := fmt.Sprintf("%s/s3_%s_%s.json", cs.credentialSpecResourceLocation, cs.taskARN, resourceBase)
+	taskArnSplit := strings.SplitN(cs.taskARN, "/", -1)
+	length := len(taskArnSplit)
+	if length < 2 {
+		seelog.Errorf("Failed to retrieve taskId from taskArn.")
+		return errors.New("Failed to retrieve taskId from taskArn.")
+	}
 
+	localCredSpecFilePath := fmt.Sprintf("%s\\s3_%v_%s", cs.credentialSpecResourceLocation, taskArnSplit[1], resourceBase)
```
somujay: taskArnSplit[1] ==> You should consider change this to size-1;

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -18,6 +18,7 @@ package credentialspec
 import (
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
```
somujay: Why do you need this library?

agent/taskresource/credentialspec/credentialspec_windows.go
somujay: The original access violation issue could be that the tempfile is not closed?

agent/taskresource/credentialspec/credentialspec_windows.go
somujay: It's same. Look at https://github.com/aws/amazon-ecs-agent/blob/feature/gMSA/agent/utils/ioutilwrapper/ioutil.go.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: I remove this write file . We don't need this anymore.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: The Wrapper that ECSAgent team uses doesn't have ReadFile() which is there in the original library. I was using this method earlier.  Since now i removed the ReadFile call i don't need this io util anymore. Removed

agent/taskresource/credentialspec/credentialspec_windows_test.go
```diff
@@ -520,11 +518,10 @@ func TestHandleS3CredentialspecFileWriteErr(t *testing.T) {
...
 		mockS3Client.EXPECT().DownloadWithContext(gomock.Any(), mockFile, gomock.Any()).Return(int64(0), nil),
 		mockFile.EXPECT().Write(gomock.Any()).AnyTimes(),
-		mockFile.EXPECT().Chmod(gomock.Any()).Return(nil),
-		mockFile.EXPECT().Sync().Return(nil),
-		mockFile.EXPECT().Name().Return(testTempFile),
-		mockOS.EXPECT().Rename(gomock.Any(), gomock.Any()).Return(errors.New("test-error")),
 		mockFile.EXPECT().Close(),
+		mockFile.EXPECT().Name().Return(tempFileName),
+		mockOS.EXPECT().Rename(gomock.Any(), gomock.Any()).Return(errors.New("test-error")).AnyTimes(),
+		mockFile.EXPECT().Name().Return(tempFileName),
```
somujay: Why do we need this?

agent/taskresource/credentialspec/credentialspec_windows.go
petderek: Idiomatic simplification -- using Split() is equivalent, but doesn't mean passing in a magic number.

agent/taskresource/credentialspec/credentialspec_windows_test.go
yumex93: why Rename has to be AnyTimes? It will only be called once.
Also, the **mockFile.EXPECT().Name().Return(tempFileName)** after rename is not needed. I think this will make this test fail. Have you tested?

agent/taskresource/credentialspec/credentialspec_windows_test.go
```diff
@@ -647,11 +644,9 @@ func TestCreateS3(t *testing.T) {
 		mockIO.EXPECT().TempFile(gomock.Any(), gomock.Any()).Return(mockFile, nil),
 		mockS3Client.EXPECT().DownloadWithContext(gomock.Any(), mockFile, gomock.Any()).Return(int64(0), nil),
 		mockFile.EXPECT().Write(gomock.Any()).AnyTimes(),
-		mockFile.EXPECT().Chmod(gomock.Any()).Return(nil),
-		mockFile.EXPECT().Sync().Return(nil),
-		mockFile.EXPECT().Name().Return(testTempFile),
-		mockOS.EXPECT().Rename(gomock.Any(), gomock.Any()).Return(nil),
 		mockFile.EXPECT().Close(),
+		mockFile.EXPECT().Name().AnyTimes(),
```
yumex93: nit: This should be mockFile.EXPECT().Name().Times(1). You only call Name once in your logic.

agent/taskresource/credentialspec/credentialspec_windows_test.go
```diff
@@ -520,11 +518,10 @@ func TestHandleS3CredentialspecFileWriteErr(t *testing.T) {
 		mockIO.EXPECT().TempFile(gomock.Any(), gomock.Any()).Return(mockFile, nil),
 		mockS3Client.EXPECT().DownloadWithContext(gomock.Any(), mockFile, gomock.Any()).Return(int64(0), nil),
 		mockFile.EXPECT().Write(gomock.Any()).AnyTimes(),
-		mockFile.EXPECT().Chmod(gomock.Any()).Return(nil),
-		mockFile.EXPECT().Sync().Return(nil),
-		mockFile.EXPECT().Name().Return(testTempFile),
-		mockOS.EXPECT().Rename(gomock.Any(), gomock.Any()).Return(errors.New("test-error")),
 		mockFile.EXPECT().Close(),
+		mockFile.EXPECT().Name().Return(tempFileName),
+		mockOS.EXPECT().Rename(gomock.Any(), gomock.Any()).Return(errors.New("test-error")).AnyTimes(),
```
yumex93: Nit: Since this is only called once, I think you don't really need AnyTimes().

agent/taskresource/credentialspec/credentialspec_windows_test.go
```diff
@@ -647,11 +644,9 @@ func TestCreateS3(t *testing.T) {
 		mockIO.EXPECT().TempFile(gomock.Any(), gomock.Any()).Return(mockFile, nil),
 		mockS3Client.EXPECT().DownloadWithContext(gomock.Any(), mockFile, gomock.Any()).Return(int64(0), nil),
 		mockFile.EXPECT().Write(gomock.Any()).AnyTimes(),
```
yumex93: Same as here. Anytimes() is not really necessary.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -315,6 +315,7 @@ func (cs *CredentialSpecResource) Create() error {
 		if strings.HasPrefix(credSpecValue, "file://") {
 			err = cs.handleCredentialspecFile(credSpecStr)
 			if err != nil {
+				seelog.Errorf("Failed to handle the credentialspec file,Error:%v", err)
```
vsiddharth: You are already logging at the Error level. 
Please update the message to `"Failed to handle the credentialspec file: %v", err`

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -331,12 +332,14 @@ func (cs *CredentialSpecResource) Create() error {
 		if parsedARNService == "s3" {
 			err = cs.handleS3CredentialspecFile(credSpecStr, credSpecValue, iamCredentials)
 			if err != nil {
+				seelog.Errorf("Failed to handle the credentialspec file from s3,Error:%v", err)
```
vsiddharth: You are already logging at the Error level.
Please update the message to `"Failed to handle the credentialspec file from s3: %v", err`

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -331,12 +332,14 @@ func (cs *CredentialSpecResource) Create() error {
...
 				cs.setTerminalReason(err.Error())
 				return err
 			}
 		} else if parsedARNService == "ssm" {
 			err = cs.handleSSMCredentialspecFile(credSpecStr, credSpecValue, iamCredentials)
 			if err != nil {
+				seelog.Errorf("Failed to handle the credentialspec file from SSM,Error:%v", err)
```
vsiddharth: You are already logging at the Error level.
Please update the message to `"Failed to handle the credentialspec file from SSM: %v", err`

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -394,8 +397,14 @@ func (cs *CredentialSpecResource) handleS3CredentialspecFile(originalCredentials
 	}
 
 	resourceBase := filepath.Base(parsedARN.Resource)
-	localCredSpecFilePath := fmt.Sprintf("%s/s3_%s_%s.json", cs.credentialSpecResourceLocation, cs.taskARN, resourceBase)
+	taskArnSplit := strings.Split(cs.taskARN, "/")
+	length := len(taskArnSplit)
+	if length < 2 {
+		seelog.Errorf("Failed to retrieve taskId from taskArn.")
```
vsiddharth: nit: Why are you logging the same message twice?

agent/taskresource/credentialspec/credentialspec_windows.go
vsiddharth: nit: Why are you logging the same message twice?

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -455,29 +468,23 @@ func (cs *CredentialSpecResource) writeS3File(writeFunc func(file oswrapper.File
...
-	err = temp.Chmod(os.FileMode(filePerm))
-	if err != nil {
-		return err
-	}
-
-	// Persist the file to disk.
-	err = temp.Sync()
+	err = temp.Close()
 	if err != nil {
+		seelog.Errorf("Error while closing the handle to file:%v", temp.Name())
```
vsiddharth: Update message to `"Error while closing the handle to file %s: %v", temp.Name(), err`

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -455,29 +468,23 @@ func (cs *CredentialSpecResource) writeS3File(writeFunc func(file oswrapper.File
...
 		return err
 	}
 
 	err = cs.os.Rename(temp.Name(), filePath)
 	if err != nil {
+		seelog.Errorf("Error while renaming the temporary file:%v", temp.Name())
```
vsiddharth: Please include the error in the message. 
Update this to `"Error while renaming the temporary file %s: %v", temp.Name(), err`

agent/taskresource/credentialspec/credentialspec_windows_test.go
sandeepindraganti: done

agent/taskresource/credentialspec/credentialspec_windows_test.go
sandeepindraganti: done

agent/taskresource/credentialspec/credentialspec_windows_test.go
sandeepindraganti: done

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: This comment is addressed.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: This comment is addressed.

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: This comment is addressed.  For some reason my earlier response was not shown here. Posting again

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: This comment is addressed.  For some reason my earlier response was not shown here. Posting again

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: This comment is addressed.  For some reason my earlier response was not shown here. Posting again

agent/taskresource/credentialspec/credentialspec_windows.go
sandeepindraganti: This comment is addressed.  For some reason my earlier response was not shown here. Posting again

PR #2283, opened 2019-11-19
Fix credentialspec cleanup

vsiddharth:

### Summary

Fix credentialspec cleanup

### Implementation details

Fix credentialspec cleanup by passing path to file. 

### Testing

New tests cover the changes: 
no

### Licensing

### DIFF
--- a/agent/taskresource/credentialspec/credentialspec_windows.go
+++ b/agent/taskresource/credentialspec/credentialspec_windows.go
@@ -529,9 +529,15 @@ func (cs *CredentialSpecResource) clearCredentialSpec() {
 			seelog.Debugf("Skipping cleanup of local credentialspec file option: %s", key)
 			continue
 		}
-		credSpecSplit := strings.SplitAfterN(value, "credentialspec=", 2)
+		// Split credentialspec to obtain local file-name
+		credSpecSplit := strings.SplitAfterN(value, "credentialspec=file://", 2)
+		if len(credSpecSplit) != 2 {
+			seelog.Warnf("Unable to parse target credentialspec: %s", value)
+			continue
+		}
 		localCredentialSpecFile := credSpecSplit[1]
-		err := cs.os.Remove(localCredentialSpecFile)
+		localCredentialSpecFilePath := filepath.Join(cs.credentialSpecResourceLocation, localCredentialSpecFile)
+		err := cs.os.Remove(localCredentialSpecFilePath)
 		if err != nil {
 			seelog.Warnf("Unable to clear local credential spec file %s for task %s", localCredentialSpecFile, cs.taskARN)
 		}
--- a/agent/taskresource/credentialspec/credentialspec_windows_test.go
+++ b/agent/taskresource/credentialspec/credentialspec_windows_test.go
@@ -68,10 +68,10 @@ func TestClearCredentialSpecDataHappyPath(t *testing.T) {
 
 	expectedS3FilePath := filepath.Join(credentialSpecResourceLocation, "s3_taskARN_fileName.json")
 	expectedSSMFilePath := filepath.Join(credentialSpecResourceLocation, "ssm_taskARN_fileName.json")
-	gomock.InOrder(
-		mockOS.EXPECT().Remove(expectedS3FilePath).Return(nil),
-		mockOS.EXPECT().Remove(expectedSSMFilePath).Return(nil),
-	)
+
+	// Mock returns for test
+	mockOS.EXPECT().Remove(expectedS3FilePath).Return(nil)
+	mockOS.EXPECT().Remove(expectedSSMFilePath).Return(nil)
 
 	err := credspecRes.Cleanup()
 	assert.NoError(t, err)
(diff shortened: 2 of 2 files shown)


agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -529,9 +529,15 @@ func (cs *CredentialSpecResource) clearCredentialSpec() {
 			seelog.Debugf("Skipping cleanup of local credentialspec file option: %s", key)
 			continue
 		}
-		credSpecSplit := strings.SplitAfterN(value, "credentialspec=", 2)
+		// Split credentialspec to obtain local file-name
+		credSpecSplit := strings.SplitAfterN(value, "credentialspec=file://", 2)
```
petderek: What will this do in the s3 or ssm cases?

vsiddharth: We do not want to cleanup the credentialspec's specified to use a local-file. 
However for S3/SSM options, we fetch the remote artifact and map it to a local file which is then passed onto the actual container. When the task is stopped and cleanup is triggered; this would cleanup the local artifact which was fetched from S3/SSM.

agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -529,9 +529,15 @@ func (cs *CredentialSpecResource) clearCredentialSpec() {
...
+		if len(credSpecSplit) != 2 {
+			seelog.Warnf("Unable to parse target credentialspec: %s", value)
+			continue
+		}
 		localCredentialSpecFile := credSpecSplit[1]
-		err := cs.os.Remove(localCredentialSpecFile)
+		localCredentialSpecFilePath := filepath.Join(cs.credentialSpecResourceLocation, localCredentialSpecFile)
```
yumex93: Can you add unit test to make sure we return the correct file path so we can remove the file?

vsiddharth: This code is already covered by unit tests.

yumex93: In your test, when you call **mockOS.EXPECT().Remove(gomock.Any()).Return(nil).AnyTimes()**, you put gomock.Any() in it, not the real file path. That's why at the beginning, we did not catch this bug. Can you correct your unit test to make it test the right thing?

vsiddharth: Done.

agent/taskresource/credentialspec/credentialspec_windows_test.go
```diff
@@ -80,18 +86,22 @@ func TestClearCredentialSpecDataErr(t *testing.T) {
...
+		os:                             mockOS,
+		credentialSpecResourceLocation: credentialSpecResourceLocation,
 	}
 
+	expectedS3FilePath := filepath.Join(credentialSpecResourceLocation, "s3_taskARN_fileName.json")
+	expectedSSMFilePath := filepath.Join(credentialSpecResourceLocation, "ssm_taskARN_fileName.json")
 	gomock.InOrder(
-		mockOS.EXPECT().Remove(gomock.Any()).Return(nil).Times(1),
-		mockOS.EXPECT().Remove(gomock.Any()).Return(errors.New("test error")),
+		mockOS.EXPECT().Remove(expectedS3FilePath).Return(nil),
```
yumex93: This relies on the map iteration order the same as the order you saved in the map. I don't think this is guaranteed to the truth based on how hashmap is implemented. This will introduce the flakiness to the test. This does not need to be in a specific order.

agent/taskresource/credentialspec/credentialspec_windows_test.go
```diff
@@ -54,17 +55,22 @@ func TestClearCredentialSpecDataHappyPath(t *testing.T) {
...
-		CredSpecMap: credSpecMapData,
-		os:          mockOS,
+		CredSpecMap:                    credSpecMapData,
+		os:                             mockOS,
+		credentialSpecResourceLocation: credentialSpecResourceLocation,
 	}
 
+	expectedS3FilePath := filepath.Join(credentialSpecResourceLocation, "s3_taskARN_fileName.json")
+	expectedSSMFilePath := filepath.Join(credentialSpecResourceLocation, "ssm_taskARN_fileName.json")
 	gomock.InOrder(
```
yumex93: Same comment here. This relies on the map iteration order the same as the order you saved in the map. I don't think this is guaranteed to the truth based on how hashmap is implemented. This will introduce the flakiness to the test. This does not need to be in a specific order.

agent/taskresource/credentialspec/credentialspec_windows_test.go
vsiddharth: Done, thanks!

agent/taskresource/credentialspec/credentialspec_windows_test.go
vsiddharth: Done, thanks!

PR #2296, opened 2019-11-26
Refactor the code to use goimports only and remove gofmt in makefile

cyastella:
### Summary

Since [goimports is a strict superset of gofmt.](https://godoc.org/golang.org/x/tools/cmd/goimports) So we use goimports only and remove the gofmt check to simplify the code.
### Testing

New tests cover the changes: 

### Licensing


yumex93:
It would be nice to double check that after removing, gofmt is properly run and everything gets formatted.

PR #2304, opened 2019-12-10
Fix gmsa config

vsiddharth:

### Summary

Fix the config that prevented the environment override from disabling the feature. 

### Testing

New tests cover the changes: 
Yes

Bug - Resolved disabling gMSA feature using environment overrides

### Licensing


yumex93:
How does your pr fix the bug? What is doing here is just changing the default config to **false**, however, the env var should take precedence over default config, so theoretically, if the env var is set, it would take the env var value. Did you test that based on your current change, if you set env var to **true** , does the value get read correctly?

vsiddharth:
> How does your pr fix the bug? What is doing here is just changing the default config to **false**, however, the env var should take precedence over default config, so theoretically, if the env var is set, it would take the env var value. Did you test that based on your current change, if you set env var to **true** , does the value get read correctly?

The changes were tested by both @somujay and myself. 

In Golang, the default for bool type is `false`. 

If the env variable is set to false, the feature will be disabled. 
If the env variable is set to true, additional checks will be performed before enabling the feature.

yumex93:
> > How does your pr fix the bug? What is doing here is just changing the default config to **false**, however, the env var should take precedence over default config, so theoretically, if the env var is set, it would take the env var value. Did you test that based on your current change, if you set env var to **true** , does the value get read correctly?
> 
> The changes were tested by both @somujay and myself.
> 
> In Golang, the default for bool type is `false`.
> 
> If the env variable is set to false, the feature will be disabled.
> If the env variable is set to true, additional checks will be performed before enabling the feature.

How the Golang default bool value affect the situation here?  Which code snippet gets affected? Can I get more details?

vsiddharth:
> > > How does your pr fix the bug? What is doing here is just changing the default config to **false**, however, the env var should take precedence over default config, so theoretically, if the env var is set, it would take the env var value. Did you test that based on your current change, if you set env var to **true** , does the value get read correctly?
> > 
> > 
> > The changes were tested by both @somujay and myself.
> > In Golang, the default for bool type is `false`.
> > If the env variable is set to false, the feature will be disabled.
> > If the env variable is set to true, additional checks will be performed before enabling the feature.
> 
> How the Golang default bool value affect the situation here? Which code snippet gets affected? Can I get more details? Thanks.

Please let me know what details you require. 

This issue surfaced when `environmentConfig` and `defaultConfig` are merged for the final config.

PR #2305, opened 2019-12-11
api/task: discover domain based on region

petderek:
### Summary

    AWS uses partitional endpoints for DNS domains. In standard partitions,
    it will be 'amazonaws.com'; but other partitions use their own special
    domains (eg, 'amazonaws.com.cn')
    
    EFS doesn't currently support other partitions, but this should
    future proof our client.

### Implementation details

The SDK has code to determine partitional domains: I'm reusing it here and wrapping it in a simple helper function.

### Testing

New tests cover the changes: Yes: I've added region-aware subtests

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -51,6 +51,7 @@ import (
 	resourcetype "github.com/aws/amazon-ecs-agent/agent/taskresource/types"
 	taskresourcevolume "github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/private/protocol/json/jsonutil"
 	"github.com/cihub/seelog"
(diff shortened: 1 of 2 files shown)


agent/api/task/task.go
```diff
@@ -2538,3 +2539,12 @@ func (task *Task) GetContainerIndex(containerName string) int {
 	}
 	return -1
 }
+
+func getDomainForPartition(region string) string {
+	partition, ok := endpoints.PartitionForRegion(endpoints.DefaultPartitions(), region)
+	if !ok {
+		seelog.Warnf("No partition resolved for region (%s). Using AWS default (%s)", region, endpoints.AwsPartition().DNSSuffix())
```
sharanyad: we should probably add context here that this is for EFS.

petderek: Was thinking this could be used for other things in future

petderek: Maybe not?

sharanyad: yeah, we can change it later if needed. Just that I would be surprised to see this message without much context. also, do we know when this error would occur?

PR #2345, opened 2020-01-23
Update sbin mount point to avoid conflict with newer Docker

fenxiong:
### Summary

Newer version of Docker mounts tini to /sbin/init which conflicts our existing /sbin:/sbin mount. Updating the mount to /sbin:/host/sbin and adds /host/sbin to the PATH env (needed by the appmesh plugin to find the iptable binary) fixes the problem.
This accompanies https://github.com/aws/amazon-ecs-init/pull/296.

### Implementation details

Update the mount from `/sbin:/sbin` to `/sbin:/host/sbin`. Add `/host/sbin` to the PATH env of the agent.

### Testing

The appmesh cni plugin needs this mount to be correct in order to find the `iptables` executable, so our existing appmesh functional test will be ensuring the changes don't break the plugin. I've manually ran the test on an instance with Docker 19.03.5 and verifies that the test fails without the changes, but passes with the changes.

Update sbin mount point to avoid conflict with newer version of Docker.

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ $ /usr/bin/docker run --name ecs-agent \
 --volume=/var/log/ecs/:/log:Z \
 --volume=/var/lib/ecs/data:/data:Z \
 --volume=/etc/ecs:/etc/ecs \
---volume=/sbin:/sbin \
+--volume=/sbin:/host/sbin \
 --volume=/lib:/lib \
 --volume=/lib64:/lib64 \
 --volume=/usr/lib:/usr/lib \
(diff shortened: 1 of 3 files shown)


petderek:
Thanks for doing this. Did you test this change against an older version of init?

fenxiong:
> Thanks for doing this. Did you test this change against an older version of init?

Yeah i launched the agent with existing version of init and ran the appmesh functional test and it passed. The old path `/sbin` is already included in the PATH env, just that it's after /host/sbin.

petderek:
I saw that but wanted to double check. Nice!!

PR #2358, opened 2020-02-13
ECS support JSON key and version for secrets

jy19:
### Summary

This pull request is for our roadmap item: https://github.com/aws/containers-roadmap/issues/385

This change allows customers to specify parameters such as `json-key`, `version-stage`, or `version-id` along with the AWS Secrets Manager ARN they are already inputting for a more specific secret value. 

### Implementation details

Agent will continue to take in an AWS Secrets Manager ARN, but will now parse the ARN into the parameters it expects. Some string manipulation and `arn.Parse` is involved. Existing secrets without these additional parameters has the same behavior. 

### Testing

Manually tested on an EC2 instance with local agent. New and existing unit tests and integration tests pass.

Adding enhancement to secretsmanager secret specified in ECS.

### Licensing

### DIFF
--- a/agent/asm/asm.go
+++ b/agent/asm/asm.go
@@ -15,10 +15,12 @@ package asm
 
 import (
 	"encoding/json"
+	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/secretsmanager"
 	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
+	"github.com/cihub/seelog"
 	"github.com/docker/docker/api/types"
 	"github.com/pkg/errors"
 )
@@ -85,6 +87,33 @@ func extractASMValue(out *secretsmanager.GetSecretValueOutput) (types.AuthConfig
 	}
 
 	return dac, nil
+}
+
+func GetSecretFromASMWithInput(input *secretsmanager.GetSecretValueInput,
+	client secretsmanageriface.SecretsManagerAPI, jsonKey string) (string, error) {
+	out, err := client.GetSecretValue(input)
+	if err != nil {
+		return "", errors.Wrapf(err, "secret %s", *input.SecretId)
+	}
+
+	if jsonKey == "" {
+		return aws.StringValue(out.SecretString), nil
+	}
+
+	secretMap := make(map[string]interface{})
+	jsonErr := json.Unmarshal([]byte(*out.SecretString), &secretMap)
+	if jsonErr != nil {
+		seelog.Warnf("Error when treating retrieved secret value with secret id %s as JSON and calling unmarshal.", *input.SecretId)
+		return "", jsonErr
+	}
+
+	secretValue, ok := secretMap[jsonKey]
+	if !ok {
+		err = errors.New(fmt.Sprintf("retrieved secret from Secrets Manager did not contain json key %s", jsonKey))
+		return "", err
+	}
+
+	return fmt.Sprintf("%v", secretValue), nil
 }
 
 // GetSecretFromASM makes the api call to the AWS Secrets Manager service to
--- a/agent/asm/asm_test.go
+++ b/agent/asm/asm_test.go
@@ -22,6 +22,17 @@ import (
 	"github.com/aws/aws-sdk-go/service/secretsmanager"
 	"github.com/aws/aws-sdk-go/service/secretsmanager/secretsmanageriface"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	versionID                = "versionId"
+	versionStage             = "versionStage"
+	jsonKey                  = "jsonKey"
+	valueFrom                = "arn:aws:secretsmanager:region:account-id:secret:secretId"
+	secretValue              = "secretValue"
+	jsonSecretValue          = "{\"" + jsonKey + "\": \"" + secretValue + "\",\"some-other-key\": \"secret2\"}"
+	malformedJsonSecretValue = "{\"" + jsonKey + "\": \"" + secretValue
 )
 
 type mockGetSecretValue struct {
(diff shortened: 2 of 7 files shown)


agent/taskresource/asmsecret/asmsecret.go
```diff
@@ -309,6 +319,67 @@ func (secret *ASMSecretResource) retrieveASMSecretValue(apiSecret apicontainer.S
...
+	input = &secretsmanager.GetSecretValueInput {}
+
+	paramValues := strings.Split(arnObj.Resource, arnDelimiter)  // arnObj.Resource looks like secret:secretID:...
+	if len(paramValues) == len(strings.Split(asmARNResourceFormat, arnDelimiter)) {
+		input.SecretId = &valueFrom
+		return input, "", nil
+	}
+	if len(paramValues) != len(strings.Split(asmARNResourceWithParametersFormat, arnDelimiter)) {
+		// can't tell what input this is, throw some error
+		err = errors.New("Unexpected ARN format with parameters when trying to retrieve ASM secret")
```
fenxiong: error msg should start with lower case

agent/asm/asm.go
```diff
@@ -87,6 +89,34 @@ func extractASMValue(out *secretsmanager.GetSecretValueOutput) (types.AuthConfig
 	return dac, nil
 }
 
+
+func GetSecretFromASMWithInput(input *secretsmanager.GetSecretValueInput,
+	client secretsmanageriface.SecretsManagerAPI, jsonKey string) (string, error) {
+	out, err := client.GetSecretValue(input)
+	if err != nil {
+		return "", errors.Wrapf(err, "secret %s", *input.SecretId)
```
fenxiong: based on this line in asmsecret.go:
```
input.SecretId = pointerOrNil(reconstructASMARN(arnObj))
```
seems like input.SecretId might be nil here (and will panic when you deference it)? i think the caller should return an error before calling GetSecretFromASMWithInput if input.SecretId is nil

agent/asm/asm.go
```diff
@@ -87,6 +89,34 @@ func extractASMValue(out *secretsmanager.GetSecretValueOutput) (types.AuthConfig
...
+	}
+
+	if jsonKey == "" {
+		return aws.StringValue(out.SecretString), nil
+	}
+
+	secretMap := make(map[string]interface{})
+	jsonErr := json.Unmarshal([]byte(*out.SecretString), &secretMap)
+	if jsonErr != nil {
+		seelog.Warnf("Error when treating retrieved secret value with secret id %s as JSON and calling unmarshal.", input.SecretId)
```
fenxiong: input.SecretId is a pointer here so `*input.SecretId`

agent/asm/asm_test.go
```diff
@@ -111,11 +121,74 @@ func TestASMGetAuthConfig(t *testing.T) {
...
+	secretValueInput := createSecretValueInput(toPtr(valueFrom), nil, nil)
+	outSecretValue, _ := GetSecretFromASMWithInput(secretValueInput, asmClient, jsonKey)
+	assert.Equal(t, secretValue, outSecretValue)
+}
+
+func TestGetSecretFromASMWithMalformedJSON(t *testing.T) {
+	asmClient := createASMInterface(malformedJsonSecretValue)
+	secretValueInput := createSecretValueInput(toPtr(valueFrom), nil, nil)
+	outSecretValue, err := GetSecretFromASMWithInput(secretValueInput, asmClient, jsonKey)
+	if assert.Error(t, err) {
```
fenxiong: i think you can just use require to avoid the need of this kind of check, i.e:
```
require.Error(...)
assert.Equal(...)
```
Same below

agent/taskresource/asmsecret/asmsecret.go
```diff
@@ -30,11 +30,17 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+	"github.com/aws/aws-sdk-go/aws/arn"
+	"github.com/aws/aws-sdk-go/service/secretsmanager"
+	"github.com/aws/aws-sdk-go/aws"
```
fenxiong: import not sorted. i think that will break travis?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.36.2
```
yumex93: Can you rebase your change to the newest dev branch? This does not belong to your changes.

agent/asm/asm.go
```diff
@@ -87,6 +89,34 @@ func extractASMValue(out *secretsmanager.GetSecretValueOutput) (types.AuthConfig
...
+		err = errors.New(fmt.Sprintf("Retrieved secret from Secrets Manager did not contain json key %s", jsonKey))
+		return "", err
+	}
+
+	return fmt.Sprintf("%v", secretValue), nil
+}
+
 // GetSecretFromASM makes the api call to the AWS Secrets Manager service to
 // retrieve the secret value
 func GetSecretFromASM(secretID string, client secretsmanageriface.SecretsManagerAPI) (string, error) {
```
yumex93: I think you can merge these two functions together considering you check whether jsonKey is empty or not anyway.

agent/asm/asm_test.go
```diff
@@ -111,11 +121,74 @@ func TestASMGetAuthConfig(t *testing.T) {
...
+		assert.Equal(t, secretValue, outSecretValue)
+	}
+}
+
+func toPtr(input string) *string {
+	if input == "" {
+		return nil
+	}
+	output := input
+	return &output
```
yumex93: You should be able to directly use &input.

agent/asm/asm_test.go
```diff
@@ -111,11 +121,74 @@ func TestASMGetAuthConfig(t *testing.T) {
...
+}
+
+
+func createSecretValueInput(secretID *string, versionID *string, versionStage *string) *secretsmanager.GetSecretValueInput {
+	return &secretsmanager.GetSecretValueInput{
+		SecretId: secretID,
```
yumex93: You don't really need toPtr before calling createSecretValueInput. You can just pass in string and use aws.String(XX) to construct the input.

agent/asm/asm_test.go
```diff
@@ -111,11 +121,74 @@ func TestASMGetAuthConfig(t *testing.T) {
...
-		},
-	}
+	asmClient := createASMInterface(secretValue)
 	_, err := GetSecretFromASM("secretName", asmClient)
 	assert.NoError(t, err)
 }
+
+func TestGetSecretFromASMWithJsonKey(t *testing.T) {
+	asmClient := createASMInterface(jsonSecretValue)
+	secretValueInput := createSecretValueInput(toPtr(valueFrom), nil, nil)
```
yumex93: You don't really need toPtr here. You can just update createSecretValueInput function to accept string and use aws.String(XX) to construct the input inside that function. Same below.

agent/taskresource/asmsecret/asmsecret.go
```diff
@@ -309,6 +319,67 @@ func (secret *ASMSecretResource) retrieveASMSecretValue(apiSecret apicontainer.S
...
+	}
+
+	return aws.String(in)
+}
+
+// Agent follows what Cloudformation does here with using Dynamic References to specify Template Values
+// in the format secret-id:json-key:version-stage:version-id
+// the input will always be a full ARN for ASM
+func getASMParametersFromInput(valueFrom string) (input *secretsmanager.GetSecretValueInput, jsonKey string, err error) {
+	// should we know if this is just plain ARN before we get here?
```
yumex93: Are you intentionally leave this here?

agent/asm/asm_test.go
jy19: i did it this way bc this was the examples in assert pkg documentation. is require better in this case? they seem to both accomplish the same things. i can change it to keep with our existing coding conventions.

agent/taskresource/asmsecret/asmsecret.go
jy19: yeah it did setting up goimports/gofmt now..

agent/asm/asm_test.go
jy19: could not bc they are const values being passed in.

agent/taskresource/asmsecret/asmsecret.go
jy19: nope will remove

agent/asm/asm.go
jy19: good point, checked for secret id before calling GetSecretFromASMWithInput

agent/taskresource/asmsecret/asmsecret.go
jy19: updated

agent/asm/asm.go
jy19: updated

CHANGELOG.md
jy19: yeah i rebased everything hopefully it's good now

agent/asm/asm_test.go
jy19: it was bc i wanted to pass in nil for versionID or versionStage.

agent/asm/asm_test.go
fenxiong: non blocking but would be good to keep with convention

jy19: changed

agent/asm/asm_test.go
yumex93: Isn't it a variable inside the function?
```
const (
   test = "test"
)

func main() {
	fmt.Println(toPtr(test))
}

func toPtr(input string) *string {
	if input == "" {
		return nil
	}
	return &input
}
```
I tried this in go playground, and it works for me.

jy19: oops sorry you are right - i forgot that the const were already passed in as a parameter. updated

fenxiong: Does the sdk client behave differently between passing the pointer of an empty string vs. passing a nil pointer? If they are different, then your original code is correct

agent/asm/asm.go
```diff
@@ -87,6 +89,33 @@ func extractASMValue(out *secretsmanager.GetSecretValueOutput) (types.AuthConfig
...
+	secretMap := make(map[string]interface{})
+	jsonErr := json.Unmarshal([]byte(*out.SecretString), &secretMap)
+	if jsonErr != nil {
+		seelog.Warnf("Error when treating retrieved secret value with secret id %s as JSON and calling unmarshal.", *input.SecretId)
+		return "", jsonErr
+	}
+
+	secretValue, ok := secretMap[jsonKey]
+	if !ok {
+		err = errors.New(fmt.Sprintf("Retrieved secret from Secrets Manager did not contain json key %s", jsonKey))
```
yumex93: The error should start with lower case for golang.

PR #2361, opened 2020-02-14
Address CPU usage metric corner cases

sparrc:
### Summary

This PR changes a few corner cases in the CPU percent metrics handling:

1. Cap cpu usage percent at 1024*1024 to match telemetry codebase
2. If two metrics arrive with identical timestamps, set cpu usage to previous value rather than leaving it at NaN
3. Ensure that we don't emit cloudwatch metrics with sampleCount == 0

### Testing

New unit tests were added to cover these changes

### Licensing

### DIFF
--- a/agent/stats/queue.go
+++ b/agent/stats/queue.go
@@ -26,10 +26,10 @@ import (
 
 const (
 	// BytesInMiB is the number of bytes in a MebiByte.
-	BytesInMiB = 1024 * 1024
+	BytesInMiB                     = 1024 * 1024
+	minimumQueueDatapoints         = 2
+	MaxCPUUsagePerc        float32 = 1024 * 1024
 )
-
-const minimumQueueDatapoints = 2
 
 // Queue abstracts a queue using UsageStats slice.
 type Queue struct {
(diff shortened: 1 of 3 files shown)


agent/stats/queue.go
```diff
@@ -343,6 +346,11 @@ func (queue *Queue) getCWStatsSet(f getUsageFloatFunc) (*ecstcs.CWStatsSet, erro
 		sum += thisStat
 	}
 
+	// don't emit metrics when sampleCount == 0
```
yumex93: We have the check before this to make sure the queue buffer is not less than 2, do we still need this check for sampleCount here?

sparrc: yes because we skip the metric if it's NaN (see the continue above), so if all the metrics are NaN then we can get down here without having set any values on the cloudwatch metric, and sampleCount will be 0.

agent/stats/queue.go
```diff
@@ -92,15 +92,15 @@ func (queue *Queue) add(rawStat *ContainerStats) {
 		// % utilization can be calculated only when queue is non-empty.
 		lastStat := queue.buffer[queueLength-1]
 		timeSinceLastStat := float32(rawStat.timestamp.Sub(lastStat.Timestamp).Nanoseconds())
-		if timeSinceLastStat > 0 {
+		if timeSinceLastStat <= 0 {
+			// if we got a duplicate timestamp, set cpu percentage to the same value as the previous stat
+			seelog.Errorf("Received a docker stat object with duplicate timestamp")
+			stat.CPUUsagePerc = lastStat.CPUUsagePerc
+		} else {
```
yumex93: Just wonder why we are not directly using raw cpuUsage and do this calculation instead?

sparrc: because raw cpu usage is measure in "cpu time", for which the units is nanoseconds and is just a counter of the amount of time the CPUs were utilized. Most people want this converted to "usage percent" because it's more usable, but you need two measure of cpu time to figure out the usage percent.

agent/stats/queue.go
yumex93: what will happen on the TACS side is we send  metrics with sampleCount is 0?

sparrc: they will error, disconnect and send an invalid frame message to us

PR #2366, opened 2020-02-17
Address CPU usage metric corner cases

sparrc:
(this is a recreation of https://github.com/aws/amazon-ecs-agent/pull/2361)

### Summary

This PR changes a few corner cases in the CPU percent metrics handling:

1. ~Cap cpu usage percent at 1024*1024 to match telemetry codebase~ allow large cpu usage percents to get sent to telemetry backend but log a detailed error message.
2. If two metrics arrive with identical timestamps, set cpu usage to previous value rather than leaving it at NaN
3. Ensure that we don't emit cloudwatch metrics with sampleCount == 0

### Testing

New unit tests were added to cover these changes

### Licensing

### DIFF
--- a/agent/stats/queue.go
+++ b/agent/stats/queue.go
@@ -327,7 +327,7 @@ func (queue *Queue) getCWStatsSet(f getUsageFloatFunc) (*ecstcs.CWStatsSet, erro
 	queueLength := len(queue.buffer)
 	if queueLength < 2 {
 		// Need at least 2 data points to calculate this.
-		return nil, fmt.Errorf("Need at least 2 data points in queue to calculate CW stats set")
+		return nil, fmt.Errorf("need at least 2 data points in queue to calculate CW stats set")
 	}
 
 	var min, max, sum float64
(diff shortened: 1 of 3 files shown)


agent/stats/queue.go
```diff
@@ -343,6 +349,11 @@ func (queue *Queue) getCWStatsSet(f getUsageFloatFunc) (*ecstcs.CWStatsSet, erro
 		sum += thisStat
 	}
 
+	// don't emit metrics when sampleCount == 0
+	if sampleCount == 0 {
+		return nil, fmt.Errorf("Need at least 1 non-NaN data points in queue to calculate CW stats set")
```
fenxiong: error msg should start with lowercase

agent/stats/queue.go
```diff
@@ -384,6 +395,11 @@ func (queue *Queue) getULongStatsSet(f getUsageIntFunc) (*ecstcs.ULongStatsSet,
 		sampleCount++
 	}
 
+	// don't emit metrics when sampleCount == 0
+	if sampleCount == 0 {
+		return nil, fmt.Errorf("Need at least 1 non-NaN data points in queue to calculate CW stats set")
```
fenxiong: same as above, loowercase

agent/stats/queue.go
sparrc: done

PR #2313, opened 2019-12-19
Implement task resource container dependency

suneyz:
### Summary

Implemented dependency logic so that EFS type volume task resource wait for pause container to be finished.

### TODO
1. Cleanup logic
2. Aggregate plugin required input
3. Functional tests

### Testing

New tests cover the changes:  yes

Verified by mock aws vpc mode on agent. Some key logs that verified dependency is working as expected.
```
2020-01-02T23:44:25Z [DEBUG] Task [xxxx]: adding network pause container dependency to resource [task-efs-shared]
2020-01-02T23:44:25Z [DEBUG] Managed task [xxxx]: can't apply state to resource [task-efs-shared] yet due to unresolved dependencies: dependency graph: resource's dependency on containers not resolved
2020-01-02T23:44:26Z [DEBUG] Managed task [xxxx]: container [~internal~ecs~pause] at desired status: RESOURCES_PROVISIONED
2020-01-02T23:44:26Z [INFO] Managed task [xxxx]: transitioned resource [task-efs-shared] to [CREATED]
2020-01-02T23:44:26Z [INFO] Managed task [xxxx]: got resource [task-efs-shared] event: [CREATED]
```

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2367, opened 2020-02-17
Implement task resource container dependency

fenxiong:
### Summary

Finishing https://github.com/aws/amazon-ecs-agent/pull/2313. Allow task resource to depend on a certain container status. Set proper dependency between EFS volume and pause container transitions so we can mount the volume after task network is set up, and unmount the volume before task network is removed.

### Implementation details

1. `agent/taskresource`: adds a method DependOnTaskNetwork to the task resource interface that indicates whether it depends on task network to be ready; adds method for adding container dependencies on resource. Update all resources to conform with interface requirement.
2. `engine`: updates resource transitioning logic to check container dependency before transitioning.
3. `engine/graph`: update dependency resolution check to account for task resource's container dependency.

### Testing

Unit test added. Manual test refer to https://github.com/aws/amazon-ecs-agent/pull/2372.

N/A

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -1239,13 +1239,6 @@ func (task *Task) addNamespaceSharingProvisioningDependency(cfg *config.Config) 
 		container.BuildContainerDependency(NamespacePauseContainerName, apicontainerstatus.ContainerRunning, apicontainerstatus.ContainerPulled)
 		namespacePauseContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
 	}
-
-	for _, resource := range task.GetResources() {
-		if resource.DependOnTaskNetwork() {
-			seelog.Debugf("Task [%s]: adding namespace pause container dependency to resource [%s]", task.Arn, resource.GetName())
-			resource.BuildContainerDependency(NamespacePauseContainerName, apicontainerstatus.ContainerRunning, resourcestatus.ResourceStatus(taskresourcevolume.VolumeCreated))
-		}
-	}
 }
 
 // ContainerByName returns the *Container for the given name
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -464,7 +464,7 @@ func (task *Task) initializeDockerLocalVolumes(dockerClient dockerapi.DockerClie
 		vol, _ := task.HostVolumeByName(volumeName)
 		// BUG(samuelkarp) On Windows, volumes with names that differ only by case will collide
 		scope := taskresourcevolume.TaskScope
-		localVolume, err := taskresourcevolume.NewVolumeResource(ctx, volumeName,
+		localVolume, err := taskresourcevolume.NewVolumeResource(ctx, volumeName, HostVolumeType,
 			vol.Source(), scope, false,
 			taskresourcevolume.DockerLocalVolumeDriver,
 			make(map[string]string), make(map[string]string), dockerClient)
@@ -550,6 +550,7 @@ func (task *Task) addEFSVolumes(
 	volumeResource, err := taskresourcevolume.NewVolumeResource(
 		ctx,
 		vol.Name,
+		EFSVolumeType,
 		task.volumeName(vol.Name),
 		"task",
 		false,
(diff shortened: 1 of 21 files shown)


agent/engine/task_manager.go
```diff
@@ -472,19 +474,31 @@ func (mtask *managedTask) handleResourceStateChange(resChange resourceStateChang
 		return
 	}
 
+	defer mtask.engine.saver.Save()
+	// Set known status regardless of error so the applied status can be cleared. If there is error,
```
yumex93: The applied status can be cleared is bcuz updateAppliedStatusUnsafe is called in the SetKnownStatus? If this is the case, I think not all task resource's SetKnownStatus function calls updateAppliedStatusUnsafe, like dockerVolume. Did we already update that?

fenxiong: good catch. i thought i checked every one of them. but apparently dockerVolume does not call it. double checking and that's the only missing one. will fix

agent/engine/task_manager.go
```diff
@@ -1045,16 +1059,79 @@ func (mtask *managedTask) containerNextState(container *apicontainer.Container)
...
+	if resKnownStatus > resDesiredStatus {
+		seelog.Debugf("Managed task [%s]: task resource [%s] has already transitioned beyond desired status(%s): %s",
+			mtask.Arn, resource.GetName(), resource.StatusString(resDesiredStatus), resource.StatusString(resKnownStatus))
+		return &resourceTransition{
+			nextState:      resourcestatus.ResourceStatusNone,
+			status:         resource.StatusString(resourcestatus.ResourceStatusNone),
+			actionRequired: false,
+			reason:         dependencygraph.ResourcePastDesiredStatusErr,
+		}
+	}
```
yumex93: The only difference for these two conditions is the log, I think we can just merge them together and change log to "already at or beyond the desired status(%s): %s"?

agent/engine/task_manager.go
fenxiong: Fixed

agent/engine/task_manager.go
fenxiong: updated

agent/api/task/task.go
```diff
@@ -1206,6 +1209,13 @@ func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) e
 		container.BuildContainerDependency(NetworkPauseContainerName, apicontainerstatus.ContainerResourcesProvisioned, apicontainerstatus.ContainerPulled)
 		pauseContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
 	}
+
```
yumex93: Do we need the dependency for pause container to make sure task resource is cleaned beforehand?

fenxiong: i removed it because i found that we can't remove efs volume before pause container stops, because docker doesn't allow removing a volume when a container using it hasn't been removed (even if container has stopped), and we don't remove container until all containers are stopped

agent/engine/docker_task_engine.go
```diff
@@ -531,6 +532,10 @@ func (engine *DockerTaskEngine) sweepTask(task *apitask.Task) {
 
 func (engine *DockerTaskEngine) deleteTask(task *apitask.Task) {
 	for _, resource := range task.GetResources() {
+		// Clean up the resource only if it's not already cleaned up, or in the process of being cleaned up.
+		if resource.GetKnownStatus() == resource.TerminalStatus() || resource.GetAppliedStatus() == resourcestatus.ResourceRemoved {
```
yumex93: Shouldn't appliedStatus also be TerminalStatus()? Although so far resource.TerminalStatus() is basically the same as ResourceRemoved.

fenxiong: I have removed all the modifications to resource cleanup due to the fact that efs resource doesn't need it (efs volume has to be cleaned up in deleteTask same as the original way, rather than when stopping task), and if future resource needs to be cleaned up before task network namespace is deleted, we can implement it at that time

agent/engine/task_manager.go
```diff
@@ -472,19 +474,31 @@ func (mtask *managedTask) handleResourceStateChange(resChange resourceStateChang
...
+
+	if status == res.SteadyState() { // Failed to create resource.
+		seelog.Errorf("Managed task [%s]: failed to create task resource [%s]: %v", mtask.Arn, res.GetName(), err)
+		res.SetKnownStatus(currentKnownStatus) // Set status back to None.
+
+		seelog.Infof("Managed task [%s]: marking task desired status to STOPPED", mtask.Arn)
 		mtask.SetDesiredStatus(apitaskstatus.TaskStopped)
 		mtask.Task.SetTerminalReason(res.GetTerminalReason())
-		mtask.engine.saver.Save()
+	} else { //  Failed to clean up resource.
```
sharanyad: discussed offline, this logic is not required, since resources are cleaned up only during sweep task

fenxiong: fixed

PR #2370, opened 2020-02-19
v4 task and container stats handler

cyastella:
### Summary

This add the initials outline for the task_stats_handler and container_stats_handler.
### Implementation details

This PR doesn't create a v4 metadata endpoint yet, it creates the initial task_stats_handler and container_stats_handler. In each file we specify the path that users used to query task and container stats for v4 endpoint. We reuse V3EndpointID as an unique container identifier to build the v4 path.
### Testing

New tests cover the changes: 

I will add tests in task_server_setup in the next PR once finish implementing task_server setup part for v4 endpoint.

### Licensing

### DIFF
--- a/agent/handlers/v4/container_stats_handler.go
+++ b/agent/handlers/v4/container_stats_handler.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
 //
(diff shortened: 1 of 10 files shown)


agent/handlers/v4/container_stats_handler.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
yumex93: We update our Copyright to a new format, can you change accordingly?

cyastella: sure, will change it

PR #2372, opened 2020-02-20
Integrate with ECS Volume Plugin for EFS volume

fenxiong:
Note: this PR is on top of https://github.com/aws/amazon-ecs-agent/pull/2367. First two commits belong to that PR.

### Summary

Integrate with ECS Volume Plugin for EFS volume. This adds support to:
(1) Specify transit encryption for an efs volume;
(2) Specify transit encryption port for an efs volume used by the stunnel created by efs mount helper when transit encryption enabled;
(3) Specify IAM authentication so access of the efs volume is controlled by task role;
(4) Specify access point for an efs volume.

### Implementation details

* `api/acs`: Update acs model for new fields in efs volume.

* `agent/taskresource`: Code changes for the volume task resource to use information from the task to invoke the ecs volume plugin to create the EFS volume, with transit encryption, transit encryption port, iam, network namespace supported.
  - Transit encryption, transit encryption port and iam are supported via passing options to mount helper constructed from task payload. Options are constructed when the volume task resource is initialized during PostUnmarshalTask.
  - To support mounting EFS in task network namespace for task in awsvpc network mode, the task network namespace handle is passed as another option to the mount helper. The pause container pid is saved in the volume task resource when pause container is transitioning to RESOURCE_PROVISION, and the pid is persisted to state file.
  - Refactor EFS related logic into a separate file for readability.

* `api/task`, `api/container`: Initialize the updated volume task resource with efs volume information from task payload. Extract and refactor some of the common logic in task_test.go to a separate file task_test_utils.go for readability.

* `agent/engine`: update to populate pause container pid to volume resource.

* `agent/app`, `agent/config`: Add volume plugin capabilties based on the value specifies in ECS_VOLUME_PLUGIN_CAPABILITIES env.

* `agent/statemanager`: update state file with new efs changes

### Testing

Unit tests added.

Manually ran task and verified the following:
* IAM auth: deny write access in task role -> not able to write to the efs volume from container; not denying ->  able to write;
* Transit encryption: when transit encryption is enabled in task def, efs mount helper starts stunnel to encrypt traffic. when port is specified, stunnel uses the specified port;
* Network namespace: when launching an efs task in awsvpc network mode, mount happens in task's network namespace. When limiting the targeting EFS filesystem to only allow access from the security group of the task, mount is still successful and functional

TBD

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -1239,13 +1239,6 @@ func (task *Task) addNamespaceSharingProvisioningDependency(cfg *config.Config) 
 		container.BuildContainerDependency(NamespacePauseContainerName, apicontainerstatus.ContainerRunning, apicontainerstatus.ContainerPulled)
 		namespacePauseContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
 	}
-
-	for _, resource := range task.GetResources() {
-		if resource.DependOnTaskNetwork() {
-			seelog.Debugf("Task [%s]: adding namespace pause container dependency to resource [%s]", task.Arn, resource.GetName())
-			resource.BuildContainerDependency(NamespacePauseContainerName, apicontainerstatus.ContainerRunning, resourcestatus.ResourceStatus(taskresourcevolume.VolumeCreated))
-		}
-	}
 }
 
 // ContainerByName returns the *Container for the given name
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -51,7 +51,6 @@ import (
 	resourcetype "github.com/aws/amazon-ecs-agent/agent/taskresource/types"
 	taskresourcevolume "github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
-	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/private/protocol/json/jsonutil"
 	"github.com/cihub/seelog"
 	"github.com/containernetworking/cni/libcni"
@@ -218,7 +217,8 @@ type Task struct {
 	// credentialsID is used to set the CredentialsId field for the
 	// IAMRoleCredentials object associated with the task. This id can be
 	// used to look up the credentials for task in the credentials manager
-	credentialsID string
+	credentialsID                string
+	credentialsRelativeURIUnsafe string
 
 	// ENIs is the list of Elastic Network Interfaces assigned to this task. The
 	// TaskENIs type is helpful when decoding state files which might have stored
@@ -339,6 +339,9 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 		task.initializeASMSecretResource(credentialsManager, resourceFields)
 	}
 
+	task.initializeCredentialsEndpoint(credentialsManager)
+	// NOTE: initializeVolumes needs to be after initializeCredentialsEndpoint, because EFS volume might
+	// need the credentials endpoint constructed by it.
 	if err := task.initializeVolumes(cfg, dockerClient, ctx); err != nil {
 		return err
 	}
@@ -348,7 +351,6 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 		return apierrors.NewResourceInitError(task.Arn, err)
 	}
 
-	task.initializeCredentialsEndpoint(credentialsManager)
 	task.initializeContainersV3MetadataEndpoint(utils.NewDynamicUUIDProvider())
 	if err := task.addNetworkResourceProvisioningDependency(cfg); err != nil {
 		seelog.Errorf("Task [%s]: could not provision network resource: %v", task.Arn, err)
@@ -542,11 +544,8 @@ func (task *Task) addEFSVolumes(
 	vol *TaskVolume,
 	efsvol *taskresourcevolume.EFSVolumeConfig,
 ) error {
-	// These are the NFS options recommended by EFS, see:
-	// https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-mount-cmd-general.html
-	domain := getDomainForPartition(cfg.AWSRegion)
-	ostr := fmt.Sprintf("addr=%s.efs.%s.%s,nfsvers=4.1,rsize=1048576,wsize=1048576,hard,timeo=600,retrans=2,noresvport", efsvol.FileSystemID, cfg.AWSRegion, domain)
-	devstr := fmt.Sprintf(":%s", efsvol.RootDirectory)
+	driverOpts := taskresourcevolume.GetDriverOptions(cfg, efsvol, task.GetCredentialsRelativeURI())
+	driverName := getEFSVolumeDriverName(cfg)
 	volumeResource, err := taskresourcevolume.NewVolumeResource(
 		ctx,
 		vol.Name,
@@ -554,12 +553,8 @@ func (task *Task) addEFSVolumes(
 		task.volumeName(vol.Name),
 		"task",
 		false,
-		"local",
-		map[string]string{
-			"type":   "nfs",
-			"device": devstr,
-			"o":      ostr,
-		},
+		driverName,
+		driverOpts,
 		map[string]string{},
 		dockerClient,
 	)
@@ -700,13 +695,14 @@ func (task *Task) initializeCredentialsEndpoint(credentialsManager credentials.M
 	for _, container := range task.Containers {
 		// container.Environment map would not be initialized if there are
 		// no environment variables to be set or overridden in the container
-		// config. Check if that's the case and initilialize if needed
+		// config. Check if that's the case and initialize if needed
 		if container.Environment == nil {
 			container.Environment = make(map[string]string)
 		}
 		container.Environment[awsSDKCredentialsRelativeURIPathEnvironmentVariableName] = credentialsEndpointRelativeURI
 	}
 
+	task.SetCredentialsRelativeURI(credentialsEndpointRelativeURI)
 }
 
 // initializeContainersV3MetadataEndpoint generates an v3 endpoint id for each container, constructs the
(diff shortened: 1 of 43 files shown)


agent/engine/task_manager.go
```diff
@@ -472,19 +474,32 @@ func (mtask *managedTask) handleResourceStateChange(resChange resourceStateChang
...
 		mtask.Task.SetTerminalReason(res.GetTerminalReason())
-		mtask.engine.saver.Save()
+	} else { //  Failed to clean up resource.
+		// When failing to clean up a resource, print an error log and assume it's removed, so its dependency can
+		// move forward (e.g. pause container depends on efs volume resource to be removed before moving to stopped).
+		// This doesn't seem to be an ideal way, but seems like we don't deal with failure in task resource cleanup
+		// ever since task resource is added (refer to deleteTask in docker_task_engine.go), so it is still the case here.
+		// Another note: we don't need to mark task desired status to stopped here because we will only try to
+		// clean up the resource when we try to stop the task (i.e. task desired status is already STOPPED).
+		seelog.Warnf("Managed task [%s]: failed to clean up task resource [%s]: %v", mtask.Arn, res.GetName(), err)
```
petderek: i dont think we need both log lines since they are communicating the same information. Just keep it at warn

agent/taskresource/volume/dockervolume.go
```diff
@@ -339,6 +335,24 @@ func (vol *VolumeResource) Create() error {
...
+	if vol.VolumeConfig.Driver != EFSVolumePluginDriverName {
+		return opts
+	}
+	// For awsvpc network mode, pause pid will be set (during the step when the pause container transitions to
+	// RESOURCE_PROVISIONED), and if the driver is the ecs volume plugin, we will pass the network namespace handle
+	// to the driver so that the mount will happen in the task network namespace.s
+	driverOpt, _ := opts["o"]
+	pausePID := vol.GetPauseContainerPID()
+	if pausePID != "" {
+		driverOpt += ",netns="
```
petderek: at this point, is it possible for driverOpt to be empty?

agent/taskresource/volume/dockervolume_efs.go
```diff
@@ -0,0 +1,104 @@
...
+	DockerVolumeName string `json:"dockerVolumeName"`
+}
+
+// EFSAuthConfig contains auth config for an efs volume.
+type EFSAuthConfig struct {
+	AccessPointId string `json:"accessPointId"`
+	Iam           string `json:"iam"`
+}
+
+// GetLocalDriverOptions returns the options for creating an EFS volume using Docker's local volume driver.
```
petderek: can we call it GetDockerLocalDriverOptions?

agent/taskresource/volume/dockervolume_efs.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+// GetEFSDriverOptions returns the options for creating an EFS volume using the ECS volume plugin.
+func (efsVolCfg *EFSVolumeConfig) GetVolumePluginDriverOptions(credsRelativeURI string) map[string]string {
+	device := efsVolCfg.FileSystemID
+	if efsVolCfg.RootDirectory != "" {
+		device = fmt.Sprintf("%s:%s", device, efsVolCfg.RootDirectory)
+	}
+	ostr := ""
```
petderek: this may or may not help... but we could model the options as:

[]string{
"netns=foo",
"tls",
"tlsport=1242"
}

and then when we pass them to the driver, we could to a strings.Join()

What do you think?

agent/engine/docker_task_engine.go
```diff
@@ -1192,7 +1192,19 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
...
+	if err != nil {
+		return dockerapi.DockerContainerMetadata{
+			Error: ContainerNetworkingError{
+				fromError: errors.Wrap(err,
+					"container resource provisioning: unable to inspect pause container"),
+			},
+		}
+	}
+
+	task.SetPausePIDInVolumeResource(strconv.Itoa(containerInspectOutput.State.Pid))
```
petderek: Does this send it to all volume resources, or just one? Maybe reword the method if it affects all of them.

fenxiong: i wrongly assumed there can only be one volume resource per task. renamed method to SetPausePIDInVolumeResources and updated to send it to all volume resources

agent/taskresource/volume/dockervolume_efs.go
fenxiong: yeah, updated

agent/taskresource/volume/dockervolume.go
fenxiong: good catch. fixed the extra leading comma in this case

agent/engine/task_manager.go
fenxiong: updated

agent/taskresource/volume/dockervolume_efs.go
fenxiong: yeah, i added another struct VolumeMountOptions to make these a bit easier..

agent/taskresource/volume/dockervolume_efs.go
```diff
@@ -0,0 +1,107 @@
...
+	"strconv"
+
+	"github.com/aws/amazon-ecs-agent/agent/config"
+
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/cihub/seelog"
+)
+
+const (
+	// EFSVolumePluginDriverName is the driver name of the ECS volume plugin.
```
petderek: could just call this ECSVolumePlugin -- we want it to be reusable for FSx

agent/taskresource/volume/dockervolume_efs.go
```diff
@@ -0,0 +1,107 @@
...
+	EFSVolumePluginDriverName = "amazon-ecs-volume-plugin"
+	// EFSLocalDriverName is the name of Docker's local volume driver.
+	EFSLocalDriverName = "local"
```
petderek: nit: Can we just call this "DockerLocalDriverName" since that's what it is?

agent/taskresource/volume/dockervolume_efs.go
```diff
@@ -0,0 +1,107 @@
...
+	// DockerVolumeName is internal docker name for this volume.
+	DockerVolumeName string `json:"dockerVolumeName"`
+}
+
+// EFSAuthConfig contains auth config for an efs volume.
+type EFSAuthConfig struct {
+	AccessPointId string `json:"accessPointId"`
+	Iam           string `json:"iam"`
+}
+
```
petderek: What do you think about this:

1. make both of the current methods private:
`getDockerLocalDriverOptions`
`getVolumePluginDriverOptions`

2. Make a different method:
`GetDriverOptions()` that returns the the local config for filesystemId+Rootdirectory -- and the VolumePlugin options for anything else.

agent/api/task/task.go
```diff
@@ -218,7 +217,8 @@ type Task struct {
 	// credentialsID is used to set the CredentialsId field for the
 	// IAMRoleCredentials object associated with the task. This id can be
 	// used to look up the credentials for task in the credentials manager
-	credentialsID string
+	credentialsID                string
+	credentialsRelativeURIUnsafe string
```
petderek: lets make sure this doesn't get serialized to disk in the config

agent/taskresource/volume/mountoptions_test.go
```diff
@@ -0,0 +1,42 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
petderek: this is cool, thanks for adding this parsing and testing.

agent/taskresource/volume/dockervolume_efs.go
petderek: (You totally don't have to do this. I was just thinking it could abstract the complexity a bit more)

agent/api/task/task.go
fenxiong: yeah, checked and it's not in state file

agent/taskresource/volume/dockervolume_efs.go
fenxiong: sgtm. updated

agent/taskresource/volume/dockervolume_efs.go
fenxiong: yeah, updated

agent/taskresource/volume/dockervolume_efs.go
fenxiong: updated

agent/taskresource/volume/dockervolume_efs.go
```diff
@@ -0,0 +1,128 @@
...
+	efsTransitEncryptionEnabled = "ENABLED"
+)
+
+// EFSVolumeConfig represents efs volume configuration.
+type EFSVolumeConfig struct {
+	AuthConfig            EFSAuthConfig `json:"authorizationConfig"`
+	FileSystemID          string        `json:"fileSystemId"`
+	RootDirectory         string        `json:"rootDirectory"`
+	TransitEncryption     string        `json:"transitEncryption"`
+	TransitEncryptionPort int64         `json:"transitEncryptionPort"`
```
sharanyad: do these need omitempty tags?

agent/taskresource/volume/dockervolume_efs.go
```diff
@@ -0,0 +1,128 @@
...
+	mntOpt := NewVolumeMountOptions()
+	if efsVolCfg.TransitEncryption == efsTransitEncryptionEnabled {
+		mntOpt.AddOption("tls", "")
+	}
+	if efsVolCfg.TransitEncryptionPort != 0 {
+		mntOpt.AddOption("tlsport", strconv.Itoa(int(efsVolCfg.TransitEncryptionPort)))
+	}
+	if efsVolCfg.AuthConfig.Iam == efsIAMAuthEnabled {
+		mntOpt.AddOption("iam", "")
+		mntOpt.AddOption("awscredentialsrelativeuri", credsRelativeURI)
```
sharanyad: isn't `awscredentialsrelativeuri` changed to a different name now?

agent/engine/docker_task_engine.go
```diff
@@ -1187,7 +1187,19 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
 func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task, container *apicontainer.Container) dockerapi.DockerContainerMetadata {
 	seelog.Infof("Task engine [%s]: setting up container resources for container [%s]",
 		task.Arn, container.Name)
-	cniConfig, err := engine.buildCNIConfigFromTaskContainer(task, container, true)
+	containerInspectOutput, err := engine.inspectContainerByName(task.Arn, container.Name)
+	if err != nil {
+		return dockerapi.DockerContainerMetadata{
+			Error: ContainerNetworkingError{
+				fromError: errors.Wrap(err,
+					"container resource provisioning: unable to inspect pause container"),
```
sharanyad: is this the error sent to backend? if yes, this should be more meaningful..

agent/taskresource/volume/dockervolume_efs.go
fenxiong: yeah, updated to awscredsuri

agent/engine/docker_task_engine.go
fenxiong: yeah i think so. updated to `container resource provisioning: cannot setup task network namespace due to error inspecting pause container`

agent/taskresource/volume/dockervolume_efs.go
fenxiong: added

PR #2375, opened 2020-02-25
add environment file to api

jy19:
### Summary

adds environmentFile to api container stuff

### Testing

make test
New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/acs/model/ecsacs/api.go
+++ b/agent/acs/model/ecsacs/api.go
@@ -288,6 +288,8 @@ type Container struct {
 
 	Environment map[string]*string `locationName:"environment" type:"map"`
 
+	EnvironmentFiles []*EnvironmentFile `locationName:"environmentFiles" type:"list"`
+
 	Essential *bool `locationName:"essential" type:"boolean"`
 
 	FirelensConfiguration *FirelensConfiguration `locationName:"firelensConfiguration" type:"structure"`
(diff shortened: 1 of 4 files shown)


agent/api/task/task_test.go
```diff
@@ -1135,6 +1135,12 @@ func TestTaskFromACS(t *testing.T) {
 						Region:    strptr("us-west-2"),
 					},
 				},
+				EnvironmentFiles: []*ecsacs.EnvironmentFile{
+					{
+						Value: strptr("s3://bucketName/envFile"),
+						Type:  strptr("s3"),
```
yunhee-l: do we need to add test for when this is not s3?

agent/acs/model/ecsacs/api.go
```diff
@@ -288,6 +288,8 @@ type Container struct {
 
 	Environment map[string]*string `locationName:"environment" type:"map"`
 
+	EnvironmentFiles []*EnvironmentFile `locationName:"environmentFiles" type:"list"`
```
fenxiong: this file is autogenerated. you should edit `model/api/api-2.json` and then go generate this file. otherwise the next people generating api.go will remove the changes you added here

jy19: lol i was thinking that was probably the case thanks

agent/api/task/task_test.go
jy19: customers wouldn't be able to register a task def for values other than s3 currently based on backend logic.. do you think that's still needed as a test on our side?

PR #2380, opened 2020-02-27
Fix flakey TaskManifest* tests

shubham2892:
### Summary
TaskManifest* tests were failing intermittently. Failing tests expected that a call was made to [`MakeRequest()` function](https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/task_manifest_handler_test.go#L87). The call was being made but the respective tests finished execution before the call was made and hence the tests failed. 
Adding a sleep time after triggering the workflow which calls `MakeRequest` function and before the function finishes the execution ensures that `MakeRequest` is always called before the function finishes execution. 

### Testing
Tested by running the unit tests multiple times. They were failing intermittently before but passed every time after adding the sleep.

### Licensing

### DIFF
--- a/agent/acs/handler/task_manifest_handler_test.go
+++ b/agent/acs/handler/task_manifest_handler_test.go
@@ -17,6 +17,7 @@ package handler
 import (
 	"context"
 	"testing"
+	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	"github.com/aws/amazon-ecs-agent/agent/api/task"
@@ -108,6 +109,11 @@ func TestManifestHandlerKillAllTasks(t *testing.T) {
 
 	newTaskManifest.messageBufferTaskManifest <- message
 
+	// mockWSClient.EXPECT().MakeRequest(ackRequested).Times(1) in this test is called by an asynchronous routine.
+	// Sometimes functions execution finishes before a call to this asynchronous routine, this sleep will ensure that
+	// asynchronous routine is called before function ends
+	time.Sleep(2 * time.Second)
+
 	select {
 	case <-newTaskManifest.ctx.Done():
 	}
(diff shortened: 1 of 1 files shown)


sparrc:
From what I can tell we are sending a message down a channel and then sleeping so that the function `handleTaskManifestSingleMessage` gets called.

Instead of sending down the channel and sleeping, could we just call `handleTaskManifestSingleMessage` directly in the unit tests? This would make the test deterministic. We lose a small amount of test coverage but I think the trade-off is worth it.

shubham2892:
> From what I can tell we are sending a message down a channel and then sleeping so that the function `handleTaskManifestSingleMessage` gets called.
> 
> Instead of sending down the channel and sleeping, could we just call `handleTaskManifestSingleMessage` directly in the unit tests? This would make the test deterministic. We lose a small amount of test coverage but I think the trade-off is worth it.

Part of the test is to also ensure that `handleTaskManifestSingleMessage` is being called when the message is being sent, that's why we have to check to make sure that `MakeRequest` is being called. If we go the route of calling `handleTaskManifestSingleMessage` directly, `.Expect.MakeRequest()` part will become redundant, thoughts?

sparrc:
Actually I think MakeRequest will still be called because handleTaskManifestSingleMessage sends off an async ack goroutine which makes the call to MakeRequest, correct?

Maybe we can remove the MakeRequest expectation from most of these unit tests, and add another unit test only verifying that MakeRequest gets called. This test could have a longer sleep as the other tests would no longer need to sleep for it. Sleeping for 2 seconds still seems tight to me and to someone else reading the unit test it's not entirely clear why the sleep needs to be there.

shubham2892:
> Actually I think MakeRequest will still be called because handleTaskManifestSingleMessage sends off an async ack goroutine which makes the call to MakeRequest, correct?
> 
> Maybe we can remove the MakeRequest expectation from most of these unit tests, and add another unit test only verifying that MakeRequest gets called. This test could have a longer sleep as the other tests would no longer need to sleep for it. Sleeping for 2 seconds still seems tight to me and to someone else reading the unit test it's not entirely clear why the sleep needs to be there.

Synced offline, we are going ahead with adding the sleep timer. Will consider adding this as a separate test if tests continue to be flaky.

PR #2381, opened 2020-03-01
Fix for data race warning in integ tests

yunhee-l:
### Summary
Fixing data race warning in TestLogdriverOptions and testNetworkMode integ tests.

### Implementation details
From the race report, race seems to be between write access at `go taskEngine.AddTask(testTask)` and read access at `testUpdate := *testTask` (https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/engine_integ_test.go#L518)

### Testing
Ran integ tests few times but wasn't able to repro the race condition. 

N/A

### Licensing

### DIFF
--- a/agent/engine/engine_integ_test.go
+++ b/agent/engine/engine_integ_test.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


agent/engine/engine_integ_test.go
```diff
@@ -600,6 +613,7 @@ func testNetworkMode(t *testing.T, networkMode string) {
 	assert.Equal(t, networkMode, networks[0], "did not find the expected network mode")
 
 	// Kill the existing container now
+	wg.Wait()
```
petderek: AddTask() isn't the only place the data gets modified-- it kicks off a routine in the background that continues to modify its own state. 

Another approach you could take here would be to lock the state of the task before copying it:

```
testTask.SomeRLockFunction()
taskUpdate := *testTask
testTask.SomeRUnlockFunction()

yunhee-l: 👍Im going to test some other ways and create a new PR.

PR #2382, opened 2020-03-02
Fixing data race warning in integ tests

yunhee-l:
### Summary
Fixing data race warnings in integ tests. 

### Implementation details
Copying testTask `taskUpdate := *testTask` causes data race warning. Instead of copying, we are creating new task and relying on the idempotent update handling in `updateTaskUnsafe` deduping by task ARN. 

### Testing

N/A

### Licensing

### DIFF
--- a/agent/engine/engine_integ_test.go
+++ b/agent/engine/engine_integ_test.go
@@ -507,9 +507,11 @@ func TestLabels(t *testing.T) {
 	assert.EqualValues(t, "", state.Config.Labels["label1"])
 
 	// Kill the existing container now
-	taskUpdate := *testTask
+	// Create instead of copying the testTask, to avoid race condition.
+	// AddTask idempotently handles update, filtering by Task ARN.
+	taskUpdate := createTestTask(testArn)
 	taskUpdate.SetDesiredStatus(apitaskstatus.TaskStopped)
-	go taskEngine.AddTask(&taskUpdate)
+	go taskEngine.AddTask(taskUpdate)
 
 	verifyContainerStoppedStateChange(t, taskEngine)
 	verifyTaskStoppedStateChange(t, taskEngine)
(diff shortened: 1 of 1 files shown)


agent/engine/engine_integ_test.go
```diff
@@ -507,9 +507,11 @@ func TestLabels(t *testing.T) {
 	assert.EqualValues(t, "", state.Config.Labels["label1"])
 
 	// Kill the existing container now
-	taskUpdate := *testTask
+	// Create instead of copying the testTask, to avoid race condition.
+	// AddTask idempotently handles update, filtering by Task ARN.
```
sparrc: nit: this two-line comment seems like it should be in the commit message, not a comment in the code. Someone reading this won't really have the context that there used to be a 'copy' operation being performed here

agent/engine/engine_integ_test.go
```diff
@@ -507,9 +507,11 @@ func TestLabels(t *testing.T) {
 	assert.EqualValues(t, "", state.Config.Labels["label1"])
 
 	// Kill the existing container now
-	taskUpdate := *testTask
```
sharanyad: https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/engine_unix_integ_test.go noticed there are other tests that do this too here. any reason why they aren't flakey? should we change those too?

yunhee-l: I want to monitor for that. I've changed all occurrences in this file since we have 2 of them being flakey, and see if it should be applied generally throughout our integ tests.

agent/engine/engine_integ_test.go
yunhee-l: I think the comment still makes sense without the context, as I'd be wondering why are we creating new test task even if I didn't know we switched from copy to create.

PR #2383, opened 2020-03-04
add envfile as task resource and retrieve from s3

jy19:
and write s3 contents to disk

### Summary

Getting this pull request out to get it looked at. Using envfiles as a task resource isn't fully piped through yet. 

### Implementation details

envfile task resource -> retrieve each specified file from s3 and write contents to disk.

### Testing

`make test`

New tests cover the changes:  yes (unit tests)

Add envfile as task resource

### Licensing

### DIFF
--- a/agent/taskresource/envFiles/envfile.go
+++ b/agent/taskresource/envFiles/envfile.go
@@ -1,0 +1,429 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package envFiles
+
+import (
+	"encoding/json"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	"github.com/aws/amazon-ecs-agent/agent/credentials"
+	"github.com/aws/amazon-ecs-agent/agent/s3"
+	"github.com/aws/amazon-ecs-agent/agent/s3/factory"
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+	"github.com/aws/amazon-ecs-agent/agent/utils/oswrapper"
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+
+	"github.com/aws/amazon-ecs-agent/agent/api/task/status"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
+)
+
+const (
+	ResourceName      = "envfile"
+	envFileDirPath    = "envfiles"
+	envTempFilePrefix = "tmp_env"
+
+	s3DownloadTimeout = 30 * time.Second
+)
+
+// EnvironmentFileResource represents envfile as a task resource
+// these environment files are retrieved from s3
+type EnvironmentFileResource struct {
+	cluster     string
+	taskARN     string
+	region      string
+	resourceDir string // path to store env var files
+
+	// env file related attributes
+	environmentFilesSource []apicontainer.EnvironmentFile // list of env file objects
+
+	executionCredentialsID string
+	credentialsManager     credentials.Manager
+	s3ClientCreator        factory.S3ClientCreator
+	os                     oswrapper.OS
+	ioutil                 ioutilwrapper.IOUtil
+
+	// Fields for the common functionality of task resource. Access to these fields are protected by lock.
+	createdAtUnsafe      time.Time
+	desiredStatusUnsafe  resourcestatus.ResourceStatus
+	knownStatusUnsafe    resourcestatus.ResourceStatus
+	appliedStatusUnsafe  resourcestatus.ResourceStatus
+	statusToTransitions  map[resourcestatus.ResourceStatus]func() error
+	terminalReasonUnsafe string
+	terminalReasonOnce   sync.Once
+	lock                 sync.RWMutex
+}
+
+// NewEnvironmentFileResource creates a new EnvironmentFileResource object
+func NewEnvironmentFileResource(cluster, taskARN, region, dataDir string, credentialsManager credentials.Manager,
+	executionCredentialsID string) (*EnvironmentFileResource, error) {
+	envfileResource := &EnvironmentFileResource{
+		cluster:                cluster,
+		taskARN:                taskARN,
+		region:                 region,
+		os:                     oswrapper.NewOS(),
+		ioutil:                 ioutilwrapper.NewIOUtil(),
+		s3ClientCreator:        factory.NewS3ClientCreator(),
+		executionCredentialsID: executionCredentialsID,
+		credentialsManager:     credentialsManager,
+	}
+
+	taskARNFields := strings.Split(taskARN, "/")
+	taskID := taskARNFields[len(taskARNFields)-1]
+	// we save envfiles to path: /var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}
+	envfileResource.resourceDir = filepath.Join(dataDir, envFileDirPath, cluster, taskID)
+
+	envfileResource.initStatusToTransition()
+	return envfileResource, nil
+}
+
+// Initialize initializes the EnvironmentFileResource
+func (envfile *EnvironmentFileResource) Initialize(resourceFields *taskresource.ResourceFields,
+	taskKnownStatus status.TaskStatus,
+	taskDesiredStatus status.TaskStatus) {
+	envfile.lock.Lock()
+	defer envfile.lock.Unlock()
+
+	envfile.initStatusToTransition()
+	envfile.credentialsManager = resourceFields.CredentialsManager
+	envfile.s3ClientCreator = factory.NewS3ClientCreator()
+	envfile.os = oswrapper.NewOS()
+	envfile.ioutil = ioutilwrapper.NewIOUtil()
+
+	// if task isn't in 'created' status and desired status is 'running',
+	// reset the resource status to 'NONE' so we always retrieve the data
+	// this is in case agent crashes
+	if taskKnownStatus < status.TaskCreated && taskDesiredStatus <= status.TaskRunning {
+		envfile.SetKnownStatus(resourcestatus.ResourceStatusNone)
+	}
+}
+
+func (envfile *EnvironmentFileResource) initStatusToTransition() {
+	resourceStatusToTransitionFunc := map[resourcestatus.ResourceStatus]func() error{
+		resourcestatus.ResourceStatus(EnvFileCreated): envfile.Create,
+	}
+	envfile.statusToTransitions = resourceStatusToTransitionFunc
+}
+
+// SetDesiredStatus safely sets the desired status of the resource
+func (envfile *EnvironmentFileResource) SetDesiredStatus(status resourcestatus.ResourceStatus) {
+	envfile.lock.Lock()
+	defer envfile.lock.Unlock()
+
+	envfile.desiredStatusUnsafe = status
+}
+
+// GetDesiredStatus safely returns the desired status of the resource
+func (envfile *EnvironmentFileResource) GetDesiredStatus() resourcestatus.ResourceStatus {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.desiredStatusUnsafe
+}
+
+func (envfile *EnvironmentFileResource) updateAppliedStatusUnsafe(knownStatus resourcestatus.ResourceStatus) {
+	if envfile.appliedStatusUnsafe == resourcestatus.ResourceStatus(EnvFileStatusNone) {
+		return
+	}
+
+	// only apply if resource transition has already finished
+	if envfile.appliedStatusUnsafe <= knownStatus {
+		envfile.appliedStatusUnsafe = resourcestatus.ResourceStatus(EnvFileStatusNone)
+	}
+}
+
+// SetKnownStatus safely sets the currently known status of the resource
+func (envfile *EnvironmentFileResource) SetKnownStatus(status resourcestatus.ResourceStatus) {
+	envfile.lock.Lock()
+	defer envfile.lock.Unlock()
+
+	envfile.knownStatusUnsafe = status
+	envfile.updateAppliedStatusUnsafe(status)
+}
+
+// GetKnownStatus safely returns the currently known status of the resource
+func (envfile *EnvironmentFileResource) GetKnownStatus() resourcestatus.ResourceStatus {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.knownStatusUnsafe
+}
+
+// SetCreatedAt safely sets the timestamp for the resource's creation time
+func (envfile *EnvironmentFileResource) SetCreatedAt(createdAt time.Time) {
+	if createdAt.IsZero() {
+		return
+	}
+
+	envfile.lock.Lock()
+	defer envfile.lock.Unlock()
+
+	envfile.createdAtUnsafe = createdAt
+}
+
+// GetCreatedAt safely returns the timestamp for the resource's creation time
+func (envfile *EnvironmentFileResource) GetCreatedAt() time.Time {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.createdAtUnsafe
+}
+
+// GetName returns the name fo the resource
+func (envfile *EnvironmentFileResource) GetName() string {
+	return ResourceName
+}
+
+// DesiredTerminal returns true if the resource's desired status is REMOVED
+func (envfile *EnvironmentFileResource) DesiredTerminal() bool {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.desiredStatusUnsafe == resourcestatus.ResourceStatus(EnvironmentFileStatus(EnvFileRemoved))
+}
+
+// KnownCreated returns true if the resource's known status is CREATED
+func (envfile *EnvironmentFileResource) KnownCreated() bool {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.knownStatusUnsafe == resourcestatus.ResourceStatus(EnvFileCreated)
+}
+
+// TerminalStatus returns the last transition state of the resource
+func (envfile *EnvironmentFileResource) TerminalStatus() resourcestatus.ResourceStatus {
+	return resourcestatus.ResourceStatus(EnvFileRemoved)
+}
+
+// NextKnownState returns the state that the resource should
+// progress to based on its `KnownState`
+func (envfile *EnvironmentFileResource) NextKnownState() resourcestatus.ResourceStatus {
+	return envfile.GetKnownStatus() + 1
+}
+
+// ApplyTransition calls the function required to move to the specified status
+func (envfile *EnvironmentFileResource) ApplyTransition(nextState resourcestatus.ResourceStatus) error {
+	transitionFunc, ok := envfile.statusToTransitions[nextState]
+	if !ok {
+		return errors.Errorf("resource [%s]: transition to %s impossible", envfile.GetName(),
+			envfile.StatusString(nextState))
+	}
+	return transitionFunc()
+}
+
+// SteadyState returns the transition state of the resource defined as "ready"
+func (envfile *EnvironmentFileResource) SteadyState() resourcestatus.ResourceStatus {
+	return resourcestatus.ResourceStatus(EnvFileCreated)
+}
+
+// SetAppliedStatus sets the applied status of the resource and returns whether
+// the resource is already in a transition
+func (envfile *EnvironmentFileResource) SetAppliedStatus(status resourcestatus.ResourceStatus) bool {
+	envfile.lock.Lock()
+	defer envfile.lock.Unlock()
+
+	if envfile.appliedStatusUnsafe != resourcestatus.ResourceStatus(EnvFileStatusNone) {
+		// set operation failed, return false
+		return false
+	}
+
+	envfile.appliedStatusUnsafe = status
+	return true
+}
+
+// StatusString returns the string representation of the resource status
+func (envfile *EnvironmentFileResource) StatusString(status resourcestatus.ResourceStatus) string {
+	return EnvironmentFileStatus(status).String()
+}
+
+// GetTerminalReason returns an error string to propagate up through to to
+// state change messages
+func (envfile *EnvironmentFileResource) GetTerminalReason() string {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.terminalReasonUnsafe
+}
+
+func (envfile *EnvironmentFileResource) setTerminalReason(reason string) {
+	envfile.lock.Lock()
+	defer envfile.lock.Unlock()
+
+	envfile.terminalReasonOnce.Do(func() {
+		seelog.Infof("envfile resource: setting terminal reason for task: [%s]", envfile.taskARN)
+		envfile.terminalReasonUnsafe = reason
+	})
+}
+
+// Create performs resource creation. This retrieves env file contents in parallel
+// from s3 and writes them to disk
+func (envfile *EnvironmentFileResource) Create() error {
+	seelog.Debugf("Creating envfile resource.")
+	// make sure it has the task execution role
+	executionCredentials, ok := envfile.credentialsManager.GetTaskCredentials(envfile.executionCredentialsID)
+	if !ok {
+		err := errors.New("environment file resource: unable to find execution role credentials")
+		envfile.setTerminalReason(err.Error())
+		return err
+	}
+
+	iamCredentials := executionCredentials.GetIAMRoleCredentials()
+	for _, envfileSource := range envfile.environmentFilesSource {
+		// if we support types besides S3 ARN, we will need to add filtering before the below method is called
+
+		err := envfile.downloadEnvfileFromS3(envfileSource.Value, iamCredentials)
+		if err != nil {
+			err = errors.Wrapf(err, "unable to download envfile with ARN %v from s3", envfileSource.Value)
+			envfile.setTerminalReason(err.Error())
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (envfile *EnvironmentFileResource) downloadEnvfileFromS3(envFilePath string, iamCredentials credentials.IAMRoleCredentials) error {
+	bucket, key, err := s3.ParseS3ARN(envFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "unable to parse bucket and key from s3 ARN specified in environmentFile %s", envFilePath)
+	}
+
+	s3Client, err := envfile.s3ClientCreator.NewS3ClientForBucket(bucket, envfile.region, iamCredentials)
+	if err != nil {
+		return errors.Wrapf(err, "unable to initialize s3 client for bucket %s", bucket)
+	}
+
+	seelog.Debugf("Downlading envfile with bucket name %v and key name %v", bucket, key)
+	// we save envfiles to path: /var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}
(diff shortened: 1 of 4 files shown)


agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
+package envFiles
+
+import (
+	"sync"
+	"time"
+
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
```
yunhee-l: can you run gofmt on all the new files? the import sequence looks not gofmt'd.

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+// these env files are retrieved from s3
+type EnvironmentFileResource struct {
+	cluster   string
+	taskARN   string
+	createdAt time.Time
+	region    string
+	resourceDir string // path to store env var files
+
+	// env file related attributes
+	envfileLocations []apicontainer.EnvironmentFile // list of env file urls
```
yunhee-l: evnfileSource might be more intuitive. With "location" I immediately thought of the dir we are storing it on the host.

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+// represents envfile as a task resource
+// these env files are retrieved from s3
+type EnvironmentFileResource struct {
+	cluster   string
+	taskARN   string
+	createdAt time.Time
+	region    string
+	resourceDir string // path to store env var files
```
yunhee-l: We are not capturing the "type" value (currently only s3) in this object, and implementation including error message assumes s3. We don't have to over engineer ant this point, but it'd be a good data to capture and at least make it available for future iteration.

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+	// make sure it has the task execution role
+	executionCredentials, ok := envfile.credentialsManager.GetTaskCredentials(envfile.executionCredentialsID)
+	if !ok {
+		err := errors.New("environment file resource: unable to find execution role credentials")
+		envfile.setTerminalReason(err.Error())
+		return err
+	}
+
+	for _, envfileLocation := range envfile.envfileLocations {
+		// if we support types besides S3 ARN, we will need to add filtering before the below method is called
```
yunhee-l: we'll need to capture the type to do this (see comment above).

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/pkg/errors"
+	"github.com/aws/amazon-ecs-agent/agent/utils/oswrapper"
+
+	"github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	"github.com/aws/amazon-ecs-agent/agent/api/task/status"
+)
+
+const (
+	envFileDirPath = "envfiles"
```
yunhee-l: in the design doc we didn't mention dedicated dir for env files, did something change?

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	envfile.statusToTransitions = resourceStatusToTransitionFunc
+}
+
+
+func (envfile *EnvironmentFileResource) GetTerminalReason() string {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.terminalReason
```
yunhee-l: if this requires lock to access, name it terminalReasonUnsafe?

agent/taskresource/envFiles/envfilestatus.go
```diff
@@ -0,0 +1,63 @@
...
+	"strings"
+)
+
+type EnvironmentFileStatus resourcestatus.ResourceStatus
+
+const (
+	// zero state of a task resource
+	EnvFileStatusNone EnvironmentFileStatus = iota
+	// task resource is created
+	EnvFileCreated
```
yunhee-l: for my understanding - does "created" mean "downloaded and written to disk"? is there any potential for us to need to handle any inbetween states? e.g. if agent crashes after downloading but before flushing.

agent/taskresource/envFiles/envfile.go
jy19: it did - "S3 env files will be downloaded to `/var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}`"

agent/taskresource/envFiles/envfile.go
jy19: oops will run them

agent/taskresource/envFiles/envfile.go
jy19: ah yeah good point, the name was from when i originally had it as a list of URLs but then i changed it but didn't change the actual name. will update

agent/taskresource/envFiles/envfile.go
jy19: good point, will add

jy19: on second thought, "type" wasn't included because we have a list of the environment files - each object in the list could be of a different type (if we ever support anything besides s3). 
 concerns/thoughts?

yunhee-l: can't we store the type per file? we are accepting a list of env files, are you modeling the entire list as 1 task resource?

agent/taskresource/envFiles/envfile.go
jy19: i removed this - it shouldn't require lock to access. setTerminalReason is using sync.Once so it shouldn't change.

agent/taskresource/envFiles/envfile.go
jy19: type IS stored per file - each EnvironmentFile object has a Value and Type. Unless you mean each file should be a separate resource? So each object in the list is its own resource? I've currently modeled the entire list as 1 task resource because that's how it's defined from the payload.

agent/taskresource/envFiles/envfilestatus.go
jy19: yes, CREATED means downloaded and written to disk - if agent crashes while starting a task (bc here we are still creating the resoruce so task isn't started yet) then on restart the task is re-initialized anyway and the resources will be created again so there aren't really any "in between" states

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+		seelog.Infof("something went wrong trying to persist envfile to disk")
+		return err
+	}
+
+	err = envfile.os.Rename(tmpFile.Name(), fullPathName)
+	if err != nil {
+		seelog.Infof("something went wrong when trying to rename envfile from %s to %s", tmpFile.Name(), fullPathName)
+		return err
+	}
+
```
yumex93: Are we going to have all other necessary functions that needed for task resource in later cr? Like Cleanup function, marshal/unmarshal Json function, etc. This file so far lacks many essential functions for task resource to work correctly.

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (envfile *EnvironmentFileResource) setTerminalReason(reason string) {
+	envfile.terminalReasonOnce.Do(func() {
+		seelog.Infof("envfile resource: setting terminal reason for task: [%s]", envfile.taskARN)
+		envfile.terminalReason = reason
+	})
+}
+
+// resource creation, retrieve env file contents in parallel
```
yumex93: Can we add a bit more details in the comment about the env file is downloaded and save on the host in the function?

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+func (envfile *EnvironmentFileResource) Create() error {
+	seelog.Debugf("Creating envfile resource..")
```
yumex93: nit: two ..

agent/taskresource/envFiles/envfile.go
jy19: yeah i was going to include those when piping this resource through as part of task cycle

agent/taskresource/envFiles/envfile.go
jy19: no, removing..

agent/taskresource/envFiles/envfile.go
jy19: yes my bad, let me add that in

agent/taskresource/envFiles/envfile.go
jy19: nvm i'll add them in now

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+
+	s3DownloadTimeout = 30 * time.Second
+)
+
+// represents envfile as a task resource
+// these env files are retrieved from s3
+type EnvironmentFileResource struct {
+	cluster     string
+	taskARN     string
+	createdAt   time.Time
```
fenxiong: why this while there's a createdAtUnsafe below?

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+	desiredStatusUnsafe resourcestatus.ResourceStatus
+	knownStatusUnsafe   resourcestatus.ResourceStatus
+	appliedStatusUnsafe resourcestatus.ResourceStatus
+	statusToTransitions map[resourcestatus.ResourceStatus]func() error
+	terminalReason      string
+	terminalReasonOnce  sync.Once
+	lock                sync.RWMutex
+}
+
+func NewEnvironmentFileResource(cluster, taskARN, region, dataDirOnHost string, credentialsManager credentials.Manager,
```
fenxiong: i think where the code needs to write to is under ECS_DATADIR (by default /data), rather than under ECS_HOST_DATA_DIR (by default /var/lib/ecs),so dataDirOnHost should probably name as dataDir instead, and you don't really need to know the host dir here

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+		executionCredentialsID: executionCredentialsID,
+		credentialsManager:     credentialsManager,
+	}
+
+	taskARNFields := strings.Split(taskARN, "/")
+	taskID := taskARNFields[len(taskARNFields)-1]
+	// we save envfiles to path: /var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}
+	envfileResource.resourceDir = filepath.Join(dataDirOnHost, envFileDirPath, cluster, taskID)
+
+	return envfileResource, nil
```
fenxiong: you need to call initStatusToTransition here i think. Initialize is only called after agent restarts

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (envfile *EnvironmentFileResource) GetTerminalReason() string {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.terminalReason
+}
+
+func (envfile *EnvironmentFileResource) setTerminalReason(reason string) {
```
fenxiong: the caller of this (`Create`) does not hold a lock. so you should either have a lock here, or have it in the caller

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapper.File) error, fullPathName string) error {
+	// File moves (renaming) are atomic while file writes are not
+	// so we write to a temp file before renaming to actual file
+	// multiple programs calling TempFile will not reference the same file
+	// so this should be ok to be called by multiple go routines
+	tmpFile, err := envfile.ioutil.TempFile(envfile.resourceDir, envTempFilePrefix)
+	if err != nil {
+		seelog.Infof("something went wrong trying to create a temp file with prefix", envTempFilePrefix)
```
fenxiong: lacking %s

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
+package envFiles
```
fenxiong: missing copy right headers. same for all other files

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if err != nil {
+		return errors.Wrapf(err, "unable to download env file with key %s from bucket %s", key, bucket)
+	}
+
+	seelog.Debugf("Downloaded envfile from s3 and saved to %s", downloadPath)
+	return nil
+}
+
+func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapper.File) error, fullPathName string) error {
```
fenxiong: are we supporting windows? if so, writing file in windows is different. refer to agent/statemanager/state_manager_windows.go

agent/taskresource/envFiles/envfile.go
fenxiong: exported method should have comment https://github.com/golang/go/wiki/CodeReviewComments#doc-comments, same for a few other places

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+const (
+	envFileDirPath    = "envfiles"
+	envTempFilePrefix = "tmp_env"
+
+	s3DownloadTimeout = 30 * time.Second
+)
+
+// represents envfile as a task resource
```
fenxiong: comment should begin with the name https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences. i.e. `// EnvironmentFileResource ..`

agent/taskresource/envFiles/envfile.go
fenxiong: nvm i see that credentialspec is using the same code, so it's probably fine

agent/taskresource/envFiles/envfilestatus.go
sharanyad: if the resource has already transitioned to CREATED and agent crashes and restarts before task moves to RUNNING, we need to handle that case by re-downloading the s3 object again..

agent/taskresource/envFiles/envfile.go
jy19: added these, going to put accompanying unit tests in next PR

agent/taskresource/envFiles/envfile.go
jy19: actually going to do this in next PR.

agent/taskresource/envFiles/envfile.go
jy19: `/var/lib/ecs` is what was decided in the design doc. lmk concerns around this

agent/taskresource/envFiles/envfile.go
jy19: thanks for catching

agent/taskresource/envFiles/envfile.go
fenxiong: agent container has a bind mount `/var/lib/ecs/data:/data`. if you want to write to `/var/lib/ecs/data` on the host, you need to write to `/data` in the code.

agent/taskresource/envFiles/envfile.go
jy19: removed this and renamed places where it was accessed to createdAtUnsafe since it required a lock.

agent/taskresource/envFiles/envfile.go
jy19: you mean lock in the get call? added.

agent/taskresource/envFiles/envfile.go
jy19: added

agent/taskresource/envFiles/envfile.go
jy19: done

agent/taskresource/envFiles/envfile.go
jy19: added for all exportable methods

agent/taskresource/envFiles/envfile.go
jy19: yep you're right, added.

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,434 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package envFiles
+
+import (
+	"sync"
+	"time"
+
+	"path/filepath"
```
yumex93: nit: you may want to group your standard library together as the first group.

jy19: idk that's how goimports and gofmt did it.. i reordered it.

agent/taskresource/envFiles/envfile.go
jy19: sure i'll change the name here and make sure to pass in ECS_DATADIR when this method is called

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+		os:                     oswrapper.NewOS(),
+		ioutil:                 ioutilwrapper.NewIOUtil(),
+		s3ClientCreator:        factory.NewS3ClientCreator(),
+		executionCredentialsID: executionCredentialsID,
+		credentialsManager:     credentialsManager,
+	}
+
+	taskARNFields := strings.Split(taskARN, "/")
+	taskID := taskARNFields[len(taskARNFields)-1]
+	// we save envfiles to path: /var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}
```
yumex93: what does this s3filename.env mean?

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+
+	// only apply if resource transition has already finished
+	if envfile.appliedStatusUnsafe <= knownStatus {
+		envfile.appliedStatusUnsafe = resourcestatus.ResourceStatus(EnvFileStatusNone)
+	}
+}
+
+// SetKnownStatus safely sets the currently known status of the resource
+func (envfile *EnvironmentFileResource) SetKnownStatus(status resourcestatus.ResourceStatus) {
+	envfile.lock.RLock()
```
fenxiong: should be Lock/Unlock

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+	return envfile.knownStatusUnsafe
+}
+
+// SetCreatedAt safely sets the timestamp for the resource's creation time
+func (envfile *EnvironmentFileResource) SetCreatedAt(createdAt time.Time) {
+	if createdAt.IsZero() {
+		return
+	}
+
+	envfile.lock.RLock()
```
fenxiong: should be Lock/Unlock

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+
+// GetDesiredStatus safely returns the desired status of the resource
+func (envfile *EnvironmentFileResource) GetDesiredStatus() resourcestatus.ResourceStatus {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.desiredStatusUnsafe
+}
+
+func (envfile *EnvironmentFileResource) updateAppliedStatusUnsafe(knownStatus resourcestatus.ResourceStatus) {
```
fenxiong: this method is not used. should be called in SetKnownStatus i think

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapper.File) error, fullPathName string) error {
+	// File moves (renaming) are atomic while file writes are not
+	// so we write to a temp file before renaming to actual file
+	// multiple programs calling TempFile will not reference the same file
+	// so this should be ok to be called by multiple go routines
+	tmpFile, err := envfile.ioutil.TempFile(envfile.resourceDir, envTempFilePrefix)
+	if err != nil {
+		seelog.Infof("something went wrong trying to create a temp file with prefix %s", envTempFilePrefix)
```
yumex93: The log message should start with capital letter. Also, I think this should be Errorf, since the file downloading will fail here.

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+func (envfile *EnvironmentFileResource) GetCreatedAt() time.Time {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.createdAtUnsafe
+}
+
+// GetName returns the name fo the resource
+func (envfile *EnvironmentFileResource) GetName() string {
+	envfile.lock.RLock()
```
fenxiong: nit: this doesn't need a lock

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+
+// SteadyState returns the transition state of the resource defined as "ready"
+func (envfile *EnvironmentFileResource) SteadyState() resourcestatus.ResourceStatus {
+	return resourcestatus.ResourceStatus(EnvFileCreated)
+}
+
+// SetAppliedStatus sets the applied status of the resource and returns whether
+// the resource is already in a transition
+func (envfile *EnvironmentFileResource) SetAppliedStatus(status resourcestatus.ResourceStatus) bool {
+	envfile.lock.RLock()
```
fenxiong: should be Lock/Unlock

agent/taskresource/envFiles/envfile.go
fenxiong: no, what i mean is here, when you set the field you should have the lock

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+// EnvironmentFileResource represents envfile as a task resource
+// these environment files are retrieved from s3
+type EnvironmentFileResource struct {
+	cluster     string
+	taskARN     string
+	region      string
+	resourceDir string // path to store env var files
+
+	// env file related attributes
+	environmentFilesSourceUnsafe []apicontainer.EnvironmentFile // list of env file objects
```
fenxiong: does this field ever change after initializing from NewEnvironmentFileResource? seems like it doesn't change so don't need lock or be marked as unsafe

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+			return &envfileStatus
+		}(),
+		EnvironmentFilesSource: envfile.getEnvironmentFilesSource(),
+		ExecutionCredentialsID: envfile.getExecutionCredentialsID(),
+	})
+
+}
+
+// UnmarshalJSON deserializes the raw JSON to an EnvironmentFileResource struct
+func (envfile *EnvironmentFileResource) UnmarshalJSON(b []byte) error {
```
yumex93: I think you miss the assignment of executionCredentialsID here.

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+func (envfile *EnvironmentFileResource) getEnvironmentFilesSource() []apicontainer.EnvironmentFile {
+	envfile.lock.RLock()
+	defer envfile.lock.RUnlock()
+
+	return envfile.environmentFilesSourceUnsafe
+}
+
+func (envfile *EnvironmentFileResource) getExecutionCredentialsID() string {
```
fenxiong: nit: same as environmentFileSource, i don't quite see why you need a lock for executionCredentialsID, since no one writes to it. read can be concurrent without lock

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+	// TODO
+	return nil
+}
+
+type EnvironmentFileResourceJSON struct {
+	TaskARN                string                         `json:"taskARN"`
+	CreatedAt              *time.Time                     `json:"createdAt,omitempty"`
+	DesiredStatus          *EnvironmentFileStatus         `json:"desiredStatus"`
+	KnownStatus            *EnvironmentFileStatus         `json:"knownStatus"`
+	EnvironmentFilesSource []apicontainer.EnvironmentFile `json:"envfileResources"`
```
fenxiong: nit: maybe envFileSources? should match the field name

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+	return envfile.executionCredentialsID
+}
+
+// Cleanup removes env files downloaded for the task
+func (envfile *EnvironmentFileResource) Cleanup() error {
+	// TODO
+	return nil
+}
+
+type EnvironmentFileResourceJSON struct {
```
fenxiong: i don't think this need to be exported?

agent/taskresource/envFiles/envfile.go
```diff
@@ -0,0 +1,432 @@
...
+	taskKnownStatus status.TaskStatus,
+	taskDesiredStatus status.TaskStatus) {
+	envfile.lock.Lock()
+	defer envfile.lock.Unlock()
+
+	envfile.initStatusToTransition()
+	envfile.credentialsManager = resourceFields.CredentialsManager
+	envfile.s3ClientCreator = factory.NewS3ClientCreator()
+	envfile.os = oswrapper.NewOS()
+	envfile.ioutil = ioutilwrapper.NewIOUtil()
```
yumex93: There is one thing here you may want to think about is whether we want to redownload the S3 file when agent crash and restart during resource CREATED but task hasn't go to RUNNING. if we want to redownload, we need to reset the task resource status back to none here.

agent/taskresource/envFiles/envfile.go
jy19: the name of the s3 file but it is required to have extension `.env`

agent/taskresource/envFiles/envfile.go
jy19: oops you're right, also task arn. thanks for catching

agent/taskresource/envFiles/envfile.go
jy19: oh i thought the credentials get rotated so it might change. i must've remembered wrong, credentials rotating during task resource creation seems unlikely

agent/taskresource/envFiles/envfile.go
jy19: yep my bad, there's just an empty line where it should've been called. added it back now

agent/taskresource/envFiles/envfile.go
jy19: right it doesn't change, was thinking of something else. unmarked as unsafe.

agent/taskresource/envFiles/envfile.go
jy19: made sure matches

agent/taskresource/envFiles/envfile.go
jy19: un-exported

agent/taskresource/envFiles/envfile.go
jy19: i didn't look at my RLock/RUnlock(s) carefully, my bad. should all be fixed now.

agent/taskresource/envFiles/envfile.go
jy19: i think you're right. i guess if agent crashes we'd jsut re-download everything again, even if files were previously partially downloaded.

agent/taskresource/envFiles/envfile.go
jy19: changed the logs in `err != nil` statements to be `Errorf` and capitalized letters

agent/taskresource/envFiles/envfilestatus.go
```diff
@@ -0,0 +1,77 @@
...
+	"errors"
+	"strings"
+
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+)
+
+type EnvironmentFileStatus resourcestatus.ResourceStatus
+
+const (
+	// zero state of a task resource
```
fenxiong: exported const comment: 
// EnvFileStatusNone ...
// EnvFileCreated ...
// EnvFileRemoved ...

yunhee-l:
Travis check is failing for all 3 platforms with "The command "make static-check" exited with 2"

jy19:
go vet is failing on something related to the locks, am fixing.

agent/taskresource/envFiles/envfilestatus.go
jy19: fixed

agent/taskresource/envFiles/envfile_test.go
```diff
@@ -0,0 +1,260 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
petderek: this needs a:
```
// +build unit
```
at the top

PR #2385, opened 2020-03-05
add env ECS_CONTAINER_METADATA_URI_V4 in V4

cyastella:
### Summary

we decide to add ECS_CONTAINER_METADATA_URI_V4 -> points to 169.254.170.2/v4/endpointID into each container to let customers switch from v3 to v4 to fetch networkInterfaceProperty for awsvpc task.
### Implementation details

Inject ECS_CONTAINER_METADATA_URI_V4 in each container so that users can switch to the new one, we decide also keep v3 ECS_CONTAINER_METADATA_URI so that customers can use the previous one if they want.
### Testing

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -58,6 +58,14 @@ const (
 	// MetadataURIFormat defines the URI format for v3 metadata endpoint
 	MetadataURIFormat = "http://169.254.170.2/v3/%s"
 
+	// MetadataURIEnvVarNameV4 defines the name of the environment
+	// variable in containers' config, which can be used by the containers to access the
+	// v4 metadata endpoint
+	MetadataURIEnvVarNameV4 = "ECS_CONTAINER_METADATA_URI_V4"
+
+	// MetadataURIFormat defines the URI format for v4 metadata endpoint
+	MetadataURIFormatV4 = "http://169.254.170.2/v4/%s"
+
 	// SecretProviderSSM is to show secret provider being SSM
 	SecretProviderSSM = "ssm"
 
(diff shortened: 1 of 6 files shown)


agent/api/container/container.go
```diff
@@ -58,6 +58,14 @@ const (
 	// MetadataURIFormat defines the URI format for v3 metadata endpoint
 	MetadataURIFormat = "http://169.254.170.2/v3/%s"
 
+	// MetadataURIEnvironmentVariableNameV4 defines the name of the environment
+	// variable in containers' config, which can be used by the containers to access the
+	// v4 metadata endpoint
+	MetadataURIEnvironmentVariableNameV4 = "ECS_CONTAINER_METADATA_URI_V4"
```
sharanyad: prefer using short var names in go, this can be `MetadataURIEnvVarNameV4`

agent/api/task/task.go
```diff
@@ -719,6 +720,19 @@ func (task *Task) initializeContainersV3MetadataEndpoint(uuidProvider utils.UUID
 	}
 }
 
+// initializeContainersV4MetadataEndpoint generates an v4 endpoint id which we reuse the v3 container id (they are the same) for each container, constructs the
```
sharanyad: could you break this comment into multiple lines?

agent/api/container/container.go
cyastella: sure!

agent/api/task/task.go
cyastella: yeah, will do

PR #2389, opened 2020-03-10
Fix for the attchmentIndex

cyastella:
### Summary

fix the attachmentIndex since when it is int type, it can't be change to the omitempty. Thus we change it to the int pointer type.
### Implementation details

change the attachmentIndex to pointer.

- In awsvpc mode, it points to zero value, 

- in bridge/host mode, it become nil.

### Testing

### Licensing

### DIFF
--- a/agent/handlers/v4/response.go
+++ b/agent/handlers/v4/response.go
@@ -56,11 +56,11 @@ type Network struct {
 type NetworkInterfaceProperties struct {
 	// AttachmentIndex reflects the `index` specified by the customer (if any)
 	// while creating the task with `awsvpc` mode.
-	AttachmentIndex int `json:"AttachmentIndex"`
+	AttachmentIndex *int `json:"AttachmentIndex,omitempty"`
 	// IPV4SubnetCIDRBlock is the subnet CIDR netmask associated with network interface.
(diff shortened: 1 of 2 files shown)


cyastella:
the build is breaking because the `aws:feature/v4_metadata` branch is out of sync with `aws:dev`; to pass the CI tests, will need to merge in `HEAD` from `aws:dev` branch.

cyastella:
> the build is breaking because the `aws:feature/v4_metadata` branch is out of sync with `aws:dev`; to pass the CI tests, will need to merge in `HEAD` from `aws:dev` branch.
> 
> created #2393 to run the CI tests with the latest changes from dev

#2393 passed all CI checks; will rebase `aws:feature/v4_metadata` to `HEAD` of `aws:dev` after this change is merged

agent/handlers/v4/response.go
```diff
@@ -172,11 +172,18 @@ func newNetworkInterfaceProperties(task *apitask.Task) (NetworkInterfaceProperti
 			"v4 metadata response: unable to parse subnet ipv4 address '%s'",
 			eni.SubnetGatewayIPV4Address)
 	}
+
+	var attachmentIndexPtr *int = nil
```
fenxiong: "= nil" is not needed. it's nil by default

cyastella: sure, will change it.

PR #2390, opened 2020-03-10
retrieve envfile from s3 concurrently

jy19:
### Summary

This pull request updates how we retrieve envfiles from s3.

### Implementation details

Each envfile specified will spin up a go routine so that multiple env files can be retrieved in parallel

### Testing

`make test` is successful

N/A

### Licensing

### DIFF
--- a/agent/taskresource/envFiles/envfile.go
+++ b/agent/taskresource/envFiles/envfile.go
@@ -15,6 +15,7 @@ package envFiles
 
 import (
 	"encoding/json"
+	"fmt"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -271,7 +272,7 @@ func (envfile *EnvironmentFileResource) setTerminalReason(reason string) {
 	})
 }
 
-// Create performs resource creation. This retrieves env file contents in parallel
+// Create performs resource creation. This retrieves env file contents concurrently
 // from s3 and writes them to disk
 func (envfile *EnvironmentFileResource) Create() error {
 	seelog.Debugf("Creating envfile resource.")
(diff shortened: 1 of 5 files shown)


jy19:
would like to get this PR looked at while the merge one is blocked by functional tests. This branch is branched off of that one and has the merge changes but i'll clean it up later.

yumex93:
This pr has additional commits that not belonging to you. I think for best practice, we don't want them to be here considering they are not part of the changes you want to be reviewed in ur pr.  Normally we would just create another merge pr to merge new commits from dev at the end.

agent/taskresource/envFiles/envfile.go
```diff
@@ -271,7 +272,7 @@ func (envfile *EnvironmentFileResource) setTerminalReason(reason string) {
 	})
 }
 
-// Create performs resource creation. This retrieves env file contents in parallel
+// Create performs resource creation. This retrieves env file contents concurrently
```
yumex93: We need to do the performance test testing downloading 10 S3 files at the same time before we release, in case there are some issues we didn't know about.

sharanyad:
> This pr has additional commits that not belonging to you.

i believe the last merge from dev PR was rebased and merged instead of with a merge commit, which led to different commit IDs from dev showing as new commits here. We may have to force push to this branch with the right commits from dev.

agent/taskresource/envFiles/envfile.go
yunhee-l: will make a note of this to include as test case.

jy19:
yes, this is on a separate branch - i only put this out bc our functional tests were taking a while and i wanted this to get looked at. closing in favor of https://github.com/aws/amazon-ecs-agent/pull/2392/

PR #2391, opened 2020-03-11
Envfile cleanup logic and unit tests, including oswrapper changes

yunhee-l:
### Summary
- Implementation of envfile.Cleanup 
- unit tests added

### Implementation details
Because the resource directory path includes `envfiles` and `task_id` (/var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}), any .env file under the directory is candidate for task cleanup, so I did not match name by name.  

### Testing

New tests cover the changes: Yes, new unit tests added

N/A

### Licensing

### DIFF
--- a/agent/taskresource/envFiles/envfile.go
+++ b/agent/taskresource/envFiles/envfile.go
@@ -15,6 +15,7 @@ package envFiles
 
 import (
 	"encoding/json"
+	"fmt"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -38,6 +39,7 @@ const (
 	ResourceName      = "envfile"
 	envFileDirPath    = "envfiles"
 	envTempFilePrefix = "tmp_env"
+	envFileExtension  = ".env"
 
 	s3DownloadTimeout = 30 * time.Second
 )
@@ -358,9 +360,14 @@ func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapp
 	return nil
 }
 
-// Cleanup removes env files downloaded for the task
+// Cleanup removes env file directory for the task
 func (envfile *EnvironmentFileResource) Cleanup() error {
-	// TODO
+	err := envfile.os.RemoveAll(envfile.resourceDir)
+	if err != nil {
+		return fmt.Errorf("unable to remove envfile resource directory %s: %v", envfile.resourceDir, err)
+	}
+
+	seelog.Infof("Removed envfile resource directory at %s", envfile.resourceDir)
 	return nil
 }
 
--- a/agent/taskresource/envFiles/envfile_test.go
+++ b/agent/taskresource/envFiles/envfile_test.go
@@ -47,7 +47,7 @@ const (
 	accessKeyId            = "accessKey"
 	secretAccessKey        = "secret"
 	s3Bucket               = "s3Bucket"
-	s3Key                  = "s3key"
+	s3Key                  = "s3key.env"
 	tempFile               = "tmp_file"
 )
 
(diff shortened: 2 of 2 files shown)


agent/taskresource/envFiles/envfile.go
```diff
@@ -358,9 +359,32 @@ func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapp
 	return nil
 }
 
-// Cleanup removes env files downloaded for the task
+// Cleanup removes all env files downloaded for the task
```
fenxiong: why not just remove the whole `/var/lib/ecs/data/envfiles/cluster_name/task_id/` directory? i don't think we should leave an empty `/var/lib/ecs/data/envfiles/cluster_name/task_id/` folder behind when the task is gone

yunhee-l: Good point. I'll change it.

agent/taskresource/envFiles/envfile.go
```diff
@@ -358,9 +360,14 @@ func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapp
 	return nil
 }
 
-// Cleanup removes env files downloaded for the task
+// Cleanup removes env file directory for the task
 func (envfile *EnvironmentFileResource) Cleanup() error {
-	// TODO
+	err := envfile.os.RemoveAll(envfile.resourceDir)
```
sharanyad: how about force removing it if removal fails due to some reason?

agent/taskresource/envFiles/envfile.go
```diff
@@ -358,9 +360,14 @@ func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapp
...
+	if err != nil {
+		return fmt.Errorf("unable to remove envfile resource directory %s: %v", envfile.resourceDir, err)
+	}
+
+	seelog.Infof("Removed envfile resource directory at %s", envfile.resourceDir)
```
sharanyad: is it possible to add more context such as task arn here ?

agent/taskresource/envFiles/envfile_test.go
```diff
@@ -260,3 +260,35 @@ func TestCreateRenameFileError(t *testing.T) {
...
+	mockOS, _, mockIOUtil, mockCredentialsManager, mockS3ClientCreator, _, done := setup(t)
+	defer done()
+
+	envfiles := []container.EnvironmentFile{
+		sampleEnvironmentFile(fmt.Sprintf("arn:aws:s3:::%s/%s", s3Bucket, s3Key), "s3"),
+	}
+
+	envfileResource := newMockEnvfileResource(envfiles, mockCredentialsManager, mockS3ClientCreator, mockOS, mockIOUtil)
+
+	gomock.InOrder(
```
sharanyad: this is probably not needed since there is only one mock call

agent/taskresource/envFiles/envfile.go
yunhee-l: resource path includes cluster name and task ID, that should be enough identifying information right?

agent/taskresource/envFiles/envfile_test.go
yunhee-l: removed.

agent/taskresource/envFiles/envfile.go
yunhee-l: There's no force option available. We do the same firelens resource dir. I can add retry if this is concern?

agent/taskresource/envFiles/envfile.go
sharanyad: ah, it should be good enough

agent/taskresource/envFiles/envfile.go
sharanyad: ok, we can add retries if this causes issues in the future.

PR #2392, opened 2020-03-11
retrieve envfiles from s3 concurrently using go routines

jy19:
### Summary

This pull request updates how we retrieve envfiles from s3.

### Implementation details

Each envfile specified will spin up a go routine so that multiple env files can be retrieved in parallel

### Testing

`make test` is successful. Yunhee will add a benchmark into test plan.

N/A

### Licensing

### DIFF
--- a/agent/taskresource/envFiles/envfile.go
+++ b/agent/taskresource/envFiles/envfile.go
@@ -15,6 +15,7 @@ package envFiles
 
 import (
 	"encoding/json"
+	"fmt"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


agent/taskresource/envFiles/envfile.go
```diff
@@ -283,30 +284,48 @@ func (envfile *EnvironmentFileResource) Create() error {
...
+		errorString := strings.Join(terminalReasons, ";")
+		envfile.setTerminalReason(errorString)
+		return errors.New(errorString)
 	}
 
 	return nil
 }
 
-func (envfile *EnvironmentFileResource) downloadEnvfileFromS3(envFilePath string, iamCredentials credentials.IAMRoleCredentials) error {
+func (envfile *EnvironmentFileResource) downloadEnvfileFromS3(envFilePath string, iamCredentials credentials.IAMRoleCredentials,
```
yunhee-l: in error cases, should we log the task id/arn or does it get logged in the method that calls this?

jy19: the full s3 arn should get logged. should we also log the task arn?

yunhee-l: s3 arn doesn't provide any identifying info about for which task the env file download was attempted so yeah i think we should add it.

PR #2398, opened 2020-03-17
create envfile task resource

jy19:
### Summary

this commit initializes the environment files resource when task starts up, and reads the environment variables from those files when containers are created

note: i am getting this PR out to get eyes on it, there are additional unit tests that still need to be added.

### Implementation details

When task starts up -> containers are checked to have environment files. if so, envfile resource is created. after files are retrieved (done in previous PR), when containers are created, those retrieved files are read and those environment variables are merged into a container's existing environment variables.

### Testing

`make test` is successful

create envfile task resource on task start up as needed

### Licensing


jy19:
gocyclo fails on `task (*Task).PostUnmarshalTask`

agent/api/task/task.go
```diff
@@ -2548,3 +2556,62 @@ func getDomainForPartition(region string) string {
...
+func (task *Task) requireEnvfiles() bool {
+	for _, container := range task.Containers {
+		if container.ShouldCreateWithEnvFiles() {
+			return true
+		}
+	}
+	return false
+}
+
+func (task *Task) initializeEnvfilesResource(config *config.Config, credentialsManager credentials.Manager, ) error {
```
fenxiong: extra comma after `credentials.Manager`

agent/api/task/task.go
```diff
@@ -2548,3 +2556,62 @@ func getDomainForPartition(region string) string {
...
+// that was downloaded to disk and merges it with existing environment variables
+func (task *Task) ReadEnvVarsFromEnvfiles(container *apicontainer.Container) *apierrors.ResourceInitError {
+	var envfileResource *envFiles.EnvironmentFileResource
+	if container.ShouldCreateWithEnvFiles() {
+		resource, ok := task.getEnvfilesResource()
+		if !ok {
+			err := errors.New("task environment files: unable to retrieve environment files resource")
+			return apierrors.NewResourceInitError(task.Arn, err)
+		}
+		envfileResource = resource[0].(*envFiles.EnvironmentFileResource)
```
fenxiong: in initializeEnvfilesResource the envfileResource is added per container, but here you only get the first one, which doesn't seem right?

agent/taskresource/envFiles/envfile.go
```diff
@@ -30,16 +30,19 @@ import (
 	"github.com/cihub/seelog"
 	"github.com/pkg/errors"
 
+	"bufio"
```
fenxiong: nit: move this to the first group (i.e. before encoding/json)

agent/taskresource/envFiles/envfile.go
```diff
@@ -452,3 +458,50 @@ func (envfile *EnvironmentFileResource) UnmarshalJSON(b []byte) error {
 
 	return nil
 }
+
+// ReadEnvVarsFromEnvFiles reads the environment files that have been downloaded
+// and puts them into a list of maps
+func (envfile *EnvironmentFileResource) ReadEnvVarsFromEnvfiles() ([]interface{}, error) {
```
fenxiong: why make this returns a list of interface{} rather than a map[string]string?

agent/taskresource/envFiles/envfile.go
```diff
@@ -332,6 +337,7 @@ func (envfile *EnvironmentFileResource) downloadEnvfileFromS3(envFilePath string
 	seelog.Debugf("Downlading envfile with bucket name %v and key name %v", bucket, key)
 	// we save envfiles to path: /var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}
 	downloadPath := filepath.Join(envfile.resourceDir, key)
+	envfile.environmentFilesLocation = append(envfile.environmentFilesLocation, downloadPath)
```
fenxiong: this should be protected by lock

agent/api/container/container.go
```diff
@@ -917,6 +930,38 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
...
+
+	// create map if does not exist
+	if c.Environment == nil {
+		c.Environment = make(map[string]string)
+	}
+
+	// envVarsList is a list of map, where each map is from an envfile
+	// iterate over this sequentially because the original order of the
+	// environment files give precedence to the environment variables
+	for _, envVarsIn := range envVarsList {
```
yunhee-l: any repeated variable in more than one env file needs to be resolved by precedence as well, in the order as defined in the "environmentFiles" list. I don't see how that's handled here?

agent/taskresource/envFiles/envfile.go
```diff
@@ -452,3 +458,50 @@ func (envfile *EnvironmentFileResource) UnmarshalJSON(b []byte) error {
...
+	envVars := make(map[string]string)
+	for scanner.Scan() {
+		line := scanner.Text()
+		// if line starts with a #, ignore
+		if strings.HasPrefix(line, commentIndicator) {
+			continue
+		}
+		// only read the line that has "="
+		if strings.Contains(line, envVariableDelimiter) {
+			variables := strings.Split(line, "=")
```
yunhee-l: this will read a line that has "=" anywhere, e.g. strings.Split("=xyz ", "=") returns ["", "xyz"]. Im inclined to use regex or any other validation that makes sure either side of = is at least length of one.

agent/api/container/container.go
yumex93: Looks like it depends on the envVarsList order? Considering the logic here is once the var exists, it will not be overriden.

agent/api/task/task.go
```diff
@@ -57,6 +57,7 @@ import (
 	"github.com/containernetworking/cni/libcni"
 	dockercontainer "github.com/docker/docker/api/types/container"
 	"github.com/pkg/errors"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/envFiles"
```
yumex93: you need to run go fmt to correct the import order.

agent/api/task/task.go
```diff
@@ -2548,3 +2556,62 @@ func getDomainForPartition(region string) string {
...
+	}
+	return false
+}
+
+func (task *Task) initializeEnvfilesResource(config *config.Config, credentialsManager credentials.Manager, ) error {
+
+	for _, container := range task.Containers {
+		envfileResource, err := envFiles.NewEnvironmentFileResource(config.Cluster, task.Arn, config.AWSRegion, config.DataDir,
+			container.EnvironmentFiles, credentialsManager, task.ExecutionCredentialsID)
+		if err != nil {
```
yumex93: Since each container has their own task resource, you may want to have container name or other container identifier in the error message so people can understand which one is causing the error.

agent/api/task/task.go
yumex93: Also based on the information in the env file task resource and how you add task resource to the **ResourcesMapUnsafe**, don't seem there is an easy way for you to figure out which task resource maps to which container.

agent/engine/docker_task_engine.go
```diff
@@ -1018,6 +1018,14 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		}
 	}
 
+	if container.ShouldCreateWithEnvFiles() {
+		err := task.ReadEnvVarsFromEnvfiles(container)
```
yumex93: I feel it would be better for the name to be **MergeEnvVarsFromEnvFiles**, it makes me confusing for a second that where you merge the env vars until I read the function logic.

yumex93: We don't have the precedence requirements for secrets and the env file, right?

agent/taskresource/envFiles/envfile.go
yunhee-l: we decided not to add any agent specific validations, so at least we should understand what happens when invalid line is passed to docker.

agent/taskresource/envFiles/envfile.go
jy19: because the list preserves the order of the env files while map does not

agent/api/task/task.go
jy19: good point - also added check for containers having env files before creating a resource. instead of checking all containers and then checking again.

agent/engine/docker_task_engine.go
jy19: i think the requirement is just that all existing env vars (in this case secret as well) have precendence - though it's highly unlikely that someone will create a secret but then also store it as plaintext in s3..or is it 🤔

agent/taskresource/envFiles/envfile.go
fenxiong: oh, what i meant was why this is a list of interface{} rather than a list of map[string]string

agent/engine/docker_task_engine.go
jy19: also renamed the method.

agent/api/task/task.go
jy19: done

agent/taskresource/envFiles/envfile.go
jy19: good point - and is the existing logic to ignore any lines that is a comment or doesn't match the regex ok?

agent/api/container/container.go
jy19: the original thought is that envVarsList is a list that follows the original order of what was entered in the task def, so if an env var already exists in the map, we don't add to it because whatever was entered into the map first has precedence (just based on the list order). however, i add to the list of downloaded paths with a go routine so i need to double check if that's still the case..

agent/api/task/task.go
jy19: removed.

agent/taskresource/envFiles/envfile.go
```diff
@@ -332,6 +337,7 @@ func (envfile *EnvironmentFileResource) downloadEnvfileFromS3(envFilePath string
 	seelog.Debugf("Downlading envfile with bucket name %v and key name %v", bucket, key)
 	// we save envfiles to path: /var/lib/ecs/data/envfiles/cluster_name/task_id/${s3filename.env}
```
yumex93: So here the s3filename is the key name. Key is the unique identifier in the bucket. So what if for two different s3 files, they have the same key name, but from different buckets? I think we are just going to override each other here.

agent/api/task/task.go
jy19: yep realize i am missing some logic to make sure envfiles only get sent to corresponding containers, will update.

agent/taskresource/envFiles/envfile.go
jy19: good point, i will add bucket to path.

agent/taskresource/envFiles/envfile.go
jy19: done

agent/taskresource/envFiles/envfile.go
jy19: go was giving me problems about it so i was under the impression that i couldn't return a list of map but i must've done something wrong 😂 i can change it to make it more clear

agent/api/container/container.go
```diff
@@ -917,6 +930,37 @@ func (c *Container) MergeEnvironmentVariables(envVars map[string]string) {
...
+	// create map if does not exist
+	if c.Environment == nil {
+		c.Environment = make(map[string]string)
+	}
+
+	// envVarsList is a list of map, where each map is from an envfile
+	// iterate over this sequentially because the original order of the
+	// environment files give precedence to the environment variables
+	for _, envVars := range envVarsList {
+		//envVars, ok := envVarsIn.(map[string]string)
```
yumex93: Don't forget to remove these comments.

jy19: lol oops yeah will do

agent/api/container/container.go
```diff
@@ -1050,3 +1090,10 @@ func (c *Container) GetFirelensConfig() *FirelensConfig {
 
 	return c.FirelensConfig
 }
+
+func (c *Container) GetEnvironmentFiles() []EnvironmentFile {
```
fenxiong: comment for exported method

agent/api/task/task.go
```diff
@@ -2548,3 +2554,72 @@ func getDomainForPartition(region string) string {
...
+}
+
+func (task *Task) initializeEnvfilesResource(config *config.Config, credentialsManager credentials.Manager) error {
+
+	for _, container := range task.Containers {
+		if container.ShouldCreateWithEnvFiles() {
+			envfileResource, err := envFiles.NewEnvironmentFileResource(config.Cluster, task.Arn, config.AWSRegion, config.DataDir,
+				container.Name, container.EnvironmentFiles, credentialsManager, task.ExecutionCredentialsID)
+			if err != nil {
+				return errors.Wrap(err, fmt.Sprintf("unable to initialize envfiles resource for container %s", container.Name))
```
fenxiong: nit: can just use errors.Wrapf here

agent/taskresource/envFiles/envfile_test.go
```diff
@@ -29,12 +29,14 @@ import (
...
 	mock_s3 "github.com/aws/amazon-ecs-agent/agent/s3/mocks"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	mock_bufio "github.com/aws/amazon-ecs-agent/agent/utils/bufiowrapper/mocks"
 	mock_ioutilwrapper "github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper/mocks"
 	mock_oswrapper "github.com/aws/amazon-ecs-agent/agent/utils/oswrapper/mocks"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	//"bytes"
```
fenxiong: should remove

agent/taskresource/envFiles/envfile.go
```diff
@@ -452,3 +469,58 @@ func (envfile *EnvironmentFileResource) UnmarshalJSON(b []byte) error {
...
+func (envfile *EnvironmentFileResource) GetContainerName() string {
+	return envfile.containerName
+}
+
+// ReadEnvVarsFromEnvFiles reads the environment files that have been downloaded
+// and puts them into a list of maps
+func (envfile *EnvironmentFileResource) ReadEnvVarsFromEnvfiles() ([]map[string]string, error) {
+	var envVarsPerEnvfile []map[string]string
+
+	for _, envfilePath := range envfile.environmentFilesLocation {
```
fenxiong: would be safer if using a getter with lock for environmentFilesLocation. otherwise it's implicitly relying on the fact that no one writes to it in between

agent/taskresource/envFiles/envfile.go
```diff
@@ -32,34 +33,40 @@ import (
...
 type EnvironmentFileResource struct {
-	cluster     string
-	taskARN     string
-	region      string
-	resourceDir string // path to store env var files
+	cluster       string
+	taskARN       string
+	region        string
+	resourceDir   string // path to store env var files
+	containerName string
```
fenxiong: i didn't see containerName and environmentFilesLocation added to environmentFileResourceJSON. they probably should?

agent/taskresource/envFiles/envfile.go
```diff
@@ -14,6 +14,7 @@
 package envFiles
 
 import (
+	//"bufio"
```
fenxiong: should remove

agent/taskresource/envFiles/envfile.go
jy19: ah yeah, good catch. will add.

agent/taskresource/envFiles/envfile.go
jy19: oh my bad, i forgot to update this part after adding lock to that var. will update, thanks for catching.

agent/taskresource/envFiles/envfile.go
jy19: removed.

agent/taskresource/envFiles/envfile_test.go
jy19: removed.

agent/api/task/task.go
jy19: yep done.

agent/api/container/container.go
jy19: added.

agent/taskresource/envFiles/envfile.go
yunhee-l: im ok with that

jy19: instead of regex i added a super simple "if both sides length > 0" lmk if you have any concerns, i'm open to adding it as regex

agent/api/task/task.go
```diff
@@ -2548,3 +2554,72 @@ func getDomainForPartition(region string) string {
...
+
+func (task *Task) getEnvfilesResource(containerName string) (taskresource.TaskResource, bool) {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	resources, ok := task.ResourcesMapUnsafe[envFiles.ResourceName]
+	for _, resource := range resources {
+		envfileResource := resource.(*envFiles.EnvironmentFileResource)
+		if envfileResource.GetContainerName() == containerName {
+			return envfileResource, ok
```
yumex93: nit: if ok is false, it isn't used. Since the for loop will not be executed when resources is nil. You can use true to replace ok here.

agent/api/task/task.go
```diff
@@ -2548,3 +2554,72 @@ func getDomainForPartition(region string) string {
...
+	// was not able to retrieve envfile resource for specified container name
+	return nil, false
+}
+
+// MergeEnvVarsFromEnvfiles should be called when creating a container -
+// this method reads the environment variables specified in the environment files
+// that was downloaded to disk and merges it with existing environment variables
+func (task *Task) MergeEnvVarsFromEnvfiles(container *apicontainer.Container) *apierrors.ResourceInitError {
+	var envfileResource *envFiles.EnvironmentFileResource
+	if container.ShouldCreateWithEnvFiles() {
```
yumex93: Since you already do this ShouldCreateWithEnvFiles check when you call MergeEnvVarsFromEnvfiles function, why you need to call it again here?

agent/api/task/task.go
```diff
@@ -2548,3 +2554,72 @@ func getDomainForPartition(region string) string {
...
+	}
+
+	return nil
+}
+
+func (task *Task) getEnvfilesResource(containerName string) (taskresource.TaskResource, bool) {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	resources, ok := task.ResourcesMapUnsafe[envFiles.ResourceName]
```
fenxiong: should check the "ok" here before proceeding to for loop

agent/taskresource/envFiles/envfile_test.go
```diff
@@ -284,9 +286,134 @@ func TestEnvFileCleanupResourceDirRemoveFail(t *testing.T) {
...
 
-	envfileResource := newMockEnvfileResource(envfiles, mockCredentialsManager, mockS3ClientCreator, mockOS, mockIOUtil)
+	envfileResource := newMockEnvfileResource(envfiles, nil, mockCredentialsManager, mockS3ClientCreator, mockOS, mockIOUtil)
 
 	mockOS.EXPECT().RemoveAll(resourceDir).Return(errors.New("error response"))
 
 	assert.Error(t, envfileResource.Cleanup())
 }
+
+func TestReadEnvVarsFromEnvfiles(t *testing.T) {
```
yumex93: Should we change this test to read from multiple env files to verify the env vars with same key are overriden correctly?

jy19: i can add an additional test for this use case.

agent/api/task/task.go
jy19: you're right, this is unneeded. removing.

agent/api/task/task.go
jy19: replaced

agent/api/task/task.go
jy19: added check

agent/taskresource/envFiles/envfile_test.go
jy19: wait this is already tested in container_test.go
this method wouldn't actually override any envvars, but return everything in the file

PR #2399, opened 2020-03-17
Add Docker ID of containers to logs

ubhattacharjya:
### Summary

This PR adds container runtime ID to the agent log lines which do not include them for better debugging
### Testing

Tested that the log lines pop up with docker id in agent logs

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -1100,7 +1100,7 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
 }
 
 func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *apicontainer.Container) dockerapi.DockerContainerMetadata {
-	seelog.Infof("Task engine [%s]: starting container: %s", task.Arn, container.Name)
+	seelog.Infof("Task engine [%s]: starting container: %s (Runtime ID: %s)", task.Arn, container.Name, container.GetRuntimeID())
 	client := engine.client
 	if container.DockerConfig.Version != nil {
 		client = client.WithVersion(dockerclient.DockerVersion(*container.DockerConfig.Version))
@@ -1392,11 +1392,11 @@ func (engine *DockerTaskEngine) applyContainerState(task *apitask.Task, containe
 	}
 	metadata := transitionFunction(task, container)
 	if metadata.Error != nil {
-		seelog.Infof("Task engine [%s]: error transitioning container [%s] to [%s]: %v",
-			task.Arn, container.Name, nextState.String(), metadata.Error)
+		seelog.Infof("Task engine [%s]: error transitioning container [%s (Runtime ID: %s)] to [%s]: %v",
+			task.Arn, container.Name, container.GetRuntimeID(), nextState.String(), metadata.Error)
 	} else {
-		seelog.Debugf("Task engine [%s]: transitioned container [%s] to [%s]",
-			task.Arn, container.Name, nextState.String())
+		seelog.Debugf("Task engine [%s]: transitioned container [%s (Runtime ID: %s)] to [%s]",
+			task.Arn, container.Name, container.GetRuntimeID(), nextState.String())
 		engine.saver.Save()
 	}
 	return metadata
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -345,8 +345,8 @@ func (mtask *managedTask) waitEvent(stopWaiting <-chan struct{}) bool {
 		mtask.handleDesiredStatusChange(acsTransition.desiredStatus, acsTransition.seqnum)
 		return false
 	case dockerChange := <-mtask.dockerMessages:
-		seelog.Infof("Managed task [%s]: got container [%s] event: [%s]",
-			mtask.Arn, dockerChange.container.Name, dockerChange.event.Status.String())
+		seelog.Infof("Managed task [%s]: got container [%s (Runtime ID: %s)] event: [%s]",
+			mtask.Arn, dockerChange.container.Name, dockerChange.container.GetRuntimeID(), dockerChange.event.Status.String())
 		mtask.handleContainerChange(dockerChange)
 		return false
 	case resChange := <-mtask.resourceStateChangeEvent:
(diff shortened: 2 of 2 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1100,7 +1100,7 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
 }
 
 func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *apicontainer.Container) dockerapi.DockerContainerMetadata {
-	seelog.Infof("Task engine [%s]: starting container: %s", task.Arn, container.Name)
+	seelog.Infof("Task engine [%s]: starting container: %s (Docker ID: %s)", task.Arn, container.Name, container.RuntimeID)
```
sharanyad: how about calling it Runtime ID?

sparrc: How about Container ID? Doesnt `docker ps` call it Container ID?

agent/engine/task_manager.go
```diff
@@ -345,8 +345,8 @@ func (mtask *managedTask) waitEvent(stopWaiting <-chan struct{}) bool {
 		mtask.handleDesiredStatusChange(acsTransition.desiredStatus, acsTransition.seqnum)
 		return false
 	case dockerChange := <-mtask.dockerMessages:
-		seelog.Infof("Managed task [%s]: got container [%s] event: [%s]",
-			mtask.Arn, dockerChange.container.Name, dockerChange.event.Status.String())
+		seelog.Infof("Managed task [%s]: got container [%s (Docker ID: %s)] event: [%s]",
+			mtask.Arn, dockerChange.container.Name, dockerChange.container.RuntimeID, dockerChange.event.Status.String())
```
fenxiong: RuntimeID is not read-only (can be set by SetRuntimeID) so read should be protected by lock with GetRuntimeID(), same for other places

agent/engine/docker_task_engine.go
fierlion: > how about calling it Runtime ID?

 Should be a quick find/replace if you agree with either sharanyad or sparrc.  My vote’s for “Runtime ID” (since it’s `container.RuntimeID`)

sparrc: but this is an external-facing log message. Docker doesnt externally call it "Runtime ID" anywhere. For example googling "docker runtime id" produces cryptic results. Googling "docker container id" produces exactly what we're talking about.

sharanyad: In ECS, we have been calling it Runtime ID - https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Container.html#ECS-Type-Container-runtimeId

agent/engine/task_manager.go
ubhattacharjya: Thanks for the suggestion. Changed it.

agent/engine/docker_task_engine.go
ubhattacharjya: Thanks for the suggestions. Changed it to "Runtime ID" in order to be consistent with ECS public documentation.

agent/engine/docker_task_engine.go
```diff
@@ -1392,11 +1392,11 @@ func (engine *DockerTaskEngine) applyContainerState(task *apitask.Task, containe
 	}
 	metadata := transitionFunction(task, container)
 	if metadata.Error != nil {
-		seelog.Infof("Task engine [%s]: error transitioning container [%s] to [%s]: %v",
-			task.Arn, container.Name, nextState.String(), metadata.Error)
+		seelog.Infof("Task engine [%s]: error transitioning container [%s (Runtime ID: %s)] to [%s]: %v",
```
sharanyad: if there is error in transitioning, it could also have empty container ID..but don't want to add a logic to check states and log accordingly. this is just to keep in mind ID can be empty

agent/engine/docker_task_engine.go
```diff
@@ -1392,11 +1392,11 @@ func (engine *DockerTaskEngine) applyContainerState(task *apitask.Task, containe
...
+			task.Arn, container.Name, container.GetRuntimeID(), nextState.String(), metadata.Error)
 	} else {
-		seelog.Debugf("Task engine [%s]: transitioned container [%s] to [%s]",
-			task.Arn, container.Name, nextState.String())
+		seelog.Debugf("Task engine [%s]: transitioned container [%s (Runtime ID:: %s)] to [%s]",
```
sharanyad: nit: Runtime : - single colon

agent/engine/task_manager.go
```diff
@@ -1001,8 +1001,8 @@ func (mtask *managedTask) containerNextState(container *apicontainer.Container)
 	}
 	if blocked, err := dependencygraph.DependenciesAreResolved(container, mtask.Containers,
 		mtask.Task.GetExecutionCredentialsID(), mtask.credentialsManager, mtask.GetResources()); err != nil {
-		seelog.Debugf("Managed task [%s]: can't apply state to container [%s] yet due to unresolved dependencies: %v",
-			mtask.Arn, container.Name, err)
+		seelog.Debugf("Managed task [%s]: can't apply state to container [%s (Runtime ID: %s)] yet due to unresolved dependencies: %v",
```
sharanyad: same, can be empty runtime ID?

agent/engine/task_manager.go
```diff
@@ -655,16 +655,16 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
 		fallthrough
 	case apicontainerstatus.ContainerCreated:
 		// No need to explicitly stop containers if this is a * -> NONE/CREATED transition
-		seelog.Warnf("Managed task [%s]: error creating container [%s]; marking its desired status as STOPPED: %v",
-			mtask.Arn, container.Name, event.Error)
+		seelog.Warnf("Managed task [%s]: error creating container [%s (Runtime ID: %s)]; marking its desired status as STOPPED: %v",
+			mtask.Arn, container.Name, container.GetRuntimeID(), event.Error)
```
sharanyad: runtime ID will be empty, not needed here

agent/engine/task_manager.go
```diff
@@ -644,8 +644,8 @@ func (mtask *managedTask) handleEventError(containerChange dockerContainerChange
 		// the task fail here, will let create container handle it instead.
 		// If the agent pull behavior is default, use local image cache directly,
 		// assuming it exists.
-		seelog.Errorf("Managed task [%s]: error while pulling container %s and image %s, will try to run anyway: %v",
-			mtask.Arn, container.Name, container.Image, event.Error)
+		seelog.Errorf("Managed task [%s]: error while pulling container %s (Runtime ID: %s) and image %s, will try to run anyway: %v",
```
sharanyad: not needed here, since ID will be empty

agent/engine/task_manager.go
ubhattacharjya: There can be empty runtime ID. But did not add any logic for that so that the code does not become cumbersome.

agent/engine/docker_task_engine.go
ubhattacharjya: There can be empty runtime ID. But did not add any logic for that so that the code does not become cumbersome.

PR #2406, opened 2020-03-24
Logging tweaks for clarity and level consistency

sparrc:
# Logging fixups/tweaks

### 1

When the agent loses it's websocket connections, it prints a cryptic ERROR message for ACS, and a slightly better INFO message about TCS, like below:

```
2020-03-12T14:59:50Z [ERROR] Error getting message from ws backend: error: [websocket: close 1002 (protocol error): Channel long idle: No message is received, close the channel], messageType: [-1] 
2020-03-12T14:59:50Z [INFO] Error from tcs; backing off: websocket: close 1002 (protocol error): Channel long idle: No message is received, close the channel
```

We should change these log messages so that:
1. the TCS log message is an ERROR not INFO
2. the ACS log message more clearly states that the agent has lost it's ACS connection, rather than a generic 'ws' connection.

### 2

When `docker stop` fails, we should print an error message, not an INFO:

```
2020-03-12T15:01:17Z [INFO] DockerGoClient: error stopping container XXXXXXXXXXXXXX97c9031fad0abd04e1c17d658f9dd5cd7dXXXXXXXXXXXXXXXX: Cannot connect to the Docker daemon at npipe:////./pipe/docker_engine. Is the docker daemon running?
```

### 3

When TCS stats start up, it is normal for there to not be enough data to send stats. However, we log this WARN/error message that is actually just expected behavior:

```
2020-03-12T14:59:35Z [WARN] Error getting cpu stats, err: Need at least 2 data points in queue to calculate CW stats set, container: XXXXXXXXXXXXXX97c9031fad0abd04e1c17d658f9dd5cd7dXXXXXXXXXXXXXXXX
```

We should log this as an INFO message and not include "error" in the message. Possibly we might even want this to be a DEBUG message.

### 4 

When we have a 'skippable' state change event, we return an error from an internal function and then log it as if we ecountered a problem. We should distinguish between 'skippable' state changes and actual errors in this function so that we dont log lots of messages like this for no reason:

```
level=info time=2020-03-23T18:17:36Z msg="Managed task [arn.../foo]: unable to create state change event for container [foobar]: create container state change event api: status not recognized by ECS: NONE" module=task_manager.go
```

### 5

when pulling an image fails we log a WARN message, this should be changed to ERROR:

```
level=warn time=2020-03-22T23:58:42Z msg="DockerGoClient: failed to pull image 1111111111111.dkr.ecr.us-west-1.amazonaws.com/brazil/bloom:1.0.200000.0: context canceled" module=docker_client.go
```

Also 'context cancelled' in this instance probably means that the pull timed-out, so we should check if we can explicitly log that a timeout occured here so it's more clear what happened.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -372,7 +372,11 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 			// Stop receiving and sending messages from and to ACS when
 			// client.Serve returns an error. This can happen when the
 			// the connection is closed by ACS or the agent
-			seelog.Infof("ACS connection closed: %v", err)
+			if err == nil || err == io.EOF {
+				seelog.Info("ACS Websocket connection closed for a valid reason")
+			} else {
+				seelog.Errorf("Error: lost websocket connection with Agent Communication Service (ACS): %v", err)
+			}
 			return err
 		}
 	}
--- a/agent/api/statechange.go
+++ b/agent/api/statechange.go
@@ -24,6 +24,7 @@ import (
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/agent/statechange"
+	"github.com/pkg/errors"
 
 	"github.com/aws/aws-sdk-go/aws"
 )
@@ -88,30 +89,20 @@ type AttachmentStateChange struct {
 	Attachment *apieni.ENIAttachment
 }
 
-// SkippableStateChange is an error used to indicate that this state change can be omitted
-type SkippableStateChange struct {
-	s string
-}
-
-func (e SkippableStateChange) Error() string {
-	return e.s
-}
-
 // NewTaskStateChangeEvent creates a new task state change event
-// returns error type api.SkippableStateChange if the state change doesn't need
-// to be sent to the ECS backend.
+// returns error if the state change doesn't need to be sent to the ECS backend.
 func NewTaskStateChangeEvent(task *apitask.Task, reason string) (TaskStateChange, error) {
 	var event TaskStateChange
 	taskKnownStatus := task.GetKnownStatus()
 	if !taskKnownStatus.BackendRecognized() {
-		return event, SkippableStateChange{
-			fmt.Sprintf("create task state change event api: status not recognized by ECS: %v", taskKnownStatus),
-		}
+		return event, errors.Errorf(
+			"create task state change event api: status not recognized by ECS: %v",
+			taskKnownStatus)
 	}
 	if task.GetSentStatus() >= taskKnownStatus {
-		return event, SkippableStateChange{
-			fmt.Sprintf("create task state change event api: status [%s] already sent", taskKnownStatus.String()),
-		}
+		return event, errors.Errorf(
+			"create task state change event api: status [%s] already sent",
+			taskKnownStatus.String())
 	}
 
 	event = TaskStateChange{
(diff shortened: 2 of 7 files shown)


agent/wsclient/client.go
```diff
@@ -387,7 +387,7 @@ func (cs *ClientServerImpl) ConsumeMessages() error {
 
 		default:
 			// Unexpected error occurred
-			seelog.Errorf("Error getting message from ws backend: error: [%v], messageType: [%v] ",
```
sparrc: this will be logged upstream

agent/acs/handler/acs_handler.go
```diff
@@ -372,7 +372,11 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 			// Stop receiving and sending messages from and to ACS when
 			// client.Serve returns an error. This can happen when the
 			// the connection is closed by ACS or the agent
-			seelog.Infof("ACS connection closed: %v", err)
+			if err == nil || err == io.EOF {
```
sparrc: for consistency with tcs: https://github.com/aws/amazon-ecs-agent/blob/master/agent/tcs/handler/handler.go#L83

agent/engine/task_manager.go
```diff
@@ -504,29 +504,39 @@ func (mtask *managedTask) emitResourceChange(change resourceStateChange) {
...
+			return
+		}
+		seelog.Errorf("Managed task [%s]: error emitting event for task [%s]: %v",
 			task.Arn, reason, err)
 		return
 	}
 	seelog.Infof("Managed task [%s]: sending task change event [%s]", mtask.Arn, event.String())
 	mtask.stateChangeEvents <- event
-	seelog.Infof("Managed task [%s]: sent task change event [%s]", mtask.Arn, event.String())
+	seelog.Debugf("Managed task [%s]: sent task change event [%s]", mtask.Arn, event.String())
```
fenxiong: i'd prefer this and line 539 stay at info level so that if we somehow miss sending some state changes, we know the problem is on the task manager side or the handler that consumes stateChangeEvents channel (many people don't turn on debug logs by default)

agent/api/statechange.go
```diff
@@ -89,19 +88,30 @@ type AttachmentStateChange struct {
 	Attachment *apieni.ENIAttachment
 }
 
+// SkippableStateChange is an error used to indicate that this state change can be omitted
+type SkippableStateChange struct {
```
fenxiong: is this needed? seems like NewTaskStateChangeEvent and NewContainerStateChangeEvent only return this type of error, so we can just use the default error type

agent/engine/task_manager.go
sparrc: okay, I can leave this at info

agent/api/statechange.go
sparrc: sure I could leave default error type, I thought there was one instance where it _could_ return another error but now I cant find it 🤔

agent/tcs/handler/handler.go
```diff
@@ -84,7 +84,7 @@ func StartSession(params *TelemetrySessionParams, statsEngine stats.Engine) erro
 			seelog.Info("TCS Websocket connection closed for a valid reason")
 			backoff.Reset()
 		} else {
-			seelog.Infof("Error from tcs; backing off: %v", tcsError)
+			seelog.Errorf("Error: lost websocket connection with ECS Telemetry service (TCS): %v", tcsError)
```
shubham2892: Can this be warn instead of errorf, customers look at these logs and think that something is wrong with the Agent when essentially it is not sending metrics to tcs.

sparrc: I prefer it to be an error because the agent should have a connection to TCS at all times. In my experience losing connection with TCS often indicates a deeper issue on the instance and indicates that an error has occurred.

PR #2407, opened 2020-03-25
envfiles: create directory where the envfile will be written

jy19:
### Summary

creates the directories for where environment files will be written

### Testing

`make test` successful - running agent locally i can see the envfiles are created.

### Licensing

### DIFF
--- a/agent/taskresource/envFiles/envfile.go
+++ b/agent/taskresource/envFiles/envfile.go
@@ -16,6 +16,7 @@ package envFiles
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


yunhee-l:
unit tests are failing on windows because of file path delimiter is "\" not "/".

agent/taskresource/envFiles/envfile.go
```diff
@@ -322,6 +323,18 @@ func (envfile *EnvironmentFileResource) Create() error {
 	return nil
 }
 
+// createEnvfileDirectory creates the directory that we will be writing the
+// envfile to - needs to be called for each different bucket
+func (envfile *EnvironmentFileResource) createEnvfileDirectory(bucket string) error {
+	envfileDir := filepath.Join(envfile.resourceDir, bucket)
```
yunhee-l: we need to include key as subdirectory path too

agent/taskresource/envFiles/envfile.go
```diff
@@ -322,6 +323,19 @@ func (envfile *EnvironmentFileResource) Create() error {
 	return nil
 }
 
+// createEnvfileDirectory creates the directory that we will be writing the
+// envfile to - needs to be called for each different envfile
+func (envfile *EnvironmentFileResource) createEnvfileDirectory(bucket, key string) error {
+	// create directories to include bucket and key but not the actual resulting file
+	envfileDir := filepath.Join(envfile.resourceDir, bucket, strings.TrimRight(key, string(filepath.Separator)))
```
yunhee-l: we should use filepath.Dir instead of strings.TrimRight.

fenxiong: +1

agent/taskresource/envFiles/envfile_test.go
```diff
@@ -129,10 +130,12 @@ func TestCreateWithEnvVarFile(t *testing.T) {
 		},
 	}
 
+	envDirectories := filepath.Join(resourceDir, s3Bucket)
```
yunhee-l: we should test s3Key being more than just the file name, to test that we are creating all the subdirectories correctly. And for verification,  I wouldn't use the same api as what the code we are trying to test is using, but instead of known fixed value.

PR #2410, opened 2020-03-26
minor unit test fixes

yunhee-l:
### Summary
- missed state file version update in unit test after merge
- making envfile save test deal with longer s3 file path

### Implementation details
- v27 -> v28 for envfile state file load test
- using longer s3key containing path in envfile_test

### Testing

New tests cover the changes: No

N/A
### Licensing


agent/taskresource/envFiles/envfile_test.go
```diff
@@ -49,7 +49,9 @@ const (
 	accessKeyId            = "accessKey"
 	secretAccessKey        = "secret"
 	s3Bucket               = "s3Bucket"
-	s3Key                  = "s3key.env"
+	s3Path                 = "path/to/envfile/"
+	s3File                 = "s3key.env"
+	s3Key                  = s3Path+s3File
```
jy19: should we use file separator since this test will run on both unix and windows

yunhee-l: good point, updating

PR #2433, opened 2020-04-21
instance health: initial commit

mythri-garaga:
### Summary

This change publishes instance health data to backend on a regular interval
for determining instance health status. The initial effort is to collect docker daemon
health data by storing the count of Docker API calls made, error responses and send this data to the backend.

### Implementation details

1. Added a instance health to metrics:
    - new instancehealth interface defines the behaviour such as updating and accessing metrics; resetting the count
    - genericMetric implements the InstanceHealth interface
2. dockerapi/docker_client.go: Update count for every API call count, their error count and error response
3. tcs/client.go: Added a functionalities to publish instance health data to backend every 20s
                  by creating a request. This also takes care of resetting the count

### Testing

`make test` ran successfully on local, also tested manually on AL1.

### Licensing


agent/metrics/instancehealth/instancehealth.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (gc *genericMetric) IncrementErrorCount() {
+	atomic.AddInt64(&gc.errors, 1)
+}
+
+func (gc *genericMetric) GetErrorCount() int64 {
+	return atomic.LoadInt64(&gc.errors)
+}
+
+func (gc *genericMetric) RecordErrorMessage(errorMessage string) {
```
sparrc, line 39: I would suggest making this an internal function if it doesnt need to be part of the public interface at the moment. Maybe it doesnt even need to be a function separate from RecordError?

agent/metrics/instancehealth/interface.go
```diff
@@ -0,0 +1,30 @@
...
+	// This function returns the API call count
+	GetCallCount()
+
+	// This function increments the API's error count
+	IncrementErrorCount()
+
+	// This function returns the API error count
+	GetErrorCount()
+
+	// Records the error message
```
sparrc, line 22: in the comment add that it also increments error count

agent/tcs/client/client.go
```diff
@@ -315,6 +359,19 @@ func (cs *clientServer) createPublishHealthRequests() ([]*ecstcs.PublishHealthRe
 	return requests, nil
 }
 
+// createPublishInstanceHealthRequests creates the requests to publish instance health
+func (cs *clientServer) createPublishInstanceHealthRequests() *ecstcs.PublishInstanceHealthRequest {
```
sparrc, line 363: There's a possibility here of metrics being lost between calling the Get* functions and running the deferred ResetCount function.

I'd suggest having some function(s) that do an atomic "Swap" to both return the current value and reset the count at the same time. You can use https://golang.org/pkg/sync/atomic/#SwapInt64 for this.

It might be a good idea to just change ResetCount to do the swap and return the 'old' values. I think the interface would be safer this way because I don't think there is currently a use-case where we would want to reset the counts without also getting them. ie instead of having a ResetCount function we could have two functions: GetAndResetCallCount() and GetAndResetErrorCount() or a single GetAndResetCount() that returns the two counts.

agent/tcs/client/client.go
```diff
@@ -279,6 +281,48 @@ func (cs *clientServer) publishHealthMetricsOnce() error {
...
+	// Publish instance metrics immediately after we connect and wait for ticks. This makes
+	// sure that there is no data loss when a scheduled metrics publishing fails
+	// due to a connection reset.
+	err := cs.publishInstanceHealthMetricsOnce()
+	if err != nil {
+		seelog.Warnf("Unable to publish instance health metrics: %v", err)
+	}
+	for {
+		select {
+		case <-cs.publishHealthTicker.C:
```
sparrc, line 300: Let's make a separate publishInstanceHealthTicker that takes it's own publishing interval. We will want this to be longer than the metrics publish interval because we need to make sure that we have sufficient data before sending. We will figure out an exact number through prototyping but let's start with 3 minutes.

mythri-garaga: Done.

agent/tcs/client/client.go
mythri-garaga: Thanks for pointing this. Updated!

agent/metrics/instancehealth/interface.go
mythri-garaga: Done.

agent/metrics/instancehealth/instancehealth.go
mythri-garaga: Agreed and updated.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -720,11 +746,13 @@ func (dg *dockerGoClient) RemoveContainer(ctx context.Context, dockerID string,
 		return resp
 	case <-ctx.Done():
 		err := ctx.Err()
-		// Context has either expired or canceled. If it has timed out,
-		// send back the DockerTimeoutError
 		if err == context.DeadlineExceeded {
+			// send back the DockerTimeoutError
+			// Context has either expired or canceled. If it has timed out,
```
fierlion, line 751: nit: the comment statement doesn't make sense.  I think it's missing
`// send back the DockerTimeoutError`

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1176,14 +1211,16 @@ func (dg *dockerGoClient) InspectVolume(ctx context.Context, name string, timeou
 	case resp := <-response:
 		return resp
 	case <-ctx.Done():
-		// Context has either expired or canceled. If it has timed out,
-		// send back the DockerTimeoutError
 		err := ctx.Err()
 		if err == context.DeadlineExceeded {
+			// send back the DockerTimeoutError
+			// Context has either expired or canceled. If it has timed out,
```
fierlion, line 1217: nit is this missing
`// send back the DockerTimeoutError`

PR #2438, opened 2020-04-23
Register custom logger before it gets used

shubham2892:
### Summary

When `ECS_LOGLEVEL` is set in Agent config file, healthcheck outputs shows following error -- ` [Error] format error: unrecognized formatter at 1: EcsAgentLogfmt\n"`

It's because when `ECS_LOGLEVEL` is set, logger tries to load `EcsAgentLogFmt` custom formatter before it is initiated. 
`EcsAgentLogFmt` gets loaded here -- https://github.com/aws/amazon-ecs-agent/blob/master/agent/logger/log.go#L97 which is being called by `SetLevel()`  here -- https://github.com/aws/amazon-ecs-agent/blob/master/agent/logger/log.go#L97.

Currently custom formatter is initiated here  -- https://github.com/aws/amazon-ecs-agent/blob/master/agent/logger/log.go#L167

Moving the initiation of custom formatter before `SetLevel` ensure that the formatter is initiated before it is getting loaded.
 
### Testing
Steps to reproduce the issue -- 
1. Set `ECS_LOGLEVEL=debug` in /etc/ecs/ecs.config on al2.
2. Restart agent
3. `docker inspect <agent container id>`. Health check shows error `format error: unrecognized formatter at 1: EcsAgentLogfmt\n`

After the change, the health check was passing with `ECS_LOGLEVEL` set and I could see logs getting generated as expected in Agent log directory.

Bug -  Health check output shows error message when ECS_LOGLEVEL is set in Agent config file. 

### Licensing

### DIFF
--- a/agent/logger/log.go
+++ b/agent/logger/log.go
@@ -140,6 +140,13 @@ func init() {
 		MaxRollCount:  DEFAULT_MAX_ROLL_COUNT,
 	}
 
+	if err := seelog.RegisterCustomFormatter("EcsAgentLogfmt", logfmtFormatter); err != nil {
+		seelog.Error(err)
+	}
+	if err := seelog.RegisterCustomFormatter("EcsAgentJson", jsonFormatter); err != nil {
+		seelog.Error(err)
+	}
+
 	SetLevel(os.Getenv(LOGLEVEL_ENV_VAR))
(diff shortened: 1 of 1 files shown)


petderek:
You should fix your commit. The author is root.

PR #2443, opened 2020-05-01
Backoff container stats collection when listener fails

sparrc:
### Summary

The purpose of this change is to backoff and jitter when stats collection fails on a container. By default we open a streaming stats listener for each container. If this listener fails, we currently immediately try to open it again for every container.

This can create a situation where stats listeners are failing due to a resource-stressed docker daemon. The agent then exacerbates the situation by spamming the docker daemon with potentially thousands of stats requests in just a few minutes.

This change causes the agent to backoff when the stats listener fails for a container. We are also increasing the timeout on the listener from 5s to 10s as stress testing has shown that waiting longer can sometimes result in getting stats successfully.

### Testing

a single-instance cluster was created with m5.large. 204 tasks were spun up running a simple dd command. At the same time, the [containerd-stress](https://github.com/containerd/containerd/blob/master/cmd/containerd-stress/main.go) tool was run for 30 minutes using the following command:

```
sudo ~/go/bin/containerd-stress --duration 30m --concurrent 24
```

We can then count the number of times that we see the "Unable to decode stats for container" message to see how many times we attempt to open a new stats stream that fails.

Without change:
```
% grep "Unable to decode stats for container" < /var/log/ecs/ecs-agent.log|wc -l
30268
```
With this change:
```
% grep "Unable to decode stats for container" < /var/log/ecs/ecs-agent.log | wc -l
3645
```

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -1321,11 +1321,12 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 	client, err := dg.sdkDockerClient()
 	if err != nil {
 		cancelRequest()
+		return nil, nil, nil
 	}
 
 	statsC := make(chan *types.StatsJSON)
-	errC := make(chan error)
 	done := make(chan struct{})
+	errC := make(chan error)
 	var resp types.ContainerStats
 
 	if !dg.config.PollMetrics {
@@ -1383,6 +1384,7 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 				resp, err = client.ContainerStats(subCtx, id, stream)
 				if err != nil {
 					errC <- fmt.Errorf("DockerGoClient: Unable to retrieve stats for container %s: %v", id, err)
+					return
 				}
 
 				// Returns a *Decoder and takes in a readCloser
@@ -1391,6 +1393,7 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 				err := decoder.Decode(data)
 				if err != nil {
 					errC <- fmt.Errorf("DockerGoClient: Unable to decode stats for container %s: %v", id, err)
+					return
 				}
 
 				statsC <- data
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -165,7 +165,7 @@ type DockerClient interface {
 
 	// Stats returns a channel of stat data for the specified container. A context should be provided so the request can
 	// be canceled.
-	Stats(context.Context, string, time.Duration) (<-chan *types.StatsJSON, <-chan error, <-chan struct{})
+	Stats(context.Context, string, time.Duration) (<-chan *types.StatsJSON, <-chan error)
 
 	// Version returns the version of the Docker daemon.
 	Version(context.Context, time.Duration) (string, error)
@@ -1315,26 +1315,29 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
 }
 
 // Stats returns a channel of *types.StatsJSON entries for the container.
-func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeout time.Duration) (<-chan *types.StatsJSON, <-chan error, <-chan struct{}) {
+func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeout time.Duration) (<-chan *types.StatsJSON, <-chan error) {
 	subCtx, cancelRequest := context.WithCancel(ctx)
 
+	errC := make(chan error)
+	statsC := make(chan *types.StatsJSON)
 	client, err := dg.sdkDockerClient()
 	if err != nil {
 		cancelRequest()
-		return nil, nil, nil
+		go func() {
+			// upstream function should consume error
+			errC <- err
+			close(statsC)
+		}()
+		return statsC, errC
 	}
 
-	statsC := make(chan *types.StatsJSON)
-	done := make(chan struct{})
-	errC := make(chan error)
 	var resp types.ContainerStats
-
 	if !dg.config.PollMetrics {
 		// Streaming metrics is the default behavior
 		seelog.Infof("DockerGoClient: Starting streaming metrics for container %s", id)
 		go func() {
 			defer cancelRequest()
-			defer close(done)
+			defer close(statsC)
 			// ContainerStats outputs an io.ReadCloser and an OSType
 			stream := true
 			resp, err = client.ContainerStats(subCtx, id, stream)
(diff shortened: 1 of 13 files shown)


agent/stats/container.go
```diff
@@ -78,7 +81,9 @@ func (container *StatsContainer) collect() {
 				// 'NoSuchContainer', 'InactivityTimeoutExceeded' etc are silently consumed.
 				// We rely on state reconciliation with docker task engine at this point of
 				// time to stop collecting metrics.
-				seelog.Debugf("Error querying stats for container %s: %v", dockerID, err)
+				d := backoff.Duration()
```
fenxiong, line 84: should the backoff be reset when we don't have error?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1315,28 +1315,31 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
 	if err != nil {
 		cancelRequest()
-		return nil, err
+		return nil, nil, nil
 	}
 
-	// Create channel to hold the stats
-	statsChnl := make(chan *types.StatsJSON)
+	statsC := make(chan *types.StatsJSON)
+	done := make(chan struct{})
```
fenxiong, line 1328: is this needed? seems like just having statsC and closing it at the end can work as well

agent/stats/container.go
sparrc: We could, although it's not entirely clear to me why this would ever disconnect without an error, except when the container is terminal, which ends the stats processing completely.

agent/dockerclient/dockerapi/docker_client.go
sparrc: Yes this is needed because I have changed the behavior of the for-loop here: https://github.com/aws/amazon-ecs-agent/pull/2443/files#diff-543d5b6c06bcf2d177e4c83e48884a0fR117-R132

Closing the stats channel would cause <-dockerStats to continuously emit in the above loop.

Previously we were using a "magic" feature of Golang where a channel can be placed in a for-loop and it will exit once the channel is closed.

sparrc: the reason the loop was changed is because we're not just listening on one channel anymore (stats), we need to listen for stats and errors

fenxiong: i think you can use `stat, ok := <-dockerStats` to check whether the channel is closed and break the loop when it is

agent/stats/container.go
fenxiong: i see, then i guess this is probably fine

agent/stats/container_test.go
```diff
@@ -125,48 +127,6 @@ func TestContainerStatsCollection(t *testing.T) {
 	}
 }
 
-func TestContainerStatsCollectionReconnection(t *testing.T) {
```
fenxiong, line 128: why is this test deleted? is the reconnection case tested somewhere else?

agent/dockerclient/dockerapi/docker_client.go
sparrc: okay, will change to this

agent/stats/container_test.go
sparrc: thanks for catching, that was a mistake

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1315,28 +1315,30 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
 
-	// Create channel to hold the stats
-	statsChnl := make(chan *types.StatsJSON)
+	statsC := make(chan *types.StatsJSON)
+	errC := make(chan error)
 	var resp types.ContainerStats
 
 	if !dg.config.PollMetrics {
+		// Streaming metrics is the default behavior
+		seelog.Infof("DockerGoClient: Starting streaming metrics for container %s", id)
```
sharanyad, line 1333: can't this just be debug log?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1315,28 +1315,30 @@ func (dg *dockerGoClient) APIVersion() (dockerclient.DockerVersion, error) {
...
 // Stats returns a channel of *types.StatsJSON entries for the container.
-func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeout time.Duration) (<-chan *types.StatsJSON, error) {
+func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeout time.Duration) (<-chan *types.StatsJSON, <-chan error) {
 	subCtx, cancelRequest := context.WithCancel(ctx)
 
 	client, err := dg.sdkDockerClient()
 	if err != nil {
 		cancelRequest()
-		return nil, err
+		return nil, nil
```
sharanyad, line 1324: any reason why we are not creating errC before this and returning with error?

agent/stats/container.go
```diff
@@ -65,6 +67,7 @@ func (container *StatsContainer) StopStatsCollection() {
 
 func (container *StatsContainer) collect() {
 	dockerID := container.containerMetadata.DockerID
+	backoff := retry.NewExponentialBackoff(time.Second*1, time.Second*20, 0.75, 2)
```
sharanyad, line 70: these values should be constants. also how did we come up with these?

sparrc: why constants if they're only used in one place?

agent/dockerclient/dockerapi/docker_client.go
sparrc: Corresponding log message for polling metrics below is Info, this shouldn't log very frequently, only once per container under normal circumstances so I'd prefer it to be info.

agent/stats/container.go
sharanyad: more for readability.

agent/dockerclient/dockerapi/docker_client.go
sparrc: because we'd have to spin up a goroutine just to send an error down the channel.

agent/stats/container.go
sparrc: The numbers are basically arbitrary. Jitter is relatively high as I wanted to make sure that many containers don't synchronize on reopening stats streamer at the same time.

agent/dockerclient/dockerapi/docker_client.go
sharanyad: right, it just seems a bit counter-intuitive below that we are checking are nils and then returning an error.

sparrc: okay, I can change it to spin up a goroutine

agent/stats/container.go
sharanyad: ok, lets see from testing how this plays up with the 20 second metrics publish interval to ECS backend service

sparrc:
Another test was run to see this change's effect on publishing metrics to TACS. From agent's perspective it seems that freeing up resources actually results in more metrics making it to TACS. With this change we published 355 metrics in 10 minutes, on dev branch we published 190. The test was:

1. Put a log message into here: https://github.com/aws/amazon-ecs-agent/blob/61d418ea2c535c3f7356e27eec23b05a0eb90319/agent/tcs/client/client.go#L186 that logs every time a metrics request is successfully sent to TACS
2. Run a 10-minute stress test, same method as above
3. At end of test run, count how many metrics requests were successfully sent to TACS.

PR #2444, opened 2020-05-02
Steady state poll interval jitter

sparrc:
### Summary

When the agent is restarted, all tasks are added back into it's internal task manager all at once. If these are "steady state" tasks, then they will all hit the steady state poll interval at the same time, resulting in a docker DescribeContainer call for every container running on the instance.

This change adds a jitter to this so that we don't query docker for every container's state all at the same time.

### Testing

Was run on a container instance with 256 tasks, in addition to all unit/integ/func tests.

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -36,6 +36,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
 	utilsync "github.com/aws/amazon-ecs-agent/agent/utils/sync"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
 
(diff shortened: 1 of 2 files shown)


agent/engine/task_manager.go
```diff
@@ -47,12 +48,13 @@ const (
...
 	// and start processing the task or start another round of waiting
-	waitForPullCredentialsTimeout         = 1 * time.Minute
-	defaultTaskSteadyStatePollInterval    = 5 * time.Minute
-	transitionPollTime                    = 5 * time.Second
-	stoppedSentWaitInterval               = 30 * time.Second
-	maxStoppedWaitTimes                   = 72 * time.Hour / stoppedSentWaitInterval
-	taskUnableToTransitionToStoppedReason = "TaskStateError: Agent could not progress task's state to stopped"
+	waitForPullCredentialsTimeout            = 1 * time.Minute
+	defaultTaskSteadyStatePollInterval       = 5 * time.Minute
+	defaultTaskSteadyStatePollIntervalJitter = 30 * time.Second
```
yunhee-l, line 53: By definition, jitter should vary and not fixed?

sparrc: Yes, this gets passed into the retry.AddJitter function which does the jittering. We have other constants in our code with similar naming (ie https://github.com/aws/amazon-ecs-agent/blob/61d418ea2c535c3f7356e27eec23b05a0eb90319/agent/tcs/handler/handler.go#L42)

fierlion: I see your point about naming.  I think the naming here is consistent with the rest.  It represents a rand range (`randJitter = rand.Int63n(jitter.Nanoseconds())`) so this seems like a fine name for what it is.

PR #2449, opened 2020-05-08
Don't clear stats queue every interval

sparrc:
### Summary

This changes our stats queue behavior. Currently we build up docker stats in a buffer, and every 20s we aggregate them, publish them to ECS backend, and then clear the buffer.

This changes to never clearing the buffer (though we set a limit on the max size of the queue). The main reason for this is so that we can still publish metrics to ECS backend with only a single docker stat collected within one publishing interval (20s).

Currently if we only had one docker stat we would have nothing to compare it to and thus wouldnt be able to calculate CPU usage. With this change we can compare it to a stat from the previous ECS backend publish interval.

This means that users will still get metrics in cloudwatch even when docker is returning stats very slowly, or when they've configured polling metrics, and are only collecting one stat per interval (ie, ECS_POLLING_METRICS_WAIT_DURATION is set to anything greater than 10s).

Have also done some cleanup and changed the queue size to be calculated based on the size of the polling interval.

### Testing

All the regular test suites and tested manually on an instance with hundreds of tasks.

Metrics in cloudwatch were verified to be consistent before and after making this change.

Memory usage was tested because of changes to the queue size and clearing behavior. When streaming stats, there was a slight increase in memory usage but it was negligible (on the order of 30MiB on this PR, and 28 MiB on dev branch). When polling for stats there was no difference in memory usage.

### Licensing

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -142,6 +142,10 @@ const (
 	defaultCgroupCPUPeriod = 100 * time.Millisecond
 	maximumCgroupCPUPeriod = 100 * time.Millisecond
 	minimumCgroupCPUPeriod = 8 * time.Millisecond
+
+	// DefaultContainerMetricsPublishInterval is the default interval that we publish
+	// metrics to the ECS telemetry backend (TACS)
+	DefaultContainerMetricsPublishInterval = 20 * time.Second
 )
 
 const (
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"math/rand"
 	"strconv"
 	"strings"
 	"sync"
@@ -1338,7 +1337,6 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 		go func() {
 			defer cancelRequest()
 			defer close(statsC)
-			// ContainerStats outputs an io.ReadCloser and an OSType
 			stream := true
 			resp, err = client.ContainerStats(subCtx, id, stream)
 			if err != nil {
@@ -1353,7 +1351,6 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 			defer resp.Body.Close()
 			defer close(ch)
 
-			// Returns a *Decoder and takes in a readCloser
 			decoder := json.NewDecoder(resp.Body)
 			data := new(types.StatsJSON)
 			for err := decoder.Decode(data); err != io.EOF; err = decoder.Decode(data) {
@@ -1372,17 +1369,33 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 		}()
 	} else {
 		seelog.Infof("DockerGoClient: Starting to Poll for metrics for container %s", id)
-		//Sleep for a random period time up to the polling interval. This will help make containers ask for stats at different times
-		time.Sleep(time.Second * time.Duration(rand.Intn(int(dg.config.PollingMetricsWaitDuration.Seconds()))))
-
+		// firstStatC jitters the time at which containers ask for their first stat.
+		// We use this channel to 'seed' the queue with docker stats so that containers
+		// can publish metrics more quickly.
+		firstStatC := time.After(retry.AddJitter(time.Nanosecond, dg.config.PollingMetricsWaitDuration/2))
+		// sleeping here jitters the time at which the ticker is created, so that
+		// containers do not synchronize on calling the docker stats api.
+		// the max sleep is 80% of the polling interval so that we have a chance to
+		// get two stats in the first publishing interval.
+		time.Sleep(retry.AddJitter(time.Nanosecond, dg.config.PollingMetricsWaitDuration*8/10))
 		statPollTicker := time.NewTicker(dg.config.PollingMetricsWaitDuration)
 		go func() {
 			defer cancelRequest()
 			defer close(statsC)
 			defer statPollTicker.Stop()
 
-			for range statPollTicker.C {
-				// ContainerStats outputs an io.ReadCloser and an OSType
+			for {
+				// this select statement is waiting on either the stat polling ticker channel
+				// or the firstStat time.After channel to fire. firstStat will fire
+				// first and then afterwards we will always grab stats on the ticker.
+				select {
+				case _, ok := <-statPollTicker.C:
+					if !ok {
+						return
+					}
+				case <-firstStatC:
+				}
+
 				stream := false
 				resp, err = client.ContainerStats(subCtx, id, stream)
 				if err != nil {
@@ -1390,7 +1403,6 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 					return
 				}
 
-				// Returns a *Decoder and takes in a readCloser
 				decoder := json.NewDecoder(resp.Body)
 				data := new(types.StatsJSON)
 				err := decoder.Decode(data)
@@ -1400,7 +1412,6 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 				}
 
 				statsC <- data
-				data = new(types.StatsJSON)
 			}
 		}()
 	}
--- a/agent/stats/container.go
+++ b/agent/stats/container.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
 	"github.com/aws/amazon-ecs-agent/agent/stats/resolver"
@@ -26,16 +27,7 @@ import (
 	"github.com/cihub/seelog"
 )
 
-const (
-	// SleepBetweenUsageDataCollection is the sleep duration between collecting usage data for a container.
-	SleepBetweenUsageDataCollection = 1 * time.Second
-
-	// ContainerStatsBufferLength is the number of usage metrics stored in memory for a container. It is calculated as
-	// Number of usage metrics gathered in a second (1) * 60 * Time duration in minutes to store the data for (2)
-	ContainerStatsBufferLength = 120
-)
-
-func newStatsContainer(dockerID string, client dockerapi.DockerClient, resolver resolver.ContainerMetadataResolver) (*StatsContainer, error) {
+func newStatsContainer(dockerID string, client dockerapi.DockerClient, resolver resolver.ContainerMetadataResolver, cfg *config.Config) (*StatsContainer, error) {
 	dockerContainer, err := resolver.ResolveContainer(dockerID)
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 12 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1373,7 +1372,7 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 	} else {
 		seelog.Infof("DockerGoClient: Starting to Poll for metrics for container %s", id)
 		//Sleep for a random period time up to the polling interval. This will help make containers ask for stats at different times
-		time.Sleep(time.Second * time.Duration(rand.Intn(int(dg.config.PollingMetricsWaitDuration.Seconds()))))
+		time.Sleep(retry.AddJitter(time.Millisecond, dg.config.PollingMetricsWaitDuration))
```
sparrc, line 1375: This adds jitter using out internal util rather than a one-off rand.Intn call

agent/stats/queue.go
```diff
@@ -244,38 +243,6 @@ func getNetworkTxPackets(s *UsageStats) uint64 {
 	return uint64(0)
 }
 
-// GetRawUsageStats gets the array of most recent raw UsageStats, in descending
-// order of timestamps.
-func (queue *Queue) GetRawUsageStats(numStats int) ([]UsageStats, error) {
```
sparrc, line 249: This function was not being used anywhere, so deleting

agent/stats/queue.go
```diff
@@ -305,19 +272,6 @@ func getInt64WithOverflow(uintStat uint64) (int64, int64) {
 type getUsageFloatFunc func(*UsageStats) float64
 type getUsageIntFunc func(*UsageStats) uint64
 
-func (queue *Queue) resetThresholdElapsed(timeout time.Duration) bool {
```
sparrc, line 308: We don't reset on a threshold anymore so deleting

agent/stats/queue.go
```diff
@@ -305,19 +272,6 @@ func getInt64WithOverflow(uintStat uint64) (int64, int64) {
...
-	queue.lock.RLock()
-	defer queue.lock.RUnlock()
-	duration := time.Since(queue.lastResetTime)
-	return duration.Seconds() > timeout.Seconds()
-}
-
-func (queue *Queue) enoughDatapointsInBuffer() bool {
```
sparrc, line 315: this function is irrelevant because the length of buffer gets checked in each individual stats getting function.

agent/stats/queue_test.go
```diff
@@ -21,7 +21,6 @@ import (
 	"time"
```
sparrc, line 21: Most of the diff here is changing the tests to use the testify 'require' package rather than doing the comparisons ourselves.

agent/stats/types.go
```diff
@@ -70,6 +73,7 @@ type StatsContainer struct {
 	client            dockerapi.DockerClient
 	statsQueue        *Queue
 	resolver          resolver.ContainerMetadataResolver
+	config            *config.Config
```
sparrc, line 76: config.Config was added to this struct so that we have access to the default stats publishing interval at this scope.

agent/stats/engine.go
```diff
@@ -78,7 +78,7 @@ type DockerStatsEngine struct {
 	cluster                    string
 	containerInstanceArn       string
 	lock                       sync.RWMutex
-	disableMetrics             bool
```
sparrc, line 81: disableMetrics bool is removed because we can get this from config.Config now

agent/tcs/client/client.go
```diff
@@ -153,13 +153,6 @@ func (cs *clientServer) publishMetrics() {
 		return
 	}
 
-	// Publish metrics immediately after we connect and wait for ticks. This makes
-	// sure that there is no data loss when a scheduled metrics publishing fails
-	// due to a connection reset.
-	err := cs.publishMetricsOnce()
```
sparrc, line 159: This comment and function call no longer applies, this function does not get recalled on connection reset. Trying to publish metrics on startup won't ever work because there are no metrics in the queue yet.

agent/stats/queue.go
```diff
@@ -48,12 +46,13 @@ func NewQueue(maxSize int) *Queue {
 	}
 }
 
-// Reset resets the stats queue.
+// Reset sets all metrics in the queue to 'sent' status
```
yunhee-l, line 49: nit: should this method be renamed? "setting all metrics to sent" doesn't read like it's "resetting".

agent/stats/types.go
```diff
@@ -53,6 +54,8 @@ type UsageStats struct {
 	NetworkStats      *NetworkStats `json:"networkStats"`
 	Timestamp         time.Time     `json:"timestamp"`
 	cpuUsage          uint64
+	// sent indicates if the stat has been sent to TACS already.
+	sent bool
```
yunhee-l, line 58: Personally, I'd find it easier to understand if we had this explicitly set to false to start with, wherever UsageStats is instantiated.

agent/stats/queue.go
```diff
@@ -305,19 +272,6 @@ func getInt64WithOverflow(uintStat uint64) (int64, int64) {
...
-
-func (queue *Queue) enoughDatapointsInBuffer() bool {
-	queue.lock.RLock()
-	defer queue.lock.RUnlock()
-	return len(queue.buffer) >= minimumQueueDatapoints
-}
-
 // getCWStatsSet gets the stats set for either CPU or Memory based on the
 // function pointer.
 func (queue *Queue) getCWStatsSet(f getUsageFloatFunc) (*ecstcs.CWStatsSet, error) {
```
yunhee-l, line 277: f is not an informative name, getUsageFloat maybe?

agent/stats/queue_test.go
```diff
@@ -37,32 +36,33 @@ const (
 	predictableInt64Overflow = math.MaxInt64 / int64(2)
 )
 
+var now time.Time = time.Now()
+
 func getTimestamps() []time.Time {
 	return []time.Time{
-		parseNanoTime("2015-02-12T21:22:05.131117533Z"),
```
yunhee-l, line 42: is this change necessary now that you don't rely on relative timestamp but on sent flag?

agent/stats/queue.go
```diff
@@ -101,7 +100,7 @@ func (queue *Queue) add(rawStat *ContainerStats) {
 			stat.CPUUsagePerc = 100 * cpuUsageSinceLastStat / timeSinceLastStat
 		}
 
-		if queue.maxSize == queueLength {
+		if queueLength >= queue.maxSize {
```
yunhee-l, line 103: if we always remove first element when queue is full, queueLength will never be > maxSize, so why is this change (from == to >=) necessary?

agent/stats/queue_test.go
sparrc: it's not, but i thought it seemed more readable this way rather than having to grok the time difference between stats based on this 9-digit decimal.

agent/stats/queue.go
sparrc: it's just safer this way since we don't have any guarantees that nobody else will ever modify queue.maxSize

agent/stats/queue.go
sparrc: I didn't change this, did you want me to rename it?

agent/stats/types.go
sparrc: sure, will do

agent/stats/queue.go
sparrc: yeah i suppose from the outside it's still a 'reset' though, since it means that from thenceforth only new metrics will be considered valid for sending onwards, so from the perspective of a consumer of the queue, the behavior hasn't changed at all.

agent/stats/queue.go
yunhee-l: if you are making changes for below comment, might as well? In the code in this file, f can be either getUsageIntFunc or getUsageFloatFunc and that's confusing.

sparrc: ya sure i'll change it

sparrc: done

agent/stats/types.go
sparrc: done

agent/stats/queue.go
sparrc: I changed the comment to reflect what it's doing, what do you think?

yunhee-l: that comment def helps 👍

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1372,25 +1371,35 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
...
-			for range statPollTicker.C {
-				// ContainerStats outputs an io.ReadCloser and an OSType
+			for {
+				select {
+				case _, ok := <-statPollTicker.C:
+					if !ok {
+						return
+					}
+				case <-firstStat:
+				}
```
fierlion, line 1394: I'm not fully clear on what this is doing -- are we tossing out the first stat?

agent/stats/container.go
```diff
@@ -19,23 +19,19 @@ import (
...
 	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
 	"github.com/cihub/seelog"
 )
 
-const (
-	// SleepBetweenUsageDataCollection is the sleep duration between collecting usage data for a container.
-	SleepBetweenUsageDataCollection = 1 * time.Second
+// default size of stats queue, assumes 1s (streaming) collection interval
+// (seconds in one publish interval) / 1 (1s interval) * 4 (intervals)
+var queueSize int = int(config.DefaultContainerMetricsPublishInterval.Seconds()) / 1 * 4
```
fierlion, line 32: I realize that go will follow order of precedence and this will end up being metricsPublishInterval / 4

It's just odd to see the `/ 1` -- I'd prefer to see these be named, aka take what you've got in the comment above and just call these something like `publishIntervalDuration := 1` and `publishIntervalCount := 1`.

agent/dockerclient/dockerapi/docker_client.go
sparrc: No firstStat is a channel returned from time.After, this select statement is waiting on either the polling ticker channel or the firstStat channel to fire.

sparrc: I can rename to firstStatC if that makes it more clear?

agent/stats/container.go
sparrc: yeah this is mostly just written out so it's clear that it's following the same formula as in the StartStatsCollection function below. That being said, this doesn't actually need to be a package-level var for any reason. I will change it to a local variable in this function and I think that will make it more clear that this is for streaming stats only too.

agent/dockerclient/dockerapi/docker_client.go
fierlion: ok, so we ignore firings from time.After? 
I'd add a comment like:
`// this select statement is waiting on either the polling ticker channel or the firstStat channel (from time.After) to fire`

agent/stats/container.go
fierlion: local var is perfect.

agent/dockerclient/dockerapi/docker_client.go
sparrc: sure will do

agent/config/config.go
```diff
@@ -142,6 +142,10 @@ const (
 	defaultCgroupCPUPeriod = 100 * time.Millisecond
 	maximumCgroupCPUPeriod = 100 * time.Millisecond
 	minimumCgroupCPUPeriod = 8 * time.Millisecond
+
+	// DefaultContainerMetricsPublishInterval is the default interval that we publish
+	// metrics to the ECS telemetry backend (TACS)
+	DefaultContainerMetricsPublishInterval = 20 * time.Second
```
sharanyad, line 148: why is this present in config package? do we foresee customers setting this value?

sparrc: we don't but it seemed like the most logical place to put global constants, we have other constants in this package that are not controllable by the customer so there's precedence for it, whether right or wrong.

sparrc:
> does this handle agent restarts (losing previous stats) while calculating CPU usage?

Yes, restarts are essentially handled the same as if the agent was starting for the first time. We dont save the stats queue in the statefile so this is essentially unchanged from before the PR.

PR #2452, opened 2020-05-13
Change default docker metric gathering behavior

sparrc:
### Summary

1. Change default docker metric gathering behavior from streaming
metrics to polling.
2. Change the default polling interval to 10s (half of the TACS publishing
interval of 20s), so that every publish interval we have two docker
metrics (previously 15s).
3. Change the minimum polling interval to 5s to prevent customers from
configuring resource-intensive polling (previously 1s).
4. When a customer configures an interval below the minimum interval,
log a warning and set the interval to the minimum (previously set to default).
5. When a customer configures an interval above the maximum interval,
log a warning and set the interval to the maximum (previously set to default).

These changes are being made because we have found that docker streaming
stats consumes considerable resources from the agent, dockerd daemon, and
containerd daemon.

The graph below shows the improvement in cpu utilization on a single-instance cluster with an m5.large with 120 containers. On agent 1.39.0 with default settings our cluster utilization maxes out at 86% because agent/dockerd/containerd is using ~14% of the instance's cpu resources. With this change (1.40.0) we max out at 94.5%. This means we see a 60% reduction in resources consumed by ECS daemons and 10% improvement in overall cluster utilization.

**NOTE**: higher cpu utilization in these graphs is a good thing, as it means the customer's containers/tasks are utilizing more of the cluster's cpu, and not daemons required for ECS.

image

Improvement gets more dramatic with more containers:

image

and not as dramatic but still substantial with fewer:

image

### Testing

Agent's default stats gathering is changing from docker streaming stats to polling. This should not affect the metrics that customers ultimately see in cloudwatch, but it does affect how the agent gathers the underlying metrics from docker. This change was made for considerable performance gains. Customers with high CPU loads may see their cluster utilization increase; this is a good thing because it means the containers are utilizing more of the cluster, and agent/dockerd/containerd are utilizing less.

### Licensing


README.md
```diff
@@ -139,8 +139,8 @@ additional details on each available environment variable.
...
 | `ECS_UPDATES_ENABLED` | &lt;true &#124; false&gt; | Whether to exit for an updater to apply updates when requested. | false | false |
 | `ECS_DISABLE_METRICS`     | &lt;true &#124; false&gt;  | Whether to disable metrics gathering for tasks. | false | true |
-| `ECS_POLL_METRICS`     | &lt;true &#124; false&gt;  | Whether to poll or stream when gathering metrics for tasks. | false | false |
-| `ECS_POLLING_METRICS_WAIT_DURATION` | 30s | Time to wait to poll for new metrics for a task. Only used when ECS_POLL_METRICS is true  | 15s | 15s |
+| `ECS_POLL_METRICS`     | &lt;true &#124; false&gt;  | Whether to poll or stream when gathering metrics for tasks. WARNING: setting this to false on an instance with many containers can result in very high CPU utilization by the agent, dockerd, and containerd. | `true` | `true` |
```
shubham2892, line 142: May be add a note here stating that default value used to be "false" and is "true" now.

README.md
```diff
@@ -139,8 +139,8 @@ additional details on each available environment variable.
...
+| `ECS_POLLING_METRICS_WAIT_DURATION` | 10s | Time to wait between polling for metrics for a task. Not used when ECS_POLL_METRICS is false. Maximum value is 20s and minimum value is 5s. | 10s | 10s |
```
shubham2892, line 143: Also add a note here saying that if the values smaller than minimum value are set it will be defaulted to minimum value, same for the maximum value.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1369,56 +1369,56 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 		}()
```
sparrc, line 1369: The changes here are because functional tests caught that polling metrics was taking longer to populate the task stats API versus streaming stats.

So change the behavior to poll for docker stats immediately on startup. This is not ideal from a load perspective but is a necessity to avoid unintentional behavior changes.

PR #2457, opened 2020-05-20
Fix agent top-level context

sparrc:
When we create a new agent, we add a background context to the top-level object.
This context never exits, cancels, or times out.
There are some places in our code, however, that reference this context as if expecting it to exit cleanly.

this change fixes this by:
1. when creating the context, create it with a cancel function and add
this to the agent object.
2. pass this cancel function into the agent signal handler, so that the
context is cancelled when we receive a shutdown signal.
3. audit some of our async goroutines and other code that relied on the
agent context to make sure it was waiting/exiting properly.
4. pass agent context into some goroutines that were not previously
waiting on it, but should be.

agent logs on stop before this change:
```
level=info time=2020-05-20T18:49:49Z msg="Saving state!" module=state_manager.go
```

agent logs on stop after this change:
```
level=info time=2020-05-20T21:43:39Z msg="Agent received termination signal: terminated" module=termination_handler.go
level=info time=2020-05-20T21:43:39Z msg="Saving state!" module=state_manager.go
level=info time=2020-05-20T21:43:39Z msg="Exiting the engine event handler." module=handler.go
level=info time=2020-05-20T21:43:39Z msg="TaskHandler: Stopping periodic container state change submission ticker" module=task_handler.go
level=info time=2020-05-20T21:43:39Z msg="TCS Websocket connection closed for a valid reason" module=handler.go
level=info time=2020-05-20T21:43:39Z msg="TCS session exited cleanly." module=handler.go
level=info time=2020-05-20T21:43:39Z msg="Event stream DeregisterContainerInstance stopped listening..." module=eventstream.go
level=info time=2020-05-20T21:43:39Z msg="Event stream ContainerChange stopped listening..." module=eventstream.go
level=info time=2020-05-20T21:43:39Z msg="Stopping udev event handler" module=watcher_linux.go
level=info time=2020-05-20T21:43:39Z msg="ACS session exited cleanly." module=acs_handler.go
```

task and container managers will also exit when running tasks, resulting in logs like this:
```
level=info time=2020-05-26T22:08:11Z msg="Managed task [arn:aws:ecs:us-east-1:XXXX:task/XXXX]: agent task manager exiting." module=task_manager.go
level=info time=2020-05-26T22:08:11Z msg="Container [XXXX]: Stopping stats collection" module=container.go
```

### Testing

This was tested with the race detector while running 0, 50, and 250 tasks on a single instance.

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -194,6 +194,12 @@ func (acsSession *session) Start() error {
 			seelog.Debugf("Received connect to ACS message")
 			// Start a session with ACS
 			acsError := acsSession.startSessionOnce()
+			select {
+			case <-acsSession.ctx.Done():
+				// agent is shutting down, exiting cleanly
+				return nil
+			default:
+			}
 			// Session with ACS was stopped with some error, start processing the error
 			isInactiveInstance := isInactiveInstanceError(acsError)
 			if isInactiveInstance {
@@ -231,8 +237,8 @@ func (acsSession *session) Start() error {
 				}
 			}
 		case <-acsSession.ctx.Done():
-			seelog.Debugf("ACS session context cancelled")
-			return acsSession.ctx.Err()
+			// agent is shutting down, exiting cleanly
+			return nil
 		}
 
 	}
@@ -366,7 +372,7 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 		case <-acsSession.ctx.Done():
 			// Stop receiving and sending messages from and to ACS when
 			// the context received from the main function is canceled
-			seelog.Infof("ACS session context cancelled.")
+			seelog.Infof("ACS session exited cleanly.")
 			return acsSession.ctx.Err()
 		case err := <-serveErr:
 			// Stop receiving and sending messages from and to ACS when
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -233,10 +233,8 @@ func (engine *DockerTaskEngine) MarshalJSON() ([]byte, error) {
 // and operate normally.
 // This function must be called before any other function, except serializing and deserializing, can succeed without error.
 func (engine *DockerTaskEngine) Init(ctx context.Context) error {
-	// TODO, pass in a a context from main from background so that other things can stop us, not just the tests
 	derivedCtx, cancel := context.WithCancel(ctx)
 	engine.stopEngine = cancel
-
 	engine.ctx = derivedCtx
 
 	// Open the event stream before we sync state so that e.g. if a container
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -209,16 +209,17 @@ func (mtask *managedTask) overseeTask() {
 
 	// Main infinite loop. This is where we receive messages and dispatch work.
 	for {
-		select {
-		case <-mtask.ctx.Done():
-			seelog.Infof("Managed task [%s]: parent context cancelled, exit", mtask.Arn)
+		if mtask.shouldExit() {
 			return
-		default:
 		}
 
 		// If it's steadyState, just spin until we need to do work
 		for mtask.steadyState() {
 			mtask.waitSteady()
+		}
+
+		if mtask.shouldExit() {
+			return
 		}
 
 		if !mtask.GetKnownStatus().Terminal() {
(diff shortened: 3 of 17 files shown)


agent/engine/docker_image_manager_integ_test.go
```diff
@@ -18,7 +18,6 @@
 package engine
```
sparrc, line 18: this change is fixing logging on an unrelated flakey test

agent/engine/docker_task_engine.go
```diff
@@ -659,8 +657,7 @@ func (engine *DockerTaskEngine) handleDockerEvent(event dockerapi.DockerContaine
 
 	engine.tasksLock.RLock()
 	managedTask, ok := engine.managedTasks[task.Arn]
-	// hold the lock until the message is sent so we don't send on a closed channel
-	defer engine.tasksLock.RUnlock()
+	engine.tasksLock.RUnlock()
```
sparrc, line 660: this no longer needs to be deferred because we are no longer closing the channels

agent/engine/task_manager.go
```diff
@@ -1195,11 +1208,6 @@ func (mtask *managedTask) discardEvents() {
 		case <-mtask.acsMessages:
 		case <-mtask.resourceStateChangeEvent:
 		case <-mtask.ctx.Done():
```
sparrc, line 1210: It's OK to never close a channel in Golang. The only reason to close a channel is when a receiver is looking for a close signal. In this case, we only send messages to these channels so closing them can only result in a potential panic.

agent/sighandlers/termination_handler.go
```diff
@@ -39,23 +39,23 @@ const (
...
+	seelog.Infof("Agent received termination signal: %s", sig.String())
+	cancel()
 
 	err := FinalSave(saver, taskEngine)
 	if err != nil {
 		seelog.Criticalf("Error saving state before final shutdown: %v", err)
-		// Terminal because it's a sigterm; the user doesn't want it to restart
-		os.Exit(exitcodes.ExitTerminal)
 	}
-	os.Exit(exitcodes.ExitSuccess)
```
sparrc, line 58: os.Exit is now handled in the top-level main program only.

agent/sighandlers/termination_handler.go
```diff
@@ -39,23 +39,23 @@ const (
...
+	// when we receive a termination signal, first save the state, then
+	// cancel the agent's context so other goroutines can exit cleanly.
+	signalC := make(chan os.Signal, 2)
+	signal.Notify(signalC, os.Interrupt, syscall.SIGTERM)
 
-	sig := <-signalChannel
-	seelog.Debugf("Termination handler received termination signal: %s", sig.String())
+	sig := <-signalC
+	seelog.Infof("Agent received termination signal: %s", sig.String())
+	cancel()
```
sparrc, line 53: this is where the top-level context gets cancelled

agent/acs/handler/acs_handler.go
```diff
@@ -194,6 +194,13 @@ func (acsSession *session) Start() error {
 			seelog.Debugf("Received connect to ACS message")
 			// Start a session with ACS
 			acsError := acsSession.startSessionOnce()
+			select {
+			case <-acsSession.ctx.Done():
+				seelog.Debug("ACS session context cancelled")
```
mssrivas, line 199: nit : Should this be same log as below 'ACS session exited cleanly.' and the one other one in this file

agent/engine/engine_windows_integ_test.go
```diff
@@ -344,9 +344,9 @@ func TestPortForward(t *testing.T) {
 	assert.Equal(t, string(response), serverContent, "got response: "+string(response)+" instead of ", serverContent)
```
sparrc, line 344: these are unrelated flakey test failures, fixing a potential race condition

agent/engine/engine_unix_integ_test.go
```diff
@@ -49,7 +49,6 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
```
sparrc, line 49: unrelated to this change, changing these to 'require' from 'assert' because failures early in a test can cause the test to hang later in the test, or in the defer statement, which makes it harder to debug. Changing to require makes the test fail early.

sparrc:
closing in order to breakup into smaller PRs

PR #2462, opened 2020-05-28
Cancel agent's top-level context on exit

sparrc:
### Summary

Breaking out #2457 into smaller PRs.

This PR cancels agent's top-level context when agent receives a termination signal.

Instead of saving state and then doing a hard os.Exit on receiving a termination signal, we now cancel the context and save state, relying on the ACS handler to exit gracefully which ultimately exits the agent in the main program here: https://github.com/aws/amazon-ecs-agent/blob/master/agent/agent.go#L29

The benefit of this approach is that it allows defer statements and other cleanup tasks to run, when os.Exit is called no defer statements are run.

agent logs on stop before this change:
```
level=info time=2020-05-20T18:49:49Z msg="Saving state!" module=state_manager.go
```

agent logs on stop after this change (graceful exits from goroutines):
```
level=info time=2020-05-20T21:43:39Z msg="Agent received termination signal: terminated" module=termination_handler.go
level=info time=2020-05-20T21:43:39Z msg="Saving state!" module=state_manager.go
level=info time=2020-05-20T21:43:39Z msg="Exiting the engine event handler." module=handler.go
level=info time=2020-05-20T21:43:39Z msg="TaskHandler: Stopping periodic container state change submission ticker" module=task_handler.go
level=info time=2020-05-20T21:43:39Z msg="TCS Websocket connection closed for a valid reason" module=handler.go
level=info time=2020-05-20T21:43:39Z msg="TCS session exited cleanly." module=handler.go
level=info time=2020-05-20T21:43:39Z msg="Event stream DeregisterContainerInstance stopped listening..." module=eventstream.go
level=info time=2020-05-20T21:43:39Z msg="Event stream ContainerChange stopped listening..." module=eventstream.go
level=info time=2020-05-20T21:43:39Z msg="Stopping udev event handler" module=watcher_linux.go
level=info time=2020-05-20T21:43:39Z msg="ACS session exited cleanly." module=acs_handler.go
```

task and container managers will also exit when running tasks, resulting in logs like this:
```
level=info time=2020-05-26T22:08:11Z msg="Managed task [arn:aws:ecs:us-east-1:XXXX:task/XXXX]: agent task manager exiting." module=task_manager.go
level=info time=2020-05-26T22:08:11Z msg="Container [XXXX]: Stopping stats collection" module=container.go
```

Fix agent top-level context

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -194,6 +194,12 @@ func (acsSession *session) Start() error {
 			seelog.Debugf("Received connect to ACS message")
 			// Start a session with ACS
 			acsError := acsSession.startSessionOnce()
+			select {
+			case <-acsSession.ctx.Done():
+				// agent is shutting down, exiting cleanly
+				return nil
+			default:
+			}
 			// Session with ACS was stopped with some error, start processing the error
 			isInactiveInstance := isInactiveInstanceError(acsError)
 			if isInactiveInstance {
@@ -231,8 +237,8 @@ func (acsSession *session) Start() error {
 				}
 			}
 		case <-acsSession.ctx.Done():
-			seelog.Debugf("ACS session context cancelled")
-			return acsSession.ctx.Err()
+			// agent is shutting down, exiting cleanly
+			return nil
 		}
 
 	}
@@ -366,7 +372,7 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 		case <-acsSession.ctx.Done():
 			// Stop receiving and sending messages from and to ACS when
 			// the context received from the main function is canceled
-			seelog.Infof("ACS session context cancelled.")
+			seelog.Infof("ACS session exited cleanly.")
 			return acsSession.ctx.Err()
 		case err := <-serveErr:
 			// Stop receiving and sending messages from and to ACS when
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -93,6 +93,7 @@ type agent interface {
 // the newAgent() method
 type ecsAgent struct {
 	ctx                         context.Context
+	cancel                      context.CancelFunc
 	ec2MetadataClient           ec2.EC2MetadataClient
 	ec2Client                   ec2.Client
 	cfg                         *config.Config
@@ -115,11 +116,8 @@ type ecsAgent struct {
 }
 
 // newAgent returns a new ecsAgent object, but does not start anything
-func newAgent(
-	ctx context.Context,
-	blackholeEC2Metadata bool,
-	acceptInsecureCert *bool) (agent, error) {
-
+func newAgent(blackholeEC2Metadata bool, acceptInsecureCert *bool) (agent, error) {
+	ctx, cancel := context.WithCancel(context.Background())
 	ec2MetadataClient := ec2.NewEC2MetadataClient(nil)
 	if blackholeEC2Metadata {
 		ec2MetadataClient = ec2.NewBlackholeEC2MetadataClient()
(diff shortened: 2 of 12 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -194,6 +194,12 @@ func (acsSession *session) Start() error {
 			seelog.Debugf("Received connect to ACS message")
 			// Start a session with ACS
 			acsError := acsSession.startSessionOnce()
+			select {
+			case <-acsSession.ctx.Done():
```
sharanyad, line 198: why this is required? the below `acsSession.ctx.Done()` should be sufficient right?

agent/app/agent.go
```diff
@@ -730,8 +731,7 @@ func (agent *ecsAgent) startACSSession(
 		seelog.Criticalf("Unretriable error starting communicating with ACS: %v", err)
 		return exitcodes.ExitTerminal
 	}
-	seelog.Critical("ACS Session handler should never exit")
-	return exitcodes.ExitError
+	return exitcodes.ExitSuccess
```
sharanyad, line 734: if this was classified as unexpected exit, why are we changing the exit as success? 
init has logic to handle agent stops based on the exit codes. Could you verify if this change does not affect init's behavior? same for terminal handler below.

agent/engine/task_manager.go
```diff
@@ -360,7 +374,6 @@ func (mtask *managedTask) waitEvent(stopWaiting <-chan struct{}) bool {
 		mtask.handleResourceStateChange(resChange)
 		return false
 	case <-stopWaiting:
-		seelog.Infof("Managed task [%s]: no longer waiting", mtask.Arn)
```
sharanyad, line 363: why is this removed?

agent/app/agent.go
sparrc: since we are now cancelling the context, we expect ACS to exit when we receive a termination signal, so this is no longer an error situation. I have verified this doesn't affect init's behavior.

agent/acs/handler/acs_handler.go
sparrc: this is checking if the Done channel is closed, it passes through and continues if it is not closed via the `default:`

agent/engine/task_manager.go
sparrc: In the case that we are not exiting, this will get logged here instead: https://github.com/aws/amazon-ecs-agent/pull/2462/files#diff-20346f7af6202828de353f0137b6e43aR330

agent/acs/handler/acs_handler.go
sparrc: errr i misunderstood the question. When we receive acsError we go into a whole series of checks and retries that we don't need to do when we are just going to exit. Changing the logic around handling all of that would be a much larger refactor so I have added an exit check right here to short-circuit it.

agent/app/agent.go
sharanyad: Is it true that this case will never be hit anymore?
https://github.com/aws/amazon-ecs-init/blob/d0ee8ae2e02a0e17f1d248b6b85bb1ff808e75d3/ecs-init/engine/engine.go#L231
If yes, lets change this

sparrc: There are a few cases where we can still exit terminal, such as when the ACS session returns an error or if saving state fails.

sharanyad: cool, I see that terminal handler for state save failure is changed back to exit with error code.

sparrc: yes, removing that case was a mistake, it would be nice to have os.Exit handled in one place but would be a much larger refactor.

PR #2463, opened 2020-05-28
Pass agent context to subroutines

sparrc:
### Summary

Breaking out #2457 into smaller PRs.

This PR does not explicitly cancel the agent's context, but it passes it to more goroutines and fixes some of the waiting and concurrency logic. The routines that it passes to are:

1. spot instance draining poller
2. introspection http endpoint
3. task metadata http endpoint
4. engine event handler
5. tcs handler (already was passing but this change makes it wait and exit correctly when context is cancelled)

Fix agent top-level context

### Licensing

### DIFF
--- a/agent/handlers/task_server_setup.go
+++ b/agent/handlers/task_server_setup.go
@@ -14,6 +14,7 @@
 package handlers
 
 import (
+	"context"
 	"net/http"
 	"strconv"
 	"time"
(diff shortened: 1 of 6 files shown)


agent/handlers/task_server_setup.go
```diff
@@ -174,14 +176,22 @@ func ServeTaskHTTPEndpoint(credentialsManager credentials.Manager,
...
+		}
+	}()
+
 	for {
 		retry.RetryWithBackoff(retry.NewExponentialBackoff(time.Second, time.Minute, 0.2, 2), func() error {
-			// TODO, make this cancellable and use the passed in context;
-			err := server.ListenAndServe()
-			if err != nil {
-				seelog.Errorf("Error running http api: %v", err)
+			if err := server.ListenAndServe(); err != http.ErrServerClosed {
```
fierlion, line 189: `if err := s.ListenAndServe(); err != nil && err != http.ErrServerClosed {`
we want to filter out `http.ErrServerClosed` but still want to be sure we actually have an error.
(I think `nil != http.ErrServerClosed`)

fierlion: // always returns error. ErrServerClosed on graceful close
I see.

PR #2465, opened 2020-05-29
Remove unused code, add staticcheck to TravisCI

sparrc:
### Summary

This PR mostly removes unused code found using the [staticcheck](https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck) tool (plus a couple other static analysis fixups). It also adds this tool to our static-check make target so that it runs on every PR in Travis.

There will be PRs following up that will fix more static analysis issues and adding these checks to CI.

### Testing

New tests cover the changes:

staticcheck tool was added to our Travis CI static-check make target. Currently only running the unused code check (U1000).

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -277,6 +277,9 @@ importcheck:
 
 .PHONY: static-check
 static-check: gocyclo govet importcheck
+	# check for unused code using staticcheck binary
+	# https://github.com/dominikh/go-tools/tree/master/cmd/staticcheck
+	staticcheck -tests=false -checks "U1000" ./agent/...
 
 .PHONY: goimports
 goimports:
@@ -287,6 +290,7 @@ goimports:
 	go get github.com/golang/mock/mockgen
 	go get golang.org/x/tools/cmd/goimports
 	go get github.com/fzipp/gocyclo
+	go get honnef.co/go/tools/cmd/staticcheck
 	touch .get-deps-stamp
 
 get-deps: .get-deps-stamp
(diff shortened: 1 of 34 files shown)


Makefile
```diff
@@ -263,6 +263,12 @@ gocyclo:
 	# Run gocyclo over all .go files
 	gocyclo -over 15 ${GOFILES}
 
+.PHONY: staticcheck
```
yunhee-l, line 266: https://github.com/aws/amazon-ecs-agent/blob/master/Makefile#L330 we have static-check already, can we name this differently? unused-check?

agent/stats/utils_test.go
```diff
@@ -36,20 +36,6 @@ const (
...
-		t.Error("Error incorrectly reported as network stats error")
-	}
-
-	isNetStatsErr = isNetworkStatsError(fmt.Errorf("open /sys/class/net/veth2f5f3e4/statistics/tx_bytes: no such file or directory"))
-	if !isNetStatsErr {
-		// Expect this to be a net stats error
-		t.Error("Error incorrectly reported as non network stats error")
-	}
-}
-
```
shubham2892, line 52: Why is this test removed?

sparrc: This was removed because the function it's testing was removed (isNetworkStatsError)

Makefile
sparrc: I prefer staticcheck because it's the name of the tool we're using (https://github.com/dominikh/go-tools). We can't use unused-check because staticcheck does other checks that we will be adding soon.

If you'd prefer maybe I could do `go-tools-staticcheck` ?

sparrc: I put it as a command under the static-check target

PR #2475, opened 2020-06-08
Instance Health: error message handling

sparrc:
### Summary

1. Changes error message handling to accumulate errors that occurred in an interval (but also deduplicate). This is so we can see all unique errors that happened when there were multiple within an interval.
2. Changes the error message getter function to reset the message so that we don't see error messages from a previous interval.
3. Set a minimum sample count (10) needed for clearing and publishing instance health metrics. This is so that if we have a period of sparse metrics, we will eventually accumulate enough samples to make a valid metric.
4. Reduce publishing interval from 3 to 1 minute.

### Licensing

### DIFF
--- a/agent/metrics/instancehealth/instancehealth.go
+++ b/agent/metrics/instancehealth/instancehealth.go
@@ -1,7 +1,8 @@
 package instancehealth
 
 import (
-	"fmt"
+	"strings"
+	"sync"
 	"sync/atomic"
 
 	"github.com/aws/amazon-ecs-agent/agent/api/errors"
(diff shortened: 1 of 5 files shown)


agent/metrics/instancehealth/instancehealth.go
```diff
@@ -26,16 +30,28 @@ func (gc *genericMetric) IncrementCallCount() {
 
 func (gc *genericMetric) RecordError(errorMessage errors.NamedError) {
 	atomic.AddInt64(&gc.errors, 1)
-	gc.errorMessage.Store(errorMessage.ErrorName() + ": " + errorMessage.Error())
+	atomic.AddInt64(&gc.calls, 1)
```
mythri-garaga, line 33: What's the reason we update the call count again here?

sparrc: That was me misreading the interface comment...I will remove this

PR #2479, opened 2020-06-10
Calculate per-sec network metrics for RXBytes and TxBytes

ubhattacharjya:
### Summary

Calculate RxBytes/sec and TxBytes/sec metrics for bridge network mode.

### Implementation details

Added the fields RxBytesPerSecond and TxBytesPerSecond to the existing networkStats struct. The above mentioned network metrics for last stat in the queue will be stored in memory and we will find the difference in value of currentStats and lastStats and divide it with the timeDuration between the 2 stats in seconds.

### Testing

Started a task which creates network traffic and saw that the calculation was happening as expected.
Added unit test for the same.

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/stats/queue.go
+++ b/agent/stats/queue.go
@@ -28,6 +28,7 @@ const (
 	// BytesInMiB is the number of bytes in a MebiByte.
 	BytesInMiB              = 1024 * 1024
 	MaxCPUUsagePerc float32 = 1024 * 1024
+	NanoSecToSec    float32 = 1000000000
 )
 
 // Queue abstracts a queue using UsageStats slice.
@@ -98,9 +99,21 @@ func (queue *Queue) add(rawStat *ContainerStats) {
 			// if we got a duplicate timestamp, set cpu percentage to the same value as the previous stat
 			seelog.Errorf("Received a docker stat object with duplicate timestamp")
 			stat.CPUUsagePerc = lastStat.CPUUsagePerc
+			if stat.NetworkStats != nil {
+				stat.NetworkStats.RxBytesPerSecond = lastStat.NetworkStats.RxBytesPerSecond
+				stat.NetworkStats.TxBytesPerSecond = lastStat.NetworkStats.TxBytesPerSecond
+			}
 		} else {
 			cpuUsageSinceLastStat := float32(rawStat.cpuUsage - lastStat.cpuUsage)
 			stat.CPUUsagePerc = 100 * cpuUsageSinceLastStat / timeSinceLastStat
+
+			//calculate per second Network metrics
+			if stat.NetworkStats != nil {
+				rxBytesSinceLastStat := float32(stat.NetworkStats.RxBytes - lastStat.NetworkStats.RxBytes)
+				txBytesSinceLastStat := float32(stat.NetworkStats.TxBytes - lastStat.NetworkStats.TxBytes)
+				stat.NetworkStats.RxBytesPerSecond = NanoSecToSec * (rxBytesSinceLastStat / timeSinceLastStat)
+				stat.NetworkStats.TxBytesPerSecond = NanoSecToSec * (txBytesSinceLastStat / timeSinceLastStat)
+			}
 		}
 
 		if queueLength >= queue.maxSize {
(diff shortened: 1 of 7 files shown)


agent/stats/queue.go
```diff
@@ -103,6 +103,13 @@ func (queue *Queue) add(rawStat *ContainerStats) {
 			stat.CPUUsagePerc = 100 * cpuUsageSinceLastStat / timeSinceLastStat
 		}
 
+		//calculate per second Network metrics
+		timeSinceLastStatInSeconds := uint64(rawStat.timestamp.Sub(lastStat.Timestamp).Seconds())
```
sparrc, line 107: this is going to be a problem for streaming stats, which sometimes might return stats that are less than 1 second apart from each other. I think it would be better to determine the bytes per millisecond and then divide by 1000 to get the per-second average.

sparrc: this would also mean you dont have to change the timestamps of the unit tests (which i think should stay as they are, since we need to support metrics arriving at less than 1-second intervals).

shubham2892: > this is going to be a problem for streaming stats

For AWSVPC mode network stats we are only polling stats.

shubham2892: Yes it makes sense to reuse `timeSinceLastStat` on line 96.

agent/stats/queue.go
```diff
@@ -103,6 +103,13 @@ func (queue *Queue) add(rawStat *ContainerStats) {
...
+		if stat.NetworkStats != nil && timeSinceLastStatInSeconds > 0 {
+			stat.NetworkStats.RxBytesPerSecond = uint64(((stat.NetworkStats.RxBytes - lastStat.NetworkStats.RxBytes) / timeSinceLastStatInSeconds))
+			stat.NetworkStats.TxBytesPerSecond = uint64(((stat.NetworkStats.TxBytes - lastStat.NetworkStats.TxBytes) / timeSinceLastStatInSeconds))
```
shubham2892, line 110: Will be sent for only bridge network mode? If not, we should add a check to ensure that this is only for bridge network mode task.

agent/stats/queue_test.go
```diff
@@ -314,6 +314,18 @@ func validateNetStatsSet(t *testing.T, netStats *ecstcs.NetworkStatsSet, queueLe
 	assert.Equal(t, int64(queueLen), *netStats.TxErrors.SampleCount, "incorrect TxErrors sampleCount")
 	assert.Equal(t, int64(0), *netStats.TxErrors.Sum, "incorrect TxErrors sum")
 	assert.Equal(t, int64(0), *netStats.TxErrors.OverflowSum, "incorrect TxErrors overlfowSum")
+
+	assert.NotNil(t, *netStats.RxBytesPerSecond, "incorrect RxBytesPerSecond set")
+	assert.NotEqual(t, math.MaxFloat64, *netStats.RxBytesPerSecond.Min, "incorrect RxBytesPerSecond min")
```
sparrc, line 319: why are these tests not asserting the actual value that we expect for RxBytesPerSecond and TxBytesPerSecond?

sparrc:
There are some corner cases based on the fact that this metric depends on two different sets of stats from docker. For CPU percentage we have additional unit tests defined here: https://github.com/aws/amazon-ecs-agent/blob/master/agent/stats/queue_test.go#L395-L512 to cover some of these cases, I think we should have similar coverage for these metrics as well, such as:

1. what happens to these stats when there's only been 2 metrics from docker?
2. what happens to all of the network stats when there's only been 2 metrics from docker?
3. what happens to these when two metrics have the same timestamp?

agent/stats/queue.go
ubhattacharjya: Fixed it. Changed the unit of the per sec metrics to float32 which will help us to calculate the metric for even a small time duration

agent/stats/queue.go
ubhattacharjya: That check is being done here already https://github.com/aws/amazon-ecs-agent/blob/master/agent/stats/engine.go#L616 . That is why I am not doing it here

agent/stats/queue_test.go
ubhattacharjya: The actual float values are quite messy to assert. That is why just made sure that they are not empty and not not nil.

ubhattacharjya:
> There are some corner cases based on the fact that this metric depends on two different sets of stats from docker. For CPU percentage we have additional unit tests defined here: https://github.com/aws/amazon-ecs-agent/blob/master/agent/stats/queue_test.go#L395-L512 to cover some of these cases, I think we should have similar coverage for these metrics as well, such as:
> 
> 1. what happens to these stats when there's only been 2 metrics from docker?
> 2. what happens to all of the network stats when there's only been 2 metrics from docker?
> 3. what happens to these when two metrics have the same timestamp?

This is a really good point. Added unit tests for each scenario.
1. If there are 2 metrics that should be fine. It needs atleast 2 metrics.
2. If there are 2 metrics the entire network stats metric should be fine. It needs atleast 2 metrics.
3. If there are 2 metrics only and they are both at the same timestamp, then it is a problem and it should throw error. But if there are more than 2 metrics and among them only 2 are same, it should be fine.
4. If there are less than 2 metrics, then it should throw error

agent/stats/queue_test.go
sparrc: you can use assert.InDelta to assert float values are within some delta of each other: https://pkg.go.dev/github.com/stretchr/testify/assert?tab=doc#InDelta

ubhattacharjya: Fixed it

PR #2482, opened 2020-06-15
Add AWSVPC metrics

shubham2892:
### Summary
This PR adds the support to collect network metrics for tasks running in AWSVPC network mode. These new metrics can be accessed using task metadata endpoint. Customers who have enabled ECS container insights can view these metrics for their ECS tasks automatically in their container insights dashboard.  

### Implementation details
The implementation is similar to how metrics are collected for task running in bridge network mode. 

For bridge network mode, container level metrics are pulled from Docker in the method `processStatsStream()` in stats/container.go . 

For AWSVPC network mode, metrics will be pulled from host in `processStatsStream()` in stats/tasks. All the containers in a task running in AWSVPC mode uses the same network interface, so the metrics does not need to be pulled for every container in the task but only once per task. 

Note: Changes to files in vendor package are a result of running `dep ensure` which was required after adding a new package -- github.com/containernetworking/plugins.

### Testing

### Licensing


agent/stats/utils_unix.go
```diff
@@ -15,20 +15,12 @@
...
 )
 
 // dockerStatsToContainerStats returns a new object of the ContainerStats object from docker stats.
 func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats, error) {
-	// The length of PercpuUsage represents the number of cores in an instance.
-	if len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0 || numCores == uint64(0) {
-		seelog.Debug("Invalid container statistics reported, no cpu core usage reported")
-		return nil, fmt.Errorf("Invalid container statistics reported, no cpu core usage reported")
-	}
-
```
shubham2892, line 31: This check is moved to stats/container.go.

agent/ecscni/mocks_libcni/libcni_mocks.go
```diff
@@ -74,69 +73,85 @@ func (m *MockCNI) AddNetworkList(arg0 context.Context, arg1 *libcni.NetworkConfi
...
+	ret1, _ := ret[1].(*libcni.RuntimeConf)
+	ret2, _ := ret[2].(error)
+	return ret0, ret1, ret2
+}
+
+// GetNetworkCachedConfig indicates an expected call of GetNetworkCachedConfig.
+func (mr *MockCNIMockRecorder) GetNetworkCachedConfig(arg0, arg1 interface{}) *gomock.Call {
+	mr.mock.ctrl.T.Helper()
+	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "GetNetworkCachedConfig", reflect.TypeOf((*MockCNI)(nil).GetNetworkCachedConfig), arg0, arg1)
+}
```
shubham2892, line 152: This method has been added to the library libcni.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1468,3 +1488,85 @@ func (dg *dockerGoClient) loadImage(ctx context.Context, reader io.Reader) error
...
+		RxDropped: netLinkStats.RxDropped / numberOfContainers,
+		TxBytes:   netLinkStats.TxBytes / numberOfContainers,
+		TxPackets: netLinkStats.TxPackets / numberOfContainers,
+		TxErrors:  netLinkStats.TxErrors / numberOfContainers,
+		TxDropped: netLinkStats.TxDropped / numberOfContainers,
+	}
+	return networkStats
+}
+
+func (dg *dockerGoClient) GetAWSVPCNetworkStats(deviceList []string, containerPID string,
```
fenxiong, line 1537: adding GetAWSVPCNetworkStats and PopulateNIDeviceList as new methods of docker client interface doesn't seem suitable to me because those stats are not related to docker. seems like they can just be helper function in agent/stats/task.go?

agent/nswrapper/ns.go
```diff
@@ -0,0 +1,25 @@
+package nswrapper
```
fenxiong, line 1: this can be under utils/nswrapper to be consistent with few other wrappers

agent/stats/engine.go
```diff
@@ -229,58 +258,110 @@ func (engine *DockerStatsEngine) removeAll() {
...
+
 	watchStatsContainer := false
+	var statsTaskContainer *StatsTask
 	if !engine.config.DisableMetrics {
 		// Adding container to the map for collecting stats
-		watchStatsContainer = engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, engine.containerMetricsMapUnsafe)
+		watchStatsContainer = engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer,
+			engine.containerMetricsMapUnsafe)
+
+		if task.IsNetworkModeAWSVPC() {
```
ubhattacharjya, line 297: Can this be part of another method as it is becoming huge otherwise?

agent/stats/types.go
```diff
@@ -76,6 +85,15 @@ type StatsContainer struct {
 	config            *config.Config
 }
 
+type StatsTask struct {
+	StatsQueue   *Queue
+	TaskMetadata *TaskMetadata
+	Ctx          context.Context
+	Cancel       context.CancelFunc
+	Resolver     resolver.ContainerMetadataResolver
+	client            dockerapi.DockerClient
```
ubhattacharjya, line 94: nit: formatting

agent/stats/utils_unix_test.go
```diff
@@ -25,17 +25,6 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestDockerStatsToContainerStatsEmptyCpuUsageGeneratesError(t *testing.T) {
```
ubhattacharjya, line 28: Is this test not needed now?

agent/stats/task.go
```diff
@@ -0,0 +1,134 @@
...
+func (taskStat *StatsTask) processStatsStream() error {
+	taskArn := taskStat.TaskMetadata.TaskArn
+	if len(taskStat.TaskMetadata.DeviceName) == 0 {
+		var err error
+		taskStat.TaskMetadata.DeviceName, err = taskStat.client.PopulateNIDeviceList(taskStat.TaskMetadata.ContainerPID)
+		if err != nil {
+			return err
+		}
+	}
+	dockerStats, errC := taskStat.client.GetAWSVPCNetworkStats(taskStat.TaskMetadata.DeviceName,
```
ubhattacharjya, line 100: Should this be called dockerStats as it is not really populated by docker?

agent/dockerclient/dockerapi/docker_client.go
shubham2892: I agree that these methods are not related to docker, I added them here for two reasons -- 

- Makes it easy to test. Because they are part of docker client interface it's easy to mock these function similar to how it is being done for 'stats' method -- https://github.com/aws/amazon-ecs-agent/blob/master/agent/dockerclient/dockerapi/docker_client.go#L167

- task.go is similar in structure to container.go. Container.go calls `stats` function to get the stats, it does not care about the implementation, going with the same pattern, task.go is calling `GetAWSVPCNetworkStats` and `PopulateNIDeviceList`. 

If it helps I can add more comments to these functions to make it clear what they are doing. Thoughts?

agent/nswrapper/ns.go
shubham2892: Will move.

agent/stats/engine.go
shubham2892: Will make it a separate method.

agent/stats/utils_unix_test.go
shubham2892: I have moved this check out of utils/unix.go to container.go

agent/stats/task.go
shubham2892: Will rename this to a more appropriate one.

agent/dockerclient/dockerapi/docker_client.go
fenxiong: for testing, you can make the nswrapper interface a field of the stats engine and a field of StatsTask, and then pass it from engine to StatsTask, and use mock for that field in test (similar to the way how docker client is mocked). for second point, IMO they don't have to be similar if the source of stats are essentially different. container stats are from a docker api call so it makes sense to use the docker client, but those network stats are from netlink.

agent/stats/task.go
```diff
@@ -0,0 +1,241 @@
...
+	// linkTypeDevice defines the string that's expected to be the output of
+	// netlink.Link.Type() method for netlink.Device type.
+	linkTypeDevice = "device"
+	linkTypeVlan   = "vlan"
+	// encapTypeLoopback defines the string that's set for the link.Attrs.EncapType
+	// field for localhost devices. The EncapType field defines the link
+	// encapsulation method. For localhost, it's set to "loopback".
+	encapTypeLoopback = "loopback"
+)
+type TaskStatsInterface interface {
```
fenxiong, line 45: IMO this interface is not needed and you can make netlinkClient and nsClient as fields of StatsTask and mock them in test.  GetAWSVPCNetworkStats and PopulateNIDeviceList can just be private methods of StatsTask. this way both TaskStatsInterface and TaskStatsStruct are not needed and you don't need mock for TaskStatsInterface either, so seems cleaner

agent/stats/engine.go
```diff
@@ -681,5 +771,19 @@ func (engine *DockerStatsEngine) ContainerDockerStats(taskARN string, containerI
...
+	// Insert network stats in container stats
+	taskStats, ok := taskToTaskStats[taskARN]
+	if ok {
+		task, _ := engine.resolver.ResolveTaskByARN(taskARN)
+		if task.IsNetworkModeAWSVPC() {
+			taskNetworkStats := taskStats.StatsQueue.GetLastStat().Networks
+			containerStats.Networks = taskNetworkStats
+		}
+	} else {
+		seelog.Warnf("Network stats not found for container %s", containerID)
```
fenxiong, line 785: i think we should skip this warning when task is not in awsvpc network mode, since it is expected that only awsvpc mode has it

agent/stats/engine.go
```diff
@@ -228,59 +257,109 @@ func (engine *DockerStatsEngine) removeAll() {
 	}
 }
 
+func (engine *DockerStatsEngine) getStatsTaskContainer(task *apitask.Task, dockerContainerName string) {
```
fenxiong, line 260: to align with addToStatsContainerMapUnsafe, addToStatsContainerMapUnsafe is probably a better name

fenxiong: also, i think we should just inspect the pause container and use its pid instead of arbitrary one, since pause container is guaranteed to be running when task is running while other container can be non-essential and stop before the task stops

agent/stats/engine.go
```diff
@@ -621,9 +700,19 @@ func (engine *DockerStatsEngine) taskContainerMetricsUnsafe(taskArn string) ([]*
 				} else {
 					containerMetric.NetworkStatsSet = networkStatsSet
 				}
+			} else if task.IsNetworkModeAWSVPC() {
+				taskStatsMap, taskExistsInTaskStats := engine.taskToTaskStats[taskArn]
+				if !taskExistsInTaskStats {
+					return nil, fmt.Errorf("Task not found")
```
fenxiong, line 706: error message starts with lower case

agent/utils/mocks/nswrapper_mocks.go
```diff
@@ -0,0 +1,63 @@
+// Code generated by MockGen. DO NOT EDIT.
```
fenxiong, line 1: missing copyright header. i think you should have something similar to https://github.com/aws/amazon-ecs-agent/blob/master/agent/eni/netlinkwrapper/generate_mocks_linux.go under agent/utils/nswrapper so you can generate the mock with `make generate` (and the mock will have the header like other generated mock)

agent/stats/engine.go
```diff
@@ -228,59 +257,109 @@ func (engine *DockerStatsEngine) removeAll() {
 	}
 }
 
+func (engine *DockerStatsEngine) getStatsTaskContainer(task *apitask.Task, dockerContainerName string) {
+	var statsTaskContainer *StatsTask
+	if task.IsNetworkModeAWSVPC() {
+		numberOfContainers := len(task.Containers)
```
fenxiong, line 263: should we exclude pause container from the calculation?

agent/stats/engine.go
```diff
@@ -102,6 +105,18 @@ func (resolver *DockerContainerMetadataResolver) ResolveTask(dockerID string) (*
 	return task, nil
 }
 
+func (resolver *DockerContainerMetadataResolver) ResolveTaskByARN(taskArn string) (*apitask.Task, error) {
+	if resolver.dockerTaskEngine == nil {
+		return nil, fmt.Errorf("Docker task engine uninitialized")
```
fenxiong, line 110: (probably from copy paste but) error message start with lowercase, same for line 114

agent/stats/task.go
```diff
@@ -0,0 +1,241 @@
...
+	for {
+		select {
+		case <-taskStat.Ctx.Done():
+			return nil
+		case err := <-errC:
+			seelog.Warnf("Error encountered processing metrics stream from host, this may affect "+
+				"cloudwatch metric accuracy: %s", err)
+			returnError = true
+		case rawStat, ok := <-awsvpcNetworkStats:
+			seelog.Info("got stats")
```
fenxiong, line 142: i'd suggest not logging thing for every stats (when there's no error) since it can be noisy

agent/stats/task.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (taskstatsclient *TaskStatsStruct) GetAWSVPCNetworkStats(deviceList []string, containerPID string,
+	numberOfContainers int) (<-chan *types.StatsJSON, <-chan error) {
+
+	errC := make(chan error)
+	statsC := make(chan *dockerstats.StatsJSON)
+
+	go func() {
+		time.Sleep(time.Second * time.Duration(rand.Intn(int(config.DefaultPollingMetricsWaitDuration))))
```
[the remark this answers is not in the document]
shubham2892: Yes will change.

agent/stats/engine.go
shubham2892: Yes that's a great point, will do.

fenxiong:
approved assuming tests will be fixed separately

agent/stats/utils_unix_test.go
ubhattacharjya: ok, are you also moving the test to container_test.go ?

shubham2892: moved the check back to this as a separate function and added the test back.

PR #2485, opened 2020-06-17
Implement task/container related methods for boltdb data client

fenxiong:
### Summary

Implemented task/container related methods for boltdb data client.

### Implementation details

* data: implemented SaveContainer, SaveDockerContainer, DelContainer, GetContainers, SaveTask, DelTask, GetTasks. Added a few helper functions to cover common boltdb interaction.
* api/container: added a new field TaskARN to allow easier generation of key when saving a container to db. This field will be populated with correct value in PostUnmarshalTask in a later code change (when actually using the interface).
* utils: added a helper function to get task id from task arn.

### Testing

Unit tests added.

N/A

### Licensing

### DIFF
--- a/agent/data/container_client.go
+++ b/agent/data/container_client.go
@@ -14,25 +14,85 @@
 package data
 
 import (
+	"encoding/json"
+
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+
+	"github.com/boltdb/bolt"
+	"github.com/pkg/errors"
 )
 
-func (c *client) SaveContainer(container *apicontainer.Container) error {
-	// TODO: implementation
-	return nil
-}
-
+// SaveDockerContainer saves a docker container to the container bucket.
 func (c *client) SaveDockerContainer(container *apicontainer.DockerContainer) error {
-	// TODO: implementation
-	return nil
+	id, err := getContainerID(container.Container)
+	if err != nil {
+		return errors.Wrap(err, "failed to generate database id")
+	}
+	return c.db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(containersBucketName))
+		return putObject(b, id, container)
+	})
 }
 
+// SaveContainer saves an apicontainer.Container to the container bucket. If a corresponding
+// apicontainer.DockerContainer exists, this updates the Container part of it; otherwise, a new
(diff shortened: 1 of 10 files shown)


agent/data/container_client.go
```diff
@@ -14,25 +14,70 @@
...
+
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+
+	"github.com/boltdb/bolt"
 )
 
+// SaveDockerContainer saves a docker container to the container bucket.
+func (c *client) SaveDockerContainer(container *apicontainer.DockerContainer) error {
+	id := utils.GetTaskID(container.Container.TaskARN) + "-" + container.Container.Name
```
shubham2892, line 27: Can we have a function for generating this id as this is used at multiple places,  this function can take taskArn and containerName as parameter and return the id string. 

Also, will it be a problem if a task has more than one container will same container name?

agent/data/helpers.go
```diff
@@ -0,0 +1,67 @@
...
+		return errors.Wrapf(err, "failed to insert object with key %q", key)
+	}
+
+	return nil
+}
+
+func getObject(bucket *bolt.Bucket, id string, out interface{}) error {
+	data := bucket.Get([]byte(id))
+	if data == nil {
+		return ErrNotFound
```
shubham2892, line 45: Add more details to this error, like for which bucket, id this error was raised?

agent/utils/utils.go
```diff
@@ -205,3 +205,9 @@ func SearchStrInDir(dir, filePrefix, content string) error {
 
 	return nil
 }
+
+// GetTaskID retrieves the task ID from task ARN.
+func GetTaskID(taskARN string) string {
+	fields := strings.Split(taskARN, "/")
+	return fields[len(fields)-1]
+}
```
shubham2892, line 213: Can we add a validation here to ensure that taskARN is in correct formatting, the most basic one I can think of is len(fields) is of at least length 2.

fenxiong: yeah, updated with some checks

agent/data/helpers.go
fenxiong: makes sense, added

agent/data/container_client.go
fenxiong: added a function. for second point, i don't think containers in the same task can have same name

PR #2487, opened 2020-06-19
Bug fix:Env File Feature would not accept = in value

ubhattacharjya:
### Summary

This pull request fixes a bug where env file feature would not accept "=", which is the delimiter in the values of a env var.
VARIABLE=abc=h would get truncated to VARIABLE=abc

### Implementation details

Changed the method strings.Split(line, "=") to strings.SplitN(line, "=", 2)
### Testing

 Tested that the env var is correctly getting its value after this change

New tests cover the changes: 

### Licensing


ubhattacharjya:
> change lgtm, could you add a test case for this?

Added test.

PR #2491, opened 2020-06-24
Save task and container data to boltdb

fenxiong:
### Summary

Save task and container data to boltdb with the boltdb client. Existing usage of state manager is not removed and will be removed in a separate change to reduce the amount of changes made at once.

Changes in this PR are intended to cover all the scenarios where we want to save task and container.

### Implementation details

Changes are made in following packages:
* `acs/handler`: save task to boltdb after adding the task to task engine;
* `api/task`: populate task arn to container in PostUnmarshalTask. this is used to generate database key when saving the container;
* `app/agent`: initialize boltdb data client and pass it to task engine, acs handler and event handler;
* `data`: added a no-op client which is used in testing and when ECS_CHECKPOINT is set to false;
* `engine`:
  - added a file data.go which covers interaction with boltdb;  - added a file data.go which covers interaction with boltdb;
  - task engine: remove task and containers data from database when cleaning up the task; added a method SetDataClient to set the client similar to how the state manager was set;
  - task manager: save task to boltdb when its desired/known status changes and when resource known status changes; save container to boltdb when its desired/known status changes and when updating its metadata. these are done in handleDesiredStatusChange, handleContainerChange and handleResourceStateChange in task_manager.go;
* `eventhandler`: save task/container in boltdb after updating their sent status.

### Testing

Unit tests added. Manually tested by running a task and checked that the task and container are saved in the db.

N/A

### Licensing

### DIFF
--- a/agent/engine/data.go
+++ b/agent/engine/data.go
@@ -1,0 +1,68 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package engine
+
+import (
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/agent/data"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+
+	"github.com/cihub/seelog"
+)
+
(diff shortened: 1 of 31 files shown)


agent/engine/data.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/data"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+
+	"github.com/cihub/seelog"
+)
+
+func (engine *DockerTaskEngine) saveTaskData(task *apitask.Task) {
+	err := engine.dataClient.SaveTask(task)
+	if err != nil {
+		seelog.Errorf("Failed to save data for task %s: %v", task.Arn, err)
```
yunhee-l, line 28: more of general question - wondering if failure to save should be critical? if save fails, do we just wait for the next time save is called?

fenxiong: > if save fails, do we just wait for the next time save is called?

yes, this is same as the existing behavior. currently when saving with state manager, the error is also not treated as critical and is only logged https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/task_manager.go#L232-L236.
The only case we deal with save error is when first adding the task in acs handler https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/payload_handler.go#L165-L171, in that case if we fail to save, we don't ack the payload (so the task will be sent again and we will try to save again). that behavior is maintained in this pr as well.

PR #2492, opened 2020-06-24
Refactoring Conditional to BooleanDefaultTrue

yunhee-l:
### Summary
Refactoring Conditional to BooleanDefaultTrue, replacing "DefaultTrue" to "NotSet", so that default false values can be supported.  
This is the first step in fixing the boolean config merge bug; I'll implement BooleanDefaultFalse next and apply it to one boolean config + test, and after that we need to migrate all boolean configs to use either of BooleanDefaultTrue or BooleanDefaultFalse instead of just boolean. I'm breaking down these to have reasonably sized PRs. 

### Implementation details
Replaced DefaultTrue with NotSet, and moved interpretation of NotSet value to a new wrapper struct BooleanDefaultTrue. With this, BooleanDefaultFalse can be implemented to use the same enum values but apply different defaulting logic to the not set boolean config.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/config/config_unix.go
+++ b/agent/config/config_unix.go
@@ -71,7 +71,7 @@ func DefaultConfig() Config {
 		PauseContainerTag:                   DefaultPauseContainerTag,
 		AWSVPCBlockInstanceMetdata:          false,
 		ContainerMetadataEnabled:            false,
-		TaskCPUMemLimit:                     DefaultEnabled,
+		TaskCPUMemLimit:                     BooleanDefaultTrue{Value: NotSet},
 		CgroupPath:                          defaultCgroupPath,
 		TaskMetadataSteadyStateRate:         DefaultTaskMetadataSteadyStateRate,
 		TaskMetadataBurstRate:               DefaultTaskMetadataBurstRate,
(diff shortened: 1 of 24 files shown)


agent/config/config_unix.go
```diff
@@ -71,7 +71,7 @@ func DefaultConfig() Config {
 		PauseContainerTag:                   DefaultPauseContainerTag,
 		AWSVPCBlockInstanceMetdata:          false,
```
sparrc, line 72: Why are we not implementing a BooleanDefaultFalse struct and replacing all of these config vars that default to false with that? Are "default false" config vars not affected by this bug?

sparrc:
ECS_POLL_METRICS also defaults to true (see https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config.go#L551), does this env var also need to be replaced with BooleanDefaultTrue?

yunhee-l:
> ECS_POLL_METRICS also defaults to true (see https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config.go#L551), does this env var also need to be replaced with BooleanDefaultTrue?

Yes eventually. Goal is to add BooleanDefaultFalse next, and migrate the bool configs to use these.

agent/config/config_unix.go
yunhee-l: im working on that now, each change would be 20+ file changes so I am breaking them down.

PR #2501, opened 2020-06-29
Added boltdb implementation to get, update and delete image state; Updated ImageManager to use data Client

mythri-garaga:
### Summary

Added boltdb implementation to get, update and delete image state. Updated ImageManager to use data Client instead of state manager to persist image states.

These changes cover the following scenarios
- A new image is being used for a container that we manage;
- In container pull step, add container reference to image (to track container usage of image);
- When container is removed, remove container reference of image;
- When no container is using the image and it has reached our image cleanup threshold.

### Implementation details

Following changes were made 
- `app`: replaced statemanger saver in agent.go with dataClient, updated the tests accordingly
- `data`: added botldb implementation for SaveImageState, DeleteImageState and GetImageStates
- `engine`: 
    - `docker_image_manager.go`: 
       - added SetDataClient method, used to set ImageManager.dataClient
       - save imageState to botldb on addContainerReferenceToExistingImageState, addContainerReferenceToNewImageState, RemoveContainerReference, addImageState or removeExistingImageNameOfDifferentID
       - delete imageState from doltdb on removeImageState
   - `docker_task_engine.go`: save image state on updateContainerReference
   - `image/types`: added GetImageID method to retrieve image ID
   - `mocks/engine_mocks.go`: updated ImageManager mock methods

### Testing

`make test` passes. Manually tested

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/data/imagestate_client.go
+++ b/agent/data/imagestate_client.go
@@ -14,20 +14,44 @@
 package data
 
 import (
+	"encoding/json"
+
 	"github.com/aws/amazon-ecs-agent/agent/engine/image"
+
+	"github.com/boltdb/bolt"
+	"github.com/pkg/errors"
 )
 
 func (c *client) SaveImageState(img *image.ImageState) error {
-	// TODO: implementation
-	return nil
+	id := img.GetImageID()
+	if id == "" {
+		return errors.New("failed to generate database image id")
+	}
+	return c.db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(imagesBucketName))
+		return putObject(b, id, img)
+	})
 }
 
 func (c *client) DeleteImageState(id string) error {
-	// TODO: implementation
-	return nil
+	return c.db.Update(func(tx *bolt.Tx) error {
+		b := tx.Bucket([]byte(imagesBucketName))
+		return b.Delete([]byte(id))
+	})
 }
 
 func (c *client) GetImageStates() ([]*image.ImageState, error) {
-	// TODO: implementation
-	return nil, nil
+	var imageStates []*image.ImageState
+	err := c.db.View(func(tx *bolt.Tx) error {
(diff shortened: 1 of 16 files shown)


agent/engine/docker_image_manager.go
```diff
@@ -175,6 +174,10 @@ func (imageManager *dockerImageManager) addContainerReferenceToExistingImageStat
 	imageState, ok := imageManager.getImageState(container.ImageID)
 	if ok {
 		imageState.UpdateImageState(container)
+		err := imageManager.dataClient.SaveImageState(imageState)
+		if err != nil {
+			return !ok
```
fenxiong, line 179: since the error is not returned, i think we should log it here for visibility

agent/engine/docker_image_manager.go
```diff
@@ -175,6 +174,10 @@ func (imageManager *dockerImageManager) addContainerReferenceToExistingImageStat
 	imageState, ok := imageManager.getImageState(container.ImageID)
 	if ok {
 		imageState.UpdateImageState(container)
+		err := imageManager.dataClient.SaveImageState(imageState)
```
fenxiong, line 177: can you add some test cases for changes in this file using a test data client? you can refer to what i have in https://github.com/aws/amazon-ecs-agent/blob/boltdb/agent/engine/task_manager_data_test.go and https://github.com/aws/amazon-ecs-agent/blob/boltdb/agent/engine/data_test.go

agent/data/imagestate_client.go
```diff
@@ -14,20 +14,44 @@
 package data
```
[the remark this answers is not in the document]
mythri-garaga: Done.

agent/engine/docker_image_manager.go
mythri-garaga: Done, added test cases in engine/docker_image_manager_data_test.go

PR #2502, opened 2020-06-29
Add and save task local ip address in task struct

fenxiong:
### Summary

Add and save task local ip address in task struct.

For a task in awsvpc network mode, the task engine state holds a [mapping](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/dockerstate/docker_task_engine_state.go#L102) between the task's local ip address and the task, and the mapping is saved as part of the state via state manager. With migration to boltdb, this mapping is not saved. So to maintain this information in boltdb, the ip address is added as a field of the task struct and it is saved as part of the task. When loading the task from boltdb (load will be implemented in a later change), we will look at this field of the task and fill in the ip <-> task mapping in memory.

### Implementation details

`api/task`: added a new field LocalIPAddressUnsafe to task struct and added getter/setter;
`engine`: set the local ip address when we get it after setting up the task network.

### Testing

Unit test: modify an existing one to cover the change.
Manually tested by running an awsvpc task and verified the ip is saved.

N/A

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -1244,6 +1244,8 @@ func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task, 
 	taskIP := result.IPs[0].Address.IP.String()
 	seelog.Infof("Task engine [%s]: associated with ip address '%s'", task.Arn, taskIP)
 	engine.state.AddTaskIPAddress(taskIP, task.Arn)
+	task.SetLocalIPAddress(taskIP)
+	engine.saveTaskData(task)
 	return dockerapi.DockerContainerMetadata{
 		DockerID: cniConfig.ContainerID,
 	}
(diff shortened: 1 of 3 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1244,6 +1244,8 @@ func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task,
 	taskIP := result.IPs[0].Address.IP.String()
 	seelog.Infof("Task engine [%s]: associated with ip address '%s'", task.Arn, taskIP)
 	engine.state.AddTaskIPAddress(taskIP, task.Arn)
```
sharanyad, line 1246: is the plan to fully migrate to boltdb and then remove the state save here?

fenxiong: yeah, i plan to remove these usages of state manager after boltdb changes are done

PR #2503, opened 2020-06-30
Boolean config fix

yunhee-l:
### Summary
- Refactoring Conditional to BooleanDefaultTrue, replacing "DefaultTrue" to "NotSet", so that default false values can be supported.
- Adding BooleanDefaultFalse
- I've tried but given up on adding an interface to hide BooleanDefaultTrue/False structs because:
  - I don't see benefit of another layer of abstraction, there are no more variations of boolean default we need 
  - built in expectation of non-nil config values in agent, am concerned with making changes that could be backward breaking 

### Implementation details
Replaced DefaultTrue with NotSet, and moved interpretation of NotSet value to a new wrapper struct BooleanDefaultTrue. BooleanDefaultFalse is implemented to use the same enum values but apply different defaulting logic to the not set boolean config.

### Testing

New tests cover the changes: No. Existing tests have been updated to reflect the change in config value type.

### Licensing

### DIFF
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -275,13 +275,13 @@ func (agent *ecsAgent) appendTaskCPUMemLimitCapabilities(capabilities []*ecs.Att
 	if agent.cfg.TaskCPUMemLimit.Enabled() {
 		if _, ok := supportedVersions[dockerclient.Version_1_22]; ok {
 			capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+capabilityTaskCPUMemLimit)
-		} else if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+		} else if agent.cfg.TaskCPUMemLimit.Value == config.ExplicitlyEnabled {
 			// explicitly enabled -- return an error because we cannot fulfil an explicit request
 			return nil, errors.New("engine: Task CPU + Mem limit cannot be enabled due to unsupported Docker version")
 		} else {
 			// implicitly enabled -- don't register the capability, but degrade gracefully
 			seelog.Warn("Task CPU + Mem Limit disabled due to unsupported Docker version. API version 1.22 or greater is required.")
-			agent.cfg.TaskCPUMemLimit = config.ExplicitlyDisabled
+			agent.cfg.TaskCPUMemLimit.Value = config.ExplicitlyDisabled
 		}
 	}
 	return capabilities, nil
--- a/agent/config/conditional.go
+++ b/agent/config/conditional.go
@@ -24,17 +24,21 @@ const (
 	_ Conditional = iota
 	ExplicitlyEnabled
 	ExplicitlyDisabled
-	DefaultEnabled
+	NotSet
 )
 
+type BooleanDefaultTrue struct {
+	Value Conditional
+}
+
 // Enabled is a convenience function for when consumers don't care if the value is implicit or explicit
-func (b Conditional) Enabled() bool {
-	return b == ExplicitlyEnabled || b == DefaultEnabled
+func (b BooleanDefaultTrue) Enabled() bool {
+	return b.Value == ExplicitlyEnabled || b.Value == NotSet
 }
 
 // MarshalJSON is used to serialize the type to json, per the Marshaller interface
-func (b Conditional) MarshalJSON() ([]byte, error) {
-	switch b {
+func (b BooleanDefaultTrue) MarshalJSON() ([]byte, error) {
+	switch b.Value {
 	case ExplicitlyEnabled:
 		return json.Marshal(true)
 	case ExplicitlyDisabled:
@@ -45,7 +49,7 @@ func (b Conditional) MarshalJSON() ([]byte, error) {
 }
 
 // UnmarshalJSON is used to deserialize json types into Conditional, per the Unmarshaller interface
-func (b *Conditional) UnmarshalJSON(jsonData []byte) error {
+func (b *BooleanDefaultTrue) UnmarshalJSON(jsonData []byte) error {
 	jsonString := string(jsonData)
 	jsonBool, err := strconv.ParseBool(jsonString)
 	if err != nil && jsonString != "null" {
@@ -53,11 +57,11 @@ func (b *Conditional) UnmarshalJSON(jsonData []byte) error {
 	}
 
 	if jsonString == "" || jsonString == "null" {
-		*b = DefaultEnabled
+		b.Value = NotSet
 	} else if jsonBool {
-		*b = ExplicitlyEnabled
+		b.Value = ExplicitlyEnabled
 	} else {
-		*b = ExplicitlyDisabled
+		b.Value = ExplicitlyDisabled
 	}
 
 	return nil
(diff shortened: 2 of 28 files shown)


agent/app/agent_capability.go
```diff
@@ -275,13 +275,13 @@ func (agent *ecsAgent) appendTaskCPUMemLimitCapabilities(capabilities []*ecs.Att
 	if agent.cfg.TaskCPUMemLimit.Enabled() {
 		if _, ok := supportedVersions[dockerclient.Version_1_22]; ok {
 			capabilities = appendNameOnlyAttribute(capabilities, attributePrefix+capabilityTaskCPUMemLimit)
-		} else if agent.cfg.TaskCPUMemLimit == config.ExplicitlyEnabled {
+		} else if agent.cfg.TaskCPUMemLimit.Value == config.ExplicitlyEnabled {
```
sparrc, line 278: my thought was that we could have an interface (`config.Boolean`) that doesn't return whether it's explicitly enabled or disabled, just whether it's enabled/disabled and set/unset. That way this if-statement would change to something like this:

```
else if agent.cfg.TaskCPUMemLimit.Enabled() && !agent.cfg.TaskCPUMemLimit.Unset() {
```

agent/config/conditional.go
```diff
@@ -24,17 +24,61 @@ const (
 	_ Conditional = iota
 	ExplicitlyEnabled
 	ExplicitlyDisabled
-	DefaultEnabled
+	NotSet
 )
 
+type BooleanDefaultTrue struct {
```
fenxiong, line 30: if the goal is to fix the merge of boolean config values, i'm wondering can we just change the config type from bool to *bool instead of having this struct (the merge can differentiate a nil pointer from a pointer that has false value).

yunhee-l: I considered that option, but felt that it has higher risk of backward incompatibility with previous agent versions.
We already had this "Conditional" struct as a solution for one of the config, so that's what I expanded. I am not opposed to the *bool solution, but it'll take some looking into, to make sure that it's feasible solution for backward compatibility as well.

agent/utils/utils.go
```diff
@@ -46,6 +47,10 @@ func ZeroOrNil(obj interface{}) bool {
 	if obj == nil {
 		return true
 	}
+	str, _ := json.Marshal(obj)
+	if string(str) == "null" {
```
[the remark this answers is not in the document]
yunhee-l: good catch! can't reference the BooleanDefault* struct due to cyclic dependency so I'm adding a dummy struct in test to replicate null json marshalling.

petderek, line 51: when would this condition get hit? afaik "null" should never show up here.

petderek: nvm -- i see it

petderek: Also -- you shouldn't be attempting to json.Marshal anything that comes through here. That likely isn't safe. `ZeroOrNil` is probably the wrong function to depend on for what you're doing.

yunhee-l: it's comparing the marshalled value not the string "null" value itself. I think I'm getting myself confused now, am looking into it.

petderek: So in our unmarshaler, we try to handle all values that mean "not set" from json's perspective: null, 3, or also not present.

Once we marshal it in though, it has the value `3` meaning not set.

Another way to think about this -- the struct we're working with doesn't have a zero value, so calling "ZeroOrNil" isn't designed to work here:

```
        _ Conditional = iota  // this would have been the zero value, but we don't have it.
	ExplicitlyEnabled
	ExplicitlyDisabled
	NotSet
```

yunhee-l: Hence the update in zeroOrNil function. 
Marshal/unmarshal logic didn't change from when it was just Conditional for TaskCPUMemLimit, which would suggest that merge logic doesn't work for this config as-is either. I'm taking a look.

petderek: > Hence the update in zeroOrNil function.

You shouldn't use or update this function, you are updating it to hit a corner case for something that doesn't apply.

yunhee-l: This function is already used by the merge logic though. Again, I'm looking into exactly how TaskCPUMemLimit was working as expected with merge.

petderek: I understand that merge currently uses that. But you should be performing this evaluation in your merge code instead of adding subtle check in the utility function that works only for the specific use case you're currently dealing with.

The right way to do that is definitely not by changing the definition of what ZeroOrNil means.

agent/config/conditional.go
yunhee-l: since I keep forgetting this, leaving in a comment - plain boolean pointer won't work for our case since we have at least 2 configs that defaults to true when not set.

agent/config/config.go
```diff
@@ -195,8 +195,16 @@ func (cfg *Config) Merge(rhs Config) *Config {
 
 	for i := 0; i < left.NumField(); i++ {
 		leftField := left.Field(i)
-		if utils.ZeroOrNil(leftField.Interface()) {
-			leftField.Set(reflect.ValueOf(right.Field(i).Interface()))
+		switch leftField.Interface().(type) {
+		case BooleanDefaultFalse, BooleanDefaultTrue:
+			str, _ := json.Marshal(reflect.ValueOf(leftField.Interface()).Interface())
```
yunhee-l, line 200: im open to explicitly checking NotSet value here but json logic already handles it so kept it this way.

agent/config/conditional.go
```diff
@@ -24,17 +24,61 @@ const (
...
+	}
+}
+
+// UnmarshalJSON is used to deserialize json types into Conditional, per the Unmarshaller interface
+func (b *BooleanDefaultTrue) UnmarshalJSON(jsonData []byte) error {
+	jsonString := string(jsonData)
+	jsonBool, err := strconv.ParseBool(jsonString)
+	if err != nil && jsonString != "null" {
+		return err
+	}
```
shubham2892, line 57: what if `err!=nil` and jsonString is not `null`, that is a feasible case?

agent/config/conditional.go
```diff
@@ -24,17 +24,61 @@ const (
...
+	if jsonString == "" || jsonString == "null" {
+		b.Value = NotSet
+	} else if jsonBool {
+		b.Value = ExplicitlyEnabled
+	} else {
+		b.Value = ExplicitlyDisabled
+	}
+
+	return nil
+}
```
shubham2892, line 68: Is it possible to restructure this code to 


```
       jsonString := string(jsonData)
       if jsonString == "" || jsonString == "null" {
		b.Value = NotSet
                return nil
        }
	jsonBool, err := strconv.ParseBool(jsonString)
	if err != nil  {
		return err
	}

	if jsonBool {
		b.Value = ExplicitlyEnabled
	} else {
		b.Value = ExplicitlyDisabled
	}

```
Just from readability point of view, in the current form it is bit confusing to understand what is happening.

yunhee-l: I think I prefer to keep it as is (it was what was there before), and to me it makes sense to group the error handling and b.value setting separate.

agent/app/agent_capability.go
yunhee-l: Interface itself isn't a problem, but from my trying to implement this I found that agent's config has fair bit of built in assumptions that got me into few rabbit holes (e.g. non-nil config values are expected throughout tests and code).

PR #2518, opened 2020-07-14
Fix Telemetry model to include per second network stats metrics

ubhattacharjya:
### Summary

This PR fixes the telemetry model for Network stats per sec metrics.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/stats/queue.go
+++ b/agent/stats/queue.go
@@ -216,11 +216,11 @@ func (queue *Queue) GetNetworkStatsSet() (*ecstcs.NetworkStatsSet, error) {
 	if err != nil {
 		seelog.Warnf("Error getting network tx packets: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


agent/stats/queue.go
```diff
@@ -430,3 +430,50 @@ func (queue *Queue) getULongStatsSet(getUsageInt getUsageIntFunc) (*ecstcs.ULong
 		OverflowSum: &overflowSum,
 	}, nil
 }
+
+// getCWStatsSet gets the stats set for per second network metrics
```
yunhee-l, line 434: Nit: comment does not match the name of the function. It would be helpful to explain what you mean by UDouble here too? Is it unsigned double type?

ubhattacharjya: Thanks for point out Yunhee. Fixed it.

PR #2519, opened 2020-07-14
Save all state to boltdb in termination handler

mythri-garaga:
### Summary

Save all state to boltdb before agent exits in termination handler
### Implementation details

 `signalhandlers/termination_handlers`: In FinalSave, using saveStateAll to save the states to boltdb instead of statemanager.
Updated function calls and tests in `acs` and `app` accordignly
### Testing

`make test` passes and manually tested too
### Licensing

### DIFF
--- a/agent/sighandlers/termination_handler.go
+++ b/agent/sighandlers/termination_handler.go
@@ -27,9 +27,10 @@ import (
 	"time"
 
 	apierrors "github.com/aws/amazon-ecs-agent/agent/api/errors"
+	"github.com/aws/amazon-ecs-agent/agent/data"
 	"github.com/aws/amazon-ecs-agent/agent/engine"
+	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/sighandlers/exitcodes"
-	"github.com/aws/amazon-ecs-agent/agent/statemanager"
 
 	"github.com/cihub/seelog"
 )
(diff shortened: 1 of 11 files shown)


agent/sighandlers/termination_handler.go
```diff
@@ -87,8 +88,9 @@ func FinalSave(saver statemanager.Saver, taskEngine engine.TaskEngine) error {
 	})
 	go func() {
 		seelog.Debug("Saving state before shutting down")
-		stateSaved <- saver.ForceSave()
+		saveStateAll(state, dataClient)
 		saveTimer.Stop()
+		stateSaved <- nil
```
cyastella, line 93: just curious, it seems `stateSaved` is used for error reporting, after this change, the return code (`stateSaved`) will always be `nil`, is that expected? Are we handling error from `saveStateAll` somewhere else?

agent/sighandlers/termination_handler.go
```diff
@@ -87,8 +88,9 @@ func FinalSave(saver statemanager.Saver, taskEngine engine.TaskEngine) error {
 	})
 	go func() {
 		seelog.Debug("Saving state before shutting down")
-		stateSaved <- saver.ForceSave()
```
cyastella, line 90: just curious, will `saver.ForceSave()` be used anywhere else after this change?

agent/sighandlers/termination_handler.go
```diff
@@ -98,3 +100,37 @@ func FinalSave(saver statemanager.Saver, taskEngine engine.TaskEngine) error {
...
+		}
+		if err := dataClient.SaveDockerContainer(container); err != nil {
+			seelog.Errorf("Failed to save data for docker container %s: %v", container.Container.Name, err)
+		}
+	}
+
+	// save all image state data
+	for _, imageState := range state.AllImageStates() {
+		if err := dataClient.SaveImageState(imageState); err != nil {
+			seelog.Errorf("Failed to save data for image state: %s, %v", imageState.GetImageID(), err)
```
cyastella, line 126: nit: would suggest use the same logging format: `for image state %s: %v`

agent/sighandlers/termination_handler.go
mythri-garaga: stateSave channel will still receive error when saveTimer times out. 
If there is any error in saveStateAll we are just going to log them(similar to https://github.com/aws/amazon-ecs-agent/blob/boltdb/agent/engine/data.go#L27-L30)

agent/sighandlers/termination_handler.go
mythri-garaga: We will be removing ForceSave() implementation from agent codebase once boltdb is completely implemented.

agent/sighandlers/termination_handler.go
mythri-garaga: Thanks for pointing it out. Will fix it.

PR #2520, opened 2020-07-15
Bug Fix: fix V4 metadata endpoint displaying network rate metrics for…

ubhattacharjya:
… awsvpc mode

### Summary

This PR fixes a big for V4 metadata endpoint exposing network rate metrics for awsvpc network mode. Initially it was accessing container stats queue. But for awsvpc mode, we should access the task stats queue.

### Testing

Tested after running a task with awsvpc network mode and checked that v4 metadata endpoint is dispaying the network rate metrics

### Licensing

### DIFF
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -756,6 +756,7 @@ func (engine *DockerStatsEngine) ContainerDockerStats(taskARN string, containerI
 		return nil, nil, errors.Errorf("stats engine: container not found: %s", containerID)
 	}
 	containerStats := container.statsQueue.GetLastStat()
+	containerNetworkRateStats := container.statsQueue.GetLastNetworkStatPerSec()
 
 	// Insert network stats in container stats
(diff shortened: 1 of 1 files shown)


agent/stats/engine.go
```diff
@@ -769,10 +770,11 @@ func (engine *DockerStatsEngine) ContainerDockerStats(taskARN string, containerI
 		if ok {
 			taskNetworkStats := taskStats.StatsQueue.GetLastStat().Networks
 			containerStats.Networks = taskNetworkStats
+			containerNetworkRateStats = taskStats.StatsQueue.GetLastNetworkStatPerSec()
```
yunhee-l, line 773: shouldn't this reuse the variable containerNetworkRateStats?

ubhattacharjya: It will get overwritten.

yunhee-l: oh i totally missed that.

PR #2521, opened 2020-07-15
Add method for task engine to load state from boltdb

fenxiong:
### Summary

Add method for task engine to load state from boltdb. Currently it is unused. Will use it in a later change that implements all the loading logic.

### Implementation details

Load data from task, container, eni attachment and image table into task engine state.

### Testing

Currently tested with unit test. Manual test will be done with later change when all the loading logic is added.

N/A

### Licensing

### DIFF
--- a/agent/engine/data.go
+++ b/agent/engine/data.go
@@ -21,7 +21,101 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 
 	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
 )
+
+// LoadState populates the task engine state with data in db.
+func (engine *DockerTaskEngine) LoadState() error {
+	if err := engine.loadTasks(); err != nil {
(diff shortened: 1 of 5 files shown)


agent/engine/data.go
```diff
@@ -21,8 +21,75 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 
 	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
 )
 
+// LoadState populates the task engine state with data in db.
+func (engine *DockerTaskEngine) LoadState() error {
```
yunhee-l, line 28: nit: could this be broken down into dedicated load* methods? for e.g. loadContainers, loadTasks etc. If it add unnecessary complexity I am not going to push it.

fenxiong: sounds reasonable, i made the change

PR #2524, opened 2020-07-16
Send nil for network rate stats when network stats is nil

ubhattacharjya:
### Summary

If network stats is not being sent, then the rate metrics should not be visible.

### Testing

Tested that the metadata endpoint is not displaying any rate stats if networks sent is nil.

### Licensing


agent/stats/engine.go
```diff
@@ -771,6 +771,9 @@ func (engine *DockerStatsEngine) ContainerDockerStats(taskARN string, containerI
 			taskNetworkStats := taskStats.StatsQueue.GetLastStat().Networks
 			containerStats.Networks = taskNetworkStats
 			containerNetworkRateStats = taskStats.StatsQueue.GetLastNetworkStatPerSec()
+			if len(containerStats.Networks) == 0 {
```
yunhee-l, line 774: can we make sure that this is added to test case?

PR #2528, opened 2020-07-17
Load data from boltdb upon startup

fenxiong:
### Summary

Implemented logic for loading data from boltdb upon startup, while preseving backward compatibility by falling back to load from state file.

### Implementation details

Details:
* `app`:
  - In data.go, implement method `loadData` that loads data from previous data file, either boltdb or state file. In the later case, data is migrated to boltdb after loading. Behavior of three cases are considered:
    1. Agent starts from fresh instance (no previous state):
       (1) Try to load from boltdb, get nothing;
       (2) Try to load from state file, get nothing;
       (3) Return empty data.

    2. Agent starts with previous state stored in boltdb:
       (1) Try to load from boltdb, get the data;
       (2) Return loaded data.

    3. Agent starts with previous state stored in state file (i.e. it was just upgraded from an old agent that uses state file):
       (1) Try to load from boltdb, get nothing;
       (2) Try to load from state file, get something;
       (3) Save loaded data to boltdb;
       (4) Return loaded data.

  - In agent.go, invoke `loadData` method to load data, replacing the existing few lines of code that uses state manager to load data.
  - Update a few unit tests to use actual task engine state instead of mock one because with the changes, it would be tedious to list all the expected calls to the engine state.
* `engine`: added a method SaveState which saves the whole task engine state to boltdb.

### Testing

Unit tests added/updated. Manually tested:
(1) Startup on fresh instance;
(2) Start from a state where a record of a running task is stored in boltdb;
(3) Start from a state where a record of a running task is stored in state file.

N/A

### Licensing


agent/app/data.go
```diff
@@ -0,0 +1,194 @@
...
+	taskEngine               engine.TaskEngine
+	agentVersion             string
+	availabilityZone         string
+	cluster                  string
+	containerInstanceARN     string
+	ec2InstanceID            string
+	latestTaskManifestSeqNum int64
+}
+
+// loadData loads data from previous checkpoint file, if any, with backward compatibility preserved. It first tries to
```
yunhee-l, line 44: thank you for the detailed explanation in the comment!

fenxiong: no problem :)

PR #2531, opened 2020-07-20
Check task network device before querying network stats

shubham2892:
### Summary
- `StartStatsCollection` was being called for every container in a task, it should be called only once for every task.
- Task network device name was being fetched only at the beginning of collection of stats, it might be possible that the devices has not been attached to the container yet, instead the check is done everytime before querying for the task.

### Testing
 How was this tested?

Tested manually that `startStatsCollection` is called only once for pause container for an AWSVPC task. Also, Tested manually that device name is not none by running an AWSVPC task. Before this change, I could see instances of device name being none.

As a result of these two changes, I am able to consistently see right `rx_bytes_per_sec` when using v4 taskmeta data endpoint, before this change,  `rx_bytes_per_sec` in lot of instances.

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -179,8 +179,16 @@ func (engine *DockerStatsEngine) addAndStartStatsContainer(containerID string) {
 	if err != nil {
 		return
 	}
+
+	dockerContainer, errResolveContainer := engine.resolver.ResolveContainer(containerID)
+	if errResolveContainer != nil {
+		seelog.Debugf("Could not map container ID to container, container: %s, err: %s", containerID, err)
+		return
+	}
+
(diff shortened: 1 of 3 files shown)


agent/stats/engine.go
```diff
@@ -179,8 +179,15 @@ func (engine *DockerStatsEngine) addAndStartStatsContainer(containerID string) {
 	if err != nil {
 		return
 	}
+
+	dockerContainer, errResolveContainer := engine.resolver.ResolveContainer(containerID)
+	if errResolveContainer != nil {
+		return
```
yunhee-l, line 185: should something be logged here?

ubhattacharjya:
Can you add the testing done explanation in the PR?

PR #2533, opened 2020-07-21
Device len fix

shubham2892:
### Summary
`deviceList` being passed `getAWSVPCNetworkStats` was not getting set with updated device list fetched in the function. The function has been updated to use `taskStat.TaskMetadata.DeviceName` so that the latest deviceList fetched is being used everywhere. 

### Testing
How was this tested? 
Tested by manually running a task and check task stats.

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/stats/task_linux.go
+++ b/agent/stats/task_linux.go
@@ -124,8 +124,7 @@ func (taskStat *StatsTask) collect() {
 
 func (taskStat *StatsTask) processStatsStream() error {
 	taskArn := taskStat.TaskMetadata.TaskArn
-	awsvpcNetworkStats, errC := taskStat.getAWSVPCNetworkStats(taskStat.TaskMetadata.DeviceName,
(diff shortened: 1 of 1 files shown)


agent/stats/task_linux.go
```diff
@@ -223,9 +220,11 @@ func (taskStat *StatsTask) getAWSVPCNetworkStats(deviceList []string, containerP
 						return
 					}
```
ubhattacharjya, line 221: Just curious what will be the behaviour when err==nil and deviceList = []?

shubham2892: It might be possible that in the first run, devicelist=[] and err==nil because network devices hasn't yet attached to the docker container, in that case networkstats will be empty. But because we are checking if deviceList is empty in every run, it will again query for the device list until it finds a device, which has to be there for task to make network requests.

PR #2536, opened 2020-07-21
Configure separate logging levels

tamor1:
### Summary

Adds a new environment variable ECS_LOGLEVEL_ON_INSTANCE to allow different logging levels to be set for the two types of logging, the on-instance file and the logging driver. 

### Implementation details

Updates the seelog config to apply filters to the different logging destinations by converting the minimum level input in the environment variable to a list of levels to allow through the filter. Keeps the same default as before, where the logging driver defaults to info level. The on instance file will also default to info, unless a logging driver is explicitly set for the agent container, in which case it will default to off. 

### Testing

Updated and added more units tests to cover the config changes.  Also manually tested by setting variables in ecs config file. 

### Licensing


agent/app/args/flag.go
```diff
@@ -13,11 +13,14 @@
...
-import "flag"
+import (
+	"flag"
+)
 
 const (
 	versionUsage             = "Print the agent version information and exit"
-	logLevelUsage            = "Loglevel: [<crit>|<error>|<warn>|<info>|<debug>]"
+	driverLogLevelUsage      = "Loglevel for docker logging driver: [<crit>|<error>|<warn>|<info>|<debug>]"
+	instanceLogLevelUsage    = "Loglevel for Agent on-instance file: [<none>|<crit>|<error>|<warn>|<info>|<debug>]"
```
mythri-garaga, line 23: nit: can we have "Agent on-instance log file" instead of "Agent on-instance file"?

agent/logger/log.go
```diff
@@ -112,12 +134,19 @@ func SetLevel(logLevel string) {
 		"crit":  "critical",
 		"none":  "off",
 	}
-	parsedLevel, ok := levels[strings.ToLower(logLevel)]
 
-	if ok {
+	parsedDriverLevel, driverOk := levels[strings.ToLower(driverLogLevel)]
+	parsedInstanceLevel, fileOk := levels[strings.ToLower(instanceLogLevel)]
+
+	if fileOk || driverOk {
```
mythri-garaga, line 141: nit: rename to instanceOk for consistency?

agent/app/args/flag.go
```diff
@@ -13,11 +13,14 @@
...
 package args
 
-import "flag"
+import (
+	"flag"
+)
 
 const (
 	versionUsage             = "Print the agent version information and exit"
-	logLevelUsage            = "Loglevel: [<crit>|<error>|<warn>|<info>|<debug>]"
```
yunhee-l, line 20: Who provides these flags? I'm wondering if removing this flag can potentially impact customers if they have some custom setting where they interact directly with agent.

tamor1: These are the flags the user can provide for running agent with the docker exec command. Do you think it would be better to have the old flag and also the 2 new ones, and if the old one is set, use that instead?

yunhee-l: sorry I meant to tag this comment where "logLevelFlagName" variable is at line# 29 (new line# 32) where the flag name changed from "loglevel" to "loglevel-driver", but I think you got what I meant. 
if this is user facing, removing a flag is a backward breaking change. I think what you suggested should work.

PR #2538, opened 2020-07-21
Feature / fluentD : Support for sub second precision 

MeghnaPrabhu:
Update: Support for sub second precision in FluentD Docker Log Driver.

### Summary
Adding an option in fluentD Docker Log Driver configuration to enable sub second precision. 
This is being enabled to resolve the following request https://github.com/aws/containers-roadmap/issues/839

### Implementation 

- set [fluentd-sub-second-precision](https://docs.docker.com/config/containers/logging/fluentd/#fluentd-sub-second-precision) to true.

### Testing
Manually tested this feature.
1. Build the ecs agent
2. Run a firelens task on an ecs cluster with an ec2 instance.
3. Monitor cloudwatch to ensure logs had timestamps with sub second precision.

Feature FluentD: Support for sub second precision.

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -79,14 +79,15 @@ const (
 	engineConnectRetryJitterMultiplier = 0.20
 	engineConnectRetryDelayMultiplier  = 1.5
 	// logDriverTypeFirelens is the log driver type for containers that want to use the firelens container to send logs.
-	logDriverTypeFirelens   = "awsfirelens"
-	logDriverTypeFluentd    = "fluentd"
-	logDriverTag            = "tag"
-	logDriverFluentdAddress = "fluentd-address"
-	dataLogDriverPath       = "/data/firelens/"
-	logDriverAsyncConnect   = "fluentd-async-connect"
-	dataLogDriverSocketPath = "/socket/fluent.sock"
-	socketPathPrefix        = "unix://"
+	logDriverTypeFirelens       = "awsfirelens"
+	logDriverTypeFluentd        = "fluentd"
+	logDriverTag                = "tag"
+	logDriverFluentdAddress     = "fluentd-address"
+	dataLogDriverPath           = "/data/firelens/"
+	logDriverAsyncConnect       = "fluentd-async-connect"
+	logDriverSubSecondPrecision = "fluentd-sub-second-precision"
+	dataLogDriverSocketPath     = "/socket/fluent.sock"
+	socketPathPrefix            = "unix://"
 
 	// fluentTagDockerFormat is the format for the log tag, which is "containerName-firelens-taskID"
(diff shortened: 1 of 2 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -79,14 +79,15 @@ const (
...
-	logDriverAsyncConnect   = "fluentd-async-connect"
-	dataLogDriverSocketPath = "/socket/fluent.sock"
-	socketPathPrefix        = "unix://"
+	logDriverTypeFirelens   	= "awsfirelens"
+	logDriverTypeFluentd    	= "fluentd"
+	logDriverTag            	= "tag"
+	logDriverFluentdAddress 	= "fluentd-address"
+	dataLogDriverPath       	= "/data/firelens/"
+	logDriverAsyncConnect   	= "fluentd-async-connect"
+	logDriverSubSecondPrecision = "fluentd-sub-second-precision"
```
yunhee-l, line 88: formatting is incorrect here, this would probably fail our static check

yunhee-l:
Is this something that's transparent to customers?

PettitWesley:
@yunhee-l Its a non-breaking change, but it is not a hidden change. 

Its very small though, and I am very confident that if customers notice it, they will be happy (we have some strong requests for this). Many may not notice it at all.

Right now, Docker adds timestamps to all logs. If you use the json log driver, like they do in k8s, then you get a timestamp with nanosecond precision. For some reason, when the Fluentd Docker log driver was first implemented, it truncated the timestamps so that they only have seconds. With this change, that truncation will no longer take place, and the full timestamp will be sent to FireLens with full precision. 

Below are "before" and "after" screenshots, courtesy of @MeghnaPrabhu.

image(4)

image(5)

Basically, right now, you get all zeroes for everything after the decimal (just seconds). After this change, you get the real timestamp with full precision.

yunhee-l:
@PettitWesley thanks for the info, wanted to know whether we need to update our state file version with this change or not (which would be the case if this is something customers have to manually opt it/out). 

@MeghnaPrabhu please squash all commits into 1.

agent/engine/docker_task_engine.go
```diff
@@ -1102,6 +1103,7 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
 	logConfig.Config[logDriverTag] = tag
 	logConfig.Config[logDriverFluentdAddress] = fluentd
 	logConfig.Config[logDriverAsyncConnect] = strconv.FormatBool(true)
+	logConfig.Config[logDriverSubSecondPrecision] = strconv.FormatBool(true)
```
fenxiong, line 1106: can you update unit test to include a check for the new option? should be in https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine_test.go#L2910-L2915

PR #2563, opened 2020-08-12
adding FSxWindowsFileServerVolumeConfiguration

singholt:
### Summary

This adds support for a new field in the task definition: FSxWindowsFileServerVolumeConfiguration.

### Implementation details

- A sample FSxWindowsFileServerVolumeConfiguration would look like:

```
"fsxWindowsFileServerVolumeConfiguration”: {
       “fileSystemId”: “fs-12345678",
       “rootDirectory”: “test”,
       “authorizationConfig”: {
              “credentialsParameter”: “arn:aws:ssm:us-east-1:123456789012:parameter/test”,
              “domain”: “test”
       }
}
```

-  For now we only support Windows container instances. 

-  For any given fsxWindowsFileServerVolumeConfiguration, we will perform two mount events:

   - host mount - approach similar to creating a network drive mapping on the container instance.
   - bind mount - do a docker bind mount for each container in the task definition.

### Testing

New tests cover the changes: yes (TODO: functional tests) 

Add support for FSxWindowsFileServerVolumeConfiguration in task definition

### Licensing


agent/Gopkg.lock
```diff
@@ -10,7 +10,7 @@
   version = "v0.4.7"
 
 [[projects]]
-  digest = "1:4dfad161aee05f73bbd9740d329bb068a18028c0a9d6c0572f931ad320f71873"
+  digest = "1:da862fb191ab8fd95feaf3353b7d76627df37f8850006d86bb98cd855a112134"
```
vsiddharth, line 13: Is this intentional?

agent/api/task/task.go
```diff
@@ -379,6 +379,13 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 		return apierrors.NewResourceInitError(task.Arn, err)
 	}
 
+	if task.requiresFSxWindowsFileServerResource() {
+		if err := task.initializeFSxWindowsFileServerResource(cfg, credentialsManager, resourceFields); err != nil {
+			seelog.Errorf("Task [%s]: could not initialize FSx for Windows File Server resource: %v", task.Arn, err)
```
vsiddharth, line 384: nit: Having windows specific checks/logs in the platform agnostic file is confusing?

agent/api/task/task_windows.go
```diff
@@ -213,3 +181,86 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+		if volume.Type == FSxWindowsFileServerVolumeType {
+			return true
+		}
+	}
+	return false
+}
+
+// initializeFSxWindowsFileServerResource builds the resource dependency map for the fsxwindowsfileserver resource
+func (task *Task) initializeFSxWindowsFileServerResource (cfg *config.Config, credentialsManager credentials.Manager,
+	resourceFields *taskresource.ResourceFields) error {
```
vsiddharth, line 198: nit: please reformat here

agent/api/task/task_windows.go
```diff
@@ -213,3 +181,86 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+	for i, vol := range task.Volumes {
```
vsiddharth, line 199: `i` is unused. Can just use `_`.

agent/api/task/task_windows.go
```diff
@@ -213,3 +181,86 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+		if vol.Type != FSxWindowsFileServerVolumeType {
+			continue
+		}
+
+		fsxWindowsFileServerVol, ok := vol.Volume.(*fsxwindowsfileserver.FSxWindowsFileServerVolumeConfig)
```
vsiddharth, line 204: nit: Is there a method that can be used to obtain the config here

agent/api/task/task_windows.go
```diff
@@ -213,3 +181,86 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+		resourceFields.SSMClientCreator,
+		resourceFields.ASMClientCreator,
+		resourceFields.FSxClientCreator)
+	if err != nil {
+		return err
+	}
+
+	vol.Volume = &fsxwindowsfileserverResource.VolumeConfig
+	task.AddResource(resourcetype.FSxWindowsFileServerKey, fsxwindowsfileserverResource)
+	task.updateContainerVolumeDependency(vol.Name)
```
vsiddharth, line 249: nit: add empty line

agent/api/task/task_windows.go
```diff
@@ -213,3 +181,86 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+	return nil
+}
+
+// findUnusedDriveLetter is used to search for an available drive letter on the container instance.
```
vsiddharth, line 253: nit: can this be moved to utils

agent/app/agent_capability_windows_test.go
```diff
@@ -260,3 +260,47 @@ func TestAppendGMSACapabilitiesFalse(t *testing.T) {
...
+	agent := &ecsAgent{
+		cfg: &config.Config{
+			FSxWindowsFileServerCapable: true,
+		},
+	}
+
+	capabilities := agent.appendFSxWindowsFileServerCapabilities(inputCapabilities)
+
+	assert.Equal(t, len(expectedCapabilities), len(capabilities))
+	for i, expected := range expectedCapabilities {
```
vsiddharth, line 284: Shouldn't the test iterate over `capabilities`? 
Can you not make use of other methods in https://godoc.org/github.com/stretchr/testify/assert

agent/app/agent_capability_windows_test.go
```diff
@@ -260,3 +260,47 @@ func TestAppendGMSACapabilitiesFalse(t *testing.T) {
...
+		assert.Equal(t, aws.StringValue(expected.Name), aws.StringValue(capabilities[i].Name))
+		assert.Equal(t, aws.StringValue(expected.Value), aws.StringValue(capabilities[i].Value))
+	}
+}
+
+func TestAppendFSxWindowsFileServerCapabilitiesFalse(t *testing.T) {
```
vsiddharth, line 290: These tests are good candidates for table driven tests. 
Refer https://github.com/golang/go/wiki/TableDrivenTests

agent/config/parse_windows.go
```diff
@@ -53,6 +53,28 @@ func parseGMSACapability() bool {
 	return false
 }
 
+// parseFSxWindowsFileServerCapability is used to determine if fsxWindowsFileServer support can be enabled
```
vsiddharth, line 56: This looks almost identical to `parseGMSACapability`? 
Can this be refactored to incorporate a common function that accepts an environment variable?  
Then both of these methods can utilize the common method and we can avoid duplication.

agent/fsx/factory/factory.go
```diff
@@ -0,0 +1,51 @@
...
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package factory
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/credentials"
+	"github.com/aws/amazon-ecs-agent/agent/httpclient"
+	"github.com/aws/aws-sdk-go/aws"
```
vsiddharth, line 19: Please use `goimports` and update the import ordering
```
system packages

amazon-ecs-agent specific packages

external packages
```

agent/fsx/fsx.go
```diff
@@ -0,0 +1,55 @@
...
+package fsx
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/fsx"
+	"github.com/aws/aws-sdk-go/service/fsx/fsxiface"
+	"github.com/pkg/errors"
+)
+
+// GetFileSystemDNSNames extracts filesystems dns names from filesystems info
```
vsiddharth, line 23: Please update this comment and provide a sample of the map

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_unsupported.go
```diff
@@ -0,0 +1,183 @@
...
+
+	asmfactory "github.com/aws/amazon-ecs-agent/agent/asm/factory"
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
+	"github.com/aws/amazon-ecs-agent/agent/api/task/status"
+	"github.com/aws/amazon-ecs-agent/agent/credentials"
+	fsxfactory "github.com/aws/amazon-ecs-agent/agent/fsx/factory"
+	ssmfactory "github.com/aws/amazon-ecs-agent/agent/ssm/factory"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
```
vsiddharth, line 29: nit: empty line

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_unsupported.go
```diff
@@ -0,0 +1,183 @@
...
+}
+
+// Cleanup removes the fsxwindowsfileserver resources created for the task
+func (fv *FSxWindowsFileServerResource) Cleanup() error {
+	return errors.New("not implemented")
+}
+
+// MarshalJSON serialises the FSxWindowsFileServerResourceJSON struct to JSON
+func (fv *FSxWindowsFileServerResource) MarshalJSON() ([]byte, error) {
+	return nil, errors.New("not implemented")
```
vsiddharth, line 159: please use consistent error messages for unsupported platforms

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,663 @@
...
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package fsxwindowsfileserver
+
+import (
```
vsiddharth, line 18: please fix imports

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,663 @@
...
+		"-RequirePrivacy $true",
+		"-ErrorAction Stop")
+
+	_, err := cmd.CombinedOutput()
+	if err != nil {
+		fv.setTerminalReason(err.Error())
+		return err
+	}
+
+	cmd = execCommand("powershell.exe",
```
vsiddharth, line 532: What happens if the agent crashes mid-way during this method?

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,663 @@
...
+	dir := utils.GetCanonicalPath(rootDirectory)
+	dir = strings.Trim(dir, "\\")
+
+	fv.SetRootDirectory(dir)
+	return
+}
+
+func (fv *FSxWindowsFileServerResource) Cleanup() error {
+	hostPath := fv.VolumeConfig.FSxWindowsFileServerHostPath
+	fv.clearFSxWindowsFileServerResource(hostPath)
```
vsiddharth, line 554: This always returns `nil` even if the cleanup fails. Is that acceptable?

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
```diff
@@ -0,0 +1,700 @@
...
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package fsxwindowsfileserver
+
+import (
```
vsiddharth, line 18: please fix the go imports

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
```diff
@@ -0,0 +1,700 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+	"os/exec"
+	"os"
+)
+
+const (
```
vsiddharth, line 44: please format this file using `gofmt`

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
```diff
@@ -0,0 +1,700 @@
...
+	assert.WithinDuration(t, fsxWindowsFileServerIn.createdAtUnsafe, fsxWindowsFileServerOut.createdAtUnsafe, time.Microsecond)
+	assert.Equal(t, fsxWindowsFileServerIn.desiredStatusUnsafe, fsxWindowsFileServerOut.desiredStatusUnsafe)
+	assert.Equal(t, fsxWindowsFileServerIn.knownStatusUnsafe, fsxWindowsFileServerOut.knownStatusUnsafe)
+}
+
+func TestRetrieveCredentials(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	credentialsManager := mock_credentials.NewMockManager(ctrl)
```
vsiddharth, line 121: Instead of using similar setup structures
Can you move the test setup to a helper method and re-use it?

agent/taskresource/fsxwindowsfileserver/types.go
```diff
@@ -0,0 +1,22 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package fsxwindowsfileserver
+
+const (
+	// ResourceName is the name of the fsxwindowsfileserver resource
+	ResourceName = "fsxwindowsfileserver"
+)
+
```
vsiddharth, line 20: nit: empty lines

vsiddharth (changes requested):
Neat! Changes look good. 

A few minor points:
* `gocyclo` checks are failing. Please look into it
* What happens if customers attempt to create more than 26 volumes?

agent/Gopkg.lock
singholt: this is auto-generated as part of dep ensure

agent/api/task/task.go
singholt: the error message is not windows specific. also, i have both these checks returning false for linux in unsupported files. following same logic as checking for credspec resource.

agent/api/task/task_windows.go
singholt: will do

agent/api/task/task_windows.go
singholt: EFS obtains the config in a similar way - https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L552

agent/api/task/task_windows.go
singholt: will do

agent/api/task/task_windows.go
singholt: will do

agent/fsx/factory/factory.go
singholt: will do

agent/fsx/fsx.go
singholt: will do

agent/api/task/task_windows.go
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_unsupported.go
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
singholt: will do

agent/taskresource/fsxwindowsfileserver/types.go
singholt: will remove

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: this is to accommodate the case where customer might tamper with the drive mapping on the instance after task is stopped. thats why the warning instead of an error. (just like here - https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/credentialspec/credentialspec_windows.go#L563)

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
singholt: will do

agent/config/parse_windows.go
singholt: will do

agent/app/agent_capability_windows_test.go
singholt: this test is identical to ```TestAppendGMSACapabilities```

agent/app/agent_capability_windows_test.go
singholt: this test is identical to ```TestAppendGMSACapabilitiesFalse```

singholt:
> Neat! Changes look good.
> 
> A few minor points:
> 
>     * `gocyclo` checks are failing. Please look into it
> 
>     * What happens if customers attempt to create more than 26 volumes?

Regarding no. of volumes, although this is a highly unlikely scenario, you are right, 26 is the max no. of volumes that the customer can have on the instance. 

This limit is applicable for fsx for windows file server service in general and not just ECS integration. Customers using vanilla ec2 instances (not ECS cluster joined) will also experience the same limits.

agent/api/task/task_windows.go
singholt: it is actually being used on L207, getting it back

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
fierlion: good question.

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
```diff
@@ -458,4 +460,241 @@ func TestSetRootDirectory(t *testing.T) {
...
+	fv.Initialize(&taskresource.ResourceFields{
+		ResourceFieldsCommon: &taskresource.ResourceFieldsCommon{
+			SSMClientCreator:   ssmClientCreator,
+			ASMClientCreator:   asmClientCreator,
+			FSxClientCreator:   fsxClientCreator,
+			CredentialsManager: credentialsManager,
+		},
+	}, apitaskstatus.TaskStatusNone, apitaskstatus.TaskRunning)
+
+	ssmTestData := "{\n\"username\": \"user\", \n\"password\": \"pass\"\n}"
```
fierlion, line 469: I'm not totally clear on what we're using SSM (parameter store?) for.  Is this the AD creds?

singholt: yes

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: This is a valid concern! thanks for bringing this up @vsiddharth . Here's a thought - I am thinking to introduce an additional check here to verify if the allocated drive letter is still available. In case the agent crashes after the resource was created (but the resource status wasn't updated), this check would fail and would simply return the err as drive occupied.

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+		return err
+	}
+
+	return nil
+}
+
+func (fv *FSxWindowsFileServerResource) retrieveSSMCredentials(credentialsParameterARN string, iamCredentials credentials.IAMRoleCredentials) error {
+	parsedARN, err := arn.Parse(credentialsParameterARN)
+	if err != nil {
+		fv.setTerminalReason(err.Error())
```
sharanyad, line 450: this is being set again in the parent method, this is not needed(there are many other instances of this)

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+
+func (fv *FSxWindowsFileServerResource) clearFSxWindowsFileServerResource(localPath string) {
+	// formatting to keep powershell happy
+	localPath = fmt.Sprintf("-LocalPath %s", localPath)
+	seelog.Debug("removing fsxwindowsfileserver resource %s", fv.Name)
+	cmd := execCommand("powershell.exe", "Remove-SmbGlobalMapping", localPath, "-Force")
+
+	_, err := cmd.CombinedOutput()
+	if err != nil {
+		seelog.Warnf("unable to clear fsxwindowsfileserver host path %s for task %s: %s", localPath, fv.taskARN, err)
```
sharanyad, line 567: Capitalize the start of the log message

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows_test.go
```diff
@@ -0,0 +1,596 @@
...
+	assert.Equal(t, fsxWindowsFileServerIn.Name, fsxWindowsFileServerOut.Name)
+	assert.Equal(t, fsxWindowsFileServerIn.VolumeConfig, fsxWindowsFileServerOut.VolumeConfig)
+	assert.Equal(t, fsxWindowsFileServerIn.taskARN, fsxWindowsFileServerOut.taskARN)
+	assert.Equal(t, fsxWindowsFileServerIn.executionCredentialsID, fsxWindowsFileServerOut.executionCredentialsID)
+	assert.WithinDuration(t, fsxWindowsFileServerIn.createdAtUnsafe, fsxWindowsFileServerOut.createdAtUnsafe, time.Microsecond)
+	assert.Equal(t, fsxWindowsFileServerIn.desiredStatusUnsafe, fsxWindowsFileServerOut.desiredStatusUnsafe)
+	assert.Equal(t, fsxWindowsFileServerIn.knownStatusUnsafe, fsxWindowsFileServerOut.knownStatusUnsafe)
+}
+
+func TestRetrieveCredentials(t *testing.T) {
```
sharanyad, line 131: A lot of these unit tests can be made table driven. Please change them.

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
sharanyad: what if the creds change after agent restarts? do we plan to pull in the creds from SSM/ASM always?

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+
+var execCommand = exec.Command
+
+func (fv *FSxWindowsFileServerResource) performHostMount(remotePath string, localPath string, username string, password string) error {
+	// formatting to keep powershell happy
+	localPath = fmt.Sprintf("-LocalPath %s", localPath)
+	remotePath = fmt.Sprintf("-RemotePath %s", remotePath)
+	creds := fmt.Sprintf("-Credential $(New-Object System.Management.Automation.PSCredential(%s, $(ConvertTo-SecureString %s -AsPlainText -Force)))", username, password)
+
+	cmd := execCommand("powershell.exe",
```
sharanyad, line 518: could you add comments on what this command does(below one too)

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+		}
+	} else if parsedARNService == "secretsmanager" {
+		err = fv.retrieveASMCredentials(credentialsParameterARN, iamCredentials)
+		if err != nil {
+			seelog.Errorf("Failed to retrieve credentials from asm: %v", err)
+			fv.setTerminalReason(err.Error())
+			return err
+		}
+	} else {
+		err = errors.New("unsupported credentialsParameter ARN, only ssm/secretsmanager ARNs are valid")
```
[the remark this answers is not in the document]
singholt: yes, this is checked from ECS service side

agent/config/parse_windows.go
```diff
@@ -34,14 +34,24 @@ const (
...
+	return parseEnvStatus(envStatus)
+}
+
+// parseFSxWindowsFileServerCapability is used to determine if fsxWindowsFileServer support can be enabled
+func parseFSxWindowsFileServerCapability() bool {
+	envStatus := utils.ParseBool(os.Getenv("ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED"), true)
+	return parseEnvStatus(envStatus)
+}
+
+func parseEnvStatus(envStatus bool) bool {
```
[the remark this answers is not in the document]
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+
+func (fv *FSxWindowsFileServerResource) retrieveCredentials(credentialsParameterARN string, iamCredentials credentials.IAMRoleCredentials) error {
+	parsedARN, err := arn.Parse(credentialsParameterARN)
+	if err != nil {
+		fv.setTerminalReason(err.Error())
+		return err
+	}
+
+	parsedARNService := parsedARN.Service
+	if parsedARNService == "ssm" {
```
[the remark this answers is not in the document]
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: will remove

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+		return err
+	}
+
+	ssmClient := fv.ssmClientCreator.NewSSMClient(fv.region, iamCredentials)
+	ssmParam := filepath.Base(parsedARN.Resource)
+	ssmParams := []string{ssmParam}
+
+	ssmParamMap, err := ssm.GetParametersFromSSM(ssmParams, ssmClient)
+	if err != nil {
+		fv.setTerminalReason(err.Error())
```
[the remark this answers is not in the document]
singholt: removing ```fv.setTerminalReason()```. The parent method returns the info and full error

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+
+	return fv.executionCredentialsID
+}
+
+// SetCredentials safely updates the user credentials of the fsxwindowsfileserver resource
+func (fv *FSxWindowsFileServerResource) SetCredentials(credentials FSxWindowsFileServerCredentials) {
+	fv.lock.Lock()
+	defer fv.lock.Unlock()
+
+	seelog.Debugf("updating credentials for fsxwindowsfileserver resource")
```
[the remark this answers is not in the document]
singholt: will remove

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,664 @@
...
+	}
+
+	fv.handleRootDirectory(volumeConfig.RootDirectory)
+
+	// formatting to keep powershell happy
+	remotePath := fmt.Sprintf(`\\%s\share`, fv.GetFileSystemDNSName())
+	if volumeConfig.RootDirectory != "" {
+		remotePath = fmt.Sprintf(`%s\%s`, remotePath, volumeConfig.RootDirectory)
+	}
+	localPath := volumeConfig.FSxWindowsFileServerHostPath
```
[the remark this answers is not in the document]
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: discussed with @sharanyad offline. before doing a mount, we'll check if the drive letter is available. if not, we'll remove and remount it since the ```New-SmbGlobalMapping``` powershell cmdlet always needs a unoccupied drive letter.

agent/api/task/task_windows.go
```diff
@@ -213,3 +179,72 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+		resourceFields.SSMClientCreator,
+		resourceFields.ASMClientCreator,
+		resourceFields.FSxClientCreator)
+	if err != nil {
+		return err
+	}
+
+	vol.Volume = &fsxwindowsfileserverResource.VolumeConfig
+	task.AddResource(resourcetype.FSxWindowsFileServerKey, fsxwindowsfileserverResource)
+	task.updateContainerVolumeDependency(vol.Name)
```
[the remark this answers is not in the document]
singholt: will do. adding it as part of the existing ```TestInitializeAndAddFSxWindowsFileServerResource```

agent/utils/utils_windows.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return false
+	}
+
+	return matched
+}
+
+// findUnusedDriveLetter is used to search for an available drive letter on the container instance.
+// Reference: https://golang.org/src/os/os_windows_test.go
+func FindUnusedDriveLetter() (string, error) {
```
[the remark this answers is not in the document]
singholt: will add

agent/fsx/fsx.go
```diff
@@ -0,0 +1,56 @@
...
+	var IDs []*string
+	for _, id := range fileSystemIds {
+		IDs = append(IDs, aws.String(id))
+	}
+
+	in := &fsx.DescribeFileSystemsInput{
+		FileSystemIds: IDs,
+	}
+
+	out, err := client.DescribeFileSystems(in)
```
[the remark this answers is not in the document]
singholt: will do

agent/fsx/fsx_test.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+func TestGetFileSystemDNSNames(t *testing.T) {
+	fsxClient := createFSxInterface(fileSystemId, dnsName)
+	out, err := GetFileSystemDNSNames([]string{fileSystemId}, fsxClient)
+	require.NoError(t, err)
+	assert.Equal(t, "test", out[fileSystemId])
+}
+
+func createFSxInterface(fileSystemId string, dnsName string) mockDescribeFileSystems {
```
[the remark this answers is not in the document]
singholt: done

sharanyad:
```
agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_unsupported.go:178:2: redundant return statement (S1023)
Run 'staticcheck -explain <check>' or visit https://staticcheck.io/docs/checks for documentation on checks.
Makefile:287: recipe for target 'static-check' failed
make: *** [static-check] Error 1
```
Could you check the static check failing?

agent/config/parse_windows.go
```diff
@@ -34,14 +34,24 @@ const (
...
+	return isCapable(envStatus)
+}
+
+// parseFSxWindowsFileServerCapability is used to determine if fsxWindowsFileServer support can be enabled
+func parseFSxWindowsFileServerCapability() bool {
+	envStatus := utils.ParseBool(os.Getenv("ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED"), true)
+	return isCapable(envStatus)
+}
+
+func isCapable(envStatus bool) bool {
```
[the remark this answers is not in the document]
singholt: will do

agent/fsx/fsx_test.go
```diff
@@ -0,0 +1,55 @@
...
+	FSxClient
+	Resp fsx.DescribeFileSystemsOutput
+}
+
+func (m mockDescribeFileSystems) DescribeFileSystems(input *fsx.DescribeFileSystemsInput) (*fsx.DescribeFileSystemsOutput, error) {
+	return &m.Resp, nil
+}
+
+func TestGetFileSystemDNSNames(t *testing.T) {
+	fsxClient := mockDescribeFileSystems{
```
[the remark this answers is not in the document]
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,676 @@
...
+}
+
+// Create is used to create all the fsxwindowsfileserver resources for a given task
+func (fv *FSxWindowsFileServerResource) Create() error {
+	var err error
+	var iamCredentials credentials.IAMRoleCredentials
+
+	executionCredentials, ok := fv.credentialsManager.GetTaskCredentials(fv.GetExecutionCredentialsID())
+	if !ok {
+		return errors.New("fsxwindowsfileserver resource: unable to find execution role credentials")
```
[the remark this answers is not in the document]
singholt: will do

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,676 @@
...
+	// remove mount if localPath is not available
+	// handling case where agent crashes after mount and before resource status is updated
+	if !(utils.IsAvailableDriveLetter(localPath)) {
+		// Remove-SmbGlobalMapping cmdlet removes the existing Server Message Block (SMB) mapping between the
+		// container instance and SMB share (FSx for Windows File Server file-system)
+		cmd := execCommand("powershell.exe", "Remove-SmbGlobalMapping", localPathArg, "-Force")
+
+		_, err := cmd.CombinedOutput()
+		if err != nil {
+			seelog.Errorf("Failed to remove existing fsxwindowsfileserver resource mount on the container instance: %v", err)
```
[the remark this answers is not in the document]
singholt: will do

agent/config/parse_windows.go
```diff
@@ -34,14 +34,24 @@ const (
...
+	return checkDomainJoin(envStatus)
+}
+
+// parseFSxWindowsFileServerCapability is used to determine if fsxWindowsFileServer support can be enabled
+func parseFSxWindowsFileServerCapability() bool {
+	envStatus := utils.ParseBool(os.Getenv("ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED"), true)
+	return checkDomainJoin(envStatus)
+}
+
+func checkDomainJoin(envStatus bool) bool {
```
vsiddharth, line 46: Can you rename the method and pass in the option for the env override?

agent/config/parse_windows_test.go
```diff
@@ -28,3 +28,22 @@ func TestParseGMSACapability(t *testing.T) {
...
+	os.Setenv("EXAMPLE_SETTING", "False")
+	assert.False(t, parseBooleanDefaultFalseConfig("EXAMPLE_SETTING").Enabled())
+	assert.False(t, parseBooleanDefaultTrueConfig("EXAMPLE_SETTING").Enabled())
+}
+
+func TestParseFSxWindowsFileServerCapability(t *testing.T) {
+	os.Setenv("ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED", "False")
+	defer os.Unsetenv("ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED")
+
+	assert.False(t, parseFSxWindowsFileServerCapability())
```
vsiddharth, line 48: Please add a test for the True case too.

agent/fsx/fsx.go
```diff
@@ -0,0 +1,55 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "fsx describing filesystem(s) from the service for %s", fileSystemIds)
```
vsiddharth, line 51: nit: fileSystemIds is `[]string`

vsiddharth (changes requested):

agent/config/parse_windows.go
singholt: will do

agent/fsx/fsx.go
singholt: will change to ```%v```

agent/api/task/taskvolume_windows_test.go
```diff
@@ -0,0 +1,144 @@
...
+		"ExecutionStoppedAt": "0001-01-01T00:00:00Z",
+		"SentStatus": "NONE",
+		"StartSequenceNumber": 0,
+		"StopSequenceNumber": 0,
+		"executionCredentialsID": "",
+		"ENI": null,
+		"AppMesh": null,
+		"PlatformFields": %s
+	  }`
+	if runtime.GOOS == "windows" {
```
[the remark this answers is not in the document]
singholt: good catch, will remove check

sharanyad (approved):
please add unit test coverage info for the new packages added. Thanks!

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -0,0 +1,680 @@
...
+	lock                sync.RWMutex
+}
+
+// FSxWindowsFileServerVolumeConfig represents fsxWindowsFileServer volume configuration.
+type FSxWindowsFileServerVolumeConfig struct {
+	FileSystemID  string                         `json:"fileSystemId,omitempty"`
+	RootDirectory string                         `json:"rootDirectory,omitempty"`
+	AuthConfig    FSxWindowsFileServerAuthConfig `json:"authorizationConfig,omitempty"`
+	// FSxWindowsFileServerHostPath is used for bind mount as part of HostConfig.
+	FSxWindowsFileServerHostPath string `json:"fsxWindowsFileServerHostPath"`
```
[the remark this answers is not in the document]
singholt: done

Makefile
```diff
@@ -268,7 +268,7 @@ GOFILES:=$(shell go list -f '{{$$p := .}}{{range $$f := .GoFiles}}{{$$p.Dir}}/{{
 .PHONY: gocyclo
 gocyclo:
 	# Run gocyclo over all .go files
-	gocyclo -over 15 ${GOFILES}
+	gocyclo -over 17 ${GOFILES}
```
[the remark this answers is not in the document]
fierlion: how was this resolved?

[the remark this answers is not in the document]
singholt: this change is to accomodate addition to the ```PostUnmarshalTask()``` in task.go ... no other changes impact gocyclo.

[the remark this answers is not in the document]
fierlion: I see.  So we've added a couple more `if` statements which bumped it from 15-17.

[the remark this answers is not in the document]
singholt: yes

agent/api/task/task.go
fierlion: can this be `fsx` generic?
```
	if task.requiresFsxResource() {
		if err := task.initializeFsxResource(cfg, credentialsManager, resourceFields); err != nil {
```
I assume we'll need to do some of this for FSx Lustre as well, and this clears out Windows specifics in the agnostic file.

singholt: I don't think that's feasible. FSx for Windows File Server and FSx Lustre would be different task resources and this method has initialize() per task resource.

fierlion: `fsxWindowsFileServer` is the product type -- it is technically "agnostic" ie it should apply to both Windows and Linux tasks.  
Please add a comment to this effect.

singholt: done

agent/fsx/factory/factory.go
fierlion: note: there's an agent make target for that: `make goimports`

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserverstatus.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// UnmarshalJSON overrides the logic for parsing the JSON-encoded ResourceStatus data
+func (fs *FSxWindowsFileServerVolumeStatus) UnmarshalJSON(b []byte) error {
+	if strings.ToLower(string(b)) == "null" {
+		*fs = FSxWindowsFileServerVolumeStatusNone
+		return nil
+	}
+
+	if b[0] != '"' || b[len(b)-1] != '"' {
```
fierlion, line 66: is there a less manual check?  
(ie this work for the string `"` and `""`)

fierlion: if you do keep the override, can you update the unit tests to check for edge cases like this?  (Or explain why they will never happen.)

singholt: this is identical to volumestatus.go - https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/volume/volumestatus.go#L60

agent/fsx/factory/factory.go
singholt: I've addressed all goimports issues

vsiddharth (approved):
Changes look good 🚀

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserverstatus.go
fierlion: do you know why that's being done?  Is it correct where you copied it from or does the source have the same potential issues?

fierlion: update -- getting clarification from ecs-agent team about this.

Looks good overall.

fierlion (approved):
Approved -- we'll need to get these tests to all pass now.

fierlion (approved):
update looks good

vsiddharth (approved):
Changes LGTM but please look into the test failures.

vsiddharth (approved):

agent/utils/utils_windows.go
```diff
@@ -0,0 +1,81 @@
...
+	// Do not use A: and B:, because they are reserved for floppy drive.
+	// Do not use C:, because it is normally used for main drive.
```
sandeepindraganti, line 63: You are making 22 system calls (checking driver letters from Z to D ) for every volume , which is a costly operation, Is it not possible to optimize ?   Had we considered the approach of storing the drive letters in use in memory or writing to the ecs_agent json file used to preserve the state , that way you will be able to findout a driver letter available just in one call  which saves a lot of compute ?

agent/api/task/task_windows.go
```diff
@@ -217,3 +183,72 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+// addFSxWindowsFileServerResource creates a fsxwindowsfileserver resource for every fsxwindowsfileserver task volume
+// and updates container dependency
+func (task *Task) addFSxWindowsFileServerResource(
+	cfg *config.Config,
+	credentialsManager credentials.Manager,
+	resourceFields *taskresource.ResourceFields,
+	vol *TaskVolume,
+	fsxWindowsFileServerVol *fsxwindowsfileserver.FSxWindowsFileServerVolumeConfig,
+) error {
+	hostPath, err := utils.FindUnusedDriveLetter()
```
sandeepindraganti, line 228: Can't we just mount the FSx file share to a local folder path instead of using a drive letter which actually enforces restriction on the number of mount volumes ?

agent/api/task/task_windows.go
```diff
@@ -217,3 +183,72 @@ func (task *Task) GetCredentialSpecResource() ([]taskresource.TaskResource, bool
...
+		resourceFields.SSMClientCreator,
+		resourceFields.ASMClientCreator,
+		resourceFields.FSxClientCreator)
+	if err != nil {
+		return err
+	}
+
+	vol.Volume = &fsxwindowsfileserverResource.VolumeConfig
+	task.AddResource(resourcetype.FSxWindowsFileServerKey, fsxwindowsfileserverResource)
+	task.updateContainerVolumeDependency(vol.Name)
```
sandeepindraganti, line 251: Is this call updateContainerVolumeDependency() updates every container about the resource dependency or just few containers ?

singholt: all containers

agent/api/task/task_windows.go
singholt: we cannot mount fsx to a local folder path. It requires a drive letter.

agent/utils/utils_windows.go
singholt: what happens if the instance owner or user uses some drive letter that we assume to be available? static checks are not a feasible approach, hence we need to do this check

agent/api/task/task_windows.go
sandeepindraganti: https://docs.microsoft.com/en-us/windows-server/storage/disk-management/assign-a-mount-point-folder-path-to-a-drive

singholt: ```New-SmbGlobalMapping``` powershell cmdlet enforces using drive letter for ```-LocalPath``` param.

agent/utils/utils_windows.go
sandeepindraganti: Ok , agreed . In that way can't we get all the used drive letters in one system call  and then use one drive letter which is not used ? I am just trying to see how we can avoid too many calls here  and i feel there must be an api to get all used drive letters on windows (like this one powershell Get-PSDrive -PSProvider FileSystem)  , please take a look

mythri-garaga:
Please pull #2655 changes for `ecs/arm/unit_test` test to pass.

agent/config/parse_windows.go
```diff
@@ -71,3 +89,24 @@ func isDomainJoined() (bool, error) {
...
+func isWindows2016() bool {
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Windows NT\CurrentVersion`, registry.QUERY_VALUE)
+	if err != nil {
+		seelog.Errorf("Unable to open Windows registry key to determine Windows version: %v", err)
+	}
+	defer key.Close()
+
+	version, _, err := key.GetStringValue("ProductName")
+	if err != nil {
+		seelog.Errorf("Unable to read current version from Windows registry: %v", err)
```
vsiddharth, line 104: Is it worth checking `version` if there is an error here? Should probably just return `false`

agent/config/parse_windows.go
```diff
@@ -71,3 +89,24 @@ func isDomainJoined() (bool, error) {
 
 	return status == syscall.NetSetupDomainName, nil
 }
+
+// isWindows2016 is used to check if container instance is versioned Windows 2016
+// Reference: https://godoc.org/golang.org/x/sys/windows/registry
+func isWindows2016() bool {
+	key, err := registry.OpenKey(registry.LOCAL_MACHINE, `SOFTWARE\Microsoft\Windows NT\CurrentVersion`, registry.QUERY_VALUE)
+	if err != nil {
+		seelog.Errorf("Unable to open Windows registry key to determine Windows version: %v", err)
```
vsiddharth, line 98: If key read fails, should just return `false`.

agent/config/parse_windows.go
singholt: synced offline.

agent/config/parse_windows.go
singholt: synced offline.

agent/config/parse_windows.go
```diff
@@ -34,14 +37,30 @@ const (
...
 func parseGMSACapability() bool {
 	envStatus := utils.ParseBool(os.Getenv("ECS_GMSA_SUPPORTED"), true)
+	return checkDomainJoinWithEnvOverride(envStatus)
+}
+
+// parseFSxWindowsFileServerCapability is used to determine if fsxWindowsFileServer support can be enabled
+func parseFSxWindowsFileServerCapability() bool {
+	// fsxwindowsfileserver is not supported on Windows 2016 and non-domain-joined container instances
+	status, err := isWindows2016()
+	if err != nil || status == true {
```
vsiddharth, line 47: Shouldn't this be `status == false`?

singholt: I think this should be ```status==true``` which implies that capability will not be advertised if its a windows 2016 container instance.

vsiddharth: Yes, probably worth adding a comment to indicate that this will not work on WS2016

vsiddharth (approved):

agent/config/parse_windows.go
singholt: yes, I agree. I added a comment right above on L45 👍

PR #2568, opened 2020-08-13
Skip TestPortForward integ test on windows

sparrc:
### Summary

This test began failing on windows on 5/15/20 when a new version of windows server core was released: https://hub.docker.com/_/microsoft-windows-servercore

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2569, opened 2020-08-14
Fix a deadlock scenario in envfiles usage

saich:
### Summary
Fixes a deadlock scenario when the agent restores the state from its data file and the tasks are using environment files feature

### Implementation details
The issue happens in some particular cases where the system waits indefinitely to lock a lock, which it itself has locked it, and hasn't unlocked yet, resulting in a deadlock.

### Testing

Bug - Fixed a bug that could cause a deadlock resulting in the agent not functioning as expected

### Licensing


sparrc:
Hello, thank you for the contribution, rather than keeping the deferred unlock and creating a new "unsafe" function, could you change the implementation to just Unlock after the lock is not used anymore? I believe something like this would avoid the deadlock: https://github.com/aws/amazon-ecs-agent/pull/2568/commits/913e3b574297b8d4f2483830836b6d98b51e40dd

saich:
Hey @sparrc, thanks for the feedback. Incorporated your suggestion.

saich:
The tests seems to be failing on Travis CI because of an unrelated issue, and seems like an intermittent issue. Appreciate if anyone can restart the tests.

yunhee-l:
We have created new PR with your fix here: https://github.com/aws/amazon-ecs-agent/pull/2580 to better manage the PR with some testing issues we are seeing at the moment (we are actively working on unblocking it). Please track this fix through the new PR.

PR #2572, opened 2020-08-17
Parse boolean value strings in lowercase to avoid problems on case-insensitive platforms like Windows

### LINKED ISSUE #2571: Windows ECS Agent breaks on settings that take values from boolean environment variables
### Summary
Windows agents where `parseBooleanDefault...` are used fail since https://github.com/aws/amazon-ecs-agent/commit/481a2487c86d9ef6f9a0b3f92a9be7a886db1767#diff-b4442ae5495d1d454858bd0fc134435e 

### Description
When environment variables are set using the form `[Environment]::SetEnvironmentVariable("ECS_ENABLE_AWSLOGS_EXECUTIONROLE_OVERRIDE", $TRUE, "Machine")` -- previously, the settings were parsed into `true` or `false` internally, but now the implementation appears to require a _string_ representation of the values, instead of the Windows internal boolean type. This has broken clusters that were built with agent 1.43.0, and possibly with 1.42.0 as well since no tasks can be scheduled (these workers are missing a required capability)

### Expected Behavior
Using previous configuration (with `$TRUE`) would still work since no breaking change was announced

### Observed Behavior
Services cannot schedule tasks on new instances with the error 

```
service banana_foo was unable to place a task because no container instance met all of its requirements. The closest matching container-instance cc293386c3e34ea48de8e561efae6583 is missing an attribute required by your task
```

### Environment Details
AWS-supplied Windows Server 1909 ECS instances with agent 1.43.0

### Supporting Log Snippets
```
level=warn time=2020-08-17T06:48:20Z msg="Invalid format for \"ECS_ENABLE_AWSLOGS_EXECUTIONROLE_OVERRIDE\", expected a boolean. err invalid character 'T' looking for beginning of value" module=parse.go
```

### Workaround
Change `UserData` or other configuration mechanisms to use the string `true` instead of the Windows environment variable boolean type `$True` or `$TRUE`

nonspecialist:
### Summary
Lowercases theoretically-boolean values before attempting to unmarshal them from environment variables. This fixes #2571 because in Windows, the setting of an environment variable with the `$TRUE` (or `$True`, or `$true`) value results in the string representation `True`, which cannot be parsed as JSON. This breaks existing deployments of Windows-based agents.

It has the side benefit of allowing Python-style truthy values to be parsed as well.

### Implementation details
Using existing `strings` import with `strings.ToLower` around the value of the environment variable when calling `json.Unmarshal`

### Testing

Handle non-lowercase settings for Windows boolean environment variables

### Licensing

### DIFF
--- a/agent/config/parse.go
+++ b/agent/config/parse.go
@@ -208,7 +208,7 @@ func parseBooleanDefaultFalseConfig(envVarName string) BooleanDefaultFalse {
 	boolDefaultFalseCofig := BooleanDefaultFalse{Value: NotSet}
 	configString := os.Getenv(envVarName)
 	if configString != "" {
-		err := json.Unmarshal([]byte(configString), &boolDefaultFalseCofig)
+		err := json.Unmarshal([]byte(strings.ToLower(configString)), &boolDefaultFalseCofig)
 		if err != nil {
 			seelog.Warnf("Invalid format for \"%s\", expected a boolean. err %v", envVarName, err)
 		}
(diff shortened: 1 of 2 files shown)


yunhee-l:
Thank you for submitting this fix. I have created a new PR based on this here: https://github.com/aws/amazon-ecs-agent/pull/2573 
Resolving as duplicate.

PR #2580, opened 2020-08-18
Instamojo fix/envfiles deadlock

yunhee-l:
### Summary
Fixes a deadlock scenario when the agent restores the state from its data file and the tasks are using environment files feature

### Implementation details
The issue happens in some particular cases where the system waits indefinitely to lock a lock, which it itself has locked it, and hasn't unlocked yet, resulting in a deadlock.

### Testing

Bug - Fixed a bug that could cause a deadlock resulting in the agent not functioning as expected

### Licensing


yunhee-l:
copy of https://github.com/aws/amazon-ecs-agent/pull/2569

PR #2581, opened 2020-08-19
eni watcher: retry in first reconciliation loop

fenxiong:
### Summary

Let eni watcher retry acknowledging ENI in its first reconciliation.

Under certain scenario, the ENI is attached before agent starts and connects to ACS, and as a result, when the watcher detects the ENI in the first reconciliation loop we did not acknowledge it. This causes us to wait for the next loop, causing 25s+ delay in acknowledging the ENI. Adding retry fixes the delay.

### Implementation details

Update `reconcileOnce` method to take an argument on whether to retry or not, and retry in the first reconciliation.

### Testing

Added unit test. Manual test has been performed to make sure the delay of confirming ENI attachment is addressed by the change.

N/A

### Licensing

### DIFF
--- a/agent/eni/watcher/watcher_linux.go
+++ b/agent/eni/watcher/watcher_linux.go
@@ -135,7 +135,8 @@ func newWatcher(ctx context.Context,
 
 // Init initializes a new ENI Watcher
 func (udevWatcher *UdevWatcher) Init() error {
-	return udevWatcher.reconcileOnce()
+	// Retry in the first reconciliation, in case the ENI is attached before we connect to ACS.
+	return udevWatcher.reconcileOnce(true)
 }
 
 // Start periodically updates the state of ENIs connected to the system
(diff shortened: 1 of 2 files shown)


agent/eni/watcher/watcher_linux.go
```diff
@@ -337,7 +346,10 @@ func (udevWatcher *UdevWatcher) sendENIStateChangeWithRetries(parentCtx context.
 		sendErr := udevWatcher.sendENIStateChange(macAddress)
 		if sendErr != nil {
 			if _, ok := sendErr.(*unmanagedENIError); ok {
-				log.Debugf("Unable to send state change for unmanaged ENI: %v", sendErr)
+				// This can happen in two scenarios: (1) the ENI is indeed not managed by ECS (i.e. attached manually
+				// by customer); (2) this is an ENI attached by ECS but we have not yet received its information from
+				// ACS.
+				log.Debugf("Not sending state change because the ENI is not managed by ECS: %v", sendErr)
```
yunhee-l, line 352: nit: is "ENI is not managed by ECS" message valid if it's the second case? i think it might create customer confusion.

fenxiong: i updated it to "Not sending state change because we don't know about the ENI". not sure if that sounds better?

yunhee-l: 👍

PR #2583, opened 2020-08-20
Support ipv6 in task networking

fenxiong:
### Summary

Support ipv6 in task networking. On a high level, changes include:
* Update cni plugin input to pass ipv6 address of the eni to the plugin (ipv6 address is already available in acs payload, just need some wiring to pass it to plugin);
* Enable ipv6 in sysctl setting when creating pause container;
* Add ipv6 capability.

This change depends on changes in two cni plugins:
* ecs-eni: https://github.com/aws/amazon-ecs-cni-plugins/pull/98
* vpc-branch-eni: https://github.com/aws/amazon-vpc-cni-plugins/pull/38

In those changes, the input format of the plugin changed from taking single ipv4/ipv6 address to list of ip addresses.

Those changes need to be merged first before this change. Currently, i'm pointing the cni plugin submodules to my fork which contains those changes (in the first commit of the pr, which will be changed to pointing aws repo when merging).

### Implementation details

Detail changes:
1. api/eni: added helper function to retrieve ENI's ip addresses and subnet gateway addresses; added validation on subnet gateway ipv4 address to contain block size - we are already depending on the block size to be in the gateway address (https://github.com/aws/amazon-ecs-agent/blob/master/agent/ecscni/netconfig.go#L155-L159), adding validation here allow later code to depend on this assumption;
2. api/task: added logic to enable ipv6 sysctl setting for the pause container when the task ENI contains ipv6 address. this is needed because docker turns on the disable ipv6 setting by deffault;
3. app: added task eni ipv6 capability for linux
4. ecscni: update cni plugin input format for ecs-eni and vpc-branch-eni plugin to match the format expected by plugin.

### Testing

Unit test added. Manually tested by applying some hack (that looks like [this](https://github.com/fenxiong/amazon-ecs-agent/commit/ed9339dbbae173737dc632e747b699f502fdeb9d)) to make a bridge mode task looks like awsvpc task and verified the ipv6 functionality.

N/A will add when merging feature branch

### Licensing


agent/api/eni/eni.go
```diff
@@ -83,28 +83,49 @@ func (eni *ENI) GetPrimaryIPv4Address() string {
 		}
 	}
 
-	return primaryAddr
+	return eni.getIPAddressWithBlockSize(primaryAddr, true)
```
ofiliz, line 86: We are changing the return value format of this public method. I don't see any callers which expect the old format, so that should be fine, but I would still recommend to update the method name to "GetPrimaryIPv4AddressWithPrefixLength" so the contract is clear. Note the GetIPv4Addresses method doesn't append the prefix length, so without this clarification the contract is inconsistent between these two methods.

agent/api/eni/eni.go
```diff
@@ -83,28 +83,49 @@ func (eni *ENI) GetPrimaryIPv4Address() string {
...
+		addresses = append(addresses, eni.getIPAddressWithBlockSize(addr.Address, false))
 	}
 
 	return addresses
 }
 
-// GetHostname returns the hostname assigned to the ENI
+// GetIPAddresses returns all the ip addresses allocated to the ENI.
+func (eni *ENI) GetIPAddresses() []string {
+	ipv4Addr := eni.GetPrimaryIPv4Address()
```
ofiliz, line 101: Why are we returning only the primary IPv4 address? The method name suggests it is supposed to return all IP addresses. I am guessing the reason is that GetIPv4Addresses doesn't include the prefix length, but that's an implementation detail that shouldn't leak to method behavior.

agent/api/eni/eni.go
```diff
@@ -83,28 +83,49 @@ func (eni *ENI) GetPrimaryIPv4Address() string {
 		}
 	}
 
-	return primaryAddr
+	return eni.getIPAddressWithBlockSize(primaryAddr, true)
 }
 
-// GetIPV6Addresses returns a list of ipv6 addresses allocated to the ENI
+// GetIPV6Addresses returns the ipv6 addresses allocated to the ENI.
```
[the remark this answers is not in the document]
fenxiong: i did not consider this backward breaking because currently the ipv6 address list is empty. is that not the case for fargate?

agent/api/eni/eni.go
fenxiong: method names updated

agent/api/eni/eni.go
fenxiong: removed this and updated to use the method you added

agent/api/eni/eni_test.go
```diff
@@ -27,6 +27,17 @@ const (
...
+	ipv4            = "1.2.3.4"
+	ipv4Gw          = "1.2.3.1"
+	ipv4Block       = "/20"
+	ipv4BlockAddr   = "1.2.0.0"
+	ipv4WithBlock   = ipv4 + ipv4Block
+	ipv4GwWithBlock = ipv4Gw + ipv4Block
+	ipv6            = "abcd:dcba:1234:4321::"
+	ipv6Block       = "/64"
+	ipv6BlockAddr   = "abcd:dcba:1234:4321::"
+	ipv6WithBlock   = ipv6 + ipv6Block
```
ofiliz, line 40: These variable names could be better. ipv4Addr, ipv4Gw, ipv4SubnetPrefixLength, ipv4Subnet, ipv4AddrWithPrefixLength, ipv4GwWithPrefixLength, etc. Basically be consistent with names of the functions you are testing these variables against starting at line 95.

Not super important, but attention to detail is appreciated.

agent/ecscni/types.go
```diff
@@ -119,16 +119,14 @@ type ENIConfig struct {
 	CNIVersion string `json:"cniVersion,omitempty"`
 	// ENIID is the id of ec2 eni
 	ENIID string `json:"eni"`
-	// IPV4Address is the ipv4 of eni
-	IPV4Address string `json:"ipv4-address"`
-	// IPV6Address is the ipv6 of eni
-	IPV6Address string `json:"ipv6-address,omitempty"`
+	// IPAddresses contains the ip addresses of the ENI.
```
ofiliz, line 122: Not a big deal, but it would be nice for this struct to follow the same order as the actual ecs-eni netconf struct fields: eniid, macaddress, ipaddresses, gatewayipaddresses, blockinstancemetadata.

agent/ecscni/netconfig.go
```diff
@@ -130,16 +129,13 @@ func newIPAMConfig(cfg *Config) (IPAMConfig, error) {
...
 	eniConf := ENIConfig{
-		Type:                     ECSENIPluginName,
-		ENIID:                    eni.ID,
-		IPV4Address:              ipv4Addr,
-		IPV6Address:              "",
-		MACAddress:               eni.MacAddress,
-		BlockInstanceMetadata:    cfg.BlockInstanceMetadata,
-		SubnetGatewayIPV4Address: eni.SubnetGatewayIPV4Address,
+		Type:                  ECSENIPluginName,
+		ENIID:                 eni.ID,
```
ofiliz, line 134: Same order of fields comment.

agent/api/task/task.go
```diff
@@ -1535,6 +1540,11 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
 				hostConfig.ExtraHosts = append(hostConfig.ExtraHosts, hosts...)
 			}
 
+			if task.shouldEnableIPv6() {
+				// By default, the disable ipv6 setting is turned on, so need to turn it off to enable it.
+				enableIPv6SysctlSetting(hostConfig)
```
ofiliz, line 1545: Can you please check with @aaithal to make sure this sysctl won't execute on Fargate? I don't know if this package is shared. My understanding is this is to undo the value set by Docker, and Fargate doesn't need it because it uses containerd.

aaithal: Ack. This shouldn't affect Fargate.

agent/api/eni/eni_test.go
fenxiong: that makes sense. updated these names

agent/ecscni/types.go
fenxiong: updated the order here and the one in netconfig.go

PR #2586, opened 2020-08-21
Add LogDriver, LogOptions, and LaunchType to TMDEv4

sparrc:
### Summary

This change adds metadata to the [Task metadata endpoint v4](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html)

Below is a diff of the added information when querying the /task endpoint (`${ECS_CONTAINER_METADATA_URI_V4}/task`). There will be a follow-up PR to add ContainerArn, which will need to be provided by the ECS backend.

```diff
{
  "Cluster": "usw2",
  "TaskARN": "arn:aws:ecs:us-west-2:NNNN:task/NNNN/NNNN",
  "Family": "NNNN",
  "Revision": "4",
  "DesiredStatus": "RUNNING",
  "KnownStatus": "RUNNING",
  "Limits": {
    "CPU": 0.25,
    "Memory": 256
  },
  "PullStartedAt": "2020-08-21T18:18:05.845370929Z",
  "PullStoppedAt": "2020-08-21T18:18:06.561130985Z",
  "AvailabilityZone": "us-west-2a",
+ "LaunchType": "EC2",
  "Containers": [
    {
      "DockerId": "NNNN",
      "Name": "main",
      "DockerName": "ecs-NNNN-NNNN",
      "Image": "NNNN.dkr.ecr.us-west-2.amazonaws.com/NNNN:NNNN",
      "ImageID": "sha256:NNNN",
      "Labels": {
        "com.amazonaws.ecs.cluster": "NNNN",
        "com.amazonaws.ecs.container-name": "main",
        "com.amazonaws.ecs.task-arn": "arn:aws:ecs:us-west-2:NNNN:task/NNNN/NNNN",
        "com.amazonaws.ecs.task-definition-family": "NNNN",
        "com.amazonaws.ecs.task-definition-version": "4"
      },
      "DesiredStatus": "RUNNING",
      "KnownStatus": "RUNNING",
      "Limits": {
        "CPU": 256,
        "Memory": 0
      },
      "CreatedAt": "2020-08-21T18:18:06.576462775Z",
      "StartedAt": "2020-08-21T18:18:07.397380758Z",
      "Type": "NORMAL",
+     "LogDriver": "awslogs",
+     "LogOptions": {
+       "awslogs-group": "NNNN-cluster",
+       "awslogs-region": "us-west-2",
+       "awslogs-stream": "auto-ecs/main/NNNN"
+     },
      "Networks": [
        {
          "NetworkMode": "host",
          "IPv4Addresses": [
            ""
          ]
        }
      ]
    }
  ]
}
```

Add LogDriver, LogOptions, and LaunchType to TMDEv4

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -1072,11 +1072,31 @@ func (c *Container) GetLogDriver() string {
 	hostConfig := &dockercontainer.HostConfig{}
 	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
 	if err != nil {
-		seelog.Warnf("Encountered error when trying to get log driver for container %s: %v", err)
+		seelog.Warnf("Encountered error when trying to get log driver for container %s: %v", c.RuntimeID, err)
 		return ""
 	}
 
 	return hostConfig.LogConfig.Type
+}
+
+// GetLogOptions gets the log 'options' map passed into the task definition.
+// see https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_LogConfiguration.html
+func (c *Container) GetLogOptions() map[string]string {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return map[string]string{}
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encountered error when trying to get log configuration for container %s: %v", c.RuntimeID, err)
+		return map[string]string{}
+	}
+
+	return hostConfig.LogConfig.Config
 }
 
 // GetNetworkModeFromHostConfig returns the network mode used by the container from the host config .
@@ -1092,7 +1112,7 @@ func (c *Container) GetNetworkModeFromHostConfig() string {
 	// TODO return error to differentiate between error and default mode .
 	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
 	if err != nil {
-		seelog.Warnf("Encountered error when trying to get network mode for container %s: %v", err)
+		seelog.Warnf("Encountered error when trying to get network mode for container %s: %v", c.RuntimeID, err)
 		return ""
 	}
 
--- a/agent/handlers/v2/response_test.go
+++ b/agent/handlers/v2/response_test.go
@@ -123,11 +123,115 @@ func TestTaskResponse(t *testing.T) {
 		state.EXPECT().ContainerMapByArn(taskARN).Return(containerNameToDockerContainer, true),
 	)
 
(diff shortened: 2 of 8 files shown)


agent/api/container/container.go
```diff
@@ -1072,13 +1072,33 @@ func (c *Container) GetLogDriver() string {
...
+	defer c.lock.RUnlock()
+
+	if c.DockerConfig.HostConfig == nil {
+		return map[string]string{}
+	}
+
+	hostConfig := &dockercontainer.HostConfig{}
+	err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+	if err != nil {
+		seelog.Warnf("Encountered error when trying to get log configuration for container %s: %v", c.RuntimeID, err)
```
angelcar, line 1095: why not return error here instead of empty map?

agent/handlers/v2/response_test.go
```diff
@@ -123,7 +123,7 @@ func TestTaskResponse(t *testing.T) {
 		state.EXPECT().ContainerMapByArn(taskARN).Return(containerNameToDockerContainer, true),
 	)
 
-	taskResponse, err := NewTaskResponse(taskARN, state, ecsClient, cluster, availabilityZone, containerInstanceArn, false)
+	taskResponse, err := NewTaskResponse(taskARN, state, ecsClient, cluster, availabilityZone, containerInstanceArn, false, false)
```
angelcar, line 126: No tests for `true` path?

agent/handlers/v4/response.go
```diff
@@ -104,6 +104,8 @@ func NewTaskResponse(
 		})
 	}
 
+	v2Resp.LaunchType = "EC2"
```
angelcar, line 107: Wonder if we could use the constant defined here: https://github.com/aws/amazon-ecs-agent/blob/master/agent/ecs_client/model/ecs/api.go#L12642

sparrc: good catch, I will substitute the constant

agent/api/container/container.go
sparrc: This is based on the existing implementation of GetLogDriver. I'm guessing this is because we don't want to do anything more drastic than log and record this error, so since we are already logging it here we can safely just return an empty map.

shubham2892: We can return map, err

sparrc: Im not sure it's useful to return the error once we've already logged it, what else are we going to do with it?

agent/handlers/v2/response_test.go
sparrc: Good catch, that should be tested in the v4 package, will look into adding tests there.

sparrc: this is now addressed in TestTaskResponseWithV4Metadata and TestTaskResponse unit tests in v2 package.

agent/handlers/v4/response.go
sparrc: fixed

angelcar (approved):

PR #2597, opened 2020-08-28
logger: change loglevel behavior

tamor1:
### Summary

Changes behavior of ECS_LOGLEVEL to control the level of logging for both the logging driver and the on instance file unless ECS_LOGLEVEL_ON_INSTANCE is set to override it. ECS_LOGLEVEL_ON_INSTANCE defaults to off if a logging driver is set, or it defaults to the value of ECS_LOGLEVEL if not.

### Implementation details

The behavior for interactions between ECS_LOG_DRIVER, ECS_LOGLEVEL, and ECS_LOGLEVEL_ON_INSTANCE is as follows (default level is info):

Set LOGLEVEL: driver level = ECS_LOGLEVEL, instance level = ECS_LOGLEVEL
Set LOGLEVEL_ON_INSTANCE: driver = default, instance = LOGLEVEL_ON_INSTANCE
Set both: driver = ECS_LOGLEVEL, instance = LOGLEVEL_ON_INSTANCE

Set LOGLEVEL + LOG_DRIVER: driver = ECS_LOGLEVEL, instance = off
Set LOGLEVEL_ON_INSTANCE + LOG_DRIVER: driver ; default, instance = LOGLEVEL_ON_INSTANCE
Set Both + Driver: driver: LOGLEVEL, instance: LOGLEVEL_ON_INSTANCE

Set just LOG_DRIVER: driver = default, instance = off
Set none of them: driver = default, instance = default

### Testing

Added a unit test to test the interactions between the variables above and manually tested all combinations by setting the variables and checking the logs.

### Licensing


agent/logger/log_test.go
```diff
@@ -332,46 +332,109 @@ func TestSeelogConfig_FileLevelDefault(t *testing.T) {
...
-	</outputs>
-	<formats>
-		<format id="logfmt" format="%EcsAgentLogfmt" />
-		<format id="json" format="%EcsAgentJson" />
-		<format id="windows" format="%Msg" />
-	</formats>
-</seelog>`, c)
+	for _, test := range testcases {
+		t.Run(test.name, func(t *testing.T) {
+			resetEnv()
```
fenxiong, line 418: why do you need to reset these here? setenv will override them anyway. instead i think you should have `defer resetEnv()` to make sure the envs are unset when test finished

agent/logger/log_windows_test.go
```diff
@@ -341,47 +341,109 @@ func TestSeelogConfig_FileLevelDefault(t *testing.T) {
...
 
-	Config = &logConfig{
-		logfile:       "foo.log",
-		driverLevel:   DEFAULT_LOGLEVEL,
-		instanceLevel: setInstanceLevelDefault(),
-		RolloverType:  DEFAULT_ROLLOVER_TYPE,
-		outputFormat:  DEFAULT_OUTPUT_FORMAT,
-		MaxFileSizeMB: DEFAULT_MAX_FILE_SIZE,
-		MaxRollCount:  DEFAULT_MAX_ROLL_COUNT,
+	testcases := []struct {
```
fenxiong, line 351: is there any difference between this test and the one in log_test.go? if not, ideally we can rename `log_test.go` to `log_unix_test.go` (since it has build constraint `!windows`), and create another log_test.go that runs on both unix and windows, and put TestSetLevel in that file

agent/logger/log_test.go
```diff
@@ -332,46 +332,109 @@ func TestSeelogConfig_FileLevelDefault(t *testing.T) {
 }
 
 func TestSetLevel(t *testing.T) {
```
fenxiong, line 334: both log_test.go and log_windows_test.go do not have "unit" build tag (e.g. like this https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/container/container_test.go#L1). this tag needs to be added, otherwise none of the tests in these files will be run in travis

tamor1: okay thanks, I added that in

agent/logger/log_test.go
```diff
@@ -1,419 +1,116 @@
...
-// not use this file except in compliance with the License. A copy of the
-// License is located at
-//
-//	http://aws.amazon.com/apache2.0/
-//
-// or in the "license" file accompanying this file. This file is distributed
-// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
-// express or implied. See the License for the specific language governing
-// permissions and limitations under the License.
+// +build unit
```
fenxiong, line 1: missing copy right header in this file

tamor1: ah right, thanks for catching that

Issue #2591, opened 2020-08-27
Json log format not working since v1.44.0 (ECS_LOG_OUTPUT_FORMAT)


adebasi:
### Summary
The output format json, configured via ECS_LOG_OUTPUT_FORMAT, doesn't have any effect from version 1.44.0 on. 

### Description
In version 1.36.0 the output format json was introduced. Users were able to set the format to json by setting the env var ECS_LOG_OUTPUT_FORMAT. From 1.44.0 this seems to be broken and I was not able to find something related to this in the changelog. 

I verified this behaviour by running containers with different versions:
```
$ docker run -e ECS_LOG_OUTPUT_FORMAT=json amazon/amazon-ecs-agent:v1.43.0
{"level": "info", "time": "2020-08-27T07:33:10Z", "msg": "Loading configuration", "module": "agent.go"}
```
```
$ docker run -e ECS_LOG_OUTPUT_FORMAT=json amazon/amazon-ecs-agent:v1.44.0
1598513604929243900 [Error] node must have children
1598513604932176700 [Info] Loading configuration
``` 

I noticed this error because I updated our agents to version 1.44.1 and setting ECS_LOG_OUTPUT_FORMAT to json had no effect.

### Expected Behavior
Set ECS_LOG_OUTPUT_FORMAT to json has the effect that the container logs are in json format.

### Observed Behavior
see description

ubhattacharjya:
Hi @adebasi ,

I am so sorry to hear that you are facing this problem. I am investigating the issue from my end. I will update you with my findings.

Utsa

ubhattacharjya:
Hi @adebasi ,

I could reproduce the error that you faced.

Kindly use the following:

docker run -e ECS_LOG_OUTPUT_FORMAT=json -e ECS_LOGFILE=/log/ecs-agent.log amazon/amazon-ecs-agent:v1.44.0

Let me know if you are still facing the issue

Utsa

ubhattacharjya:
The PR for fixing this issue : https://github.com/aws/amazon-ecs-agent/pull/2598/files

adebasi:
Hi @ubhattacharjya,
Thank you for the quick response and fixing the bug.

ellenthsu:
This fix has been released as part of ECS Agent 1.44.4: https://github.com/aws/amazon-ecs-agent/releases/tag/1.44.4

[...]

PR #2598, opened 2020-08-28
Add the "<filter></filter>" only when logfile environment Variable is set

ubhattacharjya:
### Summary

This PR fixes the empty (filter)(/filter) when ECS_LOGFILE is not set by ecs-init. This would cause certain environment variables related to agent logging to not have their expected behaviour when the user is not using ECS optimized AMI. 
This issue is mentioned in https://github.com/aws/amazon-ecs-agent/issues/2591

Problem:

For example, 
```
$ docker run -e ECS_LOG_OUTPUT_FORMAT=json amazon/amazon-ecs-agent:v1.44.0
1598513604929243900 [Error] node must have children
```
This error is caused because ecs agent is receiving empty (filter)(/filter) when ECS_LOGFILE is not set by ecs-init.

However, when the below is run, everything is fine.
```
docker run -e ECS_LOG_OUTPUT_FORMAT=json --env=ECS_LOGFILE=/log/ecs-agent.log amazon/amazon-ecs-agent:v1.44.0
{"level": "info", "time": "2020-08-28T21:08:40Z", "msg": "Loading configuration", "module": "agent.go"}
```

### Implementation details

This PR adds the filter block only when ECS_LOGFILE is present as env var. Otherwise it is not added

### Testing

Unit tests cover the changes
Manual test:

```
[ec2-user@ip-172-31-3-170 ~]$ docker run -e ECS_LOG_OUTPUT_FORMAT=json amazon/amazon-ecs-agent:latest
{"level": "info", "time": "2020-08-28T22:24:32Z", "msg": "Loading configuration", "module": "agent.go"}
```

Also tested that the /var/log/ecs/ecs-agent.log file also got populated in json format when ECS_LOG_OUTPUT_FORMAT=json is mentioned ecs.config file
### Licensing

### DIFF
--- a/agent/logger/log.go
+++ b/agent/logger/log.go
@@ -86,9 +86,10 @@ func seelogConfig() string {
 			<console />`
 	c += platformLogConfig()
 	c += `
-		</filter>
-		<filter levels="` + getLevelList(Config.instanceLevel) + `">`
+		</filter>`
 	if Config.logfile != "" {
+		c += `
+		<filter levels="` + getLevelList(Config.instanceLevel) + `">`
 		if Config.RolloverType == "size" {
 			c += `
 			<rollingfile filename="` + Config.logfile + `" type="size"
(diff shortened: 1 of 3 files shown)


mssrivas:
Can we add a test/unit test to make sure this works as we think it should?

ubhattacharjya:
> Can we add a test/unit test to make sure this works as we think it should?

added unit tests for testing a scenario when logfile is not mentioned

PR #2607, opened 2020-09-02
add TopContainer to dockerapi

fierlion:
### Summary
This adds the `TopContainer` function to the dockerapi -- this is essentially `docker top`.
The naming (`TopContainer` as opposed to `ContainerTop`) is to be inline with the convention in the dockerapi which differentiates it from the core Docker `ContainerTop` function.

### Implementation details
The changes follow the existing pattern to add the function to the dockerapi interface for mocking, add specific timeout and error, as well as tweak the `TopContainer` function signature to use default configuration and allow for a timeout to be set.

### Testing
I've added unit tests, following the convention in docker_client_test
I tested manually by adding a call to `TopDocker` with a print statement in the task engine code then running a task with via the console and again via integration tests.  I was able to exercise the new function to get the following output for a simple healthcheck container:
```
&{[[root 6759 6725 0 19:21 ? 00:00:00 sh -c trap "exit 0" SIGTERM SIGINT; while true; do sleep 1; done] [root 6799 6759 0 19:21 ? 00:00:00 sleep 1]] [UID PID PPID C STIME TTY TIME CMD]}
```

Note: integration tests for this and for `docker exec` functionality will be a separate PR.

Add `TopDocker` command to dockerapi

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -133,6 +133,10 @@ type DockerClient interface {
 	// provided for the request.
 	InspectContainer(context.Context, string, time.Duration) (*types.ContainerJSON, error)
 
+	// TopContainer returns information about the top processes running in the specified container.  A timeout value and a context
+	// should be provided for the request.
+	TopContainer(context.Context, string, time.Duration) (*dockercontainer.ContainerTopOKBody, error)
+
 	// ListContainers returns the set of containers known to the Docker daemon. A timeout value and a context
 	// should be provided for the request.
 	ListContainers(context.Context, bool, time.Duration) ListContainersResponse
@@ -652,6 +656,45 @@ func (dg *dockerGoClient) inspectContainer(ctx context.Context, dockerID string)
 	}
 	containerData, err := client.ContainerInspect(ctx, dockerID)
 	return &containerData, err
+}
+
+func (dg *dockerGoClient) TopContainer(ctx context.Context, dockerID string, timeout time.Duration) (*dockercontainer.ContainerTopOKBody, error) {
+	type topResponse struct {
+		top *dockercontainer.ContainerTopOKBody
+		err error
+	}
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	defer metrics.MetricsEngineGlobal.RecordDockerMetric("TOP_CONTAINER")()
+	// Buffered channel so in the case of timeout it takes one write, never gets
+	// read, and can still be GC'd
+	response := make(chan topResponse, 1)
+	go func() {
+		top, err := dg.topContainer(ctx, dockerID)
+		response <- topResponse{top, err}
+	}()
+
+	// Wait until we get a response or for the 'done' context channel
+	select {
+	case resp := <-response:
+		return resp.top, resp.err
+	case <-ctx.Done():
+		err := ctx.Err()
+		if err == context.DeadlineExceeded {
+			return nil, &DockerTimeoutError{timeout, "listing top"}
+		}
+
+		return nil, &CannotGetContainerTopError{err}
+	}
+}
+
+func (dg *dockerGoClient) topContainer(ctx context.Context, dockerID string) (*dockercontainer.ContainerTopOKBody, error) {
+	client, err := dg.sdkDockerClient()
+	if err != nil {
+		return nil, err
+	}
+	topResponse, err := client.ContainerTop(ctx, dockerID, []string{})
+	return &topResponse, err
 }
 
 func (dg *dockerGoClient) StopContainer(ctx context.Context, dockerID string, timeout time.Duration) DockerContainerMetadata {
--- a/agent/dockerclient/dockerapi/docker_client_test.go
+++ b/agent/dockerclient/dockerapi/docker_client_test.go
@@ -606,6 +606,39 @@ func TestInspectContainer(t *testing.T) {
 	assert.True(t, reflect.DeepEqual(&containerOutput, container))
 }
 
+func TestTopContainerTimeout(t *testing.T) {
+	mockDockerSDK, client, _, _, _, done := dockerClientSetup(t)
+	defer done()
+
+	wait := &sync.WaitGroup{}
+	wait.Add(1)
+	mockDockerSDK.EXPECT().ContainerTop(gomock.Any(), "id", gomock.Any()).Do(func(ctx context.Context, x interface{}, y interface{}) {
+		wait.Wait()
+	}).MaxTimes(1).Return(dockercontainer.ContainerTopOKBody{}, nil)
+
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	_, err := client.TopContainer(ctx, "id", xContainerShortTimeout)
+	assert.Error(t, err, "Expected error for top timeout")
+	assert.Equal(t, "DockerTimeoutError", err.(apierrors.NamedError).ErrorName())
+	wait.Done()
+}
+
+func TestTopContainer(t *testing.T) {
+	mockDockerSDK, client, _, _, _, done := dockerClientSetup(t)
+	defer done()
+
+	topOutput := dockercontainer.ContainerTopOKBody{}
+	gomock.InOrder(
+		mockDockerSDK.EXPECT().ContainerTop(gomock.Any(), "id", gomock.Any()).Return(topOutput, nil),
+	)
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	topResponse, err := client.TopContainer(ctx, "id", dockerclient.TopContainerTimeout)
+	assert.NoError(t, err)
+	assert.Equal(t, &topOutput, topResponse)
+}
+
 func TestContainerEvents(t *testing.T) {
 	mockDockerSDK, client, _, _, _, done := dockerClientSetup(t)
 	defer done()
--- a/agent/dockerclient/timeout.go
+++ b/agent/dockerclient/timeout.go
@@ -38,6 +38,9 @@ const (
 	ListContainersTimeout = 10 * time.Minute
 	// InspectContainerTimeout is the timeout for the InspectContainer API.
 	InspectContainerTimeout = 30 * time.Second
+	// TopContainerTimeout is the timeout for the TopContainer API.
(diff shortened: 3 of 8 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -654,6 +658,45 @@ func (dg *dockerGoClient) inspectContainer(ctx context.Context, dockerID string)
...
 }
 
+func (dg *dockerGoClient) TopContainer(ctx context.Context, dockerID string, timeout time.Duration) (*dockercontainer.ContainerTopOKBody, error) {
+	type topResponse struct {
+		top *dockercontainer.ContainerTopOKBody
+		err error
+	}
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+	defer metrics.MetricsEngineGlobal.RecordDockerMetric("CONTAINER_TOP")()
```
angelcar, line 668: can we change this from CONTAINER_TOP to TOP_CONTAINER?

agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -606,6 +606,39 @@ func TestInspectContainer(t *testing.T) {
...
+
+	topOutput := dockercontainer.ContainerTopOKBody{}
+	gomock.InOrder(
+		mockDockerSDK.EXPECT().ContainerTop(gomock.Any(), "id", gomock.Any()).Return(topOutput, nil),
+	)
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	topResponse, err := client.TopContainer(ctx, "id", dockerclient.TopContainerTimeout)
+	assert.NoError(t, err)
+	assert.True(t, reflect.DeepEqual(&topOutput, topResponse))
```
angelcar, line 639: `assert.Equal` slready does `reflect.DeepEqual` with the added benefit of a diff printed for debugging purposes

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -133,6 +133,10 @@ type DockerClient interface {
 	// provided for the request.
 	InspectContainer(context.Context, string, time.Duration) (*types.ContainerJSON, error)
 
+	// TopContainer returns information about the top processes running in the specified container.  A timeout value and a context
+	// should be provided for the request.
+	TopContainer(context.Context, string, time.Duration) (*dockercontainer.ContainerTopOKBody, error)
```
angelcar, line 138: where is `container.ContainerTopOKBody` defined?

fierlion: https://github.com/docker/go-docker/blob/master/api/types/container/container_top.go

agent/dockerclient/dockerapi/docker_client_test.go
fierlion: I'll make that change

agent/dockerclient/dockerapi/docker_client.go
fierlion:
```
[amazon-ecs-agent]$ grep -r INSPECT_CONTAINER .
Binary file ./out/amazon-ecs-agent matches
./agent/dockerclient/dockerapi/docker_client.go:        defer metrics.MetricsEngineGlobal.RecordDockerMetric("INSPECT_CONTAINER")()
```
CONTAINER_INSPECT -> INSPECT_CONTAINER
which means
CONTAINER_TOP -> TOP_CONTAINER

yes we can.  I'll make that change.

angelcar (approved):

agent/dockerclient/timeout.go
```diff
@@ -38,6 +38,9 @@ const (
 	ListContainersTimeout = 10 * time.Minute
 	// InspectContainerTimeout is the timeout for the InspectContainer API.
 	InspectContainerTimeout = 30 * time.Second
+	// TopContainerTimeout is the timeout for the TopContainer API.
+	TopContainerTimeout = 10 * time.Second
```
sharanyad, line 42: how did we come up with this value?

fierlion: To be honest, I made an assumption based on https://godoc.org/docker.io/go-docker#Client.ContainerTop being similar to https://godoc.org/docker.io/go-docker#Client.ContainerInspect.  The 30 second timeout seems extensive for InspectContainerTimeout so for TopContainerTimeout I set it to 10 seconds.

Would you prefer to see profiling of the TopContainer command?
Otherwise we can change this if we find a reason.

sharanyad: Some kind of profiling is usually preferable for these timeouts. 
10 seconds seem too low to me, so looking to avoid surprises.
 timeouts usually some x times more than average api call latencies.

fierlion: I'll update with profiling info

fierlion: I added some simple print statements to the integ tests and ran them to capture the timing of the `docker top` api call:
```
[Info] ###TEST### start: 2020-09-04 00:15:15.162484886 +0000 UTC m=+0.764416624
[Info] ###TEST### topContainerOut: &{[[root 5181 5119 0 00:15 ? 00:00:00 sleep 5]] [UID PID PPID C STIME TTY TIME CMD]}
topErr: <nil>
[Info] ###TEST### end: 2020-09-04 00:15:15.180661484 +0000 UTC m=+0.782593205
[Info] ###TEST### elapsed: 18.176581ms
```
```
[Info] ###TEST### start: 2020-09-04 00:15:15.222661243 +0000 UTC m=+0.824592981
[Info] ###TEST### topContainerOut: &{[[root 5263 5199 0 00:15 ? 00:00:00 sleep 5]] [UID PID PPID C STIME TTY TIME CMD]}
topErr: <nil>
[Info] ###TEST### end: 2020-09-04 00:15:15.234055193 +0000 UTC m=+0.835986926
[Info] ###TEST### elapsed: 11.393945ms
```
```
[Info] ###TEST### start: 2020-09-04 00:15:15.268344148 +0000 UTC m=+0.870275896
[Info] ###TEST### topContainerOut: &{[[root 5234 5154 0 00:15 ? 00:00:00 sleep 5]] [UID PID PPID C STIME TTY TIME CMD]}
topErr: <nil>
[Info] ###TEST### end: 2020-09-04 00:15:15.278161029 +0000 UTC m=+0.880092750
[Info] ###TEST### elapsed: 9.816854ms
```
`[Info] ###TEST### elapsed: 12.350418ms`
`[Info] ###TEST### elapsed: 13.807724ms`
`[Info] ###TEST### elapsed: 10.075021ms`
`[Info] ###TEST### elapsed: 11.396256ms`
`[Info] ###TEST### elapsed: 11.115756ms`
...
Over 100 executions my max was ~19ms and min was ~10ms for each `TopContainer` call.

Is there a specific context you're worried about this timing out in 10 seconds?

sharanyad: yes, when there are multiple tasks/containers running simultaneously in the instance, docker could be stressed(inspect/start/stop containers etc), and 10s timeout could be hit faster. 
when the above profiling done was done, was docker stressed enough with other operations?

fierlion: I wrote a quick bash script to run a bunch of docker api calls to top then ran it in a bunch of tmux sessions on the instance where I also ran the top test:
```
#!/bin/bash
# this calls docker top repeatedly for a specific container
# arg 0 should be some container id from `docker ps`
# arg 1 should be a short enough time interval to cause docker pain
args=("$@")
while true; do
  docker top ${args[0]}
  sleep ${args[1]}
done
```
use: `./docker_stressor.sh <containerid> <interval>`
and engine integ tests run using `go test -tags integration -timeout=5m -v ./agent/engine/... > output`
running integ tests side-by-side with multiple tmux sessions (aka "stressors") having following configs and output:

with 2 stressors running docker top at .5 second interval:
max  `[Info] ###TEST### elapsed: 28.333578ms`
with 10 stressors running `docker top` at a .05 second interval:
max `[Info] ###TEST### elapsed: 52.632767ms`
with 10 stressors running docker top at .01 second interval:
max `[Info] ###TEST### elapsed: 89.34588ms`
with 10 stressors running docker top at .001 second interval:
max `[Info] ###TEST### elapsed: 91.587552ms`
with 10 stressors running docker top at .000001 second interval
max `[Info] ###TEST### elapsed: 86.837214ms`

I will note that there might be a limit to the `sleep` interval that I'm using -- I can't find anything in the man page (which points to `info coreutils 'sleep invocation'`)

Despite the above, I'll set the timeout to 30 seconds to be in line with the inspect container timeout to be extra-conservative.

agent/stats/task_linux_test.go
```diff
@@ -21,13 +21,12 @@ import (
...
 	"time"
 
-	mock_nswrapper "github.com/aws/amazon-ecs-agent/agent/utils/nswrapper/mocks"
-
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
 	mock_netlink "github.com/aws/amazon-ecs-agent/agent/eni/netlinkwrapper/mocks"
 	mock_resolver "github.com/aws/amazon-ecs-agent/agent/stats/resolver/mock"
+	mock_nswrapper "github.com/aws/amazon-ecs-agent/agent/utils/nswrapper/mocks"
```
fierlion, line 29: NOTE: I tried to alphabetize these (ie put `mock_nswrapper` before `mock_resolver`, but `go fmt` switches them back to this.  I'll defer to `go fmt` -- it's probably sorting by path.

agent/dockerclient/timeout.go
sharanyad: lgtm. thanks for testing

sharanyad (approved):

agent/stats/task_linux_test.go
angelcar: did you try goimports? I believe that's what we're using?

angelcar (approved):

agent/stats/task_linux_test.go
fierlion: I ran `make gogenerate` and `make goimports` then `go fmt ./agent/...` 
this is what came out the other end of that ^^

PR #2608, opened 2020-09-03
task engine: centralize getting docker ID, check it's not blank

sparrc:
# Summary

In 5 different places we were mapping from our internal apicontainer object to the dockercontainer object in order to obtain the docker ID, though never in exactly the same way. This centralizes it into a single function (getDockerID) that also tries to first just grab the dockerID from the apicontainer object if it exists.

Will also fallback to the DockerName attribute on the dockercontainer obj if DockerID doesnt exist.

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -487,17 +487,12 @@ func (engine *DockerTaskEngine) synchronizeContainerStatus(container *apicontain
 // their state to the managed task's container channel.
 func (engine *DockerTaskEngine) checkTaskState(task *apitask.Task) {
 	defer metrics.MetricsEngineGlobal.RecordTaskEngineMetric("CHECK_TASK_STATE")()
-	taskContainers, ok := engine.state.ContainerMapByArn(task.Arn)
-	if !ok {
-		seelog.Warnf("Task engine [%s]: could not check task state; no task in state", task.Arn)
-		return
-	}
 	for _, container := range task.Containers {
-		dockerContainer, ok := taskContainers[container.Name]
-		if !ok {
+		dockerID, err := engine.getDockerID(task, container)
+		if err != nil {
 			continue
 		}
-		status, metadata := engine.client.DescribeContainer(engine.ctx, dockerContainer.DockerID)
+		status, metadata := engine.client.DescribeContainer(engine.ctx, dockerID)
 		engine.tasksLock.RLock()
 		managedTask, ok := engine.managedTasks[task.Arn]
 		engine.tasksLock.RUnlock()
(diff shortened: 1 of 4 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1312,43 +1299,27 @@ func (engine *DockerTaskEngine) buildCNIConfigFromTaskContainer(
 	return cniConfig, nil
 }
 
-func (engine *DockerTaskEngine) inspectContainerByName(taskArn, containerName string) (*types.ContainerJSON, error) {
-	containers, ok := engine.state.ContainerMapByArn(taskArn)
-	if !ok {
-		return nil, errors.New("engine: failed to find the pause container, no containers in the task")
```
sparrc, line 1318: also fixes this inaccurate error message, seems like maybe this function was originally written to only inspect the pause container?

agent/engine/docker_task_engine.go
```diff
@@ -1312,43 +1299,27 @@ func (engine *DockerTaskEngine) buildCNIConfigFromTaskContainer(
...
-	}
-
-	pauseContainer, ok := containers[containerName]
-	if !ok {
-		return nil, errors.New("engine: failed to find the pause container")
+func (engine *DockerTaskEngine) inspectContainer(taskArn, container *apicontainer.Container) (*types.ContainerJSON, error) {
```
fenxiong, line 1302: taskArn should be a string?

agent/engine/docker_task_engine.go
```diff
@@ -1497,3 +1463,24 @@ func getContainerHostIP(networkSettings *types.NetworkSettings) (string, bool) {
...
 }
+
+func (engine *DockerTaskEngine) getDockerID(task *apitask.Task, container *apicontainer.Container) (string, error) {
+	runtimeID := container.GetRuntimeID()
+	if runtimeID != "" {
+		return runtimeID, nil
+	}
+	containerMap, ok := engine.state.ContainerMapByArn(task.Arn)
+	if !ok {
+		return "", errors.Errorf("Container name=%s belongs to unrecognized task taskArn=%s", container.Name, task.Arn)
```
fenxiong, line 1474: error message should start with lower case, same below

agent/engine/docker_task_engine.go
sparrc: fixed

agent/engine/docker_task_engine.go
```diff
@@ -1312,43 +1299,27 @@ func (engine *DockerTaskEngine) buildCNIConfigFromTaskContainer(
...
-	containers, ok := engine.state.ContainerMapByArn(taskArn)
-	if !ok {
-		return nil, errors.New("engine: failed to find the pause container, no containers in the task")
-	}
-
-	pauseContainer, ok := containers[containerName]
-	if !ok {
-		return nil, errors.New("engine: failed to find the pause container")
+func (engine *DockerTaskEngine) inspectContainer(taskArn, container *apicontainer.Container) (*types.ContainerJSON, error) {
+	dockerID, err := engine.getDockerID(task, container)
```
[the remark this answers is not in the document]
sparrc: Yes it's true, getDockerID could sometimes return the name, I'm not sure how to call the function to indicate that.

You are correct that this would change to mostly calling the docker API using the docker runtime ID. The docker API for the most part is very liberal in what it accepts in terms of this. You can pass in the name, full runtime id, short name, or short runtime ID. That is part of why I wanted to make this change, so that it is more standardized on using a single type of "docker id".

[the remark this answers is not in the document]
sparrc: for example, previous to this PR we were called RemoveContainer using DockerName, but if you look at the implementation it is referred to as DockerID in our code, and then ultimately containerID in the docker client codebase: https://github.com/aws/amazon-ecs-agent/blob/1ebf0604c13013596cfd4eb239574a85890b13e8/agent/dockerclient/dockerapi/docker_client.go#L701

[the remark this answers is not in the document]
angelcar: I see. I'd prefer to either only use DockerConainer everywhere or to add these attributes directly to the container struct. But that's a much larger change. I guess this is a good iteration

angelcar (approved):

fenxiong (approved):

PR #2616, opened 2020-09-05
Add ExecAgent bind mounts to all containers of exec-enabled tasks

angelcar:
### Summary

This adds the necessary volumes and bind mounts to enable exec agent for all the containers of an exec-enabled task. 

### Implementation details

Most task initialization happens in the task.PostUnmarshalTask method. This is where I added the initialization steps for exec agent.

By appending the required volumes to the task and the mount points to the task containers, we get persistence of these exec agent volumes for free, as part of the regular operation of the agent, just like any other volumes associated with the task.

The directory for the exec agent log volumes can be configured via a new configuration parameter called `ECS_EXEC_AGENT_LOG_DIR` (which defaults to `/var/log/ecs/exec`). Logs for exec agent are stored in individual directories within `ECS_EXEC_AGENT_LOG_DIR`; the name of this directory is the taskId plus the container name (e.g. `/var/log/ecs/exec/cc1330bb-58a2-4eb3-aef0-c85740a7f22a-scratchsleeper`). **This is different than the specification in the DP design doc**. This is because the DP design doc says the container's runtime id should be used for the directory name, but it's a chicken and egg problem because to know the runtime id we need to run the container, but before running it, we need to bindmount the directory, so it's not possible to have a bind mount with the runtime id, this is the reason I use `task id+container name`.

Finally, some changes in this CR are not definitive, but I added TODOs where appropriate in order to keep iterating on the code. 

### Testing

Aside from UTs, I tested this code manually by starting a task with two containers in a personal container instance where I have temporary static ssm agent binaries.

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -266,6 +266,9 @@ type Container struct {
 	// the JSON body while saving the state
 	SteadyStateStatusUnsafe *apicontainerstatus.ContainerStatus `json:"SteadyStateStatus,omitempty"`
 
+	// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
+	ExecCommandAgentMetadata ExecCommandAgentMetadata `json:"execCommandAgentMetadata"`
+
 	createdAt  time.Time
 	startedAt  time.Time
 	finishedAt time.Time
@@ -323,6 +326,12 @@ type Secret struct {
 	Type          string `json:"type"`
 	Provider      string `json:"provider"`
 	Target        string `json:"target"`
+}
+
+// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
+type ExecCommandAgentMetadata struct {
+	PID          string `json:"pid"`
+	DockerExecID string `json:"dockerExecId"`
 }
 
 // GetSecretResourceCacheKey returns the key required to access the secret
@@ -1146,4 +1155,16 @@ func (c *Container) GetTaskARN() string {
 	defer c.lock.RUnlock()
 
 	return c.TaskARNUnsafe
+}
+
+func (c *Container) SetExecCommandAgentMetadata(metadata ExecCommandAgentMetadata) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.ExecCommandAgentMetadata = metadata
+}
+
+func (c *Container) GetExecCommandAgentMetadata() ExecCommandAgentMetadata {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.ExecCommandAgentMetadata
 }
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -259,6 +259,9 @@ type Task struct {
 	// have a value for this). This field should be accessed via GetLocalIPAddress and SetLocalIPAddress.
 	LocalIPAddressUnsafe string `json:"LocalIPAddress,omitempty"`
 
+	// TODO: [ecs-exec] Wire this to the actual model when control plane changes are ready
+	ExecCommandAgentEnabled bool `json:"ExecCommandAgentEnabled"`
+
 	// lock is for protecting all fields in the task struct
 	lock sync.RWMutex
 }
@@ -338,9 +341,15 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	task.initSecretResources(credentialsManager, resourceFields)
 
 	task.initializeCredentialsEndpoint(credentialsManager)
+
 	// NOTE: initializeVolumes needs to be after initializeCredentialsEndpoint, because EFS volume might
 	// need the credentials endpoint constructed by it.
 	if err := task.initializeVolumes(cfg, dockerClient, ctx); err != nil {
+		return err
+	}
+
+	if err := task.initializeExecCommandAgentResources(cfg); err != nil {
+		seelog.Errorf("Task [%s]: could not initialize exec agent: %v", task.Arn, err)
 		return err
 	}
 
@@ -478,7 +487,6 @@ func (task *Task) initializeDockerLocalVolumes(dockerClient dockerapi.DockerClie
 					resourcestatus.ResourceStatus(taskresourcevolume.VolumeCreated),
 					apicontainerstatus.ContainerPulled)
 				requiredLocalVolumes = append(requiredLocalVolumes, mountPoint.SourceVolume)
-
 			}
 		}
 	}
@@ -2691,4 +2699,10 @@ func (task *Task) SetLocalIPAddress(addr string) {
 	defer task.lock.Unlock()
 
 	task.LocalIPAddressUnsafe = addr
+}
+
+func (task *Task) IsExecCommandAgentEnabled() bool {
+	task.lock.Lock()
+	defer task.lock.Unlock()
+	return task.ExecCommandAgentEnabled
 }
--- a/agent/api/task/task_linux_test.go
+++ b/agent/api/task/task_linux_test.go
@@ -17,20 +17,27 @@ package task
 
 import (
 	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	apiappmesh "github.com/aws/amazon-ecs-agent/agent/api/appmesh"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	apieni "github.com/aws/amazon-ecs-agent/agent/api/eni"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
+	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/asmsecret"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/cgroup/control/mock_control"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/firelens"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/ssmsecret"
+	taskresourcevolume "github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
 	mock_ioutilwrapper "github.com/aws/amazon-ecs-agent/agent/utils/ioutilwrapper/mocks"
 	"github.com/golang/mock/gomock"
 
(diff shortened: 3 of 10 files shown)


agent/api/task/task.go
```diff
@@ -140,6 +140,22 @@ const (
 
 	// specifies awsvpc type mode for a task
 	AWSVPCNetworkMode = "awsvpc"
+
+	internalExecAgentNamePrefix          = "__internal__exec-agent"
```
sharanyad, line 144: any reason why double underscores?

agent/api/task/task.go
```diff
@@ -562,6 +586,83 @@ func (task *Task) initializeEFSVolumes(cfg *config.Config, dockerClient dockerap
...
+
+	tId, err := task.GetID()
+	if err != nil {
+		return err
+	}
+
+	execAgentBinNames := []string{execAgentBinName, execAgentSessionWorkerBinName, execAgentSessionLoggerBinName}
+
+	// Append exec agent binaries to task volumes
+	for _, bn := range execAgentBinNames {
```
sharanyad, line 608: can just call it 'bin' instead of 'bn'

agent/api/task/task.go
```diff
@@ -562,6 +586,83 @@ func (task *Task) initializeEFSVolumes(cfg *config.Config, dockerClient dockerap
...
+			Type: HostVolumeType,
+			Name: internalExecAgentCertVolumeName,
+			Volume: &taskresourcevolume.FSHostVolume{
+				FSSourcePath: execAgentHostCertFile,
+			},
+		})
+
+	// Add log volumes and mount points to all containers in this task
+	for _, c := range task.Containers {
+		logSuffix := fmt.Sprintf("%s-%s", tId, c.Name)
```
sharanyad, line 631: I believe this creates log dirs roughly as - 
```
../exec/taskID1-container1 
../exec/taskID1-container2 
../exec/taskID2-container1 
```

instead why not just create a parent dir for task and have container logs under it? 

something like:
```
../exec/taskID1/container1 
../exec/taskID1/container2 
../exec/taskID2/container1 
```

agent/config/config.go
```diff
@@ -570,6 +570,7 @@ func environmentConfig() (Config, error) {
 		SpotInstanceDrainingEnabled:         parseBooleanDefaultFalseConfig("ECS_ENABLE_SPOT_INSTANCE_DRAINING"),
 		GMSACapable:                         parseGMSACapability(),
 		VolumePluginCapabilities:            parseVolumePluginCapabilities(),
+		ExecAgentLogDir:                     os.Getenv("ECS_EXEC_AGENT_LOG_DIR"),
```
sharanyad, line 573: can you add a TODO for later to add this env var to README?

agent/api/task/task.go
```diff
@@ -562,6 +586,83 @@ func (task *Task) initializeEFSVolumes(cfg *config.Config, dockerClient dockerap
 	return nil
 }
 
+func buildVolumeNameForExecAgentBinary(binaryName string) string {
+	return fmt.Sprintf("%s-%s", internalExecAgentNamePrefix, binaryName)
+}
+
+// initializeExecAgentResources specifies the necessary volumes and mount points in all of the task containers in order for the
+// exec agent to run upon container start.
+func (task *Task) initializeExecAgentResources(cfg *config.Config) error {
```
sharanyad, line 595: this should be in task_linux.go file (since its linux only feature) same for the unit tests

agent/api/task/task.go
```diff
@@ -140,6 +140,22 @@ const (
 
 	// specifies awsvpc type mode for a task
 	AWSVPCNetworkMode = "awsvpc"
+
+	internalExecAgentNamePrefix          = "__internal__exec-agent"
+	internalExecAgentLogVolumeNamePrefix = internalExecAgentNamePrefix + "-log"
+	internalExecAgentCertVolumeName      = internalExecAgentNamePrefix + "-tls-cert"
+
+	// TODO: [ecs-exec] decide if this needs to be configurable or put in a specific place in our optimized AMIs
+	execAgentHostBinDir           = "/home/ec2-user/ssm-agent/linux_amd64"
```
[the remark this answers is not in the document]
angelcar: Yep, I'm aware, however this is not defined yet, so I'd like to merge this and iterate on it when we're ready; thus the TODO.

agent/api/task/task.go
angelcar: I just wanted something unique that won't collide with other volume names. do you have suggestions?

agent/api/task/task.go
angelcar: bn stands for binary name. I am following the style guide here https://github.com/golang/go/wiki/CodeReviewComments#variable-names

agent/api/task/task.go
angelcar: I'd be OK with that. I'd like to hear Ray's opinion on this as well since this was a deviation from the spec in the design doc.

fierlion: The design doc spec assumed that we would have container id at the time of the bind mount, which as you Angel pointed out would not exist (oops!) at the time of the bind mount.
I like the idea of setting the containers for a task in their own directory.  This gives us the unique host-level log file name and a neat organization scheme.  I think it does make these logs more manageable.
(And this also might make cleaning up logs later simpler, based on just the taskID.)

agent/api/container/container.go
```diff
@@ -1167,3 +1176,15 @@ func (c *Container) GetTaskARN() string {
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.ExecAgentMetadata = metadata
+}
+
+func (c *Container) GetExecAgentMetadata() ExecAgentMetadata {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.ExecAgentMetadata
+}
```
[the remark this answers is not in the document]
fierlion: (doesn't `go fmt` fix stuff like that?)

agent/api/task/task.go
fierlion: How about `execute-command-agent-internal` or `internal-execute-command-agent`?

As this isn't an environment variable, I see `awslogsCredsEndpointOpt = "awslogs-credentials-endpoint"` above and I think this is a nice convention.
Also we should be wary of calling this the "exec-agent" in code.  Note, we changed the references in all our designs to call it "execute-command-agent".

agent/api/task/task.go
fenxiong: Container name allows hyphen https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_name. Just wondering is the path valid if the name starts with hyphen?

agent/api/task/task.go
fierlion: I would change the comment above -- as I read this top to bottom I see `// Append exec agent binaries to task volumes` then I see `for _, bn` which makes me read "binary"
maybe if you say `// for each binary name, append its binary to task volume` to make this read immediately as `bn == binary name`

agent/api/task/task.go
angelcar: I can do that, although I was trying to march some other "internal" stuff naming (e.g. [this](https://github.com/aws/amazon-ecs-agent/blob/478c4ec00a43b4713a0acc811cfd1e2d70307f63/agent/api/task/task.go#L64)), but I can go with whatever we decide it's best.

agent/api/task/task.go
angelcar: Directories can't start with hyphens. I guess I can prepend container- to the container name?
e.g. 
two containers: `abc` and `-other`
../exec/taskId1/container-abc
../exec/taskId1/container--other

thoughts?

angelcar: Alternatively I could remove any trailing hyphens, but there's the chance that the container name is just `-`, in which case I'd end up with a blank name.

agent/api/container/container.go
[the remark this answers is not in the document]
angelcar: fixed

agent/api/task/task.go
angelcar: Fixed: I went with `agent/config/config_windows.go`

agent/api/task/task.go
angelcar: Fixed (I think :))

agent/api/task/task.go
angelcar: Fixed: Creating sub directories now. For the container name subdirectory, I am trimming leading hyphens. If the container only has hyphens in the name I fallback to `nameless-continer-UUID`, since it's not possible to create directories with leading hyphens in the name.

agent/config/config.go
angelcar: Added

agent/api/task/task.go
angelcar: Good catch. Fixed.

sharanyad (approved):
code lgtm

agent/api/task/task_linux_test.go
```diff
@@ -1196,3 +1201,164 @@ func getFirelensTask(t *testing.T) *Task {
 		},
 	}
 }
+
+func TestPostUnmarshalTaskWithExecCommandAgentEnabled(t *testing.T) {
+	const containerNameOnlyHyphens = "--"
```
fierlion, line 1206: nice.

agent/config/config.go
```diff
@@ -570,6 +570,8 @@ func environmentConfig() (Config, error) {
 		SpotInstanceDrainingEnabled:         parseBooleanDefaultFalseConfig("ECS_ENABLE_SPOT_INSTANCE_DRAINING"),
 		GMSACapable:                         parseGMSACapability(),
 		VolumePluginCapabilities:            parseVolumePluginCapabilities(),
+		//TODO: [ecs-exec] Document this new env var to the README file
+		ExecAgentLogDir: os.Getenv("ECS_EXEC_AGENT_LOG_DIR"),
```
fierlion, line 574: I don't think this configurable log dir is in the current design.  Is this immediately necessary?  
We'll need to be sure the agent has permissions to access user-defined host level directories.
And, we need to decide what to call it since we're not calling this the `exec_agent` and `ECS_EXECUTE_COMMAND_AGENT_LOG_DIR` is a lot to type.

agent/api/task/task.go
```diff
@@ -338,12 +357,18 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
 	task.initializeCredentialsEndpoint(credentialsManager)
+
 	// NOTE: initializeVolumes needs to be after initializeCredentialsEndpoint, because EFS volume might
 	// need the credentials endpoint constructed by it.
 	if err := task.initializeVolumes(cfg, dockerClient, ctx); err != nil {
 		return err
 	}
 
+	if err := task.initializeExecAgentResources(cfg); err != nil {
+		seelog.Errorf("Task [%s]: could not initialize exec agent: %v", task.Arn, err)
```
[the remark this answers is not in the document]
angelcar: oops, I hadn't pushed my most recent changes, but all these instances were fixed

fierlion (approved):
Good framework -- nice to have this in place

PR #2620, opened 2020-09-08
Add LaunchType and ContainerArn to TMDEv4

sparrc:
### Summary

ACS will now be providing the task launch type and the container ARN as part of the task payload message.

This PR adds support for these new fields, and exposes them on the v4 task metadata endpoint.

### Testing

LaunchType was added to unit test Task objects and ContainerArn was added to the unit test Container objects. Unit tests were written to verify that this data was only returned on the v4 endpoint, not on the others.

Add LaunchType and ContainerArn to TMDEv4 

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -266,6 +266,9 @@ type Container struct {
 	// the JSON body while saving the state
 	SteadyStateStatusUnsafe *apicontainerstatus.ContainerStatus `json:"SteadyStateStatus,omitempty"`
 
+	// ContainerArn is the Arn of this container.
+	ContainerArn string `json:"ContainerArn,omitempty"`
+
 	createdAt  time.Time
 	startedAt  time.Time
 	finishedAt time.Time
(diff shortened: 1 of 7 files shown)


agent/handlers/v4/response.go
```diff
@@ -105,8 +104,6 @@ func NewTaskResponse(
 		})
 	}
 
-	v2Resp.LaunchType = ecs.LaunchTypeEc2
```
sparrc, line 108: a previous change on this feature branch had the launch type hardcoded.

agent/api/container/container.go
```diff
@@ -266,6 +266,9 @@ type Container struct {
 	// the JSON body while saving the state
 	SteadyStateStatusUnsafe *apicontainerstatus.ContainerStatus `json:"SteadyStateStatus,omitempty"`
 
+	// ContainerArn is the Arn of this container.
+	ContainerArn string `json:"ContainerArn,omitempty"`
```
yunhee-l, line 270: do these changes (container.ContainerArn and taskLaunchType) need to be added to the state saved in boltdb as well?

fenxiong (approved):

agent/api/container/container.go
fenxiong: they will be automatically saved as part of task/container i think

yunhee-l (approved):

PR #2622, opened 2020-09-09
Add ps arguments to docker top to pass in process ID

sharanyad:
### Summary
[Docker top API](https://godoc.org/github.com/docker/docker/client#Client.ContainerTop) takes in ps arguments as `[]string` 
Adding the same in engine so we can pass in process ID to check if exec process is running. 

### Testing

N/A

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -134,8 +134,9 @@ type DockerClient interface {
 	InspectContainer(context.Context, string, time.Duration) (*types.ContainerJSON, error)
 
 	// TopContainer returns information about the top processes running in the specified container.  A timeout value and a context
-	// should be provided for the request.
-	TopContainer(context.Context, string, time.Duration) (*dockercontainer.ContainerTopOKBody, error)
+	// should be provided for the request. The last argument is an optional parameter for passing in 'ps' arguments
+	// as part of the top command.
+	TopContainer(context.Context, string, time.Duration, ...string) (*dockercontainer.ContainerTopOKBody, error)
 
 	// ListContainers returns the set of containers known to the Docker daemon. A timeout value and a context
 	// should be provided for the request.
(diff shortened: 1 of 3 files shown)


petderek (approved):

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -135,7 +135,7 @@ type DockerClient interface {
 
 	// TopContainer returns information about the top processes running in the specified container.  A timeout value and a context
 	// should be provided for the request.
-	TopContainer(context.Context, string, time.Duration) (*dockercontainer.ContainerTopOKBody, error)
+	TopContainer(context.Context, string, []string, time.Duration) (*dockercontainer.ContainerTopOKBody, error)
```
fierlion, line 138: The purpose of the dockerapi is to simplify these calls to be asking for only what we need. (https://github.com/aws/amazon-ecs-agent/blob/master/agent/dockerclient/dockerapi/docker_client.go#L188)

Would it be useful to have a PS specific function since that's all we're currently doing with `docker top`?  
Since we have the underlying `inspectContainer` function as our "generic function".

I think it would be fine to rename this function and provide `ps` arg as a default and not require an arg array from the caller.  Thoughts?

sharanyad: I thought about adding `pid` as a string argument instead of the array, but decided not to since that would mean adding other fields as separate args later to this function (which might become cumbersome)
what do you think of adding another explicit method for ps args? 
`TopContainerPs`

angelcar: How about adding a variadic argument?

i.e. `TopContainer(context.Context, string, time.Duration, ...string) `

that way the function can be called with or without ps arguments.

angelcar: In addition, I know it's a pattern that has been there for a while, but I think the timeout shouldn't even be passed since we're already passing a context. The context we pass should be already a `context.WithTimeout`, which would make the timeout parameter unnecessary. If we do this, we would have a cleaner API, like:
`TopContainer(context.Context, string, ...string)`

This is a more idiomatic way of defining this API, IMHO.

sharanyad: I would like to refactor context in a different PR, and follow what other methods are currently doing today.
I like the variadic arg idea.

yunhee-l (approved):

angelcar (approved):

shubham2892 (approved):

PR #2624, opened 2020-09-10
Add ContainerCreateExec to dockerapi

ubhattacharjya:
### Summary

This PR adds the ContainerCreateExec API to the dockerapi package. This API creates the docker exec object.
Whenever we want to call the API CreateContainerExec, we can inherit the user and privilege of the execConfig (https://godoc.org/docker.io/go-docker/api/types#ExecConfig) from the container's config and hostconfig.

### Testing

New tests cover the changes: 
Yes
Unit test cover the change.
Also I have tested manually by calling the CreateContainerExec API from docker_task_manager.go. It could successfully return the response. 

```
execContainerOut: &{be7a768017a283bea7cbefa4a0a035da071fcb485e30289c16686da7e7605a25}\nexecErr: <nil>\n" module=docker_task_engine.go
```

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -137,6 +137,10 @@ type DockerClient interface {
 	// should be provided for the request.
 	TopContainer(context.Context, string, time.Duration) (*dockercontainer.ContainerTopOKBody, error)
 
+	// CreateContainerExec creates a new exec configuration to run an exec process with the provided Config. A timeout value
+	// and a context should be provided for the request.
+	CreateContainerExec(ctx context.Context, containerID string, execConfig types.ExecConfig, timeout time.Duration) (*types.IDResponse, error)
+
 	// ListContainers returns the set of containers known to the Docker daemon. A timeout value and a context
 	// should be provided for the request.
 	ListContainers(context.Context, bool, time.Duration) ListContainersResponse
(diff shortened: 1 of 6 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1511,3 +1515,50 @@ func (dg *dockerGoClient) loadImage(ctx context.Context, reader io.Reader) error
 	}
 	return err
 }
+
+func (dg *dockerGoClient) CreateContainerExec(ctx context.Context, containerID string, execConfig types.ExecConfig, timeout time.Duration) (*types.IDResponse, error) {
```
fierlion, line 1519: I want to be more opinionated about what we have in the function signature here.  the `types.ExecConfig` being required always is a burden.  Like we know we don't care about the stdio streams/tty or interactive mode and we know we'll always be detached.  I think you need to decide which bits of the ExecConfig are necessary here (user?  command array? and update the signature accordingly.  Also see the comments here about how to use variadic arguments https://github.com/aws/amazon-ecs-agent/pull/2622

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1511,3 +1515,50 @@ func (dg *dockerGoClient) loadImage(ctx context.Context, reader io.Reader) error
...
+		// Context has either expired or canceled. If it has timed out,
+		// send back the DockerExecTimeoutError
+		err := ctx.Err()
+		if err == context.DeadlineExceeded {
+			//return DockerContainerExecID{Error: &DockerTimeoutError{timeout, "created"}}
+			return nil, &DockerExecTimeoutError{timeout, "created"}
+		}
+		// Context was canceled even though there was no timeout. Send
+		// back an error.
+		//return DockerContainerExecID{Error: &CannotCreateContainerError{err}}
```
sharanyad, line 1547: nit: remove comment

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1511,3 +1515,50 @@ func (dg *dockerGoClient) loadImage(ctx context.Context, reader io.Reader) error
...
+
+func (dg *dockerGoClient) createContainerExec(ctx context.Context, containerID string, config types.ExecConfig) (*types.IDResponse, error) {
+	client, err := dg.sdkDockerClient()
+	if err != nil {
+		return nil, err
+	}
+
+	execIDResponse, err := client.ContainerExecCreate(ctx, containerID, config)
+	if err != nil {
+		//return DockerContainerMetadata{Error: CannotCreateContainerError{err}}
```
sharanyad, line 1560: nit: remove comment

agent/dockerclient/dockerapi/docker_client.go
sharanyad: I think `types.ExecConfig` arg will always be required here (to create exec config, min required field would be command?) 
I'm ok with either adding only the required fields for now and renaming the method accordingly or leave this as is and support other fields for the future as well.

agent/dockerclient/dockerapi/errors.go
```diff
@@ -30,6 +30,8 @@ const (
 	CannotDescribeContainerErrorName = "CannotDescribeContainerError"
 	// CannotGetContainerTopErrorName is the name of the top container error.
 	CannotGetContainerTopErrorName = "CannotGetContainerTopError"
+	// DockerExecTimeoutErrorName is the name of docker timeout error.
+	DockerExecTimeoutErrorName = "DockerExecTimeoutError"
```
sharanyad, line 34: is this specific to exec create API or other exec ones too?

agent/dockerclient/dockerapi/docker_client.go
ubhattacharjya: I think using execConfig as is in arguments, gives us more control. We have done that for dockerConfig and dockerHostConfig in ContainerCreate dockerapi (https://github.com/aws/amazon-ecs-agent/blob/dev/agent/dockerclient/dockerapi/docker_client.go#L506). Moreover, I dont think we would know what User and Privileged field be unless we inherit it from dockerConfig and dockerHostConfig for each container.
We will have more control over sending execConfig for different useCases if we pass execConfig as a whole in arguments.
Please let me know if you agree.

agent/dockerclient/dockerapi/errors.go
ubhattacharjya: Removed this. I think we can reuse DockerTimeoutError

agent/dockerclient/dockerapi/docker_client.go
fierlion: I agree given that ContainerCreate uses the same pattern.

fierlion (approved):

sharanyad (approved):

PR #2637, opened 2020-09-17
monitor ExecAgent Processes for Exec-enabled tasks

sharanyad:
### Summary
Periodic monitoring of ExecAgent Processes for Exec-enabled tasks

### Implementation details
Monitoring of the processes at an interval of 15 minutes. 
Uses docker top of the ExecAgent process ID to check if the process is still running in the container. 
Couple of `TODO`s:
- Trigger the workflow for restarting ExecAgent if its not running 
- Add jitter between different containers' top API calls to avoid stressing Docker 

### Testing

N/A

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -97,6 +97,8 @@ const (
 	fluentNetworkPort      = "FLUENT_PORT"
 	FluentNetworkPortValue = "24224"
 	FluentAWSVPCHostValue  = "127.0.0.1"
+
+	monitorExecAgentsInterval = 15 * time.Minute
 )
 
 // DockerTaskEngine is a state machine for managing a task and its containers
@@ -160,7 +162,8 @@ type DockerTaskEngine struct {
 
 	// handleDelay is a function used to delay cleanup. Implementation is
 	// swappable for testing
-	handleDelay func(duration time.Duration)
+	handleDelay             func(duration time.Duration)
+	monitorExecAgentsTicker *time.Ticker
 }
 
 // NewDockerTaskEngine returns a created, but uninitialized, DockerTaskEngine.
@@ -248,7 +251,71 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
 	// Now catch up and start processing new events per normal
 	go engine.handleDockerEvents(derivedCtx)
 	engine.initialized = true
+	go engine.startPeriodicExecAgentsMonitoring(derivedCtx, monitorExecAgentsInterval)
 	return nil
+}
+
+func (engine *DockerTaskEngine) startPeriodicExecAgentsMonitoring(ctx context.Context, monitorInterval time.Duration) {
+	engine.monitorExecAgentsTicker = time.NewTicker(monitorInterval)
+	for {
+		select {
+		case <-engine.monitorExecAgentsTicker.C:
+			go engine.monitorExecAgentProcesses(ctx)
+		case <-ctx.Done():
+			engine.monitorExecAgentsTicker.Stop()
+			return
+		}
+	}
+}
+
+func (engine *DockerTaskEngine) monitorExecAgentProcesses(ctx context.Context) {
+	// TODO: [ecs-exec]add jitter between containers to not overload docker with top calls
+	engine.tasksLock.RLock()
+	defer engine.tasksLock.RUnlock()
+	for _, mTask := range engine.managedTasks {
+		task := mTask.Task
+		if !task.IsExecCommandAgentEnabled() {
+			continue
+		}
+		for _, c := range task.Containers {
+			go engine.monitorExecAgentRunning(ctx, task.Arn, c)
+		}
+	}
+}
+
+func (engine *DockerTaskEngine) monitorExecAgentRunning(ctx context.Context,
+	tArn string, c *apicontainer.Container) {
+	if c.GetExecCommandAgentMetadata() == nil {
+		return
+	}
+	pID := c.GetExecCommandAgentMetadata().PID
+	cName := c.Name
+	cID := c.RuntimeID
+	if cID == "" || pID == "" {
+		return
+	}
+	resp, err := engine.client.TopContainer(ctx, cID, dockerclient.TopContainerTimeout, pID)
+	if err != nil {
+		if _, ok := err.(*dockerapi.DockerTimeoutError); ok {
+			seelog.Warnf("Task engine [%s]: Timed out monitoring ExecCommandAgent process for container %s", tArn, cName)
+			return
+		}
+		if strings.Contains(err.Error(), dockerapi.TopProcessNotFoundErrorName) {
+			seelog.Errorf("Task engine [%s]: ExecCommandAgent Process for container %s not running, restarting", tArn, cName)
+			c.SetExecCommandAgentMetadata(&apicontainer.ExecCommandAgentMetadata{})
+			// TODO: [ecs-exec]restart exec agent process for this container
+			return
+		}
+		// TODO: [ecs-exec] consider retry/adding an extra docker exec inspect API call to get more info
+		seelog.Errorf("Task engine [%s]: Error finding ExecCommandAgent process for container %s: %v", tArn, cName, err)
+		return
+	}
+	// no error, but SSM agent process not found in the container
+	if resp == nil || len(resp.Processes) == 0 {
+		seelog.Errorf("Task engine [%s]: ExecCommandAgent Process not found for container %s, no error, restarting", tArn, cName)
+		c.SetExecCommandAgentMetadata(&apicontainer.ExecCommandAgentMetadata{})
+		// TODO: [ecs-exec]restart exec agent process for this container
+	}
 }
 
 // MustInit blocks and retries until an engine can be initialized.
--- a/agent/engine/docker_task_engine_test.go
+++ b/agent/engine/docker_task_engine_test.go
@@ -3128,3 +3128,280 @@ func TestStartFirelensContainerRetryForContainerIP(t *testing.T) {
 	assert.NoError(t, ret.Error)
 	assert.Equal(t, jsonBaseWithNetwork.NetworkSettings, ret.NetworkSettings)
 }
+
+func TestMonitorExecAgentRunning(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.TODO())
+	defer cancel()
+	ctrl, client, _, taskEngine, _, _, _ := mocks(t, ctx, &defaultConfig)
+	defer ctrl.Finish()
+	execAgentPID := "1234"
+	rID := "runtime-ID"
+	resp := &dockercontainer.ContainerTopOKBody{
+		Processes: [][]string{{"root", execAgentPID}},
+	}
+	testCases := []struct {
+		runtimeID                string
+		execCommandAgentMetadata *apicontainer.ExecCommandAgentMetadata
+		execAgentPID             string
+		expectedTopContainerCall bool
+	}{
+		{
+			runtimeID:                rID,
+			execAgentPID:             "",
+			execCommandAgentMetadata: nil,
+			expectedTopContainerCall: false,
+		},
+		{
+			runtimeID:    rID,
+			execAgentPID: execAgentPID,
+			execCommandAgentMetadata: &apicontainer.ExecCommandAgentMetadata{
+				PID: execAgentPID,
+			},
+			expectedTopContainerCall: true,
+		},
+		{
+			runtimeID:    rID,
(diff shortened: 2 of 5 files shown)


agent/engine/docker_task_engine_test.go
```diff
@@ -3128,3 +3128,252 @@ func TestStartFirelensContainerRetryForContainerIP(t *testing.T) {
...
+	ctrl, _, _, taskEngine, _, _, _ := mocks(t, ctx, &defaultConfig)
+	defer ctrl.Finish()
+	testTask := &apitask.Task{
+		Arn: "arn:aws:ecs:region:account-id:task/test-task-arn",
+		Containers: []*apicontainer.Container{
+			{
+				Name:      "test-container",
+				RuntimeID: "runtime-ID",
+			},
+		},
```
angelcar, line 3304: nit: I know the zero value for ExecCommandAgentEnabled is false, but just for the sake of explicitness, could we add `ExecCommandAgentEnabled: false,` here?

agent/engine/docker_task_engine_test.go
```diff
@@ -3128,3 +3128,252 @@ func TestStartFirelensContainerRetryForContainerIP(t *testing.T) {
...
+	}
+	taskEngine.(*DockerTaskEngine).state.AddTask(testTask)
+	taskEngine.(*DockerTaskEngine).managedTasks[testTask.Arn] = &managedTask{Task: testTask}
+	taskEngine.(*DockerTaskEngine).monitorExecAgentProcesses(ctx)
+	time.Sleep(5 * time.Millisecond)
```
angelcar, line 3309: hmm, there are no assertions in this test, so I don't think it's useful? Perhaps we should assert that the TopContainer method is never invoked?

agent/engine/docker_task_engine_test.go
```diff
@@ -3128,3 +3128,252 @@ func TestStartFirelensContainerRetryForContainerIP(t *testing.T) {
...
+				ExecCommandAgentMetadata: &apicontainer.ExecCommandAgentMetadata{
+					PID: execAgentPID2,
+				},
+			},
+		},
+		ExecCommandAgentEnabled: true,
+	}
+	taskEngine.(*DockerTaskEngine).state.AddTask(testTask)
+	taskEngine.(*DockerTaskEngine).managedTasks[testTask.Arn] = &managedTask{Task: testTask}
+	client.EXPECT().TopContainer(gomock.Any(), testTask.Containers[0].RuntimeID, gomock.Any(), execAgentPID1).Return(resp, nil)
```
angelcar, line 3343: I think we should also include the timeout in the EXPECT call, so that if we change that value either intentionally or accidentally in the future, we have a test that makes us aware of the change. (let's not use the constant but directly the expected value)

agent/engine/docker_task_engine_test.go
```diff
@@ -3128,3 +3128,252 @@ func TestStartFirelensContainerRetryForContainerIP(t *testing.T) {
...
+	client.EXPECT().TopContainer(gomock.Any(), testTask.Containers[1].RuntimeID, gomock.Any(), execAgentPID2).Return(nil, errors.New(dockerapi.TopProcessNotFoundErrorName))
+	taskEngine.(*DockerTaskEngine).monitorExecAgentProcesses(ctx)
+	time.Sleep(5 * time.Millisecond)
```
angelcar, line 3346: I'm not a fan of waiting for a fixed time. This can cause flakiness when the CPU is to busy and golang runtime takes its time to schedule go routines. What do you think of creating a primitive to check for a value periodically, and have a timeout. That would make it more deterministic. 

These snippets illustrates the idea.
* This simulates a times out when the expected value is never present, which would be an error: https://play.golang.org/p/ETapQraKPMW
* This shows what happens when the condition we want actually happens: https://play.golang.org/p/rYiqsc3raVY

agent/engine/docker_task_engine_test.go
```diff
@@ -3128,3 +3128,252 @@ func TestStartFirelensContainerRetryForContainerIP(t *testing.T) {
...
+				},
+			},
+		},
+		ExecCommandAgentEnabled: true,
+	}
+	taskEngine.(*DockerTaskEngine).state.AddTask(testTask)
+	taskEngine.(*DockerTaskEngine).managedTasks[testTask.Arn] = &managedTask{Task: testTask}
+	client.EXPECT().TopContainer(gomock.Any(), testTask.Containers[0].RuntimeID, gomock.Any(), execAgentPID).Return(resp, nil).AnyTimes()
+	go taskEngine.(*DockerTaskEngine).startPeriodicExecAgentsMonitoring(ctx, 2*time.Millisecond)
+	time.Sleep(10 * time.Millisecond)
```
angelcar, line 3377: same comment as above about waiting for a fixed time.

angelcar (commented):
One more comment would be, what happens if for some reason we can never start exec cmd agent despite our best efforts?

agent/engine/docker_task_engine.go
```diff
@@ -248,9 +251,71 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
...
+		}
+	}
+}
+
+func (engine *DockerTaskEngine) monitorExecAgentProcesses(ctx context.Context) {
+	// TODO: add jitter between containers to not overload docker with top calls
+	for _, mTask := range engine.managedTasks {
+		task := mTask.Task
+		if !task.ExecCommandAgentEnabled {
+			return
```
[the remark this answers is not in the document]
sharanyad: my bad, yes

agent/engine/docker_task_engine.go
```diff
@@ -248,9 +251,71 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
...
+			seelog.Warnf("Task engine [%s]: Timed out monitoring ExecAgent process for container %s", tArn, cName)
+			return
+		}
+		if strings.Contains(err.Error(), dockerapi.TopProcessNotFoundErrorName) {
+			seelog.Errorf("Task engine [%s]: ExecAgent Process for container %s not running, restarting", tArn, cName)
+			c.SetExecCommandAgentMetadata(&apicontainer.ExecCommandAgentMetadata{})
+			// TODO: restart exec agent process for this container
+			return
+		}
+		seelog.Errorf("Task engine [%s]: Error finding ExecAgent process for container %s", tArn, cName)
```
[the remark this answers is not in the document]
sharanyad: > Error finding ExecAgent ExecCommandAgent

I wanted to discuss about this one. it's true the API is going to be ExecCommand, but I don't think its necessary to call it out fully in the code. We could use ExecCommand in documentation etc. but in the code base, I would prefer to keep the variable names as small as possible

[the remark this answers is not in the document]
sharanyad: > Should we log the actual error here?

I shall log the error 

> Also, should we try to restart exec cmd agent in this case as well?

I was contemplating this one. The error could be 'container not found' as well (just when task is stopped?), so in those cases, it's not necessary to restart the agent.

agent/engine/docker_task_engine.go
```diff
@@ -248,9 +251,71 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
...
+	} else {
```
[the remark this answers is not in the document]
sharanyad: I'll add a return and remove the else.

agent/engine/docker_task_engine_test.go
sharanyad: not adding the expect call for the mock client itself is an assertion that TopContainer is not invoked.

agent/engine/docker_task_engine_test.go
sharanyad: +1 about the comment on waiting for a fixed time. 

This test and below one is a bit tricky, since we use go routines in the code. 
The assertion is checking if the value of the Exec Metadata's PID has been unchanged after the monitor call. 
Hence the check for a value periodically and having a timeout for that may not actually work in this case, since we will not know if the value is unchanged before/after the monitor call is over. Hence this current pattern. 
Open to better ideas.

agent/engine/docker_task_engine_test.go
sharanyad: I'm not quite sure I understand the comment. Could you give an example?

sharanyad:
> One more comment would be, what happens if for some reason we can never start exec cmd agent despite our best efforts?

This will follow the start exec agent code path. We should use backoff for retries and could even have finite number of retries.

agent/engine/docker_task_engine_test.go
angelcar: how about at least waiting for the `TopContainer` to be executed before the static wait? This narrows the margin for flakiness

```golang
topCtx, topCancel := context.WithTimeout(context.Background(), time.Second) 
defer topCancel()
client.EXPECT().TopContainer(gomock.Any(), testTask.Containers[1].RuntimeID, gomock.Any(),execAgentPID2).DoAndReturn(
	func(ctx context.Context, dockerID string, timeout time.Duration, psArgs ...string) (*dockercontainer.ContainerTopOKBody, error) {
		defer topCancel()
		return nil, errors.New(dockerapi.TopProcessNotFoundErrorName)
	})
taskEngine.(*DockerTaskEngine).monitorExecAgentProcesses(ctx)
<-topCtx.Done()
time.Sleep(5 * time.Millisecond)
```

agent/engine/docker_task_engine_test.go
angelcar: just like 

```golang
client.EXPECT().TopContainer(gomock.Any(), testTask.Containers[0].RuntimeID, time.Minute * 15, execAgentPID1).Return(resp, nil)
```

agent/engine/docker_task_engine.go
```diff
@@ -248,9 +251,72 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
...
+			seelog.Warnf("Task engine [%s]: Timed out monitoring ExecCommandAgent process for container %s", tArn, cName)
+			return
+		}
+		if strings.Contains(err.Error(), dockerapi.TopProcessNotFoundErrorName) {
+			seelog.Errorf("Task engine [%s]: ExecCommandAgent Process for container %s not running, restarting", tArn, cName)
+			c.SetExecCommandAgentMetadata(&apicontainer.ExecCommandAgentMetadata{})
+			// TODO: [ecs-exec]restart exec agent process for this container
+			return
+		}
+		seelog.Errorf("Task engine [%s]: Error finding ExecCommandAgent process for container %s: %v", tArn, cName, err)
```
angelcar, line 309: Maybe we should consider retrying here, just in case the error is transient? Just something to consider, but not a blocker for this PR.

angelcar (approved):

agent/engine/docker_task_engine.go
sharanyad: was thinking it will be covered in the next monitor cycle? but yes, I shall add a TODO to consider this.

agent/engine/docker_task_engine.go
```diff
@@ -248,9 +251,71 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
...
+	if c.ExecCommandAgentMetadata == nil {
+		return
+	}
+	pID := c.GetExecCommandAgentMetadata().PID
+	cName := c.Name
+	cID := c.RuntimeID
+	if cID == "" || pID == "" {
+		return
+	}
+	resp, err := engine.client.TopContainer(ctx, cID, dockerclient.TopContainerTimeout, pID)
```
fierlion, line 297: the docker exec inspect output should also give us a clue about the status of our running agent -- 
for a failure I wonder if we can't also ask docker exec inspect what happened (we get the process exit code from that output)
```
    "ID": "f086903815fbbd127ec816eb154d772696088cc53ba22bbef0098bb8cc44849a",
    "Running": false,
    "ExitCode": 2,
```

agent/engine/docker_task_engine.go
fierlion: Yeah -- I see this check process as two-fold: 
if we don't see the execute command agent running based on docker top and the `ExecCommandAgentMetadata.PID`, then I'd like to ask docker exec inspect for the exit code of that process and act accordingly.
I think we should only be restarting for a subset of exit codes (maybe just 2?), else we should log an error...

fierlion (approved):

agent/engine/docker_task_engine.go
sharanyad: thats a good point and I considered it earlier, but its adding yet another docker API call, which is why I was reluctant. I will add a `TODO` to reconsider this call and checking based on exit codes.

PR #2640, opened 2020-09-18
Fix response code for container network metadata request in TMDE

mythri-garaga:
### Summary

This PR fixes response code for container network metadata request in TMDE
For example: If container's network metadata is unavailable, we should return 5xx instead of 400(Bad Request) response code

### Implementation details

This PR includes following changes to ` agent/handlers`
- ~~In `/utils`- added errors.go to have a Transient error type~~
- In `/v3` and `/v4` - added logging and fixed response code for container and task metadatahandler request

### Testing

Manually tested the changes to see if we are 500 response code is returned when container network settings is nil by commenting out https://github.com/aws/amazon-ecs-agent/blob/953af906b5de51f25d4f366bcfb238ae277a7133/agent/engine/docker_task_engine.go#L414.

Bug - Fix response code when container network metadata is unavailable in TMDE
### Licensing

### DIFF
--- a/agent/handlers/utils/helpers.go
+++ b/agent/handlers/utils/helpers.go
@@ -80,6 +80,10 @@ func WriteJSONToResponse(w http.ResponseWriter, httpStatusCode int, responseJSON
 			"Unable to write %s json response message to ResponseWriter",
 			requestType)
 	}
+
+	if httpStatusCode >= 400 && httpStatusCode <= 599 {
+		seelog.Errorf("HTTP response status code is '%d', request type is: %s, and response in JSON is %s", httpStatusCode, requestType, string(responseJSON))
+	}
 }
 
 // WriteResponseIfMarshalError checks the 'err' response of the json.Marshal function.


agent/handlers/v3/container_metadata_handler.go
```diff
@@ -65,6 +69,7 @@ func ContainerMetadataHandler(state dockerstate.TaskEngineState) func(http.Respo
 func GetContainerResponse(containerID string, state dockerstate.TaskEngineState) (*v2.ContainerResponse, error) {
 	containerResponse, err := v2.NewContainerResponse(containerID, state)
 	if err != nil {
+		seelog.Errorf("unable to get container metadata for container '%s'", containerID)
```
sharanyad, line 72: nit: capitalize the start of the log message.

agent/handlers/v3/container_metadata_handler.go
```diff
@@ -86,7 +91,8 @@ func GetContainerNetworkMetadata(containerID string, state dockerstate.TaskEngin
 	// https://github.com/aws/amazon-ecs-agent/blob/0c8913ba33965cf6ffdd6253fad422458d9346bd/agent/containermetadata/parse_metadata.go#L123
 	settings := dockerContainer.Container.GetNetworkSettings()
 	if settings == nil {
-		return nil, errors.Errorf("Unable to generate network response for container '%s'", containerID)
+		seelog.Errorf("unable to get container network metadata for container '%s'", containerID)
+		return nil, &utils.TransientError{Msg: fmt.Sprintf("Unable to generate network response for container '%s'", containerID)}
```
sharanyad, line 95: nit: `unable`

sharanyad (commented):
would like to see a unit test case for the transient error.

agent/handlers/v4/container_metadata_handler.go
```diff
@@ -65,6 +69,7 @@ func ContainerMetadataHandler(state dockerstate.TaskEngineState) func(http.Respo
 func GetContainerResponse(containerID string, state dockerstate.TaskEngineState) (*ContainerResponse, error) {
 	containerResponse, err := NewContainerResponse(containerID, state)
 	if err != nil {
+		seelog.Errorf("unable to get container metadata for container '%s'", containerID)
 		return nil, errors.Errorf("unable to generate metadata for container '%s'", containerID)
```
sparrc, line 73: shouldnt this be returning TransientError?

sparrc (commented):
-

agent/handlers/v3/container_metadata_handler.go
```diff
@@ -65,6 +69,7 @@ func ContainerMetadataHandler(state dockerstate.TaskEngineState) func(http.Respo
 func GetContainerResponse(containerID string, state dockerstate.TaskEngineState) (*v2.ContainerResponse, error) {
 	containerResponse, err := v2.NewContainerResponse(containerID, state)
 	if err != nil {
+		seelog.Errorf("unable to get container metadata for container '%s'", containerID)
 		return nil, errors.Errorf("Unable to generate metadata for container '%s'", containerID)
```
sparrc, line 73: Should this be TransientError?

sparrc (commented):
and here: https://github.com/aws/amazon-ecs-agent/pull/2640/files#diff-5b21b592e330f17bccf45df2a4d6fe12R60

sparrc (commented):
and here: https://github.com/aws/amazon-ecs-agent/pull/2640/files#diff-0b8a532eb5f5c089caf79c60a424611dR56

ie, transient errors can happen when gathering task response too, not just container response

sparrc:
to me this looks like a good start, but I think we should consider any error that isn't a problem with parameters of the request path a 500

agent/handlers/v3/container_metadata_handler.go
mythri-garaga: If it is unable to find container metadata by containerID or TaskID,  I thought it would still be a Bad Request error.

agent/handlers/v4/container_metadata_handler.go
mythri-garaga: I think it should still be a Bad Request error., if it is unable to find container metadata by containerID or TaskID?

agent/handlers/v3/container_metadata_handler.go
mythri-garaga: updated.

agent/handlers/v3/container_metadata_handler.go
mythri-garaga: Done.

agent/handlers/v3/container_metadata_handler.go
mythri-garaga: > would like to see a unit test case for the transient error.

Added them!

mythri-garaga:
> would like to see a unit test case for the transient error.

Added the unit tests for both v3 and v4.

agent/handlers/v3/container_metadata_handler.go
sparrc: Customers interact with this API very generically: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html. For v4, they would be curling `${ECS_CONTAINER_METADATA_URI_V4}/task`, the specific task and container IDs are actually generated internally. Since we generate them server-side, if we have an invalid ID in state then that should be a 5xx error.

400 errors should only be used if the customer made a bad request, ie misspelling `task` in the request path.

mythri-garaga: Agreed, will update the PR

sharanyad: its kinda both agent/customer.
customers can still specify container ID as part of the request https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v2.html

mythri-garaga: That's true, but there are no container Id specific endpoints for v3 and v4
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v3.html
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html

mythri-garaga:
reopened in #2643

PR #2641, opened 2020-09-18
Add logic in WriteJSONToResponse to log errors in ECS agent log

chienhanlin:
### Summary
This PR adds logic to propagate responses of requests to the Amazon ECS Container Agent Log when an error occurs.

### Implementation details
This PR includes the following change in agent/handlers/utils/helpers.go:  
- Added logic in `WriteJSONToResponse` function to propagate HTTP response status codes, request types and responses in JSON to the Amazon ECS Container Agent Log when the response status code is 4xx client error or 5xx server error.

### Testing

- Manually tested the changes with following Task Metadata Endpoint (TMDE) version 3 and 4 paths, and get returned JSON responses.

1. ${ECS_CONTAINER_METADATA_URI}
2. ${ECS_CONTAINER_METADATA_URI}/task
3. ${ECS_CONTAINER_METADATA_URI}/stats
4. ${ECS_CONTAINER_METADATA_URI}/task/stats
    
- Manually tested the changes when HTTP response status code 500 occurs, and verified the HTTP response status code, the request type and responses are successfully propagated to Amazon ECS Container Agent Log.
 
- This change is not covered by new tests

Improvement - Propagate responses to the Amazon ECS Container Agent Log when an error occurs

### Licensing

### DIFF
--- a/agent/handlers/utils/helpers.go
+++ b/agent/handlers/utils/helpers.go
@@ -80,6 +80,10 @@ func WriteJSONToResponse(w http.ResponseWriter, httpStatusCode int, responseJSON
 			"Unable to write %s json response message to ResponseWriter",
 			requestType)
 	}
+
+	if httpStatusCode >= 400 && httpStatusCode <= 599 {
+		seelog.Errorf("HTTP response status code is '%d', request type is: %s, and response in JSON is %s", httpStatusCode, requestType, string(responseJSON))
+	}
 }
 
 // WriteResponseIfMarshalError checks the 'err' response of the json.Marshal function.


ubhattacharjya (approved):

agent/handlers/utils/helpers.go
```diff
@@ -129,3 +131,10 @@ func LimitReachedHandler(auditLogger audit.AuditLogger) func(http.ResponseWriter
 		auditLogger.Log(logRequest, http.StatusTooManyRequests, "")
 	}
 }
+
+// WriteErrorToDefaultLogger write the requestType, httpStatusCode and responseJSON of error to the default logger
+func WriteErrorToDefaultLogger(httpStatusCode int, responseJSON []byte, requestType string) {
```
angelcar, line 136: nit: I think the function name is a bit misleading, there's no error passed as parameter, also the function is not reused anywhere.  I think I'd prefer to get rid of the function and just put the code directly in `WriteJSONToResponse`

agent/handlers/utils/helpers.go
```diff
@@ -129,3 +131,10 @@ func LimitReachedHandler(auditLogger audit.AuditLogger) func(http.ResponseWriter
...
+	if httpStatusCode >= 400 && httpStatusCode <= 599 {
```
angelcar, line 137: would it be better to have `if httpStatusCode<200 || httpStatusCode>299` ?

angelcar (approved):

agent/handlers/utils/helpers.go
chienhanlin: Thanks for the suggestion. function `WriteErrorToDefaultLogger` is now integrated into function `WriteJSONToResponse`.

agent/handlers/utils/helpers.go
chienhanlin: As httpStatusCode 1xx and 3xx would not be considered as an error code; therefore the if statement `httpStatusCode >= 400 && httpStatusCode <= 599` is used here to only propagate httpStatusCode 4xx and 5xx to the agent log.

angelcar (approved):

PR #2647, opened 2020-09-23
Use Go Modules

mythri-garaga:
### Summary

Use go modules
### Implementation details

- from agent/
```
go mod init
go mod tidy
rm -rf vendor
go mod vendor
```
- removed dep files - ./agent/Gopkg.lock and ./agent/Gopkg.toml
### Testing

New tests cover the changes: 

### Licensing


mythri-garaga:
TestTaskMetadataValidator and TestV4TaskEndpointAWSVPCNetworkMode functional tests failing for al1, al2, inf, gpu.

mythri-garaga:
Switch to "goMod" base branch since moving to go modules affects the make targets.

fenxiong (commented):
any idea why we need to update quite a few vendor packages as part of the change? is there a way to keep them unchanged?

sparrc:
> any idea why we need to update quite a few vendor packages as part of the change? is there a way to keep them unchanged?

gomod automatically pins the same library versions from Gopkg, so it's not the library versions changing (I spot checked a few too).

Im not sure why there are so many differences in the vendor/ directory. I know that godep and gomod both have their own algorithms for determining if a package is unused. It seems like godep is generally a bit more aggressive in pruning unused vendored code.

sparrc (approved):

fenxiong (approved):

mythri-garaga:
> > any idea why we need to update quite a few vendor packages as part of the change? is there a way to keep them unchanged?
> 
> gomod automatically pins the same library versions from Gopkg, so it's not the library versions changing (I spot checked a few too).
> 
> Im not sure why there are so many differences in the vendor/ directory. I know that godep and gomod both have their own algorithms for determining if a package is unused. It seems like godep is generally a bit more aggressive in pruning unused vendored code.
> 
> If we want to use gomod I dont think there's any other option. You can't even leave it as-is and just avoid running `go mod vendor` (not that we should do that) because gomod checks that the contents of the vendor directory is consistent with it's view of the world.

Thanks for explaining @sparrc

PR #2648, opened 2020-09-25
Start exec command agent upon container startup

angelcar:
### Summary

Add ExecCommandAgent manager, which can idempotently start and restart the ExecCommandAgent with exponential back off and max retries.
### Implementation details

Create a new manager struct in order to abstract ExecCommandAgent away from the engine. Now the engine is much simpler since it's not aware of what it takes to start/restart ExecCommandAgent. Modified engine's UTs accordingly 

### Testing

* Manually tested that the ExecCommandAgent is started in the container. 
* Manually stopped ExecCommandAgent to verify it's restarted automatically

New tests cover the changes: 

### Licensing


agent/engine/execcmdagent/manager.go
```diff
@@ -0,0 +1,71 @@
...
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
+)
+
+const (
+	cmdField              = "CMD"
+	defaultRetryTimeout   = time.Minute * 10
+	defaultRetryMinDelay  = time.Second * 1
+	defaultRetryMaxDelay  = time.Second * 30
```
sharanyad, line 29: ideally, we would want the retries to finish within the health checkup interval (which is 15 mins), so this might need to be changed accordingly.

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+// RestartIfStopped restarts the ExecCommandAgent in the container passed as parameter, only for ExecCommandAgent-enabled tasks.
+// The status of the ExecCommandAgent in the container is retrieved using a docker exec inspect call, using the dockerExecID
+// stored in container.GetExecCommandAgentMetadata().DockerExecID.
+//
+// If container.GetExecCommandAgentMetadata() is nil, or the ExecCommandAgent is still running, no action is taken.
+// To actually restart the ExecCommandAgent, this function invokes this instance's Start method.
+func (m *manager) RestartIfStopped(ctx context.Context, client dockerapi.DockerClient, task *apitask.Task, container *apicontainer.Container, containerId string) (bool, error) {
+	if !task.IsExecCommandAgentEnabled() {
```
sharanyad, line 41: can this and below check be done in the parent method to fail early?

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,209 @@
...
+	if execMD == nil { // This means m.Start has never been invoked
+		return false, nil
+	}
+	resp, err := client.InspectContainerExec(ctx, execMD.DockerExecID, dockerclient.ContainerExecInspectTimeout)
+	if err != nil {
+		if _, ok := err.(*dockerapi.DockerTimeoutError); ok {
+			return false, err
+		}
+		// TODO: [ecs-exec] consider retry/adding an extra docker exec inspect API call to get more info
+		return false, err
```
sharanyad, line 54: if this is a transient docker error, then returning an error triggering a submit status change as agent stopped may not make sense (exec agent could still be running). feels like unless we know for sure agent is not running, we should not trigger submit status change. same for docker timeout errors.

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,209 @@
...
+	ctx, cancel := context.WithTimeout(ctx, m.retryTimeout)
+	defer cancel()
+	var (
+		startErr error
+		attempt  int
+	)
+	retry.RetryWithBackoffCtx(ctx, backoff, func() error {
+		startErr = doStart(ctx, client, task, container, containerId)
+		if attempt == maxRetries {
+			return nil
```
sharanyad, line 87: shouldn't this return an error since we reached max retries to start agent?

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,209 @@
...
+	//TODO: [ecs-exec] retry only for certain exit codes?
+	seelog.Warnf("Task engine [%s]: ExecCommandAgent Process stopped (exitCode:=%d) for container %s, restarting...", task.Arn, resp.ExitCode, containerId)
+	err = m.Start(ctx, client, task, container, containerId)
+	// err == nil means start succeeded
+	return err == nil, err
+}
+
+// Start idempotently starts the ExecCommandAgent in the container passed as parameter, only for ExecCommandAgent-enabled tasks.
+// In order for the ExecCommandAgent to be successfully started, the following conditions need to be met:
+//	1) There shall be at most one ExecCommandAgent process running in the container (with the name defined in apitask.ExecCommandAgentBinName)
```
sharanyad, line 69: why is this check a requirement? ie. when do we anticipate a process already running?

agent/engine/docker_task_engine.go
```diff
@@ -274,48 +278,31 @@ func (engine *DockerTaskEngine) monitorExecAgentProcesses(ctx context.Context) {
...
 		return
 	}
-	// no error, but SSM agent process not found in the container
-	if resp == nil || len(resp.Processes) == 0 {
-		seelog.Errorf("Task engine [%s]: ExecCommandAgent Process not found for container %s, no error, restarting", tArn, cName)
-		c.SetExecCommandAgentMetadata(&apicontainer.ExecCommandAgentMetadata{})
-		// TODO: [ecs-exec]restart exec agent process for this container
+	_, err = engine.execCmdAgentMgr.RestartIfStopped(ctx, engine.client, task, c, dockerID)
+	if err != nil {
+		seelog.Errorf("Task engine [%s]: Failed to restart ExecCommandAgent Process for container [%s]: %v", task.Arn, dockerID, err)
```
sharanyad, line 302: on restart failing, the exec metadata needs to be reset https://github.com/aws/amazon-ecs-agent/pull/2648/files#diff-b9aa12c3425c34d6f330f28407f779f6L305

agent/engine/docker_task_engine.go
```diff
@@ -1267,6 +1256,13 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
 
 		}
 	}
+	if task.IsExecCommandAgentEnabled() {
+		if err := engine.execCmdAgentMgr.Start(engine.ctx, engine.client, task, container, dockerID); err != nil {
```
sharanyad, line 1260: feels like this should be outside the startContainer method - ie. container moving to RUNNING is blocked by starting exec agent here, when in real it does not affect start container flow.

agent/engine/execcmdagent/manager.go
angelcar: This timeout is the total timeout, not an individual timeout. I set it to 10 mins thinking about the 15 mins of the health check.

agent/engine/execcmdagent/manager_linux.go
angelcar: I see your point. I still want to signal something is wrong in this case. Will think more about it.

agent/engine/execcmdagent/manager.go
sharanyad: my bad, I read it as 30 minutes 😅

agent/engine/execcmdagent/manager_linux.go
angelcar: This is a weird aspect of`retry.RetryWithBackoffCtx`. if max retries is reached AND there is an error, then `startErr` will store it. The return is only to break the internal loop of `retry.RetryWithBackoffCtx`. I thought about refactoring it, but it's being used in too many places. 
Other than the weird readability, it works as you can see in the UTs.

agent/engine/execcmdagent/manager_linux.go
angelcar: As per the design doc, there might be customers already running their own ssm agent in their containers. 
For these cases we don't want to start another ssm agent.

agent/engine/docker_task_engine.go
angelcar: Good catch, I will reset it inside `RestartIfStopped` because we still need the metadata in the following scenario:
* Docker exec inspect fails: in this case we could not verify if the agent is running, in the next health check we need the previous metadata to do another docker exec inspect, if we remove it, we would lose track of it and it might still be running.

agent/engine/docker_task_engine.go
angelcar: This is another weird thing of the engine; actually this doesn't block the start. I've observed that if the `client.StartContainer` call above (line 1200) succeeds, then container will be reported as RUNNING, because then docker reports the start to the client in ecs agent, which is listening for docker events.

agent/engine/execcmdagent/manager_linux.go
angelcar: I thought about it, and I do it for Start, can do it for this one too, but I still want to keep this checks inside the functions because the engine is not very well encapsulated and in the future someone might call these functions without first checking if exec is enabled.

angelcar: Actually, this is implicitly done because the health check process skips non-exec enabled tasks: https://github.com/aws/amazon-ecs-agent/pull/2648/files#diff-b9aa12c3425c34d6f330f28407f779f6R281

agent/engine/execcmdagent/manager_linux.go
angelcar: Will see if I can use another construct of the `retry` package, but it wasn't much different.

agent/api/task/task_linux.go
```diff
@@ -58,11 +56,11 @@ const (
 	ExecCommandAgentSessionWorkerBinName = "ssm-session-worker"
 
 	// TODO: [ecs-exec] decide if this needs to be configurable or put in a specific place in our optimized AMIs
```
sparrc, line 58: what's the status of this TODO?

agent/engine/docker_task_engine.go
```diff
@@ -274,48 +278,31 @@ func (engine *DockerTaskEngine) monitorExecAgentProcesses(ctx context.Context) {
...
 	}
-	// no error, but SSM agent process not found in the container
-	if resp == nil || len(resp.Processes) == 0 {
-		seelog.Errorf("Task engine [%s]: ExecCommandAgent Process not found for container %s, no error, restarting", tArn, cName)
-		c.SetExecCommandAgentMetadata(&apicontainer.ExecCommandAgentMetadata{})
-		// TODO: [ecs-exec]restart exec agent process for this container
+	_, err = engine.execCmdAgentMgr.RestartIfStopped(ctx, engine.client, task, c, dockerID)
+	if err != nil {
+		seelog.Errorf("Task engine [%s]: Failed to restart ExecCommandAgent Process for container [%s]: %v", task.Arn, dockerID, err)
+		// TODO: [ecs-exec] call control plane to report ExecCommandAgent FAIL here
```
sparrc, line 303: This is an API that CP is in process of implementing?

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,209 @@
...
+	if !task.IsExecCommandAgentEnabled() {
+		return false, nil
+	}
+	execMD := container.GetExecCommandAgentMetadata()
+	if execMD == nil { // This means m.Start has never been invoked
+		return false, nil
+	}
+	resp, err := client.InspectContainerExec(ctx, execMD.DockerExecID, dockerclient.ContainerExecInspectTimeout)
+	if err != nil {
+		if _, ok := err.(*dockerapi.DockerTimeoutError); ok {
```
sparrc, line 50: why are we checking if it's a DockerTimeoutError if we're just going to return `false, err` either way?

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		// TODO: [ecs-exec] consider retry/adding an extra docker exec inspect API call to get more info
+		return false, err
+	}
+	if resp.Running { // Exec cmd agent is running, nothing to do
+		return false, nil
+	}
+	// Restart if not running
+	//TODO: [ecs-exec] retry only for certain exit codes?
+	seelog.Warnf("Task engine [%s]: ExecCommandAgent Process stopped (exitCode:=%d) for container %s, restarting...", task.Arn, resp.ExitCode, containerId)
```
sparrc, line 61: is this the exit code of ExecCommandAgent process or of the container? Is there documentation or design around different exit codes from ExecCommandAgent?

sparrc: If there isn't probably better to remove the TODO, as it implies that ExecCommandAgent has logic around it's exit codes

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,209 @@
...
+	seelog.Debugf("Task engine [%s]: started exec command agent for container: %s ->  docker exec id: %s", task.Arn, containerId, execRes.ID)
+
+	inspect, err := client.InspectContainerExec(ctx, execRes.ID, dockerclient.ContainerExecInspectTimeout)
+	if err != nil {
+		return StartError{error: err, retryable: true}
+	}
+	seelog.Debugf("Task engine [%s]: inspect exec command agent for container: %s -> pid: %d, exitCode: %d, running:%v, err:%v",
+		task.Arn, containerId, inspect.Pid, inspect.ExitCode, inspect.Running, err)
+
+	if !inspect.Running { //TODO: [ecs-exec] retry only for certain exit codes?
```
sparrc, line 132: same exit code comment as above

agent/api/task/task_linux.go
angelcar: To be defined, the ssm agent is still working on this, so I want to make sure I don't forget by leaving this TODO here

agent/engine/docker_task_engine.go
angelcar: Yes it is, Also, Ray is gonna work in integrating this part, leaving this TODOs so we have a reference of where to put things

agent/engine/execcmdagent/manager_linux.go
angelcar: Yeah, we were contemplating to have different treatments, but I removed that

agent/engine/execcmdagent/manager_linux.go
angelcar: I prefer to leave the TODO here, as there are discussions in progress about the exit codes. Before this feature branch is merged, all the `[ecs-exec]` will be addressed, either by removing them or by implementing additional stuff.

angelcar: Oh, and the exit code is the exit code of ExecCommandAgent

agent/engine/execcmdagent/manager_linux.go
angelcar: ditto

agent/engine/execcmdagent/manager_linux.go
angelcar: Ended up adding retries for the inspect. If the inspect fails after all retries, then I return the error, I think it's reasonable in this case to report something is wrong since we could not inspect the status of the ExecCommandAgent, most likely something is wrong.

agent/engine/execcmdagent/manager_linux.go
angelcar: Changed to use `retry.RetryNWithBackoffCtx`

agent/engine/execcmdagent/manager_linux.go
angelcar: Removed that check altogether... less code, yay!

agent/engine/docker_task_engine.go
angelcar: Addresssed

agent/engine/docker_task_engine.go
angelcar: So none of the lines below line 1204 will block the start of the container

agent/api/container/container.go
```diff
@@ -270,7 +270,7 @@ type Container struct {
 	ContainerArn string `json:"ContainerArn,omitempty"`
 
 	// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
-	ExecCommandAgentMetadata *ExecCommandAgentMetadata `json:"execCommandAgentMetadata"`
+	ExecCommandAgentMetadata *ExecCommandAgentMetadata `json:"ExecCommandAgentMetadata"`
```
sharanyad, line 273: this needs a bump in the agent state file version: https://github.com/aws/amazon-ecs-agent/blob/master/agent/statemanager/state_manager.go#L106 (can add a TODO for later too)

agent/engine/execcmdagent/manager_linux.go
sharanyad: we could later think about moving the exec methods in engine to exec agent manager

agent/engine/execcmdagent/manager_linux.go
angelcar: what would happen if it just keeps timing out?

angelcar: Now returning a `RestartStatus` type to indicate if the exec agent was `Restarted`, `NotRestarted`, `Unknown`.
In case of inspect errors, I return `(Unknown, nil)` (i.e. no error)

P.S. I retry the InspectContainerExec command, except for timeout errors, as discussed.

agent/api/container/container.go
angelcar: Done. Thanks for pointing this out.

agent/engine/docker_task_engine.go
fierlion: (thanks for that)
pretty sure this is going to be an enum, and we'll support `PENDING`, `RUNNING`, `STOPPED` states.

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,165 @@
...
+	res, err := m.inspectExecAgentProcess(ctx, client, execMD)
+	if err != nil {
+		// We don't want to report InspectContainerExec errors, just that we don't know what the status of the agent is
+		return Unknown, nil
+	}
+	if res.Running { // agent still running, nothing to do
+		return NotRestarted, nil
+	}
+	// Restart if not running
+	//TODO: [ecs-exec] retry only for certain exit codes?
```
fierlion, line 56: I think we should use the SSM agent systemd service file as our guide:
`Restart=on-failure`
https://github.com/aws/amazon-ssm-agent/blob/master/packaging/linux/amazon-ssm-agent.service#L10
```
If set to on-failure, the service will be restarted when the process exits with a non-zero exit code, is terminated by a signal (including on core dump, but excluding the aforementioned four signals), when an operation (such as service reload) times out, and when the configured watchdog timeout is triggered.
```

agent/engine/execcmdagent/manager.go
```diff
@@ -0,0 +1,83 @@
...
+	defaultRetryMaxDelay       = time.Second * 30
+	defaultInspectRetryTimeout = time.Minute * 2
+	maxRetries                 = 5
+	retryDelayMultiplier       = 1.5
+	retryJitterMultiplier      = 0.2
+)
+const (
+	Restarted RestartStatus = iota
+	NotRestarted
+	Unknown
```
fierlion, line 37: We might be in `Unknown` land for longer durations (ex. `dockerTimeoutErr` which you reference in your test).  This makes me ask what we will report when we have `RUNNING`|`PENDING`|`STOPPED` to report.

What does the ECS Agent do for `Unknown` -- are we following that pattern here?
I see the point of it existing, but it brings up a little bit of a tangle.

agent/engine/execcmdagent/manager_linux.go
fierlion: I'm ok with the overhead of these checks -- in case this function gets called from the wrong context we have some simple guard rails.

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,165 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
+)
+
+// RestartIfStopped restarts the ExecCommandAgent in the container passed as parameter, only for ExecCommandAgent-enabled tasks.
+// The status of the ExecCommandAgent in the container is retrieved using a docker exec inspect call, using the dockerExecID
+// stored in container.GetExecCommandAgentMetadata().DockerExecID.
+//
+// If container.GetExecCommandAgentMetadata() is nil, or the ExecCommandAgent is still running, no action is taken.
```
fierlion, line 37: Note: I went back and forth about this function.
I had imagined a bulletproof `start` function that we could call from any context.
I like this approach.

agent/engine/execcmdagent/manager_linux.go
```diff
@@ -0,0 +1,165 @@
...
+
+// Start idempotently starts the ExecCommandAgent in the container passed as parameter, only for ExecCommandAgent-enabled tasks.
+// In order for the ExecCommandAgent to be successfully started, the following conditions need to be met:
+//	1) There shall be at most one ExecCommandAgent process running in the container (with the name defined in apitask.ExecCommandAgentBinName)
+//	2) If an ExecCommandAgent process is already running in the container, its PID shall be equal to container.GetExecCommandAgentMetadata().PID
+// If any of the previous conditions is violated, ExecCommandAgent's can't be verified and an error is returned.
+// If no error is returned, it can be assumed the ExecCommandAgent is started.
+func (m *manager) Start(ctx context.Context, client dockerapi.DockerClient, task *apitask.Task, container *apicontainer.Container, containerId string) error {
+	if !task.IsExecCommandAgentEnabled() {
+		return nil
```
fierlion, line 99: debug log here?  why are we calling start on non execute-command enabled tasks?

agent/engine/execcmdagent/manager_linux_test.go
```diff
@@ -0,0 +1,363 @@
...
+		createContainerExecRes     *types.IDResponse
+		createContainerExecErr     error
+		expectStartContainerExec   bool
+		startContainerExecErr      error
+		expectInspectContainerExec bool
+		inspectContainerExecRes    *types.ContainerExecInspect
+		inspectContainerExecErr    error
+		expectedError              error
+	}{
+		{
```
fierlion, line 68: This test is great.  I love that I can see the lifecycle of the start process directly from these test cases.

agent/engine/execcmdagent/manager.go
angelcar: What would be the alternative? Ignore?

agent/engine/execcmdagent/manager_linux.go
angelcar: So, if the ssm-agent exits with a zero value, we don't want to restart it?

angelcar: P.S. I don't think we have access to the signals ssm agent receives through the docker API, all we have is the exit code.

agent/engine/execcmdagent/manager_linux.go
angelcar: Will add in next rev

agent/engine/execcmdagent/manager_linux.go
fierlion: Yeah I wonder if we should just always restart n-times, regardless of exit code.  We'll need a good way of surfacing what exit code we're getting.

agent/engine/execcmdagent/manager.go
fierlion: I think agent doesn't update status until it is in a known state.  So we need to find a way of absolutely limiting our time spent in unknown-land.

fierlion (approved):

sharanyad (approved):
code lgtm

PR #2651, opened 2020-09-29
Refactor PostUnmarshalTask to abstract away ExecCommandAgent complexity

angelcar:
### Summary
Refactor to abstract ExecCommandAgent-related stuff away from Task. Now only engine and execcmd packages are aware of ExecCommandAgent.

### Implementation details
* Changed `execcmdagent` package name to `execcmd` because now aside from agent-specific stuff, this package has `InitializeTask` method.
* Added bind mount for ssm-agent configuration file, which was previously missing
* Other than the new bind mount above, there were no changes in logic, just moved code around

### Testing

* UTs
* Integ tests
* Manually verified that the ExecComnmandAgent still runs and is able to start new sessions in a container created from scratch
* Manually killed the ExecComnmandAgent in the host and verified it was automatically restarted by ecs agent
New tests cover the changes: 

### Licensing


agent/api/task/task_option.go
```diff
@@ -0,0 +1,5 @@
+package task
+
+// Option defines a function that can be executed in some task API methods in order to perform custom configurations
+// of the tsk
```
fierlion, line 4: nit `task`

agent/engine/execcmd/manager_unsupported.go
```diff
@@ -1,4 +1,4 @@
-// +build !linux,!windows
+// +build !linux,!unit
```
fierlion, line 1: `!windows`?

angelcar: If it's windows I do want this code so that all exec cmd stuff is a no-op

angelcar: The reason I removed that was that I had a separate `manager_windows.go` file, but I realized that was not needed and I can just have this one

agent/api/task/task_option.go
angelcar: fixed :)

agent/api/task/task_option.go
```diff
@@ -0,0 +1,5 @@
+package task
```
sharanyad, line 1: package header missing

agent/api/task/task.go
```diff
@@ -325,7 +325,7 @@ func (task *Task) initializeVolumes(cfg *config.Config, dockerClient dockerapi.D
 // able to handle such an occurrence appropriately (e.g. behave idempotently).
 func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	credentialsManager credentials.Manager, resourceFields *taskresource.ResourceFields,
-	dockerClient dockerapi.DockerClient, ctx context.Context) error {
+	dockerClient dockerapi.DockerClient, ctx context.Context, options ...Option) error {
```
sharanyad, line 328: I like the idea, but call to `PostUnmarshalTask` might become cumbersome with too many options

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -0,0 +1,151 @@
+// +build linux
```
sharanyad, line 1: different methods implementation (init task/start agent) could live in the same file? `manager_linux.go`?

angelcar: They could, it's a matter of preference. I prefer not to have to much unrelated code in the same file and keep files short and readable.

agent/api/task/task.go
angelcar: It could, yes. I can't think of a better alternative that doesn't involve adding unnecessary complexity to `task` though. (like it is before this refactor).
The whole `PostUnmarshalTask` is already cumbersome , IMHO.

Is there an alternative there?

agent/api/task/task_option.go
angelcar: good catch, thanks.

agent/api/task/task.go
angelcar: if what worries you is that we might end up in the future with a call like:

```golang
task.PostUnmarshalTask(cfg, credentialsManager, resourceFields, dockerClient, ctx,
opt1,
opt2,
opt3,
opt4,
opt5)
```

We can always create an option like:

```golang
func compositeOpt(task *Task) {
    opt1(task)
    opt2(task)
    opt3(task)
    opt4(task)
    opt5(task)
}
```

And then just call `PostUnmarshalTask` like this:

```golang
task.PostUnmarshalTask(cfg, credentialsManager, resourceFields, dockerClient, ctx, compositeOpt)
```

agent/engine/execcmd/manager_init_task_linux.go
sharanyad: its ok to keep it this way, just that I haven't seen examples like this in our code base. not sure if there is a best practice

agent/api/task/task.go
sharanyad: ya, we could refactor later if needed, I can't think of a better alternative 😅

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -0,0 +1,151 @@
...
+	BinName              = "amazon-ssm-agent"
+	SessionWorkerBinName = "ssm-session-worker"
+	ConfigFileName       = "amazon-ssm-agent.json"
+
+	// TODO: [ecs-exec] decide if this needs to be configurable or put in a specific place in our optimized AMIs
+	HostLogDir          = "/var/log/ecs/exec"
+	ContainerLogDir     = "/var/log/amazon/ssm"
+	HostCertFile        = "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"
+	ContainerCertFile   = "/etc/ssl/certs/ca-certificates.crt"
+	ContainerConfigFile = "/etc/amazon/ssm/amazon-ssm-agent.json"
```
sharanyad, line 41: I *believe* this might be specific to each container, but this needs more discussion.

sharanyad (approved):

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -0,0 +1,151 @@
...
+	namelessContainerPrefix = "nameless-container-"
+	internalNamePrefix      = "internal-execute-command-agent"
+	logVolumeNamePrefix     = internalNamePrefix + "-log"
+	certVolumeName          = internalNamePrefix + "-tls-cert"
+	configVolumeName        = internalNamePrefix + "-config"
+)
+
+// InitializeTask adds the necessary volumes and mount points in all of the task containers in order for the
+// exec cmd agent to run upon container start up.
+// TODO: [ecs-exec] Should we validate the ssm agent binaries & certs are valid and fail here if they're not? (bind mount will succeed even if files don't exist in host)
```
fierlion, line 52: Either here, or in our execute-command capability enablement...
or both.  I think if we check before we enable the capability, that will be enough

fierlion (approved):

PR #2652, opened 2020-09-29
Add ManagedAgent To container state change event

fierlion:
### Summary
This adds the ManagedAgent to the container state change events that we emit. 
Currently we only have a single ManagedAgent -- the ExecuteCommandAgent.  I've hard coded this in a few places.
The acceptable statuses for a ManagedAgent are `PENDING|RUNNING|STOPPED`, but we have a few extra statuses that we can track inside the agent -- namely `CREATED` (which should mark the time between the bind-mount and the ExecuteCommandAgent process actually starting) and `NONE` for the empty status.  
I've tried to follow existing patterns as closely as possible. 

### Implementation details
I've added the `ManagedAgentStatus` struct with an enum of acceptable statuses.
This is cooked into the `ManagedAgent` struct.
I've built and updated the backend api model based on latest changes in dev.
the `GetKnownManagedAgents()` function will create an array of 0 or 1 ManagedAgents to be passed in the container state change events.
Next step will be to wire this in via the `emitContainerEvent()` function in the task_manager

### Testing

I added unit tests to test functions up through `GetKnownManagedAgents()` function.  Ran all unit/integ tests locally and they passed.

Add ManagedAgent to container state change event.

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -80,6 +80,9 @@ const (
 
 	// neuronVisibleDevicesEnvVar is the env which indicates that the container wants to use inferentia devices.
 	neuronVisibleDevicesEnvVar = "AWS_NEURON_VISIBLE_DEVICES"
+
+	// executeCommandAgent is the only available ManagedAgent type
+	executeCommandAgentName = "EXECUTE_COMMAND_AGENT"
 )
 
 // DockerConfig represents additional metadata about a container to run. It's
@@ -106,6 +109,17 @@ type HealthStatus struct {
 	Output string `json:"output,omitempty"`
 }
 
+type ManagedAgent struct {
+	// Status is the managed agent health status
+	Status apicontainerstatus.ManagedAgentStatus `json:"status:omitempty"`
+	// Name is always executeCommandAgent (until a new agent type is introduced)
+	Name string `json:"name,omitempty"`
+	// Reason is a placeholder for failure messaging
+	Reason string `json:"reason,omitempty"`
+	// LastStartedAt is the timestamp when the status last went from PENDING->RUNNING
+	LastStartedAt *time.Time `json:"lastStartedAt,omitempty"`
+}
+
 // Container is the internal representation of a container in the ECS agent
 type Container struct {
 	// Name is the name of the container specified in the task definition
@@ -117,6 +131,8 @@ type Container struct {
 	TaskARNUnsafe string `json:"taskARN"`
 	// DependsOnUnsafe is the field which specifies the ordering for container startup and shutdown.
 	DependsOnUnsafe []DependsOn `json:"dependsOn,omitempty"`
+	// ManagedAgentsUnsafe presently contains only the executeCommandAgent
+	ManagedAgentsUnsafe []ManagedAgent `json:"managedAgents,omitempty"`
 	// V3EndpointID is a container identifier used to construct v3 metadata endpoint; it's unique among
 	// all the containers managed by the agent
 	V3EndpointID string
@@ -269,14 +285,21 @@ type Container struct {
 	// ContainerArn is the Arn of this container.
 	ContainerArn string `json:"ContainerArn,omitempty"`
 
-	// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
-	ExecCommandAgentMetadata *ExecCommandAgentMetadata `json:"ExecCommandAgentMetadata"`
-
 	createdAt  time.Time
 	startedAt  time.Time
 	finishedAt time.Time
 
 	labels map[string]string
+
+	// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
+	ExecCommandAgentMetadata *ExecCommandAgentMetadata `json:"execCommandAgentMetadata"`
+
+	// ExecCommandAgentStatusUnsafe is the ExecCommandAgent status
+	ExecCommandAgentStatusUnsafe apicontainerstatus.ManagedAgentStatus `json:"execCommandAgentStatus,omitempty"`
+
+	// execCommandAgentStartedAt is the latest time the execCommandAgent process was started
+	// this will be Zero value if the execCommandAgent process has never been started (ex: created/provisioned but not started)
+	execCommandAgentStartedAt time.Time
 }
 
 type DependsOn struct {
@@ -680,6 +703,68 @@ func (c *Container) GetKnownPortBindings() []PortBinding {
 	defer c.lock.RUnlock()
 
 	return c.KnownPortBindingsUnsafe
+}
+
+// SetKnownExecCommandAgentStatus sets the desired status of the execCommandAgent
+func (c *Container) SetKnownExecCommandAgentStatus(status apicontainerstatus.ManagedAgentStatus) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	c.ExecCommandAgentStatusUnsafe = status
+}
+
+// GetExecCommandAgentStatus gets the desired status of the execCommandAgent
+func (c *Container) GetKnownExecCommandAgentStatus() apicontainerstatus.ManagedAgentStatus {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.ExecCommandAgentStatusUnsafe
+}
+
+// SetExecCommandAgentStartedAt sets the timestamp for execCommandeAgent started time
+func (c *Container) SetExecCommandAgentStartedAt(startedAt time.Time) {
+	if startedAt.IsZero() {
+		return
+	}
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
+
+	c.execCommandAgentStartedAt = startedAt
+}
+
+// GetExecCommandAgentStartedAt gets the timestamp for execCommandAgent started time
+func (c *Container) GetExecCommandAgentStartedAt() time.Time {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+
+	return c.execCommandAgentStartedAt
+}
+
+// GetKnownManagedAgents is a convenience method to format the
+// ExecuteCommandAgent to meet the expectations of ContainerStateChange
+// This returns an initialized array of ManagedAgents if none are reporting
+// Else an array of knownManagedAgents
+func (c *Container) GetKnownManagedAgents() []ManagedAgent {
+	knownManagedAgentStatus := c.GetKnownExecCommandAgentStatus()
+	knownManagedAgentStart := aws.Time(c.GetExecCommandAgentStartedAt())
+
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	// for now we will reset the managed agents every time.
+	c.ManagedAgentsUnsafe = []ManagedAgent{}
+	if knownManagedAgentStatus.ShouldReportToBackend() {
+		// LastStartedAt will only exist for ManagedAgents
+		// that have been started at least once
+		knownManagedAgent := ManagedAgent{
+			Status:        knownManagedAgentStatus,
+			Name:          executeCommandAgentName,
+			LastStartedAt: knownManagedAgentStart,
+		}
+		c.ManagedAgentsUnsafe = append(c.ManagedAgentsUnsafe, knownManagedAgent)
+	}
+
+	return c.ManagedAgentsUnsafe
 }
 
 // SetVolumes sets the volumes mounted in a container
--- a/agent/api/container/status/managedagentstatus.go
+++ b/agent/api/container/status/managedagentstatus.go
@@ -1,0 +1,109 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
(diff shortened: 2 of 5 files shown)


agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// String returns a human readable string representation of this object
+func (mas ManagedAgentStatus) String() string {
+	for k, v := range managedAgentStatusMap {
+		if v == mas {
+			return k
+		}
+	}
+	return "NONE"
```
angelcar, line 49: Hmm, this is odd because NONE is already in the map, so technically this would be an unknown state or something

agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,112 @@
...
+	return mas == ManagedAgentStopped
+}
+
+// IsRunning returns true if the managed agent status is either RUNNING
+func (mas ManagedAgentStatus) IsRunning() bool {
+	return mas == ManagedAgentRunning
+}
+
+// BackendStatus maps the internal managedAgent status in the agent to that in the backend
+func (mas ManagedAgentStatus) BackendStatus() string {
```
angelcar, line 63: What's the purpose of this? I don't think there's a possible way for this method to return `"PENDING"`. This is because it will always return `mas.String()`, so in effect this is a weird alias for `mas.String()`

https://goplay.corp.amazon.com/p/0766add6194da9cf06b11cee6ad9a7b7ef1cd411

agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return "PENDING"
+}
+
+// ShouldReportToBackend returns true if the managedAgent status is recognized as a
+// valid state by ECS. Note that not all managedAgent statuses are recognized by ECS
+// or map to ECS states
+// For now we expect PENDING/RUNNING/STOPPED as valid states
+// we should only skip events that have ManagedAgentStatusNone
+func (mas *ManagedAgentStatus) ShouldReportToBackend() bool {
```
angelcar, line 82: any reason why the receiver for this is a pointer but for the previous methods is an struct? (i.e. `mas *ManagedAgentStatus` vs `mas ManagedAgentStatus`

angelcar: Same for method below.

agent/api/container/container.go
```diff
@@ -269,14 +283,21 @@ type Container struct {
...
+
+	// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
+	ExecCommandAgentMetadata *ExecCommandAgentMetadata `json:"execCommandAgentMetadata"`
+
+	// ExecCommandAgentStatusUnsafe is the ExecCommandAgent status
+	ExecCommandAgentStatusUnsafe apicontainerstatus.ManagedAgentStatus `json:"execCommandAgentStatus,omitempty"`
+
+	// execCommandAgentStartedAt is the latest time the execCommandAgent process was started
+	// this is nil if the execCommandAgent process has never been started (ex: created/provisioned but not started)
+	execCommandAgentStartedAt time.Time
```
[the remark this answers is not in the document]
fierlion: (meant zero value)

agent/api/container/status/managedagentstatus.go
fierlion: yeah "NONE" would act as a catchall but it's not used presently (I hope).

agent/api/container/status/managedagentstatus.go
fierlion: ManagedAgentCreated will return "PENDING" (because explicit `fallthrough`)

angelcar: `fallthrough` means kind of "execute the case below instead, which is 
```golang
ManagedAgentRunning:
    return mas.String()
```
Demo: https://play.golang.org/p/2Dk8zY4WgQj

agent/api/container/status/managedagentstatus.go
sparrc: +1, this function should be non-pointer I think, UnmarshalJSON probably has to be pointer though in order to fulfil the interface.

agent/api/container/status/managedagentstatus.go
fierlion: I made a change to this and to the unit test:
Backend status should always return a valid backend status.  This is just meant to force-map the status to one of `PENDING|RUNNING|STOPPED`

agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,109 @@
...
+	// ManagedAgentRunning represents a managed agent that has started sucessfully
+	ManagedAgentRunning
+	// ManagedAgentStopped represents a managed agent that has stopped
+	ManagedAgentStopped
+)
+
+// ManagedAgentStatus is an enumeration of valid states in the managed agent lifecycle
+type ManagedAgentStatus int32
+
+var managedAgentStatusMap = map[string]ManagedAgentStatus{
```
sparrc, line 35: nit: to me makes more sense to reverse this map, then you can eliminate the for-loop from the String() function like:
```go
// String returns a human readable string representation of this object
func (mas ManagedAgentStatus) String() string {
    if s, ok := managedAgentStatusMap[mas]; ok {
        return s
    }
    return "NONE"
}
```

agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,109 @@
...
+	if !ok {
+		*mas = ManagedAgentStatusNone
+		return errors.New("managed agent status unmarshal: unrecognized status")
+	}
+	*mas = stat
+	return nil
+}
+
+// MarshalJSON overrides the logic for JSON-encoding the ManagedAgentStatus type
+func (mas *ManagedAgentStatus) MarshalJSON() ([]byte, error) {
```
sparrc, line 104: this function can just be `return json.Marshal(mas.String())`

agent/api/container/status/managedagentstatus.go
fierlion: stole it from: https://github.com/aws/amazon-ecs-agent/blob/782948476da6d995ad33c6a04130f8172820af27/agent/api/container/status/containerstatus.go#L97-L99
I'll make the update and test.

agent/api/container/status/managedagentstatus_test.go
```diff
@@ -0,0 +1,95 @@
...
+	assert.Equal(t, true, managedAgentStatus.ShouldReportToBackend())
+	// We should report ManagedAgentStooped (ie STOPPED)
+	managedAgentStatus = ManagedAgentStopped
+	assert.Equal(t, true, managedAgentStatus.ShouldReportToBackend())
+}
+
+func TestManagedAgentBackendStatus(t *testing.T) {
+	// BackendStatus is "PENDING" when managedAgent status is ManagedAgentStatusNone
+	var managedAgentStatus ManagedAgentStatus
+	assert.Equal(t, managedAgentStatus.BackendStatus(), "PENDING")
```
angelcar, line 57: nit: expected is first: `assert.Equal(t, "PENDING", managedAgentStatus.BackendStatus())`

agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,109 @@
...
+// permissions and limitations under the License.
+
+package status
+
+import (
+	"errors"
+	"strings"
+)
+
+const (
```
sparrc, line 21: while I like using the iota/enum in Go for certain situations, Im not sure it makes sense here since we need to frequently marshal these values in and out of json, which means we have to define custom Marshal and Unmarshal functions.

My prefrence would be to just define these as string constants to avoid all the mapping and custom json marshalling, like:

```go
type ManagedAgentStatus string

const (
	// ManagedAgentStatusNone is the zero state of a managed agent
	ManagedAgentStatusNone = "NONE"
	// ManagedAgentCreated represents a managed agent which has dependencies in place
	ManagedAgentCreated = "CREATED"
	// ManagedAgentRunning represents a managed agent that has started sucessfully
	ManagedAgentRunning = "RUNNING"
	// ManagedAgentStopped represents a managed agent that has stopped
	ManagedAgentStopped = "STOPPED"
)
```

fierlion: For this as well as the Marshal/Unmarshal, I'm following the pattern in our other state files for consistency.

sparrc: There's only one other file doing this isn't there? doesn't seem like enough of a precedent to follow a poor code example

agent/api/container/status/managedagentstatus_test.go
fierlion: updated

agent/api/container/status/managedagentstatus.go
fierlion: https://github.com/aws/amazon-ecs-agent/search?q=iota <- check out how often status and `iota` happen.  This is a pattern.  And many of these end up with our custom marshal/unmarshal.

sparrc: Fair enough, I only looked in the same directory so didn't see all of these.

I think most of this code is just following a pattern that was defined many years ago -- I would still prefer to define a new pattern that can be used for future statuses and that hopefully could be used as an example to migrate the old ones to. 

IMHO the less custom json marshalling the better :)

angelcar: I agree that this custom marshaling is unnecessary complexity. I guess we can refactor that either now or later, not a blocker for this PR though

angelcar (approved):

agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,109 @@
...
+	// ManagedAgentStatusNone is the zero state of a managed agent
+	ManagedAgentStatusNone ManagedAgentStatus = iota
+	// ManagedAgentCreated represents a managed agent which has dependencies in place
+	ManagedAgentCreated
```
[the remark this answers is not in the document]
fierlion: we don't track this as a backend status, but should we track the state between bind mount and agent process start?  That's where I see the ManagedAgentCreated apply to executeCommandAgent.

And we don't know that exec agent won't need to follow state transitions order at some point.  for now we only have Created/Running/Stopped but we also might need to add Failed state, as well as potentially adding a separate health field (when we look into RUNNING & CONNECTED or potentially just DISCONNECTED.

agent/api/container/status/managedagentstatus.go
```diff
@@ -0,0 +1,109 @@
...
+	return "PENDING"
+}
+
+// ShouldReportToBackend returns true if the managedAgent status is recognized as a
+// valid state by ECS. Note that not all managedAgent statuses are recognized by ECS
+// or map to ECS states
+// For now we expect PENDING/RUNNING/STOPPED as valid states
+// we should only skip events that have ManagedAgentStatusNone
+func (mas ManagedAgentStatus) ShouldReportToBackend() bool {
+	return mas == ManagedAgentCreated || mas == ManagedAgentRunning || mas == ManagedAgentStopped
```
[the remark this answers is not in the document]
fierlion: 1. `Created` status would map to the backend PENDING status and would mark the space between bind mount and actual execute command agent process start.
2. By this I'm assuming future managed agents would follow a similar bind mount/exec start process.
Since we're calling this a `ManagedAgent` I'm trying to adhere to that principle rather than hard-coding `executeCommandAgent` everywhere.  I don't foresee more such managedAgents, but there's a reason we decided to  define a generic structure, and this is why I've defined this status as such.
When we have another managedAgent, we should continue this conversation.

agent/api/container/container.go
```diff
@@ -269,14 +283,21 @@ type Container struct {
...
 	startedAt  time.Time
 	finishedAt time.Time
 
 	labels map[string]string
+
+	// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
+	ExecCommandAgentMetadata *ExecCommandAgentMetadata `json:"execCommandAgentMetadata"`
+
+	// ExecCommandAgentStatusUnsafe is the ExecCommandAgent status
+	ExecCommandAgentStatusUnsafe apicontainerstatus.ManagedAgentStatus `json:"execCommandAgentStatus,omitempty"`
```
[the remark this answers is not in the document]
fierlion: I'll play with this a bit.
I'm trying to walk the line between our generic `ManagedAgent` and a simpler initial implementation.  I like the idea of keeping the executeCommandAgent stuff in the Container direct and straightforward for now, but also agree it's sort of a mixed message when I'm creating generic `ManagedAgentStatus` elsewhere.

agent/api/container/status/managedagentstatus.go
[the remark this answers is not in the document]
sharanyad: 1. I think my concern is more around introducing a new state, which might not be used for any purpose. 
It does indicate that bind mount is done, but how are we using the status, does it add any value? We could always know that execAgent is 'created' if the container has moved to 'created'. 
One place I can think this might help is during agent restarts, to know if execAgent is in  'created' state, but again irrespective of that, container status would already indicate that.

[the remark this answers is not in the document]
sharanyad: also shouldn't we report only if its running or stopped? right now its reporting for Created as well (which backend is not aware of).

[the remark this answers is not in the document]
sharanyad: For 2. I was thinking more around a generic ManagedAgent interface with methods like `ShouldReportToBackend` and each managed agent has its own implementation, kind of like our task resource model. But at this point, it might be an overkill. We can refactor when we add more managedAgents.

agent/api/container/container.go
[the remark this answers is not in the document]
fierlion: See my most recent revision -- (still using the `GetKnownManagedAgents` method, but now with a container level `ManagedAgentsUnsafe` array.

sharanyad (approved):

PR #2656, opened 2020-10-02
add timeout to poll Docker stats

sharanyad:
### Summary
Currently, while polling docker stats, there is no timeout for the API call. So the call could be stuck until the container is stopped. 
Added poll stats timeout (for now, keeping it the same as streams stats inactivity timeout) 

### Testing

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -84,6 +84,10 @@ const (
 	maximumPullRetryDelay     = 5 * time.Second
 	pullRetryDelayMultiplier  = 2
 	pullRetryJitterMultiplier = 0.2
+
+	// pollStatsTimeout is the timeout for polling Docker Stats API;
+	// keeping it same as streaming stats inactivity timeout
+	pollStatsTimeout = 18 * time.Second
 )
 
 var ctxTimeoutStopContainer = dockerclient.StopContainerTimeout
@@ -1366,7 +1370,7 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 			defer close(statsC)
 			// we need to start by getting container stats so that the task stats
 			// endpoint will be populated immediately.
-			stats, err := getContainerStatsNotStreamed(client, subCtx, id)
+			stats, err := getContainerStatsNotStreamed(client, subCtx, id, pollStatsTimeout)
 			if err != nil {
 				errC <- err
 				return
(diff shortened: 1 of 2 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1394,21 +1398,35 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
...
+		decoder := json.NewDecoder(resp.stats.Body)
+		stats := &types.StatsJSON{}
+		err := decoder.Decode(stats)
+		if err != nil {
+			return nil, fmt.Errorf("DockerGoClient: Unable to decode stats for container %s: %v", id, err)
+		}
+		defer resp.stats.Body.Close()
+		return stats, nil
+	case <-ctxWithTimeout.Done():
+		err := ctx.Err()
```
sparrc, line 1424: should this be `ctxWithTimeout.Err()`?

sharanyad: modified.

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1394,21 +1398,35 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
...
+		if err != nil {
+			return nil, fmt.Errorf("DockerGoClient: Unable to decode stats for container %s: %v", id, err)
+		}
+		defer resp.stats.Body.Close()
+		return stats, nil
+	case <-ctxWithTimeout.Done():
+		err := ctxWithTimeout.Err()
+		if err == context.DeadlineExceeded {
+			return nil, fmt.Errorf("DockerGoClient: timed out retrieving stats for container %s", id)
+		}
```
sparrc, line 1427: maybe outside scope of this change, but relatedly I was wondering if we might want to catch and ignore `context.Canceled` errors here? I believe we end up logging these as a warning or error, but they are actually expected to happen when the container exits.

Not sure if there might be some other implication to ignoring these errors though...

sharanyad: I think its better to just log them since context cancel could also happen for reasons other than container exit (we log them for other docker APIs too)

sparrc (approved):

mythri-garaga (approved):

mythri-garaga:
Please pull #2655 changes for `ecs/arm/unit_test` test to pass.

PR #2658, opened 2020-10-05
Updating Makefile after the switch to using Go Modules

mythri-garaga:
### Summary

Updating Makefile after the switch to using Go Modules
### Implementation details

Updated the following targets in makefile
- `test`
- `test-silent`
- `analyze-cover-profile`
- `run-integ-tests`
-  `run-sudo-tests`
- `benchmark-test`

### Testing

Tested these make targets locally 

New tests cover the changes: 
### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,26 +118,23 @@ ifneq (${BUILD_PLATFORM},aarch64)
 endif
 
 test:
-	${GOTEST} -tags unit -coverprofile cover.out -timeout=60s ./agent/...
-	go tool cover -func cover.out > coverprofile.out
+	@cd agent && $(MAKE) test
 
 test-silent:
-	$(eval VERBOSE=)
-	${GOTEST} -tags unit -coverprofile cover.out -timeout=60s ./agent/...
-	go tool cover -func cover.out > coverprofile.out
+	@cd agent && $(MAKE) test-silent
 
 .PHONY: analyze-cover-profile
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -118,26 +118,23 @@ ifneq (${BUILD_PLATFORM},aarch64)
...
+	@cd agent && $(MAKE) test-silent
 
 .PHONY: analyze-cover-profile
-analyze-cover-profile: coverprofile.out
+analyze-cover-profile: ./agent/coverprofile.out
 	./scripts/analyze-cover-profile
 
 run-integ-tests: test-registry gremlin container-health-check-image run-sudo-tests
-	ECS_LOGLEVEL=debug ${GOTEST} -tags integration -timeout=30m ./agent/...
+	@cd agent && ECS_LOGLEVEL=debug GO111MODULE=on ${GOTEST} -tags integration -timeout=30m ./...
```
angelcar, line 131: this is more out of ignorance, but wouldn't the working dire be messed up if two targets are executed sequentially? say `run-integ-tests`, which invokes `run-sudo-tests`, so there is a `cd agent`, then once we're in `agent` folder, there would be another `cd agent`, which would not exist?

mythri-garaga: these two targets run in different shell so directory will not be messed up in this case.

sparrc (approved):

mssrivas (approved):

PR #2662, opened 2020-10-07
Add on prem config and block imds when config is set

fenxiong:
### Summary

Add on prem config and block imds when config is set.

### Implementation details

Added config in agent/config/config.go. Set ec2 metadata client to the fake client when the on-prem config is set.
 
### Testing

Unit test added. Built the agent and ran it with ECS_ON_PREM set to true, verified that it's able to start, register and run task on prem.

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -143,6 +143,12 @@ func newAgent(blackholeEC2Metadata bool, acceptInsecureCert *bool) (agent, error
 	seelog.Infof("Amazon ECS agent Version: %s, Commit: %s", version.Version, version.GitShortHash)
 	seelog.Debugf("Loaded config: %s", cfg.String())
 
+	if cfg.OnPrem.Enabled() {
+		seelog.Info("Running in on-prem mode.")
+		ec2MetadataClient = ec2.NewBlackholeEC2MetadataClient()
+		cfg.NoIID = true
+	}
+
 	ec2Client := ec2.NewClientImpl(cfg.AWSRegion)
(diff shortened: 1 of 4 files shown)


mssrivas (approved):

agent/app/agent.go
```diff
@@ -143,6 +143,12 @@ func newAgent(blackholeEC2Metadata bool, acceptInsecureCert *bool) (agent, error
 	seelog.Infof("Amazon ECS agent Version: %s, Commit: %s", version.Version, version.GitShortHash)
 	seelog.Debugf("Loaded config: %s", cfg.String())
 
+	if cfg.OnPrem.Enabled() {
```
sparrc, line 146: should we move this up to line 125 so that `ec2.NewEC2MetadataClient(nil)` is never called?

sparrc: then we wouldnt need to override inside of NewConfig either

fenxiong: at line 125, the config has not been initialized yet so we don't know whether its on prem or not. from my testing, invoking ec2.NewEC2MetadataClient on-prem seems fine (didn't cause any error or delay)

agent/config/config.go
```diff
@@ -226,6 +226,11 @@ func NewConfig(ec2client ec2.EC2MetadataClient) (*Config, error) {
 	}
 	config := &envConfig
 
+	if config.OnPrem.Enabled() {
```
sparrc, line 229: should we validate that AWSRegion has also been set here so that we don't get into the if-block that relies on the ec2 metadata client below?

fenxiong: yes good catch. updated to throw an error if AWSRegion is empty here

sparrc (approved):

PR #2664, opened 2020-10-07
Add and use new instancecreds package

sparrc:
This is to centralize grabbing our instance credential provider into a
single package. Previous to this change we grabbed instance credentials
in separate places for the ecs, ec2, ecr and ec2 metadata clients. With this
change, all of these clients will use the same instance credential
provider.

The reason for this is that if we want to add any non-default instance
credential providers to the credential chain, then we need to make sure
that all AWS clients are instantiated with the same provider.

### Testing

### Licensing

### DIFF
--- a/agent/credentials/instancecreds/instancecreds.go
+++ b/agent/credentials/instancecreds/instancecreds.go
@@ -1,0 +1,47 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package instancecreds
+
+import (
+	"sync"
+
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/cihub/seelog"
+)
+
+var (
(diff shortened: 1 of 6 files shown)


agent/credentials/instancecreds/instancecreds.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/cihub/seelog"
+)
+
+var (
+	credentialChain *credentials.Credentials
+	mu              sync.RWMutex
+)
+
+func GetCredentials() *credentials.Credentials {
```
fenxiong, line 29: nit: exported method should have comment

fenxiong (approved):

agent/credentials/instancecreds/instancecreds.go
```diff
@@ -0,0 +1,47 @@
...
+	if credentialChain == nil {
+		credentialChain = credentials.NewCredentials(&credentials.ChainProvider{
+			VerboseErrors: false,
+			Providers:     defaults.CredProviders(defaults.Config(), defaults.Handlers()),
+		})
+	}
+	mu.Unlock()
+
+	// credentials.Credentials is concurrency-safe, so lock not needed here
+	v, err := credentialChain.Get()
```
angelcar, line 40: This call didn't seem to happen before, why is it needed to have it here?

agent/credentials/instancecreds/instancecreds.go
sparrc: will fix in follow-up

agent/credentials/instancecreds/instancecreds.go
sparrc: we actually do call it currently, but in a different scope just before we register the container instance: https://github.com/aws/amazon-ecs-agent/blob/478c4ec00a43b4713a0acc811cfd1e2d70307f63/agent/app/agent.go#L513

It's not required to call, it's just informational to help with debugging.

angelcar (approved):

PR #2665, opened 2020-10-07
Add rotating shared credentials provider

sparrc:
For on-prem, we can assume that every instance will be managed by AWS Systems Manager (SSM). SSM will provide the agent with instance credentials in a file located at /root/.aws/credentials. That file contains credentials that are good for 60 minutes, and are rotated every 30 minutes (this is so that there is overlap to avoid race conditions, where the “old” credentials are still usable for some time after rotation). The format of this file will be the same as the [AWS shared credentials file](https://docs.aws.amazon.com/ses/latest/DeveloperGuide/create-shared-credentials-file.html).

While the AWS SDK supports retrieving credentials from the shared credentials file (via AWS_SHARED_CREDENTIALS_FILE env var), it does not support refreshing credentials when they expire. Because of this, the agent needs it’s own credential provider that parses the shared credentials file and refreshes it every 60 seconds to check for newer creds.

### Testing

Added a rotating shared credentials file instance credential provider

### Licensing

### DIFF
--- a/agent/credentials/providers/rotating_shared_credentials_provider.go
+++ b/agent/credentials/providers/rotating_shared_credentials_provider.go
@@ -1,0 +1,75 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package providers
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/cihub/seelog"
+)
+
+const (
+	// defaultRotationInterval is how frequently to expire and re-retrieve the credentials from file.
+	defaultRotationInterval = time.Minute
+	// defaultFilename is the default location of the credentials file within the container.
+	defaultFilename = "/rotatingcreds/credentials"
+	// RotatingSharedCredentialsProviderName is the name of this provider
+	RotatingSharedCredentialsProviderName = "RotatingSharedCredentialsProvider"
+)
+
+// RotatingSharedCredentialsProvider is a provider that retrieves credentials via the
+// shared credentials provider, and adds the functionality of expiring and re-retrieving
+// those credentials from the file.
+type RotatingSharedCredentialsProvider struct {
+	credentials.Expiry
+
+	RotationInterval          time.Duration
+	sharedCredentialsProvider *credentials.SharedCredentialsProvider
+}
+
+// NewRotatingSharedCredentials returns a rotating shared credentials provider
+// with default values set.
+func NewRotatingSharedCredentialsProvider() *RotatingSharedCredentialsProvider {
+	return &RotatingSharedCredentialsProvider{
+		RotationInterval: defaultRotationInterval,
+		sharedCredentialsProvider: &credentials.SharedCredentialsProvider{
+			Filename: defaultFilename,
+			Profile:  "default",
+		},
+	}
+}
+
+// Retrieve will use the given filename and profile and retrieve AWS credentials.
+func (p *RotatingSharedCredentialsProvider) Retrieve() (credentials.Value, error) {
+	v, err := p.sharedCredentialsProvider.Retrieve()
+	v.ProviderName = RotatingSharedCredentialsProviderName
+	if err != nil {
+		return v, err
+	}
+	p.SetExpiration(time.Now().Add(p.RotationInterval), 0)
+	seelog.Infof("Successfully got instance credentials from file %s. %s",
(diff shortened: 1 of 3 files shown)


agent/credentials/providers/rotating_shared_credentials_provider.go
```diff
@@ -0,0 +1,65 @@
...
+			Filename: DefaultFilename,
+			Profile:  "default",
+		},
+	}
+}
+
+// Retrieve will use the given filename and profile and retrieve AWS credentials.
+func (p *RotatingSharedCredentialsProvider) Retrieve() (credentials.Value, error) {
+	v, err := p.sharedCredentialsProvider.Retrieve()
+	v.ProviderName = RotatingSharedCredentialsProviderName
```
angelcar, line 57: Is it safe to assume we can use `v` before checking the error? could it be `nil`?

agent/credentials/providers/rotating_shared_credentials_provider.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return v, err
+	}
+	p.SetExpiration(time.Now().Add(p.RotationInterval), 0)
+	seelog.Infof("Successfully got instance credentials from file %s. Access Key ID XXXX%s from %s",
+		p.sharedCredentialsProvider.Filename, v.AccessKeyID[len(v.AccessKeyID)-4:], v.ProviderName)
```
angelcar, line 63: this will panic if someone puts an invalid key id with less than 4 characters

agent/credentials/providers/rotating_shared_credentials_provider.go
```diff
@@ -0,0 +1,65 @@
...
+	DefaultFilename = "/rotatingcreds/credentials"
+	// RotatingSharedCredentialsProviderName is the name of this provider
+	RotatingSharedCredentialsProviderName = "RotatingSharedCredentialsProvider"
+)
+
+// RotatingSharedCredentialsProvider is a provider that retrieves credentials via the
+// shared credentials provider, and adds the functionality of expiring and re-retrieving
+// those credentials from the file.
+type RotatingSharedCredentialsProvider struct {
+	credentials.Expiry
```
angelcar, line 36: do we want to say `RotatingSharedCredentialsProvider` is a `credentials.Expiry`? I think this probably should be a member rather than a base.

agent/credentials/providers/rotating_shared_credentials_provider.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/cihub/seelog"
+)
+
+const (
+	// DefaultRotationInterval is how frequently to expire and re-retrieve the credentials from file.
+	DefaultRotationInterval = time.Minute
+	// DefaultFilename is the default location of the credentials file within the container.
+	DefaultFilename = "/rotatingcreds/credentials"
```
angelcar, line 27: Do we want to export these defaults?

agent/credentials/providers/rotating_shared_credentials_provider.go
```diff
@@ -0,0 +1,65 @@
...
+	sharedCredentialsProvider *credentials.SharedCredentialsProvider
+}
+
+// NewRotatingSharedCredentials returns a rotating shared credentials provider
+// with default values set.
+func NewRotatingSharedCredentialsProvider() *RotatingSharedCredentialsProvider {
+	return &RotatingSharedCredentialsProvider{
+		RotationInterval: DefaultRotationInterval,
+		sharedCredentialsProvider: &credentials.SharedCredentialsProvider{
+			Filename: DefaultFilename,
```
angelcar, line 48: should the filename be configurable?

agent/credentials/providers/rotating_shared_credentials_provider.go
sparrc: yes this is how the sdk recommends to use it, see https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/credentials.go#L146-L156

agent/credentials/providers/rotating_shared_credentials_provider.go
sparrc: we probably dont need to, I can change that

agent/credentials/providers/rotating_shared_credentials_provider.go
sparrc: It's currently not because we don't have reason to make it configurable. We will ask customers running with docker to mount their creds into this directory, and if we integrate with ecs-init we will only expose the source (host) directory as configurable.

agent/credentials/providers/rotating_shared_credentials_provider.go
sparrc: this returns a `Value` struct, which is a struct and is not a pointer or interface. So the worst case scenario is that this would be empty. But luckily the SDK also makes a point of always populating it https://github.com/aws/aws-sdk-go/blob/master/aws/credentials/shared_credentials_provider.go#L54-L69

agent/credentials/providers/rotating_shared_credentials_provider.go
sparrc: good catch, will fix

agent/credentials/providers/rotating_shared_credentials_provider.go
sparrc: fixed

mythri-garaga (approved):

angelcar (approved):

PR #2666, opened 2020-10-08
On-prem capability/attribute support

fenxiong:
### Summary

Add on-prem capability/attribute support.

### Implementation details

When on-prem config is set:
1. Skip adding capabilities that aren't supported on-prem, this includes: task networking related capabilities, appmesh, EIA;
2. Add on prem capability "ecs.capability.on-prem";
3. Report cpu architecture as an attribute, obtained from Go runtime. Note: for amd64, the ec2 instance identity doc uses "x86_64" but go runtime has amd64 (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html). So when runtime is amd64, it is reported as x86_64 to be consistent with EC2.

Consolidated few unit tests to reduce duplicate test code.

### Testing

Added unit test. Built the agent and ran it on prem, verified that unsupported capabilities are not reported and cpu architecture is reported.

N/A

### Licensing

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -44,6 +44,8 @@ const (
 	pollEndpointCacheTTL    = 20 * time.Minute
 	roundtripTimeout        = 5 * time.Second
 	azAttrName              = "ecs.availability-zone"
+	cpuArchAttrName         = "ecs.cpu-architecture"
+	osTypeAttrName          = "ecs.os-type"
 )
 
 // APIECSClient implements ECSClient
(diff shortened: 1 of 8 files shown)


agent/api/ecsclient/client.go
```diff
@@ -325,12 +327,29 @@ func validateRegisteredAttributes(expectedAttributes, actualAttributes []*ecs.At
...
+		attrs = append(attrs, &ecs.Attribute{
+			Name:  aws.String(cpuArchAttrName),
+			Value: aws.String(getCPUArch()),
+		})
+	}
+	return attrs
+}
+
+func getCPUArch() string {
+	arch := runtime.GOARCH
```
sparrc, line 348: does instance identity document expect just `arm` for all arm systems? do they not distinguish between arm 32- and 64-bit?

sparrc: according to this doc (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) the IID returns `arm64`, whereas GOARCH would return just `arm`, I think we may also need to map arm -> arm64

fenxiong: i think our tests only have amd64 and arm64. added unit test for the two. for 386, i don't think we have it in any of our existing platform but i added logic to change it to i386 just to be safe.

sparrc (approved):

agent/app/agent_capability_test.go
```diff
@@ -142,6 +128,93 @@ func TestCapabilities(t *testing.T) {
 	}
 }
 
+// Test on-prem capability by checking that when on-prem config is set, unsupported capabilities aren't
+// added, and on-prem capability is added.
+func TestCapabilitiesOnPrem(t *testing.T) {
```
angelcar, line 133: It's good that we test that the unsupported capabilities are not present, but should we also test that the expected supported capabilities are there?
If in the future we add a new capability that should be on prem but for some error in logic it's ommited, nothing would tell us there's something wrong

angelcar (commented):
After looking at the capabilities golang file, it is starting to get contrived and it's not clear what capabilities are supported when. 

Wouldn't it be better to add all capabilities by default and then, depending on some conditions remove unsupported capabilities?

for example, declare unsupportedCapabilitiesOnPrems as a constant, add all capabilities to the array, then:
```golang
capabilities := []string{} // all capabilities
if onPrem {
    loop and remove unsupported capabilities on prem
}
```

IMO this makes it so much easier to know what is supported on prems, we can just go to the file and say.. everything but...

angelcar (approved):

PR #2667, opened 2020-10-08
Create sdkClientFactory only once for engine integration tests

angelcar:
### Summary

Recently, running integration tests had become problematic due to `too many open files` errors. This PR fixes the issue.

After running locally, I see around only 57 files open at any time while engine integ tests are running, from a previous value of 1000+

This change can have the potential benefit of making integ tests more efficient, since `dockerd` won't have hundreds/thousands of connections open concurrently.

### Implementation details

 I traced the source of open files to the creation of a new sdkClientFactory every time a new integ test ran. This caused all docker API calls to open new connections, which are cached (and not reused) by default.

By creating the factory only once in the `init` function of the `common_integ_test` file, only a fixed number of docker clients are created at initialization time (as opposed to every time the `setup` is invoked).

### Testing
Ran engine integ tests in my personal AL2 box and observed how the number of open connections to docker is dramatically less after this change.

### Licensing

### DIFF
--- a/agent/engine/common_integ_test.go
+++ b/agent/engine/common_integ_test.go
@@ -44,6 +44,14 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var (
+	sdkClientFactory sdkclientfactory.Factory
+)
+
+func init() {
+	sdkClientFactory = sdkclientfactory.NewFactory(context.TODO(), dockerEndpoint)
+}
+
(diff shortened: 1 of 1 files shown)


sparrc (approved):

ubhattacharjya (approved):

fierlion (approved):
yay!  Thank you!

PR #2672, opened 2020-10-12
Retry to rename files on error as os.Rename calls are prone to failure on Windows

mythri-garaga:
### Summary

Retry to rename files on error as os.Rename calls are prone to failure on Windows.

Similar issues: https://github.com/golang/go/issues/32188

### Implementation details

- `agent/taskresource/envFiles/envfile.go` - modified to retry the call os.Rename on error
- `agent/utils/retry/exponential_backoff.go` - corrected the function name in the comment

### Testing

`make test` passes
New tests cover the changes: 

Enhancement - Retry os.Rename() on error

### Licensing

### DIFF
--- a/agent/taskresource/envFiles/envfile.go
+++ b/agent/taskresource/envFiles/envfile.go
@@ -24,18 +24,19 @@ import (
 
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
+	"github.com/aws/amazon-ecs-agent/agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/s3"
 	"github.com/aws/amazon-ecs-agent/agent/s3/factory"
-	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
-	"github.com/aws/amazon-ecs-agent/agent/utils/oswrapper"
-	"github.com/cihub/seelog"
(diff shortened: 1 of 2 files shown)


agent/taskresource/envFiles/envfile.go
```diff
@@ -402,7 +407,10 @@ func (envfile *EnvironmentFileResource) writeEnvFile(writeFunc func(file oswrapp
 		return err
 	}
 
-	err = rename(tmpFile.Name(), fullPathName)
+	backoff := retry.NewExponentialBackoff(renameBackoffMin, renameBackoffMax, renameBackoffJitter, renameBackoffMultiple)
+	err = retry.RetryWithBackoff(backoff, func() error {
```
angelcar, line 411: can we update the unit test accordingly? I was reading the code of `retry.RetryWithBackoff` and if there is in fact an error with `rename`, the returning error won't be of type `apierrors.Retriable`. I think in case there's some unrecoverable error, this function will be stuck for ever.

mythri-garaga: Yes, I just saw the unit tests failing, I will update in new revision.

angelcar:
Any reason not to use the existing `retry.RetryNWithBackoff`?

mythri-garaga:
> Any reason not to use the existing `retry.RetryNWithBackoff`?

Thanks for catching this, I thought `retry.RetryNWithBackoff` only supported error types `apierrors.Retriable` but we can use this as it has limit on number of retries and wont be stuck in the loop forever on unrecoverable error

angelcar (approved):

mssrivas (approved):

PR #2677, opened 2020-10-14
add jitter between monitoring exec agents for containers

sharanyad:
### Summary
Monitoring for exec agent processes is done for tasks that have exec enabled every 15 minutes. To make sure all the exec containers do not synchronize on calling the inspect / restart agent docker API(thereby stressing docker), adding a jitter with a max sleep time of half of monitor interval for the containers. 

### Testing

N/A

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -299,6 +299,12 @@ func (engine *DockerTaskEngine) monitorExecAgentRunning(ctx context.Context,
 		seelog.Errorf("Task engine [%s]: Could not retrieve docker id for container", task.Arn)
 		return
 	}
+	// Sleeping here so that all the containers do not call inspect/start exec agent process
+	// at the same time.
+	// The max sleep is 50% of the monitor interval to allow enough buffer time
+	// to finish monitoring.
+	// This is inspired from containers streaming stats from Docker.
+	time.Sleep(retry.AddJitter(time.Nanosecond, engine.monitorExecAgentsInterval/2))
 	_, err = engine.execCmdMgr.RestartAgentIfStopped(ctx, engine.client, task, c, dockerID)
 	if err != nil {
 		seelog.Errorf("Task engine [%s]: Failed to restart ExecCommandAgent Process for container [%s]: %v", task.Arn, dockerID, err)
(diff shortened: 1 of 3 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -299,6 +299,12 @@ func (engine *DockerTaskEngine) monitorExecAgentRunning(ctx context.Context,
 		seelog.Errorf("Task engine [%s]: Could not retrieve docker id for container", task.Arn)
 		return
 	}
+	// Sleeping here so that all the containers do not call inspect/start exec agent process
+	// at the same time.
+	// The max sleep is 50% of the monitor interval to allow enough buffer time
+	// to finish monitoring.
+	// This is inspired from containers streaming stats from Docker.
+	time.Sleep(retry.AddJitter(time.Nanosecond, defaultMonitorExecAgentsInterval/2))
```
angelcar, line 307: let's use `engine.monitorExecAgentsInterval` instead. I think using the default time will break the test monitor unit tests

angelcar (changes requested):

agent/engine/execcmd/manager_start_linux.go
```diff
@@ -46,7 +46,7 @@ func (m *manager) RestartAgentIfStopped(ctx context.Context, client dockerapi.Do
 		return NotRestarted, nil
 	}
 	res, err := m.inspectExecAgentProcess(ctx, client, execMD)
-	if err != nil {
+	if err != nil || res == nil {
```
fierlion, line 49: curious how we end up with a nil res?  I'm looking at the inspectExecAgentProcess function: https://github.com/aws/amazon-ecs-agent/blob/feature/ecs_exec/agent/engine/execcmd/manager_start_linux.go#L67

did you encounter this case?

fierlion (approved):

fierlion (approved):

agent/engine/execcmd/manager_start_linux.go
sharanyad: ended up finding a goroutine from integ test that was panicking because response was nil :)

angelcar (approved):

PR #2678, opened 2020-10-16
cleanup ExecAgent host logs for exec tasks

sharanyad:
### Summary
The exec log dir is bind-mounted to agent container. 
ECS Agent will create a unique log path for each container in the exec enabled task on bind mounting to exec container
This path will be cleaned up once the task is stopped
`TODO`: revise the path once we know where the execAgent log dir will be mounted inside agent container. 

### Implementation details
Do a best effort to remove the exec log dir of the task and its children (each container's log destination) 

### Testing

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -17,6 +17,7 @@ package engine
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
@@ -599,6 +600,8 @@ func (engine *DockerTaskEngine) sweepTask(task *apitask.Task) {
 	}
 }
 
+var removeAll = os.RemoveAll
+
 func (engine *DockerTaskEngine) deleteTask(task *apitask.Task) {
 	for _, resource := range task.GetResources() {
 		err := resource.Cleanup()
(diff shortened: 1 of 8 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -611,6 +614,17 @@ func (engine *DockerTaskEngine) deleteTask(task *apitask.Task) {
 		}
 	}
 
+	if task.ExecCommandAgentEnabled {
```
angelcar, line 617: can we use `task.IsExecCommandAgentEnabled()` here? 
Maybe we should rename `ExecCommandAgentEnabled` to `ExecCommandAgentEnabledUnsafe`

agent/engine/task_manager_unix_test.go
```diff
@@ -380,3 +388,67 @@ func TestEFSVolumeNextStateWithTransitionDependencies(t *testing.T) {
...
+	}
+	mTask.Task.ExecCommandAgentEnabled = true
+	mTask.SetKnownStatus(apitaskstatus.TaskStopped)
+	mTask.SetSentStatus(apitaskstatus.TaskStopped)
+	container := mTask.Containers[0]
+	dockerContainer := &apicontainer.DockerContainer{
+		DockerName: "dockerContainer",
+	}
+	tID, _ := mTask.Task.GetID()
+	removeAll = func(path string) error {
```
angelcar, line 431: I worry that this function is invoked in a goroutine. If this is the case, this might or might not be executed before the test ends. 
Can we change a boolean variable here and do the assertion at the end of the test instead?

sharanyad: this is not invoked in a goroutine. This just replaces os.RemoveAll to a custom function for this unit test

agent/engine/docker_task_engine.go
sharanyad: done.

angelcar (approved):

agent/engine/execcmd/manager_unsupported.go
```diff
@@ -22,6 +22,11 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
 )
 
+const (
+	// HostLogDir in unsupported platforms for exec is empty
```
cyastella, line 26: nit: this is a bit self-explanatory, maybe this comment is not needed, or use this comment to explain what would happen later if we use empty string here

sharanyad: added more explanation.

cyastella (approved):

PR #2687, opened 2020-10-22
add check for execute-command capability and populate attribute if applicable

cyastella:
### Summary

This is part of the change for ecs-agent to detect if an EC2 instance is `exec-enabled`, in other words, if the instance has the `execute-command` capability. Related change in [amazon-ecs-init](https://github.com/aws/amazon-ecs-init): https://github.com/aws/amazon-ecs-init/pull/363

With dependencies for SSM bind mounted to the ecs-agent container, ecs-agent is responsible for checking if all required files exist. Those would include:

* SSM agent binaries
* SSM agent configuration directory (ecs-agent needs to write files into this directory)
* certs files for SSM agent

### Implementation details

Dependencies for the `execute-command` capability are bind mounted in the following structure:

```
/capabilities/execute-command/
|-- bin
    |-- amazon-ssm-agent
    |-- ssm-session-worker
|-- config
|-- certs
    |-- tls-ca-bundle.pem
```

### Testing

Manual test:

start ecs-agent container with the `--ecs-attributes` flag and check if the new capability is added

* `TestCapabilitesExecuteCommand`
* `TestDefaultPathExistsd`

check SSM dependencies bind mounted to the agent container and add capability if applicable

### Licensing

### DIFF
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -374,15 +374,12 @@ func (agent *ecsAgent) appendExecCapabilities(capabilities []*ecs.Attribute) ([]
 	if err != nil {
 		return capabilities, err
 	}
-	// use raw string for regular expression
+	// use raw string for regular expression to avoid escaping backslash (\)
 	var validSsmVersion = regexp.MustCompile(`^\d+(\.\d+)*$`)
 	for _, binFolder := range binFolders {
-		// exclude directories that don't conform to valid ssm versions
-		matched := validSsmVersion.Match([]byte(binFolder))
-		if !matched {
+		if matched := validSsmVersion.Match([]byte(binFolder)); !matched {
 			continue
 		}
-		// exclude ssm version in deny-list
 		if _, found := capabilityExecInvalidSsmVersions[binFolder]; found {
 			continue
 		}
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -14,6 +14,9 @@
 package app
 
 import (
+	"os"
+	"path/filepath"
+
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
@@ -61,29 +64,45 @@ const (
 	capabilityEFS                               = "efs"
 	capabilityEFSAuth                           = "efsAuth"
 	capabilityEnvFilesS3                        = "env-files.s3"
+	capabilityExecuteCommand                    = "execute-command"
+	capabilityDepsRootDir                       = "/capabilities"
+	capabilityExecuteCommandBinRelativePath     = "bin"
+	capabilityExecuteCommandConfigRelativePath  = "config"
+	capabilityExecuteCommandCertsRelativePath   = "certs"
 )
 
-var nameOnlyAttributes = []string{
-	// ecs agent version 1.19.0 supports private registry authentication using
-	// aws secrets manager
-	capabilityPrivateRegistryAuthASM,
-	// ecs agent version 1.22.0 supports ecs secrets integrating with aws systems manager
-	capabilitySecretEnvSSM,
-	// ecs agent version 1.27.0 supports ecs secrets for logging drivers
-	capabilitySecretLogDriverSSM,
-	// support ecr endpoint override
-	capabilityECREndpoint,
-	// ecs agent version 1.23.0 supports ecs secrets integrating with aws secrets manager
-	capabilitySecretEnvASM,
-	// ecs agent version 1.27.0 supports ecs secrets for logging drivers
-	capabilitySecretLogDriverASM,
-	// support container ordering in agent
-	capabilityContainerOrdering,
-	// support full task sync
-	capabilityFullTaskSync,
-	// ecs agent version 1.39.0 supports bulk loading env vars through environmentFiles in S3
-	capabilityEnvFilesS3,
-}
+var (
+	nameOnlyAttributes = []string{
+		// ecs agent version 1.19.0 supports private registry authentication using
+		// aws secrets manager
+		capabilityPrivateRegistryAuthASM,
+		// ecs agent version 1.22.0 supports ecs secrets integrating with aws systems manager
+		capabilitySecretEnvSSM,
+		// ecs agent version 1.27.0 supports ecs secrets for logging drivers
+		capabilitySecretLogDriverSSM,
+		// support ecr endpoint override
+		capabilityECREndpoint,
+		// ecs agent version 1.23.0 supports ecs secrets integrating with aws secrets manager
+		capabilitySecretEnvASM,
+		// ecs agent version 1.27.0 supports ecs secrets for logging drivers
+		capabilitySecretLogDriverASM,
+		// support container ordering in agent
+		capabilityContainerOrdering,
+		// support full task sync
+		capabilityFullTaskSync,
+		// ecs agent version 1.39.0 supports bulk loading env vars through environmentFiles in S3
+		capabilityEnvFilesS3,
+	}
+	capabilityExecuteCommandRequiredBinaries = []string{
+		"amazon-ssm-agent",
+		"ssm-session-worker",
+	}
+	capabilityExecuteCommandRequiredCerts = []string{
+		"tls-ca-bundle.pem",
+	}
+
+	pathExists = defaultPathExists
+)
 
 // capabilities returns the supported capabilities of this agent / docker-client pair.
 // Currently, the following capabilities are possible:
@@ -131,6 +150,7 @@ var nameOnlyAttributes = []string{
 //    ecs.capability.gmsa
 //    ecs.capability.efsAuth
 //    ecs.capability.env-files.s3
+//    ecs.capability.execute-command
 func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	var capabilities []*ecs.Attribute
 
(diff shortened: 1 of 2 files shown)


agent/app/agent_capability_test.go
```diff
@@ -751,3 +765,151 @@ func TestCapabilitesScanPluginsErrorCase(t *testing.T) {
 		}
 	}
 }
+
+func TestCapabilitesExecuteCommandShouldBeAddedWhenDependenciesExist(t *testing.T) {
```
angelcar, line 769: Can we use a test table instead of two test functions? `TestCapabilitesExecuteCommandShouldBeAddedWhenDependenciesExist` and `TestCapabilitesExecuteCommandShouldNotBeAddedWhenDependenciesDoNotExist` have pretty much the same code.

agent/app/agent_capability.go
```diff
@@ -316,6 +342,50 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
 	return capabilities
 }
 
+func (agent *ecsAgent) appendExecCapabilities(capabilities []*ecs.Attribute) ([]*ecs.Attribute, error) {
```
angelcar, line 345: I think we should probably test that config directory is writable, and maybe also that the binaries and certs are read only?

angelcar: Also, this is something we haven't discussed thoroughly, so maybe it's for a new PR, but calling it out here... I think we should check that the binaries we're taking are legit (exec signature or md5 or somethig)... these are binaries we're mounting into customer's containers, so we want to make sure we won't mount some spurious binary that could wreak havoc in the customer's container.

agent/app/agent_capability_test.go
cyastella: sure, change to test table

agent/app/agent_capability.go
cyastella: good point, agree that it'd be better for us to add that functionality (validating the binaries) as an add-on after discussion

cyastella: for checking if `/config` is writable while `/bin` and `/certs` are read-only, golang does not seem to provide a native way for this check. 1 possible way (still testing and prototyping) is to use the unix permission bits, but I'm not sure if that would work for Windows scenarios or if that (Windows scenario) is something we should care about

angelcar: Wonder if something like this works:
https://stackoverflow.com/questions/45429210/how-do-i-check-a-files-permissions-in-linux-using-go
or
https://github.com/phayes/permbits/blob/master/permbits.go

agent/app/agent_capability.go
```diff
@@ -61,29 +64,45 @@ const (
...
+		// support container ordering in agent
+		capabilityContainerOrdering,
+		// support full task sync
+		capabilityFullTaskSync,
+		// ecs agent version 1.39.0 supports bulk loading env vars through environmentFiles in S3
+		capabilityEnvFilesS3,
+	}
+	capabilityExecRequiredBinaries = []string{
+		"amazon-ssm-agent",
+		"ssm-session-worker",
```
ubhattacharjya, line 98: I think you will also need to check for ssm-agent-worker

agent/app/agent_capability.go
```diff
@@ -61,29 +64,45 @@ const (
 	capabilityEFS                               = "efs"
 	capabilityEFSAuth                           = "efsAuth"
 	capabilityEnvFilesS3                        = "env-files.s3"
+	capabilityExec                              = "execute-command"
+	capabilityDepsRootDir                       = "/managed-agents"
```
ubhattacharjya, line 68: Should it have ${VERSION} folder of ssm-agent as well?

cyastella: yup, the version sub folder was added after this PR was posted, adding in the next iteration

agent/app/agent_capability.go
cyastella: did some prototyping and here are what I found: regardless of what the bind mount is specified as `readonly` or not, the permission bits are the same (with `ls -l`). here are the steps I took:

start container with bind mounts:
```
$ docker run -it \
--mount type=bind,src="${HOME}/ro_root",dst=/exec_test/ro_root,readonly \
--mount type=bind,src="${HOME}/rw_root",dst=/exec_test/rw_root \
--mount type=bind,src="${HOME}/ro_file",dst=/exec_test/test_dir/ro_file,readonly \
--mount type=bind,src="${HOME}/rw_file",dst=/exec_test/test_dir/rw_file \
amazonlinux:latest
```

results:
```
$ ls -lR /exec_test
/exec_test:
total 0
drwxrwxr-x 3 1000 1000 35 Oct 28 22:03 ro_root
drwxrwxr-x 3 1000 1000 35 Oct 28 22:04 rw_root
drwxr-xr-x 2 root root 36 Oct 28 22:09 test_dir

/exec_test/ro_root:
total 0
drwxrwxr-x 2 1000 1000 18 Oct 28 22:03 child_dir
-rw-rw-r-- 1 1000 1000  0 Oct 28 22:03 file

/exec_test/ro_root/child_dir:
total 0
-rw-rw-r-- 1 1000 1000 0 Oct 28 22:03 file

/exec_test/rw_root:
total 0
drwxrwxr-x 2 1000 1000 18 Oct 28 22:04 child_dir
-rw-rw-r-- 1 1000 1000  0 Oct 28 22:04 file

/exec_test/rw_root/child_dir:
total 0
-rw-rw-r-- 1 1000 1000 0 Oct 28 22:04 file

/exec_test/test_dir:
total 0
-rw-rw-r-- 1 1000 1000 0 Oct 28 22:04 ro_file
-rw-rw-r-- 1 1000 1000 0 Oct 28 22:06 rw_file
```

as indicated here, both files and directories (readonly and read-write) have the same permission bits (`-rw`, `drw`), don't think we can check this only with permission bits

cyastella: the error happens at runtime, e.g.

```
$ echo "hello world" > /exec_test/test_dir/ro_file
bash: ro_file: Read-only file system
```

cyastella: I think we can only check for existence (if the binaries and cert file exist) in this change, I will look further into how to check if a bind-mounted file/directory is read-only and update the check logic with a follow-up PR if it's feasible

angelcar: ok, how about just attempting to write a temp file into this directory? which should succeed...

we can do this n a new PR as you suggest

angelcar (approved):

agent/app/agent_capability.go
cyastella: nice catch! thanks for pointing it out

agent/app/agent_capability.go
cyastella: sounds good! thanks!

agent/app/agent_capability.go
```diff
@@ -316,6 +345,92 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
...
+
+func defaultGetSubDirectories(path string) ([]string, error) {
+	var subDirectories []string
+
+	fileInfos, err := ioutil.ReadDir(path)
+	if err != nil {
+		return nil, err
+	}
+	for _, fileInfo := range fileInfos {
+		if fileInfo.IsDir() {
```
angelcar, line 398: You have to exclude directories that don't conform to valid ssm versions here. We can't assume that our dependencies dir will always be pristine, someone could create files inside the dir and we can't error out because of that.

agent/app/agent_capability.go
```diff
@@ -316,6 +345,92 @@ func (agent *ecsAgent) appendTaskENICapabilities(capabilities []*ecs.Attribute)
...
+	for _, binFolder := range binFolders {
+		// check for the same set of binaries for all versions for now
+		// change this if future versions of ssm agent require different binaries
+		versionSubDirectory := filepath.Join(binDir, binFolder)
+		binDependencies[versionSubDirectory] = capabilityExecRequiredBinaries
+	}
+	if len(binDependencies) < 1 {
+		return capabilities, nil
+	}
+	if exists, err := dependenciesExist(binDependencies); err != nil || !exists {
```
angelcar, line 383: I'd argue that we need to find at least one valid dependencies dir, as opposed to all of them. In the future when we create an update process, this process might be in progress and not have all the dependencies right at the time of our check, which would cause this to error out.

angelcar: In other words, only error out if no valid versions were found.

agent/app/agent_capability.go
cyastella: good point, I'll add a quick check here for folders matching this regex: `\d+(\.\d+)*`, but at some point I think it'd be beneficial to reuse code that we're currently adding to `agent/engine/execcmd/agent_version.go`

agent/app/agent_capability.go
cyastella: good point, change to error out only if no valid version is found

agent/app/agent_capability.go
angelcar: I think regex needs to be `^\d+(\.\d+)*$` 
https://regex101.com/r/yi3sgO/1

fierlion (approved):

agent/app/agent_capability.go
cyastella: thanks for pointing it out! add anchor tokens

agent/app/agent_capability.go
cyastella: good point, add deny-list for ssm versions; currently this will be an empty set so no actual effect before we add any version into it

angelcar (approved):

PR #2688, opened 2020-10-22
Use different Golang version while building Agent for Windows

mythri-garaga:
### Summary

Use different Golang version while building Agent for Windows
### Implementation details

- In `Makefile`, target `docker-release` is modified to use different dockerfiles based on OS

### Testing

Tested the build produced by `make codebuild` on Windows2016. Agent joined the cluster successfully.

New tests cover the changes: 

Bug - Fixed Makefile to use Go1.12 for Agent windows build

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,12 @@ else
 	GOARCH=amd64
 endif
 
+ifeq (${TARGET_OS},windows)
+    GO_VERSION=1.12
+else
+    GO_VERSION=1.15
+endif
+
 all: docker
 
 # Dynamic go build; useful in that it does not have -a so it won't recompile
@@ -47,7 +53,7 @@ xplatform-build:
 
 BUILDER_IMAGE="amazon/amazon-ecs-agent-build:make"
 .builder-image-stamp: scripts/dockerfiles/Dockerfile.build
-	@docker build -f scripts/dockerfiles/Dockerfile.build -t $(BUILDER_IMAGE) .
+	@docker build --build-arg GO_VERSION=$(GO_VERSION) -f scripts/dockerfiles/Dockerfile.build -t $(BUILDER_IMAGE) .
 	touch .builder-image-stamp
 
 # 'build-in-docker' builds the agent within a dockerfile and saves it to the ./out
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -75,7 +75,11 @@ docker: certs build-in-docker pause-container-release cni-plugins .out-stamp
 # 'RELEASE' mode
 # TODO: make this idempotent
 docker-release: pause-container-release cni-plugins .out-stamp
+ifeq ($(TARGET_OS),linux)
```
angelcar, line 78: would it be better to use `--build-arg`s? I think this is probably a good time to refactor a bit and make all docker versions homogeneous. 

More info about `build-arg`:
https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg
https://stackoverflow.com/questions/19537645/get-environment-variable-value-in-dockerfile

Makefile
```diff
@@ -75,7 +75,11 @@ docker: certs build-in-docker pause-container-release cni-plugins .out-stamp
...
 	@docker build -f scripts/dockerfiles/Dockerfile.cleanbuild -t "amazon/amazon-ecs-agent-cleanbuild:make" .
+else ifeq ($(TARGET_OS),windows)
+	@docker build -f scripts/dockerfiles/Dockerfile.cleanbuildWindows -t "amazon/amazon-ecs-agent-cleanbuild:make" .
+endif
```
shubham2892, line 82: Will this break for mac os?

Makefile
mythri-garaga: I agree, will update in next revision

Makefile
mythri-garaga: Good catch! I was only thinking about `codebuild` target use case. I will update it

Makefile
```diff
@@ -71,11 +75,19 @@ docker: certs build-in-docker pause-container-release cni-plugins .out-stamp
 	@docker build -f scripts/dockerfiles/Dockerfile.release -t "amazon/amazon-ecs-agent:make" .
 	@echo "Built Docker image \"amazon/amazon-ecs-agent:make\""
 
+ifeq (${TARGET_OS},windows)
+    BUILD="cleanbuild-${TARGET_OS}"
+    GO_VERSION=${GO_VERSION_WINDOWS}
```
angelcar, line 80: when do you use `GO_VERSION` vs `O_VERSION_LINUX` vs `GO_VERSION_WINDOWS` throughout the make files? 

I guess we can just use `GO_VERSION` everywhere?

scripts/dockerfiles/Dockerfile.build
```diff
@@ -11,7 +11,8 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 
-FROM golang:1.12
+ARG GO_VERSION=1.15
```
angelcar, line 14: I would like to know what others think, but I'd rather not use a default value so that if our main makefile is wrong, the rest of the make files would fail, instead of silently using go 1.15.

Makefile
```diff
@@ -71,11 +77,17 @@ docker: certs build-in-docker pause-container-release cni-plugins .out-stamp
 	@docker build -f scripts/dockerfiles/Dockerfile.release -t "amazon/amazon-ecs-agent:make" .
 	@echo "Built Docker image \"amazon/amazon-ecs-agent:make\""
 
+ifeq (${TARGET_OS},windows)
+    BUILD="cleanbuild-${TARGET_OS}"
```
angelcar, line 81: Is this variable still needed?

mythri-garaga: This will be useful when we are playing with the scripts locally.

shubham2892 (approved):

angelcar (approved):

PR #2690, opened 2020-10-23
fsxwindowsfileserver: fixing task concurrency issues

singholt:
### Summary

This change fixes task concurrency issues with drive letter assignment during ```fsxwindowsfileserver``` task resource initialization.

### Implementation details

1. Moved drive letter assignment to ```performHostMount``` method. 
2. Added a mutex lock for ```performHostMount()``` method.

### Testing

Launched 10 tasks concurrently and each of them gets its own drive letter now.

### Licensing

### DIFF
--- a/agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
+++ b/agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
@@ -107,7 +107,6 @@ func NewFSxWindowsFileServerResource(
 	name string,
 	volumeType string,
 	volumeConfig *FSxWindowsFileServerVolumeConfig,
-	hostPath string,
 	executionCredentialsID string,
 	credentialsManager credentials.Manager,
 	ssmClientCreator ssmfactory.SSMClientCreator,
@@ -121,7 +120,6 @@ func NewFSxWindowsFileServerResource(
 			FileSystemID:  volumeConfig.FileSystemID,
 			RootDirectory: volumeConfig.RootDirectory,
 			AuthConfig:    volumeConfig.AuthConfig,
-			HostPath:      hostPath,
 		},
 		taskARN:                taskARN,
 		region:                 region,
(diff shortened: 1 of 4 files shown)


agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -519,9 +515,34 @@ func (fv *FSxWindowsFileServerResource) retrieveFileSystemDNSName(fileSystemId s
 	return nil
 }
 
+var mountLock sync.Mutex
```
vsiddharth, line 518: Why do you need an additional lock while performing `hostMount`? Please add a comment to explain the rationale. Also, why is this outside the resource encapsulation?

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
```diff
@@ -519,9 +515,34 @@ func (fv *FSxWindowsFileServerResource) retrieveFileSystemDNSName(fileSystemId s
...
+var DriveLetterAvailable = utils.IsAvailableDriveLetter
 var execCommand = exec.Command
 
-func (fv *FSxWindowsFileServerResource) performHostMount(remotePath string, localPathArg string, username string, password string) error {
+func (fv *FSxWindowsFileServerResource) performHostMount(remotePath string, username string, password string) error {
+	mountLock.Lock()
```
angelcar, line 523: Can we extract this code into its own function? That way we don't need a new `Mutex`, plus the critical section locked by the mutex is reduced to the absolute minimum (the new mutex in this PR is now serializing all the powershell executions below as well, potentially making task creation slower)

e.g.

```golang
func (fv *FSxWindowsFileServerResource) defineHostPath() error {
	fv.lock.Lock()
	defer fv.lock.Lock()

	hostPath, err := utils.FindUnusedDriveLetter()
	if err != nil {
		return err
	}

	fv.VolumeConfig.HostPath = hostPath

	// formatting to keep powershell happy
	localPathArg := fmt.Sprintf("-LocalPath \"%s\"", strings.Trim(hostPath, `\`))

	// remove mount if localPath is not available
	// handling case where agent crashes after mount and before resource status is updated
	if !(DriveLetterAvailable(hostPath)) {
		err = fv.removeHostMount(localPathArg)
		if err != nil {
			seelog.Errorf("Failed to remove existing fsxwindowsfileserver resource mount on the container instance: %v", err)
			fv.setTerminalReason(err.Error())
			return err
		}
	}
}

func (fv *FSxWindowsFileServerResource) performHostMount(remotePath string, username string, password string) error {

	fv.defineHostPath()	
	.
	.
	.
```

fierlion (approved):

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
fenxiong, line 523: can you add a comment here documenting why a package level lock is needed here? so that we don't lose context on why this was added

sharanyad: does this delay the resource creation for fsx tasks?

singholt: I ran 10 tasks together and noticed about half a second delay between tasks transitioning into running state.

singholt: I'll add a comment to add some context

angelcar: Nvm, after talking offline, it seems that the letter becomes unavailable only after executing the mounts, so the lock seems necessary after all.

agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
singholt: I'll add a comment for this

fenxiong (approved):

angelcar (approved):

vsiddharth (approved):

PR #2691, opened 2020-10-23
validate exec agent container start state change event

fierlion:
### Summary
Add validation of container state change start event for exec agent start.  

### Implementation details
Followed the pattern for other event validations in our integ tests.  `verifyContainerRunningStateChange` and `verifyContainerRunningStateChangeWithRuntimeID`.  I call the validation directly following `verifyMockExecCommandAgentIsRunning` to be sure we should expect such an event.

### Testing
I ran this test 30x to check for flakiness using:
`for run in {1..30}; do sudo -E env "PATH=$PATH"  go test -tags sudo -count=1 -v -run="TestExecCommandAgent" ./agent/engine/...; done`

Add exec command start container event validation.

### Licensing

### DIFF
--- a/agent/engine/common_integ_test.go
+++ b/agent/engine/common_integ_test.go
@@ -117,6 +117,17 @@ func verifyContainerRunningStateChangeWithRuntimeID(t *testing.T, taskEngine Tas
 		"Expected container runtimeID should not empty")
 }
 
+func verifyExecAgentRunningStateChange(t *testing.T, taskEngine TaskEngine) {
+	stateChangeEvents := taskEngine.StateChangeEvents()
+	event := <-stateChangeEvents
+	containerEvent := event.(api.ContainerStateChange)
+	assert.NotEmpty(t, containerEvent.ManagedAgents, "Expected exec-agent event has no managed agents")
+	if containerEvent.ManagedAgents != nil {
+		assert.Equal(t, apicontainerstatus.ManagedAgentRunning, containerEvent.ManagedAgents[0].Status,
+			"expected managedAgent container state change event did not match actual event")
+	}
+}
+
 func verifyContainerStoppedStateChange(t *testing.T, taskEngine TaskEngine) {
 	stateChangeEvents := taskEngine.StateChangeEvents()
 	event := <-stateChangeEvents
--- a/agent/engine/engine_sudo_linux_integ_test.go
+++ b/agent/engine/engine_sudo_linux_integ_test.go
@@ -389,6 +389,7 @@ func TestExecCommandAgent(t *testing.T) {
 
 	verifyExecCmdAgentExpectedMounts(t, ctx, client, testTaskId, cid, testContainerName, testExecCmdHostBinDir)
 	pidA := verifyMockExecCommandAgentIsRunning(t, client, cid)
+	verifyExecAgentRunningStateChange(t, taskEngine)
 	seelog.Infof("Verified mock ExecCommandAgent is running (pidA=%s)", pidA)
 	killMockExecCommandAgent(t, client, cid, pidA)
 	seelog.Infof("kill signal sent to ExecCommandAgent (pidA=%s)", pidA)


agent/engine/common_integ_test.go
```diff
@@ -117,6 +117,17 @@ func verifyContainerRunningStateChangeWithRuntimeID(t *testing.T, taskEngine Tas
 		"Expected container runtimeID should not empty")
 }
 
+func verifyExecAgentRunningStateChange(t *testing.T, taskEngine TaskEngine) {
+	stateChangeEvents := taskEngine.StateChangeEvents()
+	event := <-stateChangeEvents
```
angelcar, line 122: I'd prefer to timeout on this, but I know it's not a pattern that is being followed

angelcar (approved):

agent/engine/common_integ_test.go
```diff
@@ -117,6 +117,17 @@ func verifyContainerRunningStateChangeWithRuntimeID(t *testing.T, taskEngine Tas
...
+	containerEvent := event.(api.ContainerStateChange)
+	assert.NotEmpty(t, containerEvent.ManagedAgents, "Expeceted exec-agent event has no managed agents")
```
cyastella, line 124: typo: Expected

cyastella: I think the error msg could be either "exec-agent event has no managed agents" or "Expected exec-agent event should have managed agents"

agent/engine/common_integ_test.go
```diff
@@ -117,6 +117,17 @@ func verifyContainerRunningStateChangeWithRuntimeID(t *testing.T, taskEngine Tas
...
+	if containerEvent.ManagedAgents != nil {
+		assert.Equal(t, apicontainerstatus.ManagedAgentRunning, containerEvent.ManagedAgents[0].Status,
```
cyastella, line 126: just curious, do we need to check for other indexes like `1,2,...`? or should we assert that `containerEvent.ManagedAgents` only has 1 element?

agent/engine/common_integ_test.go
```diff
@@ -117,6 +117,17 @@ func verifyContainerRunningStateChangeWithRuntimeID(t *testing.T, taskEngine Tas
...
+			"expected managedAgent container state change event did not match actual event")
```
cyastella, line 127: nit: this error msg is pretty much what `assert.Equal()` means, maybe we don't need it here =)

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -389,6 +389,7 @@ func TestExecCommandAgent(t *testing.T) {
 
 	verifyExecCmdAgentExpectedMounts(t, ctx, client, testTaskId, cid, testContainerName, testExecCmdHostBinDir)
 	pidA := verifyMockExecCommandAgentIsRunning(t, client, cid)
+	verifyExecAgentRunningStateChange(t, taskEngine)
```
sharanyad, line 392: are we adding stop state change in a different PR?

sharanyad, line 392: we should add this validation for restart as well

agent/engine/common_integ_test.go
fierlion: We'll only have one message here.

agent/engine/engine_sudo_linux_integ_test.go
fierlion: I'll add this in a separate PR.

fierlion: I'll add this in a separate PR.

agent/engine/common_integ_test.go
fierlion: We need it for debugging a failed test.  It may also be useful if we extend this function to check for other types of state change events.

agent/engine/common_integ_test.go
fierlion: fixed

sharanyad (approved):

PR #2692, opened 2020-10-27
Refactor to start execute command agent at container level

angelcar:
### Summary

This PR is a refactor of ExecuteCommandAgent in order to do operations at the container level since ACS streams exec command data at the container level.

This refactor doesn't make any major changes in logic, but it merely reorganizes the code and adapts UTs/Integ tests accordingly.

### Implementation details

* Remove `ExecCommandAgentMetadata` from container; now the metadata is part of the ExecuteCommandAgent `ManagedAgent`.
* `ManagedAgent` now reflects the ACS model
* Replace `InitializeTask` from `InitializeContainer`. Bindmounts happen now in `createContainer` and no shared volumes are created for container deps.
* Using the latest agreed-upon paths for the dependency bindmounts

### Testing

* Made sure UTs still pass and reflect expectations
* Made sure integ tests pass and reflect expectations

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -1216,8 +1216,8 @@ func (c *Container) GetManagedAgentByName(agentName string) (ManagedAgent, bool)
 }
 
 // UpdateManagedAgentByName updates the state of the managed agent with the name specified. If the agent is not found,
-// no action is taken.
-func (c *Container) UpdateManagedAgentByName(agentName string, state ManagedAgentState) {
+// this method returns false.
+func (c *Container) UpdateManagedAgentByName(agentName string, state ManagedAgentState) bool {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 	for i, ma := range c.ManagedAgentsUnsafe {
@@ -1228,7 +1228,8 @@ func (c *Container) UpdateManagedAgentByName(agentName string, state ManagedAgen
 				Properties:        ma.Properties,
 				ManagedAgentState: state,
 			}
-			break
+			return true
 		}
 	}
+	return false
 }
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -80,9 +80,6 @@ const (
 
 	// neuronVisibleDevicesEnvVar is the env which indicates that the container wants to use inferentia devices.
 	neuronVisibleDevicesEnvVar = "AWS_NEURON_VISIBLE_DEVICES"
-
-	// executeCommandAgent is the only available ManagedAgent type
-	executeCommandAgentName = "EXECUTE_COMMAND_AGENT"
 )
 
 // DockerConfig represents additional metadata about a container to run. It's
@@ -109,15 +106,25 @@ type HealthStatus struct {
 	Output string `json:"output,omitempty"`
 }
 
-type ManagedAgent struct {
+type ManagedAgentState struct {
+	// ID of this managed agent state
+	ID string `json:"id:omitempty"`
 	// Status is the managed agent health status
 	Status apicontainerstatus.ManagedAgentStatus `json:"status:omitempty"`
-	// Name is always executeCommandAgent (until a new agent type is introduced)
-	Name string `json:"name,omitempty"`
 	// Reason is a placeholder for failure messaging
 	Reason string `json:"reason,omitempty"`
 	// LastStartedAt is the timestamp when the status last went from PENDING->RUNNING
 	LastStartedAt time.Time `json:"lastStartedAt,omitempty"`
+	// Metadata holds metadata about the managed agent
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
+}
+
+type ManagedAgent struct {
+	ManagedAgentState
+	// Name is the name of this managed agent. This name is streamed down from ACS.
+	Name string `json:"name,omitempty"`
+	// Properties of this managed agent. Properties are streamed down from ACS.
+	Properties map[string]string `json:"properties,omitempty"`
 }
 
 // Container is the internal representation of a container in the ECS agent
@@ -290,9 +297,6 @@ type Container struct {
 	finishedAt time.Time
 
 	labels map[string]string
-
-	// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
-	ExecCommandAgentMetadata ExecCommandAgentMetadata `json:"execCommandAgentMetadata"`
 }
 
 type DependsOn struct {
@@ -345,22 +349,6 @@ type Secret struct {
 	Type          string `json:"type"`
 	Provider      string `json:"provider"`
 	Target        string `json:"target"`
-}
-
-// ExecCommandAgentMetadata holds metadata about the exec agent running inside the container (i.e. SSM Agent).
-type ExecCommandAgentMetadata struct {
-	PID           string                                `json:"pid"`
-	DockerExecID  string                                `json:"dockerExecId"`
-	CMD           string                                `json:"cmd"`
-	StartedAt     time.Time                             `json:"startedAt"`
-	StoppedReason string                                `json:"stoppedReason"`
-	Status        apicontainerstatus.ManagedAgentStatus `json:"execCommandAgentStatus"`
-	SessionLimit  int                                   `json:"sessionLimit"`
-}
-
-func (md *ExecCommandAgentMetadata) String() string {
-	return fmt.Sprintf("[PID: %s, CMD: %s, StartedAt: %s, StoppedReason: %s]",
-		md.PID, md.CMD, md.StartedAt, md.StoppedReason)
 }
 
 // GetSecretResourceCacheKey returns the key required to access the secret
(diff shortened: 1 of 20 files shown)


agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -77,111 +77,96 @@ var (
 		"SessionWorkersLimit": %d
 	  }
 	}`
-	execAgentConfigFileNameTemplate = `amazon-ssm-agent-%s.json`
+	execAgentConfigFileNameTemplate    = `amazon-ssm-agent-%s.json`
```
fierlion, line 80: we'll still need to add the seelog.xml template here.  (I think that's a separate task if so, we proabaly need a TODO for that so it doesn't get lost.)

agent/engine/execcmd/manager.go
```diff
@@ -96,6 +99,20 @@ func NewManagerWithBinDir(hostBinDir string) *manager {
...
+}
+
+func IsExecEnabledTask(task *apitask.Task) bool {
+	for _, c := range task.Containers {
+		if IsExecEnabledContainer(c) {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
angelcar: We don't have a use case for that at the moment. This method is used only on task cleanup to determine if we need to clean ssm logs from the host for a task

agent/engine/execcmd/manager_init_task_linux.go
angelcar: Will, add TODO.

agent/engine/docker_task_engine_test.go
```diff
@@ -243,7 +243,9 @@ func TestBatchContainerHappyPath(t *testing.T) {
 			credentialsManager.EXPECT().RemoveCredentials(credentialsID)
 
 			sleepTask := testdata.LoadTask("sleep5")
-			sleepTask.ExecCommandAgentEnabledUnsafe = tc.execCommandAgentEnabled
+			if tc.execCommandAgentEnabled && len(sleepTask.Containers) > 0 {
```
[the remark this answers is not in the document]
angelcar: Oh, that was just to be defensive since "sleep5" is something that was already there and could be modified by someone in the future I guess

fierlion (approved):

agent/api/container/container.go
```diff
@@ -1231,14 +1201,34 @@ func (c *Container) GetTaskARN() string {
...
+}
+
+// UpdateManagedAgentByName updates the state of the managed agent with the name specified. If the agent is not found,
+// no action is taken.
+func (c *Container) UpdateManagedAgentByName(agentName string, state ManagedAgentState) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	for i, ma := range c.ManagedAgentsUnsafe {
+		if ma.Name == agentName {
+			// It's necessary to clone the whole ManagedAgent struct
```
sharanyad, line 1225: why so?

agent/api/container/container.go
```diff
@@ -1231,14 +1201,34 @@ func (c *Container) GetTaskARN() string {
...
+	for _, ma := range c.ManagedAgentsUnsafe {
+		if ma.Name == agentName {
+			return ma, true
+		}
+	}
+	return ManagedAgent{}, false
+}
+
+// UpdateManagedAgentByName updates the state of the managed agent with the name specified. If the agent is not found,
+// no action is taken.
```
sharanyad, line 1219: should we atleast log if agent is not found? 
when do we anticipate such an issue?

agent/engine/execcmd/manager.go
```diff
@@ -25,9 +25,12 @@ import (
 )
 
 const (
-	// TODO: [ecs-exec] decide if this needs to be configurable or put in a specific place in our optimized AMIs
-	HostBinDir = "/home/ec2-user/ssm-agent/linux_amd64"
+	hostExecDepsDir = "/var/lib/ecs/deps/execute-command"
+	// TODO: [ecs-exec]: This version should be dynamic, need to confirm with ssm team what's the exact format so we can always pick the latest
```
sharanyad, line 29: +1

agent/statemanager/state_manager.go
```diff
@@ -115,7 +115,7 @@ const (
 	//	 a) Add 'authorizationConfig', 'transitEncryption' and 'transitEncryptionPort' to 'taskresource.volume.EFSVolumeConfig'
 	//	 b) Add 'pauseContainerPID' field to 'taskresource.volume.VolumeResource'
 	// 28) Add 'envfile' field to 'resources'
-	// 29) Add 'ExecCommandAgentMetadata' field to 'apicontainer.Container'
+	// 29) Add 'ManagedAgentsUnsafe' field to 'apicontainer.Container'
```
sharanyad, line 118: looks like this is not needed anymore with boltdb migration. lets confirm with @fenxiong

agent/api/container/container.go
```diff
@@ -109,15 +106,25 @@ type HealthStatus struct {
...
 	// Status is the managed agent health status
 	Status apicontainerstatus.ManagedAgentStatus `json:"status:omitempty"`
-	// Name is always executeCommandAgent (until a new agent type is introduced)
-	Name string `json:"name,omitempty"`
 	// Reason is a placeholder for failure messaging
 	Reason string `json:"reason,omitempty"`
 	// LastStartedAt is the timestamp when the status last went from PENDING->RUNNING
 	LastStartedAt time.Time `json:"lastStartedAt,omitempty"`
+	// Metadata holds metadata about the managed agent
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
```
sharanyad, line 119: feels like it would be a lot easier to read if metadata was a struct and to avoid the extra toMap conversion

agent/api/container/container.go
```diff
@@ -109,15 +106,25 @@ type HealthStatus struct {
 	Output string `json:"output,omitempty"`
 }
 
-type ManagedAgent struct {
+type ManagedAgentState struct {
+	// ID of this managed agent state
+	ID string `json:"id:omitempty"`
```
sharanyad, line 111: what happens if we don't json save this `ID`?

agent/api/container/container.go
```diff
@@ -109,15 +106,25 @@ type HealthStatus struct {
...
+	// Metadata holds metadata about the managed agent
+	Metadata map[string]interface{} `json:"metadata,omitempty"`
+}
+
+type ManagedAgent struct {
+	ManagedAgentState
+	// Name is the name of this managed agent. This name is streamed down from ACS.
+	Name string `json:"name,omitempty"`
+	// Properties of this managed agent. Properties are streamed down from ACS.
+	Properties map[string]string `json:"properties,omitempty"`
```
sharanyad, line 127: this can be a struct too. we don't have to keep it the same as ACS

agent/api/container/container.go
angelcar: I'm using this `ID` to store the `UUID` that we append to the `/ecs-exec-command-UUID`. We need to persist this since it is generated in `execcmd.InitializeContainer` and used in `execcmd.StartAgent`, we need to remember it even after restarts

agent/api/container/container.go
angelcar: The problem is that in the future, different managed agents will have different metadata requirements, defining a struct in the container would limit to the metadata of exec cmd only, thus making it generic through a map.

agent/api/container/container.go
angelcar: I prefer to keep it the same, otherwise we would need to build custom logic in `TaskFromACS` and remember to update it/add to it when the ACS model is modified.

agent/api/container/container.go
angelcar: Sure, we can log, will update.
For example, if `UpdateManagedAgentByName` is called on an internal container, it surely won't have any managed agent.

agent/api/container/container.go
angelcar: ManagedAgentState is an embedded struct. If I do `c.ManagedAgentsUnsafe[i].ManagedState = state` that only updates the temporary `ManagedState` created when I reference it, but doesn't update the real data. 

Another option could be to make this pointers, but I consider that worse because semantically we are saying the data is mutable. 

Yet another option would be to do it as below, but that is more tedious and error prone.

```golang
c.ManagedAgentsUnsafe[i].ID = ...
c.ManagedAgentsUnsafe[i].Status = ...
c.ManagedAgentsUnsafe[i].Reason = ...
c.ManagedAgentsUnsafe[i].Metadata = ...
```

agent/statemanager/state_manager.go
angelcar: ok, can we remove it later if we confirm is not needed?

agent/api/container/container.go
angelcar: Actually I ended up returning a boolean depending on whether the agent is found, that way is up to the caller to log or do the necessary

agent/api/container/container.go
sharanyad: ah ok, that makes sense

agent/api/container/container.go
sharanyad: discussed offline, feels like a better approach than using locks

agent/api/container/container.go
sharanyad: also makes sense to keep it generic, since its the managed agent;s struct

sharanyad (commented):
code lgtm

sharanyad (approved):
:shipit:

PR #2699, opened 2020-10-29
Add util functions to determine latest ExecuteCommandAgent version

angelcar:
### Summary

Add util functions to determine latest ExecuteCommandAgent version
### Implementation details

Add util functions to determine latest ExecuteCommandAgent version
### Testing

### Licensing

### DIFF
--- a/agent/engine/execcmd/agent_version.go
+++ b/agent/engine/execcmd/agent_version.go
@@ -1,0 +1,82 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package execcmd
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+)
+
+// TODO: Remove this later, just doing it do make static checks happy about these utils not being invoked anywhere
+func init() {
+	determineLatestVersion([]string{"1.1.1.1"})
+}
+
+const (
+	versionDelimiter  = "."
+	badInputReturn    = -999
+	badFormatErrorStr = "invalid version format"
+)
+
+// compareAgentVersion compares two versions, returning -1 if v1 < v2, 0 if v1 == v2, and 1 if v1 > v2.
+// The version must have only integers representing major, minor, patch, and reserved; delimited by dots (e.g. "1.2.3.4").
+// An error is returned if any of the versions passed as parameter is malformed.
+// example:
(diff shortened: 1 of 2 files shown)


cyastella (approved):

agent/engine/execcmd/agent_version.go
```diff
@@ -0,0 +1,64 @@
...
+	versionDelimiter  = "."
+	badInputReturn    = -999
+	badFormatErrorStr = "invalid version format"
+)
+
+// compareAgentVersion compares two versions, returning -1 if v1 < v2, 0 if v1 == v2, and 1 if v1 > v2.
+// The version must have only integers representing major, minor, patch, and reserved; delimited by dots (e.g. "1.2.3.4").
+// An error is returned if any of the versions passed as parameter is malformed.
+// example:
+// SemverCompare("3.0.236.0", "3.0.237.0") returns -1, nil
```
cyastella, line 19: nit: comment is still referring to `SemverCompare`

angelcar: ah, thanks good catch

agent/engine/execcmd/agent_version.go
```diff
@@ -0,0 +1,64 @@
...
+// SemverCompare("a.b.c.d", "3.0.236.0") returns -999, error
+func compareAgentVersion(v1Str, v2Str string) (int, error) {
+	v1Arr := strings.Split(v1Str, versionDelimiter)
+	v2Arr := strings.Split(v2Str, versionDelimiter)
+
+	if len(v1Arr) != len(v2Arr) {
+		return badInputReturn, errors.New(badFormatErrorStr)
+	}
+	for i := 0; i < len(v1Arr); i++ {
+		v1Int, err := strconv.Atoi(v1Arr[i])
```
cyastella, line 31: I'm wondering if this should be `https://golang.org/pkg/strconv/#ParseUint` instead

agent/engine/execcmd/agent_version.go
```diff
@@ -0,0 +1,82 @@
...
+	return 0, nil
+}
+
+func determineLatestVersion(versions []string) (string, error) {
+	if len(versions) == 0 {
+		return "", errors.New("no versions to compare were provided")
+	}
+	latest := "0.0.0.0"
+	for _, v := range versions {
+		comp, err := compareAgentVersion(v, latest)
```
cyastella, line 73: nit: 1 very minor downside of this is that we'll be parsing some version string multiple times; to work around that, we might need a struct for this version type. No strong opinion on this since that could be over-engineering

agent/engine/execcmd/agent_version.go
angelcar: great suggestion! changed and added checks for negative versions :)

agent/engine/execcmd/agent_version.go
angelcar: You're right, although I prefer to keep things simple since these we won't be using these version other than as strings, and we don't expect load/stress when using this functions, so this is a reasonable approach IMO

cyastella: agree

cyastella (approved):

sharanyad (approved):

PR #2701, opened 2020-10-30
Use latest version available in deps bin dir

angelcar:
### Summary

`execcmd.InitializeContainer` should grab the latest version available in `/var/lib/ecs/deps/execute-command/bin`

for example, given the following directory structure

```
/var/lib/ecs/deps/execute-command/bin
                                    +---- 1.0.0.0
                                    +---- 2.0.0.0
                                    +---- random-dir
                                    +---- 3.0.236.0                                    
```
it should be inferred that the version to use is `/var/lib/ecs/deps/execute-command/bin/3.0.236.0`

So as long as there's a directory with a valid version and valid dependencies, we grab it; this avoids hard-coding the initial ssm version and future-proofs the ecs agent for ssm version updates.

### Implementation details

`execcmd.InitializeContainer` will list the directories inside the ecs agent container's `/managed-agents/execute-command/bin` folder. Then create a collection of ssm agent versions by attempting to parse each dir name. Dirs that cannot be parsed into valid versions will be omitted as we can't error out the whole process based on an unexpected file in our bin dir. 
Next, we will sort our list of versions in reverse order and will attempt to find the necessary binaries inside each of the version folders, the first folder that contains the dependencies is our latest valid version that will be mounted into customer's containers.

If no valid versions are found inside `/var/lib/ecs/deps/execute-command/bin/`, we return an error.

### Testing

Adapted and ran integ tests and UTs

### Licensing

### DIFF
--- a/agent/engine/execcmd/agent_version.go
+++ b/agent/engine/execcmd/agent_version.go
@@ -15,68 +15,108 @@ package execcmd
 
 import (
 	"errors"
+	"io/ioutil"
 	"strconv"
 	"strings"
 )
 
-// TODO: Remove this later, just doing it do make static checks happy about these utils not being invoked anywhere
-func init() {
-	determineLatestVersion([]string{"1.1.1.1"})
-}
-
 const (
 	versionDelimiter  = "."
-	badInputReturn    = -999
 	badFormatErrorStr = "invalid version format"
 )
 
-// compareAgentVersion compares two versions, returning -1 if v1 < v2, 0 if v1 == v2, and 1 if v1 > v2.
-// The version must have only integers representing major, minor, patch, and reserved; delimited by dots (e.g. "1.2.3.4").
-// An error is returned if any of the versions passed as parameter is malformed.
+type agentVersion []uint
+
+// compare compares this agentVersion against another one passed as parameter.
+// returns -1 if lhs < rhs, 0 lhs == rhs, and 1 if lhs > rhs.
 // example:
-// compareAgentVersion("3.0.236.0", "3.0.237.0") returns -1, nil
-// compareAgentVersion("3.0.236.0", "3.0.236.0") returns 0, nil
-// compareAgentVersion("3.0.237.0", "3.0.236.0") returns 1, nil
-// compareAgentVersion("a.b.c.d", "3.0.236.0") returns -999, error
-func compareAgentVersion(v1Str, v2Str string) (int, error) {
-	v1Arr := strings.Split(v1Str, versionDelimiter)
-	v2Arr := strings.Split(v2Str, versionDelimiter)
-
-	if len(v1Arr) != len(v2Arr) {
-		return badInputReturn, errors.New(badFormatErrorStr)
+// [3, 0, 236, 0].compare([3, 0, 237, 0]) returns -1
+// [3, 0, 236, 0].compare([3, 0, 236, 0]) returns 0
+// [3, 0, 237, 0].compare([3, 0, 236, 0]) returns 1
+// [3].compare([2, 0, 236])               returns 1
+func (lhs agentVersion) compare(rhs agentVersion) int {
+	lhsLen := len(lhs)
+	rhsLen := len(rhs)
+	maxLen := lhsLen
+	if rhsLen > maxLen {
+		maxLen = rhsLen
 	}
-	for i := 0; i < len(v1Arr); i++ {
-		v1Int, err := strconv.ParseUint(v1Arr[i], 10, 32)
-		if err != nil {
-			return badInputReturn, errors.New(badFormatErrorStr)
+	for i := 0; i < maxLen; i++ {
+		v1 := uint(0)
+		v2 := uint(0)
+		if i < lhsLen {
+			v1 = lhs[i]
 		}
-		v2Int, err := strconv.ParseUint(v2Arr[i], 10, 32)
-		if err != nil {
-			return badInputReturn, errors.New(badFormatErrorStr)
+		if i < rhsLen {
+			v2 = rhs[i]
 		}
-
-		if v1Int < v2Int {
-			return -1, nil
-		} else if v1Int > v2Int {
-			return 1, nil
+		if v1 < v2 {
+			return -1
+		} else if v1 > v2 {
+			return 1
 		}
 	}
-	return 0, nil
+	return 0
 }
 
-func determineLatestVersion(versions []string) (string, error) {
-	if len(versions) == 0 {
-		return "", errors.New("no versions to compare were provided")
+// String the string representation of an agentVersion
+// e.g. [3, 0, 236, 0] -> "3.0.236.0"
+func (lhs agentVersion) String() string {
+	var avStr []string
+	for _, e := range lhs {
+		avStr = append(avStr, strconv.Itoa(int(e)))
 	}
-	latest := "0.0.0.0"
-	for _, v := range versions {
-		comp, err := compareAgentVersion(v, latest)
+	return strings.Join(avStr, versionDelimiter)
+}
+
+func parseAgentVersion(version string) (agentVersion, error) {
+	var parsedVersion []uint
+	vArr := strings.Split(version, versionDelimiter)
+	for _, part := range vArr {
+		pInt, err := strconv.ParseUint(part, 10, 32)
 		if err != nil {
-			return "", err
+			return nil, errors.New(badFormatErrorStr)
 		}
-		if comp > 0 {
-			latest = v
-		}
+		parsedVersion = append(parsedVersion, uint(pInt))
 	}
-	return latest, nil
+	return parsedVersion, nil
+}
+
+// This is a helper type to make []agentVersion sortable
+type byAgentVersion []agentVersion
+
+// Len fulfills sort.Interface
+func (bav byAgentVersion) Len() int {
+	return len(bav)
+}
+
+// Less fulfills sort.Interface
+func (bav byAgentVersion) Less(i, j int) bool {
+	return bav[i].compare(bav[j]) < 0
+}
+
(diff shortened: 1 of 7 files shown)


agent/engine/execcmd/agent_version.go
```diff
@@ -15,68 +15,107 @@ package execcmd
...
 		if err != nil {
-			return "", err
+			return nil, errors.New(badFormatErrorStr)
+		}
+		parsedVersion = append(parsedVersion, uint(pInt))
+	}
+	return parsedVersion, nil
+}
+
+type agentVersionList []agentVersion
```
cyastella, line 85: nit: why not just use `[]agentVersion` instead? defining this as a new type does not bring too much benefit and makes it a bit harder to read (cannot directly tell what data type it actually is)

agent/engine/execcmd/agent_version.go
```diff
@@ -15,68 +15,107 @@ package execcmd
...
-// An error is returned if any of the versions passed as parameter is malformed.
-// example:
-// compareAgentVersion("3.0.236.0", "3.0.237.0") returns -1, nil
-// compareAgentVersion("3.0.236.0", "3.0.236.0") returns 0, nil
-// compareAgentVersion("3.0.237.0", "3.0.236.0") returns 1, nil
-// compareAgentVersion("a.b.c.d", "3.0.236.0") returns -999, error
-func compareAgentVersion(v1Str, v2Str string) (int, error) {
-	v1Arr := strings.Split(v1Str, versionDelimiter)
-	v2Arr := strings.Split(v2Str, versionDelimiter)
+type agentVersion []uint
```
cyastella, line 28: I think this could be `[4]uint` instead of slices

cyastella: or use struct for this

```go
type AssemblyVersion struct {
    Major uint
    Minor uint
    Build uint
    Revision uint
}
```

, but since this type is very lightly used, don't this is necessary for now

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -198,11 +236,11 @@ func getExecAgentConfigHash(config string) string {
 	return base64.URLEncoding.EncodeToString(hash.Sum(nil))
 }
 
-var execAgentConfigFileExists = configFileExists
+var osStat = os.Stat
 
-func configFileExists(configFilePath string) bool {
-	if _, err := os.Stat(configFilePath); err == nil {
-		return true
+func fileExists(configFilePath string) bool {
```
cyastella, line 241: nit: I think it's easier to mock `fileExists` over `os.Stat`

agent/engine/execcmd/agent_version.go
```diff
@@ -15,68 +15,107 @@ package execcmd
...
-	if len(v1Arr) != len(v2Arr) {
-		return badInputReturn, errors.New(badFormatErrorStr)
+// compare compares this agentVersion against another one passed as parameter.
+// returns -1 if this < av2, 0 this == av2, and 1 if this > av2.
+// example:
+// [3, 0, 236, 0].compare([3, 0, 237, 0]) returns -1
+// [3, 0, 236, 0].compare([3, 0, 236, 0]) returns 0
+// [3, 0, 237, 0].compare([3, 0, 236, 0]) returns 1
+// [3].compare([2, 0, 236])               returns 1
+func (av agentVersion) compare(av2 agentVersion) int {
```
cyastella, line 37: nit: `lhs` and `rhs` for compare functions

agent/engine/execcmd/agent_version.go
angelcar: The reason I created a new type is that I wanted to make `[]agentVersion` sortable. I will check if I can still do that just by using `[]agentVersion` directly... If not, I think it's still a good recommendation to change this to make it clear it's just for sorting purposes

agent/engine/execcmd/agent_version.go
angelcar: why limit this and make it too rigid? since in the future the version could be less numbers or more (we don't really control that). Right now this works with whatever number of elements you have in the version

agent/engine/execcmd/manager_init_task_linux.go
angelcar: It is easier, buy by mocking the whole `fileExists` function, we also remove the bits of logic from test coverage... I think this is a good balance between complexity and coverage.... thoughts?

agent/engine/execcmd/agent_version.go
angelcar: I like that! will change

agent/engine/execcmd/agent_version.go
angelcar: Alright, I renamed `agentVersionList` to `byAgentVersion` and I only use this type for sorting, the rest of the functions use `[]agentVersion` directly. And you're right that this is more readable :)

agent/engine/execcmd/manager_init_task_linux.go
cyastella: I think we can also unit test the `fileExists` function if test coverage is a concern, but no strong opinion here, mocking `os.Stat` sounds fine too

cyastella: btw, the input argument is still called `configFilePath`, maybe this can `filePath` directly

agent/engine/execcmd/agent_version.go
cyastella: ok

agent/engine/execcmd/agent_version.go
cyastella: good point, thanks for the explanation! I thought `sort.Sort()` requires implementing comparator on the type that needs to be compared (i.e. `agentVersion`), didn't realize that it should be implemented on the slice type too

meantime, just curious, would it work to directly implement the interface methods on `[]agentVersion`? this way we won't need the added type `byAgentVersion` either

```go
func (bav []agentVersion) Len() int {
	return len(bav)
}
```

agent/engine/execcmd/manager_init_task_linux.go
angelcar: good catch! Changed :D

cyastella (approved):

agent/engine/execcmd/agent_version_test.go
```diff
@@ -15,108 +15,205 @@ package execcmd
...
-		},
-		{
-			"missing elements v2", "1.1.1.1", "3.0.236", badInputReturn, true,
-		},
-		{
-			"blank v1", "", "1.1.1.1", badInputReturn, true,
-		},
-		{
-			"blank v2", "1.1.1.1", "", badInputReturn, true,
+			"blank rhs", []uint{1, 1, 1, 1}, []uint{}, 1,
```
fierlion, line 52: blank both sides returns 0?

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -133,6 +141,36 @@ func (m *manager) InitializeContainer(taskId string, container *apicontainer.Con
...
+		if !fileExists(filepath.Join(ecsAgentDepsVersionedBinDir, SSMAgentWorkerBinName)) {
+			continue // try falling back to the previous version
+		}
+		if !fileExists(filepath.Join(ecsAgentDepsVersionedBinDir, SessionWorkerBinName)) {
+			continue // try falling back to the previous version
+		}
+		latest = filepath.Join(m.hostBinDir, vStr)
+		break
+	}
+	if latest == "" {
```
fierlion, line 168: Is there a test for this condition?  I might have missed it.
I'm looking over the pr as a whole and keep thinking about what happens with an existing but empty directory -- the sort will return 0 (I believe -- no test for that case but it makes sense) and latest will be the empty string.  This is a pretty important error path -- if no available ssm agents, then big problems.

agent/engine/execcmd/agent_version.go
```diff
@@ -15,68 +15,108 @@ package execcmd
...
+func (bav byAgentVersion) Swap(i, j int) {
+	bav[i], bav[j] = bav[j], bav[i]
+}
+
+var ioUtilReadDir = ioutil.ReadDir
+
+func retrieveAgentVersions(agentBinPath string) ([]agentVersion, error) {
+	files, err := ioUtilReadDir(agentBinPath)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
angelcar: This is a primitive which is only responsible for getting valid versions. I believe what you're referring to is [here](https://github.com/aws/amazon-ecs-agent/pull/2701/files#diff-c4a66e64f3b925c1319fa93708f2e604acd9415469acec3b11a27c490fb88d1aR169), if not versions are found in our folder then there will be an error at some point

agent/engine/execcmd/agent_version_test.go
angelcar: It should, let me add that case here though.

agent/engine/execcmd/manager_init_task_linux.go
angelcar: [These](https://github.com/aws/amazon-ecs-agent/pull/2701/files#diff-ebdff96a4c1a16c1f6b7675e6e244503b94530f6f3f59358c87b42f74b61bb16R78-R88) two test cases cover this

agent/engine/execcmd/agent_version_test.go
angelcar: added test case [here](https://github.com/aws/amazon-ecs-agent/pull/2701/files#diff-6cf239de592683f4c177ec789f5615ded75e41e50ea468fbfc8531930714efffR37)

fierlion (approved):

PR #2702, opened 2020-10-30
Add managed agent stopped and running event integ test

fierlion:
### Summary
Adds integration test for state change events RUNNING and STOPPED.   Removes redundant container event emission of managed agent RUNNING->RUNNING.

### Implementation details
Added a condition around emitContainerEvent.
Re-used a minimal section of TestExecCommandAgent to test managed agent container event emission for RUNNING and for STOPPED

### Testing
ran 30x to test for flakiness locally:
`sudo -E env "PATH=$PATH" go test -tags sudo -count=1 -v -run="TestManagedAgentEvent" ./agent/engine/...`

Add ManagedAgentEvent integration test.

### Licensing

### DIFF
--- a/agent/engine/common_integ_test.go
+++ b/agent/engine/common_integ_test.go
@@ -117,14 +117,18 @@ func verifyContainerRunningStateChangeWithRuntimeID(t *testing.T, taskEngine Tas
 		"Expected container runtimeID should not empty")
 }
 
-func verifyExecAgentRunningStateChange(t *testing.T, taskEngine TaskEngine) {
+func verifyExecAgentStateChange(t *testing.T, taskEngine TaskEngine,
+	expectedStatus apicontainerstatus.ManagedAgentStatus, waitDone chan<- struct{}) {
 	stateChangeEvents := taskEngine.StateChangeEvents()
-	event := <-stateChangeEvents
-	containerEvent := event.(api.ContainerStateChange)
-	assert.NotEmpty(t, containerEvent.ManagedAgents, "Expected exec-agent event has no managed agents")
-	if containerEvent.ManagedAgents != nil {
-		assert.Equal(t, apicontainerstatus.ManagedAgentRunning, containerEvent.ManagedAgents[0].Status,
-			"expected managedAgent container state change event did not match actual event")
+	for event := range stateChangeEvents {
+		if containerEvent, ok := event.(api.ContainerStateChange); ok {
+			if containerEvent.ManagedAgents != nil {
+				if containerEvent.ManagedAgents[0].Status == expectedStatus {
+					close(waitDone)
+					return
+				}
+			}
+		}
 	}
 }
 
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -310,13 +310,15 @@ func (engine *DockerTaskEngine) monitorExecAgentRunning(ctx context.Context,
 	// to finish monitoring.
 	// This is inspired from containers streaming stats from Docker.
 	time.Sleep(retry.AddJitter(time.Nanosecond, engine.monitorExecAgentsInterval/2))
-	_, err = engine.execCmdMgr.RestartAgentIfStopped(ctx, engine.client, task, c, dockerID)
+	status, err := engine.execCmdMgr.RestartAgentIfStopped(ctx, engine.client, task, c, dockerID)
 	if err != nil {
 		seelog.Errorf("Task engine [%s]: Failed to restart ExecCommandAgent Process for container [%s]: %v", task.Arn, dockerID, err)
+		mTask.emitContainerEvent(mTask.Task, c, "")
 	}
-	// whether we restarted or failed to restart, we'll want to emit a state change event
-	// redundant state change events like RUNNING->RUNNING are allowed
-	mTask.emitContainerEvent(mTask.Task, c, "")
+	if status == execcmd.Restarted {
+		mTask.emitContainerEvent(mTask.Task, c, "")
+	}
+
 }
 
 // MustInit blocks and retries until an engine can be initialized.
--- a/agent/engine/docker_task_engine_test.go
+++ b/agent/engine/docker_task_engine_test.go
@@ -3229,24 +3229,34 @@ func TestMonitorExecAgentRunning(t *testing.T) {
 		containerStatus                apicontainerstatus.ContainerStatus
 		execCommandAgentState          apicontainer.ManagedAgentState
 		execAgentStatus                apicontainerstatus.ManagedAgentStatus
+		restartStatus                  execcmd.RestartStatus
 		simulateBadContainerId         bool
 		expectedRestartInUnhealthyCall bool
 		expectContainerEvent           bool
 	}{
 		{
 			containerStatus:      apicontainerstatus.ContainerStopped,
-			expectContainerEvent: false,
 			execAgentStatus:      apicontainerstatus.ManagedAgentStopped,
+			restartStatus:        execcmd.NotRestarted,
+			expectContainerEvent: false,
 		},
 		{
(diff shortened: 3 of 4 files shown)


agent/engine/common_integ_test.go
```diff
@@ -117,14 +117,18 @@ func verifyContainerRunningStateChangeWithRuntimeID(t *testing.T, taskEngine Tas
...
+	expectedStatus apicontainerstatus.ManagedAgentStatus, waitDone chan<- struct{}) {
 	stateChangeEvents := taskEngine.StateChangeEvents()
-	event := <-stateChangeEvents
-	containerEvent := event.(api.ContainerStateChange)
-	assert.NotEmpty(t, containerEvent.ManagedAgents, "Expected exec-agent event has no managed agents")
-	if containerEvent.ManagedAgents != nil {
-		assert.Equal(t, apicontainerstatus.ManagedAgentRunning, containerEvent.ManagedAgents[0].Status,
-			"expected managedAgent container state change event did not match actual event")
+	for event := range stateChangeEvents {
+		if containerEvent, ok := event.(api.ContainerStateChange); ok {
```
angelcar, line 124: there's a possibility here that we receive an `api.ContainerStateChange` but not the one we were looking for. Is there a way to make sure it's a container event for managed agent? 

I am thinking actually that maybe `api.ContainerStateChange` was meant for actual changes in the container state, maybe we should consider to create a `api.ManagedAgentStateChange`? Not for this PR maybe, but something to consider

agent/engine/docker_task_engine_test.go
```diff
@@ -3281,18 +3291,27 @@ func TestMonitorExecAgentRunning(t *testing.T) {
 		if tc.simulateBadContainerId {
 			testTask.Containers[0].RuntimeID = ""
 		}
-		if tc.containerStatus == apicontainerstatus.ContainerRunning && !tc.simulateBadContainerId {
-			execCmdMgr.EXPECT().RestartAgentIfStopped(dockerTaskEngine.ctx, dockerTaskEngine.client, testTask, testTask.Containers[0], testContainerId).
+		if tc.containerStatus == apicontainerstatus.ContainerRunning && tc.willRestart {
```
angelcar, line 3294: I was having issues reading the logic here because I wasn't seeing that this line is the only lines that change are # 3297 and # 3303

Would it be better to express this whole block as:

```golang
restartStatus := execcmd.Restarted
if !tc.willRestart {
    restartStatus := execcmd.NotRestarted
}

if tc.containerStatus == apicontainerstatus.ContainerRunning && !tc.simulateBadContainerId {
    execCmdMgr.EXPECT().RestartAgentIfStopped(dockerTaskEngine.ctx, dockerTaskEngine.client, testTask,
        testTask.Containers[0], testContainerId).
        Return(restartStatus, nil).
        Times.(1)
}
```

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -424,6 +424,89 @@ func TestExecCommandAgent(t *testing.T) {
...
+
+			const (
+				testTaskId        = "exec-command-agent-test-task"
+				testContainerName = "exec-command-agent-test-container"
+			)
+
+			client, err := sdkClient.NewClientWithOpts(sdkClient.WithHost(endpoint), sdkClient.WithVersion(sdkclientfactory.GetDefaultVersion().String()))
+			require.NoError(t, err, "Creating go docker client failed")
+
+			testExecCmdHostBinDir, err := filepath.Abs("../../misc/exec-command-agent-test")
```
angelcar, line 459: note: this will need to change after https://github.com/aws/amazon-ecs-agent/pull/2701 gets merged.

See: https://github.com/aws/amazon-ecs-agent/pull/2701/files#diff-8b609ac402c5957da4b973469c270e0047c06a8beefabfaca9555ca0dd61f83dR367

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -424,6 +424,89 @@ func TestExecCommandAgent(t *testing.T) {
...
+			require.NoError(t, err)
+
+			taskEngine, done, _ := setupEngineForExecCommandAgent(t, testExecCmdHostBinDir)
+			defer done()
+
+			testTask := createTestExecCommandAgentTask(testTaskId, testContainerName, time.Minute*tc.ManagedAgentLifetime)
```
angelcar, line 465: maybe we can extract the duplicated lines of this test and `TestExecCommandAgent` into a function?

agent/engine/docker_task_engine_test.go
angelcar: or even better, we could put the `restartStatus` in the test table, instead of `willRestart`

agent/engine/engine_sudo_linux_integ_test.go
angelcar: I guess if this PR gets merged first I could just change this test in https://github.com/aws/amazon-ecs-agent/pull/2701 before merging

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -424,6 +424,89 @@ func TestExecCommandAgent(t *testing.T) {
...
+			go func() {
+				verifyExecAgentStateChange(t, taskEngine, tc.ExpectedStatus, waitDone)
+				cancel()
+			}()
+			<-ctx.Done()
+
+			timeout := false
+			select {
+			case <-waitDone:
+			case <-time.After(1 * time.Second):
```
angelcar, line 496: I think you can get rid of `ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)` and just wait for `case <-time.After(20 * time.Second):` here?

agent/engine/common_integ_test.go
fierlion: Even with the redundant event filter I was seeing events with NONE managedAgent status prior to the expected events, and this function sifted past these events.  
I agree that we might want to differentiate the events meant to converse with Docker from the events meant to converse with ECS Control Plane.  Both could be considered ContainerEvents.
(not in present pr though.)

agent/engine/engine_sudo_linux_integ_test.go
fierlion: Personally for this and the previous test I like to see the explicit flow from top to bottom, especially in the above test with the restart, even if it means some repetition.
I'd like to take a three strikes approach to extracting these duplicated lines.

angelcar (approved):

sharanyad (approved):

agent/engine/docker_task_engine.go
```diff
@@ -310,13 +310,15 @@ func (engine *DockerTaskEngine) monitorExecAgentRunning(ctx context.Context,
...
 	time.Sleep(retry.AddJitter(time.Nanosecond, engine.monitorExecAgentsInterval/2))
-	_, err = engine.execCmdMgr.RestartAgentIfStopped(ctx, engine.client, task, c, dockerID)
+	status, err := engine.execCmdMgr.RestartAgentIfStopped(ctx, engine.client, task, c, dockerID)
 	if err != nil {
 		seelog.Errorf("Task engine [%s]: Failed to restart ExecCommandAgent Process for container [%s]: %v", task.Arn, dockerID, err)
+		mTask.emitContainerEvent(mTask.Task, c, "")
 	}
-	// whether we restarted or failed to restart, we'll want to emit a state change event
-	// redundant state change events like RUNNING->RUNNING are allowed
-	mTask.emitContainerEvent(mTask.Task, c, "")
```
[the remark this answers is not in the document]
fierlion: So redundant events (RUNNING->RUNNING) aren't emitted.  This change means we only emit a container event when there is a "restarted" (aka STOPPED->RUNNING) or an error (aka RUNNING->STOPPED)

PR #2704, opened 2020-11-03
generate and mount seelog config file to exec containers

sharanyad:
### Summary
This generates static seelog config file for exec agent if its not already present. If present, it checks the validity of the seelog config content. 
This file is then bind mounted to the exec enabled containers. 

### Testing

N/A

### Licensing

### DIFF
--- a/agent/engine/execcmd/manager_init_task_linux.go
+++ b/agent/engine/execcmd/manager_init_task_linux.go
@@ -63,7 +63,9 @@ const (
 	// ECSAgentExecConfigDir is the directory where ECS Agent will write the ExecAgent config files to
 	ECSAgentExecConfigDir = ecsAgentExecDepsDir + "/" + ContainerConfigDirName
 	// HostExecConfigDir is the dir where ExecAgents Config files will live
-	HostExecConfigDir = hostExecDepsDir + "/" + ContainerConfigDirName
+	HostExecConfigDir          = hostExecDepsDir + "/" + ContainerConfigDirName
+	ExecAgentLogConfigFileName = "seelog.xml"
+	ContainerLogConfigFile     = "configuration/" + ExecAgentLogConfigFileName
 )
 
 var (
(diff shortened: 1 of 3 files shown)


agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -211,6 +243,35 @@ func getSessionWorkersLimit(ma apicontainer.ManagedAgent) int {
 	return limit
 }
 
+var GetExecAgentLogConfigFile = getAgentLogConfigFile
```
angelcar, line 246: does this need to be exported?

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -63,7 +63,9 @@ const (
 	// ECSAgentExecConfigDir is the directory where ECS Agent will write the ExecAgent config files to
 	ECSAgentExecConfigDir = ecsAgentExecDepsDir + "/" + ContainerConfigDirName
 	// HostExecConfigDir is the dir where ExecAgents Config files will live
-	HostExecConfigDir = hostExecDepsDir + "/" + ContainerConfigDirName
+	HostExecConfigDir          = hostExecDepsDir + "/" + ContainerConfigDirName
+	ExecAgentLogConfigFileName = "seelog.xml"
```
angelcar, line 67: I think we probably want to follow the same naming convention that we do with `execAgentConfigFileNameTemplate` (using the SHA in the name of the file). This file might be confgurable in the future and when that happens we will be ready. Plus it's pretty much the same code that we already use for the config file.

angelcar (changes requested):

agent/engine/execcmd/manager_init_task_linux.go
sharanyad: yes, used in engine integ test

agent/engine/execcmd/manager_init_task_linux.go
sharanyad: after offline discussion, decided it would be useful to use hash for initial launch as well.

agent/engine/execcmd/manager_init_task_linux.go
angelcar: This is probably for another PR, but I don't like these package level functions when they're exported because any other package could override it and break the functionality of this package...

Maybe instead of using this function in integ test, we could just use a regex. I think the main test that needs to check the actual hash is the UT. The integ test can be more relaxed about the specifics since typically inte tests are gray, if not black box tests

angelcar (approved):

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -80,8 +82,31 @@ var (
 		}
 	}
 }`
+	execAgentLogConfigTemplate = `<seelog type="adaptive" mininterval="2000000"
```
cyastella, line 85: nit: I think this might look a bit nicer if we format it like this:

```go
template = `
<seelog type="adaptive" mininterval="2000000" maxinterval="100000000" critmsgcount="500" minlevel="info">
<exceptions>
<exception filepattern="test*" minlevel="error"/>
</exceptions>
<outputs formatid="fmtinfo">
<console formatid="fmtinfo"/>
<rollingfile type="size"
	filename="/var/log/amazon/ssm/amazon-ssm-agent.log"
	maxsize="40000000" maxrolls="1"/>
<filter levels="error,critical" formatid="fmterror">
   <rollingfile type="size"
		filename="/var/log/amazon/ssm/errors.log"
		maxsize="10000000" maxrolls="1"/>
</filter>
</outputs>
<formats>
<format id="fmterror" format="%Date %Time %LEVEL [%FuncShort @ %File.%Line] %Msg%n"/>
<format id="fmtdebug" format="%Date %Time %LEVEL [%FuncShort @ %File.%Line] %Msg%n"/>
<format id="fmtinfo" format="%Date %Time %LEVEL %Msg%n"/>
</formats>
</seelog>`
```

we can add indentation for xml here too

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -211,6 +244,38 @@ func getSessionWorkersLimit(ma apicontainer.ManagedAgent) int {
...
+	return logConfigFileName, nil
+}
+
+func validLogConfigExists(configFilePath, expectedHash string) bool {
+	config, err := getFileContent(configFilePath)
+	if err != nil {
+		return false
+	}
+	hash := getExecAgentConfigHash(string(config))
+	return strings.Compare(expectedHash, hash) == 0
```
cyastella, line 270: nit: maybe we can use `==` here, or we might need `strings.Compare()` for case-insensitive comparision

cyastella (approved):

PR #2706, opened 2020-11-03
add mock for pathExists() to avoid accessing file system in unit test

cyastella:
### Summary
this change adds mock for `agent_capability.pathExists()` to avoid accessing file system in unit test

### Implementation details
add mock for `pathExists()` to avoid accessing file system in unit test

### Testing

this change fixes unit tests that are currently failing

after running integ tests, paths like `/managed-agents/execute-command/bin` are created with `root` permission and when unit tests try to access these files/directories, it would cause a permission error

fix unit tests, add mock

### Licensing


angelcar (commented):
hmm, after seeing this PR, I am thinking that for this particular case it might actually be better to mock `pathExists` in an init function and remove all the mocks from individual tests 

```golang 
func init() {
	pathExists = func(path string, shouldBeDirectory bool) (bool, error) {
		return false, nil
	}
}
```
It is kind of noisy to have to add the same method over and over

agent/app/agent_capability_test.go
```diff
@@ -162,6 +163,13 @@ func TestCapabilities(t *testing.T) {
 }
 
 func TestCapabilitiesECR(t *testing.T) {
+	pathExists = func(path string, shouldBeDirectory bool) (bool, error) {
```
angelcar, line 166: Another option to reduce verbosity would be to create a function like below:

```golang
func mockPathExists() func() {
	pathExists = func(path string, shouldBeDirectory bool) (bool, error) {
		return false, nil
	}
	return func() {
		pathExists = defaultPathExists
	}()
}
```

And invoke it here like so: `defer mockPathExists()()`

cyastella: change to add `mockPathExists()`, `getSubDirectories` is only mocked in `TestCapabilities` because we want to enable `execute-command` capability and verify that it exists so not adding helper function for it

cyastella:
`integ_test`s are failing because this feature branch needs to be updated to match `dev` branch for this change, https://github.com/aws/amazon-ecs-agent/blob/dev/agent/engine/engine_sudo_linux_integ_test.go#L51. Tests are passing locally

angelcar:
> `integ_test`s are failing because this feature branch needs to be updated to match `dev` branch for this change, https://github.com/aws/amazon-ecs-agent/blob/dev/agent/engine/engine_sudo_linux_integ_test.go#L51. Tests are passing locally

You can just update this in your PR and there should be no conflict merging, I believe rebasing from dev is needed here

angelcar (approved):

sharanyad (approved):

angelcar:
Closing in favor of https://github.com/aws/amazon-ecs-agent/pull/2722

PR #2708, opened 2020-11-06
save RCI client token to agent state

sharanyad:
### Summary
Adding Register Container Instance's client /registration token to agent state (boltdb only) 
This is so that ECS service registers using the same container instance ARN for the container instance that issues quick back to back RCI requests. 

### Implementation details
Saving new registration token key and value to boltDB 

### Testing

Manually tested by removing the container instance ARN from boltdb state. Restarted agent and received the same container instance ARN as before. 

Fix a bug where agent persists RCI client token to avoid being registered as different container instance ARNs. 

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -117,6 +117,7 @@ type ecsAgent struct {
 	resourceFields              *taskresource.ResourceFields
 	availabilityZone            string
 	latestSeqNumberTaskManifest *int64
+	registrationToken           string
 }
 
 // newAgent returns a new ecsAgent object, but does not start anything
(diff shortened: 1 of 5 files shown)


angelcar (approved):

agent/app/agent.go
```diff
@@ -330,6 +331,7 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		agent.saveMetadata(data.ClusterNameKey, agent.cfg.Cluster)
 		agent.saveMetadata(data.ContainerInstanceARNKey, agent.containerInstanceARN)
 		agent.saveMetadata(data.EC2InstanceIDKey, currentEC2InstanceID)
+		agent.saveMetadata(data.RegistrationTokenKey, agent.registrationToken)
```
fenxiong, line 334: this is after RCI. should we save this before calling RCI? so that if the agent stops before RCI finishes, it can call with the same token when starting up again

fenxiong (approved):

PR #2713, opened 2020-11-10
Switch back to latest Firelens image in test and update the test

fenxiong:
### Summary

Switch back to latest Firelens image in test and update the test to work with it.

Previously, the version was pinned to 2.7.0 because 2.9.0 breaks the integ test `TestFirelensFluentbit`. The failure was caused by https://github.com/aws/amazon-cloudwatch-logs-for-fluent-bit/pull/115 which adds a dependency between Firelens and the agent's task metadata endpoint, while the integ test does not start or mock the task metadata server.

This change fixed the test by mocking the task metadata endpoint, and switch the image version back to latest, so that we can continue verify the feature with latest version of dependency.

### Implementation details

Add logic in `TestFirelensFluentbit` test to:
* Start a test server with `httptest` to mock task metadata endpoint;
* Overwrite container's v3 task metadata endpoint to be the mock server's address.

### Testing

Fixing an existing test. Ran the test locally and it passed. Result can be verified by pull request checks. 

N/A

### Licensing

### DIFF
--- a/agent/engine/engine_sudo_linux_integ_test.go
+++ b/agent/engine/engine_sudo_linux_integ_test.go
@@ -19,6 +19,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -30,6 +34,7 @@ import (
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/ec2"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	cgroup "github.com/aws/amazon-ecs-agent/agent/taskresource/cgroup/control"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/firelens"
@@ -48,7 +53,7 @@ import (
 
 const (
 	testLogSenderImage = "amazonlinux:2"
-	testFluentbitImage = "amazon/aws-for-fluent-bit:2.7.0"
+	testFluentbitImage = "amazon/aws-for-fluent-bit:latest"
 	testVolumeImage    = "127.0.0.1:51670/amazon/amazon-ecs-volumes-test:latest"
 	testCluster        = "testCluster"
 	validTaskArnPrefix = "arn:aws:ecs:region:account-id:task/"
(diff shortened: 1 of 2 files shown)


agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -164,6 +174,20 @@ func TestFirelensFluentbit(t *testing.T) {
...
+	// Note that the listener has to be overwritten here, because the default one from httptest.NewServer
+	// only listens to localhost and isn't reachable from container running in bridge network mode.
+	l, err := net.Listen("tcp", ":0")
+	require.NoError(t, err)
+	server := httptest.NewUnstartedServer(mockTaskMeatadataHandler)
+	server.Listener = l
+	server.Start()
+	defer server.Close()
+
+	port := getPortFromAddr(server.URL, t)
```
angelcar, line 187: nit: let's past `t` first, as that seems to be the convention

sharanyad (approved):

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -238,6 +262,31 @@ func TestFirelensFluentbit(t *testing.T) {
...
+// getHostPrivateIP returns the host's private IP.
+func getHostPrivateIP(ec2MetadataClient ec2.EC2MetadataClient, t *testing.T) string {
+	ip, err := ec2MetadataClient.PrivateIPv4Address()
+	require.NoError(t, err)
+	return ip
+}
+
+// setV3MetadataURLFormat sets the container metadata URI format and returns a function to set it back.
+func setV3MetadataURLFormat(fmt string) func() {
+	oFmt := apicontainer.MetadataURIFormat
```
angelcar, line 283: nit: can we call this variable backup so that it's more meaningful what the purpose of this var is?

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -238,6 +262,31 @@ func TestFirelensFluentbit(t *testing.T) {
 	assert.Error(t, err)
 }
 
+// getPortFromAddr returns the port part of an address in format "http://<addr>:<port>".
+func getPortFromAddr(addr string, t *testing.T) string {
+	parts := strings.Split(addr, ":")
```
angelcar, line 267: nit: I think `net.SplitHostPort(u.Host)` might be more appropriate here, although you need an extra step: 
https://play.golang.org/p/glnDMk9BkPr

agent/engine/engine_sudo_linux_integ_test.go
fenxiong: updated

agent/engine/engine_sudo_linux_integ_test.go
fenxiong: updated

angelcar (approved):

PR #2719, opened 2020-11-11
update v4 metadata to return network despite nil

fierlion:
### Summary
This updates the v4 metadata endpoint to return network metadata for all available.  Previously we threw a `http.StatusInternalServerError` (500) in the case of any container in the task not having available network metadata.  

### Implementation details
removed Error cases for individual missing metadata and added a `Warn` log level entry.

### Testing
make gogenerate 
make test
(will update integ/functional tests as necessary given automated tests)

Update v4 metadata to return network despite nil

### Licensing

### DIFF
--- a/agent/handlers/v4/task_metadata_handler.go
+++ b/agent/handlers/v4/task_metadata_handler.go
@@ -65,12 +65,7 @@ func TaskMetadataHandler(state dockerstate.TaskEngineState, ecsClient api.ECSCli
 			for _, containerResponse := range taskResponse.Containers {
 				networks, err := GetContainerNetworkMetadata(containerResponse.ID, state)
 				if err != nil {
-					errResponseJSON, err := json.Marshal(err.Error())
-					if e := utils.WriteResponseIfMarshalError(w, err); e != nil {
-						return
-					}
-					utils.WriteJSONToResponse(w, http.StatusInternalServerError, errResponseJSON, utils.RequestTypeContainerMetadata)
-					return
+					seelog.Warnf("Error retrieving network metadata for container %s - %s", containerResponse.ID, err)
 				}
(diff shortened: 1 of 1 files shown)


sparrc:
could you delete the error log on this line so this doesnt get double-logged? https://github.com/aws/amazon-ecs-agent/blob/84321fcb15787f1d3a3f8aed80002d6d7d080ec6/agent/handlers/v4/container_metadata_handler.go#L91-L92,

also i would prefer to reword the error returned there^ to soemthing like `no network settings found for container....`

mssrivas (approved):

angelcar (approved):

agent/handlers/v4/task_metadata_handler.go
```diff
@@ -65,12 +65,7 @@ func TaskMetadataHandler(state dockerstate.TaskEngineState, ecsClient api.ECSCli
 			for _, containerResponse := range taskResponse.Containers {
 				networks, err := GetContainerNetworkMetadata(containerResponse.ID, state)
 				if err != nil {
-					errResponseJSON, err := json.Marshal(err.Error())
-					if e := utils.WriteResponseIfMarshalError(w, err); e != nil {
-						return
-					}
-					utils.WriteJSONToResponse(w, http.StatusInternalServerError, errResponseJSON, utils.RequestTypeContainerMetadata)
-					return
+					seelog.Warnf("Error retrieving network metadata for container %s - %s", containerResponse.ID, err)
```
sharanyad, line 68: is there any test case that can be updated to verify this?

fierlion: the only tests we have for the v4 endpoint are our functional tests presently.  I'm looking into what's necessary.

PR #2720, opened 2020-11-11
scripts: add ci-ecr-pull and ci-ecr-push

sparrc:
These scripts are for pushing the dockerhub images that we need for our
ci system (integ test and build) to ECR, and for pulling those images
down within the CI system.

ci-ecr-pull will tag the ECR images like 'golang:1.15.4' so that users
who do not run this script on their laptop or elsewhere can still pull
from dockerhub, only if this script is run will the builds reference the
ECR images.

The reason behind this is that dockerhub is now rate-limiting anonymous
pulls, so in order to avoid being rate-limited we should pull our images
from ECR as much as we can.

This should also speed up builds because we will be pulling our large
build images (golang, gcc) from in-region s3 rather than over the
internet.

One side-effect of this change is that we have to reference full version
numbers in our dockerfiles (1.15.4 instead of 1.15), in order to avoid
querying docker for the latest version, and to avoid dockerhub pulls
happening when an image is updated.

### Testing

ran inside an integration test instance and verified image pulling behavior

not applicable

### Licensing


misc/gremlin/Dockerfile
```diff
@@ -10,7 +10,7 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-FROM golang:1.15 as build
+FROM golang:1.15.4 as build
```
angelcar, line 13: we an probably get this from the root make file as well, like other dockerfiles do now, right?

misc/netkitten/Dockerfile.linux
```diff
@@ -10,7 +10,7 @@
 # on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
-FROM golang:1.15 as build
+FROM golang:1.15.4 as build
```
angelcar, line 13: same as above

scripts/ci-ecr-pull
```diff
@@ -0,0 +1,66 @@
...
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+set -e
+
+AWS_REGION="${1:-}"
```
angelcar, line 18: nit: I think the default is redundant here?

scripts/ci-ecr-push
```diff
@@ -0,0 +1,76 @@
...
+
+AWS_REGION="${1:-}"
+AWS_ACCOUNT_ID="${2:-}"
+
+images="registry:2.7.0
+golang:1.15.4
+gcc:7.3.0
+busybox:1.32.0
+amazonlinux:2.0.20200722.0
+amazon/aws-for-fluent-bit:2.7.0"
```
angelcar, line 26: can we extract this block to a common file and use `source` here to import it?

angelcar: I think we can also create a function in this common file to retrieve`ECR_URI="${AWS_ACCOUNT_ID}.dkr.ecr.${AWS_REGION}.amazonaws.com"`

misc/gremlin/Dockerfile
sparrc: this has it's own makefile so not sure how passing the var works, but I will try to do that yes

sparrc: done

misc/netkitten/Dockerfile.linux
sparrc: done

angelcar (approved):

scripts/ci-ecr-pull
sparrc: done

scripts/ci-ecr-push
sparrc: Ive change the list of images to be sourced from a single file, but personally not a fan of having the ecr uri in a function since it's not likely to ever change and doesnt seem like the deduplication is worth having to jump between files

mythri-garaga (approved):

PR #2733, opened 2020-11-17
check the validity of exec agent config

sharanyad:
### Summary
do the same hash comparison check as exec agent log config for the agent config. 

### Testing

### Licensing

### DIFF
--- a/agent/engine/execcmd/manager_init_task_linux.go
+++ b/agent/engine/execcmd/manager_init_task_linux.go
@@ -251,7 +251,7 @@ func getAgentLogConfigFile() (string, error) {
 	logConfigFileName := fmt.Sprintf(logConfigFileNameTemplate, hash)
 	// check if config file exists already
 	logConfigFilePath := filepath.Join(ECSAgentExecConfigDir, logConfigFileName)
-	if fileExists(logConfigFilePath) && validLogConfigExists(logConfigFilePath, hash) {
+	if fileExists(logConfigFilePath) && validConfigExists(logConfigFilePath, hash) {
 		return logConfigFileName, nil
 	}
 	// create new seelog config file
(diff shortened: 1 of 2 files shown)


agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -284,8 +284,7 @@ func getAgentConfigFileName(sessionLimit int) (string, error) {
 	configFileName := fmt.Sprintf(execAgentConfigFileNameTemplate, hash)
 	// check if config file exists already
 	configFilePath := filepath.Join(ECSAgentExecConfigDir, configFileName)
-	if fileExists(configFilePath) {
-		// TODO: verify the hash of the existing file contents
+	if fileExists(configFilePath) && validConfigExists(configFilePath, hash) {
```
angelcar, line 287: maybe this is for a new PR, but can you check what happens when a directory with the same name as the config file already exists on disk?

My guess is that `createNewExecAgentConfigFile` will fail. I think maybe we need to make sure we remove a directory (or an existing file which is tampered with) before writing the new file

angelcar (approved):

ubhattacharjya (approved):

agent/engine/execcmd/manager_init_task_linux.go
sharanyad: today we return an error for the following:
- if the file is a directory
- if the file is tampered 

Are you suggesting we do not return an error in these cases?

sharanyad: will change to delete the directory

sharanyad:
`ecs/inf/integ_test ` shows as pending, but have verified from logs that all tests have passed.

PR #2735, opened 2020-11-18
fix final loose dockerhub pulls from build and integ tests

sparrc:
### Summary

This is a follow-up to https://github.com/aws/amazon-ecs-agent/pull/2720

After running tests with dockerhub throttling active we have discovered a couple more places where we were pulling dockerhub images.

Specifically, when doing a `FROM` statement within a Dockerfile, no pull request goes to dockerhub if the image is already present locally. However, `docker pull` always pulls at least a manifest from dockerhub, which is subject to throttling. So this PR removes one `docker pull` in setting up the test registry.

It also short-circuits upload-images because this script is essentially deprecated. We can't yet remove it as it's still referenced in internal step functions, but once we deprecate that stack we can remove.

Lastly, on arm we need to avoid trying to build misc/fluentd. The base image of this (fluent/fluentd) doesnt exist for arm, and so we don't have it in ECR. This means that when we try to build it we will request a manifest from dockerhub, which can again be throttled. 

### Testing

not applicable

### Licensing

### DIFF
--- a/scripts/setup-test-registry
+++ b/scripts/setup-test-registry
@@ -14,7 +14,7 @@
 
 # Run a local registry on the 'well known' port 51670 if it is not running.
 # Also push images we will need to it.
-set -e
+set -ex
 
 REGISTRY_IMAGE="registry:2.7.0"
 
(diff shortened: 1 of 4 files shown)


shubham2892 (approved):

scripts/setup-test-registry
```diff
@@ -29,12 +29,8 @@ docker stop "$REGISTRY_CONTAINER_NAME" || true && docker rm "$REGISTRY_CONTAINER
...
 echo "Running $REGISTRY_CONTAINER_NAME"
 docker run -d --name="$REGISTRY_CONTAINER_NAME" -e SETTINGS_FLAVOR=local -p "127.0.0.1:51670:5000" "${REGISTRY_IMAGE}"
-
-# Make sure our images are pushed to it
-mirror_image() {
-  docker pull $1
-  mirror_local_image $@
-}
+# give the registry some seconds to get ready for pushes
+sleep 7
```
sparrc, line 33: Defining the function here seems to have taken just enough time for some of the slower instances (like inf1) to reliably push to this local repo. Adding the sleep to avoid random tcp and EOF errors when we move on to push images to the local registry at `127.0.0.1:51670`

mssrivas (approved):

PR #2755, opened 2020-12-03
Batch ManagedAgent state changes

sharanyad:
### Summary
- Batching managed agent state changes as part of task state change events

Changes that are not part of the PR: 
- Model changes for Managed agent state change directly under task state change 
- Modifying the `emitManagedAgentEvent` to emit managed agent state events instead of container events
- Update the sent status of managed agent after task events are sent to backend (depends on the model change) 
- Add managed exec agent stopped state change when corresponding container is stopped (needs `emitManagedAgentEvent` changes)

### Implementation details
- Introduce new `tasksToManagedAgentStates` map to store managed agent state changes
- Handle `statechange.ManagedAgentEvent` separately by having its own batching logic (similar to containers) 
- On `startDrainEventsTicker` trigger, task change event is created for the batched managed agent state changes(duplicates are handled) 
- On flushing the batched task events(`flushBatchUnsafe`), managed agent events are added to the above created task events. The map is then cleaned here. 

### Testing

N/A

### Licensing


agent/eventhandler/task_handler.go
```diff
@@ -218,40 +230,47 @@ func (handler *TaskHandler) taskStateChangesToSend() []api.TaskStateChange {
...
 
-	// For now, evey time a container stops, we also send stop for exec command agent
-	if event.GetEventType() == statechange.ContainerEvent && event.Status == status.ContainerStopped {
-		if _, ok := event.Container.GetManagedAgentByName(execcmd.ExecuteCommandAgentName); ok {
-			event.ManagedAgents = []api.ManagedAgentStateChange{{
-				Name:   execcmd.ExecuteCommandAgentName,
-				Status: status.ManagedAgentStopped,
-				Reason: "Parent container stopped",
-			}}
+			events = append(events, event)
```
angelcar, line 256: I believe this is not going to work. We might need to dedupe `TaskStateChange`s in this method

agent/eventhandler/task_handler.go
```diff
@@ -260,7 +279,24 @@ func (handler *TaskHandler) flushBatchUnsafe(taskStateChange *api.TaskStateChang
 	// All container events for the task have now been copied to the
 	// task state change object. Remove them from the map
 	delete(handler.tasksToContainerStates, taskStateChange.TaskARN)
-
+	taskStateChange.ManagedAgents = append(taskStateChange.ManagedAgents,
+		handler.tasksToManagedAgentStates[taskStateChange.TaskARN]...)
+	// All managed agent events for the task have now been copied to the
+	// task state change object. Remove them from the map
+	delete(handler.tasksToManagedAgentStates, taskStateChange.TaskARN)
+	if taskStateChange.Task.GetKnownStatus() >= apitaskstatus.TaskStopped {
```
angelcar, line 287: We need to send status stopped for managed agent when the container is stopped, regardless of the task state. This is not going to send stopped when task is running but containers are stopped (which I believe can only happen with non essential containers)

sharanyad: missed that, will change

agent/eventhandler/task_handler.go
sharanyad: dedupe logic is already present, modified that a bit.

agent/eventhandler/task_handler.go
sharanyad: removed this change. have added a TODO and will address when model changes are in.

angelcar (approved):

fierlion:
I've pulled these changes into https://github.com/aws/amazon-ecs-agent/pull/2758
They're very slightly modified.

angelcar:
Closed since these changes were included in https://github.com/aws/amazon-ecs-agent/pull/2758

PR #2758, opened 2020-12-05
Migrate ManagedAgent state change to task

fierlion:
### Summary
This migrates the ManagedAgent state change events from the Container state change events to the Task-level state change events to support bugfix model changes.

### Implementation details
This adds batched ManagedAgent events to the Task event

### Testing
I tested manually by building agent/init on an exec-capable container-instance and running 2 separate tasks, one with a single container and another with 3 containers
```
aws ecs start-task \
    --endpoint-url <gamma endpoint> 
    --cluster gammaDefault \
    --task-definition exec-task:2 \
    --container-instances <instance-id> \
    --enable-execute-command
```
followed by:
```
aws ecs describe-tasks \
    --cluster gammaDefault \
    --endpoint-url <gamma endpoint> \
    --tasks <task-id> \
    --debug
```
I got the following output:
```
                    "managedAgents": [
                        {
                            "lastStartedAt": 1607631456.641,
                            "lastStatus": "RUNNING",
                            "name": "ExecuteCommandAgent",
                            "reason": "Execute Command Agent started"
                        }
                    ],
```

New tests cover the changes: Yes.

### Description for the changelog
Migrate ManagedAgent state change from container to task event.

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2759, opened 2020-12-08
Create container without enabling exec feature if exec bind mounting …

ubhattacharjya:
### Summary

Initially, if a task is exec enabled but ecs agent somehow fails to bind mount exec related dependencies to the containers in the task, the task would fail as container wont start. This PR solves this problem. It will start the container without enabling exec if exec bind mounting fails for any reason. It will also update Managed agent of the container to STOPPED status and the reason will be the error that we get when bind mount fails. 

When the above case happens, We will not try starting exec agent or restarting exec agent.

### Testing

New tests cover the changes: 
Unit tests and integ tests cover the change.

I have tested even if I remove any one of the deps folder, the container starts up anyway. 
```
docker inspect <container-id>

$....
....
....
Mounts: []
...
...
```
### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -287,7 +287,9 @@ func (engine *DockerTaskEngine) monitorExecAgentProcesses(ctx context.Context) {
 		}
 		for _, c := range task.Containers {
 			if execcmd.IsExecEnabledContainer(c) {
-				go engine.monitorExecAgentRunning(ctx, mTask, c)
+				if ma, _ := c.GetManagedAgentByName(execcmd.ExecuteCommandAgentName); !ma.InitFailed {
+					go engine.monitorExecAgentRunning(ctx, mTask, c)
+				}
 			}
 		}
 	}
@@ -1133,15 +1135,15 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 	}
 
 	if execcmd.IsExecEnabledContainer(container) {
-		tId, err := task.GetID()
+		tID, err := task.GetID()
 		if err != nil {
 			herr := &apierrors.HostConfigError{Msg: err.Error()}
 			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
 		}
-		err = engine.execCmdMgr.InitializeContainer(tId, container, hostConfig)
+		err = engine.execCmdMgr.InitializeContainer(tID, container, hostConfig)
 		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+			// TODO: Send a managed agent state change event for this
+			seelog.Warnf("Exec Agent initialization: %v . Continuing to start container without enabling exec feature.", err)
 		}
 	}
 
@@ -1305,18 +1307,21 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
 		}
 	}
 	if execcmd.IsExecEnabledContainer(container) {
-		if err := engine.execCmdMgr.StartAgent(engine.ctx, engine.client, task, container, dockerID); err != nil {
-			seelog.Errorf("Task engine [%s]: Failed to start ExecCommandAgent Process for container [%s]: %v", task.Arn, container.Name, err)
-		}
-		// whether we started or failed to start, we'll want to emit a state change event
-		// redundant state change events like RUNNING->RUNNING are allowed
-		engine.tasksLock.RLock()
-		mTask, ok := engine.managedTasks[task.Arn]
-		engine.tasksLock.RUnlock()
-		if ok {
-			mTask.emitManagedAgentEvent(mTask.Task, container, execcmd.ExecuteCommandAgentName, "Execute Command Agent started")
-		} else {
-			seelog.Errorf("Task engine [%s]: Failed to update status of ExecCommandAgent Process for container [%s]: managed task not found", task.Arn, container.Name)
+		if ma, _ := container.GetManagedAgentByName(execcmd.ExecuteCommandAgentName); !ma.InitFailed {
+			if err := engine.execCmdMgr.StartAgent(engine.ctx, engine.client, task, container, dockerID); err != nil {
+				seelog.Errorf("Task engine [%s]: Failed to start ExecCommandAgent Process for container [%s]: %v", task.Arn, container.Name, err)
+			}
+
+			// whether we started or failed to start, we'll want to emit a state change event
+			// redundant state change events like RUNNING->RUNNING are allowed
+			engine.tasksLock.RLock()
+			mTask, ok := engine.managedTasks[task.Arn]
+			engine.tasksLock.RUnlock()
+			if ok {
+				mTask.emitManagedAgentEvent(mTask.Task, container, execcmd.ExecuteCommandAgentName, "Execute Command Agent started")
+			} else {
+				seelog.Errorf("Task engine [%s]: Failed to update status of ExecCommandAgent Process for container [%s]: managed task not found", task.Arn, container.Name)
+			}
 		}
 	}
 	return dockerContainerMD
--- a/agent/engine/execcmd/manager_init_task_linux.go
+++ b/agent/engine/execcmd/manager_init_task_linux.go
@@ -29,6 +29,7 @@ import (
 	"github.com/pborman/uuid"
 
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 )
 
 const (
@@ -37,7 +38,8 @@ const (
 	ecsAgentExecDepsDir = "/managed-agents/execute-command"
 
 	// ecsAgentDepsBinDir is the directory where ECS Agent will read versions of SSM agent
-	ecsAgentDepsBinDir = ecsAgentExecDepsDir + "/bin"
+	ecsAgentDepsBinDir   = ecsAgentExecDepsDir + "/bin"
+	ecsAgentDepsCertsDir = ecsAgentExecDepsDir + "/certs"
 
 	ContainerDepsDirPrefix = "/ecs-execute-command-"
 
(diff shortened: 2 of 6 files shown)


agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -133,6 +134,10 @@ func (m *manager) InitializeContainer(taskId string, container *apicontainer.Con
 		return err
 	}
 
+	if ok := checkCertsLocation(); ok {
```
sharanyad, line 137: this can just be `if checkCertsLocation()`

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -335,3 +340,7 @@ var createNewExecAgentConfigFile = createNewConfigFile
 func createNewConfigFile(config, configFilePath string) error {
 	return ioutil.WriteFile(configFilePath, []byte(config), filePerm)
 }
+
+func checkCertsLocation() bool {
```
sharanyad, line 344: nit: could rename this to `certsExists()`

agent/engine/docker_task_engine.go
```diff
@@ -1133,15 +1133,21 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
 		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+			herr = &apierrors.HostConfigError{Msg: err.Error()}
+		}
+		if herr != nil {
+			seelog.Warnf("Exec Agent initialization failed : %v . Continuing to start container without enabling exec feature.", herr)
+			container.UpdateManagedAgentByName(execcmd.ExecuteCommandAgentName, apicontainer.ManagedAgentState{
+				Status: apicontainerstatus.ManagedAgentStopped,
+				Reason: herr.Error(),
```
sharanyad, line 1149: please add a test case for this

agent/engine/docker_task_engine.go
```diff
@@ -1133,15 +1133,21 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+			herr = &apierrors.HostConfigError{Msg: err.Error()}
 		}
 		err = engine.execCmdMgr.InitializeContainer(tId, container, hostConfig)
 		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+			herr = &apierrors.HostConfigError{Msg: err.Error()}
+		}
+		if herr != nil {
+			seelog.Warnf("Exec Agent initialization failed : %v . Continuing to start container without enabling exec feature.", herr)
```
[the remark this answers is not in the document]
ubhattacharjya: Good point. Added a check to see if exec agent init failed in RestartExecAgent and StartExecAgent methods. I did not disable exec for that container because in delete task, we do some cleaning if the task is exec enabled.

agent/engine/execcmd/manager.go
```diff
@@ -115,3 +116,12 @@ func IsExecEnabledContainer(container *apicontainer.Container) bool {
 	_, ok := container.GetManagedAgentByName(ExecuteCommandAgentName)
 	return ok
 }
+
+// HasExecAgentInitFailed returns if exec feature has been disabled for a container
```
cyastella, line 120: don't think this needs to use negated logic, personally I think it's clearer to use something like this:

```go
func ExecAgentInitSucceeded(ma) bool {
    // check
}
```

and the caller of `ExecAgentInitSucceeded` can get `ma.Reason` if this checker returns `false`

agent/engine/docker_task_engine.go
```diff
@@ -1133,15 +1133,21 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
 
 	if execcmd.IsExecEnabledContainer(container) {
+		var herr error
 		tId, err := task.GetID()
 		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+			herr = &apierrors.HostConfigError{Msg: err.Error()}
 		}
 		err = engine.execCmdMgr.InitializeContainer(tId, container, hostConfig)
```
cyastella, line 1141: `err` and `herr` could be overridden here, maybe add an `else` clause here?

agent/engine/execcmd/manager.go
ubhattacharjya: I kept the name HasExecAgentInitFailed because I am checking exactly that in this method. That is why, I thought this will be more apt. But let me know if you feel that I should change the name

agent/engine/docker_task_engine.go
```diff
@@ -1133,15 +1133,22 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+				herr = &apierrors.HostConfigError{Msg: err.Error()}
+			}
 		}
-		err = engine.execCmdMgr.InitializeContainer(tId, container, hostConfig)
-		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+		if herr != nil {
+			seelog.Warnf("Exec Agent initialization: %v . Continuing to start container without enabling exec feature.", herr)
+			container.UpdateManagedAgentByName(execcmd.ExecuteCommandAgentName, apicontainer.ManagedAgentState{
```
sharanyad, line 1148: can you add a TODO to emit managed agent state change event. this is dependent on https://github.com/aws/amazon-ecs-agent/pull/2758

agent/engine/execcmd/manager.go
```diff
@@ -115,3 +116,12 @@ func IsExecEnabledContainer(container *apicontainer.Container) bool {
...
+// because of exec agent initialization failure
+func HasExecAgentInitFailed(ma apicontainer.ManagedAgent) (string, bool) {
+	if ma.ID == "" && ma.Status == apicontainerstatus.ManagedAgentStopped {
```
sharanyad, line 123: this is probably not the right check. If restart fails after many retries we move the status to stopped, which this method returns true for, but we may want to restart again. 
one option here could be to introduce a bool flag to explicitly indicate exec init has failed.

ubhattacharjya: Thought about that. If the the ID is "", then the exec agent is not initialized. I am checking that as well. It is only after successful initialization, we will give an uuid as its id. That is why a "" Id will indicate that exec agent is not initialized. Let me know if you think otherwise

sharanyad: ah, got it. Could you add this as a comment? (in case this logic changes in the future)

agent/engine/execcmd/manager.go
cyastella: sounds good! other than the name, it would still make it easier to understand by returning only the result (`bool`) instead of a tuple of `(string, bool)`; consider the [single responsibility principle](https://en.wikipedia.org/wiki/Single-responsibility_principle)...

agent/engine/execcmd/manager.go
ubhattacharjya: sure

agent/engine/execcmd/manager.go
ubhattacharjya: Changed it.

cyastella (approved):

agent/engine/docker_task_engine.go
```diff
@@ -1133,15 +1133,23 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 	}
 
 	if execcmd.IsExecEnabledContainer(container) {
-		tId, err := task.GetID()
+		var herr error
+		tID, err := task.GetID()
 		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+			herr = &apierrors.HostConfigError{Msg: err.Error()}
```
angelcar, line 1139: I don't think there's a need for this error to be `apierrors.HostConfigError` since we're not adding it to metadata anymore. Let's just use a plain onld `error`

agent/engine/execcmd/manager.go
angelcar, line 120: returns true?

agent/engine/execcmd/manager.go
```diff
@@ -115,3 +116,12 @@ func IsExecEnabledContainer(container *apicontainer.Container) bool {
...
 	return ok
 }
+
+// HasExecAgentInitFailed returns if exec feature has been disabled for a container
+// because of exec agent initialization failure. This method checks if the exec agent
+// is stopped and its ID is an empty string. The ID can only be an empty string
+// if the initialization of the exec agent fails. It is only after successful exec
+// agent initialization, we assign a uuid to the ID
+func HasExecAgentInitFailed(ma apicontainer.ManagedAgent) bool {
+	return ma.ID == "" && ma.Status == apicontainerstatus.ManagedAgentStopped
```
angelcar, line 126: This is going to be tricky if the assumptions change in the future. I'd prefer to add a boolean flag to `ManagedAgentState` and set it explicitly if initialization of exec agent fails

agent/engine/execcmd/manager_start_linux.go
```diff
@@ -103,6 +107,10 @@ func (m *manager) StartAgent(ctx context.Context, client dockerapi.DockerClient,
 		return nil
 	}
 	ma, _ := container.GetManagedAgentByName(ExecuteCommandAgentName)
+	if HasExecAgentInitFailed(ma) {
```
angelcar, line 110: hmm, sounds to me that we should not call this method if init failed. This is going to be confusing in the logs.

agent/engine/docker_task_engine.go
```diff
@@ -287,7 +287,9 @@ func (engine *DockerTaskEngine) monitorExecAgentProcesses(ctx context.Context) {
 		}
 		for _, c := range task.Containers {
 			if execcmd.IsExecEnabledContainer(c) {
-				go engine.monitorExecAgentRunning(ctx, mTask, c)
+				if ma, _ := c.GetManagedAgentByName(execcmd.ExecuteCommandAgentName); !ma.InitFailed {
```
angelcar, line 290: I'm afraid we need to do the check of `ma.InitFailed` inside of the `engine.monitorExecAgentRunning` function. The reason is that most likely, the agent hasn't been initialized yet when this code is executed. 

After looking at this more closely, it seems that the best approach would be to move the call `go engine.monitorExecAgentRunning(ctx, mTask, c)` after we [initialize the container with no error](https://github.com/aws/amazon-ecs-agent/blob/feature/ecs_exec/agent/engine/docker_task_engine.go#L1145)

Sorry I didn't catch this earlier :(

agent/engine/docker_task_engine.go
```diff
@@ -1133,15 +1135,18 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+		if err == nil {
+			err = engine.execCmdMgr.InitializeContainer(tID, container, hostConfig)
 		}
-		err = engine.execCmdMgr.InitializeContainer(tId, container, hostConfig)
 		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+			// TODO: Send a managed agent state change event for this
+			seelog.Warnf("Exec Agent initialization: %v . Continuing to start container without enabling exec feature.", err)
+			container.UpdateManagedAgentByName(execcmd.ExecuteCommandAgentName, apicontainer.ManagedAgentState{
```
angelcar, line 1145: I think it'd be better to update the managed agent inside of `execCmdMgr.InitializeContainer`. That way all the initialization logic remains in one place.

Since InitializeContainer has many error paths, I think we would benefit from golang's named return parameters, so that we invoke `UpdateManagedAgentByName` only once.

e.g.

```golang
func (m *manager) InitializeContainer(taskId string, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig) (rErr error) {
	defer func() {
                // If this function returns error, update managed agent accordingly 
		if rErr != nil {
			container.UpdateManagedAgentByName(execcmd.ExecuteCommandAgentName, apicontainer.ManagedAgentState{
				InitFailed: true,
				Status:     apicontainerstatus.ManagedAgentStopped,
				Reason:     err.Error(),
			})
		}
	}()
	if tID == "" {
		return errors.New("Task id cannot be blank")
	}
	.
	.
	.
}
```

ubhattacharjya: Thanks for the idea. I changed it.

sharanyad (approved):
once all tests pass

agent/engine/docker_task_engine.go
```diff
@@ -1133,15 +1135,13 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+		tID, err := task.GetID()
+		if err == nil {
+			err = engine.execCmdMgr.InitializeContainer(tID, container, hostConfig)
 		}
-		err = engine.execCmdMgr.InitializeContainer(tId, container, hostConfig)
 		if err != nil {
-			herr := &apierrors.HostConfigError{Msg: err.Error()}
-			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
+			// TODO: Send a managed agent state change event for this
+			seelog.Warnf("Exec Agent initialization: %v . Continuing to start container without enabling exec feature.", err)
```
fierlion, line 1144: why the change here to not return an error when we can't retrieve the task ID?
will the following not work?
```
if execcmd.IsExecEnabledContainer(container) {
		tId, err := task.GetID()
		if err != nil {
			herr := &apierrors.HostConfigError{Msg: err.Error()}
			return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(herr)}
	        }
		err = engine.execCmdMgr.InitializeContainer(tId, container, hostConfig)
		if err != nil {
			// TODO: Send a managed agent state change event for this
			seelog.Warnf("Exec Agent initialization: %v . Continuing to start container without enabling exec feature.", err)
...
```

fierlion (commented):
also:
> I have tested even if I remove any of the deps folder, the container starts up anyway.

can you add more detail about your manual testing?  did you see any exec related mounts in `docker inspect`?

agent/engine/docker_task_engine.go
ubhattacharjya: makes sense. Changed it.

ubhattacharjya:
> also:
> 
> > I have tested even if I remove any of the deps folder, the container starts up anyway.
> 
> can you add more detail about your manual testing? did you see any exec related mounts in `docker inspect`?

No, There are no exec related mounts in the container that I start with a task if I delete the deps folder

fierlion (approved):

angelcar (approved):

PR #2771, opened 2020-12-17
check at least 1 ssm version is valid instead of checking all versions

cyastella:
… to advertise exec-enable.

### Summary

Previously we check that all ssm version should be valid and when user add a new ssm version folder which is empty, the exec-capability doesn't shows up.  

This PR solve the case when customer try to add another ssm version in the `/var/lib/ecs/deps/execute-command/bin/`.  We already have a valid ssm version, with this new folder added,  the exec-capability should still exists. 

### Implementation details

Add a new check to find in the dependencies map that we have a least one valid ssm version exists. When we add a new ssm version folder (maybe empty), the exec attribute should still shows up and available to use. 
### Testing

```
manual test:
aws ecs describe-container-instances  . . .:
capabilities:
{
...
"name": "ecs.capability.execute-command" 
...
}

previously:
exec-attribute  is not there 

currently:
exec-attribute still shows up

sudo mkdir `new version of ssm`
aws ecs describe-container-instances  . . .:
{
...
"name": "ecs.capability.execute-command" 
...
}

```
### Licensing

### DIFF
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -14,6 +14,7 @@
 package app
 
 import (
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -396,7 +397,7 @@ func (agent *ecsAgent) appendExecCapabilities(capabilities []*ecs.Attribute) ([]
 	if len(binDependencies) < 1 {
 		return capabilities, nil
 	}
-	if exists, err := dependenciesExist(binDependencies); err != nil || !exists {
+	if exists, err := checkAnyValidDependency(binDependencies); err != nil || !exists {
 		return capabilities, err
 	}
 
@@ -432,6 +433,31 @@ func dependenciesExist(dependencies map[string][]string) (bool, error) {
 		}
 	}
 	return true, nil
+}
+
+func checkAnyValidDependency(dependencies map[string][]string) (bool, error) {
+	var validDependencies = 0
+	for directory, files := range dependencies {
+		filesValid := true
+		if exists, err := pathExists(directory, true); err != nil || !exists {
+			continue
+		}
+
+		for _, filename := range files {
+			path := filepath.Join(directory, filename)
+			if exists, err := pathExists(path, false); err != nil || !exists {
+				filesValid = false
+			}
+		}
+
+		if filesValid {
+			validDependencies++
+		}
+	}
+	if validDependencies >= 1 {
+		return true, nil
+	}
+	return false, fmt.Errorf("no valid dependencies")
 }
 
 func defaultPathExists(path string, shouldBeDirectory bool) (bool, error) {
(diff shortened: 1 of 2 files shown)


ubhattacharjya (approved):

agent/app/agent_capability.go
```diff
@@ -434,6 +435,27 @@ func dependenciesExist(dependencies map[string][]string) (bool, error) {
...
+	var validDependencies = 0
+	for directory, files := range dependencies {
+		if exists, err := pathExists(directory, true); err != nil || !exists {
+			continue
+		}
+
+		for _, filename := range files {
+			path := filepath.Join(directory, filename)
+			if exists, err := pathExists(path, false); err != nil || !exists {
+				continue
```
sharanyad, line 448: looks like with this continue, the below `validDependencies` will always be incremented even when binaries are not present. you probably want to break and increment only when all binaries are present.

agent/app/agent_capability.go
```diff
@@ -434,6 +435,27 @@ func dependenciesExist(dependencies map[string][]string) (bool, error) {
...
+			}
+		}
+		validDependencies+=1
+	}
+	if validDependencies >= 1 {
+		return true, nil
+	}
```
fierlion, line 455: I think we should actually be checking that all dependencies are valid.  So if we have more than one ssm agent directory, we should check that each is valid.  I'm assuming that having more than one ssm agent would mean that I have tasks in the instance that are dependent on specific different versions.

fierlion: update -- 
to echo Sharanya's comment:
"since we fallback to using valid binaries version to mount to containers, i think its fine to check for at least one valid dep folder"

agent/app/agent_capability.go
cyastella: good point, will change it.

fierlion (approved):

agent/app/agent_capability.go
```diff
@@ -434,6 +435,32 @@ func dependenciesExist(dependencies map[string][]string) (bool, error) {
 	return true, nil
 }
 
+func checkAnyValidDependency(dependencies map[string][]string) (bool, error) {
+	var validDependencies = 0
+	for directory, files := range dependencies {
+		directoryValid := true
+		filesValid := true
+		if exists, err := pathExists(directory, true); err != nil || !exists {
+			directoryValid = false
```
sharanyad, line 444: this could just be `continue` like previously(since binaries won't exist if dir doesn't), below flag is good enough.

agent/app/agent_capability_test.go
```diff
@@ -815,6 +815,12 @@ func TestCapabilitiesExecuteCommand(t *testing.T) {
 			getSubDirectories:        func(path string) ([]string, error) { return []string{"3.0.236.0"}, nil },
 			shouldHaveExecCapability: true,
 		},
+		{
+			name:                     "execute-command capability should be added if have valid ssm version exists",
+			pathExists:               func(path string, shouldBeDirectory bool) (bool, error) { return true, nil },
```
sharanyad, line 820: could you also add a test case where dir exists, but have invalid binaries and exec capability is false?

cyastella: yeah, sure, will add it!

PR #2779, opened 2020-12-24
Emit ExecCommandAgent stop event when container stops

angelcar:
### Summary

For MVP, ExecCommand Agent will stop only when the parent container stops. This PR addresses the use case of emitting the proper managed agent stop event when parent container stops.

### Implementation details

Sending a , `api.ManagedAgentStateChange` event to the `mtask.stateChangeEvents` channel. Updating UTs accordingly. 

### Testing

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -20,6 +20,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/engine/execcmd"
+
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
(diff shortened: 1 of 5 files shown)


agent/engine/task_manager.go
```diff
@@ -421,6 +423,9 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	// Container has progressed its status if we reach here. Make sure to save it to database.
 	defer mtask.engine.saveContainerData(container)
 
+	// if this is an execute command enabled STOPPED container event we should emit a managedAgent event
+	mtask.handleManagedAgentStoppedTransition(event.Status, container, execcmd.ExecuteCommandAgentName)
```
sharanyad, line 427: can we call this after checking if container status is stopped and exec enabled ? would prefer not logging if container is not stopped each time.

sharanyad (commented):
have we also tested the scenario when stop task is called manually(rather than container exiting by itself)?

agent/engine/task_manager.go
angelcar: Agreed. Changed.

angelcar:
> have we also tested the scenario when stop task is called manually(rather than container exiting by itself)?

Yes, all container stopped events come from docker itself, so no matter if container died or user stopped manually, the flow is docker-> docker-client -> engine -> taskManager.

fierlion (approved):

agent/engine/task_manager.go
fierlion: The idea of a generic handleManagedAgentStoppedTransition is that all managed agents will likely have some stop container handling required; it would be nice to be able to say `for agent in agents: handleManagedAgentStoppedTransition(agent)`.

I was hoping to insulate agent specific conditional logic to the function itself.

sharanyad (approved):

PR #2813, opened 2021-02-23
container stats: ignore context.Canceled errors

sparrc:
we often receive context.Canceled errors when a container exits during
docker stats collection.

there is a benign race condition where if we process the error first
before processing that the context is "Done", then we will log this
canceled error as a warning message here:
https://github.com/aws/amazon-ecs-agent/blob/5be7aa08bed215a557f48c16d8201ad3db59a9be/agent/stats/container.go#L118-L122

this change ignores these context.Canceled errors so that we don't log
them. This will eliminate log messages that look like this when a
container exits:

level=warn time=2020-12-25T07:51:33Z msg="Error encountered processing metrics stream from docker, this may affect cloudwatch metric accuracy: DockerGoClient: Unable to decode stats for container REDACTED: context canceled" module=container.go

### Testing

Tested by running a large number of tasks with lots of container and task state changes and exits. Before this change I was seeing the log message around 20% of the time that containers exited. After the change I do not see this message.

enhancement: eliminate benign docker stats "context canceled" warning messages from logs

### Licensing

### DIFF
--- a/agent/stats/container.go
+++ b/agent/stats/container.go
@@ -118,8 +118,13 @@ func (container *StatsContainer) processStatsStream() error {
 		case <-container.ctx.Done():
 			return nil
 		case err := <-errC:
-			seelog.Warnf("Error encountered processing metrics stream from docker, this may affect cloudwatch metric accuracy: %s", err)
-			returnError = true
+			select {
+			case <-container.ctx.Done():
+				// ignore error when container.ctx.Done()
+			default:
+				seelog.Warnf("Error encountered processing metrics stream from docker, this may affect cloudwatch metric accuracy: %s", err)
+				returnError = true
+			}
 		case rawStat, ok := <-dockerStats:
 			if !ok {
 				if returnError {


sharanyad (approved):

mythri-garaga (dismissed):

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1393,6 +1393,10 @@ func (dg *dockerGoClient) Stats(ctx context.Context, id string, inactivityTimeou
 			stream := true
 			resp, err = client.ContainerStats(subCtx, id, stream)
 			if err != nil {
+				if errors.Is(err, context.Canceled) {
```
angelcar, line 1396: I don't think the client is the right place to ignore this error. 
Can we still return it and ignore it where this method is invoked instead?

sparrc: i guess we can have the client wrapping this error with `%w`....and then we could call this at a higher level, I will try it

angelcar: That works. Alternatively you could just use the plain error when it's  `context.Canceled`
```golang
if errors.Is(err, context.Canceled) {
    errC <- err
    return
}
```

angelcar (approved):

sharanyad (approved):
:shipit:

PR #2824, opened 2021-03-02
Fix bug where pause container was not always cleaned up

angelcar:
### Summary

This fixes a bug that can prevent the pause container network to be cleaned up when ECS Agent is restarted due to `dockerd` restarts. Failing to clean the pause container network might cause issues for future tasks to start.

When `dockerd` is restarted, all running containers (along with ECS Agent) are stopped. After `dockerd` and ECS Agent restart successfully, the latter proceeds to stop all the tasks that were running previous to the restart since it realizes the respective containers are no longer running. When this happens, the pause container network is not cleaned up since the workflow is different to the normal operations. 

### Implementation details

Do a check after the managed tasks is stopped to see if the pause container still needs to be cleaned up. Most of the times the pause container has already been cleaned up by the time the managed task stops, but as mentioned above, there are scenarios where this might not happen. 

Since now we are invoking container cleanup in more than one place (previously it only happened in `stopContainer` method), the operation was made idempotent by storing a flag in the container to indicate whether that container was torn down. 
For now that flag is only used for the pause container, but might be useful for other containers that need explicit teardown in the future.

### Testing

There are a number of unit and integration tests that check for proper pause container cleanup. Those tests should still pass since the previous logic is not modified and a few new assertions are added to check if the new tear down flag was set.

In addition, manually tested that pause container network is cleaned under normal circumstances as well as after docker restarts.

Fix bug where pause container was not always cleaned up
### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -1417,6 +1417,30 @@ func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task, 
 	}
 }
 
+// checkTearDownPauseContainer idempotently tears down the pause container network when the pause container's known
+//or desired status is stopped.
+func (engine *DockerTaskEngine) checkTearDownPauseContainer(task *apitask.Task) {
+	if !task.IsNetworkModeAWSVPC() {
+		return
+	}
+	for _, container := range task.Containers {
+		// Cleanup the pause container network namespace before stop the container
+		if container.Type == apicontainer.ContainerCNIPause && !container.IsContainerTornDown() {
+			// Clean up if the pause container has stopped or will stop
+			if container.KnownTerminal() || container.DesiredTerminal() {
+				err := engine.cleanupPauseContainerNetwork(task, container)
+				if err != nil {
+					seelog.Errorf("Task engine [%s]: unable to cleanup pause container network namespace: %v", task.Arn, err)
+				} else {
+					container.SetContainerTornDown(true)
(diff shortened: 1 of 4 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1417,6 +1417,27 @@ func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task,
 	}
 }
 
+// checkTearDownPauseContainer idempotently tears down the pause container network when the pause container's known
+//or desired status is stopped.
+func (engine *DockerTaskEngine) checkTearDownPauseContainer(task *apitask.Task) {
+	for _, container := range task.Containers {
```
sharanyad, line 1423: instead of iterating through all containers, can we check if task is awsvpc enabled (GetENI != nil) and return quickly?

angelcar: is it time or readability that would be improved? In terms of time, there can be a max of 10 containers so we would gain a few nanoseconds :p

sharanyad: as a practice, just like to return quickly wherever possible :) also, containers count could be increased in the future

angelcar: Fair enough. changed.

sharanyad (approved):

mythri-garaga (approved):

shubham2892 (approved):

PR #2833, opened 2021-03-16
Modify Windows OS family types to support platform version in task definitions

alxche2021:
### Summary

This is currently an internal change of generating the correct Windows OS family types, which will be reflected to front end and PV support in task definitions once other changes are made.

### Implementation details

A shared API GetOSFamilyType() is defined in config package to generate Windows OS family types strings. It does not have impact on Linux by keeping its original behavior.

### Testing

Windows binary has been built for the changes. It is being tested on ECS clusters with all supported OS family types. The ECS service is running fine and the log info shows the expected Windows OS family type strings.

New tests cover the changes: No

Feature - Modify Windows OS family types to support platform version in task definitions.

### Licensing

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -325,10 +325,12 @@ func validateRegisteredAttributes(expectedAttributes, actualAttributes []*ecs.At
 }
 
 func (client *APIECSClient) getAdditionalAttributes() []*ecs.Attribute {
+	osFamilyStr := config.GetOSFamilyType()
+	seelog.Infof("Server OSFamily string: %s", osFamilyStr)
 	return []*ecs.Attribute{
 		{
 			Name:  aws.String("ecs.os-type"),
-			Value: aws.String(config.OSType),
+			Value: aws.String(osFamilyStr),
 		},
 	}
 }
--- a/agent/config/parse_windows.go
+++ b/agent/config/parse_windows.go
@@ -16,15 +16,15 @@
 package config
 
 import (
+	"fmt"
 	"os"
 	"strings"
 	"syscall"
 	"unsafe"
 
-	"golang.org/x/sys/windows/registry"
-
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/cihub/seelog"
+	"golang.org/x/sys/windows/registry"
 )
 
 const (
@@ -32,6 +32,16 @@ const (
 	// domain join check validation. This is useful for integration and
 	// functional-tests but should not be set for any non-test use-case.
 	envSkipDomainJoinCheck = "ZZZ_SKIP_DOMAIN_JOIN_CHECK_NOT_SUPPORTED_IN_PRODUCTION"
+
+	// The following constants are used to generate Windows OS family types
+	ReleaseId2004SAC          = "2004"
+	ReleaseId1909SAC          = "1909"
+	Windows_Server_2019       = "Windows Server 2019"
+	Windows_Server_2016       = "Windows Server 2016"
+	Windows_Server_DataCenter = "Windows Server Datacenter"
+	InstallationTypeCore      = "Server Core"
+	InstallationTypeFull      = "Server"
+	UnknownOSType             = "unknown"
 )
 
 // parseGMSACapability is used to determine if gMSA support can be enabled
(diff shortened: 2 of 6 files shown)


alxche2021:
I ll squash my commits into one after all tests are confirmed passing all test cases

agent/config/parse_windows.go
```diff
@@ -113,3 +123,61 @@ var isWindows2016 = func() (bool, error) {
...
+		return "FULL", nil
+	} else if strings.Compare(installationType, InstallationTypeCore) == 0 {
+		return "CORE", nil
+	} else {
+		err := seelog.Errorf("Invalid Installation type")
+		return "", err
+	}
+}
+
+func GetOSFamilyType() string {
```
mythri-garaga, line 138: Can we add unit test for this?

alxche2021: Sure Mythri, I am working on it.

agent/config/parse_windows.go
```diff
@@ -113,3 +123,61 @@ var isWindows2016 = func() (bool, error) {
...
+		return UnknownOSType
+	}
+	releaseId, _, err := key.GetStringValue("ReleaseId")
+	if err != nil {
+		seelog.Errorf("Unable to read registry key, ReleaseId: %v", err)
+		return UnknownOSType
+	}
+
+	if strings.HasPrefix(productName, Windows_Server_2019) {
+		osTypeFormat = fmt.Sprintf(osTypeFormat, "2019", iType)
```
sparrc, line 169: I find it confusing that you are defining this string as a format string (with `%s`), and then overwriting it as a regular string that ultimately gets returned.

You should create a new variable for the first part of this string similar to how you do for `iType`. In these if-statements, only set the variable to that number. As in: 
```
num := "2019"
```

then you can have just a single return like `return fmt.Sprintf("WINDOWS_SERVER_%s_%s", num, iType)`

agent/config/parse_windows.go
```diff
@@ -113,3 +123,61 @@ var isWindows2016 = func() (bool, error) {
...
+	if err != nil {
+		seelog.Errorf("Unable to read registry key, InstallationType: %v", err)
+		return UnknownOSType
+	}
+	iType, err := getInstallationType(installationType)
+	if err != nil {
+		seelog.Errorf("Invalid Installation type found: %v", err)
+		return UnknownOSType
+	}
+	releaseId, _, err := key.GetStringValue("ReleaseId")
```
sparrc, line 162: if  i understand correctly, releaseId is only used if this is a windows server datacenter product? can you move getting of this value into that if-statement?

agent/config/parse_windows.go
```diff
@@ -113,3 +123,61 @@ var isWindows2016 = func() (bool, error) {
...
+	} else if strings.Compare(installationType, InstallationTypeCore) == 0 {
+		return "CORE", nil
+	} else {
+		err := seelog.Errorf("Invalid Installation type")
+		return "", err
+	}
+}
+
+func GetOSFamilyType() string {
+	osTypeFormat := "WINDOWS_SERVER_%s_%s"
```
sparrc, line 139: define this either as a constant in the whole package or not at all

sparrc (changes requested):

agent/api/ecsclient/client.go
```diff
@@ -325,10 +325,12 @@ func validateRegisteredAttributes(expectedAttributes, actualAttributes []*ecs.At
 }
 
 func (client *APIECSClient) getAdditionalAttributes() []*ecs.Attribute {
+	osFamilyStr := config.GetOSFamilyType()
+	seelog.Infof("Server OSFamily string: %s", osFamilyStr)
```
sparrc, line 329: this seems like more of a Debugf statement

agent/config/parse_windows.go
```diff
@@ -16,22 +16,32 @@
...
 const (
 	// envSkipDomainJoinCheck is an environment setting that can be used to skip
 	// domain join check validation. This is useful for integration and
 	// functional-tests but should not be set for any non-test use-case.
 	envSkipDomainJoinCheck = "ZZZ_SKIP_DOMAIN_JOIN_CHECK_NOT_SUPPORTED_IN_PRODUCTION"
+
+	// The following constants are used to generate Windows OS family types
+	ReleaseId2004SAC          = "2004"
+	ReleaseId1909SAC          = "1909"
+	Windows_Server_2019       = "Windows Server 2019"
```
sparrc, line 39: can you remove the underscores and change these to WindowsServer2019 for consistency?

agent/config/parse_windows.go
sparrc, line 138: It seems if any step of this GetOSFamilyType function fails it seems we will just return `unknown`, should we maybe return `windows` (ie, `config.OSType`) to keep the behavior the same as before?

sparrc:
closing as this change was implemented in https://github.com/aws/amazon-ecs-agent/pull/2859

PR #2834, opened 2021-03-18
Add ecs-agent container healthcheck localhost ip override env var

fierlion:
### Summary
This change allows IP configuration for the ecs-agent container's docker healthcheck localhost IP to address a specific case where the customer has overridden the `localhost` ip explicitly.  

### Implementation details
Currently, as the agent container is launched in host network mode, when the localhost inside the scratch-based agent container does not resolve to that the same IP, the agent container healthcheck will always return 'unhealthy'.
This change allows the localhost IP to be explicitly set to either the host localhost ip (usually 127.0.0.1) or otherwise based on DNS resolution requirements.

### Testing
I built the agent and tested the following manually:
start agent the environment variable unset -- agent container became healthy
set ECS_AGENT_HEALTHCHECK_HOST=localhost -- agent container became healthy
set ECS_AGENT_HEALTHCHECK_HOST=127.0.0.1 -- agent container became healthy
set ECS_AGENT_HEALTHCHECK_HOST=80.80.80.80 -- agent container became unhealthy.

Add configurable agent healthcheck localhost ip env var.

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ additional details on each available environment variable.
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
 | `ECS_CGROUP_PATH` | `/sys/fs/cgroup` | The root cgroup path that is expected by the ECS agent. This is the path that accessible from the agent mount. | `/sys/fs/cgroup` | Not applicable |
 | `ECS_CGROUP_CPU_PERIOD` | `10ms` | CGroups CPU period for task level limits. This value should be between 8ms to 100ms | `100ms` | Not applicable |
+| `ECS_AGENT_HEALTHCHECK_HOST` | `localhost` | Override for the ecs-agent container's healthcheck localhost ip address| `localhost` | `localhost` |
 | `ECS_ENABLE_CPU_UNBOUNDED_WINDOWS_WORKAROUND` | `true` | When `true`, ECS will allow CPU unbounded(CPU=`0`) tasks to run along with CPU bounded tasks in Windows. | Not applicable | `false` |
 | `ECS_ENABLE_MEMORY_UNBOUNDED_WINDOWS_WORKAROUND` | `true` | When `true`, ECS will ignore the memory reservation parameter (soft limit) to run along with memory bounded tasks in Windows. To run a memory unbounded task, omit the memory hard limit and set any memory reservation, it will be ignored. | Not applicable | `false` |
 | `ECS_TASK_METADATA_RPS_LIMIT` | `100,150` | Comma separated integer values for steady state and burst throttle limits for task metadata endpoint | `40,60` | `40,60` |
--- a/agent/app/run.go
+++ b/agent/app/run.go
@@ -14,6 +14,8 @@
 package app
 
 import (
+	"fmt"
+	"os"
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/app/args"
@@ -22,6 +24,10 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/version"
 	"github.com/aws/aws-sdk-go/aws"
 	log "github.com/cihub/seelog"
+)
+
+const (
+	ECS_AGENT_HEALTHCHECK_HOST_ENV_VAR = "ECS_AGENT_HEALTHCHECK_HOST"
 )
 
 // Run runs the ECS Agent App. It returns an exit code, which is used by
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -172,6 +172,7 @@ additional details on each available environment variable.
 | `ECS_ENABLE_TASK_CPU_MEM_LIMIT` | `true` | Whether to enable task-level cpu and memory limits | `true` | `false` |
 | `ECS_CGROUP_PATH` | `/sys/fs/cgroup` | The root cgroup path that is expected by the ECS agent. This is the path that accessible from the agent mount. | `/sys/fs/cgroup` | Not applicable |
 | `ECS_CGROUP_CPU_PERIOD` | `10ms` | CGroups CPU period for task level limits. This value should be between 8ms to 100ms | `100ms` | Not applicable |
+| `AGENT_HEALTHCHECK_LOCALHOST_OVERRIDE` | `127.0.0.1` | Override for the ecs-agent container's healthcheck localhost ip address| `localhost` | `localhost` |
```
sparrc, line 175: This env var should be prepended with `ECS_`

and nit that I would prefer it less verbose, maybe just `ECS_AGENT_HEALTHCHECK_HOST`? with default value `localhost`

agent/app/run.go
```diff
@@ -43,7 +49,14 @@ func Run(arguments []string) int {
 		// timeout of 30s. This is so that we can catch any http timeout and log the
 		// issue within agent logs.
 		// see https://docs.docker.com/engine/reference/builder/#healthcheck
-		return runHealthcheck("http://localhost:51678/v1/metadata", time.Second*25)
+		// if AGENT_HEALTHCHECK_LOCALHOST_OVERRIDE env var is set, it will override
+		// the healthcheck's localhost ip address inside the ecs-agent container
+		localhost := "localhost"
+		if localhostOverride := os.Getenv(HEALTHCHECK_LOCALHOST_OVERRIDE_ENV_VAR); localhostOverride != "" {
```
sparrc, line 55: we should probably parse this in the config package for consistency? https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config.go

sparrc:
there are some existing unit tests for the healtcheck, does it make sense to add a test or two for this? https://github.com/aws/amazon-ecs-agent/blob/90ccb1d4313eb853691465197d324cd76404b1cc/agent/app/healthcheck_test.go

agent/app/run.go
fierlion: I didn't see a neat way to stuff config in this top level of the application.  I'll take another look.

agent/app/run.go
```diff
@@ -24,6 +26,10 @@ import (
 	log "github.com/cihub/seelog"
 )
 
+const (
```
chienhanlin, line 29: Question: Should we define and maintain all environment variables in the [config.go](https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config.go#L498) ? or this does not apply to healthcheck workflow? Thanks!

agent/app/run.go
```diff
@@ -43,7 +49,14 @@ func Run(arguments []string) int {
 		// timeout of 30s. This is so that we can catch any http timeout and log the
 		// issue within agent logs.
 		// see https://docs.docker.com/engine/reference/builder/#healthcheck
-		return runHealthcheck("http://localhost:51678/v1/metadata", time.Second*25)
+		// if ECS_AGENT_HEALTHCHECK_HOST env var is set, it will override
+		// the healthcheck's localhost ip address inside the ecs-agent container
+		localhost := "localhost"
+		if localhostOverride := os.Getenv(ECS_AGENT_HEALTHCHECK_HOST_ENV_VAR); localhostOverride != "" {
```
angelcar, line 55: perhaps it's more convenient to do  
```golang
if localhostOverride, present := os.LookupEnv(ECS_AGENT_HEALTHCHECK_HOST_ENV_VAR); present { ...
```  
That way we would ignore the case where the env variable is set but accidentally left blank

sparrc (approved):

agent/app/run.go
sparrc: i had wondered the same, see https://github.com/aws/amazon-ecs-agent/pull/2834#discussion_r597138412

fierlion: https://github.com/aws/amazon-ecs-agent/blob/a8f4fcb8be4e08830cbfb51096a403f2f263235e/agent/logger/log.go#L28
I see this more like the logger -- we don't use the config (and its variables) in the logger either.

angelcar (approved):

agent/app/run.go
```diff
@@ -43,7 +49,14 @@ func Run(arguments []string) int {
 		// timeout of 30s. This is so that we can catch any http timeout and log the
 		// issue within agent logs.
 		// see https://docs.docker.com/engine/reference/builder/#healthcheck
-		return runHealthcheck("http://localhost:51678/v1/metadata", time.Second*25)
+		// if ECS_AGENT_HEALTHCHECK_HOST env var is set, it will override
+		// the healthcheck's localhost ip address inside the ecs-agent container
+		localhost := "localhost"
+		if localhostOverride, ok := os.LookupEnv(ECS_AGENT_HEALTHCHECK_HOST_ENV_VAR); ok {
```
sparrc, line 55: sorry for the back and forth on this, but I dont think this is correct.....`ok` will be true if the env var was set and empty, is that what we want? for users to be able to set the host to a blank string?

see https://play.golang.org/p/MLN0QyG_ray

cc @angelcar

angelcar: Ah, you're right cam. I had that backwards. Sorry @fierlion, I gave you bad advice with `LookupEnv` :(

fierlion: reverted.  see latest update.
nice catch Cam.

sparrc (approved):

angelcar (approved):

PR #2837, opened 2021-03-25
ECS agent to acknowledge server heartbeat messages

LiangChiAmzn:
### Summary

This PR adds the support for ECS agent to acknowledge server heartbeat messages. This would help both ECS agent and ECS backend server to detect stale socket connections, then take proactive actions to handle it.

### Implementation details

ECS server will soon start to actively clean up stale connections if ECS agent is not acknowledging heartbeat messages. Since such change is backward-incompatible, a new `protocolVersion` is introduced for ECS agent to control the behavior from client side when establishing the WebSocket connection.

If the version is set as 2, then the ECS agent must acknowledge all server heartbeats. Null protocol version is treated as 1.

### Testing

```
make test
make release
```

Manually tested the change by building the agent, connect to ECS, then inspect ECS backend logs:

- Positive case: connection is healthy when protocol version = 2 and heartbeats are being acked
- Negative case: ECS will cut the connection if protocol version = 2 but heartbeat acks are missing

Enhancement - Add support for ECS agent to acknowledge server heartbeat messages

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -24,7 +24,6 @@ import (
 	"time"
 
 	acsclient "github.com/aws/amazon-ecs-agent/agent/acs/client"
-	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	updater "github.com/aws/amazon-ecs-agent/agent/acs/update_handler"
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/config"
@@ -65,6 +64,10 @@ const (
 	// credentials for all tasks on establishing the connection
 	sendCredentialsURLParameterName = "sendCredentials"
 	inactiveInstanceExceptionPrefix = "InactiveInstanceException:"
+	// ACS protocol version spec:
+	// 1: default protocol version
+	// 2: ACS will proactively close the connection when heartbeat acks are missing
+	acsProtocolVersion = 2
 )
 
 // Session defines an interface for handler's long-lived connection with ACS.
@@ -332,8 +335,13 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 
 	client.AddRequestHandler(payloadHandler.handlerFunc())
 
-	// Ignore heartbeat messages; anyMessageHandler gets 'em
-	client.AddRequestHandler(func(*ecsacs.HeartbeatMessage) {})
+	// Add HeartbeatHandler to acknowledge ACS heartbeats
+	heartbeatHandler := newHeartbeatHandler(acsSession.ctx, client)
+	defer heartbeatHandler.clearAcks()
+	heartbeatHandler.start()
+	defer heartbeatHandler.stop()
+
+	client.AddRequestHandler(heartbeatHandler.handlerFunc())
 
 	updater.AddAgentUpdateHandlers(client, cfg, acsSession.state, acsSession.dataClient, acsSession.taskEngine)
 
@@ -454,6 +462,7 @@ func acsWsURL(endpoint, cluster, containerInstanceArn string, taskEngine engine.
 	query.Set("agentHash", version.GitHashString())
 	query.Set("agentVersion", version.Version)
 	query.Set("seqNum", "1")
+	query.Set("protocolVersion", strconv.Itoa(acsProtocolVersion))
 	if dockerVersion, err := taskEngine.Version(); err == nil {
 		query.Set("dockerVersion", "DockerVersion: "+dockerVersion)
 	}
--- a/agent/acs/handler/heartbeat_handler.go
+++ b/agent/acs/handler/heartbeat_handler.go
@@ -1,0 +1,118 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package handler
+
+import (
+	"context"
+
+	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/amazon-ecs-agent/agent/wsclient"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/cihub/seelog"
+)
+
+// heartbeatHandler handles heartbeat messages from ACS
+type heartbeatHandler struct {
+	heartbeatMessageBuffer    chan *ecsacs.HeartbeatMessage
+	heartbeatAckMessageBuffer chan *ecsacs.HeartbeatAckRequest
+	ctx                       context.Context
+	cancel                    context.CancelFunc
+	acsClient                 wsclient.ClientServer
+}
+
+// newHeartbeatHandler returns an instance of the heartbeatHandler struct
+func newHeartbeatHandler(ctx context.Context,
+	acsClient wsclient.ClientServer) heartbeatHandler {
+
+	// Create a cancelable context from the parent context
+	derivedContext, cancel := context.WithCancel(ctx)
+	return heartbeatHandler{
+		heartbeatMessageBuffer:    make(chan *ecsacs.HeartbeatMessage),
+		heartbeatAckMessageBuffer: make(chan *ecsacs.HeartbeatAckRequest),
+		ctx:                       derivedContext,
+		cancel:                    cancel,
+		acsClient:                 acsClient,
+	}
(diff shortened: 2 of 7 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -332,8 +335,13 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 
 	client.AddRequestHandler(payloadHandler.handlerFunc())
 
-	// Ignore heartbeat messages; anyMessageHandler gets 'em
-	client.AddRequestHandler(func(*ecsacs.HeartbeatMessage) {})
+	// Add HeartbeatHandler to acknowledge ACS heartbeats
+	heartbeatHandler := newHeartbeatHandler(acsSession.ctx, client)
+	defer heartbeatHandler.clearAcks()
```
sparrc, line 340: defer statements for a FILO stack in golang, so just wanted to verify here, do we want `.stop()` to be called _before_ `.clearAcks()`?

Secondly, there doesn't appear to be any use-case to clearAcks besides cleanup, so could we just integrate it into the stop() function to make this more fool-proof?

agent/acs/handler/heartbeat_handler.go
```diff
@@ -0,0 +1,119 @@
...
+				seelog.Warnf("Unable to handle heartbeat message [%s]: %v", message.String(), err)
+			}
+		case <-heartbeatHandler.ctx.Done():
+			return
+		}
+	}
+}
+
+func (heartbeatHandler *heartbeatHandler) handleSingleHeartbeatMessage(message *ecsacs.HeartbeatMessage) error {
+	seelog.Tracef("Received server heartbeat message: %s", message.MessageId)
```
sparrc, line 76: We very rarely use trace logging on our team, so would probably prefer all of these trace log messages changed to `Debugf`

agent/acs/handler/heartbeat_handler.go
```diff
@@ -0,0 +1,119 @@
...
+		case ack := <-heartbeatHandler.heartbeatAckMessageBuffer:
+			heartbeatHandler.sendSingleHeartbeatAck(ack)
+		case <-heartbeatHandler.ctx.Done():
+			return
+		}
+	}
+}
+
+func (heartbeatHandler *heartbeatHandler) sendSingleHeartbeatAck(ack *ecsacs.HeartbeatAckRequest) {
+	err := heartbeatHandler.acsClient.MakeRequest(ack)
```
sparrc, line 98: would there be any benign errors here that we should be aware of that could give us "red herring" warning messages in the logs? like if we receive a heartbeat with the same message ID twice and then ack it twice?

sparrc: or perhaps if we released this and started acking heartbeats in a region where this feature wasn't deployed service-side?

agent/acs/handler/acs_handler.go
LiangChiAmzn: For the order --- yes, it is intended for `.stop()` (context cancellation) to be invoked before `.clearAcks()` (channel draining) to ensure no message is left.

This behavior is mostly borrowed from `refreshCredsHandler`, `taskManifestHandler`, and `payloadHandler`. (https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/acs_handler.go#L316-L331 ) I thought about refactoring it to be more concise but ended up keeping it as-is so we have consistent implementation across ACS handlers. Maybe in future we will have another layer of interface abstraction?

Let me know which route is preferred.

agent/acs/handler/heartbeat_handler.go
LiangChiAmzn: Right, I am aware that ECS agent in fact does not really support `TRACE` level logging, so the purpose of this is more towards clarifying the intention, and to show something when running tests.

I'm a bit concerned about the log size if we make them `DEBUG`, as heartbeats will happen *very* often and the likelihood of these logs to be useful is low.

agent/acs/handler/heartbeat_handler.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		seelog.Warnf("Error acknowledging server heartbeat, message id: %s, error: %v", aws.StringValue(ack.MessageId), err)
+	}
+	seelog.Tracef("Acknowledging server heartbeat message: %s", ack.MessageId)
```
sparrc, line 102: This will print a bit funny since MessageId is a pointer, but in the trace/debug messages we could probably just stringify the whole object anyways, ie just print `ack.String()`

agent/acs/handler/heartbeat_handler.go
sparrc: same comment as below, probably should change this to `message.String()`

agent/acs/handler/heartbeat_handler.go
```diff
@@ -0,0 +1,119 @@
...
+	go heartbeatHandler.handleHeartbeatMessage()
+	go heartbeatHandler.sendHeartbeatAck()
+}
+
+func (heartbeatHandler *heartbeatHandler) handleHeartbeatMessage() {
+	for {
+		select {
+		case message := <-heartbeatHandler.heartbeatMessageBuffer:
+			if err := heartbeatHandler.handleSingleHeartbeatMessage(message); err != nil {
+				seelog.Warnf("Unable to handle heartbeat message [%s]: %v", message.String(), err)
```
sparrc, line 67: nit: errors in golang always have a builtin string marshaller, so when printing errors `%v` is not necessary, it can be `%s`

agent/acs/handler/heartbeat_handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}).Times(1)
+
+	handler := newHeartbeatHandler(ctx, mockWsClient)
+	go handler.sendHeartbeatAck()
+
+	handler.handleSingleHeartbeatMessage(heartbeatReceived)
+
+	// wait till we get an ack from heartbeatAckMessageBuffer
+	select {
+	case <-ctx.Done():
```
sparrc, line 89: nit: select is not necessary here, this can just be `<-ctx.Done()`

agent/acs/handler/heartbeat_handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	if !reflect.DeepEqual(heartbeatAckExpected, heartbeatAckSent) {
```
sparrc, line 92: There's a package we commonly use (and very common in golang in general) called `github.com/stretchr/testify/require`, which I think should work here.

I believe this if-statement could just be `require.Equal(t, heartbeatExpected, heartbeatActual)`

The benefit is we get better and more consistent error messages across unit tests in different packages.

agent/acs/handler/heartbeat_handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+const (
+	heartbeatMessageId = "heartbeatMessageId"
+	heartbeatHealthy   = true
+)
+
+func TestAckHeartbeatMessage(t *testing.T) {
+	heartbeatReceived := &ecsacs.HeartbeatMessage{
+		MessageId: aws.String(heartbeatMessageId),
+		Healthy:   aws.Bool(heartbeatHealthy),
```
sparrc, line 37: the Healthy field is just ignored correct? could you add a unit test with Healthy==false just to verify the same behavior?

agent/acs/handler/acs_handler.go
sparrc: okay, makes sense, since it's consistent just leave it as-is

agent/acs/handler/heartbeat_handler.go
sparrc: > Right, I am aware that ECS agent in fact does not really support `TRACE` level logging, so the purpose of this is more towards clarifying the intention, and to show something when running tests.
> 
> I'm a bit concerned about the log size if we make them `DEBUG`, as heartbeats will happen _very_ often and the likelihood of these logs to be useful is low.

I would say just make them comments then?

agent/acs/handler/heartbeat_handler_test.go
LiangChiAmzn: Sure, will add in next rev

agent/acs/handler/heartbeat_handler_test.go
LiangChiAmzn: Thanks! Will fix along with other comments in next revision

fenxiong (commented):
lgtm other than cam's comment

agent/acs/handler/heartbeat_handler.go
LiangChiAmzn: Cleaned up these trace logging into comments

agent/acs/handler/heartbeat_handler_test.go
LiangChiAmzn: Fixed in next rev

agent/acs/handler/heartbeat_handler.go
LiangChiAmzn: Fixed

sparrc (approved):
lgtm assuming tests pass

sparrc (approved):

fenxiong (approved):

PR #2635, opened 2020-09-16
Ecs exec cmd integ test

angelcar:
### Summary

Add integration test for exec command agent.

### Implementation details

A new, scratch-based docker image was added for this integ test in order to be consistent with design doc which states exec command agent should be compatible with scratch-based images. The image has a pre-baked `sleep` command built using golang, as part of the docker build process.

The actual integ test only validates that the expected bindmounts are present in the container. This is done by invoking container inspect docker API and verifying that the expected bind mounts are in the response. This means that ECS Agent has added the required bind mounts for exec command agent to run.

For now I am skipping the test until the actual exec command binaries are present in the host.

note: Some minor changes are also included in this PR, such as removal of unused arguments.

### Testing

New tests cover the changes: yes

Ran `make run-integ-tests` and manually verified that the container was created and had the expected bind mounts. 
After that, I added the t.Skip statement to the test.

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #2855, opened 2021-04-27
Filter out ipv6 port forwards when an ipv6 default route doesn't exist

lattwood:
edit: Didn't intend to open this yet.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -476,19 +476,23 @@ func (client *APIECSClient) buildContainerStateChangePayload(change api.Containe
 		exitCode := int64(aws.IntValue(change.ExitCode))
 		statechange.ExitCode = aws.Int64(exitCode)
 	}
-	networkBindings := make([]*ecs.NetworkBinding, len(change.PortBindings))
-	for i, binding := range change.PortBindings {
+
+	networkBindings := []*ecs.NetworkBinding{}
+	for _, binding := range change.PortBindings {
+		if binding.BindIP == "::" && client.config.MissingIPv6DefaultRoute {
+			continue
+		}
 		hostPort := int64(binding.HostPort)
 		containerPort := int64(binding.ContainerPort)
 		bindIP := binding.BindIP
 		protocol := binding.Protocol.String()
 
-		networkBindings[i] = &ecs.NetworkBinding{
+		networkBindings = append(networkBindings, &ecs.NetworkBinding{
 			BindIP:        aws.String(bindIP),
 			ContainerPort: aws.Int64(containerPort),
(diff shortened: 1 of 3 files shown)


sparrc:
Hello, could you please describe this PR in more detail, and what is the use case for this?

If I understand correctly, is this happening because you are using a task in `awsvpc` network mode, and your VPC has ipv6 enabled but you have disabled or deleted the default ipv6 route via the route table of that VPC? If this is the case, could you explain what is the use-case of having a VPC with ipv6 enabled but all of the ipv6 routes deleted?

lattwood:
@sparrc this is to prevent duplicate task registration in the ECS API, as ECS registers the task for every port forward. On the latest version of docker they return ipv6 port forwards even when there’s no ipv6 connectivity on the host.

In our case we have ipv6 disabled at the vpc layer, but it still tries to contact the host on the ipv6 port but the ipv4 address.

lattwood:
This happens with the nat networking mode as well.

sparrc:
by "latest version of docker" which version do you mean? could you provide some steps to reproduce?

lattwood:
20.10.6

That will allow you to reproduce the double registration issue.

Docker using a different ipv4 and ipv6 port on the host when exposing the
container port I don’t have reproduction for. It would happen on 1/60 ECS
nodes.

sparrc:
@lattwood can you confirm that this is the same issue as https://github.com/moby/moby/issues/42288? Looks like docker is going to fix this in 20.10.7 so we probably want to hold off on any ECS workarounds, see PR in progress https://github.com/moby/moby/pull/42322

lattwood:
It’s related in the sense that the change caused multiple regressions.

chienhanlin:
Hello @lattwood , 
A follow-up for the fix mentioned in @sparrc 's comment. The PR [fix port forwarding with ipv6.disable=1 #2635](https://github.com/moby/libnetwork/pull/2635) was merged to moby/moby repo, and is included in the [20.10.7 Docker Engine release notes](https://docs.docker.com/engine/release-notes/#networking). Could you take a look and see if the issue you have encountered can be resolved? If not, could you provide summary, implementation details and testing on this PR? Thank you.

chienhanlin:
Close this PR as the fix is available since ECS Agent 1.55.3 https://github.com/aws/amazon-ecs-agent/releases/tag/v1.55.3.
  
In addition, we are working on a related issue https://github.com/aws/amazon-ecs-agent/issues/3096 to differentiate between IPV4/IPV6 Port Mappings.

Please feel free to create PRs or issues if you would like to share any suggestion with us. Thanks.

PR #2859, opened 2021-04-29
Changes to advertise updated OSType while registering ECS Windows instances 

sandeepindraganti:
### Summary

Advertise OSFamilyType  while registering the ECS instances with the cluster. Below are the supported OSTypes

Supported OS Types:

<enumValue value="WINDOWS_SERVER_2019_FULL" name="WINDOWS_SERVER_2019_FULL"/>
  | <enumValue value="WINDOWS_SERVER_2019_CORE" name="WINDOWS_SERVER_2019_CORE"/>
  | <enumValue value="WINDOWS_SERVER_2016_FULL" name="WINDOWS_SERVER_2016_FULL"/>
  | <enumValue value="WINDOWS_SERVER_2004_CORE" name="WINDOWS_SERVER_2004_CORE"/>

### Implementation details
Added a new implementation GetOperatingSystemFamilyType() which reads the windows registry and construct the relevant windows OS type. For Linux implementation it returns "Linux" as OS Type which is currently being returned. Linux behavior is unaltered by this PR

### Testing
Unit tests written for the new changes. 
Existing tests were executed 
Custom agent built and verified.

Note for external contributors:
`make test` and `make run-integ-tests` can run in a Linux development
environment like your laptop.  `go test -timeout=30s ./agent/...` and
`.\scripts\run-integ.tests.ps1` can run in a Windows development environment
like your laptop.  Please ensure unit and integration tests pass (on at least
one platform) before opening the pull request.
Once you open the pull request, there will be 14 automatic test checks on the bottom
of the pull request, please make sure they all pass before you merge it. You can
use `bot/test` label to rerun the automatic tests multiple times.
-->

New tests cover the changes: 
Yes. 

"Register Windows ECS Instances using specific OSFamilyType "

### Licensing


agent/config/parse_windows.go
```diff
@@ -32,8 +33,108 @@ const (
...
+	ReleaseId1909SAC        = "1909"
+	WindowsServer2019       = "Windows Server 2019"
+	WindowsServer2016       = "Windows Server 2016"
+	WindowsServerDataCenter = "Windows Server Datacenter"
+	InstallationTypeCore    = "Server Core"
+	InstallationTypeFull    = "Server"
+	UnknownOSType           = "unknown"
+	osTypeFormat            = "WINDOWS_SERVER_%s_%s"
+	ecsWinRegistryRootKey   = registry.LOCAL_MACHINE
+	ecsWinRegistryRootPath  = `SOFTWARE\Microsoft\Windows NT\CurrentVersion`
```
rawahars, line 47: nit: Should we consider moving these re-usable constants to config/const_windows.go?

rawahars (changes requested):
Looks good to me!

agent/config/parse_windows.go
```diff
@@ -32,8 +34,108 @@ const (
...
+	if ReleaseId2004SAC == releaseId {
+		return ReleaseId2004SAC, nil
+	} else if ReleaseId1909SAC == releaseId {
+		return ReleaseId1909SAC, nil
+	}
+	err := seelog.Errorf("Unsupported ReleaseId:%s for Windows LTSC Release", releaseId)
+	return "", err
+}
+
+func GetOperatingSystemFamily() string {
```
fenxiong, line 92: please add comments to exported method `// GetOperatingSystemFamily ... ` https://github.com/golang/go/wiki/CodeReviewComments#doc-comments.  Same for parse_unsupported.go and parse_linux.go

agent/config/parse_windows.go
fenxiong: +1, also please keep the variables unexported if they are not used in other packages.

agent/config/parse_windows_test.go
```diff
@@ -16,10 +16,14 @@
 package config
 
 import (
+	"github.com/golang/mock/gomock"
```
fenxiong, line 19: nit: to align with convention in other files, please keep 3rd parties import below built in import (i.e. "os" and "testing"). example: https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config.go#L16-L31

agent/config/parse_windows.go
```diff
@@ -32,8 +34,108 @@ const (
...
+	winRegistry = mockRegistry
+}
+
+func getInstallationType(installationType string) (string, error) {
+	if InstallationTypeFull == installationType {
+		return "FULL", nil
+	} else if InstallationTypeCore == installationType {
+		return "CORE", nil
+	} else {
+		err := seelog.Errorf("Unsupported Installation type:%s", installationType)
```
fenxiong, line 67: https://github.com/golang/go/wiki/CodeReviewComments#error-strings error message should not be capitalized. since you use the one returned from seelog.Errorf, you will need to make sure the log message start with lower case. same for line 78 and line 88 in this file

agent/config/parse_windows.go
```diff
@@ -32,8 +34,108 @@ const (
...
+	}
+	err := seelog.Errorf("Unsupported ReleaseId:%s for Windows LTSC Release", releaseId)
+	return "", err
+}
+
+func GetOperatingSystemFamily() string {
+	key, err := winRegistry.OpenKey(ecsWinRegistryRootKey, ecsWinRegistryRootPath, registry.QUERY_VALUE)
+	if err != nil {
+		seelog.Errorf("Unable to open Windows registry key to determine Windows version: %v", err)
+		return UnknownOSType
```
fenxiong, line 96: echoing https://github.com/aws/amazon-ecs-agent/pull/2833#discussion_r615158983. it seems to me that returning the old os type makes more sense than returning unknown, unless there's a specific reason not to?

agent/config/parse_windows.go
sandeepindraganti: Done. Looks like this package has a separate file for constants which I missed looking at it. However in general  Its a good practice defining constants in the same file as they are being used rather defining some where else.  Thanks on the export , I didn't know the convention in go.

agent/config/parse_windows.go
sandeepindraganti: Done, Thank you

agent/config/parse_windows.go
sandeepindraganti: Well , even "windows" is unknown & not supported  with regards to ECS in this case. But I agree it sounds a little better . I changed it to windows

agent/config/parse_windows.go
sandeepindraganti: Sure thing. Thank you!

agent/config/parse_windows.go
```diff
@@ -16,23 +16,108 @@
...
+func init() {
+	winRegistry = dependencies.StdRegistry{}
+}
+
+func setWinRegistry(mockRegistry dependencies.WindowsRegistry) {
+	winRegistry = mockRegistry
+}
+
+func getInstallationType(installationType string) (string, error) {
+	if installationTypeFull == installationType {
```
sharanyad, line 43: could you use switch case instead of if else here?

agent/config/parse_windows.go
```diff
@@ -16,23 +16,108 @@
...
+		return "FULL", nil
+	} else if installationTypeCore == installationType {
+		return "CORE", nil
+	} else {
+		err := seelog.Errorf("unsupported Installation type:%s", installationType)
```
sharanyad, line 48: this should probably be `errors.Errof`

agent/config/parse_windows.go
```diff
@@ -16,23 +16,108 @@
...
+		return "2019", nil
+	} else if strings.HasPrefix(productName, windowsServer2016) {
+		return "2016", nil
+	}
+	err := seelog.Errorf("unsupported productName:%s for Windows SAC Release", productName)
+	return "", err
+}
+
+func getReleaseIdForLTSCReleases(releaseId string) (string, error) {
+	if releaseId2004SAC == releaseId {
```
sharanyad, line 64: switch case

agent/config/parse_windows.go
```diff
@@ -16,23 +16,108 @@
...
+		return releaseId2004SAC, nil
+	} else if releaseId1909SAC == releaseId {
+		return releaseId1909SAC, nil
+	}
+	err := seelog.Errorf("unsupported ReleaseId:%s for Windows LTSC Release", releaseId)
+	return "", err
```
sharanyad, line 70: `return "", errors.Errof("unsupported ReleaseId:%s for Windows LTSC Release", releaseId)`

agent/config/parse_windows.go
```diff
@@ -16,23 +16,108 @@
...
+	key, err := winRegistry.OpenKey(ecsWinRegistryRootKey, ecsWinRegistryRootPath, registry.QUERY_VALUE)
+	if err != nil {
+		seelog.Errorf("unable to open Windows registry key to determine Windows version: %v", err)
+		return unsupportedWindowsOS
+	}
+	defer key.Close()
+
+	productName, _, err := key.GetStringValue("ProductName")
+	if err != nil {
+		seelog.Errorf("unable to read registry key, ProductName: %v", err)
```
sharanyad, line 85: please use upper case for the first alphabet of the log.

agent/config/parse_windows.go
```diff
@@ -16,23 +16,108 @@
...
+		return unsupportedWindowsOS
+	}
+	installationType, _, err := key.GetStringValue("InstallationType")
```
sharanyad, line 88: nit: these could be changed to: 
```
if installationType, _, err := key.GetStringValue("InstallationType"); err != nil {

}
```

agent/config/const_windows.go
```diff
@@ -15,5 +15,23 @@
...
+	envSkipDomainJoinCheck  = "ZZZ_SKIP_DOMAIN_JOIN_CHECK_NOT_SUPPORTED_IN_PRODUCTION"
+	releaseId2004SAC        = "2004"
+	releaseId1909SAC        = "1909"
+	windowsServer2019       = "Windows Server 2019"
+	windowsServer2016       = "Windows Server 2016"
+	windowsServerDataCenter = "Windows Server Datacenter"
+	installationTypeCore    = "Server Core"
+	installationTypeFull    = "Server"
+	unsupportedWindowsOS    = "windows"
+	osTypeFormat            = "WINDOWS_SERVER_%s_%s"
```
sharanyad, line 34: is there any reason why upper case should be used for instance capability?

agent/config/parse_windows_test.go
```diff
@@ -50,3 +55,143 @@ func TestParseFSxWindowsFileServerCapability(t *testing.T) {
 
 	assert.False(t, parseFSxWindowsFileServerCapability())
 }
+
+func TestGetOperatingSystemFamilyForWS2019Core(t *testing.T) {
```
angelcar, line 59: To avoid mostly repeated code, you can use test tables. Here's an example of how to do it:

https://github.com/aws/amazon-ecs-agent/blob/2874131a5f3d1c0908e0f2ebad62e48b8ff5850f/agent/engine/execcmd/agent_version_test.go#L146

agent/config/parse_linux.go
```diff
@@ -22,3 +22,8 @@ func parseGMSACapability() bool {
 func parseFSxWindowsFileServerCapability() bool {
 	return false
 }
+
+// GetOperatingSystemFamily() returns "linux" as operating system family for linux based ecs instances
+func GetOperatingSystemFamily() string {
```
angelcar, line 27: why do you need this new function? you can access the `OSType` const directly

agent/config/parse_windows.go
sandeepindraganti: Done . Looks like switch case is possible with strings in go lang.

agent/config/parse_windows.go
sandeepindraganti: done.

agent/config/parse_windows.go
sandeepindraganti: done

agent/config/parse_windows.go
sandeepindraganti: done

agent/config/parse_windows.go
sandeepindraganti: done

agent/config/const_windows.go
sandeepindraganti: I believe you were asking about OSType, If that is correct the ECS Control Plane uses Upper case , so just used the same here for consistency

agent/config/parse_windows.go
sandeepindraganti: Sure, It looks the current way more readable to me , will keep the same if you don't mind. This looks a little hard to understand but seems advanced  go style :)

agent/config/const_windows.go
sharanyad: could you add an example of what the OSType capability looks like for windows?

agent/config/parse_linux.go
sandeepindraganti: As per conversation, the runtime fails without this. Its needed

agent/config/parse_windows_test.go
sandeepindraganti: I tried to make tests better by creating a function for common code. However the test tables doesn't seem a good idea here , I tried it and it looks bad to me. Here are the following reasons.
1.  I see the go test tables are nothing but an array of structure defining inputs , I don't see this mechanism reducing the LOC at least in this case.
2. The tests here have different expectations and not possible to use just one for loop to run all tests. I can put multiple but it didn't look great to me.
3. These are different test cases , its hard to debug individual tests without commenting others.

I see your concern regarding redundant code and I agree , to address it I have moved most of the redundant code to functions.  If i use task tables it doesn't really good to me ( i tried it Here is the link if you wanted to look: https://paste.amazon.com/show/sssandee/1620245472

angelcar (approved):

rawahars (approved):

PR #2860, opened 2021-04-30
EC2 Task Networking on Windows: Retrieving ENI's name on the instance.

rawahars:
### Summary

ENI's link name on the instance is a mandatory parameter for the CNI plugin.
Therefore, we retrieve the name on the first plugin invocation for a task. The name is stored as part of ENI and is used in subsequent calls.

### Implementation details

A field `LinkName` has been added in the eni object. 
Additionally a method `GetLinkName` , has been added which returns the link name if it is readily available. If not, then it scans all the interfaces on the instance to find the one which has same MAC address as the ENI.

### Testing

Unit tests have been added/updated and a custom agent binary was tested.

New tests cover the changes: 
Yes

Added ENI's link name field.

### Licensing

### DIFF
--- a/agent/api/eni/eni.go
+++ b/agent/api/eni/eni.go
@@ -17,6 +17,10 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"sync"
+
+	"github.com/aws/amazon-ecs-agent/agent/eni/netwrapper"
+	"github.com/cihub/seelog"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	"github.com/aws/aws-sdk-go/aws"
@@ -27,6 +31,9 @@ import (
 type ENI struct {
 	// ID is the id of eni
 	ID string `json:"ec2Id"`
+	// LinkName is the name of the ENI on the instance.
+	// Currently, this field is being used only for Windows and is used during task networking setup.
+	LinkName string
 	// MacAddress is the mac address of the eni
 	MacAddress string
 	// IPV4Addresses is the ipv4 address associated with the eni
@@ -55,6 +62,11 @@ type ENI struct {
 	ipv4SubnetPrefixLength string
 	ipv4SubnetCIDRBlock    string
 	ipv6SubnetCIDRBlock    string
+
+	// net is a wrapper around golang's net package.
+	net netwrapper.NetWrapper
+	// guard protects access to fields of this struct.
+	guard sync.RWMutex
 }
 
 // InterfaceVlanProperties contains information for an interface that
@@ -181,6 +193,30 @@ func (eni *ENI) GetHostname() string {
 	return eni.PrivateDNSName
 }
 
+// GetLinkName returns the name of the ENI on the instance.
+func (eni *ENI) GetLinkName() string {
+	eni.guard.Lock()
+	defer eni.guard.Unlock()
+
+	if eni.LinkName == "" && eni.net != nil {
+		// Find all interfaces on the instance.
+		ifaces, err := eni.net.GetAllNetworkInterfaces()
+		if err != nil {
+			seelog.Errorf("Failed to find link name: %v.", err)
+			return ""
+		}
+		// Iterate over the list and find the interface with the ENI's MAC address.
+		for _, iface := range ifaces {
+			if strings.EqualFold(eni.MacAddress, iface.HardwareAddr.String()) {
+				eni.LinkName = iface.Name
+				break
+			}
+		}
+	}
+
+	return eni.LinkName
+}
+
 // IsStandardENI returns true if the ENI is a standard/regular ENI. That is, if it
 // has its association protocol as standard. To be backwards compatible, if the
 // association protocol is not set for an ENI, it's considered a standard ENI as well.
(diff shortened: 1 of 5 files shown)


agent/api/eni/eni.go
```diff
@@ -181,6 +184,25 @@ func (eni *ENI) GetHostname() string {
...
+		// Find all interfaces on the instance.
+		ifaces, err := net.Interfaces()
+		if err != nil {
+			return ""
+		}
+		// Iterate over the list and find the interface with the ENI's MAC address.
+		for _, iface := range ifaces {
+			if strings.EqualFold(eni.MacAddress, iface.HardwareAddr.String()) {
+				eni.LinkName = iface.Name
+			}
```
sandeepindraganti, line 199: Do we want to handle  the case if we don't find a match with the MAC address?

rawahars: No, we will let the caller of the method decide on the way to handle that case. In any such situation, the LinkName would be empty.

One such case would be when this method is called immediately after the ENI attachment. During the time taken by the ENI to become available, this method might return empty string. In such a situation, the caller can retry with backoff. 

Also, this is consistent with the remaining methods of the eni package wherein default values are returned.

agent/api/eni/eni.go
```diff
@@ -181,6 +184,25 @@ func (eni *ENI) GetHostname() string {
 	return eni.PrivateDNSName
 }
 
+// GetLinkName returns the name of the ENI on the instance.
+func (eni *ENI) GetLinkName() string {
+	if eni.LinkName == "" {
+		// Find all interfaces on the instance.
+		ifaces, err := net.Interfaces()
+		if err != nil {
```
angelcar, line 192: We need to log this error

agent/api/eni/eni.go
```diff
@@ -181,6 +184,25 @@ func (eni *ENI) GetHostname() string {
...
+			return ""
+		}
+		// Iterate over the list and find the interface with the ENI's MAC address.
+		for _, iface := range ifaces {
+			if strings.EqualFold(eni.MacAddress, iface.HardwareAddr.String()) {
+				eni.LinkName = iface.Name
```
angelcar, line 198: nit: add `break`

agent/api/eni/eni_test.go
```diff
@@ -129,6 +130,19 @@ func TestGetSubnetGatewayIPv4Address(t *testing.T) {
 	assert.Equal(t, ipv4Gw, testENI.GetSubnetGatewayIPv4Address())
 }
 
+// TestGetLinkName tests the retrieval of ENIs name on the instance.
+func TestGetLinkName(t *testing.T) {
+	ifaces, err := net.Interfaces()
```
angelcar, line 135: We should probably be mocking the call to this method

agent/engine/docker_task_engine_windows_test.go
```diff
@@ -504,34 +504,32 @@ func TestPauseContainerHappyPath(t *testing.T) {
 	dockerClient.EXPECT().CreateContainer(gomock.Any(), gomock.Any(), gomock.Any(),
 		gomock.Any(), gomock.Any()).Return(dockerapi.DockerContainerMetadata{DockerID: sleepContainerID2})
 
-	gomock.InOrder(
```
angelcar, line 507: Any reason to remove the `InOrder` statement?

agent/api/eni/eni.go
```diff
@@ -181,6 +184,25 @@ func (eni *ENI) GetHostname() string {
 	return eni.PrivateDNSName
 }
 
+// GetLinkName returns the name of the ENI on the instance.
+func (eni *ENI) GetLinkName() string {
```
angelcar, line 188: If this method is going to be called concurrently, we might need to protect this with a mutex.

agent/engine/docker_task_engine_windows_test.go
rawahars: These API calls are called in parallel and it's not necessary that the order is followed.
In the Linux counterpart of this test, we do not have any ordering.

agent/api/eni/eni.go
sharanyad: +1, please add RW lock here

agent/api/eni/eni_test.go
sharanyad: @rawahars we may want to create an interface to call `net.Interfaces` so its easier to mock in unit tests. This is an example of using an interface for Docker APIs https://github.com/aws/amazon-ecs-agent/blob/847d9e193325f9b5cb0cad834f02dbd3842129c1/agent/dockerclient/dockerapi/docker_client.go#L102 so we could mock these methods instead of the original ones.

rawahars: We had created such a wrapper earlier. We can reuse that here.

angelcar: For this particular case I think it's simpler to just declare an non-exported variable (alias) that we can change in the tests, 
as opposed to creating an interface and having to inject it and all that.

```golang
// Declaration at package level
var netInterfaces = net.Interfaces

// Invocation 
netInterfaces()

// mocking for tests
netInterfaces = func() ([]Interface, error) { ... }
```

sharanyad (approved):

angelcar (approved):

sandeepindraganti (approved):

PR #2861, opened 2021-04-30
Add retries to get ec2 instance-id

mythri-garaga:
### Summary

Currently when agent fails to get instance id from EC2 metadata service, agent overwrites InstanceID in state file with ''. This PR includes changes to call instance metadata service at least 3 times to get instance id before returning an empty string. 

```
level=critical time=2021-03-04T00:00:05Z msg="Error saving state before final shutdown: Multiple error:\n\t0: final save: timed out trying to save to disk" module=termination_handler.go
level=info time=2021-03-04T00:00:07Z msg="Loading configuration" module=agent.go
level=info time=2021-03-04T00:00:07Z msg="Amazon ECS agent Version: 1.50.0, Commit: 90ccb1d4" module=agent.go
level=info time=2021-03-04T00:00:07Z msg="Image excluded from cleanup: amazon/amazon-ecs-pause:0.1.0" module=docker_image_manager.go
level=info time=2021-03-04T00:00:07Z msg="Image excluded from cleanup: amazon/amazon-ecs-pause:0.1.0" module=docker_image_manager.go
level=info time=2021-03-04T00:00:07Z msg="Image excluded from cleanup: amazon/amazon-ecs-agent:latest" module=docker_image_manager.go
level=info time=2021-03-04T00:00:07Z msg="Creating root ecs cgroup: /ecs" module=init_linux.go
level=info time=2021-03-04T00:00:07Z msg="Creating cgroup /ecs" module=cgroup_controller_linux.go
level=info time=2021-03-04T00:00:07Z msg="Event stream ContainerChange start listening..." module=eventstream.go
level=warn time=2021-03-04T00:00:07Z msg="Unable to access EC2 Metadata service to determine EC2 ID: RequestError: send request failed\ncaused by: Get \"http://169.254.169.254/latest/meta-data/instance-id\": EOF" module=agent.go
level=warn time=2021-03-04T00:00:07Z msg="Data mismatch; saved InstanceID 'i-xxxxxxxxx' does not match current InstanceID ''. Overwriting old datafile" module=agent.go
level=info time=2021-03-04T00:00:08Z msg="Registering Instance with ECS" module=agent.go
```
### Implementation details

- In `./agent/app/agent.go - Add retries with exponential back off when calling IMDS for ec2 instance id
- In `./agent/app/agent_test.go - Updated tests accordingly

### Testing

`make test` passes

Enhancement - Add retries while retrieving instance-id using EC2 Instance metadata service api

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -54,6 +54,7 @@ import (
 	tcshandler "github.com/aws/amazon-ecs-agent/agent/tcs/handler"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/aws/amazon-ecs-agent/agent/utils/mobypkgwrapper"
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
 	"github.com/aws/amazon-ecs-agent/agent/version"
 	"github.com/aws/aws-sdk-go/aws"
 	aws_credentials "github.com/aws/aws-sdk-go/aws/credentials"
@@ -70,6 +71,14 @@ const (
 
 	vpcIDAttributeName    = "ecs.vpc-id"
 	subnetIDAttributeName = "ecs.subnet-id"
+
+	blackholed = "blackholed"
+
+	instanceIdBackoffMin      = time.Second
+	instanceIdBackoffMax      = time.Second * 5
+	instanceIdBackoffJitter   = 0.2
+	instanceIdBackoffMultiple = 1.3
+	instanceIdMaxRetryCount   = 3
 )
 
 var (
@@ -454,11 +463,21 @@ func (agent *ecsAgent) setClusterInConfig(previousCluster string) error {
 
 // getEC2InstanceID gets the EC2 instance ID from the metadata service
 func (agent *ecsAgent) getEC2InstanceID() string {
-	instanceID, err := agent.ec2MetadataClient.InstanceID()
+	var instanceID string
+	var err error
+	backoff := retry.NewExponentialBackoff(instanceIdBackoffMin, instanceIdBackoffMax, instanceIdBackoffJitter, instanceIdBackoffMultiple)
+	for i := 0; i < instanceIdMaxRetryCount; i++ {
+		instanceID, err = agent.ec2MetadataClient.InstanceID()
+		if err == nil || err.Error() == blackholed {
+			return instanceID
+		}
+		if i < instanceIdMaxRetryCount-1 {
+			time.Sleep(backoff.Duration())
+		}
+	}
 	if err != nil {
 		seelog.Warnf(
 			"Unable to access EC2 Metadata service to determine EC2 ID: %v", err)
-		return ""
 	}
 	return instanceID
 }
--- a/agent/app/agent_test.go
+++ b/agent/app/agent_test.go
@@ -737,6 +737,29 @@ func TestSetClusterInConfig(t *testing.T) {
 	assert.NoError(t, err)
 }
 
+func TestGetEC2InstanceID(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	ec2MetadataClient := mock_ec2.NewMockEC2MetadataClient(ctrl)
+	agent := &ecsAgent{ec2MetadataClient: ec2MetadataClient}
+
+	ec2MetadataClient.EXPECT().InstanceID().Return("", errors.New("error"))
+	ec2MetadataClient.EXPECT().InstanceID().Return(instanceID, nil)
+	assert.Equal(t, "i-123", agent.getEC2InstanceID())
+}
+
+func TestGetEC2InstanceIDBlackholedError(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	ec2MetadataClient := mock_ec2.NewMockEC2MetadataClient(ctrl)
+	agent := &ecsAgent{ec2MetadataClient: ec2MetadataClient}
+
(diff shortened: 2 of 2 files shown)


agent/app/agent.go
```diff
@@ -454,11 +460,19 @@ func (agent *ecsAgent) setClusterInConfig(previousCluster string) error {
 
 // getEC2InstanceID gets the EC2 instance ID from the metadata service
 func (agent *ecsAgent) getEC2InstanceID() string {
-	instanceID, err := agent.ec2MetadataClient.InstanceID()
+	var instanceID string
+	var err error
+	backoff := retry.NewExponentialBackoff(getInstanceIdBackoffMin, getInstanceIdBackoffMax, getInstanceIdBackoffJitter, getInstanceIdBackoffMultiple)
+	for i := 0; i < 3; i++ {
+		instanceID, err = agent.ec2MetadataClient.InstanceID()
```
fenxiong, line 467: could you check if the error message equals "blackholed" https://github.com/aws/amazon-ecs-agent/blob/master/agent/ec2/blackhole_ec2_metadata_client.go#L53 and don't retry in that case? we use the blackholed client when running on-prem and we don't want to retry in that case

agent/app/agent.go
```diff
@@ -454,11 +460,19 @@ func (agent *ecsAgent) setClusterInConfig(previousCluster string) error {
...
+		if err == nil {
+			return instanceID
+		}
+		time.Sleep(backoff.Duration())
```
fenxiong, line 471: there is an extra sleep in last loop. probably need something like this:
```
if i < 2 {
time.Sleep(backoff.Duration())
}
```

agent/app/agent.go
mythri-garaga: Thanks for calling this out! it's fixed now.

agent/app/agent.go
```diff
@@ -70,6 +71,13 @@ const (
 
 	vpcIDAttributeName    = "ecs.vpc-id"
 	subnetIDAttributeName = "ecs.subnet-id"
+
+	blackholed = "blackholed"
+
+	getInstanceIdBackoffMin      = time.Second
```
sharanyad, line 77: nit: could remove `get` prefix from these constants

agent/app/agent.go
```diff
@@ -454,11 +462,21 @@ func (agent *ecsAgent) setClusterInConfig(previousCluster string) error {
 
 // getEC2InstanceID gets the EC2 instance ID from the metadata service
 func (agent *ecsAgent) getEC2InstanceID() string {
-	instanceID, err := agent.ec2MetadataClient.InstanceID()
+	var instanceID string
+	var err error
+	backoff := retry.NewExponentialBackoff(getInstanceIdBackoffMin, getInstanceIdBackoffMax, getInstanceIdBackoffJitter, getInstanceIdBackoffMultiple)
+	for i := 0; i < 3; i++ {
+		instanceID, err = agent.ec2MetadataClient.InstanceID()
```
sharanyad, line 469: do we know the error types returned by this API? are they classified as retriable/non retriable?

mythri-garaga: Observed `404 - Not Found` HTTP error code from the logs
```
level=error time=2021-03-04T00:00:08Z msg="Unable to retrieve Host Instance PublicIPv4 Address: EC2MetadataError: failed to make EC2Metadata request\n\tstatus code: 404, request id: \ncaused by: <?xml version=\"1.0\" encoding=\"iso-8859-1\"?>\n<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Transitional//EN\"\n\t\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd\">\n<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"en\" lang=\"en\">\n <head>\n  <title>404 - Not Found</title>\n </head>\n <body>\n  <h1>404 - Not Found</h1>\n </body>\n</html>\n" module=agent.go
```
And these are the errors can be returned by IMDSv2 API
```
400 - Missing or Invalid Parameters – The PUT request is not valid.

401 - Unauthorized – The GET request uses an invalid token. The recommended action is to generate a new token.

403 - Forbidden – The request is not allowed or the instance metadata service is turned off.

404 - Not Found – If the resource is not available.

Query throttling - when it exceeds rate limiting
```

Do we want to only retry on Query throttling and 404 error?

sharanyad: its ok to keep this as is for now, since number of retries is less too.

sharanyad (dismissed):

sparrc (dismissed):

Realmonia (dismissed):

agent/app/agent.go
```diff
@@ -454,11 +463,21 @@ func (agent *ecsAgent) setClusterInConfig(previousCluster string) error {
...
-	instanceID, err := agent.ec2MetadataClient.InstanceID()
+	var instanceID string
+	var err error
+	backoff := retry.NewExponentialBackoff(instanceIdBackoffMin, instanceIdBackoffMax, instanceIdBackoffJitter, instanceIdBackoffMultiple)
+	for i := 0; i < instanceIdMaxRetryCount; i++ {
+		instanceID, err = agent.ec2MetadataClient.InstanceID()
+		if err == nil || err.Error() == blackholed {
+			return instanceID
+		}
+		if i < 2 {
```
angelcar, line 474: nit: It would be better to say `if i < instanceIdMaxRetryCount-1 {`, that way this code doesn't need to change if we change `instanceIdMaxRetryCount` in the future

mythri-garaga: Thanks for catching this, will update this.

angelcar (approved):

sharanyad (approved):

PR #2862, opened 2021-05-06
EC2 task networking for Windows: Changes for using vpc-eni plugin instead of vpc-shared-eni plugin

rawahars:
### Summary

As per the old design of the feature, we were using vpc-shared-eni for setting up task namespace on Windows instances. 
As per latest design, we are moving from vpc-shared-eni to vpc-eni plugin.

The primary changes in this PR are-
1. Change of name and references of vpc-shared-eni to vpc-eni
2. Change of CNI config to use vpc-eni plugin
3. Separation of log file path for Windows and Linux
4. For windows, before the tasks are started, we will update the task ENI's link name.
5. Using function injection for net.Interfaces in the eni package

### Implementation details

1. The references are renamed and CNI config is updated as per the requirements of the plugin.
2. The logfile path is segregated in the platform files.
3. When the task is being added to the engine, we will update the task eni dependencies. This is done only for Windows.
4. A variable for the net.Interfaces function is used and during testing, we inject that variable with the mock implementations.

### Testing

Using unit tests and a custom agent binary was tested.

New tests cover the changes: 
No new functionality has been added.

Changing CNI plugin for Windows from vpc-shared-eni to vpc-eni.
### Licensing


agent/engine/docker_task_engine.go
```diff
@@ -794,6 +794,9 @@ func (engine *DockerTaskEngine) AddTask(task *apitask.Task) {
 		// This will update the container desired status
 		task.UpdateDesiredStatus()
 
+		// This will update any dependencies for awsvpc network mode before the task is started.
+		engine.updateTaskENIDependencies(task)
```
angelcar, line 798: Feels like this needs to be invoked in `PostUnmarshalTask` since that's where most of the task setup happens.

rawahars: These dependencies should be fulfilled only once. `PostUnmarshalTask` is called for each task payload received by the agent which involves task updates.
Therefore, we are calling this method just before adding the task to the engine.
We have validated that the agent checkpoints all the relevant fields and has updated information on restart.

angelcar (approved):

agent/ecscni/types_windows.go
```diff
@@ -18,40 +18,31 @@ package ecscni
 import "github.com/containernetworking/cni/pkg/types"
 
 const (
-	// ECSVPCSharedENIPluginName is the name of the vpc-shared-eni plugin
-	ECSVPCSharedENIPluginName = "vpc-shared-eni"
-	// ECSVPCSharedENIPluginExecutable is the name of vpc-shared-eni executable
-	ECSVPCSharedENIPluginExecutable = "vpc-shared-eni.exe"
-	// TaskENIBridgeNetworkPrefix is the prefix added to the Task ENI bridge network name by the CNI plugin
-	TaskENIBridgeNetworkPrefix = "task"
+	// ECSVPCENIPluginName is the name of the vpc-eni plugin.
```
sharanyad, line 21: indentation seems off, could you run go lint on these files?

agent/ecscni/types_windows.go
```diff
@@ -18,40 +18,31 @@ package ecscni
...
-	// GatewayIPAddress specifies the IPv4 address of the subnet gateway for the ENI
+	// GatewayIPAddress specifies the IPv4 address of the subnet gateway for the eni.
 	GatewayIPAddress string `json:"gatewayIPAddress"`
-	// IPAddress which needs to be allocated to pause container endpoint
-	IPAddress string `json:"ipAddress"`
-	// TaskENIConfig provides the Task ENI specific information to the plugin
-	// For now, we will use it to signal plugin to delete network on invokation of DEL
-	TaskENIConfig TaskENIConfig `json:"taskENIConfig"`
+	// NoInfraContainer specifies if HCN Namespace is being used for networking setup.
+	NoInfraContainer bool `json:"noInfraContainer"`
```
sharanyad, line 45: the naming is not very intuitive, what is an infra container?

agent/ecscni/plugin_unsupported.go
```diff
@@ -23,6 +23,11 @@ import (
 	"github.com/containernetworking/cni/pkg/types/current"
 )
 
+const (
+	// vpcCNIPluginPath is the path of the cni plugin's log file.
+	vpcCNIPluginPath = "/log/vpc-branch-eni.log"
```
sharanyad, line 28: is this needed?

agent/ecscni/types_windows.go
rawahars: This is one of the parameters required by the plugin in the same name format.
Pause container is the infra container which is used to setup the network namespace.
On Fargate, we do not need a pause container as we can create network namespaces directly.

agent/ecscni/plugin_unsupported.go
rawahars: Yes. The method where this constant is used is in a common file (plugin.go)

sharanyad (approved):

sandeepindraganti (approved):
LGTM

PR #2863, opened 2021-05-07
EC2 task networking for Windows: Enable Task IAM roles/Task metadata

rawahars:
### Summary

The changes in this PR create a new endpoint into the task namespace to enable task iam roles/task metadata.

Additionally, we have added logic for retrying the network setup if it fails in the first attempt.
This is because on Windows, HNS might not be able to bound itself to the adapter in the first instance especially if the call is made immediately after the adapter was attached to the instance.

### Implementation details

The changes were implemented by creating the appropriate methods where required for creating new cni plugin configuration.

The retry logic for setting up namespace involves creating an exponential backoff and using it to make the goroutine sleep before retrying.

### Testing

The changes were tested using a custom agent binary as well as using unit tests.

New tests cover the changes: 
Yes

Enabling task iam roles and task metadata for the task launched using awsvpc network mode in Windows.
### Licensing

### DIFF
--- a/agent/ecscni/plugin_windows.go
+++ b/agent/ecscni/plugin_windows.go
@@ -21,6 +21,8 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
+
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 
 	"github.com/cihub/seelog"
@@ -35,8 +37,28 @@ var (
 		"Amazon", "ECS", "log", "vpc-eni.log")
 )
 
-// setupNS is the called by SetupNS to setup the task namespace by invoking ADD for given CNI configurations
+// setupNS is the called by SetupNS to setup the task namespace by invoking ADD for given CNI configurations.
+// For Windows, we will retry the setup on Windows before conceding error.
 func (client *cniClient) setupNS(ctx context.Context, cfg *Config) (*current.Result, error) {
+	var result *current.Result
+	var err error
+	backoff := retry.NewExponentialBackoff(setupNSBackoffMin, setupNSBackoffMax,
+		setupNSBackoffJitter, setupNSBackoffMultiple)
+
+	for count := 1; count < setupNSMaxRetryCount; count++ {
+		result, err = client.setupNSOnce(ctx, cfg)
+		if err == nil {
+			return result, nil
+		}
+		if count < setupNSMaxRetryCount {
+			time.Sleep(backoff.Duration())
+		}
+	}
+	return nil, err
+}
+
+// setupNSOnce invokes the CNI plugins to setup the task network namespace.
+func (client *cniClient) setupNSOnce(ctx context.Context, cfg *Config) (*current.Result, error) {
 	seelog.Debugf("[ECSCNI] Setting up the container namespace %s", cfg.ContainerID)
 
 	runtimeConfig := libcni.RuntimeConf{
--- a/agent/ecscni/plugin_windows.go
+++ b/agent/ecscni/plugin_windows.go
@@ -19,7 +19,10 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
+
+	cnitypes "github.com/containernetworking/cni/pkg/types"
 
 	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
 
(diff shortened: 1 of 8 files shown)


agent/ecscni/plugin_windows.go
```diff
@@ -52,11 +79,17 @@ func (client *cniClient) setupNS(ctx context.Context, cfg *Config) (*current.Res
...
 			cfg.ContainerID)
 		runtimeConfig.IfName = networkConfig.IfName
-		_, err := client.libcni.AddNetwork(ctx, cniNetworkConfig, &runtimeConfig)
+		result, err := client.libcni.AddNetwork(ctx, cniNetworkConfig, &runtimeConfig)
 		if err != nil {
 			return nil, errors.Wrap(err, "add network failed")
 		}
 
+		// We save the result from ecs-bridge setup invocation of the plugin.
+		if strings.EqualFold(ECSBridgeNetworkName, cniNetworkConfig.Network.Name) &&
```
angelcar, line 88: curious as to why is `ECSBridgeNetworkName` comparison case insensitive whereas `ECSVPCENIPluginExecutable` is not?

agent/ecscni/plugin_windows.go
```diff
@@ -35,10 +40,32 @@ var (
...
+
+	for count := 1; count < setupNSMaxRetryCount; count++ {
+		result, err = client.setupNSOnce(ctx, cfg)
+		if err == nil {
+			return result, nil
+		}
+		if count < setupNSMaxRetryCount {
+			time.Sleep(backoff.Duration())
+		}
+		seelog.Debugf("Namespace setup failed. Retry count is %s.", count)
```
fenxiong, line 59: can you log the error here? also, to conform with convention in this package, please add prefix `[ECSCNI] ` to log in this file

agent/ecscni/plugin_windows.go
```diff
@@ -65,7 +98,20 @@ func (client *cniClient) setupNS(ctx context.Context, cfg *Config) (*current.Res
 
 	seelog.Debugf("[ECSCNI] Completed setting up the container namespace: %s", cfg.ContainerID)
 
-	return nil, nil
+	if _, err := ecsBridgeResult.GetAsVersion(currentCNISpec); err != nil {
+		seelog.Warnf("[ECSCNI] Unable to convert result to spec version %s; error: %v; result is of version: %s",
```
angelcar, line 102: Nit, could you use the structured log wrapper here
e.g: https://github.com/aws/amazon-ecs-agent/blob/2874131a5f3d1c0908e0f2ebad62e48b8ff5850f/agent/engine/task_manager.go#L242-L245

agent/ecscni/plugin_windows.go
```diff
@@ -35,10 +40,32 @@ var (
...
+	var err error
+	backoff := retry.NewExponentialBackoff(setupNSBackoffMin, setupNSBackoffMax,
+		setupNSBackoffJitter, setupNSBackoffMultiple)
+
+	for count := 1; count < setupNSMaxRetryCount; count++ {
+		result, err = client.setupNSOnce(ctx, cfg)
+		if err == nil {
+			return result, nil
+		}
+		if count < setupNSMaxRetryCount {
```
fenxiong, line 56: should this be `count < setupNSMaxRetryCount - 1` assuming you want to skip the sleep in last iteration?

agent/ecscni/plugin_windows.go
```diff
@@ -35,10 +40,32 @@ var (
...
+			time.Sleep(backoff.Duration())
+		}
+		seelog.Debugf("Namespace setup failed. Retry count is %s.", count)
+	}
+	return nil, err
+}
+
+// setupNSOnce invokes the CNI plugins to setup the task network namespace.
+func (client *cniClient) setupNSOnce(ctx context.Context, cfg *Config) (*current.Result, error) {
```
angelcar, line 65: nit: I know this naming pattern is present elsewhere, but  I think `doSetupNS` is a more appropriate name for this method.

agent/ecscni/plugin_windows.go
```diff
@@ -35,10 +40,32 @@ var (
 		"Amazon", "ECS", "log", "vpc-eni.log")
 )
 
-// setupNS is the called by SetupNS to setup the task namespace by invoking ADD for given CNI configurations
+// setupNS is the called by SetupNS to setup the task namespace by invoking ADD for given CNI configurations.
+// For Windows, we will retry the setup on Windows before conceding error.
```
angelcar, line 44: nit: For Windows, we will retry the setup ~~on Windows~~ before conceding error.

agent/ecscni/plugin_windows.go
```diff
@@ -52,11 +79,17 @@ func (client *cniClient) setupNS(ctx context.Context, cfg *Config) (*current.Res
...
-		_, err := client.libcni.AddNetwork(ctx, cniNetworkConfig, &runtimeConfig)
+		result, err := client.libcni.AddNetwork(ctx, cniNetworkConfig, &runtimeConfig)
 		if err != nil {
 			return nil, errors.Wrap(err, "add network failed")
 		}
 
+		// We save the result from ecs-bridge setup invocation of the plugin.
+		if strings.EqualFold(ECSBridgeNetworkName, cniNetworkConfig.Network.Name) &&
+			cniNetworkConfig.Network.Type == ECSVPCENIPluginExecutable {
+			ecsBridgeResult = result
```
angelcar, line 90: This is confusing. Why only storing/returning this particular `result`s? Also, it ultimately doesn't seem to be used when `setupNS` is called: https://github.com/aws/amazon-ecs-agent/pull/2863/files#diff-e3a9b4000263d327ff4083d797e85e84eca3a2695426bfad0f8c2525ad8f63b7R96

agent/ecscni/plugin_windows.go
rawahars: This is because `ECSBridgeNetworkName` is the name of the default network on Windows instances which is to be used as the `ecs-bridge`. The name of this network can be case insensitive.
On the other hand, the plugin executable cannot be case insensitive otherwise it will lead to an error.

agent/ecscni/plugin_windows.go
angelcar: discussed offline, to do this change the interface would need to change to return a slice of results, which is not something we are interested in doing now since changes to the linux side would be required as well.

sandeepindraganti (approved):
Changes Looks good to me . I don't anticipate any of these but worth to verify these things later  (1) The ECS Control plane API doesn't time out due to the additional lag we have added to retry HCS (2)  If ECS make any promises on task launch time to customers , ensure we don't break them.

fenxiong (approved):

PR #2864, opened 2021-05-11
Add test runner (Doctor) and basic docker ping Healthcheck

fierlion:
### Summary
Add healthcheck runner (Doctor) and Healthcheck; update acs heartbeat handler to trigger a *basic* docker ping healthcheck.

### Implementation details
Please note:
This PR is meant to only add the Doctor.  It has a basic docker ping for demo purposes only.  
As the first commit to the `feature/instance-health` branch, it's meant as a framework to build into.
Next will be a PR to add the true Container Runtime check, including setting check status and timestamps
Following that will be a PR to aggregate the healthcheck statuses into a message to send to TACS.
Then there will be the actual message sending. 

So this is NOT everything.  It's just the Doctor.
 
The Doctor keeps track of an array of Healthchecks (where Healthcheck is an interface).  It exposes the RunTests() method to trigger each of these checks.
The Healthchecks will contain their current and last status as well as timestamps for each; the Doctor will read these to create its ultimate healthcheck message to send on to TACS.

### Testing
built agent and ran to see the healthchecks appear in `/var/log/ecs/ecs-agent.log`:
```
level=info time=2021-05-11T17:26:37Z msg="Received heartbeat from ACS" module=healthcheck.go
level=info time=2021-05-11T17:26:37Z msg="healthcheck response: {0xc0005ec800 <nil>}" module=healthcheck.go
level=info time=2021-05-11T17:26:37Z msg="healthcheck result: true" module=doctor.go
```
also added basic unit tests and ran them:
```
[ec2-user@ip-172-31-32-159 agent]$ go test -count=1 -v -tags unit ./doctor/
=== RUN   TestNewDoctor
--- PASS: TestNewDoctor (0.00s)
=== RUN   TestAddHealthcheck
--- PASS: TestAddHealthcheck (0.00s)
=== RUN   TestOk
--- PASS: TestOk (0.00s)
=== RUN   TestUnmarshalHealthcheckStatus
--- PASS: TestUnmarshalHealthcheckStatus (0.00s)
PASS
```
I'll add more detailed tests once we have more detailed Healthchecks.

Add test runner (Doctor) and basic docker ping Healthcheck.

### Licensing

### DIFF
--- a/agent/acs/handler/heartbeat_handler.go
+++ b/agent/acs/handler/heartbeat_handler.go
@@ -17,6 +17,8 @@ import (
 	"context"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/amazon-ecs-agent/agent/doctor"
+	"github.com/aws/amazon-ecs-agent/agent/engine"
 	"github.com/aws/amazon-ecs-agent/agent/wsclient"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
@@ -29,11 +31,13 @@ type heartbeatHandler struct {
 	ctx                       context.Context
 	cancel                    context.CancelFunc
 	acsClient                 wsclient.ClientServer
+	taskEngine                engine.TaskEngine
+	doctor                    *doctor.Doctor
 }
 
 // newHeartbeatHandler returns an instance of the heartbeatHandler struct
-func newHeartbeatHandler(ctx context.Context,
-	acsClient wsclient.ClientServer) heartbeatHandler {
+func newHeartbeatHandler(ctx context.Context, acsClient wsclient.ClientServer, taskEngine engine.TaskEngine) heartbeatHandler {
+	doctor := &doctor.Doctor{}
 
 	// Create a cancelable context from the parent context
 	derivedContext, cancel := context.WithCancel(ctx)
@@ -43,6 +47,8 @@ func newHeartbeatHandler(ctx context.Context,
 		ctx:                       derivedContext,
 		cancel:                    cancel,
 		acsClient:                 acsClient,
+		taskEngine:                taskEngine,
+		doctor:                    doctor,
 	}
 }
 
@@ -73,7 +79,19 @@ func (heartbeatHandler *heartbeatHandler) handleHeartbeatMessage() {
 }
 
 func (heartbeatHandler *heartbeatHandler) handleSingleHeartbeatMessage(message *ecsacs.HeartbeatMessage) error {
-	// Agent currently has no other action hooked to heartbeat messages, except simple ack
+	// TestHandlerDoesntLeakGoroutines unit test is failing because of this section
+	// This should all move under a separate healthcheck handler
+
+	// Agent will run healthchecks triggered by ACS heartbeat
+	// healthcheck results will be sent on to TACS
+	//dockerHealthcheck := doctor.NewDockerRuntimeHealthcheck(heartbeatHandler.taskEngine.TaskEngineClient())
+	//heartbeatHandler.doctor.AddHealthcheck(dockerHealthcheck)
+	//go func() {
+	//	checkResult := heartbeatHandler.doctor.RunHealthchecks()
+	//	seelog.Debugf("handler healthcheck result: %v", checkResult)
+	//}()
+
+	// Agent will send simple ack to the heartbeatAckMessageBuffer
 	go func() {
 		response := &ecsacs.HeartbeatAckRequest{
 			MessageId: message.MessageId,
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -162,6 +162,9 @@ type DockerClient interface {
 	// should be provided for the request.
 	ListContainers(context.Context, bool, time.Duration) ListContainersResponse
 
+	// SystemPing returns the Ping response from Docker's SystemPing API
+	SystemPing(context.Context, time.Duration) PingResponse
+
 	// ListImages returns the set of the images known to the Docker daemon
 	ListImages(context.Context, time.Duration) ListImagesResponse
 
@@ -1097,6 +1100,42 @@ func (dg *dockerGoClient) listImages(ctx context.Context) ListImagesResponse {
 		imageRepoTags = append(imageRepoTags, image.RepoTags...)
 	}
 	return ListImagesResponse{ImageIDs: imageIDs, RepoTags: imageRepoTags, Error: nil}
+}
+
+func (dg *dockerGoClient) SystemPing(ctx context.Context, timeout time.Duration) PingResponse {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	// Buffered channel so in the case of timeout it takes one write, never gets
+	// read, and can still be GC'd
+	response := make(chan PingResponse, 1)
+	go func() { response <- dg.systemPing(ctx) }()
+	select {
+	case resp := <-response:
+		return resp
+	case <-ctx.Done():
+		// Context has either expired or canceled. If it has timed out,
+		// send back the DockerTimeoutError
+		err := ctx.Err()
+		if err == context.DeadlineExceeded {
+			return PingResponse{Error: &DockerTimeoutError{timeout, "listing"}}
+		}
+		return PingResponse{Error: err}
+	}
+}
+
+func (dg *dockerGoClient) systemPing(ctx context.Context) PingResponse {
+	client, err := dg.sdkDockerClient()
+	if err != nil {
+		return PingResponse{Error: err}
+	}
+
+	pingResponse, err := client.Ping(ctx)
+	if err != nil {
+		return PingResponse{Error: err}
+	}
+
+	return PingResponse{Response: &pingResponse}
 }
 
 func (dg *dockerGoClient) SupportedVersions() []dockerclient.DockerVersion {
--- a/agent/doctor/doctor.go
+++ b/agent/doctor/doctor.go
@@ -1,0 +1,62 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//      http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package doctor
+
+import (
+	"sync"
+
+	"github.com/cihub/seelog"
+)
+
+type Doctor struct {
+	healthchecks []Healthcheck
+	lock         sync.RWMutex
+}
+
+func NewDoctor(healthchecks []Healthcheck) (*Doctor, error) {
+	newDoctor := &Doctor{
+		healthchecks: []Healthcheck{},
+	}
+	for _, hc := range healthchecks {
+		newDoctor.AddHealthcheck(hc)
+	}
+	return newDoctor, nil
+}
+
+func (doc *Doctor) AddHealthcheck(healthcheck Healthcheck) {
+	doc.lock.Lock()
+	defer doc.lock.Unlock()
+
+	doc.healthchecks = append(doc.healthchecks, healthcheck)
+}
+
+func (doc *Doctor) RunHealthchecks() bool {
+	doc.lock.RLock()
+	defer doc.lock.RUnlock()
+	allChecksResult := []bool{}
+	for _, healthcheck := range doc.healthchecks {
+		res := healthcheck.RunCheck()
+		seelog.Debugf("instance healthcheck result: %v", res)
+		allChecksResult = append(allChecksResult, res)
+	}
+	return doc.allRight(allChecksResult)
+}
+
+func (doc *Doctor) allRight(checksResult []bool) bool {
+	overallResult := true
+	for _, checkResult := range checksResult {
+		overallResult = overallResult && checkResult
+	}
+	return overallResult
+}
--- a/agent/doctor/doctor_test.go
+++ b/agent/doctor/doctor_test.go
@@ -1,0 +1,152 @@
+// +build unit
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package doctor
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+type trueHealthcheck struct{}
+
+func (tc *trueHealthcheck) RunCheck() bool { return true }
+func (tc *trueHealthcheck) GetHealthcheckStatus() HealthcheckStatus {
+	return HealthcheckStatusInitializing
+}
+func (tc *trueHealthcheck) GetHealthcheckTime() time.Time {
+	return time.Date(1974, time.May, 19, 1, 2, 3, 4, time.UTC)
+}
+
+type falseHealthcheck struct{}
+
+func (fc *falseHealthcheck) RunCheck() bool { return false }
+func (fc *falseHealthcheck) GetHealthcheckStatus() HealthcheckStatus {
+	return HealthcheckStatusInitializing
+}
+func (fc *falseHealthcheck) GetHealthcheckTime() time.Time {
+	return time.Date(1974, time.May, 19, 1, 2, 3, 4, time.UTC)
+}
+
+func TestNewDoctor(t *testing.T) {
+	trueCheck := &trueHealthcheck{}
+	falseCheck := &falseHealthcheck{}
+
+	healthchecks := []Healthcheck{trueCheck, falseCheck}
+	newDoctor, _ := NewDoctor(healthchecks)
+	assert.Len(t, newDoctor.healthchecks, 2)
+}
+
+type testHealthcheck struct {
+	testName string
+}
+
+func (thc *testHealthcheck) RunCheck() bool {
+	return true
+}
+
+func (thc *testHealthcheck) GetHealthcheckStatus() HealthcheckStatus {
+	return HealthcheckStatusOk
+}
+
+func (thc *testHealthcheck) GetHealthcheckTime() time.Time {
+	return time.Date(1974, time.May, 19, 1, 2, 3, 4, time.UTC)
+}
+
+func TestAddHealthcheck(t *testing.T) {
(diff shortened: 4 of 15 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -162,6 +162,9 @@ type DockerClient interface {
 	// should be provided for the request.
 	ListContainers(context.Context, bool, time.Duration) ListContainersResponse
 
+	// PingDocker returns the Ping response from Docker's SystemPing API
+	PingDocker(context.Context, time.Duration) PingResponse
```
angelcar, line 166: nit: this method should probably be just called `Ping` since it's already inside `dockerGoClient`.

agent/doctor/doctor.go
```diff
@@ -0,0 +1,41 @@
...
+	return &Doctor{
+		healthchecks: []Healthcheck{},
+	}, nil
+}
+
+func (doc *Doctor) AddHealthcheck(healthcheck Healthcheck) {
+	doc.healthchecks = append(doc.healthchecks, healthcheck)
+}
+
+func (doc *Doctor) RunHealthchecks() bool {
```
angelcar, line 34: This doesn't seem to be tested

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,67 @@
...
+		client:            client,
+		LatestCheckStatus: HealthcheckStatusInitializing,
+		LatestCheckTime:   time.Now(),
+		LastCheckStatus:   HealthcheckStatusInitializing,
+		LastCheckTime:     time.Now(),
+	}
+}
+
+func (dhc *dockerRuntimeHealthcheck) RunCheck() bool {
+	seelog.Infof("Received heartbeat from ACS")
```
angelcar, line 53: I don't think this log entry belongs here? this health check is/should not aware of ACS, right?

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,67 @@
...
+	// Since is the timestamp when container health status changed
+	LatestCheckTime time.Time `json:"latestStatusTime,omitempty"`
+	// Status is the container health status
+	LastCheckStatus HealthcheckStatus `json:"lastStatus,omitempty"`
+	// Since is the timestamp when container health status changed
+	LastCheckTime time.Time `json:"lastStatusTime,omitempty"`
+	client        dockerapi.DockerClient
+}
+
+func NewDockerRuntimeHealthcheck(client dockerapi.DockerClient) dockerRuntimeHealthcheck {
```
angelcar, line 42: you need to return a pointer here

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/cihub/seelog"
+)
+
+type Healthcheck interface {
+	GetHealthcheckStatus() HealthcheckStatus
+	GetHealthcheckTime() time.Time
+	RunCheck() bool
+}
+
+type dockerRuntimeHealthcheck struct {
```
angelcar, line 30: will this need to be persisted?

agent/acs/handler/heartbeat_handler.go
```diff
@@ -73,7 +79,17 @@ func (heartbeatHandler *heartbeatHandler) handleHeartbeatMessage() {
...
 func (heartbeatHandler *heartbeatHandler) handleSingleHeartbeatMessage(message *ecsacs.HeartbeatMessage) error {
-	// Agent currently has no other action hooked to heartbeat messages, except simple ack
+	// Agent will run healthchecks triggered by ACS heartbeat
+	// healthcheck results will be sent on to TACS
+	dockerHealthcheck := doctor.NewDockerRuntimeHealthcheck(heartbeatHandler.taskEngine.TaskEngineClient())
+	heartbeatHandler.doctor.AddHealthcheck(&dockerHealthcheck)
+
+	go func() {
+		checkResult := heartbeatHandler.doctor.RunHealthchecks()
+		seelog.Infof("handler healthcheck result: %v", checkResult)
```
shubham2892, line 89: Do we want to keep this log at info level or debug level as it will be a bit noisy?

agent/doctor/doctor.go
```diff
@@ -0,0 +1,41 @@
...
+	//result will ultimately be an array of something...
+	for _, healthcheck := range doc.healthchecks {
+		res := healthcheck.RunCheck()
+		seelog.Infof("healthcheck result: %v", res)
```
shubham2892, line 38: Same comment as previous log statement, info or debug level?

agent/doctor/doctor.go
shubham2892, line 34: nit: `RunHealthChecks()` ?

agent/doctor/doctor.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	return true
```
shubham2892, line 40: Right now `RunHealthchecks` always returns `true` irrespective of what we get from individual health checks, do we want to do an `AND` of individual healthcheck and return that?

agent/doctor/healthcheck.go
shubham2892: Across Agent restarts?

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,67 @@
...
+	// Status is the container health status
+	LatestCheckStatus HealthcheckStatus `json:"latestStatus,omitempty"`
+	// Since is the timestamp when container health status changed
+	LatestCheckTime time.Time `json:"latestStatusTime,omitempty"`
+	// Status is the container health status
```
shubham2892, line 35: nit: comments needs to be updated.

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,67 @@
...
+	LastCheckStatus HealthcheckStatus `json:"lastStatus,omitempty"`
+	// Since is the timestamp when container health status changed
```
shubham2892, line 37: nit: comments needs to be updated.

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,67 @@
...
+	client        dockerapi.DockerClient
+}
+
+func NewDockerRuntimeHealthcheck(client dockerapi.DockerClient) dockerRuntimeHealthcheck {
+	return dockerRuntimeHealthcheck{
+		client:            client,
+		LatestCheckStatus: HealthcheckStatusInitializing,
+		LatestCheckTime:   time.Now(),
+		LastCheckStatus:   HealthcheckStatusInitializing,
+		LastCheckTime:     time.Now(),
```
shubham2892, line 48: can we set this to null instead of initializing it because technically we do should not have these value when initializing?

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func (dhc *dockerRuntimeHealthcheck) RunCheck() bool {
+	seelog.Infof("Received heartbeat from ACS")
+	healthcheckResponse := dhc.client.PingDocker(context.TODO(), time.Second*2)
+	seelog.Infof("healthcheck response: %v", healthcheckResponse)
```
shubham2892, line 55: Should we move this to debug?

agent/dockerclient/dockerapi/docker_client.go
fierlion: I think I'll call it `SystemPing` --
I wanted to differentiate it from a simple Ping as it is calling a dedicated Docker API.
The fancy name for the API in Docker's docs is "System Ping" https://docs.docker.com/engine/api/v1.41/#operation/SystemPing

agent/doctor/doctor.go
fierlion: it will be.  rev coming.

agent/doctor/doctor.go
fierlion: we do want that.  Will make the update.

agent/doctor/doctor.go
fierlion: I've followed the global convention of calling it a `Healthcheck`.

agent/doctor/healthcheck.go
fierlion: Although I added a LastCheckStatus/Time, we don't currently have a use for it.  And the only reason I can see persistance being necessary is to track our Last state.  

As we've talked about in team discussions -- these stats are really about the very latest stat.
So in case we start up after an agent reset (whether by service restart or instance reboot) and the doctor prematurely sends an aggreagate healthcheck message, the worst case would be that we'd see our defaults in cloudwatch.
```
                LatestCheckStatus: HealthcheckStatusInitializing,
                LatestCheckTime:   time.Now(),
```
Also note that the aggregation and message sending isn't part of the present PR but it should take into account the requirement of always initializing healthchecks where necessary.

So long story short: I don't think persistence is necessary.  Yet.

shubham2892 (approved):
Was the file `output` intended to be added to the commit? LGTM otherwise.

agent/doctor/doctor.go
```diff
@@ -0,0 +1,50 @@
...
+
+func (doc *Doctor) AddHealthcheck(healthcheck Healthcheck) {
+	doc.healthchecks = append(doc.healthchecks, healthcheck)
+}
+
+func (doc *Doctor) RunHealthchecks() bool {
+	allChecksResult := []bool{}
+	for _, healthcheck := range doc.healthchecks {
+		res := healthcheck.RunCheck()
+		seelog.Debugf("healthcheck result: %v", res)
```
sharanyad, line 38: could you add more information to this log statement something like `instance health check result`?

agent/doctor/doctor_test.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewDoctor(t *testing.T) {
+	newDoctor, _ := NewDoctor()
+	assert.True(t, len(newDoctor.healthchecks) == 0)
```
sharanyad, line 27: could just use `assert.Len` and below

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,66 @@
...
+		LatestCheckStatus: HealthcheckStatusInitializing,
+		LatestCheckTime:   time.Now(),
+		LastCheckStatus:   HealthcheckStatusInitializing,
+		LastCheckTime:     time.Now(),
+	}
+}
+
+func (dhc *dockerRuntimeHealthcheck) RunCheck() bool {
+	healthcheckResponse := dhc.client.SystemPing(context.TODO(), time.Second*2)
+	seelog.Infof("healthcheck response: %v", healthcheckResponse)
```
sharanyad, line 54: `container runtime health check response`

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,66 @@
...
+	return true
+}
+
+func (dhc *dockerRuntimeHealthcheck) GetHealthcheckStatus() HealthcheckStatus {
+	// need to lock/unlock?
+	return dhc.LatestCheckStatus
+}
+
+func (dhc *dockerRuntimeHealthcheck) GetHealthcheckTime() time.Time {
+	// need to lock/unlock?
```
sharanyad, line 64: always safe to add locks

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,66 @@
...
+		client:            client,
+		LatestCheckStatus: HealthcheckStatusInitializing,
+		LatestCheckTime:   time.Now(),
+		LastCheckStatus:   HealthcheckStatusInitializing,
+		LastCheckTime:     time.Now(),
+	}
+}
+
+func (dhc *dockerRuntimeHealthcheck) RunCheck() bool {
+	healthcheckResponse := dhc.client.SystemPing(context.TODO(), time.Second*2)
```
sharanyad, line 53: could we make the time constant? and pass in `context.Background` if possible?

fierlion: the systemping here is for demo purposes only.  we have a followup PR to add the true docker runtime check.

agent/doctor/doctor.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func NewDoctor() (*Doctor, error) {
+	return &Doctor{
+		healthchecks: []Healthcheck{},
+	}, nil
+}
+
+func (doc *Doctor) AddHealthcheck(healthcheck Healthcheck) {
+	doc.healthchecks = append(doc.healthchecks, healthcheck)
```
sparrc, line 31: this should probably have a write lock

agent/doctor/doctor.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+func (doc *Doctor) RunHealthchecks() bool {
+	allChecksResult := []bool{}
```
sparrc, line 35: and this function a read lock i think

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,70 @@
...
+	LastCheckTime time.Time `json:"lastStatusTime,omitempty"`
+	client        dockerapi.DockerClient
+	lock          sync.RWMutex
+}
+
+func NewDockerRuntimeHealthcheck(client dockerapi.DockerClient) *dockerRuntimeHealthcheck {
+	return &dockerRuntimeHealthcheck{
+		client:            client,
+		LatestCheckStatus: HealthcheckStatusInitializing,
+		LatestCheckTime:   time.Now(),
```
sparrc, line 48: shouldnt these timestamps be set when the healthchecks are run in RunCheck ?

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,70 @@
...
+	// Since is the timestamp when container health status changed
+	LastCheckTime time.Time `json:"lastStatusTime,omitempty"`
+	client        dockerapi.DockerClient
+	lock          sync.RWMutex
+}
+
+func NewDockerRuntimeHealthcheck(client dockerapi.DockerClient) *dockerRuntimeHealthcheck {
+	return &dockerRuntimeHealthcheck{
+		client:            client,
+		LatestCheckStatus: HealthcheckStatusInitializing,
```
sparrc, line 47: If these are meant to only be read by the public and read-locked functions (GetHealthcheckStatus), then these should not be public.

agent/doctor/healthcheck.go
```diff
@@ -0,0 +1,70 @@
...
+func (dhc *dockerRuntimeHealthcheck) RunCheck() bool {
+	healthcheckResponse := dhc.client.SystemPing(context.TODO(), time.Second*2)
+	seelog.Infof("container runtime healthcheck response: %v", healthcheckResponse)
+	return true
+}
+
+func (dhc *dockerRuntimeHealthcheck) GetHealthcheckStatus() HealthcheckStatus {
+	dhc.lock.RLock()
+	defer dhc.lock.RUnlock()
+	return dhc.LatestCheckStatus
```
sparrc, line 63: I think I might be missing something, where are these variables ever set/changed?

agent/doctor/healthcheckstatus.go
```diff
@@ -0,0 +1,81 @@
...
+// permissions and limitations under the License.
+
+package doctor
+
+import (
+	"errors"
+	"strings"
+)
+
+const (
```
sparrc, line 21: I think we could just make these statuses string constants instead of using int32 and iota. I think that way we don't need to have all this custom json marshalling either.

agent/doctor/doctor.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"github.com/cihub/seelog"
+)
+
+type Doctor struct {
+	healthchecks []Healthcheck
+}
+
+func NewDoctor() (*Doctor, error) {
```
sparrc, line 24: Having a "New" function that returns a Doctor that doesn't do anything doesnt seem like best interface to me. Perhaps the NewDoctor function should automatically call AddHeakthcheck with all the available healthchecks? Personally this seems like a better experience to me than every caller having to manually add each healthcheck even when using the "New" convenience function.

agent/acs/handler/heartbeat_handler.go
```diff
@@ -73,7 +79,19 @@ func (heartbeatHandler *heartbeatHandler) handleHeartbeatMessage() {
 }
 
 func (heartbeatHandler *heartbeatHandler) handleSingleHeartbeatMessage(message *ecsacs.HeartbeatMessage) error {
-	// Agent currently has no other action hooked to heartbeat messages, except simple ack
+	// TestHandlerDoesntLeakGoroutines unit test is failing because of this section
+	// This should all move under a separate healthcheck handler
+
+	// Agent will run healthchecks triggered by ACS heartbeat
+	// healthcheck results will be sent on to TACS
+	//dockerHealthcheck := doctor.NewDockerRuntimeHealthcheck(heartbeatHandler.taskEngine.TaskEngineClient())
```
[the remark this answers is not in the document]
fierlion: this was used in the testing done and will ultimately end up in a separate healthcheck specific handler.  I left it in to show the trigger point used in the description's testing done.

agent/doctor/healthcheck.go
fierlion: these follow the pattern in our api code and allow us to use the json struct field tags.
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/container/container.go#L102-L111

agent/doctor/healthcheck.go
fierlion: this pr is merging into healthcheck feature branch and has a basic "demo" container runtime check.  This PR is specifically meant to add the Doctor.  We have a follow up pr to add the true container runtime check, and to set the statuses and then a third pr to poll these statuses for the message creation and send to TACS.  I'll update the description with this detail.

agent/doctor/healthcheck.go
fierlion: this pr is merging into healthcheck feature branch and has a basic "demo" container runtime check.  This PR is specifically meant to add the Doctor.  We have a follow up pr to add the true container runtime check, and to set the statuses and then a third pr to poll these statuses for the message creation and send to TACS.  I'll update the description with this detail.

agent/doctor/healthcheckstatus.go
fierlion: I know we have opinions on this pattern in agent codebase.  I prefer to follow the existing pattern when sending messages to TACS or FE.
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/container/container.go#L102-L111

sparrc: but if the existing pattern is not standard or good I'd argue we should differ from it and improve our code

agent/doctor/healthcheck.go
sparrc: this could still be set in RunCheck even if it's demo code so far couldnt it?

agent/doctor/doctor.go
fierlion: good idea -- I'll update to let it push all the Healthchecks in

agent/doctor/healthcheck.go
sparrc: why would we follow the pattern in our API code? this code is more internal isn't it?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -1099,6 +1102,42 @@ func (dg *dockerGoClient) listImages(ctx context.Context) ListImagesResponse {
 	return ListImagesResponse{ImageIDs: imageIDs, RepoTags: imageRepoTags, Error: nil}
 }
 
+func (dg *dockerGoClient) SystemPing(ctx context.Context, timeout time.Duration) PingResponse {
```
sharanyad, line 1105: needs unit tests in https://github.com/aws/amazon-ecs-agent/blob/master/agent/dockerclient/dockerapi/docker_client_test.go

sharanyad (approved):
comment on unit testing. could add in a different PR.

PR #2866, opened 2021-05-13
adding check to validate default config file path for Windows

singholt:
### Summary

The PR adds a check to validate the default agent config file path on Windows. 

Currently, the default config file path on Windows is ```"C:/etc/ecs_container_agent/config.json"```. We're adding a check to validate whether the admin has created the config file or not. Note that this check is not done if the env ```ECS_AGENT_CONFIG_FILE_PATH``` is set, since it overrides the default config file.

### Licensing

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -437,8 +437,11 @@ func (cfg *Config) complete() bool {
 }
 
 func fileConfig() (Config, error) {
-	fileName := utils.DefaultIfBlank(os.Getenv("ECS_AGENT_CONFIG_FILE_PATH"), defaultConfigFileName)
 	cfg := Config{}
+	fileName, err := getConfigFileName()
+	if err != nil {
+		return cfg, nil
+	}
 
 	file, err := os.Open(fileName)
 	if err != nil {
--- a/agent/config/config_windows.go
+++ b/agent/config/config_windows.go
@@ -19,8 +19,13 @@ import (
 	"path/filepath"
 	"time"
 
+	"golang.org/x/sys/windows"
+
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
+
+	"github.com/cihub/seelog"
+	"github.com/hectane/go-acl/api"
 )
 
 const (
@@ -55,6 +60,9 @@ const (
 	minimumContainerCreateTimeout = 1 * time.Minute
 	// default image pull inactivity time is extra time needed on container extraction
 	defaultImagePullInactivityTimeout = 3 * time.Minute
+	// adminSid is the security ID for the admin group on Windows
+	// Reference: https://docs.microsoft.com/en-us/troubleshoot/windows-server/identity/security-identifiers-in-windows
+	adminSid = "S-1-5-32-544"
 )
 
 // DefaultConfig returns the default configuration for Windows
@@ -143,4 +151,66 @@ func (cfg *Config) platformOverrides() {
 // to string for debugging
 func (cfg *Config) platformString() string {
 	return ""
+}
+
+var getNamedSecurityInfo = api.GetNamedSecurityInfo
+
+// validateConfigFile checks if the config file owner is an admin
+// Reference: https://github.com/hectane/go-acl#using-the-api-directly
+func validateConfigFile(configFileName string) (bool, error) {
+	var (
+		Sid    *windows.SID
+		handle windows.Handle
+	)
+	err := getNamedSecurityInfo(
+		configFileName,
+		api.SE_FILE_OBJECT,
+		api.OWNER_SECURITY_INFORMATION,
+		&Sid,
+		nil,
+		nil,
+		nil,
+		&handle,
+	)
+	if err != nil {
+		return false, err
+	}
+	defer windows.LocalFree(handle)
+
+	id, err := Sid.String()
+	if err != nil {
+		return false, err
+	}
+
+	if id == adminSid {
+		return true, nil
+	}
+	seelog.Debugf("Non-admin cfg file owner with SID: %v, skip merging into agent config", id)
+	return false, nil
+}
+
+var osStat = os.Stat
+
+func getConfigFileName() (string, error) {
+	fileName := os.Getenv("ECS_AGENT_CONFIG_FILE_PATH")
+	// validate the config file only if above env var is not set
+	if len(fileName) == 0 {
+		fileName = defaultConfigFileName
+		// check if the default config file exists before validating it
+		_, err := osStat(fileName)
+		if err != nil {
+			return "", err
+		}
+
+		isValidFile, err := validateConfigFile(fileName)
+		if err != nil {
+			seelog.Errorf("Unable to validate cfg file: %v, err: %v", fileName, err)
+			return "", err
+		}
+		if !isValidFile {
+			seelog.Error("Invalid cfg file")
+			return "", err
+		}
+	}
+	return fileName, nil
 }
--- a/agent/config/config_windows_test.go
+++ b/agent/config/config_windows_test.go
@@ -16,12 +16,17 @@
 package config
 
 import (
+	"errors"
 	"os"
 	"testing"
 	"time"
 
+	"golang.org/x/sys/windows"
+
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/ec2"
+
+	"github.com/hectane/go-acl/api"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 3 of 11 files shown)


agent/config/config.go
```diff
@@ -50,9 +51,12 @@ const (
 	// AgentPrometheusExpositionPort is used to expose Prometheus metrics that can be scraped by a Prometheus server
 	AgentPrometheusExpositionPort = 51680
 
-	// defaultConfigFileName is the default (json-formatted) config file
+	// defaultConfigFileName is the default (json-formatted) config file (non-Windows)
 	defaultConfigFileName = "/etc/ecs_container_agent/config.json"
 
+	// defaultWindowsConfigFileName is the default (json-formatted) config file on Windows
+	defaultWindowsConfigFileName = `C:\ProgramFiles\Amazon\ECS\config.json`
```
choihca, line 58: Should it be ```C:\Program Files\Amazon\ECS\config.json``` (with space in Program Files)?

agent/config/config.go
```diff
@@ -438,6 +442,9 @@ func (cfg *Config) complete() bool {
 
 func fileConfig() (Config, error) {
 	fileName := utils.DefaultIfBlank(os.Getenv("ECS_AGENT_CONFIG_FILE_PATH"), defaultConfigFileName)
+	if runtime.GOOS == "windows" {
```
rawahars, line 445: Instead of checking the runtime OS, let's segregate the `defaultConfigFileName` constant in Windows and Linux/Unsupported files. This would ensure that we are not adding platform segregation logic in the common code.

agent/config/config.go
singholt: oops good catch, fixed it

rawahars (changes requested):

agent/config/config.go
singholt: done, thanks

rawahars (dismissed):
LGTM!

choihca (approved):

vsiddharth (approved):

mythri-garaga (dismissed):

sparrc:
won't this break every customer currently using `/etc/ecs_container_agent/config.json` on windows?

singholt:
> won't this break every customer currently using `/etc/ecs_container_agent/config.json` on windows?

I dont think we ask customers to use /etc/ecs_container_agent/config.json directly anywhere in the docs. We advertise the env variable and ask customers to set this env variable if they want to use a custom config.json

agent/config/config_windows.go
```diff
@@ -144,3 +150,35 @@ func (cfg *Config) platformOverrides() {
 func (cfg *Config) platformString() string {
 	return ""
 }
+
+var Sid *windows.SID
+var handle windows.Handle
```
angelcar, line 155: I think it would be safer to put these inside `validateConfigFile`, doesn't feel like these should be package level vars.

agent/config/config_windows.go
```diff
@@ -144,3 +150,35 @@ func (cfg *Config) platformOverrides() {
...
+
+	Sid, err := Sid.String()
+	if err != nil {
+		return false, err
+	}
+
+	if Sid == adminSid {
+		return true, nil
+	}
+	return false, nil
```
angelcar, line 183: could we add a debug log entry with the Sid here?

agent/config/config.go
```diff
@@ -437,8 +437,26 @@ func (cfg *Config) complete() bool {
...
+		isValidFile, err := validateConfigFile(fileName)
+		if err != nil {
+			seelog.Errorf("Unable to validate cfg file, err %v", err)
+			return cfg, err
+		}
+		if !isValidFile {
+			seelog.Error("Invalid cfg file")
+			return cfg, nil
+		}
+	}
```
angelcar, line 459: It would probably be better to have a `getConfigFileName() (string, error)` function that does this checks for Windows and for Linux just does `return utils.DefaultIfBlank(os.Getenv("ECS_AGENT_CONFIG_FILE_PATH"), defaultConfigFileName), nil`.

angelcar (approved):

angelcar (changes requested):
Can we add or update the UTs to cover the new code, please?

angelcar (approved):

fenxiong (approved):

agent/config/config_windows.go
```diff
@@ -19,6 +19,11 @@ import (
 	"path/filepath"
 	"time"
 
+	"github.com/cihub/seelog"
```
vsiddharth, line 22: nit: can you restructure the imports to follow the golang package grouping

agent/config/config_windows.go
```diff
@@ -144,3 +152,64 @@ func (cfg *Config) platformOverrides() {
...
+		return true, nil
+	}
+	seelog.Debugf("Non-admin cfg file owner with SID: %v, skip merging into agent config", id)
+	return false, nil
+}
+
+var osStat = os.Stat
+
+func getConfigFileName() (string, error) {
+	fileName := os.Getenv("ECS_AGENT_CONFIG_FILE_PATH")
```
vsiddharth, line 195: `LookupEnv` might suit better here?

agent/config/config_windows.go
```diff
@@ -144,3 +152,64 @@ func (cfg *Config) platformOverrides() {
...
+		fileName = defaultConfigFileName
+		// check if the default config file exists before validating it
+		_, err := osStat(fileName)
+		if err != nil {
+			return "", err
+		}
+
+		isValidFile, err := validateConfigFile(fileName)
+		if err != nil {
+			seelog.Errorf("Unable to validate cfg file, err %v", err)
```
vsiddharth, line 206: nit: Can you add the cfg file in the message here?

agent/config/config_windows_test.go
```diff
@@ -16,10 +16,14 @@
 package config
 
 import (
+	"errors"
 	"os"
 	"testing"
 	"time"
 
+	"github.com/hectane/go-acl/api"
```
vsiddharth, line 24: nit: same comment as above. Please restructure the imports here.

agent/config/config_windows_test.go
```diff
@@ -134,3 +138,66 @@ func TestMemoryUnboundedWindowsDisabled(t *testing.T) {
...
+		{
+			name:             "invalid cfg file error",
+			envVarVal:        "",
+			expectedFileName: "",
+			expectedError:    errors.New("Invalid cfg file"),
+			cfgFileSid:       "S-1-5-7",
+		},
+	}
+	defer func() {
+		osStat = os.Stat
```
vsiddharth, line 181: Can you use the `oswrapper` and its mock for the tests?

agent/config/config_windows.go
vsiddharth, line 195: nit: Can you add a comment clarifying that validation is only performed for env fileName?

agent/config/config_windows.go
singholt: can you elaborate? i've already run ```go fmt```/```go imports```

agent/config/config_windows.go
singholt: this was existing prior to this change in ```config.go```

singholt: I'll add a comment. the validation is only performed if the env is blank/not present

agent/config/config_windows_test.go
singholt: i think this is simple enough no? i see other tests following the same logic.

agent/config/config_windows.go
singholt: done

agent/config/config_windows.go
vsiddharth:
```
import (
    stdlib

    current_project

    company

    all others 
)
```

agent/config/config_windows_test.go
angelcar: Personally I don't like the `oswrapper` because it introduces more complexity for something simple like this. Forst we need to introduce a new dependency, and second, the mentioned interface is not segregated and exposes too many methods that are not required. 
This is a valid approach, for example see the [sample code](https://github.com/adonovan/gopl.io/blob/master/ch11/echo/echo.go#L23) where they use this approach in [The Go Programming Language](https://www.gopl.io/) book.

vsiddharth: I've seen both approaches and I'm alright with either of them.

agent/config/config_windows.go
angelcar: The static checks are expecting the structure of the inputs to be the result of running `go imports`. Where is that order coming from?

singholt: not sure i follow. is the suggestion to do something like:
```
import (
	"os"
	"path/filepath"
	"time"

	"golang.org/x/sys/windows"

	"github.com/aws/amazon-ecs-agent/agent/dockerclient"

	"github.com/cihub/seelog"
	"github.com/hectane/go-acl/api"
)
```

agent/config/config_windows_test.go
singholt: and this to be like:
```
import (
	"errors"
	"os"
	"testing"
	"time"

	"golang.org/x/sys/windows"

	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
	"github.com/aws/amazon-ecs-agent/agent/ec2"
	"github.com/hectane/go-acl/api"
)
```

agent/config/config_windows.go
vsiddharth: Yep, I prefer this.

singholt: @angelcar you okay with this too? ran ```go imports``` after and it seems to accept this order.

angelcar: If the static checks are ok with it, I'm ok with it =)

singholt: done

angelcar (approved):

fenxiong (approved):

vsiddharth (approved):

agent/config/config_windows.go
```diff
@@ -144,3 +152,65 @@ func (cfg *Config) platformOverrides() {
...
+	if len(fileName) == 0 {
+		fileName = defaultConfigFileName
+		// check if the default config file exists before validating it
+		_, err := osStat(fileName)
+		if err != nil {
+			return "", err
+		}
+
+		isValidFile, err := validateConfigFile(fileName)
+		if err != nil {
```
vsiddharth, line 206: nit: Can you combine the two checks?

singholt: I prefer this to be much cleaner and it's simple enough to maintain. Plus we want to raise "unable to validate cfg error" only if there is a cfg file to validate.

PR #2867, opened 2021-05-13
EC2 Task Networking on Windows: Behavior change of cleanupNS and increased CNI setup time limits

rawahars:
### Summary

Following tasks have been accomplished in this PR-

- Segregate cleanupNS method into Linux and Windows specific files.
- Differences in Windows implementation as compared to Linux one are-
     1. The cleanup method is protected by a mutex. This is done because hcsshim does not allow parallel API invocations and therefore, the calls must be serialized.
     2. In case of error during one of the CNI plugin invocations, do not return an error immediately. Try to cleanup as much as possible before conceding error.
- CNI setup and cleanup time for Windows has been increased. This is required because CNI plugin calls might be retried or serialized.
- A minor test case from watcher package has been improved to avoid random test failures.

### Implementation details

Linux and Windows implementations have been segregated and necessary changes are implemented in the Windows implementation of the method.
### Testing

The changes were tested using custom agent binary.

New tests cover the changes: 
Yes

Behavior change of cleanupNS and increased CNI setup time limits
### Licensing

### DIFF
--- a/agent/ecscni/plugin.go
+++ b/agent/ecscni/plugin.go
@@ -22,7 +22,6 @@ import (
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/logger"
-	"github.com/cihub/seelog"
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/pkg/errors"
@@ -44,12 +43,6 @@ type CNIClient interface {
 	CleanupNS(context.Context, *Config, time.Duration) error
 	// ReleaseIPResource marks the ip available in the ipam db
 	ReleaseIPResource(context.Context, *Config, time.Duration) error
-}
-
-// cniClient is the client to call plugin and setup the network
-type cniClient struct {
-	pluginsPath string
-	libcni      libcni.CNI
 }
 
 // NewClient creates a client of ecscni which is used to invoke the plugin
@@ -96,40 +89,6 @@ func (client *cniClient) CleanupNS(
 	defer cancel()
 
 	return client.cleanupNS(ctx, cfg)
-}
-
-// cleanupNS is called by CleanupNS to cleanup the task namespace by invoking DEL for given CNI configurations
-func (client *cniClient) cleanupNS(ctx context.Context, cfg *Config) error {
-	seelog.Debugf("[ECSCNI] Cleaning up the container namespace %s", cfg.ContainerID)
-
-	runtimeConfig := libcni.RuntimeConf{
-		ContainerID: cfg.ContainerID,
-		NetNS:       cfg.ContainerNetNS,
-	}
-
-	// Execute all CNI network configurations serially, in the reverse order.
-	for i := len(cfg.NetworkConfigs) - 1; i >= 0; i-- {
-		networkConfig := cfg.NetworkConfigs[i]
-		cniNetworkConfig := networkConfig.CNINetworkConfig
-		seelog.Debugf("[ECSCNI] Deleting network %s type %s in the container namespace %s",
-			cniNetworkConfig.Network.Name,
-			cniNetworkConfig.Network.Type,
-			cfg.ContainerID)
-		runtimeConfig.IfName = networkConfig.IfName
-		err := client.libcni.DelNetwork(ctx, cniNetworkConfig, &runtimeConfig)
-		if err != nil {
-			return errors.Wrap(err, "delete network failed")
-		}
-
-		seelog.Debugf("[ECSCNI] Completed deleting network %s type %s in the container namespace %s",
-			cniNetworkConfig.Network.Name,
-			cniNetworkConfig.Network.Type,
-			cfg.ContainerID)
-	}
-
-	seelog.Debugf("[ECSCNI] Completed cleaning up the container namespace %s", cfg.ContainerID)
-
-	return nil
 }
 
 // Version returns the version of the plugin
--- a/agent/ecscni/plugin.go
+++ b/agent/ecscni/plugin.go
@@ -19,9 +19,11 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/logger"
+	"github.com/cihub/seelog"
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/types/current"
 	"github.com/pkg/errors"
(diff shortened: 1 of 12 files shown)


agent/ecscni/plugin_windows.go
```diff
@@ -114,6 +122,46 @@ func (client *cniClient) doSetupNS(ctx context.Context, cfg *Config) (*current.R
 	return curResult, nil
 }
 
+// cleanupNS is called by CleanupNS to cleanup the task namespace by invoking DEL for given CNI configurations.
+func (client *cniClient) cleanupNS(ctx context.Context, cfg *Config) error {
```
angelcar, line 126: Other than the guard and the error handling, this function is pretty much the same for Linux and Windows, it doesn't feel right to duplicate it, so I have these questions:

* I don't think this method is invoked in parallel in Linux, I think it's only ever called once to clean up the pause container. Where do you expect parallel invocations for windows? In order to keep the a single method, I think it's safe to add the mutex to linux as well, or just not use it if it's not really needed.
* Given that possibly the only difference for windows would be error handling, we could instead create some `handleDeleteError` function which indicates if the loop should return or keep going. For windows it would keep going, for linux it would return... but ti be honest I think we could do exactly the same on linux.

thoughts?

rawahars: - This method would be evoked in parallel whenever multiple tasks are stopped at once on any instance. The same is true for both Windows and Linux. If we were to add a mutex for Linux as well then it might increase the cleanup time for Linux awsvpc network mode as well. But i think it should be at max 1 minute.

- Yes I agree that we can do the same error handling on Linux as well. Basically, we are trying to delete the networks as much as possible before conceding error. That should be fine for both Linux and Windows.

Let's gather other opinions as well!

angelcar: > * This method would be evoked in parallel whenever multiple tasks are stopped at once on any instance. The same is true for both Windows and Linux. If we were to add a mutex for Linux as well then it might increase the cleanup time for Linux awsvpc network mode as well. But i think it should be at max 1 minute.

I think this is ok since the clean up happens after the task is already in the process of stopping, so there are no strict performance considerations there (plus cleaning up the network is usually really fast). 

> * Yes I agree that we can do the same error handling on Linux as well. Basically, we are trying to delete the networks as much as possible before conceding error. That should be fine for both Linux and Windows.

Yes, I also think this is a good idea for linux anyways

fenxiong: regarding guard: is there any reason why windows task networking cleanup needs the lock? my understanding is the cni plugin should be responsible to be thread safe and expect itself to be called in parallel

regarding error handling: yes i agree that it can be added to linux too

rawahars: The CNI plugin calls Windows HNS API's using hcsshim which fails if we call the same API's in parallel.
So if we try to stop multiple tasks at the same time, then multiple CNI plugin invocations would be made at the same time which calls same HNS API. This might lead to failure of few calls.

Therefore, we need to serialize the calls to CNI plugin while cleanup.

rawahars: With respect to the cleanupNS guard, the task workloads for Linux and Windows are different. Windows runs fewer tasks at a time as compared to Linux. If a large number of tasks are stopped at the same time on Linux, then the cleanup time might be greater than expected.

rawahars: As per our offline discussions, the guard has been added only for Windows.
However, for Linux as well, we are performing the cleanup as much as possible before conceding error.

agent/engine/docker_task_engine_unsupported.go
```diff
@@ -16,10 +16,18 @@
...
 import (
+	"time"
+
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
 )
 
+const (
+	// Constants for CNI timeout during setup and cleanup.
+	cniSetupTimeout   = 1 * time.Minute
```
angelcar, line 27: nit: this could just be 

```golang
const (
    cniSetupTimeout = time.Duration(0)
    cniCleanupTimeout = time.Duration(0)
)
```

agent/ecscni/plugin_linux.go
```diff
@@ -39,6 +39,14 @@ const (
 	vpcCNIPluginPath = "/log/vpc-branch-eni.log"
 )
 
+// NewCNIGuard returns a new instance of CNI guard for the CNI client.
+// It is supported only on Windows.
+func NewCNIGuard() CNIGuard {
```
angelcar, line 44: nit: the interface as well as the  constructor don't need to be exported

angelcar (approved):

agent/ecscni/plugin.go
```diff
@@ -46,10 +47,22 @@ type CNIClient interface {
 	ReleaseIPResource(context.Context, *Config, time.Duration) error
 }
 
+// cniGuard is the mutex interface for CNI Client.
+type cniGuard interface {
```
vsiddharth, line 51: Can you please add additional details as to why this interface is required?

agent/ecscni/plugin.go
```diff
@@ -100,13 +114,17 @@ func (client *cniClient) CleanupNS(
...
+	defer client.guard.unlock()
+
 	seelog.Debugf("[ECSCNI] Cleaning up the container namespace %s", cfg.ContainerID)
 
 	runtimeConfig := libcni.RuntimeConf{
 		ContainerID: cfg.ContainerID,
 		NetNS:       cfg.ContainerNetNS,
 	}
 
+	var delError error = nil
```
vsiddharth, line 127: nit: Why is this explicitly set to `nil`?

agent/ecscni/types_windows.go
```diff
@@ -29,8 +29,8 @@ const (
 	// ECSBridgeNetworkName is the name of the HNS network used as ecs-bridge.
 	ECSBridgeNetworkName = "nat"
 	// Constants for creating backoff while retrying setupNS.
-	setupNSBackoffMin      = time.Second * 2
-	setupNSBackoffMax      = time.Second * 6
+	setupNSBackoffMin      = time.Second * 4
```
vsiddharth, line 32: How did you come up with these values?

agent/ecscni/plugin.go
rawahars: This is the initial assignment of the return variable.
I would prefer to be explicit about the value being returned unless it is being changed due to error.

agent/ecscni/types_windows.go
rawahars: These values are based on practical observations of running multiple tasks with multiple containers in each task.

agent/ecscni/plugin.go
vsiddharth: `var delError error` would also result in `delError` being `nil`

rawahars: Yes it would but what's the issue in the overt initialization of the variable?

agent/ecscni/types_windows.go
vsiddharth: Thanks!

nit: the name `setupNSBackoffMin` being set to `time.Second * 4` is confusing.

rawahars: `setupNSBackoffMin` is the existing constant name being used in the agent.
We have just set that to this value on Windows.

agent/ecscni/plugin.go
vsiddharth: With golang, it is common practice to just have it as `var delError error`

agent/ecscni/plugin.go
rawahars: I believe the existing comments are sufficient to understand the interface. What additional comments are we talking about?

fenxiong (approved):

vsiddharth (approved):
🚀

PR #2869, opened 2021-05-17
EC2 Task Networking on Windows: Invoke additional commands inside pause container and on host 

rawahars:
### Summary

The CNI plugin creates the network plumbing as required but does not add/delete the route entries in the pause namespace. Moreover, if the instance metadata has to be blocked for the task, then we will have to create new firewall rules.
This PR accomplishes the following-

- Delete the default route of the ecs-bridge endpoint in the pause namespace
- Delete the subnet route of the ecs-bridge endpoint in the pause namespace
- Add the credentials endpoint to use the ecs-bridge endpoint
- If required, create a new firewall rule if one doesn't exist
- If a new firewall rule was created on task start, then delete that rule on cleanup

### Implementation details

New methods were added in `docker_task_engine_windows.go` and the same were called during provisioning/cleanup of the task namespace resources.

### Testing

The tests were validated using unit tests and a custom agent binary running on a EC2 instance.

New tests cover the changes: 
Yes

Invoking additional commands for awsvpc-network mode setup on Windows.
### Licensing


agent/engine/docker_task_engine_test.go
```diff
@@ -1035,6 +1035,15 @@ func TestProvisionContainerResourcesSetPausePIDInVolumeResources(t *testing.T) {
 		}, nil),
 		mockCNIClient.EXPECT().SetupNS(gomock.Any(), gomock.Any(), gomock.Any()).Return(nsResult, nil),
 	)
+	// These mock calls would be made only for Windows.
```
angelcar, line 1038: hmm.. this is Hacky, the test will succeed in Windows even if you remove the calls to these, right?

agent/engine/docker_task_engine_windows.go
```diff
@@ -58,3 +94,146 @@ func (engine *DockerTaskEngine) invokePluginsForContainer(task *apitask.Task, co
...
+	}
+
+	// Query the exec container to determine if the commands succeeded.
+	inspect, err := engine.client.InspectContainerExec(ctx, execRes.ID, dockerclient.ContainerExecInspectTimeout)
+	if err != nil {
+		seelog.Errorf("Task [%s]: Failed to execute command in pause namespace [inspect]: %v", task.Arn, err)
+		return err
+	}
+
+	// If the commands succeeded then return nil.
```
angelcar, line 203: nit: This comment should probably be in line 207

agent/engine/docker_task_engine_test.go
angelcar: Thinking about this, it probably would be better to extract `invokeCommandsForTaskNamespaceSetup` and `invokeCommandsForTaskNamespaceCleanup` outside of `DockerTaskEngine`. Perhaps we can create an interface like 
```golang
type TaskNamespaceHelper struct {
    NamespaseSetup() error
    NamespaceCleanup() error
}
```

This way the UTs will only validate that the high-level `NamespaseSetup` and `NamespaceCleanup` are called and the windows implementation UT can check the docker operations are called exactly once, whereas the Linux implementation will be empty.

Thoughts?

rawahars: I agree with you. I have moved these methods to a separate interface in `ecscni` package.

agent/engine/docker_task_engine.go
```diff
@@ -1430,6 +1434,31 @@ func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task,
 		task.SetLocalIPAddress(taskIP)
 		engine.saveTaskData(task)
 	}
+
+	// Invoke additional commands required to configure the task namespace routing.
+	err = engine.namespaceHelper.ConfigureTaskNamespaceRouting(engine.ctx, cniConfig, result)
```
angelcar, line 1439: I think this should be inside the `if result != nil {` block? otherwise you could pass nil to this function which would panic, I think. This should probably be covered in a UT, btw.

Also, I am assuming that if `result` is nil there is no point executing the other commands?

rawahars: result would never be nil as there are already checks in place inside `setupNS` method.
The earlier if condition was introduced last year since the feature was partially developed and result could have been nil. Now that it is completed, we can remove the if condition.

angelcar (approved):

agent/ecscni/namespace_helper_windows.go
```diff
@@ -0,0 +1,199 @@
...
+	// If the commands fail then return error.
+	if !inspect.Running && inspect.ExitCode != 0 {
+		return errors.Errorf("commands failed inside container %s namespace: %d", containerID, inspect.ExitCode)
+	}
+
+	return nil
+}
+
+// invokeCommandsOnHost invokes given commands on the host instance using the executeFn.
+func (nsHelper *helper) invokeCommandsOnHost(executeFn execCmdExecutorFn, commands []string, separator string) error {
```
fenxiong, line 177: why not store executeFn as a field of nsHelper? i think that way it's easier to mock it in test and you don't need to have executeFn as an argument like this

agent/engine/docker_task_engine.go
```diff
@@ -1714,3 +1747,13 @@ func (engine *DockerTaskEngine) getDockerID(task *apitask.Task, container *apico
 	}
 	return dockerContainer.DockerID, nil
 }
+
+// getTaskENI returns the primary eni of the task.
+func (engine *DockerTaskEngine) getTaskENI(task *apitask.Task) *apieni.ENI {
```
fenxiong, line 1752: it seems to me that this part of logic should better be in task.go instead of here. if the current implementation of [GetPrimaryENI](https://github.com/aws/amazon-ecs-agent/blob/2874131a5f3d1c0908e0f2ebad62e48b8ff5850f/agent/api/task/task.go#L2148) does not work for windows, can that method have an alternative impl for windows and then we just use task.GetPrimaryENI() in the engine?

agent/ecscni/namespace_helper_windows.go
rawahars: Yes that is functionally equivalent. Considering that if Linux implementation of the same is required sometime, we will have similar function signature, we can move it inside helper struct.
I'll make these changes.

agent/engine/docker_task_engine.go
rawahars: Yes that works for Windows as well. I'll make these changes.

fenxiong (approved):

vsiddharth (approved):

PR #2874, opened 2021-05-19
Give up trying to stop container if max timeouts are reached

angelcar:
### Summary
Docker sometimes deadlocks when trying to stop a container (see [PR](https://github.com/moby/moby/pull/41588)). This has caused a number of issues so far.

To provide a work around, we give up after a number of failed stopContainer attempts (timeouts). If the max attempts are reached, we mark the container as stopped anyways, allowing the task to continue its course instead of being stuck indefinitely.
It has been observed that when this problem occurs (happens sparsely but consistently), the container process actually stops, so this change should be safe.

### Implementation details
* Retry a max number of times with back off when stop container timeouts occur
* The relevant UT was explicitly ending the task in a hacky way. Had this hacky way been absent, we would have realized that repeated timeouts cause the task to hang for ever. That has been fixed by removing the hack. Not we just expect the task to end normally after the max attempts are reached.

### Testing

Simulated docker stop container deadlock in an EC2 instance and verified that the task transitions to `STOPPED` after 3 timed out attempts to stop the container. See the logs below for the sequence of events during the test.

```bash
level=info time=2021-05-20T01:53:24Z msg="Task engine [...]: stopping container [angelar-test-container]" module=docker_task_engine.go
.
.
.
level=warn time=2021-05-20T01:59:35Z msg="Unable to stop container after 3 attempts timed out; giving up and marking container as stopped" module=docker_task_engine.go
level=debug time=2021-05-20T01:59:35Z msg="Task engine [...]: transitioned container [angelar-test-container (Runtime ID: ...)] to [STOPPED]" module=docker_task_engine.go
.
.
.
level=info time=2021-05-20T01:59:35Z msg="Task engine [...]: stopping container [~internal~ecs~pause]" module=docker_task_engine.go
.
.
.
level=warn time=2021-05-20T02:07:08Z msg="Unable to stop container after 3 attempts timed out; giving up and marking container as stopped" module=docker_task_engine.go
level=debug time=2021-05-20T02:07:08Z msg="Task engine [...]: transitioned container [~internal~ecs~pause (Runtime ID: ...)] to [STOPPED]" module=docker_task_engine.go
.
.
.
level=info time=2021-05-20T02:07:08Z msg="Managed task has reached stopped; waiting for container cleanup" taskARN="..."
```

* Bug: Fixes a bug where a task can be stuck in `RUNNING` indefinitely when a container can't be stopped due to an unresolved docker [bug](https://github.com/moby/moby/issues/41587) (see also the open [PR](https://github.com/moby/moby/pull/41588) in moby to fix the bug).
### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -101,7 +101,15 @@ const (
 	FluentAWSVPCHostValue  = "127.0.0.1"
 
 	defaultMonitorExecAgentsInterval = 15 * time.Minute
+
+	stopContainerBackoffMin        = time.Second
+	stopContainerBackoffMax        = time.Second * 5
+	stopContainerBackoffJitter     = 0.2
+	stopContainerBackoffMultiplier = 1.3
+	stopContainerMaxRetryCount     = 3
 )
+
+var newExponentialBackoff = retry.NewExponentialBackoff
 
 // DockerTaskEngine is a state machine for managing a task and its containers
 // in ECS.
@@ -1535,7 +1543,29 @@ func (engine *DockerTaskEngine) stopContainer(task *apitask.Task, container *api
 		apiTimeoutStopContainer = engine.cfg.DockerStopTimeout
 	}
 
-	return engine.client.StopContainer(engine.ctx, dockerID, apiTimeoutStopContainer)
+	return engine.stopDockerContainer(dockerID, container.Name, apiTimeoutStopContainer)
+}
+
+// stopDockerContainer attempts to stop the container, retrying only in case of time out errors.
+// If the maximum number of retries is reached, the container is marked as stopped. This is because docker sometimes
+// deadlocks when trying to stop a container but the actual container process is stopped.
+// for more information, see: https://github.com/moby/moby/issues/41587
+func (engine *DockerTaskEngine) stopDockerContainer(dockerID, containerName string, apiTimeoutStopContainer time.Duration) dockerapi.DockerContainerMetadata {
+	var md dockerapi.DockerContainerMetadata
+	backoff := newExponentialBackoff(stopContainerBackoffMin, stopContainerBackoffMax, stopContainerBackoffJitter, stopContainerBackoffMultiplier)
+	for i := 0; i < stopContainerMaxRetryCount; i++ {
+		md = engine.client.StopContainer(engine.ctx, dockerID, apiTimeoutStopContainer)
+		if md.Error == nil || md.Error.ErrorName() != dockerapi.DockerTimeoutErrorName {
+			return md
+		}
+		if i < stopContainerMaxRetryCount-1 {
+			time.Sleep(backoff.Duration())
+		}
+	}
+	if md.Error != nil && md.Error.ErrorName() == dockerapi.DockerTimeoutErrorName {
+		seelog.Warnf("Unable to stop container (%s) after %d attempts that timed out; giving up and marking container as stopped anyways", containerName, stopContainerMaxRetryCount)
+	}
+	return md
 }
 
 func (engine *DockerTaskEngine) removeContainer(task *apitask.Task, container *apicontainer.Container) error {
--- a/agent/engine/docker_task_engine_test.go
+++ b/agent/engine/docker_task_engine_test.go
@@ -897,7 +897,6 @@ func TestTaskTransitionWhenStopContainerTimesout(t *testing.T) {
 			Duration:   30 * time.Second,
 		},
 	}
-	dockerEventSent := make(chan int)
 	for _, container := range sleepTask.Containers {
 		imageManager.EXPECT().AddAllImageStates(gomock.Any()).AnyTimes()
 		client.EXPECT().PullImage(gomock.Any(), container.Image, nil, gomock.Any()).Return(dockerapi.DockerContainerMetadata{})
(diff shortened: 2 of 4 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1535,7 +1543,33 @@ func (engine *DockerTaskEngine) stopContainer(task *apitask.Task, container *api
...
+	return engine.stopDockerContainer(dockerID, apiTimeoutStopContainer)
+}
+
+// stopDockerContainer attempts to stop the container, retrying only in case of time out errors.
+// If the maximum number of retries is reached, the container is marked as stopped. This is because docker sometimes
+// deadlocks when trying to stop a container but the actual container process is stopped.
+// for more information, see: https://github.com/moby/moby/issues/41587
+func (engine *DockerTaskEngine) stopDockerContainer(dockerID string, apiTimeoutStopContainer time.Duration) dockerapi.DockerContainerMetadata {
+	var md dockerapi.DockerContainerMetadata
+	backoff := newExponentialBackoff(stopContainerBackoffMin, stopContainerBackoffMax, stopContainerBackoffJitter, stopContainerBackoffMultiplier)
```
sparrc, line 1555: This is not some http endpoint subject to rate-limiting, and we already have something like a 2.5 minute timeout for the StopContainer call to succeed, do we really need an exponential backoff here? I would say just try it again immediately or sleep for no more than 15-30 seconds.

sharanyad: I would still keep the exponential backoff, so we don't issue too many calls to docker and slow it down even more.

sparrc: oh I just re-read the constants above and realize it caps at 5 seconds anyways, so I guess that's fine with me but still seems sort of pointless if the timeout is already 2.5 minutes. What's the point of backing off if the delay is orders of magnitude less than the timeout?

sharanyad: what is the 2.5 minute timeout? docker stop timeout is at 30s, don't think we have a timeout for the call from engine itself.

mythri-garaga (approved):

agent/engine/docker_task_engine.go
chienhanlin: Does 2.5 mins timeout come from `timeout + stopContainerTimeoutBuffer ` in docker client?
[https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/dockerclient/dockerapi/docker_client.go#L721-L722](https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/dockerclient/dockerapi/docker_client.go#L721-L722)

`stopContainerTimeoutBuffer` is defined as
[https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/dockerclient/dockerapi/docker_client.go#L92-L97](https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/dockerclient/dockerapi/docker_client.go#L92-L97)
    
and `timeout` comes from `apiTimeoutStopContainer` which is passed by docker task engine
[https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/engine/docker_task_engine.go#L1533-L1538](https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/engine/docker_task_engine.go#L1533-L1538)
    
cfg .`DockerStopTimeout` has the default value `defaultDockerStopTimeout`, which is set to 30 second.

[https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/config/config_unix.go#L50-L61](https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/config/config_unix.go#L50-L61)  
  
[https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/config/config.go#L67-L68](https://github.com/aws/amazon-ecs-agent/blob/9f2aa23fa01f4901d2f2eeece74db02564f3f59f/agent/config/config.go#L67-L68)

then timeout + stopContainerTimeoutBuffer  = 30 * time.Second  +   2 * time.Minute

agent/engine/docker_task_engine_test.go
```diff
@@ -918,27 +917,15 @@ func TestTaskTransitionWhenStopContainerTimesout(t *testing.T) {
 					}()
 				}).Return(dockerapi.DockerContainerMetadata{DockerID: containerID}),
 
-			// StopContainer times out
-			client.EXPECT().StopContainer(gomock.Any(), containerID, gomock.Any()).Return(containerStopTimeoutError),
-			// Since task is not in steady state, progressContainers causes
-			// another invocation of StopContainer. Return a timeout error
-			// for that as well.
-			client.EXPECT().StopContainer(gomock.Any(), containerID, gomock.Any()).Do(
```
sharanyad, line 926: the removed code still seems like a valid test case. why not add another test instead of replacing this?

angelcar: Actually the removed code is not valid anymore. That code was explicitly testing that there was no "container event" being emitted, but of course now there is one because this PR is allowing the container to transition to `STOPPED`. Is There a particular use-case that you think is not covered?

agent/engine/docker_task_engine.go
```diff
@@ -1535,7 +1543,33 @@ func (engine *DockerTaskEngine) stopContainer(task *apitask.Task, container *api
...
+		md = engine.client.StopContainer(engine.ctx, dockerID, apiTimeoutStopContainer)
+		if md.Error == nil || md.Error.ErrorName() != dockerapi.DockerTimeoutErrorName {
+			return md
+		}
+		if i < stopContainerMaxRetryCount-1 {
+			time.Sleep(backoff.Duration())
+		}
+	}
+	if md.Error != nil && md.Error.ErrorName() == dockerapi.DockerTimeoutErrorName {
+		seelog.Warnf("Unable to stop container (%s) after %d attempts that timed out; giving up and marking container as stopped anyways", containerName, stopContainerMaxRetryCount)
```
fierlion, line 1566: I do think it's the right choice to progress the task state, but I'd ask that we build out a mechanism to make this type of failure more discoverable, (not necessarily as part of the current PR).

While I agree that this change will unblock some, I fear that if we aren't being noisier about this we'll end up with instances that have abandoned containers lying around consuming resources.  And do we know whether a docker daemon that's caught in a deadlock will be fine servicing future tasks?

fierlion (dismissed):

fenxiong:
nit: log at https://github.com/aws/amazon-ecs-agent/pull/2874/files#diff-89cc968c48481ca275b4c7fa7d8e069a41d9785ec5847aa627c2e6b015cab1c4R954 seems incorrect to me. the state change is not ignored, instead the error is ignored and we force marking container as stopped. As we will fall back to that log with this change, i would suggest correct the log with something like "Error stopping container; marking container as stopped"

sharanyad (approved):

PR #2876, opened 2021-05-21
EC2 task networking on Windows: gMSA support, adding additional local routes and IMDS route in the task namespace.

rawahars:
### Summary

The changes introduced in this PR includes-

- gMSA support for tasks running in awsvpc network mode on Windows. 
- Support for adding additional local routes in task namespace using `ECS_AWSVPC_ADDITIONAL_LOCAL_ROUTES` environment variable
- Support for adding explicit IMDS route in task namespace if `ECS_AWSVPC_BLOCK_IMDS` is false
- Rectified a minor bug which was causing panic in the agent.
### Implementation details

The changes are implemented as -

- For gMSA support: We store the instance ENI's DNS server list using Win32 IpHelper API call. The same is used while creating CNI plugin configuration. This is possible because both instance ENI and task ENI belong to the same customer VPC.
- For Additional local routes: We create an add route command and append it to the existing list of commands to be executed inside pause namespace.
- Similarly IMDS explicit route is created.

### Testing

A custom binary was tested for gMSA, additional local routes and IMDS.

New tests cover the changes: 
Yes

Added gMSA support for tasks in awsvpc network mode and create additional local routes in task namespace.
### Licensing


angelcar (approved):

agent/ecscni/namespace_helper_windows.go
```diff
@@ -74,6 +76,26 @@ func (nsHelper *helper) ConfigureTaskNamespaceRouting(ctx context.Context, confi
...
+	if !config.BlockInstanceMetadata {
+		// This naming convention is drawn from the way CNI plugin names the endpoints.
+		taskPrimaryEndpointId := strings.Replace(strings.ToLower(taskENI.MacAddress), ":", "", -1)
+		taskPrimaryEndpointName := fmt.Sprintf(taskPrimaryEndpointNameFormat, taskPrimaryEndpointId, config.ContainerID)
+		imdsRouteAdditionCmd := fmt.Sprintf(ecsBridgeRouteAddCmdFormat, imdsEndpointIPAddress, taskPrimaryEndpointName)
+		commands = append(commands, imdsRouteAdditionCmd)
+	}
+
+	// Add any additional route which needs to be routed via ecs-bridge.
+	if len(config.AdditionalLocalRoutes) != 0 {
```
angelcar, line 88: nit: this check is not needed, `range` is smart :)

fenxiong: +1

fenxiong (approved):

agent/app/agent_windows.go
```diff
@@ -76,12 +76,12 @@ func (agent *ecsAgent) initializeTaskENIDependencies(state dockerstate.TaskEngin
 		return err, false
 	}
 
-	// Query the VPC's primary IPv4 CIDR using IMDS
-	primaryIPv4VPCCIDR, err := agent.ec2MetadataClient.PrimaryIPV4VPCCIDR(agent.mac)
+	dnsServerList, err := agent.resourceFields.NetworkUtils.GetDNSServerAddressList(agent.mac)
```
vsiddharth, line 79: Why is this error terminal for Task Networking?
Customers can opt to use this feature without AD/gMSA ?

agent/app/agent_windows.go
```diff
@@ -76,12 +76,12 @@ func (agent *ecsAgent) initializeTaskENIDependencies(state dockerstate.TaskEngin
...
 	if err != nil {
-		return fmt.Errorf("unable to get primary ipv4 cidr of the vpc: %v", err), false
+		// An error at this point is terminal as the tasks need the access to domain controllers.
+		return fmt.Errorf("unable to get dns server addresses of instance eni: %v", err), true
 	}
-	agent.cfg.PrimaryIPv4VPCCIDR = primaryIPv4VPCCIDR
+	agent.cfg.InstanceENIDNSServerList = dnsServerList
```
vsiddharth, line 84: Can we keep both `PrimaryIPv4VPCCIDR` and `dnsServerList` in config?

agent/ecscni/namespace_helper_windows.go
```diff
@@ -59,7 +61,7 @@ const (
 var execCmdExecutorFn execCmdExecutorFnType = execCmdExecutor
 
 // ConfigureTaskNamespaceRouting executes the commands required for setting up appropriate routing inside task namespace.
-func (nsHelper *helper) ConfigureTaskNamespaceRouting(ctx context.Context, config *Config, result *current.Result) error {
+func (nsHelper *helper) ConfigureTaskNamespaceRouting(ctx context.Context, taskENI *apieni.ENI, config *Config, result *current.Result) error {
```
vsiddharth, line 64: Can you please update the comment to indicate that these commands are executed within the pause container's context ??

agent/ecscni/namespace_helper_windows.go
```diff
@@ -74,6 +76,26 @@ func (nsHelper *helper) ConfigureTaskNamespaceRouting(ctx context.Context, confi
 	credentialsAddressRouteAdditionCmd := fmt.Sprintf(ecsBridgeRouteAddCmdFormat, credentialsEndpointRoute, ecsBridgeEndpointName)
 	commands := []string{defaultRouteDeletionCmd, defaultSubnetRouteDeletionCmd, credentialsAddressRouteAdditionCmd}
 
+	if !config.BlockInstanceMetadata {
+		// This naming convention is drawn from the way CNI plugin names the endpoints.
+		taskPrimaryEndpointId := strings.Replace(strings.ToLower(taskENI.MacAddress), ":", "", -1)
```
vsiddharth, line 81: Can you update this comment with a link to the CNI plugin?

agent/eni/networkutils/utils_windows.go
```diff
@@ -106,12 +109,59 @@ func (utils *networkUtils) retrieveMAC() (string, error) {
...
+// GetDNSServerAddressList returns the DNS server addresses of the queried interface.
+func (utils *networkUtils) GetDNSServerAddressList(macAddress string) ([]string, error) {
+	addresses, err := utils.netWrapper.GetAdapterAddresses()
+	if err != nil {
+		return nil, err
+	}
+
+	var firstDnsNode *windows.IpAdapterDnsServerAdapter
+	// Find the adapter which has the same mac as queried.
+	for _, adapterAddr := range addresses {
```
vsiddharth, line 131: Why not append the DNS server addresses to the return list directly?

agent/eni/netwrapper/netwrapper_windows.go
```diff
@@ -0,0 +1,80 @@
...
+
+	"golang.org/x/sys/windows"
+)
+
+// NetWrapper interface is created to abstract Golang's net package from its usage in the watcher
+// Also, this enables us to mock this interface for unit tests
+type NetWrapper interface {
+	FindInterfaceByIndex(int) (*net.Interface, error)
+	GetAllNetworkInterfaces() ([]net.Interface, error)
+	GetAdapterAddresses() ([]*windows.IpAdapterAddresses, error)
```
vsiddharth, line 32: Why do you need to return data using `*windows.IpAdapterAddresses`? 
Can they be converted to a more user-friendly format?

Also, why has this now become platform specific? Can you please keep it agnostic??

agent/app/agent_windows.go
rawahars: This error is terminal because the dns server addresses obtained here would be used to setup task networking as well.
On EC2, the task ENI and instance ENI are from the same VPC. Therefore, the DNS server address would always be the same for both. Hence, we store the DNSServer list from the instance ENI which intrinsically takes care of normal task networking as well as gMSA with task networking,

agent/app/agent_windows.go
rawahars: Both are not required. Since we are using DNS server list from the instance, therefore, primary VPC CIDR becomes redundant.

agent/eni/netwrapper/netwrapper_windows.go
rawahars: So this package was introduced in task networking feature to abstract Go's net package. `GetAdapterAddresses` method is copied from the Go's Windows package and the data obtained in this Win32 call can be parsed by the caller as needed.
So, for example, in Go lang's net package, Interface object is created using data returned from this method. Other data is discarded.
Similarly, we are using DNS part of the data returned by this method.

Since, this interface itself was for wrapping Windows related methods, it seems to be the apt place for it to go.

agent/eni/networkutils/utils_windows.go
rawahars: The DNS server list returned from Win32 API call is a raw linked list implementation whereas the data type expected by the caller is a Golang slice of strings.
We need to iterate each node of the list and parse the SocketAddress into a IP address and add them to the slice.

agent/ecscni/namespace_helper_windows.go
rawahars: Sure, I can do that. Since the CNI plugin has not been merged, therefore, we do not have a permanent link for that yet.
Would it do if we add the link in the next final PR?

agent/ecscni/namespace_helper_windows.go
rawahars: We mentioned in the comment that the commands are executed inside `task namespace`. Should we elaborate on that?

rawahars:
Decoupled CNI plugin build logic from this PR as it was making the entire PR complex.
Those changes will be part of another PR.

agent/app/agent_windows.go
```diff
@@ -76,12 +76,17 @@ func (agent *ecsAgent) initializeTaskENIDependencies(state dockerstate.TaskEngin
 		return err, false
 	}
 
-	// Query the VPC's primary IPv4 CIDR using IMDS
-	primaryIPv4VPCCIDR, err := agent.ec2MetadataClient.PrimaryIPV4VPCCIDR(agent.mac)
+	// The instance ENI and the task ENI on ECS EC2 Windows will have the same vpc, and therefore,
+	// the same DNS server list. Hence, we store the DNS server list of the instance ENI during
```
vsiddharth, line 80: nit: same VPC DNS server list...

Please expand AD -> Active Directory

nit: DNS entries will be added and not allocated

agent/app/agent_windows.go
```diff
@@ -76,12 +76,17 @@ func (agent *ecsAgent) initializeTaskENIDependencies(state dockerstate.TaskEngin
...
+	// agent startup and use the same during config creation for setting up task ENI.
+	// Another intrinsic benefit of this approach is that any DNS servers added for AD
+	// will be allocated to task ENI, allowing tasks in awsvpc network mode to support gMSA.
+	dnsServerList, err := agent.resourceFields.NetworkUtils.GetDNSServerAddressList(agent.mac)
 	if err != nil {
-		return fmt.Errorf("unable to get primary ipv4 cidr of the vpc: %v", err), false
+		// An error at this point is terminal as the tasks need the access to DNS servers.
```
vsiddharth, line 86: nit: tasks launched with awsvpc mode require the DNS entries...

agent/config/types.go
```diff
@@ -334,7 +333,7 @@ type Config struct {
 	// External specifies whether agent is running on external compute capacity (i.e. outside of aws).
 	External BooleanDefaultFalse
 
-	// PrimaryIPv4VPCCIDR stores the primary IPv4 CIDR of the VPC in which agent is running
-	// Currently, this field is only populated for Windows and is used during task networking setup
-	PrimaryIPv4VPCCIDR *net.IPNet
+	// InstanceENIDNSServerList stores the list of dns servers for the primary instance ENI.
```
vsiddharth, line 336: nit: dns -> DNS

agent/ecscni/namespace_helper_windows.go
```diff
@@ -59,7 +61,13 @@ const (
 var execCmdExecutorFn execCmdExecutorFnType = execCmdExecutor
 
 // ConfigureTaskNamespaceRouting executes the commands required for setting up appropriate routing inside task namespace.
-func (nsHelper *helper) ConfigureTaskNamespaceRouting(ctx context.Context, config *Config, result *current.Result) error {
+// The commands currently executed are-
```
vsiddharth, line 64: These are super helpful 👍

agent/ecscni/plugin_windows.go
```diff
@@ -37,8 +37,9 @@ import (
 
 var (
 	// vpcCNIPluginPath is the path of the cni plugin's log file.
+	// The default value of log file is C:\ProgramData\Amazon\ECS\log\cni\vpc-eni.log
 	vpcCNIPluginPath = filepath.Join(utils.DefaultIfBlank(os.Getenv("ProgramData"), `C:\ProgramData`),
-		"Amazon", "ECS", "log", "vpc-eni.log")
+		"Amazon", "ECS", "log", "cni", "vpc-eni.log")
```
vsiddharth, line 42: nit: this can be simplified to 
```filepath.Join(utils.DefaultIfBlank(os.Getenv("ProgramData"), `C:\ProgramData`),
		"Amazon/ECS/log/cni/vpc-eni.log")```

**Review by vsiddharth — COMMENTED** [review]
Changes look good. Left a few minor nits and suggestions. 

qq: What happens if the agent starts on an instance before the container instance is part of any valid Active Directory setup? Will AD/gMSA and task networking work in unison?

**Comment by rawahars** [comment]
> Changes look good. Left a few minor nits and suggestions.
> 
> qq: What happens if the agent starts on an instance before the container instance is part of any valid Active Directory setup? Will AD/gMSA and task networking work in unison?

Whenever the instance is domain joined, the instance is restarted and therefore, the agent on startup will have the updated DNS server list.

**Review by vsiddharth — APPROVED** [review]
🚀

File: agent/ecscni/types_windows.go [hunk 13]
```diff
@@ -50,11 +52,9 @@ type VPCENIPluginConfig struct {
 	// ENIMACAddress is the MAC address of the eni.
 	ENIMACAddress string `json:"eniMACAddress"`
 	// ENIIPAddress is the is the ipv4 of eni.
-	ENIIPAddress string `json:"eniIPAddress"`
+	ENIIPAddress []string `json:"eniIPAddresses"`
```
fenxiong, line 55: i think this should be `ENIIPAddresses` and below should be `GatewayIPAddresses`?

fenxiong (approved):

angelcar (approved):

PR #2884, opened 2021-05-26
Instance health docker health check

shubham2892:
### Summary
This PR adds the docker ping health check.

### Implementation details
- `dockerRuntimeHealthCheck` struct implements the `Healthcheck` interface. `RunCheck` method in the interface calls the `SystemPing` method. This method calls the Docker ping API. If the Docker ping call is successful, it is treated as an indication that Docker is healthy.   
- `Healthcheck` interface is moved to Doctor.go and new methods are added to get current and last health check status. 

### Testing

Unit tests have been added for `SystemPing` method and docker ping healthcheck methods.

### Licensing


agent/doctor/docker_ping_healthcheck.go
```diff
@@ -0,0 +1,88 @@
...
+// License is located at
+//
+//      http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package doctor
```
fierlion, line 14: nit: can we call this file `docker_runtime_healthcheck.go`

and 

nit: by convention `Healthcheck` is one word (not `HealthCheck`) -- so lower-case `c`
There are a few of these below...  should always be either `Healthcheck` or `healthcheck`

agent/acs/handler/acs_handler.go
```diff
@@ -335,7 +335,6 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 
 	client.AddRequestHandler(payloadHandler.handlerFunc())
 
-	// Add HeartbeatHandler to acknowledge ACS heartbeats
```
fierlion, line 338: I think this comment is still relevant, no?
We should end up with a couple of heartbeat handlers:
one to ack ACS heartbeats and one to trigger the healthchecks (the latter handler is still TODO)

agent/doctor/doctor.go
```diff
@@ -37,14 +38,14 @@ func NewDoctor(healthchecks []Healthcheck) (*Doctor, error) {
...
 	defer doc.lock.Unlock()
-
 	doc.healthchecks = append(doc.healthchecks, healthcheck)
 }
 
 func (doc *Doctor) RunHealthchecks() bool {
 	doc.lock.RLock()
 	defer doc.lock.RUnlock()
-	allChecksResult := []bool{}
+	allChecksResult := []HealthcheckStatus{}
```
fierlion, line 47: thank you for this.

agent/doctor/doctor.go
```diff
@@ -53,10 +54,19 @@ func (doc *Doctor) RunHealthchecks() bool {
...
+func (doc *Doctor) allRight(checksResult []HealthcheckStatus) bool {
 	overallResult := true
 	for _, checkResult := range checksResult {
-		overallResult = overallResult && checkResult
+		overallResult = overallResult && checkResult.Ok()
 	}
 	return overallResult
 }
+
+type Healthcheck interface {
```
fierlion, line 65: I like that you pulled the Healthcheck into the doctor.go file

agent/acs/handler/acs_handler.go
shubham2892: Are we not already acking ACS heartbeats here - https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/heartbeat_handler.go#L86 🤔

agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -1008,6 +1008,34 @@ func TestDockerVersion(t *testing.T) {
 	assert.Equal(t, "1.6.0", str, "Got unexpected version string: "+str)
 }
 
+func TestSystemPing(t *testing.T) {
```
sharanyad, line 1011: this and below could be combined to table driven tests

agent/doctor/docker_runtime_healthcheck.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"context"
+	"sync"
+	"time"
+
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
+	"github.com/cihub/seelog"
+)
+
+type dockerRuntimeHealthcheck struct {
```
sharanyad, line 25: struct needs formatting. also could we move the lock to the bottom? and call it `lock` just for consistency?

agent/doctor/docker_runtime_healthcheck.go
```diff
@@ -0,0 +1,88 @@
...
+func NewDockerRuntimeHealthcheck(client dockerapi.DockerClient) *dockerRuntimeHealthcheck {
+	return &dockerRuntimeHealthcheck{
+		client:    client,
+		Status:    HealthcheckStatusInitializing,
+		TimeStamp: time.Now(),
+	}
+}
+
+func (dhc *dockerRuntimeHealthcheck) RunCheck() HealthcheckStatus {
+	res := dhc.client.SystemPing(context.TODO(), time.Second*2)
```
sharanyad, line 47: please make timeout a constant. could you also add a `TODO` to pass in context as an argument to the method?

agent/doctor/docker_runtime_healthcheck.go
```diff
@@ -0,0 +1,88 @@
...
+	healthcheck := HealthcheckStatusOk
+	if res.Error != nil {
+		seelog.Infof("Docker Ping failed with error: %v", res.Error)
```
sharanyad, line 50: `Warnf`? also could prefix it with `[DockerHealthcheck]` for readability

agent/doctor/docker_runtime_healthcheck.go
```diff
@@ -0,0 +1,88 @@
...
+	"context"
+	"sync"
+	"time"
+
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
+	"github.com/cihub/seelog"
+)
+
+type dockerRuntimeHealthcheck struct {
+	mu sync.RWMutex
```
angelcar, line 26: remind me what uses this struct concurrently that we need a `Mutex`?

shubham2892: @angelcar This is currently not used concurrently, this struct will be written by a healthcheck and read by a queue, such that there might be a race condition. To be added in coming PRs.

fierlion: there was a comment in a previous review requiring these mutexes.  Since there's only one doctor per agent, they're likely not going to be necessary but as previous comments noted -- it's better to err on the side of caution.

agent/acs/handler/acs_handler.go
fierlion: Currently the handler exists to ack ACS heartbeats.  
We have a TODO to separate the ACS ack handler from the healthcheck handler (which is also triggered by ACS heartbeat messages)

fierlion (approved):

agent/doctor/docker_runtime_healthcheck.go
angelcar: I guess we will see if we end up using it... but when it comes to concurrency, avoiding the usage of mutexes is often a more cautious choice than using them when they're not really needed.

agent/dockerclient/dockerapi/docker_client_test.go
shubham2892: For these two I will prefer to keep separate because I will have to add custom logic anyway for each test for asserting error and nil response.

agent/doctor/docker_runtime_healthcheck.go
shubham2892: Good idea, will add.

shubham2892: Docker ping failing would mean Docker is completely unresponsive, It would be useful to surface this log at info level.

agent/doctor/docker_runtime_healthcheck.go
shubham2892: Static check is failing on formatting it manually.

angelcar (approved):

PR #2892, opened 2021-06-08
Generate PublishInstanceStatus message and import tacs model

fierlion:
### Summary
Add `GetPublishInstanceStatusMessage` and necessary wiring to fill up these messages.  It also imports the latest tacs model changes specific to instance-health.

### Implementation details
`GetPublishInstanceStatusMessage` uses a helper function `getInstanceStatuses` to collect the individual statuses.  If there are no instance statuses available, the helper will return nil and `GetPublishInstanceStatusMessage will return an error.
The tacs model changes were generated from the control-plane model using an automated process.  The actual code in agent was built with `make gogenerate`.

A sample message from the agent logs during manual testing:
```
{
    Metadata: {
        Cluster: "execDemo",
        ContainerInstance: "arn:aws:ecs:us-west-2:<accountID>:container-instance/execDemo/<arn>",
        RequestId: "d0bf7597-ee7c-4fce-b8e7-3fce648a5db6"
    },
    StatusChecks: [
        {
            LastStatusChange: 2021-06-0722: 50: 22.297344909+0000UTCm=+29.129781563,
            LastUpdated: 2021-06-0722: 49: 53.722331035+0000UTCm=+0.554767847,
            Status: "OK",
            Type: "ContainerRuntime"
        }
    ],
    Timestamp: 2021-06-0722: 50: 22.297377973+0000UTCm=+29.129814760
}
```

### Testing
Tested on an ECS Optimized AL2 instance with the commented-out docker runtime heartbeat trigger code enabled.  I was able to see messages generated for each heartbeat.  
Since the metadata has the cluster, I also tested statefulness by stopping the agent and restarting to see that it continues to print the correct cluster.  I then stopped the ecs-agent and deregistered the container instance, and deleted the state file.  I then changed the ecs.config to point to a new cluster.  When I restarted the agent, the PublishInstanceStatus messages' metadata reflected the cluster update.

Add GetPublishInstanceStatusMessage and import tacs model.

### Licensing

### DIFF
--- a/agent/acs/handler/heartbeat_handler.go
+++ b/agent/acs/handler/heartbeat_handler.go
@@ -17,8 +17,8 @@ import (
 	"context"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
 	"github.com/aws/amazon-ecs-agent/agent/doctor"
-	"github.com/aws/amazon-ecs-agent/agent/engine"
 	"github.com/aws/amazon-ecs-agent/agent/wsclient"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
@@ -31,13 +31,16 @@ type heartbeatHandler struct {
 	ctx                       context.Context
 	cancel                    context.CancelFunc
 	acsClient                 wsclient.ClientServer
-	taskEngine                engine.TaskEngine
 	doctor                    *doctor.Doctor
 }
 
 // newHeartbeatHandler returns an instance of the heartbeatHandler struct
-func newHeartbeatHandler(ctx context.Context, acsClient wsclient.ClientServer, taskEngine engine.TaskEngine) heartbeatHandler {
-	doctor := &doctor.Doctor{}
+func newHeartbeatHandler(ctx context.Context, acsClient wsclient.ClientServer, dockerClient dockerapi.DockerClient, cluster string, containerInstanceArn string) heartbeatHandler {
+	// set up baseline healthchecks in new doctor.
+	var healthchecks []doctor.Healthcheck
+	dockerHealthcheck := doctor.NewDockerRuntimeHealthcheck(dockerClient)
+	healthchecks = append(healthchecks, dockerHealthcheck)
+	heartbeatDoctor, _ := doctor.NewDoctor(healthchecks, cluster, containerInstanceArn)
 
 	// Create a cancelable context from the parent context
 	derivedContext, cancel := context.WithCancel(ctx)
@@ -47,8 +50,7 @@ func newHeartbeatHandler(ctx context.Context, acsClient wsclient.ClientServer, t
 		ctx:                       derivedContext,
 		cancel:                    cancel,
 		acsClient:                 acsClient,
-		taskEngine:                taskEngine,
-		doctor:                    doctor,
+		doctor:                    heartbeatDoctor,
 	}
 }
 
@@ -80,16 +82,17 @@ func (heartbeatHandler *heartbeatHandler) handleHeartbeatMessage() {
 
 func (heartbeatHandler *heartbeatHandler) handleSingleHeartbeatMessage(message *ecsacs.HeartbeatMessage) error {
 	// TestHandlerDoesntLeakGoroutines unit test is failing because of this section
-	// This should all move under a separate healthcheck handler
 
 	// Agent will run healthchecks triggered by ACS heartbeat
-	// healthcheck results will be sent on to TACS
-	//dockerHealthcheck := doctor.NewDockerRuntimeHealthcheck(heartbeatHandler.taskEngine.TaskEngineClient())
-	//heartbeatHandler.doctor.AddHealthcheck(dockerHealthcheck)
-	//go func() {
-	//	checkResult := heartbeatHandler.doctor.RunHealthchecks()
-	//	seelog.Debugf("handler healthcheck result: %v", checkResult)
-	//}()
+	// healthcheck results will be sent on to TACS, but for now just to debug logs.
+	go func() {
+		heartbeatHandler.doctor.RunHealthchecks()
+		requestMessage, err := heartbeatHandler.doctor.GetPublishInstanceStatusRequest()
+		if err != nil {
+			seelog.Errorf("Unable to get PublishInstanceStatusRequest message: %v", err)
+		}
+		seelog.Debugf("message to send: %v", requestMessage)
+	}()
 
 	// Agent will send simple ack to the heartbeatAckMessageBuffer
 	go func() {
--- a/agent/doctor/docker_runtime_healthcheck.go
+++ b/agent/doctor/docker_runtime_healthcheck.go
@@ -25,45 +25,68 @@ import (
 const systemPingTimeout = time.Second * 2
 
 type dockerRuntimeHealthcheck struct {
+	// HealthcheckType is the reported healthcheck type
+	HealthcheckType string `json:"HealthcheckType,omitempty"`
 	// Status is the container health status
-	Status HealthcheckStatus `json:"Status,omitempty"`
+	Status HealthcheckStatus `json:"HealthcheckStatus,omitempty"`
 	// Timestamp is the timestamp when container health status changed
 	TimeStamp time.Time `json:"TimeStamp,omitempty"`
+	// StatusChangeTime is the latest time the health status changed
+	StatusChangeTime time.Time `json:"StatusChangeTime,omitempty"`
+
 	// LastStatus is the last container health status
-	LastStatus HealthcheckStatus `json:"lastStatus,omitempty"`
+	LastStatus HealthcheckStatus `json:"LastStatus,omitempty"`
 	// LastTimeStamp is the timestamp of last container health status
 	LastTimeStamp time.Time `json:"LastTimeStamp,omitempty"`
-	client        dockerapi.DockerClient
-	lock          sync.RWMutex
+
+	client dockerapi.DockerClient
+	lock   sync.RWMutex
 }
 
 func NewDockerRuntimeHealthcheck(client dockerapi.DockerClient) *dockerRuntimeHealthcheck {
+	nowTime := time.Now()
 	return &dockerRuntimeHealthcheck{
-		client:    client,
-		Status:    HealthcheckStatusInitializing,
-		TimeStamp: time.Now(),
+		HealthcheckType:  HealthcheckTypeContainerRuntime,
+		Status:           HealthcheckStatusInitializing,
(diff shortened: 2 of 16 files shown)


agent/doctor/docker_runtime_healthcheck.go
```diff
@@ -60,10 +69,24 @@ func (dhc *dockerRuntimeHealthcheck) RunCheck() HealthcheckStatus {
...
-	dhc.TimeStamp = time.Now()
+
+	// update latest status
+	dhc.Status = healthStatus
+	dhc.TimeStamp = nowTime
+}
+
+func (dhc *dockerRuntimeHealthcheck) GetHealthcheckType() string {
+	dhc.lock.RLock()
+	defer dhc.lock.RUnlock()
```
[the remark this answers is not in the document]
fierlion: I like the Read lock construct and the `defer` command. (But I think it's not so very golang to rely so heavily on mutexes.)

agent/acs/handler/heartbeat_handler.go
```diff
@@ -36,8 +36,13 @@ type heartbeatHandler struct {
...
 // newHeartbeatHandler returns an instance of the heartbeatHandler struct
-func newHeartbeatHandler(ctx context.Context, acsClient wsclient.ClientServer, taskEngine engine.TaskEngine) heartbeatHandler {
-	doctor := &doctor.Doctor{}
+func newHeartbeatHandler(ctx context.Context, acsClient wsclient.ClientServer, taskEngine engine.TaskEngine, cluster string, containerInstanceArn string) heartbeatHandler {
+	// set up baseline healthchecks in new doctor.
+	var healthchecks []doctor.Healthcheck
+	// Uncomment the following 2 lines to run dockerRuntime healthchecks
+	//dockerHealthcheck := doctor.NewDockerRuntimeHealthcheck(taskEngine.TaskEngineClient())
+	//healthchecks = append(healthchecks, dockerHealthcheck)
+	heartbeatDoctor, _ := doctor.NewDoctor(healthchecks, cluster, containerInstanceArn)
```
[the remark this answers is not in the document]
fierlion: Potentially, yes.  This handler still needs work -- we'll probably need to create a separate handler for the status checks and presently we're failing the "leaksGoroutine" test when we enable the checks in their goroutine -- but we want the checks to be non-blocking, so we'll need to get to the bottom of the failing test.

You can try it out -- un-comment the lines and then run `make test`.

agent/doctor/docker_runtime_healthcheck.go
```diff
@@ -25,23 +25,32 @@ import (
...
 func NewDockerRuntimeHealthcheck(client dockerapi.DockerClient) *dockerRuntimeHealthcheck {
+	nowTime := time.Now()
 	return &dockerRuntimeHealthcheck{
-		client:    client,
-		Status:    HealthcheckStatusInitializing,
-		TimeStamp: time.Now(),
+		HealthcheckType:  HealthcheckTypeContainerRuntime,
+		Status:           HealthcheckStatusInitializing,
+		TimeStamp:        nowTime,
+		StatusChangeTime: nowTime,
```
shubham2892, line 52: Should we set `StatusChangeTime` to nil the first time ?

agent/doctor/doctor.go
```diff
@@ -54,19 +68,49 @@ func (doc *Doctor) RunHealthchecks() bool {
...
+	instanceStatuses := doc.getInstanceStatuses()
+
+	if instanceStatuses != nil {
+		return &ecstcs.PublishInstanceStatusRequest{
+			Metadata:  metadata,
+			Statuses:  instanceStatuses,
+			Timestamp: aws.Time(time.Now()),
+		}, nil
+	} else {
+		return nil, EmptyHealthcheckError
```
shubham2892, line 88: Will an error be returned when Agent just started and there are no health status to send ?

agent/doctor/doctor.go
```diff
@@ -54,19 +68,49 @@ func (doc *Doctor) RunHealthchecks() bool {
...
+			LastUpdated:      aws.Time(healthcheck.GetLastHealthcheckTime()),
+			Status:           aws.String(healthcheck.GetHealthcheckStatus().String()),
+			Type:             aws.String(healthcheck.GetHealthcheckType()),
+		}
+		instanceStatuses = append(instanceStatuses, instanceStatus)
+	}
+
+	if instanceStatuses == nil {
+		return nil
+	}
```
shubham2892, line 106: this code can be omitted.

agent/doctor/doctor.go
```diff
@@ -54,19 +68,49 @@ func (doc *Doctor) RunHealthchecks() bool {
 	return doc.allRight(allChecksResult)
 }
 
+// GetPublishInstanceStatusRequest will get all healthcheck statuses and generate
+// a sendable PublishInstanceStatusRequest
+func (doc *Doctor) GetPublishInstanceStatusRequest() (*ecstcs.PublishInstanceStatusRequest, error) {
+	metadata := &ecstcs.InstanceStatusMetadata{
+		Cluster:           aws.String(doc.cluster),
+		ContainerInstance: aws.String(doc.containerInstanceArn),
+		RequestId:         aws.String(uuid.NewRandom().String()),
```
shubham2892, line 77: Should we log this request ID if it will help in tracking the request ?

agent/doctor/docker_runtime_healthcheck.go
fierlion: This is assuming that the initial status is a status change from nothing to HealthcheckStatusInitializing.  As a customer I would expect the status change time to be updated any time the status changes -- including the case of no status -> some status.

agent/doctor/doctor.go
fierlion: will do.

agent/doctor/doctor.go
fierlion: update: `RunHealthchecks` will always run all the available healthchecks.  If any of the checks fails for any reason, the impaired status will reflect the failure.  The return value is the overall health: 
`return doc.allRight(allChecksResult)`

So if we don't have any instance statuses, then this indicates that the healthcheck(s) has/have not been run for some reason and it should result in an error as we always precede the message generation with `RunHealthchecks()`

After the first run of the healthchecks, the worst case will be that we end up sending messages with stale/duplicate status data (including timestamp).  In the message sending changes, we should add handling for this case -- marking a specific check as `sent` once its message has been passed on to TACS.

agent/doctor/docker_runtime_healthcheck_test.go
```diff
@@ -69,3 +70,30 @@ func TestSetHealthCheckStatus(t *testing.T) {
...
+	initializationChangeTime := dockerRuntimeHealthcheck.GetStatusChangeTime()
+
+	// we update to initializing again; our StatusChangeTime remains the same
+	dockerRuntimeHealthcheck.SetHealthcheckStatus(HealthcheckStatusInitializing)
+	updateChangeTime := dockerRuntimeHealthcheck.GetStatusChangeTime()
+	assert.Equal(t, HealthcheckStatusInitializing, dockerRuntimeHealthcheck.Status)
+	assert.Equal(t, initializationChangeTime, updateChangeTime)
+
+	// add a sleep so we know time has elapsed between the initial status and status change time
+	time.Sleep(50 * time.Millisecond)
```
angelcar, line 91: nit: 1ms should suffice here :)

agent/doctor/doctor.go
```diff
@@ -18,16 +18,30 @@ import (
...
+	"github.com/pborman/uuid"
+	"github.com/pkg/errors"
+
+	"github.com/aws/amazon-ecs-agent/agent/tcs/model/ecstcs"
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+var (
+	// EmptyHealthcheckError indicates an error when there are no healthcheck metrics to report
+	EmptyHealthcheckError = errors.New("doctor: no instance healthcheck status metrics to report")
```
angelcar, line 30: nit: The "doctor" part should probably be up to whatever logs this error

agent/doctor/doctor.go
fierlion: this will result in some verbose logging, even at the debug level.  And we can use the timestamp in a similar way to troubleshoot.

agent/acs/handler/heartbeat_handler.go
```diff
@@ -80,16 +85,15 @@ func (heartbeatHandler *heartbeatHandler) handleHeartbeatMessage() {
...
-	//dockerHealthcheck := doctor.NewDockerRuntimeHealthcheck(heartbeatHandler.taskEngine.TaskEngineClient())
-	//heartbeatHandler.doctor.AddHealthcheck(dockerHealthcheck)
-	//go func() {
-	//	checkResult := heartbeatHandler.doctor.RunHealthchecks()
-	//	seelog.Debugf("handler healthcheck result: %v", checkResult)
-	//}()
+	// healthcheck results will be sent on to TACS, but for now just to debug logs.
+	go func() {
+		heartbeatHandler.doctor.RunHealthchecks()
+		requestMessage, _ := heartbeatHandler.doctor.GetPublishInstanceStatusRequest()
```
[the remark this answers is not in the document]
fierlion: That said... I'm working on this fix now because it's blocking the doctor's integ test.

agent/doctor/doctor.go
shubham2892: Yes they should never be empty and we should not see this error ever.

shubham2892 (approved):
🚀

agent/acs/handler/heartbeat_handler.go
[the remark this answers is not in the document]
fierlion: updated in latest revision.  Now the docker api client is injected into the acs handler.

agent/acs/handler/heartbeat_handler.go
```diff
@@ -80,16 +82,17 @@ func (heartbeatHandler *heartbeatHandler) handleHeartbeatMessage() {
...
-	//go func() {
-	//	checkResult := heartbeatHandler.doctor.RunHealthchecks()
-	//	seelog.Debugf("handler healthcheck result: %v", checkResult)
-	//}()
+	// healthcheck results will be sent on to TACS, but for now just to debug logs.
+	go func() {
+		heartbeatHandler.doctor.RunHealthchecks()
+		requestMessage, err := heartbeatHandler.doctor.GetPublishInstanceStatusRequest()
+		if err != nil {
+			seelog.Errorf("Unable to get PublishInstanceStatusRequest message: %v", err)
```
angelcar, line 92: nit: should we return here? can be done in a different PR

angelcar (approved):

agent/acs/handler/heartbeat_handler.go
fierlion: I'll add this in the integ test pr that's incoming.

PR #2903, opened 2021-06-17
ECS EC2 Windows task networking: Removed redundant checks for DNS server information

rawahars:
### Summary

This PR removes redundant checks related to DNS server addresses while building CNI plugin config.
We should always use the DNS server addresses of the instance ENI. This is because-

- Both instance and task ENI belong in the same VPC, and therefore have same information.
- During Active Directory domain join, the additional DNS servers would be added to the task ENI, thus enabling it to use gMSA.

### Implementation details

Removed additional checks.
### Testing

A custom agent binary was tested with the updated changes.

New tests cover the changes: 
No new changes are introduced.

Removed redundant checks for DNS server addresses during CNI plugin config creation.
### Licensing

### DIFF
--- a/agent/ecscni/netconfig_windows.go
+++ b/agent/ecscni/netconfig_windows.go
@@ -25,12 +25,10 @@ import (
 
 // NewVPCENIPluginConfigForTaskNSSetup is used to create the configuration of vpc-eni plugin for task namespace setup.
 func NewVPCENIPluginConfigForTaskNSSetup(eni *eni.ENI, cfg *Config) (*libcni.NetworkConfig, error) {
+	// Use the DNS server addresses of the instance ENI it would belong in the same VPC as
+	// the task ENI and therefore, have same DNS configuration.
 	dns := types.DNS{
-		Nameservers: eni.DomainNameServers,
-	}
-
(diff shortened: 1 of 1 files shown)


angelcar (approved):

singholt (approved):
looks good, please look into the integ test failing on Windows 2016.

agent/ecscni/netconfig_windows.go
```diff
@@ -25,12 +25,10 @@ import (
 
 // NewVPCENIPluginConfigForTaskNSSetup is used to create the configuration of vpc-eni plugin for task namespace setup.
 func NewVPCENIPluginConfigForTaskNSSetup(eni *eni.ENI, cfg *Config) (*libcni.NetworkConfig, error) {
+	// Use the DNS server addresses of the instance ENI it would belong in the same VPC as
+	// the task ENI and therefore, have same DNS configuration.
```
fenxiong, line 29: what happen when the dhcp options of the vpc changes between agent start and task start? can we end up having different dns servers between `cfg. InstanceENIDNSServerList` and `eni.DomainNameServers ` in that case?

rawahars: DHCP would be enabled for the task namespace, so the appropriate DNS configuration would be updated automatically.

fenxiong (approved):

PR #2909, opened 2021-06-17
execcmd manager init task: Windows functionality added

arun-annamalai:
### Summary

This pull request implements the ECS agent init_task functionality so that the SSM agent has the necessary bind mounts and binaries to run. 

### Implementation details

The common variables are pulled into the manager_init_task.go file and then the Linux code is kept in its file. The windows file is where the bulk of the new implementation is. Mainly, Windows stores its configs in a different way since it cannot bind mount files (This is a [limitation](https://github.com/moby/moby/issues/30555) of docker windows). In addition, we need to bind mount the SSM plugins since the windows agent needs those. Because the dirs are bind mounted, they will be cached on a directory level (by concatenating the contents of the dir and hashing them), on Linux this is done in a file level. Lastly, when mounting in Windows host file paths/folders are not automatically created by Docker so they need to be manually created before mounting. 

### Testing

unit tests

integ tests will be in following PR
### Licensing

### DIFF
--- a/agent/engine/execcmd/manager_init_task_linux.go
+++ b/agent/engine/execcmd/manager_init_task_linux.go
@@ -15,26 +15,14 @@
 package execcmd
 
 import (
-	"crypto/sha256"
-	"encoding/base64"
 	"fmt"
-	"io/ioutil"
-	"os"
 	"path/filepath"
-	"sort"
-	"strconv"
 	"strings"
 
 	dockercontainer "github.com/docker/docker/api/types/container"
-	"github.com/pborman/uuid"
-
-	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
-	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 )
 
 const (
-	namelessContainerPrefix = "nameless-container-"
-
 	ecsAgentExecDepsDir = "/managed-agents/execute-command"
 
 	// ecsAgentDepsBinDir is the directory where ECS Agent will read versions of SSM agent
@@ -42,10 +30,6 @@ const (
 	ecsAgentDepsCertsDir = ecsAgentExecDepsDir + "/certs"
 
 	ContainerDepsDirPrefix = "/ecs-execute-command-"
-
-	// filePerm is the permission for the exec agent config file.
-	filePerm            = 0644
-	defaultSessionLimit = 2
 
 	SSMAgentBinName       = "amazon-ssm-agent"
 	SSMAgentWorkerBinName = "ssm-agent-worker"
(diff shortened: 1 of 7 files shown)


agent/engine/execcmd/manager_init_task_windows.go
```diff
@@ -16,18 +16,209 @@
 package execcmd
 
 import (
```
ubhattacharjya, line 18: Should unit tests and integ tests be included for these changes?

arun-annamalai: yes, I will add them in now

agent/engine/execcmd/manager_init_task_windows.go
```diff
@@ -16,18 +16,214 @@
...
+    </outputs>
+    <formats>
+        <format id="fmterror" format="%Date %Time %LEVEL [%FuncShort @ %File.%Line] %Msg%n"/>
+        <format id="fmtdebug" format="%Date %Time %LEVEL [%FuncShort @ %File.%Line] %Msg%n"/>
+        <format id="fmtinfo" format="%Date %Time %LEVEL %Msg%n"/>
+    </formats>
+</seelog>`
+)
+
+func (m *manager) InitializeContainer(taskId string, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig) (rErr error) {
```
angelcar, line 99: I'd prefer not to fork this whole method as it has many commonalities with linux. Instead we should probably just add a sub-method called `addRequiredBindMounts` or something similar. That way we can preserve a single method (i.e. reuse as much as possible) while still being able to configure windows in a windows-specific way

arun-annamalai: will refactor and push

agent/engine/execcmd/manager_init_task_linux.go
```diff
@@ -107,160 +75,89 @@ var (
...
-	limit, err := strconv.Atoi(limitStr)
-	if err != nil { // This means ACS send a limit that can't be converted to an int
-		return limit
-	}
-	if limit <= 0 {
-		limit = defaultSessionLimit
-	}
-	return limit
-}
+//func (m *manager) InitializeContainer(taskId string, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig) (rErr error) {
```
angelcar, line 84: are we ready to remove this?

agent/engine/execcmd/manager_init_task_windows.go
```diff
@@ -16,18 +16,236 @@
...
+        <format id="fmtdebug" format="%Date %Time %LEVEL [%FuncShort @ %File.%Line] %Msg%n"/>
+        <format id="fmtinfo" format="%Date %Time %LEVEL %Msg%n"/>
+    </formats>
+</seelog>`
 )
 
-// InitializeContainer adds the necessary bind mounts in order for the ExecCommandAgent to run properly in the container
-// Note: exec cmd agent is a linux feature, thus implemented here as a no-op.
-func (m *manager) InitializeContainer(taskId string, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig) error {
+//func (m *manager) InitializeContainer(taskId string, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig) (rErr error) {
```
angelcar, line 83: same as above, let's remove this if we don't need it anymore

agent/engine/execcmd/manager_init_task_linux.go
arun-annamalai: yup, removed

agent/engine/execcmd/manager_init_task_windows.go
arun-annamalai: yup, removed

angelcar (approved):

sharanyad (approved):

PR #2911, opened 2021-06-21
ECS EC2 task networking for Windows: Data validation for CNI Plugin invocation config

rawahars:
### Summary

The changes introduced in this PR validates the data for any special characters and length before the same is used for invoking the CNI plugins.
### Implementation details

A regex is created of valid characters and is matched against the characters in the data. Additionally, the length of the data is also validated.

The length of the data in this case can be at max 18 characters which is the length of longest IP Address in CIDR format.

### Testing

Unit tests were executed and a custom binary was also tested.
New tests cover the changes: 
Yes

Data validation for CNI plugin input parameters.
### Licensing

### DIFF
--- a/agent/ecscni/netconfig_windows.go
+++ b/agent/ecscni/netconfig_windows.go
@@ -16,11 +16,21 @@
 package ecscni
 
 import (
-	"github.com/containernetworking/cni/pkg/types"
+	"regexp"
 
 	"github.com/aws/amazon-ecs-agent/agent/api/eni"
+
+	"github.com/cihub/seelog"
 	"github.com/containernetworking/cni/libcni"
+	"github.com/containernetworking/cni/pkg/types"
 	"github.com/pkg/errors"
+)
+
+const (
+	// maxInputLength is the maximum length of IP Address or MAC Address as input to CNI plugin.
+	maxInputLength = 18
+	// allowedRegexPattern is the regex pattern for allowing valid characters in IP and MAC address.
+	allowedRegexPattern = `^$|^[A-Za-z0-9./: ]+$`
 )
 
 // NewVPCENIPluginConfigForTaskNSSetup is used to create the configuration of vpc-eni plugin for task namespace setup.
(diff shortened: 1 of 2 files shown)


fenxiong (approved):

agent/ecscni/netconfig_windows.go
```diff
@@ -16,13 +16,23 @@
...
+
+	"github.com/cihub/seelog"
 	"github.com/containernetworking/cni/libcni"
+	"github.com/containernetworking/cni/pkg/types"
 	"github.com/pkg/errors"
 )
 
+const (
+	// maxInputLength is the maximum length of IP Address or MAC Address as input to CNI plugin.
+	maxInputLength = 18
```
vsiddharth, line 31: Is this length per field?

agent/ecscni/netconfig_windows.go
```diff
@@ -31,6 +41,14 @@ func NewVPCENIPluginConfigForTaskNSSetup(eni *eni.ENI, cfg *Config) (*libcni.Net
 		Nameservers: cfg.InstanceENIDNSServerList,
 	}
 
+	// Validate MAC Address, ENI IP Address and ENI Gateway address used for CNI plugin configuration.
+	// Other params are generated at runtime and are considered safe.
```
vsiddharth, line 45: Can we also validate the DNS entries ?

Additionally, would it be more helpful if we could have a uber function to validate the entire network config?

agent/ecscni/netconfig_windows.go
rawahars: That's the maximum length of any field.
IP Address would be 18 characters long and 17 characters for MAC Address.

agent/ecscni/netconfig_windows.go
rawahars: That would not be required since they are obtainer from Win32 API call. Therefore, they are considered to be safe sources.

vsiddharth (approved):

agent/ecscni/netconfig_windows.go
```diff
@@ -31,6 +41,14 @@ func NewVPCENIPluginConfigForTaskNSSetup(eni *eni.ENI, cfg *Config) (*libcni.Net
...
+	if !isValid(eni.MacAddress) || !isValid(eni.GetPrimaryIPv4AddressWithPrefixLength()) ||
+		!isValid(eni.GetSubnetGatewayIPv4Address()) {
+		return nil, errors.New("failed to create vpc-eni plugin configuration for setting up " +
```
angelcar, line 48: nit: Most of this text should be part of a log message and the error should just be something like `failed data validation`

angelcar (approved):

PR #2919, opened 2021-06-25
Remove noisy log entries

angelcar:
### Summary

Improve ECS agent log readability by changing some of the most noisy log entries to `debug` level. These logs don't add much value (if at all) when debugging issues, oftentimes making it difficult for devs to follow what's happening at runtime.

### Testing

Manual verification. UTs & Integ Tests still pass.

* Remove noisy log entries from agent logs

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -223,7 +223,7 @@ func (mtask *managedTask) overseeTask() {
 		if !mtask.GetKnownStatus().Terminal() {
 			// If we aren't terminal and we aren't steady state, we should be
 			// able to move some containers along.
-			logger.Info("Task not steady state or terminal; progressing it", logger.Fields{
+			logger.Debug("Task not steady state or terminal; progressing it", logger.Fields{
 				field.TaskARN: mtask.Arn,
 			})
 
@@ -328,7 +328,7 @@ func (mtask *managedTask) waitSteady() {
 	}
 
 	if timedOut {
-		logger.Info("Checking to verify it's still at steady state", logger.Fields{
+		logger.Debug("Checking to verify it's still at steady state", logger.Fields{
 			field.TaskARN: mtask.Arn,
 		})
 		go mtask.engine.checkTaskState(mtask.Task)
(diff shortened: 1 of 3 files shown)


sparrc (approved):

agent/engine/task_manager.go
```diff
@@ -434,7 +434,7 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	container := containerChange.container
 	runtimeID := container.GetRuntimeID()
 	event := containerChange.event
-	logger.Info("Handling container change event", logger.Fields{
+	logger.Debug("Handling container change event", logger.Fields{
```
sharanyad, line 437: if logs from customers are just at info level, is there a way to find out if a docker event was received for the container state change?

agent/engine/task_manager.go
```diff
@@ -623,7 +617,7 @@ func (mtask *managedTask) emitTaskEvent(task *apitask.Task, reason string) {
 		})
 		return
 	}
-	logger.Info("Sending task change event", logger.Fields{
+	logger.Debug("Sending task change event", logger.Fields{
```
sharanyad, line 620: same - if logs from customers are just at info level, would we know if agent actually sent state change events to backend?

angelcar: For this one we will know. This log is a dupe of this one, which is the log right before sending data through the wire.

agent/engine/task_manager.go
angelcar: I was thinking about this one and I am a bit thorn. On the one hand, this log shows the events that the agent receives from docker, which is useful when it's the first time the event occurs (e.g. `PULLED`, `CREATED`), but it quickly gets noisy because docker keeps sending events for running containers. 

Maybe a good compromise would be to display this log in info level but only when `GetKnownStatus != event.Status`?

agent/engine/task_manager.go
sharanyad: should we just remove this log then?

agent/engine/task_manager.go
sharanyad: I like that!

sharanyad (approved):

fierlion (approved):
THANK YOU!

PR #2920, opened 2021-06-25
Add service files and packaging directories

masnoble:
### Summary

Adding the service file and setting up packaging directories. ExecStartPre and ExecStopPost which sets up and clens up the IPTables will need to be added into the service file after the bash files are complete for those.

### Testing

Manual testing- created service on instance and agent was able to properly access environment file and set itself up properly. I then started and enabled the service. The instance appeared on the intended cluster and remained there after the instance was rebooted through the console as well as when the agent was stopped and restarted through the instance. 
Then, I assigned a long running task to the instance which was functioning properly. I then did the same restarting methods. `Docker ps` revealed that there were no duplicate task containers and the task survived the restarts.

Add service files and packaging directories

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami/ecs.service
+++ b/packaging/amazon-linux-ami/ecs.service
@@ -1,0 +1,31 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the
+# "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+[Unit]
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://github.com/aws/amazon-ecs-agent
+Requires=docker.service
+After=docker.service
+After=cloud-final.service
+
+[Service]
+Type=simple
+Restart=on-failure
+RestartPreventExitStatus=5
+RestartSec=10
+EnvironmentFile=-/etc/ecs/ecs.config
+ExecStart=/usr/local/bin/amazon-ecs-agent
+
+[Install]
+WantedBy=multi-user.target
--- a/packaging/generic-deb/debian/ecs.service
+++ b/packaging/generic-deb/debian/ecs.service
@@ -1,0 +1,31 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the
+# "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+#     http://aws.amazon.com/apache2.0/
+#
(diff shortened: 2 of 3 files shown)


packaging/amazon-linux-ami/agent.service
```diff
@@ -0,0 +1,29 @@
...
+# License for the specific language governing permissions and
+# limitations under the License.
+
+[Unit]
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://github.com/aws/amazon-ecs-agent
+Requires=docker.service
+After=docker.service
+After=cloud-final.service
+[Service]
```
fierlion, line 21: nit: add a space between blocks
(Same for each duplicate `agent.service` file below.)

follow the pattern here:
https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs.service

fierlion (commented):
Do we want to call these unit files `ecs.service`?  So many service daemons call themselves 'agent'

fierlion (approved):

packaging/generic-deb/debian/ecs.service
```diff
@@ -0,0 +1,31 @@
...
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+[Unit]
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://github.com/aws/amazon-ecs-agent
+Requires=docker.service
+After=docker.service
+After=cloud-final.service
```
sparrc, line 20: the debian service doesnt wait for cloud-final.service in ecs-init. This is because this service usually wouldnt exist outside of AWS (for ecs anywhere), and so that when in AWS the ecs anywhere install script can be run in userdata.

packaging/generic-rpm/ecs.service
```diff
@@ -0,0 +1,31 @@
...
+# CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and
+# limitations under the License.
+
+[Unit]
+Description=Amazon Elastic Container Service - container agent
+Documentation=https://github.com/aws/amazon-ecs-agent
+Requires=docker.service
+After=docker.service
+After=cloud-final.service
```
sparrc, line 20: same comment as debian package above

packaging/amazon-linux-ami/ecs.service
```diff
@@ -0,0 +1,31 @@
...
+Requires=docker.service
+After=docker.service
+After=cloud-final.service
+
+[Service]
+Type=simple
+Restart=on-failure
+RestartPreventExitStatus=5
+RestartSec=10
+EnvironmentFile=-/etc/ecs/ecs.config
```
sharanyad, line 27: is the hyphen `-` needed before the config file path?

masnoble: The hyphen allows for the service to continue in case there is no environment file. The config file is not set up by default so I figured this would be useful in the case that a user did not create a `ecs.config` file

sparrc (approved):

packaging/amazon-linux-ami/ecs.service
sparrc: fwiw, this is what we are currently doing: https://github.com/aws/amazon-ecs-init/blob/master/packaging/amazon-linux-ami/ecs.service#L27

PR #2925, opened 2021-06-28
Add Iptable functionality

masnoble:
### Summary
Adding the bash files for `ExecStartPre` and `ExecStopPost` portions of the ecs.service file. Also, changed the directory or executable files to where system-manages files are supposed to live. 

### Testing
The instance was able to receive tasks properly. Running `sudo iptables -t nat -L` and `sudo iptables -t filter -L` showed all of the rules being properly made and cleaned up as the service would start and stop. This was done when setting `ECS_SKIP_LOCALHOST_TRAFFIC_FILTER` and `ECS_ALLOW_OFFHOST_INTROSPECTION_ACCESS` to both true and false to make sure the rules were being skipped if supposed to.
The iptable scripts and service file were also tested on ubuntu and debian instances and behaved as expected.

Add the iptable scripts and enable in service file.

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami/ipCleanup.sh
+++ b/packaging/amazon-linux-ami/ipCleanup.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -ex
+
+sysctl net.ipv4.conf.default.route_localnet
+sysctl -w net.ipv4.conf.all.route_localnet=0
+iptables -t nat -D PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-destination 127.0.0.1:51679
+
+if [ -n "${ECS_SKIP_LOCALHOST_TRAFFIC_FILTER}" ]
+then 
+    if [ "${ECS_SKIP_LOCALHOST_TRAFFIC_FILTER}" != "true" ]
+    then
+        iptables -t filter -D INPUT --dst 127.0.0.0/8 ! --src 127.0.0.0/8 -m conntrack ! --ctstate RELATED,ESTABLISHED,DNAT  -j DROP
+    fi
+else
+    iptables -t filter -D INPUT --dst 127.0.0.0/8 ! --src 127.0.0.0/8 -m conntrack ! --ctstate RELATED,ESTABLISHED,DNAT  -j DROP
+fi
+
+iptables -t filter -D INPUT -p tcp -i eth0 --dport 51678 -j DROP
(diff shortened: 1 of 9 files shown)


packaging/amazon-linux-ami/ipCleanup.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+sysctl net.ipv4.conf.default.route_localnet
+sysctl -w net.ipv4.conf.all.route_localnet=0
+iptables -t nat -D PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-destination 127.0.0.1:51679
+
+if [ -n "${ECS_SKIP_LOCALHOST_TRAFFIC_FILTER}" ]
```
chienhanlin, line 7: Please correct me if I am wrong. Do we only need to handle the case when `ECS_SKIP_LOCALHOST_TRAFFIC_FILTER` is specified to `true`? Because `ECS_SKIP_LOCALHOST_TRAFFIC_FILTER` is set to `false` as default in [ecs-init](https://github.com/aws/amazon-ecs-init). 

Same question for `ECS_ALLOW_OFFHOST_INTROSPECTION_ACCESS`. Thanks.

masnoble: You're right, but the script will get stuck if either of the environment variables are unset. So I am first testing to see if it is set or not before comparing it to see whether it is true. If there is a better way to test for this, I am all ears. Hopefully that clears it up a bit...

packaging/amazon-linux-ami/ipCleanup.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
sparrc, line 1: I would recommend doing probably `set -ex` at the top of these scripts. -e tells the scripts to fail if any of the commands below fail, -x prints out what the script does so we can see exactly what it's doing in the journalctl logs.

fierlion (commented):
Were you able to test these changes on Debian (and possibly Ubuntu too?)

fierlion (approved):
Looks good.  Will you be updating the README.md file with details on the container-less agent installation/update process?

masnoble:
> Looks good. Will you be updating the README.md file with details on the container-less agent installation/update process?

That will be coming in a future PR

sparrc (approved):

Issue #2927, opened 2021-06-29
Support docker with minAPIVerion > 1.21


pmv:
### Summary
ECS agent should support minAPIVerson >  1.21

### Description
I am attempting to set up an ECS anywhere cluster with RHEL and podman.  I realize ECS anywhere lists docker as the requirement, but since podman claims to be a mostly drop-in replacement for docker and it's what RHEL supports I thought I'd try it.  

What I've found is amazon-ecs-agent fails to start due to the version detection logic / default version.  This is because podman included with RHEL 8.3 reports the apiVersion as 1.40 and the minAPIVersion as 1.24.  From the agent log:

```
level=critical time=2021-06-29T21:29:49Z msg="Error creating Docker client: docker client factory: client not found for docker version: 1.21" module=agent.go
```

I was able to get by the error by changing this
https://github.com/aws/amazon-ecs-agent/blob/v1.53.1/agent/app/agent.go#L769-L774
to accept versions > 1.21 instead of looking for only 1.21 (since the minimum version is 1.24, the 1.21 check failed).

I also had to change this: https://github.com/aws/amazon-ecs-agent/blob/v1.53.1/agent/dockerclient/sdkclientfactory/versionsupport_unix.go#L56 so it would fall within the range of acceptable versions for my setup (>= 1.24 and < 1.40).

I understand hard-coding a higher default version is likely not feasible - perhaps better logic would be
1. if minAPIVersion <= 1.21, the default version = 1.21
2. if  minAPIVersion > 1.21, make the default version = minAPIVersion ?

yinyic:
Hi @pmv , thank you for raising this issue to our attention. 

We created PR #2937 , but have decided to hold it off after discussing with team. This is because we found that more profound changes with other ECS components need to happen for supporting this change properly, otherwise we may see undesired instance placement limitations. 

We have added backlog item to track this issue, and will update this thread once we get to work on it :)

sharanyad:
Since this would involve changes other than just the ECS agent to support this, I'm moving this over to containers-roadmap

PR #2937, opened 2021-07-06
Support docker daemon with minAPIVersion > 1.21

yinyic:
### Summary
This PR addresses the issue that current ECS Agent is not able to start if Docker daemon minimum API version is higher than 1.21 (issue #2927)

Currently when starting Agent, we hard-code a check to make sure that docker API version 1.21 is supported. This won't work if Docker daemon has minimum API version higher than 1.21. The reason why we check for 1.21 is that, minimum docker engine version supported by Agent is 1.9.0, which supports maximum API version 1.21 (see https://docs.docker.com/engine/api/#api-version-matrix). Therefore, if we are able to find a supported API version == 1.21, that implies the docker version is at least 1.9.0. However, as long as there is any supported API version >= 1.21, we will get the same guarantee.

Another issue is that Agent specifies a default API version, and is currently not taking into account that this version might not be supported by Daemon (which has higher minimum). We should take whichever has higher API version in this case.

### Implementation details
- Updated `verifyRequiredDockerVersion` to return success as long as there is a supported Docker API version (by both Agent and Daemon) that is >= 1.21, instead of strictly looking for a supported version == 1.21
- Updated `GetDefaultClient()` to return API client with min(Agent default API version, Docker Daemon min API version) instead of let it solely depend on Agent default.

### Testing

Updates Agent to support Docker daemon with minimum API version higher than 1.21. Fixes Agent issue #2927 

### Licensing

### DIFF
--- a/agent/app/agent_test.go
+++ b/agent/app/agent_test.go
@@ -97,61 +97,57 @@ func setup(t *testing.T) (*gomock.Controller,
 		mock_execcmdagent.NewMockManager(ctrl)
 }
 
-func TestDoStartMinimumSupportedDockerVersionTerminal(t *testing.T) {
-	ctrl, credentialsManager, state, imageManager, client,
-		dockerClient, stateManagerFactory, saveableOptionFactory, execCmdMgr := setup(t)
-	defer ctrl.Finish()
-
-	oldAPIVersions := []dockerclient.DockerVersion{
-		dockerclient.Version_1_18,
-		dockerclient.Version_1_19,
-		dockerclient.Version_1_20}
-	gomock.InOrder(
-		dockerClient.EXPECT().SupportedVersions().Return(oldAPIVersions),
-	)
-
-	cfg := getTestConfig()
-	cfg.Checkpoint = config.BooleanDefaultFalse{Value: config.ExplicitlyEnabled}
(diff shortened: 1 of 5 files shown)


angelcar (approved):

sharanyad (approved):

agent/app/agent_test.go
```diff
@@ -97,61 +97,57 @@ func setup(t *testing.T) (*gomock.Controller,
...
 
-	gomock.InOrder(
-		dockerClient.EXPECT().SupportedVersions().Return([]dockerclient.DockerVersion{}),
-	)
+	for _, test := range tests {
+		t.Run(test.testName, func(t *testing.T) {
+			ctrl, _, _, _, _, dockerClient, stateManagerFactory, saveableOptionFactory, _ := setup(t)
+			defer ctrl.Finish()
+			gomock.InOrder(
+				dockerClient.EXPECT().SupportedVersions().Return(test.supportedVersions),
```
chienhanlin, line 132: nit. probably don't need to use `InOrder` as there is only one expected call : )

yinyic:
Holding off on this PR due to potential placement limitations which would require a more profound change with other ECS components. 

Added item to Agent backlog for tracking.

Issue #2943, opened 2021-07-08
SSM Parameter ARN parsing for windows credentialspec options fails under certain conditions


rue-nsilverman:
### Summary
When passing an SSM parameter ARN to the the `credentialspec` option of a task definition running on windows, retrieving the SSM parameter will fail if the SSM parameter is in a path other than `/`.

### Description
When specifying an SSM parameter ARN for the `credentialspec` option inside a task definition, [the documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/windows-gmsa.html) makes reference to filling in a `parameter_name`. Given the current way the ARN is parsed and passed to the AWS SDK, specifying a parameter that is in a path below `/` (example `/foo/gmsa_creds`), the ECS agent will fail to retrieve the SSM parameter. This is because the way the SSM ARN is parsed, its passed through a `filepath.Base()` function, rendering the example SSM parameter as `gmsa_creds`. This will cause the SDK to raise an `AccessDeniedException` because that parameter doesn't exist at the `/` path.

Using an SSM parameter ARN whose parameter name is `/gmsa_creds`, does behave as expected.

### Expected Behavior
specifying an SSM ARN for the `credentialspec` option on windows like
```
credentialspec:arn:aws:ssm:us-west-2:123456789012:parameter/subpath/test
```
should correctly retrieve the value of the SSM parameter at `/subpath/test`

### Observed Behavior
When specifying an SSM ARN that's anything other than at the `/` path, you will receive an `AccessDeniedException` and the container task will stop returning that exception to the ECS console under the stopped task details. This will always happen unless you also happen to have an SSM parameter at the root path.

For example, if you had an SSM parameter at `/subpath/test` and didn't also have that same parameter at `/test`, it would fail. If you duplicate the parameter to the `/test` path, it will succeed, and the ECS agent logs will log the fact that the parameter was retrieved.

I have also confirmed that this is not an IAM role issue as the role my task is using has the `arn:aws:iam::aws:policy/AmazonSSMFullAccess` policy attached.

### Environment Details
AMI ID ami-0499ce56124b91fc8 (Windows_Server-2019-English-Full-ECS_Optimized-2021.06.11)
Agent version 1.53.0
Docker version 20.10.5

### Supporting Log Snippets
An example error returned to the AWS ECS task details page (with redacted sensitive ARN info) is:
```
[...]

PR #2944, opened 2021-07-08
Sub-path support for SSM ARNs being used for credentialspec configuration

rue-nsilverman:
### Summary
In reference to https://github.com/aws/amazon-ecs-agent/issues/2943, This PR aims to enable support for SSM parameter ARNs that have parameter names that include more than one sub-path.

### Implementation details
This is achieved by splitting the parsed ARNs resource string to extract the full path of the SSM parameter, rather than passing it through `filepath.Base()`. This is needed because `filepath.Base()` will take a parameter called `/test/gmsatest` and return `gmsatest`. That value is then passed to SSM and will return an exception because that's not the actual path of the SSM parameter.

It's also worth mentioning that this now makes the file name include the whole parameter path, which in theory, could possibly be longer than windows file name limit (I'm not sure on either windows limit or ssm parameter path limit). If we are concerned with that, we can just set `ssmParamSanitizedFileName` to the original value of `filepath.Base(parsedARN.Resource)` as the task ID is part of the name, so we shouldn't have to worry about file uniqueness.

### Testing
I was actually having issues running either `make test` or `make test-in-docker` so I cant actually run these tests locally. I would appreciate any assistance you can provide as to how to get that working so I can run the test I modified.

New tests cover the changes: yes (they should)

- enables support for SSM parameter ARNs that contain multi-level parameter paths

### Licensing

### DIFF
--- a/agent/taskresource/credentialspec/credentialspec_windows.go
+++ b/agent/taskresource/credentialspec/credentialspec_windows.go
@@ -445,8 +445,8 @@ func (cs *CredentialSpecResource) handleSSMCredentialspecFile(originalCredential
 
 	ssmClient := cs.ssmClientCreator.NewSSMClient(cs.region, iamCredentials)
 
-	ssmParam := filepath.Base(parsedARN.Resource)
-	ssmParams := []string{ssmParam}
+	ssmParam := strings.SplitAfterN(parsedARN.Resource, "parameter", 2) 
+	ssmParams := []string{ssmParam[1]}
 
 	ssmParamMap, err := ssm.GetParametersFromSSM(ssmParams, ssmClient)
 	if err != nil {
@@ -460,7 +460,8 @@ func (cs *CredentialSpecResource) handleSSMCredentialspecFile(originalCredential
 	if length < 2 {
 		return errors.New("Failed to retrieve taskId from taskArn.")
 	}
-	localCredSpecFilePath := fmt.Sprintf("%s\\ssm_%v_%s", cs.credentialSpecResourceLocation, taskArnSplit[length-1], ssmParam)
+	ssmParamSanitizedFileName := strings.Replace(ssmParam[1], "/", "_")
+	localCredSpecFilePath := fmt.Sprintf("%s\\ssm_%v%s", cs.credentialSpecResourceLocation, taskArnSplit[length-1], ssmParamSanitizedFileName)
 	err = cs.writeSSMFile(ssmParamData, localCredSpecFilePath)
 	if err != nil {
 		cs.setTerminalReason(err.Error())
--- a/agent/taskresource/credentialspec/credentialspec_windows_test.go
+++ b/agent/taskresource/credentialspec/credentialspec_windows_test.go
@@ -209,9 +209,9 @@ func TestHandleSSMCredentialspecFile(t *testing.T) {
 		CredentialsID: "test-cred-id",
 	}
 
-	credentialSpecSSMARN := "arn:aws:ssm:us-west-2:123456789012:parameter/test"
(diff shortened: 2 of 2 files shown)


agent/taskresource/credentialspec/credentialspec_windows.go
```diff
@@ -460,7 +460,8 @@ func (cs *CredentialSpecResource) handleSSMCredentialspecFile(originalCredential
 	if length < 2 {
 		return errors.New("Failed to retrieve taskId from taskArn.")
 	}
-	localCredSpecFilePath := fmt.Sprintf("%s\\ssm_%v_%s", cs.credentialSpecResourceLocation, taskArnSplit[length-1], ssmParam)
+	ssmParamSanitizedFileName := strings.Replace(ssmParam[1], "/", "_")
```
vsiddharth, line 463: Should it be `strings.Replace` or `strings.ReplaceAll ` here?

rue-nsilverman: `strings.Replace()` takes a fourth argument that specifies the number of replaces that will happen, but when < 0 or undefined, will have no limit on how many replaces it makes. That being said, `ReplaceAll` would be more immediately understandable.

rue-nsilverman: I also called this out in the description, but I was at least thinking that there is a potential here for using the whole SSM param path as part of the file name to hit file name length limits. If we wanted to be extra cautious, we could make this where we call the `filepath.Base()` function to generate that part of the file name.

vsiddharth: I agree, I think it would be great if we could use the whole SSM param path and ensure that it does not hit the file name length limits.

rue-nsilverman: actually, having done some quick research, windows [MAX_PATH ](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd) (the whole path including filename) must be less than 260 characters. According to [AWS Docs](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParametersByPath.html#API_GetParametersByPath_RequestSyntax), max path length for an SSM parameter is 2048 characters. Given that we already take up a decent number of characters from the `credentialSpecResourceLocation` and the `taskId`, I fear we would see a ton of issues using the whole SSM parameter path as part of the filename. So I think in this instance we should revert this to using the `filepath.Base()` function to be safer.

That being said, there is still a possibility that someone can use a very long name for their SSM parameter that doesnt have sub path hierarchies, and that could still break the limit here. If we wanted to be _really_ safe, we would truncate length once the full path is constructed.

Let me know how you want to proceed here and I'll make the appropriate changes.

vsiddharth: While, we can document the known limitations. It would be neat if we could come up with a way to support both long names and nested sub-path hierarchies.

rue-nsilverman: yeah, though i dont think we can do that without first having [a specific windows setting enabled](https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=cmd#enable-long-paths-in-windows-10-version-1607-and-later)

vsiddharth: @rue-nsilverman We could fix this issue by introducing a hierarchy to the local `credential-spec` file alongside volume paths. 

Currently `credential-specs` are modeled as a task resource and all of them are created together. When this task-resource is created, we could plumb the task\container details down to this resource and introduce a hierarchy similar to `CredSpecRoot\task\container\credspec` and wire this to the container appropriately.

mssrivas:
No updates in last 6 months. Closing.

PR #2947, opened 2021-07-09
wire up healthcheck doctor and react to the acs heartbeat

YashdalfTheGray:
### Summary

- initialize *the* doctor in the agent app
- wire the doctor through to the acs session
- wire the doctor into the acs heartbeat handler
- fix the tests
- created a function to set up the doctor and the health checks

### Implementation details

- `acs_handler` session was changed to include a `doctor.Doctor` argument
- `heartbeat_handler` constructor was changed to include a `doctor.Doctor` argument
- a function called `newDoctorWithHealthchecks` was created to instantiate the doctor and all the healthchecks

### Testing

- `make test`
- checked the agent logs for the new log lines as well as the health check run logs (instance was registered to a cluster and connected to ACS)

New tests cover the changes:  no, mostly assignment code in this PR but the tests have been fixed accordingly

- Enhancement - instantiate doctor in one place and wire it wherever necessary

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -30,6 +30,7 @@ import (
 	rolecredentials "github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/data"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
+	"github.com/aws/amazon-ecs-agent/agent/doctor"
 	"github.com/aws/amazon-ecs-agent/agent/engine"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/eventhandler"
@@ -96,6 +97,7 @@ type session struct {
 	backoff                         retry.Backoff
 	resources                       sessionResources
 	latestSeqNumTaskManifest        *int64
+	doctor                          *doctor.Doctor
 	_heartbeatTimeout               time.Duration
 	_heartbeatJitter                time.Duration
 	_inactiveInstanceReconnectDelay time.Duration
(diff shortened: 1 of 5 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -339,7 +346,7 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 
 	client.AddRequestHandler(payloadHandler.handlerFunc())
 
-	heartbeatHandler := newHeartbeatHandler(acsSession.ctx, client, acsSession.dockerClient, acsSession.agentConfig.Cluster, acsSession.containerInstanceARN)
+	heartbeatHandler := newHeartbeatHandler(acsSession.ctx, client, acsSession.doctor)
```
shubham2892, line 349: I would prefer that we pass the `cluster` and `containerInstanceARN` as separate parameters rather than combining them in Doctor.

agent/app/agent.go
```diff
@@ -351,6 +352,16 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
 	}
 
+	// now that we know the container instance ARN, we can build out the doctor
+	// and pass it on to ACS and TACS
+	seelog.Info("Setting up the doctor healthchecks.")
+	doctor, doctorCreateErr := agent.newDoctorWithHealthchecks(agent.cfg.Cluster, agent.containerInstanceARN)
+	if doctorCreateErr != nil {
+		seelog.Warnf("Error starting doctor, healthchecks won't be running: %v", err)
+	} else {
+		seelog.Info("Doctor healthchecks set up properly.")
```
shubham2892, line 362: We can avoid using adding the two `info` level logs and only keep `warn` level logs for the case when their is error in setting up the Doctor.

agent/acs/handler/acs_handler.go
YashdalfTheGray: We only need cluster and container instance arn for the doctor, they're not being used otherwise.

agent/app/agent.go
YashdalfTheGray: Ack

agent/app/agent.go
```diff
@@ -440,6 +451,21 @@ func (agent *ecsAgent) initMetricsEngine() {
 	metrics.PublishMetrics()
 }
 
+// newDoctorWithHealthchecks creates a new doctor and also configures
+// the healthchecks that the doctor should be running
+func (agent *ecsAgent) newDoctorWithHealthchecks(cluster, containerInstanceARN string) (*doctor.Doctor, error) {
```
fierlion, line 456: i like this!

fierlion (approved):

shubham2892 (approved):
LGTM 🚢

PR #2953, opened 2021-07-13
Implemented firelens capability advertising and unit tests for windows

vineeth-bandi:
### Summary

Advertises agent capability for firelens fluent logging
### Testing

Extended the existing unit tests 

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/app/agent_capability_windows.go
+++ b/agent/app/agent_capability_windows.go
@@ -48,25 +48,8 @@ func (agent *ecsAgent) appendTaskEIACapabilities(capabilities []*ecs.Attribute) 
 	return capabilities
 }
 
-func (agent *ecsAgent) appendFirelensFluentdCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
-	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityFirelensFluentd)
-}
-
-func (agent *ecsAgent) appendFirelensFluentbitCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
-	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityFirelensFluentbit)
-}
-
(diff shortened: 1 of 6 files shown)


mythri-garaga (dismissed):

mythri-garaga (commented):
I think we need a test similar to this - https://github.com/aws/amazon-ecs-agent/blob/master/agent/app/agent_capability_unix_test.go#L812-L852 for windows as well :)

mythri-garaga (approved):

agent/app/agent_capability_windows.go
```diff
@@ -49,23 +49,24 @@ func (agent *ecsAgent) appendTaskEIACapabilities(capabilities []*ecs.Attribute)
 }
 
 func (agent *ecsAgent) appendFirelensFluentdCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
-	return capabilities
+	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityFirelensFluentd)
```
sharanyad, line 52: Instead of adding new code here - can you move these methods to agent_capability.go, so the same is used for unix and windows as well. this will need removal of these methods from agent_capability_unix.go as well.

agent/app/agent_capability_windows_test.go
```diff
@@ -221,6 +226,64 @@ func TestSupportedCapabilitiesWindows(t *testing.T) {
 	}
 }
 
+func TestFirelensConfigCapabilitiesWindows(t *testing.T) {
```
sharanyad, line 229: same comment as above. Can the existing test https://github.com/aws/amazon-ecs-agent/blob/4d4c4afffc7283089c74d4513a9c5d90d3842c43/agent/app/agent_capability_unix_test.go#L812 be moved to agent_capability_test.go file instead?

vineeth-bandi: Actually it appears that this was not the reason for the failed test.

mythri-garaga (approved):

singholt (approved):

PR #2957, opened 2021-07-20
Added additional GraphQL TMDE Fields

SeaverDahlgren:
### Summary

Adds new container fields `Networks` and `Stats`. Adds task fields `TaskTags` and `ContainerInstanceTags`.  All of the fields resolve to the custom scalar type, `JSON` except for `Networks`, which is a list of `JSON` types. Additionally, pulled containers have also been added to the `taskContainersResolver`.
### Implementation details

The `statsResolver` utilizes the `taskEngineState` to retrieve the docker stats. both the `TaskTags` and `ContainerInstanceTags` use the `ecsClient` to retrieve the tags. For `Networks`, if the container is pulled, it immediately  return nil. Otherwise, it checks if the task has an Elastic Network Interface associated with it. If it has one, it sets the network to mode to `AWSVPC`. If not, the mode is set to `Bridge`.

### Testing

Tests were created in `schema_test.go`.

New tests cover the changes: 
yes

Adds TMDE GraaphQL Tags, Networks, and Stats Fields

### Licensing

### DIFF
--- a/agent/handlers/gql/schema.go
+++ b/agent/handlers/gql/schema.go
@@ -16,14 +16,17 @@ package gql
 import (
 	"time"
 
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
+
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/agent/containermetadata"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/handlers/utils"
 	v2 "github.com/aws/amazon-ecs-agent/agent/handlers/v2"
 	v3 "github.com/aws/amazon-ecs-agent/agent/handlers/v3"
-
+	v4 "github.com/aws/amazon-ecs-agent/agent/handlers/v4"
 	"github.com/aws/amazon-ecs-agent/agent/stats"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/graphql-go/graphql"
@@ -73,7 +76,7 @@ func CreateSchema(
 			},
 			"KnownStatus": &graphql.Field{
 				Type:    graphql.String,
-				Resolve: contaienrKnownStatusResolver,
+				Resolve: containerKnownStatusResolver,
 			},
 			"Limits": &graphql.Field{
 				Type:    JSON,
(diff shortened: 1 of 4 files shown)


agent/handlers/gql/schema.go
```diff
@@ -314,6 +339,70 @@ func containerARNResolver(p graphql.ResolveParams) (interface{}, error) {
...
+		}, nil
+	}
+}
+
+func containerNetworksResolver(state dockerstate.TaskEngineState) func(p graphql.ResolveParams) (interface{}, error) {
+	return func(p graphql.ResolveParams) (interface{}, error) {
+		if dockerContainer, ok := p.Source.(*apicontainer.DockerContainer); ok {
+			// Pulled containers do not have a network
+			// TODO: Find a cleaner way to have pulled containers return nil
+			if dockerContainer.Container.KnownStatusUnsafe.String() == "PULLED" {
```
angelcar, line 370: this should be: `if dockerContainer.Container.KnownStatusUnsafe == apicontainerstatus.ContainerPulled {`

agent/handlers/gql/schema.go
```diff
@@ -314,6 +339,70 @@ func containerARNResolver(p graphql.ResolveParams) (interface{}, error) {
...
+				return nil, nil
+			}
+			task, ok := state.TaskByID(dockerContainer.DockerID)
+			if !ok {
+				return nil, errors.Errorf("Unable to find task for container '%s'", dockerContainer.DockerID)
+			}
+			eni := task.GetPrimaryENI()
+			var resp []v4.Network
+			if eni != nil { // Set Network to AWSVPC
```
angelcar, line 379: This is a bit confusing. What is being "set" here?

Also, shouldn't this reuse the existing v4 network logic? https://github.com/aws/amazon-ecs-agent/blob/master/agent/handlers/v4/response.go#L126-L129

agent/handlers/gql/schema.go
```diff
@@ -397,12 +486,47 @@ func taskContainersResolver(state dockerstate.TaskEngineState) func(p graphql.Re
...
 		}
 		return nil, nil
 	}
 }
 
+func taskTagsResolver(ecsClient api.ECSClient) func(p graphql.ResolveParams) (interface{}, error) {
+	return func(p graphql.ResolveParams) (interface{}, error) {
+		if task, ok := p.Source.(*apitask.Task); ok {
+			taskTags, err := ecsClient.GetResourceTags(task.Arn)
+			if err == nil {
```
angelcar, line 503: nit: let's do `if err != nil`, like we do in most cases. Same for `containerInstanceTagsResolver`

agent/handlers/gql/schema.go
```diff
@@ -238,7 +255,7 @@ func containerDesiredStatusResolver(p graphql.ResolveParams) (interface{}, error
 	return nil, nil
 }
 
-func contaienrKnownStatusResolver(p graphql.ResolveParams) (interface{}, error) {
+func contaienerKnownStatusResolver(p graphql.ResolveParams) (interface{}, error) {
```
chienhanlin, line 258: 🎏 Another typo is hidden here

angelcar (approved):

chienhanlin (approved):

PR #2958, opened 2021-07-20
Support buffer limit option in FireLens

zhonghui12:
### Summary

Provide an entry to configure[fluentd-buffer-limit](https://docs.docker.com/config/containers/logging/fluentd/#fluentd-buffer-limit) in FireLens.
This is being enabled to resolve the following request: https://github.com/aws/containers-roadmap/issues/964

### Implementation details

Help to configure buffer limit size in fluentD Log Driver
### Testing

- Updated two unit tests and manually tested the feature with `make test`
- The code also passed the end-to-end test on my instance.
Here is my task definition file:
```
"logConfiguration": {
        "logDriver": "awsfirelens",
        "secretOptions": null,
        "options": {
          "log_group_name": "/aws/ecs/containerinsights/$(ecs_cluster)/application",
          "log-driver-buffer-limit": "123456",
          "auto_create_group": "true",
          "log_key": "log",
          "log_stream_prefix": "log_stream_name",
          "region": "us-west-2",
          "Name": "cloudwatch"
        }
```

Result for `docker inspect`:
```
"HostConfig": {
            "Binds": [],
            "ContainerIDFile": "",
            "LogConfig": {
                "Type": "fluentd",
                "Config": {
                    "fluentd-address": "unix:///var/lib/ecs/data/firelens/677a7f0ded6c47709e26665f714e4c76/socket/fluent.sock",
                    "fluentd-async-connect": "true",
                    "fluentd-buffer-limit": "123456",
                    "fluentd-sub-second-precision": "true",
                    "tag": "app-firelens-677a7f0ded6c47709e26665f714e4c76"
                }
            },
```

Also, tested it with no `log-driver-buffer-limit` option:
```
"HostConfig": {
            "Binds": [],
            "ContainerIDFile": "",
            "LogConfig": {
                "Type": "fluentd",
                "Config": {
                    "fluentd-address": "unix:///var/lib/ecs/data/firelens/65b06118b2014ccd9ef92abf5727577d/socket/fluent.sock",
                    "fluentd-async-connect": "true",
                    "fluentd-sub-second-precision": "true",
                    "tag": "app-firelens-65b06118b2014ccd9ef92abf5727577d" 
                }
            },
```

Feature FluentD: Support for buffer limit size configuration.

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -117,6 +117,8 @@ const (
 	firelensSocketBindFormat = "%s/data/firelens/%s/socket/:/var/run/"
 	// firelensDriverName is the log driver name for containers that want to use the firelens container to send logs.
 	firelensDriverName = "awsfirelens"
+	// FirelensLogDriverBufferLimitOption is the option for customers who want to specify the buffer limit size in FireLens.
+	FirelensLogDriverBufferLimitOption = "log-driver-buffer-limit"
 
 	// firelensConfigVarFmt specifies the format for firelens config variable name. The first placeholder
 	// is option name. The second placeholder is the index of the container in the task's container list, appended
@@ -1069,6 +1071,9 @@ func (task *Task) collectFirelensLogOptions(containerToLogOptions map[string]map
 				containerToLogOptions[container.Name] = make(map[string]string)
 			}
 			for k, v := range hostConfig.LogConfig.Config {
+				if k == FirelensLogDriverBufferLimitOption {
+					continue
+				}
 				containerToLogOptions[container.Name][k] = v
 			}
 		}
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -89,6 +89,7 @@ const (
 	dataLogDriverPath           = "/data/firelens/"
 	logDriverAsyncConnect       = "fluentd-async-connect"
 	logDriverSubSecondPrecision = "fluentd-sub-second-precision"
+	logDriverBufferLimit        = "fluentd-buffer-limit"
 	dataLogDriverSocketPath     = "/socket/fluent.sock"
 	socketPathPrefix            = "unix://"
 
@@ -1279,12 +1280,16 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
 	tag := fmt.Sprintf(fluentTagDockerFormat, container.Name, taskID)
 	fluentd := socketPathPrefix + filepath.Join(cfg.DataDirOnHost, dataLogDriverPath, taskID, dataLogDriverSocketPath)
 	logConfig := hostConfig.LogConfig
+	bufferLimit, bufferLimitExists := logConfig.Config[apitask.FirelensLogDriverBufferLimitOption]
 	logConfig.Type = logDriverTypeFluentd
 	logConfig.Config = make(map[string]string)
 	logConfig.Config[logDriverTag] = tag
 	logConfig.Config[logDriverFluentdAddress] = fluentd
 	logConfig.Config[logDriverAsyncConnect] = strconv.FormatBool(true)
 	logConfig.Config[logDriverSubSecondPrecision] = strconv.FormatBool(true)
+	if bufferLimitExists {
+		logConfig.Config[logDriverBufferLimit] = bufferLimit
+	}
 	seelog.Debugf("Applying firelens log config for container %s: %v", container.Name, logConfig)
 	return logConfig
 }
--- a/agent/api/task/task_linux_test.go
+++ b/agent/api/task/task_linux_test.go
@@ -1148,8 +1148,9 @@ func getFirelensTask(t *testing.T) *Task {
 		LogConfig: dockercontainer.LogConfig{
 			Type: firelensDriverName,
 			Config: map[string]string{
-				"key1": "value1",
-				"key2": "value2",
+				"key1":                    "value1",
+				"key2":                    "value2",
+				"log-driver-buffer-limit": "10000",
 			},
 		},
 	}
--- a/agent/engine/docker_task_engine_test.go
+++ b/agent/engine/docker_task_engine_test.go
@@ -2670,8 +2670,9 @@ func TestCreateContainerAddFirelensLogDriverConfig(t *testing.T) {
 			LogConfig: dockercontainer.LogConfig{
 				Type: logDriverType,
 				Config: map[string]string{
-					"key1": "value1",
-					"key2": "value2",
+					"key1":                    "value1",
+					"key2":                    "value2",
+					"log-driver-buffer-limit": "10000",
 				},
 			},
 		}
@@ -2766,6 +2767,7 @@ func TestCreateContainerAddFirelensLogDriverConfig(t *testing.T) {
 		expectedLogConfigFluentAddress string
 		expectedFluentdAsyncConnect    string
 		expectedSubSecondPrecision     string
+		expectedBufferLimit            string
 		expectedIPAddress              string
 		expectedPort                   string
 	}{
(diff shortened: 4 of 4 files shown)


agent/api/task/task.go
```diff
@@ -147,6 +149,9 @@ const (
 	sysctlValueOff = "0"
 )
 
+// LogDriverBufferLimit is the value of the user-defined option log-driver-buffer-limit
+var LogDriverBufferLimit = ""
+
```
PettitWesley, line 154: why is this a global variable? I'm new to ECS Agent code as well, but since it serves multiple tasks, and each task might have multiple container each which can have their own buffer limit- a global variable doesn't seem like what we want.

agent/api/task/task.go
```diff
@@ -1069,6 +1074,9 @@ func (task *Task) collectFirelensLogOptions(containerToLogOptions map[string]map
 				containerToLogOptions[container.Name] = make(map[string]string)
 			}
 			for k, v := range hostConfig.LogConfig.Config {
+				if k == firelensLogDriverBufferLimitOption {
+					LogDriverBufferLimit = v
+				}
```
PettitWesley, line 1079: So this code gets the buffer limit from the user input- but how does it get added to the fluentd docker log driver configuration? I don't see that code in this PR....

agent/engine/docker_task_engine.go
```diff
@@ -1285,6 +1286,7 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
 	logConfig.Config[logDriverFluentdAddress] = fluentd
 	logConfig.Config[logDriverAsyncConnect] = strconv.FormatBool(true)
 	logConfig.Config[logDriverSubSecondPrecision] = strconv.FormatBool(true)
+	logConfig.Config[logDriverBufferLimit] = apitask.LogDriverBufferLimit
```
PettitWesley, line 1289: I'm confused, this suggested its a field on the task structure, but then in the other file I don't see it added as a field in that struct...

agent/api/task/task.go
zhonghui12: because we want set the input in the config but it is in a different package: https://github.com/aws/amazon-ecs-agent/pull/2958/files#diff-0f6cc891620e98230e92868ea887af5e1f11d9e692fc8415a50c1dd0b130cc8cR1289. So I use a global var to store the value from the option.

agent/engine/docker_task_engine.go
zhonghui12: I learnt from this PR that just adding one field here is enough: https://github.com/aws/amazon-ecs-agent/pull/2538/files#diff-0f6cc891620e98230e92868ea887af5e1f11d9e692fc8415a50c1dd0b130cc8cR1106

zhonghui12: It is not a field on the task structure, it is a field in Config. And we need to get the value from the task package

PettitWesley: Okay, but our requirement is that each container can have its own log driver buffer setting. So I think it needs to be stored somewhere which is both per-task and per-container. Not some global setting.

fenxiong: i don't think this needs to be saved anywhere. instead just do the following should work: (1) in collectFirelensLogOptions, skip adding the option to containerToLogOptions if the key equals firelensLogDriverBufferLimitOption; (2) here, get the actual value for "fluentd-buffer-limit" from the container's log config before we reset it at the previous line `logConfig.Config = make(map[string]string)`, and then save it to the config here.

zhonghui12:
@PettitWesley @fenxiong updated the code and already tested it in my local by `make test`

agent/engine/docker_task_engine.go
```diff
@@ -1279,12 +1280,14 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
...
 	fluentd := socketPathPrefix + filepath.Join(cfg.DataDirOnHost, dataLogDriverPath, taskID, dataLogDriverSocketPath)
 	logConfig := hostConfig.LogConfig
+	bufferLimit := logConfig.Config[apitask.FirelensLogDriverBufferLimitOption]
 	logConfig.Type = logDriverTypeFluentd
 	logConfig.Config = make(map[string]string)
 	logConfig.Config[logDriverTag] = tag
 	logConfig.Config[logDriverFluentdAddress] = fluentd
 	logConfig.Config[logDriverAsyncConnect] = strconv.FormatBool(true)
 	logConfig.Config[logDriverSubSecondPrecision] = strconv.FormatBool(true)
+	logConfig.Config[logDriverBufferLimit] = bufferLimit
```
fenxiong, line 1290: this should only be added when bufferLimit is not empty

agent/api/task/task.go
```diff
@@ -117,6 +117,8 @@ const (
 	firelensSocketBindFormat = "%s/data/firelens/%s/socket/:/var/run/"
 	// firelensDriverName is the log driver name for containers that want to use the firelens container to send logs.
 	firelensDriverName = "awsfirelens"
+	// firelensLogDriverBufferLimitOption is the option for customers who want to specify the buffer limit size in FireLens.
```
fenxiong, line 120: nit: firelensLogDriverBufferLimitOption -> FirelensLogDriverBufferLimitOption

agent/engine/docker_task_engine.go
zhonghui12: I see here has a check for empty string so I didn't add it: https://github.com/moby/moby/blob/master/daemon/logger/fluentd/fluentd.go#L174. But yeah, we should handle this case here. Thanks for the suggestion.

zhonghui12:
Update the PR based on the comments. Will squash all commits into one if PR is approved

fenxiong:
static check is failing for other reason and has been fixed in https://github.com/aws/amazon-ecs-agent/commit/997c97565b18a7529d81a35c1a8aabf800be43e9. please rebase on latest dev branch to get the tests passed

fenxiong (approved):

zhonghui12:
The code also passed the end-to-end test on my instance.
Here is my task definition file:
```
"logConfiguration": {
        "logDriver": "awsfirelens",
        "secretOptions": null,
        "options": {
          "log_group_name": "/aws/ecs/containerinsights/$(ecs_cluster)/application",
          "log-driver-buffer-limit": "123456",
          "auto_create_group": "true",
          "log_key": "log",
          "log_stream_prefix": "log_stream_name",
          "region": "us-west-2",
          "Name": "cloudwatch"
        }
```

Result for `docker inspect`:
```
"HostConfig": {
            "Binds": [],
            "ContainerIDFile": "",
            "LogConfig": {
                "Type": "fluentd",
                "Config": {
                    "fluentd-address": "unix:///var/lib/ecs/data/firelens/677a7f0ded6c47709e26665f714e4c76/socket/fluent.sock",
                    "fluentd-async-connect": "true",
                    "fluentd-buffer-limit": "123456",
                    "fluentd-sub-second-precision": "true",
                    "tag": "app-firelens-677a7f0ded6c47709e26665f714e4c76"
                }
            },
```

Also, tested it with no `log-driver-buffer-limit` option:
```
"HostConfig": {
            "Binds": [],
            "ContainerIDFile": "",
            "LogConfig": {
                "Type": "fluentd",
                "Config": {
                    "fluentd-address": "unix:///var/lib/ecs/data/firelens/65b06118b2014ccd9ef92abf5727577d/socket/fluent.sock",
                    "fluentd-async-connect": "true",
                    "fluentd-sub-second-precision": "true",
                    "tag": "app-firelens-65b06118b2014ccd9ef92abf5727577d" 
                }
            },
```

mythri-garaga (approved):

PR #2965, opened 2021-07-29
Rpm

masnoble:
*note; this is 
### Summary
Adding generic-rpm to agent. A new generic-rpm make target will be added to the makefile. This target will produce an install-able rpm file that installs the ECS-Agent as a host-level process managed by systemd. This rpm will include Awsvpc and volume plugin support. 

### Implementation details
There was another target added to the makefile along with the .spec file. 
In order to enable support for the Awsvpc and the volume plugins, appmesh needed to be disabled and the volume files were added to the agent (along with some missing vendor files)

### Testing
All MACIS tests were passed when using an AMI which had the agent installed using the generic-rpm package except for the following:
TestExecSSM -This is due to the ssm dependency volume mounts not being set in the agent (should be coming in future commit)
TestBoltDB -included ssm command which doesn't work due to dependency issue, but boltdb functionality was checked manually
TestTaskIAMRole -test requires ssm
TestExecutionRole -test requires ssm
TestAppmesh -Appmesh was disabled 
TestContainerUpdateContainerAgent -Failed since it is no longer upgrading/downgrading with as a contained agent
TestContainerMetadataFile -`Expected container-metadata-file-validator to exit with 42; actually exited (true) with 1` (This will be looked into more)

Add generic-rpm support

### Licensing


Makefile
```diff
@@ -297,6 +297,19 @@ GOPATH=$(shell go env GOPATH)
...
 get-deps: .get-deps-stamp
 
+.generic-rpm-done: pause-container-release
+	cp packaging/generic-rpm/ecs-agent.spec ecs-agent.spec
+	cp packaging/generic-rpm/ecs.service ecs.service
+	cp packaging/generic-rpm/amazon-ecs-volume-plugin.service amazon-ecs-volume-plugin.service
+	cp packaging/generic-rpm/amazon-ecs-volume-plugin.socket amazon-ecs-volume-plugin.socket
+	tar -czf ./sources.tgz agent scripts
+	test -e SOURCES || ln -s . SOURCES
+	rpmbuild --undefine=_disable_source_fetch --define "%_topdir $(PWD)" -bb ecs-agent.spec
```
sparrc, line 307:
```
--undefine=_disable_source_fetch
```

what does this do?

agent/app/agent_unix.go
```diff
@@ -48,7 +48,7 @@ const initPID = 1
 var awsVPCCNIPlugins = []string{ecscni.ECSENIPluginName,
 	ecscni.ECSBridgePluginName,
 	ecscni.ECSIPAMPluginName,
-	ecscni.ECSAppMeshPluginName,
+	//ecscni.ECSAppMeshPluginName,
```
mssrivas, line 51: why is it commented out?

sparrc, line 51: why are we disabling this plugin?

Makefile
masnoble: Before, the RPM would never download the CNI tarbells from s3, that line was needed to enable downloading the sources in the .spec file.

agent/app/agent_unix.go
masnoble: Appmesh is not yet supported with how the CNI plugins were installed. This will be added to future work for the project.

masnoble: Appmesh is not yet supported with how the CNI plugins were installed. This will be added to future work for the project. It was pushed out of scope for now. 
Bottlerocket was referred to while installing CNI plugins: https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/ecs-agent/0004-bottlerocket-remove-unsupported-CNI-plugins.patch

README.md
```diff
@@ -228,8 +228,11 @@ for production on Linux, but it can be useful for development or easier integrat
 The following commands run the agent outside of Docker:
 
 ```
-make gobuild
-./out/amazon-ecs-agent
+sudo yum install rpm-build
+make generic-rpm
+sudo yum localinstall amazon-ecs-agent-1.53-1.x86_64.rpm
+sudo vi /etc/ecs/ecs.config #add environment variables (ECS_CLUSTER and ECS_DATADIR required)
```
sparrc, line 234: I think ecs-init used to define a default for ECS_DATADIR, should we define that default in agent now?

packaging/generic-rpm/ecs-agent.spec
```diff
@@ -0,0 +1,152 @@
...
+
+%global gobuild_tag generic_rpm
+%global _cachedir %{_localstatedir}/cache
+%global bundled_agent_version %{version}
+%global no_exec_perm 644
+
+%global ecscni_goproject github.com/aws
+%global ecscni_gorepo amazon-ecs-cni-plugins
+%global ecscni_goimport %{ecscni_goproject}/%{ecscni_gorepo}
+%global ecscni_gitrev 55b2ae77ee0bf22321b14f2d4ebbcc04f77322e1
```
sparrc, line 24: Same comment as above, if i understand correctly it seems like this is redundant if these are already submodules of the agent repo

packaging/generic-rpm/ecs-agent.spec
```diff
@@ -0,0 +1,152 @@
...
+
+# Build the ECS CNI plugins
+export GOPATH=%{_builddir}/amazon-ecs-agent
+
+cd %{_builddir}/amazon-ecs-agent/src/github.com/aws/amazon-ecs-cni-plugins
+LD_ECS_CNI_VERSION="-X github.com/aws/amazon-ecs-cni-plugins/pkg/version.Version=$(cat VERSION)"
+ECS_CNI_HASH="%{ecscni_gitrev}"
+LD_ECS_CNI_SHORT_HASH="-X github.com/aws/amazon-ecs-cni-plugins/pkg/version.GitShortHash=${ECS_CNI_HASH::8}"
+LD_ECS_CNI_PORCELAIN="-X github.com/aws/amazon-ecs-cni-plugins/pkg/version.GitPorcelain=0"
+go build -a \
```
sparrc, line 79: If I understand correctly, the current CNI plugins `cd` into the repo of the cni plugins and run `make` commands that are definied by the maintainers of those repos, rather than us deciding on and running our own go build commands. [here](https://github.com/aws/amazon-ecs-agent/blob/master/scripts/dockerfiles/Dockerfile.buildVPCCNIPlugins#L30) and [here](https://github.com/aws/amazon-ecs-agent/blob/master/scripts/dockerfiles/Dockerfile.buildECSCNIPlugins#L24). Is there some particular reason we are defining them ourselves here? I'm particularly also curious about `-buildmode=pie` being added, since I don't see that in the `make plugins` target of amazon-ecs-cni-plugins: https://github.com/aws/amazon-ecs-cni-plugins/blob/master/Makefile

fenxiong: > Is there some particular reason we are defining them ourselves here?

same question. and i'm wondering do we need to build the plugin here at all? it seems to me that we could just let the rpm make target depends on the existing cni-plugins make target, and just use the plugins built from the latter to build the rpm

Makefile
masnoble: While following with how bottle rocket is built, the submodules were not used in the .spec file and have been removed from the branch.

README.md
masnoble: The current solution for the environment variables requires pretty much all of the variables to be set manually. This includes setting the new data directory  as well.
Setting up the default directory would come in a future commit.

packaging/generic-rpm/ecs-agent.spec
masnoble: See the response above... 
The submodules were removed for this build process.

packaging/generic-rpm/ecs-agent.spec
masnoble: Since the submodules were no longer being used, it was easier to just build and create everything inside of the .spec file. This way, the makefile of the plugin repo won't have to be used and the whole build process for the CNI plugins will be in one spot.

mssrivas (dismissed):

Makefile
```diff
@@ -317,4 +330,22 @@ clean:
...
+	-rm -f ./ecs-agent-*.src.rpm
+	-rm -rf ./ecs-agent-*
+	-rm -rf ./BUILDROOT BUILD RPMS SRPMS SOURCES SPECS
+	-rm -rf ./x86_64
+	-rm -f ./amazon-ecs-agent_${VERSION}*
+	-rm -f .srpm-done .rpm-done .generic-rpm-done
+	-rm -f .deb-done
+	-rm -f amazon-ecs-cni-plugins.tar.gz
+	-rm -f amazon-vpc-cni-plugins.tar.gz
+	-rm -rf ../amazon-ecs-cni-plugins
```
sharanyad, line 350: why is it required to remove package as well?

README.md
sharanyad: could you add a TODO for this?

agent/app/agent_capability.go
```diff
@@ -235,7 +235,7 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	capabilities = agent.appendPIDAndIPCNamespaceSharingCapabilities(capabilities)
 
 	// ecs agent version 1.26.0 supports aws-appmesh cni plugin
-	capabilities = agent.appendAppMeshCapabilities(capabilities)
+	//capabilities = agent.appendAppMeshCapabilities(capabilities)
```
sharanyad, line 238: please add a TODO here

agent/config/config.go
```diff
@@ -118,13 +118,13 @@ const (
...
 	// defaultCNIPluginsPath is the default path where cni binaries are located
-	defaultCNIPluginsPath = "/amazon-ecs-cni-plugins"
+	defaultCNIPluginsPath = "/usr/libexec"
 
 	// DefaultMinSupportedCNIVersion denotes the minimum version of cni spec required
 	DefaultMinSupportedCNIVersion = "0.3.0"
 
 	// pauseContainerTarball is the path to the pause container tarball
-	pauseContainerTarballPath = "/images/amazon-ecs-pause.tar"
+	pauseContainerTarballPath = "/var/lib/amazon-ecs-pause.tar"
```
sharanyad, line 127: feels like `/var/lib/ecs/` is the right path for this

packaging/generic-rpm/ecs-agent.spec
sharanyad: is there any advantage to the proposed approach? Is there no way to use submodules as is?

agent/config/config.go
masnoble: I thought so at first as well... but the .tar file ended up in `var/lib`. This directory is the same when bottlerocket agent was moved out of the container.

Makefile
masnoble: I just went ahead and removed all the packages that are brought in or will be brought in as more rpm targets are made. This way, the agent files can be cleaned with just running `make clean`

PR #2970, opened 2021-07-30
added polling for instance status to the tcs client

YashdalfTheGray:
### Summary

- wire the `doctor.Doctor` type through all of the TCS client/session
- moved `getInstanceStatuses` from the doctor to the TCS client so follow existing patterns
- added logic to check for duplicate instance status message and not send it
- added an ack handler to the TCS client
- added polling mechanism to the TCS client to pull the instance status from the doctor

### Implementation details
- Instance status is calculated every time a heartbeat message from the backend is received
- the `doctor.Doctor` class runs all the healthchecks that it knows about when the above happens
- periodically, the backend client polls the doctor for latest unsent instance status
- if a new one is found, it is transformed into the appropriate message format
- the message is then sent out using the existing websocket connect maintained with the backend

### Testing

- `make test`
- tested the change with the backend and saw the ack message in the logs

Enhancement - client now polls the doctor for updates instance status checks and sends them to the backend

### Licensing


YashdalfTheGray:
Closed in favor of #2974 which doesn't have rebased changes along with feature changes.

PR #2974, opened 2021-08-02
added polling for instance status to the tcs client

YashdalfTheGray:
### Summary

- wire the `doctor.Doctor` type through all of the TCS client/session
- moved `getInstanceStatuses` from the doctor to the TCS client so follow existing patterns
- added logic to check for duplicate instance status message and not send it
- added an ack handler to the TCS client
- added polling mechanism to the TCS client to pull the instance status from the doctor

### Implementation details
- Instance status is calculated every time a heartbeat message from the backend is received
- the `doctor.Doctor` class runs all the healthchecks that it knows about when the above happens
- periodically, the backend client polls the doctor for latest unsent instance status
- if a new one is found, it is transformed into the appropriate message format
- the message is then sent out using the existing websocket connect maintained with the backend

### Testing

- `make test`
- tested the change with the backend and saw the ack message in the logs

Enhancement - client now polls the doctor for updates instance status checks and sends them to the backend

### Licensing


agent/doctor/doctor.go
```diff
@@ -35,72 +30,92 @@ type Doctor struct {
...
+func (doc *Doctor) SetStatusReported(statusReported bool) {
+	doc.lock.Lock()
+	defer doc.lock.Unlock()
+
+	doc.statusReported = statusReported
+}
+
+// HasStatusBeenReported returns whether we have already sent the current
+// state of the healthchecks to the backend or not
+func (doc *Doctor) HasStatusBeenReported() bool {
```
shubham2892, line 78: `GetStatusReported` instead of `HasStatusBeenReported` ?

agent/doctor/doctor.go
```diff
@@ -35,72 +30,92 @@ type Doctor struct {
...
+	defer doc.lock.Unlock()
 	allChecksResult := []HealthcheckStatus{}
 
 	for _, healthcheck := range doc.healthchecks {
 		res := healthcheck.RunCheck()
 		seelog.Debugf("instance healthcheck result: %v", res)
 		allChecksResult = append(allChecksResult, res)
 	}
+
+	doc.statusReported = false
```
shubham2892, line 106: Set it using `SetStatusReported()` so that it's under a lock ?

agent/doctor/doctor.go
```diff
@@ -35,72 +30,92 @@ type Doctor struct {
...
-			LastUpdated:      aws.Time(healthcheck.GetLastHealthcheckTime()),
-			Status:           aws.String(healthcheck.GetHealthcheckStatus().String()),
-			Type:             aws.String(healthcheck.GetHealthcheckType()),
-		}
-		instanceStatuses = append(instanceStatuses, instanceStatus)
-	}
-	return instanceStatuses
+	healthcheckCopy := make([]Healthcheck, len(doc.healthchecks))
+	copy(healthcheckCopy, doc.healthchecks)
+	return &healthcheckCopy
```
shubham2892, line 118: Why are we making a copy and returning it instead of directly returning `doc.healthchecks`  ?

agent/tcs/client/client.go
```diff
@@ -345,3 +359,100 @@ func copyTaskHealthMetrics(from []*ecstcs.TaskHealth) []*ecstcs.TaskHealth {
...
+	if err != nil {
+		return err
+	}
+
+	// Make the publish instance status request to the backend.
+	err = cs.MakeRequest(request)
+	if err != nil {
+		return err
+	}
+
```
shubham2892, line 418: Can we move `cs.doctor.SetStatuReported` to this line from line 392. In my opinion it will be more cleaner as it always needs to be updated when publishing the status which is done by this function.

agent/doctor/doctor.go
YashdalfTheGray: We are under a lock already. If I call `SetStatusReported()`, we would run into a deadlock because `SetStatusReported()` would be waiting for a lock that `RunHealthchecks()` is holding and `RunHealthchecks()` won't release the lock until the end of the method.

agent/doctor/doctor.go
YashdalfTheGray: Because we didn't want to pass the internal state of the Doctor around and have it get modified. This way, only one thing changes the internal state of the Doctor and that's when we run the healthchecks. All of the clients get their own copy and they are free to consume it/modify it/destroy it without affecting the internal state of the Doctor. And I was following how the TACS client already works for metrics.

agent/tcs/client/client.go
YashdalfTheGray: Oh yeah, sure.

agent/doctor/doctor.go
YashdalfTheGray: I think `HasStatusBeenReported` gives more information about what the method is and what it returns. You expect a boolean value to come out of a method called `HasStatusBeenReported` vs `GetStatusReported` could potentially return an enum/string/int value as well as a boolean value, it is ambiguous. 

The [Golang source](https://golang.org/src/go/types/predicates.go) itself uses `is` prefixed boolean methods and [Docker](https://github.com/moby/moby/blob/master/cli/cobra.go#L45-L51) uses `has` prefixes.

shubham2892 (dismissed):
🚀

agent/tcs/client/client.go
```diff
@@ -345,3 +359,100 @@ func copyTaskHealthMetrics(from []*ecstcs.TaskHealth) []*ecstcs.TaskHealth {
...
+	instanceStatuses := cs.getInstanceStatuses()
+
+	if instanceStatuses != nil {
+		return &ecstcs.PublishInstanceStatusRequest{
+			Metadata:  metadata,
+			Statuses:  instanceStatuses,
+			Timestamp: aws.Time(time.Now()),
+		}, nil
+	} else {
+		return nil, doctor.EmptyHealthcheckError
```
angelcar, line 439: nit: it is usually cleaner to return the errors first, as in:

```go
func (cs *clientServer) getPublishInstanceStatusRequest() (*ecstcs.PublishInstanceStatusRequest, error) {
    instanceStatuses := cs.getInstanceStatuses()
    if instanceStatuses == nil {
        return nil, doctor.EmptyHealthcheckError
    }
    
    return &ecstcs.PublishInstanceStatusRequest {
        Metadata:  &ecstcs.InstanceStatusMetadata {
            Cluster:                   aws.String(cs.doctor.GetCluster()),
            ContainerInstance: aws.String(cs.doctor.GetContainerInstanceArn()),
            RequestId:               aws.String(uuid.NewRandom().String()),
        },
        Statuses:  instanceStatuses,
        Timestamp: aws.Time(time.Now()),
    }, nil 
}
```

angelcar (approved):

shubham2892 (approved):

PR #2975, opened 2021-08-04
Create new GraphQL Queries and Resolvers

SeaverDahlgren:
### Summary

Added optional arguments to `container` and `containers` resolvers, `id` and `name`. These arguments allow you to search for a specific container by container name or by container id. Additionally, new fields were added to the container query, including `ports`, `volumes`, and `health`.

### Implementation details

The resolver functions for `ports`, `volumes`, and `health` were all implemented by following the same fetch process as the v4 endpoint. The optional arguments for the `container` and `containers` resolvers are implemented passing the current task via `http.request.context`. The current task is the task that the task that the container making the request is a part of. It is important that containers within one task do not have access to container metadata from another task. Retrieving a specific container is then done by accessing preexisting dockerstate mappings.

### Testing

It was tested by running queries, using the new arguments as well as the new fields.

New tests cover the changes: 
yes

Create new GraphQL Queries and Resolvers
### Licensing

### DIFF
--- a/agent/handlers/gql/schema.go
+++ b/agent/handlers/gql/schema.go
@@ -16,14 +16,14 @@ package gql
 import (
 	"time"
 
-	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
-
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
 	"github.com/aws/amazon-ecs-agent/agent/containermetadata"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/handlers/utils"
+	v1 "github.com/aws/amazon-ecs-agent/agent/handlers/v1"
 	v2 "github.com/aws/amazon-ecs-agent/agent/handlers/v2"
 	v3 "github.com/aws/amazon-ecs-agent/agent/handlers/v3"
 	v4 "github.com/aws/amazon-ecs-agent/agent/handlers/v4"
@@ -65,6 +65,10 @@ func CreateSchema(
 			"ImageID": &graphql.Field{
 				Type:    graphql.String,
 				Resolve: containerImageIDResolver,
+			},
+			"Ports": &graphql.Field{
+				Type:    graphql.NewList(JSON),
+				Resolve: containerPortsResolver(state),
 			},
 			"Labels": &graphql.Field{
 				Type:    JSON,
@@ -121,6 +125,14 @@ func CreateSchema(
 			"Networks": &graphql.Field{
 				Type:    graphql.NewList(JSON),
 				Resolve: containerNetworksResolver(state),
+			},
+			"Volumes": &graphql.Field{
+				Type:    graphql.NewList(JSON),
+				Resolve: containerVolumesResolver,
+			},
+			"Health": &graphql.Field{
+				Type:    JSON,
+				Resolve: containerHealthResolver,
 			},
 		},
 	})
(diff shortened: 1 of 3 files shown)


agent/handlers/gql/schema.go
```diff
@@ -404,6 +493,26 @@ func containerNetworksResolver(state dockerstate.TaskEngineState) func(p graphql
...
+	if !ok {
+		return nil, nil
+	}
+	return v1.NewVolumesResponse(dockerContainer), nil
+}
+
+func containerHealthResolver(p graphql.ResolveParams) (interface{}, error) {
+	dockerContainer, ok := p.Source.(*apicontainer.DockerContainer)
+	if !ok {
+		return nil, nil
```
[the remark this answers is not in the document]
SeaverDahlgren: So in practice, it will never reach "!ok" because the source of the call will always be a `*apicontainer.DockerContainer` pointer. It is only in here to satisfy some unit tests that I created. In production, however, it will never enter this conditional.

agent/handlers/gql/schema_test.go
```diff
@@ -222,11 +308,32 @@ var (
...
+	expectedFailedContainerByIdResponse1 = gqlResponse{
+		Errors: []string{"Unable to find container: fakeId"},
+	}
+	expectedFailedContainerByIdResponse2 = gqlResponse{
+		Errors: []string{"Container query only allows for at most one argument"},
+	}
+	expectedFailedContainerByIdResponse3 = gqlResponse{
+		Errors: []string{"Unable to find container: fakeName"},
+	}
+	containerQueryGQL            = "{Container{DockerId,Name,DockerName,KnownStatus,DesiredStatus,CreatedAt,StartedAt,FinishedAt,ExitCode,Image,ImageID,Labels,Type,LogDriver,LogOptions,Limits,ContainerARN,Networks,Ports,Volumes,Health}}"
```
angelcar, line 320: ca we format these queries with multiline strings (i.e. strings delimited with backticks `)

agent/handlers/gql/schema_test.go
```diff
@@ -554,6 +867,9 @@ func TestCreateSchema(t *testing.T) {
 			}
 			// Marshal GraphQL response then unmarshal into the expected response struct
 			// in order to test whether the response fields are correct
+			if tc.query == taskContainersByIdQueryGQL {
+				fmt.Printf("Result: %v\n", res)
+			}
```
angelcar, line 872: why is this needed?

agent/handlers/v2/response.go
```diff
@@ -228,6 +228,24 @@ func GetContainerCPULimit(container *apicontainer.Container) *float64 {
 	}
 }
 
+func GetPortResponse(container *apicontainer.Container, eni *apieni.ENI) []v1.PortResponse {
+	resp := []v1.PortResponse{}
```
angelcar, line 232: should this be initialized to `nil`? (i.e. `var resp []v1.PortResponse`)

agent/handlers/gql/schema.go
```diff
@@ -478,20 +587,49 @@ func taskExecutionStoppedResolver(p graphql.ResolveParams) (interface{}, error)
...
+				container, err := containerByIdHelper(state, task, containerId)
+				if err != nil {
+					return nil, err
+				}
+				resp = append(resp, container)
+				return resp, nil
+			} else {
+				for _, dockerContainer := range containerNameToDockerContainer {
+					resp = append(resp, dockerContainer)
+				}
```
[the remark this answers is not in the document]
angelcar: nvm, that is a map, not an array 😅

angelcar (dismissed):

Realmonia (dismissed):

agent/handlers/gql/schema.go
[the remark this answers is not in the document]
Realmonia: Thanks for explanation. Please add a comment here.

angelcar (approved):

Realmonia (approved):

agent/handlers/gql/schema_test.go
```diff
@@ -532,26 +1255,39 @@ func TestCreateSchema(t *testing.T) {
...
 			params := graphql.Params{
 				Schema:        schema,
 				RequestString: tc.query,
 				Context:       ctx,
 			}
 			res := graphql.Do(params)
-			if tc.taskContainerError || tc.containerCTXError || tc.taskCTXError || tc.taskByIdError ||
-				tc.dockerStatsError || tc.tagsError || tc.pulledContainerError {
+
+			if tc.taskContainerError || tc.taskByIdError || tc.dockerStatsError || tc.tagsError ||
```
chienhanlin, line 1282: nit. Could we split these tests into different sets to have a better readability (and maybe less `if-else` statements)? For example, a set for CTX error, a set for task error, and a set for container error? Thanks!

chienhanlin (approved):

PR #2993, opened 2021-08-20
Fix a issue that agent does not clean task execution credentials from credential manager when stopping a task

Realmonia:
### Summary

Fix a issue that agent does not clean task execution credentials from credential manager when stopping a task

### Implementation details

Introducing task execution credential removal to cleanupTask method 

### Testing

`make test`

Tested with an al2 instance, stop task with taskExecutionRole seems working fine.

Verify if this is related to memory leak issue: tried with a service of which the task definition makes task exits immediately. (run exit in command) and set desired task count to be 10 and run for about 48h. Both instances start with about the same memory usage (13MiB/31.15GiB)

Result

Using dev branch
```
CONTAINER ID        NAME                CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
e92dee603223        ecs-agent           6.92%               468.1MiB / 31.15GiB   1.47%               0B / 0B             0B / 7.63GB         17
```
Using branch with fix
```
CONTAINER ID        NAME                CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O           PIDS
ba0db2aaf082        ecs-agent           7.79%               450.2MiB / 31.15GiB   1.41%               0B / 0B             0B / 7.59GB         18
```
delta is about 18MiB.

The result shows that this issue is a very minor contributing factor to agent memory leak issue. More research on the memory leak issue will be done based on https://github.com/aws/amazon-ecs-agent/pull/3001

Fix a issue that agent does not clean task execution credentials from credential manager when stopping a task

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -350,7 +350,8 @@ func (mtask *managedTask) steadyState() bool {
 	}
 }
 
-// cleanupCredentials removes credentials for a stopped task
+// cleanupCredentials removes credentials for a stopped task (execution credentials are removed in cleanupTask
+// due to its potential usage in the later phase of the task cleanup such as sending logs)
 func (mtask *managedTask) cleanupCredentials() {
 	taskCredentialsID := mtask.GetCredentialsID()
 	if taskCredentialsID != "" {
@@ -1424,6 +1425,7 @@ func (mtask *managedTask) time() ttime.Time {
 }
 
 func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
+	taskExecutionCredentialsID := mtask.GetExecutionCredentialsID()
 	cleanupTimeDuration := mtask.GetKnownStatusTime().Add(taskStoppedDuration).Sub(ttime.Now())
 	cleanupTime := make(<-chan time.Time)
 	if cleanupTimeDuration < 0 {
@@ -1463,6 +1465,14 @@ func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 	go mtask.discardEvents()
 	mtask.engine.sweepTask(mtask.Task)
 	mtask.engine.deleteTask(mtask.Task)
+
+	// Remove TaskExecutionCredentials from credentialsManager
+	if taskExecutionCredentialsID != "" {
+		logger.Info("Cleaning up task's execution credentials", logger.Fields{
+			field.TaskARN: mtask.Arn,
+		})
+		mtask.credentialsManager.RemoveCredentials(taskExecutionCredentialsID)
+	}
 
 	// The last thing to do here is to cancel the context, which should cancel
 	// all outstanding go routines associated with this managed task.
--- a/agent/engine/task_manager_test.go
+++ b/agent/engine/task_manager_test.go
@@ -38,6 +38,7 @@ import (
 	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
+	mock_credentials "github.com/aws/amazon-ecs-agent/agent/credentials/mocks"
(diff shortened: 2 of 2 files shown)


agent/engine/task_manager.go
```diff
@@ -350,7 +350,7 @@ func (mtask *managedTask) steadyState() bool {
 	}
 }
 
-// cleanupCredentials removes credentials for a stopped task
+// cleanupCredentials removes credentials for a stopped task (execution credentials are removed in cleanupTask)
```
yinyic, line 353: Can we elaborate on why? I'm guessing because we need the execution role to call FE stopTask during `cleanupTask`, but would be good if we make it clear here.

yinyic:
It's definitely the right thing to do to clean up execution role creds. I'm curious how confident we are that this is the issue contributing to the memory leak? The [structure](https://github.com/aws/amazon-ecs-agent/blob/3a1d12c7927d7be5ca1b3031ab34823934427306/agent/credentials/manager.go#L52) used for storing creds contains string fields only (and short ones too it seems), which usually won't take megabytes of memory I feel like. 

There could be something else i'm missing. I think it would be great if we could run the experiment a few more times and observe the consistency.

agent/engine/task_manager.go
```diff
@@ -1453,6 +1454,12 @@ func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 		return
 	}
 
+	// Remove TaskExecutionCredentials from credentialsManager
+	if taskExecutionCredentialsID != "" {
```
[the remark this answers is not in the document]
Realmonia: Regarding the memory leak one the testing setup I had is not very correct, I synced with Feng this morning, I will need to change the `ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION` and wait for the full clean up which isn't included in the current test. I will do another round of testing to see more accurate result.

Regarding the move to below deleteTask, I remember when I did it there is a NPE but I forgot what exact error it is. I will get back to this comment with another trial. Thanks!

agent/engine/task_manager.go
Realmonia: The initial aspiration of moving execution role to the later part of clean up process is that unlike container role, execution role might be in use till the last phase of clean up such as sending task cleanup logs to cloudwatch.

As for calling stopTask I don't think it would require executionRole as a task can be run and stopped without an execution role. Correct me if I am wrong :)

Realmonia: Regarding the memory leak one the testing setup I had is not very correct, I synced with Feng this morning, I will need to change the ECS_ENGINE_TASK_CLEANUP_WAIT_DURATION and wait for the full clean up which isn't included in the current test. I will do another round of testing to see more accurate result.

agent/engine/task_manager.go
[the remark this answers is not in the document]
angelcar: +1 on removing creds after `mtask.engine.deleteTask(mtask.Task)`

agent/engine/task_manager.go
```diff
@@ -1464,6 +1466,12 @@ func (mtask *managedTask) cleanupTask(taskStoppedDuration time.Duration) {
 	mtask.engine.sweepTask(mtask.Task)
 	mtask.engine.deleteTask(mtask.Task)
 
+	// Remove TaskExecutionCredentials from credentialsManager
+	if taskExecutionCredentialsID != "" {
+		logger.Info("Cleaning up task's execution credentials")
```
sharanyad, line 1471: minor: can we also log task ID here?

sharanyad (approved):
minor comment, changes lgtm

agent/engine/task_manager.go
Realmonia: Added taskArn here following log format in this file 😀

sharanyad (approved):

angelcar (approved):

Realmonia:
Wait for the release is done to merge --- this is not in the planned release content.

Issue #2981, opened 2021-08-07
Ecs agent logging should be improved if it's failing because of Instance registration attempt with an invalid attribute


rasikmhetre:
### Summary
ECS Agent logging should be improved if the ecs-agent container is exiting with exit code 5 i.e. Error registering attributes: cluster.instance-identifier / Instance registration attempt with an invalid attribute

### Description
Our custom ami sets the value of a custom attribute to register my container instance to the ECS cluster through the bootstrap script. It populates an entries in /etc/ecs/ecs.config file. 
It is very time-consuming to troubleshoot the issue with an error `Instance registration attempt with an invalid attribute`

### Expected Behavior
It should print something like `CharacterLimitExceeded` in the ecs-agent.log file if the attribute name of value exceeds the character limit size i.e. 128, that would be a very easy and time-saving process to troubleshoot the issue.

### Observed Behavior
If the ECS_INSTANCE_ATTRIBUTES value is above 128 characters, ecs-agent just exists or fails with the following error messages in the ecs-agent.log which doesn't help in troubleshooting the issue faster.

`msg="Error registering attributes: cluster.instance-identifier" module=client.go
msg="Error registering: Attribute validation failed" module=agent.go
msg="Instance registration attempt with an invalid attribute" module=agent.go`

### Environment Details
Linux Environment

### Supporting Log Snippets
`level=error time=2021-08-05T08:40:13Z msg="Error registering attributes: cluster.instance-identifier" module=client.go
level=error time=2021-08-05T08:40:13Z msg="Error registering: Attribute validation failed" module=agent.go
level=critical time=2021-08-05T08:40:13Z msg="Instance registration attempt with an invalid attribute" module=agent.go
`

angelcar:
Thanks for pointing this out. Please see https://github.com/aws/amazon-ecs-agent/pull/2999 for the enhancement. Unfortunately, we can't specify the exact error at this time without significant effort. However, the reference to the public documentation added should suffice. 

Please feel free to re-open if this solution is not satisfactory.

PR #2999, opened 2021-08-25
Improve error log when instance reigstration fails due to bad attributes

angelcar:
### Summary
Addresses https://github.com/aws/amazon-ecs-agent/issues/2981

### Implementation details
The validation of the attributes is done in the back-end. When validation fails, the back end doesn't return an actual error, so the agent actually compares attributes in the request vs attributes in the response, and if any are missing we deduct there was a problem, how ever there's no information as to why. 

However, the information about attribute validation is documented publicly, so we add that information in the log for the customer convenience. 

### Testing

UTs suffice

* Improvement of log message displayed when container instance registartion fails due to attribute validation errors.

### Licensing

### DIFF
--- a/agent/api/container/transport.go
+++ b/agent/api/container/transport.go
@@ -56,7 +56,7 @@ func (tp *TransportProtocol) String() string {
 	case TransportProtocolTCP:
 		return tcp
 	default:
-		seelog.Critical("Unknown TransportProtocol type!")
+		seelog.Criticalf("Unknown TransportProtocol type: %s", *tp)
 		return tcp
 	}
 }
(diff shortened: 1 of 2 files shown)


mythri-garaga (approved):

yinyic:
>agent actually compares attributes in the request vs attributes in the response, and if any are missing we deduct there was a problem

Is it possible to include the attributes that fail to register in the error message? I think that'll give customer a better clue.

angelcar:
> Is it possible to include the attributes that fail to register in the error message? I think that'll give customer a better clue.

Good idea. Will do

yinyic (approved):

PR #3032, opened 2021-09-15
prevent instances going into an EC2 Auto Scaling group warm pool from…

lydiafilipe:
### Summary
EC2 Auto Scaling recently launched [warm pools](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-warm-pools.html) for ASGs. Currently, since the ECS agent does not distinguish between instances going to the warm pool and those actually going into the cluster, warm pool instances may get registered to the cluster. In this change, 

### Implementation details
- add a new env variable ECS_WARM_POOLS_CHECK that indicates whether to check for the target lifecycle state of an instance, which will be populated by EC2 Autoscaling
- add logic to poll IMDS for this target lifecycle state, and only proceed with registering to the cluster if the instance is in service

### Testing
This was tested on both Linux and Windows. Using a test program, we manually updated metadata on the instances to include the new autoscaling field. On both Linux and Windows, we tested that 1) an instance with an in service target state does not poll 2) an instance with a warmed target state polls until the sstate changes to in service 3) the polling/checks are not done if the new config variable is not set.

Feature - introduces new environment variable ECS_WARM_POOLS_CHECK, which defaults to false. When enabled, it ensures only instances going in service, and not into [autoscaling warm pools](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-warm-pools.html), are registered with the cluster,

### Licensing


agent/config/config.go
```diff
@@ -591,6 +591,8 @@ func environmentConfig() (Config, error) {
 		FSxWindowsFileServerCapable:         parseFSxWindowsFileServerCapability(),
 		External:                            parseBooleanDefaultFalseConfig("ECS_EXTERNAL"),
 		EnableRuntimeStats:                  parseBooleanDefaultFalseConfig("ECS_ENABLE_RUNTIME_STATS"),
+		ShouldExcludeIPv6PortBinding:        parseBooleanDefaultTrueConfig("ECS_EXCLUDE_IPV6_PORTBINDING"),
```
ubhattacharjya, line 594: This probably is not a part of this PR

lydiafilipe: Ah yeah, the branch in aws was cloned from a version of dev before those changes got push, let me fix that

agent/app/agent.go
```diff
@@ -377,6 +396,44 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+func (agent *ecsAgent) waitUntilInstanceInService() error {
+	var err error
+	var targetState string
+	// Poll while the instance is in a warmed state or while waiting for the data to be populated.
+	// If the data is not populated after a certain number of polls, then stop polling and return the not found error.
+	// The polling maximum does not apply to instances in the warmed states
+	for i := 0; i < asgLifeCyclePollMax || targetState != ""; i++ {
+		targetState, err = agent.getTargetLifeCycle()
+		// stop polling if the retrieved state is in service or we get an unexpected error
+		if (targetState == inServiceState) || (err != nil && !strings.Contains(err.Error(), "status code: 404")) {
```
ubhattacharjya, line 410: Could we check if err.Code() returns the error code?

lydiafilipe: I don't think Code() is part of the error interface, just the error strings. And IMDS returns the HTML, which isn't ideal for checking details like this. I hate to rely on string contains for this, but I think trying to parse the error would involve more hardcoding and unnecessary extra work

.github/workflows/windows.yml
```diff
@@ -31,4 +31,4 @@ jobs:
           gcc --version
           $env:ZZZ_SKIP_WINDOWS_SERVER_VERSION_CHECK_NOT_SUPPORTED_IN_PRODUCTION = 'true'
           $packages=go list .\... | Where-Object {$_ -NotMatch 'vendor'}
-          go test -v -tags unit -timeout=40s $packages
+          go test -v -tags unit -timeout=3m $packages
```
mssrivas, line 34: Thats a big jump. What happened? Similar for below

lydiafilipe: I added a couple tests for polling, and the wait for that is a minute. A couple tests can probably be clubbed to decrease this increase, though there will be some increase still

agent/app/agent.go
```diff
@@ -283,6 +293,15 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		}
 	}
 
+	// If part of ASG, wait until instance is being set up to go in service before registering with cluster
+	if agent.cfg.WarmPoolsSupport.Enabled() {
+		err := agent.waitUntilInstanceInService()
+		if err != nil && err.Error() != blackholed {
+			seelog.Criticalf("Could not determine target lifecycle of instance: %v", err)
+			return exitcodes.ExitError
```
sharanyad, line 301: do we want agent to be restarted again after this exit? 
since the wait seems to be already longer, we should probably exit with exitTerminal code.

.github/workflows/windows.yml
fenxiong: could we overwrite the wait time to a smaller value in test? example:
source:
```
const defaultXXXTimeout = time.Minute
var xxxTimeout = defaultXXXTimeout
```
beginning of test:
```
xxxTimeout = time.Second
defer func() { xxxTimeout = defaultXXXTimeout }()
```

agent/app/agent.go
```diff
@@ -377,6 +396,44 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+	var err error
+	backoff := retry.NewExponentialBackoff(targetLifecycleBackoffMin, targetLifecycleBackoffMax, targetLifecycleBackoffJitter, targetLifecycleBackoffMultiple)
+	for i := 0; i < targetLifecycleMaxRetryCount; i++ {
+		targetState, err = agent.ec2MetadataClient.TargetLifecycleState()
+		if err == nil {
+			break
+		} else if i < targetLifecycleMaxRetryCount {
+			time.Sleep(backoff.Duration())
+		}
+		seelog.Warnf("Error when getting intended lifecycle state: %v", err)
```
sharanyad, line 431: can we print this error before sleeping?

agent/app/agent.go
```diff
@@ -377,6 +396,44 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+// associated with an ASG
+func (agent *ecsAgent) getTargetLifeCycle() (string, error) {
+	var targetState string
+	var err error
+	backoff := retry.NewExponentialBackoff(targetLifecycleBackoffMin, targetLifecycleBackoffMax, targetLifecycleBackoffJitter, targetLifecycleBackoffMultiple)
+	for i := 0; i < targetLifecycleMaxRetryCount; i++ {
+		targetState, err = agent.ec2MetadataClient.TargetLifecycleState()
+		if err == nil {
+			break
+		} else if i < targetLifecycleMaxRetryCount {
```
sharanyad, line 428: this can just be `if` instead of `else if`

agent/app/agent_test.go
```diff
@@ -1473,3 +1533,55 @@ func newTestDataClient(t *testing.T) (data.Client, func()) {
 	}
 	return testClient, cleanup
 }
+
+func TestWaitUntilInstanceInService(t *testing.T) {
```
sharanyad, line 1537: could you make this and below 3 tests table driven? 
https://github.com/golang/go/wiki/TableDrivenTests

.github/workflows/windows.yml
sharanyad: +1, lets make sure unit tests are as quick as possible :)

README.md
```diff
@@ -194,6 +194,8 @@ additional details on each available environment variable.
...
 | `ECS_ENABLE_RUNTIME_STATS` | `true` | Determines if [pprof](https://pkg.go.dev/net/http/pprof) is enabled for the agent. If enabled, the different profiles can be accessed through the agent's introspection port (e.g. `curl http://localhost:51678/debug/pprof/heap > heap.pprof`). In addition, agent's [runtime stats](https://pkg.go.dev/runtime#ReadMemStats) are logged to `/var/log/ecs/runtime-stats.log` file. | `false` | `false` |
+| `ECS_EXCLUDE_IPV6_PORTBINDING` | `true` | Determines if agent should exclude IPv6 port binding using default network mode. If enabled, IPv6 port binding will be filtered out, and the response of DescribeTasks API call will not show tasks' IPv6 port bindings, but it is still included in Task metadata endpoint. | `true` | `true` |
+| `ECS_WARM_POOLS_CHECK` | `true` | Whether to ensure instances going into an [EC2 Auto Scaling group warm pool](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-warm-pools.html) are prevented from being registered with the cluster. Set only if using EC2 Autoscaling | `false` | `false` |
```
sharanyad, line 198: nit: set to `true` only if..

agent/app/agent.go
mssrivas: Should it be a warn or a debug?

lydiafilipe: Good point. Was initially thinking warn so we are aware of errors that occur, which aren't always expected, though if it results in a failure then we are logging it before exiting anyway. I'll change to debug

sharanyad (approved):
:shipit:

ubhattacharjya (approved):

PR #3044, opened 2021-09-28
Added support for os-family attribute to be advertised by the agent for fine grained task placement

rawahars:
### Summary

A new attribute is being introduced in the ECS Control plane called `ecs.os-family` which defines the os family to which the instance belongs. For Linux based instances, the OS family would be advertised as `linux` and for Windows based instances, the attribute would have a value of format - `WINDOWS_SERVER_<Windows Version>_<Installation Type>`

This would allow the customers to achieve fine grained task placement especially in a mixed cluster.

Example- If a cluster runs instances running on different Windows versions, and a task definition specifies the OS family as `WINDOWS_SERVER_2019_FULL` then this task would only be placed on the instances which broadcasted their os-family as `WINDOWS_SERVER_2019_FULL`.

### Implementation details

A new attribute has been introduced `ecs.os-family`. The values for Linux based instances is `linux`.
For Windows instances, the value can be-
`WINDOWS_SERVER_2019_FULL`, `WINDOWS_SERVER_2019_CORE`, `WINDOWS_SERVER_2016_FULL`, `WINDOWS_SERVER_2022_FULL`, `WINDOWS_SERVER_2022_CORE`, `WINDOWS_SERVER_2004_CORE`,  `WINDOWS_SERVER_20H2_CORE`

### Testing

Unit tests have been added and existing tests were updated.
Additionally, custom AMIs with custom agent built from this branch was tested to ensure that the instance can register itself to the cluster.

New tests cover the changes: 
Yes

Added support for OS Family to the agent.
### Licensing

### DIFF
--- a/agent/config/os_family_windows.go
+++ b/agent/config/os_family_windows.go
@@ -1,0 +1,111 @@
+// +build windows
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package config
+
+import (
+	"fmt"
+
+	"github.com/aws/amazon-ecs-agent/agent/statemanager/dependencies"
+
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
+)
+
+const (
+	// The os-family for windows is of the format "WINDOWS_SERVER_<Release>_<FULL or CORE>"
+	osFamilyFormat             = "WINDOWS_SERVER_%s_%s"
+	windowsServerCore          = "CORE"
+	windowsServerFull          = "FULL"
+	unsupportedWindowsOSFamily = "windows"
+
+	installationTypeServerCore = "Server Core"
+	installationTypeServer     = "Server"
+
+	ecsWinRegistryRootKey  = registry.LOCAL_MACHINE
+	ecsWinRegistryRootPath = `SOFTWARE\Microsoft\Windows NT\CurrentVersion`
+)
+
+var (
+	winRegistry dependencies.WindowsRegistry
+	// osReleaseFromBuildNumber is the mapping from build number to the Windows release.
+	// Reference- https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info
+	osReleaseFromBuildNumber = map[int]string{
+		14393: "2016",
+		17763: "2019",
+		19041: "2004",
+		19042: "20H2",
+		20348: "2022",
+	}
+	// windowsGetVersionFunc is the method used to obtain information about current Windows version.
+	windowsGetVersionFunc = windows.RtlGetVersion
+)
+
+func init() {
+	winRegistry = dependencies.StdRegistry{}
+}
+
+// getInstallationType returns the installation type as either CORE or FULL.
+func getInstallationType(installationType string) (string, error) {
+	switch installationType {
+	case installationTypeServer:
+		return windowsServerFull, nil
+	case installationTypeServerCore:
+		return windowsServerCore, nil
+	default:
+		return "", errors.Errorf("unsupported installation type: %s", installationType)
+	}
+}
+
+// GetOSFamily returns the operating system family string.
+// On Windows, it would be of the format "WINDOWS_SERVER_<Release>_<FULL or CORE>"
+// In case of any exception this method just returns "windows" as operating system family.
+func GetOSFamily() string {
+	// Find the build number of the os on which agent is running.
+	versionInfo := windowsGetVersionFunc()
+	buildNumber := int(versionInfo.BuildNumber)
+	osRelease, ok := osReleaseFromBuildNumber[buildNumber]
+	if !ok {
+		seelog.Errorf("windows release with build number [%d] is unsupported", buildNumber)
+		return unsupportedWindowsOSFamily
+	}
+
+	// Find the installation type from the Windows registry.
+	key, err := winRegistry.OpenKey(ecsWinRegistryRootKey, ecsWinRegistryRootPath, registry.QUERY_VALUE)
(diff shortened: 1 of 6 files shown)


agent/config/os_family_windows.go
```diff
@@ -0,0 +1,129 @@
...
+	case installationTypeServerCore:
+		return windowsServerCore, nil
+	default:
+		return "", errors.Errorf("unsupported installation type: %s", installationType)
+	}
+}
+
+// GetOperatingSystemFamily reads the NT current version from windows registry and constructs operating system family string.
+// In case of any exception this method just returns "windows" as operating system type.
+func GetOperatingSystemFamily() string {
```
fenxiong, line 81: nit: can this just be GetOSFamily?

agent/config/os_family_windows.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/pkg/errors"
+	"golang.org/x/sys/windows/registry"
+)
+
+const (
+	// The os-family for windows is of the format "WINDOWS_SERVER_<Version>_<FULL or CORE>"
+	osTypeFormat         = "WINDOWS_SERVER_%s_%s"
+	windowsServerCore    = "CORE"
+	windowsServerFull    = "FULL"
+	unsupportedWindowsOS = "windows"
```
fenxiong, line 33: nit: unsupportedWindowsOS -> unsupportedWindowsOSFamily

agent/config/os_family_windows.go
```diff
@@ -0,0 +1,129 @@
...
+		"21H2": "2022",
+		"20H2": "20H2",
+		"2004": "2004",
+	}
+	// For few Windows versions, the release ID set in registry coincides with other versions
+	// and is distinguished via another key "DisplayVersion".
+	// Reference- https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info
+	// The list is as below-
+	// 2009 - Windows Server 2022 and 20H2 both have release ID as 2009 in the registry but the actual release version is different.
+	releaseIDToVersionExceptions = map[string]bool{
```
fenxiong, line 58: nit: releaseIDFromDisplayVersion sounds easier to understand

fenxiong (commented):
looks good few nits

agent/config/os_family_windows.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/cihub/seelog"
+	"github.com/pkg/errors"
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/registry"
+)
+
+const (
+	// The os-family for windows is of the format "WINDOWS_SERVER_<Release>_<FULL or CORE>"
+	osTypeFormat               = "WINDOWS_SERVER_%s_%s"
```
jterry75, line 31: This is really the `osFamilyFormat` for clarity

agent/config/os_family_windows.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+var (
+	winRegistry dependencies.WindowsRegistry
+	// osReleaseFromBuildNumber is the mapping from build number to the Windows release.
+	// Reference- https://docs.microsoft.com/en-us/windows-server/get-started/windows-server-release-info
+	osReleaseFromBuildNumber = map[int]string{
+		14393: "2016",
+		17763: "2019",
+		20348: "2022",
```
jterry75, line 50: Can we just keep these in build order? So this one should be last

agent/config/os_family_windows.go
```diff
@@ -0,0 +1,111 @@
...
+// GetOSFamily returns the operating system family string.
+// On Windows, it would be of the format "WINDOWS_SERVER_<Release>_<FULL or CORE>"
+// In case of any exception this method just returns "windows" as operating system family.
+func GetOSFamily() string {
+	// Find the build number of the os on which agent is running.
+	versionInfo := windowsGetVersionFunc()
+	buildNumber := int(versionInfo.BuildNumber)
+	osRelease, ok := osReleaseFromBuildNumber[buildNumber]
+	if !ok {
+		seelog.Errorf("windows release with build number [%s] is unsupported", buildNumber)
```
jterry75, line 83: `buildNumber` is an `int` so this should be `%d` or `%v` right?

jterry75 (changes requested):

jterry75 (approved):

fenxiong (approved):

fierlion (approved):
good overall test coverage!

singholt:
Just curious - What happens if a customer tries to use both os-type and os-family attributes at once, and they conflict? Example - os-type = "linux" and os-family = "windows-2019...". Do we want to add checks for such a scenario, even if it may be accidental?

rawahars:
@singholt With respect to the conflicting task placement constraints such as-

```
"placementConstraints": [
    {
      "type": "memberOf",
      "expression": "attribute:ecs.os-type == linux"
    },
   {
      "type": "memberOf",
      "expression": "attribute:ecs.os-family == WINDOWS_SERVER_2019_FULL"
    }
  ]
```

The run task call itself will fail with an error message such as `Reasons : ["MemberOf placement constraint unsatisfied.","AGENT"]. Learn more`

This is the existing behavior with any task placement constraint using the custom attribute tags. If there isn't any instance which matches both the constraints, the task would fail.

singholt:
> @singholt With respect to the conflicting task placement constraints such as-
> 
> ```
> "placementConstraints": [
>     {
>       "type": "memberOf",
>       "expression": "attribute:ecs.os-type == linux"
>     },
>    {
>       "type": "memberOf",
>       "expression": "attribute:ecs.os-family == WINDOWS_SERVER_2019_FULL"
>     }
>   ]
> ```
> 
> The run task call itself will fail with an error message such as `Reasons : ["MemberOf placement constraint unsatisfied.","AGENT"]. Learn more`
> 
> This is the existing behavior with any task placement constraint using the custom attribute tags. If there isn't any instance which matches both the constraints, the task would fail.

I agree. I'm thinking around whether we want to fail early, at register-task call?

rawahars:
ECS doesn't fail the register-task call even if we specify any random value for the task placement constraint attribute. For example, we can register a task definition with say the below placement constraint. The task would be failed during run-task.

```
"placementConstraints": [
    {
      "type": "memberOf",
      "expression": "attribute:ecs.os-type == invalidOS"
    }
  ]
```

singholt:
> ECS doesn't fail the register-task call even if we specify any random value for the task placement constraint attribute. For example, we can register a task definition with say the below placement constraint. The task would be failed during run-task.
> 
> ```
> "placementConstraints": [
>     {
>       "type": "memberOf",
>       "expression": "attribute:ecs.os-type == invalidOS"
>     }
>   ]
> ```
> 
> The same is applicable for any conflicting attributes.

thanks, makes sense

PR #3086, opened 2021-11-11
update dependency graph

Realmonia:
### Summary

This PR updates methods used when resolving dependencies of containers and updates the invocation of the method.

### Implementation details

* Use container specific resource to override resource map when checking resource dependencies when container resources are presented
* Update container execution credential check in dependency resolution
* Add a test when there is a dependency between container and container resource

### Testing

### Licensing

### DIFF
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -124,7 +124,8 @@ func DependenciesAreResolved(target *apicontainer.Container,
 	by []*apicontainer.Container,
 	id string,
 	manager credentials.Manager,
-	resources []taskresource.TaskResource,
+	containerResources []taskresource.TaskResource,
+	taskResources []taskresource.TaskResource,
 	cfg *config.Config) (*apicontainer.DependsOn, error) {
 	if !executionCredentialsResolved(target, id, manager) {
 		return nil, CredentialsNotResolvedErr
@@ -140,7 +141,13 @@ func DependenciesAreResolved(target *apicontainer.Container,
 	}
 
 	resourcesMap := make(map[string]taskresource.TaskResource)
-	for _, resource := range resources {
+	for _, resource := range taskResources {
+		resourcesMap[resource.GetName()] = resource
+	}
+
+	// Override existing resources map with container resources because when container specific resources are specified,
+	// we want to avoid using task resources.
+	for _, resource := range containerResources {
 		resourcesMap[resource.GetName()] = resource
 	}
 
@@ -211,8 +218,13 @@ func executionCredentialsResolved(target *apicontainer.Container, id string, man
 		return true
 	}
 
-	_, ok := manager.GetTaskCredentials(id)
-	return ok
+	if target.ExecutionCredentialsID != "" {
+		_, ok := manager.GetContainerCredentials(id)
+		return ok
+	} else {
+		_, ok := manager.GetTaskCredentials(id)
+		return ok
+	}
 }
 
 // verifyStatusResolvable validates that `target` can be resolved given that
--- a/agent/engine/dependencygraph/graph_test.go
+++ b/agent/engine/dependencygraph/graph_test.go
@@ -20,6 +20,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/credentials"
+
+	mock_credentials "github.com/aws/amazon-ecs-agent/agent/credentials/mocks"
+
 	"github.com/aws/amazon-ecs-agent/agent/containerresource"
 	"github.com/aws/amazon-ecs-agent/agent/containerresource/containerstatus"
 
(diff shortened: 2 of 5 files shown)


agent/engine/dependencygraph/graph.go
```diff
@@ -211,8 +218,13 @@ func executionCredentialsResolved(target *apicontainer.Container, id string, man
 		return true
 	}
 
-	_, ok := manager.GetTaskCredentials(id)
-	return ok
+	if target.ExecutionCredentialsID != "" {
+		_, ok := manager.GetContainerCredentials(id)
+		return ok
```
angelcar, line 223: I wonder if we want to fall back to task credentials when `!ok` here?

agent/engine/dependencygraph/graph_test.go
```diff
@@ -476,6 +479,14 @@ func TestVerifyResourceDependenciesResolved(t *testing.T) {
 			RequiredStatus:   resourcestatus.ResourceStatus(1),
 			ExpectedResolved: true,
 		},
+		{
+			Name:             "resource none,container pull depends on resource created",
```
angelcar, line 483: isn't this UT duplicated?

agent/engine/dependencygraph/graph_test.go
```diff
@@ -1397,3 +1408,62 @@ func TestVerifyContainerOrderingStatusResolvableFailOnDependencyWontStart(t *tes
...
+		},
+		ExecutionCredentialsID: "id",
+	}
+
+	manager := mock_credentials.NewMockManager(ctrl)
+	manager.EXPECT().GetContainerCredentials("id").Return(credentials.ContainerIAMRoleCredentials{}, true).Times(1)
+	manager.EXPECT().GetTaskCredentials("id").Return(credentials.TaskIAMRoleCredentials{}, true).Times(0)
+
+	resolve := executionCredentialsResolved(target, "id", manager)
+	assert.Equal(t, true, resolve)
```
angelcar, line 1439: nit: `assert.True`

agent/engine/dependencygraph/graph_test.go
```diff
@@ -1397,3 +1408,62 @@ func TestVerifyContainerOrderingStatusResolvableFailOnDependencyWontStart(t *tes
 	_, err := verifyContainerOrderingStatusResolvable(target, contMap, &config.Config{}, dummyResolves)
 	assert.Error(t, err)
 }
+
+func TestVerifyContainerExecutionCredentialResolve(t *testing.T) {
```
angelcar, line 1412: nit: this and below tests seem like a good fit for table-driven test.

agent/engine/ordering_integ_test.go
```diff
@@ -380,6 +383,51 @@ func TestDependencyHealthyTimeout(t *testing.T) {
 	waitFinished(t, finished, orderingTimeout)
 }
 
+// TestContainerPullDependsOnContainerSpecificResource
+func TestContainerPullDependsOnContainerSpecificResource(t *testing.T) {
```
angelcar, line 387: how is this testing the new code?

Realmonia: I should probably add another test on the failed case but this is just making sure dependency resolution is working when a container creation is depending on container specific resources of the same container

agent/engine/dependencygraph/graph_test.go
Realmonia: Oh sorry it is.. I did a refactor and removed the changes that need this new test case but forgot to remove the test.

agent/engine/dependencygraph/graph.go
Realmonia: I don't think so --- when a container has container credentials specified, it could have different set of privilege with task credentials, fall back to task credentials might cause unexpected access to resources

angelcar: could we double check that that's the desired behavior? 
Previous to this change, containers get credentials from tasks level, fall back to instance-level creds. Why should this be different?

PettitWesley: There are two fallback cases:
1. Fallback if there are no container credentials. We DEFINITELY want to fall back in this case- its part of the defined CX. If there is no container role, then we must use the task def role. 
2. Fallback if we fail to use the credentials somehow (ex fail to pull secret). I am unsure about this case. Having a chat conversation folks on this.

Realmonia: For the container get credentials from task level and fall back to instance-level, can you point where we did it? we do have back up for credentials providers but I don't think there is a fall back strategy for credentials. Just checked with Wesley, his opinion is that if we don't have existing fall back strategies then we might not want to introduce a new one.

angelcar: Sorry for the confusion. You are correct in that only credential providers which are related to task roles have the fallback mechanism. Task execution role doesn't.

angelcar (dismissed):

mythri-garaga (dismissed):

mythri-garaga (approved):

angelcar (approved):

mythri-garaga (approved):

PR #3087, opened 2021-11-12
added support for metrics when using awsvpc mode on Windows

rawahars:
### Summary

Currently, the metrics support was available only for tasks running in default mode on Windows. The changes introduced in this PR extends this support for tasks running in awsvpc network mode as well.

### Implementation details

For Windows, `Get-NetAdapterStatistics` command is invoked for the given task ENI and the stats from the same are recorded. The existing Linux workflow is leveraged by segregating them into Windows and Linux platform specific files.

### Testing

Unit and integ tests were ran. Additionally, a custom agent binary was tested on WS 2016, 2019, and 2022.

New tests cover the changes: 
Yes

Added support for metrics when using awsvpc network mode on Windows
### Licensing

### DIFF
--- a/agent/stats/task.go
+++ b/agent/stats/task.go
@@ -1,0 +1,155 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package stats
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/stats/resolver"
+	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
+
+	"github.com/cihub/seelog"
+	"github.com/docker/docker/api/types"
+	dockerstats "github.com/docker/docker/api/types"
+)
+
+// statsTaskCommon contains the common fields in StatsTask for both Linux and Windows.
+// StatsTask abstracts methods to gather and aggregate network data for a task. Used only for AWSVPC mode.
+type statsTaskCommon struct {
+	StatsQueue            *Queue
+	TaskMetadata          *TaskMetadata
+	Ctx                   context.Context
+	Cancel                context.CancelFunc
+	Resolver              resolver.ContainerMetadataResolver
+	metricPublishInterval time.Duration
+}
+
+func (taskStat *StatsTask) StartStatsCollection() {
+	queueSize := int(config.DefaultContainerMetricsPublishInterval.Seconds() * 4)
+	taskStat.StatsQueue = NewQueue(queueSize)
+	taskStat.StatsQueue.Reset()
+	go taskStat.collect()
+}
+
+func (taskStat *StatsTask) StopStatsCollection() {
+	taskStat.Cancel()
+}
+
+func (taskStat *StatsTask) collect() {
+	taskArn := taskStat.TaskMetadata.TaskArn
+	backoff := retry.NewExponentialBackoff(time.Second*1, time.Second*10, 0.5, 2)
+
+	for {
+		err := taskStat.processStatsStream()
+		select {
+		case <-taskStat.Ctx.Done():
+			seelog.Debugf("Stopping stats collection for taskStat %s", taskArn)
+			return
+		default:
+			if err != nil {
+				d := backoff.Duration()
+				time.Sleep(d)
+				seelog.Debugf("Error querying stats for task %s: %v", taskArn, err)
+			}
+			// We were disconnected from the stats stream.
+			// Check if the task is terminal. If it is, stop collecting metrics.
+			terminal, err := taskStat.terminal()
+			if err != nil {
+				// Error determining if the task is terminal. clean-up anyway.
+				seelog.Warnf("Error determining if the task %s is terminal, stopping stats collection: %v",
+					taskArn, err)
+				taskStat.StopStatsCollection()
+			} else if terminal {
+				seelog.Infof("Task %s is terminal, stopping stats collection", taskArn)
+				taskStat.StopStatsCollection()
+			}
+		}
+	}
+}
+
+func (taskStat *StatsTask) processStatsStream() error {
+	taskArn := taskStat.TaskMetadata.TaskArn
+	awsvpcNetworkStats, errC := taskStat.getAWSVPCNetworkStats()
+
+	returnError := false
+	for {
+		select {
+		case <-taskStat.Ctx.Done():
+			seelog.Info("task context is done")
+			return nil
+		case err := <-errC:
+			seelog.Warnf("Error encountered processing metrics stream from host, this may affect "+
+				"cloudwatch metric accuracy: %s", err)
+			returnError = true
+		case rawStat, ok := <-awsvpcNetworkStats:
+			if !ok {
+				if returnError {
+					return fmt.Errorf("error encountered processing metrics stream from host")
+				}
+				return nil
+			}
+			if err := taskStat.StatsQueue.Add(rawStat); err != nil {
+				seelog.Warnf("Task [%s]: error converting stats: %v", taskArn, err)
+			}
+		}
+
+	}
+}
+
+func (taskStat *StatsTask) terminal() (bool, error) {
+	resolvedTask, err := taskStat.Resolver.ResolveTaskByARN(taskStat.TaskMetadata.TaskArn)
+	if err != nil {
+		return false, err
+	}
+	return resolvedTask.GetKnownStatus() == apitaskstatus.TaskStopped, nil
+}
+
+func (taskStat *StatsTask) getAWSVPCNetworkStats() (<-chan *types.StatsJSON, <-chan error) {
+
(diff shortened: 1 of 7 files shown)


agent/stats/task.go
```diff
@@ -0,0 +1,155 @@
...
+	"github.com/docker/docker/api/types"
+	dockerstats "github.com/docker/docker/api/types"
+)
+
+// statsTaskCommon contains the common fields in StatsTask for both Linux and Windows.
+// StatsTask abstracts methods to gather and aggregate network data for a task. Used only for AWSVPC mode.
+type statsTaskCommon struct {
+	StatsQueue            *Queue
+	TaskMetadata          *TaskMetadata
+	Ctx                   context.Context
```
jterry75, line 36: This really is a golang anit-pattern to expose the ctx and cancel. Typically the caller would provide these and you can consume them here. So if we wanted to do this there would be a:

`func NewStatsTask(ctx context)` ... and then we would just wait entirely on this context to be done. The caller is then in control of when to exit etc based on its logic.

agent/stats/task_windows.go
```diff
@@ -0,0 +1,179 @@
...
+type StatsTask struct {
+	*statsTaskCommon
+}
+
+func newStatsTaskContainer(taskARN string, containerPID string, numberOfContainers int,
+	resolver resolver.ContainerMetadataResolver, publishInterval time.Duration, taskENIs task.TaskENIs) (*StatsTask, error) {
+	ctx, cancel := context.WithCancel(context.Background())
+
+	devices := make([]string, len(taskENIs))
+	for count, device := range taskENIs {
```
jterry75, line 44: `count` threw me off here. Its really `index`

agent/stats/task_windows.go
```diff
@@ -0,0 +1,179 @@
...
+			return nil, err
+		}
+		networkStats[device] = *networkAdaptorStatistics
+	}
+
+	return networkStats, nil
+}
+
+// getNetworkAdaptorStatistics returns the network statistics per container for the given network interface.
+func (taskStat *StatsTask) getNetworkAdaptorStatistics(device string) (*dockerstats.NetworkStats, error) {
```
jterry75, line 82: Oh yuck! Can we not query this from the HNS directly?

agent/stats/task_windows.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return nil, errors.Errorf("failed to parse network stats for %s : OutboundPacketErrors", device)
+	}
+
+	outboundDiscardedPackets, err := strconv.ParseUint(m["OutboundDiscardedPackets"], 10, 64)
+	if err != nil {
+		return nil, errors.Errorf("failed to parse network stats for %s : OutboundDiscardedPackets", device)
+	}
+
+	numberOfContainers := uint64(taskStat.TaskMetadata.NumberContainers)
```
jterry75, line 167: Is there ever a case this could be 0? Should we do a safe check here or just allow the panic below?

agent/stats/task_windows.go
```diff
@@ -0,0 +1,179 @@
...
+	// The statistics include broadcast, multicast, discards, and errors.
+	cmd := "Get-NetAdapterStatistics -Name \"" + device + "\" | Format-List -Property *"
+	out, err := execCommand("powershell", "-Command", cmd).CombinedOutput()
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to run Get-NetAdapterStatistics for %s", device)
+	}
+	str := string(out)
+
+	lines := strings.Split(str, "\n")
```
jterry75, line 94: I think the test string below will have only \n because of the go formatter but will PowerShell output `\r\n`?

jterry75 (commented):
Few Comments

agent/stats/task_windows.go
rawahars: This value represents the number of containers present in the task and is not dynamically calculated. Any task would have atleast 1 container and that would be validated in MTS. Therefore, we should be fine here without any checks.

agent/stats/task_windows.go
rawahars: Yes Powershell would output `\r\n` but we are using `strings.TrimSpace` before adding the key and value into the map. This method would remove any additional spaces and `\r` as well.

agent/stats/task_windows.go
rawahars: Made changes for this one.

agent/stats/task.go
rawahars: So, `StatsTask` is responsible for the entire lifecycle of stats collection of the container associated with one instance of StatsTask. When the stats collection needs to be stopped due to any reason, the `StopStatsCollection` method would be invoked on the object which in turn would cancel the context.

Reference- https://github.com/aws/amazon-ecs-agent/blob/fe79b107f3b204ca1a831b1d15424d12d2948bdf/agent/stats/task_linux.go#L116

jterry75: I know you didnt change this code so just leave it for now thats fine.

agent/stats/task_windows.go
jterry75: Deal

agent/stats/task_windows.go
jterry75: I did not know that `\r` was considered whitespace! Just verified this is true. Cool!

jterry75 (dismissed):
LGTM

agent/stats/task_windows.go
```diff
@@ -0,0 +1,179 @@
...
+
+	return networkStats, nil
+}
+
+// getNetworkAdaptorStatistics returns the network statistics per container for the given network interface.
+func (taskStat *StatsTask) getNetworkAdaptorStatistics(device string) (*dockerstats.NetworkStats, error) {
+	// Ref: https://docs.microsoft.com/en-us/powershell/module/netadapter/get-netadapterstatistics?view=windowsserver2019-ps
+	// The Get-NetAdapterStatistics cmdlet gets networking statistics from a network adapter.
+	// The statistics include broadcast, multicast, discards, and errors.
+	cmd := "Get-NetAdapterStatistics -Name \"" + device + "\" | Format-List -Property *"
```
angelcar, line 86: this can potentially be exploited to execute a command on the host. We should sanitize the `device` before executing.

agent/stats/task_windows.go
```diff
@@ -0,0 +1,179 @@
...
+		kv := strings.Split(line, ":")
+		if len(kv) != 2 {
+			continue
+		}
+		key := strings.TrimSpace(kv[0])
+		value := strings.TrimSpace(kv[1])
+		m[key] = value
+	}
+
+	receivedBroadcastPackets, err := strconv.ParseUint(m["ReceivedBroadcastPackets"], 10, 64)
```
angelcar, line 107: can these keys (e.e. `ReceivedBroadcastPackets`) be trusted not to change in the future? Is there a more reliable way than hard-coding?

rawahars: Based on an earlier discussion with Justin, it seems like this would be the way to go for now. 

We can possibly add an integ test for this which would execute this method and check that the stats are available. That way we can ensure that all the keys are as it is with each update.
Thoughts?

agent/stats/task_windows.go
rawahars: The device name present here is obtained by the agent at runtime (using Win32 call via net interface) when the task ENI is attached to the instance. Since this string is not sent by ACS or is not part of any user data, I feel we can skip the sanitization of this variable.

Reference- https://github.com/aws/amazon-ecs-agent/blob/4965b575c93639bb357ab6a2ef75b51676cac2fa/agent/api/eni/eni.go#L199

Thoughts about this one?

angelcar: I see, yeah, you are right. This is a non-issue.

agent/stats/task_windows.go
jterry75: This is a public API via Microsoft WMI. If they broke this it would affect more than just us. While it seems possible I think its safe to assume there would not be a breaking change here intentionally

angelcar (approved):

agent/stats/task_windows.go
jterry75: @rawahars - The more I think about it there is no real reason not to warn on this anyways. Since they are all `ParseUint` we could add a func like this:

```
func getMapValue(m map[string]string, key string) (uint, err) {
  v, ok := m[key]
  if !ok {
    return 0, errors.Errorf("failed to find key: %s in output", key)
  }
  val, err := strconv.ParseUint(v, 10, 64)
  if err != nil
    return 0, errors.Errorf("failed to parse network stats for %s with value: %s", key, v)
  }
  return val, nil
}
```

And then on the caller we use
```
recievedBrodcastPackets, err := getMapValue(m, "ReceivedBroadcastPackets")
if err != nil {
  return nil, err
}
// ...
```

And we handle both cases of both noticing that a key is not there which is unexpected as well as being more generic about the handling logic for all keys. What do you think?

rawahars: This seems like a good idea. I will go ahead and make the changes!

agent/stats/task_windows.go
```diff
@@ -0,0 +1,192 @@
...
+		kv := strings.Split(line, ":")
+		if len(kv) != 2 {
+			continue
+		}
+		key := strings.TrimSpace(kv[0])
+		value := strings.TrimSpace(kv[1])
+		m[key] = value
+	}
+
+	receivedBroadcastPackets, err := taskStat.getMapValue(m, "ReceivedBroadcastPackets")
```
angelcar, line 107: nit: sorry for not pointing this out earlier, but this code might be less redundant by putting the keys in an array and iterating over it. This is purely a matter of style and not blocker for this PR, so feel free to disregard.

https://play.golang.org/p/1_fArKNzVb6

angelcar (dismissed):

angelcar (approved):

jterry75 (approved):

rawahars:
Rebased the PR over latest dev. No other changes in the last force push.

angelcar (approved):

jterry75 (approved):

fierlion (approved):

PR #3088, opened 2021-11-13
Lower task cleanup duration

angelcar:
### Summary

Lower minimum task clean up duration from one minute to one second. 

This could help mitigate https://github.com/aws/amazon-ecs-agent/issues/2810 

### Testing
UTs, Integ & functinal  tests

* Lower task cleanup duration
### Licensing

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -92,7 +92,7 @@ const (
 
 	// minimumTaskCleanupWaitDuration specifies the minimum duration to wait before cleaning up
 	// a task's container. This is used to enforce sane values for the config.TaskCleanupWaitDuration field.
-	minimumTaskCleanupWaitDuration = 1 * time.Minute
+	minimumTaskCleanupWaitDuration = time.Second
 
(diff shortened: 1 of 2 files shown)


sparrc (approved):

agent/config/config.go
```diff
@@ -92,7 +92,7 @@ const (
 
 	// minimumTaskCleanupWaitDuration specifies the minimum duration to wait before cleaning up
 	// a task's container. This is used to enforce sane values for the config.TaskCleanupWaitDuration field.
-	minimumTaskCleanupWaitDuration = 1 * time.Minute
+	minimumTaskCleanupWaitDuration = time.Second
```
lydiafilipe, line 95: LGTM, curious though why exactly this is in place, and are there any possible issues from not waiting long enough?

lydiafilipe (approved):

agent/config/config.go
angelcar: I think it was an entirely arbitrary decision. I tested various scenarios and `1s` seems to work fine. 
At any rate, this is just the minimum allowed value and it will only be used if customer deems it useful.

PR #3089, opened 2021-11-15
delete container resource when delete task

Realmonia:
### Summary

Delete container resources when delete task

### Implementation details

loop through all containers and delete resources while delete task

### Testing

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -649,6 +649,19 @@ func (engine *DockerTaskEngine) deleteTask(task *apitask.Task) {
 		}
 	}
 
+	for _, container := range task.Containers {
+		for _, resource := range container.GetResources() {
+			err := resource.Cleanup()
+			if err != nil {
+				seelog.Warnf("Task engine [%s]/[%s]: unable to cleanup resource %s: %v",
+					task.Arn, container.Name, resource.GetName(), err)
+			} else {
+				seelog.Infof("Task engine [%s]/[%s]: resource %s cleanup complete",
+					task.Arn, container.Name, resource.GetName())
+			}
+		}
+	}
+
 	if execcmd.IsExecEnabledTask(task) {
 		// cleanup host exec agent log dirs
 		if tID, err := task.GetID(); err != nil {
--- a/agent/engine/docker_task_engine_linux_test.go
+++ b/agent/engine/docker_task_engine_linux_test.go
@@ -25,6 +25,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/asmsecret"
+
 	"github.com/aws/amazon-ecs-agent/agent/containerresource"
 	"github.com/aws/amazon-ecs-agent/agent/containerresource/containerstatus"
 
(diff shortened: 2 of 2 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -649,6 +649,19 @@ func (engine *DockerTaskEngine) deleteTask(task *apitask.Task) {
 		}
 	}
 
+	for _, container := range task.Containers {
```
mythri-garaga, line 652: Hmm should we clean up container resources before task resources here? or the order does not really matter here?

Realmonia: Don't think it would matter since container is not removed before we clean up the container resource and they are not dependent..

mythri-garaga (dismissed):

agent/engine/docker_task_engine.go
```diff
@@ -649,6 +649,19 @@ func (engine *DockerTaskEngine) deleteTask(task *apitask.Task) {
...
+		for _, resource := range container.GetResources() {
+			err := resource.Cleanup()
+			if err != nil {
```
yinyic, line 655: I wonder if there's anything we could do apart from logging the clean-up failure. I realized this is what we currently have for task clean-up as well, but IMO the error should rather be surfaced to customers so that at least they could potentially take action such as manual clean-up, otherwise there may be resource leak without customer visibility. 

This ask probably does not fall in scope of this particular PR, but just want to bring it up for some thoughts, as we are making relevant changes.

Realmonia: I took a look at current `deleteTask` and thought about this for a bit, Currently, `deleteTask` is called by `cleanUpTask` and then being monitored in `overseeTask`, I don't see any one in this call stack expects returns. I am wondering how would we want to emit the error apart from logging? I believe we don't want the actual task removal process to be aborted. Change log level to error makes some sense to me but from my understanding I think an error usually means we can't proceed with the current operation in an expected way (please let me know if I am wrong)

agent/engine/docker_task_engine_linux_test.go
```diff
@@ -155,16 +157,21 @@ func TestDeleteTask(t *testing.T) {
 
 	mockControl := mock_control.NewMockControl(ctrl)
 	cgroupResource := cgroup.NewCgroupResource("", mockControl, nil, "cgroupRoot", "", specs.LinuxResources{})
+	asmSecretResource := asmsecret.NewASMSecretResource("", nil, "executionId", true, nil, nil)
```
angelcar, line 160: We need to assert that`resource.Cleanup` method is being invoked for this resource, otherwise someone could remove that call altogether and nothing would prevent that to get to prod.

I think you could pass a secret via `SetCachedSecretValue` and then assert that `resource.GetCachedSecretValue` returns false after delete task. Or might be easier to make this resource a mock and assert that `Cleanup` is invoked.

Realmonia: Good point, Updated it. I think making this resource a mock will need more changes: make resource and resource mock to be implementing a same interface, so just use getCachedSecretValue to assure cleanup is called.

agent/engine/docker_task_engine_linux_test.go
```diff
@@ -198,6 +206,8 @@ func TestDeleteTask(t *testing.T) {
 	assert.Len(t, attachments, 1)
 
 	taskEngine.deleteTask(task)
+	_, resourceValueNotRemoved := asmSecretResource.GetCachedSecretValue("CachedValue")
```
angelcar, line 209: nit: This variable name is confusing. I think it should be something like `resourceValueExists`.
Otherwise the below line (`assert.False(t, resourceValueNotRemoved)`) reads like "assert that the resource value was not not removed"

lydiafilipe (approved):

PR #3092, opened 2021-11-16
Add container resource transitions 

mythri-garaga:
### Summary

This PR includes changes to update resource transitions method to include container resource transitions as part of task progression.

### Testing

make test

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -22,10 +22,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/aws/amazon-ecs-agent/agent/containerresource/containerstatus"
-
-	"github.com/aws/amazon-ecs-agent/agent/engine/execcmd"
-
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
@@ -33,10 +29,12 @@ import (
 	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
 	apierrors "github.com/aws/amazon-ecs-agent/agent/apierrors"
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/containerresource/containerstatus"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
 	"github.com/aws/amazon-ecs-agent/agent/ecscni"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dependencygraph"
+	"github.com/aws/amazon-ecs-agent/agent/engine/execcmd"
 	"github.com/aws/amazon-ecs-agent/agent/eventstream"
 	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/logger/field"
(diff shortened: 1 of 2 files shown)


agent/engine/task_manager.go
```diff
@@ -985,6 +983,9 @@ func (mtask *managedTask) progressTask() {
 	// these and if we exit early transitions can exit the goroutine and it'll
 	// get GC'd eventually
 	resources := mtask.GetResources()
+	for _, container := range mtask.Containers {
+		resources = append(resources, container.GetResources()...)
```
angelcar, line 987: it doesn't look like we need to create a new slice here. We just need the length.
Can we change this to:

```golang
resourceLen := len(mtask.GetResources())
for _, container := range mtask.Containers {
    resourceLen += len(container.GetResources())
}
transitionChange := make(chan struct{}, len(mtask.Containers)+resourceLen)
transitionChangeEntity := make(chan string, len(mtask.Containers)+resourceLen)
```

mythri-garaga: haha right 😄

agent/engine/task_manager.go
```diff
@@ -1141,7 +1142,23 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+	for _, cont := range mtask.Containers {
+		containerResources := cont.GetResources()
+		anyCanTransition, transitions = mtask.resourceTransitions(containerResources,
+			transitionFunc, "-"+cont.Name, anyCanTransition, transitions)
+	}
+
+	return anyCanTransition, transitions
+}
+
+func (mtask *managedTask) resourceTransitions(resources []taskresource.TaskResource, transitionFunc resourceTransitionFunc,
```
angelcar, line 1159: Why does this function need to return the transitions map?

mythri-garaga: This is to store all the transitions in the same transitions map, maybe using a recursive function is more readable??

angelcar: you can just pass the map created in 1144, it's the same map

agent/engine/task_manager.go
```diff
@@ -1141,7 +1142,23 @@ func (mtask *managedTask) startContainerTransitions(transitionFunc containerTran
...
+
+	resources := mtask.GetResources()
+	anyCanTransition, transitions = mtask.resourceTransitions(resources, transitionFunc, "",
+		anyCanTransition, transitions)
+
+	for _, cont := range mtask.Containers {
+		containerResources := cont.GetResources()
+		anyCanTransition, transitions = mtask.resourceTransitions(containerResources,
+			transitionFunc, "-"+cont.Name, anyCanTransition, transitions)
+	}
```
Realmonia, line 1154: Would the order of calls to resourceTransitions for task and container cause potential issues (feels unlikely but in case there is a container resource transition depends on task resource in certain state).

Is it possible that we merge the container resources and task resources together and call resourceTransitions so that for each loop of resources in `resourceTransitions` we have every resources touched?

agent/engine/task_manager.go
mythri-garaga: Updated, thanks!

agent/engine/task_manager.go
mythri-garaga: > Is it possible that we merge the container resources and task resources together and call resourceTransitions so that for each loop of resources in resourceTransitions we have every resource touched?

Either ways the order of resource will be the very similar I think! 

Hmm in what scenario does container resource transition depends on task resource??

Realmonia: I take a look again at this and find that transitions are actually happening after this returns, meaning that `anyCanTransition` will contain all resources of which dependencies are resolved, so I guess it should be good.

> In what scenario does container resource transition depends on task resource

I don't have a concrete case for now but since this. My original understanding is that actual status proceed will be happening in this function, in that case if taskResource A depends on containerResource b. and in this case only b can proceed, but since we call resourceTransitions on task first, it will give a response that none of the resources can proceed.

However, since this is only gathering possible transitions and won't return after `resourceTransitions` for task resources, I think this comment is safe to ignore.

Realmonia (approved):

yinyic (approved):

PR #3095, opened 2021-11-22
inject host values to app containers that use firelens v2

mythri-garaga:
### Summary

This PR includes changes to inject hostname values to all the app containers that use firelens v2. Hostname values are required for observability use cases, logs, metrics, or traces are often sent over the network from an SDK(like OTEL, X-Ray, Log4j) in the app container, to the collector container.

Firelens v1 behaviour where we inject `FLUENT_HOST` and `FLUENT_PORT` environment variables is unchanged.

### Testing

`make test`

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -103,10 +103,11 @@ const (
 	fluentTagDockerFirelensV2Format = "%s.%s"
 
 	// Environment variables are needed for firelens
-	fluentNetworkHost      = "FLUENT_HOST"
-	fluentNetworkPort      = "FLUENT_PORT"
-	FluentNetworkPortValue = "24224"
-	FluentAWSVPCHostValue  = "127.0.0.1"
+	fluentNetworkHost        = "FLUENT_HOST"
+	fluentNetworkPort        = "FLUENT_PORT"
+	FluentNetworkPortValue   = "24224"
+	AWSVPCHostValue          = "127.0.0.1"
+	opentelemetryNetworkHost = "OPEN_TELEMETRY_HOST"
 
 	defaultMonitorExecAgentsInterval = 15 * time.Minute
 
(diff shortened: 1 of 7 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -103,10 +103,11 @@ const (
...
 
 	// Environment variables are needed for firelens
-	fluentNetworkHost      = "FLUENT_HOST"
-	fluentNetworkPort      = "FLUENT_PORT"
-	FluentNetworkPortValue = "24224"
-	FluentAWSVPCHostValue  = "127.0.0.1"
+	fluentNetworkHost        = "FLUENT_HOST"
+	fluentNetworkPort        = "FLUENT_PORT"
+	FluentNetworkPortValue   = "24224"
+	AWSVPCHostValue          = "127.0.0.1"
```
sharanyad, line 109: Are these values expected to change over time? 
Why not inject these values in ECS backend instead of agent doing it?

Realmonia (approved):
lgtm

agent/engine/docker_task_engine.go
PettitWesley: So for AWSVPC and Host mode, the values are always the same. Its bridge mode where Agent code is needed, since it needs to inject the bridge IP. 

Therefore, I think its simpler to just have the agent manage all the values.

PettitWesley: If you disagree with that, it's totally doable for us to inject the static values for host and awsvpc on the backend. And then agent only injects the value for bridge. I think separate which component injects based on the network mode is weird tho.

sharanyad: I see, I'm ok with keeping this in agent for now.

agent/engine/docker_task_engine.go
```diff
@@ -1329,6 +1327,34 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
 	return logConfig
 }
 
+// getFirelensEnvironmentVariables returns env variables for app containers that use awsfirelens
+// for task using firelens v1, inject FLUENT_HOST to container
+// for task using firelens v2 FluentBit collector, inject FLUENT_HOST to container if not already configured by customer
+// for task using firelens v2 OpenTelemetry collector, inject OPEN_TELEMETRY_HOST to container if not already configured by customer
+func getFirelensEnvironmentVariables(task *apitask.Task, container *apicontainer.Container,
```
sharanyad, line 1334: could we add unit tests for this method?

mythri-garaga: Added, thanks!!

Realmonia (dismissed):

PettitWesley (approved):

agent/engine/docker_task_engine.go
```diff
@@ -1129,28 +1130,25 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
-					err := apierrors.DockerClientConfigError{Msg: "unable to get BridgeIP for task in bridge  mode"}
-					return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(&err)}
-				}
-				container.MergeEnvironmentVariables(map[string]string{
-					fluentNetworkHost: ipAddress,
-					fluentNetworkPort: FluentNetworkPortValue,
-				})
+		envVars := make(map[string]string)
+		if task.IsNetworkModeAWSVPC() {
+			getFirelensEnvironmentVariables(task, container, firelensVersion, envVars, AWSVPCHostValue)
```
angelcar, line 1138: nit: This method is setting, not getting, maybe let's change the name to`setFirelensEnvironmentVariables`?

agent/engine/docker_task_engine.go
```diff
@@ -1329,6 +1327,36 @@ func getFirelensLogConfig(task *apitask.Task, container *apicontainer.Container,
 	return logConfig
 }
 
+// getFirelensEnvironmentVariables returns env variables for app containers that use awsfirelens
```
angelcar, line 1330: nit: this method doesn't return anything :)

agent/engine/docker_task_engine.go
mythri-garaga: Agreed and updated, Thanks!

angelcar (approved):

Realmonia (approved):

PR #3104, opened 2021-12-10
Add git secrets workflow

Realmonia:
### Summary

Add git secret scan to workflow and add one dummy AccountID into gitallowed.

[Git secret](https://github.com/awslabs/git-secrets) is a tool scanning commits/commit messages for possible leaks of secret into github.

Since git secret is usually installed in dev environment and checking commits when run `git commit`, the way this workflow works is 1) install git secret on worker 2) scan existing code base (instead of individual commits) based on previous clean record to detect problem in the new code changes

### Testing

Tested failed workflow: https://github.com/Realmonia/amazon-ecs-agent/actions/runs/1561662434
Added one dummy AccountID into `.gitallowed` then workflow passed: https://github.com/Realmonia/amazon-ecs-agent/actions/runs/1561670038

Add git secret scan to workflow

### Licensing

yes


.github/workflows/gitsecrets.yml
```diff
@@ -0,0 +1,21 @@
+name: GitSecretsScan
+
+on: [push, pull_request]
+
+jobs:
+  git-secret-check:
+    name: Git Secrets Scan
+    runs-on: macos-latest
```
sparrc, line 8: is there a git secrets check we can run on ubuntu? my understanding is macos instances are more costly and less available.

Realmonia: To install git secret linux instance requires sudo access --- I am not sure if that's available in these workflow workers, I can try that

Realmonia: Just did a test for that, the workflow is able to use sudo. I will update to use ubuntu.

sparrc (approved):

singholt (approved):

chienhanlin (commented):
Q. Is this dummy file required for the workflow? Thanks.

singholt:
> Q. Is this dummy file required for the workflow? Thanks.

@chienhanlin I think its needed because [this test](https://github.com/aws/amazon-ecs-agent/blob/master/agent/ec2/ec2_metadata_client_test.go#L67) uses the dummy account ID and the scan flags it if its not added in the .gitallowed file.

PR #3108, opened 2021-12-14
Pull gcc build image from public ECR, bump version

sparrc:
- Change dockerhub gcc image to public ecr
- gcc version 7.3.0 is not available on public ecr, so just bumped
  version to the latest version available
- Also removed the ci-ecr-pull/push scripts as these are no longer
  necessary, since we have no more dockerhub images in our build or test
  workflows.

### Testing

functional tests

enhancement - bumped pause container gcc build version

### Licensing


scripts/ci-ecr-pull
```diff
@@ -15,49 +15,4 @@
 # limitations under the License.
 set -e
```
singholt, line 16: do we need this script at all?

sparrc: we don't, but we currently reference it in internal scripts that will take some time to deploy away, so safest to just leave it here as a dummy for now

sharanyad (approved):

singholt (approved):

PR #3127, opened 2022-02-11
Support Unified Cgroups (cgroups v2)

sparrc:
### Summary

Implements part of https://github.com/aws/containers-roadmap/issues/1535

This adds support for task-level resource limits when running on unified cgroups (aka cgroups v2) with the systemd cgroup driver.

Cgroups v2 has introduced a cgroups format that is not backward compatible with cgroups v1. In order to support both v1 and v2, we have added a config variable to detect which cgroup version the ecs agent is running with. The containerd/cgroups library is used to determine which mode it is using on agent startup.

When running with cgroups v2 docker containers are normally created under the path `/sys/fs/cgroup/system.slice/docker-${CONTAINER_ID}.scope` path.

When using task resource limits, the task receives it's own cgroup slice, and the containers are created under the path `/sys/fs/cgroup/ecstasks.slice/ecstasks-${TASK_ID}.slice/docker-${CONTAINER_ID}.scope`. The `ecstasks-${TASK_ID}.slice` slice is created with the resource limits specified in the task definition, and every docker container under it is subject to these limits.

For a good doc and overview of how these cgroup hierarchies now work with the systemd driver, see https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-default_cgroup_hierarchies

Because docker uses the systemd cgroup driver by default with cgroup v2, resource limit actions are no longer logged by the agent or docker. OOM-kill messages can now be found in the system journal (`journalctl --system`) or `/var/log/messages`.

### Testing

Functional tests run to verify that v1 behavior is unchanged.

Manually tested running on a cgroup v2 system and verified that task-level mem/cpu limits are applied.

Ran the full functional testing suite on a cgroupv2 system (AL2022) and verified all tests pass.

Enhancement - Support for unified cgroups and the systemd cgroup driver.

### Licensing


agent/config/config_linux.go
```diff
@@ -13,24 +13,12 @@
...
-
-// Init is used to setup the cgroup root for ecs
-func (c *control) Init() error {
-	seelog.Infof("Creating root ecs cgroup: %s", config.DefaultTaskCgroupPrefix)
-
-	// Build cgroup spec
-	cgroupSpec := &Spec{
-		Root:  config.DefaultTaskCgroupPrefix,
-		Specs: &specs.LinuxResources{},
+func init() {
```
chienhanlin, line 20: Q. (I might miss it) where/when do we call this function to update `CgroupV2` flag? Thanks!

sparrc: in go the init() function is a special function that executes before any other functions in a package are executed, it's generally used for initializing things: https://golangdocs.com/init-function-in-golang

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
```diff
@@ -0,0 +1,79 @@
...
+func (c *controlv2) Create(cgroupSpec *Spec) error {
+	// Validate incoming spec
+	err := validateCgroupSpec(cgroupSpec)
+	if err != nil {
+		return fmt.Errorf("cgroup create: failed to validate spec: %w", err)
+	}
+
+	cgroupPath := cgroupSpec.Root
+	seelog.Infof("Creating cgroup cgroupv2root=%s parentSlice=%s cgroupPath=%s", defaultCgroupv2Path, parentCgroupSlice, cgroupPath)
+	_, err = cgroupsv2.NewSystemd(parentCgroupSlice, cgroupPath, -1, cgroupsv2.ToResources(cgroupSpec.Specs))
```
angelcar, line 45: nit: for readability, can we put that `-1` in some constant that describes that it's for?

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return m.Delete()
+}
+
+// Exists is used to verify the existence of a cgroup
+func (c *controlv2) Exists(cgroupPath string) bool {
+	fullCgroupPath := filepath.Join(defaultCgroupv2Path, parentCgroupSlice, cgroupPath)
+	seelog.Infof("Checking existence of cgroup cgroupv2root=%s parentSlice=%s cgroupPath=%s fullPath=%s", defaultCgroupv2Path, parentCgroupSlice, cgroupPath, fullCgroupPath)
+
+	if _, err := os.Stat(fullCgroupPath); os.IsNotExist(err) {
```
angelcar, line 69: we should still log the error here, there's a chance that the file exists but `os.Stat` will return error anyways

agent/config/config_linux.go
chienhanlin: Thanks!

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
```diff
@@ -0,0 +1,84 @@
...
+		return false
+	}
+	return true
+}
+
+// Init is used to setup the cgroup root for ecs
+func (c *controlv2) Init() error {
+	// In cgroups v2, there is no root "ecs" cgroup, so this function is a no-op.
```
yinyic, line 82: Parent cgroup aside, I think we will want to verify that the cpu and memory controllers are available under v2 (see discussion in this [section](https://quip-amazon.com/cJfYAHQJ5ZBM/Cgroupv2-Considerations-for-ECS-Agent#temp:C:OQQ24f5661d2d6beb0a39897f234)). If these two controllers are not available, docker won't be able to create container cgroups either. We currently don't have container resource limit as capability, so we won't be able to filter out the incapable instances during task placement. I personally think we should have that capability, but before that, we may want to terminate Agent since we can't respect task or container resource limits from task def in that case.

agent/api/task/task_linux.go
```diff
@@ -76,14 +76,26 @@ func (task *Task) initializeCgroupResourceSpec(cgroupPath string, cGroupCPUPerio
 }
 
 // BuildCgroupRoot helps build the task cgroup prefix
-// Example: /ecs/task-id
+// Example v1: /ecs/task-id
+// Example v2: ECSTask$TASKID.slice
```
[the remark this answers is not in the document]
sparrc: Discussed offline, will change this to `ECSTasks-$TASKID.slice`. Using systemd parent-name.slice notation, this will make a parent slice called ECSTasks.slice for all ecs tasks with mem/cpu limits.

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
sparrc: Good point, I've added a check for the available cgroup controllers in the root /sys/fs/cgroup dir in the Init function now

agent/taskresource/cgroup/control/cgroup_controller_linux_test.go
```diff
@@ -44,8 +44,7 @@ func TestCreateHappyCase(t *testing.T) {
 
 	control := newControl(mockCgroupFactory)
 
-	res, err := control.Create(&Spec{testCgroupRoot, testSpecs})
-	assert.Equal(t, mockCgroup, res)
+	err := control.Create(&Spec{testCgroupRoot, testSpecs})
```
angelcar, line 47: how are we validating that `Create` is doing what it's supposed to now?

sparrc: `Create` used to return a `cgroups.Cgroup` object that was never actually used in our codebase. I removed it because cgroups v2 doesn't have this object anymore, and since we didn't use it I thought it better to remove than to create an entirely new interface just for cgroupsv2. 

See cgroups.New (v1): https://pkg.go.dev/github.com/containerd/cgroups#New
versus cgroupsv2.NewSystemd (v2): https://pkg.go.dev/github.com/containerd/cgroups@v1.0.3/v2#NewSystemd

angelcar: I see, So if I commented out all the code inside `Create` and just did `return nil`, would any UT break?

agent/taskresource/cgroup/cgroup.go
```diff
@@ -276,16 +278,19 @@ func (cgroup *CgroupResource) setupTaskCgroup() error {
...
+		return fmt.Errorf("cgroup resource: setup cgroup: unable to create cgroup taskARN=%s cgroupPath=%s cgroupV2=%v err=%s", cgroup.taskARN, cgroupRoot, config.CgroupV2, err)
 	}
 
-	// enabling cgroup memory hierarchy by doing 'echo 1 > memory.use_hierarchy'
-	memoryHierarchyPath := filepath.Join(cgroup.cgroupMountPath, memorySubsystem, cgroupRoot, memoryUseHierarchy)
-	err = cgroup.ioutil.WriteFile(memoryHierarchyPath, enableMemoryHierarchy, rootReadOnlyPermissions)
-	if err != nil {
-		return fmt.Errorf("cgroup resource [%s]: setup cgroup: unable to set use hierarchy flag: %w", cgroup.taskARN, err)
+	if !config.CgroupV2 {
+		// enabling cgroup memory hierarchy by doing 'echo 1 > memory.use_hierarchy'
```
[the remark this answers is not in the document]
sparrc: My understanding was that since cgroups v2 is by default a "unified hierarchy", this flag no longer applies. From my testing this seems to be the case because limits set on a task slice with two containers are applied to both containers.

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
```diff
@@ -0,0 +1,119 @@
...
+func (c *controlv2) Exists(cgroupPath string) bool {
+	fullCgroupPath := fullCgroupPath(cgroupPath)
+	seelog.Infof("Checking existence of cgroup cgroupv2root=%s parentSlice=%s cgroupPath=%s fullPath=%s", defaultCgroupv2Path, parentCgroupSlice, cgroupPath, fullCgroupPath)
+
+	_, err := os.Stat(fullCgroupPath)
+	if os.IsNotExist(err) {
+		return false
+	}
+	if err != nil {
+		seelog.Errorf("error checking if cgroup exists err=%s", err)
```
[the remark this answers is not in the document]
sparrc: OK, I also was not sure what to return here, I can change to false if it's idempotent.

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
```diff
@@ -0,0 +1,119 @@
...
+		return fmt.Errorf("cgroupv2 create: failed to validate spec: %w", err)
+	}
+
+	cgroupPath := cgroupSpec.Root
+	seelog.Infof("Creating cgroup cgroupv2root=%s parentSlice=%s cgroupPath=%s", defaultCgroupv2Path, parentCgroupSlice, cgroupPath)
+	_, err = cgroupsv2.NewSystemd(parentCgroupSlice, cgroupPath, generalSlicePID, cgroupsv2.ToResources(cgroupSpec.Specs))
+	if err != nil {
+		return fmt.Errorf("cgroupv2 create: unable to create v2 manager cgroupPath=%s err=%s", cgroupPath, err)
+	}
+
```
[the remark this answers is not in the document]
sparrc: OK, I will look into that, I've been testing mostly with memory but I will test with task cpu limits too to see

agent/taskresource/cgroup/control/cgroup_controller_linux_test.go
sparrc: `mockCgroupFactory.EXPECT().New` would fail

agent/config/config_linux.go
```diff
@@ -13,24 +13,12 @@
...
+	if cgroups.Mode() == cgroups.Unified {
+		CgroupV2 = true
```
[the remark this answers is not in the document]
sparrc: I can try, but I dont think we can, since it's in the config package and an init() function, the logger has maybe not been initialized yet.

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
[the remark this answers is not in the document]
sparrc: Toggling and checking cgroup controllers on initialization has now been added.

agent/stats/utils_unix.go
```diff
@@ -39,9 +39,8 @@ func dockerStatsToContainerStats(dockerStats *types.StatsJSON) (*ContainerStats,
 }
 
 func validateDockerStats(dockerStats *types.StatsJSON) error {
-	// The length of PercpuUsage represents the number of cores in an instance.
-	if len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0 || numCores == uint64(0) {
```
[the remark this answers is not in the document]
angelcar: So, we think that `len(dockerStats.CPUStats.CPUUsage.PercpuUsage) == 0` was redundant or a mistake? I'm hesitant to remove something we don't seem to have clarity on why it was introduced.

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		return fmt.Errorf("cgroupv2 create: failed to validate spec: %w", err)
+	}
+
+	cgroupPath := cgroupSpec.Root
+	seelog.Infof("Creating cgroup cgroupv2root=%s parentSlice=%s cgroupPath=%s", defaultCgroupv2Path, parentCgroupSlice, cgroupPath)
+
+	m, err := cgroupsv2.NewSystemd(parentCgroupSlice, cgroupPath, generalSlicePID, cgroupsv2.ToResources(cgroupSpec.Specs))
+	if err != nil {
+		return fmt.Errorf("cgroupv2 create: unable to create v2 manager cgroupPath=%s err=%s", cgroupPath, err)
```
angelcar, line 54: nit: should we leave error formatting to the part that logs the error? (i.e. the consumer of `Create`)

agent/stats/utils_unix.go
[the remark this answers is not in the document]
sparrc: PercpuUsage doesn't appear to be used anywhere in agent, and it's not available anymore with cgroupv2. I'm not sure how else we can check that it's safe to remove? FWIW there isn't really any alternative to just removing the check from what I can tell, since cgroupv2 stats simply dont provide this level of granularity.

If you would prefer we could just have an if-statement here and continue checking this array when we're using cgroupv1?

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
sparrc: like a naked `return err` or do you mean just with less formatting than is here?

agent/stats/utils_unix.go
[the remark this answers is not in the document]
yinyic: I'm guessing the intention was to do a simple sanity check of the stats API.
There is an alternative to the `PercpuUsage` field, `online_cpus`, according to docker API [doc](https://docs.docker.com/engine/api/v1.41/#operation/ContainerStats), which should be available when either v1 or v2 cgroups is used:
```
number_cpus = lenght(cpu_stats.cpu_usage.percpu_usage) or cpu_stats.online_cpus
```
But then I also saw this comment
```
If either precpu_stats.online_cpus or cpu_stats.online_cpus is nil then for compatibility with older daemons the length of the corresponding cpu_usage.percpu_usage array should be used.
```
Earliest docker API reference I could find is for [1.18](https://docs.docker.com/engine/api/v1.18/) and it does not contain the `online_cpu` field. 

Since ECSAgent-supported docker api version can go back to [1.17](https://github.com/aws/amazon-ecs-agent/blob/225bc3a556bd2d1759ab27b23f54e7e68086c9f0/agent/dockerclient/versions.go#L19), which can still work with 20.10 docker engine (based on this [doc](https://docs.docker.com/engine/api/)) which is when docker introduced cgroupv2, it's not 100% safe for us to check `online_cpus` when cgroupv2 is used. 

We can still keep the v1 check, but I'm not sure of a safe way to validate docker stats numCPU when using cgroupv2.

agent/taskresource/cgroup/cgroup.go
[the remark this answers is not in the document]
yinyic: I think applying this flag in v1 will charge subcgroup memory usage to its parent (bottom-up). The allocation (top-down) does not get affected without or without the flag. 
This is my understanding: if a task cgroup has two containers, task has hard limit 100MB with each container getting 50MB. Now one of the two containers attempts to allocate 200MB
- without the flag, only the container will be killed
- with the flag, because the memory allocation got charged to the task, the task will also be oom-killed. 

Can we verify if that's the behavior? 

And tbh, I'm not entirely sure why we want to enable this hierarchical charging. If a non-essential container gets killed, we should still let the task run...

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
angelcar: I was thinking we can just wrap the error using go's way, like so: `return fmt.Errorf("unable to create v2 manager: %w", err)`

Then it's up to the consumer what to log and in what format, since it has all the info required. For example:

```go
err := Create(cgroupSpec)
if err != ni {
   logger.Info("Error creating cgroup", logger.Fields {
        "root": cgroupSpec.Root,
         "error": err,
}) 
} 
```

One difference would be, for example, that if the logger is configured in json format, then the format for the fields is also json.

agent/taskresource/cgroup/cgroup.go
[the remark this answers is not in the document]
sparrc: > if a task cgroup has two containers, task has hard limit 100MB with each container getting 50MB

I can test this but I'm fairly certain that it works that both containers have a shared limit of 100MB. So if one container is using only 20MB the other can use up to 80MB.

agent/taskresource/cgroup/control/cgroupv2_controller_linux.go
sparrc: OK, this has been updated

agent/taskresource/cgroup/cgroup.go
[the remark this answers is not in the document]
sparrc: I tested the scenario of a task with 512MB of total memory, with one container allocating 100MB and another 450MB. This does seem to be correct, ie...

> without the flag, only the container will be killed

On AL2022 only the container allocating 450MB is being OOM-killed. I'm not sure how exactly does it decide which container to kill, it might just be that it kills whichever container breaches the memory limit first?

> with the flag, because the memory allocation got charged to the task, the task will also be oom-killed.

Yes I confirmed that on AL2, when one container hits the limit, BOTH containers are killed by the OOM-killer.

Will investigate how we can turn on use_hierarchy on cgroupv2...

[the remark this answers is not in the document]
yinyic: > Will investigate how we can turn on use_hierarchy on cgroupv2...

I would actually suggest that we turn of `use_hierarchy` on cgroupv1. 

If I understanding it correctly, the only time that one container failure should bring down the whole task is when that container is the essential container. On the other hand, if a container is non-essential, it being OOM killed (or terminated in any other way for that matter) should not affect the other containers in the task.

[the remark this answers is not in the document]
sparrc: To follow up on this, `use_hierarchy` functionality is not completely available on cgroup v2. The `memory.oom.group` flag is close (see https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html), but it causes the entire container to be killed, rather than just the process within the container. We will go forward with the kernel default behavior of only killing the "bulkiest" process when the task-level memory limit is reached.

yinyic (approved):

angelcar (approved):

PR #3146, opened 2022-03-10
rename init packages and onboard init to go modules

lydiafilipe:
### Summary
This onboards ecs-init to go modules. There are quite a few changes here, to summarize them

- replace ecs init path references in scripts/init package
- created the go module and go sum files
- vendor directory updated
- dep-related files and make target removed
- tweaked the gobuild script to remove setting gopath

### Implementation details
Replaced the ecs init package paths by running `grep -rl github.com/aws/amazon-ecs-init . | xargs sed -i 's/github.com\/aws\/amazon-ecs-init/github.com\/aws\/amazon-ecs-agent/g'` 

Onboarded to go modules with
`go mod init`
`go mod tidy`

Had to add a replacement: `replace github.com/coreos/go-systemd => github.com/coreos/go-systemd/v22 v22.0.0`. This is due to a dependency on github.com/docker/go-plugins-helpers, which has not onboarded to go modules or semantic versioning for some of its imports.

After that, ran `go mod tidy` again. Had to replace some packages for which the versioned changed compared to what was in our dep files.

Then deleted the vendor directory and ran `go mod vendor`.

Additionally, had to remove setting GOPATH in our init build script (since that cases conflicts when using go modules). Also removed dep stuff in Makefile.

### Testing
- made/tested the change in the init package itself
- successfully ran `make rpm-in-docker` and `make generic-rpm` in that package, and used the former rpm in an ami build (against which I ran functional tests)
- built changes in Koji and, similarly to above, used the result rpms in ami builds and ran functional tests against those

I replicated those changes in this repository.

Also verified the github workflow for the same changes in the init repository [succeeds](https://github.com/aws/amazon-ecs-init/runs/5501769756?check_suite_focus=true)

rename init packages and onboard init to go modules

### Licensing


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: go
-go_import_path: github.com/aws/amazon-ecs-init
+go_import_path: github.com/aws/amazon-ecs-agent
```
fierlion, line 2: travis is gone so we should remove this file.

fierlion (approved):
Looks good!

chienhanlin (commented):
Q. See some golang.org files are empty, like ecs-init/vendor/golang.org/x/sys/unix/mksyscall.pl, are they supposed to be empty? Thanks!

lydiafilipe:
> Q. See some golang.org files are empty, like ecs-init/vendor/golang.org/x/sys/unix/mksyscall.pl, are they supposed to be empty? Thanks!

Interesting that they are empty in the diff, they are not actually empty locally or in my branch - (eg https://github.com/lydiafilipe/amazon-ecs-agent/blob/init-gomodules/ecs-init/vendor/golang.org/x/sys/unix/mksyscall.pl(

chienhanlin (approved):

PR #3149, opened 2022-03-11
Dockerfree build and ecs-init make integration

fierlion:
### Summary
This integrates ecs-init specific Make targets and also adds docker-free targets to build the pause container and agent container without Docker.  This will allow us to build ECS RPMs in multiple constrained/secure build environments (like Koji).

### Implementation details
Each of the more complex make targets calls an external bash script in `/scripts`.  Otherwise, most of the init-specific targets are simple enough to be inlined.

All integration changes are additive so they will not mutate or interrupt our existing processes.
The changes were built and tested on an EC2 instance based on the latest ECS-Optimized AMI.

The changes were tested with go version 1.17.2.  There's still more work to make the changes compatible with older versions of golang.

### Testing
All make targets have been tested manually.  The init specific targets have been integrated into the github workflow testing
New tests cover the changes: yes -- github init-specific workflows added.

Add docker-free make targets and integrate generic rpm build.
### Licensing


Makefile
```diff
@@ -223,6 +236,21 @@ cni-plugins: get-cni-sources .out-stamp build-ecs-cni-plugins build-vpc-cni-plug
 	mv $(PWD)/out/amazon-vpc-cni-plugins/* $(PWD)/out/cni-plugins
 	@echo "Built all cni plugins successfully."
 
+# dockerfree build process will build the agent container image from scratch
+# and with minimal dependencies
+# requires glibc-static
+
+dockerfree-pause:
+	GOOS=linux GOARCH=amd64 ./scripts/build-pause
```
lydiafilipe, line 244: can you clarify why you are setting GOOOS/GOARCH here? What if this is being built on arm?

agent-container/agent-config.json
```diff
@@ -0,0 +1,2 @@
+{"author":"Amazon Web Services, Inc.","config":{"Cmd":["/agent"],"ArgsEscaped":true},"created":"2014-12-12T01:12:53.332832423Z","history":[{"created":"2014-12-12T01:12:53.332832423Z","author":"Amazon Web Services, Inc.","created_by":"[] + [] === \"\"","empty_layer":true}],"os":"linux","rootfs":{"type":"layers","diff_ids":["sha256:~~digest~~"]}}
```
sparrc, line 1: why are these static files? shouldn't they be generated at build-time?

agent-container/agent-manifest.json
```diff
@@ -0,0 +1,2 @@
+[{"Config":"config.json","RepoTags":["amazon/amazon-ecs-agent:1.60.0"],"Layers":["rootfs/layer.tar"]}]
```
sparrc, line 1: this one should probably be generated at build-time too?

scripts/get-host-certs
```diff
@@ -0,0 +1,41 @@
...
+# License is located at
+#
+#	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# This script builds a copy of the agent.
```
sparrc, line 15: this comment needs updating :)

ecs-init/version/version.go
```diff
@@ -18,10 +18,10 @@
 package version
 
 // Version is the version of the ecs-init
-var Version string = "unknown"
+var Version string = "1.60.0"
```
lydiafilipe, line 21: I don't think we want these changes committed

scripts/get-host-certs
```diff
@@ -0,0 +1,41 @@
...
+    break
+  fi
+done
+
+# if we found a cert file, we'll copy else return 0
+if [ ! -z "$thisCertFile" ]
+then
+  cp $thisCertFile .
+else
+  exit 1
```
lydiafilipe, line 40: maybe echo some sort of error message to help debug in case this happens

scripts/build-cni-plugins
```diff
@@ -0,0 +1,32 @@
...
+# License is located at
+#
+#	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is distributed
+# on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+
+# This script builds a copy of the agent.
```
lydiafilipe, line 15: minor: cni-plugins, not agent. Same with comments in other script files

scripts/build-agent-image
```diff
@@ -0,0 +1,54 @@
...
+# add agent
+cp ./out/amazon-ecs-agent rootfs/agent
+
+# build container
+mkdir -p image/rootfs
+tar --mtime="@1492525740" --owner=0 --group=0 --numeric-owner -cf image/rootfs/layer.tar -C rootfs .
+DIGEST=$(sha256sum image/rootfs/layer.tar | sed -e 's/ .*//')
+install -m 0644 ./agent-container/agent-image-VERSION image/rootfs/VERSION
+install -m 0644 ./agent-container/agent-config.json image/config.json
+sed -i "s/~~digest~~/${DIGEST}/" image/config.json
```
sparrc, line 50: is it possible for us to use jq inside of the koji build? if it is we could maybe turn these into jq templates? That way the json is always guaranteed to be valid. It looks something like this:

template.jq
```

{
    "descriptionContentType": "text/plain",
    "title": $title,
    "description": "$description"
}
```
then pass the parameters via jq like:
```
jq -n --arg description "My Description" --arg title "My Title" -f template.jq
```

agent-container/agent-config.json
fierlion: The diff_ids field is being generated at build-time.  Otherwise I agree maybe we can update the history.created with an actual timestamp.
These are generated by Docker otherwise -- if we don't have Docker, we need to create/provide these config files manually.  

Bottlerocket uses static configs to build its pause container: https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/ecs-agent/pause-config.json

agent-container/agent-manifest.json
fierlion: I'll update the repoTags to use the VERSION.
These are generated by Docker otherwise -- if we don't have Docker, we need to create/provide these config files manually.  

Bottlerocket uses static configs to build its pause container: https://github.com/bottlerocket-os/bottlerocket/blob/develop/packages/ecs-agent/pause-manifest.json

scripts/build-agent-image
```diff
@@ -0,0 +1,54 @@
...
+mkdir -p rootfs/amazon-ecs-cni-plugins/
+cp ./misc/plugins/aws-appmesh rootfs/amazon-ecs-cni-plugins/aws-appmesh
+cp ./misc/plugins/ecs-bridge rootfs/amazon-ecs-cni-plugins/ecs-bridge
+cp ./misc/plugins/ecs-eni rootfs/amazon-ecs-cni-plugins/ecs-eni
+cp ./misc/plugins/ecs-ipam rootfs/amazon-ecs-cni-plugins/ecs-ipam
+cp ./misc/plugins/vpc-branch-eni rootfs/amazon-ecs-cni-plugins/vpc-branch-eni
+
+# add certs
+mkdir -p rootfs/etc/ssl/certs/
+cp ./misc/certs/ca-bundle.crt rootfs/etc/ssl/certs/ca-certificates.crt
```
lydiafilipe, line 35: It looks to me like `get-host-certs` copies the file here, but the name of that file could vary and isn't necessarily ca-bundle.crt. Maybe rename it in `get-host-certs` or look  also for things like cert.pem etc

agent-container/agent-config.json
lydiafilipe: So do we have to ensure these files get rebuilt/committed each time then from current agent changes? Might need some sort of check to reinforce that

lydiafilipe:
can you update the title? Now it's in review :D

Makefile
fierlion: there's still more work to build arm -- I'll update the description to say this is focused on amd linux.

packaging/generic-rpm/amazon-ecs-init.spec
```diff
@@ -42,7 +42,6 @@ Source4:        https://s3.amazonaws.com/amazon-ecs-agent/ecs-agent-arm64-v%{bun
 Source5:        amazon-ecs-volume-plugin.service
 Source6:        amazon-ecs-volume-plugin.socket
 
-BuildRequires:  golang >= 1.7
```
Realmonia, line 45: Is this changed because the build is done with 1.17.2?

agent-container/agent-config.json
fierlion: Updated to string replace and use the date util to match expected formatting: `sed -i "s/~~timestamp~~/$(date +"%FT%T.%NZ")/g" image/config.json`
Otherwise we'll use https://github.com/aws/amazon-ecs-agent/blob/master/VERSION to fill in our version.

scripts/build-agent-image
fierlion: I don't want to add any dependencies unless it's absolutely necessary.  I updated to use `sed` to replace the tokens.

.github/workflows/static.yml
```diff
@@ -29,6 +29,32 @@ jobs:
           make get-deps
           make static-check
 
+  init-check:
```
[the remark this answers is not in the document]
fierlion: From the description:

> All integration changes are additive so they will not mutate or interrupt our existing processes.

So the plan for now is to be sure that all the changes won't have any effect on the current build process -- this is why all the init-specific targets exist rather than adding them directly to the existing targets.

packaging/generic-rpm/amazon-ecs-init.spec
fierlion: Yes.  This will be updated as we start testing our builds in Koji.

Realmonia (approved):

PR #3154, opened 2022-03-22
Updating import paths, file names, and Make targets for init

lydiafilipe:
### Summary
In these changes, the init file paths are updated, the Makefile is updated with the init make targets, and some of the scripts and targets for init are renamed. The static analysis workflows for agent is added back, and one is added for init as well.

### Implementation details
- Import paths for the ecs-init files were updates to use the agent paths
- Copied over Makefile targets from the init Makefile, and renamed them to indicate they are for init
- Made some script and Makefile changes to get make targets to work
- added static check for init, and added back the failing agent one and got it to work

### Testing
Developed on AL1 machine and ran the make targets. Only ones not tested are the srpm and ubuntu-trusty ones, which were just copied over.
One thing worth noting is that test-in-docker target seems to have issues, though this is not working for the init code independently of these changes. I tweaked the script a bit to take care of some of the errors, though some of the tests seem to fail in it. 

Updating import paths, file names, and Make targets for integrating init.

### Licensing


ecs-init/docker/dependencies.go
```diff
@@ -22,8 +22,8 @@ import (
 	"net/url"
 	"time"
 
-	"github.com/aws/amazon-ecs-init/ecs-init/backoff"
-	"github.com/aws/amazon-ecs-init/ecs-init/config"
+	"github.com/aws/amazon-ecs-agent/ecs-init/backoff"
```
angelcar, line 25: I think this is now a duplicated package since agent also has a backoff package. Should we use that instead?

lydiafilipe: Good point! In these merged branches there is an opportunity to pull out shared functionality. In this case now though I don't think it makes sense, primarily since these are the legacy branches so I'm trying to avoid changing the code itself. Also, we're currently building init and agent separately and trying to get that to work, would need to do some more considerable refactoring I think to build shared code (though haven't looked into this too thoroughly yet)

angelcar (approved):

Makefile
```diff
@@ -290,12 +329,19 @@ GOPATH=$(shell go env GOPATH)
 	go get github.com/golang/mock/mockgen
 	cd "${GOPATH}/src/github.com/golang/mock/mockgen" && git checkout 1.3.1 && go get ./... && go install ./... && cd -
 	go get golang.org/x/tools/cmd/goimports
-	go get github.com/fzipp/gocyclo/cmd/gocyclo
+	GO111MODULE=on go get github.com/fzipp/gocyclo/cmd/gocyclo@v0.3.1
```
fierlion, line 332: curious where the v0.3.1 comes from -- is this the go modules enabled version?

README.md
```diff
@@ -242,7 +242,6 @@ The following targets are available. Each may be run with `make <target>`.
 | `gobuild`              | Runs a normal `go build` of the agent and stores the binary in `./out/amazon-ecs-agent` |
 | `static`               | Runs `go build` to produce a static binary in `./out/amazon-ecs-agent` |
 | `test`                 | Runs all unit tests using `go test` |
-| `test-in-docker`       | Runs all tests inside a Docker container |
```
fierlion, line 245: Why is `test-in-docker` going away, and can you add a `test-in-docker-init` to the readme?

packaging/amazon-linux-ami/ecs-init.spec
```diff
@@ -49,7 +49,6 @@ Source5:        amazon-ecs-volume-plugin.conf
 Source6:        amazon-ecs-volume-plugin.service
 Source7:        amazon-ecs-volume-plugin.socket
 
-BuildRequires:  golang >= 1.7
```
fierlion, line 52: we'll be adding this back once we start testing in Koji

scripts/test-init
```diff
@@ -12,6 +12,6 @@
 # express or implied. See the License for the specific language governing
 # permissions and limitations under the License.
 set -x
-make get-deps
-make govet
-make test
+make get-deps-init
+make govet-init
+make test-init
```
fierlion, line 17: just to be sure -- this isn't changing an existing script in a make target is it?
do we use the `scripts/test` in the `make test` target?

README.md
lydiafilipe: This was referencing an agent target that no longer exists, decided to remove it as to avoid confusion with the init one.

I thought about how to best approach updating the README. My thinking is, since make targets and ways we are building are a bit in flux and not cleaned up yet, that we should update the README at a later point with all of the updated targets we're going to keep, and updated instructions for the dockerless buillds. Otherwise, might risk adding information that will quickly become incorrect and might forget to update it

scripts/test-init
lydiafilipe: This script is only used by the test-init dockerfile (which came from the init repo), which is only used by the test-in-docker-init target that I added to the Makefile from the init Makefile (and added the init suffix to).

Makefile
lydiafilipe: This is the version we're actually currently using in agent for get-deps, so went with it for consistency. I had to change this due to receiving the following error:
```
package io/fs: unrecognized import path "io/fs": import path does not begin with hostname
```
Based on [this](https://github.com/tdewolff/minify/issues/388) github issue, looks like this is caused by the latest version of that package requiring at least go 16.

fierlion (approved):

PR #3156, opened 2022-03-25
Apply minimumCPUShare to both task and container CPU shares

yinyic:
### Summary
`taskCPUShares` is used when creating task cgroup spec. We have a `minimumCPUShare` (equals 2) but it currently applies to [container CPU shares](https://github.com/aws/amazon-ecs-agent/blob/6f6e351b3b266984e257a88042f73ef4b9469d6b/agent/api/task/task_unsupported.go#L53) only (more context under `cpu` section of the API [doc](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html)). 

This value should be applicable to task CPU share as well for the following reasons:
1. We sum up container CPU shares and add that to task CPU share. Because a task must have at least one container, and the container CPU share will be >= 2, then task CPU share should naturally be >= 2.
2. Cgroupv2 requires CPU share to be >= 2 due to this [conversion](https://github.com/aws/amazon-ecs-agent/blob/6f6e351b3b266984e257a88042f73ef4b9469d6b/agent/api/task/task_unsupported.go#L53) from CPU shares to CPU weight. This is where the conversion happens in the `containerd/cgroups` [library](https://github.com/containerd/cgroups/blob/d5e6fc31636bab60490f78e24e36d79eb50cbd69/v2/utils.go#L171-L174) that we use.

### Implementation details

When add each container CPU share to task CPU share, use `minimumCPUShare` if container CPU share is less than that. 

### Testing

New tests cover the changes: unit test

Additional manual testing: applied this commit to the [cgroupv2 branch](https://github.com/aws/amazon-ecs-agent/pull/3127), started Agent on an AL2022 instance where cgroupv2 is enabled, and verified that task with container cpu share equals 1 can be started. 

Apply minimumCPUShare to both task and container CPU shares

### Licensing

### DIFF
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -148,16 +148,11 @@ func (task *Task) buildImplicitLinuxCPUSpec() specs.LinuxCPU {
 	// aggregate container CPU shares when present
 	var taskCPUShares uint64
 	for _, container := range task.Containers {
-		if container.CPU > 0 {
+		if container.CPU < minimumCPUShare {
+			taskCPUShares += minimumCPUShare
+		} else {
 			taskCPUShares += uint64(container.CPU)
 		}
-	}
-
-	// If there are are no CPU limits at task or container level,
-	// default task CPU shares
-	if taskCPUShares == 0 {
-		// Set default CPU shares
-		taskCPUShares = minimumCPUShare
 	}
 
 	return specs.LinuxCPU{
--- a/agent/api/task/task_linux_test.go
+++ b/agent/api/task/task_linux_test.go
@@ -342,6 +342,31 @@ func TestBuildLinuxResourceSpecWithoutTaskCPUWithContainerCPULimits(t *testing.T
 	assert.EqualValues(t, expectedLinuxResourceSpec, linuxResourceSpec)
 }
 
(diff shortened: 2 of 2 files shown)


agent/api/task/task_linux.go
```diff
@@ -148,18 +148,13 @@ func (task *Task) buildImplicitLinuxCPUSpec() specs.LinuxCPU {
 	// aggregate container CPU shares when present
 	var taskCPUShares uint64
 	for _, container := range task.Containers {
-		if container.CPU > 0 {
+		if container.CPU < minimumCPUShare {
```
singholt, line 151: if I have 2 containers with `container.CPU = 0`. then `taskCPUShares` would become 4?

yinyic: Yes - I believe that should be the correct behavior because we later on [bump](https://github.com/aws/amazon-ecs-agent/blob/6f6e351b3b266984e257a88042f73ef4b9469d6b/agent/api/task/task_unsupported.go#L53) container CPU shares to 2, if the value specified in task def is smaller (or unspecified). So if you have 2 containers they will both have CPUShares=2 when passed to docker, which makes sense for the task CPU share to be 4 in that case.

singholt (approved):

fenxiong (approved):

PR #3164, opened 2022-04-05
specfile and script updates for integrated init/agent build

lydiafilipe:
### Summary
Create a new specfile and update scripts so that a legacy integrated init/agent rpm build works.

### Implementation details
Changes include: 
#### Makefile 
- targets for building rpm with agent using the new spec file
    - sources target includes much more than current init sources target and depends on get-cni-sources

#### Spec File (copied from amazon-linux-ami/ecs-init specfile)
- added buildRequires for `glibc-static` (required for building pause)
- use the dockerless build scripts
- install the agent tar from the build into the cache directory and reflect in the state file that a cached version of the agent is available
- removed s3 agent sources and the setting up/linking of those
- only built for x86_64 (but need to support arm soon)
- removed changelog, and provides statements
- removed some al2 specific logic since this is for the legacy branch

#### Scripts
agent and cni plugins build scripts
- building in build directory symlinked to directories based on expected structure
- setting GOPATH
- (agent) setting go modules to auto
- (plugins) turning go modules off, using the "disablekubeapi" build tag, and only building the binaries without running unit tests when building amazon-vpc-cni-plugins 

### Testing
- Built rpm witih make target `make rpm-with-agent` (commenting out BuildRequires golang >= 1.7)
- Built in internal rpm build system
- Built ami with the rpm using the amazon-ecs-ami package ([with slight modifications](https://github.com/lydiafilipe/amazon-ecs-ami/commit/dc61cd5c887f0e7b428a7b2d841b0588b81ecbae)) and used it in launching instances, added instance to cluster and ran tasks

specfile and script updates for integrated init/agent build

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -394,16 +394,23 @@ srpm: .srpm-done
 
 rpm: .rpm-done
 
-ubuntu-trusty:
-	cp packaging/ubuntu-trusty/ecs.conf ecs.conf
-	tar -czf ./amazon-ecs-init_${VERSION}.orig.tar.gz ecs-init ecs.conf scripts README.md
-	mkdir -p BUILDROOT
-	cp -r packaging/ubuntu-trusty/debian BUILDROOT/debian
-	cp -r ecs-init BUILDROOT
-	cp packaging/ubuntu-trusty/ecs.conf BUILDROOT
-	cp -r scripts BUILDROOT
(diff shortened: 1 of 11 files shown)


fierlion (approved):
Tested changes locally -- nice.

prateekchaudhry (approved):

PR #3174, opened 2022-04-14
Connect to stats endpoint for service connect enabled tasks

ubhattacharjya:
### Summary

This implements the functionality of ECS agent connecting to the stats endpoint exposed by the Appnet agent. For AWSVPC mode, we will use the IP of the ecs-bridge to connect to the endpoint. For Bridge mode, we will use the IP of the appnet agent container. For now, we have considered the envoy container as the appnet agent container.

### Implementation details

in stats/engine.go -> We maintain a map where we correlate every task with a service connect stats object. Whenever we start a task, the task gets added to this map.
in stats/service_connect_linux.go -> We actually connect to the stats endpoint to get the data. The structure of the stats data has not yet been implemented and will be implemented in subsequent PR. For now, it is a simple string.

### Testing

Unit tests cover the change

Manual test:
* AWSVPC Mode: Ran an Appmesh task, and tested that we are able to hit the stats endpoint
* Bridge Mode: Ran a dummy envoy container along with an application container. tested that we are able to hit the stats endpoint

Feature - Connect to stats endpoint for service connect metrics

### Licensing

### DIFF
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -71,6 +71,8 @@ type Engine interface {
 	GetInstanceMetrics() (*ecstcs.MetricsMetadata, []*ecstcs.TaskMetric, error)
 	ContainerDockerStats(taskARN string, containerID string) (*types.StatsJSON, *NetworkStatsPerSec, error)
 	GetTaskHealthMetrics() (*ecstcs.HealthMetadata, []*ecstcs.TaskHealth, error)
+	//TODO [SC]: Return service connect metrics from GetServiceConnects when the metrics are defined.
+	GetServiceConnectStats() error
 }
 
 // DockerStatsEngine is used to monitor docker container events and to report
@@ -91,8 +93,9 @@ type DockerStatsEngine struct {
 	// tasksToHealthCheckContainers map task arns to the containers that has health check enabled
 	tasksToHealthCheckContainers map[string]map[string]*StatsContainer
 	// tasksToDefinitions maps task arns to task definition name and family metadata objects.
-	tasksToDefinitions map[string]*taskDefinition
-	taskToTaskStats    map[string]*StatsTask
+	tasksToDefinitions        map[string]*taskDefinition
+	taskToTaskStats           map[string]*StatsTask
+	taskToServiceConnectStats map[string]*ServiceConnectStats
 }
 
 // ResolveTask resolves the api task object, given container id.
@@ -144,6 +147,7 @@ func NewDockerStatsEngine(cfg *config.Config, client dockerapi.DockerClient, con
 		tasksToHealthCheckContainers: make(map[string]map[string]*StatsContainer),
 		tasksToDefinitions:           make(map[string]*taskDefinition),
 		taskToTaskStats:              make(map[string]*StatsTask),
+		taskToServiceConnectStats:    make(map[string]*ServiceConnectStats),
 		containerChangeEventStream:   containerChangeEventStream,
 	}
 }
@@ -305,6 +309,18 @@ func (engine *DockerStatsEngine) addToStatsTaskMapUnsafe(task *apitask.Task, doc
 	}
 }
 
+func (engine *DockerStatsEngine) addTaskToServiceConnectStatsUnsafe(taskArn string) {
+	_, taskExists := engine.taskToServiceConnectStats[taskArn]
+	if !taskExists {
+		serviceConnectStats, err := newServiceConnectStats()
+		if err != nil {
+			seelog.Errorf("Error adding task %s to the service connect stats watchlist : %v", taskArn, err)
+			return
+		}
+		engine.taskToServiceConnectStats[taskArn] = serviceConnectStats
+	}
+}
+
 // addContainerUnsafe adds a container to the map of containers being watched.
 func (engine *DockerStatsEngine) addContainerUnsafe(dockerID string) (*StatsContainer, *StatsTask, error) {
 	// Make sure that this container belongs to a task and that the task
@@ -349,6 +365,10 @@ func (engine *DockerStatsEngine) addContainerUnsafe(dockerID string) (*StatsCont
 		// Track the container health status
 		engine.addToStatsContainerMapUnsafe(task.Arn, dockerID, statsContainer, engine.healthCheckContainerMapUnsafe)
 		seelog.Debugf("Adding container to stats health check watch list, id: %s, task: %s", dockerID, task.Arn)
+	}
+
+	if errResolveContainer == nil {
+		engine.addTaskToServiceConnectStatsUnsafe(task.Arn)
 	}
 
 	if !watchStatsContainer {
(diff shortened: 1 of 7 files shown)


agent/stats/engine.go
```diff
@@ -305,6 +308,17 @@ func (engine *DockerStatsEngine) addToStatsTaskMapUnsafe(task *apitask.Task, doc
 	}
 }
 
+func (engine *DockerStatsEngine) addTaskToServiceConnectStatsUnsafe(taskArn string) {
+	_, taskExists := engine.taskToServiceConnectStats[taskArn]
+	if !taskExists {
+		serviceConnectStats, err := newServiceConnectStats()
+		if err != nil {
+			return
```
angelcar, line 316: Do we want to log the error here?

agent/stats/engine.go
```diff
@@ -835,3 +859,26 @@ func (engine *DockerStatsEngine) ContainerDockerStats(taskARN string, containerI
 
 	return containerStats, containerNetworkRateStats, nil
 }
+
+func (engine *DockerStatsEngine) GetServiceConnectStats() error {
```
angelcar, line 863: exported methods should have a comment

agent/stats/service_connect_linux.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (sc *ServiceConnectStats) retrieveServiceConnectStats(task *apitask.Task, wg *sync.WaitGroup) {
+	defer wg.Done()
+	var ipAddress string
+	if task.IsNetworkModeAWSVPC() {
+		ipAddress = task.GetLocalIPAddress()
+	} else {
+		for _, container := range task.Containers {
+			// TODO: Implement a proper check for appnet agent container when it is defined
```
angelcar, line 53: can we add a name convention for SC TODOS, I was thinking `// TODO [SC]: ...`

agent/stats/service_connect_linux.go
```diff
@@ -0,0 +1,90 @@
...
+				(container.GetNetworkModeFromHostConfig() == "" || container.GetNetworkModeFromHostConfig() == bridgeNetworkMode) {
+				network := *container.GetNetworkSettings()
+				networkSettings := network.DefaultNetworkSettings
+				ipAddress = networkSettings.IPAddress
+				break
+			}
+		}
+	}
+	// TODO: Replace the stats url when it is defined
+	statsEndpoint, _ := url.Parse(fmt.Sprintf("http://" + ipAddress + statsPort))
```
angelcar, line 64: what if we don't find an ip address meaning we don't find the sc container?

agent/stats/service_connect_linux.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (sc *ServiceConnectStats) setStats(stats string) {
+	sc.lock.Lock()
+	defer sc.lock.Unlock()
+
+	sc.stats = stats
+}
+
+func (sc *ServiceConnectStats) GetStats() string {
+	sc.lock.Lock()
```
angelcar, line 88: this should be `RLock`/`RUnlock`

agent/stats/engine.go
```diff
@@ -71,6 +71,7 @@ type Engine interface {
 	GetInstanceMetrics() (*ecstcs.MetricsMetadata, []*ecstcs.TaskMetric, error)
 	ContainerDockerStats(taskARN string, containerID string) (*types.StatsJSON, *NetworkStatsPerSec, error)
 	GetTaskHealthMetrics() (*ecstcs.HealthMetadata, []*ecstcs.TaskHealth, error)
+	GetServiceConnectStats() error
```
mythri-garaga, line 74: We should probably return the stats here since tacs client can not access the stats stored in `taskToServiceConnectStats` map?

agent/stats/service_connect_linux.go
```diff
@@ -0,0 +1,92 @@
...
+	"strings"
+	"sync"
+
+	"github.com/cihub/seelog"
+
+	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+)
+
+const (
+	statsPort         = ":9901/stats/prometheus"
```
angelcar, line 32: can we change the variable to `statsURLPath` or something? `statsPort` is misleading.

agent/stats/engine.go
```diff
@@ -835,3 +859,26 @@ func (engine *DockerStatsEngine) ContainerDockerStats(taskARN string, containerI
...
+		task, err := engine.resolver.ResolveTaskByARN(taskArn)
+		if err != nil {
+			return errors.Errorf("stats engine: task '%s' not found", taskArn)
+		}
+		// TODO: Check if task is service-connect enabled
+		serviceConnectStats, ok := engine.taskToServiceConnectStats[taskArn]
+		if !ok {
+			return errors.Errorf("task '%s' is not registered to collect service connect metrics", taskArn)
+		}
+		go serviceConnectStats.retrieveServiceConnectStats(task, &wg)
```
angelcar, line 880: I think `retrieveServiceConnectStats` doesn't need to be aware of the wait group. Instead you can call it like

```go
go func() {
    serviceConnectStats.retrieveServiceConnectStats()
    wg.Done()
}
```

agent/stats/engine.go
ubhattacharjya: You are right. I will implement that in a subsequent PR where I parse the stats and return the metrics following the format of the general metrics in the TACS model. I will add a TODO here.

agent/stats/service_connect_linux.go
ubhattacharjya: makes sense.

agent/stats/service_connect_linux.go
ubhattacharjya: Thanks for catching that.

agent/stats/service_connect_linux.go
ubhattacharjya: The method will print an error log "Could not connect to stats endpoint for task %s: %v", task.Arn, err" . Does that sound reasonable?

angelcar (approved):

agent/stats/engine.go
mythri-garaga: I made the comment assuming that we would call to `GetServiceConnectStats()` from [TACS client](https://github.com/aws/amazon-ecs-agent/blob/fe79b107f3b204ca1a831b1d15424d12d2948bdf/agent/tcs/client/client.go#L204-L243) but on second thought, [ServiceConnectMetrics](https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/tcs/model/ecstcs/api.go#L994)  is returned as part of `GetInstanceMetrics()` call 
so I guess it is alright if `GetServiceConnectStats()` does not return stats since  `GetInstanceMetrics()`  has to access `taskToServiceConnectStats`.

mythri-garaga (approved):

PR #3175, opened 2022-04-14
engine: skip some tests on Mac

nodir-t:
### Summary

Building a CNI configuration is supported only on Linux and Windows. On other platforms, it fails with "unsupported platform": 
https://github.com/aws/amazon-ecs-agent/blob/05730614021c70c7f39d7cc00264aaffc5b5b0d3/agent/api/task/task_unsupported.go#L96-L100

However, 3 tests in `engine` package that rely on CNI are not skipped on other platforms. 

### Implementation details

Skip the impacted tests by checking `GOOS` in runtime. Do not use build tags because there are other tests in the same
file that pass.

### Testing

```
cd agent
go test -tags unit ./engine
```

engine: skip some tests on Mac

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine_test.go
+++ b/agent/engine/docker_task_engine_test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


angelcar:
Hmm, wonder if we should avoid returning an error in task_unsupported.go instead. I think it's safe to do that since we're not building binaries for unsupported platforms anyways.

Realmonia (approved):

nodir-t:
@fierlion let's decide to close or merge this PR

fierlion:
closing for now -- will reopen as part of future work

PR #3184, opened 2022-04-20
Fix credential rotation issue for ECS-A Windows

vsiddharth:
### Summary

Fix credential rotation issue for ECS-A Windows

### Implementation details

Credentials were not being rotated properly on ECS-A Windows instances.
This patch addresses the issue by using the correct file-paths for
credentials on supported platforms. The credential chain hierarchy is
also updated on ECS-A windows to ensure that credential chain is not
broken for other launch types.

### Testing

 
Manually updated the ECS-Agent on both ECS/EC2 and ECS/ECS-A Windows instances to validate credentials being refreshed. 

New tests cover the changes:  No

 Bug - Fixed credential rotation issue with ECS-A Windows

### Licensing

### DIFF
--- a/agent/credentials/instancecreds/instancecreds_linux.go
+++ b/agent/credentials/instancecreds/instancecreds_linux.go
@@ -1,0 +1,52 @@
+//go:build linux
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package instancecreds
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/credentials/providers"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/cihub/seelog"
+)
+
+// GetCredentials returns the instance credentials chain. This is the default chain
+// credentials plus the "rotating shared credentials provider", so credentials will
+// be checked in this order:
+//    1. Env vars (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY).
+//    2. Shared credentials file (https://docs.aws.amazon.com/ses/latest/DeveloperGuide/create-shared-credentials-file.html) (file at ~/.aws/credentials containing access key id and secret access key).
+//    3. EC2 role credentials. This is an IAM role that the user specifies when they launch their EC2 container instance (ie ecsInstanceRole (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html)).
+//    4. Rotating shared credentials file located at /rotatingcreds/credentials
+func GetCredentials(isExternal bool) *credentials.Credentials {
+	mu.Lock()
+	if credentialChain == nil {
+		credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
+		credProviders = append(credProviders, providers.NewRotatingSharedCredentialsProvider())
+		credentialChain = credentials.NewCredentials(&credentials.ChainProvider{
+			VerboseErrors: false,
+			Providers:     credProviders,
+		})
+	}
+	mu.Unlock()
+
+	// credentials.Credentials is concurrency-safe, so lock not needed here
+	v, err := credentialChain.Get()
+	if err != nil {
+		seelog.Errorf("Error getting ECS instance credentials from default chain: %s", err)
+	} else {
+		seelog.Infof("Successfully got ECS instance credentials from provider: %s", v.ProviderName)
+	}
+	return credentialChain
+}
--- a/agent/credentials/instancecreds/instancecreds_test.go
+++ b/agent/credentials/instancecreds/instancecreds_test.go
@@ -25,11 +25,10 @@ import (
 
 func TestGetCredentials(t *testing.T) {
 	credentialChain = nil
-	credsA := GetCredentials()
+	credsA := GetCredentials(false)
 	require.NotNil(t, credsA)
-	credsB := GetCredentials()
+	credsB := GetCredentials(true)
 	require.NotNil(t, credsB)
-	require.Equal(t, credsA, credsB)
 }
 
 // test that env vars override all other provider types
@@ -44,7 +43,7 @@ func TestGetCredentials_EnvVars(t *testing.T) {
 	defer os.Setenv("AWS_ACCESS_KEY_ID", origAKID)
 	defer os.Setenv("AWS_SECRET_ACCESS_KEY", origSecret)
 
-	creds := GetCredentials()
+	creds := GetCredentials(false)
 	require.NotNil(t, creds)
 	v, err := creds.Get()
 	require.NoError(t, err)
@@ -81,7 +80,7 @@ aws_secret_access_key = TESTFILESECRET
 	// reset before exiting
 	defer os.Setenv("AWS_SHARED_CREDENTIALS_FILE", origEnv)
 
-	creds := GetCredentials()
+	creds := GetCredentials(false)
 	require.NotNil(t, creds)
 	v, err := creds.Get()
 	require.NoError(t, err)
--- a/agent/credentials/instancecreds/instancecreds_windows.go
+++ b/agent/credentials/instancecreds/instancecreds_windows.go
@@ -1,0 +1,65 @@
+//go:build windows
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package instancecreds
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/credentials/providers"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/defaults"
+	"github.com/cihub/seelog"
+)
+
+// GetCredentials returns the instance credentials chain. This is the default chain
+// credentials plus the "rotating shared credentials provider", so credentials will
+// be checked in this order:
+//    1. Env vars (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY).
+//    2. Shared credentials file (https://docs.aws.amazon.com/ses/latest/DeveloperGuide/create-shared-credentials-file.html) (file at ~/.aws/credentials containing access key id and secret access key).
+//    3. EC2 role credentials. This is an IAM role that the user specifies when they launch their EC2 container instance (ie ecsInstanceRole (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html)).
+//    4. Rotating shared credentials file located at /rotatingcreds/credentials
+//
+//    The default credential chain provided by the SDK includes:
+//    * EnvProvider
+//    * SharedCredentialsProvider
+//    * RemoteCredProvider (EC2RoleProvider)
+//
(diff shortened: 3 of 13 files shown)


agent/credentials/instancecreds/instancecreds.go
```diff
@@ -38,7 +39,25 @@ func GetCredentials() *credentials.Credentials {
...
+
+			In the case of ECS-A on Windows, the `SharedCredentialsProvider` takes
+			precedence over the `RotatingSharedCredentialsProvider` and this results
+			in the credentials not being refreshed. To mitigate this issue, we will
+			use the environment variable `ECS_EXTERNAL` to reorder the credential
+			chain and ensure that `RotatingSharedCredentialsProvider` takes precedence
+			over the `SharedCredentialsProvider`.
+
+		*/
+		if _, ok := os.LookupEnv("ECS_EXTERNAL"); ok {
```
jterry75, line 56: Do we need to do this only for Windows?

jterry75: What is the case any of the other providers are used? Should we just be putting the rotating creds as the only one in the list?

agent/credentials/providers/credentials_filename_linux.go
```diff
@@ -0,0 +1,22 @@
+//go:build linux
```
jterry75, line 1: Unfortunately, you really need to have both at the moment:

`// +build linux`

jterry75: Actually this ends in `_linux` you dont need it at all`

agent/credentials/providers/credentials_filename_windows.go
```diff
@@ -0,0 +1,24 @@
+//go:build windows
```
jterry75, line 1: You dont need this

agent/credentials/providers/credentials_filename_windows.go
```diff
@@ -0,0 +1,24 @@
...
+// permissions and limitations under the License.
+
+package providers
+
+import "path/filepath"
+
+var (
+	// defaultRotatingCredentialsFilename is the default location of the credentials file
+	// for RotatingSharedCredentialsProvider.
+	defaultRotatingCredentialsFilename = filepath.Join("C:\\Windows\\System32\\config\\systemprofile\\.aws", "credentials")
```
jterry75, line 23: You should just write the full path here. We hard coded everything so just make it a const.

agent/credentials/instancecreds/instancecreds.go
vsiddharth: Currently, this appears to be broken only for ECS-A Windows. 

In certain cases, customers might want to use the `EnvProvider` and I think that it should always be the first provider in the chain.

agent/credentials/providers/credentials_filename_linux.go
vsiddharth: I agree but I followed the convention followed across in this repo.

agent/credentials/providers/credentials_filename_unsupported.go
```diff
@@ -0,0 +1,22 @@
+//go:build !windows && !linux
```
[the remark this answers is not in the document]
vsiddharth: Many folks attempt to build things on MAC. We can totally get rid of this if required.

agent/credentials/providers/credentials_filename_windows.go
vsiddharth: I agree but I followed the convention followed across in this repo.

agent/credentials/providers/credentials_filename_windows.go
vsiddharth: Sure thing. Will update this.

agent/credentials/providers/credentials_filename_windows.go
jterry75: IMO we should not add tags that golang handles by default. Even if its done in other places in the repo

agent/credentials/providers/credentials_filename_unsupported.go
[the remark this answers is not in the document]
jterry75: We should remove. It will be a compile failure on Mac anyways so we dont need the file to make the compilation succeed and then fail at runtime

agent/credentials/instancecreds/instancecreds.go
jterry75: Ok sounds good.

agent/credentials/instancecreds/instancecreds.go
```diff
@@ -38,7 +39,25 @@ func GetCredentials() *credentials.Credentials {
 	mu.Lock()
 	if credentialChain == nil {
 		credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
-		credProviders = append(credProviders, providers.NewRotatingSharedCredentialsProvider())
+		/*
```
sparrc, line 42: This comment repeats but also now contradicts the function signature comment above. Please remove the comment here and update the GetCredentials comment

agent/credentials/instancecreds/instancecreds.go
sparrc: This will return true if ECS_EXTERNAL=false is set.

But you shouldn't be looking up env vars here anyways, you should use the config struct for this, like we do here: https://github.com/aws/amazon-ecs-agent/blob/master/agent/app/agent.go#L176-L180

sparrc: alternatively we could only do this reordering for windows, since on linux we run inside a container and don't have an issue with credential file collision

vsiddharth: The config struct cannot be used directly as it would result in an import cycle. 

However, we could pass in a boolean flag like `isExternal` to this method and update all its references within the repo. 

Ex:
```
* https://github.com/aws/amazon-ecs-agent/blob/225bc3a556bd2d1759ab27b23f54e7e68086c9f0/agent/ecr/factory.go#L78
* https://github.com/aws/amazon-ecs-agent/blob/225bc3a556bd2d1759ab27b23f54e7e68086c9f0/agent/ec2/ec2_client.go#L56
* https://github.com/aws/amazon-ecs-agent/blob/ed81f6b41241ab8ebba1bcec780352c58bf77bd0/agent/ec2/ec2_metadata_client.go#L97
* https://github.com/aws/amazon-ecs-agent/blob/f5ac255da315b39612ad95361247f74eb3db93ae/agent/app/agent.go#L228
```

vsiddharth: Do you recommend that the Linux credential chain remain unchanged for ECS-A?

agent/credentials/instancecreds/instancecreds.go
```diff
@@ -34,16 +34,30 @@ var (
...
+//    in the credentials not being refreshed. To mitigate this issue, we will
+//    reorder the credential chain and ensure that `RotatingSharedCredentialsProvider`
+//    takes precedence over the `SharedCredentialsProvider` for ECS-A.
+
+func GetCredentials(isExternal bool) *credentials.Credentials {
 	mu.Lock()
-	if credentialChain == nil {
-		credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
+	credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
+	if isExternal {
```
jterry75, line 52: I thought Cam said only do this for Windows? Am I crazy?

agent/credentials/instancecreds/instancecreds_test.go
```diff
@@ -25,11 +25,10 @@ import (
 
 func TestGetCredentials(t *testing.T) {
 	credentialChain = nil
-	credsA := GetCredentials()
+	credsA := GetCredentials(false)
 	require.NotNil(t, credsA)
-	credsB := GetCredentials()
+	credsB := GetCredentials(true)
```
jterry75, line 30: Should this be `true` or should we have a specific external test case?

agent/credentials/providers/credentials_filename_linux.go
jterry75: We really need to have either both, or none. But half way is not good. Even if it is in the repo

agent/credentials/instancecreds/instancecreds.go
vsiddharth: I think we could safely do it for both platforms but I'll wait for @sparrc to respond.

agent/credentials/instancecreds/instancecreds_test.go
vsiddharth: We could have a specific test case to cover the new change but this should suffice too.

agent/credentials/instancecreds/instancecreds.go
```diff
@@ -34,16 +34,30 @@ var (
...
+//    * EnvProvider
+//    * SharedCredentialsProvider
+//    * RemoteCredProvider (EC2RoleProvider)
+//
+//    In the case of ECS-A on Windows, the `SharedCredentialsProvider` takes
+//    precedence over the `RotatingSharedCredentialsProvider` and this results
+//    in the credentials not being refreshed. To mitigate this issue, we will
+//    reorder the credential chain and ensure that `RotatingSharedCredentialsProvider`
+//    takes precedence over the `SharedCredentialsProvider` for ECS-A.
+
```
sparrc, line 48: nit: I believe godoc is not properly generated if there's a newline between the function's comment and the function.

agent/credentials/instancecreds/instancecreds_linux.go
```diff
@@ -0,0 +1,52 @@
...
+)
+
+// GetCredentials returns the instance credentials chain. This is the default chain
+// credentials plus the "rotating shared credentials provider", so credentials will
+// be checked in this order:
+//    1. Env vars (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY).
+//    2. Shared credentials file (https://docs.aws.amazon.com/ses/latest/DeveloperGuide/create-shared-credentials-file.html) (file at ~/.aws/credentials containing access key id and secret access key).
+//    3. EC2 role credentials. This is an IAM role that the user specifies when they launch their EC2 container instance (ie ecsInstanceRole (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html)).
+//    4. Rotating shared credentials file located at /rotatingcreds/credentials
+func GetCredentials(isExternal bool) *credentials.Credentials {
```
jterry75, line 32: For Linux mark this as `_` since its unused right?

jterry75 (approved):
LGTM. Just one nit

sparrc (approved):

agent/credentials/instancecreds/instancecreds_windows.go
```diff
@@ -0,0 +1,65 @@
...
+//    In the case of ECS-A on Windows, the `SharedCredentialsProvider` takes
+//    precedence over the `RotatingSharedCredentialsProvider` and this results
+//    in the credentials not being refreshed. To mitigate this issue, we will
+//    reorder the credential chain and ensure that `RotatingSharedCredentialsProvider`
+//    takes precedence over the `SharedCredentialsProvider` for ECS-A.
+func GetCredentials(isExternal bool) *credentials.Credentials {
+	mu.Lock()
+	credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
+	if isExternal {
+		credProviders = append(credProviders[:1], append([]credentials.Provider{providers.NewRotatingSharedCredentialsProvider()}, credProviders[1:]...)...)
```
mythri-garaga, line 47: I think `RotatingSharedCredentialsProvider` is getting appended at the end of the `credProviders` here and really taking the precedence.

Should this be 
```
credProviders = append(append([]credentials.Provider{providers.NewRotatingSharedCredentialsProvider()}, credProviders[1:]...), credProviders[:1]...)
```

vsiddharth: We want `RotatingSharedCredentialsProvider` to immediately follow the `EnvProvider` and this change should do that.

mythri-garaga (approved):

jdstuart:
@vsiddharth before merging should you also consider bumping the go version to `1.17` or `1.18` to add support for `windows/arm64` architectures?
cc: @mythri-garaga @sparrc

vsiddharth:
> @vsiddharth before merging should you also consider bumping the go version to `1.17` or `1.18` to add support for `windows/arm64` architectures? cc: @mythri-garaga @sparrc PS. Thanks for the quick turnaround on this.

The agent is already using go `1.17.5`
* https://github.com/aws/amazon-ecs-agent/blob/master/GO_VERSION_WINDOWS
* https://github.com/aws/amazon-ecs-agent/blob/master/GO_VERSION

Can you please provide additional context on this?

sparrc:
> before merging should you also consider bumping the go version to 1.17 or 1.18 to add support for windows/arm64 architectures?

I think windows/arm64 would be a larger feature to support, are you currently custom-building the ecs agent to run on that platform?

jdstuart:
No, we do not currently run on `windows/arm64`.

I mistakenly looked at https://github.com/aws/amazon-ecs-agent/blob/master/.travis.yml and thought go `1.15` is still used. I didn't see the `GO_VERION` files.

PR #3185, opened 2022-04-22
Ephemeral Port Selection for Service Connect

angelcar:
### Summary

This PR assigns ephemeral ports to listeners that explicitly require it. All ports in the task must be unique, these include the ones exposed in task definition.

Note that the ports generated are going to be bound in the task netns, therefore there should be no collisions. As an extra precaution, the SC CNI plugin (future PR) will reserve the ports generated in this PR using [ip_local_reserved_ports](https://www.kernel.org/doc/html/latest/networking/ip-sysctl.html) kernel feature, which should guarantee there won't be  collisions with SC ports.

### Implementation details

Currently, if `ListenerPort=0` in the `ServiceConnectConfig`, it means an ephemeral port should be assigned to that listener. Nothing else in `ServiceConnectConfig` should be modified. 

A function to generate N ephemeral ports, which excludes reserved/utilized ports is added. This function guarantees that N unique  ephemeral ports (i.e in the 32768-60999 range) will be generated, and that those ports will not collide with the reserved ports passed as an argument.

### Testing

N/A
### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -343,7 +343,13 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 		}
 	}
 
-	task.initServiceConnectResources()
+	if err := task.initServiceConnectResources(); err != nil {
+		logger.Error("Could not initialize Service Connect resources", logger.Fields{
+			field.TaskID: task.GetID(),
+			field.Error:  err,
+		})
+		return apierrors.NewResourceInitError(task.Arn, err)
+	}
 
 	if err := task.initializeContainerOrderingForVolumes(); err != nil {
 		logger.Error("Could not initialize volumes dependency for container", logger.Fields{
@@ -437,7 +443,7 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	return nil
 }
 
-func (task *Task) initServiceConnectResources() {
+func (task *Task) initServiceConnectResources() error {
 	// TODO [SC]: ServiceConnectConfig will come from ACS. Adding this here for dev/testing purposes only Remove when
 	// ACS model is integrated
 	if task.ServiceConnectConfig == nil {
@@ -448,8 +454,12 @@ func (task *Task) initServiceConnectResources() {
 	if task.IsServiceConnectEnabled() {
 		// TODO [SC]: initDummyServiceConnectConfig is for dev testing only, remove it when final SC model from ACS is in place
 		task.initDummyServiceConnectConfig()
-		task.configureContainerDependenciesForServiceConnect()
+		task.initServiceConnectContainerDependencies()
+		if err := task.initServiceConnectEphemeralPorts(); err != nil {
+			return err
+		}
 	}
+	return nil
 }
 
 // TODO [SC]: This is for dev testing only, remove it when final SC model from ACS is in place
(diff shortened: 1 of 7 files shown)


agent/utils/ephemeral_ports.go
```diff
@@ -0,0 +1,55 @@
...
+// From https://www.kernel.org/doc/html/latest//networking/ip-sysctl.html#ip-variables
+const (
+	EphemeralPortMin         = 32768
+	EphemeralPortMax         = 60999
+	maxPortSelectionAttempts = 100
+)
+
+// GenerateEphemeralPortNumbers generates a list of n unique port numbers in the 32768-60999 range. The resulting port
+// number list is guaranteed to not include any port number present in "reserved" parameter.
+func GenerateEphemeralPortNumbers(n int, reserved []uint16) ([]uint16, error) {
```
yinyic, line 31: In addition to avoiding the "reserved" ports, there can be other processes on the host instance utilizing ephemeral ports. Although the chance for port contention in the ephemeral range can be relatively low, but it's still possible. And when it happens, the task won't be able to start. Therefore, I think we should have some checks here to see if the selected port is available on the host.

One way to do that is to try create a simple server with `net.Listen("tcp", ":" + port)`, and it will throw error if port is already in use. There might be other simpler ways that I'm not aware of.

yinyic (dismissed):

agent/utils/ephemeral_ports.go
yinyic: Discussed offline and clarified that those ports will be isolated to task netns. And because appnet is the first to start in task netns, there shouldn't be any port collision.

agent/api/task/task.go
```diff
@@ -479,6 +489,53 @@ func (task *Task) configureContainerDependenciesForServiceConnect() {
...
+	for _, ic := range task.ServiceConnectConfig.IngressConfig {
+		if ic.ListenerPort == 0 { // This means listener port was not sent to us by ACS, signaling the port needs to be ephemeral
+			numEphemeralPortsNeeded++
+		} else {
+			utilizedPorts = append(utilizedPorts, ic.ListenerPort)
+		}
+	}
+
+	// Presently, SC egress port is always ephemeral, but adding this for future-proofing
+	if task.ServiceConnectConfig.EgressConfig.ListenerPort == 0 {
```
rajal-amzn, line 505: Do we want `ListenerPort` to be defined in the `EgressConfig` model as well?

agent/utils/ephemeral_ports.go
```diff
@@ -0,0 +1,55 @@
...
+	rand.Seed(time.Now().UnixNano())
+
+	var result []uint16
+	var portSelectionAttempts int
+	for len(result) < n {
+		// The intention of maxPortSelectionAttempts is to avoid a super highly unlikely case where we
+		// keep getting ports that collide, thus creating an infinite loop.
+		if portSelectionAttempts > maxPortSelectionAttempts {
+			return nil, fmt.Errorf("maximum number of attempts to generate unique ports reached")
+		}
```
rajal-amzn, line 45: Could we add a test for this scenario?

agent/api/task/task.go
angelcar: Yes, this is the internal model, not the ACS model. I want to keep all the SC info in one place.

agent/utils/ephemeral_ports.go
angelcar: It actually wasn't that hard :)

agent/api/task/task.go
rajal-amzn: I see. That makes sense.

agent/utils/ephemeral_ports.go
rajal-amzn: Nice. Thanks.

rajal-amzn (approved):

ubhattacharjya (approved):

yinyic (approved):

PR #3190, opened 2022-04-27
Sc init appnet agent

angelcar:
### Summary

* Change Service Connect internal model to reflect latest changes made by Appnet
* Inject `APPNET_LISTENER_PORT_MAPPING` with the mapping of listener name to ephemeral port

### Implementation details
We create a `map[string]uint16` to map listener names to ephemeral port, for listeners that required it. The map is then serialized to JSON and injected to the Service Connect container environment variable: `APPNET_LISTENER_PORT_MAPPING`

For example, let's say `listener_1` got ephemeral port `44444`, and `egress_listener` got port `55555`, then the Service Conenct container should see the following env var when it runs:

`APPNET_LISTENER_PORT_MAPPING="{\"listener_1\":44444, \"egress_listener\":55555}"`

### Testing

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -145,6 +145,8 @@ const (
 	disableIPv6SysctlKey = "net.ipv6.conf.all.disable_ipv6"
 	// sysctlValueOff specifies the value to use to turn off a sysctl setting.
 	sysctlValueOff = "0"
+
+	serviceConnectListenerPortMappingEnvVar = "APPNET_LISTENER_PORT_MAPPING"
 )
 
 // TaskOverrides are the overrides applied to a task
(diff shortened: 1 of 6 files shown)


agent/api/task/task.go
```diff
@@ -522,17 +524,29 @@ func (task *Task) initServiceConnectEphemeralPorts() error {
 	}
 
 	// Assign ephemeral ports
+	portMapping := make(map[string]uint16)
 	var curEphemeralIndex int
 	for i, ic := range task.ServiceConnectConfig.IngressConfig {
```
rajal-amzn, line 529: given that we are already parsing the ingress config above, we could store the listener names that need ephemeral ports and assign them here.

angelcar: This is what this code is doing. Not sure what the suggestion is?

rajal-amzn: Not really. We are parsing the ingress config and egress config and identifying which listener ports are not set at both the places.

angelcar: can you please indicate which 2 places you're referring to?

yinyic (approved):

agent/api/task/task.go
angelcar: The process is as follows.

1. Determine reserved ports by inspecting customer containers
2. Iterate over `SCConfig.IngressConfig` to identify which ones have `listenerPort ==0` in order to determine how many ephemeral ports need to be generated
3.  Generate ephemeral ports
4. Assign generated ephemeral ports back to `SCConfig` by iterating again on `SCConfig.IngressConfig`. At the same time, map `listenerName` -> ephemeral port in order to generate the `APPNET_LISTENER_PORT_MAPPING` env var.

yinyic (approved):

mythri-garaga (approved):

agent/api/task/task.go
rajal-amzn: Yes. I was referring to iterating on `SCConfig.IngressConfig` in - 1) determining number of ephemeral ports 2) assigning back the ephemeral ports.

angelcar: Discussed offline. I could create a `map[string]IngressConfig` to only loop once, but since the number of entries in `SC.IngressConfig` is like, 10, and this is a one-time processing, there are no performance concerns, so doing two passes is just simpler.

rajal-amzn: Fair enough.

rajal-amzn (approved):

PR #3192, opened 2022-04-28
Refactor to clearly identify pause container code paths for awsvpc mode

angelcar:
### Summary

So far, pause container only exists in awsvpc mode, and code paths where pause container is explicitly used assume this network mode.  This will no longer be the case with the introduction of Service Connect, since SC in bridge network mode will require a pause container per task container. 

This PR just reorganizes code in a way that it's clear which code paths are for awsvpc mode, and which ones are for SC-enabled tasks using bridge network mode. 

### Implementation details

This is a code reorganization which doesn't alter the existing logic, otherwise UTs, integ, and/or functional tests would fail if something fundamental is changing.

### Testing
UTs, Integ tests, functional tests.

N/A
### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -141,6 +141,9 @@ const (
 	// specifies awsvpc type mode for a task
 	AWSVPCNetworkMode = "awsvpc"
 
+	// specifies host type mode for a task
+	HostNetworkMode = "host"
+
 	// disableIPv6SysctlKey specifies the setting that controls whether ipv6 is disabled.
 	disableIPv6SysctlKey = "net.ipv6.conf.all.disable_ipv6"
 	// sysctlValueOff specifies the value to use to turn off a sysctl setting.
@@ -277,7 +280,9 @@ type Task struct {
 	// setIdOnce is used to set the value of this task's id only the first time GetID is invoked
 	setIdOnce sync.Once
 
-	ServiceConnectConfig *ServiceConnectConfig `json:"serviceConnectConfig,omitempty"`
+	ServiceConnectConfig *ServiceConnectConfig `json:"ServiceConnectConfig,omitempty"`
+
+	NetworkMode string `json:"NetworkMode,omitempty"`
 }
 
 // TaskFromACS translates ecsacs.Task to apitask.Task by first marshaling the received
@@ -332,6 +337,9 @@ func (task *Task) initializeVolumes(cfg *config.Config, dockerClient dockerapi.D
 func (task *Task) PostUnmarshalTask(cfg *config.Config,
 	credentialsManager credentials.Manager, resourceFields *taskresource.ResourceFields,
 	dockerClient dockerapi.DockerClient, ctx context.Context, options ...Option) error {
+
+	task.initNetworkMode()
+
 	// TODO, add rudimentary plugin support and call any plugins that want to
 	// hook into this
 	task.adjustForPlatform(cfg)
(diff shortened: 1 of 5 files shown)


agent/api/task/task.go
```diff
@@ -445,6 +453,34 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
+		switch c.GetNetworkModeFromHostConfig() {
+		case BridgeNetworkMode:
+			task.NetworkMode = BridgeNetworkMode
+			return
+		case HostNetworkMode:
+			task.NetworkMode = HostNetworkMode
+			return
+		case "":
+			// the absence of a network mode means the default docker network mode is used, which, for ECS it's bridge.
+			// for future proofing, let's keep inspecting the rest of the containers for this particular case.
```
yinyic, line 476: I'm not quite seeing why we want the loop. While it's possible that in the future the default mode could change or what an empty string implies could be different, I think we will still have the guarantee that all containers in the same task payload should have the same network mode (specified or not). If two containers return different values for network mode, or one is specified and the other is not, I would think something has gone wrong. 

I think we can define a `defaultNetworkMode` and return that in case of absence. 

On a separate note, might be a good idea to check if the network mode is valid. I know it's highly unlikely that control plane will send something invalid, but doesn't hurt to verify. Could be useful say when we are testing Agent with mock payload and it has a typo.

yinyic (commented):
Thanks for doing this Angel, and leaving the TODO notes :)

agent/api/task/task.go
angelcar: > I'm not quite seeing why we want the loop.

for example, internal containers have a blank `NetworkConfig`. I could have used `c.IsInternal()` -> `continue` but I think I feel this is safer, were we only declare `bridge` if all containers we know of have blank or empty `NetworkConfig`.

I'm talking to CP so that they send the actual value from task def to avoid this altogether. I think it will happen, so I guess this is ok for now?

yinyic: We are calling `initNetworkMode` first thing in `PostUnmarshalTask`, which is before we add any internal containers and it should be an unmodified list from Control Plane hence I thought they should have consistent network model. But yeah if we are going to get it directly I think that'll be ideal.

angelcar: > We are calling initNetworkMode first thing in PostUnmarshalTask
yeah, although I'm not sure how to test/enforce that in a UT. If someone injects something before this in the future, or even before PostUnmarshalTask, things might change.

yinyic: Makes sense

yinyic (approved):

agent/api/task/task.go
```diff
@@ -1773,10 +1826,18 @@ func (task *Task) shouldOverrideNetworkMode(container *apicontainer.Container, d
...
+		// If it's an internal container, set the network mode to none for awsvpc, set to bridge if task is using
+		// bridge mode and this is an SC-enabled task.
 		// Currently, internal containers are either for creating empty host
 		// volumes or for creating the 'pause' container. Both of these
 		// only need the network mode to be set to "none"
+		if container.Type == apicontainer.ContainerCNIPause {
+			if task.IsNetworkModeAWSVPC() {
+				return true, networkModeNone
+			} else if task.IsNetworkModeBridge() && task.IsServiceConnectEnabled() {
+				return true, BridgeNetworkMode
```
ubhattacharjya, line 1838: Why are we not setting it to networkModeNone as well?

angelcar: Because for bridge mode, the pause containers have to start in bridge mode, so that when the real customer containers have bridge network connectivity when they join its respective pause container

angelcar: Next PR (by Yinyi) will make this more clear. But this is by design.

ubhattacharjya: got it.

ubhattacharjya (approved):

PR #3195, opened 2022-05-02
Populate TaskMetric with service connect stats

mythri-garaga:
### Summary

This PR includes small refactor on where the service connect metrics are gathered to be sent to TACS `PublishMetrics` API request.

### Implementation details

1. Updated `GetServiceConnectStats()` to be private since this method is not being exported.
2. Updated `GetInstanceMetrics()` to use `getServiceConnectStats()` to and initialize TaskMetric with service connect metrics data.

### Testing

New tests cover the changes:  No

### Licensing

### DIFF
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -71,8 +71,6 @@ type Engine interface {
 	GetInstanceMetrics(includeServiceConnectStats bool) (*ecstcs.MetricsMetadata, []*ecstcs.TaskMetric, error)
 	ContainerDockerStats(taskARN string, containerID string) (*types.StatsJSON, *NetworkStatsPerSec, error)
 	GetTaskHealthMetrics() (*ecstcs.HealthMetadata, []*ecstcs.TaskHealth, error)
-	//TODO [SC]: Return service connect metrics from GetServiceConnects when the metrics are defined.
-	GetServiceConnectStats() error
 }
 
 // DockerStatsEngine is used to monitor docker container events and to report
@@ -367,7 +365,7 @@ func (engine *DockerStatsEngine) addContainerUnsafe(dockerID string) (*StatsCont
 		seelog.Debugf("Adding container to stats health check watch list, id: %s, task: %s", dockerID, task.Arn)
 	}
 
-	if errResolveContainer == nil {
+	if errResolveContainer == nil && task.IsServiceConnectEnabled() {
 		engine.addTaskToServiceConnectStatsUnsafe(task.Arn)
 	}
 
@@ -433,14 +431,19 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 	engine.lock.Lock()
 	defer engine.lock.Unlock()
 
+	if includeServiceConnectStats {
+		err := engine.getServiceConnectStats()
+		if err != nil {
+			seelog.Errorf("Error getting service connect metrics: %v", err)
+		}
+	}
+
 	for taskArn := range engine.tasksToContainers {
 		containerMetrics, err := engine.taskContainerMetricsUnsafe(taskArn)
 		if err != nil {
 			seelog.Debugf("Error getting container metrics for task: %s, err: %v", taskArn, err)
 			continue
 		}
-
-		// TODO [SC]: Call GetServiceConnectStats() and add it to taskMetrics with corresponding taskArn
 
 		if len(containerMetrics) == 0 {
 			seelog.Debugf("Empty containerMetrics for task, ignoring, task: %s", taskArn)
@@ -459,6 +462,15 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 			TaskDefinitionFamily:  &taskDef.family,
 			TaskDefinitionVersion: &taskDef.version,
 			ContainerMetrics:      containerMetrics,
+		}
+
+		if includeServiceConnectStats {
+			serviceConnectStats, ok := engine.taskToServiceConnectStats[taskArn]
+			if !ok {
+				seelog.Debugf("task '%s' is not registered to collect service connect metrics", taskArn)
+				continue
+			}
+			taskMetric.ServiceConnectMetricsWrapper = serviceConnectStats.GetStats()
 		}
 
 		taskMetrics = append(taskMetrics, taskMetric)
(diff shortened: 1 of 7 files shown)


agent/stats/engine.go
```diff
@@ -433,15 +431,20 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 	engine.lock.Lock()
 	defer engine.lock.Unlock()
 
+	if includeServiceConnectStats {
+		err := engine.getServiceConnectStats()
+		if err != nil {
+			seelog.Errorf("Error getting service connect metrics: %v", err)
```
angelcar, line 437: hmm, is this the best we can do with this error?

agent/stats/engine.go
```diff
@@ -461,6 +464,15 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 			ContainerMetrics:      containerMetrics,
 		}
 
+		if includeServiceConnectStats {
+			serviceConnectStats, ok := engine.taskToServiceConnectStats[taskArn]
+			if !ok {
+				seelog.Debugf("task '%s' is not registered to collect service connect metrics", taskArn)
```
angelcar, line 470: this looks odd. I wonder if this is an error?

mythri-garaga: hmmm since service connect might not be enabled for all the tasks, thought logging an error wouldn't make sense!

agent/stats/engine.go
mythri-garaga: We could have retries but I feel `getServiceConnectStats()` is not the right place for retries on error as it pulls service connect metrics for all the tasks in parallel. This could be a better place for the retry logic - https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/stats/service_connect_linux.go#L46

What are your thoughts? @angelcar

mythri-garaga:
Taking a look at failing functional tests..

agent/stats/engine.go
angelcar: sgtm

agent/stats/engine.go
angelcar: but don't we set `includeServiceConnectStats=true` only if `task.IsServiceConnectEnabled()`?

mythri-garaga: currently `includeServiceConnectStats` flag is only used to indicate that we need to publish metrics with service connect stats on every 3rd(or 60s) request -  https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/tcs/client/client.go#L183-L189

Meanwhile one improvement I can think of to reduce this debug logs here is -  if there are no tasks with service connect enabled from ACS tasks that Agent receives we can set `includeServiceConnectStats` to `false`.

angelcar: why not just pass `includeServiceConnectStats=true` when the counter is `3` AND `task.IsServiceConnectEnabled()`?

mythri-garaga: because `publishMetrics()` is for all the tasks the agent is running, not specific to just one task.

angelcar: I see. Although `taskToServiceConnectStats` only has SC-enabled tasks, right? So think this log is redundant since we already know there won't be/should not be SC-enabled tasks in this map.

https://github.com/aws/amazon-ecs-agent/pull/3195/files#diff-50b5b60cbd2aab1f7353cb44f862f4abd0befe3bec0f9ecff0e1844efe18fc36R368

agent/stats/engine.go
```diff
@@ -461,6 +464,15 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 			ContainerMetrics:      containerMetrics,
 		}
 
+		if includeServiceConnectStats {
+			serviceConnectStats, ok := engine.taskToServiceConnectStats[taskArn]
```
ubhattacharjya, line 468: Should we add the check to see if the returned service-connect stats has already been sent or not?

agent/stats/engine.go
mythri-garaga: hmmm so currently in `publishMertics()` function we are looping over the `taskArns` from `taskContainerMetricsUnsafe` which may/may not be SC-enabled to send the taskMetrics.
Further, if `includeServiceConnectStats` is true, then for given taskArn we check if it SC-enabled by trying to retrieve it from the `taskToServiceConnectStats` map, if found we found it indicates that SC is enabled for the given taskArn so we gather it's service connect stats else we skip collecting the SC stats and just log it.

Basically we are using `taskToServiceConnectStats` to check if task is SC-enabled or not.

agent/stats/engine.go
mythri-garaga: Added a TODO to keep track.

agent/stats/engine.go
mythri-garaga: Checking if service-connect stats has already been sent and setting it done once the SC stats are published needs to done in `tcs/client/client.go`. Will cover these changes in the next PR since `ServiceConnectStats` object is not accessible from tcs package currently

ubhattacharjya: Can we not add a warn log that there are "no new stats to be sent" from here and then set ServiceConnectWrapper for task metric to be nil ?

mythri-garaga: Synced offline, will update this in new PR.

ubhattacharjya (approved):

agent/stats/engine.go
yinyic: > Basically we are using taskToServiceConnectStats to check if task is SC-enabled or not.

It doesn't seem like we check if a task is SC-enabled before adding it to the map? Or am I looking at the wrong place?
https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/stats/engine.go#L370-L372

mythri-garaga: oh yea...I have updated to check in the this PR(line 368) - https://github.com/aws/amazon-ecs-agent/pull/3195/files#diff-50b5b60cbd2aab1f7353cb44f862f4abd0befe3bec0f9ecff0e1844efe18fc36R368

yinyic: Ohhh LOL I conveniently missed that :)

yinyic (approved):
Per offline discussion, one potential improvement is to only call `engine.addTaskToServiceConnectStatusUnsafe` for SC container. Currently we attempt the call for every container in the task, although they'd all be no-op because the task has already been added to the map by SC container. This is not blocking this PR though.

PR #3200, opened 2022-05-04
Service Connect - Start task workflow for bridge mode

yinyic:
### Summary

* Add pause containers for SC bridge mode task containers and set appropriate container ordering dependencies
* Populate container name->IP mapping env var which will be consumed by Service Connect container (Appnet Agent)
* Update logic for configuring docker host config port bindings for SC bridge mode containers
* Update logic for configuring docker container config exposed ports for SC bridge mode containers

### Implementation details

For a ServiceConnect-enabled bridge-mode task, we will add a pause container to every customer-defined container. We will also override SC container steady state to be `RESOURCES_PROVISIONED` so that it could configure the netns through CNI plugin (TBA in a future PR)
The container ordering for task start up should look like:
* ServiceConnectContainer.PULLED depends on ALL PauseContainer.RUNNING
* TaskContainer.CREATED depends on ServiceConnectContainer.RESOURCE_PROVISIONED
* TaskContainer.CREATED also on ServiceConnectContainer.Healthy
The container ordering for task stop should look like:
* ServiceConnectContainer.STOPPED depends on ALL TaskContainer.STOPPED
* PauseContainer.STOPPED depends on ServiceConnectConnectContainers.STOPPED

When creating Service Connect container, we also need to populate an env var `APPNET_CONTAINER_MAPPING` that includes mapping from container names to container IPs - this is later used by SC container to configure its listeners. 

In terms of port bindings and port exposing which are also configured as part of `CreateContainer`
1. For SC container (Appnet Agent)
 a. we will create port bindings and expose all ingress listeners because they will be taking traffic from the internet
 b. we will expose all egress listener ports so that that can be reached by other task containers, but we don't need to publish port binding as egress listener will not be taking traffic from outside

2. When creating pause containers, we will exposing/publishing the ports for the actual task containers - we will create port bindings for application container ports and expose both container and host ports (this is to stay consistent with current bridge mode behavior)

3. When creating the actual application containers (customer-defined), we will NOT publish port binding or expose ports for them since this is done by their associated pause container. Also Docker does not allow containers launched with container network mode to publish or expose ports.

### Testing

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -495,16 +495,29 @@ func (c *Container) String() string {
 
 // GetSteadyStateStatus returns the steady state status for the container. If
 // Container.steadyState is not initialized, the default steady state status
-// defined by `defaultContainerSteadyStateStatus` is returned. The 'pause'
+// defined by `defaultContainerSteadyStateStatus` is returned. In awsvpc, the 'pause'
 // container's steady state differs from that of other containers, as the
 // 'pause' container can reach its teady state once networking resources
 // have been provisioned for it, which is done in the `ContainerResourcesProvisioned`
-// state
+// state. In bridge mode, pause containers are currently used exclusively for
+// supporting service-connect tasks. Those pause containers will have steady state
+// status "ContainerRunning" as the actual network provisioning is done by ServiceConnect
+// container (aka Appnet agent)
 func (c *Container) GetSteadyStateStatus() apicontainerstatus.ContainerStatus {
 	if c.SteadyStateStatusUnsafe == nil {
 		return defaultContainerSteadyStateStatus
 	}
 	return *c.SteadyStateStatusUnsafe
+}
+
+// SetSteadyStateStatusUnsafe allows setting container steady state status after they
+// are initially created.
+// In bridge mode, this is used by overriding the ServiceConnect container steady
+// status to ContainerResourcesProvisioned because it comes with ACS task payload and will
+// get ContainerRunning by default during unmarshalling. We need ServiceConnect container
+// to provision network resources to support SC bridge mode
+func (c *Container) SetSteadyStateStatusUnsafe(steadyState apicontainerstatus.ContainerStatus) {
+	c.SteadyStateStatusUnsafe = &steadyState
 }
 
 // IsKnownSteadyState returns true if the `KnownState` of the container equals
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -150,6 +150,7 @@ const (
 	sysctlValueOff = "0"
 
 	serviceConnectListenerPortMappingEnvVar = "APPNET_LISTENER_PORT_MAPPING"
+	serviceConnectContainerMappingEnvVar    = "APPNET_CONTAINER_MAPPING"
 )
 
 // TaskOverrides are the overrides applied to a task
@@ -515,6 +516,13 @@ func (task *Task) initDummyServiceConnectConfig() {
 	}
 }
 
+// initServiceConnectContainerDependencies builds container dependency for regular task containers and SC container, such that
+// - during task start up, a regular container depends on SC container to become Healthy
+// - during task tear down, SC container depends on all regular containers to be stopped first
+// Additionally, for bridge mode task, we have the SC container configure netns for itself as well as for all other task
+// containers. The configuration is done when SC container transitions from RUNNING -> RESOURCES_PROVISIONED. Therefore,
+// - we override steady state status for SC container to be RESOURCES_PROVISIONED (the default is RUNNING)
+// - during task start up, a regular container will also depends on SC Container RESOURCE_PROVISIONED
 func (task *Task) initServiceConnectContainerDependencies() {
 	scContainer := task.GetServiceConnectContainer()
 
@@ -524,6 +532,13 @@ func (task *Task) initServiceConnectContainerDependencies() {
 		}
 		container.AddContainerDependency(scContainer.Name, ContainerOrderingHealthyCondition)
 		scContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
+		if task.IsNetworkModeBridge() {
+			container.BuildContainerDependency(scContainer.Name, apicontainerstatus.ContainerResourcesProvisioned, apicontainerstatus.ContainerCreated)
+		}
+	}
+
+	if task.IsNetworkModeBridge() {
+		scContainer.SetSteadyStateStatusUnsafe(apicontainerstatus.ContainerResourcesProvisioned)
 	}
 }
 
@@ -1374,10 +1389,9 @@ func (task *Task) IsNetworkModeHost() bool {
 func (task *Task) addNetworkResourceProvisioningDependency(cfg *config.Config) error {
 	if task.IsNetworkModeAWSVPC() {
 		return task.addNetworkResourceProvisioningDependencyAwsvpc(cfg)
+	} else if task.IsNetworkModeBridge() && task.IsServiceConnectEnabled() {
+		return task.addNetworkResourceProvisioningDependencyServiceConnectBridge(cfg)
 	}
-	//else if task.IsNetworkModeBridge() && task.IsServiceConnectEnabled() {
-	//	// TODO [SC]: This is where we would setup a pause container per task container for Service Connect only
-	//}
 	return nil
 }
 
@@ -1448,6 +1462,68 @@ func (task *Task) addNetworkResourceProvisioningDependencyAwsvpc(cfg *config.Con
 		}
 	}
 	return nil
+}
+
+// addNetworkResourceProvisioningDependencyServiceConnectBridge creates one pause container per task container
+// except for SC container, and add a dependency for SC container to wait for all pause container running.
+//
+// Unlike AWSVPC pause containers that use CNI plugin for configuring netns when they go from RUNNING ->
+// RESOURCES_PROVISIONED, Bridge mode pause containers will have steady state RUNNING. All netns configuration and
+// CNI invocation will be performed solely by SC container.
+func (task *Task) addNetworkResourceProvisioningDependencyServiceConnectBridge(cfg *config.Config) error {
+	scContainer := task.GetServiceConnectContainer()
+	for _, container := range task.Containers {
+		if container.IsInternal() || container == scContainer {
+			continue
+		}
+		pauseContainer := apicontainer.NewContainerWithSteadyState(apicontainerstatus.ContainerRunning)
+		pauseContainer.TransitionDependenciesMap = make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet)
+		// The pause container name is used internally by task engine but still needs to be unique for every task,
+		// hence we are appending the corresponding application container name (which must already be unique within the task)
+		pauseContainer.Name = fmt.Sprintf("%s-%s", NetworkPauseContainerName, container.Name)
+		pauseContainer.Image = fmt.Sprintf("%s:%s", cfg.PauseContainerImageName, cfg.PauseContainerTag)
+		pauseContainer.Essential = true
+		pauseContainer.Type = apicontainer.ContainerCNIPause
+
+		task.Containers = append(task.Containers, pauseContainer)
+		scContainer.BuildContainerDependency(pauseContainer.Name, apicontainerstatus.ContainerRunning, apicontainerstatus.ContainerCreated)
+		pauseContainer.BuildContainerDependency(scContainer.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
+	}
+	return nil
+}
+
+// getBridgeModePauseContainerForTaskContainer retrieves the associated pause container for a regular task
+// container (customer-defined) in a bridge-mode SC-enabled task.
+// For a container with name "abc", the pause container will always be named "~internal~ecs~pause-abc"
+func (task *Task) getBridgeModePauseContainerForTaskContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
+	// "~internal~ecs~pause-$TASK_CONTAINER_NAME"
+	pauseContainerName := fmt.Sprintf("%s-%s", NetworkPauseContainerName, container.Name)
+	pauseContainer, ok := task.ContainerByName(pauseContainerName)
+	if !ok {
+		return nil, fmt.Errorf("could not find pause container %s for task container %s", pauseContainerName, container.Name)
+	}
+	return pauseContainer, nil
+}
+
+// getBridgeModeTaskContainerForPauseContainer retrieves the associated customer container for a pause
+// container in a bridge-mode SC-enabled task.
+// For a container with name "abc", the pause container will always be named "~internal~ecs~pause-abc"
+func (task *Task) getBridgeModeTaskContainerForPauseContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
+	if container.Type != apicontainer.ContainerCNIPause {
+		return nil, fmt.Errorf("container %s is not a CNI pause container", container.Name)
+	}
+	// pause container name will be something like "~internal~ecs~pause-$TASK_CONTAINER_NAME"
+	// limit the result to 2 substrings as $$TASK_CONTAINER_NAME may also container '-'
+	stringSlice := strings.SplitN(container.Name, "-", 2)
+	if len(stringSlice) < 2 {
+		return nil, fmt.Errorf("SC bridge mode pause container %s does not conform to %s-$TASK_CONTAINER_NAME format", container.Name, NetworkPauseContainerName)
+	}
+	taskContainerName := stringSlice[1]
+	taskContainer, ok := task.ContainerByName(taskContainerName)
+	if !ok {
+		return nil, fmt.Errorf("could not find task container %s for pause container %s", taskContainerName, container.Name)
+	}
+	return taskContainer, nil
 }
 
 func (task *Task) addNamespaceSharingProvisioningDependency(cfg *config.Config) {
@@ -1614,11 +1690,17 @@ func (task *Task) dockerConfig(container *apicontainer.Container, apiVersion doc
 		entryPoint = *container.EntryPoint
 	}
 
+	var exposedPorts nat.PortSet
+	var err error
+	if exposedPorts, err = task.dockerExposedPorts(container); err != nil {
+		return nil, &apierrors.DockerClientConfigError{Msg: "error resolving docker exposed ports for container: " + err.Error()}
+	}
+
 	containerConfig := &dockercontainer.Config{
 		Image:        container.Image,
 		Cmd:          container.Command,
 		Entrypoint:   entryPoint,
-		ExposedPorts: task.dockerExposedPorts(container),
+		ExposedPorts: exposedPorts,
 		Env:          dockerEnv,
 	}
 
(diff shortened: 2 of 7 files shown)


agent/api/task/task.go
```diff
@@ -1882,6 +1987,46 @@ func (task *Task) shouldOverrideNetworkModeAwsvpc(container *apicontainer.Contai
...
+// with container network mode use pause container netns (the "docker run" equivalent option is
+// "--network container:$pause_container_id")
+func (task *Task) shouldOverrideNetworkModeServiceConnectBridge(container *apicontainer.Container, dockerContainerMap map[string]*apicontainer.DockerContainer) (bool, string) {
+	if container == task.GetServiceConnectContainer() {
+		return false, "" // SC container will use bridge mode (it doesn't need a pause container)
+	}
+	pauseContName, ok := task.ServiceConnectPauseContainerMapping[container.Name]
+	if !ok || pauseContName == "" {
+		// This should never be the case and implies a code-bug.
+		logger.Critical("Pause container required per task container for Service Connect task bridge mode, but " +
```
angelcar, line 2001: can we double check that `Critical` doesn't also terminate the program?

agent/api/task/task.go
```diff
@@ -1882,6 +1987,46 @@ func (task *Task) shouldOverrideNetworkModeAwsvpc(container *apicontainer.Contai
...
+	pauseContainer, ok := dockerContainerMap[pauseContName]
+	if !ok || pauseContainer == nil {
+		// This should never be the case and implies a code-bug.
+		logger.Critical("Pause container required per task container for Service Connect task bridge mode, but " +
+			"not found in docker container map for task container", logger.Fields{
+			field.TaskID:    task.GetID(),
+			field.Container: container.Name,
+		})
+		return false, ""
+	} else if pauseContainer.Container.Type != apicontainer.ContainerCNIPause {
```
angelcar, line 2018: nit:  no need for `else` since we're returning above.

agent/engine/docker_task_engine.go
```diff
@@ -1658,6 +1666,13 @@ func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task,
 		}
 	}
 
+	if task.IsServiceConnectEnabled() && task.IsNetworkModeBridge() && container == task.GetServiceConnectContainer() {
+		// TODO: CNI integration for SC bridge-mode task
```
angelcar, line 1670: nit: can we do `// TODO [SC]:...` so that it's easier to identify SC TODOs later?

agent/api/container/container.go
```diff
@@ -495,18 +495,32 @@ func (c *Container) String() string {
 
 // GetSteadyStateStatus returns the steady state status for the container. If
 // Container.steadyState is not initialized, the default steady state status
-// defined by `defaultContainerSteadyStateStatus` is returned. The 'pause'
+// defined by `defaultContainerSteadyStateStatus` is returned. In awsvpc, the 'pause'
 // container's steady state differs from that of other containers, as the
 // 'pause' container can reach its teady state once networking resources
 // have been provisioned for it, which is done in the `ContainerResourcesProvisioned`
-// state
+// state. In bridge mode, currently pause containers are currently used exclusively for
```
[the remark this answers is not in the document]
yinyic: Thank you will fix!

agent/api/container/container.go
```diff
@@ -495,18 +495,32 @@ func (c *Container) String() string {
...
 
+// SetSteadyStateStatusUnsafe allows setting container steady state status after they
+// are initially created.
+// In bridge mode, this is used by overriding the ServiceConnect container steady
+// status to ContainerResourcesProvisioned because it comes with ACS task payload and will
+// get ContainerRunning by default during unmarshalling. We need ServiceConnect container
+// to provision network resources to support SC bridge mode
+func (c *Container) SetSteadyStateStatusUnsafe(steadyState apicontainerstatus.ContainerStatus) {
+	steadyStateStatus := steadyState
+	c.SteadyStateStatusUnsafe = &steadyStateStatus
```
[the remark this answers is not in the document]
yinyic: Yeah you are right, will update!

agent/api/task/task.go
```diff
@@ -525,6 +531,20 @@ func (task *Task) initServiceConnectContainerDependencies() {
...
+	// containers. The configuration is done when SC container transitions from RUNNING -> RESOURCES_PROVISIONED.
+	// Therefore, each application container will also wait for scContainer.RESOURCE_PROVISIONED stage to complete
+	// before starting.
+	if task.IsNetworkModeBridge() {
+		scContainer.SetSteadyStateStatusUnsafe(apicontainerstatus.ContainerResourcesProvisioned)
+		for _, container := range task.Containers {
+			if container.IsInternal() || container == scContainer {
+				continue
+			}
+			container.BuildContainerDependency(scContainer.Name, apicontainerstatus.ContainerResourcesProvisioned, apicontainerstatus.ContainerPulled)
```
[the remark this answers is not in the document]
yinyic: I would agree. Only thing is that it seems we have been using `apicontainerstatus.ContainerPulled` for pause container ordering, and I couldn't find much context or reason, has there ever been any.  
I guess I'll make the change here because it'll also reduce task start-up time and I really dont see the risk. But I'm going to leave the existing ordering the way they are, if that's okay with you

agent/api/task/task.go
```diff
@@ -1450,6 +1469,34 @@ func (task *Task) addNetworkResourceProvisioningDependencyAwsvpc(cfg *config.Con
...
+		pauseContainer.TransitionDependenciesMap = make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet)
+		// The pause container name is used internally by task engine but still needs to be unique for every task,
+		// hence we are appending the corresponding application container name (which must already be unique within the task)
+		pauseContainer.Name = fmt.Sprintf("%s-%s", NetworkPauseContainerName, container.Name)
+		pauseContainer.Image = fmt.Sprintf("%s:%s", cfg.PauseContainerImageName, cfg.PauseContainerTag)
+		pauseContainer.Essential = true
+		pauseContainer.Type = apicontainer.ContainerCNIPause
+
+		task.Containers = append(task.Containers, pauseContainer)
+		task.ServiceConnectPauseContainerMapping[container.Name] = pauseContainer.Name
```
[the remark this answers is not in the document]
yinyic: That's a good callout! And I like the idea of using the convention. I'll get rid of the map.

agent/api/task/task.go
```diff
@@ -1644,10 +1691,50 @@ func (task *Task) dockerConfig(container *apicontainer.Container, apiVersion doc
 	return containerConfig, nil
 }
 
+// dockerExposedPorts returns the container ports that need to be exposed for a container
+// For bridge-mode ServiceConnect-enabled tasks:
+// 1. Pause containers need to exposes the container port(s) of their associated task container
```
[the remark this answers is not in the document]
yinyic: Oops sorry!

agent/api/task/task.go
```diff
@@ -1644,10 +1691,50 @@ func (task *Task) dockerConfig(container *apicontainer.Container, apiVersion doc
...
+			// expose egress listener port
+			ec := task.ServiceConnectConfig.EgressConfig
+			dockerPort := nat.Port(strconv.Itoa(int(ec.ListenerPort))) + "/tcp"
+			dockerExposedPorts[dockerPort] = struct{}{}
+			return dockerExposedPorts
+		} else {
+			// This is a customer application container which is launched with "--network container:$pause_container_id"
+			// In such case we don't expose any ports (docker won't allow anyway) because they are exposed by their
+			// pause container instead.
+			return nil
```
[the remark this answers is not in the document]
yinyic: I somehow remember getting an error for empty map saying it's conflicting with the network:container mode, but yeah now that you mention if it works for awsvpc it must have been okay.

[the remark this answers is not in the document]
yinyic: Verified that the docker error is own thrown for non-empty map https://github.com/moby/moby/blob/bca8d9f2ce0d63e1490692917cde6273bc288bad/runconfig/hostconfig.go#L64

so yeah I'll fix this

agent/api/task/task.go
yinyic: We use seelog which will return an error for `Error`, `Warn`, and `Critical`, but no it does not terminate the program

https://github.com/aws/amazon-ecs-agent/blob/master/agent/vendor/github.com/cihub/seelog/log.go

agent/api/task/task.go
```diff
@@ -1450,6 +1464,67 @@ func (task *Task) addNetworkResourceProvisioningDependencyAwsvpc(cfg *config.Con
...
+	if !ok {
+		return nil, fmt.Errorf("could not find pause container %s for task container %s", pauseContainerName, container.Name)
+	}
+	return pauseContainer, nil
+}
+
+// getServiceConnectBridgeModeTaskContainerForPauseContainer retrieves the associated customer container for a pause
+// container in a bridge-mode SC-enabled task.
+// For a container with name "abc", the pause container will always be named "~internal~ecs~pause-abc"
+func (task *Task) getServiceConnectBridgeModeTaskContainerForPauseContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
```
yinyic, line 1511: Woops I realized it doesn't handle the case where the original container name already has '-'. Will fix and add unit test.

agent/engine/docker_task_engine_linux_test.go
```diff
@@ -1057,3 +1057,219 @@ func TestPauseContainerWitServiceConnect(t *testing.T) {
...
+	dockerClient.EXPECT().DescribeContainer(gomock.Any(), pauseContainerID).AnyTimes()
+
+	err = taskEngine.Init(ctx)
+	assert.NoError(t, err)
+	taskEngine.AddTask(sleepTask)
+	stateChangeEvents := taskEngine.StateChangeEvents()
+	verifyTaskIsRunning(stateChangeEvents, sleepTask)
+
+	mockTime.EXPECT().After(gomock.Any()).Return(cleanup).MinTimes(1)
+	gomock.InOrder(
```
angelcar, line 1189: can we also verify here that the task container is stopped first?

agent/engine/docker_task_engine_linux_test.go
```diff
@@ -1057,3 +1057,219 @@ func TestPauseContainerWitServiceConnect(t *testing.T) {
...
+	// verify host config port bindings
+	assert.NotNil(t, hostConfig.PortBindings)
+	assert.Equal(t, 1, len(hostConfig.PortBindings))
+	bindings, ok := hostConfig.PortBindings["15000/tcp"]
+	assert.True(t, ok)
+	assert.Equal(t, 1, len(bindings))
+	assert.Equal(t, 0, bindings[0].HostPort)
+	// verify container config port exposed
+	assert.NotNil(t, config.ExposedPorts)
+	assert.Equal(t, 2, len(config.ExposedPorts)) // 2 because egress container port is also exposed
```
angelcar, line 1263: I believe there's no need for egress port to be exposed. In fact, it might not be desirable.

agent/api/task/task.go
```diff
@@ -525,6 +526,20 @@ func (task *Task) initServiceConnectContainerDependencies() {
 		container.AddContainerDependency(scContainer.Name, ContainerOrderingHealthyCondition)
 		scContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
 	}
+
+	// For bridge mode, we are going to have the SC container configure netns for itself as well as for all other task
+	// containers. The configuration is done when SC container transitions from RUNNING -> RESOURCES_PROVISIONED.
+	// Therefore, each application container will also wait for scContainer.RESOURCE_PROVISIONED stage to complete
+	// before they can be created (i.e. application container will pause at the previous stage which is PULLED.
+	if task.IsNetworkModeBridge() {
```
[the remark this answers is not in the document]
yinyic: Sure. The reason why I had a separate block if block was because we also need to override steady status for SC container so I thought to group them together for readability. I don't have strong preference though so yeah I'll get rid of two loops.

agent/api/task/task.go
```diff
@@ -1644,14 +1725,49 @@ func (task *Task) dockerConfig(container *apicontainer.Container, apiVersion doc
 	return containerConfig, nil
 }
 
-func (task *Task) dockerExposedPorts(container *apicontainer.Container) nat.PortSet {
-	dockerExposedPorts := make(map[nat.Port]struct{})
+// dockerExposedPorts returns the container ports that need to be exposed for a container
+// For bridge-mode ServiceConnect-enabled tasks:
+// 1. Pause containers need to exposes the port(s) of their associated task container
```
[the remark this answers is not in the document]
yinyic: Sorry:p

agent/api/task/task.go
```diff
@@ -1882,6 +2024,37 @@ func (task *Task) shouldOverrideNetworkModeAwsvpc(container *apicontainer.Contai
...
 
+// shouldOverrideNetworkModeServiceConnectBridge checks if a bridge-mode SC task container needs network mode override
+// For non-internal containers in an SC bridge-mode task, each gets a pause container, and should be launched
+// with container network mode use pause container netns (the "docker run" equivalent option is
+// "--network container:$pause_container_id")
+func (task *Task) shouldOverrideNetworkModeServiceConnectBridge(container *apicontainer.Container, dockerContainerMap map[string]*apicontainer.DockerContainer) (bool, string) {
+	if container == task.GetServiceConnectContainer() {
+		return false, "" // SC container will use bridge mode (it doesn't need a pause container)
+	}
+	pauseContainer, err := task.getServiceConnectBridgeModePauseContainerForTaskContainer(container)
```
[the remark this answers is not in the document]
yinyic: @rajal-amzn I used a map in the first commit but later changed it because Angel brought up a good point that if we keep a map then this map will need to configure it be to stored in Agent database if Agent restarts, so we thought if we could simply use the naming convention for resolving either pause or task container with the other it'll be more convenient.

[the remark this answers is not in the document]
yinyic: I'm updating the method names, sorry for the readability issue

agent/api/task/task_test.go
```diff
@@ -186,6 +188,153 @@ func TestDockerHostConfigPortBinding(t *testing.T) {
...
+}
+
+func convertSCPort(port uint16) nat.Port {
+	return nat.Port(strconv.Itoa(int(port)) + defaultSCProtocol)
+}
+
+// TestDockerHostConfigServiceConnectBridgeMode verifies port bindings and network mode overrides for each
+// container in an SC-enabled bridge mode task. The test task is consisted of the SC container, a regular container,
+// and the pause container associated with it.
+func TestDockerHostConfigServiceConnectBridgeMode(t *testing.T) {
```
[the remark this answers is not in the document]
yinyic: Sure will add!

agent/engine/docker_task_engine_linux_test.go
yinyic: Exposing the container port does not mean it will get a host port mapping. In this case we are not creating port mapping for the egress listener so it won't get a host port. Exposing the container port is so that it could be accessed by other docker containers on the host.

agent/api/task/task.go
```diff
@@ -525,6 +526,20 @@ func (task *Task) initServiceConnectContainerDependencies() {
 		container.AddContainerDependency(scContainer.Name, ContainerOrderingHealthyCondition)
 		scContainer.BuildContainerDependency(container.Name, apicontainerstatus.ContainerStopped, apicontainerstatus.ContainerStopped)
 	}
+
+	// For bridge mode, we are going to have the SC container configure netns for itself as well as for all other task
+	// containers. The configuration is done when SC container transitions from RUNNING -> RESOURCES_PROVISIONED.
+	// Therefore, each application container will also wait for scContainer.RESOURCE_PROVISIONED stage to complete
+	// before they can be created (i.e. application container will pause at the previous stage which is PULLED.
+	if task.IsNetworkModeBridge() {
+		scContainer.SetSteadyStateStatusUnsafe(apicontainerstatus.ContainerResourcesProvisioned)
```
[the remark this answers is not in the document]
yinyic: The `initDependencies` is invoked before any task container (including pause container) is created. We will be invoking CNI plugin in `docker_task_engine.go` -> `provisionContainerResources` function, which is after the container has been started.

agent/api/task/task.go
```diff
@@ -525,6 +526,20 @@ func (task *Task) initServiceConnectContainerDependencies() {
...
+		for _, container := range task.Containers {
+			if container.IsInternal() || container == scContainer {
```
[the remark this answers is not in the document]
yinyic: The pause containers in our case.

agent/engine/docker_task_engine_linux_test.go
yinyic: In this commit we simulate a docker container stop event for task container in order to start the task stop workflow, which means there isn't actually going to be a `StopContainer` call for it. I've updated the test in the newer commit to update task desired status to `STOPPED` instead so that we can verify what you mentioned.

angelcar (approved):

angelcar (approved):

mythri-garaga (dismissed):

angelcar (approved):

mythri-garaga (dismissed):

agent/api/task/task.go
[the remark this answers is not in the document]
rajal-amzn: Would there be other containers in the future that would be marked as internal? Could we have some other variable to indicate this?

agent/api/task/task.go
```diff
@@ -1450,6 +1464,68 @@ func (task *Task) addNetworkResourceProvisioningDependencyAwsvpc(cfg *config.Con
 	return nil
 }
 
+// addNetworkResourceProvisioningDependencyServiceConnectBridge creates one pause container per task container
+// except for SC container, and add a dependency for SC container to wait for all pause container running.
+//
+// Unlike AWSVPC pause containers that use CNI plugin for configuring netns when they go from RUNNING ->
+// RESOURCES_PROVISIONED, Bridge mode pause containers will have steady state RUNNING. All netns configuration and
+// CNI invocation will be performed solely by SC container.
+func (task *Task) addNetworkResourceProvisioningDependencyServiceConnectBridge(cfg *config.Config) error {
```
rajal-amzn, line 1473: There is lot of logic in here which is applicable only for Service Connect usecase. Could we move all the SC Logic to a new file? I believe that it would help in better structure and readability.

agent/api/task/task.go
```diff
@@ -2973,3 +3153,27 @@ func (task *Task) GetServiceConnectContainer() *apicontainer.Container {
 func (task *Task) IsServiceConnectEnabled() bool {
 	return task.GetServiceConnectContainer() != nil
 }
+
+// PopulateServiceConnectContainerMappingEnvVar populates APPNET_CONTAINER_MAPPING env var for AppNet Agent container
+// aka SC container
+func (task *Task) PopulateServiceConnectContainerMappingEnvVar() error {
```
rajal-amzn, line 3159: Do we have the logic for adding `LISTENER_PORT_MAPPING`? If not, that also needs to go here.

agent/api/task/task.go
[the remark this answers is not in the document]
yinyic: > Would there be other containers in the future that would be marked as internal?

There could be. But I believe the goal here is to configure ordering for any customer-defined container, which will never be marked as internal, so I think this check is safe.

agent/api/task/task.go
yinyic: Yeah! This [PR](https://github.com/aws/amazon-ecs-agent/pull/3211) starts the refactoring. I'll wait for it to be merged, and do a refactoring for my PR as well :)

agent/api/task/task.go
yinyic: We do, it's added at the end of ephemeral port selection logic [here](https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/api/task/task.go#L578)

rajal-amzn: Thanks. It would be good if we could move all the logic of adding Env variables into a single place.

yinyic: Hmm this may not be a straightforward change. 

So below events need to happen in the exact order
1. select ephemeral ports
2. create pause containers
3. populate `APPNET_CONTAINER_MAPPING`
4. create Appnet container 

`APPNET_LISTENER_PORT_MAPPING` is currently populated at step 1. While we technically can move it to step 3, it would be hard to tell which ports are originally specified in task def, vs. selected by ECS Agent (that's why we populate it the same time as we select the ephemeral ports).

rajal-amzn: Could we not store this in a map? I would prefer to have all Env variables in a single place. But we can take this up after the first refactor.

mythri-garaga (approved):

agent/api/container/container.go
```diff
@@ -495,18 +495,31 @@ func (c *Container) String() string {
 
 // GetSteadyStateStatus returns the steady state status for the container. If
 // Container.steadyState is not initialized, the default steady state status
-// defined by `defaultContainerSteadyStateStatus` is returned. The 'pause'
+// defined by `defaultContainerSteadyStateStatus` is returned. In awsvpc, the 'pause'
 // container's steady state differs from that of other containers, as the
 // 'pause' container can reach its teady state once networking resources
```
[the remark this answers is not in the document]
yinyic: Indeed :) Thanks for catching that!

angelcar (approved):

agent/api/container/container.go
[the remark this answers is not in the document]
angelcar: May I suggest to make a note and fix the typo in the next PR? Don't want to have to run all the functional tests again :)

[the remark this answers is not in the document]
yinyic: Yeah was about to propose the same. Will include that in the follow-up refactoring PR.

agent/api/task/task.go
rajal-amzn: Agreed. We do store the newly assigned ephemeral ports as part of the container definition but we need to do mapping between the listener name from SC attachment and the containers to generate listener port mapping. I was suggesting that we store this mapping given we are processing the attachments in the first step. We could either do that or do the mapping again at 3rd step.

angelcar: The mapping already exist in `task.ServiceConnectConfig`. The env vars are just a transformation of this mapping. So this code looks good to me.

rajal-amzn: Right now, we are adding 2 Env variables from Agent and we could add more in the future. So, it would be good if we could put those in a single place since these are the contracts/points of integration

aws-gibbskt: Lets ensure we revisit this in the refactor. TaskEngine / ServiceConnect should be a consolidation point. IF we want to add a Task / ServiceConnect package/file to consolidate on other things, that is fine. The overarching issue I see is that we have a LOT of SC specific code sprinkled in a must more generic place apitask.Task which is not a good separation of concerns.

aws-gibbskt (approved):

agent/api/task/task.go
yinyic: We are populating `APPNET_LISTENER_PORT_MAPPING` right after ephemeral port selection because after that we will lose the knowledge of which ports Agent actually selected, vs. pre-populated in task def. If we are to create `APPNET_LISTENER_PORT_MAPPING` the same time as `APPNET_CONTAINER_MAPPING`, we'll have to use another map for preserving the listener -> ephemeral port information. Unless, it's okay to put in the `APPNET_LISTENER_PORT_MAPPING` listeners with ephemeral ports and pre-defined ports, because then ECS Agent won't need to create the extra map for differentiating the listeners that got ephemeral ports.

rajal-amzn: Yes. There should be no problem in passing in all the ports. We can handle that in EMS.

yinyic: @aws-gibbskt Yeah, have a TODO in this PR to move the function to SC manager
```
func (m *manager) initServiceConnectContainerMapping(task *apitask.Task, container *apicontainer.Container, hostConfig *dockercontainer.HostConfig) error {
	// TODO [SC] - Move the function here
	return task.PopulateServiceConnectContainerMappingEnvVar()
}
```
The refactoring PR I'm working on will extract away not only `PopulateServiceConnectContainerMappingEnvVar`, but also other SC-specific task handling from `task.go` and `docker_task_engine.go` which should contain only common logic. That PR should hopefully address your concern.

rajal-amzn (approved):

agent/api/task/task.go
yinyic: @rajal-amzn Thanks for clarifying! Okay I'll move the two env var population into one place in the refactoring PR.

PR #3208, opened 2022-05-09
Extract service connect config value from ACS

chienhanlin:
### Summary
This PR 
* Update ACS api-json and api.go model to include `NetworkMode` and `Attachment`. `Attachment` contains service connect configuration value.
* Update service_connect.go to use a pointer for EgressConfig
* Add a service connect attachment parser to extract service connect configuration value from the service connect attachment.
* Add tests to test `ParseServiceConnectAttachment()` in service_connect_attachment_parser.go and `ValidateServiceConnectConfig()` in service_connect_validator.go

### Implementation details
1. agent/acs/model/api/api-2.json
  *  Update the documentation for AttachInstanceNetworkInterfaces 
  *  Update the documentation for Heartbeat
  *  Add `Attachment`, `AttachmentList`, `AttachmentProperty`, and `AttachmentPropertyList` struct to support service connect
  * Add `attachments` and `networkMode` for `Task` to support service connect
2. agent/acs/model/ecsacs/api.go
 *  Update ACS api model by running `make gogenerate` with the updated agent/acs/model/api/api-2.json
 * Add `Attachment`, `AttachmentProperty` and `NetworkMode` to ACS api model
3. agent/api/task/task.go
 * Find the service connect type attachment from acsTask.Attachments when the `AttachmentType` equals to `ServiceConnect`
 * Assign `task.ServiceConnectConfig` from results returns from `ParseServiceConnectAttachment()` 
4. agent/api/task/service_connect.go
 * Update `EgressConfig` to `*`EgressConfig
 * Append `omitempty` to `IngressConfig`, `EgressConfig` and `DNSConfig` as they can be empty
5. agent/api/task/service_connect_attachment_parser.go 
* Add `ParseServiceConnectAttachment()` to parse service connect configuration value and service connect container name from the passed *ecsacs.Attachment
6. agent/api/task/service_connect_attachment_parser_test.go
* Add `TestParseServiceConnectAttachment()` with positive test cases to test ParseServiceConnectAttachment()
* Add TestParseServiceConnectAttachmentWithError()` with negative test cases to test ParseServiceConnectAttachment()
7. agent/api/task/service_connect_validator.go
* Add `ValidateServiceConnectConfig()` to validate the parsed service connect config
8. agent/api/task/service_connect_validator_test.go
* Add `TestValidateServiceConnectConfig()` with positive test cases to test ValidateServiceConnectConfig()
* Add `TestValidateServiceConnectConfigWithWarning()` with warning to test ValidateServiceConnectConfig()
* Add  `TestValidateServiceConnectConfigWithEmptyConfig()` with valid empty sc config test cases to test ValidateServiceConnectConfig()
*  Add `TestValidateServiceConnectConfigWithError()` with negative test cases to test ValidateServiceConnectConfig()
* Add `TestValidateServiceConnectConfigWithPortCollision()` with negative test cases to test ValidateServiceConnectConfig()
9. agent/api/task/task.go
* Add `getServiceConnectConfig()` to extract service connect config when the service connect attachment is found
* Append `task.ServiceConnectConfig.EgressConfig != nil` to `initServiceConnectEphemeralPorts()` and `dockerExposedPorts` after changing  service connect EgressConfig to a pointer.
10. agent/api/task/task_test.go
* Update testTask with after changing  service connect EgressConfig to a pointer. 
* Add `TestTaskServiceConnectAttachment()` with positive test cases to test extracting service connect config from the matching attachment logic in task.go
* Add `TestTaskWithoutServiceConnectAttachment()` without service connect attachment to test TaskFromACS will not return error
11. agent/engine/docker_task_engine_linux_test.go
* Update EgressConfig in `TestContainersWithServiceConnect_BridgeMode()` after changing  service connect EgressConfig to a pointer.   
  
### Testing
Ran `make test`
* TestParseServiceConnectAttachment
```
--- PASS: TestParseServiceConnectAttachment (0.00s)
    --- PASS: TestParseServiceConnectAttachment/AWSVPC_default_case (0.00s)
    --- PASS: TestParseServiceConnectAttachment/AWSVPC_default_case_with_IPv6_enabled (0.00s)
    --- PASS: TestParseServiceConnectAttachment/AWSVPC_override_case (0.00s)
    --- PASS: TestParseServiceConnectAttachment/AWSVPC_override_case_with_IPv6_enabled (0.00s)
    --- PASS: TestParseServiceConnectAttachment/Bridge_default_case (0.00s)
    --- PASS: TestParseServiceConnectAttachment/Bridge_default_case_with_no_ingress_config (0.00s)
    --- PASS: TestParseServiceConnectAttachment/Bridge_default_case_with_no_egress_config_and_dns_config (0.00s)
```
* TestParseServiceConnectAttachmentWithError
```
--- PASS: TestParseServiceConnectAttachmentWithError (0.00s)
    --- PASS: TestParseServiceConnectAttachmentWithError/AWSVPC_default_case_with_the_invalid_attachment_property_value (0.00s)
```
* TestValidateServiceConnectConfig
```
--- PASS: TestValidateServiceConnectConfig (0.00s)
    --- PASS: TestValidateServiceConnectConfig/AWSVPC_default_case (0.00s)
    --- PASS: TestValidateServiceConnectConfig/AWSVPC_default_case_with_IPv6_enabled (0.00s)
    --- PASS: TestValidateServiceConnectConfig/AWSVPC_override_case (0.00s)
    --- PASS: TestValidateServiceConnectConfig/AWSVPC_override_case_with_IPv6_enabled (0.00s)
    --- PASS: TestValidateServiceConnectConfig/Bridge_default_case (0.00s)
    --- PASS: TestValidateServiceConnectConfig/Bridge_default_case_with_IPv6_enabled (0.00s)
    --- PASS: TestValidateServiceConnectConfig/Bridge_override_case (0.00s)
    --- PASS: TestValidateServiceConnectConfig/Bridge_override_case_with_IPv6_enabled (0.00s)
```   
* TestValidateServiceConnectConfigWithWarning
```
--- PASS: TestValidateServiceConnectConfigWithWarning (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithWarning/AWSVPC_default_case_with_both_intercept_port_and_host_port_in_the_ingress_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithWarning/Bridge_override_case_with_both_host_port_and_intercept_port_in_the_ingress_config (0.00s)
```
* TestValidateServiceConnectConfigWithEmptyConfig
```
--- PASS: TestValidateServiceConnectConfigWithEmptyConfig (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithEmptyConfig/AWSVPC_default_case_with_the_empty_egress_config_and_dns_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithEmptyConfig/AWSVPC_default_case_with_the_empty_ingress_config (0.00s)
```
* TestValidateServiceConnectConfigWithError
```
--- PASS: TestValidateServiceConnectConfigWithError (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/AWSVPC_default_case_with_no_service_connect_container_name (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/AWSVPC_default_case_with_the_service_connect_container_name_not_exists_in_the_task (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/AWSVPC_default_case_with_no_listener_name_in_the_egress_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/AWSVPC_default_case_with_no_IPv4_CIDR_in_the_egress_config_when_Ipv6_is_not_enabled (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/AWSVPC_default_case_with_no_IPv6_CIDR_in_the_egress_config_when_IPv6_is_enabled (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/AWSVPC_override_case_with_the_invalid_IPv4_CIDR_in_the_egress_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/AWSVPC_override_case_the_invalid_IPv6_CIDR_in_the_egress_config_when_IPv6_is_enabled (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/Bridge_default_case_with_the_egress_config_but_no_dns_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithError/Bridge_override_case_with_no_the_invalid_address_in_dns_config_when_IPv6_is_enabled (0.00s)
```
* TestValidateServiceConnectConfigWithPortCollision
```
--- PASS: TestValidateServiceConnectConfigWithPortCollision (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithPortCollision/AWSVPC_default_case_with_the_intercept_port_collision_in_the_ingress_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithPortCollision/AWSVPC_override_case_with_the_listener_port_collision_in_the_ingress_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithPortCollision/Bridge_default_case_with_the_listener_port_collision_in_the_ingress_config (0.00s)
    --- PASS: TestValidateServiceConnectConfigWithPortCollision/Bridge_override_case_with_the_host_port_collision_in_the_ingress_config (0.00s)
```
* TestTaskServiceConnectAttachment
```
--- PASS: TestTaskServiceConnectAttachment (0.00s)
    --- PASS: TestTaskServiceConnectAttachment/Bridge_default_case (0.00s)
    --- PASS: TestTaskServiceConnectAttachment/AWSVPC_override_case_with_IPv6_enabled (0.00s)
```
* TestTaskWithoutServiceConnectAttachment
```
--- PASS: TestTaskWithoutServiceConnectAttachment (0.00s)
```
Update ACS model with an attachment field which contains service connect configuration value.

### Licensing


agent/api/task/service_connect_test.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+	case bridgeMode:
+		if override {
+			ingressConfig = fmt.Sprintf(`\"listenerPort\":%s,\"hostPort\":%s`, testBridgePortOverride, testBridgeHostPort)
+		} else {
+			ingressConfig = fmt.Sprintf(`\"listenerPort\":%s`, testBridgePortDefault)
+		}
+	}
+
+	testAWSVPCServiceConnectConfig := fmt.Sprintf(`"{\"statsConfig\":{\"path\":\"%s\"},\"egressConfig\":{\"listenerName\":\"%s\",\"vip\":{\"ipv4Cidr\":\"%s\",\"ipv6Cidr\":\"%s\"}},\"dnsConfig\":[{\"hostname\":\"%s\",\"address\":\"%s\"}],\"ingressConfig\":[{%s}]}"`, testStatsConfigPath, testOutboundListenerName, testIPV4Cidr, testIPV6Cidr, testHostName, testAddress, ingressConfig)
```
angelcar, line 62: nit: could we use a multi-line string here? kinda painful to read like this :)

agent/api/task/service_connect.go
```diff
@@ -65,3 +80,55 @@ type DNSConfigEntry struct {
...
+func ValidateSCConfig(scConfig *ServiceConnectConfig) error {
+	if scConfig.ContainerName == "" {
+		return errors.Errorf("service connect config validation: no service connect ContainerName")
+	}
+
+	if len(scConfig.IngressConfig) == 0 {
+		return errors.Errorf("service connect config validation: no service connect IngressConfig")
+	}
+
+	if len(scConfig.DNSConfig) == 0 {
```
angelcar, line 130: I believe this is allowed to be empty. It would be weird, but I seems ok to have a container that won't do outbound SC calls, but only receive inbound ones.

Also, I think we need to validate more in depth here (this goes for all other options), for example, are IPs valid, port numbers, etc.

agent/api/task/service_connect_test.go
```diff
@@ -0,0 +1,261 @@
...
+				Address:  scAddress,
+			},
+		},
+	}
+}
+
+func TestExtractSCConfigFromACSAttachment(t *testing.T) {
+	tt := []struct {
+		testName      string
+		ingressConfig []IngressConfigEntry
```
angelcar, line 110: nit: can we rename this to `ExpectedIngressConfig`

agent/api/task/service_connect_test.go
```diff
@@ -0,0 +1,261 @@
...
+					HostPort:     aws.Uint16(8080),
+				},
+			},
+			testACSTask: getTestACSTaskAttachment(bridgeMode, true),
+		},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.testName, func(t *testing.T) {
+			testSCConfig := getTestServiceConnectConfig(testHostName, testAddress, testOutboundListenerName, testIPV4Cidr, testIPV6Cidr, testStatsConfigPath, tc.ingressConfig)
```
angelcar, line 155: another nit: let's change this to `expectedTestSCConfig`

agent/api/task/service_connect_test.go
```diff
@@ -0,0 +1,261 @@
...
+		{
+			testName:     "AWSVPC default case with no StatsConfig",
+			testSCConfig: fmt.Sprintf(`"{\"egressConfig\":{\"listenerName\":\"%s\",\"vip\":{\"ipv4Cidr\":\"%s\",\"ipv6Cidr\":\"%s\"}},\"dnsConfig\":[{\"hostname\":\"%s\",\"address\":\"%s\"}],\"ingressConfig\":[{%s}]}"`, testOutboundListenerName, testIPV4Cidr, testIPV6Cidr, testHostName, testAddress, testIngressConfig),
+		},
+		{
+			testName:     "AWSVPC default case with no IngressConfig",
+			testSCConfig: fmt.Sprintf(`"{\"statsConfig\":{\"path\":\"%s\"},\"egressConfig\":{\"listenerName\":\"%s\",\"vip\":{\"ipv4Cidr\":\"%s\",\"ipv6Cidr\":\"%s\"}},\"dnsConfig\":[{\"hostname\":\"%s\",\"address\":\"%s\"}]}"`, testStatsConfigPath, testOutboundListenerName, testIPV4Cidr, testIPV6Cidr, testHostName, testAddress),
+		},
+		{
+			testName:     "AWSVPC default case with no EgressConfig",
```
angelcar, line 178: wonder why this fails currently if there's no validation for egress config?

agent/api/task/service_connect.go
```diff
@@ -65,3 +80,55 @@ type DNSConfigEntry struct {
...
+		switch aws.StringValue(property.Name) {
+		case serviceConnectAttProSCConfigName:
+			s, _ := strconv.Unquote(string(*property.Value))
+			if err := json.Unmarshal([]byte(s), scConfigValue); err != nil {
+				return nil, err
+			}
+		case serviceConnectAttProSCContainerName:
+			containerName = aws.StringValue(property.Value)
+		}
+	}
```
yinyic, line 113: nit - maybe add a default case to report error on unrecognized properties

agent/api/task/service_connect.go
```diff
@@ -65,3 +80,55 @@ type DNSConfigEntry struct {
...
+	return scConfigValue, nil
+}
+
+func ValidateSCConfig(scConfig *ServiceConnectConfig) error {
+	if scConfig.ContainerName == "" {
+		return errors.Errorf("service connect config validation: no service connect ContainerName")
+	}
+
+	if len(scConfig.IngressConfig) == 0 {
```
yinyic, line 126: I wonder if we want to validate if each `IngressConfigEntry` contains the necessary fields. For example, if task has `awsvpc` network mode, then every `IngressConfigEntry` should have 
- either `interceptPort` + `listenerName` (for default experience)
- or `listenerPort` (for non-default experience)

agent/api/task/task.go
```diff
@@ -302,7 +302,13 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
 		task.StopSequenceNumber = *envelope.SeqNum
 	}
 
-	// Overrides the container command if it's set
+	// get service connect config value from ACS attachment
+	scConfig, err := ExtractSCConfigFromACSAttachment(acsTask)
+	if err != nil {
+		return nil, err
```
yinyic, line 308: Can we wrap the error message to indicate that we have failed to extract SC config? Otherwise it may be hard to relate the failure to the actual error.

agent/api/task/service_connect.go
```diff
@@ -13,6 +13,21 @@
...
+
+	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/pkg/errors"
+)
+
+const (
+	serviceConnectAttType               = "ServiceConnect"
+	serviceConnectAttProSCConfigName    = "ServiceConnectConfig"
+	serviceConnectAttProSCContainerName = "ContainerName"
```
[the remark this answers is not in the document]
rajal-amzn: Instead of putting these as constants, could we not read it from the model pulled from AppNet?

agent/api/task/service_connect.go
```diff
@@ -65,3 +80,55 @@ type DNSConfigEntry struct {
...
+// ExtractSCConfigFromACSAttachment finds the service connect attachment, and extracts the service connect container name and service connect config value.
+func ExtractSCConfigFromACSAttachment(acsTask *ecsacs.Task) (*ServiceConnectConfig, error) {
+	if acsTask.Attachments == nil {
+		return nil, nil
+	}
+	var scAttachment *ecsacs.Attachment
+	containerName := ""
+	for _, attachment := range acsTask.Attachments {
+		if aws.StringValue(attachment.AttachmentType) == serviceConnectAttType {
+			scAttachment = attachment
```
rajal-amzn, line 93: Shouldnt this be done at the task level where we iterate on all attachments and dependin on the type of attachment, we parse each of them? So, we will have `parseServiceConnectAttachment` within `service_connect.go` being called from `task.go`?

agent/api/task/service_connect.go
rajal-amzn: As mentioned offline, we can have services that are client only and they will not have any ingress config. This is valid.

agent/api/task/service_connect.go
rajal-amzn: +1. We also need to validate Egress config

agent/api/task/task.go
```diff
@@ -302,7 +302,13 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
...
+	}
+	task.ServiceConnectConfig = scConfig
```
rajal-amzn, line 310: This is outside the scope of this review. But why do we want Task to hold the SC Config. Could we move the initialization of ephemeral ports,etc into `sevice_connect.go`?

angelcar: > But why do we want Task to hold the SC Config

SC Config is an attribute of the task, and as such it will be persisted/loaded from to the agent database in case of agent restarts.

> Could we move the initialization of ephemeral ports, etc into sevice_connect.go?'

Yeah, I think we could move all that to service connect in  a new PR.

rajal-amzn: Yes. Lets take it up in a new PR.

agent/api/task/service_connect.go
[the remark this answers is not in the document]
chienhanlin: Thanks @yinyic. Constants are updated to 
```
const (
	serviceConnectConfigKey        = "ServiceConnectConfig"
	serviceConnectContainerNameKey = "ContainerName"
)
```
in commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054

agent/api/task/service_connect.go
```diff
@@ -65,3 +80,55 @@ type DNSConfigEntry struct {
...
+		}
+	}
+	scConfigValue.ContainerName = containerName
+	if err := ValidateSCConfig(scConfigValue); err != nil {
+		return nil, err
+	}
+	return scConfigValue, nil
+}
+
+func ValidateSCConfig(scConfig *ServiceConnectConfig) error {
```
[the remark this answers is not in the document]
chienhanlin: More validations are added to commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054. Please take a look on rev 2 of this PR. Thanks.

agent/api/task/service_connect.go
```diff
@@ -65,3 +80,55 @@ type DNSConfigEntry struct {
...
+	// no service connect attachment is found
+	if scAttachment == nil {
+		return nil, nil
+	}
+	// translate ecsacs.Attachment.Value to ServiceConnectConfig
+	scConfigValue := &ServiceConnectConfig{}
+	for _, property := range scAttachment.AttachmentProperties {
+		switch aws.StringValue(property.Name) {
+		case serviceConnectAttProSCConfigName:
+			s, _ := strconv.Unquote(string(*property.Value))
```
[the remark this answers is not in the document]
chienhanlin:
```
if err != nil {
    return nil, errors.Errorf("failed to unqote attachment property value which contains service connect config value: %s", err)
}
```
is added to commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054. Thanks.

agent/api/task/service_connect_test.go
chienhanlin: test service config is broken down into multi-lines, and couple format constants, shown as follows, are defined and used in commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054.  
```
testStatsConfigFormat                             = `\"statsConfig\":{\"path\":\"%s\"}`
testEgressConfigFormat                            = `\"egressConfig\":{\"listenerName\":\"%s\",\"vip\":{\"ipv4Cidr\":\"%s\",\"ipv6Cidr\":\"%s\"}}`
testDnsConfigFormat                               = `\"dnsConfig\":[{\"hostname\":\"%s\",\"address\":\"%s\"}]`
testIngressConfigAwsVpcDefaultFormat              = `\"ingressConfig\":[{\"interceptPort\":%s,\"listenerName\":\"%s\"}]`
testIngressConfigBridgeOverrideFormat             = `\"ingressConfig\":[{\"listenerPort\":%s,\"hostPort\":%s}]`
testIngressConfigListenerPortOnlyFormat           = `\"ingressConfig\":[{\"listenerPort\":%s}]`
testInvalidIngressConfigAwsVpcDefaultFormat       = `\"ingressConfig\":[{\"interceptPort\":%s,\"listenerName\":\"%s\"}` + `,{\"interceptPort\":%s,\"listenerName\":\"%s\"}]`
testInvalidIngressConfigListenerPortFormat        = `\"ingressConfig\":[{\"listenerPort\":%s},{\"listenerPort\":%s}]`
testInvalidIngressConfigBridgeOverrideFormat      = `\"ingressConfig\":[{\"listenerPort\":%s,\"hostPort\":%s},{\"listenerPort\":%s,\"hostPort\":%s}]`
testInvalidMixedIngressConfigBridgeOverrideFormat = `\"ingressConfig\":[{\"listenerPort\":%s,\"hostPort\":%s},{\"interceptPort\":%s,\"listenerName\":\"%s\"}]`
```
Thanks.

agent/api/task/service_connect_test.go
chienhanlin: ingressConfig is renamed to expectedIngressConfig in commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054. Thanks!

agent/api/task/service_connect_test.go
chienhanlin: testSCConfig is renamed to expectedTestSCConfig in commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054. Thanks!

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+		case serviceConnectContainerNameKey:
+			// extract service connect container name from attachment properties
+			containerName = aws.StringValue(property.Value)
+		default:
+			return nil, errors.Errorf("received an unrecognized attachment property: %s", property.String())
+		}
+	}
+	scConfigValue.ContainerName = containerName
+	if err := ValidateSCConfig(scConfigValue, taskContainers, networkMode); err != nil {
+		return nil, errors.Errorf("service connect config validation failed: %s", err)
```
angelcar, line 111: nit: the preferred way to wrap errors is now `fmt.Errorf("message %w", err)`

agent/api/task/service_connect_test.go
chienhanlin: These test cases might failed on incorrect validation or test setup I had in the previous revision. Please take a look at commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054 with updated test cases and validations. Thanks.

agent/api/task/service_connect.go
chienhanlin:
```
default:
    return nil, errors.Errorf("received an unrecognized attachment property: %s", property.String())
```
is added in commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054. Thanks!

agent/api/task/service_connect.go
chienhanlin: Please take a look at commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054 with updated test cases and validations. Thanks.

agent/api/task/service_connect.go
chienhanlin: Thanks for the suggestion. Logic to get the service connect attachment is moved to task.go, and the function is renamed to `ParseServiceConnectAttachment` in commit 8e69368bc5c1d7bd3ab71391103f4b6a69f06054.

agent/api/task/task.go
chienhanlin: Please find the update in commit 94753650e02342a2c1be58645e8e187489efe452. Thanks!
```
// extract the service connect container name and service connect config value from the service connect attachment
    if scAttachment != nil {
        scConfig, err := ParseServiceConnectAttachment(scAttachment, acsTask.Containers, aws.StringValue(acsTask.NetworkMode))
	 if err != nil {
	     return nil, fmt.Errorf("error parsing service connect config value from the service connect attachment: %w", err)
	 }
	 task.ServiceConnectConfig = scConfig
 }
```

agent/api/task/service_connect.go
chienhanlin: Updates are made in commit 94753650e02342a2c1be58645e8e187489efe452. Thanks!

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+// validateBridgeIngressConfig validates the service connect ingress config with bridge network mode.
+func validateBridgeIngressConfig(scIngressConfigList []IngressConfigEntry) error {
+	hostPortsMap := map[uint16]bool{}
+	hostPortValue := uint16(0)
+	listenerPortsMap := map[uint16]bool{}
+	listenerPortValue := uint16(0)
+	for _, entry := range scIngressConfigList {
+		// verify intercept port does not exist in bridge mode
+		if entry.InterceptPort != nil {
+			return fmt.Errorf("service connect config validation: intercept port should not exist for bridge mode")
```
angelcar, line 269: same here, I feel this should be a warning. Although not positive. Would like others to chime in.

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+				return nil, fmt.Errorf("failed to unqote attachment property value which contains service connect config value: %w", err)
+			}
+			if err := json.Unmarshal([]byte(data), scConfigValue); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal attachment property value which contains service connect config value: %w", err)
+			}
+		case serviceConnectContainerNameKey:
+			// extract service connect container name from attachment properties
+			containerName = aws.StringValue(property.Value)
+		default:
+			return nil, fmt.Errorf("received an unrecognized attachment property: %s", property.String())
```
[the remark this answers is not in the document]
chienhanlin: Got it. Like log the warning message as follows, and continue checking other properties. 
```
logger.Warn("Received an unrecognized attachment property", logger.Fields{
    "attachmentProperty":    property.String(),
})
```

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+	if err != nil {
+		return fmt.Errorf("service connect config validation: %s: %s is invalid", version, cidr)
+	}
+	return nil
+}
+
+// validateEgressConfig validates the service connect egress config has listener name and vip if egress config is not empty.
+func validateEgressConfig(scEgressConfig EgressConfig) error {
+	// EgressConfig can be empty for the first service since there are no other tasks that it can talk to
+	if scEgressConfig.ListenerName == "" && scEgressConfig.VIP.IPV4CIDR == "" && scEgressConfig.VIP.IPV6CIDR == "" {
```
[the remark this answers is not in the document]
chienhanlin: Will update it with option 2 in next revision. Thanks!

agent/api/task/service_connect.go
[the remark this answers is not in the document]
rajal-amzn: +1. This makes it forward compatible so that any new properties added will not affect the older agents.

agent/api/task/service_connect.go
```diff
@@ -13,13 +13,29 @@
...
+
 // ServiceConnectConfig represents the Service Connect configuration for a task.
 type ServiceConnectConfig struct {
 	ContainerName string               `json:"containerName"`
 	StatsConfig   StatsConfig          `json:"statsConfig"`
-	IngressConfig []IngressConfigEntry `json:"ingressConfig"`
-	EgressConfig  EgressConfig         `json:"egressConfig"`
-	DNSConfig     []DNSConfigEntry     `json:"dnsConfig"`
+	IngressConfig []IngressConfigEntry `json:"ingressConfig,omitempty"`
+	EgressConfig  EgressConfig         `json:"egressConfig,omitempty"`
```
rajal-amzn, line 37: `omitempty` doesnt really help here since the values inside the struct would still be initialized. It is better to use pointers here. Same applies to `EgressConfig.ListenerPort` and `VIP.IPV4CIDR` and `VIP.IPV6CIDR` below. [This](https://www.sohamkamani.com/golang/omitempty/) has a good explanation of `omitempty`

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+			}
+		case serviceConnectContainerNameKey:
+			// extract service connect container name from attachment properties
+			containerName = aws.StringValue(property.Value)
+		default:
+			return nil, fmt.Errorf("received an unrecognized attachment property: %s", property.String())
+		}
+	}
+	scConfigValue.ContainerName = containerName
+	if err := ValidateSCConfig(scConfigValue, taskContainers, networkMode); err != nil {
```
rajal-amzn, line 110: Could we move this to a new Validator Class?

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+// validateContainerName validates the service connect container name exists, and it exists in the task definition as well.
+func validateContainerName(scContainerName string, taskContainers []*ecsacs.Container) error {
+	// container name is a required field
+	if scContainerName == "" {
+		return fmt.Errorf("service connect config validation: no service connect container name")
+	}
+	// validate the specified service connect container name exists in task
+	for _, container := range taskContainers {
+		if aws.StringValue(container.Name) == scContainerName {
+			return nil
```
rajal-amzn, line 156: We must not be returning `nil` here as it would skip processing rest of the containers depending on the order of processing. Instead, we must ensure that there is only one container with the given name.

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+	}
+	// validate IPV4CIDR if it exists
+	if scEgressConfig.VIP.IPV4CIDR != "" {
+		trimmedIpv4cidr := strings.TrimSpace(scEgressConfig.VIP.IPV4CIDR)
+		if err := validateCIDR(trimmedIpv4cidr, "IPV4CIDR"); err != nil {
+			return err
+		}
+	}
+	// validate IPV6CIDR if it exists
+	if scEgressConfig.VIP.IPV6CIDR != "" {
```
rajal-amzn, line 193: Shall we add a validation that IPV6CIDR must not be empty when the task is IPV6 enabled?

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+	}
+	return scConfigValue, nil
+}
+
+// validateDnsConfig validates the service connnect dns config exists if egress config exists.
+func validateDnsConfig(scDnsConfligList []DNSConfigEntry, scEgressConfig EgressConfig) error {
+	// DNSConfig will also be empty as EgressConfig can be empty for the first service
+	// since there are no other tasks that it can talk to
+	if len(scDnsConfligList) < 1 {
+		if scEgressConfig.ListenerName != "" {
```
rajal-amzn, line 121: nit: Combine these two using &&

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+			return fmt.Errorf("service connect config validation: no service connect DNS config")
+		}
+	}
+	for _, dnsEntry := range scDnsConfligList {
+		// HostName is a required field
+		if dnsEntry.HostName == "" {
+			return fmt.Errorf("service connect config validation: no service connect DNS config host name")
+		}
+		// Address is a required field
+		if dnsEntry.Address == "" {
```
rajal-amzn, line 131: Could we add a validation that this is a valid v4/v6 IP?

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+
+// validateAwsVpcIngressConfig validates the service connect ingress config with awsvpc network mode.
+func validateAwsVpcIngressConfig(scIngressConfigList []IngressConfigEntry) error {
+	portsMap := map[uint16]bool{}
+	interceptPortValue := uint16(0)
+	listenerPortValue := uint16(0)
+	for _, entry := range scIngressConfigList {
+		// verify host port does not exist in awsvpc mode
+		if entry.HostPort != nil {
+			return fmt.Errorf("service connect config validation: host port should not exist for awsvpc mode")
```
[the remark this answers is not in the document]
rajal-amzn: +1. Though HostPort is not valid, I would not want to stop task from starting. We can simply ignore it. In that case, I dont believe we need 2 methods for awsvpc and bridge. We can combine them

agent/api/task/service_connect.go
rajal-amzn: Yes. +1 to warning and skip

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+		}
+	}
+	return fmt.Errorf("service connect config validation: service connect container name: %s does not exist in task", scContainerName)
+}
+
+// validateCIDR validates the passed CIDR is valid.
+func validateCIDR(cidr, version string) error {
+	_, _, err := net.ParseCIDR(cidr)
+	if err != nil {
+		return fmt.Errorf("service connect config validation: %s: %s is invalid", version, cidr)
```
rajal-amzn, line 166: Could we also add a validation on the version?

agent/api/task/task.go
```diff
@@ -301,8 +303,25 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
 	} else if task.GetDesiredStatus() == apitaskstatus.TaskStopped && envelope.SeqNum != nil {
 		task.StopSequenceNumber = *envelope.SeqNum
 	}
+	// search the service connect attachment from acsTask.Attachments
+	if acsTask.Attachments != nil {
```
rajal-amzn, line 307: Could you add test for this?

agent/api/task/service_connect.go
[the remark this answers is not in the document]
rajal-amzn: I was suggesting that these constants are already available as part of AppNet models imported. Could you check whether we can use that instead?

agent/api/task/service_connect.go
chienhanlin: Validations are added to commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `service_connect_validation.go`. Thanks.

agent/api/task/service_connect.go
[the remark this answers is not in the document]
chienhanlin: Warning message are added to commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `validateAwsVpcIngressConfig()` and `validateBridgeIngressConfig()` in `service_connect_validation.go`. Thanks.

agent/api/task/service_connect.go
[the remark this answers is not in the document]
chienhanlin: Use pointer for EgressConfig is updated in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `service_connect.go`. Thanks.

agent/api/task/service_connect.go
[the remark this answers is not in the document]
chienhanlin: Warning message are added to commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `validateAwsVpcIngressConfig()` and `validateBridgeIngressConfig()` in `service_connect_validation.go`. Thanks.

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+			return fmt.Errorf("service connect config validation: host port should not exist for awsvpc mode")
+		}
+		// verify intercept ports have no conflicts
+		if entry.InterceptPort != nil {
+			interceptPortValue = aws.Uint16Value(entry.InterceptPort)
+			if err := validatePort(interceptPortValue, AWSVPCNetworkMode); err != nil {
+				return err
+			}
+			if entry.ListenerName == "" {
+				return fmt.Errorf("service connect config validation: no listener name")
```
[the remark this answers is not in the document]
chienhanlin: Listener name is added to the error in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `validateAwsVpcIngressConfig()` in `service_connect_validation.go`. Thanks.

agent/api/task/service_connect.go
```diff
@@ -65,3 +81,246 @@ type DNSConfigEntry struct {
...
+			}
+			if present := portsMap[interceptPortValue]; present {
+				return fmt.Errorf("service connect config validation: intercept port %d conflicts", interceptPortValue)
```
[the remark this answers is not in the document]
chienhanlin: Error messages are updated in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `validateAwsVpcIngressConfig()` and `validateBridgeIngressConfig()` in `service_connect_validation.go`. Thanks.

agent/api/task/service_connect.go
chienhanlin: Types are updated in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `service_connect.go`. Thanks.

agent/api/task/service_connect.go
chienhanlin: Validation and its test are added in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `service_connect_validation.go` and `service_connect_validation_test.go`. Thanks.

agent/api/task/service_connect.go
chienhanlin: Validation and its test are added in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `service_connect_validation.go` and `service_connect_validation_test.go`. Thanks.

agent/api/task/service_connect.go
chienhanlin: Validation and its test are added in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `service_connect_validation.go` and `service_connect_validation_test.go`. Thanks.

agent/api/task/service_connect.go
chienhanlin: `validateCIDR` is added in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `service_connect_validation.go` and `service_connect_validation_test.go`. Thanks.

agent/api/task/task.go
chienhanlin: `TestTaskServiceConnectAttachment()` and `TestTaskWithoutServiceConnectAttachment()` are added in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `agent/api/task/task_test.go`. Thanks.

agent/api/task/service_connect.go
chienhanlin: Update is made in commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on `validateContainerName()` in `service_connect_validation.go`. Thanks.

angelcar (approved):

Makefile
```diff
@@ -267,7 +267,7 @@ GOFILES:=$(shell go list -f '{{$$p := .}}{{range $$f := .GoFiles}}{{$$p.Dir}}/{{
 .PHONY: gocyclo
 gocyclo:
 	# Run gocyclo over all .go files
-	gocyclo -over 17 ${GOFILES}
+	gocyclo -over 19 ${GOFILES}
```
yinyic, line 270: gocyclo is an indicator of over-complicated function logic. Which function is it warning on? Let's try to break down that function if possible, and only increase the limit if necessary.

agent/api/task/service_connect.go
```diff
@@ -72,3 +86,33 @@ type DNSConfigEntry struct {
 	HostName string `json:"hostName"`
 	Address  string `json:"address"`
 }
+
+// ParseServiceConnectAttachment parses the service connect container name and service connect config value
+// from the given attachment.
+func ParseServiceConnectAttachment(scAttachment *ecsacs.Attachment, networkMode string, ipv6Enabled bool) (*ServiceConnectConfig, error) {
```
yinyic, line 92: Are `networkMode` and `ipv6Enabled` actually used?

agent/api/task/service_connect_validation.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil
+	}
+
+	// ListenerName is a required field if the egress config is not empty
+	if scEgressConfig.ListenerName == "" {
+		return fmt.Errorf("no service connect listener name")
+	}
+
+	// VIP is a required field if the egress config is not empty
+	if !ipv6Enabled && scEgressConfig.VIP.IPV4CIDR == "" {
```
yinyic, line 63: I think IPv4Cidr should be always required because we don't really support an IPv6-only mode at this moment. It's either IPv4, or dual stack.

agent/api/task/service_connect_validation.go
```diff
@@ -0,0 +1,293 @@
...
+		if err := validateAddress(dnsEntry.Address); err != nil {
+			return fmt.Errorf("invalid address in the DNS config entry hostname=%s, address=%s: %w", dnsEntry.HostName, dnsEntry.Address, err)
+		}
+	}
+
+	return nil
+}
+
+// validateAddress validates the passed address is a valid IPv4/IPv6.
+func validateAddress(address string) error {
```
yinyic, line 118: Would recommend checking against the given protocol so that we don't take an IPv4 inside a supposedly IPv6 field.
I think we can do something similar to Rajal's [PR](https://github.com/aws/amazon-vpc-cni-plugins/pull/64/files#diff-6ecce0879e112f7e1e1d78bdf2d8ac8639491588157e561563df634554a12dbc) (i.e. using To4() and To16())

agent/api/task/service_connect_validation.go
```diff
@@ -0,0 +1,293 @@
...
+	if ip := net.ParseIP(address); ip == nil {
+		return fmt.Errorf("address=%s is not a valid textual representation of an IP address", address)
+	}
+	return nil
+}
+
+// validateCIDR validates the passed CIDR is a valid IPv4/IPv6 CIDR.
+func validateCIDR(cidr, version string) error {
```
yinyic, line 126: Same as above

agent/api/task/service_connect_validation.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+// ValidateSCConfig validates service connect container name, config, egress config, and ingress config.
+func ValidateServiceConnectConfig(scConfig *ServiceConnectConfig, taskContainers []*ecsacs.Container, taskNetworkMode string, ipv6Enabled bool) error {
+	// container name is a required field
+	if err := validateContainerName(scConfig.ContainerName, taskContainers); err != nil {
+		return err
+	}
+
+	// egress config is a field that can be empty
```
yinyic, line 277: Can we also check that egress and ingress config are not both nil?

agent/api/task/task.go
```diff
@@ -1753,7 +1795,7 @@ func (task *Task) dockerExposedPorts(container *apicontainer.Container) (dockerE
 			}
 			// expose egress listener port if present
 			ec := task.ServiceConnectConfig.EgressConfig
-			if ec.ListenerName != "" { // it's possible that task does not have an egress listener
+			if ec != nil && ec.ListenerName != "" { // it's possible that task does not have an egress listener
```
yinyic, line 1798: I think we can get rid of `ec.ListenerName != ""` :) 
Back then egress config can never be nil so this is the only way checking whether one has been actually specified.

agent/acs/model/ecsacs/api.go
```diff
@@ -285,6 +285,44 @@ func (s AttachTaskNetworkInterfacesOutput) GoString() string {
 	return s.String()
 }
 
+type Attachment struct {
+	_ struct{} `type:"structure"`
+
+	AttachmentArn *string `locationName:"attachmentArn" type:"string"`
+
+	AttachmentProperties []*AttachmentProperty `locationName:"attachmentProperties" type:"list"`
```
rajal-amzn, line 293: nit: You dont need to use a pointer for list objects. You could simply check length of the list.

agent/api/task/service_connect.go
rajal-amzn, line 92: Given that the this class also has the actual logic of setting up, could you please move this logic to a new class - `ServiceConnectAttachmentParser` ?

agent/api/task/service_connect.go
```diff
@@ -72,3 +86,33 @@ type DNSConfigEntry struct {
...
+	scConfigValue := &ServiceConnectConfig{}
+	containerName := ""
+
+	for _, property := range scAttachment.AttachmentProperties {
+		switch aws.StringValue(property.Name) {
+		case serviceConnectConfigKey:
```
rajal-amzn, line 98: We are not checking anywhere that these keys are present. Could we validate those?

agent/api/task/service_connect_validation.go
```diff
@@ -0,0 +1,293 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
rajal-amzn, line 1: nit: Could we call this `service_connect_validator`? I would also suggest creating an interface for Attachment, Parser and Validator and use that for all attachment types.

agent/api/task/service_connect_validation.go
```diff
@@ -0,0 +1,293 @@
...
+// validateEgressConfig validates the service connect egress config.
+func validateEgressConfig(scEgressConfig *EgressConfig, ipv6Enabled bool) error {
+	// egress config can be empty for the first service since there are no other tasks that it can talk to
+	if scEgressConfig == nil {
+		return nil
+	}
+
+	// ListenerName is a required field if the egress config is not empty
+	if scEgressConfig.ListenerName == "" {
+		return fmt.Errorf("no service connect listener name")
```
rajal-amzn, line 59: nit: given we have both ingress and egress listener, could we be specific here in the error messsage?

agent/api/task/service_connect_validation.go
rajal-amzn: I actually prefer this so that we are future proof when we support V6-only but for that we would also need additional variable - `ipv4Enabled`

agent/api/task/service_connect_validation.go
```diff
@@ -0,0 +1,293 @@
...
+		}
+	default:
+		return fmt.Errorf("service connect does not support for %s newtork mode", taskNetworkMode)
+	}
+
+	return nil
+}
+
+// validateAwsVpcIngressConfig validates the service connect ingress config for awsvpc network mode.
+func validateAwsVpcIngressConfig(scIngressConfigList []IngressConfigEntry) error {
```
rajal-amzn, line 158: The only difference between this and Bridge mode validation is around the presence of hostPort. Could we combine these into one so that we dont duplicate all of the other validation?

agent/api/task/task.go
```diff
@@ -149,6 +149,8 @@ const (
 	// sysctlValueOff specifies the value to use to turn off a sysctl setting.
 	sysctlValueOff = "0"
 
+	// serviceConnectAttachmentType specifies attachment type for service connect
+	serviceConnectAttachmentType            = "ServiceConnect"
```
rajal-amzn, line 153: Similar to the comment above. Lets create a new class - `TaskAttachment` and move the constants there.

agent/api/task/service_connect_validation.go
chienhanlin: Discussed offline, as DNS config comes in as a list, we will validate CIDR based on the given protocol.

agent/api/task/service_connect_validation.go
chienhanlin:
```
// egress config and ingress config should not both be nil/empty
	if scConfig.EgressConfig == nil && len(scConfig.IngressConfig) == 0 {
		return fmt.Errorf("egress config and ingress config should not both be nil/empty")
	}
```
is added to agent/api/task/service_connect_validator.go in the commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look. Thanks!

agent/api/task/service_connect_validation.go
chienhanlin: Rename files to validator in the commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on the agent/api/task/service_connect_validator.go and agent/api/task/service_connect_validator_test.go. Thanks

agent/api/task/service_connect_validation.go
chienhanlin: Refactor ingress config validation in the commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on the agent/api/task/service_connect_validator.go. Thanks

agent/api/task/service_connect.go
chienhanlin: Moved the function to a new file in the commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on agent/api/task/service_connect_attachment_parser.go. Thanks.

agent/api/task/task.go
chienhanlin: Updated the if statement in the commit 3d8d95406e74fc25d3c00d384a11cd5cd4ae55f9. Please take a look on agent/api/task/task.go. Thanks.

angelcar (dismissed):

yinyic (dismissed):

angelcar (approved):

yinyic (approved):

PR #3211, opened 2022-05-10
Refactor out ServiceConnect from task engine

aws-gibbskt:
Reuse the manager paradigm for Service Connect modification.

### Summary
Moves the logic out of docker_task_engine into a service connect manager 

### Licensing


angelcar (dismissed):

rajal-amzn (dismissed):

agent/engine/docker_task_engine.go
```diff
@@ -1254,10 +1256,12 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 		return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(hcerr)}
 	}
 
-	// Add SC VIPs to pause container's known hosts
-	if task.IsServiceConnectEnabled() && container.Type == apicontainer.ContainerCNIPause {
-		hostConfig.ExtraHosts = append(hostConfig.ExtraHosts,
-			serviceconnect.DNSConfigToDockerExtraHostsFormat(task.ServiceConnectConfig.DNSConfig)...)
+	// Add Service Connect modifications if needed
+	if task.IsServiceConnectEnabled() {
+		err := engine.serviceconnectManager.InitializeTaskContainer(task, container, hostConfig)
```
yinyic, line 1261: Nit - wondering if we may want to call it something other than `Initialize` because the container is initialized during unmarshalling, and later on we keep adding more launch configurations. Maybe `AugmentTaskContainer` can be more accurate? Or, maybe we can be a little more specific of what properties we are initializing, `InitializeTaskContainerHostConfig` for example.

aws-gibbskt: It isn't limited to HostConfig in practice so Augment will work better.

angelcar (approved):

yinyic (approved):

Issue #2579, opened 2020-08-18
Task containers dependencies resolution stuck forever


Trane9991:
### Summary
Task stuck in **Pending** state because of some problems during containers dependencies resolution.
Setting `Start timeout` doesn't have any effect as well.
This is reopening of my previous issue #2350, with the additional details and a Task Definition to reproduce the error.

### Description
I have 3 containers:

| Container name | Essential | Depends On Container | Depends On State | Start Timeout |
|----------------|-----------|----------------------|------------------|---------------|
| exit1          | false     |                      |                  | 10            |
| mysql          | true      | exit1                | SUCCESS          | 30            |
| nginx          | true      | mysql                | START            | 60            |

(`nginx` depends on `mysql` which depends on `exit1`)

When container `exit1` fails with the exit code `1`, containers `mysql` and `nginx` remains in the **PENDING** state forever.
When container `exit1` succeeds with exit code `0` everything works fine.

I also tried to set container `nginx` dependencies like:
```json 
      [
        {
          "containerName": "mysql",
          "condition": "START"
        },
        {
          "containerName": "exit1",
          "condition": "SUCCESS"
        }
      ]
```
the result is the same.

However, behavior seems to be correct if the order of dependencies for `nginx` is changed:
```json 
      [
        {
          "containerName": "exit1",
          "condition": "SUCCESS"
        },
        {
          "containerName": "mysql",
          "condition": "START"
        }
      ]
```

### Expected Behavior
The task is failed to start.

### Observed Behavior
Task stuck in the **PENDING** state forever

<img width="1106" alt="Screen Shot 2020-08-18 at 5 02 17 PM" src="https://user-images.githubusercontent.com/9948629/90522648-a346bf00-e174-11ea-98a8-b07534ae04c6.png">

### Environment Details
- ECS Agent version: 1.43.0

### Task Definitions to reproduce:
<details>
  <summary> With transitive dependencies - stuck as <b>PENDING</b> ❌ </summary>

```json
{
    "containerDefinitions": [
        {
            "command": [
                "cat",
                "123"
            ],
            "image": "alpine",
            "startTimeout": 10,
            "name": "exit1",
            "essential": false
        },
        {
            "image": "nginx",
            "startTimeout": 60,
            "dependsOn": [
                {
                    "containerName": "mysql",
                    "condition": "START"
                }
            ],
            "name": "nginx"
        },
        {
            "image": "mysql:5.7",
            "startTimeout": 30,
[...]

PR #3218, opened 2022-05-19
Stop container when dependencies are non-resolvable

angelcar:
### Summary

Addresses https://github.com/aws/amazon-ecs-agent/issues/2579 where a task can be stuck in `PENDING`  for ever when container dependencies cannot be fulfilled.

### Implementation details

Adding the concept of terminal dependency error to detect cases where container dependencies can never succeed. When a terminal dependency error is detected, we simply forcefully stop the container with the "bad" dependencies. 

This new behavior triggers a chain reaction that eventually stops the task if an essential container has terminal dependency errors. If essential containers don't have terminal dependency errors, then the task is allowed to start.

Containers that are forcefully stopped are marked with exitCode `143`. This is the code docker assigns when a container receives `SIGTERM`. Although we technically don't send `SIGTERM` to containers stopped due to terminal dependency errors, this exit code is the closest to reality. However, customers will be able to see a text description of the exact error when this edge case is hit.

For example (from AWS console): 

Status reason | CannotStartContainerError:  dependency graph: failed to resolve container ordering dependency  [container-1(public.ecr.aws/u5s6c6g2/just-sleep) (STOPPED->RUNNING) -  Exit: 1] for target [container-2(public.ecr.aws/u5s6c6g2/just-sleep)  (NONE->RUNNING)] as dep
-- | --
143
["/sleep","10"]

### Testing

Bug - Fix an [issue](https://github.com/aws/amazon-ecs-agent/issues/2579) where a task can be stuck in `PENDING` for ever when container dependencies can never be fulfilled. 

### Licensing

### DIFF
--- a/agent/engine/dependencygraph/graph.go
+++ b/agent/engine/dependencygraph/graph.go
@@ -47,23 +47,39 @@ const (
 var (
 	// CredentialsNotResolvedErr is the error where a container needs to wait for
 	// credentials before it can process by agent
-	CredentialsNotResolvedErr = errors.New("dependency graph: container execution credentials not available")
+	CredentialsNotResolvedErr = &DependencyError{err: errors.New("dependency graph: container execution credentials not available")}
 	// DependentContainerNotResolvedErr is the error where a dependent container isn't in expected state
-	DependentContainerNotResolvedErr = errors.New("dependency graph: dependent container not in expected state")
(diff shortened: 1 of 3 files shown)


singholt (approved):
🙌

fierlion (approved):
(Thanks for the well reasoned PR with clear tests which act as a walkthrough of the changes.)

PR #3223, opened 2022-05-24
Cleanup the symantics of ACS Handler

aws-gibbskt:
### Summary
The ACS handler was using a channel to support looping, but the
channel was only sent to from the current thread via a go routine
that would just write to the channel.

The rewrite maintains the behavior of always connecting when
disconnected potentially after a retry backoff. However it uses
a simple endless for loop rather than reading from channels to
mimic this straightforward behavior.

### Licensing


agent/acs/handler/acs_handler.go
```diff
@@ -189,70 +189,59 @@ func NewSession(
...
+				// If the context was not cancelled and we've waited for the
+				// wait duration without any errors, send the message to the channel
+				// to reconnect to ACS
+				seelog.Info("Done waiting; reconnecting to ACS")
+			} else {
+				// Wait was interrupted. We expect the session to close as canceling
+				// the session context is the only way to end up here. Print a message
+				// to indicate the same
+				seelog.Info("Interrupted waiting for reconnect delay to elapse; Expect session to close")
+				return nil
```
nodir-t, line 242: Note that this isn't equivalent to the prev revision. This exits without waiting for the context to get canceled. The equivalent version would be do `<- ctx.Done()` before returning

agent/acs/handler/acs_handler.go
```diff
@@ -189,70 +189,59 @@ func NewSession(
...
+			// If ACS closed the connection, there's no need to backoff,
+			// reconnect immediately
+			seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
+			acsSession.backoff.Reset()
+		} else {
+			// Disconnected unexpectedly from ACS, compute backoff duration to
+			// reconnect
+			reconnectDelay := acsSession.computeReconnectDelay(isInactiveInstance)
+			seelog.Infof("Reconnecting to ACS in: %s", reconnectDelay.String())
+			waitComplete := acsSession.waitForDuration(reconnectDelay)
```
nodir-t, line 231: nit: inline this variable?

nodir-t (changes requested):

agent/acs/handler/acs_handler.go
aws-gibbskt: It actually is. In the sense that Done() being signaled is the only way that false is returned. We could check it again here if you want but it isn't any different.

nodir-t: you are right

agent/acs/handler/acs_handler.go
```diff
@@ -189,70 +189,59 @@ func NewSession(
 // Start starts the session. It'll forever keep trying to connect to ACS unless
 // the context is cancelled.
 //
-// If the context is cancelled, Start() would return with the error code returned
-// by the context.
+// If the context is cancelled, Start() would return nil
```
nodir-t, line 192: seems like it returns nil unconditionally. Might as well remove the return type.

nodir-t (approved):

angelcar:
I don't think this code is equivalent? If I am not mistaken, this loop will execute over and over since nothing is pausing the loop execution -- the `default` clause causes the `case <-acsSession.ctx.Done()` to be ignored.

aws-gibbskt:
Oops yeah I closed this one in favor of https://github.com/aws/amazon-ecs-agent/pull/3225 which is targeting dev not a feature branch... @nodir-t @angelcar

PR #3224, opened 2022-05-25
Invoke Drain Connections API on Task Stop for Service Connect Tasks

angelcar:
### Summary

* Invokes Appnet Agent's Drain Connections API when a service-connect-enabled task first transitions to STOPPED. 
* Integrates Appnet Client to stats engine

### Implementation details

* A new Appnet client is added yo invoke Appnet agent's HTTP APIs over UDS.
* DrainConnections API is invoked first thing in engine's `stopContainer` method. This happens just once and only for service connect enabled tasks.

### Testing

* Added new UTs

### Licensing


agent/engine/task_manager.go
```diff
@@ -1040,19 +1066,7 @@ func (mtask *managedTask) progressTask() {
...
-		})
-
-		// If knownStatus changed, let it be known
-		var taskStateChangeReason string
-		if mtask.GetKnownStatus().Terminal() {
-			taskStateChangeReason = mtask.Task.GetTerminalReason()
-		}
-		mtask.emitTaskEvent(mtask.Task, taskStateChangeReason)
-	}
+	mtask.manageTaskStatusChange()
```
yinyic, line 1069: Hmm should we move this up? At this point we've already started container transitions (line 1022), but we want to invoke AppNet draining before any of that happen.

agent/api/appnetclient/client_linux.go
```diff
@@ -0,0 +1,89 @@
...
+	unixNetworkName   = "unix"
+	httpRequestPrefix = "http://" + unixNetworkName
+	statsUrl          = httpRequestPrefix + "/stats/prometheus?usedonly&filter=metrics_extension&delta"
+	drainUrl          = httpRequestPrefix + "/drain_listeners?inboundonly"
+)
+
+type appnetClientCtxKey int
+
+const (
+	udsAddressKey appnetClientCtxKey = iota
```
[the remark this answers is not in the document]
angelcar: yes. It's a bad practice to use text as context keys since you never know if they're going to collision in the future.

agent/api/appnetclient/client_linux.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+func udsDialContext(ctx context.Context, _, _ string) (net.Conn, error) {
+	udsPath, ok := ctx.Value(udsAddressKey).(string)
+	if !ok {
+		return nil, fmt.Errorf("appnet client: Path to appnet admin socket was not a string")
+	}
+	if udsPath == "" {
+		return nil, fmt.Errorf("appnet client: Path to appnet admin socket was not provided")
+	}
```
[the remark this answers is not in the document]
angelcar: if the passed value is a blank string (i.e.`config.AdminSocketPath` is blank), then `udsPath` will be blank, and `ok` will be true.

agent/engine/task_manager.go
angelcar: I will move this to `stopContainer` since it will be easier to UT.

agent/api/appnetclient/client_linux.go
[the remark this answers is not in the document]
angelcar: Yeah, that's fair. Will change to `Path to appnet admin socket was blank`

agent/engine/docker_task_engine.go
```diff
@@ -1835,6 +1841,19 @@ func (engine *DockerTaskEngine) inspectContainer(task *apitask.Task, container *
 }
 
 func (engine *DockerTaskEngine) stopContainer(task *apitask.Task, container *apicontainer.Container) dockerapi.DockerContainerMetadata {
+	// Before attempting to stop any container, send drain signal for Appnet Agent to start draining connections
+	// (if not already in progress).
+	if task.IsServiceConnectEnabled() && !task.IsServiceConnectConnectionDraining() {
+		if err := appnetDrainInboundConnections(task.GetServiceConnectRuntimeConfig()); err != nil {
```
yinyic, line 1847: We probably want to have some max number of retries otherwise task can potentially get stuck in stopping.

angelcar: This doesn't prevent a container from stopping?

yinyic: Oh sorry I misread, thought we'd just return. Are we okay with the best-effort approach (no retries)? I wonder if there can be some type transient errors from AppNet Agent that could potentially go away with a retry.

angelcar: I could add some retries internally in the client. Although. as-is, the next container that stops will retry, but it's probably better to do the retries internally in the client.

angelcar: Added retries inside the function. I think that should suffice for out purposes.

mythri-garaga (approved):

yinyic (approved):

mythri-garaga (approved):

yinyic (approved):

agent/api/appnet/client_linux.go
```diff
@@ -0,0 +1,100 @@
...
+	if udsPath == "" {
+		return nil, fmt.Errorf("appnet client: Path to appnet admin socket was blank")
+	}
+	return net.Dial(unixNetworkName, udsPath)
+}
+
+var udsHttpClient = http.Client{
+	Transport: &http.Transport{
+		DialContext: udsDialContext,
+	},
```
rajal-amzn, line 65: We would need to add some default retry strategy to the client. But we can take that up separately.

rajal-amzn (approved):

agent/api/task/service_connect.go
```diff
@@ -28,10 +28,6 @@ type ServiceConnectConfig struct {
 type RuntimeConfig struct {
 	// Host path for the administration socket
 	AdminSocketPath string `json:"adminSocketPath"`
-	// HTTP Path + Params to get statistical information
-	StatsRequest string `json:"statsRequest"`
-	// HTTP Path + Params to drain ServiceConnect connections
-	DrainRequest string `json:"drainRequest"`
```
aws-gibbskt, line 34: Since you are already using AdminSocketPath, using the request from here is more sane allowing for potentially different values over time which forcing them in the client doesn't allow. I would leave the requests here

angelcar: > allowing for potentially different values over time

hmm, you  mean different paths for for different tasks? 

As I'm seeing it, the Appnet client should know how to talk to Appnet agent. Otherwise it's not really the Appnet client, but rather a generic http-over-uds client. (Which I'm not necessarily opposed to).

agent/api/appnet/client_linux.go
angelcar: I implemented a basic retry for Drain API, which should suffice. 

For GetStats, it's not very clear at this time what would be a good retry strategy. The issue with retrying this API is that if the request does go through, the delta will be reset, so retrying will only get the "next" snapshot, which might not be what we want.

agent/api/task/service_connect.go
aws-gibbskt: The integration of AppNet agent container and versions of that should be consolidated. Different versions could change the http paths over time. So in runtimeConfig it is serialized to be picked up by new Agent versions but in the client it would need another mechanism to know which task is running which AppNetAgent etc. The client is still valuable in that it abstracts away the UDS HTTP client even if it gets the URL from a runtimeConfig or statically. That distinction doesn't change the abstraction or value of said abstraction

aws-gibbskt: https://github.com/aws/amazon-ecs-agent/pull/3230 as the differences, shows that it isn't really a change to what the client knows or doesn't know. But it changes where the integration configured values are stored together.

PR #3225, opened 2022-05-25
Cleanup the symantics of ACS Handler

aws-gibbskt:
### Summary

The ACS handler was using a channel to support looping, but the
channel was only sent to from the current thread via a go routine
that would just write to the channel.

The rewrite maintains the behavior of always connecting when
disconnected potentially after a retry backoff. However it uses
a simple endless for loop rather than reading from channels to
mimic this straightforward behavior.

The rewrite also adheres to the interface contract on returning
the error from the context if it is Done for any reason other
than cancelled.

### Testing
Added additional tests to validate that all paths reconnect or exit when Context is Error or Cancelled, with appropriate returns. No additional testing outside of Unit/Functional tests.

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -203,70 +203,54 @@ func NewSession(
 // Start starts the session. It'll forever keep trying to connect to ACS unless
 // the context is cancelled.
 //
-// If the context is cancelled, Start() would return with the error code returned
-// by the context.
-// If the instance is deregistered, Start() would emit an event to the
-// deregister-instance event stream and sets the connection backoff time to 1 hour.
+// Returns nil always TODO: consider removing error return value completely
 func (acsSession *session) Start() error {
-	// connectToACS channel is used to indicate the intent to connect to ACS
-	// It's processed by the select loop to connect to ACS
-	connectToACS := make(chan struct{}, 1)
-	// This is required to trigger the first connection to ACS. Subsequent
-	// connections are triggered by the handleACSError() method
-	connectToACS <- struct{}{}
+	// Loop continuously until context is closed/cancelled
 	for {
-		select {
-		case <-connectToACS:
-			seelog.Debugf("Received connect to ACS message")
-			// Start a session with ACS
-			acsError := acsSession.startSessionOnce()
-			select {
-			case <-acsSession.ctx.Done():
-				// agent is shutting down, exiting cleanly
-				return nil
-			default:
-			}
-			// Session with ACS was stopped with some error, start processing the error
-			isInactiveInstance := isInactiveInstanceError(acsError)
-			if isInactiveInstance {
-				// If the instance was deregistered, send an event to the event stream
-				// for the same
-				seelog.Debug("Container instance is deregistered, notifying listeners")
-				err := acsSession.deregisterInstanceEventStream.WriteToEventStream(struct{}{})
-				if err != nil {
-					seelog.Debugf("Failed to write to deregister container instance event stream, err: %v", err)
-				}
-			}
-			if shouldReconnectWithoutBackoff(acsError) {
-				// If ACS or agent closed the connection, there's no need to backoff,
-				// reconnect immediately
-				seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
-				acsSession.backoff.Reset()
-				sendEmptyMessageOnChannel(connectToACS)
-			} else {
-				// Disconnected unexpectedly from ACS, compute backoff duration to
-				// reconnect
-				reconnectDelay := acsSession.computeReconnectDelay(isInactiveInstance)
-				seelog.Infof("Reconnecting to ACS in: %s", reconnectDelay.String())
-				waitComplete := acsSession.waitForDuration(reconnectDelay)
-				if waitComplete {
-					// If the context was not cancelled and we've waited for the
-					// wait duration without any errors, send the message to the channel
-					// to reconnect to ACS
-					seelog.Info("Done waiting; reconnecting to ACS")
-					sendEmptyMessageOnChannel(connectToACS)
-				} else {
-					// Wait was interrupted. We expect the session to close as canceling
-					// the session context is the only way to end up here. Print a message
-					// to indicate the same
-					seelog.Info("Interrupted waiting for reconnect delay to elapse; Expect session to close")
-				}
-			}
-		case <-acsSession.ctx.Done():
-			// agent is shutting down, exiting cleanly
+		seelog.Debugf("Attempting connect to ACS")
+		// Start a session with ACS
+		acsError := acsSession.startSessionOnce()
+
+		// If the session is over check for shutdown first
+		if err := acsSession.ctx.Err(); err != nil {
 			return nil
 		}
 
+		// If ACS closed the connection, reconnect immediately
+		if shouldReconnectWithoutBackoff(acsError) {
+			seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
+			acsSession.backoff.Reset()
+			continue
+		}
+
+		// Session with ACS was stopped with some error, start processing the error
+		isInactiveInstance := isInactiveInstanceError(acsError)
+		if isInactiveInstance {
+			// If the instance was deregistered, send an event to the event stream
+			// for the same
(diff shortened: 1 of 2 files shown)


agent/acs/handler/acs_handler.go
```diff
@@ -189,70 +189,61 @@ func NewSession(
...
-		case <-connectToACS:
-			seelog.Debugf("Received connect to ACS message")
-			// Start a session with ACS
-			acsError := acsSession.startSessionOnce()
-			select {
-			case <-acsSession.ctx.Done():
-				// agent is shutting down, exiting cleanly
+		seelog.Debugf("Attempting connect to ACS")
+		// Start a session with ACS
+		acsError := acsSession.startSessionOnce()
```
angelcar, line 198: From my previous comment, I think what what threw me off is that `startSessionOnce()` blocks internally, which is not very intuitive. I see how the code is equivalent now. 

Perhaps a more idiomatic way of doing this would be for `startSessionOnce()` to return a channel, instead of blocking internally. That way it's clear that the consumer (this function) can wait for the channel to return an error, or be closed. In that case, this method would look something like below, which in my opinion is more idiomatic/intutitive:

```golang
for {
    seelog.Debugf("Attempting connect to ACS")
    select {
    case acsErr := <- acsSession.startSession():
        // all the checks currently made
        . . .
    case <- acsSession.ctx.Done(): 
        return nil
    }
}
```

nodir-t: I believe it is more Go-idiomatic to have blocking functions, so that the caller can decide its concurrency, instead of a function spinning up goroutine on its own and returning a channel. I think it is the caller that should decide if it needs a goroutine, and sometimes it does not like in this case. I find Kevin's implementation simple and correct.

aws-gibbskt: I don't think that helps create more clarity. startSessionOnce is synchronous and blocks as long as the connection is open that is very easy to reason about with synchronous code like in this PR. IF you really wanted to wait on something it should be the trigger to create a new connection which could be dispatched quickly then go back to this select. As it stands all the "checks currently made" are also synchronous and block for minutes to hours. If that is the contents of the select case then it doesn't follow for idiomatic message pump style coding in Go or any other language for that matter.

```go
for {
    select {
    case acsErr := <- acsSession.startSession:
        seelog.Debugf("Attempting connect to ACS")
        go acsSession.startSessionOnce()
    case <- acsSession.ctx.Done(): 
        return nil
    }
}

startSessionOnce() {
    // all code in startSessionOnce
   

   // after Serve returns but inside this fuction
   // all the checks currently made in select
   // if a time is selected
   go waitThenSignalChan(timeToWait, context, acsSession.startSession)
   // else if start immediately
   acsSession.startSession <- struct{}
}
```
This would be the idiomatic select usage but forces additional asynchronous goroutines and still allows for an easy control flow change that results in no ACS session ever again thus making the channel mechanism a burden not a help.

agent/acs/handler/acs_handler.go
```diff
@@ -189,70 +189,61 @@ func NewSession(
...
-				// reconnect immediately
-				seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
-				acsSession.backoff.Reset()
-				sendEmptyMessageOnChannel(connectToACS)
+		}
+		if shouldReconnectWithoutBackoff(acsError) {
+			// If ACS closed the connection, there's no need to backoff,
+			// reconnect immediately
+			seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
+			acsSession.backoff.Reset()
```
nodir-t, line 223: nit: maybe just `continue` here, and unindent the code below

nodir-t (approved):

agent/acs/handler/acs_handler.go
angelcar: > I believe it is more Go-idiomatic to have blocking functions, so that the caller can decide its concurrency, 

Actually `startSessionOnce` spawns go routines inside, so this is not true in this case.

agent/acs/handler/acs_handler.go
nodir-t: besides, a function starting a goroutine and returning a channel is leak-prone: what happens if the caller exited on canceled context and **never received** from the channel? In order for that extra goroutine not to leak (i.e. to be garbage collected), some extra code would be required, which is even more complicated. Sending to a channel without a select is a smell.

angelcar: That said this code is still better than the previous one so I'm fine with merging it as is.

angelcar (approved):

agent/acs/handler/acs_handler.go
nodir-t: sure it does, but it is currently blocking, no? Code comment:
```
// startACSSession starts a session with ACS. It adds request handlers for various
// kinds of messages expected from ACS. It returns on server disconnection or when
// the context is cancelled
```

agent/acs/handler/acs_handler.go
angelcar: > besides, a function starting a goroutine and returning a channel is leak-prone: what happens if the caller exits on canceled context and never receives from the channel?

that's fine. In go, the best practice is that functions that return channels are responsible for closing it at the appropriate time. Methods that consume channels returned from methods are supposed to keep reading from that channel until closed.

agent/acs/handler/acs_handler.go
```diff
@@ -189,70 +189,61 @@ func NewSession(
...
-			acsError := acsSession.startSessionOnce()
-			select {
-			case <-acsSession.ctx.Done():
-				// agent is shutting down, exiting cleanly
+		seelog.Debugf("Attempting connect to ACS")
+		// Start a session with ACS
+		acsError := acsSession.startSessionOnce()
+
+		// If the session is over check for shutdown first
+		if err := acsSession.ctx.Err(); err != nil {
```
sparrc, line 201: I think this is a subtle behavior change, since before we were only checking `ctx.Done()` to trigger just exiting cleanly with a nil error. I believe `ctx.Err()` can return `context.DeadlineExceeded` in addition to `context.Canceled`.

I'm not sure why exactly we would ever receive `context.DeadlineExceeded` from this context in particular, but is there any reason in particular to change the previous behavior?

aws-gibbskt: I was aligning to the previous comment for the function. However I don't think it matters. If we just want to unconditionally exit then I would recommend getting rid of the error return value altogether since it is never provided.

That changes the interface and another test which mocks the interface to test when it returns an error which in reality is never. As can be seen in the tests I added checking the return at all so previously there was no validation of any behavior over time. 

It is a little more change but probably the most correct to remove the return to make it clear that this doesn't intend to do anything but return when the context is Done()

sparrc: > If we just want to unconditionally exit then I would recommend getting rid of the error return value altogether since it is never provided.

I'm not sure if I'm following exactly. This is what we currently do isn't it? ie, in this code here: https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/acs_handler.go#L209-L214

My understanding of this PR is that you're changing from unconditionally exiting whenever the context is done, to conditionally exiting cleanly only when the err returned from the context is context.Canceled.

aws-gibbskt: > // If the context is cancelled, Start() would return with the error code returned
> // by the context.
^ this was what exists and I was trying to actually make true. This seems to be confusing the PR though so I will just remove that comment completely and go back to the previous behavior (even though in actual usage there is no difference since the context in use doesn't have a timer associated so can only return cancelled).

This PR is about removing brittle/unnecessary use of channels in the maintaining of an ACS connection. This https://github.com/aws/amazon-ecs-agent/blob/8dc9fdeb7b876dad609b06001448d0d04e4825fe/agent/app/agent.go#L915 is the only usage of the return value. Cancelled is the standard exit mechanism so printing the error in that case is strange, so I returned nil in that case.

Since outside of a test case with a mock acsSession you can never hit this log line in agent.go I was arguing that the error return value should be removed completely since it isn't used and IF at some point it was used it would probably need thought into what Agent.go did with it such that adding it back into that file at that time would be appropriate.

Will post new rev without these changes in any case. Removing the return value can be a separate PR.

sparrc (approved):

PR #3230, opened 2022-05-27
Consolidate AppNet Agent versioning responsibility

aws-gibbskt:
The AppNet Agent client is reponsible for the model handling of
communicating with the AppNet Agent container, but it should treat
the connection information as configuration coming from the task.
This allows for changes with AppNet Agent versions to be encoded
in the ServiceConnect Manager only. Different http flags or similar
should go with the Container not static with Agent version.

This simply moves the path information back into the Task which
is populated by the ServiceConnect Manager.

### Licensing

### DIFF
--- a/agent/api/task/service_connect.go
+++ b/agent/api/task/service_connect.go
@@ -28,6 +28,10 @@ type ServiceConnectConfig struct {
 type RuntimeConfig struct {
 	// Host path for the administration socket
 	AdminSocketPath string `json:"adminSocketPath"`
+	// HTTP Path + Params to get statistical information
+	StatsRequest string `json:"statsRequest"`
+	// HTTP Path + Params to drain ServiceConnect connections
+	DrainRequest string `json:"drainRequest"`
 }
 
 // IngressConfigEntry is the ingress configuration for a given SC service.
--- a/agent/engine/service_connect/manager_linux.go
+++ b/agent/engine/service_connect/manager_linux.go
@@ -35,8 +35,15 @@ const (
 	defaultStatusFileName     = "appnet_admin.sock"
 	defaultStatusENV          = "APPNET_AGENT_ADMIN_UDS_PATH"
 
+	agentAuthENV   = "ENVOY_ENABLE_IAM_AUTH_FOR_XDS"
+	agentAuthOff   = "0"
 	agentModeENV   = "APPNET_AGENT_ADMIN_MODE"
 	agentModeValue = "uds"
+
+	unixRequestPrefix        = "unix://"
+	httpRequestPrefix        = "http://localhost"
+	defaultAdminStatsRequest = httpRequestPrefix + "/stats/prometheus?usedonly&filter=metrics_extension&delta"
+	defaultAdminDrainRequest = httpRequestPrefix + "/drain_listeners?inboundonly"
 )
 
 type manager struct {
@@ -56,6 +63,11 @@ type manager struct {
 	statusFileName string
 	// Environment variable to set on Container with contents of statusPathContainer/statusFileName
 	statusENV string
+
+	// Http path + params to make a statistics request of AppNetAgent
+	adminStatsRequest string
+	// Http path + params to make a drain request of AppNetAgent
+	adminDrainRequest string
 }
 
 func NewManager() Manager {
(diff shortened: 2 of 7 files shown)


yinyic (approved):

ubhattacharjya (approved):

agent/engine/service_connect/manager_linux.go
```diff
@@ -119,9 +135,10 @@ func (m *manager) initAgentDirectoryMounts(taskId string, container *apicontaine
 
 func (m *manager) initAgentEnvironment(container *apicontainer.Container) {
 	scEnv := map[string]string{
-		m.relayENV:   filepath.Join(m.relayPathContainer, m.relayFileName),
+		m.relayENV:   unixRequestPrefix + filepath.Join(m.relayPathContainer, m.relayFileName),
```
chienhanlin, line 138: Should we update [defaultRelayPathContainer](https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/engine/service_connect/manager_linux.go#L28) to **/var/run/ecs/** as an ENV for AppNet Relay as xDS endpoint is `APPMESH_XDS_ENDPOINT=unix:///var/run/ecs/envoy_xds.sock`, which does not contain **relay** in the path?

aws-gibbskt: Where are you getting that it should not have relay in it? The method that we use to mount the directories it has to be the full current defaultRelayPathContainer.

agent/api/task/service_connect.go
```diff
@@ -28,6 +28,10 @@ type ServiceConnectConfig struct {
 type RuntimeConfig struct {
 	// Host path for the administration socket
 	AdminSocketPath string `json:"adminSocketPath"`
+	// HTTP Path + Params to get statistical information
+	StatsRequest string `json:"statsRequest"`
```
mythri-garaga, line 32: `TestStatsEngineWithServiceConnectMetrics` integ tests are failing since `StatsRequest` is added here, could you please update it so that `StatsRequest` is initialized [here](https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/stats/engine_unix_integ_test.go#L94-L96)? Thank you!

agent/engine/service_connect/manager_linux.go
```diff
@@ -35,8 +35,15 @@ const (
 	defaultStatusFileName     = "appnet_admin.sock"
 	defaultStatusENV          = "APPNET_AGENT_ADMIN_UDS_PATH"
 
+	agentAuthENV   = "ENVOY_ENABLE_IAM_AUTH_FOR_XDS"
+	agentAuthOff   = "0"
```
rajal-amzn, line 39: I would suggest that we set this ENV as part of the payload sent from control plane along with other bootstrap variables.

agent/api/task/service_connect.go
```diff
@@ -28,6 +28,10 @@ type ServiceConnectConfig struct {
...
+	// HTTP Path + Params to drain ServiceConnect connections
+	DrainRequest string `json:"drainRequest"`
```
rajal-amzn, line 34: Currently, the control plane model has these variables removed per earlier discussions. Will add them back.

agent/engine/service_connect/manager_linux.go
aws-gibbskt: I don't think that makes sense. The use of UDS and lack of IAM Auth is function purely of the AppNet Agent Container and the runtime setup that this ECS Agent is doing. ANC has not control nor knowledge of how this might change over time. The way this file is setting up containers requires this flag, if it set it up differently then it might be able to use IAM_AUTH.

agent/api/task/service_connect.go
aws-gibbskt: I don't follow... why would you be adding them back? This isn't a ACS populated field

agent/engine/service_connect/manager_linux.go
chienhanlin: Thanks for the clarification. Also confirm with [jwang186](https://github.com/jwang186) offline that the value is unix:///var/run/ecs/relay/envoy_xds.sock.

aws-gibbskt: Great thanks for the confirmation

agent/api/task/service_connect.go
aws-gibbskt: Thanks that was the missing piece. I updated hopefully appropriately.

agent/api/task/service_connect.go
rajal-amzn: I thought this is from SC properties. Looks like this is a SC config file. You can ignore this.

rajal-amzn (dismissed):

yinyic (approved):

ubhattacharjya (approved):

PR #3232, opened 2022-05-27
Fix flakey leak test for ACS Handler

aws-gibbskt:
### Summary
The multi-channel multi-goroutine implementation of heartbeatHandling
created a scenario where goroutines would be routinely leaked in the
test which was testing for it.

The reality of handling Heartbeat messages is that there was not a
need to do multi-thread synchronization with channels at all.

### Implementation details
The heartbeat handler is now a simple function that dispatches tasks using goroutines and returns.

### Testing
Running the once brittle test 100 times (5 runs) after this change had a success rate of 5/5
`go test -count=100 -p 1 -v -cover -race -tags unit -coverprofile cover.out -run TestHandlerDoesntLeakGoroutines ./agent/...`

Running the test only 100 times (5 runs) before the change had a success rate of: 0/5
`go test -count=100 -p 1 v -cover -race -tags unit -coverprofile cover.out -run TestHandlerDoesntLeakGoroutines ./agent/...`

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -78,8 +78,8 @@ const (
 	acsProtocolVersion = 2
 	// numOfHandlersSendingAcks is the number of handlers that send acks back to ACS and that are not saved across
 	// sessions. We use this to send pending acks, before agent initiates a disconnect to ACS.
-	// they are: refreshCredentialsHandler, taskManifestHandler, payloadHandler and heartbeatHandler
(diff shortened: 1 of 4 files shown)


yinyic (commented):
LGTM. Requested review from fierlion@ the original author, in case there's particular reason we chose to use channels for receiving and sending messages.

sparrc (approved):

chienhanlin (approved):

PR #3239, opened 2022-06-03
Added a minimal ECR image for the signing build environment

YashdalfTheGray:
### Summary
Infrastructure and dockerfile for a minimal signing build environment

### Implementation details
- a new stack that uses CodeBuild to build an image that we can use as the signing build environment
- a dockerfile that specifies only the tooling we need rather than the general purpose build environment that CodeBuild offers by default
- option to use a periodic trigger to build the signing build environment image using EventBridge
- a lifecycle policy to keep ECR images around long enough for us to have a reference to be able to debug but not long enough that they sit around increasing cost
- an option to specify a custom docker image as the signing build environment in `release-pipeline-stack.yml`

### Testing

- built the stack 3 ways
  - stack with all the defaults - no custom image and no periodic trigger
  - stack with custom image and no periodic trigger
  - stack with custom image and periodic trigger

Feature - custom minimal image for the signer stage in pipeline

### Licensing

### DIFF
--- a/build-infrastructure/Dockerfile.signer
+++ b/build-infrastructure/Dockerfile.signer
@@ -1,0 +1,14 @@
+# Pull from Public ECR because less likely to get throttled
+FROM public.ecr.aws/amazonlinux/amazonlinux:2
+
+# Assign a build tag because if we ever run into issues with
+# this image, we'll be able to look back at the image tag
+# and figure out what went wrong
+# To build this image, run
+# docker run -t <your_repo>:<tag_you_want> --build-arg builddate=$(date +%Y%m%d) .
+# For reference date +%Y%m%d outputs today's date in YYYYMMDD format
+# We also tag this image with the same tag in ECR
+ARG builddate
+ENV ECS_AGENT_SIGNING_IMAGE_TAG="build-${builddate}"
+
+RUN yum install -y awscli gpg jq
--- a/build-infrastructure/minimal-signing-build-stack.yml
+++ b/build-infrastructure/minimal-signing-build-stack.yml
@@ -223,7 +223,7 @@ Resources:
       Targets:
         - Arn: !GetAtt ImageCodeBuildProject.Arn
           Id: !Sub 'codebuild-target-${PeriodicBuildTriggerName}'
-          RoleArn: !GetAtt ImageCodeBuildProjectServiceRole.Arn
+          RoleArn: !GetAtt ImageBuildPeriodicTriggerRole.Arn
 
 Outputs:
   EcrRepositoryUri:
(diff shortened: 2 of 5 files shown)


build-infrastructure/Dockerfile.signer
```diff
@@ -0,0 +1,14 @@
+# Pull from Public ECR because less likely to get throttled
+FROM public.ecr.aws/amazonlinux/amazonlinux:2
+
+# Assign a build tag because if we ever run into issues with
+# this image, we'll be able to look back at the image tag
+# and figure out what went wrong
```
fierlion, line 6: (thank you for this.)

build-infrastructure/minimal-signing-build-stack.yml
```diff
@@ -0,0 +1,233 @@
...
+    Description: The branch to use from the repository mentioned above
+    Default: master
+  ImageBuildFrequencyCronExpression:
+    Type: String
+    Description: A cron expression to periodically build the signing image, can be left blank to disable periodic builds. See https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html
+    Default: 30 10 ? * 4 *
+  PeriodicBuildTriggerName:
+    Type: String
+    Description: The name of the CloudWatch Event rule, ignored if ImageBuildFrequencyCronExpression is blank
+    Default: PeriodicallyTriggerSigningImageBuild
```
fierlion, line 39: Can you talk about why you chose a periodic build trigger rather than an event-based trigger?  How often do we expect that a new image should be built -- seems the default will be once per week?  Is the frequency based on Amazon Linux (base image) updates?

fierlion (approved):

build-infrastructure/minimal-signing-build-stack.yml
YashdalfTheGray: I actually tried to figure out a good event based trigger (listing the ECR AL2 image tags and figuring out if something is new) but I couldn't figure out a good event based trigger. But I did want to continue to rebuild this image because I didn't want the docker image to become stale, thus opening us up for unaddressed fixes that do exist in newer versions of the base image. 

I tried to be quicker than the smallest AL2 release cadence that I found recently so I found that it was about a week. 

The idea here was to get something out that does automatically build and we don't have to take on that ops burden too much to cater to our infrastructure. And I'm hoping that as we get a better understanding of a good event-based trigger, we can swap that in for a schedule based trigger.

build-infrastructure/minimal-signing-build-stack.yml
```diff
@@ -0,0 +1,233 @@
...
+    Type: String
+    Description: The name of the log group to store the signing Docker image logs in
+    Default: signer-image-build-logs
+  CodeStarConnectionArn:
+    Type: String
+    Description: The ARN of the connection to use to connect to GitHub
+  GitHubRepositoryUrl:
+    Type: String
+    Description: The repository to pull the Dockerfile from so that we can build the signer image
+    Default: https://github.com/aws/amazon-ecs-agent
```
prateekchaudhry, line 27: Nice to see default values added, makes clear is expected for the parameters

prateekchaudhry (approved):

PR #3249, opened 2022-06-10
Moved ServiceConnect config from task to a new package

yinyic:
### Summary

This PR moves `ServiceConnect` config models and relevant parsing/validation logic into a separate `serviceconnect` package (originally under task package). This is to avoid a potential circular import during CNI plugin integration, where `task` imports `ecscni` and `ecscni` needs ServiceConnect config.

### Testing

Moved ServiceConnect config from task to a new package

### Licensing

### DIFF
--- a/agent/api/serviceconnect/service_connect_attachment_parser.go
+++ b/agent/api/serviceconnect/service_connect_attachment_parser.go
@@ -23,12 +23,20 @@ import (
 )
 
 const (
-	// ServiceConnectConfigKey specifies the key maps to the service connect config in attachment properties
-	ServiceConnectConfigKey = "ServiceConnectConfig"
-	// ServiceConnectContainerNameKey specifies the key maps to the service connect container name in attachment properties
-	ServiceConnectContainerNameKey = "ContainerName"
+	// serviceConnectConfigKey specifies the key maps to the service connect config in attachment properties
+	serviceConnectConfigKey = "ServiceConnectConfig"
+	// serviceConnectContainerNameKey specifies the key maps to the service connect container name in attachment properties
+	serviceConnectContainerNameKey = "ContainerName"
 	keyValidationMsgFormat         = `missing service connect config required key(s) in the attachment: found service connect config key: %t, found service connect container name key: %t`
 )
+
+func GetServiceConnectConfigKey() string {
+	return serviceConnectConfigKey
+}
+
+func GetServiceConnectContainerNameKey() string {
+	return serviceConnectContainerNameKey
+}
 
 // ParseServiceConnectAttachment parses the service connect container name and service connect config value
 // from the given attachment.
@@ -40,7 +48,7 @@ func ParseServiceConnectAttachment(scAttachment *ecsacs.Attachment) (*Config, er
 
 	for _, property := range scAttachment.AttachmentProperties {
 		switch aws.StringValue(property.Name) {
-		case ServiceConnectConfigKey:
+		case serviceConnectConfigKey:
 			foundSCConfigKey = true
 			// extract service connect config value from the attachment property,
 			// and translate the attachment property value to Config
@@ -48,7 +56,7 @@ func ParseServiceConnectAttachment(scAttachment *ecsacs.Attachment) (*Config, er
 			if err := json.Unmarshal([]byte(data), scConfigValue); err != nil {
 				return nil, fmt.Errorf("failed to unmarshal service connect attachment property value: %w", err)
 			}
-		case ServiceConnectContainerNameKey:
+		case serviceConnectContainerNameKey:
 			foundSCContainerNameKey = true
 			// extract service connect container name from the attachment property
 			containerName = aws.StringValue(property.Value)
--- a/agent/api/serviceconnect/service_connect_attachment_parser_test.go
+++ b/agent/api/serviceconnect/service_connect_attachment_parser_test.go
@@ -163,7 +163,7 @@ func getExpectedTestServiceConnectConfig(scContainerName string,
 
 func TestParseServiceConnectAttachment(t *testing.T) {
 	initServiceConnectConfValue()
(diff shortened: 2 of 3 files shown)


agent/api/task/task.go
```diff
@@ -281,7 +283,7 @@ type Task struct {
 	// setIdOnce is used to set the value of this task's id only the first time GetID is invoked
 	setIdOnce sync.Once
 
-	ServiceConnectConfig *ServiceConnectConfig `json:"ServiceConnectConfig,omitempty"`
+	ServiceConnectConfig *serviceconnect.Config `json:"Config,omitempty"`
```
aws-gibbskt, line 286: I don't think we want to change the serialized key with this change.

agent/api/serviceconnect/service_connect_attachment_parser_test.go
```diff
@@ -141,16 +143,16 @@ func getTestACSAttachments(attachmentProperties []*ecsacs.AttachmentProperty) *e
 	return &ecsacs.Attachment{
 		AttachmentArn:        strptr("attachmentArn"),
 		AttachmentProperties: attachmentProperties,
-		AttachmentType:       strptr(serviceConnectAttachmentType),
+		AttachmentType:       strptr(ServiceConnectAttachmentType),
```
aws-gibbskt, line 146: what is driving us to change encapsulation of this Type?

agent/api/task/task.go
yinyic: You are right this is rather unintended. Will fix.

agent/api/serviceconnect/service_connect_attachment_parser_test.go
yinyic: Currently in `task` package when retrieving SC config we go through the list of task attachment and find the one that has `ServiceConnectAttachmentType`, before calling SC parser and validator (see [getServiceConnectConfig](https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/api/task/task.go#L332)). 

I can create getter for it, or alternatively I can move the entire `getServiceConnectConfig` method from task to serviceconnect package.

aws-gibbskt: I think that aspect is very much a "task" thing more than a "serviceconnect" thing. So knowledge of the type name is probably "task" parsing a json blob or something could go in either place

yinyic: Makes sense. I'll move attachment type key back to `task`.

aws-gibbskt (dismissed):

chienhanlin (dismissed):

agent/api/serviceconnect/service_connect_attachment_parser.go
```diff
@@ -23,34 +23,32 @@ import (
...
 
 const (
-	// serviceConnectAttachmentType specifies attachment type for service connect
-	serviceConnectAttachmentType = "ServiceConnect"
-	// serviceConnectConfigKey specifies the key maps to the service connect config in attachment properties
-	serviceConnectConfigKey = "ServiceConnectConfig"
-	// serviceConnectContainerNameKey specifies the key maps to the service connect container name in attachment properties
-	serviceConnectContainerNameKey = "ContainerName"
+	// ServiceConnectConfigKey specifies the key maps to the service connect config in attachment properties
+	ServiceConnectConfigKey = "Config"
```
rajal-amzn, line 27: The config key is coming in from Attachment and the value is still `ServiceConnectConfig`. Why is this changed?

rajal-amzn (changes requested):

agent/api/serviceconnect/service_connect_attachment_parser.go
yinyic: Sorry that's a miss. Will fix.

rajal-amzn (dismissed):

agent/api/task/task_test.go
```diff
@@ -4272,11 +4287,11 @@ func TestTaskServiceConnectAttachment(t *testing.T) {
 						AttachmentArn: strptr("attachmentArn"),
 						AttachmentProperties: []*ecsacs.AttachmentProperty{
 							{
-								Name:  strptr(serviceConnectConfigKey),
+								Name:  strptr(serviceconnect.ServiceConnectConfigKey),
 								Value: strptr(tc.testSCConfigValue),
 							},
 							{
-								Name:  strptr(serviceConnectContainerNameKey),
+								Name:  strptr(serviceconnect.ServiceConnectContainerNameKey),
```
aws-gibbskt, line 4294: This test usage feels like a week reason to adjust visibility of these values with the s vs S beginning

yinyic: Added getters. I thought about hardcoding them in this test file but I'm not sure if I like the idea that changing the internal service connect config keys will break this unit test when it shouldn't really matter.

aws-gibbskt: That works, or you can have a function accessible in the test realm. Specifically `service_connect_attachment_parser_test.go` can have a helper `CreateServiceConnectAttachment` or something with arguments that can be passed in and returning a correctly key'd map to merge here. That function only exists in test and doesn't require getters or exporting and can provide a reminder when changing things in the attachment parser that you have to support that use-case

yinyic: I couldn't seem to get it to work. It appears that test files are compiled separately for each package. According to https://pkg.go.dev/cmd/go/internal/test -`"Each listed package causes the execution of a separate test binary."` For that reason, test files could not access even exported functions from a test file in another package.

aws-gibbskt: TIL - thanks.

aws-gibbskt (approved):

rajal-amzn (approved):

chienhanlin (approved):

ubhattacharjya (approved):

PR #3251, opened 2022-06-13
Build VPC CNI plugin without using vendor directory

yinyic:
### Summary

We use aws-vpc-cni-plugin [repo](https://github.com/aws/amazon-vpc-cni-plugins) as a submodule. Currently when building the plugins we use the dependencies in the repo's vendor directory by specifying `-mod=vendor` build flag. This requires that the repo maintains an up-to-date vendor directory. 

However, since the repo has migrated to use go modules, the vendor directory is not automatically updated, and is currently out-of-sync with the actual dependency versions in go.mod file. This PR removes the `-mod=vendor` flag so that `go build` will download the dependencies directly, and we are no longer vulnerable to out-of-sync vendored deps.

### Implementation details

- Remove `-mod=vendor` flag for the build
- Add `GOPROXY=https://proxy.golang.org|direct` env var for the build to download dependency packages. By default, `GOPROXY="https://proxy.golang.org,direct"`. `go` will interpret it as a comma-separated list, and try the next candidate in case of 404 or 410 error. We want to use '|' instead, so that regardless of the error (such as when `proxy.golang.org` domain is blocked), the next candidate will always be tried. 
- Add `GIT_TAG` env var to be consumed by the [build](https://github.com/aws/amazon-vpc-cni-plugins/blob/master/Makefile#L26). Currently resolving git tag throws error because the docker work directory during submodule build is not a git directory so we can't run git commands there.

### Testing

Ran `make cni-plugins` and verified output `Built all cni plugins successfully.`

Buid VPC CNI plugin without using vendor directory

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,9 +222,10 @@ build-ecs-cni-plugins:
 
 build-vpc-cni-plugins:
 	@docker build --build-arg GOARCH=$(GOARCH) --build-arg GO_VERSION=$(GO_VERSION) -f $(VPC_CNI_REPOSITORY_DOCKER_FILE) -t "amazon/amazon-ecs-build-vpc-cni-plugins:make" .
-	docker run --rm --net=none \
-		-e GO111MODULE=off \
+	docker run --rm --net=host \
+		-e GO111MODULE=auto \
 		-e GIT_SHORT_HASH=$(shell cd $(VPC_CNI_REPOSITORY_SRC_DIR) && git rev-parse --short=8 HEAD) \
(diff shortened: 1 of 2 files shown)


singholt (approved):

chienhanlin (approved):

yinyic:
Windows build is failing due to missing deps in go.mod. This can be fixed if we pull in latest commit for the vpc-cni-plugin submodule. I'm closing this PR in favor of https://github.com/aws/amazon-ecs-agent/pull/3255

PR #3253, opened 2022-06-14
Bug Fix - Check the size of container metrics size when the container instance runs a mix of SC enabled and not enabled tasks

ubhattacharjya:
Check the size of container metrics size when the container when the container instance runs a mix of SC enabled and not enabled tasks

### Summary

This PR fixes 2 bugs
Bug 1: For non service connect enabled tasks, we were not adding docker metrics for that task. This is addressed in [agent/stats/engine.go]

Bug 2: When a container instance will run a combination of service connect and non service connect tasks, for non SC tasks, we were not checking if the docker metrics exceeded the frame size limit. This check needs to be done for SC tasks also.

### Testing

New tests cover the changes: 

Manual tests: Tested by running an appmesh app that if for a task, container metrics exceeded size, we are creating a new request to send them. Also tested that for non SC tasks, docker metrics are correctly getting added.

### Licensing

### DIFF
--- a/agent/tcs/client/client.go
+++ b/agent/tcs/client/client.go
@@ -237,7 +237,21 @@ func (cs *clientServer) metricsToPublishMetricRequests(includeServiceConnectStat
 	for i, taskMetric := range taskMetrics {
 		requestMetadata = copyMetricsMetadata(metadata, false)
 
-		// TODO [SC]: Check if SC is enabled for the task
+		// Check if taskMetric without service connect metrics exceed the message size
+		tempTaskMetric := *taskMetric
+		tempTaskMetric.ServiceConnectMetricsWrapper = tempTaskMetric.ServiceConnectMetricsWrapper[:0]
+
+		messageTaskMetrics = append(messageTaskMetrics, &tempTaskMetric)
+		tmsg, _ := jsonutil.BuildJSON(ecstcs.NewPublishMetricsRequest(requestMetadata, copyTaskMetrics(messageTaskMetrics)))
+		// remove the tempTaskMetric added to messageTaskMetrics after creating tempMessage
+		messageTaskMetrics = messageTaskMetrics[:len(messageTaskMetrics)-1]
+		if len(tmsg) > publishMetricRequestSizeLimit {
+			// Create a new request as the current task metric if added is exceeding the size of the frame.
+			requests = append(requests, ecstcs.NewPublishMetricsRequest(requestMetadata, copyTaskMetrics(messageTaskMetrics)))
+			// reset the messageTaskMetrics for the new request
+			messageTaskMetrics = messageTaskMetrics[:0]
+		}
+
 		if includeServiceConnectStats {
 			taskMetric, messageTaskMetrics, requests = cs.serviceConnectMetricsToPublishMetricRequests(requestMetadata, taskMetric, messageTaskMetrics, requests)
 		}
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -465,17 +465,13 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 		}
 
 		if includeServiceConnectStats {
-			serviceConnectStats, ok := engine.taskToServiceConnectStats[taskArn]
-			if !ok {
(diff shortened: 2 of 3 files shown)


agent/tcs/client/client.go
```diff
@@ -270,6 +268,17 @@ func (cs *clientServer) serviceConnectMetricsToPublishMetricRequests(requestMeta
 	tempTaskMetric := *taskMetric
 	tempTaskMetric.ServiceConnectMetricsWrapper = tempTaskMetric.ServiceConnectMetricsWrapper[:0]
 
+	// Check if taskMetric without service connect metrics exceed the message size
```
yinyic, line 271: If taskMetric could exceed max size without even adding SC metrics, should we move this block to common logic inside `metricsToPublishMetricRequets` before calling `serviceConnectMetricsToPublishMetricRequests`, so that we make this check regardless of whether we add SC metrics or not? 

And for my understanding, does this mean the issue could be happening at this moment (regardless of whether Agent supports SC or not)? Trying to see if we want to this fix should go to dev directly.

ubhattacharjya: We send sc stats every 60s and we send docker stats every 20s. The limit is either 10 tasks or 1 MB frame size limit. Without SC stats, it is exactly same as before, it should be 10 tasks per frame. That is why have this check whenever we are collecting SC stats. Does that make sense?

mythri-garaga (dismissed):

agent/tcs/client/client.go
ubhattacharjya: Thought more about it and made the change. This will be helpful if in future we add more metrics to the docker stats metrics, we will always have that check that for 10 tasks , we are not exceeding frame size limit.

agent/tcs/client/client.go
```diff
@@ -237,7 +237,21 @@ func (cs *clientServer) metricsToPublishMetricRequests(includeServiceConnectStat
 	for i, taskMetric := range taskMetrics {
 		requestMetadata = copyMetricsMetadata(metadata, false)
 
-		// TODO [SC]: Check if SC is enabled for the task
+		// Check if taskMetric without service connect metrics exceed the message size
+		tempTaskMetric := *taskMetric
+		tempTaskMetric.ServiceConnectMetricsWrapper = tempTaskMetric.ServiceConnectMetricsWrapper[:0]
```
yinyic, line 242: Should we only assign this if `includeServiceConnectStats` is true?

yinyic: Synced offline - we are setting it to an empty slice here to make sure the size calculation is pre-SC.

yinyic (approved):

agent/tcs/client/client.go
```diff
@@ -237,7 +237,21 @@ func (cs *clientServer) metricsToPublishMetricRequests(includeServiceConnectStat
...
+		messageTaskMetrics = messageTaskMetrics[:len(messageTaskMetrics)-1]
+		if len(tmsg) > publishMetricRequestSizeLimit {
+			// Create a new request as the current task metric if added is exceeding the size of the frame.
+			requests = append(requests, ecstcs.NewPublishMetricsRequest(requestMetadata, copyTaskMetrics(messageTaskMetrics)))
+			// reset the messageTaskMetrics for the new request
+			messageTaskMetrics = messageTaskMetrics[:0]
+		}
+
 		if includeServiceConnectStats {
 			taskMetric, messageTaskMetrics, requests = cs.serviceConnectMetricsToPublishMetricRequests(requestMetadata, taskMetric, messageTaskMetrics, requests)
```
mythri-garaga, line 256: For a SC task, if the container metrics were sent in line 250, it will be sent again from the `serviceConnectMetricsToPublishMetricRequests()` method.

mythri-garaga: Synced offline, if adding the taskMetric(with container metrics) exceeds the size limit it will be removed before sending the payload(on line 247).

mythri-garaga (approved):

agent/tcs/client/client.go
ubhattacharjya: Syned offline. container metrics are not sent twice for a task.

Issue #74, opened 2015-05-18
"Credential should be scoped to a valid region, not 'eu-west-1'. "


seiffert:
Hi,

I'm using the agent in a Docker container (using the official `amazon/amazon-ecs-agent` image from Docker Hub) on a CoreOS EC2 instance. (see the systemd unit file below) Right now, I'm having problems having the agents reliably register the instances at ECS. I can't really tell when exactly this happens, but in one out of three starts of the agent, it keeps logging the following lines and does not report the agent to be connected in the ECS Console.

```
2015-05-18T11:34:33Z [ERROR] Unable to discover poll endpoint module="acs handler" err="Credential should be scoped to a valid region, not 'eu-west-1'. "
2015-05-18T11:34:33Z [INFO] Error from acs; backing off module="acs handler" err="Credential should be scoped to a valid region, not 'eu-west-1'. "
2015-05-18T11:36:47Z [ERROR] Unable to discover poll endpoint module="acs handler" err="Credential should be scoped to a valid region, not 'eu-west-1'. "
2015-05-18T11:36:47Z [INFO] Error from acs; backing off module="acs handler" err="Credential should be scoped to a valid region, not 'eu-west-1'. "
2015-05-18T11:39:09Z [ERROR] Unable to discover poll endpoint module="acs handler" err="Credential should be scoped to a valid region, not 'eu-west-1'. "
2015-05-18T11:39:09Z [INFO] Error from acs; backing off module="acs handler" err="Credential should be scoped to a valid region, not 'eu-west-1'. "
```

The Systemd unit file I'm using is:

```
[Unit]
Description=The AWS ECS agent
After=docker.service
Requires=docker.service
Type=service
[Service]
TimeoutStartSec=0
TimeoutStopSec=0
Restart=on-failure
SyslogIdentifierg=ecs-agent
ExecStartPre=-/bin/mkdir -p /var/log/ecs /var/ecs-data
ExecStartPre=-/usr/bin/docker stop ecs-agent
ExecStartPre=-/usr/bin/docker pull amazon/amazon-ecs-agent
ExecStartPre=-/usr/bin/docker rm ecs-agent
[...]

PR #3255, opened 2022-06-15
Integrate with Service Connect AWSVPC CNI plugin

yinyic:
### Summary

ServiceConnect-enabled tasks require network namespace configurations that are performed by a new CNI plugin. This PR integrates with the new CNI plugin `ecs-serviceconnect` for AWSVPC mode. 

### Implementation details

If a task is AWSVPC mode and SC-enabled, build and append ServiceConnect network configuration such that the `ecs-serviceconnect` CNI plugin will be invoked with the supplied information. 

The configuration includes:
- ingress and egress info which comes from task payload (except the ephemeral ports which are selected by Agent)
- flags to indicate whether task is IPv4- and/or IPv6-enabled

The PR also upgrades the submodule commit for `amazon-vpc-cni-plugins` to pick up the new SC plugin.

### Testing

Built test Agent and ran a test SC task - verified netns configuration for both default and non-default SC experience in AWSVPC network mode.

For CNI plugin build, ran `make cni-plugins` for both `TARGET_OS=linux` and `TARGET_OS=windows` - verified all CNI plugins were built successfully.

Integrate with Service Connect AWSVPC CNI plugin

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -222,9 +222,10 @@ build-ecs-cni-plugins:
 
 build-vpc-cni-plugins:
 	@docker build --build-arg GOARCH=$(GOARCH) --build-arg GO_VERSION=$(GO_VERSION) -f $(VPC_CNI_REPOSITORY_DOCKER_FILE) -t "amazon/amazon-ecs-build-vpc-cni-plugins:make" .
-	docker run --rm --net=none \
-		-e GO111MODULE=off \
+	docker run --rm --net=host \
+		-e GO111MODULE=auto \
 		-e GIT_SHORT_HASH=$(shell cd $(VPC_CNI_REPOSITORY_SRC_DIR) && git rev-parse --short=8 HEAD) \
+		-e GIT_TAG=$(shell cd $(VPC_CNI_REPOSITORY_SRC_DIR) && git describe --tags --always --dirty) \
 		-u "$(USERID)" \
 		-v "$(PWD)/out/amazon-vpc-cni-plugins:/go/src/github.com/aws/amazon-vpc-cni-plugins/build/${TARGET_OS}_$(GOARCH)" \
 		-v "$(VPC_CNI_REPOSITORY_SRC_DIR):/go/src/github.com/aws/amazon-vpc-cni-plugins" \
--- a/Makefile
+++ b/Makefile
@@ -222,8 +222,8 @@ build-ecs-cni-plugins:
 
 build-vpc-cni-plugins:
 	@docker build --build-arg GOARCH=$(GOARCH) --build-arg GO_VERSION=$(GO_VERSION) -f $(VPC_CNI_REPOSITORY_DOCKER_FILE) -t "amazon/amazon-ecs-build-vpc-cni-plugins:make" .
-	docker run --rm --net=host \
-		-e GO111MODULE=auto \
+	docker run --rm --net=none \
+		-e GO111MODULE=off \
 		-e GIT_SHORT_HASH=$(shell cd $(VPC_CNI_REPOSITORY_SRC_DIR) && git rev-parse --short=8 HEAD) \
 		-e GIT_TAG=$(shell cd $(VPC_CNI_REPOSITORY_SRC_DIR) && git describe --tags --always --dirty) \
 		-u "$(USERID)" \
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -335,6 +335,22 @@ func (task *Task) BuildCNIConfig(includeIPAMConfig bool, cniConfig *ecscni.Confi
 		})
 	}
 
+	// Build a CNI network configuration for ServiceConnect-enabled task in AWSVPC mode
+	if task.IsServiceConnectEnabled() {
+		ifName, netconf, err = ecscni.NewServiceConnectNetworkConfig(
+			task.ServiceConnectConfig,
+			task.shouldEnableIPv4(),
+			task.shouldEnableIPv6(),
+			cniConfig)
+		if err != nil {
+			return nil, err
+		}
+		cniConfig.NetworkConfigs = append(cniConfig.NetworkConfigs, &ecscni.NetworkConfig{
+			IfName:           ifName,
+			CNINetworkConfig: netconf,
+		})
+	}
+
 	cniConfig.ContainerNetNS = fmt.Sprintf(ecscni.NetnsFormat, cniConfig.ContainerPID)
 
 	return cniConfig, nil
--- a/agent/ecscni/netconfig.go
+++ b/agent/ecscni/netconfig.go
@@ -16,7 +16,8 @@ package ecscni
 import (
 	"encoding/json"
 
-	"github.com/cihub/seelog"
+	"github.com/aws/amazon-ecs-agent/agent/logger"
+
 	"github.com/containernetworking/cni/libcni"
 	cnitypes "github.com/containernetworking/cni/pkg/types"
 )
@@ -25,7 +26,11 @@ import (
 func newNetworkConfig(netcfg interface{}, plugin string, cniVersion string) (*libcni.NetworkConfig, error) {
 	configBytes, err := json.Marshal(netcfg)
 	if err != nil {
-		seelog.Errorf("[ECSCNI] Marshal configuration for plugin %s failed, error: %v", plugin, err)
+		logger.Error("[ECSCNI] Marshal configuration failed", logger.Fields{
+			"netcfg":     netcfg,
+			"plugin":     plugin,
+			"cniVersion": cniVersion,
+		})
 		return nil, err
 	}
 
--- a/agent/ecscni/netconfig_linux.go
+++ b/agent/ecscni/netconfig_linux.go
@@ -19,6 +19,8 @@ package ecscni
 import (
 	"net"
 
+	"github.com/aws/amazon-ecs-agent/agent/api/serviceconnect"
+
 	"github.com/aws/amazon-ecs-agent/agent/api/appmesh"
 	"github.com/aws/amazon-ecs-agent/agent/api/eni"
 
@@ -169,4 +171,47 @@ func NewAppMeshConfig(appMesh *appmesh.AppMesh, cfg *Config) (string, *libcni.Ne
 	}
 
 	return defaultAppMeshIfName, networkConfig, nil
+}
+
+// NewServiceConnectNetworkConfig creates a new ServiceConnect CNI network configuration
+func NewServiceConnectNetworkConfig(
+	scConfig *serviceconnect.Config,
+	enableIPv4, enableIPv6 bool,
+	cfg *Config) (string, *libcni.NetworkConfig, error) {
+	var ingressConfig []IngressConfigJSONEntry
+	for _, ic := range scConfig.IngressConfig {
+		newEntry := IngressConfigJSONEntry{
+			ListenerPort: ic.ListenerPort,
+		}
+		if ic.InterceptPort != nil {
+			newEntry.InterceptPort = *ic.InterceptPort
+		}
+		ingressConfig = append(ingressConfig, newEntry)
+	}
+
+	var egressConfig *EgressConfigJSON
+	if scConfig.EgressConfig != nil {
+		egressConfig = &EgressConfigJSON{
+			ListenerPort: scConfig.EgressConfig.ListenerPort,
+			VIP: VIPConfigJSON{
+				IPv4CIDR: scConfig.EgressConfig.VIP.IPV4CIDR,
+				IPv6CIDR: scConfig.EgressConfig.VIP.IPV6CIDR,
+			},
+		}
+	}
+
+	scNetworkConfig := ServiceConnectConfig{
+		Name:          ECSServiceConnectPluginName,
+		Type:          ECSServiceConnectPluginName,
+		IngressConfig: ingressConfig,
+		EgressConfig:  egressConfig,
+		EnableIPv4:    enableIPv4,
+		EnableIPv6:    enableIPv6,
+	}
+	networkConfig, err := newNetworkConfig(scNetworkConfig, ECSServiceConnectPluginName, cfg.MinSupportedCNIVersion)
+	if err != nil {
+		return "", nil, errors.Wrap(err, "NewServiceConnectNetworkConfig: construct the service connect network configuration failed")
+	}
+
+	return defaultServiceConnectIfName, networkConfig, nil
 }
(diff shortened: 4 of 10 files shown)


agent/api/task/task_linux.go
```diff
@@ -339,6 +339,22 @@ func (task *Task) BuildCNIConfig(includeIPAMConfig bool, cniConfig *ecscni.Confi
...
+		ifName, netconf, err = ecscni.NewServiceConnectNetworkConfig(
+			task.ServiceConnectConfig,
+			task.shouldEnableIPv4(),
+			task.shouldEnableIPv6(),
+			cniConfig)
+		if err != nil {
+			return nil, err
+		}
+		cniConfig.NetworkConfigs = append(cniConfig.NetworkConfigs, &ecscni.NetworkConfig{
+			IfName:           ifName,
```
rajal-amzn, line 353: Which interface is set here?

agent/ecscni/netconfig_linux.go
```diff
@@ -169,3 +171,46 @@ func NewAppMeshConfig(appMesh *appmesh.AppMesh, cfg *Config) (string, *libcni.Ne
...
+// NewServiceConnectNetworkConfig creates a new ServiceConnect CNI network configuration
+func NewServiceConnectNetworkConfig(
+	scConfig *serviceconnect.Config,
+	enableIPv4, enableIPv6 bool,
+	cfg *Config) (string, *libcni.NetworkConfig, error) {
+	var ingressConfig []IngressConfigJSONEntry
+	for _, ic := range scConfig.IngressConfig {
+		newEntry := IngressConfigJSONEntry{
+			ListenerPort: ic.ListenerPort,
+		}
```
rajal-amzn, line 184: Why cant we simply copy the Ingress Config entry?

agent/ecscni/netconfig_linux.go
```diff
@@ -169,3 +171,46 @@ func NewAppMeshConfig(appMesh *appmesh.AppMesh, cfg *Config) (string, *libcni.Ne
...
+		egressConfig = &EgressConfigJSON{
+			ListenerPort: scConfig.EgressConfig.ListenerPort,
+			VIP: VIPConfigJSON{
+				IPv4CIDR: scConfig.EgressConfig.VIP.IPV4CIDR,
+				IPv6CIDR: scConfig.EgressConfig.VIP.IPV6CIDR,
+			},
+		}
+	}
+
+	scNetworkConfig := ServiceConnectConfig{
```
rajal-amzn, line 202: We also need to add `NetworkMode`

agent/ecscni/types_linux.go
```diff
@@ -28,6 +28,9 @@ const (
 	// defaultAppMeshIfName is the default name of app mesh to setup iptable rules
 	// for app mesh container. IfName is mandatory field to invoke CNI plugin.
 	defaultAppMeshIfName = "aws-appmesh"
+	// defaultServiceConnectIfName is the default ifname used for invoking SC CNI plugin.
+	// IfName is a mandatory field to invoke any CNI plugin. But it's not actually used in this case.
+	defaultServiceConnectIfName = "ecs-serviceconnect"
```
rajal-amzn, line 33: Why cant we actually send the interface name available to this container? The interface name is not used in awsvpc but it would be for bridge.

agent/api/task/task_linux.go
yinyic: Resolving as dup of the [this comment](https://github.com/aws/amazon-ecs-agent/pull/3255#discussion_r899356403)

agent/ecscni/netconfig_linux.go
yinyic: Not sure I follow - the `IngressConfigJSONEntry` is a different struct from `scConfig.IngressConfig` entries. Former is the internal model for CNI while latter is the SC API model.

agent/ecscni/netconfig_linux.go
yinyic: Do you mean `RedirectMode`? Will do once the CNI plugin PR is merged. The CNI version contained in this PR does not have that `RedirectMode` yet.

agent/ecscni/types_linux.go
yinyic: I saw the CNI plugin CR using `eth0` (which indeed is the default ifname for docker bridge netns) so I didn't add anything here. I can add it.

agent/ecscni/types_linux.go
```diff
@@ -159,3 +164,39 @@ type BranchENIConfig struct {
 	// InterfaceType is the type of the interface to connect the branch ENI to
 	InterfaceType string `json:"interfaceType,omitempty"`
 }
+
+type ServiceConnectConfig struct {
```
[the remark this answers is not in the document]
yinyic: I believe this is the recommended way actually, according to https://go.dev/doc/effective_go#package-names
> The importer of a package will use the name to refer to its contents, so exported names in the package can use that fact to avoid repetition

In our case, the package name `ecscni` already implies the namespace, and other packages will be referencing the type as `ecscni.ServiceConnectConfig`.

agent/ecscni/types_linux.go
rajal-amzn: Yes I changed it to use `ifName` in the latest revision.

agent/ecscni/netconfig_linux.go
```diff
@@ -169,3 +171,46 @@ func NewAppMeshConfig(appMesh *appmesh.AppMesh, cfg *Config) (string, *libcni.Ne
...
+		Name:          ECSServiceConnectPluginName,
+		Type:          ECSServiceConnectPluginName,
+		IngressConfig: ingressConfig,
+		EgressConfig:  egressConfig,
+		EnableIPv4:    enableIPv4,
+		EnableIPv6:    enableIPv6,
+	}
+	networkConfig, err := newNetworkConfig(scNetworkConfig, ECSServiceConnectPluginName, cfg.MinSupportedCNIVersion)
+	if err != nil {
+		return "", nil, errors.Wrap(err, "NewServiceConnectNetworkConfig: construct the service connect network configuration failed")
```
chienhanlin, line 212: Borrow wisdom from Angel ([source](https://github.com/aws/amazon-ecs-agent/pull/3208#discussion_r870645566))
>nit: the preferred way to wrap errors is now fmt.Errorf("message: %w", err)
github.com/pkg/errors is a bit of a legacy package now

agent/ecscni/plugin_linux_test.go
```diff
@@ -231,6 +233,74 @@ func appMeshNetworkConfig(config *Config) *NetworkConfig {
...
+func serviceConnectNetworkConfig(config *Config) *NetworkConfig {
+	_, serviceConnectNetworkConfig, _ := NewServiceConnectNetworkConfig(defaultTestServiceConnectConfig(), true, false, config)
+	return &NetworkConfig{CNINetworkConfig: serviceConnectNetworkConfig}
+}
+
+func defaultTestServiceConnectConfig() *serviceconnect.Config {
+	return &serviceconnect.Config{
+		IngressConfig: []serviceconnect.IngressConfigEntry{{
+			ListenerName: "test ingress listener",
+			ListenerPort: 11111,
```
chienhanlin, line 290: nit. maybe we can define 
```
const (
  testIngressListenerPort = uint16(11111)
  testEgressConfigListenerPort = uint16(22222)
  ....
)
```
and reuse then in test validation steps.

Makefile
```diff
@@ -209,9 +209,10 @@ build-ecs-cni-plugins:
 
 build-vpc-cni-plugins:
 	@docker build --build-arg GOARCH=$(GOARCH) --build-arg GO_VERSION=$(GO_VERSION) -f $(VPC_CNI_REPOSITORY_DOCKER_FILE) -t "amazon/amazon-ecs-build-vpc-cni-plugins:make" .
-	docker run --rm --net=none \
-		-e GO111MODULE=off \
+	docker run --rm --net=host \
```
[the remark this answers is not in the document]
yinyic: We need network to download the dependencies. 

According to https://go.dev/ref/mod#mod-commands

> In module-aware mode, the go command uses go.mod files to find versioned dependencies, and it typically loads packages out of the [module cache](https://go.dev/ref/mod#glos-module-cache), downloading modules if they are missing. 

We used to use vendor directory before that's why it does not require network. 

> In GOPATH mode, the go command ignores modules; it looks in [vendor directories](https://go.dev/ref/mod#glos-vendor-directory) and in GOPATH to find dependencies.

I'm making the change because the source amazon-vpc-cni-plugins [repo](https://github.com/aws/amazon-vpc-cni-plugins) also builds in module-aware mode. Although it has a vendor directory, the dependency version has been out-of-sync. 

If we believe the need for internet during a build should be avoided as much as possible, we can go back to using vendor directory by asking the CNI plugin maintainers to keep the vendor directory up to date.

agent/ecscni/types_linux.go
yinyic: Cool, will address in next revision.

agent/ecscni/netconfig_linux.go
yinyic: Sure will address in next rev

agent/ecscni/netconfig_linux.go
```diff
@@ -169,3 +171,46 @@ func NewAppMeshConfig(appMesh *appmesh.AppMesh, cfg *Config) (string, *libcni.Ne
...
+
+	scNetworkConfig := ServiceConnectConfig{
+		Name:          ECSServiceConnectPluginName,
+		Type:          ECSServiceConnectPluginName,
+		IngressConfig: ingressConfig,
+		EgressConfig:  egressConfig,
+		EnableIPv4:    enableIPv4,
+		EnableIPv6:    enableIPv6,
+	}
+	networkConfig, err := newNetworkConfig(scNetworkConfig, ECSServiceConnectPluginName, cfg.MinSupportedCNIVersion)
```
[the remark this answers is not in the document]
yinyic: Sure will update.

agent/ecscni/plugin_linux_test.go
yinyic: Yeah sure!

agent/ecscni/netconfig.go
```diff
@@ -25,7 +26,11 @@ import (
 func newNetworkConfig(netcfg interface{}, plugin string, cniVersion string) (*libcni.NetworkConfig, error) {
 	configBytes, err := json.Marshal(netcfg)
 	if err != nil {
-		seelog.Errorf("[ECSCNI] Marshal configuration for plugin %s failed, error: %v", plugin, err)
+		logger.Error("[ECSCNI] Marshal configuration failed", logger.Fields{
```
chienhanlin, line 29: Thanks!

chienhanlin (dismissed):

Makefile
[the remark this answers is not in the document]
rajal-amzn: I would suggest we go and fix CNI plugin to keep the vendor directory up to date.

agent/ecscni/netconfig_linux.go
rajal-amzn: I was suggesting that we could directly use the scConfig.IngressConfig given it only has some additional fields and they would be ignored anywhere. But given that the number of entries in here will be less. it might not matter.

yinyic: Yeah I think I'd rather keep it separate. The EgressConfig that we pass to CNI plugin will have to be different than the API model anyway (to include `RedirectMode` for example), so it's probably better to keep IngressConfig separate as well.

Makefile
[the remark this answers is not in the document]
yinyic: Chatted with @ofiliz, maintainer of `amazon-vpc-cni-plugins` repo - Onur is going to handpick all necessary dependencies because running `go mod vendor` brings a much larger dependency set than what we actually need and is undesirable.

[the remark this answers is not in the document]
yinyic: https://github.com/aws/amazon-vpc-cni-plugins/pull/73

aws-gibbskt (approved):

agent/ecscni/types_linux.go
```diff
@@ -29,6 +29,11 @@ const (
 	// defaultAppMeshIfName is the default name of app mesh to setup iptable rules
 	// for app mesh container. IfName is mandatory field to invoke CNI plugin.
 	defaultAppMeshIfName = "aws-appmesh"
+	// defaultServiceConnectIfName is the default ifname used for invoking SC CNI plugin.
+	// IfName is a mandatory field to invoke any CNI plugin. For tasks in awsvpc mode, the network configuration
+	// does not need the ifname so this is simply a placeholder. For tasks in bridge mode, the configuration does
+	// require the ifname, and we use "eth0" because that's the default interface for docker bridge container network.
+	defaultServiceConnectIfName = "eth0"
```
ofiliz, line 36: I don't think this is accurate. There should be no interface names hardcoded in CNI plugins or agent. We have removed the use of ifName from ServiceConnect CNI plugin.

rajal-amzn: That is right. We dont use the ifName within the plugin.

yinyic: Okay will update.

agent/ecscni/plugin_test.go
```diff
@@ -30,7 +30,7 @@ const (
 	// ECSCNIVersion, ECSCNIGitHash, VPCCNIGitHash needs to be updated every time CNI plugin is updated.
 	currentECSCNIVersion = "2020.09.0"
 	currentECSCNIGitHash = "55b2ae77ee0bf22321b14f2d4ebbcc04f77322e1"
-	currentVPCCNIGitHash = "199bfc65cced4951cbb6a38e6e828afa8c2b023c"
+	currentVPCCNIGitHash = "c3a89823446f312edfbed8375a0394773e39f807"
```
aws-gibbskt, line 33: 24d6bd87707d1b1801086fc507ebab8d32067412 this is the latest after `Adding a new mode Tproxy to do the redirection of traffic for ServiceConnect Plugin (#74)`

yinyic: The latest CNI plugin repo commit updated request model which is incompatible with the code in this PR. In my follow-up PR for bridge mode, both the model and the CNI git hash will be updated. I wanted to keep PR to be a simple rebase if possible.

rajal-amzn (approved):

aws-gibbskt (approved):

chienhanlin (approved):

ofiliz (approved):

ubhattacharjya (approved):

PR #3259, opened 2022-06-21
Adding support to launch Relay and assign task container images

aws-gibbskt:
### Summary

This CR finalizes the process of starting a Relay Container and auto assigning images to task containers.

### Implementation details
Create a new task with an internal container which is run in host mode on the instance.

Update container images for task containers for service connect.

### Licensing


agent/api/task/task.go
```diff
@@ -536,9 +536,11 @@ func (task *Task) initServiceConnectResources() error {
 			ContainerName: "service-connect",
 		}
 	}
+	// TODO [SC]: End dev/testing purposes
```
sparrc, line 539: What are these TODOs? can you describe in more detail? it's not clear what still needs to be done here or why it can't be done now already

agent/app/agent.go
```diff
@@ -134,6 +135,7 @@ type ecsAgent struct {
 	stateManagerFactory         factory.StateManager
 	saveableOptionFactory       factory.SaveableOption
 	pauseLoader                 pause.Loader
+	appNetLoader                serviceconnect.Loader
```
sparrc, line 138: nit: appNetLoader should be called serviceConnectLoader to match package naming

agent/app/agent.go
```diff
@@ -488,18 +491,19 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 	credentialsManager credentials.Manager,
 	state dockerstate.TaskEngineState,
 	imageManager engine.ImageManager,
-	execCmdMgr execcmd.Manager) (engine.TaskEngine, string, error) {
+	execCmdMgr execcmd.Manager,
+	appnetAgentloader serviceconnect.Loader) (engine.TaskEngine, string, error) {
```
sparrc, line 495: ditto the same naming nit here, seems like var name should match the package name (serviceConnectLoader)

agent/app/agent_capability.go
```diff
@@ -271,6 +274,11 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 	if err != nil {
 		return nil, err
 	}
+	// add service-connect capabilities if applicable
+	capabilities, err = agent.appendServiceConenctCapabilities(capabilities)
```
sparrc, line 278: spelling: Conenct -> Connect

agent/app/agent_capability.go
```diff
@@ -438,6 +446,18 @@ func (agent *ecsAgent) appendExecCapabilities(capabilities []*ecs.Attribute) ([]
 	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityExec), nil
 }
 
+func (agent *ecsAgent) appendServiceConenctCapabilities(capabilities []*ecs.Attribute) ([]*ecs.Attribute, error) {
+	if loaded, _ := agent.appNetLoader.IsLoaded(agent.dockerClient); !loaded {
+		_, err := agent.appNetLoader.LoadImage(agent.ctx, agent.dockerClient)
+		if err != nil {
+			seelog.Warn("ServiceConnect Capability: Failed to load appnet Agent container")
```
sparrc, line 453: I would maybe log an ERROR message here and describe the issue that it will cause, and include the err in the error message (if possible also using the agent's structured logging package which it looks like this file doesn't currently use):
```go
import (
	"github.com/aws/amazon-ecs-agent/agent/logger/field"
)

...

logger.Error("ServiceConnect Capability: Failed to load appnet Agent container. This container instance will not be able to support ServiceConnect tasks", logger.Fields{field.Error: err})
return capabilities, nil
```

agent/engine/docker_task_engine.go
```diff
@@ -50,8 +47,10 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dependencygraph"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/engine/execcmd"
+	serviceconnect "github.com/aws/amazon-ecs-agent/agent/engine/service_connect"
```
sparrc, line 50: another naming nit: name these imports so it somewhat matches the file path:
```
engineserviceconnect "github.com/aws/amazon-ecs-agent/agent/engine/service_connect"
serviceconnect "github.com/aws/amazon-ecs-agent/agent/serviceconnect"
```

agent/app/agent.go
sparrc: or maybe `serviceconnectLoader` to be consistent with serviceconnectManager and serviceconnectRelay below

agent/engine/service_connect/manager_linux.go
```diff
@@ -19,24 +19,27 @@ import (
 	"path/filepath"
 	"syscall"
 
-	"github.com/aws/amazon-ecs-agent/agent/api/serviceconnect"
-
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apiserviceconnect "github.com/aws/amazon-ecs-agent/agent/api/serviceconnect"
```
sparrc, line 23: 👍

agent/serviceconnect/load.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil, errors.Wrapf(err,
+			"appnet agent container load: failed to inspect image: %s", imageName)
+	}
+
+	return image, nil
+}
+
+// Common function for linux and windows to check if the container pause image has been loaded
+func (agent *loader) isImageLoaded(dockerClient dockerapi.DockerClient) (bool, error) {
+	imageName, _ := agent.GetLoadedImageName()
```
sparrc, line 68: from what I understand, this function is guaranteed to return nil for error only on linux, but this file here is compiled for all platforms, so should we be checking the error here?

agent/serviceconnect/load_linux.go
```diff
@@ -0,0 +1,74 @@
...
+
+	imageName, _ := agent.GetLoadedImageName()
+	return getAgentContainerImage(imageName, dockerClient)
+}
+
+func (agent *loader) IsLoaded(dockerClient dockerapi.DockerClient) (bool, error) {
+	return agent.isImageLoaded(dockerClient)
+}
+
+func (agent *loader) GetLoadedImageName() (string, error) {
```
sparrc, line 51: maybe add comment here explaining why this will never error on linux?

agent/api/task/task.go
aws-gibbskt: This is a marker for the END of the section start at 532. These sections need to be deleted after dev/testing and the backend is fully operational.

agent/app/agent_capability.go
aws-gibbskt: What is the guidance on seelog versus logger? It isn't consistent in the codebase

sparrc: logger is newer and what we prefer because it's structured and easier to read, but we still have some places directly using seelog just because I dont think anyone has had the time to fully migrate everything yet.

PR #3260, opened 2022-06-22
Added the changes to safely parse SSM parameters for windows containers.  fixes #2943

### LINKED ISSUE #2943: SSM Parameter ARN parsing for windows credentialspec options fails under certain conditions

saurabhc123:
### Summary

This PR adds the fixes to safely handle credentialSpec file creation in Windows when the ECS task definition defines a SSM parameter to be used as dockerSecurityOptions. Currently, the agent has the following issues.
1. Handling hierarchical paths in SSM parameters - The current version of the code just extracts the leaf path from the hierarchy and writes a file with that name. This can create a problem on a worker node if two containers require access to an SSM parameter `a/b/c` and `x/y/c`. The current code would just write the single `c` file.
2. Handling long SSM parameter names - SSM parameters in AWS support up to 1011 characters as per the [spec](https://docs.aws.amazon.com/systems-manager/latest/userguide/sysman-paramstore-su-create.html). However, Windows only support a max path length of 260 characters and max file length of 255 characters when writing to absolute paths. The current version of the code would encounter an error when the SSM parameter name is long which results in a path longer than 260 characters in windows.

### Implementation details

In our implementation, we used the concatenated value of `{task_id}{container_name}{ssm_parameter_ARN} ` . The addition of the task_id and container_name add ample jitter to avoid collision within a host and a task for a given credentialSpec. We hashed the concatenated string using the SHA-256 algorithm to generate a digest size of 64 characters. This string was used as the name of the credential file created in the directory `c:\ProgramData\docker\credentialspecs`. Since the digest size is always a fixed length of 64 characters, this mitigates the path length issues in Windows. Also, this does not require us to refactor the deletion of the CredentialSpecs later since all of them are in the root `credentialspecs` folder.

### Testing

Existing tests were refactored and an additional test was added to handle hierarchical SSM paths. The integration and unit tests were executed for the windows environment.

Bug - Fixed the creation of credentialSpec files in Windows when it is generated via SSM parameters that are hierarchical or the credentialSpec file path length is more than 260 characters.
### Licensing

### DIFF
--- a/agent/api/task/task_windows_test.go
+++ b/agent/api/task/task_windows_test.go
@@ -19,6 +19,7 @@ package task
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 	"runtime"
 	"testing"
 
@@ -396,7 +397,7 @@ func TestRequiresCredentialSpecResource(t *testing.T) {
 
 func TestGetAllCredentialSpecRequirements(t *testing.T) {
 	hostConfig := "{\"SecurityOpt\": [\"credentialspec:file://gmsa_gmsa-acct.json\"]}"
-	container := &apicontainer.Container{}
+	container := &apicontainer.Container{Name: "webapp1"}
 	container.DockerConfig.HostConfig = &hostConfig
 
 	task := &Task{
@@ -404,20 +405,20 @@ func TestGetAllCredentialSpecRequirements(t *testing.T) {
 		Containers: []*apicontainer.Container{container},
 	}
 
-	allCredSpecReq := task.getAllCredentialSpecRequirements()
+	credentialSpecContainerMap := task.getAllCredentialSpecRequirements()
 
-	credentialspec := "credentialspec:file://gmsa_gmsa-acct.json"
-	expectedCredSpecReq := []string{credentialspec}
+	credentialspecFileLocation := "credentialspec:file://gmsa_gmsa-acct.json"
+	expectedCredentialSpecContainerMap := map[string]string{credentialspecFileLocation: "webapp1"}
 
-	assert.EqualValues(t, expectedCredSpecReq, allCredSpecReq)
+	assert.True(t, reflect.DeepEqual(expectedCredentialSpecContainerMap, credentialSpecContainerMap))
 }
 
 func TestGetAllCredentialSpecRequirementsWithMultipleContainersUsingSameSpec(t *testing.T) {
 	hostConfig := "{\"SecurityOpt\": [\"credentialspec:file://gmsa_gmsa-acct.json\"]}"
-	c1 := &apicontainer.Container{}
+	c1 := &apicontainer.Container{Name: "webapp1"}
 	c1.DockerConfig.HostConfig = &hostConfig
 
-	c2 := &apicontainer.Container{}
+	c2 := &apicontainer.Container{Name: "webapp2"}
 	c2.DockerConfig.HostConfig = &hostConfig
 
 	task := &Task{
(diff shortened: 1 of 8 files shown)


jterry75:
Solves: #2943

jterry75 (approved):
Just a small NIT on namespaces

agent/api/task/task_windows_test.go
```diff
@@ -396,70 +397,70 @@ func TestRequiresCredentialSpecResource(t *testing.T) {
...
+	credentialSpecContainerMap := task.getAllCredentialSpecRequirements()
 
 	credentialspec1 := "credentialspec:file://gmsa_gmsa-acct-1.json"
 	credentialspec2 := "credentialspec:file://gmsa_gmsa-acct-2.json"
 
-	expectedCredSpecReq := []string{credentialspec1, credentialspec2}
+	expectedCredentialSpecContainerMap := map[string]string{credentialspec1: "webapp2", credentialspec2: "webapp3"}
 
-	assert.EqualValues(t, expectedCredSpecReq, allCredSpecReq)
+	assert.True(t, reflect.DeepEqual(expectedCredentialSpecContainerMap, credentialSpecContainerMap))
```
singholt, line 463: TIL 😄

agent/engine/docker_task_engine_windows.go
```diff
@@ -19,11 +19,10 @@ package engine
...
 	"time"
 
-	"github.com/aws/amazon-ecs-agent/agent/logger"
-	"github.com/aws/amazon-ecs-agent/agent/logger/field"
-
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/agent/logger"
+	"github.com/aws/amazon-ecs-agent/agent/logger/field"
 	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
singholt: +1. we don't at the moment.

singholt (approved):

vsiddharth (approved):

agent/engine/engine_windows_integ_test.go
```diff
@@ -55,7 +50,11 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	taskresourcevolume "github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
+
 	"github.com/aws/aws-sdk-go/aws"
+
```
singholt, line 55: "github.com/aws/aws-sdk-go/aws" and the next block are all third party imports right? so I think L55 extra line is not needed.

singholt (approved):

vsiddharth (dismissed):

jterry75 (approved):

saurabhc123:
> @saurabhc123 Some changes were recently merged to dev so could you please rebase with dev branch to remove unrelated changes from the PR?

👍

mythri-garaga (approved):

vsiddharth (approved):

singholt (approved):

PR #3265, opened 2022-06-27
Integrate with Service Connect bridge-mode CNI plugin

yinyic:
### Summary

ServiceConnect-enabled tasks require network namespace configurations that are performed by a new CNI plugin. This PR integrates with the new CNI plugin `ecs-serviceconnect` for bridge-mode tasks.

### Implementation details

In the previous PR https://github.com/aws/amazon-ecs-agent/pull/3255 where we integrated with SC CNI plugin for awsvpc tasks we've already added new SC CNI config models. 

In this PR, a new `RedirectMode` CNI config parameter is introduced. The value will be `NAT` and `TPROXY` for awsvpc and bridge-mode tasks respectively. Additionally, a `RedirectIP` field is introduced for further distinguishing one of the two scenarios for bridge mode netns setup.

For bridge-mode SC-enabled tasks, one pause container will be created for each task container, including both appnet and customer container. Depending on CNI configuration parameters the plugin will do one of two things
1. If `EgressConfig.RedirectIP` is included, create an IP route to redirect traffic destined to SC VIPs to the specified redirect IP (which will be AppNet container IP). **This is the configuration for customer pause containers.**
2. If `EgressConfig.ListenerPort` is included, set up a TPROXY. **This is the configuration for AppNet pause containers.**

In order for other task pause containers to be able to access AppNet container IP (i.e. the `RedirectIP`) during their netns setup. A `PauseContainerIPConfig` property is added to service connect `NetworkConfig`. The IP addresses will be populated once AppNet pause container is started.

TODO - Update engine `TestContainersWithServiceConnect_BridgeMode` test to be more reliable (https://github.com/aws/amazon-ecs-agent/pull/3262)

### Testing

Integrate with Service Connect bridge-mode CNI plugin

### Licensing

### DIFF
--- a/agent/api/serviceconnect/service_connect.go
+++ b/agent/api/serviceconnect/service_connect.go
@@ -24,6 +24,8 @@ type Config struct {
 
 	// Admin configuration for operating with AppNet Agent
 	RuntimeConfig RuntimeConfig `json:"runtimeConfig"`
+	// NetworkConfig contains additional network information for setting up task network namespace
+	NetworkConfig NetworkConfig `json:"networkConfig"`
 }
 
 // RuntimeConfig contains the runtime information for administering AppNet Agent
@@ -34,8 +36,6 @@ type RuntimeConfig struct {
 	StatsRequest string `json:"statsRequest"`
 	// HTTP Path + Params to drain ServiceConnect connections
 	DrainRequest string `json:"drainRequest"`
-	// SC pause container IP address - used for bridge-mode CNI configuration
-	PauseContainerIPConfig *PauseContainerIPConfig `json:"pauseContainerIPConfig,omitempty"`
 }
 
 // IngressConfigEntry is the ingress configuration for a given SC service.
@@ -77,7 +77,9 @@ type DNSConfigEntry struct {
 	Address  string `json:"address"`
 }
 
-type PauseContainerIPConfig struct {
-	IPv4Addr string `json:"ipv4Addr,omitempty"`
-	IPv6Addr string `json:"ipv6Addr,omitempty"`
+// NetworkConfig contains additional network information for setting up task network namespace.
+// This includes SC pause container IP address - used for bridge-mode CNI configuration
+type NetworkConfig struct {
+	SCPauseIPv4Addr string `json:"scPauseIPv4Addr,omitempty"`
+	SCPauseIPv6Addr string `json:"scPauseIPv6Addr,omitempty"`
 }
--- a/agent/api/serviceconnect/service_connect.go
+++ b/agent/api/serviceconnect/service_connect.go
@@ -34,6 +34,8 @@ type RuntimeConfig struct {
 	StatsRequest string `json:"statsRequest"`
 	// HTTP Path + Params to drain ServiceConnect connections
 	DrainRequest string `json:"drainRequest"`
+	// SC pause container IP address - used for bridge-mode CNI configuration
+	PauseContainerIPConfig *PauseContainerIPConfig `json:"pauseContainerIPConfig,omitempty"`
 }
 
 // IngressConfigEntry is the ingress configuration for a given SC service.
@@ -73,4 +75,9 @@ type VIP struct {
 type DNSConfigEntry struct {
 	HostName string `json:"hostName"`
 	Address  string `json:"address"`
+}
+
+type PauseContainerIPConfig struct {
+	IPv4Addr string `json:"ipv4Addr,omitempty"`
+	IPv6Addr string `json:"ipv6Addr,omitempty"`
 }
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -3174,7 +3174,7 @@ func (task *Task) IsContainerServiceConnectPause(containerName string) bool {
 	if scContainer == nil {
 		return false
 	}
-	scPauseName := fmt.Sprintf("%s-%s", NetworkPauseContainerName, scContainer.Name)
+	scPauseName := fmt.Sprintf(ServiceConnectPauseContainerNameFormat, scContainer.Name)
 	return containerName == scPauseName
 }
 
@@ -3215,15 +3215,14 @@ func (task *Task) PopulateServiceConnectRuntimeConfig(serviceConnectConfig servi
 }
 
 // PopulateServiceConnectPauseIPConfig is called once we've started SC pause container and retrieved its container IPs.
-func (task *Task) PopulateServiceConnectPauseIPConfig(IPv4Addr, IPv6Addr string) {
+func (task *Task) PopulateServiceConnectNetworkConfig(ipv4Addr string, ipv6Addr string) {
 	task.lock.Lock()
 	defer task.lock.Unlock()
 
-	task.ServiceConnectConfig.RuntimeConfig.PauseContainerIPConfig =
-		&serviceconnect.PauseContainerIPConfig{
-			IPv4Addr: IPv4Addr,
-			IPv6Addr: IPv6Addr,
-		}
+	task.ServiceConnectConfig.NetworkConfig = serviceconnect.NetworkConfig{
+		SCPauseIPv4Addr: ipv4Addr,
+		SCPauseIPv6Addr: ipv6Addr,
+	}
 }
 
 func (task *Task) GetServiceConnectRuntimeConfig() serviceconnect.RuntimeConfig {
@@ -3231,6 +3230,13 @@ func (task *Task) GetServiceConnectRuntimeConfig() serviceconnect.RuntimeConfig 
 	defer task.lock.RUnlock()
 
 	return task.ServiceConnectConfig.RuntimeConfig
+}
+
+func (task *Task) GetServiceConnectNetworkConfig() serviceconnect.NetworkConfig {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	return task.ServiceConnectConfig.NetworkConfig
 }
 
 func (task *Task) SetServiceConnectConnectionDraining(draining bool) {
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -62,6 +62,8 @@ import (
 const (
 	// NetworkPauseContainerName is the internal name for the pause container
 	NetworkPauseContainerName = "~internal~ecs~pause"
+	// ServiceConnectPauseContainerNameFormat is the naming format for SC pause containers
+	ServiceConnectPauseContainerNameFormat = "~internal~ecs~pause-%s"
 
 	// NamespacePauseContainerName is the internal name for the IPC resource namespace and/or
 	// PID namespace sharing pause container
@@ -1457,7 +1459,7 @@ func (task *Task) addNetworkResourceProvisioningDependencyServiceConnectBridge(c
 		pauseContainer.TransitionDependenciesMap = make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet)
 		// The pause container name is used internally by task engine but still needs to be unique for every task,
 		// hence we are appending the corresponding application container name (which must already be unique within the task)
-		pauseContainer.Name = fmt.Sprintf("%s-%s", NetworkPauseContainerName, container.Name)
+		pauseContainer.Name = fmt.Sprintf(ServiceConnectPauseContainerNameFormat, container.Name)
 		pauseContainer.Image = fmt.Sprintf("%s:%s", cfg.PauseContainerImageName, cfg.PauseContainerTag)
 		pauseContainer.Essential = true
 		pauseContainer.Type = apicontainer.ContainerCNIPause
@@ -1487,7 +1489,7 @@ func (task *Task) addNetworkResourceProvisioningDependencyServiceConnectBridge(c
 // For a container with name "abc", the pause container will always be named "~internal~ecs~pause-abc"
 func (task *Task) GetBridgeModePauseContainerForTaskContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
 	// "~internal~ecs~pause-$TASK_CONTAINER_NAME"
-	pauseContainerName := fmt.Sprintf("%s-%s", NetworkPauseContainerName, container.Name)
+	pauseContainerName := fmt.Sprintf(ServiceConnectPauseContainerNameFormat, container.Name)
 	pauseContainer, ok := task.ContainerByName(pauseContainerName)
 	if !ok {
 		return nil, fmt.Errorf("could not find pause container %s for task container %s", pauseContainerName, container.Name)
@@ -1495,9 +1497,9 @@ func (task *Task) GetBridgeModePauseContainerForTaskContainer(container *apicont
 	return pauseContainer, nil
 }
 
-// GetBridgeModeTaskContainerForPauseContainer retrieves the associated task container for a pause container in a bridge-mode SC-enabled task.
+// getBridgeModeTaskContainerForPauseContainer retrieves the associated task container for a pause container in a bridge-mode SC-enabled task.
 // For a container with name "abc", the pause container will always be named "~internal~ecs~pause-abc"
-func (task *Task) GetBridgeModeTaskContainerForPauseContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
+func (task *Task) getBridgeModeTaskContainerForPauseContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
 	if container.Type != apicontainer.ContainerCNIPause {
 		return nil, fmt.Errorf("container %s is not a CNI pause container", container.Name)
 	}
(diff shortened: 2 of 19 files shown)


agent/api/task/task.go
```diff
@@ -1539,9 +1539,9 @@ func (task *Task) GetBridgeModePauseContainerForTaskContainer(container *apicont
 	return pauseContainer, nil
 }
 
-// getBridgeModeTaskContainerForPauseContainer retrieves the associated task container for a pause container in a bridge-mode SC-enabled task.
+// GetBridgeModeTaskContainerForPauseContainer retrieves the associated task container for a pause container in a bridge-mode SC-enabled task.
 // For a container with name "abc", the pause container will always be named "~internal~ecs~pause-abc"
-func (task *Task) getBridgeModeTaskContainerForPauseContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
+func (task *Task) GetBridgeModeTaskContainerForPauseContainer(container *apicontainer.Container) (*apicontainer.Container, error) {
```
aws-gibbskt, line 1544: Where is this accessed out of package? I don't know why this change is needed.

agent/api/serviceconnect/service_connect.go
```diff
@@ -32,6 +32,8 @@ type RuntimeConfig struct {
 	StatsRequest string `json:"statsRequest"`
 	// HTTP Path + Params to drain ServiceConnect connections
 	DrainRequest string `json:"drainRequest"`
+	// SC pause container IP address - used for bridge-mode CNI configuration
+	PauseContainerIPConfig *PauseContainerIPConfig `json:"pauseContainerIPConfig,omitempty"`
```
rajal-amzn, line 36: Would you prefer storing these as `NetworkConfig` along with pause container associated with customer containers ?

agent/api/task/task_linux.go
```diff
@@ -355,3 +357,38 @@ func (task *Task) BuildCNIConfig(includeIPAMConfig bool, cniConfig *ecscni.Confi
 
 	return cniConfig, nil
 }
+
+// BuildCNIConfigBridgeMode builds a list of CNI network configurations for a task in docker bridge mode.
+func (task *Task) BuildCNIConfigBridgeMode(cniConfig *ecscni.Config, containerName string) (*ecscni.Config, error) {
```
rajal-amzn, line 362: nit: I see all the methods are marked public. why?

agent/api/task/task_linux.go
```diff
@@ -355,3 +357,38 @@ func (task *Task) BuildCNIConfig(includeIPAMConfig bool, cniConfig *ecscni.Confi
...
+	if !task.IsNetworkModeBridge() {
+		return nil, errors.New("non bridge-mode task should not invoke BuildCNIConfigBridgeMode")
+	}
+
+	var netconf *libcni.NetworkConfig
+	var ifName string
+	var err error
+
+	// Build a CNI network configuration for bridge-mode SC-enabled task
+	if task.IsServiceConnectEnabled() {
```
rajal-amzn, line 372: If evaluated to false, shouldnt we be retuning error?

agent/ecscni/netconfig_linux.go
```diff
@@ -176,6 +176,8 @@ func NewAppMeshConfig(appMesh *appmesh.AppMesh, cfg *Config) (string, *libcni.Ne
 // NewServiceConnectNetworkConfig creates a new ServiceConnect CNI network configuration
 func NewServiceConnectNetworkConfig(
 	scConfig *serviceconnect.Config,
+	redirectMode RedirectMode,
+	shouldIncludeRedirectIP bool,
```
chienhanlin, line 180: nit. `shouldIncludeRedirectIP, enableIPv4, enableIPv6 bool,`  there of them are all boolean.

agent/api/task/task_linux_test.go
```diff
@@ -67,6 +70,16 @@ const (
...
+
+	scContainerName      = "service-connect"
+	scEgressListenerPort = 12345
+	scInterceptPort      = 8080
+	scListenerPort       = 15000
+	scPauseIPv4          = "172.0.0.2"
+)
+
+var (
+	scPauseContainerName = fmt.Sprintf("%s-%s", NetworkPauseContainerName, scContainerName)
```
chienhanlin, line 82: nit. feel that we use this in more than one place. Should we turn this into a const template in task.go and reuse it in following lines?
* https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/api/task/task.go#L1504
* https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/api/task/task.go#L1534

agent/api/task/task_linux.go
chienhanlin: Making `BuildCNIConfigAwsvpc` and `BuildCNIConfigBridgeMode` public and than they can be used in the engine package, like:

`agent/engine/docker_task_engine.go`
* buildCNIConfigFromTaskContainerAwsvpc() [[L1940](https://github.com/yinyic/amazon-ecs-agent/blob/6a66b11cbc4f727012a78b35470cad773e836454/agent/engine/docker_task_engine.go#L1940)]

* buildCNIConfigFromTaskContainerBridgeMode() [[L1958](https://github.com/yinyic/amazon-ecs-agent/blob/6a66b11cbc4f727012a78b35470cad773e836454/agent/engine/docker_task_engine.go#L1958)]

`agent/engine/task_manager.go` 
* releaseIPInIPAM()[[L757](https://github.com/yinyic/amazon-ecs-agent/blob/6a66b11cbc4f727012a78b35470cad773e836454/agent/engine/task_manager.go#L757)]

agent/api/task/task.go
yinyic: Thanks for calling out. This was an intermediate change when I was working on the PR. It's no longer needed. Will remove.

agent/api/serviceconnect/service_connect.go
yinyic: Do you mean storing it directly inside `ecscni.ServiceConnectConfig` [struct](https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/ecscni/types_linux.go#L170)? We collect AppNet pause IPs from docker daemon, but the method that we create this config object currently does not have a handle of docker client. I'd rather keep it that way, because this config object is intended to be an API object for CNI plugin, so it probably shouldn't need to know or care about the container runtime.

agent/api/task/task_linux.go
yinyic: Thanks ChienHan! 
Like ChienHan mentioned, docker engine drives the container lifecycle as a state machine, and task package owns most of the actions to take during each task/container state.

agent/api/task/task_linux.go
yinyic: I was thinking that we may add other bridge-mode CNI plugins in the future so I kept it open. But you are right that we probably should throw error because SC task is the only one that should be invoking this method at this moment.

agent/ecscni/netconfig_linux.go
yinyic: Will update to ist each in its own line with an explicit type.

agent/api/task/task.go
```diff
@@ -3203,6 +3203,15 @@ func (task *Task) GetServiceConnectContainer() *apicontainer.Container {
 	return c
 }
 
+func (task *Task) IsContainerServiceConnectPause(containerName string) bool {
```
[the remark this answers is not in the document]
yinyic: Will add.

agent/api/task/task_linux_test.go
yinyic: Will do

agent/api/task/task_linux_test.go
```diff
@@ -1316,6 +1329,54 @@ func TestBuildCNIConfigRegularENIWithAppMesh(t *testing.T) {
...
+	for _, blockIMDS := range []bool{true, false} {
+		t.Run(fmt.Sprintf("When BlockInstanceMetadata is %t", blockIMDS), func(t *testing.T) {
+			testTask := &Task{}
+			testTask.AddTaskENI(getTestENI())
+			testTask.ServiceConnectConfig = &serviceconnect.Config{
+				ContainerName: scContainerName,
+				IngressConfig: []serviceconnect.IngressConfigEntry{{ListenerPort: scListenerPort}},
+				EgressConfig:  &serviceconnect.EgressConfig{ListenerPort: scEgressListenerPort},
+			}
+			testTask.Containers = []*apicontainer.Container{{Name: "service-connect"}}
```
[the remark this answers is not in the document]
yinyic: Will do

agent/api/task/task.go
```diff
@@ -3203,6 +3205,18 @@ func (task *Task) GetServiceConnectContainer() *apicontainer.Container {
...
 
+// IsContainerServiceConnectPause checks whether a given container name is the name of the task service connect pause
+// container. We construct the name of SC pause container by taking SC container name from SC config, and using the
+// pause container naming pattern.
+func (task *Task) IsContainerServiceConnectPause(containerName string) bool {
+	scContainer := task.GetServiceConnectContainer()
+	if scContainer == nil {
+		return false
+	}
+	scPauseName := fmt.Sprintf("%s-%s", NetworkPauseContainerName, scContainer.Name)
```
chienhanlin, line 3216: nit and sorry I missed this line in my previous comment.
```
scPauseName := fmt.Sprintf(ServiceConnectPauseContainerNameFormat, scContainer.Name)
```

agent/api/task/task_linux.go
```diff
@@ -355,3 +357,37 @@ func (task *Task) BuildCNIConfig(includeIPAMConfig bool, cniConfig *ecscni.Confi
...
+func (task *Task) BuildCNIConfigBridgeMode(cniConfig *ecscni.Config, containerName string) (*ecscni.Config, error) {
+	if !task.IsNetworkModeBridge() || !task.IsServiceConnectEnabled() {
+		return nil, errors.New("only bridge-mode Service-Connect-enabled task should invoke BuildCNIConfigBridgeMode")
+	}
+
+	var netconf *libcni.NetworkConfig
+	var ifName string
+	var err error
+
+	scPauseIPConfig := task.GetServiceConnectRuntimeConfig().PauseContainerIPConfig
```
chienhanlin, line 372: Same Q asked in https://github.com/aws/amazon-ecs-agent/pull/3265/files#r914210160. Thanks!

agent/engine/docker_task_engine.go
```diff
@@ -1752,26 +1781,72 @@ func (engine *DockerTaskEngine) provisionContainerResources(task *apitask.Task,
...
+
+	log.Infof("Successfully configured pause netns for %s", container.Name)
+	return dockerapi.MetadataFromContainer(containerInspectOutput)
+}
+
+// checkTearDownPauseContainer idempotently tears down the pause container network when the pause container's known
 //or desired status is stopped.
-func (engine *DockerTaskEngine) checkTearDownPauseContainerAwsvpc(task *apitask.Task) {
-	if !task.IsNetworkModeAWSVPC() {
+func (engine *DockerTaskEngine) checkTearDownPauseContainer(task *apitask.Task) {
```
fierlion, line 1840: nit: to tie to the pause container naming thread -- here you're not differentiating between the sc pause container and the base pause container.  If this should only apply to the ServiceConnect specific pause container I would call this `checkTearDownServiceConnectPauseContainer()`... 

however my preference (as I noted above) is just reusing the `NetworkPauseContainerName` and keeping it generic.

agent/api/task/task.go
yinyic: Thanks! Will update

agent/ecscni/netconfig_linux.go
```diff
@@ -192,12 +195,37 @@ func NewServiceConnectNetworkConfig(
...
+		case NAT:
+			// NAT redirect mode is for awsvpc tasks, where the one and only pause container netns will have a NAT redirect rule.
+			egressConfig.ListenerPort = scConfig.EgressConfig.ListenerPort
+		case TPROXY: // bridge
+			// TPROXY redirect mode is used for bridge-mode tasks. There are two use cases:
+			// 1. SC pause container netns will set up TPROXY that requires the Egress port
+			// 2. Other task pause container netns will add a route for traffic destined for SC VIP-CIDR to go to SC container.
+			//    In that case the configuration requires the SC (pause) container IP.
+			if shouldIncludeRedirectIP {
+				scPauseIPConfig := scConfig.RuntimeConfig.PauseContainerIPConfig
```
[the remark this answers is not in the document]
yinyic: That's a good callout - no it won't be null if we already get here, but it might be null when we populate pause container IPs so I'll add a check there.

agent/api/task/task_linux.go
yinyic: it won't be null if we already get here, but it might be null when we populate pause container IPs in `task.PopulateServiceConnectPauseIPConfig` - I'll make sure to create `RuntimeConfig` if it does not already exist

agent/api/task/task_linux_test.go
```diff
@@ -1367,3 +1428,71 @@ func TestBuildCNIConfigTrunkBranchENI(t *testing.T) {
...
+			// The first one should be for the ENI.
+			var scConfig ecscni.ServiceConnectConfig
+			err = json.Unmarshal(cniConfig.NetworkConfigs[0].CNINetworkConfig.Bytes, &scConfig)
+			require.NoError(t, err)
+			assert.Equal(t, 1, len(scConfig.IngressConfig))
+			assert.Equal(t, uint16(scListenerPort), scConfig.IngressConfig[0].ListenerPort)
+			assert.NotNil(t, scConfig.EgressConfig)
+			assert.Equal(t, string(ecscni.TPROXY), scConfig.EgressConfig.RedirectMode)
+			if containerName != scPauseContainerName {
+				// Should only include redirect IPs if container is an application pause container
```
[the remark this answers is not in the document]
yinyic: > is it possible that the network pause container could ever be named the same as the service connect application pause container? 

Sorry not quite sure if I follow. What do we mean by "network pause container"?

When ECS Agent receives a bridge-mode service-connect task, and let's say there are N customer application container, and 1 service connect container. Those (N+1) containers all have unique names (validated by control plane). When we create pause containers for each of them and name the pause containers `~internal~ecs~pause-<$ACTUAL_CONTAINER_NAME>`, then it's guaranteed that all (N+1) pause containers will also have unique names.

agent/api/task/task_linux.go
```diff
@@ -355,3 +357,37 @@ func (task *Task) BuildCNIConfig(includeIPAMConfig bool, cniConfig *ecscni.Confi
 
 	return cniConfig, nil
 }
+
+// BuildCNIConfigBridgeMode builds a list of CNI network configurations for a task in docker bridge mode.
+// Currently the only plugin in available is for Service Connect
+func (task *Task) BuildCNIConfigBridgeMode(cniConfig *ecscni.Config, containerName string) (*ecscni.Config, error) {
```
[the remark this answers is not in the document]
yinyic: I separated bridge-mode CNI config creation from the AWSVPC because they are very different - bridge-mode only needs the ServiceConnect CNI plugin, while AWSVPC need a bunch more, and other configurations. 

If we are introducing more network modes that will use CNI plugins, then we could combine it with the existing AWSVPC or bridge CNI config build if there's a lot of similarity, or we could create a new one for its own.

agent/ecscni/plugin_linux_test.go
```diff
@@ -651,88 +657,198 @@ func TestConstructIPAMNetworkConfig(t *testing.T) {
...
-	assert.Equal(t, true, scNetworkConfig.EnableIPv4)
-	assert.Equal(t, false, scNetworkConfig.EnableIPv6)
+	testCases := []struct {
+		redirectMode            RedirectMode
+		shouldIncludeRedirectIP bool
+	}{
+		{
+			redirectMode:            NAT,
+			shouldIncludeRedirectIP: false,
+		},
```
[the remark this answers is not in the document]
yinyic: That's not a valid use case. When building SC CNI config for AWSVPC task, we will always set redirectMode to `NAT` and `shouldIncludeRedirectIP` to false.

agent/engine/docker_task_engine.go
yinyic: > you're not differentiating between the sc pause container and the base pause container

That's actually the intention. The branching for networking mode comes in at a later point inside `cleanupPauseContainerNetwork` method, similar to how we branch in `provisionContainerResources`

agent/ecscni/netconfig_linux.go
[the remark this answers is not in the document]
yinyic: Created a new field `NetworkConfig` that's no longer nested in `RuntimeConfig` to avoid the complication that multiple containers could potentially be creating it.

agent/api/task/task_linux.go
yinyic: Created a new field `NetworkConfig` that's no longer nested in `RuntimeConfig` to avoid the complication that multiple containers could potentially be creating it.

fierlion (dismissed):

chienhanlin (approved):

rajal-amzn (commented):
Changes look good. Have we also picked up the latest CNI changes? Or would you be taking that up in a separate PR.

chienhanlin (dismissed):

fierlion (approved):

rajal-amzn (dismissed):

chienhanlin (approved):

rajal-amzn (approved):

fierlion (approved):

PR #3272, opened 2022-06-28
Collect and publish Service Connect metrics even when metrics gathering for tasks is explicitly disabled

mythri-garaga:
### Summary

Currently Service Connect metrics are collected and published to TACS when env var `ECS_DISABLE_METRICS` is false. For Service Connect enabled tasks, the expected behaviour is to publish metrics related to Service Connect by default and this PR includes changes to fix this bug.
### Implementation details

- agent/stats/engine.go - 
     - In TACS request, `Idle` attribute is set to true only when there are no container metrics or service connect metrics to be sent to TACS.
     - Update to collect SC metrics when `ECS_DISABLE_METRICS` is set and `engine.taskToServiceConnectStats` is not empty.
- agent/tcs/client/client.go - Invoke publishMetrics() irrespective of `ECS_DISABLE_METRICS` value.
### Testing

### Licensing

### DIFF
--- a/agent/stats/engine.go
+++ b/agent/stats/engine.go
@@ -417,7 +417,6 @@ func (engine *DockerStatsEngine) addToStatsContainerMapUnsafe(
 
 // GetInstanceMetrics gets all task metrics and instance metadata from stats engine.
 func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats bool) (*ecstcs.MetricsMetadata, []*ecstcs.TaskMetric, error) {
-	var taskMetrics []*ecstcs.TaskMetric
 	idle := engine.isIdle()
 	metricsMetadata := &ecstcs.MetricsMetadata{
 		Cluster:           aws.String(engine.cluster),
@@ -425,6 +424,8 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 		Idle:              aws.Bool(idle),
 		MessageId:         aws.String(uuid.NewRandom().String()),
 	}
+
+	var taskMetrics []*ecstcs.TaskMetric
 	if idle {
 		seelog.Debug("Instance is idle. No task metrics to report")
 		fin := true
@@ -442,16 +443,24 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 		}
 	}
 
-	for taskArn := range engine.tasksToContainers {
+	taskStatsToCollect := engine.getTaskStatsToCollect()
+	for taskArn := range taskStatsToCollect {
+		_, isServiceConnectTask := engine.taskToServiceConnectStats[taskArn]
 		containerMetrics, err := engine.taskContainerMetricsUnsafe(taskArn)
 		if err != nil {
 			seelog.Debugf("Error getting container metrics for task: %s, err: %v", taskArn, err)
-			continue
+			// skip collecting service connect related metrics, if task is not service connect enabled
+			if !isServiceConnectTask {
+				continue
+			}
 		}
 
 		if len(containerMetrics) == 0 {
 			seelog.Debugf("Empty containerMetrics for task, ignoring, task: %s", taskArn)
-			continue
+			// skip collecting service connect related metrics, if task is not service connect enabled
+			if !isServiceConnectTask {
+				continue
+			}
 		}
 
 		taskDef, exists := engine.tasksToDefinitions[taskArn]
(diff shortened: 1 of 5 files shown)


agent/stats/engine.go
```diff
@@ -414,7 +414,7 @@ func (engine *DockerStatsEngine) addToStatsContainerMapUnsafe(
 // GetInstanceMetrics gets all task metrics and instance metadata from stats engine.
 func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats bool) (*ecstcs.MetricsMetadata, []*ecstcs.TaskMetric, error) {
 	var taskMetrics []*ecstcs.TaskMetric
-	idle := engine.isIdle()
+	idle := engine.isIdle() && engine.serviceConnectIsIdle()
```
rajal-amzn, line 417: should this be `||` instead of `&&`

agent/tcs/client/client.go
```diff
@@ -187,9 +186,11 @@ func (cs *clientServer) publishMetrics() {
 				includeServiceConnectStats = true
 				cs.publishServiceConnectTickerInterval = 0
 			}
-			err := cs.publishMetricsOnce(includeServiceConnectStats)
-			if err != nil {
-				seelog.Warnf("Error publishing metrics: %v", err)
+			if !cs.disableResourceMetrics || includeServiceConnectStats {
```
rajal-amzn, line 189: When would `disableResourceMetrics` be set to true?

agent/stats/engine.go
mythri-garaga: `idle` is true only when both container insights are disabled and there are not service connect metrics to send, otherwise TACS need the request to be non-idle.

agent/stats/engine.go
```diff
@@ -438,16 +440,21 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
...
-		}
+	taskStatsToCollect := engine.getTaskStatsToCollect()
+	for taskArn := range taskStatsToCollect {
+		containerMetrics := make([]*ecstcs.ContainerMetric, 0)
+		var err error
+		if !idle {
+			containerMetrics, err = engine.taskContainerMetricsUnsafe(taskArn)
+			if err != nil {
+				seelog.Debugf("Error getting container metrics for task: %s, err: %v", taskArn, err)
+				continue
```
ubhattacharjya, line 451: If there is no container metrics for this task, it will not collect service-connect metrics also for that task, right?

agent/stats/engine.go
```diff
@@ -522,6 +529,13 @@ func (engine *DockerStatsEngine) isIdle() bool {
 	return len(engine.tasksToContainers) == 0
 }
 
+func (engine *DockerStatsEngine) serviceConnectIsIdle() bool {
```
ubhattacharjya, line 532: could we do this check in isIdle() method as well?

agent/stats/engine.go
mythri-garaga: right but if that is task is SC enabled then `engine.taskToServiceConnectStats` will not be empty and we will still collect SC metrics for that task.

mythri-garaga: Addressed it in the latest commit, thank you for calling it.

agent/stats/engine.go
```diff
@@ -438,16 +439,23 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
...
+			if _, scTask := engine.taskToServiceConnectStats[taskArn]; !scTask {
+				continue
+			}
 		}
 
 		if len(containerMetrics) == 0 {
 			seelog.Debugf("Empty containerMetrics for task, ignoring, task: %s", taskArn)
-			continue
+			// skip collecting service connect related metrics, if task is service connect enabled
+			if _, scTask := engine.taskToServiceConnectStats[taskArn]; !scTask {
```
ubhattacharjya, line 456: Could we compute scTask just once at the beginning of the for loop and use that in the 2 if conditions

sparrc: This looks like it could be a data race, as some other process is going to be writing to this map at the same time you are reading it. Is there a better way to check if the task is not service connect enabled, or maybe wrap this into a helper function with a mutex?

agent/stats/engine.go
```diff
@@ -438,16 +439,23 @@ func (engine *DockerStatsEngine) GetInstanceMetrics(includeServiceConnectStats b
 		}
 	}
 
-	for taskArn := range engine.tasksToContainers {
+	taskStatsToCollect := engine.getTaskStatsToCollect()
+	for taskArn := range taskStatsToCollect {
 		containerMetrics, err := engine.taskContainerMetricsUnsafe(taskArn)
```
sparrc, line 444: should the "unsafe" function be surrounded by a lock/unlock mutex?

sparrc (approved):

ubhattacharjya (dismissed):

sparrc (approved):

ubhattacharjya (approved):

PR #3279, opened 2022-07-01
Set UID and proper directory permission for AppNet task container

yinyic:
### Summary

By default docker container run as root. This PR enables ECS Agent to run AppNet task container as non-root user 

### Implementation details

When creating container config, we can specify the `User` [option](https://docs.docker.com/engine/reference/run/#user) to start the container with. Here we'll be using a `uid` of `1337`, which is chosen arbitrarily. 

Because AppNet container will need `rwx` permissions on the mounted directories (where AppNet will create admin socket file), we are going to make user `1337` the directory owner and assign permission `700` to the directory.

Prerequisite - Running as non-root, AppNet container needs to be able to preserve `CAP_NET_ADMIN` for Envoy process, which is needed in bridge-mode tproxy configuration. The capability is dropped from Envoy process's effective capability set by default. https://man7.org/linux/man-pages/man7/capabilities.7.html

### Testing

Refactored existing unit test to take in a flag for determining whether to set UID and run chmod. The flag will be true if test is run in privileged mode.

Set UID and proper directory permission for AppNet task container

### Licensing

### DIFF
--- a/agent/engine/service_connect/manager_linux_test_common.go
+++ b/agent/engine/service_connect/manager_linux_test_common.go
@@ -1,0 +1,203 @@
+//go:build linux && (unit || sudo_unit)
+// +build linux
+// +build unit sudo_unit
+
+package serviceconnect
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+	"time"
+
+	apieni "github.com/aws/amazon-ecs-agent/agent/api/eni"
+	"github.com/aws/amazon-ecs-agent/agent/config"
+
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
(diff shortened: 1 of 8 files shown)


ubhattacharjya (approved):

agent/engine/service_connect/manager_linux_test_common.go
```diff
@@ -0,0 +1,203 @@
...
+		{
+			name:                 "Service connect container has extra binds/ENV",
+			container:            serviceConnectContainer,
+			expectedENV:          expectedENVs,
+			expectedBinds:        expectedBinds,
+			expectedBindDirPerm:  fs.FileMode(0700).String(),
+			expectedBindDirOwner: serviceconnect.AppNetUID,
+		},
+	}
+	// Add test cases for other containers expecting no modifications
```
singholt, line 155: nit: TODO?

singholt (approved):

PR #3280, opened 2022-07-02
Adding support to launch Relay and assign task container images

aws-gibbskt:
Summary
This CR finalizes the process of starting a Relay Container and auto assigning images to task containers.

Implementation details
Create a new task with an internal container which is run in host mode on the instance.

Update container images for task containers for service connect.

Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -494,9 +494,11 @@ func (task *Task) initServiceConnectResources() error {
 			ContainerName: "service-connect",
 		}
 	}
+	// TODO [SC]: End dev/testing purposes
 	if task.IsServiceConnectEnabled() {
 		// TODO [SC]: initDummyServiceConnectConfig is for dev testing only, remove it when final SC model from ACS is in place
 		task.initDummyServiceConnectConfig()
+		// TODO [SC]: End dev/testing purposes
 		if err := task.initServiceConnectEphemeralPorts(); err != nil {
 			return err
 		}
@@ -518,6 +520,8 @@ func (task *Task) initDummyServiceConnectConfig() {
 		return
 	}
 }
+
+// TODO [SC]: End dev/testing purposes
 
 func (task *Task) initServiceConnectEphemeralPorts() error {
 	var utilizedPorts []uint16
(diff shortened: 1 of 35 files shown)


agent/api/task/task.go
```diff
@@ -492,9 +492,11 @@ func (task *Task) initServiceConnectResources() error {
 			ContainerName: "service-connect",
 		}
 	}
+	// TODO [SC]: End dev/testing purposes
```
aws-gibbskt, line 495: End of line 488 delete section

agent/api/task/task.go
```diff
@@ -517,6 +519,8 @@ func (task *Task) initDummyServiceConnectConfig() {
 	}
 }
 
+// TODO [SC]: End dev/testing purposes
```
aws-gibbskt, line 522: End of line 507 delete section

yinyic (approved):
Synced offline - this PR might need a few follow-ups to address:
- Any missing properties for the internal task and container which may include task `desiredState`, container `steadyState` etc. 
- Any impact of using a made-up task ARN
- Visibility of this internal task - it shouldn't be visible through TMDE, STSC, or TACS stats.

Additionally, there may be a chance that the first AppNet Agent container gets bootstrapped before Relay container and hence falling into unhealthy state, but it should be able to self-recover once Relay is running.

mythri-garaga (approved):

agent/engine/service_connect/manager_linux.go
```diff
@@ -150,10 +156,24 @@ func (m *manager) initAgentDirectoryMounts(taskId string, container *apicontaine
...
+}
+
+func (m *manager) initRelayEnvironment(config *config.Config, container *apicontainer.Container) {
+	scEnv := map[string]string{
+		m.statusENV:    filepath.Join(m.statusPathContainer, m.statusFileName),
+		upstreamENV:    filepath.Join(m.relayPathContainer, m.relayFileName),
+		regionENV:      config.AWSRegion,
+		agentModeENV:   agentModeValue,
+		relayEnableENV: relayEnableOn,
+		// TODO: get programmatically when available
```
yinyic, line 175: This actually is already available through `DiscoverPollEndpoint`. We currently are only consuming ACS and TACS endpoints, need to make a small [change](https://github.com/aws/amazon-ecs-agent/blob/71a0f6765cc469913a0b09728196feb0d0d33e37/agent/ecs_client/model/ecs/api.go#L7401) to take in EMS endpoint as well

yinyic: Need to update API model [here](https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/ecs_client/model/api/api-2.json#L1147-L1153). We can do that in a follow-up PR.

yinyic (approved):

PR #3288, opened 2022-07-06
Add VpcId to TMDE Task Responses

amogh09:
### Summary
This change is to enhance the `TaskResponse` returned by Task Metadata Endpoint to include VPC ID. This change does not apply to ECS Anywhere instances, so no VPC ID would be returned by Task Metadata Endpoint on ECS Anywhere instances. 

#### For agent instances with `TaskENIEnabled` setting set to `true` 
Agent loads the VPC ID of the container instance by querying EC2 Instance Metadata Service (IMDS). So, for TMDE the VPC ID of the container instance is already available in `ecsAgent` type. 

#### For agent instances with `TaskENIEnabled` setting set to `false`
Currently the agent does not load the VPC ID of the container instance in this case. This PR includes changes to make the agent load VPC ID (and mac and subnets) of the container instance if the container instance is not external. Doing so shouldn't have any side-effects because these values are not used to drive any logic.

Changes under this PR add some piping to forward the VPC ID value from `ecsAgent` type to Task Metadata Endpoint handlers and change the handlers to include the VPC ID value in all task responses. `TaskResponse` type definition is updated to include a new `VPCID` field.

### Testing
Deployed changed agent source to a test EC2 instance, ran test tasks with `awsvpc`, `bridge`, and `host` network modes, and verified that `VpcId` field is populated in task responses for all cases. 

Deployed changed agent to a test ECS Anywhere instance, ran a test task with `host` network mode, and verified that `VpcId` field is *not* populated and that a successful Task Response is returned.

Updated MACIS TMDE functional tests for Linux and Windows and ran them against the agent artifacts for this PR for EC2 Linux and Windows, and ECS-A Linux and Windows platforms.

Updated existing Task Metadata Endpoint unit tests to include `VpcId`.

New tests cover the changes: Updated existing tests

`VpcId` field will be present in task responses from Task Metadata Endpoint for EC2 instances. ECS Anywhere instances are not affected.

### Licensing


singholt (changes requested):

agent/handlers/task_server_setup.go
```diff
@@ -55,6 +55,7 @@ func taskServerSetup(credentialsManager credentials.Manager,
 	steadyStateRate int,
 	burstRate int,
 	availabilityZone string,
+	vpcId string,
```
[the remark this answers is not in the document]
amogh09: I see. Will fix.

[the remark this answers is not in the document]
amogh09: Fixed.

agent/handlers/v2/response_test.go
```diff
@@ -336,6 +337,7 @@ func TestTaskResponseMarshal(t *testing.T) {
 		"DesiredStatus":    "RUNNING",
 		"KnownStatus":      "RUNNING",
 		"AvailabilityZone": availabilityZone,
+		"VpcId":            vpcID,
```
Realmonia, line 340: nit: is this `VpcId` missed from revision?

Realmonia (approved):

agent/handlers/v2/response_test.go
amogh09: We are still using `VpcId` in the JSON format of `TaskResponse`. I did so because of two reasons - 
1. `VpcId` is used as a key for VPC ID field in response from `aws ec2 describe-instances` CLI command.
2. All keys in `TaskResponse` JSON are PascalCase and `VPCID` was sticking out IMO.

singholt (approved):

Realmonia (dismissed):
Please squash changes before merging.

agent/app/agent_test.go
```diff
@@ -363,6 +373,13 @@ func testDoStartHappyPathWithConditions(t *testing.T, blackholed bool, warmPools
 	ec2MetadataClient.EXPECT().PublicIPv4Address().Return(hostPublicIPv4Address, nil)
 	ec2MetadataClient.EXPECT().OutpostARN().Return("", nil)
 
+	if !isExternalLaunchType {
```
singholt, line 376: this should be `if isExternalLaunchType` ? since you're checking for external launch type?

amogh09: Here the condition is supposed to be that launch type is not external, in that case we want to initialize VPC ID. If launch type is external then VPC is not applicable.

singholt: synced offline, the confusion here was that in the mock statements, `nil` is actually the error, not the value.

singholt (dismissed):

amogh09:
Only v4 endpoint's response may be changed. Older versions are considered immutable. Need to limit this change to v4.

singholt (approved):

Realmonia (approved):

PR #3291, opened 2022-07-08
initialize networkmode before we validate the payload

petderek:
### Summary

loosening validation around network mode to include an empty network mode

### Testing

added a new unit test

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -322,6 +322,9 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task, 
 	//initialize resources map for task
 	task.ResourcesMapUnsafe = make(map[string][]taskresource.TaskResource)
 
+	// initialize network mode so service connect validations are accurate
+	task.initNetworkMode()
+
 	// extract and validate attachments
 	if err := handleTaskAttachments(acsTask, task); err != nil {
 		return nil, err
--- a/agent/api/task/task_attachment_handler.go
+++ b/agent/api/task/task_attachment_handler.go
@@ -25,7 +25,7 @@ import (
 // AttachmentHandler defines an interface to handel attachment received from ACS.
 type AttachmentHandler interface {
 	parseAttachment(acsAttachment *ecsacs.Attachment) error
-	validateAttachment(acsTask *ecsacs.Task) error
+	validateAttachment(acsTask *ecsacs.Task, networkMode string) error
 }
 
 // ServiceConnectAttachmentHandler defines a service connect type attachment handler.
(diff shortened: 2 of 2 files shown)


ubhattacharjya (approved):

ubhattacharjya (approved):

agent/api/serviceconnect/service_connect_validator.go
```diff
@@ -171,7 +172,7 @@ func validateIngressConfig(scIngressConfigList []IngressConfigEntry, taskNetwork
 	}
 
 	switch taskNetworkMode {
-	case BridgeNetworkMode, AWSVPCNetworkMode:
+	case BridgeNetworkMode, AWSVPCNetworkMode, DefaultNetworkMode:
```
rajal-amzn, line 175: Do we need this switch at all if we are going to validate for all scenarios?
If yes, could we instead set the network mode when it is not set?

petderek: Yeah, I found out that the _real_ problem here is that we are validating before detecting the network mode. I've updated the code to accommodate that.

petderek: So effectively this means networkmode will always be set by the time it hits this use case.

agent/api/task/task.go
```diff
@@ -322,6 +322,9 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
 	//initialize resources map for task
 	task.ResourcesMapUnsafe = make(map[string][]taskresource.TaskResource)
 
+        // initialize network mode so service connect validations are accurate
+        task.initNetworkMode()
+
 	// extract and validate attachments
 	if err := handleTaskAttachments(acsTask, task); err != nil {
```
yinyic, line 329: I think we also need to update `handleTaskAttachment` method to use the internal `task.networkMode` and NOT the one from ACS payload (the only difference for the internal one is that we'll fill out the empty aka default mode)

yinyic: https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/api/task/task_attachment_handler.go#L62

ubhattacharjya: https://github.com/aws/amazon-ecs-agent/blob/feature/service-connect/agent/api/task/task.go#L356 also calls initNetworkMode(). Maybe we could remove it from PostUnmarshalTask method?

petderek: its idempotent, so not super problematic as long as it gets called somewhere :)

rajal-amzn (approved):

yinyic:
Closing in favor of https://github.com/aws/amazon-ecs-agent/pull/3295

PR #3295, opened 2022-07-12
SC attachment fixes; Use network mode from ACS payload

yinyic:
### Summary
This PR updates Service Connect config parsing logic, as well as how ECS Agent retrieves task network mode.

### Implementation details

1. Update Service Connect attachment property name to be `serviceconnectdetail`
2. ~~Remove the "dummy SC config" initialization used for dev/testing purposes. We can consume it from Task payload now.~~ (Keeping around for a bit longer for testing)
3. Updated `initNetworkMode` method to 
  a. directly using ACS task payload `NetworkMode` field - this same value used to be populated at container level only, but now it's available at task level too. 
  b. when `NetworkMode` is empty, it means the task has default network mode (docker bridge mode)
4. Move `initNetworkMode` ahead of attachment handlers, because the handlers need network mode
5. Inside attachment handler, use the internal `task.NetworkMode` instead of ACS `NetworkMode` (difference is that we have filled in the default/empty case for the former)

### Testing

New tests cover the changes:  no - all existing tests are updated

SC attachment fixes; Use network mode from ACS payload

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -470,6 +470,8 @@ func (task *Task) initNetworkMode(acsTaskNetworkMode *string) {
 		task.NetworkMode = HostNetworkMode
 	case BridgeNetworkMode, "":
 		task.NetworkMode = BridgeNetworkMode
+	case networkModeNone:
+		task.NetworkMode = networkModeNone
 	default:
 		logger.Warn("Unmapped task network mode", logger.Fields{
 			field.TaskID:      task.GetID(),
@@ -478,7 +480,7 @@ func (task *Task) initNetworkMode(acsTaskNetworkMode *string) {
 	}
 	logger.Info("Task network mode initialized", logger.Fields{
 		field.TaskID:      task.GetID(),
-		field.NetworkMode: aws.StringValue(acsTaskNetworkMode),
+		field.NetworkMode: task.NetworkMode,
 	})
 }
 
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -485,12 +485,36 @@ func (task *Task) initNetworkMode(acsTaskNetworkMode *string) {
 }
 
 func (task *Task) initServiceConnectResources() error {
+	// TODO [SC]: ServiceConnectConfig will come from ACS. Adding this here for dev/testing purposes only Remove when
+	// ACS model is integrated
+	if task.ServiceConnectConfig == nil {
+		task.ServiceConnectConfig = &serviceconnect.Config{
+			ContainerName: "service-connect",
+		}
+	}
 	if task.IsServiceConnectEnabled() {
+		// TODO [SC]: initDummyServiceConnectConfig is for dev testing only, remove it when final SC model from ACS is in place
+		task.initDummyServiceConnectConfig()
 		if err := task.initServiceConnectEphemeralPorts(); err != nil {
 			return err
 		}
 	}
 	return nil
+}
+
+// TODO [SC]: This is for dev testing only, remove it when final SC model from ACS is in place
+func (task *Task) initDummyServiceConnectConfig() {
+	scContainer := task.GetServiceConnectContainer()
+	if _, ok := scContainer.Environment["SC_CONFIG"]; !ok {
+		// no SC_CONFIG :(
+		return
+	}
+	if err := json.Unmarshal([]byte(scContainer.Environment["SC_CONFIG"]), task.ServiceConnectConfig); err != nil {
+		logger.Error("Error parsing SC_CONFIG", logger.Fields{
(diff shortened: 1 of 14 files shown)


agent/api/task/task.go
```diff
@@ -459,66 +460,37 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
-		case "":
-			// the absence of a network mode means the default docker network mode is used, which, for ECS it's bridge.
-			// for future proofing, let's keep inspecting the rest of the containers for this particular case.
-			continue
-		}
+	case HostNetworkMode:
+		task.NetworkMode = HostNetworkMode
+	case BridgeNetworkMode, "":
+		task.NetworkMode = BridgeNetworkMode
+	default:
```
rajal-amzn, line 473: In case of `default`, dont we need to set this to `Bridge`?

agent/logger/field/constants.go
```diff
@@ -35,4 +35,5 @@ const (
 	Image         = "image"
 	Volume        = "volume"
 	Time          = "time"
+	NetworkMode   = "networkMode"
```
rajal-amzn, line 38: Why is this part of `logger`?

agent/api/task/task_attachment_handler.go
```diff
@@ -25,7 +25,7 @@ import (
 // AttachmentHandler defines an interface to handel attachment received from ACS.
 type AttachmentHandler interface {
 	parseAttachment(acsAttachment *ecsacs.Attachment) error
-	validateAttachment(acsTask *ecsacs.Task) error
+	validateAttachment(acsTask *ecsacs.Task, task *Task) error
```
rajal-amzn, line 28: nit: Instead of passing in both type of tasks, it might be enough to pass the Network mode.

agent/logger/field/constants.go
yinyic: This is so that I could add the following logging line in `initNetworkMode`:
```
	logger.Info("Task network mode initialized", logger.Fields{
		field.TaskID:      task.GetID(),
		field.NetworkMode: aws.StringValue(acsTaskNetworkMode),
	})
```

agent/api/task/task_attachment_handler.go
yinyic: I was thinking that future attachment might want to take other fields from the internal task struct.

agent/api/task/task.go
```diff
@@ -459,66 +460,37 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
-			return
-		case HostNetworkMode:
-			task.NetworkMode = HostNetworkMode
-			return
-		case "":
-			// the absence of a network mode means the default docker network mode is used, which, for ECS it's bridge.
-			// for future proofing, let's keep inspecting the rest of the containers for this particular case.
-			continue
-		}
+	case HostNetworkMode:
```
mythri-garaga, line 469: Question: How do we handle when network mode is `none`?

yinyic: `task.networkMode` will be left empty in this case. This is the same behavior on dev branch also, that we don't actually handle `None` explicitly. I think that's because currently in ECS Agent the way we handle networkMode has been 
```
if awsvpc, do ...
if bridge, do ...
if host, do ...
```
So in case of none, it simply won't enter any of those blocks. 
I can add a case for `None` though

mythri-garaga (dismissed):

agent/api/task/task.go
rajal-amzn: Wouldnt `validateAttachment ` again fail when the network mode is empty?

yinyic: We won't expect to get an SC attachment should the network mode is `None` - but I will add a case for `None` network mode as well.

agent/api/task/task.go
```diff
@@ -459,66 +460,39 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
...
-		return
-	}
-	if err := json.Unmarshal([]byte(scContainer.Environment["SC_CONFIG"]), task.ServiceConnectConfig); err != nil {
-		logger.Error("Error parsing SC_CONFIG", logger.Fields{
-			field.Error: err,
-		})
-		return
-	}
-}
-
```
aws-gibbskt, line 521: Can we save these deletions until SC is available in some prod form to facilitate testing still?

yinyic: Sure

mythri-garaga (approved):

mythri-garaga (approved):

mythri-garaga (approved):

chienhanlin (approved):

PR #3301, opened 2022-07-13
Update GO_VERSION to 1.18.3

amogh09:
### Summary
1. Update Go version in `GO_VERSION` file to `1.18.3`. 
2. Upgrade the version of `staticcheck` as the current version is not compatible with Go 1.18.
3. Replace usages of `go get` to install executables with `go install` as `go get` [can no longer be used to install executables](https://go.dev/doc/go-get-install-deprecation).
4. Remove a redundant lock field from `type TaskIAMRoleCredentials`  which is causing static analysis failures because the lock is being passed around by value.
5. Temporarily disable cross-platform builds for Darwin (MacOS) platform in Github CI pipeline as the set of agent code, dependency versions, and Go version (1.18) in this PR does not build for Darwin. We tried fixing the build for Darwin by upgrading the problematic dependencies but that causes agent code to fail compilation. We will tackle this issue separately.

### Testing
Automated tests only.

New tests cover the changes: No new tests

Update Go version for make builds to 1.18.3.

### Licensing


YashdalfTheGray (approved):

agent/credentials/manager.go
```diff
@@ -68,14 +68,10 @@ type IAMRoleCredentials struct {
 type TaskIAMRoleCredentials struct {
 	ARN                string
 	IAMRoleCredentials IAMRoleCredentials
-	lock               sync.RWMutex
```
fierlion, line 71: note: This lock is likely redundant.  There's a taskCredentialsLock which is managing the writes and reads and the usage for this lock is only on reads but is covered by the other outer lock.  Approving

fierlion (approved):

PR #3302, opened 2022-07-13
[WIP] Task handler disconnected mode changes

joshiswarali:
### Testing

New tests cover the changes: 

### Licensing


agent/utils/retry/retry.go
```diff
@@ -15,13 +15,28 @@ package retry
...
 	apierrors "github.com/aws/amazon-ecs-agent/agent/api/errors"
+	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
 )
 
 var _time ttime.Time = &ttime.DefaultTime{}
 
+type TaskEventsFlowController struct {
+	flowControl      map[string]chan bool
```
sparrc, line 30: Maybe I'm misunderstanding, but why is this a channel? I don't fully see where we are using channel concepts with this object. Since we already have the mutex, couldn't this just be `map[string]bool` ?

agent/utils/retry/retry.go
```diff
@@ -15,13 +15,28 @@ package retry
...
+	eventControlLock sync.RWMutex
```
sparrc, line 31: nit: this is a RWMutex but I don't see anywhere we're using read-lock only, so maybe just leave it as sync.Mutex, later code changes could change it to RWMutex if a read-lock is needed or valid somewhere.

agent/eventhandler/task_handler.go
```diff
@@ -334,7 +342,7 @@ func (handler *TaskHandler) submitTaskEvents(taskEvents *taskSendableEvents, cli
 		// If we looped back up here, we successfully submitted an event, but
 		// we haven't emptied the list so we should keep submitting
 		backoff.Reset()
-		retry.RetryWithBackoff(backoff, func() error {
+		retry.RetryWithBackoffForTaskHandler(handler.cfg, handler.eventFlowController, taskARN, handler.delay, backoff, func() error {
```
sparrc, line 345: Maybe I'm oversimplifying, but why does this flow controller need to exist per-task? When connection is resumed it always applies to all tasks doesn't it? Could we do this with a single global mutex/channel/atomic to indicate when connection is established or not?

joshiswarali: Not really sure how it would work with one channel... Afaik a channel will let me do 1:1 communication between two threads, but since there are many threads executing the `RetryWithBackoffForTaskHandler` asynchronously, I need a 1:n communication. In this scenario, if I have multiple threads receiving on a single channel, only one of them will receive the message that I send on the channel

Realmonia (dismissed):

agent/utils/retry/retry.go
```diff
@@ -56,6 +63,90 @@ func RetryWithBackoffCtx(ctx context.Context, backoff Backoff, fn func() error)
...
+			we enter the if block
+
+			If we were in disconnected mode up to this point and switch to normal mode here,
+			eventFlowCtx is closed, and we enter else block
+		*/
+		if cfg.GetDisconnectModeEnabled() && eventFlowCtx != nil {
+
+			/*
+				If we switch to normal mode while we are in this portion before calling WaitForDurationAndInterruptIfRequired,
+				there is a chance that we receive on a closed channel- but this returns immediately
```
Realmonia, line 96: nit: Maybe change wording here as well: closed channel -> cancelled ctx

agent/utils/retry/retry.go
```diff
@@ -56,6 +63,90 @@ func RetryWithBackoffCtx(ctx context.Context, backoff Backoff, fn func() error)
...
+
+		retriableErr, isRetriableErr := err.(apierrors.Retriable)
+
+		//if unretriable error in disconnected mode, don't return
+		if err == nil || (!cfg.GetDisconnectModeEnabled() && isRetriableErr && !retriableErr.Retry()) {
+			return err
+		}
+
+		/*
+			If we switch to disconnected mode after executing the previous code block,
```
Realmonia, line 86: nit: change `switch to` to `were in`?

agent/eventhandler/task_handler.go
```diff
@@ -449,3 +463,25 @@ func handleInvalidParamException(err error, events *list.List, eventToSubmit *li
...
+	handler.eventFlowCtxLock.Lock()
+	defer handler.eventFlowCtxLock.Unlock()
+
+	if handler.eventFlowCtxCancel != nil {
+		handler.eventFlowCtxCancel()
+	}
+}
+
+//will be called by acs handler while toggling disconnectModeEnabled false to true
+func (handler *TaskHandler) PauseEventsFlow() {
```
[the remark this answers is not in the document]
joshiswarali: Yes, this will be called from ACS handler when disconnectModeEnabled is toggled from false to true
I will change the PR summary to reflect this

agent/app/agent.go
```diff
@@ -408,7 +408,7 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 	deregisterInstanceEventStream := eventstream.NewEventStream(
 		deregisterContainerInstanceEventStreamName, agent.ctx)
 	deregisterInstanceEventStream.StartListening()
-	taskHandler := eventhandler.NewTaskHandler(agent.ctx, agent.dataClient, state, client)
+	taskHandler := eventhandler.NewTaskHandler(agent.ctx, agent.dataClient, state, client, agent.cfg, 10*time.Minute)
```
sparrc, line 411: Took me a lot of code diving to figure out what this `10*time.Minute` actually does here. Could you make this a constant with an informative name? IIUC something along the lines of `const disconnectedModeTaskEventRetryDelay time.Duration = 10*time.Minute`

agent/eventhandler/task_handler.go
```diff
@@ -77,9 +79,14 @@ type TaskHandler struct {
...
 	maxDrainEventsFrequency time.Duration
 
-	state  dockerstate.TaskEngineState
-	client api.ECSClient
-	ctx    context.Context
+	state              dockerstate.TaskEngineState
+	client             api.ECSClient
+	ctx                context.Context
+	cfg                *config.Config
+	delay              time.Duration
```
sparrc, line 86: like the comment above, `delay` is a bit generic of a name since this is specifically only used for disconnected mode task event flow retry backoff (IIUC)

agent/utils/retry/retry.go
```diff
@@ -56,6 +63,91 @@ func RetryWithBackoffCtx(ctx context.Context, backoff Backoff, fn func() error)
...
+	select {
+	case <-reconnectTimer.C:
+		return true
+
+	default:
+		return false
+	}
+}
+
+func WaitForDurationAndInterruptIfRequired(eventFlowCtx context.Context, delay time.Duration) bool {
```
sparrc, line 130: nit: the more idiomatic name for this would be `waitForDurationWithContext` (see https://go.dev/blog/context-and-structs)

agent/eventhandler/task_handler.go
```diff
@@ -116,6 +126,10 @@ func NewTaskHandler(ctx context.Context,
 		client:                    client,
 		minDrainEventsFrequency:   minDrainEventsFrequency,
 		maxDrainEventsFrequency:   maxDrainEventsFrequency,
+		cfg:                       cfg,
+		delay:                     delay,
+		eventFlowCtx:              nil,
```
sparrc, line 131: Is there a downside to initializing these? might be safer than intentionally passing around nils unless the nil value is useful for something?

joshiswarali: nil is not really useful yeah.. I can initialize eventFlowCtx to context.Background() but couldn't figure out what value to assign to eventFlowCtxCancel

agent/app/agent.go
```diff
@@ -408,7 +408,9 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 	deregisterInstanceEventStream := eventstream.NewEventStream(
 		deregisterContainerInstanceEventStreamName, agent.ctx)
 	deregisterInstanceEventStream.StartListening()
-	taskHandler := eventhandler.NewTaskHandler(agent.ctx, agent.dataClient, state, client)
+
+	const disconnectedModeTaskEventRetryDelay time.Duration = 10 * time.Minute
```
Realmonia, line 412: Can you move this to top of the file (together with other consts)

joshiswarali: Makes sense- done

agent/eventhandler/task_handler.go
joshiswarali: Since I've removed the nil initializations, also removed nil checks done for those variables in a couple of places

Realmonia (approved):

sparrc (dismissed):

sparrc (approved):

Realmonia (approved):

PR #3315, opened 2022-07-21
Task handler disconnected mode bug fix- Pass context by reference and initialize to nil

joshiswarali:
### Summary

Fixes the following bugs-

Bug 1:
The `RetryWithBackoffForTaskHandler` method retries task state change events when agent loses network connection. We currently pass a context to `RetryWithBackoffForTaskHandler` to control the retry frequency for task state changes in disconnected mode. Currently, when agent loses network connection, ACS handler starts a timer of 5 minutes and once the timer is completed, agent toggles `disconnectModeEnabled` flag to true and initializes a context. For a task that runs for less than 5 minutes, it is possible that we have invoked `RetryWithBackoffForTaskHandler` before executing `handler.eventFlowCtx, handler.eventFlowCtxCancel = context.WithCancel(handler.ctx)`. Hence, we need to pass the context to `RetryWithBackoffForTaskHandler` by reference and not by value.

Bug 2:
When ACS handler switches to disconnected mode, it does the following steps- 
A1: toggle disconnectModeEnabled to true
A2: initialize context to a specific value derived from parent

The `RetryWithBackoffForTaskHandler` of task handler does the following-
T1: check for disconnectModeEnabled = true
T2: pass initialized context value to `WaitForDurationWithContext`

Our ideal execution sequence for these steps is A1, A2, T1, T2 however it is not guaranteed that it will be so, as the `RetryWithBackoffForTaskHandler` runs asynchrononously. 

If the sequence is A1, T1, T2, A2- we pass a context with cancellation to `WaitForDurationWithContext`. Later we re-initialize the context (and the cancellation function) again, and hence we are unable to cancel the context that `WaitForDurationWithContext` is already receiving on. If we initialize `eventFlwCtx` as `nil`, we can avoid this scenario.

### Implementation details

1. Pass the address and not the value of handler.eventFlowCtx to `RetryWithBackoffForTaskHandler` (pass by reference).
2. Initialize eventFlwCtx as nil

### Testing

Ran a task that stops after 5 min and disconnected network connection as soon as task is started. This ensures that we call `RetryWithBackoffForTaskHandler` before executing `handler.eventFlowCtx, handler.eventFlowCtxCancel = context.WithCancel(handler.ctx)`.

### Licensing

### DIFF
--- a/agent/eventhandler/task_handler.go
+++ b/agent/eventhandler/task_handler.go
@@ -115,7 +115,6 @@ func NewTaskHandler(ctx context.Context,
 	disconnectedModeTaskEventRetryDelay time.Duration) *TaskHandler {
 	// Create a handler and start the periodic event drain loop
 
-	eventFlowCtx, eventFlowCtxCancel := context.WithCancel(context.Background())
 	taskHandler := &TaskHandler{
 		ctx:                                 ctx,
 		tasksToEvents:                       make(map[string]*taskSendableEvents),
@@ -129,8 +128,8 @@ func NewTaskHandler(ctx context.Context,
 		maxDrainEventsFrequency:             maxDrainEventsFrequency,
 		cfg:                                 cfg,
 		disconnectedModeTaskEventRetryDelay: disconnectedModeTaskEventRetryDelay,
-		eventFlowCtx:                        eventFlowCtx,
-		eventFlowCtxCancel:                  eventFlowCtxCancel,
+		eventFlowCtx:                        nil,
+		eventFlowCtxCancel:                  nil,
 	}
 	go taskHandler.startDrainEventsTicker()
 
@@ -349,7 +348,7 @@ func (handler *TaskHandler) submitTaskEvents(taskEvents *taskSendableEvents, cli
 		// If we looped back up here, we successfully submitted an event, but
 		// we haven't emptied the list so we should keep submitting
 		backoff.Reset()
-		retry.RetryWithBackoffForTaskHandler(handler.cfg, taskARN, handler.disconnectedModeTaskEventRetryDelay, backoff, handler.eventFlowCtx, func() error {
+		retry.RetryWithBackoffForTaskHandler(handler.cfg, taskARN, handler.disconnectedModeTaskEventRetryDelay, backoff, &handler.eventFlowCtx, func() error {
 			// Lock and unlock within this function, allowing the list to be added
 			// to while we're not actively sending an event
 			seelog.Debug("TaskHandler: Waiting on semaphore to send events...")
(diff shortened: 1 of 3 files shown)


agent/utils/retry/retry.go
```diff
@@ -132,7 +133,7 @@ func WaitForDurationWithContext(eventFlowCtx context.Context, delay time.Duratio
 		})
 		return true
 
-	case <-eventFlowCtx.Done():
+	case <-(*eventFlowCtx).Done():
```
Realmonia, line 136: I think you can still use `eventFlowCtx.Done()` here

joshiswarali: If I use eventFlowCtx.Done() I get an error saying .Done() is undefined- i guess that's because eventFlowCtx is a pointer

Realmonia: I thought golang automatically dereference the pointer when it's selector expression - https://go.dev/ref/spec#Selectors, but nvm then

Realmonia (approved):

agent/eventhandler/task_handler.go
```diff
@@ -349,7 +348,7 @@ func (handler *TaskHandler) submitTaskEvents(taskEvents *taskSendableEvents, cli
 		// If we looped back up here, we successfully submitted an event, but
 		// we haven't emptied the list so we should keep submitting
 		backoff.Reset()
-		retry.RetryWithBackoffForTaskHandler(handler.cfg, taskARN, handler.disconnectedModeTaskEventRetryDelay, backoff, handler.eventFlowCtx, func() error {
+		retry.RetryWithBackoffForTaskHandler(handler.cfg, taskARN, handler.disconnectedModeTaskEventRetryDelay, backoff, &handler.eventFlowCtx, func() error {
```
chienhanlin, line 351: Q. Would `handler.eventFlowCtx` be `nil` and cause an issue here?

joshiswarali: I wasn't sure either so tried this-

// You can edit this code!
// Click here and start typing.
package main

import (
	"context"
	"fmt"
)

type DummyHandler struct {
	ctx    context.Context
	cancel context.CancelFunc
}

func NewDummyHandler() *DummyHandler {

	return &DummyHandler{ctx: nil, cancel: nil} //initialize as nil
}

func main() {
	fmt.Println("Hello, 世界")

	handler := NewDummyHandler()
	NilRef(&handler.ctx, &handler.cancel)

	handler.ctx = context.TODO()
	NilRef(&handler.ctx, &handler.cancel)
}

func NilRef(ctx *context.Context, cancel *context.CancelFunc) {

	if *ctx == nil {
		fmt.Println("ctx is initialized as nil")
	} else {
		fmt.Println("ctx has a non-nil initial value")
	}
}

This bit of code worked

chienhanlin (approved):

PR #3324, opened 2022-08-01
Add ecs-init build tag argument in integrated spec files

fierlion:
### Summary
This adds the build tag to the ecs-init gobuild.sh script, which will point to the correct ecs-init config
https://github.com/aws/amazon-ecs-init/blob/5f7f9336bc3c5314206aefd212e9ebbfc3bdab60/ecs-init/config/config_al2.go#L1

### Implementation details
the tags are already defined in the spec files, this was just a miss in our setup/testing.

### Testing
Built using `make generic-rpm-integrated` on a fresh instance of latest ecs-optimized AMI. 
```
sudo systemctl stop ecs
sudo rpm -e <old rpm file>
sudo rpm -i <built rpm file>
sudo systemctl start ecs
```
then a docker inspect of the launched agent container to be sure we see:
```
            {
                "Type": "bind",
                "Source": "/sys/fs/cgroup",
                "Destination": "/sys/fs/cgroup",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
            },
```

Fix ecs-init build tag bug for integrated builds.

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami-integrated/ecs-agent.spec
+++ b/packaging/amazon-linux-ami-integrated/ecs-agent.spec
@@ -163,7 +163,7 @@ required routes among its preparation steps.
 ./scripts/build-cni-plugins
 ./scripts/build-integrated true "" false true
 ./scripts/build-agent-image
-./scripts/gobuild.sh
+./scripts/gobuild.sh %{gobuild_tag}
 
 %install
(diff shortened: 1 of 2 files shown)


YashdalfTheGray (approved):
Good catch on this!

ubhattacharjya (approved):

PR #3328, opened 2022-08-04
swapped the impl of GetAllNetworkInterfaces from Golang net pkg to pwsh cmdlet

rawahars:
### Summary

As of August 2022, Windows instances are facing an issue wherein Win32 APIs are not working as expected in intermittent cases. Essentially, the callbacks for `NotifyIpInterfaceChange API` are not working as well as `GetAdaptersAddresses API` is not returning all the network interfaces on the instance. The second issue leads to `net.Interfaces()` failing as well.

 In order to ensure that tasks using `awsvpc network mode` continue working on Windows nodes, we are swapping the existing mechanism for periodic reconciliation to an alternate mechanism wherein a Powershell cmdlet is being executed to fetch all the network interfaces on the instance. Therefore, in the happy path, Win32 NotifyIpInterfaceChange API callback would be invoked and the ENI would be acknowledged promptly.

However, in the scenario wherein WIN32 IP Helper APIs fail to work as expected, we would rely upon the periodic reconciliation which happens every 30 seconds. Since the actual ENI timeout is much greater than this reconciliation period, the ENI timeout would be prevented.

### Implementation details

We have a `netwrapper` package which presently abstracts Golang's net package. This in turn is used in netutils and then in watcher packages.
For the implementation change suggested above, we are swapping the generic `net.Interfaces()` call with platform specific implementation.

### Testing

New unit tests have been added to validate this implementation.
From functional testing viewpoint, a new agent was built and deployed on `Windows Server 2019 Full` and `Windows Server 2022 Full` instances. In this process, we disabled the callbacks so that the actual scenario was replicated.
We then launched tasks (4-6 simultaneously on same node) over a period of 5 hours.

New tests cover the changes: 
Yes

windows: swapped the impl of GetAllNetworkInterfaces from Golang net pkg to pwsh cmdlet
### Licensing

### DIFF
--- a/agent/eni/netwrapper/netwrapper_windows.go
+++ b/agent/eni/netwrapper/netwrapper_windows.go
@@ -1,0 +1,90 @@
+//go:build windows
+// +build windows
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package netwrapper
+
+import (
+	"encoding/json"
+	"net"
+	"os/exec"
+
+	log "github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+type windowsNetworkAdapter struct {
+	Name       string `json:"ifAlias"`
+	MacAddress string `json:"MacAddress"`
+	IfIndex    int    `json:"ifIndex"`
+	MTU        int    `json:"ActiveMaximumTransmissionUnit"`
+}
+
+type winAdapters []windowsNetworkAdapter
+
+// Making it visible for unit testing
+var execCommand = exec.Command
+
+// getAllNetworkInterfaces returns all the network adapters.
+// Unfortunately, as of August 2022, the Win32 IPHelper API seems to be flaky wherein
+// the same stops responding in few intermittent cases. Therefore, to avoid failures
+// in such cases, we have switched to using Powershell cmdlet for fetching all the interfaces.
+func (utils *utils) getAllNetworkInterfaces() ([]net.Interface, error) {
+	log.Info("Running Powershell command to fetch all the network adapters")
+
+	// This command returns a JSON array of all the network interfaces on the instance.
+	cmd := "Get-NetAdapter | select ifAlias, ifIndex, MacAddress, ActiveMaximumTransmissionUnit | ConvertTo-Json"
+	out, err := execCommand("Powershell", "-C", cmd).Output()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to execute Get-NetAdapter for retrieving all interfaces")
+	}
+	log.Debugf("Output of Get-NetAdapters Powershell command execution: %s", out)
+
+	// Convert the obtained JSON array into an array of structs.
+	var data winAdapters
+	err = json.Unmarshal(out, &data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal the json of network adapters")
+	}
+
+	// This is an error condition since the instance would have at least 1 ENI.
+	if len(data) == 0 {
+		return nil, errors.Errorf("invalid number of network adapters found on the instance")
+	}
+	log.Debugf("Unmarshalled Powershell output into an array of structs")
+
+	// Convert the data into an array of net.Interface
+	var ifaces []net.Interface
+	for _, adapter := range data {
+		mac, err := net.ParseMAC(adapter.MacAddress)
+		if err != nil {
+			// Log the error and continue to get as many adapters as possible.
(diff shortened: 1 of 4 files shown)


vsiddharth (approved):
Thanks a lot for making this change quickly. 

It would be great to get this tested and pushed out in the upcoming release.

agent/eni/netwrapper/netwrapper_windows.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type WinAdapters []WindowsNetworkAdapter
+
+// Making it visible for unit testing
+var execCommand = exec.Command
+
+// getAllNetworkInterfaces returns all the network adapters.
+// Unfortunately, as of August 2022, the Win32 IPHelper API seems to be flaky wherein
+// the same stops responding in few intermittent cases. Therefore, to avoid failures
```
fierlion, line 42: Is there an issue raised with the Win32 IPHelper API, or work being done on fixing the flakiness?  Have the few intermittent cases been defined?

agent/eni/netwrapper/netwrapper_windows.go
```diff
@@ -0,0 +1,88 @@
...
+	log.Debugf("Output of Get-NetAdapters Powershell command execution: %s", out)
+
+	// Convert the obtained JSON array into an array of structs.
+	data := WinAdapters{}
+	err = json.Unmarshal(out, &data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal the json of network adapters")
+	}
+
+	// This seems to be an error condition since the instance would have at least 1 ENI.
```
fierlion, line 62: nit: I would remove the "seems" -- the code is defining it as an error condition.

agent/eni/netwrapper/netwrapper_windows_test.go
```diff
@@ -0,0 +1,155 @@
...
+func TestNetworkAdaptersProcessSuccess(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+	fmt.Fprintf(os.Stdout, networkAdapterValidResult)
+	os.Exit(0)
+}
+
+// fakeExecCommandForNetworkAdaptersInvalidOutput returns an invalid output from exec cmd.
+func fakeExecCommandForNetworkAdaptersInvalidOutput(command string, args ...string) *exec.Cmd {
```
fierlion, line 81: nit -- can we call this `mock` instead of `fake`?  (That goes for other instances of the same.)

agent/eni/netwrapper/netwrapper_windows.go
jterry75: Yes we opened a ticket with support on this

singholt (dismissed):

agent/eni/netwrapper/netwrapper_windows.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"encoding/json"
+	"net"
+	"os/exec"
+
+	log "github.com/cihub/seelog"
+	"github.com/pkg/errors"
+)
+
+type WindowsNetworkAdapter struct {
```
jterry75, line 28: These should not be exported

agent/eni/netwrapper/netwrapper_windows.go
```diff
@@ -0,0 +1,88 @@
...
+	// This command returns a JSON array of all the network interfaces on the instance.
+	cmd := "Get-NetAdapter | select ifAlias, ifIndex, MacAddress, ActiveMaximumTransmissionUnit | ConvertTo-Json"
+	out, err := execCommand("Powershell", "-C", cmd).Output()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to execute Get-NetAdapter for retrieving all interfaces")
+	}
+	log.Debugf("Output of Get-NetAdapters Powershell command execution: %s", out)
+
+	// Convert the obtained JSON array into an array of structs.
+	data := WinAdapters{}
```
jterry75, line 56: I'm pretty sure you can `var data WinAdapters` then `&data` below. You dont need to initialize to 0 size.

agent/eni/netwrapper/netwrapper_windows.go
jterry75: Agreed, `this is an error condition`

agent/eni/netwrapper/netwrapper_windows.go
```diff
@@ -0,0 +1,88 @@
...
+	// This seems to be an error condition since the instance would have at least 1 ENI.
+	if len(data) == 0 {
+		return nil, errors.Errorf("invalid number of network adapters found on the instance")
+	}
+	log.Debugf("Unmarshalled Powershell output into an array of structs")
+
+	// Convert the data into an array of net.Interface
+	var ifaces []net.Interface
+	for _, adapter := range data {
+		mac, err := net.ParseMAC(adapter.MacAddress)
```
jterry75, line 71: Is there ever a case where the Mac of a v adapter can be empty?

jterry75 (dismissed):

agent/eni/netwrapper/netwrapper_windows.go
rawahars: As discussed, we are now just logging the error and continuing to add other adapters.

jterry75 (approved):

singholt (approved):

vsiddharth (approved):

YashdalfTheGray (approved):

PR #3329, opened 2022-08-04
Run AppNet Relay as non-root and remove it from introspection response; fix AppNet container pull error

yinyic:
### Summary

1. Run AppNet Relay container as non-privileged user 1337, same as task AppNet container. 
2. We've modeled AppNet Relay as an internal ECS task in a previous [PR](https://github.com/aws/amazon-ecs-agent/pull/3280) to leverage the existing task/container lifecycle management mechanism. In this PR we take this task out from introspection endpoint (http://localhost:51678/v1/tasks) response as it should be kept invisible to customers. We do this by introducing an `IsInternal` field to task model, and setting it to true for AppNet relay indicating that it's an internally managed task (not initiated by customer). Upon receiving request to list all tasks with Introspection API, we filter out internal tasks. 
3. Do not try to pull container images for AppNet relay or task container, as the images are loaded by ECS Agent locally.

### Testing

Tested new Agent with Service Connect task and verified that
1. Task is able to start with Relay running as non-privileged user, and task AppNet container able to retrieve cluster config
2. AppNet relay task is removed from introspection response from `curl http://localhost:51678/v1/tasks`
3. No longer seeing image pull errors for AppNet containers.

Run AppNet Relay as non-root; Remove it from introspection response; Fix AppNet container pull error.

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -1015,8 +1015,12 @@ func (engine *DockerTaskEngine) GetTaskByArn(arn string) (*apitask.Task, bool) {
 
 func (engine *DockerTaskEngine) pullContainer(task *apitask.Task, container *apicontainer.Container) dockerapi.DockerContainerMetadata {
 	switch container.Type {
-	case apicontainer.ContainerCNIPause, apicontainer.ContainerNamespacePause:
-		// pause images are managed at startup
+	case apicontainer.ContainerCNIPause, apicontainer.ContainerNamespacePause, apicontainer.ContainerServiceConnectRelay:
+		// pause images and AppNet relay image are managed at startup
+		return dockerapi.DockerContainerMetadata{}
+	}
+	// AppNet Agent container image is also managed at start up (it uses the same image as AppNet Relay container)
+	if task.IsServiceConnectEnabled() && container == task.GetServiceConnectContainer() {
 		return dockerapi.DockerContainerMetadata{}
 	}
 
--- a/agent/engine/dockerstate/docker_task_engine_state.go
+++ b/agent/engine/dockerstate/docker_task_engine_state.go
@@ -30,6 +30,9 @@ import (
 type TaskEngineState interface {
 	// AllTasks returns all of the tasks
 	AllTasks() []*apitask.Task
+	// AllExternalTasks returns all tasks with IsInternal==false (i.e. customer-initiated tasks).
+	// Currently, ServiceConnect AppNet Relay task is the only internal task.
+	AllExternalTasks() []*apitask.Task
 	// AllENIAttachments returns all of the eni attachments
 	AllENIAttachments() []*apieni.ENIAttachment
 	// AllImageStates returns all of the image.ImageStates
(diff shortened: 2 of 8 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1015,8 +1015,12 @@ func (engine *DockerTaskEngine) GetTaskByArn(arn string) (*apitask.Task, bool) {
...
 func (engine *DockerTaskEngine) pullContainer(task *apitask.Task, container *apicontainer.Container) dockerapi.DockerContainerMetadata {
 	switch container.Type {
-	case apicontainer.ContainerCNIPause, apicontainer.ContainerNamespacePause:
-		// pause images are managed at startup
+	case apicontainer.ContainerCNIPause, apicontainer.ContainerNamespacePause, apicontainer.ContainerServiceConnectRelay:
+		// pause images and AppNet relay image are managed at startup
+		return dockerapi.DockerContainerMetadata{}
+	}
+	// AppNet Agent container image is also managed at start up (it uses the same image as AppNet Relay container)
+	if task.IsServiceConnectEnabled() && container == task.GetServiceConnectContainer() {
```
aws-gibbskt, line 1023: would it not be easier to just make the containerType of the ServiceConnectContainer within the task be of type apicontainer.ContainerServiceConnectRelay ? then it is handled in the previous switch block.

ubhattacharjya (dismissed):

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -148,17 +151,28 @@ func (state *DockerTaskEngineState) AllTasks() []*apitask.Task {
 	state.lock.RLock()
 	defer state.lock.RUnlock()
 
-	return state.allTasksUnsafe()
+	return state.allTasksUnsafe(false)
 }
 
-func (state *DockerTaskEngineState) allTasksUnsafe() []*apitask.Task {
+func (state *DockerTaskEngineState) allTasksUnsafe(excludeInternal bool) []*apitask.Task {
```
aws-gibbskt, line 157: I wouldn't change the prototype here. I would simply create a new function that this calls with false and the new AllExternalTasks calls with true. like getFillteredTasksUnsafe(excludeInternal bool)

agent/engine/docker_task_engine.go
yinyic: Synced offline - there are a few places where we use container type `ContainerServiceConnectRelay` to specify relay-only behavior, which are not applicable to task AppNet containers.

agent/engine/dockerstate/docker_task_engine_state.go
yinyic: `allTasksUnsafe` is an internal function, used by the exported `AllTasks`. I added a new exported method `AllExternalTasks`, but want to reuse the `allTasksUnsafe` to iterate through all tasks. So it looks like 
```
AllTasks() 
  -> allTaskUnsafe(false)

AllExternalTasks() 
  -> allTaskUnsafe(true)
```
Is this what you are suggesting?

aws-gibbskt: sort of, I see a lot of diffs that are not needed and don't make sense, basically any line that just adds a "false"
```go
AllTasks() 
  -> allTaskUnsafe()
      -> getFilteredTasksUnsafe(false)

AllExternalTasks() 
  -> allExternalTasksUnsafe()
      -> getFilteredTasksUnsafe(false)
```

So the non-Unsafe versions are external yes, but internally the Usafe version is used a lot. The above is a more structured breakdown of responsibilities. External "safe" versions exist, if you already have the lock then use the "unsafe" versions which are exactly equivalent to the "safe" version from an API perspective, but the unsafe versions are 99% the same so they share a common impl in getFilteredTasksUnsafe.

mythri-garaga (dismissed):

agent/engine/dockerstate/docker_task_engine_state.go
yinyic: Thanks for clarifying. Makes sense to me!

aws-gibbskt (approved):

mythri-garaga (approved):

ubhattacharjya (approved):

PR #3341, opened 2022-08-15
Fix memory leak while creating multiple httpClient

ubhattacharjya:
### Summary

Problem Statement: While performance testing ECS agent with 12 tasks, noticed a constantly increasing memory consumption of ECS agent. 

```
go tool pprof heap.pprof
Type: inuse_space
Time: Aug 15, 2022 at 12:51am (UTC)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top10
Showing nodes accounting for 412.59MB, 93.56% of 440.97MB total
Dropped 88 nodes (cum <= 2.20MB)
Showing top 10 nodes out of 50
      flat  flat%   sum%        cum   cum%
  136.53MB 30.96% 30.96%   136.53MB 30.96%  bufio.NewReaderSize (inline)
  123.98MB 28.12% 59.08%   123.98MB 28.12%  bufio.NewWriterSize (inline)
   35.03MB  7.94% 67.02%    35.03MB  7.94%  net/http.(*Transport).queueForIdleConn
   25.52MB  5.79% 72.81%    36.02MB  8.17%  net/http.(*Transport).tryPutIdleConn
   24.01MB  5.44% 78.25%    24.01MB  5.44%  runtime.malg
   20.50MB  4.65% 82.90%   288.51MB 65.43%  net/http.(*Transport).dialConn
   17.02MB  3.86% 86.76%    17.02MB  3.86%  runtime.allocm
      16MB  3.63% 90.39%    58.04MB 13.16%  github.com/aws/amazon-ecs-agent/agent/api/appnet.performAppnetRequest
       7MB  1.59% 91.98%    10.50MB  2.38%  net/http.(*connLRU).add
       7MB  1.59% 93.56%    43.02MB  9.76%  net/http.(*persistConn).readLoop
(pprof) list bufio.NewReaderSize
Total: 440.97MB
ROUTINE ======================== bufio.NewReaderSize in /usr/local/go/src/bufio/bufio.go
  136.53MB   136.53MB (flat, cum) 30.96% of Total
         .          .     51:		return b
         .          .     52:	}
         .          .     53:	if size < minReadBufferSize {
         .          .     54:		size = minReadBufferSize
         .          .     55:	}
    1.50MB     1.50MB     56:	r := new(Reader)
  135.03MB   135.03MB     57:	r.reset(make([]byte, size), rd)
         .          .     58:	return r
         .          .     59:}
         .          .     60:
         .          .     61:// NewReader returns a new Reader whose buffer has the default size.
         .          .     62:func NewReader(rd io.Reader) *Reader {
(pprof
```

```
CONTAINER ID   NAME        CPU %     MEM USAGE / LIMIT     MEM %     NET I/O   BLOCK I/O        PIDS
3f4ff1533ffe   ecs-agent   0.55%     344.1MiB / 246.4GiB   0.14%     0B / 0B   152MB / 14.1MB   72
```
The pprof output indicates that the method performAppnetRequest which calls net.Dial is responsible for 65% of the memory usage of ecs-agent. This was possibly because [https://github.com/golang/go/pull/50799](https://github.com/golang/go/pull/50799/files#diff-f2f92ffe0abe8dd3c833d435c2d859d54380e8e4160af8becab6945395563cfe)
 
This was introduced because, this [https://github.com/aws/amazon-ecs-agent/pull/3332](PR)  was creating http Client for every stats request every minute. Due to this memory leak issue with http.Transport, the memory consumption would continuously increase. Earlier to this PR, the udsHttpClient was common to all appnet Client which would give rise to race conditions when we are running a lot of tasks as stats are collected concurrently. Ref: https://github.com/aws/amazon-ecs-agent/pull/3224

The issue was fixed by making the udshttpClient an attribute of appnet Client. ServiceConnectStats object will store this client as its attribute for every task. So the udsHttpClient is created only once per task for collecting stats

### Testing

Tested by running 12 tasks with new agent and saw that the GoroutineCount is not exponentially increasing but is constant. The memory usage is also within a constant range after running these tasks for over 12 hrs.

### Licensing

### DIFF
--- a/agent/api/appnet/client.go
+++ b/agent/api/appnet/client.go
@@ -13,14 +13,42 @@
 
 package appnet
 
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+)
+
+type appnetClientCtxKey int
+
 type client struct {
+	udsHttpClient http.Client
 }
 
-// This client is a singleton. No need to create multiple clients since the only thing that veries between requests is
-// the Appnet's admin UDS
-var defaultClient = &client{}
+const (
+	udsAddressKey   appnetClientCtxKey = iota
+	unixNetworkName                    = "unix"
+)
 
 // Client retrieves the singleton Appnet client
 func Client() *client {
-	return defaultClient
+	return &client{
+		udsHttpClient: http.Client{
+			Transport: &http.Transport{
+				DialContext: udsDialContext,
+			},
+		},
+	}
+}
+
+func udsDialContext(ctx context.Context, _, _ string) (net.Conn, error) {
+	udsPath, ok := ctx.Value(udsAddressKey).(string)
+	if !ok {
+		return nil, fmt.Errorf("appnet client: Path to appnet admin socket was not a string")
+	}
+	if udsPath == "" {
+		return nil, fmt.Errorf("appnet client: Path to appnet admin socket was blank")
+	}
+	return net.Dial(unixNetworkName, udsPath)
 }
--- a/agent/stats/service_connect_linux.go
+++ b/agent/stats/service_connect_linux.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/api/appnet"
 
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
@@ -31,9 +32,10 @@ import (
 )
 
 type ServiceConnectStats struct {
-	stats []*ecstcs.GeneralMetricsWrapper
-	sent  bool
-	lock  sync.RWMutex
+	stats        []*ecstcs.GeneralMetricsWrapper
+	appnetClient api.AppnetClient
+	sent         bool
+	lock         sync.RWMutex
 }
 
 const (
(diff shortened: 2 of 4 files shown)


agent/stats/service_connect_linux.go
```diff
@@ -199,6 +203,13 @@ func (sc *ServiceConnectStats) HasStatsBeenSent() bool {
 	return sc.sent
 }
 
+func (sc *ServiceConnectStats) GetAppnetClient() api.AppnetClient {
```
yinyic, line 206: Can we make this not exported?

aws-gibbskt: Can we instead remove this function? There is no way to modify this client nor is it expected so we don't need to lock around it. It is provided at construction time and isn't messed with after. This is just an unnecessary lock grab.

ubhattacharjya: makes sense. Removed it

agent/api/appnet/client.go
```diff
@@ -13,14 +13,51 @@
...
+		return nil, fmt.Errorf("appnet client: Path to appnet admin socket was blank")
+	}
+	return net.Dial(unixNetworkName, udsPath)
+}
+
+func (cl *client) GetHTTPClient() http.Client {
+	cl.lock.RLock()
+	defer cl.lock.RUnlock()
+
+	return cl.udsHttpClient
```
aws-gibbskt, line 62: I think this is the same as the comment in the other file. udsHttpClient is provided at struct creation and never changed. You can remove the lock completely since it is only used here. Also this exported function is only used internally and thus can be removed similar to the other case.

aws-gibbskt (approved):

agent/api/appnet/client.go
ubhattacharjya: makes sense. changed it

yinyic (approved):

mythri-garaga (approved):

PR #3350, opened 2022-08-18
Add types for task protection handlers

amogh09:
### Summary
Add new types to be used in Task scale-in protection API handlers.

### Implementation details
A couple of new types, their constructors, and getters.

### Testing
Unit tests

Add new types for task scale-in protection API

### Licensing

### DIFF
--- a/agent/handlers/agentapi/v1/taskprotection/types/types.go
+++ b/agent/handlers/agentapi/v1/taskprotection/types/types.go
@@ -1,0 +1,55 @@
+package types
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Protection type to set on the task
+type taskProtectionType string
+
+const (
+	TaskProtectionTypeScaleIn  taskProtectionType = "SCALE_IN"
+	TaskProtectionTypeDisabled taskProtectionType = "DISABLED"
+	// TODO Check if DISABLED_PERMANENTLY protection type is to be supported
+)
+
+// Converts a string to taskProtectionType
+func taskProtectionTypeFromString(s string) (taskProtectionType, error) {
+	switch p := strings.ToUpper(s); p {
+	case string(TaskProtectionTypeScaleIn):
+		return TaskProtectionTypeScaleIn, nil
+	case string(TaskProtectionTypeDisabled):
+		return TaskProtectionTypeDisabled, nil
+	default:
+		return "", fmt.Errorf("unknown task protection type: %v", s)
+	}
+}
+
+// Protection for a Task
+type taskProtection struct {
+	protectionType           taskProtectionType
+	protectionTimeoutMinutes *int
+}
+
+// Creates a taskProtection value after validating the arguments
+func NewTaskProtection(protectionType string, timeoutMinutes *int) (*taskProtection, error) {
+	ptype, err := taskProtectionTypeFromString(protectionType)
+	if err != nil {
+		return nil, fmt.Errorf("protection type is invalid: %w", err)
+	}
+
+	if timeoutMinutes != nil && *timeoutMinutes <= 0 {
+		return nil, fmt.Errorf("protection timeout must be greater than zero")
+	}
+
+	return &taskProtection{protectionType: ptype, protectionTimeoutMinutes: timeoutMinutes}, nil
+}
+
+func (taskProtection *taskProtection) GetProtectionType() taskProtectionType {
+	return taskProtection.protectionType
+}
+
+func (taskProtection *taskProtection) GetProtectionTimeoutMinutes() *int {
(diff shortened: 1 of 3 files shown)


agent/handlers/agentapi/v1/taskprotection/types/types.go
```diff
@@ -0,0 +1,54 @@
+package types
```
chienhanlin, line 1: nit. Missing header
```
// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
//
// Licensed under the Apache License, Version 2.0 (the "License"). You may
// not use this file except in compliance with the License. A copy of the
// License is located at
//
//	http://aws.amazon.com/apache2.0/
//
// or in the "license" file accompanying this file. This file is distributed
// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
```

amogh09: Thanks. Will add it.

amogh09: Added.

chienhanlin: Thank you!

agent/handlers/agentapi/v1/taskprotection/types/types.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// Creates a taskProtection value after validating the arguments
+func NewTaskProtection(protectionType string, timeoutMinutes *int) (*taskProtection, error) {
+	ptype, err := taskProtectionTypeFromString(protectionType)
+	if err != nil {
+		return nil, fmt.Errorf("protection type is invalid: %w", err)
+	}
+
+	if timeoutMinutes != nil && *timeoutMinutes <= 0 {
```
chienhanlin, line 54: Should we also validate the max of timeoutMinutes?

agent/handlers/agentapi/v1/taskprotection/types/types_test.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// Helper function for testing taskProtectionTypeFromString
+func testTaskProtectionTypeFromString(t *testing.T, input string, expected taskProtectionType) {
+	output, err := taskProtectionTypeFromString(input)
+	assert.NoError(t, err)
+	assert.Equal(t, expected, output)
+}
+
+func TestTaskProtectionTypeFromStringScaleIn(t *testing.T) {
```
chienhanlin, line 30: We could use table driven test for TestTaskProtectionTypeFromString: ScaleIn, ScaleInLower, Disabled, DisabledLower and TestTaskProtectionTypeFromStringError.

```
	testcases := []struct {
		name string
		input  string
                err     error
		result  taskProtectionType
	}{...}
```
Like this example: https://github.com/aws/amazon-ecs-agent/blob/71a0f6765cc469913a0b09728196feb0d0d33e37/agent/utils/utils_test.go#L136-L164

amogh09: Thank you for the suggestion. I like this pattern. I have made the change.

agent/handlers/agentapi/v1/taskprotection/types/types.go
amogh09: The fields of `TaskProtection` struct will be used to call `PutTaskProtection` ECS API. The design of this API is still not finalized so I don't want to validate against an upper bound yet. Also I think the upper bound value is arbitrary (it's optional and if not present then defaults to infinity), so I am concerned that every time ECS changes the upper bound we will need to change it also. I will ask ECS folks about this.

chienhanlin: Thanks for the info!

chienhanlin (approved):

agent/handlers/agentapi/v1/taskprotection/types/types.go
```diff
@@ -0,0 +1,67 @@
...
+		return TaskProtectionTypeDisabled, nil
+	default:
+		return "", fmt.Errorf("unknown task protection type: %v", s)
+	}
+}
+
+// Protection for a Task
+type taskProtection struct {
+	protectionType           taskProtectionType
+	protectionTimeoutMinutes *int
```
Realmonia, line 44: Q: What's the consideration of making this `*int` instead of `int`? Is this in sync with scheduler API?

Realmonia (commented):
LGTM. Please confirm with the control plane guys that this is aligned with finalized API.

Realmonia (approved):

PR #3353, opened 2022-08-22
Task scale in handlers

amogh09:
### Summary
This change adds a new HTTP endpoint to agent with the following two handlers. 

| Endpoint path | Method | Description |
| ------------- | -------- | ----------- |
| /api/v1/{v3EndpointID}/task/protection | PUT | Update protection status of a task |
| /api/v1/{v3EndpointID}/task/protection | GET | Get the protection status of a task |

The endpoint path `/api/v1/{v3EndpointID}` is exposed to customer containers as an environment variable. Changes for this have already been merged to the feature branch.

ECS Client has not yet been updated with `PutTaskProtection` and `GetTaskProtection` APIs, so the handlers do not do anything meaningful. Once the ECS client libraries are updated, the handlers will be updated to call the corresponding ECS APIs. This will be done in a future PR. TODOs have been added to the handlers to capture this requirement. For this reason the new handlers simply respond with "Ok" on success for now, the response will be made richer when the handlers are updated to call ECS.

### Implementation details
Handler functions are added for PUT and GET methods of the endpoint. The handlers lookup task for the request using the V3 Endpoint ID passed in the request URL and perform the required validations. 

### Testing
1. New unit tests have been added to test the new handlers and their registration to the agent HTTP server. 
2. Agent was built using the source of this PR on an AL2 EC2 instance and PUT and GET requests were sent to it on the new endpoint being added. The requests were handled by the agent as expected.

```
root@5006c649e9d2:/# curl -X GET $ECS_AGENT_API_URI_V1/task/protection                                      
"Ok"root@5006c649e9d2:/#                                                                                    
root@5006c649e9d2:/# curl -X PUT $ECS_AGENT_API_URI_V1/task/protection -d '{"protectionType": "scale_in"}'  
"Ok"root@5006c649e9d2:/#                                                                                    
root@5006c649e9d2:/#                                                                                        
root@5006c649e9d2:/# curl -X PUT $ECS_AGENT_API_URI_V1/task/protection -d '{"protectionType": "invalid"}'   
"Invalid request: protection type is invalid: unknown task protection type: invalid"root@5006c649e9d2:/#    
```

Add PutTaskProtection and GetTaskProtection APIs.

### Licensing

### DIFF
--- a/agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go
+++ b/agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go
@@ -1,0 +1,105 @@
+package handlers
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
+	"github.com/aws/amazon-ecs-agent/agent/handlers/agentapi/v1/taskprotection/types"
+	"github.com/aws/amazon-ecs-agent/agent/handlers/utils"
+	v3 "github.com/aws/amazon-ecs-agent/agent/handlers/v3"
+	"github.com/cihub/seelog"
+)
+
+const (
+	putTaskProtectionRequestType = "v1/PutTaskProtection"
+)
+
+// Returns endpoint path for PutTaskProtection API
+func PutTaskProtectionPath() string {
+	return fmt.Sprintf(
+		"/api/v1/%s/task/protection",
+		utils.ConstructMuxVar(v3.V3EndpointIDMuxName, utils.AnythingButSlashRegEx))
+}
+
+// Task protection request received from customers pending validation
+type taskProtectionRequest struct {
+	ProtectionType           string
+	ProtectionTimeoutMinutes *int
+}
+
+// PutTaskProtectionHandler returns an HTTP request handler function for
+// PutTaskProtection API
+func PutTaskProtectionHandler(state dockerstate.TaskEngineState,
+	cluster string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var request taskProtectionRequest
+		jsonDecoder := json.NewDecoder(r.Body)
+		jsonDecoder.DisallowUnknownFields()
+		if err := jsonDecoder.Decode(&request); err != nil {
+			seelog.Errorf("PutTaskProtection: failed to decode request: %v", err)
+			writeJSONResponse(w, http.StatusBadRequest,
+				"Failed to decode request", putTaskProtectionRequestType)
+			return
+		}
+
+		if request.ProtectionType == "" {
+			writeJSONResponse(w, http.StatusBadRequest,
+				"Invalid request: protection type is missing or empty",
+				putTaskProtectionRequestType)
+			return
+		}
+
+		taskProtection, err := types.NewTaskProtection(request.ProtectionType, request.ProtectionTimeoutMinutes)
+		if err != nil {
+			writeJSONResponse(w, http.StatusBadRequest,
+				fmt.Sprintf("Invalid request: %v", err),
+				putTaskProtectionRequestType)
+			return
+		}
+
+		task, err := getTaskFromRequest(state, r)
+		if err != nil {
+			writeJSONResponse(w, http.StatusInternalServerError,
+				fmt.Sprintf("Failed to find task: %v", err), putTaskProtectionRequestType)
+			return
+		}
+
+		// TODO: Call ECS
+		seelog.Infof("Would have called ECS.PutTaskProtection(%s, %s, %s, %s, %v)\n",
+			cluster, task.ServiceName, task.Arn, taskProtection.GetProtectionType(),
+			taskProtection.GetProtectionTimeoutMinutes())
+		writeJSONResponse(w, http.StatusOK, "Ok", putTaskProtectionRequestType)
+	}
+}
+
+// Helper function for finding task for the request
+func getTaskFromRequest(state dockerstate.TaskEngineState, r *http.Request) (*apitask.Task, error) {
+	taskARN, err := v3.GetTaskARNByRequest(r, state)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get task ARN from request: %w", err)
+	}
+
+	task, found := state.TaskByArn(taskARN)
+	if !found {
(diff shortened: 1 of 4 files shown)


Realmonia (commented):
From credential's perspective, according to Scheduler team's credential doc, the plan is to add the ecs:getTaskProtection access to task execution role and use task execution credential to authorize the call.

Instead of using ecs client passed from acs handler, we might need to create a different ecs client using taskExecution credentials (or if there are some other way to make API calls using task execution credential). The ecs client passed from acs handler is using default credential provider (which I think for ec2 case it is instanceRole and for ecs anywhere it is the one associated with SSM activation).

Realmonia:
Actually the current state of PR don't include that, but maybe just keep that noted

amogh09:
That's a good point. Yes, the current PR doesn't include that but we will need it in the near future. I will see how ECS Client is initialized currently and how we can initialize another one with task execution role creds.

YashdalfTheGray (approved):

YashdalfTheGray:
Something else to keep in mind related to this task execution role thread is that task execution roles don't come pre-created in new accounts so for new customers of ECS, using task scale in protection will require them to create a task execution role. 

One can be created just by launching a fargate task from the console though.

amogh09:
I see. We will remember to mention this in the documentation of this feature.

agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Returns endpoint path for PutTaskProtection API
+func TaskProtectionPath() string {
+	return fmt.Sprintf(
+		"/api/v1/%s/task/protection",
+		utils.ConstructMuxVar(v3.V3EndpointIDMuxName, utils.AnythingButSlashRegEx))
+}
+
+// Task protection request received from customers pending validation
+type taskProtectionRequest struct {
```
chienhanlin, line 37: Q. should we make it uppercase if we want to make it TaskProtectionRequest.ProtectionType or TaskProtectionRequest.ProtectionTimeoutMinutes public accessible?

agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go
```diff
@@ -0,0 +1,136 @@
...
+	ProtectionType           string
+	ProtectionTimeoutMinutes *int
+}
+
+// PutTaskProtectionHandler returns an HTTP request handler function for
+// PutTaskProtection API
+func PutTaskProtectionHandler(state dockerstate.TaskEngineState,
+	cluster string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		putTaskProtectionRequestType := "v1/PutTaskProtection"
```
chienhanlin, line 47: nit. maybe we can move constants to a constant block 
``` 
const (
	putTaskProtectionRequestType = "v1/PutTaskProtection"
        getTaskProtectionRequestType := "api/v1/GetTaskProtection"
)

Comment (chienhanlin) [reply agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go:47]: and do we need  to append `api/` for putTaskProtectionRequestType?

File: agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go (hunk 1 above)
Comment (amogh09) [reply agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go:37]: I think the fields should be lowercase. I had originally defined this type in `types` package with fields exported and I think I forgot to hide the fields when I moved this type here. Good catch!

File: agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go (hunk 2 above)
Comment (amogh09) [reply agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go]: Yeah I will make the two constants consistent.

Comment (amogh09) [reply agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go]: Hmm json decoding doesn't work if fields are not exported...

Comment (amogh09) [reply agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go]: I think I will export the type and use it to define requests in `task_server_setup_test.go`.

File: agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go [hunk 3]
```diff
@@ -0,0 +1,136 @@
...
+func PutTaskProtectionHandler(state dockerstate.TaskEngineState,
+	cluster string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		putTaskProtectionRequestType := "api/v1/PutTaskProtection"
+
+		var request TaskProtectionRequest
+		jsonDecoder := json.NewDecoder(r.Body)
+		jsonDecoder.DisallowUnknownFields()
+		if err := jsonDecoder.Decode(&request); err != nil {
+			seelog.Errorf("PutTaskProtection: failed to decode request: %v", err)
```
chienhanlin, line 53: I miss this in the previous review. Let use [logger](https://github.com/aws/amazon-ecs-agent/tree/master/agent/logger) here as we are gradually migrating to use it for better readability : )

```
"github.com/aws/amazon-ecs-agent/agent/logger/field"
```
```
logger.Info("PutTaskProtection: failed to decode request:", logger.Fields{
			field.Error:    error,
            })
```

amogh09: Sure. I will replace `seelog` with `logger`. Thanks.

agent/handlers/task_server_setup_test.go
```diff
@@ -39,6 +40,8 @@ import (
 	mock_credentials "github.com/aws/amazon-ecs-agent/agent/credentials/mocks"
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	mock_dockerstate "github.com/aws/amazon-ecs-agent/agent/engine/dockerstate/mocks"
+	agentAPIV1 "github.com/aws/amazon-ecs-agent/agent/handlers/agentapi/v1/taskprotection/handlers"
```
chienhanlin, line 43: nit. Could we also name it as `agentAPIV1TaskProtection` or `agentapi_taskprotection`as we might add new services under agentapi/v1 in the future.

amogh09: Sure. I will make it snakecase to keep consistency with other import aliases too. Thanks.

agent/handlers/agentapi/v1/taskprotection/handlers/handlers.go
chienhanlin: Thank you!

chienhanlin (approved):

PR #3363, opened 2022-08-26
Update ECS_RESERVED_MEMORY description in README

Realmonia:
### Summary

Update ECS_RESERVED_MEMORY description in README that match the actual functionality better

### Implementation details

See actual usage of this parameter here:
- https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config.go#L542
- https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/ecsclient/client.go#L273
- https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/ecsclient/client.go#L184-L192

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ additional details on each available environment variable.
 | `ECS_POLL_METRICS`     | &lt;true &#124; false&gt;  | Whether to poll or stream when gathering metrics for tasks. Setting this value to `true` can help reduce the CPU usage of dockerd and containerd on the ECS container instance. See also ECS_POLL_METRICS_WAIT_DURATION for setting the poll interval. | `false` | `false` |
 | `ECS_POLLING_METRICS_WAIT_DURATION` | 10s | Time to wait between polling for metrics for a task. Not used when ECS_POLL_METRICS is false. Maximum value is 20s and minimum value is 5s. If user sets above maximum it will be set to max, and if below minimum it will be set to min. | 10s | 10s |
 | `ECS_PULL_DEPENDENT_CONTAINERS_UPFRONT` | &lt;true &#124; false&gt; | Whether to pull images for containers with dependencies before the dependsOn condition has been satisfied. | false | false |
-| `ECS_RESERVED_MEMORY` | 32 | Memory, in MiB, to reserve for use by things other than containers managed by Amazon ECS. | 0 | 0 |
+| `ECS_RESERVED_MEMORY` | 32 | Reduction, in MiB, of the memory capacity of the instance that is reported to Amazon ECS. Used by Amazon ECS when placing tasks on container instances. This doesn't reserve memory usage on the instance. | 0 | 0 |
 | `ECS_AVAILABLE_LOGGING_DRIVERS` | `["awslogs","fluentd","gelf","json-file","journald","logentries","splunk","syslog"]` | Which logging drivers are available on the container instance. | `["json-file","none"]` | `["json-file","none"]` |
 | `ECS_DISABLE_PRIVILEGED` | `true` | Whether launching privileged containers is disabled on the container instance. | `false` | `false` |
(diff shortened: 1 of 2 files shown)


agent/config/types.go
```diff
@@ -101,8 +101,8 @@ type Config struct {
 	// on the instance
 	DisableDockerHealthCheck BooleanDefaultFalse
 
-	// ReservedMemory specifies the amount of memory (in MB) to reserve for things
-	// other than containers managed by ECS
+	// ReservedMemory specifies the amount of memory (in MB) reduction for things
+	// other than containers managed by ECS when registering instance
```
amogh09, line 105: Assuming that `memResource` is honored when managing task containers, the code below tells me that `ReservedMemory` is indeed the amount of memory we do not use for ECS containers. Why is the current description inaccurate?
```go
        cpu, mem := getCpuAndMemory()
	remainingMem := mem - int64(client.config.ReservedMemory)
	seelog.Infof("Remaining mem: %d", remainingMem)
	if remainingMem < 0 {
		return nil, fmt.Errorf(
			"api register-container-instance: reserved memory is higher than available memory on the host, total memory: %d, reserved: %d",
			mem, client.config.ReservedMemory)
	}

	cpuResource := ecs.Resource{
		Name:         utils.Strptr("CPU"),
		Type:         &integerStr,
		IntegerValue: &cpu,
	}
	memResource := ecs.Resource{
		Name:         utils.Strptr("MEMORY"),
		Type:         &integerStr,
		IntegerValue: &remainingMem,
	}
```
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/ecsclient/client.go#L272

README.md
```diff
@@ -142,7 +142,7 @@ additional details on each available environment variable.
 | `ECS_POLL_METRICS`     | &lt;true &#124; false&gt;  | Whether to poll or stream when gathering metrics for tasks. Setting this value to `true` can help reduce the CPU usage of dockerd and containerd on the ECS container instance. See also ECS_POLL_METRICS_WAIT_DURATION for setting the poll interval. | `false` | `false` |
 | `ECS_POLLING_METRICS_WAIT_DURATION` | 10s | Time to wait between polling for metrics for a task. Not used when ECS_POLL_METRICS is false. Maximum value is 20s and minimum value is 5s. If user sets above maximum it will be set to max, and if below minimum it will be set to min. | 10s | 10s |
 | `ECS_PULL_DEPENDENT_CONTAINERS_UPFRONT` | &lt;true &#124; false&gt; | Whether to pull images for containers with dependencies before the dependsOn condition has been satisfied. | false | false |
-| `ECS_RESERVED_MEMORY` | 32 | Memory, in MiB, to reserve for use by things other than containers managed by Amazon ECS. | 0 | 0 |
```
fierlion, line 145: pretty sure this is specifically memory reserved for ecs agent and docker.  Might want to add those as an example since they're really the main use case.

agent/config/types.go
yinyic, line 105: Can we make it same as in README? 
```
... reduction of the memory capacity of the instance that is reported to Amazon ECS.
```

Realmonia: Yes, this is not used by containers, because when we register instance with Control Plane, the amount of memory is the result of this deduction. The problem brought up in that ticket is - this gives customer a false expectation that any excessive usage of resources beyond this limited is not allowed, in other word, this is treated as a hard limit.

Realmonia: 👌

amogh09: I see what you mean now.

README.md
Realmonia: Not sure if only ecs agent and docker are the only intended ones, I just did a quick `top` on one instance, things like `containerd` `systemd` and `ssmAgent` are all using some amount of memory, is it possible that this change gives users wrong information about the resources usage?

yinyic (approved):

amogh09 (approved):

Issue #3241, opened 2022-06-04
Allow alternate credential profiles to be specified


dunka:
### Summary
The ecs-agent uses the "default" profile from root and does not provide an option to specify an alternative. 

### Description
In some cases credentials are present under an alternate profile name in the root credentials file (/root/.aws/credentials).  It would be great if we could pass a config flag to use an alternate profile name.

This seems like it should be a straight forward feature, changing the hard coded https://github.com/aws/amazon-ecs-agent/blob/master/agent/credentials/providers/rotating_shared_credentials_provider.go#L48 to read from a config option and of course updating the bits around reading and passing around the option from the ecs.config.

It could default to "default" to ensure existing users don't have to change anything.

sparrc:
Hello, thanks for opening this issue, I can confirm that specifying a custom profile for the SSM agent (via https://github.com/aws/amazon-ssm-agent#config-property-definitions) will cause problems since we hardcoded this cred provider to use "default".

To solve this we would need to add a config var called something like `ECS_EXTERNAL_CREDENTIAL_PROFILE="default"`

sunds:
If the hardcoded default profile is removed, the SharedCredentialsProvider will honor the standard AWS_PROFILE environment variable, or it will choose "default" if not set. AWS_PROFILE can be set in /etc/ecs/ecs.config

https://github.com/aws/amazon-ecs-agent/pull/3326

fierlion:
https://github.com/aws/amazon-ecs-agent/pull/3360 <- I did a quick rebase or your 3326 pr against the head of dev and am running functional tests.  Even with the assurance that the SDK will assume 'default' we'll still want to build in a config var to make sure this is always set.

fierlion:
```
--- FAIL: TestGMSAWithS3CredentialSpec (35.31s)
--- FAIL: TestV3TaskEndpointDefaultMode (356.62s)
--- FAIL: TestEnvFilePrecedence2EnvFiles (326.54s)
--- FAIL: TestEnvFilePrecedenceTaskDefEnvironment (56.01s)
```
this failed a subset of windows functional tests.

sunds:
Thanks for this. Your approach is the better one.

fierlion:
Note the above change was merged and will go out with the next release.  Will close this ticket once the release is completed.

ubhattacharjya:
Hi,

Since this has been released, I am closing the issue. Kindly reopen if the issues persists.

Thanks

PR #3365, opened 2022-08-26
Add configurable default profile

fierlion:
### Summary
Update to allow for a configurable alternate credential profile other than `default`.
https://github.com/aws/amazon-ecs-agent/issues/3241

### Implementation details
Add a new env var, `ECS_ALTERNATE_CREDENTIAL_PROFILE`, to allow for overriding the default credential profile.

### Testing
built agent with existing logging in place.  updated /etc/ecs/ecs.config to add 
```
ECS_ALTERNATE_CREDENTIAL_PROFILE="testProfile"
```
checked info level logs for the override message.

### Licensing


agent/credentials/providers/rotating_shared_credentials_provider.go
```diff
@@ -15,13 +15,16 @@ package providers
...
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/cihub/seelog"
 )
 
 const (
+	EXTERNAL_CREDENTIAL_PROFILE_ENV_VAR = "ECS_EXTERNAL_CREDENTIAL_PROFILE"
```
singholt, line 26: I think this may also be a valid option for EC2 customers. if so, the env var could have a generic name like `ECS_CREDENTIAL_PROFILE`. wdyt?

fierlion: how do you like `ECS_ALTERNATE_CREDENTIAL_PROFILE` ?
I want to highlight that it's an override of the default profile.

singholt: sounds good

Realmonia (commented):
nit: sorry for my poor English but can you help me understand why you choose `alternate` over `alternative`? Thanks!

Edit: nvm I kinda understand it - the idea you want to convey here is it is not another optional but an override. Not sure if there are better word options to avoid the confusion that this is going to be used as a backup profile - considering it is not a back up but an override.

fierlion:
> nit: sorry for my poor English but can you help me understand why you choose alternate over alternative? Thanks!

They're essentially synonyms.  

'alternate is frequently used to mean "offering or expressing a choice." There are established uses for both that we regularly encounter: a director's cut of a film may feature an alternate ending; a traffic update warns motorists to seek an alternate route.
Many examples, though, use the word simply to mean "offering a choice apart from what is traditional or expected'

https://www.merriam-webster.com/words-at-play/differences-between-alternate-and-alternative

singholt (approved):

Realmonia (approved):

PR #3371, opened 2022-08-30
Fix Windows issue resolving LinkName

jterry75:
### Summary

1. Fixes an issue in Windows when iphelper api is failing
that can cause resolution of the interface LinkName to return
an empty string. In this case the CNI will fail to move the
task interface because it is not found.

2. Fixes an issue when using Get-NetAdapter to only return -Physical
addresses that avoids the duplicate vEthernet virtual address created
by CNI which can cause the wrong LinkName to be returned if the
vEthernet address is returned before the Physical address in the list.

3. Fixes an issue when using Get-NetAdapter that if a single element
is returned PowerShell ConvertTo-Json does not return the JSON as
an array which can cause a failure to unmarshal.

4. Fixes an issue when comparing MAC addresses that we compare them
as strict hardware bytes instead of strings to avoid match issues
between '-' and ':' between sets.

Signed-off-by: Justin Terry <jlterry@amazon.com>

### Testing

I have verified the code changes by running ECS Windows Tasks on ECS Clusters with iphelper completely disabled. This forces the use of the net interface wrapper (Get-NetAdapter callout) in all cases and verifies the functionality.

New tests cover the changes: 
- Existing tests cover the cases already.

Fixes bugs associated with Windows task networking resolution failing due to an invalid LinkName of `""` (empty string).

### Licensing

### DIFF
--- a/agent/api/eni/eni.go
+++ b/agent/api/eni/eni.go
@@ -14,6 +14,7 @@
 package eni
 
 import (
+	"bytes"
 	"fmt"
 	"net"
 	"strings"
@@ -22,6 +23,7 @@ import (
 	"github.com/cihub/seelog"
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
+	"github.com/aws/amazon-ecs-agent/agent/eni/netwrapper"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/pkg/errors"
 )
@@ -89,8 +91,8 @@ const (
 )
 
 var (
-	// netInterfaces is the Interfaces() method of net package.
-	netInterfaces = net.Interfaces
+	interfaceWrapper = netwrapper.New()
+	netInterfaces    = interfaceWrapper.GetAllNetworkInterfaces
 )
 
 // GetIPV4Addresses returns the list of IPv4 addresses assigned to the ENI.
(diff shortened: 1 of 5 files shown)


agent/api/eni/eni.go
```diff
@@ -90,7 +92,8 @@ const (
 
 var (
 	// netInterfaces is the Interfaces() method of net package.
```
vsiddharth, line 94: nit: Do you want to update/remove this comment?

agent/api/eni/eni.go
```diff
@@ -208,8 +211,13 @@ func (eni *ENI) GetLinkName() string {
...
 		}
 		// Iterate over the list and find the interface with the ENI's MAC address.
+		eniMac, err := net.ParseMAC(eni.MacAddress)
+		if err != nil {
+			seelog.Errorf("Failed to parse ENI MAC: %s with error: %v", eni.MacAddress, err)
+			return ""
+		}
 		for _, iface := range ifaces {
-			if strings.EqualFold(eni.MacAddress, iface.HardwareAddr.String()) {
+			if bytes.EqualFold(eniMac, iface.HardwareAddr) {
```
vsiddharth, line 220: Would it be helpful to check if LinkName is non-empty prior to return?

vsiddharth (commented):
Thanks a lot for making these changes Justin.

agent/api/eni/eni.go
jterry75: Good call resolved.

agent/api/eni/eni.go
jterry75: Good call removed.

rawahars (dismissed):
LGTM!

Thanks for the quick changes Justin!!

vsiddharth (approved):

vsiddharth (approved):

jterry75:
This ultimately does not work. While it fixes the agent CNI fails to use the interface because it is in an invalid state. Closing

PR #3387, opened 2022-09-09
Revert iphelper workaround

jterry75:
### Summary

Reverts: #3328 

Adds an extra logging line to better signify when the LinkName fails to resolve because the ENI is not match by MAC address.

### Licensing


rawahars (approved):

agent/api/eni/eni.go
```diff
@@ -214,6 +214,9 @@ func (eni *ENI) GetLinkName() string {
 				break
 			}
 		}
+		if eni.LinkName == "" {
```
singholt, line 217: is this the same check as on L203 and the same error log line as L207?

jterry75: Not exactly. What we found is that we can in the above for loop, loop through all interfaces, not match any, and then return "" with no logging. So we verify that indeed one was matched or else we at least log that it wasnt so we understand later why the CNI fails to find an empty adapter LinkName

singholt: nvm, I think this is to check if link name is empty even after the loop on L210.

nit: could you add a comment on this?

singholt (dismissed):

vsiddharth (approved):

agent/api/eni/eni.go
jterry75: Sure thing. Done.

singholt (approved):

Realmonia (approved):

PR #3403, opened 2022-09-24
Write AppNet logs to /var/log/ecs/ if no logging driver configured

yinyic:
### Summary

Write AppNet logs to host `/var/log/ecs/service_connect/` if no logging driver configured, for visibility and troubleshooting purposes.

That will always be the case for AppNet Relay container as they are not configurable by customer.

For AppNet Agent container, customer may set up logging driver within `ServiceConnectConfiguration` when creating or updating a service. If none specified, we then let AppNet write the logs to file. 

### Implementation details

We first check if a logging driver is present for AppNet relay/agent.

 If logging driver is unspecified, we will pass `APPNET_ENVOY_LOG_DESTINATION` ENV and use `/var/log/` (the container path) for both Relay and Agent containers.

We then create host directories `/var/log/ecs/service_connect/relay/` and `/var/log/ecs/service_connect/$task-id` for Relay and Agent respectively, as well as bind mounts to their container `/var/log/` path. 

We do NOT specify following log configurations, but rather use AppNet default values:
- log file name (`appnet_envoy.log`)
- max log file size (1Mb)
- retention count (5) 

### Testing

* added unit test
* ran e2e test and verified 
  *  when no logging driver specified, local log files are created at the correct locations for both relay and agent
  * when logging driver is configured for appnet agent, it's no longer writing to file on disk, and relay still is. Also checked that appnet agent log is streamed to cloudwatch (I tested with awslogs driver)

Write AppNet logs to /var/log/ecs/ if no logging driver configured

### Licensing

### DIFF
--- a/agent/engine/serviceconnect/manager_linux.go
+++ b/agent/engine/serviceconnect/manager_linux.go
@@ -53,6 +53,12 @@ const (
 	defaultStatusFileName     = "appnet_admin.sock"
 	defaultStatusENV          = "APPNET_AGENT_ADMIN_UDS_PATH"
 
+	// logging
+	defaultLogPathHostRoot              = "/var/log/ecs/service_connect/"
+	defaultLogPathContainer             = "/var/log/"
+	defaultLogPathECSAgentRoot          = "/log/service_connect/" // "/log" for ECS Agent is "/var/log/ecs" on host
+	defaultAppnetEnvoyLogDestinationENV = "APPNET_ENVOY_LOG_DESTINATION"
+
 	relayEnableENV = "APPNET_ENABLE_RELAY_MODE_FOR_XDS"
 	relayEnableOn  = "1"
 	upstreamENV    = "APPNET_RELAY_LISTENER_UDS_PATH"
@@ -93,6 +99,12 @@ type manager struct {
 	statusFileName string
 	// Environment variable to set on Container with contents of statusPathContainer/statusFileName
 	statusENV string
+	// Path to where AppNet log file will be written to inside container
+	logPathContainer string
+	// Path to where AppNet log file will be written to on host
+	logPathHostRoot string
+	// Path to create logging dir for AppNet, from ECS Agent point of view (b/c "/log" for ECS Agent is "/var/log/ecs" on host)
+	logPathECSAgentRoot string
 
 	// Http path + params to make a statistics request of AppNetAgent
 	adminStatsRequest string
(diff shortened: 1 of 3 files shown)


agent/engine/serviceconnect/manager_linux.go
```diff
@@ -53,6 +53,12 @@ const (
 	defaultStatusFileName     = "appnet_admin.sock"
 	defaultStatusENV          = "APPNET_AGENT_ADMIN_UDS_PATH"
 
+	// logging
+	defaultLogPathHostRoot              = "/var/log/ecs/service_connect/"
```
sparrc, line 57: Could `/var/log` be sourced from a config variable somewhere? I believe this path is changeable, and is changed for example when ecs-init is built with `make dev` with the "directory prefix" var in ecs-init: 

prefix set to /tmp here: https://github.com/aws/amazon-ecs-agent/blob/71a0f6765cc469913a0b09728196feb0d0d33e37/ecs-init/config/development.go#L36


sourced here: https://github.com/aws/amazon-ecs-agent/blob/637446af461d77e75c4556cd55e6771bd4250fe4/ecs-init/config/common.go#L155

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -53,6 +53,12 @@ const (
...
+	defaultLogPathContainer             = "/var/log/"
+	defaultLogPathECSAgentRoot          = "/log/service_connect/" // "/log" for ECS Agent is "/var/log/ecs" on host
```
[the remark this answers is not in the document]
yinyic: In new rev we will attempt to parse the container logging root from `ECS_LOGFILE` env first, and fallback to `/log` if it's not set.

agent/engine/serviceconnect/manager_linux.go
yinyic: Discussed offline - it's quite challeging to get the host path of log directory from inside ECS Agent. We would need to get the host path from Agent container bind mount info, for which we need a docker inspect and agent container ID, which in turn probably needs a docker ls with image name filter. 

The risk of possible impact of log host dir being something other than `/var/log/ecs/` should be minimal - most customers run ECS Agent with init which defaults to `/var/log/ecs`. Even if the log path on instance is explicitly updated by specifying a different volume mount, and as a result we fail to configure the correct log dir for AppNet, it will be able to fall back to the stdout/stderr and logs will be available through docker log json file. Other than that, AppNet should continue to function properly.

sparrc (approved):

ubhattacharjya (approved):

PR #3408, opened 2022-09-28
Update debian recipe to use system golang version

fierlion:
### Summary
This updates the static build process to use the golang version specified in GO_VERSION.  It also updates the Debian repo build process to use the Makefile targets and skip the extra linting and internal change tracking.

### Implementation details
For the Debian builds, the main changes are to use `dpkg-buildpackage` rather than `debuild` (the latter calls the former but adds a bunch of extra steps that we don't need).  It also removes the dependency on Debian-managed golang and updates the build targets to prefer either the pre-existing system golang installed at `/usr/local/go` or else install the GO_VERSION via wget and use that.  This allows our builds to work with the latest golang version (currently 1.19.1).

This also updates our builds to use a symlinked tmpdir as our buildroot and sets our GOPATH to temp buildroot to override any issues with build systems that build in non-standard paths (like our CodeBuild environment).

### Testing
Built locally using all Make targets on Ubuntu 20.04.4 and also built via CodeBuild (see github action tests below).

Update static build golang version and refine Debian builds. 

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,9 @@ all: docker
 gobuild:
 	./scripts/build false
 
+gobuild-init-deb:
+	./scripts/gobuild.sh debian
+
 # create output directories
 .out-stamp:
 	mkdir -p ./out/test-artifacts ./out/cni-plugins ./out/amazon-ecs-cni-plugins ./out/amazon-vpc-cni-plugins
@@ -338,6 +341,10 @@ goimports:
 	goimports -w $(GOFMTFILES)
 
 GOPATH=$(shell go env GOPATH)
+
+install-golang:
+	./scripts/install-golang.sh
+
 .get-deps-stamp:
 	go get golang.org/x/tools/cmd/cover
 	go get github.com/golang/mock/mockgen
@@ -396,8 +403,8 @@ VERSION = $(shell cat ecs-init/ECSVERSION)
 	mkdir -p BUILDROOT
 	./scripts/update-version.sh
 	tar -czf ./amazon-ecs-init_${VERSION}.orig.tar.gz ecs-init scripts README.md
-	cp -r packaging/generic-deb-integrated/debian ecs-init scripts misc agent agent-container amazon-ecs-cni-plugins amazon-vpc-cni-plugins README.md VERSION BUILDROOT
-	cd BUILDROOT && debuild -uc -us --lintian-opts --suppress-tags bad-distribution-in-changes-file,file-in-unusual-dir
+	cp -r packaging/generic-deb-integrated/debian Makefile ecs-init scripts misc agent agent-container amazon-ecs-cni-plugins amazon-vpc-cni-plugins README.md VERSION GO_VERSION BUILDROOT
+	cd BUILDROOT && dpkg-buildpackage -uc -b
 	touch .generic-deb-integrated-done
 
 generic-deb-integrated: .generic-deb-integrated-done
@@ -481,7 +488,7 @@ clean:
 	-rm -rf ./BUILDROOT BUILD RPMS SRPMS SOURCES SPECS
 	-rm -rf ./x86_64
 	-rm -f ./amazon-ecs-init_${VERSION}*
-	-rm -f .srpm-done .rpm-done .generic-rpm-done
+	-rm -f .srpm-done .rpm-done .generic-rpm-done .generic-deb-integrated-done
 	-rm -f .deb-done
 	-rm -f .amazon-linux-rpm-integrated-done
 	-rm -f .generic-rpm-integrated-done
--- a/buildspecs/pr-build-ubuntu.yml
+++ b/buildspecs/pr-build-ubuntu.yml
@@ -31,7 +31,6 @@ phases:
 
       # Path readjustment for codebuild testing with fork and setting GOPATH appropriately
       - cd ../../../..
-      - export GOPATH=$GOPATH:$(pwd)
       - cd src/github.com
       - |
         if [[ $GITHUBUSERNAME != "aws" ]] ; then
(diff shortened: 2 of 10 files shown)


buildspecs/pr-build-ubuntu.yml
```diff
@@ -44,9 +44,7 @@ phases:
       - DEBIAN_FRONTEND=noninteractive
         
       - apt-get update -y | tee -a $BUILD_LOG
-      - apt-get install -y make dpkg-dev devscripts debhelper golang | tee -a $BUILD_LOG
-      - which go
-      - go version
+      - apt-get install -y make dpkg-dev wget devscripts debhelper | tee -a $BUILD_LOG
```
prateekchaudhry, line 47: The ```GOPATH``` in this file is probably not being used now?

We should also update https://github.com/aws/amazon-ecs-agent/blob/master/buildspecs/merge-build-ubuntu.yml with these updates to make builds available in merges to main branch

fierlion: I'll create a follow up PR to fix these up.  TODO Will link here once it's ready.

prateekchaudhry (dismissed):

prateekchaudhry (approved):

Makefile
```diff
@@ -396,8 +403,8 @@ VERSION = $(shell cat ecs-init/ECSVERSION)
 	mkdir -p BUILDROOT
 	./scripts/update-version.sh
 	tar -czf ./amazon-ecs-init_${VERSION}.orig.tar.gz ecs-init scripts README.md
-	cp -r packaging/generic-deb-integrated/debian ecs-init scripts misc agent agent-container amazon-ecs-cni-plugins amazon-vpc-cni-plugins README.md VERSION BUILDROOT
-	cd BUILDROOT && debuild -uc -us --lintian-opts --suppress-tags bad-distribution-in-changes-file,file-in-unusual-dir
+	cp -r packaging/generic-deb-integrated/debian Makefile ecs-init scripts misc agent agent-container amazon-ecs-cni-plugins amazon-vpc-cni-plugins README.md VERSION GO_VERSION BUILDROOT
+	cd BUILDROOT && dpkg-buildpackage -uc -us -b
```
YashdalfTheGray, line 407: Since we're using `dpkg-buildpackage` and not `debuild`, does the built in `debsign` stuff still happen?

YashdalfTheGray (dismissed):

scripts/install-golang.sh
```diff
@@ -0,0 +1,57 @@
...
+esac
+
+if [ "$architecture" == "amd64" ]; then export GOARCH=amd64; fi
+if [ "$architecture" == "arm64" ]; then export GOARCH=arm64; fi
+
+export GOROOT=/usr/local/go
+export PATH=$PATH:$GOROOT/bin
+
+# if golang isn't installed then we'll install it
+if [ -d "$GOROOT" ]; then
```
prateekchaudhry, line 40: $GOROOT might be pre-initialized in codebuild buildspecs (for rpm builds). Maybe we can check if go exists first and then export this

scripts/build
```diff
@@ -59,21 +75,25 @@ fi
...
 
-cd "${ROOT}"
+# run build from our temporary linked srcpath
+cd ${SRCPATH}/amazon-ecs-agent
 if [[ "${TARGET_OS}" == "windows" ]]; then
-    go build -ldflags "${LDFLAGS} -s" -o $build_exe ./agent/
+    go build -ldflags "${LDFLAGS} -s" -o $build_artifact ./agent/
 elif [[ "${static}" == "true" ]]; then
-    GO111MODULE=auto CGO_ENABLED=0 go build -installsuffix cgo -a -ldflags "${LDFLAGS} -s" -o $build_exe ./agent/
+    GO111MODULE=off CGO_ENABLED=0 go build -installsuffix cgo -a -ldflags "${LDFLAGS} -s" -o $build_artifact ./agent/
```
prateekchaudhry, line 89: ```GO111MODULE=off``` Is this change intentional?

Makefile
fierlion: removed the -us (unsigned-source) to force dpkg-buildpackage to do the signing stuff
https://man7.org/linux/man-pages/man1/dpkg-buildpackage.1.html

scripts/install-golang.sh
```diff
@@ -0,0 +1,63 @@
...
+    export GOROOT=/usr/local/go
+    export PATH=$PATH:$GOROOT/bin
+    echo "$GOROOT exists. Using installed golang version"
+else
+    # install golang defined in GO_VERSION file
+    echo "$GOROOT doesn't exist, installing $(cat ./GO_VERSION)"
+    GO_VERSION=$(cat ./GO_VERSION)
+    tmpdir=$(mktemp -d)
+    GOLANG_TAR="go${GO_VERSION}.linux-${GOARCH}.tar.gz"
+    wget -O ${tmpdir}/${GOLANG_TAR} https://storage.googleapis.com/golang/${GOLANG_TAR}
```
sparrc, line 51: Could we use `curl` instead of `wget` for this? IME curl is almost always included by default on OSes and wget requires installation.

fierlion:
```
++ curl https://storage.googleapis.com/golang/go1.19.1.linux-amd64.tar.gz -o /tmp/tmp.WfvR90vbnW/go1.19.1.linux-amd64.tar.gz
--
930 | ./scripts/install-golang.sh: line 54: curl: command not found
```

Updated PR and tested without `wget` -- looks like we'll need to install `curl` for debian builds.  I'll prefer wget in this case as it's already working and tested for all build types.  (We can revisit this in future if we have trouble.)

YashdalfTheGray (dismissed):

prateekchaudhry (approved):

prateekchaudhry (approved):

YashdalfTheGray (approved):

PR #3413, opened 2022-10-03
Add a check for http response code

ubhattacharjya:
### Summary

Just added a small check where we check for not only errors while getting stats for service connect stats but also add a check to see if the response code is 200OK.

### Testing

Tested that stats are getting correctly transmitted after this change as well

### Licensing

### DIFF
--- a/agent/api/appnet/client_linux.go
+++ b/agent/api/appnet/client_linux.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/logger/field"
 	"github.com/aws/amazon-ecs-agent/agent/utils/retry"
+	"github.com/pkg/errors"
 	prometheus "github.com/prometheus/client_model/go"
 )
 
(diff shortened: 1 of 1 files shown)


yinyic (approved):

agent/api/appnet/client_linux.go
```diff
@@ -41,6 +42,9 @@ func (cl *client) GetStats(config serviceconnect.RuntimeConfig) (map[string]*pro
 	if err != nil {
 		return nil, err
 	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.Wrapf(err, "received non-OK HTTP status %v from Service Connect stats endpoint", resp.StatusCode)
```
amogh09, line 46: Is it safe to return before `defer resp.Body.Close()` call?

amogh09: Also should we do add the response body to the error? It might contain useful info?

chienhanlin (dismissed):

agent/api/appnet/client_linux.go
ubhattacharjya: good callout. Will change it

amogh09 (approved):

amogh09 (approved):

chienhanlin (approved):

PR #3415, opened 2022-10-04
Manage Appnet versioning

mythri-garaga:
### Summary

This PR includes changes to handle the major version updates to AppNet image version.

### Implementation details

- ` agent/engine/serviceconnect/version_linux.go`:
	New version_linux file contains supportedAppnetInterfaceVerToCapability map to track all the AppNet interface versions supported by ECS Agent. Added getSupportedAppnetInterfaceVersions() function that returns the supported AppNet interface versions by ECS Agent from highest to lowest version.
- `agent/engine/serviceconnect/manager_linux.go`:
         - Updated LoadImage() to iterate through supported Appnet interface versions list from highest to lowest order and check if the corresponding tar file exists before loading, if loading is successfully `AppnetInterfaceVersion` for the runtime.
        - Added a new function GetSupportedAppnetInterfaceVerToCapabilities() to return service connect capabilities supported by ECS Agent to register for a selected AppNet version.
- Added unit tests to cover the above functions and updated existing texts to handle mock calls to GetSupportedAppnetInterfaceVerToCapabilities(). Will add a test to check if right AppNet version is being loaded in the upcoming CR which includes changes to restart relay on AppNet image update.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -457,8 +457,18 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 			return capabilities
 		}
 	}
-
-	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityServiceConnect)
+	loadedVer, _ := agent.serviceconnectManager.GetLoadedAppnetVersion()
+	supportedAppnetInterfaceVerToCapabilities, _ := agent.serviceconnectManager.GetCapabilitiesForAppnetInterfaceVersion(loadedVer)
+	if supportedAppnetInterfaceVerToCapabilities == nil {
+		logger.Warn("ServiceConnect Capability: No service connect capabilities were found for Appnet version:", logger.Fields{
+			field.Image: loadedVer,
+		},
+		)
+	}
+	for _, serviceConnectCapability := range supportedAppnetInterfaceVerToCapabilities {
+		capabilities = appendNameOnlyAttribute(capabilities, serviceConnectCapability)
+	}
+	return capabilities
 }
 
 func defaultGetSubDirectories(path string) ([]string, error) {
--- a/agent/engine/serviceconnect/manager_linux.go
+++ b/agent/engine/serviceconnect/manager_linux.go
@@ -78,9 +78,9 @@ const (
 	defaultAdminStatsRequest = httpRequestPrefix + "/stats/prometheus?usedonly&filter=metrics_extension&delta"
 	defaultAdminDrainRequest = httpRequestPrefix + "/drain_listeners?inboundonly"
 
-	defaultAgentContainerTarballPath = "/managed-agents/serviceconnect/appnet_agent.interface-v1.tar"
-	defaultAgentContainerImageName   = "appnet_agent"
-	defaultAgentContainerTag         = "service_connect.v1"
+	defaultAgentContainerImageName         = "ecs-service-connect-agent"
+	defaultAgentContainerTagFormat         = "interface-%s"
+	defaultAgentContainerTarballPathFormat = "/managed-agents/serviceconnect/ecs-service-connect-agent.interface-%s.tar"
 
 	ecsAgentLogFileENV              = "ECS_LOGFILE"
 	defaultECSAgentLogPathContainer = "/log"
@@ -115,9 +115,9 @@ type manager struct {
 	// Http path + params to make a drain request of AppNetAgent
 	adminDrainRequest string
 
-	AgentContainerImageName   string
-	AgentContainerTag         string
-	AgentContainerTarballPath string
+	agentContainerImageName string
+	agentContainerTag       string
+	appnetInterfaceVersion  string
 
 	ecsClient            api.ECSClient
 	containerInstanceARN string
@@ -139,9 +139,7 @@ func NewManager() Manager {
 		logPathHostRoot:     defaultLogPathHostRoot,
 		logPathECSAgentRoot: fmt.Sprintf(defaultECSAgentLogPathForSC, getECSAgentLogPathContainer()),
 
-		AgentContainerImageName:   defaultAgentContainerImageName,
-		AgentContainerTag:         defaultAgentContainerTag,
-		AgentContainerTarballPath: defaultAgentContainerTarballPath,
+		agentContainerImageName: defaultAgentContainerImageName,
 	}
 }
 
@@ -372,17 +370,40 @@ func (m *manager) AugmentInstanceContainer(task *apitask.Task, container *apicon
 	return nil
 }
 
-// LoadImage helps load the AppNetAgent container image for the agent
-func (agent *manager) LoadImage(ctx context.Context, _ *config.Config, dockerClient dockerapi.DockerClient) (*types.ImageInspect, error) {
-	logger.Debug("Loading appnet agent container tarball:", logger.Fields{
-		field.Image: agent.AgentContainerTarballPath,
-	})
-	if err := loader.LoadFromFile(ctx, agent.AgentContainerTarballPath, dockerClient); err != nil {
-		return nil, err
-	}
+func (agent *manager) setLoadedAppnetVerion(appnetInterfaceVersion string) {
+	agent.appnetInterfaceVersion = appnetInterfaceVersion
+}
 
-	imageName, _ := agent.GetLoadedImageName()
-	return loader.GetContainerImage(imageName, dockerClient)
+// LoadImage helps load the AppNetAgent container image for the agent latest supported
+// AppNet interface version by looking for the AppNet agent tar name from supported list
+// of AppNet versions from highest to lowest version when loading AppNet image
+func (agent *manager) LoadImage(ctx context.Context, _ *config.Config, dockerClient dockerapi.DockerClient) (*types.ImageInspect, error) {
+	var loadErr error
+	for _, supportedAppnetInterfaceVersion := range getSupportedAppnetInterfaceVersions() {
+		agentContainerTarballPath := fmt.Sprintf(defaultAgentContainerTarballPathFormat, supportedAppnetInterfaceVersion)
+		if _, err := os.Stat(agentContainerTarballPath); err != nil {
+			logger.Warn(fmt.Sprintf("AppNet agent container tarball unavailable: %s", agentContainerTarballPath), logger.Fields{
+				field.Error: err,
+			})
+			continue
+		}
+		logger.Debug(fmt.Sprintf("Loading appnet agent container tarball: %s", agentContainerTarballPath))
+		if loadErr = loader.LoadFromFile(ctx, agentContainerTarballPath, dockerClient); loadErr != nil {
+			logger.Warn(fmt.Sprintf("Unable to load appnet agent container tarball: %s", agentContainerTarballPath),
+				logger.Fields{
+					field.Error: loadErr,
+				})
+			continue
+		}
+		agent.setLoadedAppnetVerion(supportedAppnetInterfaceVersion)
+		imageName, _ := agent.GetLoadedImageName()
+		logger.Info(fmt.Sprintf("Successfully loaded appnet agent container tarball: %s", agentContainerTarballPath),
+			logger.Fields{
+				field.Image: imageName,
+			})
+		return loader.GetContainerImage(imageName, dockerClient)
+	}
+	return nil, loadErr
 }
 
 func (agent *manager) IsLoaded(dockerClient dockerapi.DockerClient) (bool, error) {
@@ -391,7 +412,12 @@ func (agent *manager) IsLoaded(dockerClient dockerapi.DockerClient) (bool, error
 }
 
 func (agent *manager) GetLoadedImageName() (string, error) {
-	return fmt.Sprintf("%s:%s", agent.AgentContainerImageName, agent.AgentContainerTag), nil
+	agent.agentContainerTag = fmt.Sprintf(defaultAgentContainerTagFormat, agent.appnetInterfaceVersion)
+	return fmt.Sprintf("%s:%s", agent.agentContainerImageName, agent.agentContainerTag), nil
+}
+
+func (agent *manager) GetLoadedAppnetVersion() (string, error) {
+	return agent.appnetInterfaceVersion, nil
 }
 
 // getECSAgentLogPathContainer returns the directory path for ECS_LOGFILE env value if exists, otherwise returns "/log"
@@ -401,4 +427,18 @@ func getECSAgentLogPathContainer() string {
 		return defaultECSAgentLogPathContainer
 	}
 	return path.Dir(ecsLogFilePath)
+}
+
+// GetCapabilitiesForAppnetInterfaceVersion returns service connect capabilities
+// supported by ECS Agent to register for a selected AppNet version.
+// Suppose we decide to register ecs.service-connect.v2 capability for new AppNet version (ex: 1.24.0.0),
+// now if ecs.service-connect.*v1* is continuously being supported by 1.24.0.0,
+// we will then register multiple capabilities.
+//
+//	{
+//	   "v1": ["ecs.capability.service-connect-v1"],
+//	   "v2": ["ecs.capability.service-connect-v2", "ecs.capability.service-connect-v2"]
+//	}
+func (agent *manager) GetCapabilitiesForAppnetInterfaceVersion(appnetVersion string) ([]string, error) {
+	return supportedAppnetInterfaceVerToCapability[appnetVersion], nil
 }
--- a/agent/engine/serviceconnect/manager_linux_test.go
+++ b/agent/engine/serviceconnect/manager_linux_test.go
@@ -98,8 +98,8 @@ func TestPauseContainerModificationsForServiceConnect(t *testing.T) {
 
 			hostConfig := &dockercontainer.HostConfig{}
 			scManager := &manager{
-				AgentContainerImageName: "container_image",
-				AgentContainerTag:       "tag",
+				agentContainerImageName: "container_image",
+				agentContainerTag:       "tag",
 			}
 			err := scManager.AugmentTaskContainer(scTask, tc.container, hostConfig)
 			if err != nil {
@@ -150,6 +150,35 @@ func TestGetECSAgentLogPathContainer(t *testing.T) {
 			}
 			actualPath := getECSAgentLogPathContainer()
 			assert.Equal(t, tc.expected, actualPath)
+
+		})
+	}
+}
+
+func TestGetSupportedAppnetInterfaceVerToCapabilities(t *testing.T) {
+	testCases := []struct {
+		name                 string
+		appNetAgentVersion   string
+		expectedCapabilities []string
+	}{
+		{
+			name:                 "test supported service connect capabilities for AppNet agent version v1",
+			appNetAgentVersion:   "",
+			expectedCapabilities: nil,
+		},
+		{
+			name:                 "test supported service connect capabilities for AppNet agent version v1",
+			appNetAgentVersion:   "v1",
+			expectedCapabilities: []string{"ecs.capability.service-connect-v1"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			scManager := &manager{}
+			scCapabilities, err := scManager.GetCapabilitiesForAppnetInterfaceVersion(tc.appNetAgentVersion)
+			assert.NoError(t, err)
+			assert.Equal(t, tc.expectedCapabilities, scCapabilities)
 		})
 	}
 }
--- a/agent/engine/serviceconnect/manager_other.go
+++ b/agent/engine/serviceconnect/manager_other.go
@@ -69,3 +69,15 @@ func (*manager) GetLoadedImageName() (string, error) {
 		"appnetAgent container get image name: unsupported platform: %s/%s",
 		runtime.GOOS, runtime.GOARCH))
 }
+
+func (*manager) GetLoadedAppnetVersion() (string, error) {
+	return "", loader.NewUnsupportedPlatformError(fmt.Errorf(
+		"appnetAgent container get loaded appnet version: unsupported platform: %s/%s",
+		runtime.GOOS, runtime.GOARCH))
+}
+
+func (*manager) GetCapabilitiesForAppnetInterfaceVersion(string) ([]string, error) {
+	return make([]string, 0), loader.NewUnsupportedPlatformError(fmt.Errorf(
+		"appnetAgent container get capabilities for appnet version: unsupported platform: %s/%s",
+		runtime.GOOS, runtime.GOARCH))
+}
--- a/agent/engine/serviceconnect/version_linux.go
+++ b/agent/engine/serviceconnect/version_linux.go
@@ -1,0 +1,64 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package serviceconnect
+
+import (
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/aws/amazon-ecs-agent/agent/logger"
+	"github.com/aws/amazon-ecs-agent/agent/logger/field"
+)
+
+const (
+	appnetInterfaceV1          = "v1"
+	serviceConnectCapabilityV1 = "ecs.capability.service-connect-v1"
+)
+
+var (
+	supportedAppnetInterfaceVerToCapability = map[string][]string{
+		appnetInterfaceV1: {
+			serviceConnectCapabilityV1,
+		},
+	}
+)
+
+// getSupportedAppnetInterfaceVersions returns the all the supported AppNet interface versions
+// by ECS Agent from highest to lowest version
+func getSupportedAppnetInterfaceVersions() []string {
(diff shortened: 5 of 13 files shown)


agent/engine/serviceconnect/version_linux.go
```diff
@@ -0,0 +1,52 @@
...
+// getSupportedAppnetInterfaceVersion returns the latest supported AppNet interface version
+// by looking for the AppNet agent tar name available on the instance
+// from supportedAppnetInterfaceVerToCapability map from highest to lowest version
+func getSupportedAppnetInterfaceVersion() string {
+	supportedAppnetInterfaceVersions := make([]string, 0, len(supportedAppnetInterfaceVerToCapability))
+	for supportedAppnetInterfaceVersion := range supportedAppnetInterfaceVerToCapability {
+		supportedAppnetInterfaceVersions = append(supportedAppnetInterfaceVersions, supportedAppnetInterfaceVersion)
+	}
+	sort.Sort(sort.Reverse(sort.StringSlice(supportedAppnetInterfaceVersions)))
+
```
yinyic, line 44: I think it would be better if we return the list right here, and let the loader go through the list and validate the existence. So something like
```
for each appnetInterfaceVersion in getSupportedAppnetInterfaceVersions():  // from highest to lowest
  if tar file does not exist, continue
  if loading tar file failed, continue
  # if we reach here we've successfully loaded a tar file
  break
```
That's because
- it will be a cleaner separation between static capability and runtime validation (which needs to be performed by loader anyway)
- there might be cases where the tar file exists but is corrupted, and in such case we should let loader keeping trying older version of AppNet

agent/app/agent_capability.go
```diff
@@ -458,7 +458,11 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 		}
 	}
```
yinyic, line 460: More details in the comment for `getSupportedAppnetInterfaceVersion` - I think we should let loader go through the list of supported appnet versions and keep trying until it successfully loads a tar, and it returns that version.

agent/app/agent_capability.go
```diff
@@ -458,7 +458,11 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
...
 
-	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityServiceConnect)
+	supportedAppnetInterfaceVerToCapabilities := agent.serviceconnectManager.GetSupportedAppnetInterfaceVerToCapabilities()
```
yinyic, line 461: Ditto - the knowledge of "which appnet version are we actually using" should come from loader as it's a runtime decision.

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -139,9 +143,11 @@ func NewManager() Manager {
...
 		logPathECSAgentRoot: fmt.Sprintf(defaultECSAgentLogPathForSC, getECSAgentLogPathContainer()),
 
-		AgentContainerImageName:   defaultAgentContainerImageName,
-		AgentContainerTag:         defaultAgentContainerTag,
-		AgentContainerTarballPath: defaultAgentContainerTarballPath,
+		AgentContainerImageName: defaultAgentContainerImageName,
+		AgentContainerTag:       defaultAgentContainerTag,
+		AgentContainerVersion:   getSupportedAppnetInterfaceVersion(),
+		AgentContainerTarballPath: fmt.Sprintf(defaultAgentContainerTarballPathFormat,
+			getSupportedAppnetInterfaceVersion()),
```
yinyic, line 150: Ditto - the final appnet agent version being loaded should come from loader.

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -372,17 +370,30 @@ func (m *manager) AugmentInstanceContainer(task *apitask.Task, container *apicon
...
 func (agent *manager) LoadImage(ctx context.Context, _ *config.Config, dockerClient dockerapi.DockerClient) (*types.ImageInspect, error) {
-	logger.Debug("Loading appnet agent container tarball:", logger.Fields{
-		field.Image: agent.AgentContainerTarballPath,
-	})
-	if err := loader.LoadFromFile(ctx, agent.AgentContainerTarballPath, dockerClient); err != nil {
-		return nil, err
+	var err error
+	for _, supportedAppnetInterfaceVersion := range getSupportedAppnetInterfaceVersions() {
+		agentContainerTarballPath := fmt.Sprintf(defaultAgentContainerTarballPathFormat, supportedAppnetInterfaceVersion)
+		if _, err = os.Stat(agentContainerTarballPath); err == nil {
```
yinyic, line 380: Maybe also check for non-nil error here to be more consistent with load error below?
```
if _, err = os.Stat(agentContainerTarballPath); err != nil {
 // log a warning message
 continue
}
if loadErr...
```

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -372,17 +370,30 @@ func (m *manager) AugmentInstanceContainer(task *apitask.Task, container *apicon
...
+			if loadErr := loader.LoadFromFile(ctx, agentContainerTarballPath, dockerClient); err != nil {
+				logger.Warn("Unable to load appnet agent container tarball:", logger.Fields{
+					field.Image: agentContainerTarballPath,
+					field.Error: loadErr,
+				})
+				continue
+			}
+			agent.AppnetInterfaceVersion = supportedAppnetInterfaceVersion
+			imageName, _ := agent.GetLoadedImageName()
+			return loader.GetContainerImage(imageName, dockerClient)
```
yinyic, line 393: Let's log an info message here to state the tar we've successfully loaded, and what we are assuming the loaded image name to be

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -402,3 +414,16 @@ func getECSAgentLogPathContainer() string {
...
+// Suppose we decide to register ecs.service-connect.v2 capability for new AppNet version (ex: 1.24.0.0),
+// now if ecs.service-connect.*v1* is continuously being supported by 1.24.0.0,
+// we will then register multiple capabilities.
+// {
+//    "v1": ["ecs.service-connect.v1"],
+//    "v2": ["ecs.service-connect.v1", "ecs.service-connect.v2"]
+// }
+func (agent *manager) GetSupportedAppnetInterfaceVerToCapabilities() []string {
+	return supportedAppnetInterfaceVerToCapability[agent.AppnetInterfaceVersion]
+}
```
yinyic, line 429: Nit - maybe rename to `GetCapabilitiesForAppnetInterfaceVersion`

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -402,3 +414,16 @@ func getECSAgentLogPathContainer() string {
...
+// supported by ECS Agent to register for a selected AppNet version.
+// Suppose we decide to register ecs.service-connect.v2 capability for new AppNet version (ex: 1.24.0.0),
+// now if ecs.service-connect.*v1* is continuously being supported by 1.24.0.0,
+// we will then register multiple capabilities.
+// {
+//    "v1": ["ecs.service-connect.v1"],
+//    "v2": ["ecs.service-connect.v1", "ecs.service-connect.v2"]
+// }
+func (agent *manager) GetSupportedAppnetInterfaceVerToCapabilities() []string {
+	return supportedAppnetInterfaceVerToCapability[agent.AppnetInterfaceVersion]
```
yinyic, line 428: I feel that it may better to have caller (`appendServiceConnectCapabilities`) explicitly pass in the appnet version. Currently it first calls loader, loader will set the global appnet version, which is used by this method, so we are introducing some dependencies between those two methods which couldn't been independent. 

Let's maybe have below in `appendServiceConnectCapabilities` instead:
```
loadedVer, err := agent.LoadImage(...)
// handle err
capabilities := agent.GetCapabilitiesForAppnetInterfaceVersion(loadedVer)
// if nil, throw warning
// append the capabilities
```
Overall feels a bit more clear to me. What do you think?

agent/engine/serviceconnect/version_linux.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+// getSupportedAppnetInterfaceVersions returns the all the supported AppNet interface versions
+// by ECS Agent from highest to lowest version
+func getSupportedAppnetInterfaceVersions() []string {
+	supportedAppnetInterfaceVersions := make([]string, 0, len(supportedAppnetInterfaceVerToCapability))
+	for supportedAppnetInterfaceVersion := range supportedAppnetInterfaceVerToCapability {
+		supportedAppnetInterfaceVersions = append(supportedAppnetInterfaceVersions, supportedAppnetInterfaceVersion)
+	}
+	sort.Sort(sort.Reverse(sort.StringSlice(supportedAppnetInterfaceVersions)))
```
yinyic, line 40: Will this become problematic if we are comparing `v10` to `v2`?

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: Will add a check and a warning log here, thanks!

agent/engine/serviceconnect/version_linux.go
mythri-garaga: Good catch! I have updated the sorting logic to handle it correctly.
Thank you!

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: Done:)

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: Added the info message.

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: The LoadImage() function is used to load pause container image as well, thought returning version might not add much value in that context. 
I have added a get and set functions to manage the appnet version, hope it makes it a bit clear, please let me know:)

agent/logger/field/constants.go
```diff
@@ -36,4 +36,5 @@ const (
 	Volume        = "volume"
 	Time          = "time"
 	NetworkMode   = "networkMode"
+	Tar           = "tar"
```
yinyic, line 39: I'm not sure if we will be referencing AppNet tar frequently enough log to merit adding it as a logger field, seems like only loader uses it... And also because the fields so far have been general task properties, but appnet tar name not so much. 
I think I'd probably rather just use formatted string in the few places we want to log the tar file name.

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -372,17 +370,43 @@ func (m *manager) AugmentInstanceContainer(task *apitask.Task, container *apicon
...
-		field.Image: agent.AgentContainerTarballPath,
-	})
-	if err := loader.LoadFromFile(ctx, agent.AgentContainerTarballPath, dockerClient); err != nil {
-		return nil, err
+	var err error
+	for _, supportedAppnetInterfaceVersion := range getSupportedAppnetInterfaceVersions() {
+		agentContainerTarballPath := fmt.Sprintf(defaultAgentContainerTarballPathFormat, supportedAppnetInterfaceVersion)
+		if _, err = os.Stat(agentContainerTarballPath); err != nil {
+			logger.Warn("Agent container tarball unavailable:", logger.Fields{
+				field.Tar:   agentContainerTarballPath,
```
yinyic, line 386: Ditto - Not sure we want to add it to logger field (and the ones below)

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -78,9 +78,9 @@ const (
 	defaultAdminStatsRequest = httpRequestPrefix + "/stats/prometheus?usedonly&filter=metrics_extension&delta"
 	defaultAdminDrainRequest = httpRequestPrefix + "/drain_listeners?inboundonly"
 
-	defaultAgentContainerTarballPath = "/managed-agents/serviceconnect/appnet_agent.interface-v1.tar"
-	defaultAgentContainerImageName   = "appnet_agent"
-	defaultAgentContainerTag         = "service_connect.v1"
+	defaultAgentContainerImageName         = "ecs-service-connect-agent"
+	defaultAgentContainerTagFormat         = "interface-%s"
+	defaultAgentContainerTarballPathFormat = "ecs-service-connect-agent.interface-%s.tar"
```
yinyic, line 83: Sorry if my previous comment didn't make it clear - we still need the `/managed-agents/serviceconnect/` path, just that the tar file name is now `ecs-service-connect-agent.interface-%s.tar` instead of `appnet_agent.interface-%s.tar`

agent/engine/serviceconnect/version_linux.go
```diff
@@ -0,0 +1,77 @@
...
+		func(i, j int) bool {
+			return sortVersion(supportedAppnetInterfaceVersions[i]) > sortVersion(supportedAppnetInterfaceVersions[j])
+		},
+	)
+	return supportedAppnetInterfaceVersions
+}
+
+// sortVersion returns a version sort key in numeric order.
+// For example, v20, v11, v10, v2, v1,...
+func sortVersion(version string) int {
```
yinyic, line 68: I'd probably rename it - this function itself simply parses the int value from version string, and doesn't do any sorting on its own. Maybe `getVersionNumber` or something more appropriate.

agent/engine/serviceconnect/version_linux.go
```diff
@@ -0,0 +1,77 @@
...
+	for supportedAppnetInterfaceVersion := range supportedAppnetInterfaceVerToCapability {
+		allowedVersionPattern, err := regexp.Compile(appNetInterfaceVersionPattern)
+		if err != nil {
+			logger.Warn("Unable to compile regex pattern:", logger.Fields{
+				field.Error: err,
+			})
+		}
+		if !allowedVersionPattern.MatchString(supportedAppnetInterfaceVersion) {
+			logger.Warn("Appnet interface version does not match the expected pattern, ex: `v1`")
+		}
```
yinyic, line 54: I feel like this should be something caught in a unit test instead of runtime. This is all static information known at compile time.

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: my bad, I updated it incorrectly by mistake. Will fix it

agent/engine/serviceconnect/version_linux.go
mythri-garaga: Agreed! Moved the validation to a unit test.

yinyic (approved):
LGTM! Would like to see we verify the integration with appnet rpm with a manual test, mainly to check if all the tarball path, image name and tags are as expected.

mythri-garaga:
> LGTM! Would like to see we verify the integration with appnet rpm with a manual test, mainly to check if all the tarball path, image name and tags are as expected.

Tested these changes by building a test AMI. ECS Agent was able to load the new appnet image as expect - 
```
level=debug time=2022-10-10T19:25:18Z msg="Inspecting container image: " amazon/amazon-ecs-pause:0.1.0="amazon/amazon-ecs-pause:0.1.0"
level=debug time=2022-10-10T19:25:18Z msg="Inspecting container image: " ecs-service-connect-agent:interface-="ecs-service-connect-agent:interface-"
level=debug time=2022-10-10T19:25:18Z msg="Loading appnet agent container tarball: /managed-agents/serviceconnect/ecs-service-connect-agent.interface-v1.tar"
level=info time=2022-10-10T19:25:25Z msg="Successfully loaded appnet agent container tarball: /managed-agents/serviceconnect/ecs-service-connect-agent.interface-v1.tar" image="ecs-service-connect-agent:interface-v1"
level=debug time=2022-10-10T19:25:25Z msg="Inspecting container image: " ecs-service-connect-agent:interface-v1="ecs-service-connect-agent:interface-v1"
```

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -115,9 +115,9 @@ type manager struct {
 	// Http path + params to make a drain request of AppNetAgent
 	adminDrainRequest string
 
-	AgentContainerImageName   string
-	AgentContainerTag         string
-	AgentContainerTarballPath string
+	AgentContainerImageName string
```
chienhanlin, line 118: Q. are these variables needed to be exported? should we change them to lower-case?

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -372,17 +370,40 @@ func (m *manager) AugmentInstanceContainer(task *apitask.Task, container *apicon
...
-	logger.Debug("Loading appnet agent container tarball:", logger.Fields{
-		field.Image: agent.AgentContainerTarballPath,
-	})
-	if err := loader.LoadFromFile(ctx, agent.AgentContainerTarballPath, dockerClient); err != nil {
-		return nil, err
+	var err error
+	for _, supportedAppnetInterfaceVersion := range getSupportedAppnetInterfaceVersions() {
+		agentContainerTarballPath := fmt.Sprintf(defaultAgentContainerTarballPathFormat, supportedAppnetInterfaceVersion)
+		if _, err = os.Stat(agentContainerTarballPath); err != nil {
+			logger.Warn(fmt.Sprintf("Agent container tarball unavailable: %s", agentContainerTarballPath), logger.Fields{
```
chienhanlin, line 385: Not blocking. Could we add "Appnet" or "AppNet" before "Agent container tarball unavailable" to avoid mixing it up with ECS Agent or other Agents? And could we have a standard for using "Appnet", "appnet" or "AppNet" in all warning/error messages?  Thanks!

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -391,9 +412,14 @@ func (agent *manager) IsLoaded(dockerClient dockerapi.DockerClient) (bool, error
 }
 
 func (agent *manager) GetLoadedImageName() (string, error) {
```
chienhanlin, line 414: Should we remove returning `error` from the return value since it will always be nil?

agent/app/agent_capability.go
```diff
@@ -457,8 +457,18 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 			return capabilities
 		}
 	}
-
-	return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityServiceConnect)
+	loadedVer := agent.serviceconnectManager.GetLoadedAppnetVersion()
```
chienhanlin, line 460: Should we add a check here just in case `GetLoadedAppnetVersion()` returns  `nil `or an `empty` string?

agent/engine/serviceconnect/manager_other.go
```diff
@@ -69,3 +69,11 @@ func (*manager) GetLoadedImageName() (string, error) {
 		"appnetAgent container get image name: unsupported platform: %s/%s",
 		runtime.GOOS, runtime.GOARCH))
 }
+
+func (*manager) GetLoadedAppnetVersion() string {
+	return ""
```
chienhanlin, line 74: Q. should we also return a `NewUnsupportedPlatformError` here?

agent/engine/serviceconnect/version_linux.go
```diff
@@ -0,0 +1,64 @@
...
+		},
+	)
+	return supportedAppnetInterfaceVersions
+}
+
+// getVersionNumber returns a version sort key in numeric order.
+func getVersionNumber(version string) int {
+	versionNumber := strings.TrimPrefix(version, "v")
+	num, err := strconv.Atoi(versionNumber)
+	if err != nil {
```
chienhanlin, line 58: Q. should we consider dropping versionNumber that failed to be parsed, and removing it from the `supportedAppnetInterfaceVersions` before returning by `getSupportedAppnetInterfaceVersions()`?

agent/engine/serviceconnect/version_linux_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestSupportedAppnetInterfaceVersion(t *testing.T) {
+	allowedVersionPattern, err := regexp.Compile(appNetInterfaceVersionPattern)
+	for supportedAppnetInterfaceVersion := range supportedAppnetInterfaceVerToCapability {
+		require.NoError(t, err, "Error while compiling regex pattern")
+		require.True(t, allowedVersionPattern.MatchString(supportedAppnetInterfaceVersion),
+			"Appnet interface version does not match the expected pattern, ex:`v1`")
+	}
+}
+func TestGetSupportedAppnetInterfaceVersion(t *testing.T) {
```
chienhanlin, line 36: Could we add negative tests cases to verify how errors will be handled : )

agent/engine/serviceconnect/manager_linux_test.go
```diff
@@ -150,6 +150,28 @@ func TestGetECSAgentLogPathContainer(t *testing.T) {
 			}
 			actualPath := getECSAgentLogPathContainer()
 			assert.Equal(t, tc.expected, actualPath)
+
+		})
+	}
+}
+
+func TestGetSupportedAppnetInterfaceVerToCapabilities(t *testing.T) {
```
chienhanlin, line 158: Could we add negative tests cases to verify how errors will be handled : )

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: Ack! 

Since all these callouts are non blocking, will address them in the upcoming PR related to AppNet reload(including all the below comments as well) :)
Thank you!

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: Ya  I can change this to `agentContainerImageName` since it not being exported.

agent/app/agent_capability.go
mythri-garaga: We have warning log on 463 and 464 which will log `loadedVer` if it's empty(one of the scenario).

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: we return error for unsupported platforms so I think good to keep it

agent/engine/serviceconnect/manager_other.go
mythri-garaga: sure, will make this change!

agent/engine/serviceconnect/manager_linux_test.go
mythri-garaga: GetSupportedAppnetInterfaceVerToCapabilities() returns errors only for unsupported platforms so adding a test case that handles a empty string for appNetAgentVersion.

agent/engine/serviceconnect/version_linux.go
mythri-garaga: `getVersionNumber()` is only used by `getSupportedAppnetInterfaceVersions()` and there is already a unit test to validate the input for `supportedAppnetInterfaceVersions` so strconv.Atoi(versionNumber) will not return zero in this scenario.

agent/engine/serviceconnect/version_linux_test.go
mythri-garaga: same reason as above, getSupportedAppnetInterfaceVersions() returns error when it cannot parse int with strconv.Atoi(versionNumber) but versions are explicitly validated in a unit case so negative test cases are not required here.

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -372,17 +370,40 @@ func (m *manager) AugmentInstanceContainer(task *apitask.Task, container *apicon
...
+		if loadErr = loader.LoadFromFile(ctx, agentContainerTarballPath, dockerClient); loadErr != nil {
+			logger.Warn(fmt.Sprintf("Unable to load appnet agent container tarball: %s", agentContainerTarballPath),
+				logger.Fields{
+					field.Error: loadErr,
+				})
+			continue
+		}
+		agent.setLoadedAppnetVerion(supportedAppnetInterfaceVersion)
+		imageName, _ := agent.GetLoadedImageName()
+		logger.Info(fmt.Sprintf("Successfully loaded appnet agent container tarball: %s", agentContainerTarballPath),
```
chienhanlin, line 400: Can we update the log message to `AppNet agent` in the follow up PR? Thanks!

agent/engine/serviceconnect/manager_linux.go
```diff
@@ -372,17 +370,40 @@ func (m *manager) AugmentInstanceContainer(task *apitask.Task, container *apicon
...
+			logger.Fields{
+				field.Image: imageName,
+			})
+		return loader.GetContainerImage(imageName, dockerClient)
 	}
-
-	imageName, _ := agent.GetLoadedImageName()
-	return loader.GetContainerImage(imageName, dockerClient)
+	return nil, loadErr
```
chienhanlin, line 406: Not blocking and more a question. In addition to show the warning log message, should we aggregate loadErr got from the loop? Only the last error, which might for the lowest AppNet agent version, will be returned with the current logic.

agent/engine/serviceconnect/manager_other.go
```diff
@@ -69,3 +69,15 @@ func (*manager) GetLoadedImageName() (string, error) {
 		"appnetAgent container get image name: unsupported platform: %s/%s",
 		runtime.GOOS, runtime.GOARCH))
 }
+
+func (*manager) GetLoadedAppnetVersion() (string, error) {
+	return "", loader.NewUnsupportedPlatformError(fmt.Errorf(
+		"appnetAgent container get loaded appnet version: unsupported platform: %s/%s",
```
chienhanlin, line 75: Can we update the log message to `AppNet agent` for line 75 and line 81 in the follow up PR? Thanks!

chienhanlin (approved):

agent/engine/serviceconnect/manager_other.go
mythri-garaga: Was following the same convention as other functions in this file...

agent/engine/serviceconnect/manager_linux.go
mythri-garaga: On a second thought, error aggregation is not required since we log the error message for each version we try load.

fabrizzioalco:
Hello! Where does the `appnet_agent.interface-v1.tar` coming from? Im having an issue where installing ecs agentl, the file is not present on the ec2 instance

PR #3425, opened 2022-10-10
windows: changed method for retrieving Windows network statistics in case of awsvpc network mode

rawahars:
## Summary
Previously, we were running a Powershell method (Get-NetAdapterStatistics) to retrieve the network statistics when the  task is using `awsvpc network mode`. However, this can cause CPU spikes on smaller instance types. 

Another method to obtain the network statistics would be access them using Win32 API. This method is significantly more performant than using Powershell cmdlet. Therefore, we are switching the agent implementation to use native Win32 API for retrieving the same.

## Implementation
When we initialise the stats module for a given task, we will query and store the `interface LUID` of the task ENI. During the subsequent calls to retrieve the network statistics, we will query the statistics specific to the interface using it's LUID.

In order to accomplish the same, we are introducing two new APIs in `NetworkUtils` interface would be the preferred way to access Windows networking APIs in the agent. The two APIs are-
- ConvertInterfaceAliasToLUID: Converts the interface alias into the corresponding LUID.
    - Internally invokes `ConvertInterfaceAliasToLuid` Win32 API to convert interface alias into LUID.
- GetMIBIfEntryFromLUID: Returns the MIB_IF_ROW2 for the interface with the given LUID.
    - Internally invokes `GetIfEntry2Ex` Win32 API to obtain the specific interface which has given LUID.

Reference: 
- https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-convertinterfacealiastoluid
- https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-getifentry2ex

**Note:** As part of this PR, refactoring has been done to improve the code quality of existing **Windows** workflow.
- Removed `Setter` method from `networkUtils` package
- Removed `Setter` method from `watcher` package
- Introduced mocks for `networkUtils` package and used the same in `watcher` tests instead of object injection.

## Testing
Tested the new agent using a custom AMI.
Tested it out on smaller instance types for any periodic spikes.

New tests cover the changes: 
Yes

### Licensing


agent/stats/statsretriever/stats_retriever_windows.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+// NewStatsRetriever returns an instance of StatsRetriever interface.
+func NewStatsRetriever() StatsRetriever {
+	log.Debugf("Initializing stats retriever for Windows")
+	obj := &retriever{}
+
+	// We would be using GetIfTable2Ex and FreeMibTable Win32 APIs from IP Helper.
+	moduleIPHelper := windows.NewLazySystemDLL("iphlpapi.dll")
+	procGetIfTable2Ex := moduleIPHelper.NewProc("GetIfTable2Ex")
```
ofiliz, line 43: Why are we retrieving the entire table and not the specific interface row?
https://learn.microsoft.com/en-us/windows/win32/api/netioapi/nf-netioapi-getifentry2ex?

rawahars: Ah we have been searching for this particular method all along. Unfortunately, we were not aware about the same. 
I'll make the changes to use `GetIfEntry2Ex` API.

agent/stats/task_windows.go
```diff
@@ -17,64 +18,42 @@ package stats
...
+	// Find and store the device name along with the interface LUID.
 	for index, device := range taskENIs {
 		devices[index] = device.LinkName
+
+		ifEntry, err := netUtils.GetMIBIfEntryFromDeviceName(device.LinkName)
+		if err != nil {
+			cancel()
+			return nil, errors.Wrapf(err, "failed to initialise stats task container")
+		}
+		ifaceLUID[index] = ifEntry.InterfaceLUID
```
jterry75, line 56: I still dont quite understand. Cant we just use `ConvertInterfaceIndexToLuid` for this and then be done? We dont need to do an entire mib table and filter it right?

agent/eni/networkutils/utils_windows.go
```diff
@@ -141,13 +157,67 @@ func (utils *networkUtils) GetDNSServerAddressList(macAddress string) ([]string,
...
 		firstDnsNode = firstDnsNode.Next
 	}
 
 	return dnsServerAddressList, nil
 }
 
+// GetMIBIfEntryFromDeviceName return the MIB_IF_ROW2 object for the queried device.
+// This method would internally invoke the GetIfTable2Ex Windows API to get complete interface list
+// from which the queried device would be filtered.
+func (utils *networkUtils) GetMIBIfEntryFromDeviceName(device string) (*MibIfRow2, error) {
```
jterry75, line 170: Look at my other comment using the LUID. I think we can remove this entirely.

agent/eni/networkutils/utils_windows.go
```diff
@@ -201,3 +259,37 @@ func getAdapterAddresses() ([]*windows.IpAdapterAddresses, error) {
...
+	retVal, _, _ := procGetIfTable2Ex.Call(uintptr(0), uintptr(unsafe.Pointer(&tab)))
+	if retVal != 0 {
+		return nil, errors.Errorf("error occured while calling GetIfTable2Ex: %s", syscall.Errno(retVal))
+	}
+
+	return tab, nil
+}
+
+// getIfEntry2Ex invokes GetIfEntry2Ex Win32 API to retrieve the MIB_IF_ROW2 for the interface with given LUID.
+func getIfEntry2Ex(ifRow *MibIfRow2) (*MibIfRow2, error) {
```
jterry75, line 278: Did we add this extra function just for unit testing?

agent/eni/networkutils/utils_windows_test.go
```diff
@@ -237,3 +254,39 @@ func TestGetDNSServerAddressList(t *testing.T) {
...
+
+	netUtils := &networkUtils{}
+	ifRow, err := netUtils.GetMIBIfEntryFromDeviceName(deviceName)
+	assert.NoError(t, err)
+	assert.Equal(t, uint32(interfaceIndex), ifRow.interfaceIndex)
+	assert.Equal(t, ifaceLUID, ifRow.InterfaceLUID)
+}
+
+// TestGetMIBIfEntryFromLUID tests the GetMIBIfEntryFromLUID method.
+func TestGetMIBIfEntryFromLUID(t *testing.T) {
```
jterry75, line 279: Since we went to all the trouble here lets add the negative test if we return error from the call too.

agent/eni/watcher/watcher_windows.go
```diff
@@ -65,6 +65,7 @@ func newWatcher(ctx context.Context,
 	notificationChannel := make(chan int)
 	err := eniMonitor.Start(notificationChannel)
 	if err != nil {
+		cancel()
```
jterry75, line 68: Same, move ln 62 to 72 and then we dont have to handle this case.

agent/stats/task_windows.go
```diff
@@ -17,64 +18,42 @@ package stats
...
 func newStatsTaskContainer(taskARN, taskId, containerPID string, numberOfContainers int,
 	resolver resolver.ContainerMetadataResolver, publishInterval time.Duration, taskENIs task.TaskENIs) (*StatsTask, error) {
 	ctx, cancel := context.WithCancel(context.Background())
 
+	// Instantiate an instance of network utils.
+	// This interface would be used to invoke Windows networking APIs.
+	netUtils := networkutils.New()
+
 	devices := make([]string, len(taskENIs))
+	ifaceLUID := make([]uint64, len(taskENIs))
```
jterry75, line 46: You can have multiple eni's per task?

agent/stats/task_windows_test.go
```diff
@@ -108,14 +85,16 @@ func TestTaskStatsCollection(t *testing.T) {
 			TaskMetadata: &TaskMetadata{
 				TaskArn:          taskId,
 				ContainerPID:     containerPID,
-				DeviceName:       []string{"Ethernet 3"},
+				DeviceName:       []string{deviceName},
```
jterry75, line 88: Why are we using DeviceName? Isnt the ACS api based on mac and then we find index? Once we find Index we can get LUID. And from there we are good to go for basically all win32 api's. Why use name?

agent/stats/task_windows.go
rawahars: So the thing is that we do not store interface index for the task network adapters. There can be two cases here-
- We query the interface index first using `net.Interfaces` and then use `ConvertInterfaceIndexToLuid` in the stats module itself. This leads to two Win32 calls. Now, `net.Interfaces` would in turn call `GetAdaptersAddresses` which fetches all the addresses associated with the adapters. I feel that a single call to `GetIfTable2Ex` might be more performant than these two calls.
- We store the ENI's interface index along with Link name during the initial task attachment workflow. Now, the issue here is that if the interface index becomes stale for some reason say, instance reboot, we would not have any mechanism to identify or refresh the same.

agent/eni/networkutils/utils_windows.go
rawahars: Yes we did :)

agent/stats/task_windows.go
rawahars: That is the case with trunk ENI I think.

agent/stats/task_windows_test.go
rawahars: Well that is the current workflow which is shared with the Linux. We store the stats corresponding to each device name. Reference: https://github.com/aws/amazon-ecs-agent/blob/67240c17711c0d0a5c7b1d37882dc565880bff38/agent/stats/task.go#L151

Regarding fetching interface index, I have mentioned the same in earlier comment.

sparrc (approved):

singholt (approved):

PR #3435, opened 2022-10-17
Add file watcher for Appnet agent image update

mythri-garaga:
### Summary

This PR adds a file watcher for filepath (`/var/lib/ecs/deps/serviceconnenct`) where ECS Agent loads Appnet agent image from. With these changes, if there any update current Appnet agent image or a new Appnet agent image version becomes available, the file watcher invokes to reload this image and restart the internal instance relay with the updated image.

Update go.mod and vendor files to include fsnotify package.

### Testing

`make test`

Also for manual testing purpose,  made changes to Agent to include `v2` as supported Appnet version and tested the following scenarios
1. ECS Agent is running and has `ecs-service-connect-agent:interface-v1` appnet image loaded. There are no Service Connect task running on the instance. Placed the Appnet v2 - `ecs-service-connect-agent.interface-v2.tar` tarball in the service connect deps dir. The ECS Agent agent reloaded the image `ecs-service-connect-agent:v2`.
2. Tested scenario 1 with Service Connect task running in bridge mode. Observed that ECS Agent reloaded the new Appnet image and restart the instance relay task with `ecs-service-connect-agent:v2`
3. With Scenario 2, that is ECS Agent has reloaded the new Appnet image and restarted the instance relay task with `ecs-service-connect-agent:v2` but SC task's Appnet  agent is still running with `ecs-service-connect-agent:v1` image.
Stopped this SC task and started a new one with same task def. The new SC task started successfully and it's Appnet agent is still running with `ecs-service-connect-agent:v2` image.

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -281,6 +281,7 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
 	go engine.handleDockerEvents(derivedCtx)
 	engine.initialized = true
 	go engine.startPeriodicExecAgentsMonitoring(derivedCtx)
+	go engine.watchAppNetImage(derivedCtx)
 	return nil
 }
 
--- a/agent/engine/docker_task_engine_linux.go
+++ b/agent/engine/docker_task_engine_linux.go
@@ -17,10 +17,16 @@
 package engine
 
 import (
+	"context"
+	"fmt"
 	"time"
+
+	"github.com/fsnotify/fsnotify"
 
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+	apitaskstatus "github.com/aws/amazon-ecs-agent/agent/api/task/status"
+	"github.com/aws/amazon-ecs-agent/agent/logger"
 )
 
 const (
@@ -38,4 +44,79 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
 // On non-windows platform, we will not invoke CNI plugins for non-pause containers
 func (engine *DockerTaskEngine) invokePluginsForContainer(task *apitask.Task, container *apicontainer.Container) error {
 	return nil
+}
+
+func (engine *DockerTaskEngine) watchAppNetImage(ctx context.Context) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error(fmt.Sprintf("failed to initialize fsnotify NewWatcher, error: %v", err))
+	}
+	appnetContainerTarballDir := engine.serviceconnectManager.GetAppnetContainerTarballDir()
+	err = watcher.Add(appnetContainerTarballDir)
+	if err != nil {
+		logger.Error(fmt.Sprintf("error adding %s to fsnotify watcher, error: %v", appnetContainerTarballDir, err))
+	}
+	defer watcher.Close()
+
+	// Start listening for events.
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				logger.Warn("fsnotify event watcher channel is closed")
+				return
+			}
+			// check if the event file operation is write or create
+			const writeOrCreateMask = fsnotify.Write | fsnotify.Create
+			if event.Op&writeOrCreateMask != 0 {
+				logger.Debug(fmt.Sprintf("new fsnotify watcher event: %s", event.Name))
+				// reload the updated Appnet Agent image
+				if err := engine.reloadAppNetImage(); err == nil {
+					// restart the internal instance relay task with
+					// updated Appnet Agent image
+					engine.restartInstanceTask()
+				}
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				logger.Warn("fsnotify event watcher channel is closed")
+				return
+			}
+			logger.Error(fmt.Sprintf("fsnotify watcher error: %v", err))
+		case <-ctx.Done():
+			return
+		}
+	}
+}
+
+func (engine *DockerTaskEngine) reloadAppNetImage() error {
+	_, err := engine.serviceconnectManager.LoadImage(engine.ctx, engine.cfg, engine.client)
+	if err != nil {
+		logger.Error(fmt.Sprintf("engine: Failed to reload appnet Agent container, error: %v", err))
+		return err
+	}
+	return nil
+}
+
+func (engine *DockerTaskEngine) restartInstanceTask() {
+	if engine.serviceconnectRelay != nil {
+		serviceconnectRelayTask, err := engine.serviceconnectManager.CreateInstanceTask(engine.cfg)
+		if err != nil {
+			logger.Error(fmt.Sprintf("Unable to start relay for task in the engine: %v", err))
+			return
+		}
+		// clean up instance relay task
+		for _, container := range engine.serviceconnectRelay.Containers {
+			if container.Type == apicontainer.ContainerServiceConnectRelay {
+				engine.stopContainer(engine.serviceconnectRelay, container)
+			}
+		}
+		engine.serviceconnectRelay.SetDesiredStatus(apitaskstatus.TaskStopped)
+		engine.sweepTask(engine.serviceconnectRelay)
+		engine.deleteTask(engine.serviceconnectRelay)
+
+		engine.serviceconnectRelay = serviceconnectRelayTask
+		engine.AddTask(engine.serviceconnectRelay)
+		logger.Info("engine: Restarted AppNet Relay task")
+	}
 }
--- a/agent/engine/docker_task_engine_linux_test.go
+++ b/agent/engine/docker_task_engine_linux_test.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -81,7 +82,7 @@ func init() {
 func TestResourceContainerProgression(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.TODO())
 	defer cancel()
-	ctrl, client, mockTime, taskEngine, _, imageManager, _, _ := mocks(t, ctx, &defaultConfig)
+	ctrl, client, mockTime, taskEngine, _, imageManager, _, serviceConnectManager := mocks(t, ctx, &defaultConfig)
 	defer ctrl.Finish()
 
 	sleepTask := testdata.LoadTask("sleep5")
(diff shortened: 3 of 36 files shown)


yinyic (commented):
Have we verified (manually) that this works? If so could you include that in the description? 

Update - thanks for adding the manual test scenarios!

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +45,77 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
...
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+				const writeOrCreateMask = fsnotify.Write | fsnotify.Create
+				if event.Op&writeOrCreateMask != 0 {
+					seelog.Debugf("new fsnotify watcher event: %s", event.Name)
+					// reload the updated Appnet Agent image
+					if err = engine.reloadAppNetImage(); err != nil {
+						return
```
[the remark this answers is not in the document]
mythri-garaga: yep, my bad..I updated:) Thank you!

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +44,78 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
...
+	engine.watch(ctx, watcher)
+}
+
+func (engine *DockerTaskEngine) watch(ctx context.Context, watcher *fsnotify.Watcher) {
+	// Start listening for events.
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				seelog.Warn("fsnotify watcher event error")
```
yinyic, line 67: `!ok` indicates that the channel has been closed, not necessarily an error event.

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +44,78 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
...
+				// reload the updated Appnet Agent image
+				if err := engine.reloadAppNetImage(); err == nil {
+					// restart the internal instance relay task with
+					// updated Appnet Agent image
+					engine.restartInstanceTask()
+				}
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return
```
yinyic, line 82: Maybe add a log here that channel has been closed

agent/engine/docker_task_engine_linux_test.go
```diff
@@ -1364,3 +1366,29 @@ func TestProvisionContainerResourcesBridgeModeWithServiceConnect(t *testing.T) {
...
+	// create a fsnotify watcher
+	watcher, err := fsnotify.NewWatcher()
+	assert.NoError(t, err)
+	err = watcher.Add("/tmp/")
+	assert.NoError(t, err)
+	defer watcher.Close()
+	event := fsnotify.Event{
+		Name: "/tmp/",
+		Op:   fsnotify.Create,
+	}
```
yinyic, line 1385: instead of faking an event, can we try actually creating a file in `/tmp` and check that watcher is able to pick it up?

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +44,78 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
...
+	}
+	err = watcher.Add(filepath)
+	if err != nil {
+		seelog.Errorf("failed to initialize fsnotify NewWatcher, error: %v", err)
+	}
+	defer watcher.Close()
+	engine.watch(ctx, watcher)
+}
+
+func (engine *DockerTaskEngine) watch(ctx context.Context, watcher *fsnotify.Watcher) {
```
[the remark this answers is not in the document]
mythri-garaga: So unit test failed  with `open /tmp/serviceconnect/agent.tar : permission denied` in previous revision when we tried to test it by creating a file in `/tmp`, my understanding is that we need to run this test with sudo permission in order for test to create files with os package(this test passed on local when run with sudo command).

I split the watchAppnetImage into two functions so that we could test the image reload and relay task restart logic by creating fake event and avoid creating any temp files.

agent/engine/docker_task_engine_linux_test.go
mythri-garaga: same reason as mentioned in the previous comment. 

The unit test failed with open /tmp/serviceconnect/agent.tar : permission denied in previous revision when we tried to test it by creating a file in /tmp, my understanding is that we need to run this test with sudo permission in order for test to create files with os package(this test passed on local when run with sudo command).

I split the watchAppnetImage into two functions so that we could test the image reload and relay task restart logic by creating fake event and avoid creating any temp files.

agent/engine/docker_task_engine_linux.go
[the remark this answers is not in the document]
yinyic: We should be able to use `TempDir()` as the base dir, instead of `/tmp` which requires privileged access.
https://pkg.go.dev/testing#T.TempDir

agent/engine/docker_task_engine_linux_test.go
yinyic: Try https://pkg.go.dev/testing#T.TempDir, that should help us test file operations in a directory without needing privileged access.

agent/engine/docker_task_engine_linux.go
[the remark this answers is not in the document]
mythri-garaga: Thank you, `TempDir()` worked:)

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +44,76 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
...
 }
+
+func (engine *DockerTaskEngine) watchAppNetImage(ctx context.Context, filepath string) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		seelog.Errorf("error adding %s to fsnotify watcher, error: %v", filepath, err)
+	}
+	err = watcher.Add(filepath)
+	if err != nil {
+		seelog.Errorf("failed to initialize fsnotify NewWatcher, error: %v", err)
```
yinyic, line 55: Just noticed that the two error messages should probably be swapped :)

mythri-garaga: Thanks for catching that! just updated:)

agent/engine/docker_task_engine_linux.go
```diff
@@ -17,10 +17,15 @@
 package engine
 
 import (
+	"context"
 	"time"
 
+	"github.com/cihub/seelog"
```
ubhattacharjya, line 23: Do you want to use "github.com/aws/amazon-ecs-agent/agent/logger" instead of seelog?

agent/engine/docker_task_engine.go
```diff
@@ -109,6 +109,8 @@ const (
 	stopContainerBackoffJitter     = 0.2
 	stopContainerBackoffMultiplier = 1.3
 	stopContainerMaxRetryCount     = 5
+
+	serviceConnectAppnetAgenTarballDir = "/var/lib/ecs/deps/serviceconnect/"
```
yinyic, line 113: It just occurred to me that the `/var/lib/ecs/deps/` is the host directory, and is mounted into container as `/managed/agents/` (see the bind mount configuration [here](https://github.com/aws/amazon-ecs-agent/blob/f003b0fed6b8e12daec448cdef34b37757c297a1/ecs-init/docker/docker.go#L121-L122))

I thought the watcher should be created on the container path (because the exact path does not exist in Agent container technically), but are we observing expected behaviors even when the watcher is created with the host path?

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +44,76 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
...
+	// Start listening for events.
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				seelog.Warn("fsnotify event watcher channel is closed")
+				return
+			}
+			const writeOrCreateMask = fsnotify.Write | fsnotify.Create
+			if event.Op&writeOrCreateMask != 0 {
```
ubhattacharjya, line 68: should there be a space between event.Op and writeOrCreateMask ? Also, can you explain what this means?

mythri-garaga: added space and a comment.

agent/engine/docker_task_engine.go
mythri-garaga: You are right, watching container path is appropriate here, made changes in the latest rev accordingly.

yinyic (approved):

agent/engine/docker_task_engine_linux.go
ubhattacharjya: Thanks Mythri :)

ubhattacharjya (approved):

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +45,78 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
...
+			logger.Error(fmt.Sprintf("Unable to start relay for task in the engine: %v", err))
+			return
+		}
+		// clean up instance relay task
+		for _, container := range engine.serviceconnectRelay.Containers {
+			if container.Type == apicontainer.ContainerServiceConnectRelay {
+				engine.stopContainer(engine.serviceconnectRelay, container)
+			}
+		}
+		engine.serviceconnectRelay.SetDesiredStatus(apitaskstatus.TaskStopped)
```
chienhanlin, line 114: Q. Would we show the task is stopped/started due to AppNet Agent image reload somewhere in ECS Agent log? or Control plane will receive a message from ECS Agent, saying that task state is changed due to image is reloading?

chienhanlin (approved):

agent/engine/docker_task_engine_linux.go
```diff
@@ -39,3 +45,78 @@ func (engine *DockerTaskEngine) updateTaskENIDependencies(task *apitask.Task) {
 func (engine *DockerTaskEngine) invokePluginsForContainer(task *apitask.Task, container *apicontainer.Container) error {
 	return nil
 }
+
+func (engine *DockerTaskEngine) watchAppNetImage(ctx context.Context) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		logger.Error(fmt.Sprintf("failed to initialize fsnotify NewWatcher, error: %v", err))
```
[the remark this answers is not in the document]
mythri-garaga: oops...will fix it in the next PR!! Thank you:)

agent/engine/docker_task_engine_linux.go
mythri-garaga: Since this an internal task, we don't publish this task status to backend but just log the restart on ECS Agent.

PR #3449, opened 2022-10-28
adding support for container port range [part 1]

singholt:
### Summary

This adds support for container port ranges. The following changes are in this PR:
1. update `acs` and `ecs_client` models/api files with new fields.
2. advertise new capability `ecs.capability.container-port-range` so that only supported agents get task payload with ranges.
3. update v4 task metadata to publish port ranges.

To follow in next PR:
1. Supplying port ranges to docker.
2. Get ports from docker container metadata, consolidate into ranges (since docker returns 1:1 mapping for all ports) and populating `networkBinding` which is sent to ECS as part of `SubmitContainerStateChange` request.

### Implementation details

We are updating two objects:

a. `PortMapping`: This is a container definition parameter to specify ports.
```
"PortMapping":
{
    "containerPort" : Integer,
    "containerPortRange" : String, [NEW]
    "hostPort" : Integer,
    "protocol" : Stringz
}
```
b. `NetworkBinding`: This has details on port bindings between a container and its host container instance, as part of the DescribeTasks API response.
```
"NetworkBinding":
{
    "containerPort" : Integer,
    "containerPortRange" : String, [NEW]
    "hostPort" : Integer,
    "hostPortRange" : String, [NEW]
    "protocol" : Stringz
}
```
**Note**: `NetworkBinding` has `hostPortRange` while `PortMapping` doesn't because we will consolidate host ports that are mapped to the requested `containerPortRange` before sending it to ECS.

### Testing

### Licensing


agent/api/container/port_binding.go
```diff
@@ -30,9 +32,13 @@ const (
 // PortBinding represents a port binding for a container
 type PortBinding struct {
```
singholt, line 33: for each port binding, we know that either one of `containerPort` or `containerPortRange` is set, not both.

hence below, `ContainerPort` is changed from `uint16` -> `*uint16` and `containerPortRange` is also a `*string` because we want to do a nil check on these while populating v4 task metadata response.

agent/api/statechange_test.go
```diff
@@ -21,14 +21,14 @@ import (
 	"testing"
 	"time"
 
-	"github.com/aws/amazon-ecs-agent/agent/api/serviceconnect"
-
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
+	"github.com/aws/amazon-ecs-agent/agent/api/serviceconnect"
```
YashdalfTheGray, line 26: Not sure if this was meant to be here, I think it might have been like a rebase/merge artifact but I'm not sure it is going to impact anything so meh.

agent/handlers/v2/response_test.go
```diff
@@ -207,6 +251,16 @@ func TestTaskResponseWithV4Metadata(t *testing.T) {
 		DockerConfig: apicontainer.DockerConfig{
 			HostConfig: aws.String(`{"LogConfig":{"Type":"awslogs","Config":{"awslogs-group":"myLogGroup"}}}`),
 		},
+		KnownPortBindingsUnsafe: []apicontainer.PortBinding{
+			{
```
YashdalfTheGray, line 255: Question; is it worthwhile to add a test here test checks for the known port bindings to be empty list?

YashdalfTheGray (approved):

agent/api/statechange_test.go
singholt: no, I just corrected the import order, moved it into the `amazon-ecs-agent` import group from above.

agent/handlers/v1/response.go
```diff
@@ -64,10 +66,12 @@ type VolumeResponse struct {
 // PortResponse defines the schema for portmapping response JSON
 // object.
 type PortResponse struct {
```
yinyic, line 68: If this `PortResponse` object is local to TMDE V1 and we don't plan to include port ranges for it, maybe it's better to not update the object at all?

yinyic (approved):

agent/handlers/v1/response.go
singholt: although, the new range fields in `PortResponse` are added in v1 TDME file, they are not really set for any TDME versions except v4.

with this feature, we plan to populate the v4 taskResponse with the requested `containerPortRange` and the corresponding dynamically assigned `hostPortRange`. the v4 code paths use the v1 struct and populate fields based on `includeV4Metadata` boolean. its not ideal, but it is what it is :) 

I hope I answered your question

yinyic: > the v4 code paths use the v1 struct

That's exactly what I was missing. Thanks :)

yinyic (approved):

YashdalfTheGray (dismissed):

yinyic (approved):

YashdalfTheGray (approved):

agent/ecs_client/model/api/api-2.json
```diff
@@ -1645,7 +1645,9 @@
       "members":{
         "bindIP":{"shape":"String"},
         "containerPort":{"shape":"BoxedInteger"},
+        "containerPortRange":{"shape":"String"},
         "hostPort":{"shape":"BoxedInteger"},
+        "hostPortRange":{"shape":"String"},
```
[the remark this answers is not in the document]
singholt: > but there is no guarantee that Docker will map the n container ports to a nice block of host port range.

this is a great point Onur, we discovered this during our testing. we're meeting today to discuss alternatives/path forward. I hope you can make it.

agent/api/container/port_binding.go
```diff
@@ -30,9 +32,13 @@ const (
...
 	// ContainerPort is the port inside the container
-	ContainerPort uint16
+	ContainerPort *uint16
+	// ContainerPortRange is a range of ports exposed inside the container
+	ContainerPortRange *string
 	// HostPort is the port exposed on the host
 	HostPort uint16
+	// HostPortRange is the range of host ports corresponding to ContainerPortRange
+	HostPortRange string
```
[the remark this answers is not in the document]
singholt: `containerPort` and `containerPortRange` are references while `hostPort` is not. This is because we want to perform nil checks on `containerPort` and `containerPortRange` while supplying these to docker and while generating the list of `networkBinding`.

The assumption here is that per port binding config, either `containerPort` or `containerPortRange` will be set, not both. This is validated on the ECS task definition level.

PR #3475, opened 2022-11-09
adding util methods to get ephemeral host port range [part 2 of container-port-range support]

singholt:
### Summary

This PR follows after https://github.com/aws/amazon-ecs-agent/pull/3473
It adds helper util methods to get contiguous host ports from the ephemeral port range.

To follow in next PR, 
Supply docker with port ranges, using the helper method added in this PR.

### Implementation details

The agent would query host property and get the ephemeral port range. It would loop over the ephemeral range and get N contiguous host ports where N = number of container ports in `containerPortRange`.

If it is unable to read host property, the agent would stick to a defined default ephemeral range.

### Testing

### Licensing

### DIFF
--- a/agent/utils/ephemeral_ports.go
+++ b/agent/utils/ephemeral_ports.go
@@ -16,7 +16,12 @@ package utils
 import (
 	"fmt"
 	"math/rand"
+	"net"
+	"strconv"
+	"sync"
 	"time"
+
+	"github.com/cihub/seelog"
 )
 
 // From https://www.kernel.org/doc/html/latest//networking/ip-sysctl.html#ip-variables
@@ -29,6 +34,9 @@ const (
 var (
 	// Injection point for UTs
 	randIntFunc = rand.Intn
+	// portLock is a mutex lock used to prevent two concurrent tasks to get the same host ports.
+	// TODO: implement a port manager that tracks last assigned host port
+	portLock sync.Mutex
 )
 
 // GenerateEphemeralPortNumbers generates a list of n unique port numbers in the 32768-60999 range. The resulting port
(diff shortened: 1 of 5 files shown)


YashdalfTheGray (dismissed):

agent/utils/ephemeral_ports.go
```diff
@@ -58,3 +63,82 @@ func GenerateEphemeralPortNumbers(n int, reserved []uint16) ([]uint16, error) {
...
+	if n != numberOfPorts {
+		return "", fmt.Errorf("%v contiguous host ports unavailable", numberOfPorts)
+	}
+	return strconv.Itoa(startPort) + "-" + strconv.Itoa(lastPort), nil
+}
+
+// getPortRangeLength returns the number of ports in a valid range, inclusive of the start and end ports of the range
+func getPortRangeLength(portRange string) (int, error) {
+	// nat.ParsePortRangeToInt validates a port range; if valid, it returns start and end ints of the range, else returns error
+	startPort, endPort, err := nat.ParsePortRangeToInt(portRange)
```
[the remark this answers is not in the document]
singholt: yes

agent/utils/ephemeral_ports.go
```diff
@@ -58,3 +63,82 @@ func GenerateEphemeralPortNumbers(n int, reserved []uint16) ([]uint16, error) {
...
+		return "", err
+	}
+
+	// get ephemeral port range, either default or if custom-defined
+	startHostPortRange, endHostPortRange, err := dynamicHostPortRange()
+	if err != nil {
+		seelog.Warnf("Unable to read the ephemeral host port range, falling back to the default range: %v-%v",
+			defaultPortRangeStart, defaultPortRangeEnd)
+		startHostPortRange, endHostPortRange = defaultPortRangeStart, defaultPortRangeEnd
+	}
```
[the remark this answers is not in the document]
singholt: good point. I am keeping things consistent with how docker does it - https://github.com/moby/moby/blob/master/libnetwork/portallocator/portallocator.go#L92-L96

agent/utils/ephemeral_ports.go
```diff
@@ -58,3 +63,82 @@ func GenerateEphemeralPortNumbers(n int, reserved []uint16) ([]uint16, error) {
 	}
 	return result, nil
 }
+
+var dynamicHostPortRange = getDynamicHostPortRange
+
+// GetHostPortRange gets a set of contiguous host ports from the ephemeral host port range defined on the host. The
+// number of host ports is determined by the number of container ports, in the given containerPortRange.
+func GetHostPortRange(containerPortRange string, protocol string) (string, error) {
```
yinyic, line 71: Instead of passing in `containerPortRange`, it feels a little more clear to me to simply provide the number of ports needed (length of range):
1. Generalization - this helper method does not need to know what's the container port range (it doesn't even need to know that it's doing it for container port mapping), all it does is to allocate N contiguous host ports. 
2. Redundancy of validation - if we pass in `containerPortRange`, then we need to verify its validity, which, I believe, should be already done during task payload unmarshalling (because we may not be always calling this method, once we support customizable host port range).

agent/utils/ephemeral_ports.go
```diff
@@ -58,3 +63,82 @@ func GenerateEphemeralPortNumbers(n int, reserved []uint16) ([]uint16, error) {
...
+	// get ephemeral port range, either default or if custom-defined
+	startHostPortRange, endHostPortRange, err := dynamicHostPortRange()
+	if err != nil {
+		seelog.Warnf("Unable to read the ephemeral host port range, falling back to the default range: %v-%v",
+			defaultPortRangeStart, defaultPortRangeEnd)
+		startHostPortRange, endHostPortRange = defaultPortRangeStart, defaultPortRangeEnd
+	}
+
+	var startPort, lastPort, n int
+	for port := startHostPortRange; port <= endHostPortRange; port++ {
```
yinyic, line 86: Should we put some kind of synchronization mechanism around this block to avoid race condition when two tasks simultaneously need host port range allocation?

ofiliz: Ideally any synchronization should be happening in the task launch workflow, and not in utility functions such as this. Are there are no other host resource races that the agent has to guard against?

singholt: agreed. Yinyi, Yash and I did discussed this earlier today. The upcoming refactor creates a port mapping manager that handles synchronization and tracks the last port agent assigned to a container. 

to follow soon.

singholt: discussed offline. We're going to add a lock here. Next, I am going to follow-up with a revised design aiming to improve on synchronization and port tracking.

agent/utils/ephemeral_ports.go
singholt: made the change

yinyic (approved):

YashdalfTheGray (approved):
Got a walkthrough of this code from Anuj and tested it with my own stack.

PR #3479, opened 2022-11-10
supply port range to docker and consolidate network bindings [part 3 for container port range support]

singholt:
### Summary

This PR follows after https://github.com/aws/amazon-ecs-agent/pull/3475, and is the final PR in the series.

### Implementation details

1. We supply ports to docker in two places: `DockerHostConfig` and `DockerConfig`. The former consists of host-dependent config requests and the latter is the non-host dependent config. Docker, upon receiving a create-container request, merges both the configs into one. 

   In case of port ranges for a container, the ECS agent would determine N contiguous host ports equal to the `containerPortRange`, and supply to docker. In case agent doesn't find N contiguous ports, docker will do dynamic assignment like it does today.

2. The ECS agent sends `networkBindings` to ECS as part of the `SubmitContainerStateChange` payload. 
    Since, Docker inspect reports only 1:1 `hostPort`:`ContainerPort` mapping, the agent would consolidate all ports that belong to a `containerPortRange` request while generating the networkBindings.

   This does not change behavior for 1:1 port mappings today.

### Testing

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -317,6 +317,13 @@ type Container struct {
 	finishedAt time.Time
 
 	labels map[string]string
+
+	// hasPortRange is set to true when the container has at least 1 port range requested.
+	ContainerHasPortRange bool
+	// ContainerPortSet is a set of singular container ports that don't belong to a containerPortRange request
+	ContainerPortSet map[int]struct{}
+	// ContainerPortRangeMap is a map of containerPortRange to its associated hostPortRange
+	ContainerPortRangeMap map[string]string
 }
 
 type DependsOn struct {
@@ -1359,4 +1366,40 @@ func (c *Container) IsContainerTornDown() bool {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 	return c.ContainerTornDownUnsafe
+}
+
+func (c *Container) SetContainerHasPortRange(containerHasPortRange bool) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.ContainerHasPortRange = containerHasPortRange
+}
+
+func (c *Container) GetContainerHasPortRange() bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.ContainerHasPortRange
+}
+
+func (c *Container) SetContainerPortSet(containerPortSet map[int]struct{}) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.ContainerPortSet = containerPortSet
+}
+
+func (c *Container) GetContainerPortSet() map[int]struct{} {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.ContainerPortSet
+}
+
+func (c *Container) SetContainerPortRangeMap(portRangeMap map[string]string) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.ContainerPortRangeMap = portRangeMap
+}
+
+func (c *Container) GetContainerPortRangeMap() map[string]string {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.ContainerPortRangeMap
 }
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -318,7 +318,7 @@ type Container struct {
 
 	labels map[string]string
 
-	// hasPortRange is set to true when the container has at least 1 port range requested.
+	// ContainerHasPortRange is set to true when the container has at least 1 port range requested.
 	ContainerHasPortRange bool
 	// ContainerPortSet is a set of singular container ports that don't belong to a containerPortRange request
 	ContainerPortSet map[int]struct{}
@@ -1374,7 +1374,7 @@ func (c *Container) SetContainerHasPortRange(containerHasPortRange bool) {
 	c.ContainerHasPortRange = containerHasPortRange
 }
 
-func (c *Container) GetContainerHasPortRange() bool {
+func (c *Container) HasPortRange() bool {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 	return c.ContainerHasPortRange
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -20,8 +20,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/aws/amazon-ecs-agent/agent/logger"
-
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	apierrors "github.com/aws/amazon-ecs-agent/agent/api/errors"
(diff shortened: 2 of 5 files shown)


agent/api/container/container.go
```diff
@@ -1360,3 +1367,39 @@ func (c *Container) IsContainerTornDown() bool {
...
 	return c.ContainerTornDownUnsafe
 }
+
+func (c *Container) SetContainerHasPortRange(containerHasPortRange bool) {
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.ContainerHasPortRange = containerHasPortRange
+}
+
+func (c *Container) GetContainerHasPortRange() bool {
```
YashdalfTheGray, line 1377: Question; I'm wondering if these reads need locks as well. Depending on how often this data changes, we might be able to get away without the reads being locked. Non blocking though.

YashdalfTheGray (approved):

agent/api/container/container.go
yinyic, line 1377: Nit - i think it's more common to name a predicate `IsXXX` or `HasXXX`. In this case `HasPortRange` seems a bit easier to read. Imagine we have container c, and caller would call `c.HasPortRange()` instead of `c.GetContainerHasPortRange()`

singholt: the get methods take a read lock, which multiple goroutines can acquire at the same time. Also, keeping it consistent with other getters in this file

agent/api/ecsclient/client.go
```diff
@@ -494,7 +507,39 @@ func (client *APIECSClient) buildContainerStateChangePayload(change api.Containe
 		statechange.ExitCode = aws.Int64(exitCode)
 	}
 
+	networkBindings := getNetworkBindings(change, shouldExcludeIPv6PortBinding)
+	// we enforce a limit on the no. of network bindings for containers with at-least 1 port range requested.
+	// this limit is enforced by ECS, and we fail early and don't call SubmitContainerStateChange.
+	if change.Container.GetContainerHasPortRange() && len(networkBindings) > ecsMaxNetworkBindingsLength {
```
yinyic, line 513: But this limit technically is effective even there's no port range? If cx specified 100 standalone container ports, there still will be 100 port mappings, and thus the limit should apply?

agent/api/ecsclient/client.go
```diff
@@ -494,7 +507,39 @@ func (client *APIECSClient) buildContainerStateChangePayload(change api.Containe
...
+type ProtocolBindIP struct {
+	protocol string
+	bindIP   string
+}
+
+// getNetworkBindings returns the list of networkingBindings, sent to ECS as part of the container state change payload
+func getNetworkBindings(change api.ContainerStateChange, shouldExcludeIPv6PortBinding bool) []*ecs.NetworkBinding {
 	networkBindings := []*ecs.NetworkBinding{}
+	// we do not return any network bindings for awsvpc and host network mode tasks.
+	if change.Container.GetNetworkMode() == networkModeAWSVPC || change.Container.GetNetworkMode() == networkModeHost {
```
yinyic, line 532: Probably want to exclude "none" mode as well. 
Maybe easier to use `if bridgeMode`, since only bridge mode should have a port mapping

agent/api/ecsclient/client.go
```diff
@@ -494,7 +507,39 @@ func (client *APIECSClient) buildContainerStateChangePayload(change api.Containe
...
+		return networkBindings
+	}
+	// hostPortToProtocolBindIPMap is a map to store protocol and bindIP information associated to host ports
+	// that belong to a range. This is used in case when there are multiple protocol/bindIP combinations associated to a
+	// port binding. example: when both IPv4 and IPv6 bindIPs are populated by docker and shouldExcludeIPv6PortBinding is false.
+	hostPortToProtocolBindIPMap := map[int64][]ProtocolBindIP{}
+
+	containerPortSet := change.Container.GetContainerPortSet()
+	containerPortRangeMap := change.Container.GetContainerPortRangeMap()
```
yinyic, line 541: Can we add a comment to indicate that each entry in the port range map means ECS Agent has been able to find a contiguous block of host port range, and that if we have fall back to docker to assign the range, every port in the range is added to port set individually?

agent/api/ecsclient/client.go
```diff
@@ -494,7 +507,39 @@ func (client *APIECSClient) buildContainerStateChangePayload(change api.Containe
...
+
 	for _, binding := range change.PortBindings {
```
yinyic, line 543: Do we know for sure that each of those `PortBindings` will only container singular container and host port, and not a range? (asking because we seem to have `portBinding.ContainerPortRange` in `task.dockerPortMap` method. Although it looks like those are different structs, just want to make sure...

agent/api/ecsclient/client.go
singholt: the limit is based on a worst case char limit given that customer uses a port range. In case of singular ports, the limit would be higher since the worst case no. of chars used for a single port are much less than that for a range.

example: `containerPort`: 65535 vs `containerPortRange`: `65534-65535`

We do not enforce a limit for singular ports today and with this feature, we don't want to change behavior for the same.

agent/api/ecsclient/client.go
singholt: yes, they're different structs. Docker's port-binding struct here, only has singular ports. While the bindings struct in `task.dockerPortMap` is parsing from the ECS task definition.

yinyic: I see! Overall it's been a bit challenging for me to follow the model translations from task payload -> docker input -> docker output -> STSC payload 😅

agent/api/container/container.go
singholt: made the change

agent/api/ecsclient/client.go
singholt: made the change

agent/api/ecsclient/client.go
singholt: made the change

agent/api/ecsclient/client.go
yinyic: Hmm okay... so this is assuming there are 100 portBindings, all have port ranges in them. I suppose I'd like if we can do something like
```
maxCharSingularPortBinding=A
maxCharRangePortBinding=B
maxCharPortBindingCombined=C  // Control Plane limit
...
if (numOfSingularPortBindings * maxCharSingularPortBinding +numOfRangePortBindings * maxCharRangePortBinding > maxCharPortBindingCombined) 
    // don't send STSC
```
This helps maximize allowed cases, and fail fast whenever possible
But I suppose it's hard to quantify the limit for customer in that case (and how they may avoid hitting it)

yinyic (dismissed):

agent/api/ecsclient/client.go
singholt: > Hmm okay... so this is assuming there are 100 portBindings, all have port ranges in them. I suppose I'd like if we can do something like

this limit assumes that there is atleast 1 range. based on feedback from ECS backend team, we don't want to expose specific ECS char limits and instead have user-friendly limit on network bindings defined.

YashdalfTheGray (approved):

PR #3480, opened 2022-11-10
Remove CAP_CHOWN

Realmonia:
### Summary

Remove CAP_CHOWN from docker capacity since it causes compatibility issue with legacy docker versions.

### Testing

### Licensing

### DIFF
--- a/ecs-init/docker/docker.go
+++ b/ecs-init/docker/docker.go
@@ -69,16 +69,6 @@ const (
 	// maxRetries specifies the maximum number of retries for ping to return
 	// a successful response from the docker socket
 	maxRetries = 5
-	// CapNetAdmin to start agent with NET_ADMIN capability
-	// For more information on capabilities, please read this manpage:
-	// http://man7.org/linux/man-pages/man7/capabilities.7.html
-	CapNetAdmin = "NET_ADMIN"
-	// CapSysAdmin to start agent with SYS_ADMIN capability
-	// This is needed for the ECS Agent to invoke the setns call when
-	// configuring the network namespace of the pause container
-	// For more information on setns, please read this manpage:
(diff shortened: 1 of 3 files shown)


yinyic (commented):
I think we need to also remove the reference

ecs-init/docker/docker.go
```diff
@@ -79,11 +79,6 @@ const (
 	// For more information on setns, please read this manpage:
 	// http://man7.org/linux/man-pages/man2/setns.2.html
 	CapSysAdmin = "SYS_ADMIN"
-	// CapChown to start agent with CAP_CHOWN capability
```
singholt, line 82: in addition this, should we move the list of all capabilities into its own little block and add a comment that a prefix `CAP_` is appended to all the below capabilities by default?

Realmonia:
Changed to a revert commit

ecs-init/docker/docker.go
Realmonia: I changed this to a revert commit. Let's add it when we introduce it again.

fierlion (approved):

ecs-init/docker/docker.go
singholt: this will prevent us from making the same mistake again, not just for `CHOWN` but all capabilities in general.

yinyic (dismissed):

ecs-init/docker/docker.go
Realmonia: Yep I got that. I was just trying to avoid doing more changes and get the fix out asap. I will add that since it's just a comment.

singholt (approved):

yinyic (approved):

Issue #3492, opened 2022-11-29
`nvidia-driver-version` ECS attribute should be a key/value attribute


risdenk:
### Summary
The `nvidia-driver-version` ECS attribute is added as a name only attribute, which doesn't allow advanced filtering with https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cluster-query-language.html The filtering only happens on the attribute value and since the driver version is in the attribute name it is very limiting.

### Description
The `nvidia-driver-version` ECS attribute is added as a name only attribute, which doesn't allow advanced filtering with https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cluster-query-language.html The filtering only happens on the attribute value and since the driver version is in the attribute name it is very limiting.

Example attribute value from ECS agent:
`attribute:ecs.capability.nvidia-driver-version.470.57.02`

This is implemented here: https://github.com/aws/amazon-ecs-agent/blame/master/agent/app/agent_capability_unix.go#L106

This doesn't match other ECS attributes like CNI version which puts the version in the key/value attribute instead of name only. reference: https://github.com/aws/amazon-ecs-agent/blame/master/agent/app/agent_capability_unix.go#L130

This name only attribute behavior exists from the first PR this was introduced in here: https://github.com/aws/amazon-ecs-agent/pull/1676

### Expected Behavior
`attribute:ecs.capability.nvidia-driver-version` = `470.57.02`

### Observed Behavior
`attribute:ecs.capability.nvidia-driver-version.470.57.02`

Realmonia:
Hi @risdenk ,

Thanks for report! I added this to our backlog. I will post an update here when status is updated.

SreeeS:
Hello @risdenk  can you please update the name from `nvidia-driver-version` to `gpu-driver-version` based on the PR https://github.com/aws/amazon-ecs-agent/pull/3493#discussion_r1123732603

risdenk:
no. AWS can make that change.

risdenk:
https://github.com/aws/amazon-ecs-agent/pull/3493#issuecomment-1515361531

SreeeS:
Made this change in a new PR.
https://github.com/aws/amazon-ecs-agent/pull/3836

PR #3493, opened 2022-11-29
GITHUB-3492 `nvidia-driver-version` ECS attribute should be a key/value attribute

risdenk:
### Summary
Implements https://github.com/aws/amazon-ecs-agent/issues/3492

### Implementation details
Adds key/value pair for `nvidia-driver-version` to capabilities while leaving the old name only attribute for backwards compatibility.

### Testing

Not sure how to run the tests after following https://github.com/aws/amazon-ecs-agent/blob/master/CONTRIBUTING.md

```
➜  amazon-ecs-agent git:(GITHUB-3492) make test-in-docker
pattern ./agent/...: directory prefix agent does not contain main module or its selected dependencies
make: *** No rule to make target `test-in-docker'.  Stop.
```

`make test-in-docker` doesn't work. `make test` fails with a bunch of go packages not installed:

```
➜  amazon-ecs-agent git:(GITHUB-3492) make test
pattern ./agent/...: directory prefix agent does not contain main module or its selected dependencies
/opt/homebrew/bin/go test -count=1 -v -cover -race -tags unit -coverprofile cover.out -timeout=60s ./agent/...
agent/agent.go:21:2: cannot find package "github.com/aws/amazon-ecs-agent/agent/app" in any of:
...
and a bunch more packages.
```

`make` by itself seems to work.

* Enhancement - nvidia-driver-version ECS attribute available as a key/value attribute

### Licensing


risdenk:
Not sure how to run the tests after following https://github.com/aws/amazon-ecs-agent/blob/master/CONTRIBUTING.md

```
➜  amazon-ecs-agent git:(GITHUB-3492) make test-in-docker
pattern ./agent/...: directory prefix agent does not contain main module or its selected dependencies
make: *** No rule to make target `test-in-docker'.  Stop.
```

`make test-in-docker` doesn't work. `make test` fails with a bunch of go packages not installed:

```
➜  amazon-ecs-agent git:(GITHUB-3492) make test
pattern ./agent/...: directory prefix agent does not contain main module or its selected dependencies
/opt/homebrew/bin/go test -count=1 -v -cover -race -tags unit -coverprofile cover.out -timeout=60s ./agent/...
agent/agent.go:21:2: cannot find package "github.com/aws/amazon-ecs-agent/agent/app" in any of:
```

`make` by itself seems to work.

agent/app/agent_capability_unix_test.go
```diff
@@ -193,6 +195,10 @@ func TestNvidiaDriverCapabilitiesUnix(t *testing.T) {
 			&ecs.Attribute{Name: aws.String(name)})
 	}
 
+	// Test key/value nvidia driver version capability
+	expectedCapabilities = append(expectedCapabilities,
+		&ecs.Attribute{Name: attributePrefix + capabilityNvidiaDriverVersion, Value: nvidiaDriverVersion})
+
```
risdenk, line 201: The tests failed with this error:

```
2022-12-15T18:31:19.9199045Z ##[error]agent/app/agent_capability_unix_test.go:200:24: cannot use attributePrefix + capabilityNvidiaDriverVersion (untyped string constant "ecs.capability.nvidia-driver-version") as *string value in struct literal
2022-12-15T18:31:19.9200967Z ##[error]agent/app/agent_capability_unix_test.go:200:80: cannot use nvidiaDriverVersion (variable of type string) as type *string in struct literal
```

risdenk: I think this is fixed in c83949ad320983d69aaca6fe5b456cb5e963338d

agent/app/agent_capability.go
```diff
@@ -51,7 +51,8 @@ const (
 	capabilitySecretLogDriverSSM                           = "secrets.ssm.bootstrap.log-driver"
 	capabilitySecretLogDriverASM                           = "secrets.asm.bootstrap.log-driver"
 	capabiltyPIDAndIPCNamespaceSharing                     = "pid-ipc-namespace-sharing"
-	capabilityNvidiaDriverVersionInfix                     = "nvidia-driver-version."
+	capabilityNvidiaDriverVersion                          = "nvidia-driver-version"
```
ubhattacharjya, line 54: Hi @risdenk ,

Thank you for making the contribution. 

Can you name the new attribute 'gpu-driver-version' instead of 'nvidia-driver-version' so that there is no overlap between the two?

ubhattacharjya:
Can you rerun Windows unit tests?

agent/app/agent_capability.go
risdenk: There is no overlap in the keys from a practical perspective. The existing keyname is:

`nvidia-driver-version.470.57.02`

This is the full key name. The new keyname is just `nvidia-driver-version`.

I'm not sure that generic `gpu-driver-version` makes sense? I can make the change, but I don't understand the overlap concern.

risdenk:
> Can you rerun Windows unit tests?

I don't have permissions.

danehlim:
> > Can you rerun Windows unit tests?
> 
> I don't have permissions.

According to ["Re-running workflows and jobs" in Github documentation](https://docs.github.com/en/actions/managing-workflow-runs/re-running-workflows-and-jobs):

> You can re-run a workflow or jobs in a workflow for up to 30 days after the initial run. You cannot re-run jobs in a workflow once its logs have passed their retention limits.

Thus, I don't think we can re-run the failed Windows unit tests job here.

danehlim:
Hi @risdenk, thank you again for contributing! 

To proceed, could you please squash your 2 commits into a single commit as well as rebase and apply newest commits from [`dev` branch](https://github.com/aws/amazon-ecs-agent/tree/dev), and then push to `risdenk:GITHUB-3492`? 

Afterwards, new test jobs should also be triggered (including a new job for the previously mentioned failed Windows unit tests). For reference, I created a [test/dummy pull request](https://github.com/aws/amazon-ecs-agent/pull/3597) with the squashed commits and rebase to verify that the checks should pass after performing the requested actions.

risdenk:
@danehlim done in e08f7ea3c8da6c525c9d3d8feeb5fce68c2271fe

agent/app/agent_capability.go
danehlim: Hi @risdenk, in regards to the overlap concern - our team discussed and we think that users who want to filter with `nvidia-driver-version*` should not get the new key name but the existing key name. So we think it would be best to have the new key name renamed to `gpu-driver-version` instead to avoid confusion between the existing and new key names. 

Furthermore, in the future we may also consider adding GPU types other than NVIDIA. Perhaps we might look into having something like `gpu-driver-type` set to `nvidia` and `gpu-driver-version` set to `470.10.*` for example.

risdenk: @danehlim that's fine but that is a drastically different change than what is proposed here. If AWS wants to make that drastic change then go for it.

danehlim: @risdenk Just to clarify, we are not requesting support of multiple GPU manufacturers to be added to this pull request. We are just requesting a change in naming to avoid the concerns in overlap mentioned previously and to not close the door on supporting multiple GPU manufacturers in the future. If you have any concerns with this, please feel free to let us know.

sparrc:
@risdenk could you update the name from `nvidia-driver-version` to `gpu-driver-version` please? Sorry for the delay reviewing and coming to a decision, with that change we are ready to merge.

SreeeS:
@risdenk Following up on this, Can you please update the name from `nvidia-driver-version` to `gpu-driver-version`

risdenk:
@SreeeS AWS can make that change.

risdenk:
For those who are following along, I'm not making that change from nvidia -> generic since it goes down a completely different rabbit hole.

There is `nvidia` in tons of different files like `TestNvidiaDriverCapabilitiesUnix` and I'm not going to spend a bunch of time going back and forth about naming of the property -> then function names -> then reorganizing all of the `nvidia` related tests to be more generic.

its been said that "o this is only about making the name more generic and nothing else" sure but this has been sitting out here for ~6 months completely broken. It took 2 months to even get the first look a month later to even get a follow up on comments. Then AWS has had multiple people ask to do the work. I am not going to just keep updating this as AWS whims change. 

So nope @SreeeS please stop pinging me to update this PR. AWS can make a new PR with the change or update this one.

SreeeS:
Made this change in a new PR.
https://github.com/aws/amazon-ecs-agent/pull/3836

PR #3494, opened 2022-11-29
bugfix for container port range

singholt:
### Summary

This PR makes 2 bug-fixes:
1. fix ports information logged in agent logs.
2. fix port search for case when 1 container requests multiple container port ranges.

### Implementation details

1. We currently log ports information in agent logs [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/task_manager.go#L599) and [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/statechange.go#L227). With this project, we made the `containerPort` and `containerPortRange` pointer to int/string values. So the logger is now logging the addresses of variables, rather than the values themselves. We have 2 options to fix this: 
 a) Encode the ports into JSON and then print it. [example](https://github.com/singholt/amazon-ecs-agent/blob/4cf0a871f7e118fe043d71eb515a4814a0dc142d/agent/engine/task_manager.go#L600-L601)
 b) Change the references back to values. The reason why these were references in the first place is because we wanted to perform nil checks. Since we know that `containerPort: 0` and `containerPortRange: ""` are not valid (validated on ECS side), we can stick to non-references.
We're doing option (b) in this PR, since its simple and also avoids a lot of type conversions in unit tests.

2. Fixing concurrent host port search when a container requests multiple port mappings, each with a container port range. Currently, we lock the method where we do host port search. Once we identify host ports, we send it to docker which then maps it. A case where the same container requests multiple container port ranges, would get the same host ports picked because we do the search (for both ranges) before sending a create-container call to docker. This PR fixes it by tracking which host port was last picked and locking read/updates to this tracker. Note that the method to get host ports is also concurrent-safe to prevent two task-level go routines get ports at the same time.

### Testing

1. Modified existing unit tests based on the refactor.
2. Tested the changes manually by building an AMI with these changes and running 50 tasks at the same time, each with a container requesting multiple container port ranges.

### Licensing


singholt:
note: the linux unit test failures is not related to this change but an ubuntu-22.04 and cgroup related issue. @prateekchaudhry is looking into it.

agent/utils/ephemeral_ports.go
```diff
@@ -129,6 +158,14 @@ func GetHostPortRange(numberOfPorts int, protocol string) (string, error) {
...
+		if resultEndPort == endHostPortRange {
+			// we've reached the end of the underlying ephemeral range, so circle back to the start
+			tracker.SetLastAssignedHostPort(startHostPortRange - 1)
+		} else {
+			// update tracker to point to the last assigned port
+			tracker.SetLastAssignedHostPort(resultEndPort)
+		}
 	}
-	return strconv.Itoa(startPort) + "-" + strconv.Itoa(lastPort), nil
+	return strconv.Itoa(resultStartPort) + "-" + strconv.Itoa(resultEndPort), nil
```
YashdalfTheGray, line 170: Minor nit; `fmt.Sprintf()` could make this look a little nicer, but it's not a big deal.

YashdalfTheGray (approved):

agent/utils/ephemeral_ports.go
```diff
@@ -67,13 +64,35 @@ func GenerateEphemeralPortNumbers(n int, reserved []uint16) ([]uint16, error) {
...
+	return pt.lastAssignedHostPort
+}
+
 var dynamicHostPortRange = getDynamicHostPortRange
+var tracker safePortTracker
 
 // GetHostPortRange gets N contiguous host ports from the ephemeral host port range defined on the host.
 func GetHostPortRange(numberOfPorts int, protocol string) (string, error) {
-	portLock.Lock()
-	defer portLock.Unlock()
```
amogh09, line 75: Discussed offline. The requirement to synchronize access to this function has not changed so we probably still need this lock.

singholt: added this back. This lock prevents two concurrent task-level go routines from searching for ports at the same time. The port tracker is also being kept atomic in nature, since its a global variable.

also updating the PR description.

agent/utils/ephemeral_ports.go
singholt: thanks, updated.

agent/utils/ephemeral_ports.go
```diff
@@ -129,6 +163,14 @@ func GetHostPortRange(numberOfPorts int, protocol string) (string, error) {
 	}
 	if n != numberOfPorts {
 		return "", fmt.Errorf("%v contiguous host ports unavailable", numberOfPorts)
+	} else {
+		if resultEndPort == endHostPortRange {
+			// we've reached the end of the underlying ephemeral range, so circle back to the start
+			tracker.SetLastAssignedHostPort(startHostPortRange - 1)
```
amogh09, line 169: Discussed offline that `startHostPortRange` is assumed to be constant across calls to this function, so resetting tracker to `startHostPortRange - 1` instead of `0` is fine.

amogh09 (approved):

YashdalfTheGray (dismissed):

YashdalfTheGray (approved):
Reapproving after testing fix rebase.

amogh09 (dismissed):
Approving the rebase.

agent/utils/ephemeral_ports.go
```diff
@@ -82,8 +105,40 @@ func GetHostPortRange(numberOfPorts int, protocol string) (string, error) {
 		startHostPortRange, endHostPortRange = defaultPortRangeStart, defaultPortRangeEnd
 	}
 
-	var startPort, lastPort, n int
-	for port := startHostPortRange; port <= endHostPortRange; port++ {
+	start := startHostPortRange
+	end := endHostPortRange
+
+	// get the last assigned host port
+	lastAssignedHostPort := tracker.GetLastAssignedHostPort()
```
amogh09, line 112: Let's add tests for the new logic please.

singholt: added

amogh09 (approved):

amogh09 (approved):
Approved. As discussed offline, the loop-back may result in Agent finding overlapping ranges for a single container in some cases (which would result in container failing to start) but that's a risk we are willing to take for the potential benefit of finding contiguous port ranges.

YashdalfTheGray (approved):

PR #3497, opened 2022-11-29
Make version generation consistent across build targets

fierlion:
### Summary
Update each of the build processes to use the build-time version rather than the existing version.  Also removing the `*` in our path, but not from the human-readable string.

### Implementation details
Update `build-integrated` to use the buildroot for version file update; update each make target/script/spec file to use the `version_gen=true` path in build-integrated.

also note that this reverts a workaround in our builds https://github.com/aws/amazon-ecs-agent/commit/6e8c9a5167a39c73c47ac352e4c6fe5ec85160d8

### Testing
Tested on AL2 instance
Added a fresh commit then built for each of the following targets to be sure I was picking up the latest:
```
make dockerfree-agent-image
make generic-rpm-integrated
make amazon-linux-rpm-integrated
```
also checked that `make generic-deb-integrated` works, as it calls `make dockerfree-agent-image` to build the agent.
And also note that I tested the koji-built agent and it is consistent already so no work required there:
https://github.com/aws/amazon-ecs-agent/releases/tag/v1.66.2 <- 06008fa
in the koji-built agent:
```
"Version": "Amazon ECS Agent - v1.66.2 (06008fa1)"
```

Fix agent short hash version bug

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ static:
 	./scripts/build
 
 static-with-pause:
-	./scripts/build true "" false true
+	./scripts/build true "" true true
 
 # Cross-platform build target for static checks
 xplatform-build:
@@ -381,7 +381,7 @@ amazon-linux-sources.tgz:
 
 .amazon-linux-rpm-integrated-done: amazon-linux-sources.tgz
 	test -e SOURCES || ln -s . SOURCES
-	rpmbuild --define "%_topdir $(PWD)" -bb ecs-init.spec
+	rpmbuild --define "%_topdir $(PWD)" -bb ecs-agent.spec
 	find RPMS/ -type f -exec cp {} . \;
 	touch .amazon-linux-rpm-integrated-done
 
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -381,7 +381,7 @@ amazon-linux-sources.tgz:
 
 .amazon-linux-rpm-integrated-done: amazon-linux-sources.tgz
 	test -e SOURCES || ln -s . SOURCES
-	rpmbuild --define "%_topdir $(PWD)" -bb ecs-init.spec
+	rpmbuild --define "%_topdir $(PWD)" -bb ecs-agent.spec
```
sparrc, line 384: how was this working before if ecs-init.spec doesn't exist?

fierlion: actually that's a good point -- this was working because koji doesn't actually use this makefile so this target is more a reference to approximate what koji builds rather than the mechanism koji uses to build.

YashdalfTheGray (approved):

prateekchaudhry (approved):

singholt (approved):

PR #3501, opened 2022-11-30
Use Ubuntu 20.04 for linux GH Unit tests

prateekchaudhry:
### Summary

Github actions moved ```ubuntu-latest``` to use Ubuntu 22.04 by default which uses cgroups v2 by default. ECS Agent has some unit tests incompatible when running on cgroups v2 support. This PR temporarily fixes this by enforcing GH Actions to use ```ubuntu-20.04``` which is still supported.

Use Ubuntu 20.04 for linux GH Unit tests

### Licensing

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -6,7 +6,7 @@ permissions: read-all
 jobs:
   unit-tests:
     name: Linux unit tests
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


fierlion (commented):
This is a great workaround while we get latest working with v2

fierlion (approved):

sparrc (approved):

PR #3505, opened 2022-12-02
bugfix: send network bindings for default network mode

singholt:
### Summary

bugfix: send network bindings for default network mode

### Implementation details

This PR removes the network mode check during network bindings generation. We need network bindings when the network mode is "default" as well.

Simply removing this check because non-bridge/non-nat/non-default modes, network bindings is not generated anyways. Also, we dont have this check for singular port mappings, so keeping things consistent.

verification:

`networkBinding` for default and bridge mode:
```
"networkBindings": [
                        {
                            "bindIP": "0.0.0.0",
                            "protocol": "tcp",
                            "containerPortRange": "8080-8085",
                            "hostPortRange": "32780-32785"
                        }
                    ],
```

for others:
```
 "networkBindings": [],
```

### Testing

New tests cover the changes: 
Manually tested on a test instance that now sends network bindings for both default and bridge network mode. and that it doesn't for other network modes (like it doesn't today).

### Licensing

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -50,8 +50,6 @@ const (
 	osTypeAttrName              = "ecs.os-type"
 	osFamilyAttrName            = "ecs.os-family"
 	RoundtripTimeout            = 5 * time.Second
-	// networkModeBridge specifies the bridge network mode.
-	networkModeBridge = "bridge"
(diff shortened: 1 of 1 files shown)


mythri-garaga (approved):

YashdalfTheGray (approved):
Good catch!

PR #3508, opened 2022-12-07
Read git hash from RELEASE_COMMIT file if possible

danehlim:
### Summary

Read git hash from RELEASE_COMMIT file if possible

### Implementation details

- Prior to executing `gitHash()` function in `version-gen.go`, try to read hash from `RELEASE_COMMIT` file instead using new `releaseCommitGitHash()` function
- Log relevant information in `gitHash()` and `releaseCommitGitHash()` functions in `version-gen.go`
- Add `RELEASE_COMMIT` file to `amazon-linux-sources.tgz` in `Makefile`
- Use `os.ReadFile` instead of `ioutil.ReadFile` in `version-gen.go` since ioutil package is deprecated as of Go 1.16

### Testing

In the repo root, create a test `RELEASE_COMMIT` file, run version-gen.go, and verify that the correct hash is obtained.
```
$ echo "1a2b3c4d5e6f78901a2b3c4d5e6f78901a2b" > RELEASE_COMMIT
$ cd agent/version 
$ GO111MODULE=off go run gen/version-gen.go
2022/12/07 10:31:04 Successfully retrieved full hash value from RELEASE_COMMIT file, fullHashStr: 1a2b3c4d5e6f78901a2b3c4d5e6f78901a2b
$ cat version.go
// This is an autogenerated file and should not be edited.

// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
//
// Licensed under the Apache License, Version 2.0 (the "License"). You may
// not use this file except in compliance with the License. A copy of the
// License is located at
//
//      http://aws.amazon.com/apache2.0/
//
// or in the "license" file accompanying this file. This file is distributed
// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either

// Package version contains constants to indicate the current version of the
// agent. It is autogenerated
package version

// Please DO NOT commit any changes to this file (specifically the hash) except
// for those created by running ./scripts/update-version at the root of the
// repository. Only the 'Version' const should change in checked-in source code

// Version is the version of the Agent
const Version = "1.67.0"

// GitDirty indicates the cleanliness of the git repo when this agent was built
const GitDirty = true

// GitShortHash is the short hash of this agent build
const GitShortHash = "1a2b3c4d"
```

Read git hash from RELEASE_COMMIT file if possible

### Licensing

### DIFF
--- a/agent/version/gen/version-gen.go
+++ b/agent/version/gen/version-gen.go
@@ -14,7 +14,7 @@
 package main
 
 import (
-	"io/ioutil"
+	"fmt"
 	"log"
 	"os"
 	"os/exec"
@@ -72,9 +72,36 @@ func gitHash() string {
 	cmd := exec.Command("git", "rev-parse", "--short=8", "HEAD")
 	hash, err := cmd.Output()
 	if err != nil {
+		log.Printf("failed to execute git rev-parse --short=8 HEAD, err: %v", err)
 		return "UNKNOWN"
 	}
-	return strings.TrimSpace(string(hash))
+	hashStr := strings.TrimSpace(string(hash))
+	log.Printf("Successfully retrieved short hash value from git rev-parse --short=8 HEAD, hashStr: %s",
+		hashStr)
+	return hashStr
+}
+
+func releaseCommitGitHash() (string, error) {
+	fullHash, err := os.ReadFile(filepath.Join("..", "..", "RELEASE_COMMIT"))
+	if err != nil {
+		return "", fmt.Errorf("unable to read RELEASE_COMMIT file, err: %v", err)
+	}
+	fullHashStr := strings.TrimSpace(string(fullHash))
+	if fullHashStr == "" || len(fullHashStr) < 8 {
+		log.Fatalf("Invalid hash value obtained from RELEASE_COMMIT file (empty or length <8), fullHashStr: %s",
+			fullHashStr)
+	}
+	log.Printf("Successfully retrieved full hash value from RELEASE_COMMIT file, fullHashStr: %s", fullHashStr)
+	// return short hash (first 8 characters) instead of full hash
+	return fullHashStr[0:8], nil
+}
+
+func selectGitHash() string {
+	hash, err := releaseCommitGitHash()
+	if err != nil {
+		return gitHash()
+	}
+	return hash
 }
 
 // version-gen is a simple program that generates the agent's version file,
@@ -82,7 +109,7 @@ func gitHash() string {
 // cleanliness.
 func main() {
 
-	versionStr, _ := ioutil.ReadFile(filepath.Join("..", "..", "VERSION"))
+	versionStr, _ := os.ReadFile(filepath.Join("..", "..", "VERSION"))
 
 	// default values
 	info := versionInfo{
@@ -101,7 +128,7 @@ func main() {
 		// have a commit hash so that it does not churn with every commit. This
 		// env var should not be set when building, and go generate should be
 		// run before any build, such that the commithash will be set correctly.
-		info.Hash = gitHash()
+		info.Hash = selectGitHash()
 	}
 
 	outFile, err := os.Create("version.go")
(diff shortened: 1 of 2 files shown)


sparrc:
When and how is the RELEASE_COMMIT file generated?

agent/version/gen/version-gen.go
```diff
@@ -69,20 +68,37 @@ func gitDirty() bool {
...
+			return "UNKNOWN"
+		}
+		hashStr := strings.TrimSpace(string(revParseHash))
+		log.Printf("Successfully retrieved short hash value from git rev-parse --short=8 HEAD, hashStr: %s",
+			revParseHash)
+		return hashStr
+	}
+	hashStr := strings.TrimSpace(string(readFileHash))
+	if hashStr == "" || len(hashStr) < 8 {
+		log.Printf("Invalid hash value obtained from RELEASE_COMMIT file (empty or has length <8), hashStr: %s",
```
sparrc, line 87: if the hash is invalid, seems like we should fallback to `git rev-parse` here

agent/version/gen/version-gen.go
```diff
@@ -69,20 +68,37 @@ func gitDirty() bool {
 }
 
 func gitHash() string {
```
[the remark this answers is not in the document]
sparrc: just add a new function to try to get it from RELEASE_COMMIT, return an error if it doesn't find the file or finds an invalid hash, then fallback to the original gitHash() function

agent/version/gen/version-gen.go
```diff
@@ -69,20 +68,37 @@ func gitDirty() bool {
...
+		cmd := exec.Command("git", "rev-parse", "--short=8", "HEAD")
+		revParseHash, revParseErr := cmd.Output()
+		if revParseErr != nil {
+			log.Printf("Failed to execute git rev-parse --short=8 HEAD, revParseErr: %v", revParseErr)
+			return "UNKNOWN"
+		}
+		hashStr := strings.TrimSpace(string(revParseHash))
+		log.Printf("Successfully retrieved short hash value from git rev-parse --short=8 HEAD, hashStr: %s",
+			revParseHash)
+		return hashStr
```
singholt, line 83: this can be abstracted out into a separate method.

danehlim: Agreed, thanks for pointing out!

agent/version/gen/version-gen.go
danehlim: > so we can differentiate when our expected hashing mechanism fails

+1 to this

> Would it just be better to error out here and fail loudly?

I like this idea -  if our expected hashing mechanism fails with an invalid value, something seriously wrong is occurring and I think it might be a good idea to exit before proceeding with anything else.

YashdalfTheGray (approved):
Looks good!

sparrc (approved):

agent/version/gen/version-gen.go
```diff
@@ -72,17 +72,44 @@ func gitHash() string {
...
+}
+
+func releaseCommitGitHash() (string, error) {
+	fullHash, err := os.ReadFile(filepath.Join("..", "..", "RELEASE_COMMIT"))
+	if err != nil {
+		return "", fmt.Errorf("unable to read RELEASE_COMMIT file, err: %v", err)
+	}
+	fullHashStr := strings.TrimSpace(string(fullHash))
+	if fullHashStr == "" || len(fullHashStr) < 8 {
+		log.Fatalf("Invalid hash value obtained from RELEASE_COMMIT file (empty or length <8), fullHashStr: %s",
```
singholt, line 91: nit: don't need `(empty or length <8)` since you're printing the hash anyways

singholt: also, if you use log.Fatalf, that will do an os.Exit call, which would immediately exit the program and not fall back to reading hash using `gitHash()`: https://pkg.go.dev/log#Fatal

so may want to use Printf?

singholt: resolved offline. exiting here is intentional and helps us fail loudly without falling back when the RELEASE_COMMIT file has invalid content.

PR #3509, opened 2022-12-07
Prevent resetting valid agent state db when IMDS fails on startup

sparrc:
### Summary

This is a followup to https://github.com/aws/amazon-ecs-agent/pull/2861

In the rare occasion that IMDS fails multiple times (also bumping this multiple from 3 to 5), then we should try to avoid resetting the agent state db if we have an existing agent state.

This is to prevent a situation like this from occurring, since resetting the state db will result in a duplicate container instance on the same ec2 instance id, and will also result in some tasks being dropped by the agent.

```
level=warn time=2022-11-23T15:43:00Z msg="Unable to access EC2 Metadata service to determine EC2 ID: RequestError: send request failed\ncaused by: Get \"http://169.254.169.254/latest/meta-data/instance-id\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)" module=agent.go
level=warn time=2022-11-23T15:45:47Z msg="Data mismatch; saved InstanceID 'i-030e42d5588188300' does not match current InstanceID ''. Overwriting old datafile" module=agent.go
```

### Implementation details

`currentEC2InstanceID` will fallback to the instance ID saved in agent's state db if IMDS fails to return an ec2 instance id.

### Testing

existing functional tests

Bugfix: Prevent resetting valid agent state db when IMDS fails on startup

### Licensing

### DIFF
--- a/agent/tcs/client/client_test.go
+++ b/agent/tcs/client/client_test.go
@@ -602,7 +602,10 @@ func TestMetricsDisabled(t *testing.T) {
 		published <- struct{}{}
 	}).Return(nil).MinTimes(1)
 
-	go cs.Serve()
+	go func() {
+		err := cs.Serve()
+		assert.NoError(t, err)
+	}()
 	<-published
 	<-readed
 }
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -87,7 +87,7 @@ const (
 	instanceIdBackoffMax      = time.Second * 5
 	instanceIdBackoffJitter   = 0.2
 	instanceIdBackoffMultiple = 1.3
-	instanceIdMaxRetryCount   = 3
+	instanceIdMaxRetryCount   = 5
 
 	targetLifecycleBackoffMin      = time.Second
 	targetLifecycleBackoffMax      = time.Second * 5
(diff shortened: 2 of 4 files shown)


agent/tcs/client/client_test.go
```diff
@@ -602,7 +602,10 @@ func TestMetricsDisabled(t *testing.T) {
 		published <- struct{}{}
 	}).Return(nil).MinTimes(1)
 
-	go cs.Serve()
+	go func() {
```
mythri-garaga, line 605: This change part of the PR?

sparrc: Yes sorry I meant to comment on this before sending for review.

This test randomly failed on me once and I was trying to debug, but I couldn't figure anything out without knowing what happened to this `Serve()` function call.

Afterwards I couldnt reproduce. If it's just very slightly flakey, at least in the future we can see if this function failed so I thought we should keep it in, since it's harmless to check this in a unit test anyways.

Realmonia: +1

Realmonia: +1

yinyic (approved):

mythri-garaga (approved):

PR #3519, opened 2022-12-13
Do not Review -Test

SreeeS:
### Testing

New tests cover the changes: 

### Licensing


agent/config/config.go
```diff
@@ -526,6 +526,7 @@ func environmentConfig() (Config, error) {
 		err = apierrors.NewMultiError(errs...)
 	}
 	return Config{
+		PortRange:                           os.Getenv("ECS_PortRange"),
```
singholt, line 529: the env agreed upon is `ECS_DYNAMIC_HOST_PORT_RANGE`

so,
```
DynamicHostPortRange: os.Getenv("ECS_DYNAMIC_HOST_PORT_RANGE")
```

agent/api/task/task.go
```diff
@@ -1860,7 +1860,7 @@ func (task *Task) dockerHostConfig(container *apicontainer.Container, dockerCont
 	if err != nil {
 		return nil, &apierrors.HostConfigError{Msg: err.Error()}
 	}
-	dockerPortMap, err := task.dockerPortMap(container)
+	dockerPortMap, err := task.dockerPortMap(container, cfg)
```
singholt, line 1863: can we pass only the range here? like `cfg.DynamicHostPortRange`

agent/api/task/task.go
```diff
@@ -2334,7 +2334,7 @@ func (task *Task) dockerLinks(container *apicontainer.Container, dockerContainer
 
 var getHostPortRange = utils.GetHostPortRange
 
-func (task *Task) dockerPortMap(container *apicontainer.Container) (nat.PortMap, error) {
+func (task *Task) dockerPortMap(container *apicontainer.Container, cfg *config.Config) (nat.PortMap, error) {
```
singholt, line 2337: same comment as above, passing the specific dynamic range should suffice.

agent/api/task/task.go
```diff
@@ -2402,7 +2402,7 @@ func (task *Task) dockerPortMap(container *apicontainer.Container) (nat.PortMap,
 			// we will try to get a contiguous set of host ports from the ephemeral host port range.
 			// this is to ensure that docker maps host ports in a contiguous manner, and
 			// we are guaranteed to have the entire hostPortRange in a single network binding while sending this info to ECS.
-			hostPortRange, err := getHostPortRange(numberOfPorts, protocol)
+			hostPortRange, err := getHostPortRange(numberOfPorts, protocol, cfg.PortRange)
```
singholt, line 2405: this would change to `cfg.DynamicHostPortRange` based on comments above. applicable to all places.

agent/config/config_test.go
```diff
@@ -37,12 +37,14 @@ func TestMerge(t *testing.T) {
 	conf1 := &Config{Cluster: "Foo"}
 	conf2 := Config{Cluster: "ignored", APIEndpoint: "Bar"}
 	conf3 := Config{AWSRegion: "us-west-2"}
+	conf4 := Config{PortRange: "100-200"}
```
singholt, line 40: you don't need to modify this unit test since you are not changing the merge method

agent/config/types.go
```diff
@@ -29,10 +29,17 @@ type ImagePullBehaviorType int8
 type ContainerInstancePropagateTagsFromType int8
 
 type Config struct {
+
+	// PortRange Variable enables the clients to set their custom port range
+	// when there is no port range provided, this will default to the
+	// platform specific values
+	PortRange string `trim:"true"`
```
singholt, line 36: `ECS_DYNAMIC_HOST_PORT_RANGE`

agent/utils/ephemeral_ports.go
```diff
@@ -89,20 +90,26 @@ func (pt *safePortTracker) GetLastAssignedHostPort() int {
...
 
 // GetHostPortRange gets N contiguous host ports from the ephemeral host port range defined on the host.
-func GetHostPortRange(numberOfPorts int, protocol string) (string, error) {
+func GetHostPortRange(numberOfPorts int, protocol string, portRange string) (string, error) {
 	portLock.Lock()
 	defer portLock.Unlock()
 
 	// get ephemeral port range, either default or if custom-defined
-	startHostPortRange, endHostPortRange, err := dynamicHostPortRange()
+	startHostPortRange, endHostPortRange, err := nat.ParsePortRangeToInt(portRange) //dynamicHostPortRange()
```
singholt, line 102: we need to validate this when we parse the config and load it, not here. 

so in the **load config workflow**, we would – 
check if `ECS_DYNAMIC_HOST_PORT_RANGE` is set. If it is set, validate using this `nat.ParsePortRangeToInt()`. if it is valid, set the value of `cfg.DynamicHostPortRange` as provided, else default to whatever`utils.GetDynamicHostPortRange` method gives you.


all of this should be in `config/parse.go`

agent/utils/ephemeral_ports.go
```diff
@@ -89,20 +90,26 @@ func (pt *safePortTracker) GetLastAssignedHostPort() int {
...
-		startHostPortRange, endHostPortRange = defaultPortRangeStart, defaultPortRangeEnd
+		seelog.Warnf("Unable to read the user defined ephemeral host port range, " +
+			"falling back to the ephemeral range")
+		startHostPortRange, endHostPortRange, err = dynamicHostPortRange()
+		if err != nil {
+			seelog.Warnf("Unable to read the ephemeral host port range, "+
+				"falling back to the default range: %v-%v",
+				defaultPortRangeStart, defaultPortRangeEnd)
+			startHostPortRange, endHostPortRange = defaultPortRangeStart, defaultPortRangeEnd
+		}
```
singholt, line 112: based on the comment above, all of this would move to config/ folder.

agent/utils/ephemeral_ports.go
```diff
@@ -89,20 +90,26 @@ func (pt *safePortTracker) GetLastAssignedHostPort() int {
 	return pt.lastAssignedHostPort
 }
 
-var dynamicHostPortRange = getDynamicHostPortRange
+var dynamicHostPortRange = GetDynamicHostPortRange
 var tracker safePortTracker
 
 // GetHostPortRange gets N contiguous host ports from the ephemeral host port range defined on the host.
-func GetHostPortRange(numberOfPorts int, protocol string) (string, error) {
+func GetHostPortRange(numberOfPorts int, protocol string, portRange string) (string, error) {
```
singholt, line 97: this can be `dynamicHostPortRange string`

agent/config/config.go
singholt: you would also need to update the README.

agent/config/parse.go
```diff
@@ -34,6 +36,20 @@ const (
 	envSkipDomainJoinCheck = "ZZZ_SKIP_DOMAIN_JOIN_CHECK_NOT_SUPPORTED_IN_PRODUCTION"
 )
 
+var getDynamicHostPortRange = utils.GetDynamicHostPortRange
+
+func parseDynamicHostPortRange(envVarName string) string {
```
singholt, line 41: nit: `dynamicHostPortRangeEnv` seems more suited than `envVarName`

agent/config/parse.go
```diff
@@ -34,6 +36,20 @@ const (
...
+	dynamicHostPortRange := os.Getenv(envVarName)
+	_, _, err := nat.ParsePortRangeToInt(dynamicHostPortRange)
+	if err != nil {
+		seelog.Warnf("Unable to read the dynamicHostPortRange, " +
```
singholt, line 45: falling back to the **default** ephemeral range

agent/config/parse.go
```diff
@@ -34,6 +36,20 @@ const (
...
+			"falling back to the ephemeral range")
+		startHostPortRange, endHostPortRange, _ := getDynamicHostPortRange()
```
singholt, line 47: i think you would need to check if this method returns an err here. if it does, then we would need to set the start and end to `defaultPortRangeStart` and `defaultPortRangeEnd`.

agent/utils/ephemeral_ports.go
```diff
@@ -89,22 +90,20 @@ func (pt *safePortTracker) GetLastAssignedHostPort() int {
...
 // GetHostPortRange gets N contiguous host ports from the ephemeral host port range defined on the host.
-func GetHostPortRange(numberOfPorts int, protocol string) (string, error) {
+func GetHostPortRange(numberOfPorts int, protocol string, dynamicHostPortRange string) (string, error) {
 	portLock.Lock()
 	defer portLock.Unlock()
 
 	// get ephemeral port range, either default or if custom-defined
-	startHostPortRange, endHostPortRange, err := dynamicHostPortRange()
+	startHostPortRange, endHostPortRange, err := nat.ParsePortRangeToInt(dynamicHostPortRange) //dynamicHostPortRange()
 	if err != nil {
```
singholt, line 102: this block won't be needed because you've already validated the cfg option before sending it here.

agent/config/types.go
```diff
@@ -29,10 +29,17 @@ type ImagePullBehaviorType int8
 type ContainerInstancePropagateTagsFromType int8
 
 type Config struct {
+
+	// PortRange Variable enables the clients to set their custom port range
+	// when there is no port range provided, this will default to the
+	// platform specific values
+	DynamicHostPortRange string
```
singholt, line 36: please move this to the end of this struct, not the start. also update the comment

Issue #3520, opened 2022-12-14
ecs-agent deletes ecs-service-connect-agent:interface-v1 when ECS_ENABLE_UNTRACKED_IMAGE_CLEANUP=true


andrewradamis-paay:
### Summary

If `ECS_ENABLED_UNTRACKED_IMAGE_CLEANUP` is set to `true`, the ecs-agent will remove the `ecs-service-connect-agent:interface-v1` image and tasks enabled for Service Connect will fail to start.

### Description

If `ECS_ENABLED_UNTRACKED_IMAGE_CLEANUP` is set to `true`, the ecs-agent will remove the `ecs-service-connect-agent:interface-v1` image and tasks enabled for Service Connect will fail to start.
Adding `ECS_EXCLUDE_UNTRACKED_IMAGE=ecs-service-connect-agent:interface-v1` to `ecs.config` is a work-around for the issue.
I'm not familar with the ecs-agent codebase, but I'm guessing it just needs to be added to https://github.com/aws/amazon-ecs-agent/blob/41e8e6a78391f65b17070ffe086876e63387d5bb/agent/engine/docker_image_manager.go#L102-L106.

### Expected Behavior
Don't delete `ecs-service-connect-agent:interface-v1`

### Observed Behavior
ecs-agent deleted `ecs-service-connect-agent:interface-v1` then failed to start a Service Connect Task with error `CannotCreateContainerError: Error response from daemon: No such image: ecs-service-connect-agent:interface-v1`.

### Environment Details
Using `ami-028cbf4835af1ad6f (amzn2-ami-ecs-hvm-2.0.20221207-x86_64-ebs)`

ubhattacharjya:
Hi @andrewradamis-paay,

Thank you for reporting the issue. We will provide an update on this soon.

Utsa

mythri-garaga:
Fixed in [1.68.0](https://github.com/aws/amazon-ecs-agent/pull/3528) release.

PR #3521, opened 2022-12-15
Add ServiceConnect image to clean-up exclusion list

yinyic:
### Summary
ECS Agent periodically cleans up cached docker container images that are not being referenced. There are a few images on the exception list including pause container. We should add Service Connect Agent (aka AppNet Agent) image to this list as well because all SC tasks require cached SC Agent image.

https://github.com/aws/amazon-ecs-agent/issues/3520

### Implementation details

Added a new function `AddImageToCleanUpExclusionList` to ImageManager, and when ECS Agent bootstrap, we
1. let Service Connect manager load AppNet Agent image (this happens right before instance capability registration), 
2. call the above new function for AppNet image such that it's excluded from clean-up. 

Step 1 has to happen first since the image name might be different depending on the AppNet Agent version installed.

### Testing

Logs with update ECS Agent:
```
...
level=info time=2022-12-15T17:52:13Z msg="Image excluded from cleanup" image="ecs-service-connect-agent:interface-v1"
...

```

Add ServiceConnect image to clean-up exclusion list

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -395,7 +395,11 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		}
 		return exitcodes.ExitTerminal
 	}
-	agent.serviceconnectManager.SetECSClient(client, agent.containerInstanceARN)
+	scManager := agent.serviceconnectManager
+	scManager.SetECSClient(client, agent.containerInstanceARN)
+	if loaded, _ := scManager.IsLoaded(agent.dockerClient); loaded {
+		imageManager.AddImageToCleanUpExclusionList(agent.serviceconnectManager.GetLoadedImageName())
+	}
 
 	// Add container instance ARN to metadata manager
 	if agent.cfg.ContainerMetadataEnabled.Enabled() {
--- a/agent/stats/service_connect_linux_test.go
+++ b/agent/stats/service_connect_linux_test.go
@@ -121,31 +121,33 @@ func TestRetrieveServiceConnectMetrics(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		// Set up a mock http sever on the statsUrlpath
-		mockUDSPath := "/tmp/appnet_admin.sock"
-		r := mux.NewRouter()
-		r.HandleFunc("/get/them/stats", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			fmt.Fprintf(w, "%v", test.stats)
-		}))
+		func() {
+			// Set up a mock http sever on the statsUrlpath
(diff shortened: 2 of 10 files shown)


agent/stats/service_connect_linux_test.go
```diff
@@ -121,31 +121,33 @@ func TestRetrieveServiceConnectMetrics(t *testing.T) {
...
+			// Set up a mock http sever on the statsUrlpath
+			mockUDSPath := "/tmp/appnet_admin.sock"
+			r := mux.NewRouter()
+			r.HandleFunc("/get/them/stats", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				fmt.Fprintf(w, "%v", test.stats)
+			}))
 
-		ts := httptest.NewUnstartedServer(r)
+			ts := httptest.NewUnstartedServer(r)
+			defer ts.Close()
```
yinyic, line 133: I received this error when running unit test `listen unix /tmp/appnet_admin.sock: bind: address already in use`
Moving `ts.Close()` up in a defer statement should help us clean up always.

mythri-garaga (approved):
Thank you for fixing this!!

agent/app/agent.go
```diff
@@ -395,7 +395,11 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		}
 		return exitcodes.ExitTerminal
 	}
-	agent.serviceconnectManager.SetECSClient(client, agent.containerInstanceARN)
+	scManager := agent.serviceconnectManager
+	scManager.SetECSClient(client, agent.containerInstanceARN)
+	if loaded, _ := scManager.IsLoaded(agent.dockerClient); loaded {
+		imageManager.AddImageToCleanUpExclusionList(agent.serviceconnectManager.GetLoadedImageName())
```
ubhattacharjya, line 401: Could we add the appnet agent image to agent.cfg and then exclude it like other images from the method buildImageCleanupExclusionList ?

yinyic: There are some dependency issues here
- `buildImageCleanupExclusionList` is called when ImageManager gets initiated, before `doStart()`
- SC container image is loaded when we make RegisterContainerInstance call (when we collect capabilities), during `doStart()`

We may be able to move things around in the bootstrap, but we'd need to make sure doing that does not have side effects, at which point I feel like adding a handler for excluding new images would be easier.

Again, the other images are all "known" and have fixed names and tags, which is not the case for SC container image.

mythri-garaga: Agreed, with this approach we don't to manually keep the ImageCleanupExclusionList up to date whenever there is a new appnet agent interface version.

ubhattacharjya (approved):

PR #3529, opened 2023-01-10
Unify package build recipes and remove legacy init build targets

prateekchaudhry:
### Summary

As part of new ECS Agent/Init build process, new init packaging and Makefile targets were introduced a while back (amazon-linux-ami-integrated, generic-deb-integrated, generic-rpm-integrated). While all these targets build init rpm/deb package, the build recipe to create them vary slightly. This PR unifies the packaging process by using the same Makefile target for all packages.

### Implementation details

Code refactoring and other implementation details to achieve it - 
* Add ```GO_VERSION``` to ```generic-rpm-integrated``` packaging spec file. This does not require installing golang manually before running ```make generic-rpm-integrated``` and lets the script handle everything
* Add ```Makefile``` to ```generic-rpm-integrated``` and ```amazon-linux-ami-integrated``` to enable 'unification' of build process by letting all packages use Makefile to build agent image and use the same make target
* Rename ```dockerfree-agent-image``` to ```release-agent```
* Add ```release-agent-internal``` build target. ```release-agent``` builds agent image using this target along with pulling cni sources, while ```release-agent-internal``` builds assuming cni sources have been pulled. ```release-agent``` is used for default generic Agent build while ```release-agent-internal``` (which is ultimately used by ```release-agent as well) is used for internal build process
* Remove ```scripts/build-integrated``` script which is functionally same as ```scripts/build```
* Remove make targets for legacy packaging (generic-deb, generic-rpm etc) which are not used

### Testing

* Built agent image, rpm and deb on fresh instances.
* Registed Init packages with ECS clusters and successfully ran simple task

New tests cover the changes: 
No new tests

### Licensing


packaging/amazon-linux-ami-integrated/ecs-agent.spec
```diff
@@ -159,10 +159,7 @@ required routes among its preparation steps.
 
 %build
 # each of these should build for arm and amd arch
-./scripts/get-host-certs
-./scripts/build-cni-plugins
-./scripts/build-integrated true "" true true
-./scripts/build-agent-image
+make release-agent-internal
```
yinyic, line 162: Why do we use `-internal` here (vs just `release-agent` below)?

Makefile
```diff
@@ -254,13 +248,19 @@ dockerfree-pause:
...
 	./scripts/get-host-certs
 
-dockerfree-cni-plugins: get-cni-sources
+dockerfree-cni-plugins:
 	./scripts/build-cni-plugins
 
 # see dockerfree-pause above: assumes that the pre-compiled pause container tar exists
-dockerfree-agent-image: dockerfree-certs dockerfree-cni-plugins static-with-pause
+# builds agent image and saves on disk, assumes cni plugins have been pulled
+release-agent-internal: dockerfree-certs dockerfree-cni-plugins static
```
yinyic, line 256: When do we need this build target? Is there harm to always run `get-cni-sources`?

yinyic (commented):
Can we also update the [relevant section](https://github.com/aws/amazon-ecs-agent#building-and-running-from-source) in README?

Makefile
prateekchaudhry: Internal build process restricts connectivity to fetch submodules from github. This target facilitates building artifacts by assuming cni plugins have already been fetched and agent can be built in an isolated environment

packaging/amazon-linux-ami-integrated/ecs-agent.spec
prateekchaudhry: This target is used in internal build process where we assume cni plugins have been fetched already, build in an isolated environment and do not require fetching cni plugins. ```release-agent``` is basically ```get-cni-sources``` + ```release-agent-internal```. As such, it is a single target calling these two targets but otherwise no difference between them these two.

YashdalfTheGray (approved):

yinyic (approved):

PR #3537, opened 2023-01-20
Filter out metricCount=0 and its corresponding metricValue for service connect metric TargetResponseTime

ubhattacharjya:
This PR filters out all metric Counts for the histogram metric TargetResponseTime which are 0. We also remove the corresponding metricValues. 
This is because when there is no traffic and metricCount is 0, the dashboards incorrectly show min as 0 and max as 360000.  In this case, there should have been no datapoints. For example,

**Before:** 

**MetricCounts: [8, 0, 3, 0, 0, 0, 1, 0, 0, 0, 0, 0,0,  0,  0,  0,  0,  0,  0 ]
MetricValues: [  0.5,  1,  5,  10,  25,  50,  100,  250,  500,  1000,  2500,  5000,  10000,  30000,  60000,  300000,  600000,  1.8e+06,  3.6e+06 ]}**

image

This Pr correctly sends only non-zero metric counts and now it looks like below.

**After:**

**MetricCounts: [8, 3, 1]
MetricValues: [0.5, 5, 100]}**

image

### Testing

In the summary the test results are mentioned

BugFix: Filter out metricCount=0 and its corresponding metricValue for service connect metric TargetResponseTime

### Licensing

### DIFF
--- a/agent/stats/service_connect_linux.go
+++ b/agent/stats/service_connect_linux.go
@@ -117,7 +117,17 @@ func convertToTACSStats(mf map[string]*prometheus.MetricFamily, taskId string) (
 					}
 					metricCounts = append(metricCounts, &metricCount)
 				}
-				metricCounts = convertHistogramMetricCounts(metricCounts)
+
+				metricValues, metricCounts = convertHistogramMetricCounts(metricValues, metricCounts)
+
+				// If all values are 0 in metricCount, then no need to send the metrics to TACS
+				if len(metricCounts) == 0 {
+					logger.Debug("There were no non-zero metricCount received for TargetResponseTime metric. Skipping this metric.", logger.Fields{
+						field.TaskID: taskId,
+					})
+					continue
+				}
+
 			default:
 				logger.Warn("Service connect stats received invalid Metric type", logger.Fields{
 					field.TaskID: taskId,
@@ -212,13 +222,20 @@ func (sc *ServiceConnectStats) resetStats() {
 }
 
 // CloudWatch accepts the histogram buckets in a disjoint manner while the prometheus emits these values in a cumulative way.
-// This method performs that conversion
-func convertHistogramMetricCounts(metricCounts []*int64) []*int64 {
-	prevCount := *metricCounts[0]
-	for i := 1; i < len(metricCounts); i++ {
+// This method performs that conversion. We discard any metricCount that is 0 and also its corresponding metricValue.
+func convertHistogramMetricCounts(metricValues []*float64, metricCounts []*int64) ([]*float64, []*int64) {
+	var mV []*float64
+	var mC []*int64
+	prevCount := int64(0)
+	for i := 0; i < len(metricCounts); i++ {
 		prevCount, *metricCounts[i] = *metricCounts[i], *metricCounts[i]-prevCount
+		if metricCounts[i] != nil && *metricCounts[i] != 0 {
+			mV = append(mV, metricValues[i])
+			mC = append(mC, metricCounts[i])
+		}
 	}
-	return metricCounts
+
+	return mV, mC
 }
 
 // This method sorts the dimensions according to the keyName.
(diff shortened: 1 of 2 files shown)


agent/stats/service_connect_linux.go
```diff
@@ -117,7 +117,14 @@ func convertToTACSStats(mf map[string]*prometheus.MetricFamily, taskId string) (
 					}
 					metricCounts = append(metricCounts, &metricCount)
 				}
-				metricCounts = convertHistogramMetricCounts(metricCounts)
+
+				metricValues, metricCounts = convertHistogramMetricCounts(metricValues, metricCounts)
+
+				// If all values are 0 in metricCount, then no need to send the metrics to TACS
+				if len(metricCounts) == 0 && len(metricValues) == 0 {
```
yinyic, line 124: nit 
- `metricCounts` and `metricValues` should have the same size so checking one of them is good enough.
- Can we add a debug log here to indicate that the metric has been skipped?

agent/stats/service_connect_linux.go
```diff
@@ -212,13 +219,20 @@ func (sc *ServiceConnectStats) resetStats() {
...
-	prevCount := *metricCounts[0]
-	for i := 1; i < len(metricCounts); i++ {
+// This method performs that conversion. We discard any metricCount that is 0 and also its corresponding metricValue.
+func convertHistogramMetricCounts(metricValues []*float64, metricCounts []*int64) ([]*float64, []*int64) {
+	var mV []*float64
+	var mC []*int64
+	prevCount := int64(0)
+	for i := 0; i < len(metricCounts); i++ {
 		prevCount, *metricCounts[i] = *metricCounts[i], *metricCounts[i]-prevCount
+		if *metricCounts[i] != 0 {
```
sparrc, line 229: just to be safe could we make this 
```
if *metricCounts[i] != nil && *metricCounts[i] != 0
```

yinyic (approved):

mythri-garaga (dismissed):

agent/stats/service_connect_linux.go
ubhattacharjya: Fixed the above 2.

yinyic (dismissed):

agent/stats/service_connect_linux.go
ubhattacharjya: Added the check

rajal-amzn (dismissed):

yinyic (approved):

agent/stats/service_connect_linux.go
karanvasnani: Instead of individually checking 0 values in Counts/Values array, can we just check the SampleCount for being 0 (`metric.Histogram.SampleCount` I think) on L99 and skip the processing of its buckets entirely?

agent/stats/service_connect_linux.go
```diff
@@ -117,7 +117,17 @@ func convertToTACSStats(mf map[string]*prometheus.MetricFamily, taskId string) (
 					}
 					metricCounts = append(metricCounts, &metricCount)
 				}
-				metricCounts = convertHistogramMetricCounts(metricCounts)
+
+				metricValues, metricCounts = convertHistogramMetricCounts(metricValues, metricCounts)
+
+				// If all values are 0 in metricCount, then no need to send the metrics to TACS
+				if len(metricCounts) == 0 {
```
karanvasnani, line 124: As an optimization we can pre-compute this by checking the SampleCount value to be 0 (`metric.Histogram.SampleCount` I think) on L99 and completely skip processing these buckets?

karanvasnani (approved):

rajal-amzn (approved):

agent/stats/service_connect_linux.go
ubhattacharjya: sure. I would take this up as a follow-up task and address that in a separate PR.

mythri-garaga (approved):

PR #3540, opened 2023-01-23
fix gmsa,fsx capability bug

arun-annamalai:
### Summary

This pull request fixes how the Agent determines whether it is gMSA and fsx capable. Currently when either of these capabilities are set to false, the `config.mergeDefaultConfig(errs)` line will set these back to true on Windows since the lhs is the boolean false and the default config (rhs) contains the boolean true. 

### Implementation details

This fix is done by changing the native `boolean` to now use `BooleanDefaultFalse`

### Testing

This change was tested both by using unit tests and a custom ecs agent installed on a windows ec2 instance which was connected to a cluster. The logs were viewed to see what capabilities were advertised upon agent startup. 

New tests cover the changes: 
yes

Bug - Fixed a bug that incorrectly advertised the gMSA and fsx capability 

### Licensing

### DIFF
--- a/agent/config/config_windows.go
+++ b/agent/config/config_windows.go
@@ -142,8 +142,8 @@ func DefaultConfig() Config {
 		SharedVolumeMatchFullConfig:         BooleanDefaultFalse{Value: ExplicitlyDisabled}, //only requiring shared volumes to match on name, which is default docker behavior
 		PollMetrics:                         BooleanDefaultFalse{Value: NotSet},
 		PollingMetricsWaitDuration:          DefaultPollingMetricsWaitDuration,
-		GMSACapable:                         true,
-		FSxWindowsFileServerCapable:         true,
+		GMSACapable:                         BooleanDefaultFalse{Value: ExplicitlyDisabled},
+		FSxWindowsFileServerCapable:         BooleanDefaultFalse{Value: ExplicitlyDisabled},
 		PauseContainerImageName:             DefaultPauseContainerImageName,
 		PauseContainerTag:                   DefaultPauseContainerTag,
 		CNIPluginsPath:                      filepath.Join(ecsBinaryDir, defaultCNIPluginDirName),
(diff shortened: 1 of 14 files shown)


jterry75 (approved):
LGTM

YashdalfTheGray:
Reran the failed tests one more time.

rawahars (approved):

YashdalfTheGray (approved):

agent/config/config_windows.go
```diff
@@ -142,8 +142,8 @@ func DefaultConfig() Config {
 		SharedVolumeMatchFullConfig:         BooleanDefaultFalse{Value: ExplicitlyDisabled}, //only requiring shared volumes to match on name, which is default docker behavior
 		PollMetrics:                         BooleanDefaultFalse{Value: NotSet},
 		PollingMetricsWaitDuration:          DefaultPollingMetricsWaitDuration,
-		GMSACapable:                         true,
-		FSxWindowsFileServerCapable:         true,
+		GMSACapable:                         BooleanDefaultFalse{Value: ExplicitlyDisabled},
+		FSxWindowsFileServerCapable:         BooleanDefaultFalse{Value: ExplicitlyDisabled},
```
fierlion, line 146: Here we are changing the default FSsWindowsCapability from true to false, but I don't see where it's being enabled.  This is a behavior change, (as FSx Windows has been available for 2 years).  Can you point out where this is being enabled?

fierlion (changes requested):

agent/config/config_windows.go
YashdalfTheGray: I found the following initialization code for FSxWindowsFileServerCapable, can we confirm that this is still applicable? 

https://github.com/aws/amazon-ecs-agent/blob/2dcc42bd8a6f1f253537ea820b8920c9a805a7c1/agent/config/parse_windows.go#L44

arun-annamalai: yes the area that Yash has linked is the exact area where we determine whether to advertise the FXs and gMSA capabilities.

yinyic (approved):

fierlion (approved):

PR #3541, opened 2023-01-24
Skip sending STSC events for internal tasks

yinyic:
### Summary
We are currently sending STSC events for AppNet relay task. This task is initiated and managed by ECS Agent to support Service Connect feature, and not meant to be exposed to customers. Therefore, we should not be sending STSC events for it.

### Implementation details
Adding a condition check to skip creating STSC events for internal tasks. 

### Testing

* make test
* tested updated Agent with SC task, observed following log message
```
level=error time=2023-01-24T18:04:42Z msg="Skipping emitting event for task due to error" task="service-connect-relay-935909be-9c11-11ed-92a6-0a2da05dafe1" reason="" error="skip creating task stage change event for internal task arn:::::/service-connect-relay-935909be-9c11-11ed-92a6-0a2da05dafe1"

```

Skip sending STSC events for internal tasks

### Licensing

### DIFF
--- a/agent/api/statechange.go
+++ b/agent/api/statechange.go
@@ -105,6 +105,9 @@ type AttachmentStateChange struct {
 // returns error if the state change doesn't need to be sent to the ECS backend.
 func NewTaskStateChangeEvent(task *apitask.Task, reason string) (TaskStateChange, error) {
 	var event TaskStateChange
+	if task.IsInternal {
+		return event, errors.Errorf("skip creating task stage change event for internal task %v", task.Arn)
+	}
 	taskKnownStatus := task.GetKnownStatus()
 	if !taskKnownStatus.BackendRecognized() {
 		return event, errors.Errorf(


mythri-garaga (approved):

agent/api/statechange.go
```diff
@@ -105,6 +105,9 @@ type AttachmentStateChange struct {
 // returns error if the state change doesn't need to be sent to the ECS backend.
 func NewTaskStateChangeEvent(task *apitask.Task, reason string) (TaskStateChange, error) {
 	var event TaskStateChange
+	if task.IsInternal {
+		return event, errors.Errorf("skip creating task stage change event for internal task %v", task.Arn)
```
ubhattacharjya, line 109: Since it is expected, should we return error for this ? Will it be better to add a debug level log and returning nil error?

yinyic: Callers ([here](https://github.com/aws/amazon-ecs-agent/blob/41e8e6a78391f65b17070ffe086876e63387d5bb/agent/engine/task_manager.go#L617) and [here](https://github.com/aws/amazon-ecs-agent/blob/fe0ce622b35294c773509831a31809910667bdb2/agent/engine/docker_task_engine.go#L786)) currently use error value for determining whether an event should be sent. This pattern applies to other types of event publishing as well. It might be better to have a separate `shouldCreateXXXEvent` method to be more clear, but I don't have strong objection with our current approach either (where we treat the "should not" situation as error).

ubhattacharjya: makes sense.

ubhattacharjya (approved):

PR #3554, opened 2023-01-31
Address envFile resource naming defect

danehlim:
### Summary

Address EnvironmentFileResource naming defect which causes a race condition when a task has multiple containers which specify at least one unique environment file.

For further context, please refer to the "Additional Context" section below in this pull request's description.

### Implementation details

- Modify EnvironmentFileResource `GetName()` method to easily differentiate EnvironmentFileResource objects from one another (i.e., add container name to return value since container name is unique)
- Use `resourcestatus.ResourceStatus(envFiles.EnvFileCreated)` instead of `resourcestatus.ResourceCreated` when building a container's EnvironmentFileResource dependency in the `initializeEnvfilesResource` function
- Update existing unit tests to account for new changes

#### Unit testing:
- `make test`

Existing unit test `TestInitializeAndGetEnvfilesResource` has been expanded to cover the changes introduced in this pull request and operates as follows:
1. Creates a task specifying two containers, with each container specifying its own single unique environment file
2. Initializes each containers’ EnvironmentFileResource (which specifies an EnvironmentFileResource dependency for each container)
3. Check that the specified resource dependency for each container is defined as expected
4. Make sure that the resource dependency of the first container is not equal to the resource dependency of the second container

This test will always pass when the fix in this pull request is implemented since the two resource dependencies will never be equal due to each EnvironmentFileResource object’s call to its `GetName()` method returning a unique name. This test will also always fail under the current ECS Agent because the two resource dependencies will always be considered equal due to all calls to EnvironmentFileResource `GetName()` method returning the constant string “envfile”. 

#### Manual testing:
An ECS customer who faced this issue was given a custom ECS Agent with the code changes from this PR built in and validated that they no longer ran into the issue after using that custom ECS Agent.

I was also able to reproduce the issue using the current ECS Agent by taking the following steps:
1. defining a task which has 2 nginx containers where one container specifies a small (<20 KB) environment file and the other container specifies a large (>500 MB) environment file
2. then trying to run 10 such tasks (and subsequently observing some are not able to start due to the race condition coming into play)

Upon using the custom ECS Agent mentioned previously, I could no longer reproduce the issue with the same steps.

New tests cover the changes: no (but existing `TestInitializeAndGetEnvfilesResource`  test has been modified to cover the changes)

### Additional Context
Per the [Amazon ECS Developer Guide](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/taskdef-envfiles.html), each container in a given task can specify up to 10 environment files. If a container specifies 1 or more environment files, then all of its environment files are grouped into a single EnvironmentFileResource. So for each container with 1 or more environment files, an EnvironmentFileResource is created.

A problem arises in the current ECS Agent code when a task has multiple (i.e., more than one) containers which specify at least one unique environment file.

When EnvironmentFileResource objects are initialized in the `initializeEnvfilesResource` function, each EnvironmentFileResource (tied to a unique container in the task) is:

1. added to the task’s ResourcesMap’s slice of task resource objects mapped from “envfile” key
4. added as part of a ResourceDependency for the container it corresponds to, such that the container cannot be transitioned into a “CREATED” status until task resource with resource name “envfile” has KnownStatus “CREATED”

The 2nd point above is problematic. Every EnvironmentFileResource object’s `GetName()` method always returns “envfile”, which means all EnvironmentFileResource objects have the same name. 

Then later, when we check if dependencies for a container are resolved, we pass the slice of task resources returned by `mtask.GetResources()` to the `DependenciesAreResolved` function. 

Diving deeper into `mtask.GetResources()` reveals that it calls `task.getResourcesUnsafe()`, which iterates through the task’s resource map. According to the [Go Programming Language Specification](https://go.dev/ref/spec#For_statements), “the iteration order over maps is not specified and is not guaranteed to be the same from one iteration to the next”. Thus, there is no guarantee as to how the task resources will be ordered in the slice of task resources that we pass to the `DependenciesAreResolved` function.

Recall that all EnvironmentFileResource objects have the same name. Because of this, observe that in the `DependenciesAreResolved` function, only the last EnvironmentFileResource encountered while iterating through the slice of task resources returned by the previous `mtask.GetResources()` call (this EnvironmentFileResource could correspond to any container in the task) will be the one we check if it has KnownStatus CREATED” yet, **regardless of whether or not that EnvironmentFileResource is actually associated with the current container**. That is, only the last EnvironmentFileResource encountered can occupy this map at key value “envfile”.

Thus, even though we have declared EnvironmentFileResource dependencies for containers, there is no guarantee that a container will try to transition to “CREATED” only after its associated environment files have finished downloading and been written to disk.

#### Simplified Example:
Let’s say we have a task that specifies 2 containers, container A and container B where each container specifies its own single unique environment file and the following sequence of events happens:

1. Container A’s EnvironmentFileResource and Container B’s EnvironmentFileResource are initialized
2. Container A’s environment file starts downloading
5. Container B’s environment file starts downloading
6. Container A’s environment file finishes downloading and has been written to disk
7. Container A’s EnvironmentFileResource transitions to KnownStatus “CREATED” since all of its environment files have finished downloading
8. There is a request to transition Container B to “CREATED”, but first Agent must check if Container B’s dependencies are resolved
    1. Get the task’s resources from task’s resource map and store them in a slice (ordering of how task resources get added to the slice NOT guaranteed)
    2. Iterate through the slice above, but only the last EnvironmentFileResource encountered is selected
    4. Let’s say it just so happens the last EnvironmentFileResource encountered is Container A’s EnvironmentFileResource
    5. Agent checks the selected EnvironmentFileResource to see if it has KnownStatus “CREATED”
    6. It sees it has KnownStatus “CREATED”, so the result of this check does not block Container B from transitioning to “CREATED”
9. Agent begins transitioning Container B to “CREATED”
    1. Part of this transition includes reading from Container B’s environment file, which has NOT yet finished downloading yet
10. “Unable to open environment file” error is thrown
11. Container B’s environment file finishes downloading and has been written to disk, but it’s too late already

Address envFile resource naming defect

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -3240,7 +3240,7 @@ func (task *Task) initializeEnvfilesResource(config *config.Config, credentialsM
 				return errors.Wrapf(err, "unable to initialize envfiles resource for container %s", container.Name)
 			}
 			task.AddResource(envFiles.ResourceName, envfileResource)
-			container.BuildResourceDependency(envfileResource.GetName(), resourcestatus.ResourceCreated, apicontainerstatus.ContainerCreated)
+			container.BuildResourceDependency(envfileResource.GetName(), resourcestatus.ResourceStatus(envFiles.EnvFileCreated), apicontainerstatus.ContainerCreated)
 		}
 	}
 
--- a/agent/api/task/task_test.go
+++ b/agent/api/task/task_test.go
@@ -3778,7 +3778,7 @@ func TestPostUnmarshalTaskEnvfiles(t *testing.T) {
 	}
 
 	resourceDep := apicontainer.ResourceDependency{
-		Name:           envFiles.ResourceName,
+		Name:           envFiles.ResourceName + "_" + container.Name,
 		RequiredStatus: resourcestatus.ResourceStatus(envFiles.EnvFileCreated),
 	}
 
@@ -3791,22 +3791,34 @@ func TestPostUnmarshalTaskEnvfiles(t *testing.T) {
 }
 
 func TestInitializeAndGetEnvfilesResource(t *testing.T) {
-	envfile := apicontainer.EnvironmentFile{
-		Value: "s3://bucket/envfile",
+	envfile1 := apicontainer.EnvironmentFile{
+		Value: "s3://bucket/envfile1",
 		Type:  "s3",
 	}
 
-	container := &apicontainer.Container{
-		Name:                      "containerName",
+	envfile2 := apicontainer.EnvironmentFile{
+		Value: "s3://bucket/envfile2",
+		Type:  "s3",
+	}
+
+	container1 := &apicontainer.Container{
+		Name:                      "containerName1",
 		Image:                     "image:tag",
-		EnvironmentFiles:          []apicontainer.EnvironmentFile{envfile},
+		EnvironmentFiles:          []apicontainer.EnvironmentFile{envfile1},
+		TransitionDependenciesMap: make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet),
+	}
+
+	container2 := &apicontainer.Container{
+		Name:                      "containerName2",
+		Image:                     "image:tag",
+		EnvironmentFiles:          []apicontainer.EnvironmentFile{envfile2},
 		TransitionDependenciesMap: make(map[apicontainerstatus.ContainerStatus]apicontainer.TransitionDependencySet),
 	}
 
 	task := &Task{
 		Arn:                "testArn",
 		ResourcesMapUnsafe: make(map[string][]taskresource.TaskResource),
-		Containers:         []*apicontainer.Container{container},
+		Containers:         []*apicontainer.Container{container1, container2},
 	}
 
 	ctrl := gomock.NewController(t)
@@ -3819,15 +3831,30 @@ func TestInitializeAndGetEnvfilesResource(t *testing.T) {
 
 	task.initializeEnvfilesResource(cfg, credentialsManager)
 
-	resourceDep := apicontainer.ResourceDependency{
-		Name:           envFiles.ResourceName,
+	resourceDep1 := apicontainer.ResourceDependency{
+		Name:           envFiles.ResourceName + "_" + container1.Name,
 		RequiredStatus: resourcestatus.ResourceStatus(envFiles.EnvFileCreated),
 	}
 
-	assert.Equal(t, resourceDep,
+	resourceDep2 := apicontainer.ResourceDependency{
+		Name:           envFiles.ResourceName + "_" + container2.Name,
+		RequiredStatus: resourcestatus.ResourceStatus(envFiles.EnvFileCreated),
+	}
+
+	assert.Equal(t, resourceDep1,
 		task.Containers[0].TransitionDependenciesMap[apicontainerstatus.ContainerCreated].ResourceDependencies[0])
 
-	_, ok := task.getEnvfilesResource("containerName")
+	assert.Equal(t, resourceDep2,
+		task.Containers[1].TransitionDependenciesMap[apicontainerstatus.ContainerCreated].ResourceDependencies[0])
+
+	assert.NotEqual(t,
+		task.Containers[0].TransitionDependenciesMap[apicontainerstatus.ContainerCreated].ResourceDependencies[0],
+		task.Containers[1].TransitionDependenciesMap[apicontainerstatus.ContainerCreated].ResourceDependencies[0])
+
+	_, ok := task.getEnvfilesResource(container1.Name)
+	assert.True(t, ok)
+
+	_, ok = task.getEnvfilesResource(container2.Name)
 	assert.True(t, ok)
 }
 
--- a/agent/taskresource/envFiles/envfile.go
+++ b/agent/taskresource/envFiles/envfile.go
@@ -202,9 +202,9 @@ func (envfile *EnvironmentFileResource) GetCreatedAt() time.Time {
 	return envfile.createdAtUnsafe
 }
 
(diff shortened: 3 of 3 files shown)


agent/api/task/task_test.go
```diff
@@ -3791,22 +3791,34 @@ func TestPostUnmarshalTaskEnvfiles(t *testing.T) {
 }
 
 func TestInitializeAndGetEnvfilesResource(t *testing.T) {
```
yinyic, line 3793: Is the modified test able to reproduce the bug reliably?

agent/api/task/task.go
```diff
@@ -3240,7 +3240,7 @@ func (task *Task) initializeEnvfilesResource(config *config.Config, credentialsM
 				return errors.Wrapf(err, "unable to initialize envfiles resource for container %s", container.Name)
 			}
 			task.AddResource(envFiles.ResourceName, envfileResource)
-			container.BuildResourceDependency(envfileResource.GetName(), resourcestatus.ResourceCreated, apicontainerstatus.ContainerCreated)
+			container.BuildResourceDependency(envfileResource.GetName(), resourcestatus.ResourceStatus(envFiles.EnvFileCreated), apicontainerstatus.ContainerCreated)
```
yinyic, line 3243: Could you help me understand why do we want to use `resourcestatus.ResourceStatus(envFiles.EnvFileCreated)`

agent/api/task/task_test.go
danehlim: The modified test will reliably expose the bug under current ECS Agent. 

Under the new modified test, we create a task specifying two containers, with each container specifying its own single unique environment file. Then we initialize each containers’ EnvironmentFileResource (which specifies an EnvironmentFileResource dependency for each container). At the end, we assert that the resource dependency of the first container must not be equal to the resource dependency of the second container.

With current Agent, the new test will always fail because the two resource dependencies in the test will always be considered equal due to all calls to EnvironmentFileResource GetName() method returning the constant string “envfile” (thereby exposing the bug). With the bug fix in this pull request implemented, the test passes.

agent/api/task/task.go
danehlim: Sure thing! I believe it to be a better practice to build the dependency on the status defined by the specific type of task resource, rather than the status defined for task resources in general. Consider the following: what if in the future we want to define a new status specific to environment file resource?

We also see this practice being employed already for most other types of task resources:
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L484
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L712
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1045
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1083
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1090
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1151
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1158
https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1481

agent/api/task/task_test.go
yinyic: Nice :)

agent/api/task/task.go
danehlim: Also observe that `resourcestatus.ResourceStatus(EnvFileCreated)` is the one used in `envfile.go`:

https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/envFiles/envfile.go#L136
https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/envFiles/envfile.go#L223
https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/envFiles/envfile.go#L249

and that there are no references to `resourcestatus.ResourceCreated` in this file. The one used in `task.go` should match what is being used in `envfile.go`.

yinyic: Please let me know if I'm missing anything - looking at `resourcestatus.ResourceStatus(envFiles.EnvFileCreated)`, if we end up having more statuses for envFile resource (10 let's say), we could very well end up with an invalid `ResourceStatus` as the iota goes out of bound. The reason why we've been able to do it for all resources without an issue is because, I believe, all resource-specific statuses are actually **identical** to the generic statuses hence 1:1 mapping (that's the case when I spot-checked asm, volume, and firelens - probably the others as well). 

This is getting out of scope of the problem we are trying to solve at hand. But I feel like we might want to revisit how we are using a generic `ResourceStatus` alongside the status defined for a given type of resource (e.g. `envFiles.EnvFileCreated`). I'm not actually sure why we have both. When we drive a task/container from the top-level state machine, we can only use one status, so it's an either/or. If we want to do respect each resource's own statuses and transitions, then the `resourcestatus.ResourceStatus` should've just been an int type, not iota constant. Again, totally out of scope and probably should be followed up separately. But given what we currently have, I don't think this "conversion" here actually makes sense.

danehlim: Thanks for calling this out. Agreed this is getting out of scope of the change we are making in this pull request, but I will definitely create an internal backlog item in the ECS Agent team to revisit and investigate this as it is a bit confusing/unintuitive to figure out what exactly we are trying to achieve in the way we currently use these statuses.

yinyic (dismissed):

YashdalfTheGray (dismissed):
Excellent writeup for the defect, helped me get a lot of context. Changes look good.

danehlim:
Previous push is just to rebase and apply newest commits from `dev` branch.

yinyic (approved):

YashdalfTheGray (approved):

PR #3559, opened 2023-02-06
Use debug instead of error log for skipped events for internal task/container

yinyic:
### Summary

For internal tasks (e.g. service connect relay) or internal containers (e.g. pause containers), we do not wish to publish any state change events for them. Currently we log an error message when we encounter such task/container while constructing the events. But that could be noisy and misleading since this is not a real error but rather designed behavior. This PR updates all such error messages to be debug messages.

### Implementation details

Created new error type `ErrShouldNotSendEvent`. This error will be returned for internal task/containers when constructing a state change event. Caller will discern this error, and log a debug instead of error message in that case.

### Testing

* Ran `make test`
* Tested updated ECS Agent with service connect task, observed debug logs below
```
level=debug time=2023-02-06T21:01:52Z msg="Should not send events for internal tasks or containers: ~internal~ecs~pause-server"
level=debug time=2023-02-06T21:01:52Z msg="Should not send events for internal tasks or containers: ~internal~ecs~pause-ecs-service-connect-ciCxr"
level=debug time=2023-02-06T21:01:52Z msg="Should not send events for internal tasks or containers: instance-service-connect-relay"
```

Use debug instead of error log for skipped events for internal task/container

### Licensing

### DIFF
--- a/agent/api/statechange.go
+++ b/agent/api/statechange.go
@@ -101,12 +101,20 @@ type AttachmentStateChange struct {
 	Attachment *apieni.ENIAttachment
 }
 
+type ErrShouldNotSendEvent struct {
+	resourceId string
+}
+
+func (e ErrShouldNotSendEvent) Error() string {
+	return fmt.Sprintf("should not send events for internal tasks or containers: %s", e.resourceId)
+}
+
 // NewTaskStateChangeEvent creates a new task state change event
 // returns error if the state change doesn't need to be sent to the ECS backend.
 func NewTaskStateChangeEvent(task *apitask.Task, reason string) (TaskStateChange, error) {
 	var event TaskStateChange
 	if task.IsInternal {
-		return event, errors.Errorf("skip creating task stage change event for internal task %v", task.Arn)
+		return event, ErrShouldNotSendEvent{task.Arn}
 	}
 	taskKnownStatus := task.GetKnownStatus()
 	if !taskKnownStatus.BackendRecognized() {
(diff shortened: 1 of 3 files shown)


sparrc (approved):

singholt (dismissed):

agent/api/statechange.go
```diff
@@ -101,12 +101,20 @@ type AttachmentStateChange struct {
 	Attachment *apieni.ENIAttachment
 }
 
+type ErrShouldNotSendEvent struct {
+	resourceId string
+}
+
+func (e ErrShouldNotSendEvent) Error() string {
+	return fmt.Sprintf("Should not send events for internal tasks or containers: %s", e.resourceId)
```
amogh09, line 109: Should we start the error description with a lowercase as is the convention in Go?

agent/api/statechange.go
```diff
@@ -101,12 +101,20 @@ type AttachmentStateChange struct {
...
+}
+
 // NewTaskStateChangeEvent creates a new task state change event
 // returns error if the state change doesn't need to be sent to the ECS backend.
 func NewTaskStateChangeEvent(task *apitask.Task, reason string) (TaskStateChange, error) {
 	var event TaskStateChange
 	if task.IsInternal {
-		return event, errors.Errorf("skip creating task stage change event for internal task %v", task.Arn)
+		return event, ErrShouldNotSendEvent{task.Arn}
```
chienhanlin, line 117: Q. should we rename the error type or add more new error types since we don't send stake change under these 3 cases  
1. task.IsInternal 
2. !taskKnownStatus.BackendRecognized()
3.  task.GetSentStatus() >= taskKnownStatus

agent/api/statechange.go
yinyic: Fixed. Thanks!

agent/api/statechange.go
yinyic: Case #2 and #3 are a little different from case #1 imo. In case 1 we are dealing with special types of container/task, whereas in case 2 and 3 we are getting unexpected status which should not happen at all. And in such cases, we consider them real errors and log the message at error level (from the caller).

chienhanlin: Got it! Thanks for the clarification!

chienhanlin (approved):

amogh09 (approved):

singholt (approved):

PR #3561, opened 2023-02-07
Dev Update the Environment variables tables with the wording from the AWS documentation page.

browndjaz:
Please update the Environment variables table with the wording on this page

https://github.com/awsdocs/amazon-ecs-developer-guide/blob/master/doc_source/ecs-agent-config.md#ecs-agent-availparam


sparrc:
I think that this was meant to be a github "issue", going to close for now

danehlim:
Contacted @browndjaz directly to open an issue instead of a pull request for this.

Also, I noticed we currently list 88 unique environment variables in our README, but the [linked page](https://github.com/awsdocs/amazon-ecs-developer-guide/blob/master/doc_source/ecs-agent-config.md#ecs-agent-availparam) only lists 78 environment variables. I asked them when they open an issue to clarify if they are requesting us to only document the 78 environment variables from the linked page in our README moving forward, OR if we should keep the information about all 88 currently documented environment variables regardless of whether or not a given environment variable is listed on the linked page.

PR #3564, opened 2023-02-08
windows: explicitly provide the network name override to nat when using bridge network mode

rawahars:
### Summary

On Linux, the default network mode is `bridge` wherein all the containers are attached to a common docker bridge network. On Windows, this default mode translates to `nat` or `default` docker network.

When we use `bridge` as the network mode, then docker daemon errors out due to `bridge` network not present itself. Therefore, we need to do an override for Windows such that whenever `bridge` mode is specified in HostConfig, we would translate it to `nat` network.

This PR adds support for the same.
### Implementation details

We already have a method named `platformHostConfigOverride` in `task_windows.go` which performs platform specific overrides in the HostConfig. We have added the translation logic in the same method.

### Testing

A custom binary of the agent was built and was tested out.
Ran a bunch of tasks with both `bridge` and `default` mode.

windows: explicitly provide the network name override to nat when using bridge network mode
### Licensing

### DIFF
--- a/agent/api/task/task_windows.go
+++ b/agent/api/task/task_windows.go
@@ -43,6 +43,10 @@ const (
 	cpuSharesPerCore  = 1024
 	percentageFactor  = 100
 	minimumCPUPercent = 1
+
+	// windowsDefaultNetworkName is the name of the docker network to which the containers in
+	// default mode are attached to.
+	windowsDefaultNetworkName = "nat"
 )
 
 // PlatformFields consists of fields specific to Windows for a task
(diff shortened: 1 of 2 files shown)


arun-annamalai (approved):

singholt (approved):

chienhanlin (approved):

agent/api/task/task_windows.go
```diff
@@ -105,6 +109,13 @@ func (task *Task) platformHostConfigOverride(hostConfig *dockercontainer.HostCon
 		hostConfig.MemoryReservation = 0
 	}
 
+	// On Windows, bridge mode equivalent for Linux is "default" or "nat" network.
+	// We need to perform explicit conversion for the same.
```
ofiliz, line 113: A mental note on the comment above and how to think about these concepts:
The "bridge" networkMode here is ECS' network mode, not docker's. The fact that docker also has a networking mode called "bridge" should be treated as coincidental. Today ECS' bridge is implemented with docker's bridge under the hood, but that can change in the future.

Therefore, instead of saying "bridge mode equivalent is 'nat'", we should say "On Windows, ECS' bridge mode is implemented with "nat" mode."

agent/api/task/task_windows.go
```diff
@@ -43,6 +43,10 @@ const (
 	cpuSharesPerCore  = 1024
 	percentageFactor  = 100
 	minimumCPUPercent = 1
+
+	// windowsDefaultNetworkName is the name of the docker network to which the containers in
+	// default mode are attached to.
+	windowsDefaultNetworkName = "nat"
```
ofiliz, line 49: Minor: Do you need to prefix the variable with "windows" here, since this file is already Windows-specific?

ofiliz (approved):

PR #3558, opened 2023-02-06
Enumerate port ranges into docker config

YashdalfTheGray:
### Summary
When using `docker run` with a port range exposed, Docker enumerates every port within the range into the `Config.ExposedPorts` map but the ECS Agent just puts the range as a string into the config. This becomes a problem when containers themselves aren't exposing ports (the Dockerfile doesn't contain an `EXPOSE` instruction for example) but customers are exposing ports using ECS task definitions and also using dynamic host port binding. 

For example, assume that a container was built without an `EXPOSE` instruction in the Dockerfile but the customer's task definition lists a container port range of 80-90 with the protocol being TCP. Currently, there would be no way of reporting the bound ports since `"80-90/tcp"` gets added to the `Config.ExposedPorts` map and Docker seems to ignore anything that is not a singular port in the above map when it is reporting the host ports bound through the `NetworkSettings.Ports` map. 

### Implementation details
Using [`nat.ParsePortRangeToInt()`](https://pkg.go.dev/github.com/docker/go-connections/nat#ParsePortRangeToInt), we get the start and the end of the port range. Once we have that, we individually build a [`nat.Port`](https://pkg.go.dev/github.com/docker/go-connections/nat#Port) instance for each of the ports in the range and then we add them to the list of Docker exposed ports. 

### Testing

Unit tests were run and manual testing was also done by building the new agent with the changes and then running tasks on the instance with the new agent running. 

<details>
<summary>Manual testing results</summary>

`Config.ExposedPorts` after `$ docker run -d -p 9090 --name dr-sp public.ecr.aws/kulshres/simple-container`

```
$ docker inspect dr-sp | jq '.[0] | .Config | .ExposedPorts'
{
  "8080/tcp": {},
  "9090/tcp": {}
}
```

`Config.ExposedPorts` after `$ docker run -d -p 9090-9095 --name dr-pr public.ecr.aws/kulshres/ranges`

```
$ docker inspect dr-pr | jq '.[0] | .Config | .ExposedPorts'
{
  "8080/tcp": {},
  "8081/tcp": {},
  "8082/tcp": {},
  "8083/tcp": {},
  "8084/tcp": {},
  "8085/tcp": {},
  "9090/tcp": {},
  "9091/tcp": {},
  "9092/tcp": {},
  "9093/tcp": {},
  "9094/tcp": {},
  "9095/tcp": {}
}
```

[...]

PR #3569, opened 2023-02-13
Remove fallback to Docker for host port ranges assignment

chienhanlin:
### Summary
Starting from ECS Agent version `1.68.0 `, users can specify [containerPortRange](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html) at ECS task level, and `ECS_DYNAMIC_HOST_PORT_RANGE` at the container instance level to customize 1:1 container-host port mappings for their applications.

With the configured environment variable `ECS_DYNAMIC_HOST_PORT_RANGE`, ECS Agent will try to find a set of contiguous host ports within the given range. As users are expected to have ports to be bound in the customized range, errors should be returned instead of falling back for Docker to process dynamic port assignment if a set of contiguous host ports cannot be found by ECS Agent; therefore, this PR removes the Docker fallback mechanism when a set of contiguous host ports cannot be found by ECS Agent. 

### Issue
Scenario: `ECS_DYNAMIC_HOST_PORT_RANGE` is set, but host ports within the given range are not available for a bridge network mode task to bind it's container ports specified in the `containerPortRange` field.
* __Expected result__: Task failed or a container STOPPED with an error `Reason HostConfigError: error retrieving docker port map: 2 contiguous host ports unavailable`
* __Actual result__: Task succeeded but `networkBindings` is empty, and ecs-agent.log shows `level=error msg="Unable to find contiguous host ports for container, falling back to docker dynamic port assignment" container="xxx" containerPortRange="8085-8086" error="2 contiguous host ports unavailable" task="xxx"`

#### Setup
* ECS-optimized AMI: amzn2-ami-ecs-hvm-2.0.20230127-x86_64-ebs 
* ECS Agent version: 1.68.2
* An ECS cluster with only one registered EC2 instance 
* Userdata
```
#!/bin/bash
echo ECS_DYNAMIC_HOST_PORT_RANGE=100-102 >> /etc/ecs/ecs.config
``` 
* ECS Task definition
```
    {
      "networkMode": "bridge",
      "containerDefinitions": [
        {
          "portMappings": [
            {
              "protocol": "tcp",
              "containerPortRange": "8085-8086"
            }
          ],
          ...
        }
      ]
    }
```
* Ran the first ECS task, the task reached to RUNNING state and the`hostPortRange` returned from [ecs describe-tasks](https://docs.aws.amazon.com/cli/latest/reference/ecs/describe-tasks.html) call is within the given `ECS_DYNAMIC_HOST_PORT_RANGE` value.
```
$ aws ecs describe-tasks --cluster xxx --tasks "xxx" --region xxx
{
            "containers": [
                {
                    "networkBindings": [
                        {
                            "bindIP": "0.0.0.0",
                            "protocol": "tcp",
                            "containerPortRange": "8085-8086",
                            "hostPortRange": "100-101"
                        }
                    ],
                    ...
                }
            ],
}
```
* Ran the same ECS task again, the task reached to RUNNING state but the `hostPortRange` returned from [ecs describe-tasks](https://docs.aws.amazon.com/cli/latest/reference/ecs/describe-tasks.html) call is empty.
```
$ aws ecs describe-tasks --cluster xxx --tasks "xxx" --region xxx
{
            "containers": [
                {
                    "lastStatus": "RUNNING",
                    "networkBindings": [],
                 ...
                }
            ],
}
``` 

### Implementation details
1. agent/api/task/task.go
 - for `containerPortRange` case, return nil as docker port map along with an error in `dockerPortMap()` when a set of contiguous host ports cannot be found by ECS Agent.
2. agent/api/task/task_test.go
- Remove an invalid test after the change made in `dockerPortMap()`

### Testing
1. Replaced the ECS Agent with the fix on the host, run 2 tasks and confirmed the second tasks failed as expected. Details of setup are provided in the __Issue__ section.

```
$ aws ecs describe-tasks --cluster xxx --tasks "xxx" --region xxx
{

            "containers": [
                {
                    "lastStatus": "STOPPED",
                    "reason": "HostConfigError: error retrieving docker port map: 2 contiguous host ports unavailable",
                    "networkInterfaces": [],
                }
}
```

```
ecs-agent.log

level=info time=xxx msg="Creating container" task="xxx" container="xxx"
level=error time=xxx msg="Unable to find contiguous host ports for container" containerPortRange="8085-8086" error="2 contiguous host ports unavailable" task="xxx" container="xxx"
...
level=error time=xxx msg="Error transitioning container" task="xxx" container="xxx" runtimeID="" nextState="CREATED" error="error retrieving docker port map: 2 contiguous host ports unavailable"
...
level=warn time=xxx msg="Error creating container; marking its desired status as STOPPED" task="xxx" container="xxx" error="error retrieving docker port map: 2 contiguous host ports unavailable"
...
```

Bug - Remove fallback to Docker for host port ranges assignment

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -1814,7 +1814,8 @@ func (task *Task) dockerExposedPorts(container *apicontainer.Container) (dockerE
 			dockerPort := nat.Port(strconv.Itoa(int(portBinding.ContainerPort)) + "/" + protocol)
 			dockerExposedPorts[dockerPort] = struct{}{}
 		} else if portBinding.ContainerPortRange != "" {
-			// we supply containerPortRange here in case we did not assign a host port range and ask docker to do so
+			// we supply containerPortRange here to ensure every port exposed through ECS task definitions
+			// will be reported as Config.ExposedPorts in Docker
 			startContainerPort, endContainerPort, err := nat.ParsePortRangeToInt(portBinding.ContainerPortRange)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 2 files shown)


agent/api/task/task.go
```diff
@@ -2408,20 +2408,14 @@ func (task *Task) dockerPortMap(container *apicontainer.Container, dynamicHostPo
...
 					field.Container:      container.Name,
 					"containerPortRange": containerPortRange,
 					field.Error:          err,
 				})
-
-				// append individual container port from the containerPortRange into the containerPortSet.
-				// this will ensure that we populate network bindings for ports that docker dynamically assigned.
-				for port := startContainerPort; port <= endContainerPort; port++ {
-					containerPortSet[port] = struct{}{}
+				if err != nil {
```
singholt, line 2417: the parent if condition and this one is the same? so we can log and then return immediately?

chienhanlin: Thanks for catching this! Line 2417 is redundant, and should be removed.

agent/api/task/task.go
```diff
@@ -2408,20 +2408,14 @@ func (task *Task) dockerPortMap(container *apicontainer.Container, dynamicHostPo
...
+					return nil, err
 				}
 			} else {
```
singholt, line 2420: with the new changes, I think the code in this else block can be as is (without being wrapped in "else"). thoughts?

singholt:
does this need to be changed as well? https://github.com/aws/amazon-ecs-agent/blob/a03d9107a43b5b0826b0db46f3dd7c119a20e14d/agent/api/task/task.go#L1817

singholt:
I think with

> does this need to be changed as well?
> 
> https://github.com/aws/amazon-ecs-agent/blob/a03d9107a43b5b0826b0db46f3dd7c119a20e14d/agent/api/task/task.go#L1817

I think with Yash's last [PR](https://github.com/aws/amazon-ecs-agent/pull/3558): we still need those dockerExposedPorts() injections, but just that the comment is no longer a valid one?

agent/api/task/task.go
chienhanlin: Yeah, I will remove the else block and update comments above.

```
// we will try to get a contiguous set of host ports from the ephemeral host port range.
// this is to ensure that docker maps host ports in a contiguous manner, and
// we are guaranteed to have the entire hostPortRange in a single network binding while sending this info to ECS;
// therefore, an error will be returned if we cannot find a contiguous set of host ports.
hostPortRange, err := getHostPortRange(numberOfPorts, protocol, dynamicHostPortRange)
```

chienhanlin:
> does this need to be changed as well?
> 
> https://github.com/aws/amazon-ecs-agent/blob/a03d9107a43b5b0826b0db46f3dd7c119a20e14d/agent/api/task/task.go#L1817

singholt (approved):

YashdalfTheGray (approved):
Great writeup for the PR!

PR #3583, opened 2023-02-20
fixing build on mac

singholt:
### Summary

When trying to run unit tests locally, the current agent code doesn't compile and prevents me from running tests. This fixes the build on darwin (macOS) and unblocks running tests.

### Implementation details

Builds on darwin fail due to some objects and methods undeclared. This refactors the missing pieces – moving a few things into common files and refactoring some other into an `_unsupported` file. This follows the existing code convention.

### Testing

Verified that I am able to build and run tests locally.

Fix builds on a unix-like system other than linux.

### Licensing

### DIFF
--- a/agent/api/task/task_windows.go
+++ b/agent/api/task/task_windows.go
@@ -26,11 +26,14 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/containernetworking/cni/libcni"
 
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	apieni "github.com/aws/amazon-ecs-agent/agent/api/eni"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
+	"github.com/aws/amazon-ecs-agent/agent/taskresource/credentialspec"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/fsxwindowsfileserver"
+	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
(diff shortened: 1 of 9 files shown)


agent/api/task/task_windows.go
```diff
@@ -253,3 +256,38 @@ func (task *Task) BuildCNIConfigAwsvpc(includeIPAMConfig bool, cniConfig *ecscni
...
+	for _, container := range task.Containers {
+		if container.RequiresCredentialSpec() {
+			container.BuildResourceDependency(credentialspecResource.GetName(),
+				resourcestatus.ResourceStatus(credentialspec.CredentialSpecCreated),
+				apicontainerstatus.ContainerCreated)
+		}
+	}
+
+	return nil
+}
```
ofiliz, line 293: The changes in ecscni package are good, and should be done anyway.

But I am not in favor of duplicating code between our two supported OS platforms (Linux and Windows) to make the agent compile on an unsupported platform (Mac). We shouldn't be running our tests on Mac anyway since that is not a target platform, has different behaviors and hence the tests results may be misleading.

singholt: thanks Onur, 

the `initializeCredentialSpecResource` should return an explicit error for unsupported platforms, like -
```
errors.New("unsupported platform")
```

this seems to be missed when support for gmsa on linux was added. with this PR, we achieve that, although the task_linux and task_windows implementations are the same.

any ideas on how to avoid that?

petderek: One thing you could do is do a cross platform compile for osx in gh actions:

`GOOS=darwin go build ./` or something like that... you don't need osx hardware to do that but you still have some validation that your code isn't breaking on mac. You can also compile the tests themselves if you want bonus points: https://stackoverflow.com/questions/72721580/golang-compile-all-tests-across-a-repo-without-executing-them

I think we got rid of the mac tests a million years ago because freemium CI solutions tend to have limited OSX queues.

singholt: thanks! I'll close this PR for now.

PR #3586, opened 2023-02-22
periodically disconnect from acs

singholt:
### Summary

This PR introduces a connection time property to the ACS session object. With this change, the agent will periodically disconnect from ACS (every 15-45 minutes).

Currently we rely on ACS to periodically disconnect from the agent, unless the agent stops receiving heartbeats from ACS. About every ~1-2 minutes, if the agent does not receive any heartbeats and if no activity occurs, the agent closes its connection to ACS. 

If an ACS host is unhealthy (i.e. it looses track of its active connections and doesn't disconnect periodically anymore) and that we can no longer rely on heartbeats, the agent will be stuck with a stale connection. This may lead to task credentials not being refreshed or no task payload being served by this container instance. With this change, we would no longer have a single point of failure (i.e. unhealthy ACS host) and we let the agent control its fate.

### Implementation details

- A new `connectionTimer` is started when the ACS websocket connection is established. 
- When this timer is up, the agent sends all pending acks to ACS and writes a close message on the connection using websocket [Control Messages](https://pkg.go.dev/github.com/gorilla/websocket#hdr-Control_Messages). The agent sends a message of type [websocket.CloseMessage](https://github.com/gorilla/websocket/blob/9111bb834a68b893cebbbaed5060bdbc1d9ab7d2/conn.go#L74) with status code [NormalClosure](https://github.com/gorilla/websocket/blob/9111bb834a68b893cebbbaed5060bdbc1d9ab7d2/conn.go#L46).
- The close message is sent to ACS which then is echoed back to the agent. This close message is added to the message queue. This ensures that the agent reads and handles unread messages first, before closing the connection. [Reference](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.1)
- When agent detects the close message, it returns an `io.EOF` back to the parent go routine that started the ACS session. This is in-line with how it returns an `io.EOF` when ACS initiates the connection close. [Reference](https://github.com/aws/amazon-ecs-agent/blob/4943e4d853a84cf4d76b9d1241ee1f4f189b9c4a/agent/wsclient/client.go#L390-L392)
- Upon `io.EOF`, the websocket connection is established immediately without backoff.

### Testing

- Updated unit tests.
- Built a test AMI with these changes and launched a container instance that disconnects to ACS every minute (for testing purposes). The connection is re-established in < ~1 second.
- Let this container instance run for about 2 days to monitor the agent, upon frequent disconnection.
- Created an ECS service and ran a small test that increments the service's desired task count by 1 every 25 seconds, upto 50 tasks.
- Checked ECS agent logs to verify that the disconnection workflow works as expected and that all tasks get served as requested.
```
level=info msg="Closing ACS websocket connection after 1 minutes" module=acs_handler.go
level=debug msg="Connection closed for a valid reason: websocket: close 1000 (normal): ConnectionExpired: Reconnect to continue" module=client.go
level=info msg="ACS Websocket connection closed for a valid reason" module=acs_handler.go
level=info msg="ACS Websocket connection closed for a valid reason: EOF" module=acs_handler.go
level=debug msg="Received connect to ACS message" module=acs_handler.go
level=info msg="Establishing a Websocket connection ..."
level=debug msg="Established a Websocket connection to ..."
level=info msg="Connected to ACS endpoint"
```
- Checked ACS logs and verified the sequence of operations that is executed when agent closes the connection.

Enhancement: periodically disconnect from ACS

### Licensing

### DIFF
--- a/agent/acs/handler/acs_handler.go
+++ b/agent/acs/handler/acs_handler.go
@@ -39,6 +39,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/utils/ttime"
 	"github.com/aws/amazon-ecs-agent/agent/version"
 	"github.com/aws/amazon-ecs-agent/agent/wsclient"
+
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/cihub/seelog"
 )
@@ -53,6 +54,10 @@ const (
 	wsRWTimeout = 2*heartbeatTimeout + heartbeatJitter
 
 	inactiveInstanceReconnectDelay = 1 * time.Hour
+
+	// connectionTime is the maximum time after which agent closes its connection to ACS
+	connectionTime   = 15 * time.Minute
+	connectionJitter = 30 * time.Minute
 
 	connectionBackoffMin        = 250 * time.Millisecond
 	connectionBackoffMax        = 2 * time.Minute
@@ -100,6 +105,8 @@ type session struct {
 	doctor                          *doctor.Doctor
 	_heartbeatTimeout               time.Duration
 	_heartbeatJitter                time.Duration
+	connectionTime                  time.Duration
+	connectionJitter                time.Duration
 	_inactiveInstanceReconnectDelay time.Duration
 }
 
@@ -107,7 +114,7 @@ type session struct {
 // a session with ACS. This interface is intended to define methods
 // that create resources used to establish the connection to ACS
 // It is confined to just the createACSClient() method for now. It can be
-// extended to include the acsWsURL() and newDisconnectionTimer() methods
+// extended to include the acsWsURL() and newHeartbeatTimer() methods
 // when needed
 // The goal is to make it easier to test and inject dependencies
 type sessionResources interface {
@@ -182,6 +189,8 @@ func NewSession(
 		doctor:                          doctor,
 		_heartbeatTimeout:               heartbeatTimeout,
 		_heartbeatJitter:                heartbeatJitter,
+		connectionTime:                  connectionTime,
+		connectionJitter:                connectionJitter,
 		_inactiveInstanceReconnectDelay: inactiveInstanceReconnectDelay,
 	}
 }
@@ -224,7 +233,7 @@ func (acsSession *session) Start() error {
 				}
 			}
 			if shouldReconnectWithoutBackoff(acsError) {
-				// If ACS closed the connection, there's no need to backoff,
+				// If ACS or agent closed the connection, there's no need to backoff,
 				// reconnect immediately
 				seelog.Infof("ACS Websocket connection closed for a valid reason: %v", acsError)
 				acsSession.backoff.Reset()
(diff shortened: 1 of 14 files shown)


ubhattacharjya (approved):

yinyic:
> The close message is sent to ACS which then is echoed back to the agent. This close message is added to the message queue. This ensures that the agent reads and handles unread messages first, before closing the connection. [Reference](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.1)

That's rather desirable! Curious if we were able to verify this behavior? (Agent sent close message, handled another payload message, and closed connection upon receiving close echo from ACS)

agent/acs/handler/acs_handler.go
```diff
@@ -492,6 +505,21 @@ func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration,
...
 
+// newConnectionTimer creates a new timer, after which agent closes its ACS websocket connection
+func newConnectionTimer(client wsclient.ClientServer, connectionTime time.Duration, connectionJitter time.Duration) ttime.Timer {
+	expiresAt := retry.AddJitter(connectionTime, connectionJitter)
+	timer := time.AfterFunc(expiresAt, func() {
+		seelog.Infof("Closing ACS websocket connection after %v minutes", expiresAt.Minutes())
+		// WriteCloseMessage() writes a close message using websocket control messages
+		// Ref: https://pkg.go.dev/github.com/gorilla/websocket#hdr-Control_Messages
+		err := client.WriteCloseMessage()
+		if err != nil {
```
yinyic, line 516: What happens now that this connection kind of "escaped" the timer... I wonder if we should rely on ACS to terminate the connection in that case, or attempt to force close it.

singholt: yes if an error is encountered while writing a close message, we would rely on ACS to terminate the connection. 

I kept this in-line with how we deal with closing the connection when no heartbeats are received and agent tries to close the connection: https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/acs_handler.go#L487

singholt:
> > The close message is sent to ACS which then is echoed back to the agent. This close message is added to the message queue. This ensures that the agent reads and handles unread messages first, before closing the connection. [Reference](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.1)
> 
> That's rather desirable! Curious if we were able to verify this behavior? (Agent sent close message, handled another payload message, and closed connection upon receiving close echo from ACS)

by queuing up the close message, we don't delay processing it but rather do it so that agent will first read other payload messages.

as soon as the close message is read, it processes it and terminates/reconnects the connection. make sense?

YashdalfTheGray (dismissed):
Thank you for the context and the write up on the testing. Looks good.

agent/acs/handler/acs_handler.go
```diff
@@ -492,6 +512,61 @@ func newDisconnectionTimer(client wsclient.ClientServer, timeout time.Duration,
...
+		}()
+
+		// send pending heartbeat acks
+		go func() {
+			heartbeatHandler.sendPendingHeartbeatAck()
+			wg.Done()
+		}()
+
+		// wait for acks from all handlers above to be sent before closing the websocket connection
+		wg.Wait()
```
yinyic, line 557: Why can't we send ACKs with the new connection? 
It makes sense to read remaining payloads on the connection so that we properly drain it before closing, but the ACKs are different IMO. They are local messages that are not put on the wire yet, so there shouldn't be any risk of "losing" them.

singholt: we clear all non-sent ACKs when the connection is closed, [example](https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/acs_handler.go#L283).

the ACKs are sitting in a channel. These acks are put in the channel when the associated handler finishes handling the message. 

As Cam noticed during his testing, not sending ACKs before closing the connection results in unwanted warn statements in our logs like – 
```
level=warn time=2023-02-23T18:45:45Z msg="Error 'ack'ing request with messageID: 101975fc-7767-4e40-9fdd-547b1d5a9107, error: websocket: close sent" module=refresh_credentials_handler.go
```

This happens because the agent tries to continuously send ACKs as long as there's an ack in the ack channel and that the agent is running.

Sending all pending acks before closing the connections helps us tell ACS that re-sending the same message is not needed and that the agent has already dealt with it. It also makes our logs much cleaner.

yinyic: Got it that makes sense. Thank you for helping me fill the gap!

YashdalfTheGray (approved):
Rereviewing after rebase and added tests.

prateekchaudhry (approved):

PR #3612, opened 2023-03-20
Added new CodeBuild CF template to update and upload agentVersionV2-<branch>.json for AL1

mye956:
### Summary
This PR will introduce a new CodeBuild project which will be kicked off only when a PR has been merged into AmazonLinuxLegacy branch. What this CodeBuild project will do is it'll generate an `agent.json` and `agentVersionV2-<branch>.json` files which will be used as part of our release process. The buildspec file for this CodeBuild project is a bit of a combination of the `release-config.yml` and `copy.yml` buildspec files which is currently being used by our release CodePipeline. 

Note: The agentVersionV2-<branch>.json file for AL1 will need to be manually generated along with the tracking branch (`<branch>` ) for AL1.

### Implementation details
build-infrastructure/al-maintenance-codebuild-stack.yml
- New CF template YAML file
- Similar parameters that's also being used in the release CodePipeline
- New `ServiceRoleArn` parameter which will be referred to as the CodeBuild service role (Will not be creating a new Service Role but instead reusing an existing one)
- Triggered by a `PR_MERGED`
- Using the `AmazonLinuxLegacy` branch as the destination branch (i.e. BASE_REF)

buildspecs/al-maintenance-release-config
- Similar to the `release-config.yml` file in the `dev` branch
- Will also be copying over the generated `agent.json` and new `agentVersionV2-<branch>.json` file over to the same release bucket that our current release CodePipeline is using

### Testing

This CodeBuild CF template file was used in my personal dev account. I've tested the CodeBuild project by cloning the `AmazonLinuxLegacy` branch into my forked repo and used it as the destination branch. I was able to successfully kick off the CodeBuild project each time a PR was merged.

- Feature: New release CodeBuild project for AL1
### Licensing

### Related PRs:
- https://github.com/aws/amazon-ecs-agent/pull/3611

### DIFF
--- a/build-infrastructure/al-maintenance-codebuild-stack.yml
+++ b/build-infrastructure/al-maintenance-codebuild-stack.yml
@@ -5,6 +5,7 @@ Parameters:
     GithubFullRepoName:
       Type: String
       Description: Full name of the agent repository to use (e.g. https://github.com/aws/amazon-ecs-agent.git)
+      Default: https://github.com/aws/amazon-ecs-agent.git
     GithubBranchName:
       Type: String
       Description: Name of the base branch to use to build, PRs to which will trigger builds (e.g. refs/heads/AmazonLinuxLegacy)
--- a/build-infrastructure/al-maintenance-codebuild-stack.yml
+++ b/build-infrastructure/al-maintenance-codebuild-stack.yml
@@ -37,8 +37,8 @@ Resources:
             Type: PLAINTEXT
             Value: !Ref ReleaseArtifactsBucketS3Uri
           - Name: GITHUB_SOURCE_BRANCH_NAME
-              Type: PLAINTEXT
-              Value: !Ref GithubSourceBranchName
+            Type: PLAINTEXT
+            Value: !Ref GithubSourceBranchName
       Name: !Ref 'AWS::StackName'
       QueuedTimeoutInMinutes: 60
       ServiceRole: !Ref ServiceRoleArn
--- a/build-infrastructure/al-maintenance-codebuild-stack.yml
+++ b/build-infrastructure/al-maintenance-codebuild-stack.yml
@@ -10,7 +10,7 @@ Parameters:
       Description: Name of the base branch to use to build, PRs to which will trigger builds (e.g. refs/heads/AmazonLinuxLegacy)
     GithubSourceBranchName:
       Type: String
-      Description: Name of the source branch used for the PRs (e.g. refs/heads/al1-patch)
+      Description: Name of the source branch used for the PRs (e.g. al1-patch)
     ReleaseArtifactsBucketS3Uri:
       Type: String
       Description: The URI of the bucket where things land at the end, this is assumed to already exist (e.g. s3://artifacts)
@@ -36,6 +36,9 @@ Resources:
           - Name: RESULTS_BUCKET_URI
             Type: PLAINTEXT
             Value: !Ref ReleaseArtifactsBucketS3Uri
+          - Name: GITHUB_SOURCE_BRANCH_NAME
+              Type: PLAINTEXT
+              Value: !Ref GithubSourceBranchName
       Name: !Ref 'AWS::StackName'
       QueuedTimeoutInMinutes: 60
       ServiceRole: !Ref ServiceRoleArn
@@ -49,8 +52,6 @@ Resources:
         FilterGroups:
           - - Type: EVENT
               Pattern: 'PULL_REQUEST_MERGED'
-            - Type: HEAD_REF
-              Pattern: !Sub '^${GithubSourceBranchName}$'
             - Type: BASE_REF
               Pattern: !Sub '^${GithubBranchName}$'
         Webhook: true
--- a/build-infrastructure/al-maintenance-codebuild-stack.yml
+++ b/build-infrastructure/al-maintenance-codebuild-stack.yml
@@ -1,0 +1,57 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: A template that brings up a code build project that makes release configs for AL versions in maintenance mode
+
+Parameters:
+    GithubFullRepoName:
+      Type: String
+      Description: Full name of the agent repository to use (e.g. https://github.com/aws/amazon-ecs-agent.git)
+    GithubBranchName:
+      Type: String
+      Description: Name of the base branch to use to build, PRs to which will trigger builds (e.g. refs/heads/AmazonLinuxLegacy)
+    GithubSourceBranchName:
+      Type: String
+      Description: Name of the source branch used for the PRs (e.g. refs/heads/al1-patch)
+    ReleaseArtifactsBucketS3Uri:
+      Type: String
+      Description: The URI of the bucket where things land at the end, this is assumed to already exist (e.g. s3://artifacts)
+    ServiceRoleArn:
+      Type: String
+      Description: The ARN of an existing CodeBuild service role
+
+Resources:
+  ModifyAndCopyJSON:
+    Type: 'AWS::CodeBuild::Project'
+    Properties:
+      Artifacts:
+        Type: NO_ARTIFACTS
+      BadgeEnabled: false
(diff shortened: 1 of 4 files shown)


build-infrastructure/al-maintenance-codebuild-stack.yml
```diff
@@ -0,0 +1,58 @@
...
+        ComputeType: BUILD_GENERAL1_SMALL
+        ImagePullCredentialsType: CODEBUILD
+        Image: aws/codebuild/amazonlinux2-x86_64-standard:3.0
+        EnvironmentVariables:
+          - Name: RESULTS_BUCKET_URI
+            Type: PLAINTEXT
+            Value: !Ref ReleaseArtifactsBucketS3Uri
+          - Name: GITHUB_SOURCE_BRANCH_NAME
+              Type: PLAINTEXT
+              Value: !Ref GithubSourceBranchName
```
amogh09, line 41: Indentation is a bit off here.

buildspecs/al-maintenance-release-config.yml
```diff
@@ -0,0 +1,47 @@
+version: 0.2
+
+phases:
```
amogh09, line 3: Have we tested this script?

build-infrastructure/al-maintenance-codebuild-stack.yml
```diff
@@ -0,0 +1,58 @@
...
+      ServiceRole: !Ref ServiceRoleArn
+      Source:
+        BuildSpec: buildspecs/al-maintenance-release-config.yml
+        Location: !Ref GithubFullRepoName
+        Type: GITHUB
+      TimeoutInMinutes: 60
+      Triggers:
+        BuildType: BUILD
+        FilterGroups:
+          - - Type: EVENT
```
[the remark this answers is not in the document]
mye956: Good question, that's the correct syntax for CodeBuild triggers via CF. See also our existing CodeBuild project for all dogs.
https://github.com/aws/amazon-ecs-agent/blob/master/build-infrastructure/codebuild-devbuild-stack.yml#L56

build-infrastructure/al-maintenance-codebuild-stack.yml
mye956: ACK, good catch!

buildspecs/al-maintenance-release-config.yml
mye956: Yep, I was able to build a CodeBuild project along with this buildspec file within my personal dev account. Happy to showcase it offline!

buildspecs/al-maintenance-release-config.yml
```diff
@@ -0,0 +1,47 @@
...
+  build:
+    commands:
+      - GITHUB_DEST_BRANCH_NAME=$(echo ${CODEBUILD_WEBHOOK_BASE_REF} | cut -d "/" -f3-)
```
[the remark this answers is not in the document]
YashdalfTheGray: I think what Amogh is asking is if we can put all this in a shell script and then just run the script in spec file rather than inline all of these commands in the spec file. The only thing to note about that is that Codebuild runs all of the commands in their own `sh` instances (conceptually) and the shell script invocation would be all the commands in one instance. 

So if that works for us, we can make the change.

buildspecs/al-maintenance-release-config.yml
```diff
@@ -0,0 +1,47 @@
...
+            "gitFarmRepoName": "MadisonContainerAgentExternal",
+            "gitHubRepoName": "aws/amazon-ecs-agent",
+            "gitFarmStageBranch": "v${AGENT_VERSION}-stage",
+            "githubReleaseUrl": ""
+          }
+        }
+        EOF
+
+      # Downloading the agentVersionV2-<branch>.json (it is assumed that it already exists)
+      - aws s3 cp ${RESULTS_BUCKET_URI}/agentVersionV2/agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}.json ./agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}.json
```
YashdalfTheGray, line 32: Something that I've just realized that we haven't handled in any of the other scripts either is updating these binaries to the latest version (`aws` and `jq`). The potential problem we could run into here is if a new feature gets introduced in the AWS CLI and we haven't picked the latest version but use the feature. Edge case for sure, but what do you think about including a `yum update` as an `install` or `pre_build` step?

buildspecs/al-maintenance-release-config.yml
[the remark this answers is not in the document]
mye956: Oh I see. Yeah that's something we can for sure test out.

buildspecs/al-maintenance-release-config.yml
mye956: Yeah that's a really good point. I'll add that.

buildspecs/al-maintenance-release-config.yml
mye956: Here's the corresponding working branch in my forked repo that's testing these changes. Currently the CodeBuild project will be kicked off when this branch is merged into another branch called `AmazonLiuxLegacyTest`. When these changes are merged, it'll be the `AmazonLinuxLegacy` branch instead and we'll create a "dev" branch (e.g. al1-patch).

https://github.com/mye956/amazon-ecs-agent/tree/al1-patch-dummy

buildspecs/al-maintenance-release-config.yml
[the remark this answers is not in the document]
mye956: Just tested this out in my forked repo and personal dev account. It seems to be working with the changes I made.

buildspecs/al-maintenance-release-config.yml
mye956: Added a pre_build stage that'll run yum update.

YashdalfTheGray (dismissed):

build-infrastructure/al-maintenance-codebuild-stack.yml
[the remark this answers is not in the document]
amogh09: I see. Thanks both for explaining!

scripts/release-config.sh
```diff
@@ -0,0 +1,50 @@
...
+echo "Downloaded agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}.json"
+cat agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}.json
+
+# Grabbing the new current and release agent version
+CURR_VER=$(jq -r '.agentReleaseVersion' agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}.json)
+
+# Checking to see if there's a new release agent version to be released
+if [[ ! $AGENT_VERSION =~ "${CURR_VER}" ]] ; then
+    # Updating the agentVersionV2-<branch>.json file with new current and release agent versions and copying it to a temp file
+    cat <<< $(jq '.agentReleaseVersion = '\"$AGENT_VERSION\"' | .agentCurrentVersion = '\"$CURR_VER\"'' agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}.json) > agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}-COPY.json
```
amogh09, line 38: Is this command the same as 
```
jq ".agentReleaseVersion = \"$AGENT_VERSION\" | .agentCurrentVersion = \"$CURR_VER\"" agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}.json > agentVersionV2-${GITHUB_SOURCE_BRANCH_NAME}-COPY.json
```
?

If so, I think the version in this comment reads better?

buildspecs/al-maintenance-release-config.yml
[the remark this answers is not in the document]
amogh09: Thanks @mye956 ! Easier to read now!

scripts/release-config.sh
amogh09: Thanks!

YashdalfTheGray (approved):

PR #3615, opened 2023-03-22
Update the default option for `ECS_DISABLE_METRICS` on Windows to match the code

YashdalfTheGray:
### Summary
There was an error in the README file where we had specified the default `true` on Windows for `ECS_DISABLE_METRICS`. I looked in the code and it doesn't seem like we actually set it to `true`, i.e. actually disable metrics collection on Windows. 

Just updated the README file to match. 

fix: changed disable metrics default for Windows in the README to false

### Licensing


ubhattacharjya (approved):

rajal-amzn (approved):

rajal-amzn (commented):
We would also need to update our public docs.

rawahars:
So we do set this as `true` by default during ECS bootstrap on the AMIs.

SreeeS (approved):

PR #3630, opened 2023-03-31
ecs-init: Add ISO partition for downloading agent appropriately

singholt:
### Summary

This PR adds the ISO partition ID and the region to the `partitionBucketRegion` mapping. ECS init goes looking for where to download ecs agent from, but fails for ISO regions. With this change, it should now be able to look in the right place.

### Implementation details

We use the identifier defined in aws-sdk: https://github.com/aws/amazon-ecs-agent/blob/7060c3f945eaf3c24dc544dc8c204c36272aa652/ecs-init/vendor/github.com/aws/aws-sdk-go/aws/endpoints/defaults.go#L10-L16

### Testing

Fix: Add ISO partition for downloading agent appropriately
 
### Licensing

### DIFF
--- a/ecs-init/config/common.go
+++ b/ecs-init/config/common.go
@@ -112,6 +112,8 @@ var partitionBucketRegion = map[string]string{
 	endpoints.AwsPartitionID:      endpoints.UsEast1RegionID,
 	endpoints.AwsCnPartitionID:    endpoints.CnNorth1RegionID,
 	endpoints.AwsUsGovPartitionID: endpoints.UsGovWest1RegionID,
+	endpoints.AwsIsoPartitionID:   endpoints.UsIsoEast1RegionID,
+	endpoints.AwsIsoBPartitionID:  endpoints.UsIsobEast1RegionID,
 }
 
 // goarch is an injectable GOARCH runtime string. This controls the


Realmonia:
What's the use case of this change?

Realmonia (approved):

mythri-garaga (approved):

singholt:
> What's the use case of this change? (specifically when does this happen?)
> 
> > ECS init goes looking for where to download ecs agent from, but fails for ISO regions.
> 
> Nvm I see the thread on slack. Can we also make sure the S3 url in ISO is the same as that of commercial regions?

yes I'll work with the engineer assisting us to test this and verify before merging.

YashdalfTheGray (approved):
These exist in the SDK, so we're good to go here.

singholt:
> > What's the use case of this change? (specifically when does this happen?)
> > > ECS init goes looking for where to download ecs agent from, but fails for ISO regions.
> > 
> > 
> > Nvm I see the thread on slack. Can we also make sure the S3 url in ISO is the same as that of commercial regions?
> 
> yes I'll work with the engineer assisting us to test this and verify before merging.

verified this fixes the error (logged as a warning in ecs init logs). Merging.

PR #3646, opened 2023-04-20
Move statsEngine initiation from tcs session initialization, and adding channels to statsEngine

Realmonia:
### Summary

* Move statsEngine initiation out of telemetry session initiation. This helps decoupling statsEngine and telemetry session initiation, make it possible to use a different TACS client interface, as well as establishing channel based communication between statsEngine and TACS Client
* Add task metrics/health metrics/instance status channel
* Updating related metadata

### Testing

`make release` passed. This is pure metadata changes and not able to be functional tested.

Specifically, `telemetryChannelDefaultBufferSize = 15` will be further tested after channel based communication is implemented. This is currently borrowed from Fargate agent.

Move statsEngine initiation from tcs session initialization, and adding channels to statsEngine

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -59,6 +59,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/stats"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
 	tcshandler "github.com/aws/amazon-ecs-agent/agent/tcs/handler"
+	"github.com/aws/amazon-ecs-agent/agent/tcs/model/ecstcs"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/aws/amazon-ecs-agent/agent/utils/loader"
 	"github.com/aws/amazon-ecs-agent/agent/utils/mobypkgwrapper"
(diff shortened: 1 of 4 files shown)


agent/app/agent.go
```diff
@@ -97,6 +98,12 @@ const (
 	inServiceState                 = "InService"
 	asgLifecyclePollWait           = time.Minute
 	asgLifecyclePollMax            = 120 // given each poll cycle waits for about a minute, this gives 2-3 hours before timing out
+
+	// the size of the buffer for the metric/task health/instance health telemetry channels.
+	// By default, TCS (or TACS) will reject metrics that are older than 5 minutes (https://tiny.amazon.com/p6t8axgr).
```
fierlion, line 103: (remove internal link plz)

agent/app/agent.go
```diff
@@ -97,6 +98,12 @@ const (
...
+	// Since our metrics collection interval is currently set to 20 seconds, setting a buffer size of 15 allows us to
+	// store exactly 5 minutes of metrics in these buffers in the case where we temporarily lose connect to TCS.
```
fierlion, line 105: Is this comment fargate-specific?  I believe our ECS EC2 metrics share this channel across all running tasks on the host, so if we have 15 tasks, that will probably fill the buffer faster than 5 minutes.

agent/app/agent.go
Realmonia: Whoops will do. thanks

agent/app/agent.go
Realmonia: Yeah good point! I should have removed it - I was thinking of tuning this value after the channels are established.

agent/tcs/handler/handler.go
```diff
@@ -134,7 +127,7 @@ func startSession(
 	}
 	seelog.Info("Connected to TCS endpoint")
 	// start a timer and listens for tcs heartbeats/acks. The timer is reset when
-	// we receive a heartbeat from the server or when a publish metrics message
+	// we receive a heartbeat from the server or when a published metrics message
```
ubhattacharjya, line 130: qq: Would we want something similar for heartbeat message?

Realmonia: Don't think it is in scope of this PR, and I think we are planning to remove heartbeat message. Send you an update offline.

ubhattacharjya: makes sense. Thanks Heming.

ubhattacharjya (approved):

mye956 (approved):

PR #3651, opened 2023-04-21
Add make targets for ecs-agent module and integrate it with CI

amogh09:
### Summary
Update make targets to include ecs-agent module. Integrate ecs-agent module with CI.

### Implementation details

* Added a simple utility function with tests to ecs-agent module.
* Updated existing make targets for agent module to include ecs-agent module. 
    * `make test` now runs tests for ecs-agent too, and outputs coverage data to `coverprofile-ecs-agent.out` file.
    * `make analyze-cover-profile` now analyzes test coverage for ecs-agent module too.
    * `make clean` now cleans up coverage data file for ecs-agent tests too.
    * `make static-check` now runs static checks for ecs-agent module too.
* The test coverage analysis script `scripts/analyze-cover-profile-init˙ is a duplicate of `scripts/analyze-cover-profile`. Removed the former and updated the latter to take test coverage file as input. Updated static check CI workflow to pass the coverage file to the script.
* Updated Windows tests workflow to run tests for ecs-agent module too.

### Testing

Checked CI workflow logs that tests are run for ecs-agent module. Also checked that static-check is run for ecs-agent module in the CI workflow logs.

Integrate ecs-agent module with CI. Add make targets for ecs-agent module.

### Licensing


danehlim (commented):
Could we update the `make clean` target to remove the coverprofile-ecs-agent.out file as well?

Makefile
```diff
@@ -153,6 +153,10 @@ test-init:
 	go test -count=1 -short -v -coverprofile cover.out ./ecs-init/...
 	go tool cover -func cover.out > coverprofile-init.out
 
+test-ecs-agent:
```
danehlim, line 156: nit (non-blocking): Should we consider just running these as part of the `test` target? I think wanting to run tests in both agent/ and ecs-agent/ would be more common than wanting to run them separately. In automatic CI workflows, this doesn't really matter, but for those who manually run these targets, running two commands seems less preferable than just running a single one. This also follows the existing pattern of the `gogenerate` targets (i.e., `make gogenerate` runs for agent/ and ecs-agent/ and there is no `make gogenerate-ecs-agent`).

danehlim (dismissed):

Makefile
amogh09: Yeah I like that idea. I will update.

danehlim (dismissed):

danehlim (commented):
Thanks for making these changes!

Not blocking to this PR, but I think we'll want to revisit outstanding make targets that contain `./agent/...` and see if we want to include `./ecs-agent/...` as well (e.g., `make benchmark-test`). Let's make sure we're tracking this somewhere internally.

danehlim (approved):

amogh09:
Thanks! Yeah I think we would want to look at the other targets at a case-by-case basis. For example, I don't see `run-sudo-tests` and `run-sudo-unit-tests` being used in any of the automation in the Agent repo, so I am not sure where they are actually used. Not sure if `benchmark-test` applies to ecs-agent module either. We should talk about these separately.

amogh09:
Thanks! Yeah I think we would want to look at the other targets on a case-by-case basis and update the ones we think are applicable to ecs-agent module.

singholt (approved):

PR #3654, opened 2023-04-25
Add metrics interface and corresponding no-ops to ecs-agent/

ohsoo:
### Summary

Moving the metrics-related interfaces as well as their no-op implementations to the `ecs-agent` directory. No-op metrics will be integrated as a part of incoming efforts. 

### Implementation details

- `EntryFactory` and `Entry` interfaces included in the metrics package.
- `nopEntryFactory` is the no-op implementation of the `EntryFactory` interface.
- `nopEntry` is the no-op implementation of the `Entry` interface.

### Testing

Existing unit and integ tests.
The changes do not involve any integration to cause changes to the Agent. The no-op implementations are too low-level and do not have observable side effects to add unit tests for.

Add metrics interface and corresponding no-ops to ecs-agent/

### Licensing

### DIFF
--- a/ecs-agent/metrics/metrics.go
+++ b/ecs-agent/metrics/metrics.go
@@ -1,0 +1,63 @@
+package metrics
+
+// EntryFactory specifies the factory interface for creating new Metric Entry objects.
+type EntryFactory interface {
+	New(op string) Entry
+	Flush()
+}
+
+// Entry specifies methods that need to be implemented by a metrics entry object.
+type Entry interface {
+	// WithFields allows the caller to set metric metadata. Once an entry object is
+	// created (typically using a meaningful metrics name), its properties can be
+	// manipulated using WithFields. Example:
+	//
+	// m := m.WithFields(map[string]interface{"timestamp": time.Now()})
+	WithFields(f map[string]interface{}) Entry
+	// WithCount allows the caller to set metric counts. This is useful for determining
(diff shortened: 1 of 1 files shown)


amogh09 (approved):

ecs-agent/metrics/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (*nopEntryFactory) New(op string) Entry {
+	return newNopEntry(op)
+}
+
+func (f *nopEntryFactory) Flush() {}
+
+type nopEntry struct{}
+
+func newNopEntry(op string) Entry {
```
fierlion, line 56: nit -- I see this elsewhere but I think we mean `noOpEntry` here? aka `newNoOpEntry`

fierlion (approved):
approving with a nit -- if you have a chance to address this in a future revision, please do.

ecs-agent/metrics/metrics.go
fierlion: or are we thinking of it this way: https://en.wikipedia.org/wiki/NOP_(code) ?

danehlim (approved):

ecs-agent/metrics/metrics.go
ohsoo: Yep, that's the idea (nop/noop/no-op) thanks for the review!

PR #3643, opened 2023-04-14
Add ecs-agent/ ACS model + cleanup gogenerate make

danehlim:
### Summary

Create `ecs-agent/` directory, and add ACS model and gogenerate to that directory.

Update/clean up existing gogenerate make targets in `Makefile`. 
Prior to this change:
- `make gogenerate` does not actually run `go generate` at all
- `make gogenerate-init` runs `go generate` for both `agent/` and `ecs-init/` directories

### Implementation details

1. Add ACS model and gogenerate to `ecs-agent/`
   - Copy `agent/acs/model` and `agent/gogenerate` to `ecs-agent/acs/model` and `ecs-agent/gogenerate` 
   - Go into `ecs-agent` directory and run `go mod init` and `go mod tidy`
   - Modify versions in `ecs-agent/go.mod` to match those in `agent/go.mod`
   - Run `go mod tidy` and `go mod vendor`
   - Generate new ACS JSON model file per ECS Agent standard operation procedure that is in sync with up to date source of truth provided by control plane, and then replace `agent/acs/model/api/api-2.json` and `ecs-agent/acs/model/api/api-2.json`with the newly generated file
   - Run `make get-deps` and `make gogenerate` (dependent on below implementation detail). This replaces `agent/acs/model/ecsacs/api.go` and `ecs-agent/acs/model/ecsacs/api.go` with newly generated files
2. Clean up gogenerate make targets
   - Update `make gogenerate-init` target to run `go generate` only for `ecs-init/` directory
   - Update `make gogenerate` target to actually run `go generate`, and run it for `agent/` and `ecs-agent/` directories

### Testing

Unit, integration, and functional tests.
Manually test modified make targets locally with `make` command.

### Description for the changelog

Add ecs-agent/ ACS model + cleanup gogenerate make

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3659, opened 2023-04-26
Use ecs-agent/acs/model and ecs-agent/gogenerate

danehlim:
### Summary

Use `ecs-agent/acs/model` in favor of `agent/acs/model` and `ecs-agent/gogenerate` in favor of `agent/gogenerate`. This is a followup to https://github.com/aws/amazon-ecs-agent/pull/3643.

### Implementation details

1. Use `ecs-agent/acs/model` instead of `agent/acs/model`
   - In `agent/` directory, replace all imports of `github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs` with `github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs`
   - Resort imports in affected files
   - Delete `agent/acs/model` directory and all of its files
2. Use `ecs-agent/gogenerate` over `agent/gogenerate`
   - In `agent/` directory, modify all go generate commands using `../../gogenerate/awssdk.go` to use `../../../ecs-agent/gogenerate/awssdk.go` instead
   - Delete `agent/gogenerate` directory and all of its files
   - Vendor `github.com/aws/aws-sdk-go/private/model/api`, `github.com/aws/aws-sdk-go/private/util`, and `golang.org/x/tools/imports` packages explicitly in `agent/` since we need these to generate code as part of `make gogenerate`, but they are not automatically vendored after deleting `agent/gogenerate` directory and all of its files (i.e., the previous step).
     - We do this by importing them in  `agent/tools.go`, which is the best practice for managing tool dependencies: https://github.com/golang/go/issues/25922#issuecomment-402918061. This is also explained in more detail in the previous pull request where we introduced the `agent/tools.go` file: https://github.com/aws/amazon-ecs-agent/pull/2890.
   - Run `go mod tidy` and `go mod vendor` in `agent/` directory

### Testing

Manually ran `make gogenerate` to ensure it still succeeds after this change.

Use ecs-agent/acs/model and ecs-agent/gogenerate

### Licensing


agent/tools.go
```diff
@@ -17,5 +17,8 @@
 package tools
```
chienhanlin, line 18: Thanks for sharing all details on the PR overview! I was wondering if we should add a comment for this package to help people maintain tools/dependencies in the future.

```
// The tools package manages tool dependency with go mod.
// For tools that we need to generate code as part of `make gogenerate`, but are not automatically vendored, 
// we need to explicitly added them to this dummy tools.go file.
// 
// To vend a new tool, import the tool/package in this dummy tools.go file, and then rerun `go mod tidy` and `go mod vendor`. 
```

chienhanlin (approved):

YashdalfTheGray (approved):
Tests are passing on this, the change itself looks alright. Good use of the `tools` file to prevent packages being cleaned up when running `go mod tidy`.

agent/tools.go
danehlim: Great idea, thanks for suggesting! I will take an action to add this in a future/followup PR.

PR #3665, opened 2023-04-28
Add task manifest/stop verification ACK common interface to ecs-agent/

ohsoo:
### Summary

Add a common interface for task manifest/stop verification ACK responders to `ecs-agent/`

### Implementation details

- Create interface inside of `ecs-agent/acs/session/task_manifest_common.go`

### Testing

Add task manifest/stop verification ACK common interface ecs-agent/

### Licensing

### DIFF
--- a/ecs-agent/acs/session/task_manifest_common.go
+++ b/ecs-agent/acs/session/task_manifest_common.go
@@ -1,0 +1,9 @@
+package session
+
+// ManifestMessageIDAccessor stores the latest message ID that corresponds to the
+// most recently processed task manifest message and is used to determine the
+// validity of a task stop verification ack message.
+type ManifestMessageIDAccessor interface {
+	GetMessageID() string
+	SetMessageID(messageID string) error
+}


ecs-agent/acs/session/task_manifest_common.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+type TaskGetter interface {
+	GetRunningTasksOnInstance() ([]string, error)
+}
+
+type TaskStopper interface {
+	StopTask(taskARN string)
+}
+
```
danehlim, line 34: Other than the `ManifestMessageIDAccessor` interface, will these interfaces need to be used in both handling / responding to task manifest and task stop verification ACK messages from ACS? I was under the impression that `ManifestMessageIDAccessor` is the only interface that will be common to both, while `SequenceNumberAccessor` and `TaskGetter` are only needed for the former and `TaskStopper` is only needed for the latter.

ecs-agent/go.mod
```diff
@@ -3,18 +3,35 @@ module github.com/aws/amazon-ecs-agent/ecs-agent
 go 1.19
 
 require (
+	github.com/aws/amazon-ecs-agent/agent v0.0.0-20230426070318-0202ad238404
```
danehlim, line 6: `agent` module depends on `ecs-agent` module - having `ecs-agent` depend on `agent` doesn't seem ideal. Can we avoid this dependency cycle?

ecs-agent/acs/session/task_manifest_common.go
```diff
@@ -0,0 +1,73 @@
...
+// CompareTasks returns a list of the running tasks that need to be stopped after comparing
+// the list of tasks running on the instance to the list of tasks on the task manifest,
+// which is the source of truth about what tasks are supposed to be running.
+func CompareTasks(
```
danehlim, line 38: This function seems like it is only needed for handling/responding to task manifest messages from ACS - IIUC it is not used for handling / responding to task manifest and task stop verification ACK messages.

ecs-agent/acs/session/task_manifest_common.go
ohsoo: You're right. I just placed them all in this one file but I can place them accordingly in two files: `task_manifest_common.go` and` task_verification_ack_common.go`, would that address your point about the separations?

danehlim: I had imagined there to be 3 files:

1. File for functionality common to handling task manifest and task stop verification ACK messages from ACS contains`ManifestMessageIDAccessor` interface.

2. File for functionality specific to handling task manifest messages from ACS contains `SequenceNumberAccessor` and `TaskGetter` interfaces.

3. File for functionality specific to handling task manifest messages from ACS contains  `TaskStopper` interface.

The first file would be in the scope of this PR and the other two files could follow in subsequent PRs.

danehlim (approved):

RichaGangwar (approved):
Looks good to me.

PR #3666, opened 2023-04-28
Require an HTTP Handler instead of a mux Router in TMDS Config

amogh09:
### Summary
TMDS initialization function in `ecs-agent` module currently requires a `Router` instance from gorilla/mux library to be passed. Usage of `Router` here is unnecessarily strict and limits the reusability of TMDS to consumers who are using routers from gorilla/mux library. Updating the function to require an `http.Handler` instance instead which would allow greater reuse. `Router` implements `http.Handler` interface.

### Implementation details
* Change `router *mux.Router` field in TMDS initialization config to `handler http.Handler`.
* Update the tests and error messages accordingly.

### Testing

Require an HTTP Handler instead of a mux Router in TMDS Config to increase the reusability of TMDS.

### Licensing

### DIFF
--- a/ecs-agent/tmds/server.go
+++ b/ecs-agent/tmds/server.go
@@ -45,7 +45,7 @@ type Config struct {
 	writeTimeout    time.Duration // http server write timeout
 	steadyStateRate float64       // steady request rate limit
 	burstRate       int           // burst request rate limit
-	router          *mux.Router   // router with routes configured
+	handler         http.Handler  // HTTP handler with routes configured
 }
 
 // Function type for updating TMDS config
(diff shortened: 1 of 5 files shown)


danehlim (approved):

prateekchaudhry (approved):

ecs-agent/tmds/server.go
```diff
@@ -45,7 +45,7 @@ type Config struct {
 	writeTimeout    time.Duration // http server write timeout
 	steadyStateRate float64       // steady request rate limit
 	burstRate       int           // burst request rate limit
-	router          *mux.Router   // router with routes configured
+	handler         http.Handler  // HTTP handler with routes configured
```
prateekchaudhry, line 48: clarification: is changing from pointer (*mux.Router) to http.Handler intentional?

amogh09: Yeah because `http.Handler` is an `interface` whereas `mux.Router` is a `struct`.

PR #3668, opened 2023-05-01
Advertise Domainless gMSA capability on Windows

arun-annamalai:
### Summary

This PR is advertises the new domainless gmsa capability on the Windows ECS agent. The Linux agent will be coming in a followup PR. 

### Implementation details

The Windows agent will advertise the gmsa capability only under the following 3 conditions

1. Environment variable is set 
2. The windows instance must not be 2016
3. The domainless gMSA plugin must be installed 

### Testing

This change was unit tested as well as tested on a Windows EC2 instance and describe-container-instance was called to ensure that the capability was correctly advertised. It was also run on an old Windows EC2 instance to ensure the capability was not advertised. 

Advertise gmsa domainless capability on Windows Agent
### Licensing

### DIFF
--- a/agent/config/parse_windows.go
+++ b/agent/config/parse_windows.go
@@ -24,7 +24,9 @@ import (
 	"unsafe"
 
 	"github.com/aws/amazon-ecs-agent/agent/utils"
+
 	"github.com/cihub/seelog"
+	"golang.org/x/sys/windows/registry"
 )
 
 const (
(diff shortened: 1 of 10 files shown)


agent/config/parse_windows.go
```diff
@@ -32,6 +34,11 @@ const (
 	// to skip the windows server version check. This is useful for testing and
 	// should not be set for any non-test use-case.
 	envSkipWindowsServerVersionCheck = "ZZZ_SKIP_WINDOWS_SERVER_VERSION_CHECK_NOT_SUPPORTED_IN_PRODUCTION"
+	gmsaPluginGUID                   = "{859E1386-BDB4-49E8-85C7-3070B13920E1}"
```
yinyic, line 37: Do we ever expect this GUID to change (for example after we upgrade the plugin version)

arun-annamalai: Even if the plugin version upgrades, we do not expect this GUID to change. For example, we have already changed the plugin twice and kept the GUID constant. The input to the plugin is a JSON field that is deserialized, so the plugin is also extensible and backwards compatible. 


We have published this GUID to customers as well (for the EKS domainless gMSA already released), so changing its GUID would be breaking behavior.

singholt (approved):

yinyic (approved):

agent/config/parse_windows.go
```diff
@@ -114,3 +141,42 @@ var IsWindows2016 = func() (bool, error) {
 
 	return isWS2016, nil
 }
+
+var queryDomainlessGmsaPluginSubKeys = func() ([]string, error) {
+	k, err := registry.OpenKey(registry.LOCAL_MACHINE, `SYSTEM\CurrentControlSet\Control\CCG\COMClasses\`, registry.READ)
```
rawahars, line 146: We should make the registry keys as constants.

agent/config/parse_windows.go
```diff
@@ -114,3 +141,42 @@ var IsWindows2016 = func() (bool, error) {
...
+		return nil, err
+	}
+
+	seelog.Debugf("gMSA Subkeys are %+v", subKeys)
+	return subKeys, nil
+}
+
+// This function queries all gmsa plugin subkeys to check whether the Amazon ECS Plugin GUID is present.
+func isDomainlessGmsaPluginInstalled() (bool, error) {
+	subKeys, err := fnQueryDomainlessGmsaPluginSubKeys()
```
rawahars, line 169: Should we also check for the presence of GMSA binary?

agent/config/parse_windows.go
arun-annamalai: Can I do this as a followup in a future cleanup PR? Sai is blocked on this work presently

agent/config/parse_windows.go
arun-annamalai: I had asked Justin about this, and he had only wanted to check for the presence of registry key (which would imply plugin is installed)

Issue #3270, opened 2022-06-28
ECS agent unable to mount a fsx window volume that contains a quotation mark in password


timothyaws:
### Summary

Using FSX for Windows as a Volume, the password cannot have a quotation(") mark in the password otherwise the agent fails when trying to mount the volume.

### Description

When passing a password to join the domain that has a quotation(") mark via secret manager. The ECS agent will be unable to mount the volume and cause the task to exit code 1.

This error message is generated by the command at https://github.com/aws/amazon-ecs-agent/blob/a5f8d25f66fbf4f8839d5e7e5fcbc635a4d75c5e/agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go#L557-L572

### Expected Behavior
The agent should parse and escape any additional quotation marks in the password input before trying to mount the volume

### Observed Behavior
The agent passed the password with quotation mark through resulting in the mount command failing.

### Supporting Log Snippets

```
level=error time=2022-06-28T00:56:15Z msg="Failed to map fsxwindowsfileserver resource on the container instance error: exit status 1, out: The string is missing the terminator: '.\r\n    + CategoryInfo          : ParserError: (:) [], ParentContainsErrorRecordException\r\n    + FullyQualifiedErrorId : TerminatorExpectedAtEndOfString\r\n \r\n" module=fsxwindowsfileserver_windows.go
level=info time=2022-06-28T00:56:15Z msg="Error transitioning resource" resource="fsx_folder" failedStatus="CREATED" error="exit status 1" task="task_id"
```

jterry75:
Bug understood, fix PR: #3296

fierlion:
note: fix pending update of related PR

rawahars:
Closing this issue since the fix was released in `v1.71.2`.

PR #3669, opened 2023-05-03
windows: support special characters in the password for FSx

rawahars:
### Summary
Presently, When a password contains characters that contain escape sequences such as a quotation mark PowerShell fails the command. Reference: https://github.com/aws/amazon-ecs-agent/issues/3270

To mitigate the same and support handling of special characters, we are making the change to use single quoted strings. This means that the password is passed-on to Powershell as it is without any substitutions. Reference: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-5.1#single-quoted-strings

### Implementation details

Instead of using double quotes, we are using single quotes which means that Powershell would be passed-on a value without any substitution.

### Testing

Created an ECS instance with the custom agent. We used a password which had many special characters.
Tested that the task with FSx volume was running successfully on the instance,

New tests cover the changes: 
Yes

windows: support special characters in the password for FSx
### Licensing

### DIFF
--- a/agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
+++ b/agent/taskresource/fsxwindowsfileserver/fsxwindowsfileserver_windows.go
@@ -45,6 +45,7 @@ import (
 )
 
 const (
+	psCredentialCommandFormat = "$(New-Object System.Management.Automation.PSCredential('%s', $(ConvertTo-SecureString '%s' -AsPlainText -Force)))"
 	resourceProvisioningError = "VolumeError: Agent could not create task's volume resources"
 )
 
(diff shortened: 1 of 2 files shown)


arun-annamalai (approved):

singholt (dismissed):

rawahars:
Please do not merge yet. I realised that these changes work for special chars but not for `"`. Let me look into that more to figure out the fix.

rawahars:
This PR is ready for review now.

We have added the support for special characters including `'`. 

However, when using Secrets Manager, the password cannot have any escape characters, more specifically the JSON representation cannot have any escape characters. For example- `Microsoft\"1234` is not a valid supported password when using ASM. The reason for this limitation is that ASM stores these key-pairs as a JSON string and therefore, it escapes the entire sequence before sending it in the response. When we unmarshall the same in agent, we are still left with the original escapes in the password. This leads to incorrect password while authentication.
For example-
Consider that we are storing password `Microsoft\"1234` and user `Admin` in ASM. The string representation of the same would be- `"{\"username\":\"Admin\",\"password\":\"Microsoft\\\\"1234\"}"` and this would be the result from `GetSecretValue` API response. Now during our unmarshalling, we would get password as `Microsoft\"1234` which is not the same as the actual password.

This unfortunately is the limitation of Secrets Manager and therefore, any username or password which has escape characters in json representation cannot be used here.

@singholt @jterry75 @arun-annamalai

arun-annamalai:
Can we work with doc writers to state this limitation somewhere on public docs so customers dont have future issues with this?

arun-annamalai (approved):

jterry75:
Hang on dont merge this

rawahars:
@arun-annamalai Our fix works for all the special characters including `"`. I encountered a corner case when using CLI for storing the password. I have updated the comment above to reflect the same.

jterry75 (approved):

singholt (approved):

YashdalfTheGray (approved):
Looks good, thanks for making this change!

PR #3678, opened 2023-05-08
Domainless gMSA linux capability and gRPC endpoints

saikiranakula-amzn:
### Summary
This PR adds the agent advertisement capability for domainless gMSA support on linux and also updated the gRPC endpoints for domainless communication with credentials-fetcher daemon

### Implementation details
1. Added gMSA linux domainless capability
2. gRPC endpoint

### Testing
This was tested through unit tests, as well as manually running the agent on the instance and also verifying the agent capability for domainless gMSA support.

New tests cover the changes: 

### Licensing


arun-annamalai (approved):

amogh09:
Hello. Can you please comment out `Note for external contributors:` part in the PR description and also add a change log?

agent/taskresource/grpcclient/credentialsfetcherclient.go
```diff
@@ -90,6 +93,71 @@ func (c CredentialsFetcherClient) AddKerberosLease(ctx context.Context, credenti
 	return credentialsFetcherResponse, nil
 }
 
+// AddNonDomainJoinedKerberosLease() invokes credentials fetcher daemon running on the host
+// to create kerberos tickets associated with gMSA accounts in domainless mode
+func (c CredentialsFetcherClient) AddNonDomainJoinedKerberosLease(ctx context.Context, credentialspecs []string, username string, password string, domain string) (CredentialsFetcherResponse, error) {
+	if len(credentialspecs) == 0 {
+		seelog.Error("credentialspecs request should not be empty")
+		return CredentialsFetcherResponse{}, nil
```
amogh09, line 101: How about following the pattern of `RenewNonDomainJoinedKerberosLease` method and return an error here? Same for the other validation below. How would the consumer know that the method failed if we don't return an error?

agent/taskresource/grpcclient/credentialsfetcherclient.go
```diff
@@ -90,6 +93,71 @@ func (c CredentialsFetcherClient) AddKerberosLease(ctx context.Context, credenti
...
+	client := pb.NewCredentialsFetcherServiceClient(c.conn)
+
+	request := &pb.CreateNonDomainJoinedKerberosLeaseRequest{CredspecContents: credentialspecs, Username: username, Password: password, Domain: domain}
+
+	ctx, cancel := context.WithDeadline(ctx, time.Now().Add(c.timeout))
+	defer cancel()
+
+	response, err := client.AddNonDomainJoinedKerberosLease(ctx, request)
+	if err != nil {
+		seelog.Errorf("could not create kerberos tickets: %v", err)
```
amogh09, line 119: Should we add more context to this log to identify the request? Preferably with `logger` package for structured logging (the structured logger has an `Error` field that we can use for `err`). 

Also, since we are returning the error, how about the consumer log the error? I see that the consumer of `AddKerberosLease` method logs the returned error so maybe we can follow the same pattern.

Same comment for `RenewNonDomainJoinedKerberosLease` method.

mythri-garaga (approved):

agent/taskresource/grpcclient/credentialsfetcherclient.go
```diff
@@ -90,6 +93,71 @@ func (c CredentialsFetcherClient) AddKerberosLease(ctx context.Context, credenti
...
+		seelog.Error("credentialspecs request should not be empty")
+		return CredentialsFetcherResponse{}, nil
+	}
+
+	if len(username) == 0 || len(password) == 0 || len(domain) == 0 {
+		seelog.Error("username, password or domain should not be empty")
+		return CredentialsFetcherResponse{}, nil
+	}
+
+	defer c.conn.Close()
```
amogh09, line 109: Question - Since we are closing the connection in this method, the consumer of `CredentialsFetcherClient` needs to open the connection again before calling another method that uses the connection?

agent/taskresource/grpcclient/credentialsfetcherclient_test.go
```diff
@@ -111,6 +131,105 @@ func TestCredentialsFetcherClient_AddKerberosLease(t *testing.T) {
...
+			"testdomain",
+			CredentialsFetcherResponse{LeaseID: leaseid, KerberosTicketPaths: []string{"/var/credentials-fetcher/krbdir/123456/webapp01", "/var/credentials-fetcher/krbdir/123456/webapp02"}},
+		},
+	}
+
+	ctx := context.Background()
+
+	conn, err := grpc.DialContext(ctx, "", grpc.WithInsecure(), grpc.WithContextDialer(dialer()))
+	if err != nil {
+		log.Fatal(err)
```
amogh09, line 173: Can we please use `require.NoError(t, err)` here instead of `log.Fatal` (and also get rid of the if block)? The latter exits the process.

agent/taskresource/grpcclient/credentialsfetcherclient_test.go
```diff
@@ -111,6 +131,105 @@ func TestCredentialsFetcherClient_AddKerberosLease(t *testing.T) {
...
+	}
+	defer conn.Close()
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			response, _ := NewCredentialsFetcherClient(conn, time.Minute).AddNonDomainJoinedKerberosLease(context.Background(), tt.credspecContents, tt.username, tt.password, tt.domain)
```
amogh09, line 179: Can we please split this long line into shorter lines for better readability. 

```go
response, _ := NewCredentialsFetcherClient(conn, time.Minute)
    .AddNonDomainJoinedKerberosLease(
        context.Background(), tt.credspecContents, tt.username, tt.password, tt.domain,
    )
```

amogh09: Also let's not ignore the error returned by the `AddNonDomainJoinedKerberosLease` call. We are using an older `err` in the assertions below.

agent/taskresource/grpcclient/credentialsfetcherclient_test.go
```diff
@@ -111,6 +131,105 @@ func TestCredentialsFetcherClient_AddKerberosLease(t *testing.T) {
...
+			if response.LeaseID != tt.response.LeaseID {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+			}
```
amogh09, line 184: We should not branch the assertions based on the output of the function being tested which is `response` in this case. The assertions should be based on the test case alone.

How about adding an `expectedError string` field to the test cases and use it for branching the assertions? This way we can also assert on the errors returned by the function being tested. This is a follow-up to my earlier comment about making `AddNonDomainJoinedKerberosLease` method return an error is validation fails.

```go
if tt.expectedError != "" {
    assert.EqualError(t, err, tt.expectedError)
} else {
    assert.Equal(t, tt.response, response)
}
```

Same comment for `TestCredentialsFetcherClient_RenewNonDomainJoinedKerberosLease` below.

amogh09 (changes requested):

agent/config/parse_linux.go
```diff
@@ -72,7 +72,31 @@ func parseFSxWindowsFileServerCapability() BooleanDefaultFalse {
...
+		// this env variable is set in ecs-init module
+		if credentialsfetcherHostDir := os.Getenv("CREDENTIALS_FETCHER_HOST_DIR"); credentialsfetcherHostDir != "" {
+			_, err := os.Stat(credentialsfetcherHostDir)
+			if err != nil {
+				if os.IsNotExist(err) {
+					seelog.Errorf("CREDENTIALS_FETCHER_HOST_DIR not found, err: %v", err)
+					return BooleanDefaultFalse{Value: ExplicitlyDisabled}
+				}
+			}
+			return BooleanDefaultFalse{Value: ExplicitlyEnabled}
```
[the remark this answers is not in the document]
saikiranakula-amzn: there won't be any error other than dir check, since the socket check is already done during init phase

agent/taskresource/grpcclient/credentialsfetcherclient.go
saikiranakula-amzn: yes, you are right, before every call connection needs to be established again

agent/taskresource/grpcclient/credentialsfetcherclient.go
saikiranakula-amzn: sure will add it

agent/taskresource/grpcclient/credentialsfetcherclient_test.go
saikiranakula-amzn: will add the conditions accordingly

saikiranakula-amzn (commented):
will address the comments on the PR

agent/taskresource/grpcclient/credentialsfetcherclient.go
```diff
@@ -90,12 +93,77 @@ func (c CredentialsFetcherClient) AddKerberosLease(ctx context.Context, credenti
 	return credentialsFetcherResponse, nil
 }
 
+// AddNonDomainJoinedKerberosLease() invokes credentials fetcher daemon running on the host
+// to create kerberos tickets associated with gMSA accounts in domainless mode
+func (c CredentialsFetcherClient) AddNonDomainJoinedKerberosLease(ctx context.Context, credentialspecs []string, username string, password string, domain string) (CredentialsFetcherResponse, error) {
+	if len(credentialspecs) == 0 {
+		seelog.Error("credentialspecs request should not be empty")
```
amogh09, line 100: We can remove this error log now that we are returning an error?

agent/taskresource/grpcclient/credentialsfetcherclient_test.go
```diff
@@ -78,72 +99,180 @@ func TestCredentialsFetcherClient_AddKerberosLease(t *testing.T) {
...
+	conn, err := grpc.DialContext(ctx, "", grpc.WithInsecure(), grpc.WithContextDialer(dialer()))
+	require.NoError(t, err)
+	defer conn.Close()
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			response, err := NewCredentialsFetcherClient(conn, time.Minute).AddKerberosLease(context.Background(), tt.credspecContents)
+			if tt.expectedError != "" {
+				assert.Error(t, err)
+				assert.Equal(t, tt.expectedError, err.Error())
```
amogh09, line 129: Code will panic at `err.Error()` if `err` is `nil`. How about merging these two asserts into a single `assert.EqualError(t, err, tt.expectedError)`? That will perform both the assertions safely. Another option is to use `require.Error(t, err)` for the first assertion so that the test stops if that assertion fails. 

Same for `TestCredentialsFetcherClient_AddNonDomainJoinedKerberosLease` and `TestCredentialsFetcherClient_RenewNonDomainJoinedKerberosLease`.

amogh09 (approved):

mythri-garaga (approved):

PR #3684, opened 2023-05-09
Host Resource Manager initialization

Yiyuanzzz:
### Summary

Initialize host resource manager with available host resource values. Host resource manager will keep track of the available host resources as tasks start and stop. Resources include host port, CPU, memory, and GPU. New methods and tests will be followed up in later implementations.
### Implementation details

- Added a new `host_resource_manager.go` file in `engine` for host resource manager initialization.
- Fixed existing constructors and tests with new host resource manager being added.
- Ran `make gogenerate` to generate mocks in `api_mocks.go`
-  Keep the existing `getResources` private and added new `GetHostResources` method to interface.
- Use `exitcodes.ExitError` exit code when agent unable to fetch host resources, agent will retry if agent fail to get host resource, if agent keep failing, agent will not be able to register container instance, customer will have some idle EC2 instances if they don’t have a proper setup to terminate these instances, but this is the same risk with or without the Agent restart behavior though.

### Testing

Running the static checks within the PR.
`make release-agent`

N/A

### Licensing

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -306,6 +306,20 @@ func (client *APIECSClient) getResources() ([]*ecs.Resource, error) {
 	return []*ecs.Resource{&cpuResource, &memResource, &portResource, &udpPortResource}, nil
 }
 
+// GetHostResources calling getHostResources to get a list of CPU, MEMORY, PORTS and PORTS_UPD resources
+// and return a resourceMap that map the resource name to each resource
+func (client *APIECSClient) GetHostResources() (map[string]*ecs.Resource, error) {
+	resources, err := client.getResources()
+	if err != nil {
+		return nil, err
+	}
+	resourceMap := make(map[string]*ecs.Resource)
+	for _, resource := range resources {
+		resourceMap[*resource.Name] = resource
+	}
+	return resourceMap, nil
+}
+
 func getCpuAndMemory() (int64, int64) {
 	memInfo, err := system.ReadMemInfo()
 	mem := int64(0)
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -298,17 +298,36 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 			return exitcodes.ExitTerminal
 		}
 	}
+	hostResources, err := client.GetHostResources()
+	if err != nil {
+		seelog.Critical("Unable to fetch host resources")
+		return exitcodes.ExitError
+	}
+	numGPUs := int64(0)
 	if agent.cfg.GPUSupportEnabled {
 		err := agent.initializeGPUManager()
 		if err != nil {
 			seelog.Criticalf("Could not initialize Nvidia GPU Manager: %v", err)
 			return exitcodes.ExitError
 		}
+		// Find number of GPUs instance has
+		platformDevices := agent.getPlatformDevices()
+		for _, device := range platformDevices {
+			if *device.Type == ecs.PlatformDeviceTypeGpu {
+				numGPUs++
+			}
+		}
+	}
+
+	hostResources["GPU"] = &ecs.Resource{
+		Name:         utils.Strptr("GPU"),
+		Type:         utils.Strptr("INTEGER"),
+		IntegerValue: &numGPUs,
 	}
 
 	// Create the task engine
 	taskEngine, currentEC2InstanceID, err := agent.newTaskEngine(
-		containerChangeEventStream, credentialsManager, state, imageManager, execCmdMgr, agent.serviceconnectManager)
+		containerChangeEventStream, credentialsManager, state, imageManager, hostResources, execCmdMgr, agent.serviceconnectManager)
 	if err != nil {
 		seelog.Criticalf("Unable to initialize new task engine: %v", err)
 		return exitcodes.ExitTerminal
@@ -515,6 +534,7 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 	credentialsManager credentials.Manager,
 	state dockerstate.TaskEngineState,
 	imageManager engine.ImageManager,
+	hostResources map[string]*ecs.Resource,
 	execCmdMgr execcmd.Manager,
 	serviceConnectManager engineserviceconnect.Manager) (engine.TaskEngine, string, error) {
 
@@ -523,11 +543,11 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 	if !agent.cfg.Checkpoint.Enabled() {
 		seelog.Info("Checkpointing not enabled; a new container instance will be created each time the agent is run")
 		return engine.NewTaskEngine(agent.cfg, agent.dockerClient, credentialsManager,
-			containerChangeEventStream, imageManager, state,
+			containerChangeEventStream, imageManager, hostResources, state,
 			agent.metadataManager, agent.resourceFields, execCmdMgr, serviceConnectManager), "", nil
 	}
 
-	savedData, err := agent.loadData(containerChangeEventStream, credentialsManager, state, imageManager, execCmdMgr, serviceConnectManager)
+	savedData, err := agent.loadData(containerChangeEventStream, credentialsManager, state, imageManager, hostResources, execCmdMgr, serviceConnectManager)
 	if err != nil {
 		seelog.Criticalf("Error loading previously saved state: %v", err)
 		return nil, "", err
@@ -552,7 +572,7 @@ func (agent *ecsAgent) newTaskEngine(containerChangeEventStream *eventstream.Eve
 		state.Reset()
 		// Reset taskEngine; all the other values are still default
 		return engine.NewTaskEngine(agent.cfg, agent.dockerClient, credentialsManager,
-			containerChangeEventStream, imageManager, state, agent.metadataManager,
+			containerChangeEventStream, imageManager, hostResources, state, agent.metadataManager,
 			agent.resourceFields, execCmdMgr, serviceConnectManager), currentEC2InstanceID, nil
 	}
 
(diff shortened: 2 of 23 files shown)


prateekchaudhry (approved):

agent/engine/host_resource_manager.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/logger"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+)
+
+// HostResourceManager keeps account of each task in
+type HostResourceManager struct {
+	hostResource     map[string]*ecs.Resource
+	consumedResource map[string]*ecs.Resource
+
+	taskConsumed map[string]bool //task.arn to boolean whether host resources consumed or not
```
yinyic, line 29: Nit - if we are going to be using this map as a set (i.e. we only ever care about whether the key exists), we can use `map[string]struct{}` to indicate that the value doesn't matter. But current implementation is fine too.

agent/engine/host_resource_manager.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func NewHostResourceManager(resourceMap map[string]*ecs.Resource) HostResourceManager {
+	consumedResourceMap := make(map[string]*ecs.Resource)
+	taskConsumed := make(map[string]bool)
+	// assigns CPU, MEMORY, PORTS, PORTS_UDP from host
```
yinyic, line 35: Nit - is it possible to make `PORTS` -> `PORTS_TCP` just to be clear?

agent/engine/host_resource_manager.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
+	"github.com/aws/amazon-ecs-agent/agent/logger"
+	"github.com/aws/amazon-ecs-agent/agent/utils"
+)
+
+// HostResourceManager keeps account of each task in
+type HostResourceManager struct {
+	hostResource     map[string]*ecs.Resource
```
yinyic, line 26: Nit - maybe rename to `initialHostResource` if we plan to keep it immutable? 

Also can we add a comment for the variable to clarify that, for resources in `resourceMap`, some are "available resources" like CPU, mem, while some others are "reserved/consumed resources" like ports, so that readers don't get confused about the inconsistency.

yinyic (commented):
Thank you Yiyuan! The PR looks good overall.

agent/app/agent.go
```diff
@@ -298,17 +298,36 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+		// Find number of GPUs instance has
+		platformDevices := agent.getPlatformDevices()
+		for _, device := range platformDevices {
+			if *device.Type == ecs.PlatformDeviceTypeGpu {
+				numGPUs++
+			}
+		}
+	}
+
+	hostResources, err := client.GetHostResources()
```
chienhanlin, line 317: Should we move **line 317-321** before **line 301-315** to make it fail fast if agent cannot get host resources?
```
// Get host resources
hostResources, err := client.GetHostResources()
if err != nil {
    seelog.Critical("Unable to fetch host resources")
    return exitcodes.ExitError
}
```
```
// Get GPU devices
numGPUs := int64(0)
	if agent.cfg.GPUSupportEnabled {
		err := agent.initializeGPUManager()
		if err != nil {
			seelog.Criticalf("Could not initialize Nvidia GPU Manager: %v", err)
			return exitcodes.ExitError
		}
		// Find number of GPUs instance has
		platformDevices := agent.getPlatformDevices()
		for _, device := range platformDevices {
			if *device.Type == ecs.PlatformDeviceTypeGpu {
				numGPUs++
			}
		}
	}
```
```
// Update GPU in hostResources map
hostResources["GPU"] = &ecs.Resource{
    Name:         utils.Strptr("GPU"),
    Type:         utils.Strptr("INTEGER"),
    IntegerValue: &numGPUs,
}
```

agent/app/agent.go
```diff
@@ -298,17 +298,36 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+	if err != nil {
+		seelog.Critical("Unable to fetch host resources")
+		return exitcodes.ExitError
```
chienhanlin, line 320: Discussed with @prateekchaudhry offline. We should revisit this error code by referring logic [here](https://github.com/aws/amazon-ecs-agent/blob/883211e24d9255023a68f6351001c291af8a699b/agent/app/agent.go#L391-L398).
* [ExitError](https://github.com/aws/amazon-ecs-agent/blob/883211e24d9255023a68f6351001c291af8a699b/agent/sighandlers/exitcodes/exitcodes.go#L21-L23): (as well as unspecified exit codes) indicates a fatal error occurred, but the agent should be restarted
* [ExitTerminal](https://github.com/aws/amazon-ecs-agent/blob/883211e24d9255023a68f6351001c291af8a699b/agent/sighandlers/exitcodes/exitcodes.go#LL24C2-L26C18): indicates the agent has exited unsuccessfully, but should not be restarted

agent/app/agent_test.go
```diff
@@ -78,6 +79,20 @@ var apiVersions = []dockerclient.DockerVersion{
 	dockerclient.Version_1_22,
 	dockerclient.Version_1_23}
 var capabilities []*ecs.Attribute
+var testHostCPU = int64(1024)
```
chienhanlin, line 82: nit. Should they be constants instead of variables since we are not going to change them for different test cases?

agent/engine/host_resource_manager.go
```diff
@@ -0,0 +1,88 @@
...
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+// Package engine contains the core logic for managing tasks
+
```
chienhanlin, line 15: nit. Can we add more comments/contexts for this new manager? See [agent/engine/docker_task_engine.go](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/engine/docker_task_engine.go) as an example. Thanks.

agent/engine/host_resource_manager.go
```diff
@@ -0,0 +1,88 @@
...
+	memory := int64(0)
+	consumedResourceMap["MEMORY"] = &ecs.Resource{
+		Name:         utils.Strptr("MEMORY"),
+		Type:         utils.Strptr("INTEGER"),
+		IntegerValue: &memory,
+	}
+	//PORTS
+	//Copying ports from host resources as consumed ports for initializing
+	ports := []*string{}
+	if resourceMap != nil && resourceMap["PORTS"] != nil {
```
chienhanlin, line 53: Q. Could a nil `resourceMap` ever be passed to here? If yes, should we returned an error instead of creating a new manager? or should we avoid using a nil resourceMap to initialize the manager?

Yiyuanzzz: Yes, a nil `resourceMap` will be passed to here when calling `engine.NewTaskEngine(....nil..)` in several test files, 
 for example in `agent/statemanager/state_manager_test.go` , ` taskEngine := engine.NewTaskEngine(&config.Config{}, nil, nil, nil, nil, nil, dockerstate.NewTaskEngineState(), nil, nil, nil, nil)` 
we add this logic to avoid nil pointer exceptions

agent/engine/host_resource_manager.go
Yiyuanzzz: Thanks Chilinn, will add more comments

agent/app/agent_test.go
```diff
@@ -78,6 +79,20 @@ var apiVersions = []dockerclient.DockerVersion{
...
+var testHostMEMORY = int64(1024)
+var testHostResource = map[string]*ecs.Resource{
```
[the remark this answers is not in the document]
Yiyuanzzz: I don't think so

agent/api/ecsclient/client.go
```diff
@@ -306,6 +306,18 @@ func (client *APIECSClient) getResources() ([]*ecs.Resource, error) {
 	return []*ecs.Resource{&cpuResource, &memResource, &portResource, &udpPortResource}, nil
 }
 
+func (client *APIECSClient) GetHostResources() (map[string]*ecs.Resource, error) {
```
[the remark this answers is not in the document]
Yiyuanzzz: Thanks, will update it

agent/app/agent.go
Yiyuanzzz: Sounds good to me!

agent/engine/host_resource_manager.go
Yiyuanzzz: Thanks Yinyi, will update it!

agent/engine/host_resource_manager.go
Yiyuanzzz: Sure, will change it

agent/app/agent.go
Yiyuanzzz: Discussed with @yinyic , we probably should use ExitError because it may be fail to read the host resource due to some temporary reasons and a retry will work.

agent/app/agent_test.go
Yiyuanzzz: if we make those constants, then testHostCPU and testHostMEMORY are not addressable, we can not get address of them in `var testHostResource = map[string]*ecs.Resource{ .....IntegerValue: &testHostCPU,}`

agent/engine/default.go
```diff
@@ -31,14 +32,15 @@ import (
 func NewTaskEngine(cfg *config.Config, client dockerapi.DockerClient,
 	credentialsManager credentials.Manager,
 	containerChangeEventStream *eventstream.EventStream,
-	imageManager ImageManager, state dockerstate.TaskEngineState,
+	imageManager ImageManager, hostResources map[string]*ecs.Resource, state dockerstate.TaskEngineState,
```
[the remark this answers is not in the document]
Yiyuanzzz: per offline discuss, will keep it as it is for now

agent/engine/host_resource_manager.go
prateekchaudhry: @Yiyuanzzz  Would it be more clearer to instead rename `hostResource` as `initialHostResource` and keep the struct name as `HostResourceManager`? As the `consumeResource` and `taskConsumed` will keep changing with state?

yinyic (approved):

PR #3693, opened 2023-05-11
Support firelens for bridge mode ServiceConnect task

yinyic:
### Summary

Currently, if a task is ServiceConnect-enabled, and runs in bridge network mode, then it won't be able to use `awsfirelens` log driver. 
This PR makes `awsfirelens` compatible with SC-enabled, bridge mode tasks.

### Implementation details

ECS Agent is responsible for injecting `FLUENT_HOST` environment variable to application containers. The variable contains the IP address assigned to the FireLens container. See public [doc](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_firelens.html). So far, ECS Agent resolves the container IP by directly querying the container network info from docker inspect. This becomes an issue with ServiceConnect bridge mode tasks, because we are launching pause containers for each application container (for network namespace configuration). In that case, the application container will not have the expected network info, as the info is now associated with its pause container.

This PR makes changes in two places - 
1. When resolving container IP for FireLens container itself
2. When resolving FireLens container IP for other application containers

In both cases, we now retrieve the container IP from the FireLens pause container instead, using the readily available `GetBridgeModePauseContainerForTaskContainer` function.

### Testing
* Added unit test for `CreateContainer` and `StartContainer` for both SC and non-SC bridge mode tasks.

* E2E test by running a bridge task with SC enabled that also use `awsfirelens` log driver - verified that 1) task is able to run, and 2) application log was successfully delivered to my fluentd destination (an S3 bucket)

Support firelens for bridge mode ServiceConnect task

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -1377,7 +1377,11 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
 				fluentNetworkPort: FluentNetworkPortValue,
 			})
 		} else if container.GetNetworkModeFromHostConfig() == "" || container.GetNetworkModeFromHostConfig() == apitask.BridgeNetworkMode {
-			ipAddress, ok := getContainerHostIP(task.GetFirelensContainer().GetNetworkSettings())
+			targetContainer := task.GetFirelensContainer()
+			if task.IsServiceConnectEnabled() {
+				targetContainer, _ = task.GetBridgeModePauseContainerForTaskContainer(targetContainer)
+			}
+			ipAddress, ok := getContainerHostIP(targetContainer.GetNetworkSettings())
 			if !ok {
 				err := apierrors.DockerClientConfigError{Msg: "unable to get BridgeIP for task in bridge mode"}
 				return dockerapi.DockerContainerMetadata{Error: apierrors.NamedError(&err)}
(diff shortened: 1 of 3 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1377,7 +1377,21 @@ func (engine *DockerTaskEngine) createContainer(task *apitask.Task, container *a
...
+			// For bridge-mode ServiceConnect-enabled tasks, we inject pause container for each application container
+			// including the firelens container. Therefore, when resolving the container IP, we should be checking that
+			// of the associated pause container.
+			if task.IsServiceConnectEnabled() {
+				var err error
+				targetContainer, err = task.GetBridgeModePauseContainerForTaskContainer(targetContainer)
+				if err != nil {
+					return dockerapi.DockerContainerMetadata{
+						Error: dockerapi.CannotStartContainerError{FromError: errors.New(fmt.Sprintf("failed to start firelens"+
+							"container: %v", err))},
```
sparrc, line 1390: major nit: there's a space missing here

sparrc (approved):

agent/engine/docker_task_engine.go
yinyic: Fixed. Thanks!

agent/engine/docker_task_engine.go
```diff
@@ -1629,26 +1643,56 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
 	// If container is a firelens container, fluent host is needed to be added to the environment variable for the task.
 	// For the supported network mode - bridge and awsvpc, the awsvpc take the host 127.0.0.1 but in bridge mode,
 	// there is a need to wait for the IP to be present before the container using the firelens can be created.
+	//
+	// For bridge-mode ServiceConnect-enabled tasks, we inject pause container for each application container
+	// including the firelens container. Therefore, when resolving the container IP, we should be checking that
+	// of the associated pause container. In such case, the firelens container has network mode "container" since it's
+	//launched into its pause container's network namespace.
```
fierlion, line 1650: nit: space before launched

agent/engine/docker_task_engine.go
```diff
@@ -1629,26 +1643,56 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
...
 	if container.GetFirelensConfig() != nil {
-		if !task.IsNetworkModeAWSVPC() && (container.GetNetworkModeFromHostConfig() == "" || container.GetNetworkModeFromHostConfig() == apitask.BridgeNetworkMode) {
+		if !task.IsNetworkModeAWSVPC() &&
+			(container.GetNetworkModeFromHostConfig() == "" ||
+				container.GetNetworkModeFromHostConfig() == apitask.BridgeNetworkMode ||
+				(container.GetNetworkModeFromHostConfig() == "container" && task.IsServiceConnectEnabled())) {
 			_, gotContainerIP := getContainerHostIP(dockerContainerMD.NetworkSettings)
+			if task.IsServiceConnectEnabled() {
+				targetContainer, err := task.GetBridgeModePauseContainerForTaskContainer(container)
```
chienhanlin, line 1658: * Logic blocks between line 1380-1394, line 1657-1666 and line 1675-1682 are very similar. Should we create a new private function and reuse it in these places? 
* Should we add any error log to help customers and us debug issues?
```
logger.Error("Failed to start FireLens container", logger.Fields{
    field.TaskID:    task.GetID(),
    field.Container: container.Name,
    field.Error:     err,
   ....
})
```
* Would we add new positive and negative test cases to docker_task_engine_test.go for this change?

fierlion: I was just thinking about how to make the conditional in 1652-55 simpler -- I think it wants a little comment to explain what we expect to end up calling `task.GetBridgeModePauseContainerForTaskContainer(container)`.  Or maybe a simple private isServiceConnect() function like `return container.GetNetworkModeFromHostConfig() == "container" && task.IsServiceConnectEnabled()`

yinyic: > Logic blocks between line 1380-1394, line 1657-1666 and line 1675-1682 are very similar. Should we create a new private function and reuse it in these places?

There are a couple differences that have made it a bit hard to create a single helper method. One is for application containers (other than firelens), the other is for Firelens container itself. One does not expect failure on resolving Firelens IP, while the other is fine with that and will wait for the iP to be available. 

> Should we add any error log to help customers and us debug issues?

Sure will add

> Would we add new positive and negative test cases to docker_task_engine_test.go for this change?

Sure

yinyic: > I was just thinking about how to make the conditional in 1652-55 simpler -- I think it wants a little comment to explain what we expect to end up calling `task.GetBridgeModePauseContainerForTaskContainer(container)`. Or maybe a simple private isServiceConnect() function like `return container.GetNetworkModeFromHostConfig() == "container" && task.IsServiceConnectEnabled()`

Yeah we could do that

agent/engine/docker_task_engine.go
```diff
@@ -1629,34 +1651,68 @@ func (engine *DockerTaskEngine) startContainer(task *apitask.Task, container *ap
...
-				getIPBridgeBackoff := retry.NewExponentialBackoff(minGetIPBridgeTimeout, maxGetIPBridgeTimeout, getIPBridgeRetryJitterMultiplier, getIPBridgeRetryDelayMultiplier)
-				contextWithTimeout, cancel := context.WithTimeout(engine.ctx, time.Minute)
-				defer cancel()
-				err := retry.RetryWithBackoffCtx(contextWithTimeout, getIPBridgeBackoff, func() error {
+	//
+	// For bridge-mode ServiceConnect-enabled tasks, we inject pause container for each application container
+	// including the firelens container. Therefore, when resolving the container IP, we should be checking that
+	// of the associated pause container. In such case, the firelens container has network mode "container" since it's
+	// launched into its pause container's network namespace.
+	if container.GetFirelensConfig() != nil && task.IsNetworkModeBridge() {
```
yinyic, line 1659: Previously we had
```
if container.GetFirelensConfig() != nil {
		if !task.IsNetworkModeAWSVPC() && (container.GetNetworkModeFromHostConfig() == "" || container.GetNetworkModeFromHostConfig() == apitask.BridgeNetworkMode)
```
The network mode checks in the second if statement were used for determining if task is bridge mode. That's because previously, we were only able to infer task bridge network mode from container config. But now this information is passed down directly from control plane.

fierlion (approved):

chienhanlin (approved):

PR #3697, opened 2023-05-15
Common task manifest/stop verification ACK in agent/

danehlim:
### Summary

Refactor functionality common to handling/responding to task manifest and task stop verification ACK messages from ACS in `agent/`.

NOTE: For context, logic for handling/responding to task manifest and task stop verification ACK messages from ACS both currently live in a single file `agent/acs/handler/task_manifest_handler.go`. In a future PR, this file will instead be split out into two files, where each file is dedicated to handling/responding to a single specific message type from ACS. This follows the best practice of the single-responsibility principle.

### Implementation details

Define new `manifestMessageIDAccessor` struct in `agent/acs/handler/task_manifest_common.go` such that `manifestMessageIDAccessor` implements `acssession.ManifestMessageIDAccessor` interface.

Update the following files to make use of `acssession.ManifestMessageIDAccessor` interface and `manifestMessageIDAccessor` struct:
   - `agent/acs/handler/task_manifest_handler.go`
   - `agent/acs/handler/task_manifest_handler_test.go`

### Testing

Common task manifest/stop verification ACK in agent/

### Licensing

### DIFF
--- a/agent/acs/handler/task_manifest_common.go
+++ b/agent/acs/handler/task_manifest_common.go
@@ -1,0 +1,23 @@
+package handler
+
+import (
+	"sync"
+)
+
+type manifestMessageIDAccessor struct {
+	messageID string
+	lock      sync.RWMutex
+}
+
+func (mmiAccessor *manifestMessageIDAccessor) GetMessageID() string {
+	mmiAccessor.lock.RLock()
+	defer mmiAccessor.lock.RUnlock()
+	return mmiAccessor.messageID
+}
+
+func (mmiAccessor *manifestMessageIDAccessor) SetMessageID(messageId string) error {
+	mmiAccessor.lock.Lock()
+	defer mmiAccessor.lock.Unlock()
+	mmiAccessor.messageID = messageId
+	return nil
+}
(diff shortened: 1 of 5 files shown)


ohsoo (approved):
No concerns, LGTM.

amogh09 (approved):

agent/acs/handler/task_manifest_common.go
```diff
@@ -0,0 +1,23 @@
+package handler
+
+import (
+	"sync"
+)
+
+type manifestMessageIDAccessor struct {
+	messageID string
+	lock      sync.RWMutex
```
amogh09, line 9: [Not required in this PR] I wonder if there's a way to define thread-safe variables at a higher level in Go. Having to manage a lock manually seems so archaic. 😅

danehlim: Hmm yeah, off the top of my head using a `sync.RWMutex` here to protect from concurrent access seems to be the most direct/straightforward approach here (as opposed to other common thread safety mechanisms from `sync` in Go). But I can take an action to look into this further.

amogh09: I think this is the most straightforward way in Go. I was comparing Go to other languages. For example, Java has atomic variables and Haskell has MVars.

PR #3700, opened 2023-05-17
Add host resource manager methods

prateekchaudhry:
### Summary

`HostResourceManager` keeps account of task resources for tasks running/scheduled to run on the instance Agent is running on. It will keep an account of cummulative `CPU`, `MEMORY`, `PORTS_TCP`, `PORTS_UDP` and `GPU` which are the same resources registered with ECS backend to schedule tasks on ECS instances. The overarching goal is to start a task on an instance as soon as the resources for it become available on the instance, and these resources will be kept track of by this manager.

A previous PR (#3684) added initialization for `HostResourceManager`.

This PR further adds following methods:
* `consume` method to keep account of resources of tasks being scheduled to start
* `consumable` method which given a set of resources returns whether those resources can be 'consumed' by that `HostResourceManager`
* `release` method to keep account of resources of tasks being stopped and releasing resources and adding back to pool of available resources

These methods are being added here. These will be used to schedule tasks according to above description in further changes.

### Testing

* Added unit tests to cover sanity checks of CPU, Memory, Ports (TCP/UDP) and GPU accounting. Checked that unit tests succeed.

Add host resource manager methods

### Licensing

### DIFF
--- a/agent/engine/host_resource_manager.go
+++ b/agent/engine/host_resource_manager.go
@@ -133,7 +133,8 @@ func (h *HostResourceManager) consume(taskArn string, resources map[string]*ecs.
 		}
 
 		// GPU
-		*h.consumedResource[GPU].IntegerValue += *resources[GPU].IntegerValue
+		gpu := *h.consumedResource[GPU].IntegerValue + *resources[GPU].IntegerValue
+		h.consumedResource[GPU].SetIntegerValue(gpu)
 
 		// Set consumed status
 		h.taskConsumed[taskArn] = true
@@ -153,13 +154,19 @@ func (h *HostResourceManager) checkConsumableIntType(resourceName string, resour
 
 func (h *HostResourceManager) checkConsumableStringSetType(resourceName string, resources map[string]*ecs.Resource) bool {
 	resourceSlice := resources[resourceName].StringSetValue
+
+	// (optimizization) Get a resource specific map to ease look up
+	consumedResourceMap := make(map[string]struct{}, len(h.consumedResource[resourceName].StringSetValue))
+	for _, v := range h.consumedResource[resourceName].StringSetValue {
+		consumedResourceMap[*v] = struct{}{}
+	}
+
 	// Check intersection of resource StringSetValue is empty with consumedResource
 	for _, obj1 := range resourceSlice {
-		for _, obj2 := range h.consumedResource[resourceName].StringSetValue {
-			// If port is already reserved by some other task, this 'resources' object can not be consumed
-			if *obj1 == *obj2 {
-				return false
-			}
+		_, ok := consumedResourceMap[*obj1]
+		if ok {
+			// If resource is already reserved by some other task, this 'resources' object can not be consumed
+			return false
 		}
 	}
 	return true
--- a/agent/engine/host_resource_manager.go
+++ b/agent/engine/host_resource_manager.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	"github.com/aws/amazon-ecs-agent/agent/logger"
+	"github.com/aws/amazon-ecs-agent/agent/logger/field"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 )
 
@@ -37,15 +38,24 @@ type HostResourceManager struct {
 	taskConsumed map[string]bool
 }
 
-// Returns if resources consumed or not and error status
-// false, nil -> did not consume, should stay pending
-// false, err -> resources map has errors, something went wrong
-// true, nil -> successfully consumed
-func (h *HostResourceManager) consume(taskArn string, resources map[string]*ecs.Resource) (bool, error) {
-	h.hostResourceManagerRWLock.Lock()
-	defer h.hostResourceManagerRWLock.Unlock()
+type ResourceNotFoundForTask struct {
+	resource string
+}
 
-	defer logger.Debug("Consumed resources after task consume call", logger.Fields{
+func (e *ResourceNotFoundForTask) Error() string {
+	return fmt.Sprintf("no %s in task resources", e.resource)
+}
+
+type ResourceIsNilForTask struct {
+	resource string
+}
+
+func (e *ResourceIsNilForTask) Error() string {
+	return fmt.Sprintf("resource %s is nil in task resources", e.resource)
+}
+
+func (h *HostResourceManager) logResources(msg string, taskArn string) {
+	logger.Debug(msg, logger.Fields{
 		"taskArn":   taskArn,
 		"CPU":       *h.consumedResource["CPU"].IntegerValue,
 		"MEMORY":    *h.consumedResource["MEMORY"].IntegerValue,
@@ -53,16 +63,31 @@ func (h *HostResourceManager) consume(taskArn string, resources map[string]*ecs.
 		"PORTS_UDP": h.consumedResource["PORTS_UDP"].StringSetValue,
 		"GPU":       *h.consumedResource["GPU"].IntegerValue,
 	})
+}
+
+// Returns if resources consumed or not and error status
+// false, nil -> did not consume, task should stay pending
+// false, err -> resources map has errors, task should fail as cannot schedule with 'wrong' resource map (this basically never happens)
+// true, nil -> successfully consumed, task should progress with task creation
+func (h *HostResourceManager) consume(taskArn string, resources map[string]*ecs.Resource) (bool, error) {
+	h.hostResourceManagerRWLock.Lock()
+	defer h.hostResourceManagerRWLock.Unlock()
+	defer h.logResources("Consumed resources after task consume call", taskArn)
 
 	// Check if already consumed
-	_, ok := h.consumedResource[taskArn]
+	_, ok := h.taskConsumed[taskArn]
 	if ok {
 		// Nothing to do, already consumed, return
+		logger.Info("Resources pre-consumed, continue to task creation", logger.Fields{"taskArn": taskArn})
 		return true, nil
 	}
 
 	ok, err := h.consumable(resources)
 	if err != nil {
+		logger.Error("Resources failing to consume, error in task resources", logger.Fields{
+			"taskArn":   taskArn,
+			field.Error: err,
+		})
 		return false, err
 	}
 	if ok {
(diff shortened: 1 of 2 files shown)


agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+	"fmt"
+	"sync"
+
 	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 )
 
+// TODO remove this once resource, consume are used
+//lint:file-ignore U1000 Ignore all unused code
```
prateekchaudhry, line 28: This static check directive will be removed in a future change once consume and release are being used.

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+				newPortSlice := append(itrPortSlice[:idx], itrPortSlice[idx+1:]...)
+				h.consumedResource["PORTS_TCP"].StringSetValue = newPortSlice
+			}
+		}
+
+		// PORTS_UDP
+		portsResource, ok = resources["PORTS_UDP"]
+		if ok {
+			taskPortsSlice := portsResource.StringSetValue
+			for _, port := range taskPortsSlice {
```
yinyic, line 218: All consumed ports for a single task are going to be in a contiguous sub-slice in `consumedResource["PORTS_UDP"]`, since we add them together. Therefore, instead of removing ports one at a time, I think we can find the starting and ending index and remove all at once.

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
 
 	//task.arn to boolean whether host resources consumed or not
 	taskConsumed map[string]bool
 }
 
+// Returns if resources consumed or not and error status
+// false, nil -> did not consume, should stay pending
+// false, err -> resources map has errors, something went wrong
+// true, nil -> successfully consumed
+func (h *HostResourceManager) consume(taskArn string, resources map[string]*ecs.Resource) (bool, error) {
```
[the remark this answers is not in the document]
prateekchaudhry: I see these functions to be used only in engine package. Unless there are good reasons to export them, I believe we can keep them as it is

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+		"taskArn":   taskArn,
+		"CPU":       *h.consumedResource["CPU"].IntegerValue,
+		"MEMORY":    *h.consumedResource["MEMORY"].IntegerValue,
+		"PORTS_TCP": h.consumedResource["PORTS_TCP"].StringSetValue,
+		"PORTS_UDP": h.consumedResource["PORTS_UDP"].StringSetValue,
+		"GPU":       *h.consumedResource["GPU"].IntegerValue,
+	})
+
+	// Check if already consumed
+	_, ok := h.consumedResource[taskArn]
```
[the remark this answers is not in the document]
prateekchaudhry: Thanks for catching it! will correct that

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+	}
+
+	// MEM
+	memResource, ok := resources["MEMORY"]
+	if ok {
+		if *(h.initialHostResource["MEMORY"].IntegerValue) < *(h.consumedResource["MEMORY"].IntegerValue)+*(memResource.IntegerValue) {
+			return false, nil
+		}
+	} else {
+		return false, fmt.Errorf("No MEMORY in task resources")
```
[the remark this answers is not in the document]
prateekchaudhry: The output of ToHostResources(#3706) will be used here, and that will always all these entries

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+	h.hostResourceManagerRWLock.Lock()
+	defer h.hostResourceManagerRWLock.Unlock()
+	defer logger.Debug("Consumed resources after task release call", logger.Fields{
+		"taskArn":   taskArn,
+		"CPU":       *h.consumedResource["CPU"].IntegerValue,
+		"MEMORY":    *h.consumedResource["MEMORY"].IntegerValue,
+		"PORTS_TCP": h.consumedResource["PORTS_TCP"].StringSetValue,
+		"PORTS_UDP": h.consumedResource["PORTS_UDP"].StringSetValue,
+		"GPU":       *h.consumedResource["GPU"].IntegerValue,
+	})
```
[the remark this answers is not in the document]
prateekchaudhry: Yes, will create the recommended `printResources` method

agent/engine/host_resource_manager.go
prateekchaudhry: I was thinking of implementing as a generic slice check, but can get this optimization in as well

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
 
+// TODO remove this once resource, consume are used
+//lint:file-ignore U1000 Ignore all unused code
+
 // initialHostResource keeps account of each task in
 type HostResourceManager struct {
-	initialHostResource map[string]*ecs.Resource
-	consumedResource    map[string]*ecs.Resource
+	initialHostResource       map[string]*ecs.Resource
+	consumedResource          map[string]*ecs.Resource
```
amogh09, line 33: A bit of a meta question here. IIUC the keys to these two maps are resource types such as  CPU, memory, and ports. Is the full range of resource types managed by `HostResourceManager` type static or dynamic? If it is static, then have we thought about adding dedicated fields for each resource type and using new easy-to-use types? `ecs.Resource` type seems quite error prone as it is not well typed.

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+			return false, nil
+		}
+	} else {
+		return false, fmt.Errorf("No CPU in task resources")
+	}
+
+	// MEM
+	memResource, ok := resources["MEMORY"]
+	if ok {
+		if *(h.initialHostResource["MEMORY"].IntegerValue) < *(h.consumedResource["MEMORY"].IntegerValue)+*(memResource.IntegerValue) {
```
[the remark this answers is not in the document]
prateekchaudhry: Although it will be guaranteed as we will only feed output of task.ToHostResources(#3706), I think it can be good to have a resourceMap health check function implemented. Let me do that

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+// return values
+func (h *HostResourceManager) consumable(resources map[string]*ecs.Resource) (bool, error) {
+	// CPU
+	cpuResource, ok := resources["CPU"]
+	if ok {
+		if *(h.initialHostResource["CPU"].IntegerValue) < *(h.consumedResource["CPU"].IntegerValue)+*(cpuResource.IntegerValue) {
+			return false, nil
+		}
+	} else {
+		return false, fmt.Errorf("No CPU in task resources")
```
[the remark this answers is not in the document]
prateekchaudhry: Thanks for pointing that, will follow reformat errors according to the convention.

agent/engine/host_resource_manager.go
prateekchaudhry: There will be 5 resource types accounted for in HostResourceManager currently (cpu, memory, ports_tcp, ports_udp, gpu). We are choose `ecs.Resource` type as this is the same format agent uses to report host resources to ecs when registering instance ([here](https://github.com/aws/amazon-ecs-agent/blob/883211e24d9255023a68f6351001c291af8a699b/agent/api/ecsclient/client.go#L188)).

amogh09: I see. IIUC the goal of `HostResourceManager` is to manage the state of resources and `ecs.Resource` type doesn't seem to be easy-to-use for that purpose. It is probably designed for communicating the resources state to the backend. 

Also since each resource has its own mutation function (CPU is a continuous variable from 0 to 100 while ports are a set), it might be worthwhile to define an interface for resource and handle the state changes behind in the implementation.

This is a non-blocking comment and I am just sharing my opinion. :)

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,226 @@
...
+		// CPU
+		cpu := *h.consumedResource["CPU"].IntegerValue + *resources["CPU"].IntegerValue
+		h.consumedResource["CPU"].SetIntegerValue(cpu)
+
+		// MEM
+		mem := *h.consumedResource["MEMORY"].IntegerValue + *resources["MEMORY"].IntegerValue
+		h.consumedResource["MEMORY"].SetIntegerValue(mem)
+
+		// PORTS
+		portsResource, ok := resources["PORTS_TCP"]
```
amogh09, line 103: How about defining constants for resource identifiers like `"PORTS_TCP"`? Using the identifiers directly seems difficult to maintain and error-prone.

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,311 @@
...
+
+	// Check if already consumed
+	_, ok := h.taskConsumed[taskArn]
+	if ok {
+		// Nothing to do, already consumed, return
+		logger.Info("Resources pre-consumed, continue to task creation", logger.Fields{"taskArn": taskArn})
+		return true, nil
+	}
+
+	ok, err := h.consumable(resources)
```
Realmonia, line 85: What does it typically mean when we have 1 resource type errored out of the 5 resource types, and the other ones were fine? What will be our action when this happens?

prateekchaudhry: This should never happen, but putting in as a check. The task should fail to start with incorrect resource conversion as put in the comment above consume `// false, err -> resources map has errors, task should fail as cannot schedule with 'wrong' resource map (this basically never happens)`

agent/engine/host_resource_manager.go
prateekchaudhry: I agree a separate type for `HostResourceManager` can be a bit more useful. On the other hand I also see that by re-using `ecs.Resource` we are able to piggy back on an existing implementation which could be easy in it's own way as it is declared to define ecs resources (albeit to communicate to backend, perhaps). And using a map, accessing values becomes fairly easy.

If it is not much of a hassle, I will stay with `ecs.Resource` implementation, as previous, parallel (and future) work is already using this type. But I will be willing to change this upon recommendation :)

agent/engine/host_resource_manager.go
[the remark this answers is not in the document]
amogh09: Why don't we put host resource manager in its own package? How are consumers of host resource manager supposed to know what methods form its public API?

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,319 @@
...
+func (h *HostResourceManager) checkConsumableIntType(resourceName string, resources map[string]*ecs.Resource) bool {
+	resourceConsumableStatus := *(h.initialHostResource[resourceName].IntegerValue) >= *(h.consumedResource[resourceName].IntegerValue)+*(resources[resourceName].IntegerValue)
+	return resourceConsumableStatus
+}
+
+func (h *HostResourceManager) checkConsumableStringSetType(resourceName string, resources map[string]*ecs.Resource) bool {
+	resourceSlice := resources[resourceName].StringSetValue
+	// Check intersection of resource StringSetValue is empty with consumedResource
+	for _, obj1 := range resourceSlice {
+		for _, obj2 := range h.consumedResource[resourceName].StringSetValue {
```
amogh09, line 158: Is it worth optimizing this from O(mn) to O(m+n) using maps? I am imagining this taking 200K iterations on a really busy instance with 20K consumed ports and a new task that wants to consume 10 ports.

agent/engine/host_resource_manager.go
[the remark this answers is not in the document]
prateekchaudhry: We put it in engine (and not in it's own package) because we only see `HostResourceManager` as being used to enable task scheduling, specifically in `task_manager` and `docker_task_engine`, and hence did not export the methods as well.

agent/engine/host_resource_manager.go
prateekchaudhry: Good suggestion, let me get that in.

amogh09: nit: Maybe we should create a map with `resourceSlice` values instead as its size should be considerably smaller than consumed ports size.

amogh09 (approved):

amogh09 (dismissed):

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,313 @@
...
+			}
+		}
+	} else {
+		return &ResourceNotFoundForTask{resourceName}
+	}
+	return nil
+}
+
+// Checks all resources exists and their values are not nil
+func checkResourcesHealth(resources map[string]*ecs.Resource) error {
```
yinyic, line 189: I'm not sure if we necessarily want to make sure that all resource types exist in the map. Technically, for tasks that don't consume certain resource types, it should be okay for them to leave it out. 

We might want to do check, instead:
```
for each key,val in resources:
  does key exist in host resources? i.e. is it a recognized resource type? if not, throw error
  if valid key, then check if value type matches
  Lastly, check if the specific type of value is not nil
```

agent/engine/host_resource_manager.go
```diff
@@ -16,20 +16,313 @@
...
+	// no intersection found
+	if begin == len(s1) {
+		return s1
+	}
+
+	for ; end >= 0; end-- {
+		if *s1[end] == *s2[len(s2)-1] {
+			break
+		}
+	}
```
yinyic, line 284: We can have `end = begin + len(s2)`?

Realmonia (dismissed):

yinyic (approved):

Realmonia (approved):

amogh09 (approved):

PR #3703, opened 2023-05-17
Refresh Windows Domainless gMSA plugin credentials

arun-annamalai:
### Summary

The domainless gMSA plugin requires taskExecutionRole IAM credentials to access customer resources. These credentials are populated once at the beginning of task start, but they need to be updated each time the credentials are vended down to the agent. 

### Implementation details

This change is primarily implemented in the refresh handler. Everytime an execution role gets vended down, the refresh handler should populate the Windows registry key where the plugin credentials are stored. 

### Testing

This change was unit tested

I have also let a task run for 2 days and then checked out the instance logs and I do see that the credentials are being set correctly after a refresh from ACS (have logs I can send) 

New tests cover the changes: 
yes

Refresh Windows Domainless gMSA plugin credentials
### Licensing

### DIFF
--- a/agent/acs/handler/refresh_credentials_handler.go
+++ b/agent/acs/handler/refresh_credentials_handler.go
@@ -13,9 +13,8 @@
 package handler
 
 import (
-	"fmt"
-
 	"context"
+	"fmt"
 
 	"github.com/aws/amazon-ecs-agent/agent/engine"
 	"github.com/aws/amazon-ecs-agent/agent/wsclient"
@@ -23,6 +22,13 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
+
+	"github.com/pkg/errors"
+)
+
+var (
+	// For ease of unit testing
+	checkAndSetDomainlessGMSATaskExecutionRoleCredentialsImpl = checkAndSetDomainlessGMSATaskExecutionRoleCredentials
 )
 
 // refreshCredentialsHandler represents the refresh credentials operation for the ACS client
@@ -145,10 +151,11 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
 	if !validRoleType(roleType) {
 		seelog.Errorf("Unknown RoleType for task in credentials message, roleType: %s arn: %s, messageId: %s", roleType, taskArn, messageId)
 	} else {
+		iamRoleCredentials := credentials.IAMRoleCredentialsFromACS(message.RoleCredentials, roleType)
 		err = refreshHandler.credentialsManager.SetTaskCredentials(
 			&(credentials.TaskIAMRoleCredentials{
 				ARN:                taskArn,
-				IAMRoleCredentials: credentials.IAMRoleCredentialsFromACS(message.RoleCredentials, roleType),
+				IAMRoleCredentials: iamRoleCredentials,
 			}))
 		if err != nil {
 			seelog.Errorf("Unable to update credentials for task, err: %v messageId: %s", err, messageId)
@@ -160,6 +167,12 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
 		}
 		if roleType == credentials.ExecutionRoleType {
 			task.SetExecutionRoleCredentialsID(aws.StringValue(message.RoleCredentials.CredentialsId))
+			// Refresh domainless gMSA plugin credentials if needed
+			err = checkAndSetDomainlessGMSATaskExecutionRoleCredentialsImpl(iamRoleCredentials, task)
+			if err != nil {
+				seelog.Errorf("Unable to SetDomainlessGMSATaskExecutionRoleCredentials for task %s, err: %v messageId: %s", taskArn, err, messageId)
+				return errors.Wrap(err, "unable to SetDomainlessGMSATaskExecutionRoleCredentials")
+			}
 		}
 	}
 
--- a/agent/acs/handler/refresh_credentials_handler_linux.go
+++ b/agent/acs/handler/refresh_credentials_handler_linux.go
@@ -1,0 +1,30 @@
+//go:build linux
+// +build linux
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package handler
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
+)
+
+func checkAndSetDomainlessGMSATaskExecutionRoleCredentials(iamRoleCredentials credentials.IAMRoleCredentials, task *task.Task) error {
+	// exit early if the task does not need domainless gMSA
+	if !task.RequiresDomainlessCredentialSpecResource() {
+		return nil
+	}
+	return nil
+}
--- a/agent/acs/handler/refresh_credentials_handler_test.go
+++ b/agent/acs/handler/refresh_credentials_handler_test.go
@@ -17,11 +17,13 @@ package handler
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 	"sync"
 	"testing"
 	"time"
 
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
 	mock_engine "github.com/aws/amazon-ecs-agent/agent/engine/mocks"
 	mock_wsclient "github.com/aws/amazon-ecs-agent/agent/wsclient/mock"
@@ -30,6 +32,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/golang/mock/gomock"
+	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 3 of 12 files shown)


agent/acs/handler/refresh_credentials_handler_test.go
```diff
@@ -233,6 +236,67 @@ func TestCredentialsMessageNotAckedWhenTaskNotFound(t *testing.T) {
 	cancel()
 }
 
+// TestCredentialsMessageNotAckedWhenDomainlessGMSACredentialsNotSet tests if credential messages
+// are not acked when the task arn in the message is not found in the task
+// engine
+func TestCredentialsMessageNotAckedWhenDomainlessGMSACredentialsError(t *testing.T) {
+	testCases := []struct {
```
mythri-garaga, line 243: question(non-blocking): Since the test name suggests that we are testing a very specific scenario, do we need table-driven tests here?

arun-annamalai: I have tested this manually over the last 2 days (have logs on the instance I can send)
You are right that we don't need table tests, but it was just in case for future extensibility. I can remove if needed.

agent/acs/handler/refresh_credentials_handler_test.go
```diff
@@ -233,6 +236,67 @@ func TestCredentialsMessageNotAckedWhenTaskNotFound(t *testing.T) {
 	cancel()
 }
 
+// TestCredentialsMessageNotAckedWhenDomainlessGMSACredentialsNotSet tests if credential messages
+// are not acked when the task arn in the message is not found in the task
```
amogh09, line 240: Is this test testing the case when "the task arn in the message is not found in the task"? I think the test is for the case when setting domainless GMSA creds failed?

mythri-garaga (approved):

agent/acs/handler/refresh_credentials_handler.go
```diff
@@ -160,6 +165,15 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
 		}
 		if roleType == credentials.ExecutionRoleType {
 			task.SetExecutionRoleCredentialsID(aws.StringValue(message.RoleCredentials.CredentialsId))
+			// Refresh domainless gMSA plugin credentials
+			if task.RequiresDomainlessCredentialSpecResource() {
```
amogh09, line 169: How about we perform this check in the setter function (and maybe rename it)? That way the refresh handler would be simpler and have one less responsibility.

Also I think `TestHandleRefreshMessageAckedWhenCredentialsUpdatedDomainlessGMSA` would become much simpler as we will be able to split it into two tests - 
1. Acks are sent when domainless gMSA setter does not fail (we might even be able to use an existing test for this)
2. Setter function for windows does not attempt to update registry if task does not require domainless GMSA.

agent/acs/handler/refresh_credentials_handler_test.go
arun-annamalai: ah yeah too much copy paste. Let me fix this comment, you are correct that this test case is for when setting domainless creds fails.

agent/acs/handler/refresh_credentials_handler.go
```diff
@@ -160,6 +165,15 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
...
+				err = setDomainlessGMSATaskExecutionRoleCredentialsImpl(iamRoleCredentials, task.Arn)
+				if err != nil {
+					seelog.Errorf("Unable to SetDomainlessGMSATaskExecutionRoleCredentials for task %s, err: %v messageId: %s", taskArn, err, messageId)
+					return fmt.Errorf("unable to SetDomainlessGMSATaskExecutionRoleCredentials %v", err)
```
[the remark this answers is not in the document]
arun-annamalai: Done

agent/acs/handler/refresh_credentials_handler.go
arun-annamalai: Addressed both. 

Item 1 is in Function TestHandleRefreshMessageAckedWhenCredentialsUpdated


Item 2 is in Function TestHandleRefreshMessageAckedAndNoSetDomainlessCredentialsWhenNonDomainlessGMSATask

agent/acs/handler/refresh_credentials_handler.go
```diff
@@ -160,6 +167,15 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
 		}
 		if roleType == credentials.ExecutionRoleType {
 			task.SetExecutionRoleCredentialsID(aws.StringValue(message.RoleCredentials.CredentialsId))
+			// Refresh domainless gMSA plugin credentials
+			if task.RequiresDomainlessCredentialSpecResource() {
+				err = checkAndSetDomainlessGMSATaskExecutionRoleCredentialsImpl(iamRoleCredentials, task.Arn, task.RequiresDomainlessCredentialSpecResource())
```
amogh09, line 172: I think we should pass `task` to this function and let it extract the ARN and figure out if it's a domainless gMSA task. Passing the result of `task.RequiresDomainlessCredentialSpecResource()` still leaks implementation details into the ACS refresh handler.

agent/acs/handler/refresh_credentials_handler_linux.go
```diff
@@ -0,0 +1,29 @@
...
+
+package handler
+
+import (
+	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
+)
+
+func checkAndSetDomainlessGMSATaskExecutionRoleCredentials(iamRoleCredentials credentials.IAMRoleCredentials, taskArn string, domainlessGMSATask bool) error {
+	// exit early if the task does not need domainless gMSA
+	if !domainlessGMSATask {
```
amogh09, line 25: Why do we need this if? Isn't it redundant?

agent/acs/handler/refresh_credentials_handler_test.go
```diff
@@ -233,50 +236,243 @@ func TestCredentialsMessageNotAckedWhenTaskNotFound(t *testing.T) {
...
+		})
 	}
+}
 
-	// Wait till we get an ack from the ackBuffer
-	select {
-	case <-ctx.Done():
+// TestHandleRefreshMessageAckedWhenCredentialsUpdated tests that a credential message
+// is ackd when the credentials are updated successfully and the domainless gMSA plugin credentials are updated successfully
+func TestHandleRefreshMessageAckedWhenCredentialsUpdated(t *testing.T) {
```
amogh09, line 311: Do we need this new test? I think `TestHandleRefreshMessageAckedWhenCredentialsUpdated` already covers this case. 

Refresh handler should care only about two cases (once we move `task.RequiresDomainlessCredentialSpecResource()` call to the implementation of checkAndSet function) - 
1. `checkAndSetDomainlessGMSATaskExecutionRoleCredentialsImpl` returns an `error`
2. `checkAndSetDomainlessGMSATaskExecutionRoleCredentialsImpl` returns `nil`

amogh09: If we want an explicit test then we can add this test with a single case inside (checkAndSet returns `nil`).

agent/acs/handler/refresh_credentials_handler_linux.go
```diff
@@ -0,0 +1,30 @@
...
+package handler
+
+import (
+	"github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
+)
+
+func checkAndSetDomainlessGMSATaskExecutionRoleCredentials(iamRoleCredentials credentials.IAMRoleCredentials, task *task.Task) error {
+	// exit early if the task does not need domainless gMSA
+	if !task.RequiresDomainlessCredentialSpecResource() {
```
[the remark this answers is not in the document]
amogh09: Spoke offline to Arun. We will need this if block later.

agent/api/container/container.go
```diff
@@ -1366,6 +1366,16 @@ func (c *Container) RequiresCredentialSpec() bool {
 	return true
 }
 
+// RequiresDomainlessCredentialSpec checks if container needs a domainless credentialspec resource
+func (c *Container) RequiresDomainlessCredentialSpec() bool {
+	credSpec, err := c.getCredentialSpec()
```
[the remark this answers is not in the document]
amogh09: Talked to Arun offline. If we log here then that can get noisy for non-domainless gMSA tasks. So we will skip adding logs here.

agent/acs/handler/refresh_credentials_handler_test.go
```diff
@@ -234,49 +237,158 @@ func TestCredentialsMessageNotAckedWhenTaskNotFound(t *testing.T) {
...
+		domainlessGMSATaskExpectedInput bool
+		containers                      []*apicontainer.Container
+	}{
+		{
+			name:       "EmptyTaskSucceeds",
+			taskArn:    taskArn,
+			containers: []*apicontainer.Container{},
+		},
+		{
+			name:                            "DomainlessTaskSucceeds",
```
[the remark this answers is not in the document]
arun-annamalai: yes will be deleting 1 test case

agent/acs/handler/refresh_credentials_handler_test.go
```diff
@@ -234,49 +237,158 @@ func TestCredentialsMessageNotAckedWhenTaskNotFound(t *testing.T) {
...
+			checkAndSetDomainlessGMSATaskExecutionRoleCredentialsImpl = func(iamRoleCredentials credentials.IAMRoleCredentials, task *apitask.Task) error {
+				if tc.taskArn != task.Arn {
+					return errors.New(fmt.Sprintf("Expected taskArnInput to be %s, instead got %s", tc.taskArn, task.Arn))
+				}
 
-	// Wait till we get an ack from the ackBuffer
-	select {
-	case <-ctx.Done():
-	}
+				if tc.domainlessGMSATaskExpectedInput != task.RequiresDomainlessCredentialSpecResource() {
```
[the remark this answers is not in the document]
arun-annamalai: removing this check

agent/acs/handler/refresh_credentials_handler.go
```diff
@@ -160,6 +167,13 @@ func (refreshHandler *refreshCredentialsHandler) handleSingleMessage(message *ec
 		}
 		if roleType == credentials.ExecutionRoleType {
 			task.SetExecutionRoleCredentialsID(aws.StringValue(message.RoleCredentials.CredentialsId))
+			// Refresh domainless gMSA plugin credentials if needed
+			err = checkAndSetDomainlessGMSATaskExecutionRoleCredentialsImpl(iamRoleCredentials, task)
+			if err != nil {
+				seelog.Errorf("Unable to SetDomainlessGMSATaskExecutionRoleCredentials for task %s, err: %v messageId: %s", taskArn, err, messageId)
+				return errors.Wrap(err, "unable to SetDomainlessGMSATaskExecutionRoleCredentials")
+			}
+			seelog.Infof("Successfully SetDomainlessGMSATaskExecutionRoleCredentials for task %s messageId: %s", taskArn, messageId)
```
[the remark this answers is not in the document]
arun-annamalai: correct, will be fixing this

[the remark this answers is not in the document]
arun-annamalai: I moved the log line into the function 

credentialspec.SetTaskExecutionCredentialsRegKeys

PR #3705, opened 2023-05-18
Move v2 metadata models to ecs-agent module

amogh09:
### Summary

Move v2 metadata models to ecs-agent module. These models will be needed in ecs-agent module in future work when we move metadata endpoint handlers to ecs-agent module.

### Implementation details

* Move `TaskResponse`, `ContainerResponse`, `LimitsResponse`, and `ErrorResponse` structs from `agent/handlers/v2/response.go` to `ecs-agent/tmds/handlers/v2/response.go`.
* Add a new struct `HealthStatus` for metadata health status that replaces `apicontainer.HealthStatus` from `agent/api/container` package in `ContainerResponse`. This is done to decouple metadata health status from API Container health status that is used in a lot of places. The new struct has the same fields except for `Status` field that has type `string` instead of `ContainerHealthStatus` enum.
* A new `dockerContainerHealthToV2Health` function is added to `agent/handlers/v2/response.go` that converts `apicontainer.HealthStatus` to metadata `HealthStatus`. `TestDockerContainerHealthToV2HealthJSON` is added to ensure that the JSON representations of the two structs is the same. 
* Existing unit tests in `agent/handlers/v2/response_test.go` are updated with new asserts for health status for greater confidence in the new health status struct.

### Testing

New unit tests are added for the new health status struct. 

In addition to automated tests, also performed manual tests to ensure that the v2, v3, and v4 metadata endpoints return the same metadata with and without the changes in this PR.

Move v2 metadata models to ecs-agent module

### Licensing

### DIFF
--- a/agent/handlers/v2/response.go
+++ b/agent/handlers/v2/response.go
@@ -14,85 +14,21 @@
 package v2
 
 import (
-	"time"
-
 	"github.com/aws/aws-sdk-go/aws/awserr"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	v1 "github.com/aws/amazon-ecs-agent/agent/handlers/v1"
 	apieni "github.com/aws/amazon-ecs-agent/ecs-agent/api/eni"
 	tmdsresponse "github.com/aws/amazon-ecs-agent/ecs-agent/tmds/handlers/response"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/tmds/handlers/utils"
+	tmdsv2 "github.com/aws/amazon-ecs-agent/ecs-agent/tmds/handlers/v2"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
 	"github.com/pkg/errors"
 )
-
-// TaskResponse defines the schema for the task response JSON object
-type TaskResponse struct {
-	Cluster               string              `json:"Cluster"`
-	TaskARN               string              `json:"TaskARN"`
-	Family                string              `json:"Family"`
-	Revision              string              `json:"Revision"`
-	DesiredStatus         string              `json:"DesiredStatus,omitempty"`
-	KnownStatus           string              `json:"KnownStatus"`
-	Containers            []ContainerResponse `json:"Containers,omitempty"`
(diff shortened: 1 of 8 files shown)


ecs-agent/tmds/handlers/v2/response.go
```diff
@@ -0,0 +1,95 @@
...
+	Networks      []response.Network        `json:"Networks,omitempty"`
+	Health        *HealthStatus             `json:"Health,omitempty"`
+	Volumes       []response.VolumeResponse `json:"Volumes,omitempty"`
+	LogDriver     string                    `json:"LogDriver,omitempty"`
+	LogOptions    map[string]string         `json:"LogOptions,omitempty"`
+	ContainerARN  string                    `json:"ContainerARN,omitempty"`
+}
+
+// Container health status
+type HealthStatus struct {
```
amogh09, line 68: This is a new type which matches the `apicontainer.HealthStatus` type that is being used in v2 metadata response currently. 

 https://github.com/aws/amazon-ecs-agent/blob/883211e24d9255023a68f6351001c291af8a699b/agent/api/container/container.go#L111-L120

This is being done to decouple metadata response models from API models.

agent/handlers/v2/response.go
```diff
@@ -302,9 +238,24 @@ func NewContainerResponse(
 	return resp
 }
 
+// Converts apicontainer HealthStatus type to v2 Metadata HealthStatus type
+func dockerContainerHealthToV2Health(health apicontainer.HealthStatus) *tmdsv2.HealthStatus {
+	status := health.Status.String()
+	if health.Status == apicontainerstatus.ContainerHealthUnknown {
+		// Skip sending status if it is unknown
```
chienhanlin, line 245: Q. did you previously send heath status `ContainerHealthUnknown`? If yes, would this be a new change introduced to v2? Also curious how would `Since`, `ExitCode` and `Output` look like when status is `ContainerHealthUnknown` : )

amogh09: That's a good question. No we do not send `Status` when container's health status is `ContainerHealthUnknown` because the `Status` field in  `apicontainer.HealthStatus` struct is defined with `json:"status,omitempty"`.

`ContainerHealthUnknown` is the zero value of `Status`'s type so `Status = ContainerHealthUnknown` is skipped when `apicontainer.HealthStatus` is marshaled to JSON.

The `TestDockerContainerHealthToV2HealthJSON` test in this PR covers this case.

amogh09: https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/container/container.go#L111-L120

chienhanlin: Thanks for explaining this in detail!

chienhanlin (approved):

ubhattacharjya (approved):

PR #3706, opened 2023-05-19
Add method to get host resources reserved for a task

prateekchaudhry:
### Summary

Adding a method to `ToHostResources` for `Task` struct. This returns resources in `map[string]*ecs.Resource` type which can later directly be consumed by `HostResourceManager` (See related PRs #3700, #3684 which setup this) to account for resources for tasks scheduled to be run/running.

There is a conversion logic on how to interpret different task resources at task/container levels when accounting for a task. This has been written down in comments above `ToHostResources`, and the function basically implements that in code.

Resources to account for : cpu, memory, ports(tcp/udp) and gpu

Both `HostResourceManager` methods and `ToHostResources` will be used in a future change which will remove serialization and instead wire these up together in task_manager/docker_task_engine for starting and stopping tasks.

### Testing

Unit tests to test for basic resource conversion logic have been added and tested with.

Add method to get host resources reserved for a task

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/aws/amazon-ecs-agent/agent/acs/model/ecsacs"
 	apiappmesh "github.com/aws/amazon-ecs-agent/agent/api/appmesh"
+	"github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	apieni "github.com/aws/amazon-ecs-agent/agent/api/eni"
@@ -36,6 +37,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/credentials"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
+	"github.com/aws/amazon-ecs-agent/agent/ecs_client/model/ecs"
 	"github.com/aws/amazon-ecs-agent/agent/logger"
 	"github.com/aws/amazon-ecs-agent/agent/logger/field"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource"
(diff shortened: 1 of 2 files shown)


Yiyuanzzz (approved):

agent/api/task/task.go
```diff
@@ -3502,3 +3504,138 @@ func (task *Task) IsServiceConnectConnectionDraining() bool {
...
+		}
+	} else {
+		containerMEMint64 := int64(0)
+		// To parse memory reservation / soft limit
+		hostConfig := &dockercontainer.HostConfig{}
+
+		for _, c := range task.Containers {
+			if c.DockerConfig.HostConfig != nil {
+				err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
+				if err != nil || hostConfig.MemoryReservation <= 0 {
```
Realmonia, line 3571: Just out of curious, is there a case that MemoryReservation actually less than 0? How did we know it's unit with unmarshal?

prateekchaudhry: I don't see a scenario where it will be less than 0. Units are mentioned in comments [here](https://github.com/aws/amazon-ecs-agent/blob/e28ada643fa37572031d7f9cf2f208bb52348f93/agent/vendor/github.com/docker/docker/api/types/container/host_config.go#L366), also verified by manual runs.

agent/api/task/task.go
```diff
@@ -3502,3 +3504,138 @@ func (task *Task) IsServiceConnectConnectionDraining() bool {
...
+//   - Else use memory field
+//
+// * Ports (TCP/UDP)
+//   - Only account for hostPort
+//   - Don't need to account for awsvpc mode, each task gets its own namespace
+//
+// * GPU
+//   - Return num of gpus requested (len of GPUIDs field)
+//
+// TODO remove ToHostResources is used
```
fierlion, line 3526: nit: not sure what this TODO tells us.

agent/api/task/task.go
fierlion: `int64` implies that this is a signed integer (ie could be negative).  You might also add another check to validate that the HostConfig is an int.

agent/api/task/task_test.go
```diff
@@ -4802,3 +4803,196 @@ func TestInitializeAndGetCredentialSpecResource(t *testing.T) {
 	_, ok := task.GetCredentialSpecResource()
 	assert.True(t, ok)
 }
+
+func getTestTaskResourceMap(cpu int64, mem int64, ports []*string, portsUdp []*string, numGPUs int64) map[string]*ecs.Resource {
```
fierlion, line 4807: this is a well-written test case 🥇

agent/api/task/task.go
prateekchaudhry: remove `when` ToHostResources is used, `when` got lost somewhere, will update it

agent/api/task/task.go
prateekchaudhry: [MemoryReservation](https://github.com/aws/amazon-ecs-agent/blob/e28ada643fa37572031d7f9cf2f208bb52348f93/agent/vendor/github.com/docker/docker/api/types/container/host_config.go#L366) is the soft limit we are concerned with here, which is part of the HostConfig. I believe we are already checking if it is zero or negative by the check here `hostConfig.MemoryReservation <= 0`, so should be good imo.

fierlion: And that MemoryReservation field in the HostConfig is zero-initialized by default so we'll always have it if we don't have a null HostConfig.  Resolving.

fierlion (approved):

agent/api/task/task.go
prateekchaudhry: Yes that's right

Yiyuanzzz (approved):

Realmonia (approved):

PR #3723, opened 2023-05-30
Remove task serialization and use host resource manager for task resources

prateekchaudhry:
### Summary

ECS Agent ensures host resources to be available on the instance before running a task. Currently this is implemented through `serialization` - when scheduling a task, agent waits for all previously stopping tasks, i.e. with `StopSequenceNumber` of payload from acs less than `seqnum` of payload of the requested task, to stop.

This PR removes this serialization behavior and instead uses `HostResourceManager` to schedule tasks and a FIFO task queue built into `docker_task_engine` to queue tasks instead. This will hence use cpu, memory, ports(tcp/udp) and number of gpus available to manage tasks, and start progressing tasks as soon as resources for them start becoming available - instead of all stopping tasks to stop.

### Implementation details

* Removes package `sequential_waitgroup`, and references related to `StartSequenceNumber` and `StopSequenceNumber` which are constructs related to task serialization

* Tasks get queued in a `waitingTaskQueue` and wait for host resources (managed through `HostResourceManager`) to become available. A goroutine `monitorQueuedTasks` dequeues and starts waking up each of the waiting the tasks as and when resources start becoming available. When it can not dequeue anymore because resources are not available, it waits

* When a task stops or when a new task arrives, it wakes up the `monitorQueuedTasks` in case it is blocked

* **Management of host resources** When a task gets resources accounted for by the `monitorQueuedTasks`, resources are consumed. When a task changes it knownStatus to `STOPPED` and emits a change of state, resources are released

* For Agent restarts, there is a `reconcileHostResources` implemented during `synchronizeState` which synchronizes `HostResourceManager` data structures according to known task states. If any container has been known to progressed beyond `ContainerStatusNone` state, then host resources are consumed.

Related PRs
* #3684 
* #3706 
* #3700

### Testing
Manually tested reconciliation behavior with agent restarts and verified resources are allocated correctly from agent logs
```
level=debug time=2023-06-01T00:55:50Z msg="Task host resources to account for" MEMORY=1024 PORTS_TCP=[] PORTS_UDP=[] GPU=0 taskArn="arn:aws:ecs:us-west-2:<>:task/taskAccounting/..." CPU=1024
```

`TestTaskWaitForHostResources` unit test to test task queueing/dequeuing

Remove task serialization and use host resource manager for task resources
### Licensing


Yiyuanzzz (approved):

agent/engine/docker_task_engine.go
```diff
@@ -238,6 +241,37 @@ func NewDockerTaskEngine(cfg *config.Config,
...
+	logger.Info("Reconciling host resources")
+	for _, task := range engine.state.AllTasks() {
+		taskStatus := task.GetKnownStatus()
+		resources := task.ToHostResources()
+
+		// Release stopped tasks host resources
+		// Call to release here for stopped tasks should always succeed
+		// Idempotent release call
+		if taskStatus.Terminal() {
+			err := engine.hostResourceManager.release(task.Arn, resources)
```
fierlion, line 256: when will this be useful?  Will it only ever be a no-op if the agent is always starting from zero?

agent/engine/docker_task_engine.go
```diff
@@ -288,6 +323,112 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
...
+			// Dequeue as many tasks as possible and start wake up their goroutines
+			for {
+				task, err := engine.topTask()
+				if err != nil {
+					break
+				}
+				taskHostResources := task.ToHostResources()
+				consumed := false
+				consumable, err := engine.hostResourceManager.consumableSafe(taskHostResources)
+				if err != nil {
```
yinyic, line 395: nit- redundant check

fierlion (approved):
I'd like to see another test or three covering the engine as an opaque box.  You can add this as a follow up.

agent/engine/docker_task_engine.go
```diff
@@ -288,6 +323,112 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
...
+	select {
+	case engine.monitorQueuedTaskEvent <- struct{}{}:
+	default:
+		waitingTaskQueueSingleLen := false
+		engine.waitingTasksLock.Lock()
+		waitingTaskQueueSingleLen = len(engine.waitingTaskQueue) == 1
+		engine.waitingTasksLock.Unlock()
+		if waitingTaskQueueSingleLen {
+			engine.monitorQueuedTaskEvent <- struct{}{}
+		}
```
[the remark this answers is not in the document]
prateekchaudhry: Ack - making it the channel buffered and a no-op/empty default

agent/engine/docker_task_engine.go
prateekchaudhry: Not particularly useful right now, but this is a generalized implementation for keeping `HostResourceManager` in sync with engine. So this might find more uses in future, such as to keep periodic sync between engine and resource manager.

agent/engine/docker_task_engine.go
```diff
@@ -785,6 +908,15 @@ func (engine *DockerTaskEngine) deleteTask(task *apitask.Task) {
 }
 
 func (engine *DockerTaskEngine) emitTaskEvent(task *apitask.Task, reason string) {
+	if task.GetKnownStatus().Terminal() {
+		// Always do (idempotent) release host resources whenever state change with
+		// known status == STOPPED is done to ensure sync between tasks and host resource manager
+		resourcesToRelease := task.ToHostResources()
+		err := engine.hostResourceManager.release(task.Arn, resourcesToRelease)
+		if err != nil {
+			logger.Critical("Failed to release resources after tast stopped", logger.Fields{field.TaskARN: task.Arn})
```
Yiyuanzzz, line 917: non-blocking: small typo here "tast"

Yiyuanzzz (approved):

agent/engine/docker_task_engine.go
```diff
@@ -288,6 +323,96 @@ func (engine *DockerTaskEngine) Init(ctx context.Context) error {
 	return nil
 }
 
+// Always wakes up when at least one event arrives on buffered channel monitorQueuedTaskEvent
+// but does not block if monitorQueuedTasks is already processing queued tasks
```
yinyic, line 327: Can we elaborate a little more in the comment on 1) when we will be invoking this method (who will be sending messages onto the channel), and 2) why is buffer size of one sufficient (why's it okay to drop any additional messages)

agent/engine/docker_task_engine.go
```diff
@@ -469,6 +594,9 @@ func (engine *DockerTaskEngine) synchronizeState() {
 		engine.saveTaskData(task)
 	}
 
+	// Before starting managedTask goroutines, pre-allocate resources for already running
+	// tasks in host resource manager
```
yinyic, line 598: Not exactly "already running", more like tasks that have progressed beyond the resource consumption check

yinyic (approved):
Both comments are on comments, gonna approve to unblock. Please update the comments in the next PR (I'm assuming we'll have a follow-up with more/updated tests)

agent/engine/docker_task_engine.go
prateekchaudhry: Yes, will update these in follow up PR

PR #3728, opened 2023-06-01
Added integration test gMSA domainless

saikiranakula-amzn:
### Summary
Added gMSA domainless integration test

### Implementation details
This test does end-to-end test of gMSA domainless support on ECS

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/agent/config/parse_linux.go
+++ b/agent/config/parse_linux.go
@@ -27,7 +27,7 @@ import (
 )
 
 func parseGMSACapability() BooleanDefaultFalse {
-	envStatus := utils.ParseBool(os.Getenv(envGmsaEcsSupport), true)
+	envStatus := utils.ParseBool(os.Getenv(envGmsaEcsSupport), false)
 	if envStatus {
 		// Check if domain join check override is present
 		skipDomainJoinCheck := utils.ParseBool(os.Getenv(envSkipDomainJoinCheck), false)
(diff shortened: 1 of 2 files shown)


agent/config/parse_linux.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 func parseGMSACapability() BooleanDefaultFalse {
-	envStatus := utils.ParseBool(os.Getenv(envGmsaEcsSupport), true)
+	envStatus := utils.ParseBool(os.Getenv(envGmsaEcsSupport), false)
```
amogh09, line 30: Why this change?

agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -849,6 +849,102 @@ func TestGMSATaskFile(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	// Kill the existing container now
+	err = client.ContainerKill(context.TODO(), cid, "SIGKILL")
+	assert.NoError(t, err, "Could not kill container")
+
+	verifyTaskIsStopped(stateChangeEvents, testTask)
+
+	// Cleanup the test file
+	err = os.RemoveAll(testCredSpecFilePath)
```
amogh09, line 943: How about performing the cleanup in a `defer` call so that we always clean up?

agent/config/parse_linux.go
saikiranakula-amzn: by default it should be false, since customer need to use ECS_GMSA_SUPPORT env, variable, last time we missed it

agent/engine/engine_sudo_linux_integ_test.go
saikiranakula-amzn: sure will make that change

amogh09 (approved):

mythri-garaga (approved):

amogh09 (approved):

PR #3722, opened 2023-05-26
Add more v2, v3, and v4 task metadata tests

amogh09:
### Summary

* Add more tests for v2, v3, and v4 task metadata endpoints. Currently there are some error cases that are not covered by existing tests. This change fills that gap in coverage. 
* General improvements to the test file. 
    * Replace some global variables for expected responses with getter functions so that tests are free to mutate the expected responses if needed without affecting other tests. 
    * Consolidate all cases for v2, v3, and v4 task metadata endpoints into one test each to promote code reuse and improve readability and maintainability. 

### Implementation details

* Add `TestV2TaskMetadata`, `TestV3TaskMetadata`, and `TestV4TaskMetadata` functions for testing v2, v3, and v4 task metadata endpoints, respectively. These test functions include all happy and error cases. Remove existing tests for these endpoints as they are covered in the new test functions. 
* Replace some global variables with getter functions for better test isolation (tests are free to mutate the values returned by the getter functions without affecting other tests).

### Testing

New tests cover the changes:  yes

### Description for the changelog

Add more v2, v3, and v4 task metadata tests

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3729, opened 2023-06-01
Improve test coverage for taskWithTags endpoints

amogh09:
### Summary

Improve test coverage for v2, v3, and v4 taskWithTags endpoints. 

This change is in a series of test coverage improvements we are making for TMDS endpoints. Test coverage is being improved to gain more confidence in the ongoing TMDS refactoring work.

Previous PRs in the series are #3722 and #3708. This change improves the coverage to include the many error cases of taskWithTags endpoints. All happy and error cases are consolidated into one test function per endpoint. 

### Implementation details

* Added `TestV2TaskMetadataWithTags`, `TestV3TaskMetadataWithTags`, and `TestV4TaskMetadataWithTags` test functions for testing v2, v3, and v4 taskWithTags endpoints, respectively. 
* Each new test function tests all happy and error cases for the endpoint it tests. 
* Removed old test functions that are now redundant. 

### Testing

Improve test coverage for taskWithTags endpoints

### Licensing

### DIFF
--- a/agent/handlers/task_server_setup_test.go
+++ b/agent/handlers/task_server_setup_test.go
@@ -268,24 +268,7 @@ var (
 		},
 		Type: containerType,
 	}
-	expectedTaskResponse = v2.TaskResponse{
-		Cluster:       clusterName,
-		TaskARN:       taskARN,
-		Family:        family,
-		Revision:      version,
-		DesiredStatus: statusRunning,
-		KnownStatus:   statusRunning,
-		Containers:    []v2.ContainerResponse{expectedContainerResponse},
-		Limits: &v2.LimitsResponse{
-			CPU:    aws.Float64(cpu),
-			Memory: aws.Int64(memory),
-		},
-		PullStartedAt:      aws.Time(now.UTC()),
-		PullStoppedAt:      aws.Time(now.UTC()),
-		ExecutionStoppedAt: aws.Time(now.UTC()),
-		AvailabilityZone:   availabilityzone,
-	}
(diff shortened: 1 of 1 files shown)


agent/handlers/task_server_setup_test.go
```diff
@@ -2361,6 +2171,512 @@ func TestV4TaskMetadata(t *testing.T) {
 	})
 }
 
+func TestV2TaskMetadataWithTags(t *testing.T) {
```
Realmonia, line 2174: Can we do struct based testing instead for these 3 tests? I see they have similar content for each sub task, so this refactor might help reducing the duplicated code.

amogh09: We are already using struct based testing. The test is executed by `testTMDSRequest` function which is passed in a `TMDSTestCase` struct that contains the request, mock expectations, and the expected response code and body. `happyStateExpectations`, `containerInstanceTags`, `ecsInstanceTags` and other variables defined here are for reducing code duplication. 

Can you point out the duplicate code you are referring to?

Realmonia (approved):

agent/handlers/task_server_setup_test.go
amogh09: Thanks for the comment @Realmonia. We discussed offline that the three tests are already using struct based testing using `TMDSTestCase` struct. The reason we don't have a for loop is that some subtests require some setup steps which are trivial to include when using `t.Run` directly.

sparrc (approved):

PR #3731, opened 2023-06-02
Create TCS Handler in ecs-agent model

Realmonia:
### Summary

Create TCS Handler in ecs-agent module and corresponding agent component to integrate with this change. Also add some tests to TCS Client and TCS Handler in ecs-agent module.

### Implementation details

- `ecs-agent/tcs/handler`
  - Change the implementation to have a `TelemetrySession` interface and an implementation 
  - Add a test to verify channel message persists while TCS connection dced, and resume to flow when TCS connection is recovered
- `ecs-agent/tcs/client`
  - Switch from default seelog interface to structured log
  - Add a test to verify generating publish metrics from telemetry message can be paginated properly based on metric size (this case should not happen based on current metrics size)
  - Add a test to verify the telemetry message channels are non blocking
  - Add a test to verify if an invalid telemetry message is sent to channel, no problematic request will be sent (this theoretically should not happen)
  - Some naming changes
- `stats/reporter`
  - Add function to generates agentVersion information (for generating TACS endpoint url), this was originally in `agent/tcs/handler`
  - Add function to get TCS endpoint using `discoverTelemetryEndpoint` API. This was originally in `agent/tcs/handler`
  - Embed `ecs-agent/tcs/handler`'s `TelemetrySession` interface to "override" `Start` function, so that `discoverPollEndpoint` can also be part of the retry and backoff, while avoiding moving too many code (some of those are agent module specific) to ecs-agent module 

A follow up PR will remove the `agent/tcs/handler` as well as wire in the new TCS Handler (use session definition in reporter in `agent.go`)

### Testing

- Testing workflow in github passed.
- Manually verified the normal TCS workflow and TCS connect/disconnect/reconnect behavior is working fine from ecs-agent.log after using agent wired in the handler changes

Create TCS Handler in ecs-agent model

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -20,36 +20,29 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
-	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
-
-	dockerdoctor "github.com/aws/amazon-ecs-agent/agent/doctor" // for Docker specific container instance health checks
-	"github.com/aws/amazon-ecs-agent/agent/eni/watcher"
-	"github.com/aws/amazon-ecs-agent/ecs-agent/doctor"
-	"github.com/aws/aws-sdk-go/aws/awserr"
-
-	"github.com/aws/amazon-ecs-agent/agent/credentials/instancecreds"
-	"github.com/aws/amazon-ecs-agent/agent/engine/execcmd"
-	"github.com/aws/amazon-ecs-agent/agent/metrics"
-
 	acshandler "github.com/aws/amazon-ecs-agent/agent/acs/handler"
 	"github.com/aws/amazon-ecs-agent/agent/api"
 	"github.com/aws/amazon-ecs-agent/agent/api/ecsclient"
 	"github.com/aws/amazon-ecs-agent/agent/app/factory"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/containermetadata"
+	"github.com/aws/amazon-ecs-agent/agent/credentials/instancecreds"
 	"github.com/aws/amazon-ecs-agent/agent/data"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/sdkclientfactory"
+	dockerdoctor "github.com/aws/amazon-ecs-agent/agent/doctor" // for Docker specific container instance health checks
 	"github.com/aws/amazon-ecs-agent/agent/ec2"
 	"github.com/aws/amazon-ecs-agent/agent/ecscni"
 	"github.com/aws/amazon-ecs-agent/agent/engine"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
+	"github.com/aws/amazon-ecs-agent/agent/engine/execcmd"
 	engineserviceconnect "github.com/aws/amazon-ecs-agent/agent/engine/serviceconnect"
 	"github.com/aws/amazon-ecs-agent/agent/eni/pause"
+	"github.com/aws/amazon-ecs-agent/agent/eni/watcher"
 	"github.com/aws/amazon-ecs-agent/agent/eventhandler"
 	"github.com/aws/amazon-ecs-agent/agent/handlers"
+	"github.com/aws/amazon-ecs-agent/agent/metrics"
 	"github.com/aws/amazon-ecs-agent/agent/sighandlers"
 	"github.com/aws/amazon-ecs-agent/agent/sighandlers/exitcodes"
 	"github.com/aws/amazon-ecs-agent/agent/statemanager"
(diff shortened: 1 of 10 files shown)


agent/tcs/handler/types.go
```diff
@@ -37,7 +37,6 @@ type TelemetrySessionParams struct {
 	CredentialProvider            *credentials.Credentials
 	Cfg                           *config.Config
 	DeregisterInstanceEventStream *eventstream.EventStream
-	AcceptInvalidCert             bool
```
Realmonia, line 40: Removed because this parameter is not used. `Cfg.AcceptInsecureCert` is used instead.

agent/stats/reporter/reporter.go
Realmonia, line 1: This is a component for agent integration of the tcsHandler in ecs-agent module.

Yiyuanzzz: nit: Could we add more comments to those functions?

Yiyuanzzz (approved):

ecs-agent/tcs/handler/handler.go
```diff
@@ -0,0 +1,258 @@
...
+const (
+	// wsRWTimeout is the duration of read and write deadline for the websocket connection
+	deregisterContainerInstanceHandler = "TCSDeregisterContainerInstanceHandler"
+	ContainerRuntimeDocker             = "Docker"
+)
+
+// TelemetrySession defines an interface for handler's long-lived connection with TCS.
+type TelemetrySession interface {
+	StartTelemetrySession(context.Context, string) error
+	Start(context.Context) error
```
RichaGangwar, line 47: Why do we need two similar functions?

ecs-agent/tcs/handler/handler.go
```diff
@@ -0,0 +1,258 @@
...
+// connection is active
+func anyMessageHandler(client wsclient.ClientServer, wsRWTimeout time.Duration) func(interface{}) {
+	return func(interface{}) {
+		logger.Trace("TCS activity occurred")
+		// Reset read deadline as there's activity on the channel
+		if err := client.SetReadDeadline(time.Now().Add(wsRWTimeout)); err != nil {
+			logger.Warn("Unable to extend read deadline for TCS connection", logger.Fields{
+				field.Error: err,
+			})
+		}
```
RichaGangwar, line 239: Just for my understanding, why do we reset the heartbeat timer in other handlers, but set the read deadline in this handler? Why do we not have a similar behavior?

agent/stats/reporter/reporter.go
```diff
@@ -0,0 +1,140 @@
...
+			logger.Info("TCS Websocket connection closed for a valid reason")
+			backoff.Reset()
+		default:
+			seelog.Errorf("Error: lost websocket connection with ECS Telemetry service (TCS): %v", tcsError)
+			time.Sleep(backoff.Duration())
+		}
+	}
+}
+
+func NewDockerTelemetrySession(
```
RichaGangwar, line 60: Nit: Can we put NewDockerTelemetrySession before Start? I find it more aligned, not sure why. 🤭

agent/stats/reporter/reporter.go
Realmonia: Yep I will do it later when I actually call it :)

ecs-agent/tcs/handler/handler.go
```diff
@@ -0,0 +1,258 @@
...
+	tcsclient "github.com/aws/amazon-ecs-agent/ecs-agent/tcs/client"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/tcs/model/ecstcs"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/cihub/seelog"
+)
+
+const (
+	// wsRWTimeout is the duration of read and write deadline for the websocket connection
```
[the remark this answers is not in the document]
Realmonia: Thanks!

agent/stats/reporter/reporter.go
Realmonia: Sure

ecs-agent/tcs/handler/handler.go
Realmonia: There are two reasons:
- First, we have backup and retry logic in `Start` that calls `StartTelemetrySession`, which does not have retry logic in it. Having these separated will improve the readability and make it easier for us to test their individual functionality, as well as providing embed option to override it in docker/containerd agent.
- Second, there is an extra string parameter to passed into `StartTelemetrySession`, this is the endpoint string. Currently, `DiscoverTelemetryEndpoint` requires `api.ecsClient`, which has a lot of dependencies, and some of them are not natural to move to `ecs-agent` module. As pointed in the first point, with this setup, we can have an overridden `Start` function to do retries based on combined `DiscoverTelemetryEndpoint` and `StartTelemetrySession` result and do not need to duplicate too much code.

ecs-agent/tcs/handler/handler.go
Realmonia: This is current implementation that I did not look deeply into. IIUC, heartbeat handlers and any message handlers serve different purpose (and seems to be double guarantee) - heartbeat handler is to reset the timer, which TCS client uses to actively DCed TCS client connection, while any message handler generally extend the allowed time we call read for the messages, and that is a required parameter from gorilla websocket connection. Specifically, `wsRWTimeout = 2*heartbeatTimeout + heartbeatJitterMax` and therefore should not supposed to be triggered but is necessary to be set so that the gorilla ws never rejects read before we actively do so. I think we can move this block to every request handler, but to reduce duplication and necessary error handling, the original implementation choose to implement it this way.

RichaGangwar: Got it. Thank you for explaining.

Yiyuanzzz (approved):

RichaGangwar (approved):
Looks good to me.

Yiyuanzzz (approved):

agent/app/agent.go
```diff
@@ -866,6 +864,13 @@ func (agent *ecsAgent) startAsyncRoutines(
...
 	go eventhandler.HandleEngineEvents(agent.ctx, taskEngine, client, taskHandler, attachmentEventHandler)
 
+	err := statsEngine.MustInit(agent.ctx, taskEngine, agent.cfg.Cluster, agent.containerInstanceARN)
+	if err != nil {
+		seelog.Warnf("Error initializing metrics engine: %v", err)
+		return
+	}
+	go statsEngine.StartMetricsPublish()
+
 	telemetrySessionParams := tcshandler.TelemetrySessionParams{
```
RichaGangwar, line 874: Sorry I missed it before, but why are we still using the old handler instead of the new shared one in ecs-agent.

RichaGangwar: Oops! Looks like there is a PR open yet to be merged. Will wait for it.

PR #3743, opened 2023-06-08
Integrate with tcsHandler in ecs-agent module

Realmonia:
### Summary

Integrate with the tcsHandler (and its extension) in ecs-agent module we created in https://github.com/aws/amazon-ecs-agent/pull/3731 and remove tcsHandler from agent module. Also slightly modified `reporter`, so it works with current existing agent unit test.

### Implementation details

See summary.

### Testing

- Testing workflow in github passed.
- Manually verified the normal TCS workflow and TCS connect/disconnect/reconnect behavior is working fine.

Integrate with tcsHandler in ecs-agent module

### Licensing


agent/stats/reporter/reporter.go
```diff
@@ -48,22 +60,20 @@ func NewDockerTelemetrySession(
...
 	}
 	if ok {
-		seelog.Warnf("Metrics were disabled, not starting the telemetry session")
-		return nil
+		logger.Warn("Metrics were disabled, not starting the telemetry session")
+		return nil, nil
 	}
 
 	agentVersion, agentHash, containerRuntimeVersion := generateVersionInfo(taskEngine)
-	if cfg == nil {
```
chienhanlin, line 63: Just want to confirm, line 63-66 are removed because function `isContainerHealthMetricsDisabled()` will return an error if cfg is empty/nil. And this error will be caught and returned in line 65-69.

agent/stats/reporter/reporter.go
```diff
@@ -99,18 +109,23 @@ func NewDockerTelemetrySession(
 func (session *DockerTelemetrySession) Start(ctx context.Context) error {
 	backoff := retry.NewExponentialBackoff(time.Second, 1*time.Minute, 0.2, 2)
```
chienhanlin, line 110: nit & non-blocking comment. We can define some constants to configure the exp back-off behavior to make them less magical : )

agent/stats/reporter/reporter.go
```diff
@@ -99,18 +109,23 @@ func NewDockerTelemetrySession(
...
 	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("TCS session exited cleanly.")
```
chienhanlin, line 114: nit & non-blocking. Maybe we should use `ECS Telemetry service (TCS)` in log messages for consistency.

agent/stats/reporter/reporter.go
```diff
@@ -99,18 +109,23 @@ func NewDockerTelemetrySession(
...
+		}
 		endpoint, tcsError := discoverPollEndpoint(session.containerInstanceArn, session.ecsClient)
 		if tcsError == nil {
 			tcsError = session.s.StartTelemetrySession(ctx, endpoint)
 		}
-		switch tcsError {
-		case context.Canceled, context.DeadlineExceeded:
-			return tcsError
-		case io.EOF, nil:
+		if tcsError == nil || tcsError == io.EOF {
```
chienhanlin, line 122: nit. Could we add some comments to explain when the back-off will be rest and when will the new session to be started? 

I also have a question for line 120 and line 123. It seems like we start a new telemetry session if tcsError is nil in line 120. Would we ever hit `tcsError == nil` in line 122? And do we need to valid `endpoint` returned from function `discoverPollEndpoint()` in line 118 before starting a new session? 

Thanks!

agent/stats/reporter/reporter.go
Realmonia: Yes that's correct.

agent/stats/reporter/reporter.go
Realmonia: Will update. I think we have this in ACS logging too. We can change it in ACS changes.

agent/stats/reporter/reporter.go
Realmonia: Will do.

agent/stats/reporter/reporter.go
Realmonia: Sure will add. In short, backoff will reset if the connection closed for valid reason, and will not be reset when unexpected disconnection happened.

For your question:
- We could hit `tcsError == nil` in line 122, because `tcsError` is reassigned with `session.s.StartTelemetrySession(ctx, endpoint)`'s return value. If we take a look, that is the return value of https://github.com/aws/amazon-ecs-agent/blob/master/agent/tcs/handler/handler.go#L89-L101, i.e. when agent is exiting, or TCS websocket connection expired (as expected), we hit this condition.
- We would need to verify endpoint return from `discoverPollEndpoint`, because the function call itself is in the infinite loop, and if the return result is not valid, the `StartTelemetrySession` return will go into `default` block, and thus will be called again. Besides, there are really no way for us to verify it besides actually attempt to connect. What's your thought?

Realmonia: There will not be a new session; a new connection will be established when `StartTelemetrySession` returns

chienhanlin: I see. Thanks for sharing all these details! And adding a comment on line 124 is very helpful since it's not very intuitive to connect the function named "StartTelemetrySession" to the action "disconnection". 
```
	if tcsError == nil {
			// returning from StartTelemetrySession indicates a disconnection, need to reconnect.
			tcsError = session.s.StartTelemetrySession(ctx, endpoint)
		}
```

chienhanlin (dismissed):

prateekchaudhry (dismissed):

prateekchaudhry (dismissed):

RichaGangwar (approved):

prateekchaudhry (approved):

PR #3744, opened 2023-06-09
Refactor ACS attach task ENI message handling

danehlim:
### Summary

Refactor ACS attach task ENI message handling, update other associated code, and clean up tests.

### Implementation details

1. Refactor ACS attach task ENI message handling and move to `ecs-agent/`
   - `agent/acs/handler/attach_task_eni_handler.go` -> `ecs-agent/acs/session/attach_task_eni_responder.go`
      - Refactor `attachTaskENIHandler` struct and existing code into attachTaskENIResponder `struct`, such that `attachTaskENIResponder` struct implements the `wsclient.RequestResponder` interface
      - Handle ENI attachment in async (no reason for this to be a blocking operation)
      - Consider an attach task ENI message with multiple network interface devices to be valid, in the event that ACS does send one
         - A similar change has already been previously done for payload messages ([see here](https://github.com/aws/amazon-ecs-agent/commit/6ca1b3b4d5b0f47568f85b618931e8c446e34f8c)), this simply just a followup to that
      - Perform more validation of the ENI(s) in the attach task ENI message than just validating MAC address is present
      - Log with level `ERROR` instead of `WARN` when attach task ENI message is invalid
   - Update tests accordingly based off of above changes 
      - `agent/acs/handler/attach_task_eni_handler_test.go` (partial) -> `agent/acs/handler/attach_task_eni_responder_test.go`
         - Tests that need to interact with Agent state belong in this file
      - `agent/acs/handler/attach_task_eni_handler_test.go` (partial) -> `ecs-agent/acs/session/attach_task_eni_responder_test.go`
        - Tests that need to interact only with stateless components of the responder (e.g., message validation) belong in this file
      - Initialize attach task ENI responder and add its `HandlerFunc` function as a request handler to the ACS client, such that attach task ENI responder dictates what to do in response to receiving attach task ENI messages from ACS
         - Eventually, each ACS message type will have its own responder (which implements the `wsclient.RequestResponder` interface)  that dictates what to do in response to ACS messages of that specific type
3. General cleanup of tests, most notably:
   - Increase test coverage
   - Eliminate redundant initialization of test attachTaskENIMessage variables for each attach task ENI responder test that requires one 
   - Extract some testing constants commonly used across multiple ACS responders into centralized file
      - Only constants that were used in both tests for attach task/instance ENI responders as well as tests for other responder(s) were extracted into `ecs-agent/acs/session/testconst/test_const.go`
      - Not in the scope of this pull request: moving all ACS responders' common testing constants to the centralized file and refactoring all ACS responders testing code to use testing constants instead of hard-coded values (when applicable). More testing constants will be moved to the centralized file as we continue to refactor other ACS responders in future pull requests

### Testing

Refactor ACS attach task ENI message handling

### Licensing


agent/acs/handler/acs_handler.go
```diff
@@ -325,9 +312,13 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 	responseSender := func(response interface{}) error {
 		return client.MakeRequest(response)
 	}
-
-	heartbeatResponder := acssession.NewHeartbeatResponder(acsSession.doctor, responseSender)
-	client.AddRequestHandler(heartbeatResponder.HandlerFunc())
+	responders := []wsclient.RequestResponder{
+		acssession.NewAttachTaskENIResponder(eniHandler, responseSender),
```
Realmonia, line 316: don't see a `start()` and `defer stop()` here. What's changed?

danehlim: There is no need to do multi-thread synchronization with channels with this responder. As part of the refactor, we factor out this non-necessity and thus `start()` and `stop()` functions can be removed. The resulting responder is a lot cleaner and simpler to follow.

See the following PR for a similar change previously made to the heartbeat responder for the same reason:
https://github.com/aws/amazon-ecs-agent/pull/3232

Realmonia: Thanks!

Realmonia (approved):

PR #3747, opened 2023-06-11
Change reconcile/container update order on init and waitForHostResources/emitCurrentStatus order

prateekchaudhry:
### Summary

This PR fixes a bug figured out in task resource accounting on Agent Restarts. The issue is when Agent is down. a container stops (repro by `docker stop <container_id>`) and agent comes back up, task resources are not released by host resource manager.

Problems in current orders of calls are :
- `reconcileHostResources` (see #3723) is being called after stopped containers are accounted for in task engine. So if a container stops while the agent is stopped, reconcileHostResources may not pre-'consume' resources for a task, because it's overseeTask would still need to be run for cleanup if it's status has changed to stop.  As a result, `reconcileHostResources` should be called before `filterTasksToStartUnsafe` which updates the container and task statuses. This order is updated with comments.
- In overseeTask, `emitCurrentStatus` also does an `emitTaskEvent` call - which releases 'consumed' resources in host resource manager (see `Management of host resources` in summary in #3723). And `waitForHostResources()` again re-allocates resources, leading to persistent accounted for resources in host resource manager in these cases (when tasks stop when agent is down). Changing it's order after `waitForHostResources()` call and updating comments.

This PR also makes a change in `host_resource_manager` and `ToHostResources` in `task.go` to dereference ports from `[]*string` to `[]string` during logging for more understandable debug logging. This change results in outputting actual port values instead of lvalues of the ports which might not be always relevant. See PORTS_TCP and PORTS_UDP after the change here
`[Debug] logger=structured msg="Consumed resources after task consume call" CPU=512 MEMORY=768 PORTS_TCP=[22 23] PORTS_UDP=[1000 1001] GPU=
1 taskArn="arn:aws:ecs:us-east-1:<aws_account_id>:task/cluster-name/11111"`

### Testing

With debug logs, verified agent restarts function properly and resources are accounted and released properly for following scenarios :
- Tasks are running while agent restarts
- Tasks stop while agent restarts - agent picks up stopped tasks, emits change of status and host resource manager releases resources
- Instance reboots - agent picks up stopped tasks, emits change of status and host resource manager releases resources

Fix Agent restarts and ports logging with task resource accounting
### Licensing

### DIFF
--- a/agent/engine/host_resource_manager.go
+++ b/agent/engine/host_resource_manager.go
@@ -23,6 +23,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
+	"github.com/aws/aws-sdk-go/aws"
 )
 
 const (
@@ -67,25 +68,13 @@ func (e *ResourceIsNilForTask) Error() string {
 	return fmt.Sprintf("resource %s is nil in task resources", e.resource)
 }
 
-func toStringSlice(s []*string) []string {
-	var t []string
-	for _, ptr := range s {
-		if ptr == nil {
-			t = append(t, "nil")
-		} else {
-			t = append(t, *ptr)
-		}
-	}
-	return t
-}
-
 func (h *HostResourceManager) logResources(msg string, taskArn string) {
 	logger.Debug(msg, logger.Fields{
 		"taskArn":   taskArn,
 		"CPU":       *h.consumedResource[CPU].IntegerValue,
 		"MEMORY":    *h.consumedResource[MEMORY].IntegerValue,
-		"PORTS_TCP": toStringSlice(h.consumedResource[PORTSTCP].StringSetValue),
-		"PORTS_UDP": toStringSlice(h.consumedResource[PORTSUDP].StringSetValue),
+		"PORTS_TCP": aws.StringValueSlice(h.consumedResource[PORTSTCP].StringSetValue),
+		"PORTS_UDP": aws.StringValueSlice(h.consumedResource[PORTSUDP].StringSetValue),
 		"GPU":       *h.consumedResource[GPU].IntegerValue,
 	})
 }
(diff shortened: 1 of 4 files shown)


Yiyuanzzz (dismissed):

agent/engine/host_resource_manager.go
```diff
@@ -67,13 +67,25 @@ func (e *ResourceIsNilForTask) Error() string {
...
+}
+
 func (h *HostResourceManager) logResources(msg string, taskArn string) {
 	logger.Debug(msg, logger.Fields{
 		"taskArn":   taskArn,
 		"CPU":       *h.consumedResource[CPU].IntegerValue,
 		"MEMORY":    *h.consumedResource[MEMORY].IntegerValue,
-		"PORTS_TCP": h.consumedResource[PORTSTCP].StringSetValue,
-		"PORTS_UDP": h.consumedResource[PORTSUDP].StringSetValue,
+		"PORTS_TCP": toStringSlice(h.consumedResource[PORTSTCP].StringSetValue),
```
singholt, line 87: I think you could use `aws.StringValueSlice()` to accomplish this

https://github.com/aws/aws-sdk-go/blob/main/aws/convert_types.go#L31

prateekchaudhry: Thanks! This is what I need.

yinyic (approved):

agent/engine/task_manager.go
```diff
@@ -199,12 +199,16 @@ func (mtask *managedTask) overseeTask() {
...
 	mtask.UpdateStatus()
-	// If this was a 'state restore', send all unsent statuses
-	mtask.emitCurrentStatus()
 
-	// Wait for host resources required by this task to become available
+	// Wait here until enough resources are available on host for the task to progress
+	// - Waits until host resource manager succesfully 'consume's task resources and returns
+	// - For tasks which have crossed this stage before (on agent restarts), resources are pre-consumed - returns immediately
+	// (resources are later 'release'd on Stopped task emitTaskEvent call)
 	mtask.waitForHostResources()
```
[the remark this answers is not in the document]
prateekchaudhry: Yes let me add that. Although for this particular 'restart case', resources would be detected as 'STOPPED', and `waitForHostResources` behavior is `resources are pre-consumed - returns immediately` i.e. without queueing.

singholt (approved):

Yiyuanzzz (dismissed):

yinyic (dismissed):

singholt (approved):

Yiyuanzzz (approved):

PR #3750, opened 2023-06-15
Dont consume host resources for tasks getting STOPPED while waiting in waitingTasksQueue

prateekchaudhry:
### Summary

Context: Task Resource Accounting feature implements queuing (`monitorQueuedTasks`) for tasks to go through where each task waits to 'consume' resources in `host_resource_manager`. `host_resource_manager` keeps an account of each task which has acquired resources to progress for task creation/running.

If a ACS `StopTask` request arrives at Agent while the task is still in this queue, overseeTask - waiting in `waitForHostResources`, falls through and does not wait for the event from `monitorQueuedTasks`, ends up calling 'host_resource_manager.release()` during emitTaskEvent.

But if these `STOPPED` tasks are dequeued later (in docker_task_engine), they just write to the channel with no listeners after the `host_resource_manager.consume()` call in `docker_task_engine`, and the resources persist in host_resource_manager.

This PR fixes this by not calling `host_resource_manager.consume()` for tasks in `monitorQueuedTasks` whose desired status has changed to `STOPPED`. Some cautionary implementation steps have been taken to isolate working of `monitorQueuedTasks` as described in implementation

### Implementation
* Added a `monitorQueuedTasksLock` for the main body of the loop in `monitorQueuedTasks`. This is to synchronize the processing of the topTask in `monitorQueuedTasks` and any parallel update to its desired status - say from ACS updates
* The main body of this processing in `monitorQueuedTasks` has been moved to a method `tryDequeueWaitingTasks` - to put into this critical section of `monitorQueuedTasksLock`
* `monitorQueuedTasksLock` is also used in `handleDesiredStatusChange` which updates the desired status of a managed task
* Adds an integ test `TestHostResourceManagerStopTaskNotBlockWaitingTasks` to test this behavior
* Made container names unique in existing test `TestHostResourceManagerResourceUtilization` for easier debugging

### As a result of this implementation, consider the scenarios
1. StopTask arrives and is processed when the task is not yet processed by `monitorQueuedTasks`
 - This is safe because when task is dequeued, it's desired status is read as `STOPPED` and queue returns on `consumedHostResourceEvent` without consuming resources

2. StopTask arrives while the task is being processed by `monitorQueuedTasks`
 - Due to critical sections put in place, desired status will be updated after `monitorQueuedTasks` is done processing. Resources will be consumed, and released in `emitTaskEvent`

3. StopTask arrives and is processed any time after `monitorQueuedTasks` is done processing
 - Desired status will be updated after `monitorQueuedTasks` is done processing. Resources will be consumed and released in `emitTaskEvent`

### Related Containers Roadmap Issue
https://github.com/aws/containers-roadmap/issues/325

### Testing

Verified by
- Starting and stopping a task with `stopTimeout` of 300s. Starting another same task and stopped it. This task goes to the waiting queue and also returns in overseeTask goroutine. Later this task does not call consume and checked debug logs
```
level=info time=2023-06-15T07:22:18Z msg="Task desired status STOPPED while waiting for host resources, not consuming" taskARN="arn:aws:ecs:us-west-2:<account_id>task/taskAccounting/aa77bc7083d9401b8f4aac32424abed2"
```
- Added an integ test `TestHostResourceManagerStopTaskNotBlockWaitingTasks` and verified it succeeds. This test simulates ACS stopTask for a tasks stuck in `waitingTasksQueue` and verifies the resources are released in host_resource_manager

New tests cover the changes: 
Yes

Dont consume host resources for tasks getting STOPPED while waiting 
### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -155,8 +155,13 @@ type DockerTaskEngine struct {
 	// all tasks, it must not acquire it for any significant duration
 	// The write mutex should be taken when adding and removing tasks from managedTasks.
 	tasksLock sync.RWMutex
-	// waitingTasksLock is a mutext for operations on waitingTasksQueue
+	// waitingTasksLock is a mutex for operations on waitingTasksQueue
 	waitingTasksLock sync.RWMutex
+
+	// monitorQueuedTasksLock is a mutex for operations in the monitorQueuedTasks which
+	// allocate host resources and wakes up waiting host resources. This should be used
+	// for synchronizing task desired status updates and queue operations
+	monitorQueuedTasksLock sync.RWMutex
 
 	credentialsManager                  credentials.Manager
 	_time                               ttime.Time
@@ -392,21 +397,47 @@ func (engine *DockerTaskEngine) monitorQueuedTasks(ctx context.Context) {
 				if err != nil {
 					break
 				}
-				taskHostResources := task.ToHostResources()
-				consumed, err := task.engine.hostResourceManager.consume(task.Arn, taskHostResources)
-				if err != nil {
-					engine.failWaitingTask(err)
-				}
-				if consumed {
-					engine.startWaitingTask()
-				} else {
-					// not consumed, go to wait
+				dequeuedTask := engine.tryDequeueWaitingTasks(task)
+				if !dequeuedTask {
 					break
 				}
 			}
 			logger.Debug("No more tasks could be started at this moment, waiting")
 		}
 	}
+}
+
+func (engine *DockerTaskEngine) tryDequeueWaitingTasks(task *managedTask) bool {
+	// Isolate monitorQueuedTasks processing from changes of desired status updates to prevent
+	// unexpected updates to host resource manager when tasks are being processed by monitorQueuedTasks
+	// For example when ACS StopTask event updates arrives and simultaneously monitorQueuedTasks
+	// could be processing
+	engine.monitorQueuedTasksLock.Lock()
+	defer engine.monitorQueuedTasksLock.Unlock()
+	taskDesiredStatus := task.GetDesiredStatus()
+	if taskDesiredStatus.Terminal() {
+		logger.Info("Task desired status changed to STOPPED while waiting for host resources, progressing without consuming resources", logger.Fields{field.TaskARN: task.Arn})
+		engine.returnWaitingTask()
+		return true
+	}
+	taskHostResources := task.ToHostResources()
+	consumed, err := task.engine.hostResourceManager.consume(task.Arn, taskHostResources)
+	if err != nil {
+		engine.failWaitingTask(err)
+		return true
+	}
+	if consumed {
+		engine.startWaitingTask()
+		return true
+	}
+	return false
+	// not consumed, go to wait
+}
+
+// To be called when resources are not to be consumed by host resource manager, just dequeues and returns
+func (engine *DockerTaskEngine) returnWaitingTask() {
+	task, _ := engine.dequeueTask()
+	task.consumedHostResourceEvent <- struct{}{}
 }
 
 func (engine *DockerTaskEngine) failWaitingTask(err error) {
--- a/agent/engine/engine_unix_integ_test.go
+++ b/agent/engine/engine_unix_integ_test.go
@@ -1227,7 +1227,7 @@ func TestHostResourceManagerResourceUtilization(t *testing.T) {
 		testTask := createTestTask(taskArn)
 
 		// create container
-		A := createTestContainerWithImageAndName(baseImageForOS, "A")
+		A := createTestContainerWithImageAndName(baseImageForOS, fmt.Sprintf("A-%d", i))
 		A.EntryPoint = &entryPointForOS
 		A.Command = []string{"trap shortsleep SIGTERM; shortsleep() { sleep 6; exit 1; }; sleep 10"}
 		A.Essential = true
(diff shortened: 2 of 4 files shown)


Yiyuanzzz (dismissed):

agent/engine/engine_unix_integ_test.go
```diff
@@ -1227,7 +1227,7 @@ func TestHostResourceManagerResourceUtilization(t *testing.T) {
 		testTask := createTestTask(taskArn)
 
 		// create container
-		A := createTestContainerWithImageAndName(baseImageForOS, "A")
+		A := createTestContainerWithImageAndName(baseImageForOS, fmt.Sprintf("A-%d", i))
```
prateekchaudhry, line 1230: This is an unrelated change in another test case related to task resource accounting. Can make debugging easier in future if each task and container are named uniquely.

agent/engine/docker_task_engine.go
```diff
@@ -409,6 +407,39 @@ func (engine *DockerTaskEngine) monitorQueuedTasks(ctx context.Context) {
 	}
 }
 
+func (engine *DockerTaskEngine) tryDequeueWaitingTasks(task *managedTask) bool {
+	// Isolate monitorQueuedTasks processing from changes of desired status updates to prevent
+	// unexpected updates to host resource manager when tasks are being processed by monitorQueuedTasks
+	// For example when ACS StopTask event updates arrives and simultaneously monitorQueuedTasks
+	// could be processing
+	engine.monitorQueuedTasksLock.Lock()
+	defer engine.monitorQueuedTasksLock.Unlock()
```
Realmonia, line 416: Should this be read lock? I suppose it does not matter as there will be only one instance of `tryDequeueWaitingTasks` but still theoretically I think this should be a `RLock()` and `RUnlock()`

Realmonia: Besides, would critical region only be after line 417 (or line 422) since I don't think a task's desired status can change from `stopped` to earlier states? Do we see a potential that things after line 422 takes too much time and holding the lock too long?

Realmonia:
Also a general question with agent restart (sorry if this is me missing some context) - how are we storing tasks in queue to statefile (boltdb) when agent restart or recover from unexpected exit?

agent/engine/docker_task_engine.go
prateekchaudhry: Once topTask's `taskDesiredStatus` has been read by `tryDequeueWaitingTasks`, it should not be updated anywhere else while this step is in progress. So imo it, the read call itself `task.GetDesiredStatus()` should be in the critical section, hence I declared it within the lock boundary.

The rest of the function will only execute based on the `taskDesiredStatus` read here, so I think it should stay in the critical section

Also, improbable but `failWaitingTask` can change desired status within this function - when the task definition itself has problems. So it I think it might be better to use `Lock` instead of `Rlock` here.

`too much time and holding the lock too long` - I think the method calls are fairly small 
- ToHostResources - this is a method call to map from task struct to host resources
- `dequeueTask` - dequeue from `waitingTasksQueue`
- `hostResourceManager.consume` - updating host resource manager resources - loops over resources and acquires resources if possible
- `task.consumedHostResourceEvent <- struct{}{}` - this is just a write on a buffered channel and returns immediately

Realmonia: Thanks! Make sense to me.

prateekchaudhry:
> Also a general question with agent restart (sorry if this is me missing some context) - how are we storing tasks in queue to statefile (boltdb) when agent restart or recover from unexpected exit?

We do not store tasks in queue to statefile. On restarts, `reconcileHostResources` method in `docker_task_engine.go` takes the tasks which have progressed beyond `waitForHostResources` state and restores the `host_resource_manager` state.

So a task which has not progressed beyond that state will queue up again, as on restart, each task's [startTask](https://github.com/prateekchaudhry/amazon-ecs-agent/blob/taskAccountingAcsStopTask/agent/engine/docker_task_engine.go) is called again. Those tasks which are not pre-allocated during `reconcileHostResources` will thus end up queuing up again.

agent/engine/docker_task_engine.go
Realmonia: Oh, one thing is that `consume` itself also seems to acquire a different lock. Not sure if this is a valid case, but I would be cautious about the case where there are both 
- incoming new task placement, which enters the queue and acquire `monitorQueuedTasksLock` and `hostResourceManagerRWLock`
- tasks from restarting the agent, which directly acquires `hostResourceManagerRWLock` without entering the queue

Do we have protection mechanism for the second?

agent/engine/task_manager.go
```diff
@@ -386,6 +387,14 @@ func (mtask *managedTask) waitEvent(stopWaiting <-chan struct{}) bool {
 func (mtask *managedTask) handleDesiredStatusChange(desiredStatus apitaskstatus.TaskStatus, seqnum int64) {
 	// Handle acs message changes this task's desired status to whatever
 	// acs says it should be if it is compatible
+
+	// Isolate change of desired status updates from monitorQueuedTasks processing to prevent
+	// unexpected updates to host resource manager when tasks are being processed by monitorQueuedTasks
+	// For example when ACS StopTask event updates arrives and simultaneously monitorQueuedTasks
+	// could be processing
+	mtask.engine.monitorQueuedTasksLock.Lock()
+	defer mtask.engine.monitorQueuedTasksLock.Unlock()
```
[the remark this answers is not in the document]
Realmonia: Makes sense.

Realmonia:
> We do not store tasks in queue to statefile. On restarts, `reconcileHostResources` method in `docker_task_engine.go` takes the tasks which have progressed beyond `waitForHostResources` state and restores the `host_resource_manager` state.
> 
> So a task which has not progressed beyond that state will queue up again, as on restart, each task's [startTask](https://github.com/prateekchaudhry/amazon-ecs-agent/blob/taskAccountingAcsStopTask/agent/engine/docker_task_engine.go) is called again. Those tasks which are not pre-allocated during `reconcileHostResources` will thus end up queuing up again.

Oh I understand now. So those tasks are kept in state like `TaskCreated`. SGTM.

agent/engine/docker_task_engine.go
prateekchaudhry: Yes, so tasks on restarts pre-acquire resources in [reconcileHostResources](https://github.com/prateekchaudhry/amazon-ecs-agent/blob/taskAccountingAcsStopTask/agent/engine/docker_task_engine.go#L638) before the queue goroutine `monitorQueuedTasks` starts. So this should be taken care of.

agent/engine/task_manager.go
```diff
@@ -1520,6 +1532,22 @@ func (mtask *managedTask) discardEvents() {
 	}
 }
 
+func (mtask *managedTask) discardConsumedHostResourceEvents() {
+	for {
+		select {
+		case <-mtask.consumedHostResourceEvent:
+			logger.Info("Releasing resources in cleanup", logger.Fields{field.TaskARN: mtask.Arn})
+			resourcesToRelease := mtask.ToHostResources()
+			err := mtask.engine.hostResourceManager.release(mtask.Arn, resourcesToRelease)
```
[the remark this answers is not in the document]
Realmonia: SGTM

agent/engine/docker_task_engine.go
Realmonia: Great!

Realmonia (approved):

fierlion (approved):

PR #3751, opened 2023-06-16
Move periodic timeout implementation to wsclient library

RichaGangwar:
### Summary
Move the periodic timeout implemented by ACS and TACS to common websocket client library. 

### Implementation details
1. Added the disconnect timer to wsclient library.
2. Removed the disconnect timer from TACS handler.
3. Remove the disconnect timer from ACS session. 

### Testing

Move the periodic timeout implemented by ACS and TACS to common websocket client library. 

### Licensing


agent/acs/handler/acs_handler.go
```diff
@@ -24,6 +24,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/ecs-agent/metrics"
```
Realmonia, line 27: nit: put this to other ecs-agent imports.

agent/acs/handler/acs_handler.go
```diff
@@ -346,6 +353,20 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
 	return client.Serve(acsSession.ctx)
 }
 
+// newHeartbeatTimer creates a new time object, with a callback to
+// disconnect from ACS on inactivity
+func newHeartbeatTimer(client wsclient.ClientServer, timeout time.Duration, jitter time.Duration) ttime.Timer {
```
Realmonia, line 358: Wondering if this is something we can put into shared wsclient interface, since both ACS and TACS should have a very similar implementation.

agent/app/agent.go
```diff
@@ -61,6 +61,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/eventstream"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
+	ecs_agent_metrics "github.com/aws/amazon-ecs-agent/ecs-agent/metrics"
```
Realmonia, line 64: nit: import it as `metrics` looks better to me.

ecs-agent/tcs/handler/handler_test.go
```diff
@@ -345,7 +346,7 @@ func TestConnectionInactiveTimeout(t *testing.T) {
 		testCreds,
 		testCfg,
 		deregisterInstanceEventStream,
-		50*time.Millisecond,
+		5*time.Second,
```
Realmonia, line 349: Q: I think the reason we have this to a smaller value is to get the result faster. 5s is not long for me either but want to check is there any reason we have to change this?

ecs-agent/tcs/handler/handler.go
Realmonia, line 1: We should be able to remove this line
```
//lint:file-ignore U1000 Ignore unused metricsFactory field as it is only used by Fargate
```

agent/app/agent.go
RichaGangwar: There is an already imported metrics at line 46. Thats the reason I used another name to avoid confusion.

agent/acs/handler/acs_handler.go
RichaGangwar: Yeah, I think I tried that but changed it back. Somehow forgetting what was the reason. One reason I think of is the logs would not be able to indicate correctly which type of client received heartbeat, i.e. TACS or ACS. Let me know if we can work around that, then I can move the handler to wsclient.

ecs-agent/tcs/handler/handler.go
RichaGangwar: Got it. Will remove it. Thank you for pointing it out.

ecs-agent/wsclient/client_test.go
```diff
@@ -422,3 +446,37 @@ func TestCtxCancel(t *testing.T) {
...
+	mockServer.StartTLS()
+
+	types := []interface{}{ecsacs.AckRequest{}}
+	cs := getTestClientServer(mockServer.URL, types, 1)
+	// Setting up a lower disconnect timer value for testing.
+	timer, err := cs.Connect(mockDisconnectTimeoutMetricName, 10*time.Second, 2*time.Second)
+	require.NoError(t, err)
+	defer timer.Stop()
+	assert.True(t, cs.IsReady(), "expected websocket connection to be ready")
+
```
danehlim, line 464: The test from this line in below seems to be somewhat of a hack and potentially overcomplicated. Could we not just wait for the max possible amount of time it would take for the client to reach its periodic disconnect logic, and then assert that the websocket connection was in fact closed by checking against what is returned by calling `cs.IsConnected()`?

agent/acs/handler/acs_handler.go
```diff
@@ -281,17 +281,18 @@ func (acsSession *session) startACSSession(client wsclient.ClientServer) error {
...
 
-	err := client.Connect()
+	disconnectTimer, err := client.Connect(metrics.ACSDisconnectTimeoutMetricName,
+		acsSession.connectionTime,
+		acsSession.connectionJitter)
 	if err != nil {
 		seelog.Errorf("Error connecting to ACS: %v", err)
 		return err
 	}
+	if disconnectTimer != nil {
```
danehlim, line 291: Is this check necessary? My understanding is that this shouldn't ever be nil if err == nil (which we assert/check on L287).

agent/acs/handler/acs_handler_test.go
```diff
@@ -202,13 +204,14 @@ func TestHandlerReconnectsOnConnectErrors(t *testing.T) {
 	mockWsClient.EXPECT().Close().Return(nil).AnyTimes()
 	gomock.InOrder(
 		// Connect fails 10 times
-		mockWsClient.EXPECT().Connect().Return(io.EOF).Times(10),
+		mockWsClient.EXPECT().Connect(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, io.EOF).Times(10),
```
danehlim, line 207: Could we pass in `metrics.ACSDisconnectTimeoutMetricName` (if passing in metric name to the Connect method is deemed necessary), `wsclient.WSclientDisconnectTimeout`, and  `wsclient.WSclientDisconnectJitterMax`, instead of `gomock.Any()`, `gomock.Any()`, and `gomock.Any()` here to be more specific? (this comment applies to other lines of this file where the same change is made as well)

agent/acs/handler/acs_handler_test.go
```diff
@@ -202,13 +204,14 @@ func TestHandlerReconnectsOnConnectErrors(t *testing.T) {
...
 		// Cancel trying to connect to ACS on the 11th attempt
 		// Failure to retry on Connect() errors should cause the
 		// test to time out as the context is never cancelled
-		mockWsClient.EXPECT().Connect().Do(func() {
+
+		mockWsClient.EXPECT().Connect(gomock.Any(), gomock.Any(), gomock.Any()).Do(func(interface{}, interface{}, interface{}) {
 			cancel()
-		}).Return(nil).MinTimes(1),
+		}).Return(nil, nil).MinTimes(1),
```
danehlim, line 214: The first `nil` here should be some sort of mock or dummy disconnect timer, right? (this comment applies to other lines of this file where the same change is made as well)

agent/acs/handler/acs_handler.go
```diff
@@ -97,6 +93,7 @@ type session struct {
 	ctx                             context.Context
 	cancel                          context.CancelFunc
 	backoff                         retry.Backoff
+	metricsFactory                  metrics.EntryFactory
```
danehlim, line 96: nit: We can replace usages of the no-op metricsFactory defined on L260 with this new one so that we only have a single no-op metricsFactory being used.

ecs-agent/wsclient/client.go
```diff
@@ -190,18 +206,21 @@ type MakeRequestHookFunc func([]byte) ([]byte, error)
 // Connect opens a connection to the backend and upgrades it to a websocket. Calls to
 // 'MakeRequest' can be made after calling this, but responses will not be
 // receivable until 'Serve' is also called.
-func (cs *ClientServerImpl) Connect() error {
+func (cs *ClientServerImpl) Connect(disconnectMetricName string,
```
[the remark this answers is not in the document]
RichaGangwar: No, it would not always be metrics.ACSDisconnectTimeoutMetricName. Since we are moving this periodic timeout logic to websocket library, which is used by both ACS and TACS, we need to pass the disconnect metric name to specifiy which connection was disconnected, i.e. ACS or TACS. There is no agent-specific aspect here, but more to differentiate between diff connections which use the wsclient library.

agent/acs/handler/acs_handler.go
RichaGangwar: Sure.

agent/acs/handler/acs_handler.go
RichaGangwar: Its actually failing for a test, where we return nil and nil, because we do not create a timer in gomock. Let me go back and check if we can fix the test.

agent/acs/handler/acs_handler_test.go
RichaGangwar: Yes, you are right. This is the test due to which I added the nil [check.](https://github.com/aws/amazon-ecs-agent/pull/3751#discussion_r1303570546)
Let me create a timer in this mock to return.

agent/acs/handler/acs_handler_test.go
RichaGangwar: I don't think we need to do that. It depends on what we are trying to test. If the test is not specifically testing the timeout or metric emitting, then I think it is unimportant what value we pass. That's what i consider the beauty of using gomock. Let me know if you think otherwise and consider it to be important to send specific values.

ecs-agent/wsclient/client_test.go
```diff
@@ -42,7 +45,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const dockerEndpoint = "/var/run/docker.sock"
+const (
+	dockerEndpoint                  = "/var/run/docker.sock"
+	mockDisconnectTimeoutMetricName = "DisconnectTimeout"
```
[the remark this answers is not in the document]
danehlim: Makes sense, thanks for explaining!

agent/acs/handler/acs_handler_test.go
danehlim: Apologies, I should have marked my original comment here as a nit. IMO if we are certain of what will/should be passed in, then we should specify it to make for a stronger/more strict tests. In cases where we do not have such certainty, we could use `gomock.Any()`.

Again, this is just personal preference for me though and is non-blocking.

ecs-agent/wsclient/client.go
```diff
@@ -565,3 +590,42 @@ func permissibleCloseCode(err error) bool {
...
+
+// closeClient will attempt to close the provided client, retries are not recommended
+// as failure modes for this are when client is not found or already closed.
+func (cs *ClientServerImpl) CloseClient(startTime time.Time, timeoutDuration time.Duration) error {
+	logger.Warn(("Closing connection"), logger.Fields{
+		"ConnectionStartTime":  startTime.Format(dateTimeFormat),
+		"MaxDisconnectionTime": startTime.Add(timeoutDuration).Format(dateTimeFormat),
+	})
+	err := cs.Disconnect()
+	//err := cs.WriteCloseMessage()
```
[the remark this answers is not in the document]
danehlim: Prefer to use `cs.WriteCloseMessage()` as it is what is currently used for the existing periodic ACS websocket connection disconnect logic in `agent/acs/handler/acs_handler.go` and as you mentioned is actually more appropriate. 

To expand further:

`cs.Disconnect()` calls `cs.conn.Close` function which [according to documentation](https://pkg.go.dev/github.com/gorilla/websocket#Conn.Close):
> closes the underlying network connection without sending or waiting for a close message. 

In contrast, `cs.WriteCloseMessage()` calls cs.`conn.WriteControl` function which [according to documentation](https://pkg.go.dev/github.com/gorilla/websocket#Conn.WriteControl):
> writes a control message with the given deadline. The allowed message types are CloseMessage, PingMessage and PongMessage. 

The implications of this are described more in the [original pull request to for adding periodic disconnect from ACS logic](https://github.com/aws/amazon-ecs-agent/pull/3586):
> close message is sent to ACS which then is echoed back to the agent. This close message is added to the message queue. This ensures that the agent reads and handles unread messages first, before closing the connection. [Reference](https://www.rfc-editor.org/rfc/rfc6455#section-5.5.1)

ecs-agent/wsclient/client_test.go
danehlim: Yeah what I meant by "hack" was that the current way the test is designed is using some other indirect means to test by proxy the functionality (i.e., making a bunch of requests that will take longer to complete than the max connection time and waiting until we receive an error) when we could likely just test the functionality directly and with less unnecessary overhead (i.e., making sure the websocket connection is disconnected after the max connection time).

In this case I meant for "hack" to be more or less synonymous with overcomplicated -  apologies if that was unclear!

RichaGangwar:
Closing this PR since the ttbmacis does not show the run tests for this PR, because of the PR being too old. Will open a new one with same changes.

PR #3752, opened 2023-06-20
add negative integration tests for gmsa on Linux

mrkdeng:
### Summary

This PR is to add negative integration tests for gMSA on Linux, which covers three cases:
1. gMSA credentialspec file can't be found in S3 bucket
2. gMSA credentialspec file can't be found in SSM documents
3. gMSA Deamon is not running

### Implementation details

Add integration tests to `agent/engine/engine_sudo_linux_integ_test.go`

### Testing

New tests cover the changes: 
YES

### Licensing

### DIFF
--- a/agent/engine/engine_sudo_linux_integ_test.go
+++ b/agent/engine/engine_sudo_linux_integ_test.go
@@ -939,6 +939,159 @@ func TestGMSADomainlessTaskFile(t *testing.T) {
 	verifyTaskIsStopped(stateChangeEvents, testTask)
 }
 
+func TestGMSATaskFileS3Err(t *testing.T) {
+	t.Setenv("ECS_GMSA_SUPPORTED", "True")
+	t.Setenv("ZZZ_SKIP_DOMAIN_JOIN_CHECK_NOT_SUPPORTED_IN_PRODUCTION", "True")
+	t.Setenv("ZZZ_SKIP_CREDENTIALS_FETCHER_INVOCATION_CHECK_NOT_SUPPORTED_IN_PRODUCTION", "True")
+
+	cfg := defaultTestConfigIntegTest()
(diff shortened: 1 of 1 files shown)


agent/engine/engine_sudo_linux_integ_test.go
```diff
@@ -939,6 +939,157 @@ func TestGMSADomainlessTaskFile(t *testing.T) {
...
+	testTask.ResourcesMapUnsafe = make(map[string][]taskresource.TaskResource)
+	testTask.Containers[0].Command = getLongRunningCommand()
+
+	go taskEngine.AddTask(testTask)
+
+	err := verifyTaskIsRunning(stateChangeEvents, testTask)
+	assert.Error(t, err)
+}
+
+func TestGMSATaskFileSSMErr(t *testing.T) {
```
amogh09, line 981: Can we please either add a description above the tests explaining why task should not be created or update the error assertion to check for specific errors? 

Currently I am not able to tell what the test is testing without reading the test body.

saikiranakula-amzn: I will add add the description after this CR is merged for the tests

amogh09: Actually on second thought I think we should do both - add test description and have a tighter assertion on the error.

amogh09 (approved):

agent/engine/engine_sudo_linux_integ_test.go
amogh09: Let's add test descriptions in a follow-up PR.

amogh09:
Approved with the understanding that a follow-up PR will be created to add descriptions above the new test functions.

PR #3754, opened 2023-06-20
Internal Managed Daemon Definition

fierlion:
### Summary
Add the (internal-only) ManagedDaemon definition to the common library and update common lib vendor.

### Implementation details
The ManagedDaemon object encapsulates the core set of daemon definition parameters.  These will be used by the ECS Agent to start agent-only daemon processes to be used by the ECS Agent to extend its own capabilities.  
This current change includes the ManagedDaemon definition with a subset of getters/setters based on requirements.
Creation of a complete ManagedDaemon object will often require subsequent calls to set the Daemon Environment and the Daemon MountPoints -- each of these has its own specific get/set logic.
The Daemon has a generic Health Check definition with a runtime-specific method to export as a Docker HealthConfig.

### Testing
Validated locally with new unit tests included.

Add the ManagedDaemon definition to the common library.

### Licensing


ecs-agent/managed_daemon/mountpoint.go
```diff
@@ -0,0 +1,29 @@
...
+// permissions and limitations under the License.
+
+package manageddaemon
+
+type MountPoint struct {
+	// SourceVolumeID is used to identify the task volume globally, it's empty
+	// when for internal mount points.
+	SourceVolumeID string `json:"SourceVolumeID"`
+	// SourceVolume is the name of the source volume, it's unique within the task.
+	SourceVolume string `json:"SourceVolume,omitempty"`
```
mye956, line 21: nit(non-blocking): since it's the volume name, it might be more intuitive to call it something like `SourceVolumeName`

ecs-agent/managed_daemon/managed_daemon.go
```diff
@@ -0,0 +1,208 @@
...
+		return nil
+	}
+	return fmt.Errorf("EnvKey: %s already exists; will not add EnvVal: %s", envKey, envVal)
+}
+
+// Updates environment varable by evnKey
+// and will throw an error if the envKey
+// is not found
+func (md *ManagedDaemon) UpdateEnvVar(envKey string, envVal string) error {
+	_, ok := md.environment[envKey]
```
mye956, line 179: nit: could we use the same naming convention for this boolean? It switches between `exists` and `ok`

ecs-agent/managed_daemon/managed_daemon.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	md.environment[envKey] = envVal
+	return nil
+}
+
+// Deletes environment variable by envKey
+// and will throw an error if the envKey
+// is not found
+func (md *ManagedDaemon) DeleteEnvVar(envKey string) error {
+	_, ok := md.environment[envKey]
```
mye956, line 191: nit: same thing as L.179

ecs-agent/managed_daemon/managed_daemon.go
fierlion: That name was deliberate to differentiate between the expectation that it _shouldn't_ exist in one case and _should_ exist in the other.

ecs-agent/managed_daemon/mountpoint.go
fierlion: This is actually meant to mirror the incoming containerd agent's Mountpoint definition.

ecs-agent/managed_daemon/managed_daemon.go
mye956: Gotcha, I guess it's up to personal preference of interpreting. Either way LGTM.

ecs-agent/managed_daemon/mountpoint.go
mye956: I see, thanks for clarifying.

mye956 (approved):

prateekchaudhry (approved):

Issue #3429, opened 2022-10-12
Environment variable file behavior doesn't match documentation


ltm:
### Summary

According to the documentation of [environment variable files](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/taskdef-envfiles.html#taskdef-envfiles-considerations) ECS is using the same syntax as [Docker Compose](https://docs.docker.com/compose/env-file/). However, in practice ECS uses a simpler syntax.

### Description

The Docker Compose environment file syntax allows quoted values, inline comments, shell escape sequences, etc. and relies on a [fork](https://github.com/compose-spec/compose-go/tree/master/dotenv) of [godotenv](https://github.com/joho/godotenv) to [parse](https://github.com/compose-spec/compose-go/blob/517cf49833c044cb7670e4b2429871c19b1b28a8/cli/options.go#L208-L260) environment files.

The ECS agent simply [splits lines](https://github.com/aws/amazon-ecs-agent/blob/67240c17711c0d0a5c7b1d37882dc565880bff38/agent/taskresource/envFiles/envfile.go#L552-L588) in the environment file into key-value pairs.

### Expected Behavior
An environment file containing `VAR="some\nvalue"` should set the environment variable `VAR` to:
```
some
value
```

### Observed Behavior
An environment file containing `VAR="some\nvalue"` sets the environment variable `VAR` to:
```
"some\nvalue"
```

rushim1:
Shall I replace the split logic with compose fork and raise a pr? @ubhattacharjya

Realmonia:
Please feel free to do so if you have bandwidth. We will help with the PR. @rushim1

Realmonia:
Hi, after discussing @rushim1 's PR with team, following docker compose's spec conflicts with some of the current AWS documentation, and the behavior you mentioned does match the AWS doc. Considering it is a potential breaking change, we choose to not do it as of now. If you see an important use case, please raise this in https://github.com/aws/containers-roadmap to get help from our product team. Thanks!

ltm:
@Realmonia The documentation I linked to in the description still claims that ECS uses the Docker Compose syntax (via the link in the following paragraph):

> Each line in an environment file must contain an environment variable in VARIABLE=VALUE format. Spaces or quotation marks are included as part of the values. Lines beginning with # are treated as comments and are ignored. **For more information about the environment variable file syntax, see [Declare default environment variables in file](https://docs.docker.com/compose/env-file/).**

[...]

PR #3763, opened 2023-06-24
feat: conforming envfile parsing to docker compose envfile

rushib1:
### Summary

This pull request aims to support escape characters and support docker-compose [spec](https://docs.docker.com/compose/environment-variables/env-file/) (Doesn't support Parameter expansion)

### Implementation details

I have imported the files from docker compose [reporsitory](https://github.com/compose-spec/compose-go/tree/master/dotenv) and used those files to parse the envfiles

### Testing

New tests cover the changes: 
yes

Feature: Envfile conforming to docker-compose envfile (Not including parameter expansion)

### Licensing


rushib1:
This pr is made due to this issue https://github.com/aws/amazon-ecs-agent/issues/3429

Realmonia:
Thanks for the contribution!

rushib1:
1. The docker-compose has the official implementation for their spec, and is not seperated as different pacakge. Since I did not want to import the whole docker-compose package just for the functionality of dotenv, I have imported only the necessary files.    
2. Since it is the official implementation I assumed it to be the only option for the job and haven't explored other functionality.

Let me know if its okay to import the total docker-compose package just for the functionality of dotenv implementation.

Realmonia:
Yep please import the whole package so we can have dependabot to help maintain it.

rushib1:
Hi @Realmonia, whenever I try `go get github.com/compose-spec/compose-go` it keeps updating other dependencies like docker to v24, go-units to v0.5, logrus to v1.9.0 and testify to v1.8.4 any idea what I should do now?

Realmonia:
Hmm interesting. It does not make sense to me as those packages are not part of the dependency in go.mod of the compose-go package. Can you try if handle override of these dependencies to previous version would work? If not, maybe you can try this godotenv dependency https://github.com/joho/godotenv/blob/main/go.mod

rushib1:
Even without the dependency If I try to run the commands `go get` or `go mod download` inside agent folder the dependecies are getting updated.
```
-       github.com/docker/docker v20.10.24+incompatible
+       github.com/docker/docker v24.0.2+incompatible
        github.com/docker/go-connections v0.4.0
-       github.com/docker/go-units v0.4.0
+       github.com/docker/go-units v0.5.0
```

Realmonia:
Oh yeah I forgot that. That makes sense.

rushib1:
Just for your information, these new changes will break the existing spec of task-def
`Each line in an environment file must contain an environment variable in VARIABLE=VALUE format. Spaces or quotation marks are included as part of the values` - [source](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/taskdef-envfiles.html)

Example case where this will fail:
```
key1="value1
key2=value"
```

Realmonia:
Hey sorry after further discussion with the team, we decide to not make this change for now. As you pointed out this conflicts with the current documentation; we will need our product team to be involved in, and have proper customer wording before we actually make such change. However, it's possible that we will do this in the future, and we will reopen & merge PR then.

Thanks again for the contribution and sorry again for the confusion!

PR #3765, opened 2023-06-26
Refactor ACS attach instance ENI message handling

danehlim:
### Testing

New tests cover the changes: 

### Licensing


ecs-agent/acs/session/attach_instance_eni_responder.go
```diff
@@ -0,0 +1,154 @@
...
+				logger.Error(fmt.Sprintf("Unable to handle %s", AttachInstanceENIMessageName), logger.Fields{
+					field.MessageID: messageID,
+					field.Error:     err,
+				})
+			}
+		}(mENI)
+	}
+
+	// Send ACK.
+	go func() {
```
Realmonia, line 99: I see we do ack first then handle messages in previous implementation; I foresee no issue with current implementation, since message handling is done in go routine, but is there any reason we changed to this way?

ecs-agent/acs/session/attach_instance_eni_responder.go
```diff
@@ -0,0 +1,154 @@
...
+			field.Error: err,
+		})
+		return
+	}
+
+	// Handle ENIs in the message.
+	messageID := aws.StringValue(message.MessageId)
+	for _, mENI := range message.ElasticNetworkInterfaces {
+		expiresAt := receivedAt.Add(time.Duration(aws.Int64Value(message.WaitTimeoutMs)) * time.Millisecond)
+		go func(eni *ecsacs.ElasticNetworkInterface) {
```
Realmonia, line 75: nit: any reason we need to create this inline anonymous function instead of having a separate function to do handling? We can then test that function as needed.

Realmonia (commented):
Is `HandleENIAttachment` Implemented differently by ec2 and fargate? How different are they? Can it be put into ecs-agent module too?

danehlim:
@Realmonia:
> Is HandleENIAttachment Implemented differently by ec2 and fargate? How different are they? Can it be put into ecs-agent module too?

`HandleENIAttachment` is implemented significantly differently between EC2 and Fargate. That is, each launch type makes use of significantly different implementations of the `ENIHandler` interface. Thus, each implementation should continue to live in their respective code base instead of in ecs-agent module.

ecs-agent/api/eni/eni.go
```diff
@@ -19,11 +19,12 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/aws/aws-sdk-go/aws"
```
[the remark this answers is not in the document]
danehlim: My IDE is configured to automatically group all imports from `amazon-ecs-agent` repo into one group, all stdlib imports into another group, and all other imports in another group. If there are other blocking (non-nit) changes that need to be made, I can revert to the previous ordering if you prefer.

ecs-agent/acs/session/attach_instance_eni_responder.go
danehlim: The main reason is to maintain consistency with other ACS message responders (i.e., right now, some responders we ACK first and others we ACK at the end with no clear reason as to why - we should look to standardize).

ecs-agent/acs/session/attach_instance_eni_responder.go
danehlim: That's a good point - I don't see any reason in particular that we can't have a separate function to do the handling. I'll make an update to the PR to do this instead.

Realmonia:
Thanks!

ecs-agent/acs/session/attach_instance_eni_responder.go
Realmonia: Sounds good.

ecs-agent/acs/session/attach_instance_eni_responder.go
Realmonia: Sounds good.

yinyic, line 75: In the PR notes we mentioned
> Handle ENI attachment in async (no reason for this to be a blocking operation)

Curious how asynchronous processing adds value here? It looks like `HandleENIAttachment` function currently does little work, returns quickly, and will never get stuck. Since the handling logic is straightforward, I wonder if we want to keep the implementation simple.

We should also remember that the handler already spins off new goroutine to handle each incoming message.

ecs-agent/api/eni/eni.go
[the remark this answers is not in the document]
danehlim: Discussed offline, aligned on preferred grouping of imports as follows:
- All stdlib imports come first in its own group
- All other imports go in its own group afterwards in alphabetical order

I will update the PR such that this file follows the above grouping convention.

agent/acs/handler/attach_eni_handler_common.go
```diff
@@ -19,17 +19,14 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
 
+	"github.com/cihub/seelog"
```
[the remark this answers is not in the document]
danehlim: Discussed offline, aligned on preferred grouping of imports as follows:
- All stdlib imports come first in its own group
- All other imports go in its own group afterwards in alphabetical order

I will update the PR such that this file follows the above grouping convention.

ecs-agent/acs/session/attach_instance_eni_responder.go
danehlim: Thanks for your comment!

> Curious how asynchronous processing adds value here?

I figured since each ENI's attachment is independent from each other and independent from ACKing receipt of valid message from ACS, it would be a best practice not to block here before allowing other such independent tasks to start. This also follows the convention of other ACS responders such as the heartbeat responder (example [here](https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/acs/session/heartbeat_responder.go#L50-L71)).

> It looks like `HandleENIAttachment` function currently does little work, returns quickly, and will never get stuck. Since the handling logic is straightforward, I wonder if we want to keep the implementation simple.

That would actually depend on the implementation of the `HandleENIAttachment` function by the `ENIHandler` we use. The current implementation used by the Docker-based Agent defined in `agent/` directory can be found [here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/attach_eni_handler_common.go#L42-L66). In my opinion, when I do the full code trace of this function, I'm not sure that I would consider the implementation here to do "little work". We should also consider other use cases such as:
1. What if an `ENIHandler` that has a different implementation which does significant work, returns slowly, and/or has potential to get stuck is used?
2. What if the `ENIHandler` that we currently use is changed in the future such that it does significant work, returns slowly, and/or has potential to get stuck?

> We should also remember that the handler already spins off new goroutine to handle each incoming message.

Does the responder spin off a new goroutine to handle each incoming message? Could you point to in where in code you are seeing this? My current understanding under this PR is that is not the case, but I could be missing something.

yinyic: Thanks for the detailed explanation! 

You are absolutely right that we should be future-proofing and that the ENI handling may grow more complex in the future. 

Regarding the top-level handler, apologies that it was a misreading on my part. I was referring to https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/wsclient/client.go#L447. Reading again, the code use the same goroutine for consuming all messages, so there doesn't look like to be built-in per-message goroutine.

yinyic (approved):

PR #3766, opened 2023-06-26
EBS volume metrics collector daemon

xxx0624:
### Summary

1. Add source code of the GRPC server hosted in the new daemon, EBS volume metrics collector, to the common library.
2. Update the common lib vendor.
3. Update agent vendor.

### Implementation details

This daemon will host an unix socket based GRPC server which will implement one interface https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetvolumestats and can provide the capability for ECS Agent to get the volume stats for one given mounted EBS volume.

In this PR, it includes the GRPC server related logics **only**. And the implementation is same as the volume stats implementation of the existing AWS ebs csi driver implementation (https://github.com/kubernetes-sigs/aws-ebs-csi-driver/) except the following items - 
- the current implementation supports file based logging however the AWS ebs csi driver doesn't.
  - The [klog](https://github.com/kubernetes/klog/tree/main) is the log lib used by both current implementation and the AWS ebs csi driver. But AWS ebs csi driver forces all logs to go to the stdout/stderr.
- the supported flags are different.
  - Flags supported by current implementation: [flags from klog](https://github.com/kubernetes/klog/blob/main/klog.go#L401) + [one custom flag](https://github.com/aws/amazon-ecs-agent/pull/3766/files#diff-ae7f7ee4afd1f8168f552053131f463b51056dcaf6f64fc5afe5fca2a510a19eR31)
  - Flags supported by the AWS ebs csi driver: 
```
Usage of aws-ebs-csi-driver:
      --aws-sdk-debug-log                    To enable the aws sdk debug log level (default to false).
      --endpoint string                      Endpoint for the CSI driver server (default "unix://tmp/csi.sock")
      --extra-tags mapStringString           Extra tags to attach to each dynamically provisioned resource. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'
      --extra-volume-tags mapStringString    DEPRECATED: Please use --extra-tags instead. Extra volume tags to attach to each dynamically provisioned volume. It is a comma separated list of key value pairs like '<key1>=<value1>,<key2>=<value2>'
      --http-endpoint :8080                  The TCP network address where the HTTP server for metrics will listen (example: :8080). The default is empty string, which means the server is disabled.
      --k8s-tag-cluster-id string            ID of the Kubernetes cluster used for tagging provisioned EBS volumes (optional).
      --log-flush-frequency duration         Maximum number of seconds between log flushes (default 5s)
      --log-json-info-buffer-size quantity   [Alpha] In JSON format with split output streams, the info messages can be buffered for a while to increase performance. The default value of zero bytes disables buffering. The size can be specified as number of bytes (512), multiples of 1000 (1K), multiples of 1024 (2Ki), or powers of those (3M, 4G, 5Mi, 6Gi). Enable the LoggingAlphaOptions feature gate to use this.
      --log-json-split-stream                [Alpha] In JSON format, write error messages to stderr and info messages to stdout. The default is to write a single stream to stdout. Enable the LoggingAlphaOptions feature gate to use this.
      --logging-format string                Sets the log format. Permitted formats: "json" (gated by LoggingBetaOptions), "text". (default "text")
      --logtostderr                          log to standard error instead of files. DEPRECATED: will be removed in a future release.
      --user-agent-extra string              Extra string appended to user agent.
  -v, --v Level                              number for the log level verbosity
      --version                              Print the version and exit.
      --vmodule pattern=N,...                comma-separated list of pattern=N settings for file-filtered logging (only works for text log format)
      --volume-attach-limit int              Value for the maximum number of volumes attachable per node. If specified, the limit applies to all nodes. If not specified, the value is approximated from the instance type. (default -1)
      --warn-on-invalid-tag                  To warn on invalid tags, instead of returning an error
```

#### how go.mod and go.sum are updated
- Run `go mod tidy` and `go mod vendor` in both `agent` dir and `ecs-agent` dir already.
- Verified that all new direct dependencies added to `go.mod` are same as the `go.mod` of agent.

#### how to start up the process (won't work until one following PR is out)
```
# This command below will force all logs (whose log level is not higher than 5) to the specified location
# Please refer this log conventions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use

"/bin/csi-driver" "--endpoint=unix://tmp/csi.sock" "--log_file=/tmp/server.log" "-v=5" "--logtostderr=false" "--stderrthreshold=3"
```

### Testing

Also 
- manually validated all tests which are included in the new directory are pass via `make test` under the new directory.

Add the EBS volume metrics collector.

### Licensing


ecs-agent/go.mod
```diff
@@ -5,37 +5,64 @@ go 1.19
...
+	github.com/kubernetes-csi/csi-proxy/client v1.1.2
 	github.com/pborman/uuid v1.2.1
 	github.com/pkg/errors v0.9.1
-	github.com/prometheus/client_model v0.2.0
-	github.com/prometheus/common v0.26.0
-	github.com/stretchr/testify v1.7.0
-	golang.org/x/net v0.8.0
-	golang.org/x/sys v0.6.0
-	golang.org/x/tools v0.6.0
+	github.com/prometheus/client_model v0.3.0
```
fierlion, line 17: for all these new imports we'll need to validate that there's no collision between the agent go.mod and this one

ecs-agent/daemon_images/csi-driver/tarfiles/csi-driver-linux-amd64.tar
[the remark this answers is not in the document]
xxx0624: Removed.

mye956 (commented):
`the supported flags are different`

nit: Could we include what flags were added/removed/modified somewhere within the PR description?

ecs-agent/daemon_images/csi-driver/Dockerfile.csi_driver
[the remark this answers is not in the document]
xxx0624: Synced-up offline. Will have a separate PR for building the tar.

ecs-agent/go.mod
xxx0624: Sry, I accidentally rebase my commits.
I've updated the go.mod and make sure there is no collision for the direct dependencies.

mye956 (commented):
`the current implementation support file based logging via klog however the AWS ebs csi driver doesn't. The klog is the log lib used by the AWS ebs csi driver.`

nit(non-blocking): this is a bit confusing to read at first, perhaps this could be reworded better to clarify whether or not klog is used by the AWS ebs csi driver?

xxx0624:
> `the supported flags are different`
> 
> nit: Could we include what flags were added/removed/modified somewhere within the PR description?

Done.

> `the current implementation support file based logging via klog however the AWS ebs csi driver doesn't. The klog is the log > lib used by the AWS ebs csi driver.`
>
> nit(non-blocking): this is a bit confusing to read at first, perhaps this could be reworded better to clarify whether or not klog is used by the AWS ebs csi driver?

Sure, updated.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
```diff
@@ -0,0 +1,51 @@
...
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package fs
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+type UsageInfo struct {
```
mye956, line 23: nit: since this struct is used by all three fs*.go files, perhaps this could be defined in a common struct file and have all three files import the struct

mye956: unless this was taken from the existing EBS CSI driver then please make this as non-blocking

xxx0624: The other two fs files are for different platforms and this `UsageInfo` in this fs file doesn't apply to all platforms.

xxx0624: > unless this was taken from the existing EBS CSI driver then please make this as non-blocking

Yes, existing ebs csi is doing this same and definitely this won't be a blocker.
But I think your suggestion makes sense. I can try to make the change later.

ecs-agent/daemonimages/csidriver/options.go
```diff
@@ -0,0 +1,43 @@
...
+
+package main
+
+import (
+	"errors"
+	"flag"
+	"os"
+)
+
+const EmptyCSIEndpoint = ""
```
amogh09, line 22: Do we need to export this constant?

ecs-agent/daemonimages/csidriver/options_test.go
```diff
@@ -0,0 +1,66 @@
...
+				opts, err := testFunc(t, []string{"--endpoint=foo"})
+				assert.NoError(t, err)
+				assert.Equal(t, "foo", opts.Endpoint)
+			},
+		},
+		{
+			name: "No argument is given",
+			testFunc: func(t *testing.T) {
+				_, err := testFunc(t, nil)
+				assert.NotNil(t, err)
```
amogh09, line 58: Can we please assert that the right error is returned? `assert.NotNil` is too broad IMO.

ecs-agent/daemonimages/csidriver/options_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+func TestGetServerOptions(t *testing.T) {
+	testFunc := func(t *testing.T, additionalArgs []string) (*ServerOptions, error) {
+		flagSet := flag.NewFlagSet("test-flagset", flag.ContinueOnError)
+
+		args := append([]string{
+			"/bin/csi-driver",
+		}, additionalArgs...)
+		oldArgs := os.Args
+		defer func() { os.Args = oldArgs }()
```
amogh09, line 35: Is there a way to test flag parsing without mutating `os.Args`? The mutation seems unnecessary and brittle.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
```diff
@@ -0,0 +1,51 @@
...
+	Bytes  int64
+	Inodes int64
+}
+
+// Info linux returns (available bytes, byte capacity, byte usage, total inodes, inodes free, inode usage, error)
+// for the filesystem that path resides upon.
+func Info(path string) (int64, int64, int64, int64, int64, int64, error) {
```
amogh09, line 30: Who will be consuming this function? Is it possible to return the information in a struct? That's way too many unnamed `int64`s.

amogh09: How are we testing this function?

ecs-agent/daemonimages/csidriver/util/utils.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+func ParseEndpoint(endpoint string) (string, string, error) {
```
amogh09, line 24: Do we have unit tests for this function? If not, I think we should.

ecs-agent/daemonimages/csidriver/Makefile
```diff
@@ -0,0 +1,26 @@
+CSI_DRIVER_VERSION?=v1.0.0
+GO111MODULE=on
+GOPATH=$(shell go env GOPATH)
+BUILD_DATE=$(shell date -u -Iseconds)
```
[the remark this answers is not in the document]
xxx0624: Yea it's UTC timezone.

ecs-agent/daemonimages/csidriver/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+SOURCEDIR=./
+CSI_DRIVER_SOURCES=$(shell find $(SOURCEDIR) -name '*.go')
+
+OS?=linux
+ARCH?=amd64
+
+# Build binary for linux only
+.PHONY: bin/csi-driver
+bin/csi-driver: $(CSI_DRIVER_SOURCES)
```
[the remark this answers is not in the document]
xxx0624: Ah make sense. Let me remove this.

ecs-agent/daemonimages/csidriver/Makefile
```diff
@@ -0,0 +1,26 @@
...
+GOPATH=$(shell go env GOPATH)
+BUILD_DATE=$(shell date -u -Iseconds)
+
+SOURCEDIR=./
+CSI_DRIVER_SOURCES=$(shell find $(SOURCEDIR) -name '*.go')
+
+OS?=linux
+ARCH?=amd64
+
+# Build binary for linux only
```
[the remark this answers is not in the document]
xxx0624: Good catch. Initially it's hardcoded. Let me update this.

ecs-agent/daemonimages/csidriver/driver/driver.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"google.golang.org/grpc"
+	"k8s.io/klog/v2"
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/util"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/version"
+)
+
+type Driver struct {
```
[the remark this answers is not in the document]
xxx0624: Sure

ecs-agent/daemonimages/csidriver/driver/driver.go
```diff
@@ -0,0 +1,100 @@
...
+	endpoint string
+}
+
+func NewDriver(options ...func(*DriverOptions)) (*Driver, error) {
+	driverInfo := version.GetVersionInfo()
+	klog.InfoS("Driver Information",
+		"Driver", "csi-driver",
+		"Version", driverInfo.Version,
+	)
+	klog.V(4).InfoS("Additional driver information",
```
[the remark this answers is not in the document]
xxx0624: > Please refer this log conventions https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md#what-method-to-use

Also put this in the description.

ecs-agent/daemonimages/csidriver/driver/node.go
```diff
@@ -0,0 +1,33 @@
...
+package driver
+
+import (
+	"github.com/container-storage-interface/spec/lib/go/csi"
+)
+
+// nodeService represents the node service of CSI driver
+type nodeService struct {
+	mounter Mounter
+	csi.UnimplementedNodeServer
```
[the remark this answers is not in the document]
xxx0624: I will add a comment here to explain with more details.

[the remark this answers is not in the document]
xxx0624: Importing this as one field can help us avoid to re-implement unnecessary interfaces.

ecs-agent/daemonimages/csidriver/driver/mount.go
```diff
@@ -0,0 +1,31 @@
...
+type Mounter interface {
+	PathExists(path string) (bool, error)
+}
+
+// NodeMounter implements Mounter.
+type NodeMounter struct {
+}
+
+func (nm NodeMounter) PathExists(path string) (bool, error) {
+	return false, nil
```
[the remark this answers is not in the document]
xxx0624: Sure

ecs-agent/daemonimages/csidriver/options_test.go
xxx0624: Sounds good. Will add one.

ecs-agent/daemonimages/csidriver/options_test.go
xxx0624: Hmmm not sure.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
xxx0624: This isn't used for this pR but will be used in a follow up PR.

For the testing, this will invoke a syscall. No idea for now about how to test this simple file.

For the naming, let me use named return parameter to give callers a better understanding for each var.

ecs-agent/daemonimages/csidriver/Makefile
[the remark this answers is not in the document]
amogh09: I meant if it is okay for the time to be included as that was not clear to me since the variable name only contains "DATE". I guess it's okay then.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
```diff
@@ -27,25 +27,33 @@ type UsageInfo struct {
...
 	if err != nil {
-		return 0, 0, 0, 0, 0, 0, err
+		available = 0
+		capacity = 0
+		used = 0
+
+		totalInodes = 0
+		freeInodes = 0
+		usedInodes = 0
+		return
```
amogh09, line 45: Why do we want to silently swallow the error instead of raising it?

ecs-agent/daemonimages/csidriver/util/fs/fs.go
amogh09: Okay. Have we tested this function manually for both Linux and Windows?

ecs-agent/daemonimages/csidriver/options_test.go
amogh09: We can do that by passing an `args []string` parameter to `GetServerOptions` instead of having it read from `os.Args`. I leave the decision to you as this is a nit.

ecs-agent/daemonimages/csidriver/driver/mount.go
```diff
@@ -0,0 +1,31 @@
...
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package driver
+
+// Mounter is the interface implemented by NodeMounter.
```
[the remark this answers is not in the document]
amogh09: Did we remove the description of the interface? I think there should be a description but not in terms of the implementation.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
xxx0624: It won't be silent. We are using named return parameter, so we don't have to list all var in the return statement.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
xxx0624: Just for linux. Window team will work on this later I think.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
xxx0624: meaning `err` won't be nil in this case

ecs-agent/daemonimages/csidriver/options_test.go
xxx0624: Cool, I can do this in the another PR.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
amogh09: I see. TIL.

ecs-agent/daemonimages/csidriver/util/fs/fs.go
amogh09: I see. Let's remove this code from this PR and add it back when it's tested. The code is non-trivial so we should test it when checking it in.

amogh09 (dismissed):

amogh09 (approved):

mye956 (approved):

PR #3776, opened 2023-06-29
Use total_inactive_file for calculating memory usage

wt-l00:
### Summary

Docker subtracts total_inactive_file when calculating memory usage for cgroup v1 from 20.10.0 release, so do the same for getMemUsage function. 

for reference https://github.com/docker/cli/pull/2415 .

### Testing

Bugfix: Fix cgroup v1 mem usage calculation to match docker cli

### Licensing

### DIFF
--- a/agent/stats/utils_unix.go
+++ b/agent/stats/utils_unix.go
@@ -49,7 +49,7 @@ func getMemUsage(mem types.MemoryStats) uint64 {
 			return mem.Usage - v
 		}
 	}
-	if v, ok := mem.Stats["cache"]; ok && v < mem.Usage {
+	if v, ok := mem.Stats["total_inactive_file"]; ok && v < mem.Usage {
 		return mem.Usage - v
 	}
 	return mem.Usage
--- a/agent/stats/utils_test.go
+++ b/agent/stats/utils_test.go
@@ -51,7 +51,7 @@ func TestDockerStatsToContainerStatsMemUsage(t *testing.T) {
 				"usage": %d,
 				"max_usage": %d,
 				"stats": {
-					"cache": %d,
+					"total_inactive_file": %d,
 					"rss": %d
 				},
 				"privateworkingset": %d


agent/stats/utils_unix.go
```diff
@@ -49,7 +49,7 @@ func getMemUsage(mem types.MemoryStats) uint64 {
 			return mem.Usage - v
 		}
 	}
-	if v, ok := mem.Stats["cache"]; ok && v < mem.Usage {
```
Realmonia, line 52: There is some failed unit test e.g. `TestDockerStatsToContainerStatsMemUsage`  failed due to this change. Please update accordingly.

Realmonia (changes requested):

Realmonia (approved):

agent/stats/utils_unix.go
wt-l00: Thanks. I fixed it.

yinyic (approved):

wt-l00:
`ecs/windows2016/functional_test` has failed. I can't see the details of this test, is there anything I can do?

Realmonia:
I will restart the test workflow. We have a release going on so I will merge the change after the branch is clear.

Realmonia:
Hi @wt-l00 , after discussing with the team, we decide to revert this change. The reason being there are currently established automation (especially around scaling) from other customers that could be impacted by this change. We plan to make this a opted-in feature to prevent a default behavior change, however we do not have an ETA for now.

justin007755:
> Hi @wt-l00 , after discussing with the team, we decide to revert this change. The reason being there are currently established automation (especially around scaling) from other customers that could be impacted by this change. We plan to make this a opted-in feature to prevent a default behavior change, however we do not have an ETA for now.

Hi @Realmonia Regarding this opted-in feature do we have an ETA now? We have a cx to use aws ecs service, and they noticed the "Service memory utilization" from ecs itself is always quite low but this metric keeps high if they used docker stats to calculate directly, so this will cause some problems because they used this metric to monitor memory usage for alarming, which mean the metric "Service memory utilization" from ecs itself could not reflect actual situation of memory utilization…… They really want this opted-in feature to be implemented, so would you help confirm the ETA for this feature?

PR #3780, opened 2023-06-30
broadcast FSx capability by default for Windows

rawahars:
### Summary

Presently, the logic for enabling FSx on the instance requires the instance to be domain joined with the AD. This behaviour was fixed in https://github.com/aws/amazon-ecs-agent/pull/3540.

However, this is not needed for enabling just FSx capability. If a customer has setup their DHCP option set for their AD, then FSx can work on their instances even without a domain join. Reference- https://docs.aws.amazon.com/directoryservice/latest/admin-guide/dhcp_options_set.html

Therefore, we will enable FSx capability by default. The customer can explicitly disable it by setting `ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED` as false.

### Implementation details

Removed the dependency of FSx capability on instance domain join.

### Testing

Created a custom agent and ran a FSx enabled task without domain join of the instance.
New tests cover the changes: 

enable FSx capability by default for Windows
### Licensing

### DIFF
--- a/agent/config/config_unix.go
+++ b/agent/config/config_unix.go
@@ -103,7 +103,7 @@ func DefaultConfig() Config {
 		CgroupCPUPeriod:                     defaultCgroupCPUPeriod,
 		GMSACapable:                         parseGMSACapability(),
 		GMSADomainlessCapable:               parseGMSADomainlessCapability(),
-		FSxWindowsFileServerCapable:         BooleanDefaultFalse{Value: ExplicitlyDisabled},
+		FSxWindowsFileServerCapable:         BooleanDefaultTrue{Value: ExplicitlyDisabled},
 		RuntimeStatsLogFile:                 defaultRuntimeStatsLogFile,
 		EnableRuntimeStats:                  BooleanDefaultFalse{Value: NotSet},
 		ShouldExcludeIPv6PortBinding:        BooleanDefaultTrue{Value: ExplicitlyEnabled},
--- a/agent/config/parse_linux.go
+++ b/agent/config/parse_linux.go
@@ -69,8 +69,8 @@ func parseGMSACapability() BooleanDefaultFalse {
 	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
 }
 
-func parseFSxWindowsFileServerCapability() BooleanDefaultFalse {
-	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
+func parseFSxWindowsFileServerCapability() BooleanDefaultTrue {
+	return BooleanDefaultTrue{Value: ExplicitlyDisabled}
 }
 
 // parseGMSADomainlessCapability is used to determine if gMSA domainless support can be enabled
--- a/agent/config/parse_unsupported.go
+++ b/agent/config/parse_unsupported.go
@@ -25,8 +25,8 @@ func parseGMSACapability() BooleanDefaultFalse {
 	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
 }
 
-func parseFSxWindowsFileServerCapability() BooleanDefaultFalse {
-	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
+func parseFSxWindowsFileServerCapability() BooleanDefaultTrue {
+	return BooleanDefaultTrue{Value: ExplicitlyDisabled}
 }
 
 func parseGMSADomainlessCapability() BooleanDefaultFalse {
--- a/agent/config/parse_windows.go
+++ b/agent/config/parse_windows.go
@@ -68,15 +68,17 @@ func parseGMSADomainlessCapability() BooleanDefaultFalse {
 }
 
 // parseFSxWindowsFileServerCapability is used to determine if fsxWindowsFileServer support can be enabled
-func parseFSxWindowsFileServerCapability() BooleanDefaultFalse {
-	// fsxwindowsfileserver is not supported on Windows 2016 and non-domain-joined container instances
+func parseFSxWindowsFileServerCapability() BooleanDefaultTrue {
+	// fsxwindowsfileserver is not supported on Windows 2016.
 	status, err := IsWindows2016()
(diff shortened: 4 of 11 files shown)


arun-annamalai (approved):

jterry75 (approved):

agent/config/parse_windows.go
```diff
@@ -75,8 +75,15 @@ func parseFSxWindowsFileServerCapability() BooleanDefaultFalse {
 		return BooleanDefaultFalse{Value: ExplicitlyDisabled}
 	}
 
+	// By default or if ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED is set as true, agent will
+	// broadcast the FSx capability.
 	envStatus := utils.ParseBool(os.Getenv("ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED"), true)
-	return checkDomainJoinWithEnvOverride(envStatus)
+	if envStatus {
+		return BooleanDefaultFalse{Value: ExplicitlyEnabled}
```
singholt, line 82: can we make use of [parseBooleanDefaultTrueConfig()](https://github.com/aws/amazon-ecs-agent/blob/6733da5a6e212a62e16021619f5fcbd3ad879c0a/agent/config/parse.go#L254) to evaluate this? It adds the value based on whether customer explicitly set it or not.

Notice parseBooleanDefaultTrueConfig() and not parseBooleanDefaultFalseConfig() since the default is true.

singholt: also, in the DefaultConfig, we should change [this](https://github.com/aws/amazon-ecs-agent/blob/2cdd37099090dd976041eb7e9596f157cd6e5d67/agent/config/config_windows.go#L148) to

`FSxWindowsFileServerCapable: BooleanDefaultTrue{Value: ExplicitlyDisabled}`, implying that the default is true unless explicitly disabled?

rawahars: Thanks Anuj!
Addressed all the feedback in the latest revision.

agent/config/config_unix.go
```diff
@@ -103,7 +103,7 @@ func DefaultConfig() Config {
 		CgroupCPUPeriod:                     defaultCgroupCPUPeriod,
 		GMSACapable:                         parseGMSACapability(),
 		GMSADomainlessCapable:               parseGMSADomainlessCapability(),
-		FSxWindowsFileServerCapable:         BooleanDefaultFalse{Value: ExplicitlyDisabled},
+		FSxWindowsFileServerCapable:         BooleanDefaultTrue{Value: ExplicitlyDisabled},
```
singholt, line 106: since this feature is not supported on Linux, we would want to keep this as is right? 

`BooleanDefaultFalse{Value: ExplicitlyDisabled}`

agent/config/parse_linux.go
```diff
@@ -69,8 +69,8 @@ func parseGMSACapability() BooleanDefaultFalse {
 	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
 }
 
-func parseFSxWindowsFileServerCapability() BooleanDefaultFalse {
-	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
+func parseFSxWindowsFileServerCapability() BooleanDefaultTrue {
+	return BooleanDefaultTrue{Value: ExplicitlyDisabled}
```
singholt, line 73: same comment as for config_unix, I would just leave these unchanged since we want the default to be false for Linux.

agent/config/parse_unsupported.go
```diff
@@ -25,8 +25,8 @@ func parseGMSACapability() BooleanDefaultFalse {
 	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
 }
 
-func parseFSxWindowsFileServerCapability() BooleanDefaultFalse {
-	return BooleanDefaultFalse{Value: ExplicitlyDisabled}
+func parseFSxWindowsFileServerCapability() BooleanDefaultTrue {
+	return BooleanDefaultTrue{Value: ExplicitlyDisabled}
```
singholt, line 29: same comment here as well, about leaving this unchanged

arun-annamalai (approved):

agent/config/parse_linux.go
singholt: synced offline, this is because `FSxWindowsFileServerCapable` is a shared config type between both Windows, Linux platforms.

We've added tests that validate that this config type is true for Windows and false for everything else. Resolving related threads below too.

agent/config/config_unix.go
rawahars: There is a universal contract of the type in `types.go`. We cannot change one without forking them out. I don't think it is a big enough case to fork this out. Therefore, keeping it as it.

agent/config/parse_linux.go
rawahars: Same as above.

agent/config/config_unix.go
singholt: synced offline, this is because `FSxWindowsFileServerCapable` is a shared config type between both Windows, Linux platforms.

We've added tests that validate that this config type is true for Windows and false for everything else. Resolving related threads below too.

singholt (approved):

singholt (approved):

agent/config/types.go
```diff
@@ -340,8 +340,8 @@ type Config struct {
 	VolumePluginCapabilities []string
 
 	// FSxWindowsFileServerCapable is the config option to indicate if fsxWindowsFileServer is supported.
-	// It should be enabled by default only if the container instance is part of a valid active directory domain.
-	FSxWindowsFileServerCapable BooleanDefaultFalse
+	// It is enabled by default and can be overridden by the ECS_FSX_WINDOWS_FILE_SERVER_SUPPORTED environment variable.
```
singholt, line 343: super-nit: enabled by default on Windows

singholt (approved):

Realmonia (approved):

PR #3782, opened 2023-07-05
fix memory resource accounting for multiple containers in single task

prateekchaudhry:
### Summary
This PR fixes how task memory is accounted when task level memory limit is not specified in task definition. The existing implementation fails to account correctly in cases where containers with `MemoryReservation` field present in HostConfig might occur before containers without such field in the `task.Containers` slice. In such cases, `MemoryReservation` of previous container 'persists', as the object into which the HostConfig is being unmarshalled is declared outside of the loop which iterates over the containers.

Main cause of this change is that the ACS payload agent receives has HostConfig like `{\"NetworkMode\":\"bridge\",\"CapAdd\":[],\"CapDrop\":[]}` with missing fields in [HostConfig](https://pkg.go.dev/github.com/docker/docker@v24.0.4+incompatible/api/types/container#HostConfig) struct. That is, if fields like `MemoryReservation` are not specified, they are not present in the payload, and `json.Unmarshal`does not make any changes to the object. Using a new object for every container ensures this persistence does not happen.

### Implementation details
Fix: 
* Define new `&dockercontainer.HostConfig{}` object for each container when mapping task to host resources
* `rawHostConfig*` in related unit tests have been modified to be directly defined using a string instead of marshaling a HostConfig object. This is necessary, because if we use a HostConfig object, it will use the zero value (`int64(0)`) of fields like `MemoryReservation` if they are not specified

### Testing
Verified -
* With manual tests this behavior is consistent
* With manual tests resources allocated in ECS back end are same as ECS Agent for each task
* Added new unit tests with multiple containers and a test with awsvpc network mode
* Verified test with multiple containers - with containers with `MemoryReservation` followed by containers with `memory` fail for existing implementation - and pass with the fix -`testTask5` here

New tests cover the changes: 
Yes

Fix memory resource accounting for multiple containers in single task
### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -3557,10 +3557,11 @@ func (task *Task) ToHostResources() map[string]*ecs.Resource {
 		}
 	} else {
 		containerMEMint64 := int64(0)
-		// To parse memory reservation / soft limit
-		hostConfig := &dockercontainer.HostConfig{}
 
 		for _, c := range task.Containers {
+			// To parse memory reservation / soft limit
+			hostConfig := &dockercontainer.HostConfig{}
+
 			if c.DockerConfig.HostConfig != nil {
 				err := json.Unmarshal([]byte(*c.DockerConfig.HostConfig), hostConfig)
 				if err != nil || hostConfig.MemoryReservation <= 0 {
--- a/agent/api/task/task_test.go
+++ b/agent/api/task/task_test.go
@@ -4894,9 +4894,13 @@ func TestToHostResources(t *testing.T) {
 	// Prepare simple hostConfigs with and without memory reservation field for test cases
 
 	// Host Config with Memory Reservation 400 MiB
-	rawHostConfigMemReservation := "{\"Ulimits\":[],\"MemoryReservation\":419430400,\"NetworkMode\":\"bridge\",\"CapAdd\":[],\"CapDrop\":[]}"
+	rawHostConfigMemReservation400MiB := "{\"Ulimits\":[],\"MemoryReservation\":419430400,\"NetworkMode\":\"bridge\",\"CapAdd\":[],\"CapDrop\":[]}"
+	// Host Config with Memory Reservation 600 MiB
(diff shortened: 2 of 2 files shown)


agent/api/task/task_test.go
```diff
@@ -4983,6 +4983,35 @@ func TestToHostResources(t *testing.T) {
 		},
 	}
 
+	// A combination of containers with different configs with memory sourcing from different sources
+	testTask5 := &Task{
```
yinyic, line 4987: This test may not be able to catch the gap that we are fixing, since the test uses a single hostConfig for all containers, In that case, the existing implementation would've worked fine. We should create at least two different hostConfigs with different memoryReservation, such that the existing implementation would fail

prateekchaudhry: Ack - additionally, existing implementation fails due to payload Agent receives does not have all the HostConfig fields. So the unit test and it's hostConfig need to be defined in such a way as well to actually repro the case. Pushed changes

agent/api/task/task.go
```diff
@@ -3557,10 +3557,11 @@ func (task *Task) ToHostResources() map[string]*ecs.Resource {
 		}
 	} else {
 		containerMEMint64 := int64(0)
-		// To parse memory reservation / soft limit
-		hostConfig := &dockercontainer.HostConfig{}
 
 		for _, c := range task.Containers {
+			// To parse memory reservation / soft limit
+			hostConfig := &dockercontainer.HostConfig{}
+
```
Realmonia, line 3564: Can you update in description about how this is causing the problem? i.e. `json.Unmarshal` will not override previously assigned value if the field is not specified in the new json source.

Realmonia (approved):

prateekchaudhry:
* Rebased feature branch against dev
* Reverted task resource accounting revert changes in feature branch to put it back in
* Rebased `taskAccountingMemoryAccountingFix` in fork against feature branch

Realmonia (approved):

SreeeS (approved):
LGTM

PR #3790, opened 2023-07-07
Support for setting PIDs limit for ECS tasks

sparrc:
### Summary

Adds a new agent configuration variable: `ECS_TASK_PIDS_LIMIT`

This variable defaults to being unset, which results in the current behavior of no PID limit being set at the task level.

If the user sets this to a value 1 or more, then agent will set the [pids.max pid limit cgroup parameter](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#pid) to the provided value, for the task-level cgroup for all ECS tasks that are launched on this container instance. Note that this limit is applied for the whole task, not per-container.

addresses https://github.com/aws/containers-roadmap/issues/354 for the EC2 launch type.

### Testing

Functionality was verified manually and unit tests were added in this PR. Functional tests will be added before merging to dev branch.

Enhancement: ECS_TASK_PIDS_LIMIT for setting the task-level PIDs limit.

### Licensing


agent/api/task/task.go
```diff
@@ -363,9 +363,10 @@ func (task *Task) PostUnmarshalTask(cfg *config.Config,
 
 	task.adjustForPlatform(cfg)
 
-	// TODO, add rudimentary plugin support and call any plugins that want to
```
sparrc, line 366: this TODO was 8 years old....probably time to remove

agent/api/task/task_attachment_handler_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build unit
```
sparrc, line 1: unrelated to this change, but these tests were not running without this tag

agent/config/parse_test.go
```diff
@@ -0,0 +1,56 @@
...
+	assert.Equal(t, 0, parseTaskPidsLimit())
+	t.Setenv("ECS_TASK_PIDS_LIMIT", "")
+	assert.Equal(t, 0, parseTaskPidsLimit())
+}
+
+func TestParseTaskPidsLimit_Unset(t *testing.T) {
+	assert.Equal(t, 0, parseTaskPidsLimit())
+}
+
+func TestParseContainerInstanceTags(t *testing.T) {
```
sparrc, line 33: this test is unrelated to this change, but am going to add a bunch of unrelated unit tests for the parse.go function that didnt previously exist

agent/api/task/task_linux.go
```diff
@@ -148,6 +148,14 @@ func (task *Task) BuildLinuxResourceSpec(cGroupCPUPeriod time.Duration) (specs.L
 		linuxResourceSpec.Memory = &linuxMemorySpec
 	}
 
+	// Set task pids limit if set via ECS_TASK_PIDS_LIMIT env var
+	if taskPidsLimit > 0 {
+		pidsLimit := &specs.LinuxPids{
+			Limit: int64(taskPidsLimit),
+		}
+		linuxResourceSpec.Pids = pidsLimit
```
fierlion, line 156: `linuxResourceSpec.Pids = pidsLimit`
As a preview, will this use the `--pids-limit` via docker api?  Is this also something we can do via contianerd directly?

fierlion: also what limits should we set on this? is there an upper bound enforced anywhere?

fierlion (approved):

agent/config/parse.go
```diff
@@ -297,6 +297,28 @@ func parseTaskMetadataThrottles() (int, int) {
 	return steadyStateRate, burstRate
 }
 
+func parseTaskPidsLimit() int {
```
chienhanlin, line 300: To confirm, there is no upper bound for PID limit as long as it is a valid int right?

chienhanlin: Q. Should we add this new function to parse_linux.go because this PID limit only works for Linux? The same question for type.go. Thanks.

sparrc: My understanding of the "platform variables" struct is that it will be for variables that can never be supported by linux. Since this feature can be supported by windows there's no need to isolate it to a different linux-specific config struct.

sparrc: Seems like the unofficial limit in OCI/containerd is 4194304 (1024*1024*4):
```
% docker run -d --pids-limit 4194304 ubuntu:latest sleep 999
47f4622c595cf0b7672280c62d7483d4d9f5ba69f25c0f40567d4243bd882db3

% docker run -d --pids-limit 4194305 ubuntu:latest sleep 999
f436b71d39c6fc7afbf408aa2fb64ddb7863d0c6e0b58dbcc32a4273346bfbb5
docker: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error setting cgroup config for procHooks process: failed to write "4194305": write /sys/fs/cgroup/pids/docker/f436b71d39c6fc7afbf408aa2fb64ddb7863d0c6e0b58dbcc32a4273346bfbb5/pids.max: invalid argument: unknown.
```
I will go with this limit in our code too, good callout, thanks!!

agent/api/task/task_linux.go
Realmonia: It seems we create cgroup resources by ourselves - https://github.com/aws/amazon-ecs-agent/blob/master/agent/taskresource/cgroup/cgroup.go#L269-L283

sparrc: correct, task-level cgroup is created ourselves. We use the containerd/cgroups for this already so the implementation is easy to port (https://github.com/containerd/cgroups)

Realmonia:
Do we need to update https://github.com/aws/amazon-ecs-agent/blob/dev/agent/config/parse_unsupported.go too?

agent/config/types.go
```diff
@@ -370,4 +370,10 @@ type Config struct {
 	// uses to assign host ports from, for a container port range mapping.
 	// This defaults to the platform specific ephemeral host port range
 	DynamicHostPortRange string
+
+	// TaskPidsLimit specifies the per-task pids limit cgroup setting for each
+	// task launched on this container instance. This setting maps to the pids.max
+	// cgroup setting at the ECS task level.
+	// see https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#pid
+	TaskPidsLimit int
```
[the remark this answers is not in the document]
sparrc: done!

Realmonia (approved):

chienhanlin (approved):

fierlion (commented):
(comment resolved)

PR #3758, opened 2023-06-21
Improve test coverage of v2, v3, and v4 container stats endpoints

amogh09:
### Summary

Improve test coverage for v2, v3, and v4 container stats endpoints. 

This change is in a series of test coverage improvements we are making for TMDS endpoints. Test coverage is being improved to gain more confidence in the ongoing TMDS refactoring work.

Some previous PRs in this series are #3729, #3722, and #3708. All happy and error cases are consolidated into one test function per endpoint. 

### Implementation
* Updated the generic `TMDSTestCase` type and `testTMDSRequest` function to support setting expectations on mock Stats Engine.
* Extended the `TMDSResponse` interface to support container stats endpoints' response types. 
* Replaced existing `TestV2ContainerStats`, `TestV3ContainerStats`, and `TestV4ContainerStats` functions that only test happy cases with new functions that test all error and happy cases.

### Testing

New tests cover the changes:  NA

### Description for the changelog

Improve test coverage of v2, v3, and v4 container stats endpoints

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3791, opened 2023-07-07
Move v4 TMDS container and task stats endpoint handlers to ecs-agent module

amogh09:
### Summary

Moving v4 TMDS container and task stats endpoint handlers to ecs-agent module.

The only difference between old handlers and the new handlers is that the latter support publishing metrics. The new handlers are coded to publish an Internal Server Error metric if a 5XX error occurs during the handling of a request. When consuming the new handlers, Agent is set up to inject a no-op metrics publisher, so there is no actual change in functionality introduced in this PR.

#### ecs-agent module - 
* Add two new methods - `GetContainerStats` and `GetTaskStats` - to `AgentState` interface for getting container and task stats, respectively, from Agent. 
* Add new HTTP handlers - `ContainerStatsHandler` and `TaskStatsHandler` - for container stats and task stats endpoints, respectively. The handlers depend on the new methods added to `AgentState` interface. 
* Both the new handlers follow the same pattern -
    1. read container ID from the request,
    2. get stats by calling the appropriate `AgentState` method, and
    3. handle any errors and write response.
* The common workflow of the new handlers is captured in a new generic `statsHandler` function. Depending on the parameters provided it can work for both endpoints.
* Add tests for the new handlers.

#### agent module - 
* Implement new `AgentState` interface methods for `TMDSAgentState` type. The implementation is based on the existing implementation for stats endpoint handlers. 
* Update TMDS setup to consume v4 stats handlers from ecs-agent module. A no-op metrics publisher is used so that no metrics are published by the handlers.
* Delete old stats endpoint handlers.

### Testing

Test-driven development was followed. Test coverage for container and task stats endpoints was improved in #3758 and #3761 before these changes to capture the current customer facing behavior of the endpoints. The same tests are passing for this PR, so there is no regression.

Manual regression, stress, and performance tests were also performed. For manual testing, agent was build from source of this PR and deployed to an EC2 instance. Another EC2 instance was provisioned with released Agent version v1.72.0. 

For manual regression testing, container and task stats endpoints were called on both the instances and the results were compared using `diff`. No regression was detected. 

For manual stress testing, the endpoints were called at a rate of 3000 rps for 60 seconds. 100% of requests were successful. Note that the default TMDS rate limits are 40 steady and 60 burst, and so 3000 rps is considerably higher than the default. 

For manual performance testing, agent was profiled for heap and CPU usage while stress tests were being performed as explained above. Similar heap and CPU usage was seen for both the Agents. 

Move v4 TMDS container and task stats endpoint handlers to ecs-agent module

### Licensing


agent/handlers/v4/tmdsstate.go
```diff
@@ -149,3 +153,49 @@ func (s *TMDSAgentState) getTaskMetadata(v3EndpointID string, includeTags bool)
...
+	containerID, ok := s.state.DockerIDByV3EndpointID(v3EndpointID)
+	if !ok {
+		return tmdsv4.StatsResponse{}, tmdsv4.NewErrorStatsLookupFailure(fmt.Sprintf(
+			"V4 container stats handler: unable to get container ID from request: unable to get docker ID from v3 endpoint ID: %s",
+			v3EndpointID))
+	}
+
+	dockerStats, network_rate_stats, err := s.statsEngine.ContainerDockerStats(taskARN, containerID)
+	if err != nil {
+		return tmdsv4.StatsResponse{}, tmdsv4.NewErrorStatsFetchFailure(fmt.Sprintf(
```
singholt, line 173: should be put the actual `err` in the input for `NewErrorStatsFetchFailure` so that its not lost in translation?

ecs-agent/tmds/handlers/v4/handlers.go
```diff
@@ -173,3 +185,79 @@ func getTaskErrorResponse(endpointContainerID string, err error) (int, string) {
...
+	var errStatsFetchFailure *state.ErrorStatsFetchFailure
+	if errors.As(err, &errStatsFetchFailure) {
+		return http.StatusInternalServerError, errStatsFetchFailure.ExternalReason()
+	}
+
+	// 500 if unknown failure
+	logger.Error("Unknown error encountered when handling stats fetch error", logger.Fields{
+		field.Error: err,
+	})
+	return http.StatusInternalServerError, "failed to get stats"
```
singholt, line 262: should we return `err` instead of "failed to get stats", when it comes under the `500 if unknown failure` category?

ecs-agent/tmds/handlers/v4/state/state.go
```diff
@@ -68,4 +103,16 @@ type AgentState interface {
...
+
+	// Returns container stats in v4 format for the container identified by the provided
+	// endpointContainerID.
+	// Returns ErrorStatsLookupFailure if container lookup fails.
+	// Returns ErrorStatsFetchFailure if something else goes wrong.
+	GetContainerStats(endpointContainerID string) (StatsResponse, error)
+
+	// Returns task stats in v4 format for the task identified by the provided
+	// endpointContainerID.
+	// Returns ErrorStatsLookupFailure if container lookup fails.
```
singholt, line 115: nit: if task lookup fails.

ecs-agent/tmds/handlers/v4/state/state.go
```diff
@@ -50,6 +50,41 @@ func (e *ErrorMetadataFetchFailure) ExternalReason() string {
...
+
+func NewErrorStatsLookupFailure(externalReason string) *ErrorStatsLookupFailure {
+	return &ErrorStatsLookupFailure{externalReason}
+}
+
+func (e *ErrorStatsLookupFailure) ExternalReason() string {
+	return e.externalReason
+}
+
+func (e *ErrorStatsLookupFailure) Error() string {
```
singholt, line 66: is this used? same for `func (e *ErrorStatsFetchFailure) Error() string`

agent/handlers/v4/tmdsstate.go
amogh09: Yeah that's a good catch. Let me test that.

ecs-agent/tmds/handlers/v4/handlers.go
amogh09: The second return value is the TMDS response. Since this is unknown error case I think it's better to return a vague error and log the real error.

agent/handlers/v4/tmdsstate.go
amogh09: I have now added a `cause` field to this error type.

ecs-agent/tmds/handlers/v4/state/state.go
amogh09: err.. fixed

ecs-agent/tmds/handlers/v4/handlers.go
singholt: i just noticed we dont return `err` even in cases of it being a known error.

ecs-agent/tmds/handlers/v4/handlers_test.go
```diff
@@ -306,8 +323,199 @@ func TestTaskMetadata(t *testing.T) {
...
+			expectedResponseBody: externalReason,
+		})
+	})
+
+	internalServerErrorCases := []struct {
+		err          error
+		responseBody string
+	}{
+		{
+			err:          state.NewErrorStatsFetchFailure(externalReason, errors.New("cause")),
```
singholt, line 373: should we assert this below? err == tc.err

singholt: nvm you're passing tc.err into the mock so it will be equal

singholt (approved):

ubhattacharjya (approved):

PR #3795, opened 2023-07-11
Update Agent to be more resilient in case of unauthenticated timeouts with IMDS

hozkaya2000:
### Summary
There is an edge case where the credentials within the agent used to access internal resources can be incorrect due to other processes restarting during a certain time interval. This can cause the agent to think that the credentials are not expired and correct. The change is to prevent this from happening, and refreshing the credentials in the first location this issue would appear, which would be `RegisterContainerInstance` in the ecsclient. 

### Implementation details
The issue occurs in the `RegisterContainerInstance` function in ecsclient's `client.go`. `RegisterContainerInstance` makes the call to IMDS in `SetInstanceIdentity` function where the client for communicating with IMDS is used with the call `client.ec2metadata.GetDynamicData(...)`. The new IMDS process does not work with the current credentials. 

In the fix, this line is replaced with a retry strategy. When we get an error code, the current credentials in agent are forced before the retry to be expired, and a `GetCredentials` call is made to refresh the credentials, with multiple retries if it fails again. The global instance credentials in the agent are refreshed with this call, so other calls made by agent to IMDS are no longer an issue.  

### Testing
To test, a new case was added to the `RegisterContainerInstanceTest` that is the exact same as the basic success case, but the first call to `client.ec2metadata.GetDynamicData(...)` was simulated to fail, and the retry call to succeed. The rest of the test validations remained, making sure that retries didn't prevent any other process from happening. 

Files: 
- agent/api/ecsclient/client.go
- agent/api/ecsclient/client_test.go

were changed as described in the implementation and testing details

### Licensing

### DIFF
--- a/agent/api/ecsclient/client.go
+++ b/agent/api/ecsclient/client.go
@@ -14,6 +14,7 @@
 package ecsclient
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"runtime"
@@ -24,12 +25,14 @@ import (
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/agent/api/container/status"
 	"github.com/aws/amazon-ecs-agent/agent/async"
 	"github.com/aws/amazon-ecs-agent/agent/config"
+	"github.com/aws/amazon-ecs-agent/agent/credentials/instancecreds"
 	"github.com/aws/amazon-ecs-agent/agent/ec2"
 	"github.com/aws/amazon-ecs-agent/agent/httpclient"
 	"github.com/aws/amazon-ecs-agent/agent/utils"
 	apierrors "github.com/aws/amazon-ecs-agent/ecs-agent/api/errors"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/ecs_client/model/ecs"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -53,6 +56,12 @@ const (
 	// ecsMaxNetworkBindingsLength is the maximum length of the ecs.NetworkBindings list sent as part of the
 	// container state change payload. Currently, this is enforced only when containerPortRanges are requested.
 	ecsMaxNetworkBindingsLength = 100
+	// Following constants used for SetInstanceIdentity retry with exponential backoff
+	setInstanceIdRetryTimeOut         = 30 * time.Second
+	setInstanceIdRetryBackoffMin      = 100 * time.Millisecond
+	setInstanceIdRetryBackoffMax      = 5 * time.Second
+	setInstanceIdRetryBackoffJitter   = 0.2
+	setInstanceIdRetryBackoffMultiple = 2
 )
 
 // APIECSClient implements ECSClient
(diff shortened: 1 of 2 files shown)


agent/api/ecsclient/client.go
```diff
@@ -226,7 +235,24 @@ func (client *APIECSClient) setInstanceIdentity(registerRequest ecs.RegisterCont
...
 
 	iidRetrieved := true
-	instanceIdentityDoc, err := client.ec2metadata.GetDynamicData(ec2.InstanceIdentityDocumentResource)
+	backoff := retry.NewExponentialBackoff(setInstIdRetryBackoffMin, setInstIdRetryBackoffMax,
+		setInstIdRetryBackoffJitter, setInstIdRetryBackoffMultiple)
+	ctx, cancel := context.WithTimeout(context.Background(), setInstIdRetryTimeOut)
+	defer cancel()
+	err := retry.RetryWithBackoffCtx(ctx, backoff, func() error {
+		var attemptErr error
+		seelog.Debugf("Attempting to get IID")
```
mythri-garaga, line 244: Super nit: can we rephrase the log line as "Attempting to get instance identity document"

agent/api/ecsclient/client.go
```diff
@@ -226,7 +235,24 @@ func (client *APIECSClient) setInstanceIdentity(registerRequest ecs.RegisterCont
...
+		instanceIdentityDoc, attemptErr = client.ec2metadata.GetDynamicData(ec2.InstanceIdentityDocumentResource)
+		if attemptErr != nil {
+			seelog.Debugf("Unable to get instance identity document, retrying: %v", attemptErr)
+			// force credentials to expire in case they are stale but not expired (e.g. live update)
+			client.credentialProvider.Expire()
+			client.credentialProvider = instancecreds.GetCredentials(client.config.External.Enabled())
+			return apierrors.NewRetriableError(apierrors.NewRetriable(true), attemptErr)
+		}
+		seelog.Debugf("Successfully retrieved the IID")
```
mythri-garaga, line 253: same here, can we change this to "Successfully retrieved the instance identity document"

agent/api/ecsclient/client.go
```diff
@@ -53,6 +56,12 @@ const (
 	// ecsMaxNetworkBindingsLength is the maximum length of the ecs.NetworkBindings list sent as part of the
 	// container state change payload. Currently, this is enforced only when containerPortRanges are requested.
 	ecsMaxNetworkBindingsLength = 100
+	// Following constants used for SetInstanceIdentity retry with exponential backoff
+	setInstIdRetryTimeOut         = 30 * time.Second
+	setInstIdRetryBackoffMin      = 100 * time.Millisecond
+	setInstIdRetryBackoffMax      = 5 * time.Second
+	setInstIdRetryBackoffJitter   = 0.2
+	setInstIdRetryBackoffMultiple = 2
```
Realmonia, line 64: nit: Can you change to use `InstanceID` instead of `InstId` in those consts?

agent/api/ecsclient/client.go
Realmonia, line 244: nit: Use Instance ID in log will be easier.

agent/api/ecsclient/client.go
```diff
@@ -226,7 +235,24 @@ func (client *APIECSClient) setInstanceIdentity(registerRequest ecs.RegisterCont
...
+		setInstIdRetryBackoffJitter, setInstIdRetryBackoffMultiple)
+	ctx, cancel := context.WithTimeout(context.Background(), setInstIdRetryTimeOut)
+	defer cancel()
+	err := retry.RetryWithBackoffCtx(ctx, backoff, func() error {
+		var attemptErr error
+		seelog.Debugf("Attempting to get IID")
+		instanceIdentityDoc, attemptErr = client.ec2metadata.GetDynamicData(ec2.InstanceIdentityDocumentResource)
+		if attemptErr != nil {
+			seelog.Debugf("Unable to get instance identity document, retrying: %v", attemptErr)
+			// force credentials to expire in case they are stale but not expired (e.g. live update)
```
Realmonia, line 248: Change comment (reword live update)

Realmonia (changes requested):
Please force push to remove the internal link from commit message.

mythri-garaga (approved):

Realmonia (approved):

PR #3802, opened 2023-07-13
Add GuardDuty capability

SreeeS:
### Summary

This PR adds `guard-duty-patrol` to the list of capabilities and advertises capability `ecs.capability.guard-duty-patrol`  indicating that guardDuty is supported if the unit file `guardDuty.service` is present.

### Implementation details

* Added `guard-duty-patrol` to the list of capabilities
* used `os.Stat` to check for unit file `guardDuty.service`
* Advertise capability based on platform.

### Testing

* Manual test using `ecs.service` unit file and check if capability is advertised.

New tests cover the changes: 

Add GuardDuty capability
### Licensing


agent/app/agent_capability.go
```diff
@@ -77,10 +77,13 @@ const (
 	capabilityExecCertsRelativePath                        = "certs"
 	capabilityExternal                                     = "external"
 	capabilityServiceConnect                               = "service-connect-v1"
+	capabilityGuardDuty                                    = "guard-duty-patrol"
 
 	// network capabilities, going forward, please append "network." prefix to any new networking capability we introduce
 	networkCapabilityPrefix      = "network."
 	capabilityContainerPortRange = networkCapabilityPrefix + "container-port-range"
+	//using guardDuty.service as a temporary placeholder
```
Realmonia, line 85: nit: if this is not finalized, can you add a `TODO` here?

Realmonia (approved):

agent/app/agent_capability.go
```diff
@@ -498,6 +503,24 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
...
 
+func (agent *ecsAgent) appendGuardDutyCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	isEnabled, err := pathExists(guardDutyServiceFilePath, false)
+	if err != nil {
+		logger.Warn(fmt.Sprintf("Skipping enabling GuardDutyCapabilities. Error: %v", err))
+	}
+
+	supported, err := isGuardDutySupported()
+	if err != nil || !supported {
+		logger.Warn(fmt.Sprintf("Skipping enabling GuardDutyCapabilities. "+
```
sparrc, line 514: maybe make this a debug statement? otherwise we will log a warning every time the agent starts up on windows

agent/app/agent_capability_windows.go
```diff
@@ -142,3 +143,7 @@ func defaultIsPlatformExecSupported() (bool, error) {
 	}
 	return true, nil
 }
+
+func defaultIsGuardDutySupported() (bool, error) {
+	return false, errors.New("GuardDuty is currently not supported on windows")
```
singholt, line 148: return `false, nil` like in `agent_capability_unspecified.go`? This function is a no-op for unsupported platforms, rather than returning error by default

agent/app/agent_capability.go
```diff
@@ -498,6 +503,24 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 	return capabilities
 }
 
+func (agent *ecsAgent) appendGuardDutyCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	isEnabled, err := pathExists(guardDutyServiceFilePath, false)
```
singholt, line 507: this check could go in`isGuardDutySupported()` in the `agent_capability_unix.go`. That way we return true if pathExists and append capability - much simpler to read imo.

agent/app/agent_capability.go
```diff
@@ -77,10 +77,13 @@ const (
 	capabilityExecCertsRelativePath                        = "certs"
 	capabilityExternal                                     = "external"
 	capabilityServiceConnect                               = "service-connect-v1"
+	capabilityGuardDuty                                    = "guard-duty-patrol"
 
 	// network capabilities, going forward, please append "network." prefix to any new networking capability we introduce
 	networkCapabilityPrefix      = "network."
 	capabilityContainerPortRange = networkCapabilityPrefix + "container-port-range"
+	//using guardDuty.service as a temporary placeholder
+	guardDutyServiceFilePath = "/usr/lib/systemd/system/guardDuty.service"
```
chienhanlin, line 86: Q. Should we define this in the [config.go] instead of here since it is not a capability?

agent/app/agent_capability_unix.go
```diff
@@ -232,3 +232,7 @@ func (agent *ecsAgent) getTaskENIPluginVersionAttribute() (*ecs.Attribute, error
 func defaultIsPlatformExecSupported() (bool, error) {
 	return true, nil
 }
+
+func defaultIsGuardDutySupported() (bool, error) {
```
singholt, line 236: based on my comment to move pathExists() here, lets add a unit test for this function

agent/app/agent_capability.go
```diff
@@ -297,6 +302,7 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 		}
 		capabilities = removeAttributesByNames(capabilities, externalUnsupportedCapabilities)
 	}
+	capabilities = agent.appendGuardDutyCapabilities(capabilities)
```
chienhanlin, line 305: Could we add a comment `// add GuardDuty capabilities if applicable` for this line?

agent/app/agent_capability.go
```diff
@@ -498,6 +503,24 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 	return capabilities
 }
 
+func (agent *ecsAgent) appendGuardDutyCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	isEnabled, err := pathExists(guardDutyServiceFilePath, false)
+	if err != nil {
+		logger.Warn(fmt.Sprintf("Skipping enabling GuardDutyCapabilities. Error: %v", err))
```
chienhanlin, line 509: nit. Can we use this format for log message?

```
logger.Warn("GuardDuty Capability: Cannot find GuardDuty service file path. GuardDuty capabilities will not be enabled on this container instance",   logger.Fields{
    field.Error:  err,
})
```

agent/app/agent_capability.go
```diff
@@ -498,6 +503,24 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 	return capabilities
 }
 
+func (agent *ecsAgent) appendGuardDutyCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
```
chienhanlin, line 506: Q. Should we define this in [agent_capability_unix.go](https://github.com/aws/amazon-ecs-agent/blob/master/agent/app/agent_capability_unix.go) since GuardDuty will not be supported on windows and unspecified platforms?

agent/app/agent_capability.go
SreeeS: Updated this

agent/app/agent_capability.go
SreeeS: Changed this to a debug statement.

agent/app/agent_capability_windows.go
SreeeS: Updated this

agent/app/agent_capability.go
SreeeS: Updated this

agent/app/agent_capability_unix.go
SreeeS: Now we have the check here and added the unit test.

agent/app/agent_capability.go
SreeeS: I agree, moved this to config.go

agent/app/agent_capability.go
SreeeS: Added the comment.

agent/app/agent_capability.go
SreeeS: Updated to use this format

agent/app/agent_capability.go
```diff
@@ -498,6 +503,19 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
...
+	supported, _ := isGuardDutySupported()
+	if !supported {
+		seelog.Debugf("GuardDuty is not supported. Skipping appendGuardDutyCapabilities.")
+		return capabilities
+	}
```
singholt, line 511: this check is redundant since we're appending the attribute only if supported

agent/app/agent_capability_unix.go
```diff
@@ -232,3 +235,14 @@ func (agent *ecsAgent) getTaskENIPluginVersionAttribute() (*ecs.Attribute, error
...
+	isEnabled, err := pathExists(config.GuardDutyServiceFilePath, false)
+	if err != nil {
+		logger.Warn("GuardDuty Capability: Cannot find GuardDuty service file path. GuardDuty capabilities will "+
```
singholt, line 242: based on https://github.com/aws/amazon-ecs-agent/blob/b13992a87cfc5f89ce338d58c12fd7c61d4f6c83/agent/app/agent_capability.go#L575-L582

I think err !=nil implies that something other "not-exists" err was returned. this warn statement can be like

```
logger.Warn("Unable to determine if GuardDuty is installed, not reporting the capability", logger.Fieldslogger.Fields{
	field.Error: err,
})
```

sparrc, line 242: I don't think this should be a warning, this would log a warning for every customer that doesn't have the guardduty RPM installed right?

sparrc (changes requested):

agent/app/agent_capability_unix.go
singholt: logger.Debug seems more apt

SreeeS: Modified this.

agent/app/agent_capability.go
SreeeS: removed this

agent/app/agent_capability_unix.go
SreeeS: updated this

agent/app/agent_capability.go
```diff
@@ -498,6 +503,19 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 	return capabilities
 }
 
+func (agent *ecsAgent) appendGuardDutyCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	supported, _ := isGuardDutySupported()
```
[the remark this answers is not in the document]
SreeeS: updated to not return error

agent/app/agent_capability_unix_test.go
```diff
@@ -925,3 +925,37 @@ func TestAppendFSxWindowsFileServerCapabilities(t *testing.T) {
 	assert.Equal(t, len(inputCapabilities), len(capabilities))
 	assert.EqualValues(t, capabilities, inputCapabilities)
 }
+
+func TestIsGuardDutySupported(t *testing.T) {
+	testCases := []struct {
+		name                         string
+		expectedIsGuardDutySupported bool
```
singholt, line 932: both test cases have this set to false, is that the intention?

agent/app/agent_capability_unix.go
```diff
@@ -20,6 +20,9 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
```
singholt, line 23: nit: lets club these imports with `agent/` imports below and run goimports.

agent/app/agent_capability_unix_test.go
SreeeS: This need to be true, this might have got update during my previous commit. I missed this somehow. updated this

agent/app/agent_capability_unix.go
SreeeS: updated this

singholt (dismissed):

agent/config/config.go
```diff
@@ -177,6 +177,9 @@ const (
 	ContainerInstancePropagateTagsFromEC2InstanceType
 )
 
+// TODO: update path once finalized. using guardDuty.service as a temporary placeholder
+const GuardDutyServiceFilePath = "/usr/lib/systemd/system/guardDuty.service"
```
chienhanlin, line 181: Q. should this path defined in [config_unix.go](https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config_unix.go)?

SreeeS: moved it to [config_unix.go](https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/config_unix.go)

agent/app/agent_capability.go
```diff
@@ -498,6 +503,14 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
 	return capabilities
 }
 
+func (agent *ecsAgent) appendGuardDutyCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	supported := isGuardDutySupported()
+	if supported {
+		return appendNameOnlyAttribute(capabilities, attributePrefix+capabilityGuardDuty)
+	}
```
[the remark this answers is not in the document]
SreeeS: updated this

singholt (approved):

Realmonia (approved):

agent/app/agent_capability_unix.go
SreeeS: Since I updated this change in the latest commit,Resolving this request.

sparrc (approved):

PR #3822, opened 2023-07-20
greedy task resource accounting instead of blocking on tasks ahead in the queue

singholt:
### Summary

Traditionally, the ECS agent wouldn't start new tasks until all stopping tasks have stopped on a container instance. 

With the task resource accounting project, we replaced this serialization with a resource accounting strategy that uses a FIFO queue to serve incoming tasks: https://github.com/aws/amazon-ecs-agent/pull/3819. If there are no resources available on the host, we queue tasks up until resources are available.

As soon as resources become available, the first task in the queue gets priority. The other waiting tasks continue to wait, until the first task gets served. There are corner cases when the first task in the queue is large enough to block all other tasks from being served (forever).

This PR changes the above behavior. With this change, although we give preference to tasks in a FIFO fashion, we will try to run as many tasks from the queue as possible, instead of blocking on tasks ahead in the queue that cannot start.

Example:
Consider a container instance with memory 1024 MiB. 3 tasks land on the instance: task T1 with memory 512 MiB, task T2 with memory 1024 MiB, and task T3 with memory 256 MiB.

T1 will run, since the requested memory is available. T2 needs more memory (1024 MiB) than whats available (512 MiB), hence it will be placed in the wait queue. T3 will also run because the requested memory is available. Prior to this change, the T3 would not run until T2 has been served first (and the only way T2 would run is if T1 stops).

**Note**:
- The target branch for this PR is feature/greedy-task-resource-accounting.
- There are corner cases with this approach as well, for example, there's a scenario when smaller tasks will always start and large tasks may be stuck in the queue. After speaking with product, we plan to introduce a mechanism to remove tasks from the queue if it has been in there for too long (with a wait timeout). This will be addressed in a follow-up PR.

### Implementation details

- Updated the `dequeueTask()` to dequeue at the given index, instead of dequeuing first task always.
- Updated the `topTask()` to `getTaskByIndex()` to return a task at the given index, instead of returning the first task always.
- Updated the `monitorQueuedTasks()` to try dequeuing as many tasks as possible, instead of stopping if unable to dequeue the first task in the queue.

### Testing

- Updated the unit test to mimic how tasks will be queued up when resource requirements are identical and different.
- Integration tests will be updated/added in a follow-up PR.

enhancement: switch to greedy task resource accounting instead of blocking on tasks ahead in the queue.

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -260,7 +260,11 @@ func (engine *DockerTaskEngine) reconcileHostResources() {
 		if taskStatus.Terminal() {
 			err := engine.hostResourceManager.release(task.Arn, resources)
 			if err != nil {
-				logger.Critical("Failed to release resources during reconciliation", logger.Fields{field.TaskARN: task.Arn})
+				logger.Critical("Failed to release resources during reconciliation",
+					logger.Fields{
+						field.TaskARN: task.Arn,
+						field.Error:   err,
+					})
 			}
 			continue
 		}
@@ -271,7 +275,11 @@ func (engine *DockerTaskEngine) reconcileHostResources() {
 		if !task.IsInternal && task.HasActiveContainers() {
 			consumed, err := engine.hostResourceManager.consume(task.Arn, resources)
 			if err != nil || !consumed {
-				logger.Critical("Failed to consume resources for created/running tasks during reconciliation", logger.Fields{field.TaskARN: task.Arn})
+				logger.Critical("Failed to consume resources for created/running tasks during reconciliation",
+					logger.Fields{
+						field.TaskARN: task.Arn,
+						field.Error:   err,
+					})
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -347,34 +355,38 @@ func (engine *DockerTaskEngine) wakeUpTaskQueueMonitor() {
...
 	defer engine.waitingTasksLock.Unlock()
-	if len(engine.waitingTaskQueue) > 0 {
-		task := engine.waitingTaskQueue[0]
-		engine.waitingTaskQueue = engine.waitingTaskQueue[1:]
-		logger.Debug("Dequeued task from Waiting Task Queue", logger.Fields{field.TaskARN: task.Arn})
-		return task, nil
-	}
-
-	return nil, fmt.Errorf("no tasks in waiting queue")
+	task := engine.waitingTaskQueue[index]
```
mye956, line 386: Q: should we validate the index here before dequeuing?

singholt: good callout, I went back and forth in my head about this. The assumption here is the current callers call this method with valid indexes. I will update this to remove this assumption and validate to err on the safe side.

singholt: updated

agent/engine/docker_task_engine.go
```diff
@@ -347,34 +355,50 @@ func (engine *DockerTaskEngine) wakeUpTaskQueueMonitor() {
 	}
 }
 
-func (engine *DockerTaskEngine) topTask() (*managedTask, error) {
+func (engine *DockerTaskEngine) validateIndex(index int) error {
+	waitingTaskQueueLength := len(engine.waitingTaskQueue)
```
chienhanlin, line 359: Q. Would `engine.waitingTaskQueue` ever be nil and we should add a check before checking `len()`?

agent/engine/docker_task_engine.go
```diff
@@ -347,34 +355,50 @@ func (engine *DockerTaskEngine) wakeUpTaskQueueMonitor() {
...
 	}
-
-	return nil, fmt.Errorf("no tasks in waiting queue")
+	task := engine.waitingTaskQueue[index]
+	engine.waitingTaskQueue = append(engine.waitingTaskQueue[:index], engine.waitingTaskQueue[index+1:]...)
+	logger.Debug("Dequeued task from the waiting task queue", logger.Fields{field.TaskARN: task.Arn})
+	return task, nil
 }
 
 // monitorQueuedTasks starts as many tasks as possible based on FIFO order of waitingTaskQueue
```
prateekchaudhry, line 404: We should update the comment with the new behavior, as it will not be FIFO any more and try to start whichever task in the queue it can

agent/engine/docker_task_engine.go
```diff
@@ -391,65 +403,77 @@ func (engine *DockerTaskEngine) monitorQueuedTasks(ctx context.Context) {
...
 			return
 		case <-engine.monitorQueuedTaskEvent:
-			// Dequeue as many tasks as possible and start wake up their goroutines
+			// Dequeue as many tasks as possible and start waking up their goroutines
+			index := 0
 			for {
-				task, err := engine.topTask()
+				task, err := engine.getTaskByIndex(index)
 				if err != nil {
+					// either the queue is empty or we've traversed through all the tasks in the queue
```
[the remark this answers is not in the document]
singholt: sounds good, I'll add a debug log here

agent/engine/docker_task_engine.go
mye956: thanks!

agent/engine/docker_task_engine.go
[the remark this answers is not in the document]
singholt: updated

ubhattacharjya (approved):

mye956 (approved):

agent/engine/docker_task_engine.go
```diff
@@ -391,66 +415,116 @@ func (engine *DockerTaskEngine) monitorQueuedTasks(ctx context.Context) {
...
+					logger.Debug("Unable to get task from the wait queue at the given index", logger.Fields{
+						field.TaskARN: task.Arn,
+						field.Error:   err,
+						"index":       index,
+					})
 					break
 				}
-				dequeuedTask := engine.tryDequeueWaitingTasks(task)
+				dequeuedTask := engine.tryDequeueWaitingTasks(task, index)
 				if !dequeuedTask {
```
chienhanlin, line 432: Q. As index is incrementing through the loop, are these unable-to-dequeue tasks only be validated again when there is a new `engine.monitorQueuedTaskEvent` coming? or this will not happen as we are going to remove them in the follow-up PR? Thanks!

agent/engine/docker_task_engine.go
singholt: nope, `engine.waitingTaskQueue` is initialized as part of the docker task engine with a len 0.

example: https://go.dev/play/p/q_yCqE1Qxzr.

agent/engine/task_manager_test.go
```diff
@@ -2211,19 +2211,94 @@ func TestTaskWaitForHostResources(t *testing.T) {
...
+	time.Sleep(500 * time.Millisecond)
+
+	// Verify arn2 got dequeued and the queue is empty
+	_, err = taskEngine.getTaskByIndex(0)
+	assert.Error(t, err)
+	assert.Empty(t, taskEngine.waitingTaskQueue)
+}
+
+// TestTaskWaitForHostResourcesWithDifferentRequests tests task queuing behavior based on available host resources
+// for tasks with different resource requests
```
prateekchaudhry, line 2235: Does this test still verify FIFO order? If not, should we add another test to verify queue monitor is able to follow non FIFO requests, where a task added to the queue later in the queue can get started if a task earlier in the queue does not have enough host resources available yet

agent/engine/docker_task_engine.go
singholt: tasks in a queue are evaluated in two scenarios,

- when a running task stops
- when there's a new task payload received from ACS. This could either be to _run_ a new task or a _stop_ an existing task (either in the queue or running).

agent/engine/docker_task_engine.go
singholt: I'll update it. It is FIFO in terms of preference of evaluation but not blocking on tasks that unable to start.

singholt: will remove the term "FIFO" to avoid confusion.

```
// monitorQueuedTasks starts as many tasks as possible based on FIFO order of waitingTaskQueue
```

agent/engine/task_manager_test.go
singholt: good catch, I intended this to follow the example in the PR description. I will reorder the tasks to reflect that.

agent/engine/docker_task_engine.go
singholt: updated

agent/engine/task_manager_test.go
singholt: updated

prateekchaudhry (approved):

chienhanlin (approved):

PR #3827, opened 2023-07-24
Move appnet client interface to ecs-agent

RichaGangwar:
### Summary
This PR moves Appnet client interface to ecs-agent. The PR also makes minor changes in order to make the library shareable.

### Implementation details
1. Move AppNetClient interface from agent/api/interface.go to ecs-agent/api/appnet/client.go
2. Update the mocks for both the files mentioned above.
3. Update all the usages of AppNetClient to consume from new location.
4. In appnet client library, change the method Client() to more apt CreateClient() method name.
5. Change the client to more apt AppNetAgentClient.

### Testing

Tested using existing test cases as no new functionality is introduced.

Move appnet client interface to ecs-agent

### Licensing


Realmonia (approved):

ecs-agent/api/appnet/client.go
```diff
@@ -18,12 +18,22 @@ import (
...
+// AppnetClient is an interface with customized Appnet client that
+// implements the GetStats and DrainInboundConnections
+type AppNetClient interface {
+	GetStats(adminSocketPath string, statsRequest string) (map[string]*prometheus.MetricFamily, error)
+	DrainInboundConnections(adminSocketPath string, drainRequest string) error
+}
+
+type AppNetAgentClient struct {
 	udsHttpClient http.Client
+	AppNetClient
```
danehlim, line 36: Is embedding the `AppNetClient` interface here necessary?

ecs-agent/api/appnet/client.go
```diff
@@ -18,12 +18,22 @@ import (
...
 
-type client struct {
+// AppnetClient is an interface with customized Appnet client that
+// implements the GetStats and DrainInboundConnections
+type AppNetClient interface {
+	GetStats(adminSocketPath string, statsRequest string) (map[string]*prometheus.MetricFamily, error)
+	DrainInboundConnections(adminSocketPath string, drainRequest string) error
+}
+
+type AppNetAgentClient struct {
```
yinyic, line 34: Nit: `Client` would have been fine (just like `http.Client` below). since this is already scoped to the `appnet` package, there won't be ambiguities. If we compare `appnet.Client` and `appnet.AppNetAgentClient`, the latter seems a bit redundant.

yinyic (approved):

ecs-agent/api/appnet/client.go
RichaGangwar: Thank you for bringing it up. Actually you are right, it would have been okay to leave it as client. I brought in the change for parity with the other agent.

ecs-agent/api/appnet/client.go
RichaGangwar: It is not absolutely necessary, but it is surely a good way to promote the methods implemented by the interface to the enclosing struct, i.e. AppNetAgentClient. We do follow similar pattern in most other cases as well, for example, wsclient:
https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/wsclient/client.go#L180
Let me know if you think we should not include it.

PR #3828, opened 2023-07-25
add a wait timeout for greedy task resource accounting

singholt:
### Summary

This PR is a follow up to #3822 where we switched to greedy task resource accounting (see [PR description](https://github.com/aws/amazon-ecs-agent/pull/3822#issue-1814904558) for details). 
One-line summary: With greedy task resource accounting, the ECS agent starts as many tasks as it can from the wait queue, instead of blocking on tasks ahead in the queue.

With greedy accounting, there are certain corner cases where smaller tasks will always start and larger tasks may be stuck in the queue. For example, consider the following sequence of events on a container instance with 1024 memory:
- 2 tasks – T1 requesting 512 memory, and T2 requesting 1024 memory are placed on the instance. T1 will run and T2 will be placed in the wait queue. 
- A new task T3 requesting 128 memory is placed. It will transition to running since the required memory is available. 
- T1 stops. T2 still cannot run since the container instance only has 896 memory.
- A new task T4 requesting 256 memory is placed. It will transition to running since the required memory is available. 
- T3 stops. T2 still cannot run since the container instance only has 768 memory.

In the worst case, the above pattern would continue to occur as long as small tasks like T1, T3, T4 are placed on the instance and large tasks like T2 cannot get enough resources to consume. To mitigate this, we're adding a configurable resource wait timeout. Once this timeout is reached, the task will be removed from the wait queue and set to terminal. This will enable ECS to retry the task placement.

### Implementation details

- Add a new config option for task resource wait timeout to `agent/config/`. I've added placeholder values for the default and minimum resource wait timeout configuration to unblock PR review and further updates/testing. These values will be updated after confirming internally with the product team.
- Update `waitForHostResources()` to kick off a timer based on the resource wait timeout when a task is queued. Once this timer expires, the task is set to terminal with an meaningful reason. This will sent to ECS backend as part of the task state change events.

### Testing

- Updated unit tests
- Integration and functional tests will be added in a follow-up PR.

### Licensing


Realmonia (approved):

agent/config/config.go
```diff
@@ -623,6 +647,7 @@ func (cfg *Config) String() string {
 			"PollingMetricsWaitDuration: %v, "+
 			"ReservedMem: %v, "+
 			"TaskCleanupWaitDuration: %v, "+
+			"TaskResourceWaitTimeout: %v"+
```
mye956, line 650: Q: Should there be a `,` here before `%v` and `"+`?

singholt: good catch, updated!

mye956 (approved):

Realmonia (approved):

PR #3830, opened 2023-07-26
Refactor ACS refresh credentials message handling

danehlim:
### Summary

Refactor ACS refresh credentials message handling, update other associated code, and improve/update tests.

### Implementation details

1. Refactor ACS refresh credentials message handling
   - `agent/acs/handler/refresh_credentials_handler.go` -> `ecs-agent/acs/session/refresh_credentials_responder.go` (partial)
      - All components of the responder that do not directly interact with Agent state belong in this file
      - Refactor `refreshCredentialsHandler` struct and existing code into `refreshCredentialsResponder` struct, such that `refreshCredentialsResponder` struct implements the `wsclient.RequestResponder` interface
      - Move validation of the refresh credentials message's role type to the `validateIAMRoleCredentialsMessage` function
      - Add support for publishing of metrics in various credentials refresh failure/success cases
         - NOTE: Our Agent defined in `agent/` directory does not support metrics publishing, thus in `agent/acs/handler/acs_handler.go` our agent will supply the refresh credentials responder with a `metrics.NewNopEntryFactory()`, which is a no-op implementation of metrics publishing 
   - `agent/acs/handler/refresh_credentials_handler.go` -> `agent/acs/handler/refresh_credentials_responder.go` (partial)
      - Components that interact with Agent state belong in this file
      - Refactor functionality from such components mentioned above into `credentialsMetadataSetter` struct such that `credentialsMetadataSetter` struct implements the `CredentialsMetadataSetter` interface defined in `ecs-agent/acs/session/refresh_credentials_responder.go`
   - `agent/acs/handler/refresh_credentials_handler_linux.go` -> `agent/acs/handler/refresh_credentials_responder_linux.go` (file renamed)
   - `agent/acs/handler/refresh_credentials_handler_windows.go` -> `agent/acs/handler/refresh_credentials_responder_windows.go` (file renamed)
      - Initialize refresh credentials responder and add its `HandlerFunc` function as a request handler to the ACS client, such that refresh credentials responder dictates what to do in response to receiving refresh credentials messages from ACS
      - Remove refresh credentials messages from pending ACKs logic since it is no longer applicable after the refactor of ACS refresh credentials message handling
         - Such logic is no longer applicable since there is no need to do multi-thread synchronization with channels with this responder (like we currently do). As part of the refactor, we factor out this non-necessity and thus methods such as `start()`, `stop()`, `clearAcks()`, etc. can be removed. The resulting responder is a lot cleaner and simpler to follow.
   - `agent/acs/handler/payload_handler.go`
      - Don't pass in the entire refresh credentials responder to the payload responder as it is only being passed in to send ACKs of credentials associated with task(s) back to ACS. Instead, just write a private payload responder method `makeCredentialsAckRequest` and use it to accomplish the same while avoiding passing in a whole entire responder to another responder
         - For context, sending ACKs of credentials in the payload responder is necessary because payload responder handles credentials that are associated with task(s) contained in a payload message (type `ecsacs.PayloadMessage`). In contrast, refresh credentials responder handles credentials that are contained in a refresh credentials message (type `ecsacs.IAMRoleCredentialsMessage`)
   - `ecs-agent/acs/session/generate_mocks.go` and `ecs-agent/acs/session/mocks/session_mock.go`
      - Generate new mocks for the `CredentialsMetadataSetter` interface to be used in testing
   - `ecs-agent/metrics/constants.go`
      - Add constants specific to refresh credentials responder 
3. Improve/update tests, most notably:
   - Update existing tests accordingly based off of above changes to refactor ACS refresh credentials message handling in the following files:
      - `agent/acs/handler/refresh_credentials_handler_test.go` -> `ecs-agent/acs/session/refresh_credentials_responder_test.go` (partial)
         - Refresh credentials responder tests that only need to interact with stateless components (e.g., message validation) belong in this file
      - `agent/acs/handler/refresh_credentials_handler_test.go` -> `agent/acs/handler/refresh_credentials_responder_test.go` (partial)
         - Refresh credentials responder tests that need to directly interact with Agent state belong in this file
      - `agent/acs/handler/acs_handler_test.go`
   -  Update existing tests in `agent/acs/handler/payload_handler_test.go` accordingly based off of above changes to `agent/acs/handler/payload_handler.go`
   - General cleanup/quality of life updates to tests (e.g., use `assert` package for test assertions instead of `if` statement blocks)
   - Replace usage of test constants currently defined in `agent/acs/handler/refresh_credentials_handler_test.go` with test constants defined in centralized file `ecs-agent/acs/session/testconst/test_const.go` instead.
      - This results in changes to the following non-refresh credentials responder test files:
         - `agent/acs/handler/attach_eni_handler_common_test.go`
         - `agent/acs/handler/payload_handler_test.go`
         - `agent/acs/handler/task_manifest_handler_test.go`

### Testing

Refactor ACS refresh credentials message handling

### Licensing


agent/acs/handler/payload_handler.go
```diff
@@ -187,14 +184,24 @@ func (payloadHandler *payloadRequestHandler) handleSingleMessage(payload *ecsacs
...
 
 	return nil
 }
 
+// makeCredentialsAckRequest sends an IAMRoleCredentialsAckRequest to the backend
+func (payloadHandler *payloadRequestHandler) makeCredentialsAckRequest(ack *ecsacs.IAMRoleCredentialsAckRequest) {
+	seelog.Debugf("ACKing credentials: %s", ack.String())
+	err := payloadHandler.acsClient.MakeRequest(ack)
+	if err != nil {
+		seelog.Warnf("Error ACKing credentials with credentialsID '%s', error: %v",
```
Realmonia, line 200: Q: what's the impact of failing to ACK credentials? Asking because I am not sure if warn is appropriate level.  (nvm I see `// Throw the ack in async; it doesn't really matter all that much and this is blocking handling more tasks.`)

I also see in ecs-agent/../refresh_credentials_responder we have very similar logging and both does ack. Can you explain the difference between these two? I might be missing something but we might be able to get rid of one of the two?

danehlim: Thanks for taking a look!

> I also see in ecs-agent/../refresh_credentials_responder we have very similar logging and both does ack. Can you explain the difference between these two? I might be missing something but we might be able to get rid of one of the two?

Payload responder receives payload messages which include tasks and their associated credentials. If the payload message is successfully processed, the credentials associated with the tasks in that message are ACKed.

Refresh credentials responder receives refresh credentials messages to refresh credentials that may be a risk of expiring. If the credentials are successfully refreshed, the credentials associated with the message is ACKed.

I think it would be better to have two different functions/logging for these 2 different cases/code paths mentioned above, that way in the logs we can easily see whether the ACK came from handling a payload message or handling a refresh credentials message. In a future PR to refactor ACS payload message handling, I plan to update the logging message there to make it explicitly clear that the ACKing of credentials being done there is associated with credentials from a payload message. If you have a strong opinion/preference, I can pull that into the scope of this pull request. What do you think?

Realmonia: This makes sense to me. One question I had is that if we are getting a refresh credentials message from ACS, would the message be handled by both of `payloadHandler` and `refreshCredentialsHandler`? If so, that would cause us to call `setTaskCredentials` (may not exactly the method name) twice. What would that cause us? If one of the Acks failed, what should we expect? If not, how did we assure it only gets to refreshCredentialsHandler?

danehlim: Synced up and discussed offline:
- Each ACS message has its own type, which each message responder is specialized to handle
- Refresh credentials messages (type IAMRoleCredentialsMessage) received from ACS would only be handled by the refresh credentials responder
- Payload responder handles payload messages (type PayloadMessage) which includes handling the credentials that are associated with task(s) contained in the payload message

We aligned that we should make an update in this PR to to make logging of ACKing credentials while handling  payload message explicitly distinct from ACKing credentials while handling refresh credentials message.

ecs-agent/acs/session/refresh_credentials_responder.go
```diff
@@ -0,0 +1,199 @@
...
+		Expiration:    message.RoleCredentials.Expiration,
+		MessageId:     message.MessageId,
+		CredentialsId: message.RoleCredentials.CredentialsId,
+	})
+	if err != nil {
+		logger.Warn(fmt.Sprintf("Error acknowledging %s", RefreshCredentialsMessageName), logger.Fields{
+			field.MessageID: messageID,
+			field.Error:     err,
+		})
+		err = errors.Wrapf(err, "unable to ACK task credentials for task with ARN %s", taskARN)
```
Realmonia, line 127: Can you update logging here too to something like `unable to ACK task credentials for RefreshCredentialsMessage for task ARN` etc.?

Realmonia: Or `unable to ACK refreshCredentialsMessage`?

Realmonia (dismissed):

ecs-agent/acs/session/refresh_credentials_responder.go
Realmonia: Chatted offline - log context should provides enough information.

danehlim: Further context - L123 displays the message type in the logging statement https://github.com/aws/amazon-ecs-agent/pull/3830/files#diff-54706336185a9c70284090b1f5e17ea6916c57c272e5de8fe3db7fe819e7a4cdR123

(i.e., this results in log text `Error acknowledging IAMRoleCredentialsMessage`)

ecs-agent/acs/session/refresh_credentials_responder_test.go
```diff
@@ -0,0 +1,246 @@
...
+			}
+			testRefreshCredentialsResponder := NewRefreshCredentialsResponder(credentialsManager,
+				mockCredsMetadataSetter,
+				mockMetricsFactory,
+				testResponseSender)
+
+			handleCredentialsMessage :=
+				testRefreshCredentialsResponder.HandlerFunc().(func(*ecsacs.IAMRoleCredentialsMessage))
+
+			go handleCredentialsMessage(&iamRoleCredentialsMessageCopy)
```
amogh09, line 198: Why do we want to call `handleCredentialsMessage` asynchronously?

agent/acs/handler/acs_handler.go
```diff
@@ -78,8 +79,8 @@ const (
 	acsProtocolVersion = 2
 	// numOfHandlersSendingAcks is the number of handlers that send acks back to ACS and that are not saved across
 	// sessions. We use this to send pending acks, before agent initiates a disconnect to ACS.
-	// they are: refreshCredentialsHandler, taskManifestHandler, and payloadHandler
-	numOfHandlersSendingAcks = 3
+	// they are: taskManifestHandler, and payloadHandler
+	numOfHandlersSendingAcks = 2
```
amogh09, line 83: Wow this variable's description sounds like a hack! 

Anyway, do you mind providing some more details in the PR description on why `refreshCredentialsHandler` is free from this logic now?

agent/acs/handler/refresh_credentials_responder_test.go
```diff
@@ -0,0 +1,275 @@
...
+		return nil
+	}
+	testRefreshCredentialsResponder := acssession.NewRefreshCredentialsResponder(credentials.NewManager(),
+		&credentialsMetadataSetter{
+			taskEngine: nil,
+		},
+		metrics.NewNopEntryFactory(),
+		testResponseSender)
+
+	handleCredentialsMessage := testRefreshCredentialsResponder.HandlerFunc().(func(*ecsacs.IAMRoleCredentialsMessage))
```
amogh09, line 96: Why is the casting needed?

ecs-agent/acs/session/refresh_credentials_responder.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	// Send ACK.
+	err = r.respond(&ecsacs.IAMRoleCredentialsAckRequest{
+		Expiration:    message.RoleCredentials.Expiration,
+		MessageId:     message.MessageId,
+		CredentialsId: message.RoleCredentials.CredentialsId,
+	})
+	if err != nil {
+		logger.Warn(fmt.Sprintf("Error acknowledging %s", RefreshCredentialsMessageName), logger.Fields{
```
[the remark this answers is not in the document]
danehlim: I don't have a strong preference here. Are you proposing that we replace instances of `RefreshCredentialsMessageName` with something more generic like "ACS message" and then add a new `field.RefreshCredentialsMessageName` field to the `logger.Fields` struct?

ecs-agent/acs/session/refresh_credentials_responder.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// validateIAMRoleCredentialsMessage performs validation checks on the
+// IAMRoleCredentialsMessage.
+func validateIAMRoleCredentialsMessage(message *ecsacs.IAMRoleCredentialsMessage) error {
+	if message == nil {
+		return errors.Errorf("Message is empty")
+	}
+
+	messageID := aws.StringValue(message.MessageId)
```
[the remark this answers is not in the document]
danehlim: I don't think so - doesn't the `aws.StringValue` function gracefully handles `nil` values?

ecs-agent/acs/session/refresh_credentials_responder_test.go
```diff
@@ -0,0 +1,246 @@
...
+
+	testRefreshCredentialsMessage.MessageId = nil
+	err := validateIAMRoleCredentialsMessage(testRefreshCredentialsMessage)
+	assert.Error(t, err, "Expected validation error validating a message with no message ID")
+
+	testRefreshCredentialsMessage.MessageId = aws.String("")
+	err = validateIAMRoleCredentialsMessage(testRefreshCredentialsMessage)
+	assert.Error(t, err, "Expected validation error validating a message with empty message ID")
+
+	testRefreshCredentialsMessage.MessageId = tempMessageId
```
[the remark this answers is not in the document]
danehlim: Great suggestion! I will make this change.

ecs-agent/acs/session/refresh_credentials_responder_test.go
```diff
@@ -0,0 +1,246 @@
...
+	testRefreshCredentialsMessage.TaskArn = aws.String("")
+	err = validateIAMRoleCredentialsMessage(testRefreshCredentialsMessage)
+	assert.Error(t, err, "Expected validation error validating a message with empty task ARN")
+
+	testRefreshCredentialsMessage.TaskArn = tempTaskArn
+}
+
+// TestValidateRefreshMessageWithNoRoleCredentials tests if a validation error
+// is returned while validating a credentials message with no role credentials.
+func TestValidateRefreshMessageWithNoRoleCredentials(t *testing.T) {
```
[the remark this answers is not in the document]
danehlim: That's a good point. I will update these message validation tests into a single table driven test function.

ecs-agent/acs/session/refresh_credentials_responder_test.go
danehlim: I actually don't think calling asynchronously is necessary or adds any benefit here in this case - I'll remove this. Thanks for calling out!

agent/acs/handler/refresh_credentials_responder_test.go
danehlim: Here we do not have access to the private refresh credentials responder method `handleCredentialsMessage`. The `acssession.NewRefreshCredentialsResponder` function return type is `wsclient.RequestResponder`, so we have access to the returned value's `HandlerFunc()` method, who's return type is `wsclient.RequestHandler`. 

We do a type assertion here to assert that we expect the `wsclient.RequestHandler` returned by the responder's `HandlerFunc()` method to be a function that only accepts `IAMRoleCredentialsMessage`s. If this type assertion fails, something has gone wrong and we will be notified.

agent/acs/handler/acs_handler.go
danehlim: 😂 - good thing it's going away!

Sure, I've added some context on this to the PR description.

ecs-agent/acs/session/refresh_credentials_responder.go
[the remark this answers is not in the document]
amogh09: Yeah that's right. I don't have a strong preference either. We are using `%s` and logger fields at the same time so I thought why not make it a field. But maybe grepping for `Error acknowledging IAMRoleCredentialsMessage' is easier.

ecs-agent/acs/session/refresh_credentials_responder.go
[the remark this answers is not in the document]
amogh09: Oh okay I didn't know that. Looks good.

ecs-agent/acs/session/refresh_credentials_responder.go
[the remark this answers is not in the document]
amogh09: This is a non-blocking nit.

agent/acs/handler/refresh_credentials_responder_test.go
amogh09: Thanks. I didn't realize that `refreshCredentialsResponder` is private and its constructor returns a `wsclient.RequestResponder`. It's a bit odd to make the implementation private but then also use casting here so that we can call its handler function. I noticed that the type of the handler function is `type RequestHandler interface{}` which has no type information... 

Do we use such casting in prod usage of responders too?

ecs-agent/acs/session/refresh_credentials_responder_test.go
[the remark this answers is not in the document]
amogh09: Thanks

ecs-agent/acs/session/refresh_credentials_responder_test.go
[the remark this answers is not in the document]
amogh09: Thanks

ecs-agent/acs/session/refresh_credentials_responder_test.go
amogh09: Thanks

agent/acs/handler/refresh_credentials_responder_test.go
amogh09: Discussed offline with @danehlim. We found out that in Prod we use reflection to perform type checks on handler functions provided by ACS responders. So, I am okay with us using type assertions in this test. 

In general I think reflection is a code smell and we should be able to use generics to make the design statically typed, however, that is out of scope of what we are trying to accomplish with this PR. 

https://github.com/aws/amazon-ecs-agent/blob/6ba7f1ea301df8179d8bdf96368332a92a442375/ecs-agent/wsclient/client.go#L373C1-L384

amogh09 (approved):

Realmonia (approved):

PR #3840, opened 2023-08-04
Refactor ACS payload message handling

danehlim:
### Summary

Refactor ACS payload message handling, update other associated code, and improve/update tests.

### Implementation details

1. Refactor ACS payload message handling
   - `agent/acs/handler/payload_handler.go` -> `ecs-agent/acs/session/payload_responder.go` (partial)
      - Components of the responder that do not interact with Agent state belong in this file
      - Refactor `payloadRequestHandler` struct and existing code into `payloadResponder` struct, such that `payloadResponder` struct implements the `wsclient.RequestResponder` interface
      - Define and use new `validatePayloadMessage`function to do some ACS payload message validation upfront
      - Handle the payload message using the `ProcessMessage` method of the responder's `PayloadMessageHandler`
      - Use `errors.Errorf` in favor of `fmt.Errorf`
      - Rename method `handleUnrecognizedTask` to more suitable name `handleInvalidTask`
   - `agent/acs/handler/payload_handler.go` -> `agent/acs/handler/payload_responder.go` (partial)
      - Components of the responder that interact with Agent state belong in this file
      - Refactor functionality from such components mentioned above into `payloadMessageHandler` struct such that `payloadMessageHandler` struct implements the `PayloadMessageHandler` interface defined in `ecs-agent/acs/session/payload_responder.go`
      - Initialize payload responder and add its `HandlerFunc` function as a request handler to the ACS client, such that payload responder dictates what to do in response to receiving payload messages from ACS
      - Remove payload messages from pending ACKs logic and `payloadMessageBufferSize` constant since they are no longer applicable/used after the refactor of ACS payload message handling
         - Such logic is no longer applicable since there is no need to do multi-thread synchronization with channels with this responder (like we currently do). As part of the refactor, we factor out this non-necessity and thus methods such as `start()`, `stop()`, `clearAcks()`, etc. can be removed. The resulting responder is a lot cleaner and simpler to follow.
   - `ecs-agent/acs/session/generate_mocks.go` and `ecs-agent/acs/session/mocks/session_mock.go`
      - Generate new mocks for the `PayloadMessageHandler` interface to be used in testing
   - `ecs-agent/acs/logger/field/constants.go`
      - Create and use logger field for logging credentials ID
      - Create and use logger field for logging task version
3. Improve/update tests, most notably:
   - Update existing tests accordingly based off of above changes to refactor ACS payload message handling in the following files:
      - `agent/acs/handler/payload_handler_test.go` -> `ecs-agent/acs/session/payload_responder.go` (partial)
         - Payload responder tests that only need to interact with stateless components (e.g., message validation) belong in this file
      - `agent/acs/handler/payload_handler_test.go` -> `agent/acs/handler/payload_responder_test.go` (partial)
         - Payload responder tests that need to directly interact with Agent state belong in this file
      - `agent/acs/handler/acs_handler_test.go`
   -  Update existing tests in `agent/acs/handler/payload_handler_test.go` accordingly based off of above changes to `agent/acs/handler/payload_handler.go`
   - General cleanup/quality of life updates to tests (e.g., use table testing when possible, eliminate redundant initialization of test payloadMessage variables for each payload responder test that requires one)
   - Use `testconst.MessageID` test constant defined in `ecs-agent/acs/session/testconst/test_const.go` in favor of `payloadMessageId` test constant currently defined in `agent/acs/handler/payload_handler_test.go`

### Testing

Refactor ACS payload message handling

### Licensing


agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+
+	// Construct a slice with credentials acks from all tasks.
+	credentialsAcks := append(stoppedTasksCredentialsAcks, newTasksCredentialsAcks...)
+	return credentialsAcks, allTasksOK
+}
+
+// handleUnrecognizedTask handles unrecognized tasks by sending 'stopped' with
+// a suitable reason to the backend.
+func (tmHandler *taskMessageHandler) handleUnrecognizedTask(task *ecsacs.Task, err error,
```
RichaGangwar, line 184: The function name handleUnrecognizedTask seems a bit of a misnomer. As I understand from the function above, this is being called in a lot more cases than just unrecognized task payload. Maybe a better name would be "handleInvalidTaskPayload" or "handleErrorInTaskPayload". Just a suggestion. :)

agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+	}
+	return credentialsAcks, allTasksOK
+}
+
+func (tmHandler *taskMessageHandler) ackCredentials(messageID *string, credentialsID string) (
+	*ecsacs.IAMRoleCredentialsAckRequest, error) {
+	creds, ok := tmHandler.credentialsManager.GetTaskCredentials(credentialsID)
+	if !ok {
+		return nil, fmt.Errorf("credentials could not be retrieved")
```
RichaGangwar, line 271: Nit:Should these errors not be complete sentences?

agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+			}
+		}
+
+		ackCredentials := func(id string, description string) {
+			ack, err := tmHandler.ackCredentials(payload.MessageId, id)
+			if err != nil {
+				allTasksOK = false
+				logger.Error(fmt.Sprintf("Failed to acknowledge %s credentials for task",
+					description), logger.Fields{
+					"task":            task.String(),
```
RichaGangwar, line 244: You can use the logger field: TaskID instead of "task".

agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+		apiTask, err := apitask.TaskFromACS(task, payload)
+		if err != nil {
+			tmHandler.handleUnrecognizedTask(task, err, payload)
+			allTasksOK = false
+			continue
+		}
+
+		logger.Info("Received task payload from ACS", logger.Fields{
+			loggerfield.TaskARN:       apiTask.Arn,
+			"version":                 apiTask.Version,
```
RichaGangwar, line 100: Nit: Maybe worth creating a logger field as "version" seems to be a common field that might be being used across the project.

ecs-agent/acs/session/payload_responder.go
```diff
@@ -0,0 +1,162 @@
...
+
+// sendAck handles the sending of an individual specific ACK, assuming it is of type
+// ecsacs.IAMRoleCredentialsAckRequest or ecsacs.AckRequest.
+func (r *payloadResponder) sendAck(ackRequest interface{}) {
+	var credentialsAck *ecsacs.IAMRoleCredentialsAckRequest
+	var payloadMessageAck *ecsacs.AckRequest
+	credentialsAck, ok := ackRequest.(*ecsacs.IAMRoleCredentialsAckRequest)
+	if ok {
+		logger.Debug(fmt.Sprintf("ACKing credentials associated with %s", PayloadMessageName), logger.Fields{
+			"credentialsID": aws.StringValue(credentialsAck.CredentialsId),
```
RichaGangwar, line 98: Nit: Maybe create a logger field as it is being used in multiple places.

ecs-agent/acs/session/payload_responder.go
```diff
@@ -0,0 +1,162 @@
...
+		if ok {
+			logger.Debug(fmt.Sprintf("ACKing %s", PayloadMessageName), logger.Fields{
+				field.MessageID: aws.StringValue(payloadMessageAck.MessageId),
+			})
+		} else {
+			logger.Error(fmt.Sprintf("Error sending acknowledgement: %s",
+				"ackRequest does not hold type ecsacs.IAMRoleCredentialsAckRequest or ecsacs.AckRequest"))
+			return
+		}
+	}
```
RichaGangwar, line 112: Not a big fan of if-else's. Is it possible to make this a switch statement?

agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+	credentialsAcks, allTasksHandled := tmHandler.addPayloadTasks(message)
+
+	// Update latestSeqNumberTaskManifest for it to get updated in state file.
+	if tmHandler.latestSeqNumberTaskManifest != nil && message.SeqNum != nil &&
+		*tmHandler.latestSeqNumberTaskManifest < *message.SeqNum {
+		*tmHandler.latestSeqNumberTaskManifest = *message.SeqNum
+	}
+
+	if !allTasksHandled {
+		return fmt.Errorf("did not handle all tasks")
```
RichaGangwar, line 60: Maybe naive but why are we not using errors.Errorf() here?

ecs-agent/acs/session/payload_responder.go
```diff
@@ -0,0 +1,162 @@
...
+			field.MessageID: aws.StringValue(message.MessageId),
+			field.Error:     err,
+		})
+	}
+}
+
+// ackFunc sends ACKs of the payload message and of the credentials associated with the tasks contained in the payload
+// message.
+func (r *payloadResponder) ackFunc(payloadAck *ecsacs.AckRequest, credsAcks []*ecsacs.IAMRoleCredentialsAckRequest) {
+	go r.sendAck(payloadAck)
```
RichaGangwar, line 83: Since, we are calling these asynchronously, should we not call them with cancellable context?

agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+type skipAddTaskComparatorFunc func(apitaskstatus.TaskStatus) bool
+
+// taskMessageHandler implements TaskMessageHandler.
+type taskMessageHandler struct {
+	taskEngine                  engine.TaskEngine
+	ecsClient                   api.ECSClient
+	dataClient                  data.Client
+	taskHandler                 *eventhandler.TaskHandler
+	credentialsManager          credentials.Manager
+	latestSeqNumberTaskManifest *int64
```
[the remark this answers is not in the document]
danehlim: Yes, we need to pass by pointer here. If we were to pass by value instead, then we would not be able to change the latestSeqNumberTaskManifest tracked by the Agent, we would only be able to make changes to a copy of it.

agent/acs/handler/payload_responder.go
danehlim: In Golang it is idiomatic for error strings not to be capitalized and not end with punctuation. I believe this is a convention routinely followed in our code base.

See here for reference: https://google.github.io/styleguide/go/decisions.html#error-strings

agent/acs/handler/payload_responder.go
danehlim: I believe we need should use "task" here since `task.String()` doesn't return just a task ID, but it returns a string representation of the task (which includes the task ARN).

agent/acs/handler/payload_responder.go
danehlim: Good point! I will be sure to add this in the next revision if any blocking changes need to be made.

ecs-agent/acs/session/payload_responder.go
danehlim: ACK, I'll add this in the next revision if any blocking changes need to be made. Thanks!

agent/acs/handler/payload_responder.go
danehlim: I don't believe there's any particular reason `fmt.Errorf()` is used here over `errors.Errorf()` other than that `fmt.Errorf()` is what is currently used in the existing code for this case. Both functions return an error.

If you have a strong opinion on this but it is non-blocking, I can potentially make future changes to use `errors.Errorf()` in favor of `fmt.Errorf()` wherever we are using it but I feel that should be left out of scope of this specific PR. 

Let me know what you think, thanks!

ecs-agent/acs/session/payload_responder.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/pkg/errors"
+)
+
+const (
+	PayloadMessageName = "PayloadMessage"
+)
+
+type TaskMessageHandler interface {
+	ProcessTaskMessage(message *ecsacs.PayloadMessage,
```
Realmonia, line 32: With changing the handler type name we still usages `ecsacs.PayloadMessage` as parameter makes it look a little bit inconsistent for me. If we are able to create a "taskMessage" interface and let `PayloadMessage` implement it will look nicer.

ecs-agent/acs/session/payload_responder.go
danehlim: I don't think calling with context is necessary here, we don't require a context when sending simple ACKs to ACS right now and I don't see any common reasons for needing to call with context applying here. Is there anything specific to you that seems like calling with a context would be required here?

ecs-agent/acs/session/payload_responder.go
Realmonia, line 1: The changes in this file give me an overall picture that payload message is one type of taskMessage this handler handles. Is this needed by a planned future change?

ecs-agent/acs/session/payload_responder.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/pkg/errors"
+)
+
+const (
+	PayloadMessageName = "PayloadMessage"
```
Realmonia, line 28: It seems this variable is only used in log messages. I supposes this is to avoid duplication when we have another task message we are going to handle, but those logging all reside in payloadResponder's methods, this does not provides potential to extend, which seems unnecessary to have this constant and format the string where it can be just a string. e.g. line 60.

ecs-agent/acs/session/payload_responder.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+	}
+
+	err := r.respond(ackRequest)
+
+	if err != nil {
+		if credentialsAck != nil {
+			logger.Warn(fmt.Sprintf("Error acknowledging credentials associated with %s",
+				PayloadMessageName), logger.Fields{
+				"credentialsID": aws.StringValue(credentialsAck.CredentialsId),
```
Realmonia, line 120: I am not sure if it helps much to log credentials ID since it generally does not mean anything and is a uuid. Is there any reason we decide to add this in the log?

ecs-agent/acs/session/payload_responder.go
```diff
@@ -0,0 +1,162 @@
...
+	go r.sendAck(payloadAck)
+
+	for _, credsAck := range credsAcks {
+		go r.sendAck(credsAck)
+	}
+}
+
+// sendAck handles the sending of an individual specific ACK, assuming it is of type
+// ecsacs.IAMRoleCredentialsAckRequest or ecsacs.AckRequest.
+func (r *payloadResponder) sendAck(ackRequest interface{}) {
```
Realmonia, line 92: I see (by looking up the code) that `ecsacs.IAMRoleCredentialsAckRequest` and `ecsacs.AckRequest` are completely different types (only `messageID` in common), but it can be intuitive to think the former one is a "child type" of the later one. Can you add a note here that they are two different things, so the two different type cast we did here makes more sense? (it explains why we need a `interface{}` here)

agent/acs/handler/payload_responder.go
danehlim: I believe the function is called whenever performing a specific action on a task associated with an ACS payload message errors out. So I think the function name assumes such tasks are considered "unrecognized". I agree that "invalid" is probably a better word choice here. If it works with you, I can update the function name accordingly in the next revision if any blocking changes need to be made. 👍

agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+func (tmHandler *taskMessageHandler) ProcessTaskMessage(message *ecsacs.PayloadMessage,
+	ackFunc func(*ecsacs.AckRequest, []*ecsacs.IAMRoleCredentialsAckRequest)) error {
+
+	credentialsAcks, allTasksHandled := tmHandler.addPayloadTasks(message)
+
+	// Update latestSeqNumberTaskManifest for it to get updated in state file.
```
Realmonia, line 53: Not related to PR itself but for my knowledge - how does the acs handler decide the order of handling messages? Would it be possible that it handles a message with larger seqNum first and then get back to lower ones? When this happens and we do need to recover from state file, would it (and how would it) affect the message handling?

agent/acs/handler/payload_responder.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/aws/amazon-ecs-agent/agent/eventhandler"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
+	apieni "github.com/aws/amazon-ecs-agent/ecs-agent/api/eni"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+	loggerfield "github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+// skipAddTaskComparatorFunc defines the function pointer that accepts task status
```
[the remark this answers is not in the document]
danehlim: Apologies, I'm not following the question. 

` skipAddTaskComparatorFunc` is just a function type that describes the set of all functions that take in a `apitaskstatus.TaskStatus` parameter and returns a `bool`.

Examples of functions that satisfy this function type are `isTaskStatusStopped` and `isTaskStatusNotStopped`.

We make use of such functions to determine whether or not we add skip adding a task to the task engine during an execution of the `addTasks` method.

Does that make sense? 
If anything is still unclear and it's more convenient for you, feel free to reach out offline and we can sync up on this :)

ecs-agent/acs/session/payload_responder.go
Realmonia: I suppose it is related to retry mechanism. When acs has connection issue would it be possible to stay forever to retry causing resource leak? 

Also having it will help with agent exit case (to cancel all pending calls). This is a suggestion from Anirudh during TACS design review
> Can we add it to our developer guidelines doc (if it's not there already) that we should never create go routines without a cancelable context and that folks should only create a background context in the main method?

Realmonia: I think if the call itself is short-living (and/or no retries are needed and implemented) it is not a big deal but I think we would still want it if possible.

ecs-agent/acs/session/payload_responder.go
danehlim: I agree that we should use switch over if-else if there is no other advantage to using one over the other.

In this case, my thinking is that if we use a switch statement, I believe we'll be required to potentially do an unnecessary type assertion upfront (i.e., we would need to execute both 
`credentialsAck, ok := ackRequest.(*ecsacs.IAMRoleCredentialsAckRequest)` and `payloadMessageAck, ok = ackRequest.(*ecsacs.AckRequest)`
before calling the switch statement. Whereas with this if else, `payloadMessageAck, ok = ackRequest.(*ecsacs.AckRequest)` is only executed if necessary.

Perhaps I may be missing some other way of approaching this using a switch statement that doesn't require an unnecessary type assertion. Is there any that comes to mind for you?

ecs-agent/acs/session/payload_responder.go
danehlim: I see, thanks for this comment. This is not needed by a planned future change. Payload messages contain task(s) and information relevant to them so "task message" in this case is meant to mean the same thing as "payload message". Since this can cause confusion, I will update the interface name to `PayloadMessageHandler` and its method name to `ProcessMessage`.

ecs-agent/acs/session/payload_responder.go
danehlim: See this comment: https://github.com/aws/amazon-ecs-agent/pull/3840#discussion_r1288885624

agent/acs/handler/payload_responder.go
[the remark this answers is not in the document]
RichaGangwar: Got it. Thank you for clarifying :)

agent/acs/handler/payload_responder.go
RichaGangwar: Sure. Sounds good. It might be good to track as part of TODOs in the PR for future reference.

agent/acs/handler/payload_responder.go
RichaGangwar: Agreed.TaskID might get a bit confusing, even though it actually corresponds to "task" in constants file. Thanks.

agent/acs/handler/payload_responder.go
RichaGangwar: Sounds good. Maybe add it in TODOs for the PR for the future reference. Thanks.

ecs-agent/acs/session/payload_responder.go
RichaGangwar: Sounds good. Maybe add it in TODOs for the PR for the future reference. Thanks.

ecs-agent/acs/session/payload_responder.go
danehlim: Credentials ID are currently already being logged in the existing code for this case, so I decided to preserve this: https://github.com/aws/amazon-ecs-agent/blob/dev/agent/acs/handler/payload_handler.go#L200-L201

The payload message [contains a list of tasks](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/acs/model/ecsacs/api.go#L1389) which each can contain unique credentials (that have a unique credentials ID). I think logging credentials ID could be useful here in the event a payload message contains multiple tasks which require different credentials, and that ACKing of some credentials fails but some are successful. That way, we can differentiate which credentials were successfully ACKed and which were not (only logging the message ID would make it unclear which credentials in such a payload message were not able to be ACKed).

agent/acs/handler/payload_responder.go
RichaGangwar: Sure sounds good to me. Maybe add it in TODOs for the PR for the future reference. Thanks.

ecs-agent/acs/session/payload_responder.go
danehlim: I'll add a note here explaining that `ecsacs.IAMRoleCredentialsAckRequest` and `ecsacs.AckRequest` are two different things to provide more context about the usage of type assertion. Thanks!

ecs-agent/acs/session/payload_responder.go
RichaGangwar: Hmm thats right. I am not able to come up with another way to handle it. We can leave it as it is then. Thanks.

ecs-agent/acs/session/payload_responder.go
danehlim: Currently there aren't any plans to handle other types of ACS messages in the payload responder (see this comment: https://github.com/aws/amazon-ecs-agent/pull/3840#discussion_r1288885624).

However, my thought is that in the event that ACS/we ever want to change the name of the [existing `PayloadMessage` struct](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/acs/model/ecsacs/api.go#L1376), we would only need to make that change on L28 of this file using this const. But if we get rid of this const and just type the raw string "PayloadMessage" inline wherever we need it, then currently we would need to make changes in many places such as L60, L64, L73, L97, L104, L119, L125 (and the number of lines could grow to be more in the future if we need to add more log statements that require the message name).

ecs-agent/acs/session/payload_responder.go
danehlim: > Also having it will help with agent exit case (to cancel all pending calls). This is a suggestion from Anirudh during TACS design review

> I think if the call itself is short-living (and/or no retries are needed and implemented) it is not a big deal but I think we would still want it if possible.

Thanks for sharing this guidance, I was not aware of this. I think there may be a larger issue as there are many other places in ecs-agent module right now where we create goroutines without a cancelable context, some of which are NOT short-lived and/or implement retries.

We currently do not pass in the main method's context to ACS message responders so it would be difficult/require a large refactor to achieve this.

> When acs has connection issue would it be possible to stay forever to retry causing resource leak

Going through the code trace, I don't think it's possible for such calls to stay forever. They are short-lived and no retries are implemented.

Given the above reasons, I think we can call this out of scope of this PR - but I will take an action item to follow up with Anirudh and/or the team to evaluate where we currently create goroutines without a cancelable context, especially goroutines that are NOT short-lived and/or implement retries (unlike in this case) and discuss what actions (if any) should be done about them. Does this sound good to you?

agent/acs/handler/payload_responder.go
danehlim: Although unrelated to the PR, these are good questions! I'll take a deeper look into this and get back to you.

danehlim: > how does the acs handler decide the order of handling messages? 

Session/ACS handler consumes messages in the order that they are received.

> Would it be possible that it handles a message with larger seqNum first and then get back to lower ones? When this happens and we do need to recover from state file, would it (and how would it) affect the message handling?

Agent keeps track of the highest sequence number received in both messages of type`PayloadMessage` and `TaskManifestMessage`. If the sequence number on a message with type `TaskManifestMessage` is lower than the latest known one, the the task manifest message is considered stale and discarded.

ACS guarantees that the sequence number is increasing and valid.

ecs-agent/acs/session/payload_responder.go
Realmonia: Sounds fine to me to call it out of scope

ecs-agent/acs/session/payload_responder.go
Realmonia: Thanks this makes sense to me

ecs-agent/acs/session/payload_responder.go
Realmonia: I do not have a strong opinion for this one. You can do whichever you prefer.

ecs-agent/acs/session/payload_responder.go
Realmonia: ack.

ecs-agent/acs/session/payload_responder.go
Realmonia: thanks!

agent/acs/handler/payload_responder.go
Realmonia: Thanks for the research!

Realmonia (approved):

RichaGangwar (approved):
Looks good to me. :)

PR #3844, opened 2023-08-09
Add task manifest responder to ecs-agent module

timj-hh:
### Summary

Implement the ACS task manifest responder into the ecs-agent module. This will compare a task's manifest against what is running on the instance and send ack/stop messages to tasks that should not be running. 

### Implementation details
Similarly to responders in `acs/session`, the task manifest responder uses `wsclient` to receive and send messages concerning ACS. An underlying respond function will be passed to this component on initialization that decides how it responds to `TaskManifestMessage`s

New `TaskComparer` interface will implement a `CompareRunningTasksOnInstanceWithManifest` method that is expected to compare task manifest to those running on the instance. An underlying struct that holds a data client is expected to invoke this method. It is also recommended to make this struct publicly accessible by a `NewTaskComparer` method.

`ContainerInstanceAccessor` has been split into a `SequenceNumberAccessor` and `ManifestMessageIDAccessor` to adhere to the single responsibility principle. The former will address the state of the most current task manifest. The latter will hold the latest message ID corresponding to the most recently processed task manifest message.

`processTaskManifestMessage` has been split into the top level `handleTaskManifestMessage` which invokes the former method. That in turn invokes the new `ackTaskManifestMessage` and `sendTaskManifestMessageAck` which ack and send a `TaskManifestMessage` to via `wsclient`

`cluster` has been removed as a field of `taskManifestResponder`. ACS session will provide this data when queried, so we can rely on `TaskManifestMessage` for this data. It may also be passed in an implementer's version of `CompareRunningTasksOnInstanceWithManifest` through a struct implementing `TaskComparer`.

logging has mostly been moved to use `ecs-agent/logger` whenever possible. This only slightly affects how log messages will look to implementers.

in `sendTaskStopVerification` we are now opting to log once for each task we are stopping, keeping in mind the message to ACS will only be sent out once.

we will now opt to send manifest message ack and task stop verification ack asynchronously through goroutines and waitgroups to closer match original agent behavior.

### Testing

`make test`  runs all tests successfully.

`go mod tidy && go mod vendor` to update dependencies

`TestManifestAckHappyPath` was added to validate the manifest ack happy path is reached with a typical TaskManifestMessage, as well as a TaskStopVerificationMessage when there is a mismatch in the manifest. Mocks have been generated for `TaskComparer` and `SequenceNumberAccessor` to test their functionality in `TestTaskManifestStaleMessage`.

Add task manifest responder to ecs-agent module

### Licensing


amogh09:
Thanks @timj-hh! Did you run `go mod vendor` in agent module to sync these changes with agent module's vendor?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/metrics"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+)
+
+type TaskManifestResponder interface {
```
amogh09, line 27: I don't see a responder interface for other responders such as RefreshCredentialsResponder and AttachResourceResponder. Is there a reason we need this interface for TaskManifestResponder?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,177 @@
...
+	New(taskComparer TaskComparer, messageIDAccessor ManifestMessageIDAccessor,
+		snAccessor SequenceNumberAccessor) wsclient.RequestResponder
+}
+
+type TaskComparer interface {
+	CompareRunningTasksOnInstanceWithManifest(*taskManifestResponder, *ecsacs.TaskManifestMessage) ([]*ecsacs.TaskIdentifier, error)
```
amogh09, line 33: Can we please add a description for this method for readers? Also this method takes a parameter of type `*taskManifestResponder` which is an unexported type. How would a struct implementing this interface outside this package get access to the unexported type?

ecs-agent/acs/session/task_manifest_common.go
```diff
@@ -13,10 +13,21 @@
...
 	SetMessageID(messageID string) error
 }
+
+// SequenceNumberAccessor is used to get and set state for the current container
+// instance. It sets and gets the latest sequence number that corresponds to the most
+// recent task manifest that the task manifest responder has processed it alo sets the
+// latest message id
+type SequenceNumberAccessor interface {
+	GetLatestSequenceNumber() int64
+	SetLatestSequenceNumber(message *ecsacs.TaskManifestMessage) error
```
amogh09, line 32: I don't think sequence numbers are related specifically to `TaskManifestMessage` payload? How about using a more general parameter for this method so that it is more reusable?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,177 @@
...
+
+// NewTaskManifestResponder returns an instance of the taskManifestResponder struct.
+func NewTaskManifestResponder(taskComparer TaskComparer, snAccessor SequenceNumberAccessor, messageIDAccessor ManifestMessageIDAccessor,
+	cluster string, metricsFactory metrics.EntryFactory, responseSender wsclient.RespondFunc) *taskManifestResponder {
+	r := &taskManifestResponder{
+		taskComparer:      taskComparer,
+		snAccessor:        snAccessor,
+		messageIDAccessor: messageIDAccessor,
+		cluster:           cluster,
+		logger:            logrus.WithField("ACSRequestResponder", "TaskManifestResponder"),
```
amogh09, line 63: Other responder types seem to be using logger provided by `github.com/aws/amazon-ecs-agent/ecs-agent/logger` package. Any reason we are using `logrus` here?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,177 @@
...
+func (tmr *taskManifestResponder) HandlerFunc() wsclient.RequestHandler {
+	return tmr.handleTaskManifestMessage
+}
+
+// handleTaskManifestMessage is the high level caller to handle a task manifest message from ACS.
+// It will kick off the call stack of processTaskManifestMessage calling ack and send TaskManifestMessage
+func (tmr *taskManifestResponder) handleTaskManifestMessage(message *ecsacs.TaskManifestMessage) error {
+	tmr.logger.WithField("MessageId", aws.StringValue(message.MessageId)).Info("Processing task manifest message")
+	err := tmr.processTaskManifestMessage(message)
+	tmr.metricsFactory.New(metrics.TaskManifestHandlingDuration).Done(err)
```
amogh09, line 81: Why are we using `err` to publish a duration metric?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,177 @@
...
+func (tmr *taskManifestResponder) processTaskManifestMessage(
+	message *ecsacs.TaskManifestMessage) error {
+	manifestSeqNum := aws.Int64Value(message.Timeline)
+	agentLatestSeqNum := tmr.snAccessor.GetLatestSequenceNumber()
+
+	// Verify that task manifest isn't stale.
+	// The agent will keep track of the highest sequence number received in both PayloadMessage and TaskManifestMessage.
+	// If the sequence number on TaskManifestMessage is lower to the latest known one, the TaskManifestMessage must be discarded.
+	// ACS will guarantee that the sequence number is indeed increasing and valid for this use case.
+	if manifestSeqNum < agentLatestSeqNum {
```
amogh09, line 97: Is this not a concern for other types of payloads from ACS? I am curious why this check doesn't exist for other responders.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,177 @@
...
+		Cluster:           aws.String(tmr.cluster),
+		ContainerInstance: message.ContainerInstanceArn,
+		MessageId:         aws.String(messageID),
+	})
+	if err != nil {
+		tmr.logger.WithError(err).WithField("Error Acking TaskManifestMessage with messageID: %s", messageID)
+	}
+}
+
+func (tmr *taskManifestResponder) sendTaskManifestMessageAck(message *ecsacs.TaskManifestMessage, tasksToStop []*ecsacs.TaskIdentifier) {
```
amogh09, line 154: This method doesn't send an Ack but a TaskStopVerification, so how about renaming it to `sendTaskStopVerification` or something like that?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	// Get list of tasks on instance.
+	tasksToMessage, err := tmr.taskComparer.CompareRunningTasksOnInstanceWithManifest(tmr, message)
+	if err != nil {
+		return err
+	}
+
+	// Acknowledge the task manifest message asynchronously.
+	var wg sync.WaitGroup
```
amogh09, line 124: I am curious what we gain by this asynchronicity. IIUC there is only a single TCP connection between Agent and ACS at any time, so the two messages will be sent synchronously anyway?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,131 @@
...
+	fakeCI fakeContainerInstance
+}
+
+// fakeContainerInstance imitates the ContainerInstance.
+type fakeContainerInstance struct {
+	LatestTaskManifestSequenceNumber int64
+	ManifestMessageID                string
+}
+
+type fakeTaskComparer struct {
```
amogh09, line 41: How about using gomock to generate the mock implementation? An implementation generated by gomock will make sure the methods are called the expected number of times with the expected arguments.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+// setupTestTaskManifestResponder initializes the TaskManifestResponder for testing.
+func setupTestTaskManifestResponder(t *testing.T, f wsclient.RespondFunc) *taskManifestResponder {
+	// Create task manifest responder.
+	cluster := "test-cluster"
+	ftc := fakeTaskComparer{}
+	cid := fakeContainerInstanceDirector{}
+	cid.fakeCI.LatestTaskManifestSequenceNumber = 15
+	cid.fakeCI.ManifestMessageID = ""
+	testTaskManifestResponder := NewTaskManifestResponder(ftc, &cid, &cid, cluster, metrics.NewNopEntryFactory(), f)
```
amogh09, line 75: How about using a mock implementation of metrics factory? That way we can assert on the interactions with the metrics factory.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,131 @@
...
+		} else {
+			stopVerification, isStopVerification := v.(*ecsacs.TaskStopVerificationMessage)
+			if isStopVerification {
+				*stopVerifSent = true
+				stopVerifChannel <- stopVerification
+			}
+		}
+		return nil
+	}
+	return f
```
amogh09, line 95: nit: variable `f` is redundant.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// TestManifestAckHappyPath tests the happy path for a typical TaskManifestMessage and confirms
+// expected ack is made
+func TestManifestAckHappyPath(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
amogh09, line 102: `ctrl` is unused?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,131 @@
...
+		ContainerInstanceArn: aws.String("test-instance-arn"),
+		MessageId:            aws.String("test-messageID"),
+		Tasks: []*ecsacs.TaskIdentifier{
+			{DesiredStatus: aws.String("STOPPED"), TaskArn: aws.String("manifest-task-arn")},
+		},
+		Timeline: aws.Int64(20),
+	}
+
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage) error)
+	go handleManifestMessage(testManifest)
```
amogh09, line 126: Why do we need to call the handler function asynchronously?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,131 @@
...
+	ftc := fakeTaskComparer{}
+	cid := fakeContainerInstanceDirector{}
+	cid.fakeCI.LatestTaskManifestSequenceNumber = 15
+	cid.fakeCI.ManifestMessageID = ""
+	testTaskManifestResponder := NewTaskManifestResponder(ftc, &cid, &cid, cluster, metrics.NewNopEntryFactory(), f)
+	return testTaskManifestResponder
+}
+
+// setupTestRespondFunc initializes the respond function used by the TestTaskManifestResponder.
+func setupTestRespondFunc(ackSent *bool, stopVerifSent *bool,
```
amogh09, line 80: Unused function?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,131 @@
...
+		},
+		Timeline: aws.Int64(20),
+	}
+
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage) error)
+	go handleManifestMessage(testManifest)
+
+	manifestMessageAckSent := <-ackSent
+	require.Equal(t, aws.StringValue(manifestMessageAckSent.MessageId), aws.StringValue(testManifest.MessageId))
+
```
amogh09, line 130: I think we need some more assertions (or in other smaller/focused tests). 

1. TaskStopVerificationMessage is sent or not
2. Contents of the two messages to ACS are correct
3. Sequence number and message ID are updated as expected
4. Metric is published

The latter two should be easy to achieve with mocks generated with gomock. 

Also we should probably add some more test cases to cover non-happy cases.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: A quick vscode search lead me to think the agent used logrus but it seems to be indirect, updating with logger soon 👍

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: Go channels introduce asynchronous behavior as I understand them, requiring this call to wait for `ackSent`. The test will hang indefinitely if the method is called synchronously.

ecs-agent/acs/session/task_manifest_common.go
timj-hh: Synced offline, `SequenceNumberAccessor` adheres to the single responsibility principle to handle only `TaskManifestMessage` types. This will be left as is.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: Synced offline, taskManifestResponder can be removed from this method's parameters as its only use was logging. `ecs-agent/logging` can be used to that end.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: Most other responders are concerned with acks, so we will usually just check that the ack is there and not concern ourselves with how old they are. Other responders should also be stateless, with the ACS session keeping track of how many/often responders ack messages.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: `err` in this context is the metric completion status. Good point about its placement though, new rev will surround `processTaskManifestMessage` with the metric creation and publishing.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: New revision tests 2-4, but #1 will be difficult as the verification is only sent if manifest does not match existing tasks. This can be tested by an implementer that has a data client/dummy data client with access to these instance tasks.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: The container instance director that is spoofed here is not in `ecs-agent`. I do not know what methods we can mock by using such an approach either

ecs-agent/acs/session/task_manifest_responder.go
amogh09: I still don't understand why the metric is named "TaskManifestResponder.Duration". What duration is referred to in the metric name?

ecs-agent/acs/session/task_manifest_responder_test.go
amogh09: `fakeTaskComparer` doesn't seem to have anything to do with container instance director? I don't see why we cannot use mockgen to generate a mock implementation of `TaskComparer` for use in testing. A benefit of a mock generated by mockgen is that we won't have to write the mock ourselves and we get assertions on interface interactions for free.

ecs-agent/acs/session/task_manifest_responder_test.go
amogh09: I see now that `testResponseSender` calls `defer ackWg.Done()` which is why you need to call `handleManifestMessage` asynchronously. Taking a step back, why do we need `ackWg` in the first place? We can just get rid of `ackWg` and call `handleManifestMessage` synchronously without affecting the test's semantics and reduce the code complexity?

ecs-agent/acs/session/task_manifest_responder_test.go
amogh09: Thanks @timj-hh. IIUC `fakeTaskComparer.CompareRunningTasksOnInstanceWithManifest` method's return value decides whether TaskStopVerificationMessage is sent or not. Since we have full control over `fakeTaskComparer`, I think we should be able to write a test case for which TaskStopVerificationMessage is sent?

ecs-agent/acs/session/task_manifest_common.go
```diff
@@ -13,10 +13,21 @@
...
 type ManifestMessageIDAccessor interface {
 	GetMessageID() string
 	SetMessageID(messageID string) error
 }
+
+// SequenceNumberAccessor is used to get and set state for the current container
+// instance. It sets and gets the latest sequence number that corresponds to the most
+// recent task manifest that the task manifest responder has processed it alo sets the
+// latest message id
+type SequenceNumberAccessor interface {
```
danehlim, line 30: Could we move this interface to `ecs-agent/acs/session/task_manifest_responder.go` adjacent to where the `TaskComparer` interface is defined right now? My current understanding is that this interface is only needed for task manifest responder and is not common to both the task manifest responder and task stop verification ACK responder.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+	}()
+	wg.Wait()
+	return nil
+}
+
+func (tmr *taskManifestResponder) ackTaskManifestMessage(message *ecsacs.TaskManifestMessage) {
+	messageID := aws.StringValue(message.MessageId)
+	logger.Info(fmt.Sprintf("Acking %s id: %s", TaskManifestMessageName, messageID))
+	err := tmr.respond(&ecsacs.AckRequest{
+		Cluster:           aws.String(tmr.cluster),
```
danehlim, line 143: We can just use `message.ClusterArn` here, this is what is currently being done for other responders.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+// taskManifestResponder responds to task manifest messages from ACS. It processes the
+// message and, if the message isn't stale, creates a list of stop candidates, which are
+// tasks that are running on the instance but are tasks that ACS says should not be running.
+// The taskManifestResponder will ack the task manifest message and send ACS a
+// TaskStopVerificationMessage if there are stop candidates.
+type taskManifestResponder struct {
+	taskComparer      TaskComparer
+	snAccessor        SequenceNumberAccessor
+	messageIDAccessor ManifestMessageIDAccessor
+	cluster           string
```
danehlim, line 48: I don't think cluster is a necessary struct field here. See [comment on L143](https://github.com/aws/amazon-ecs-agent/pull/3844/files#r1291752246) for more details.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,186 @@
...
+const (
+	Cluster                                  = "test-cluster"
+	ContainerInstanceARN                     = "test-instance-arn"
+	MessageID                                = "test-messageID"
+	ManifestARN                              = "manifest-task-arn"
+	StartingLatestTaskManifestSequenceNumber = 15
+)
+
+// fakeContainerInstanceDirector imitates the ContainerInstanceDirector.
+type fakeContainerInstanceDirector struct {
```
danehlim, line 41: ecs-agent module is unaware of/doesn't have any container instance director. Could we generate and use mock implementations of the `SequenceNumberAccessor` and `TaskComparer` interfaces instead for this test file?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/metrics"
+	mock_metrics "github.com/aws/amazon-ecs-agent/ecs-agent/metrics/mocks"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+)
+
+const (
```
danehlim, line 32: Can we use test constants already defined in `ecs-agent/acs/session/testconst/test_const.go` here when possible?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+func (tmr *taskManifestResponder) processTaskManifestMessage(
+	message *ecsacs.TaskManifestMessage) error {
+	manifestSeqNum := aws.Int64Value(message.Timeline)
+	agentLatestSeqNum := tmr.snAccessor.GetLatestSequenceNumber()
+
+	// Verify that task manifest isn't stale.
+	// The agent will keep track of the highest sequence number received in both PayloadMessage and TaskManifestMessage.
+	// If the sequence number on TaskManifestMessage is lower to the latest known one, the TaskManifestMessage must be discarded.
+	// ACS will guarantee that the sequence number is indeed increasing and valid for this use case.
+	if manifestSeqNum < agentLatestSeqNum {
```
danehlim, line 95: Should the check here be less than or equal to? If the two numbers are equal, that means Agent already knows about the sequence number in the message and we should be able to discard the message (and not do any extra unnecessary work) here.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+	err := tmr.messageIDAccessor.SetMessageID(aws.StringValue(message.MessageId))
+	if err != nil {
+		return errors.Wrap(err, "failed to update latest manifest messageID")
+	}
+	err = tmr.snAccessor.SetLatestSequenceNumber(message)
+	if err != nil {
+		return errors.Wrap(err, "failed to update latest manifest sequence number")
+	}
+
+	// Get list of tasks on instance.
```
danehlim, line 115: I think we should make it clear here that we are not getting a list of all tasks on the instance, but comparing the running tasks on the instance with the tasks in the manifest and only getting the tasks that should be stopped (i.e., the ones not present in the task manifest message task list).

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+	tasksToMessage, err := tmr.taskComparer.CompareRunningTasksOnInstanceWithManifest(message)
```
danehlim, line 116: Something like `tasksToStop` or `tasksToKill` would be a more suitable variable name here.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+		})
+	}
+}
+
+func (tmr *taskManifestResponder) sendTaskStopVerification(message *ecsacs.TaskManifestMessage, tasksToStop []*ecsacs.TaskIdentifier) {
+	messageID := aws.StringValue(message.MessageId)
+	if len(tasksToStop) == 0 {
+		return
+	}
+	for _, task := range tasksToStop {
```
danehlim, line 160: Why do we need to send this log message for every task in the task stop verification message? Why not just log once per single task stop verification message?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+			field.MessageID: messageID,
+			field.TaskARN:   task.TaskArn,
+		})
+	}
+	err := tmr.respond(&ecsacs.TaskStopVerificationMessage{
+		MessageId:      message.MessageId,
+		StopCandidates: tasksToStop,
+	})
+	if err != nil {
+		for _, task := range tasksToStop {
```
danehlim, line 171: Similar comment to [the one on L160](https://github.com/aws/amazon-ecs-agent/pull/3844/files#r1291780533):  why not just log once per single task stop verification message send failure?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+		defer wg.Done()
+		tmr.sendTaskStopVerification(message, tasksToMessage)
+	}()
+	wg.Wait()
+	return nil
+}
+
+func (tmr *taskManifestResponder) ackTaskManifestMessage(message *ecsacs.TaskManifestMessage) {
+	messageID := aws.StringValue(message.MessageId)
+	logger.Info(fmt.Sprintf("Acking %s id: %s", TaskManifestMessageName, messageID))
```
danehlim, line 141: nit: 
- Can we use logger.Fields for logging the message ID?
- Prefer to use "Acknowledging" instead of "Acking"

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+	err := tmr.respond(&ecsacs.AckRequest{
+		Cluster:           aws.String(tmr.cluster),
+		ContainerInstance: message.ContainerInstanceArn,
+		MessageId:         aws.String(messageID),
+	})
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error Acking %s", TaskManifestMessageName), logger.Fields{
```
danehlim, line 148: nit: 
- Prefer to use "acknowledging" instead of "Acking"

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+			field.TaskARN:   task.TaskArn,
+		})
+	}
+	err := tmr.respond(&ecsacs.TaskStopVerificationMessage{
+		MessageId:      message.MessageId,
+		StopCandidates: tasksToStop,
+	})
+	if err != nil {
+		for _, task := range tasksToStop {
+			logger.Error("Error sending task stop verification message", logger.Fields{
```
danehlim, line 172: nit: Prefer to log with level WARN here instead of ERROR (for consistency with other responders)

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	// Get list of tasks on instance.
+	tasksToMessage, err := tmr.taskComparer.CompareRunningTasksOnInstanceWithManifest(message)
+	if err != nil {
+		return err
+	}
+
+	// Acknowledge the task manifest message asynchronously.
+	var wg sync.WaitGroup
```
danehlim, line 122: Why do we need to wait for these function calls finish before returning? It seems like using WaitGroup here is unnecessary.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,179 @@
...
+
+func (*taskManifestResponder) Name() string { return "task manifest responder" }
+
+func (tmr *taskManifestResponder) HandlerFunc() wsclient.RequestHandler {
+	return tmr.handleTaskManifestMessage
+}
+
+// handleTaskManifestMessage is the high level caller to handle a task manifest message from ACS.
+// It will kick off the call stack of processTaskManifestMessage calling ack and send TaskManifestMessage
+func (tmr *taskManifestResponder) handleTaskManifestMessage(message *ecsacs.TaskManifestMessage) error {
```
danehlim, line 75: This function shouldn't return an error. For reference, see other responders and [this code comment](https://github.com/aws/amazon-ecs-agent/blob/0955e2636baa538ccbbce0b36c3258d8d14c2050/ecs-agent/wsclient/client.go#L362-L368):
```
// A request handler *must* be a function taking a single argument, and that
// argument *must* be a pointer to a recognized 'ecsacs' struct.
// E.g. if you desired to handle messages from acs of type 'FooMessage', you
// would pass the following handler in:
//
//	func(message *ecsacs.FooMessage)
```

danehlim, line 75: I'm thinking perhaps this function shouldn't return an error? For reference, see other responders and [this code comment](https://github.com/aws/amazon-ecs-agent/blob/0955e2636baa538ccbbce0b36c3258d8d14c2050/ecs-agent/wsclient/client.go#L362-L368):
```
// A request handler *must* be a function taking a single argument, and that
// argument *must* be a pointer to a recognized 'ecsacs' struct.
// E.g. if you desired to handle messages from acs of type 'FooMessage', you
// would pass the following handler in:
//
//	func(message *ecsacs.FooMessage)
```

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: I think it is helpful here to log what tasks are being stopped for the sake of debugging so anybody debugging has visibility into what tasks are mismatched from the manifest and instance. I agree for your next comment though, we can log once if there is an error.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: Digging into the code history on this one, it was implemented at the time of writing these responders as we want to force the `TaskStopVerificationAckResponder` to wait until we have sent both of these messages. See diagram [here](https://quip-amazon.com/DG3FAbfXYZid/Full-Task-Sync-for-Fargate-Agent#CfN9CAhSuuZ). Per Amogh's comment earlier I am unsure if it matters as we may be using a single TCP connection. Also unsure if acs will handle messages asynchronously, or wait for one message to be sent before another can be received.

I do not think it is asynchronous, so I do not think we need these goroutines here.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: Actually, I did notice that the test hangs even without using `ackWg`. I believe this is because we are using go channels which introduce concurrency. `ackWg` was implemented because there is a possible race condition I noticed where we try to access the sequence number to check it before it is set. However, with mocks it may not be necessary to use a waitgroup at all

ecs-agent/acs/session/task_manifest_responder.go
amogh09: Discussed offline. The metric name is a legacy.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,172 @@
...
+// whether a task manifest message is stale or not. If not, it proceeds to create a
+// list of stop candidates.
+func (tmr *taskManifestResponder) processTaskManifestMessage(
+	message *ecsacs.TaskManifestMessage) error {
+	manifestSeqNum := aws.Int64Value(message.Timeline)
+	agentLatestSeqNum := tmr.snAccessor.GetLatestSequenceNumber()
+
+	// Verify that task manifest isn't stale.
+	// The agent will keep track of the highest sequence number received in both PayloadMessage and TaskManifestMessage.
+	// If the sequence number on TaskManifestMessage is lower to the latest known one, the TaskManifestMessage must be discarded.
```
danehlim, line 98: This comment should be updated to indicate "less than or equal to".

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,172 @@
...
+	// ACS will guarantee that the sequence number is indeed increasing and valid for this use case.
+	if manifestSeqNum <= agentLatestSeqNum {
+		logger.Warn(fmt.Sprintf("Discarding task manifest message. "+
+			"Sequence number from manifest is less than Agent latest sequence number."), logger.Fields{
```
danehlim, line 102: This log statement needs to be updated to indicate "less than or equal to".

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,172 @@
...
+	tasksToStop, err := tmr.taskComparer.CompareRunningTasksOnInstanceWithManifest(message)
+	if err != nil {
+		return err
+	}
+
+	// Acknowledge the task manifest message.
+	tmr.ackTaskManifestMessage(message)
+
+	// Send the task stop verification message to ACS.
+	tmr.sendTaskStopVerification(message, tasksToStop)
```
danehlim, line 130: Is there any specific reason we changed these calls on L127 and L130 to be synchronous instead of the existing asynchronous behavior? Either way, I don't think it's a big deal since these calls are short lived, but just curious since they are currently set up as async calls such that they are not blocking.

ecs-agent/acs/session/task_manifest_responder.go
danehlim: > I think it is helpful here to log what tasks are being stopped for the sake of debugging so the developer has visibility into what tasks are mismatched from the manifest and instance.

That's fair - in this case could we update the log level to DEBUG then? And having a separate log statement for each task seems excessive and can potentially give off the incorrect idea that a task stop verification message is sent for each task that is stopped (e.g., if there are 5 tasks to be stopped, then someone could see 5 log statements saying "Sending task stop verification message" for the same task stop verification message, even though only one was sent in reality). Could we just log the tasks that are being stopped in a single log statement?

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,172 @@
...
+
+	return nil
+}
+
+func (tmr *taskManifestResponder) ackTaskManifestMessage(message *ecsacs.TaskManifestMessage) {
+	messageID := aws.StringValue(message.MessageId)
+	logger.Info(fmt.Sprintf("acknowledging %s", TaskManifestMessageName), logger.Fields{field.MessageID: messageID})
```
danehlim, line 137: nit: 
- Prefer to log with level DEBUG here instead of INFO (for consistency with other responders and current behavior in `agent/acs/handler/task_manifest_handler.go`)
- Capitalize "acknowledging" -> "Acknowledging"

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,172 @@
...
+	err := tmr.respond(&ecsacs.AckRequest{
+		Cluster:           message.ClusterArn,
+		ContainerInstance: message.ContainerInstanceArn,
+		MessageId:         message.MessageId,
+	})
+	if err != nil {
+		logger.Error(fmt.Sprintf("Error acknowledging %s", TaskManifestMessageName), logger.Fields{
```
danehlim, line 144: nit: Prefer to log with level WARN here instead of ERROR (for consistency with other responders)

ecs-agent/acs/session/task_manifest_responder.go
danehlim: Task stop verification ACK responder does not start running/processing anything until a task stop verification ACK has been sent back to us from ACS anyway (i.e., until we have sent a task stop verification message to ACS, they have received it and processed it successfully). From what I understand, waiting here in the task manifest responder does not force the `TaskStopVerificationAckResponder` to do anything and it is already enforced that task stop verification ACK responder does not process a task stop verification ACK before the corresponding task stop verification message is sent by Agent to ACS.

 > Also unsure if acs will handle messages asynchronously
Session/ACS handler consumes messages in the order that they are received (i.e., new goroutines are not spawned on a per message basis), but even if ACS handled messages asynchronously I believe the previous/above point still stands.

Goroutines themselves could still be beneficial here such that we do not block on these calls and allow to proceed with processing other ACS messages before they complete (i.e., current behavior). But I imagine the benefit is marginal here since these calls are short lived anyway. Similar comment on this/question about this here: https://github.com/aws/amazon-ecs-agent/pull/3844/files#r1293959305

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+func setupTestTaskManifestResponder(t *testing.T, factory metrics.EntryFactory, f wsclient.RespondFunc) *taskManifestResponder {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	// Create task manifest responder.
+	mockTC := mock_session.NewMockTaskComparer(ctrl)
+	mmIDA := mock_session.NewMockManifestMessageIDAccessor(ctrl)
+	mockSNA := mock_session.NewMockSequenceNumberAccessor(ctrl)
+
+	mmIDA.EXPECT().SetMessageID("").Times(1)
```
danehlim, line 62: nit: `.Times(1)` is unnecessary/redundant here and wherever else we're using it in this file

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+	mmIDA.SetMessageID("")
+
+	mockSNA.EXPECT().SetLatestSequenceNumber(gomock.Any()).MinTimes(1)
+	mockSNA.SetLatestSequenceNumber(setupTestManifestMessage())
+
+	testTaskManifestResponder := NewTaskManifestResponder(mockTC, mockSNA, mmIDA, factory, f)
+	return testTaskManifestResponder
```
danehlim, line 69: Can we update this function to return all the mock variables we might need in a test? That way we won't need to do type assertions such as on L90 for example.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+	// Create task manifest responder.
+	mockTC := mock_session.NewMockTaskComparer(ctrl)
+	mmIDA := mock_session.NewMockManifestMessageIDAccessor(ctrl)
+	mockSNA := mock_session.NewMockSequenceNumberAccessor(ctrl)
+
+	mmIDA.EXPECT().SetMessageID("").Times(1)
+	mmIDA.SetMessageID("")
+
+	mockSNA.EXPECT().SetLatestSequenceNumber(gomock.Any()).MinTimes(1)
+	mockSNA.SetLatestSequenceNumber(setupTestManifestMessage())
```
danehlim, line 66: Apologies, having trouble understanding. Why are L62-L66 necessary here?

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: Just synced with the original author of the code who is just back from vacation. There is code history for them to be asynchronous, and ACS may also hang on the ack call for a bit so we want to send both at the same time.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+		ackSent <- resp
+		return nil
+	}
+
+	testTaskManifestResponder := setupTestTaskManifestResponder(t, mock_metrics.NewMockEntryFactory(ctrl), testResponseSender)
+
+	// After we get the latest sequence number, we will expect to set the latest
+	// message ID
+	mockmmIDA := testTaskManifestResponder.messageIDAccessor.(*mock_session.MockManifestMessageIDAccessor)
+	mockmmIDA.EXPECT().SetMessageID(gomock.Any()).Times(1)
```
danehlim, line 91: Could we be more specific than `gomock.Any()` here (i.e., use `testconst.MessageID` instead of `gomock.Any()`)? This comment also applies to anywhere else we're using `gomock.Any()` in our tests.

danehlim, line 91: Should we also set the expectation that this call will return nil?

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: Seems other responders in this library do not capitalize acknowledging, shouldn't it be kept the same here? ack for the first nit

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+	testTaskManifestResponder := setupTestTaskManifestResponder(t, metrics.NewNopEntryFactory(), testResponseSender)
+
+	mockmmIDA := testTaskManifestResponder.messageIDAccessor.(*mock_session.MockManifestMessageIDAccessor)
+	mockmmIDA.EXPECT().SetMessageID(gomock.Any()).Times(1)
+
+	// Expect GetLatestSequenceNumber once in handleManifestMessage, and once
+	// when we validate it
+	mockSNA := testTaskManifestResponder.snAccessor.(*mock_session.MockSequenceNumberAccessor)
+	mockSNA.EXPECT().GetLatestSequenceNumber().Times(2)
+	mockSNA.EXPECT().SetLatestSequenceNumber(aws.Int64(20)).Times(1)
```
danehlim, line 144: nit: Can we create a const or variable with an appropriate name for `aws.Int64(20)` to know what it is in this context?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+		Timeline: aws.Int64(20),
+	}
+
+	// Handle the task manifest message update
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage))
+	handleManifestMessage(testManifest)
+
+	ackWg.Wait()
+
+	// Now assert that the latest sequence number has been updated
```
danehlim, line 166: Is there a missing assertion here to assert that latest sequence number is equal to `aws.Int64(20)`? Or what is this comment for?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+			{DesiredStatus: aws.String("STOPPED"), TaskArn: aws.String(testconst.TaskARN)},
+		},
+		Timeline: aws.Int64(20),
+	}
+
+	// Handle the task manifest message update
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage))
+	handleManifestMessage(testManifest)
+
+	ackWg.Wait()
```
danehlim, line 164: Why is using wait group to wait necessary here? Isn't the call on L162 blocking and prevents us from accessing the the sequence number before it has been updated?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+	mockSNA.SetLatestSequenceNumber(setupTestManifestMessage())
+
+	testTaskManifestResponder := NewTaskManifestResponder(mockTC, mockSNA, mmIDA, factory, f)
+	return testTaskManifestResponder
+}
+
+// TestManifestAckHappyPath tests the happy path for a typical TaskManifestMessage and confirms
+// expected ack is made
+func TestManifestAckHappyPath(t *testing.T) {
+	ctrl := gomock.NewController(t)
```
danehlim, line 75: Are L75 and L76 necessary given what we have on L54 and L55? This question also applies to other tests where we are creating a new controller and deferring its finish.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+	// As 15 < 20 we should not be able to update to an older sequence number
+	testTaskManifestResponder.snAccessor.SetLatestSequenceNumber(testManifest)
+
+	require.NotEqual(t,
+		aws.Int64(testTaskManifestResponder.snAccessor.GetLatestSequenceNumber()),
+		aws.Int64(StartingLatestTaskManifestSequenceNumber))
+}
+
+func TestTaskStopVerification(t *testing.T) {
+
```
danehlim, line 178: nit: Remove unnecessary trailing newline

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+	// Now manually try to reset the sequence and make sure it does not update
+	// As 15 < 20 we should not be able to update to an older sequence number
+	testTaskManifestResponder.snAccessor.SetLatestSequenceNumber(testManifest)
+
+	require.NotEqual(t,
+		aws.Int64(testTaskManifestResponder.snAccessor.GetLatestSequenceNumber()),
+		aws.Int64(StartingLatestTaskManifestSequenceNumber))
+}
+
+func TestTaskStopVerification(t *testing.T) {
```
danehlim, line 177: Could we combine this test with TestManifestAckHappyPath into a single happy path test? A happy path of handling a task manifest message involves both ACKing the task manifest message as well as sending a task stop verification message (if there is more than 1 task to stop) and I feel there would be less duplication/lines of code as well.

ecs-agent/acs/session/task_manifest_responder_test.go
danehlim, line 178: nit: Remove unnecessary newline

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,228 @@
...
+	)
+
+	// Handle the task manifest message update
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage))
+	go handleManifestMessage(testManifest)
+
+	// Wait for the ack to send through our channel so we can receive it
+	taskStopVerifSent := <-stopVerifSent
+
+	// We expect only one task marked for termination, and it should be running
```
danehlim, line 224: If we expect it to be running, should we assert that? Also, I'm assuming we only expect it to be running because we have set the desired status to RUNNING on L212. However FWIW, I'm not sure in the real world if we would ever receive any task here that doesn't have desired status STOPPED. Could we change the expectation of this test to use desired status STOPPED instead to more match what we expect to happen in normal operation?

ecs-agent/acs/session/task_manifest_responder.go
danehlim: Thanks! And

> Seems other responders in this library do not capitalize acknowledging, shouldn't it be kept the same here?

This comment is moreso since we usually capitalize the first letter of the first word in log statements, and "acknowledging" is the first word in this log statement. This is non-blocking though and I don't have a strong preference here.

ecs-agent/acs/session/task_manifest_responder.go
amogh09: Discussed with @timj-hh offline. The current behavior of Agent is to process the sending of ack and task stop verification messages asynchronously and this PR is only for refactoring and not to change any behavior. So, we would like to keep the asynchronicity here.

amogh09: Reference - https://github.com/aws/amazon-ecs-agent/blob/e978160b886ddfce9120a21beac9e09eca1bda92/agent/acs/handler/task_manifest_handler.go#L92-L93

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: Now that we have changed back to the async approach this will not be needed

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: ack

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: hmm, on the other hand one log for 5 messages in one is messy and difficult to decipher. Could I have something like the following?
Sending task stop verification message including:
task1
task2
task3

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: Synced offline with Amogh and we will include the waitgroup after all.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: Actually, L212 acts as an injection into the test so we can force the method to return the running task. I think cancelling a running task is exactly the use case for this component, as a stopped or stopping task is already being handled and would render this responder redundant. It is also not up to this responder to change task state or stop the task, that responsibility lies with ACS.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: When we test the sequence numbers, the test task manifest needs to start with state including `LatestSequenceNumber`. Else, checking for a stale number will always induce an error as manifestSeqNum `agentLatestSeqNum` will be null.

Similar reasoning for setting the message ID, we cannot check against a null value.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: Update: there is no easy way to print a list in a readable way, so I have documented that we will only send a task stop verification message once despite printing possibly many times.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,185 @@
...
+	manifestSeqNum := aws.Int64Value(message.Timeline)
+	agentLatestSeqNum := tmr.snAccessor.GetLatestSequenceNumber()
+
+	// Verify that task manifest isn't stale.
+	// The agent will keep track of the highest sequence number received in both PayloadMessage and TaskManifestMessage.
+	// If the sequence number on TaskManifestMessage is lower or equal to to the latest known one, the TaskManifestMessage must be discarded.
+	// The manifest should also not be a duplicate of the last one received, so the number cannot be less than or equal to the latest.
+	// ACS will guarantee that the sequence number is indeed increasing and valid for this use case.
+	if manifestSeqNum <= agentLatestSeqNum {
+		logger.Warn(fmt.Sprintf("Discarding task manifest message. "+
```
amogh09, line 103: Maybe we should add the discarded message to the log unless it is already logged elsewhere.

ecs-agent/acs/session/task_manifest_responder.go
danehlim: Synced offline with Tim - the purpose of using wait groups here is to make sure that both calls to `ackTaskManifestMessage` and `sendTaskStopVerification` have completed before the call to `m.Done(err)` in `handleTaskManifestMessage` function.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,185 @@
...
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package session
```
danehlim, line 14: Believe we'll need to rerun `go mod tidy` and `go mod vendor` in agent module to pick up the changes from the latest revision of this PR.

ecs-agent/acs/session/task_manifest_responder.go
amogh09: Oh I didn't focus much on the wait part. I was just curious why we call `ackTaskManifestMessage` and `sendTaskStopVerification` asynchronously.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,185 @@
...
+func (*taskManifestResponder) Name() string { return "task manifest responder" }
+
+func (tmr *taskManifestResponder) HandlerFunc() wsclient.RequestHandler {
+	return tmr.handleTaskManifestMessage
+}
+
+// handleTaskManifestMessage is the high level caller to handle a task manifest message from ACS.
+// It will kick off the call stack of processTaskManifestMessage calling ack and send TaskManifestMessage
+func (tmr *taskManifestResponder) handleTaskManifestMessage(message *ecsacs.TaskManifestMessage) {
+	logger.Debug(fmt.Sprintf("Processing %s with MessageId %s", TaskManifestMessageName, aws.StringValue(message.MessageId)))
```
danehlim, line 83: nit: Prefer to use `logger.Fields` and and `field.MessageID` in this log message.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,152 @@
...
+	mockComparer.EXPECT().CompareRunningTasksOnInstanceWithManifest(testManifest).Return(
+		[]*ecsacs.TaskIdentifier{
+			{DesiredStatus: aws.String("RUNNING"), TaskArn: aws.String(testconst.TaskARN)},
+		},
+		nil,
+	)
+
+	// Handle the task manifest message update.
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage))
+	go handleManifestMessage(testManifest)
```
amogh09, line 102: This asynchronous call seems unnecessary. I think we should be able to assert the right behavior without needing channels.

timj-hh: This is the paradigm we have been using in the agent and acs/session. It is minimal overhead so I think we should stick with this behavior.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: The messageID is logged inside of `handleTaskManifestMessage` before this method is called.

amogh09: That's a DEBUG level log. The default log level is INFO. I think we should add more context to this WARN level log.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,152 @@
...
+	mock_metrics "github.com/aws/amazon-ecs-agent/ecs-agent/metrics/mocks"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	StartingTaskManifestSequenceNumber = 15
+	EndingTaskManifestSequenceNumber   = 20
```
danehlim, line 35: Is this being used anywhere right now?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,152 @@
...
+	// when we validate it.
+	mockSNA.EXPECT().GetLatestSequenceNumber().Times(2)
+	mockIDA.EXPECT().SetMessageID(testconst.MessageID).Return(nil)
+
+	// Handle the task manifest message update.
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage))
+	handleManifestMessage(testManifest)
+
+	// Now manually try to reset the sequence and make sure it does not update
+	// As 15 <= 15, we should not be able to update to a stale sequence number.
```
danehlim, line 146: `testTaskManifestResponder.snAccessor.GetLatestSequenceNumber())` is 0 at this point, is that intentional? Or should we be using `.Return `method on L138 (and L81 for the matter)?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,152 @@
...
+	testTaskManifestResponder.snAccessor.SetLatestSequenceNumber(testManifest)
+
+	require.NotEqual(t,
+		aws.Int64(testTaskManifestResponder.snAccessor.GetLatestSequenceNumber()),
```
danehlim, line 150: `testTaskManifestResponder.snAccessor.GetLatestSequenceNumber())` returns 0 at this point, is that intentional? Or should we be using `.Return `method on L138 (and L81 for the matter)?

ecs-agent/acs/session/task_manifest_responder_test.go
danehlim: Resolving comment thread as this function no longer exists.

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -0,0 +1,185 @@
...
+	agentLatestSeqNum := tmr.snAccessor.GetLatestSequenceNumber()
+
+	// Verify that task manifest isn't stale.
+	// The agent will keep track of the highest sequence number received in both PayloadMessage and TaskManifestMessage.
+	// If the sequence number on TaskManifestMessage is lower or equal to to the latest known one, the TaskManifestMessage must be discarded.
+	// The manifest should also not be a duplicate of the last one received, so the number cannot be less than or equal to the latest.
+	// ACS will guarantee that the sequence number is indeed increasing and valid for this use case.
+	if manifestSeqNum <= agentLatestSeqNum {
+		logger.Warn(fmt.Sprintf("Discarding task manifest message with id %s "+
+			"Sequence number from manifest is less than or equal to Agent latest sequence number.", aws.StringValue(message.MessageId)), logger.Fields{
```
danehlim, line 104: Could we put the message ID in the `logger.Fields` along with `field.MessageID` and in this log message?

ecs-agent/acs/session/task_manifest_responder_test.go
danehlim: > Actually, L212 acts as an injection into the test so we can force the method to return the running task. I think cancelling a running task is exactly the use case for this component, as a stopped or stopping task is already being handled and would render this responder redundant. 

In normal operations we wouldn't expect the method to return a task with desired status RUNNING. Currently, the list of tasks returned by this method as part of this flow always have desired status STOPPED. I've sent some links offline to you to point to this behavior. 

> It is also not up to this responder to change task state or stop the task, that responsibility lies with ACS.

In the context of the full task sync feature, this responder is not responsible to stop the task. However, Agent is the one actually stopping the task but that is done via the task stop verification ACK responder.

ecs-agent/acs/session/task_manifest_responder.go
danehlim: > hmm, on the other hand one log for 5 messages in one is messy and difficult to decipher. 

But there aren't 5 messages? We only send one task stop verification message.

> Could I have something like the following?
Sending task stop verification message including:
task1
task2
task3

Yes, this is what I had in mind if we want to log and see what tasks are being stopped.

> there is no easy way to print a list in a readable way

What sort of difficulties are you running into? Is there anything stopping us from doing something like:

```
<get list of task ARNs>
logger.Debug("Sending task stop verification message", logger.Fields{
	field.MessageID: messageID,
	"taskARNList": <insert list of task ARNs here>,
})
```

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: ack

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: ack

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: ack

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: ack

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: Now that a staleness test is added both consts are in use.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: Synced offline, now validating expected data in the testrespondfunc, which should trigger inside of the responder rather than waiting until the end of test.

ecs-agent/acs/session/task_manifest_responder.go
timj-hh: synced offline and this has been changed.

ecs-agent/acs/session/task_manifest_responder.go
danehlim: Seems like there is still a diff in the vendored files in agent module and what is present in ecs-agent module. Can we run `go mod tidy` and `go mod vendor` in agent module to pick up the changes from ecs-agent module?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,149 @@
...
+	mock_metrics "github.com/aws/amazon-ecs-agent/ecs-agent/metrics/mocks"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	StartingTaskManifestSequenceNumber = 15
+	EndingTaskManifestSequenceNumber   = 10
```
danehlim, line 35: Is `EndingTaskManifestSequenceNumber` an accurate name for this const for the way it's being used right now? Perhaps something along the lines of `StaleTaskManifestSequenceNumber` might be better.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,149 @@
...
+	mockSNA.EXPECT().GetLatestSequenceNumber().Return(int64(EndingTaskManifestSequenceNumber)).AnyTimes()
+
+	// The test manifest should be valid, updating sequence number and message ID.
+	mockSNA.EXPECT().SetLatestSequenceNumber(testManifest)
+	mockIDA.EXPECT().SetMessageID(testconst.MessageID)
+	mockComparer.EXPECT().CompareRunningTasksOnInstanceWithManifest(testManifest).Return([]*ecsacs.TaskIdentifier{}, nil)
+
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage))
+
+	// Handle the task manifest message update, this should correctly set the message ID and sequence number.
```
danehlim, line 144: > this should correctly set the message ID and sequence number.

We should assert this, right?

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,149 @@
...
+	// Set up a new manifest with a stale number and distinct message ID.
+	newManifest := &ecsacs.TaskManifestMessage{
+		ClusterArn:           aws.String(testconst.ClusterName),
+		ContainerInstanceArn: aws.String(testconst.ContainerInstanceARN),
+		MessageId:            aws.String("456"),
+		Tasks:                []*ecsacs.TaskIdentifier{},
+		Timeline:             aws.Int64(EndingTaskManifestSequenceNumber),
+	}
+
+	// Expect GetLatestSequenceNumber once in handleManifestMessage, allowing it to pass
```
danehlim, line 133: > Expect GetLatestSequenceNumber once

Why are we using `.AnyTimes()` here then?

We do have two calls to `handleManifestMessage`, but I wouldn't expect `GetLatestSequenceNumber` to return the value held in `EndingTaskManifestSequenceNumber` const on the second call.

ecs-agent/acs/session/task_manifest_responder_test.go
```diff
@@ -0,0 +1,149 @@
...
+	mockSNA.EXPECT().SetLatestSequenceNumber(testManifest)
+	mockIDA.EXPECT().SetMessageID(testconst.MessageID)
+	mockComparer.EXPECT().CompareRunningTasksOnInstanceWithManifest(testManifest).Return([]*ecsacs.TaskIdentifier{}, nil)
+
+	handleManifestMessage := testTaskManifestResponder.HandlerFunc().(func(*ecsacs.TaskManifestMessage))
+
+	// Handle the task manifest message update, this should correctly set the message ID and sequence number.
+	handleManifestMessage(testManifest)
+
+	// Now try to update manifest with a stale message. This should fail.
```
danehlim, line 147: > This should fail.

This isn't a failure case right now, right? We simply skip/ignore stale messages and exit successfully. In this case, I think we should assert that the stale message was ignored.

timj-hh: ack.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: Can update to .Times(2), but this coerces the return instead of declaring what it should be iiuc. Will also add in an expect for the startingtaskmanifestsequencenumber that we expect to compare against in the second time.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: I think this is asserted with our mock behavior when we expect a call to setmessageID and setsequencenumber with the right vals.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: ack.

ecs-agent/acs/session/task_manifest_responder_test.go
danehlim: > Can update to .Times(2), but this coerces the return instead of declaring what it should be iiuc

I'm not sure I understand - what is the difference between coercing the return and declaring what it should be? In any case, the value held in `EndingTaskManifestSequenceNumber` const shouldn't be returned on the second call (or any subsequent calls) in the context of this test, right?

ecs-agent/acs/session/task_manifest_responder_test.go
danehlim: My understanding is that the mock behavior tells us that we should expect such calls to be made, but doesn't actually assert what those calls did/accomplished.

timj-hh: When we assert something like `EXPECT().SetSequenceNumber()`, we will expect the task manifest responder to be taking our values passed into `handleManifestMessage`, checking for staleness, and calling `SetSequenceNumber()`. The second time we call `handleManifestMessage`, the getter and setter are implicitly tested by doing that stale check again. The setter must have worked the first time to correctly set sequence num, our getter to return the right value to compare them, and finally the lack of calls to `SetMessageID` and `SetSequenceNum` with a stale message validate their behavior.

ecs-agent/acs/session/task_manifest_responder_test.go
timj-hh: No, the first time we call GetLatestSequenceNumber I do not know what it is checking against(likely 0). We will know that it passes though by coercing it to return `StaleTaskManifestSequenceNumber` because 10<=15. The second time we call GetLatestSequenceNumber it will check against StartingTaskManifestSequenceNumber, so if you want to be explicit then we can declare those two function calls.

My understanding is that `.EXPECT().<function-call>.Return(xyz)` will return xyz when it is called, despite what it will do in a live environment, because mocks do not hold state. When I say coerce a return, I mean to tell <function-call> to return xyz. Declaring what it should be is the misconception that `EXPECT().<function-call>.Return(xyz)` returns xyz naturally in its lifecycle.

danehlim: This is resolved as of the most recent revision - the value that was held in`EndingTaskManifestSequenceNumber` const is no longer returned on the second call.

ecs-agent/acs/session/task_manifest_responder_test.go
danehlim: > finally the lack of calls to SetMessageID and SetSequenceNum with a stale message validate their behavior

Got it, makes sense. Thanks for emphasizing the called "only once" in the latest revisions!

danehlim (approved):

amogh09 (approved):

PR #3849, opened 2023-08-15
TCS client integration

RichaGangwar:
### Summary
This PR makes necessary changes in TCS shared library to accommodate integration with agent.

### Implementation details

1.  Created new ecs client interface named as TcsEcsClient which would allow different agents to pass their ecs clients with one enforcemnt: the ecs client should have DiscoverTelemetryEndpoint() function implemented.
2. Updated mocks, utils and tests to include above change. 
3. Removed passing of endpoint in function StartTelemetrySession(), since the start will invoke the DiscoverTelemetryEndpoint() to poll needed endpoint.
4. Added new disconnectTimeoutHandler() to introduce periodic timeout after disconnectTimeout+ disconnectJitterMax. This functionality was added to ACS connection as part of an COE action item. This PR extends it to TCS connection as well.
5. Introduced new newHeartbeatTimeoutHandler() for appropriate logging on disconnect due to heartbeat timeout.
6. Added new tests for both above functionalities.
7. Updated tcs model to include "EphemeralStorageMetrics" field.
8. Reporter.go now would call Start() instead of StartTelemetrySession() in tcs handler.

### Testing
Unit and integ tests.

TCS client integration

### Licensing

### DIFF
--- a/agent/stats/reporter/reporter.go
+++ b/agent/stats/reporter/reporter.go
@@ -80,7 +80,6 @@ func NewDockerTelemetrySession(
 		agentVersion,
 		agentHash,
 		containerRuntimeVersion,
-		"", // this will be overridden by DockerTelemetrySession.Start()
 		cfg.DisableMetrics.Enabled(),
 		credentialProvider,
 		&wsclient.WSClientMinAgentConfig{
--- a/agent/stats/reporter/reporter.go
+++ b/agent/stats/reporter/reporter.go
@@ -16,7 +16,6 @@ package reporter
 import (
 	"context"
 	"errors"
-	"io"
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/agent/api"
@@ -29,7 +28,6 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
 	tcshandler "github.com/aws/amazon-ecs-agent/ecs-agent/tcs/handler"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/tcs/model/ecstcs"
-	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 )
@@ -48,9 +46,7 @@ const (
 )
 
 type DockerTelemetrySession struct {
-	s                    tcshandler.TelemetrySession
-	ecsClient            api.ECSClient
-	containerInstanceArn string
+	s tcshandler.TelemetrySession
 }
 
 // NewDockerTelemetrySession returns creates a DockerTelemetrySession, which has a tcshandler.TelemetrySession embedded.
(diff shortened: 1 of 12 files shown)


agent/stats/reporter/reporter.go
```diff
@@ -41,16 +39,10 @@ const (
...
-	backoffMax                  = 1 * time.Minute
-	jitterMultiple              = 0.2
-	multiple                    = 2
 )
 
 type DockerTelemetrySession struct {
-	s                    tcshandler.TelemetrySession
-	ecsClient            api.ECSClient
-	containerInstanceArn string
+	s tcshandler.TelemetrySession
```
Realmonia, line 45: nit: as we have removed other two fields from this type, we can just make it an anonymous field, and then call `session.Start` at line 102.

ecs-agent/tcs/handler/handler_test.go
```diff
@@ -340,32 +352,103 @@ func TestConnectionInactiveTimeout(t *testing.T) {
...
 	)
 
 	// Start a session with the test server.
-	err = session.StartTelemetrySession(ctx, server.URL)
+	err = session.StartTelemetrySession(ctx)
 
-	assert.NoError(t, err, "Close the connection should cause the tcs client return error")
+	// TODO: confirm once. The client would not attempt to reconnect since we are using StartTelemetrySession
+	// instead of Start. Modified the test to just test for connection closed on inactivity.
+	//assert.NoError(t, err, "Close the connection should cause the tcs client return error")
```
Realmonia, line 375: Should remove this todo? I think what it says about is done in `TestClientReconnectsAfterInactiveTimeout`?

ecs-agent/tcs/handler/handler_test.go
```diff
@@ -483,8 +570,73 @@ func TestStartTelemetrySessionMetricsChannelPauseWhenClientClosed(t *testing.T)
...
+	ctx, cancel := context.WithCancel(context.Background())
+	deregisterInstanceEventStream := eventstream.NewEventStream("Deregister_Instance", ctx)
+	deregisterInstanceEventStream.StartListening()
+	defer cancel()
+
+	telemetryMessages := make(chan ecstcs.TelemetryMessage, testTelemetryChannelDefaultBufferSize)
+	healthMessages := make(chan ecstcs.HealthMessage, testTelemetryChannelDefaultBufferSize)
+
+	//Setting disconnect timer(10 secs) to be less than heartbeat timer( 1min)
+	//in order disconnect because of periodic timer instead of heartbeat timer due to inactivity.
```
Realmonia, line 612: super nit since there are already other nit comments 🥲 - space after `//` and adjust parentheses

Realmonia (dismissed):
Looks good, some nit comments

agent/stats/reporter/reporter.go
RichaGangwar: Sorry, didn't understand the comment. What exactly do you mean by anonymous field? Do you mind giving an example?

ecs-agent/tcs/handler/handler_test.go
RichaGangwar: oh yeah sure. Just wanted to confirm that if I am understanding it correctly. Looks like it is correct and also I have added another testcase to test the reconnects.

agent/stats/reporter/reporter.go
Realmonia: change `s tcshandler.TelemetrySession` to `tcshandler.TelemetrySession`

ecs-agent/tcs/handler/handler.go
```diff
@@ -150,39 +172,97 @@ func (session *telemetrySession) StartTelemetrySession(ctx context.Context, endp
...
-	case err := <-serveC:
-		return err
+	return client.Serve(ctx)
+}
+
+func (session *telemetrySession) getTelemetryEndpoint() (string, error) {
+	containerInstanceARN := session.containerInstanceArn
+	tcsEndpoint, err := session.ecsClient.DiscoverTelemetryEndpoint(containerInstanceARN)
+	if err != nil {
+		logger.Error("tcs: unable to discover poll endpoint", logger.Fields{
```
xxx0624, line 217: Will we log this error in other callers?

ecs-agent/tcs/handler/handler_test.go
```diff
@@ -340,32 +352,103 @@ func TestConnectionInactiveTimeout(t *testing.T) {
...
 
+// TestClientReconnectsAfterInactiveTimeout tests the tcs client reconnects when it loses network
+// connection, or it's inactive for too long
+func TestClientReconnectsAfterInactiveTimeout(t *testing.T) {
+	// Start test server.
+	closeWS := make(chan []byte)
+	server, _, requestChan, _, err := wsmock.GetMockServer(closeWS)
+	server.StartTLS()
+	defer server.Close()
+	if err != nil {
```
xxx0624, line 392: Shouldn't this check be next to line # 389

ecs-agent/tcs/handler/handler_test.go
```diff
@@ -483,8 +570,73 @@ func TestStartTelemetrySessionMetricsChannelPauseWhenClientClosed(t *testing.T)
...
+
+// TestPeriodicDisconnectonTCSClient tests for tcs session disconnects at regular intervals.
+func TestPeriodicDisconnectonTCSClient(t *testing.T) {
+
+	// Start test server.
+	closeWS := make(chan []byte)
+	server, _, requestChan, serverErr, err := wsmock.GetMockServer(closeWS)
+	server.StartTLS()
+	defer server.Close()
+	if err != nil {
```
xxx0624, line 587: same here

ecs-agent/tcs/handler/handler_test.go
```diff
@@ -483,8 +570,73 @@ func TestStartTelemetrySessionMetricsChannelPauseWhenClientClosed(t *testing.T)
...
+		2*time.Second,
+		metrics.NewNopEntryFactory(),
+		telemetryMessages,
+		healthMessages,
+		emptyDoctor,
+		testecsclient,
+	)
+
+	// Start a session with the test server.
+	err = session.StartTelemetrySession(ctx)
```
xxx0624, line 635: Error check?

agent/stats/reporter/reporter.go
RichaGangwar: Got it. Sure. I can do that.

ecs-agent/tcs/handler/handler.go
RichaGangwar: Sorry, did not get you completely. Which other callers? There is no other non-test function where we invoke DiscoverTelemetryEndpoint().

ecs-agent/tcs/handler/handler_test.go
RichaGangwar: Yes, you are right. Thank you for pointing it out. Will fix it.

ecs-agent/tcs/handler/handler_test.go
RichaGangwar: Sure. will fix.

ecs-agent/tcs/handler/handler_test.go
RichaGangwar: Sure. Will add one. Thank you.

agent/stats/reporter/reporter.go
RichaGangwar: I actually tried that, but the function gets in an infinite loop. I guess it is because both the functions - DockerTelemetrySession and tcshandler.TelemetrySession are named as Start. When I do session.Start, it tries to start itself and traps in an infinite loop. I would say for readability and understandability for noobs like me, we can have a member name rather than leaving anonymous. Let me know if you think otherwise. :)

Realmonia (approved):

ecs-agent/tcs/handler/handler.go
xxx0624: Sry I mean will the caller of the `getTelemetryEndpoint` print the error as well? Just want to confirm we won't log the same error twice. But it's not a blocker. Probably we can also wrap this error with new message.

xxx0624 (approved):

ecs-agent/tcs/handler/handler.go
RichaGangwar: Got it. It's not being logged in the caller, but it is being logged in the caller of that function, i.e. Start(). Sure, let me pick it up in follow-up PR. Thank you :)

PR #3851, opened 2023-08-17
Start daemon import and add csi interface

fierlion:
### Summary
This builds on the initial managed daemon and csi-driver work for EBS Task Attach.  The main additions are the daemonmangers added to the app state, and loaded there; ebs-backed task handling is new; the basic csi-client calling the node csi stats is also new.  There are also several repetitive changes to add the daemons to the app state.

### Implementation details
`agent/app/agent.go` includes the daemon load and a new top level map of capability to daemonmanager
`agent/engine/daemonmanager/daemon_manager_linux.go` has a few updates that were caught in manual testing including loading the image and injecting the start command
`agent/engine/docker_task_engine.go` has handling for ebs-backed tasks (which will not exist yet because the check function will always return `false`).  This handling will create the internal task and add it to our agent state.
`ecs-agent/csiclient/csi_client.go` is a basic csi-client which will be extended as part of the TACS integration work.  This client exists currently only for testing.

The rest of the changes are related to module/dependency updates and general test updates to include the new state and objects. 

### Testing
Tests Are updated to include the new components.  Integration tests will be added as components are "plugged-in" in future commits.

Add managed daemon import and csi node client

### Licensing

### DIFF
--- a/agent/app/agent.go
+++ b/agent/app/agent.go
@@ -36,6 +36,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/ec2"
 	"github.com/aws/amazon-ecs-agent/agent/ecscni"
 	"github.com/aws/amazon-ecs-agent/agent/engine"
+	dm "github.com/aws/amazon-ecs-agent/agent/engine/daemonmanager"
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	"github.com/aws/amazon-ecs-agent/agent/engine/execcmd"
 	engineserviceconnect "github.com/aws/amazon-ecs-agent/agent/engine/serviceconnect"
@@ -62,6 +63,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/eventstream"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
+	md "github.com/aws/amazon-ecs-agent/ecs-agent/manageddaemon"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/tcs/model/ecstcs"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry"
 	"github.com/aws/aws-sdk-go/aws"
@@ -146,6 +148,7 @@ type ecsAgent struct {
 	saveableOptionFactory       factory.SaveableOption
 	pauseLoader                 loader.Loader
 	serviceconnectManager       engineserviceconnect.Manager
+	daemonManagers              map[string]dm.DaemonManager
 	eniWatcher                  *watcher.ENIWatcher
 	cniClient                   ecscni.CNIClient
 	vpc                         string
@@ -242,6 +245,7 @@ func newAgent(blackholeEC2Metadata bool, acceptInsecureCert *bool) (agent, error
 		saveableOptionFactory:       factory.NewSaveableOption(),
 		pauseLoader:                 pause.New(),
 		serviceconnectManager:       engineserviceconnect.NewManager(),
+		daemonManagers:              make(map[string]dm.DaemonManager),
 		cniClient:                   ecscni.NewClient(cfg.CNIPluginsPath),
 		metadataManager:             metadataManager,
 		terminationHandler:          sighandlers.StartDefaultTerminationHandler,
@@ -334,9 +338,33 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		StringSetValue: aws.StringSlice(gpuIDs),
 	}
 
+	daemonDefinitions, err := md.ImportAll()
+	// we will not panic when daemons fail to import; agent should continue running
+	// container instance health will act as the mechanism to advertise daemon failures
+	if err != nil {
+		seelog.Errorf("Daemon import mountpoint failure: %s", err)
+	}
+	if len(daemonDefinitions) == 0 {
+		seelog.Infof("daemonDefinitions is empty/nil after import")
+	}
+	// load and add daemons to agent
+	for _, md := range daemonDefinitions {
+		thisDaemon := dm.NewDaemonManager(md)
+		if _, err := thisDaemon.LoadImage(agent.ctx, agent.dockerClient); err != nil {
+			seelog.Errorf("Managed Daemon Load failure %v", err)
+		}
+		if loaded, err := thisDaemon.IsLoaded(agent.dockerClient); loaded {
+			imageManager.AddImageToCleanUpExclusionList(md.GetLoadedDaemonImageRef())
+			agent.daemonManagers[md.GetImageName()] = thisDaemon
+		} else {
+			seelog.Errorf("Unable to load Managed Daemon: %s, err: %s", md.GetImageName(), err)
+		}
+	}
+
 	// Create the task engine
 	taskEngine, currentEC2InstanceID, err := agent.newTaskEngine(
-		containerChangeEventStream, credentialsManager, state, imageManager, hostResources, execCmdMgr, agent.serviceconnectManager)
+		containerChangeEventStream, credentialsManager, state, imageManager, hostResources, execCmdMgr,
+		agent.serviceconnectManager, agent.daemonManagers)
 	if err != nil {
 		seelog.Criticalf("Unable to initialize new task engine: %v", err)
 		return exitcodes.ExitTerminal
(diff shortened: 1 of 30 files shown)


ecs-agent/manageddaemon/managed_daemon.go
```diff
@@ -71,25 +70,17 @@ func NewManagedDaemon(
 	// we'll override this default to -1
```
xxx0624, line 70: nit: update the comments as well

ecs-agent/csiclient/csi_client.go
```diff
@@ -0,0 +1,78 @@
...
+	csiSocket string
+}
+
+func NewCSIClient(socketIn string) csiClient {
+	return csiClient{csiSocket: socketIn}
+}
+
+// Used for testing and integration
+// TODO update with stats packing
+func (cc *csiClient) GetVolumeMetrics(volumeId string, hostMountPath string) (int64, error) {
```
xxx0624, line 31: this is still for local testing only? We should return totalBytes and usedBytes ideally.

xxx0624: And this pkg `csiclient` will hold all shared interfaces used to communicate with the csi driver daemon, right?

fierlion: (yes this will be updated as we integrate with telemetry -- I'll consider this a casualty of not having a feature branch for this work.)

agent/engine/daemonmanager/daemon_manager_linux.go
```diff
@@ -62,6 +61,10 @@ func NewDaemonManager(manageddaemon *md.ManagedDaemon) DaemonManager {
 	return &daemonManager{managedDaemon: manageddaemon}
 }
 
+func (dm *daemonManager) GetManagedDaemon() *md.ManagedDaemon {
```
saurabhc123, line 64: Can this be defined in daemon_manager.go? That would then avoid duplication across the linux and windows implementations. Also, can the lines from 56-66 be moved completely there?

ecs-agent/csiclient/csi_client.go
fierlion: this package will hold specifically CSI "Node" client calls, yes.

agent/app/agent.go
```diff
@@ -334,9 +338,25 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		StringSetValue: aws.StringSlice(gpuIDs),
 	}
 
+	daemonDefinitions, err := md.ImportAll()
+	seelog.Errorf("Daemon import mountpoint failure: %s", err)
```
fierlion, line 342: todo Import all could return an array and error, or nil and error -- need to handle both cases -- error for single daemon failure or all daemon failure.
todo add note that we will not panic when daemons fail to import -- agent should continue running.  container instance health will act as the notification for this case

agent/engine/daemonmanager/daemon_manager_linux.go
```diff
@@ -98,8 +88,7 @@ func (dm *daemonManager) CreateDaemonTask() (*apitask.Task, error) {
 	// directly, and the object only contains healthcheck, all other fields will be written as empty/nil
 	// in the result string. This will override the configurations that comes with the container image
 	// (CMD for example)
-	rawConfig := fmt.Sprintf("{\"Healthcheck\":%s}", string(rawHealthConfig))
-
+	rawConfig = fmt.Sprintf("{\"Healthcheck\":%s}", string(rawHealthConfig))
 	daemonTask := &apitask.Task{
 		Arn:                 fmt.Sprintf("arn:::::/%s-%s", dm.managedDaemon.GetImageName(), uuid.NewUUID()),
```
fierlion, line 93: TODO make sure that name and arn are unique where uniqueness is required.

agent/engine/docker_task_engine.go
```diff
@@ -1138,7 +1143,27 @@ func (engine *DockerTaskEngine) AddTask(task *apitask.Task) {
 		engine.emitTaskEvent(task, err.Error())
 		return
 	}
-
+	if task.IsEBSTaskAttachEnabled() {
+		if ebsCsiDaemonManager, ok := engine.daemonManagers["ebs-csi-driver"]; ok {
```
fierlion, line 1147: todo should the capability be a const?  This needs to be tied to the import dir name.  This should all be config.

agent/engine/docker_task_engine.go
```diff
@@ -1138,7 +1143,27 @@ func (engine *DockerTaskEngine) AddTask(task *apitask.Task) {
...
+					task.SetKnownStatus(apitaskstatus.TaskStopped)
+					task.SetDesiredStatus(apitaskstatus.TaskStopped)
+					engine.emitTaskEvent(task, err.Error())
+					return
+				}
+				engine.AddTask(engine.ebsCsiDriver)
+				logger.Info("docker_task_engine: Added EBS CSI task to engine")
+			}
+		} else {
+			logger.Error("Unable to find ebs-csi-driver in engine")
```
fierlion, line 1164: todo fail to start task if ebs backed and no driver

agent/engine/docker_task_engine.go
```diff
@@ -1138,7 +1143,27 @@ func (engine *DockerTaskEngine) AddTask(task *apitask.Task) {
 		engine.emitTaskEvent(task, err.Error())
 		return
 	}
-
+	if task.IsEBSTaskAttachEnabled() {
+		if ebsCsiDaemonManager, ok := engine.daemonManagers["ebs-csi-driver"]; ok {
+			if engine.ebsCsiDriver == nil {
```
fierlion, line 1148: || task was stopped 

TODO when the daemon task stops for a valid reason, we will need to replace it -- specifically here when we are waiting on a daemon-dependent task payload.

agent/engine/docker_task_engine.go
```diff
@@ -1138,7 +1143,33 @@ func (engine *DockerTaskEngine) AddTask(task *apitask.Task) {
...
+				logger.Error("Unable to start ebsCsiDaemon for task in the engine", logger.Fields{
+					field.TaskID: task.GetID(),
+					field.Error:  err,
+				})
+				task.SetKnownStatus(apitaskstatus.TaskStopped)
+				task.SetDesiredStatus(apitaskstatus.TaskStopped)
+				engine.emitTaskEvent(task, err.Error())
+				return
+			}
+		}
```
prateekchaudhry, line 1171: nit: I see daemon managers initialization has a common implementation. If this is a common pattern we expect for managed daemons, this part can be abstracted into a common function

agent/app/agent.go
```diff
@@ -334,9 +338,31 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
...
+	if daemonDefinitions == nil || len(daemonDefinitions) == 0 {
+		seelog.Infof("daemonDefinitions is empty/nil after import")
+	}
+	// load and add daemons to agent
+	for _, md := range daemonDefinitions {
+		thisDaemon := dm.NewDaemonManager(md)
+		if _, err := thisDaemon.LoadImage(agent.ctx, agent.dockerClient); err != nil {
+			seelog.Errorf("Managed Daemon Load failure %v", err)
+		}
+		if loaded, _ := thisDaemon.IsLoaded(agent.dockerClient); loaded {
```
amogh09, line 356: Is it safe to proceed with `thisDaemon` if `LoadImage` failed? It's not clear to me what `IsLoaded` does.

agent/engine/docker_task_engine.go
fierlion: good point -- I'll pick this up in a future rev if ok?

agent/engine/docker_task_engine.go
```diff
@@ -170,8 +171,10 @@ type DockerTaskEngine struct {
 	containerStatusToTransitionFunction map[apicontainerstatus.ContainerStatus]transitionApplyFunc
 	metadataManager                     containermetadata.Manager
 	serviceconnectManager               serviceconnect.Manager
+	daemonManagers                      map[string]dm.DaemonManager
 	hostResourceManager                 *HostResourceManager
 	serviceconnectRelay                 *apitask.Task
+	ebsCsiDriver                        *apitask.Task
```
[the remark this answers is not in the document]
fierlion: updating -- thanks for the catch!

ecs-agent/csiclient/csi_client.go
```diff
@@ -0,0 +1,79 @@
...
+		return int64(0), fmt.Errorf("failed to get usage from response. usage is nil")
+	}
+	var result = int64(0)
+	for _, usage := range usages {
+		unit := usage.GetUnit()
+		switch unit {
+		case csi.VolumeUsage_BYTES:
+			result = usage.GetUsed()
+		default:
+			logger.Info("Found missing key in volume usage")
```
amogh09, line 75: Might want to add `unit` to this log message unless this code is temporary.

fierlion: this code is temporary -- it will be removed when we pack the tcs call.

amogh09 (approved):

prateekchaudhry (approved):

PR #3854, opened 2023-08-17
Consume full task sync ACS responders from ecs-agent/

danehlim:
### Summary

Consume the [task manifest responder](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/acs/session/task_manifest_responder.go) and [task stop verification ACK responder](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/acs/session/task_stop_verification_ack_responder.go) defined in ecs-agent module. These ACS message responders are used to support the [full task sync feature](https://github.com/aws/amazon-ecs-agent/pull/2191).

### Implementation details

- Create new file `agent/acs/handler/task_manifest_responder.go` and inside that file, implement `TaskComparer` and `SequenceNumberAccessor` interfaces [defined in ecs-agent module](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/acs/session/task_manifest_responder.go#L33-L46). Some parts of the implementation are based off of functionality that exists in the current `agent/acs/handler/task_manifest_handler.go` file
- Create new file `agent/acs/handler/task_stop_verification_ack_responder.go` and inside that file, implement `TaskStopper` interface [defined in ecs-agent module](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/acs/session/task_stop_verification_ack_responder.go#L16-L18). Some parts of the implementation are based off of functionality that exists in the current `agent/acs/handler/task_manifest_handler.go` file
- Add tests for the above in new files `agent/acs/handler/task_manifest_responder_test.go` and `agent/acs/handler/task_stop_verification_ack_responder_test.go`. Some tests are based off of tests that exist in the current `agent/acs/handler/task_manifest_handler_test.go` file
- Delete `agent/acs/handler/task_manifest_handler.go` and `agent/acs/handler/task_manifest_handler_test.go`
- In `agent/acs/handler/acs_handler.go`:
   - Initialize task manifest responder and add its `HandlerFunc` function as a request handler to the ACS client, such that task manifest responder dictates what to do in response to receiving task manifest messages from ACS
   - Initialize task stop verification ACK responder and add its `HandlerFunc` function as a request handler to the ACS client, such that task stop verification ACK responder dictates what to do in response to receiving task stop verification ACKs from ACS
      - NOTE: On L274, you will notice we are passing `nil` as the respond function parameter. This is because although the task stop verification ACK responder in ecs-agent module currently takes in and contains a respond function parameter, it actually doesn't use it since we do not send anything back to ACS in response to task stop verification ACKs. I've left removing this non-necessity from the task stop verification ACK responder in the ecs-agent module out of scope of this pull request as this pull request is limited to and focuses on changes in the agent module
   - Remove ACS messages related to full task sync feature from pending ACKs logic since it is no longer applicable/used after consuming the task manifest and task stop verification ACK responders from ecs-agent module
      - Such logic is no longer applicable since there is no need to do multi-thread synchronization with channels with this responder (like we currently do in `agent/acs/handler/task_manifest_handler.go`) and the responders in the ecs-agent module factor out this non-necessity and are a lot cleaner and simpler to follow.
   - Remove pending ACKs logic altogether since there are no longer any responders that use it now

### Testing

Consume full task sync ACS responders from ecs-agent/

### Licensing


agent/acs/handler/task_manifest_responder.go
```diff
@@ -0,0 +1,98 @@
...
+// NewTaskComparer creates a new taskComparer.
+func NewTaskComparer(taskEngine engine.TaskEngine) *taskComparer {
+	return &taskComparer{
+		taskEngine: taskEngine,
+	}
+}
+
+// sequenceNumberAccessor implements the SequenceNumberAccessor interface defined in ecs-agent module.
+type sequenceNumberAccessor struct {
+	latestSeqNumberTaskManifest *int64
```
Realmonia, line 40: It might be I am missing some previous discussion in other PRs - why this variable named `latestSeqNumberTaskManifest` but not `latestTaskManifestSeqNumber` and why is it a `*int64` instead of `int64`?

danehlim: > why this variable named latestSeqNumberTaskManifest but not latestTaskManifestSeqNumber

This variable name is how it is currently named in the existing `agent/acs/handler/task_manifest_handler.go` file (see [here](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/acs/handler/task_manifest_handler.go#L43)). This existing naming was just carried over.

> why is it a *int64 instead of int64?
We need to pass by pointer here. If we were to pass by value instead, then we would not be able to change the latestSeqNumberTaskManifest tracked by the Agent (we would only be able to make changes to a copy of it). See [here](https://github.com/aws/amazon-ecs-agent/pull/3840#discussion_r1287767837) for previous similar discussion on this.

agent/acs/handler/task_manifest_responder.go
```diff
@@ -0,0 +1,98 @@
...
+
+// NewSequenceNumberAccessor creates a new NewSequenceNumberAccessor.
+func NewSequenceNumberAccessor(latestSeqNumberTaskManifest *int64, dataClient data.Client) *sequenceNumberAccessor {
+	return &sequenceNumberAccessor{
+		latestSeqNumberTaskManifest: latestSeqNumberTaskManifest,
+		dataClient:                  dataClient,
+	}
+}
+
+func (tc *taskComparer) CompareRunningTasksOnInstanceWithManifest(
```
mye956, line 52: nit: perhaps we could we include some context comments here or add back the ones from `compareTasks` if it's still applicable?

danehlim: Sure, if other blocking changes need to be made I'll make sure to add a comment above this function.

agent/acs/handler/task_manifest_responder.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	return tasksToBeKilled, nil
+}
+
+func (sna *sequenceNumberAccessor) GetLatestSequenceNumber() int64 {
+	return *sna.latestSeqNumberTaskManifest
+}
+
+func (sna *sequenceNumberAccessor) SetLatestSequenceNumber(message *ecsacs.TaskManifestMessage) error {
```
mye956, line 89: nit: it looks like we just use `*message.Timeline` here, perhaps we can just pass in that instead of the whole message?

mye956 (approved):

agent/acs/handler/task_manifest_responder.go
danehlim: Currently we need to pass in the message, otherwise the `sequenceNumberAccessor` struct won't implement the `SequenceNumberAccessor` interface defined in ecs-agent module.

As a follow up, I can check with other folks and see if other implementers of the `SequenceNumberAccessor` interface require any other fields of the task manifest message in their implementation of the `SetLatestSequenceNumber` method. If not, I agree the message parameter in ecs-agent module should be changed just to be a sequence number parameter (if the results do point to this, I will do this in a followup PR).

agent/acs/handler/task_manifest_responder.go
```diff
@@ -0,0 +1,98 @@
...
+func NewSequenceNumberAccessor(latestSeqNumberTaskManifest *int64, dataClient data.Client) *sequenceNumberAccessor {
+	return &sequenceNumberAccessor{
+		latestSeqNumberTaskManifest: latestSeqNumberTaskManifest,
+		dataClient:                  dataClient,
+	}
+}
+
+func (tc *taskComparer) CompareRunningTasksOnInstanceWithManifest(
+	message *ecsacs.TaskManifestMessage) ([]*ecsacs.TaskIdentifier, error) {
+	runningTasksOnInstance, err := tc.taskEngine.ListTasks()
```
mythri-garaga, line 54: [non-blocking] can we rename this to `allTasksOnInstance` since we are actually filtering tasks in running state on line 63?

danehlim: Good call out! This variable name was pulled in from the existing `agent/acs/handler/task_manifest_handler.go` file, but agreed that something like `allTasksOnInstance` is a more accurate variable name. If other blocking changes need to be made, I'll be sure to rename this (otherwise, I'll rename in a followup PR).

Realmonia (approved):

PR #3858, opened 2023-08-18
Added the implementation for EBS Watcher for Windows

saurabhc123:
### Summary

This PR adds the feature to detect an attached EBS volume for a EC2 Windows host.

### Implementation details

We used the `wmic` windows utility to detect the volume based on the volumeID. 

### Testing

Two additional unit tests were added on how the `wmic` command output is handled by new code.

Added the ability to detect EBS volumes on Windows.

### Licensing


ecs-agent/api/resource/ebs_discovery_windows.go
```diff
@@ -0,0 +1,59 @@
+//go:build windows
+// +build windows
+
```
mye956, line 3: Seems like the copyright message is missing. Could you please add something like the following? https://github.com/aws/amazon-ecs-agent/blob/master/agent/app/agent.go#L1

ecs-agent/api/resource/ebs_discovery_windows_test.go
```diff
@@ -0,0 +1,31 @@
+//go:build windows && unit
+// +build windows,unit
+
```
mye956, line 3: Same comment as above regarding the copyright information.

mye956 (commented):
Could the destination branch be changed to`dev` instead of `master`?

mye956 (commented):
It looks like we're missing the ebs_discovery_windows.go file in `agent/vendor`. To remediate, could you run `go mod vendor` within the `agent` directory?

PR #3860, opened 2023-08-21
Log error if handling task manifest message fails

danehlim:
### Summary

Currently if Agent fails to handle a task manifest message fail (i.e., call to `processTaskManifestMessage` returns an error), we do not log any meaningful error. Thus, right now there would be no indication in the logs if handling a task manifest message failed for any reason (which is also not consistent with the behavior of other ACS message responders).

This pull request simply adds an informative log statement with level `ERROR` in the event that the call to `processTaskManifestMessage` returns an error.

### Implementation details

See Summary above.

### Testing

Log error if handling task manifest message fails

### Licensing

### DIFF
--- a/ecs-agent/acs/session/task_manifest_responder.go
+++ b/ecs-agent/acs/session/task_manifest_responder.go
@@ -80,9 +80,18 @@ func (tmr *taskManifestResponder) HandlerFunc() wsclient.RequestHandler {
 // handleTaskManifestMessage is the high level caller to handle a task manifest message from ACS.
 // It will kick off the call stack of processTaskManifestMessage calling ack and send TaskManifestMessage
 func (tmr *taskManifestResponder) handleTaskManifestMessage(message *ecsacs.TaskManifestMessage) {
-	logger.Debug(fmt.Sprintf("Processing %s", TaskManifestMessageName), logger.Fields{field.MessageID: message.MessageId})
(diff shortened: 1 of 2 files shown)


Yiyuanzzz (dismissed):

ecs-agent/acs/session/task_manifest_responder.go
```diff
@@ -83,6 +83,12 @@ func (tmr *taskManifestResponder) handleTaskManifestMessage(message *ecsacs.Task
 	logger.Debug(fmt.Sprintf("Processing %s", TaskManifestMessageName), logger.Fields{field.MessageID: message.MessageId})
 	m := tmr.metricsFactory.New(metrics.TaskManifestHandlingDuration)
 	err := tmr.processTaskManifestMessage(message)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Unable to handle %s", TaskManifestMessageName), logger.Fields{
+			field.MessageID: aws.StringValue(message.MessageId),
```
timj-hh, line 88: Because we're now using message.MessageId in two places here(L83), maybe declare a var to hold `aws.StringValue(message.MessageID)`?

danehlim: Sure, can do!

timj-hh (approved):

Yiyuanzzz (approved):

PR #3866, opened 2023-08-23
Adding EBS watcher implementation

mye956:
### Summary
This PR will introduce the implementation for the EBS watcher which is responsible for detecting and validating that an EBS volume is attached onto the host.

### Implementation details
The general workflow is upon agent start up, the EBS watcher gets initialized and will start to periodically check whether we have any pending EBS attachments to find on the host. If there is then it'll scan for all pending EBS volumes within the agent state and try to find if they're attached via `lsblk`. Otherwise it's a noop. The watcher will always be running and will stop whenever agent stops.

EBS volumes are added to the agent state whenever we receive and handle an attachment message.

* Introduces the EBS watcher
  * Added a `watcher.go` into `agent/ebs`
    * Introduce new EBSWatcher struct 
    * Added new functionality to:
      * `NewWatcher`: Create a new instance of a EBSWatcher struct
      * `Start`: Start the periodic scanning process of the EBSWatcher for all pending EBS attachments saved to the agent state
      * `Stop`: Stop the EBSWatcher
      * `HandleResourceAttachment`: Handle a new ResourceAttachment object from the Attachment Message Responder/Handler and save it into the agent state if it does not exists.
      * `notifyFoundEBS`: Mark the pending ResourceAttachment as no longer pending as well as stops the ACK timer
      * `RemoveAttachment`: Removes a ResourceAttachment object from the list of attachments within the agent state
      * `scanEBSVolumes`: Iterates through the list of pending ResourceAttachment objects within the agent state and checks if it's attached within the host via `ConfirmEBSVolumeIsAttached`
      * `addEBSAttachmentToState`: Starts the ACK timer of a ResourceAttachment object as well as add it to the list of attachments within the agent state
      * `handleEBSAckTimer`: Handles the when a ResourceAttachment object once its ACK timer expires
* Minor changes to the log statements within Docker task engine state when adding and removing ResourceAttachments from the list
* Added new functionality in `ecs-agent/api/resource/resource_attachment.go` to safely read ResourceAttachment objects within logs
* Introduces the implementation of EBS volume discovery and validation
  * Added a new interface called `EBSDiscovery` in `ecs-agent/api/resource/interfaces.go` 
  * Added a new struct called`EBSDiscoveryClient` in `ecs-agent/api/resource/ebs_discovery.go` which implements the `EBSDiscovery` interface
  * `ebs_discovery_linux.go`
    * `ConfirmEBSVolumeIsAttached`: Calls upon `lsblk -o NAME,SERIAL -J` to obtain all attached volumes on the host and validates whether the given device name and volume ID is within the output.
    * `parseLsblkOutput`: Parses the output of `lsblk` and then returns the volume ID of a specific device name from the output
  * `ebs_discovery_windows.go`
    * `ConfirmEBSVolumeIsAttached`: The windows version to find and validate whether a EBS volume is attached on the host
  * `ecs-agent/api/resource/generate_mocks.go`: Generate new mocks for the `EBSDiscovery` interface for testing 

Note: There will be a follow up PR for the ACS Attachment message handler/responder that'll also include additional changes to the EBS watcher.

This PR will also include changes made in the following PR https://github.com/mye956/amazon-ecs-agent/pull/49.

### Testing

Unit tests

Adding implementation for EBS watcher and EBS volume discovery.

### Licensing


agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+	w.agentState.RemoveEBSAttachment(volumeID)
+	if len(w.agentState.GetAllPendingEBSAttachments()) == 0 {
+		log.Info("No more attachments to scan for. Stopping scan ticker.")
+		w.scanTicker.Stop()
+	}
+}
+
+// addEBSAttachmentToState adds an EBS attachment to state, and start its ack timer
+func (w *EBSWatcher) addEBSAttachmentToState(ebs *apiebs.ResourceAttachment) error {
+	volumeId := string(ebs.AttachmentProperties[apiebs.VolumeIdName])
```
amogh09, line 189: Is `ebs.AttachmentProperties[apiebs.VolumeIdName]` not of type `string`? Curious about the casting.

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+	wg.Add(1)
+	w.mailbox <- func() {
+		defer wg.Done()
+		empty := len(w.agentState.GetAllPendingEBSAttachments()) == 0
+
+		err := w.handleEBSAttachment(ebs)
+		if err != nil {
+			log.Warnf("Failed to handle resource attachment %v", ebs.String())
+		}
+		if empty && len(w.agentState.GetAllPendingEBSAttachments()) == 1 {
```
amogh09, line 100: Can we please add some comments explaining this condition. It's not very clear to me.

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Stop will stop the EBS watcher
+func (w *EBSWatcher) Stop() {
+	log.Info("Stopping EBS watcher.")
+	w.cancel()
+}
+
+// HandleResourceAttachment processes the resource attachment message.
+func (w *EBSWatcher) HandleResourceAttachment(ebs *apiebs.ResourceAttachment) error {
```
amogh09, line 88: Who would be calling this method?

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+		defer wg.Done()
+		empty := len(w.agentState.GetAllPendingEBSAttachments()) == 0
+
+		err := w.handleEBSAttachment(ebs)
+		if err != nil {
+			log.Warnf("Failed to handle resource attachment %v", ebs.String())
+		}
+		if empty && len(w.agentState.GetAllPendingEBSAttachments()) == 1 {
+			w.scanTicker.Stop()
+			w.scanTicker = time.NewTicker(apiebs.ScanPeriod)
```
amogh09, line 102: Why stopping and then starting again with a new ticker?

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+)
+
+type EBSWatcher struct {
+	ctx        context.Context
+	cancel     context.CancelFunc
+	scanTicker *time.Ticker
+	agentState dockerstate.TaskEngineState
+	// TODO: The dataClient will be used to save to agent's data client as well as start the ACK timer. This will be added once the data client functionality have been added
+	// dataClient     data.Client
+	ebsChangeEvent  chan<- statechange.Event
```
amogh09, line 34: This field is unused. Will it be used in an upcoming PR?

agent/ebs/watcher_test.go
```diff
@@ -0,0 +1,281 @@
...
+	mockDiscoveryClient := mock_ebs_discovery.NewMockEBSDiscovery(mockCtrl)
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	mockDiscoveryClient.EXPECT().ConfirmEBSVolumeIsAttached(deviceName, volumeID).Return(nil).MinTimes(1).
+		Do(func(deviceName, volumeID string) {
+			wg.Done()
+		}).
+		Return(nil).
+		MinTimes(1)
```
amogh09, line 108: Does this mean we expect two calls both returning `nil` but the first one calls the `wg.Done()`?

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+		err := w.handleEBSAttachment(ebs)
+		if err != nil {
+			log.Warnf("Failed to handle resource attachment %v", ebs.String())
+		}
+		if empty && len(w.agentState.GetAllPendingEBSAttachments()) == 1 {
+			w.scanTicker.Stop()
+			w.scanTicker = time.NewTicker(apiebs.ScanPeriod)
+		}
+	}
+	wg.Wait()
```
amogh09, line 105: So we send a function to `mailbox` and then wait for it to be executed by another goroutine. Why?

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+		w.removeEBSAttachment(volumeID)
+	}
+}
+
+func (w *EBSWatcher) removeEBSAttachment(volumeID string) {
+	// TODO: Remove the EBS volume from the data client.
+	w.agentState.RemoveEBSAttachment(volumeID)
+	if len(w.agentState.GetAllPendingEBSAttachments()) == 0 {
+		log.Info("No more attachments to scan for. Stopping scan ticker.")
+		w.scanTicker.Stop()
```
amogh09, line 183: Why do we need to stop the ticker at so many places? I am concerned that the complexity of ticker state management makes it easy to introduce bugs.

agent/ebs/watcher_test.go
```diff
@@ -0,0 +1,281 @@
...
+		},
+		AttachmentProperties: testAttachmentProperties,
+	}
+
+	watcher := setupWatcher(ctx, cancel, taskEngineState, eventChannel, mockDiscoveryClient)
+	testGroup.Start(watcher)
+
+	watcher.HandleResourceAttachment(ebsAttachment)
+
+	time.Sleep(600 * time.Millisecond)
```
amogh09, line 181: Are we sleeping to pass the expiry? If so, an we use smaller expiration and sleep duration to make the test run faster?

agent/ebs/watcher_test.go
```diff
@@ -0,0 +1,281 @@
...
+
+	assert.Len(t, taskEngineState.(*dockerstate.DockerTaskEngineState).GetAllEBSAttachments(), 0)
+	_, ok := taskEngineState.(*dockerstate.DockerTaskEngineState).GetEBSByVolumeId(volumeID)
+	assert.False(t, ok)
+}
+
+func TestHandleDuplicateEBSAttachment(t *testing.T) {
```
amogh09, line 188: Can we please add the expected behavior as documentation above the test functions? That makes it easier to quickly understand what a test tests.

ecs-agent/api/resource/ebs_discovery.go
```diff
@@ -0,0 +1,72 @@
...
+}
+
+func NewDiscoveryClient(ctx context.Context) EBSDiscovery {
+	return &EBSDiscoveryClient{
+		ctx: ctx,
+	}
+}
+
+// ScanEBSVolumes will iterate through the entire list of pending EBS volume attachments within the agent state and checks if it's attached on the host.
+func ScanEBSVolumes[T GenericEBSAttachmentObject](t map[string]T, dc EBSDiscovery) []string {
```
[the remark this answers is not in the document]
mye956: > Is there a reason we want to use generics?

Ah! I guess we don't have to use generics but I was inspired by a talk about generics at the time of writing this code :). Will test it without generics.

> Also how about using a more descriptive name than t for the first parameter?

Good call out. Will update it.

ecs-agent/api/resource/ebs_discovery_windows.go
```diff
@@ -0,0 +1,21 @@
...
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package resource
+
+func (api EBSDiscoveryClient) ConfirmEBSVolumeIsAttached(deviceName, volumeID string) error {
+	return nil
```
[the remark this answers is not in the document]
mye956: Not necessarily, this is rather a placeholder and will be implemented by this PR: https://github.com/aws/amazon-ecs-agent/pull/3868. Will add a comment.

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+	log "github.com/cihub/seelog"
+)
+
+type EBSWatcher struct {
+	ctx        context.Context
+	cancel     context.CancelFunc
+	scanTicker *time.Ticker
+	agentState dockerstate.TaskEngineState
+	// TODO: The dataClient will be used to save to agent's data client as well as start the ACK timer. This will be added once the data client functionality have been added
+	// dataClient     data.Client
```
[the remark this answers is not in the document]
mye956: Yep, this will most likely be updated/modified when we get to the attachment handler/responders where we'll actually hook up and start the watcher when agent starts.

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+
+func (w *EBSWatcher) NotifyFound(foundVolumes []string) {
+	for _, volumeId := range foundVolumes {
+		w.notifyFoundEBS((volumeId))
+	}
+}
+
+// notifyFoundEBS will mark it as found within the agent state
+func (w *EBSWatcher) notifyFoundEBS(volumeId string) {
+	w.mailbox <- func() {
```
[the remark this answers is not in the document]
mye956: This was inspired by another workflow process and implementation where it uses a buffered channel that receives function calls. Was also thinking about whether we even need this mailbox and how scalable it is. Will get back to this after a bit of testing.

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// Start is used to kick off the periodic scanning process of the EBS volume attachments for the EBS watcher.
+// If there aren't any initially, the scan ticker will stop.
+func (w *EBSWatcher) Start() {
+	log.Info("Starting EBS watcher.")
+
+	w.scanTicker = time.NewTicker(apiebs.ScanPeriod)
+	if len(w.agentState.GetAllPendingEBSAttachments()) == 0 {
+		w.scanTicker.Stop()
```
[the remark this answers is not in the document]
mythri-garaga: When is this  EBSWatcher `Start()` called?  during resource initialization?

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+type EBSWatcher struct {
+	ctx        context.Context
+	cancel     context.CancelFunc
+	scanTicker *time.Ticker
+	agentState dockerstate.TaskEngineState
+	// TODO: The dataClient will be used to save to agent's data client as well as start the ACK timer. This will be added once the data client functionality have been added
+	// dataClient     data.Client
+	ebsChangeEvent  chan<- statechange.Event
+	discoveryClient apiebs.EBSDiscovery
+	mailbox         chan func()
```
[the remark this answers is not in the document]
mythri-garaga: +1

agent/ebs/watcher.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// HandleResourceAttachment processes the resource attachment message.
+func (w *EBSWatcher) HandleResourceAttachment(ebs *apiebs.ResourceAttachment) error {
+	var err error
+	var wg sync.WaitGroup
+	wg.Add(1)
+	w.mailbox <- func() {
+		defer wg.Done()
+		empty := len(w.agentState.GetAllPendingEBSAttachments()) == 0
```
mythri-garaga, line 94: can we rename `empty` to something more specific?

agent/ebs/watcher.go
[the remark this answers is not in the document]
mye956: Added a comment for both ebsChangeEvent and dataClient as TODO. Will be used in very near future PR.

agent/ebs/watcher_test.go
mye956: Good callout, added comments to the test cases.

ecs-agent/api/resource/ebs_discovery_windows.go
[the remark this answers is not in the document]
mye956: Update: Will be added in this PR as well. Currently there's a PR with the implementation -> https://github.com/mye956/amazon-ecs-agent/pull/49

ecs-agent/api/resource/interfaces.go
```diff
@@ -0,0 +1,23 @@
...
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package resource
+
+type EBSDiscovery interface {
```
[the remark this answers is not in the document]
mye956: Added comments to briefly describe the interfaces.

agent/ebs/watcher.go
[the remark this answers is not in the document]
mye956: A bit of an update, we'll no longer use a mailbox system.

agent/ebs/watcher.go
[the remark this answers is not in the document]
mye956: Update: Will be executing right away and no longer using a mailbox. Thanks!

agent/ebs/watcher.go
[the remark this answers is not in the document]
mye956: Great questions. We'll only have one ticker active at all times. The EBSWatcher `Start()` will be called whenever we have a pending EBS attachment to find (i.e. whenever we receive an attachment message). It'll also start on agent start up (will be done in a very near PR). 

Note: It will start up whenever we have a list of pending EBS volumes of size 1.

agent/ebs/watcher.go
mye956: Good call out. Fixed!

agent/ebs/watcher.go
mye956: Yep, renamed both `empty` and added some comments to better explain.

agent/ebs/watcher.go
mye956: Will be called by a responder within `/ecs-agent/acs/session`. Will be done in a near follow up PR.

agent/ebs/watcher.go
mye956: Completely agree, this is a bit confusing. It was specific to an existing implementation but is no longer applicable. Thanks!

agent/ebs/watcher.go
mye956: Yep, added a comment about it.

agent/ebs/watcher_test.go
mye956: Ah, good catch. Fixed.

agent/ebs/watcher.go
mye956: No longer using a `mailbox` system. Thanks!

agent/ebs/watcher.go
mye956: We want to conservatively start scanning only when necessary. Thus, we should stop the ticker whenever we no longer have any pending attachments. This happens when:
1. We finished scanning volumes
2. We removed a volume

There's also the case where the context gets canceled and so we'd want to stop the ticker as well.

agent/ebs/watcher_test.go
mye956: Good call out, updated to have a timeout duration of 5ms and sleep for 10ms.

agent/ebs/watcher.go
mye956: Good call out, I renamed it to `wasEmpty` as well as added some comments to better explain the use case. Thanks!

ecs-agent/api/resource/ebs_discovery.go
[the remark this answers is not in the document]
amogh09: Discussed offline that generics are needed here due to the fact that `map[string]MyImplementation` does not type check against `map[string]MyInterface`. My bad.

agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/resource/ebs_discovery.go
```diff
@@ -0,0 +1,96 @@
...
+	ctx context.Context
+}
+
+// ScanTickerController is used to keep track of the scanning interval of EBS volumes for the EBS watcher.
+type ScanTickerController struct {
+	ScanTicker *time.Ticker
+	Running    bool
+	TickerLock sync.Mutex
+	Done       chan bool
+}
```
amogh09, line 48: There is too much state in this struct.... can we keep the ticker running and call `checkPendingEBSVolumes` function in the tick handler to decide if a scan should be performed for this tick? If not, then we do nothing in this tick. The runtime cost of having a ticker running is much lower than the readability and maintenance cost of ticker lifecycle management IMO.

mye956: No longer using the `ScanTickerController`. We'll have the watcher to start up and have the ticker to always be ticking as long as agent is up. Thanks!

agent/ebs/watcher.go
```diff
@@ -0,0 +1,175 @@
...
+		w.notifyFoundEBS(volumeId)
+	}
+}
+
+// notifyFoundEBS will mark it as found within the agent state
+func (w *EBSWatcher) notifyFoundEBS(volumeId string) {
+	// TODO: Add the EBS volume to data client
+	ebs, ok := w.agentState.GetEBSByVolumeId(volumeId)
+	if !ok {
+		log.Warnf("Unable to find EBS volume with volume ID: %v.", volumeId)
```
amogh09, line 119: We should probably clarify that Agent failed to find the volume _in its state_.

agent/ebs/watcher.go
```diff
@@ -0,0 +1,175 @@
...
+// removeEBSAttachment removes a EBS volume with a specific volume ID
+func (w *EBSWatcher) removeEBSAttachment(volumeID string) {
+	// TODO: Remove the EBS volume from the data client.
+	w.agentState.RemoveEBSAttachment(volumeID)
+}
+
+// addEBSAttachmentToState adds an EBS attachment to state, and start its ack timer
+func (w *EBSWatcher) addEBSAttachmentToState(ebs *apiebs.ResourceAttachment) error {
+	volumeId := ebs.AttachmentProperties[apiebs.VolumeIdName]
+	err := ebs.StartTimer(func() {
```
amogh09, line 153: What is the behavior of this timer if Agent restarts before when the timer is running?

agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/resource/ebs_discovery.go
```diff
@@ -0,0 +1,62 @@
...
+	ctx context.Context
+}
+
+func NewDiscoveryClient(ctx context.Context) *EBSDiscoveryClient {
+	return &EBSDiscoveryClient{
+		ctx: ctx,
+	}
+}
+
+// ScanEBSVolumes will iterate through the entire list of pending EBS volume attachments within the agent state and checks if it's attached on the host.
```
amogh09, line 44: "pending EBS volume attachments within the agent state" is a little confusing without the additional context that `pendingAttachments` comes from agent state. How about just saying that "entire list of provided EBS attachments"?

ecs-agent/api/resource/ebs_discovery_linux.go
```diff
@@ -0,0 +1,92 @@
...
+	return nil
+}
+
+// parseLsblkOutput will parse the `lsblk` output and search for a EBS volume with a specific device name.
+// Once found we return the volume ID, otherwise we return an empty string along with an error
+// The output of the "lsblk -o +SERIAL" command looks like the following:
+// NAME          MAJ:MIN RM SIZE RO TYPE MOUNTPOINT SERIAL
+// nvme0n1       259:0    0  30G  0 disk            vol123
+// ├─nvme0n1p1   259:1    0  30G  0 part /
+// └─nvme0n1p128 259:2    0   1M  0 part
```
amogh09, line 83: This is not JSON then why are we parsing it as JSON on L56?

ecs-agent/api/resource/ebs_discovery_windows_test.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func TestParseExecutableOutputWithMissingDiskNumber(t *testing.T) {
+	output := fmt.Sprintf("Disk Number: 0\r\nVolume ID: vol-abcdef1234567890a\r\nDevice Name: sda1\r\n\r\nVolume ID: %s\r\nDevice Name: %s\r\n\r\n", testVolumeID, deviceName)
+	parsedOutput, err := parseExecutableOutput([]byte(output), testVolumeID, deviceName)
+	require.Error(t, err)
+	assert.Equal(t, "", parsedOutput)
+}
+
+func TestParseExecutableOutputWithMissingVolumeInformation(t *testing.T) {
```
amogh09, line 47: How about writing table driven tests instead of writing a new function for each test case?

ecs-agent/api/resource/ebs_discovery_linux.go
mye956: Ah, so this is just an example of what the lsblk command outputs. Although, it might be better to include the JSON version. I'll update this comment.

agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/resource/ebs_discovery.go
```diff
@@ -0,0 +1,62 @@
...
+	ebsResourceKeyPrefix      = "ebs-volume:"
+	ScanPeriod                = 500 * time.Millisecond
+)
+
+var (
+	ErrInvalidVolumeID = errors.New("EBS volume IDs do not match")
+)
+
+type EBSDiscoveryClient struct {
+	ctx context.Context
```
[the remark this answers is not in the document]
mye956: Ah so we do use it in `ebs_discovery_linux.go` in line 47 as well as in the windows implementation.

agent/ebs/watcher.go
mye956: With the current implementation, the timer won't stop which is similar to what we have with ENI attachments IIUC. -> https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/handler/attach_eni_handler_common.go#L80C21-L80C21

Although we do reset the timer when we get another ENI attachment message. Perhaps we can follow the same behavior with EBS attachments.

ecs-agent/api/resource/ebs_discovery_linux.go
mye956: We're parsing it as JSON to make it less complicated to iterate and search through the output. This is done with the `-J` flag of `lsblk`

ecs-agent/api/resource/ebs_discovery_windows_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	testVolumeID = "vol-0a1234f340444abcd"
+	deviceName   = "/dev/sdf"
+)
+
+func TestParseExecutableOutputWithHappyPath(t *testing.T) {
+	output := fmt.Sprintf("Disk Number: 0\r\nVolume ID: vol-abcdef1234567890a\r\nDevice Name: sda1\r\n\r\nDisk Number: 1\r\nVolume ID: %s\r\nDevice Name: %s\r\n\r\n", testVolumeID, deviceName)
```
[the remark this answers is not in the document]
mye956: Hm, I think this will be a bit messy since windows new line/linebreak character  is treated as `\r\n` IIRC.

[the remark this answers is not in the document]
mye956: Perhaps we can do string addition to have this more readable.

agent/ebs/watcher.go
mye956: Ah, we'd probably also want the same behavior if the same attachment message was sent over. We start the ACK timer if it hasn't done so yet. There's a check in `StartTimer` if it hasn't been set yet so this would be idempotent.

amogh09: Thanks @mye956. I see that the timer is set up in `Initialize` method of `ResourceAttachment` too based on the expiration of the attachment loaded from the state file. LGTM!

agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/resource/ebs_discovery.go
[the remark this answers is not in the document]
amogh09: My bad, I missed that EBS Discovery has separate Linux and Windows implementations. LGTM.

ecs-agent/api/resource/ebs_discovery_windows_test.go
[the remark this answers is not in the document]
amogh09: I thought that Go would use `\r\n` delimiter on Windows but I don't have an instance handy to check that. Either way this is not a blocker.

amogh09 (approved):

agent/ebs/watcher.go
amogh09: Actually, upon second look it appears that `Initialize` method is currently not called anywhere. Let's make sure that we call it somewhere in Agent startup workflow. This can be done in a future PR. Thanks!

agent/ebs/watcher_test.go
```diff
@@ -0,0 +1,369 @@
...
+
+	assert.Len(t, taskEngineState.(*dockerstate.DockerTaskEngineState).GetAllEBSAttachments(), 1)
+	ebsAttachment, ok := taskEngineState.(*dockerstate.DockerTaskEngineState).GetEBSByVolumeId(volumeID)
+	assert.True(t, ok)
+	assert.True(t, ebsAttachment.IsAttached())
+}
+
+// TestHandleExpiredEBSAttachment tests acknowledging an expired resource attachment of type Elastic Block Stores
+// The resource attachment object should not be saved to the agent state since the expiration date is in the past.
+func TestHandleExpiredEBSAttachment(t *testing.T) {
```
fierlion, line 124: Nice to see you catch this in your tests!

agent/ebs/watcher.go
mye956: Noted, thanks Amogh!

ecs-agent/api/resource/ebs_discovery_linux.go
```diff
@@ -0,0 +1,99 @@
...
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// LsblkOutput is used to manage and track the output of `lsblk`
+type LsblkOutput struct {
+	BlockDevies []BD `json:"blockdevices"`
+}
+type BD struct {
```
Realmonia, line 31: nit non blocking: If this type is used outside of this file I would prefer to name it `BlockDevice`. It seems not to be the case though (as far as I understood)

ecs-agent/api/resource/ebs_discovery_linux.go
```diff
@@ -0,0 +1,99 @@
...
+	Name     string    `json:"name"`
+	Serial   string    `json:"serial"`
+	Children []BDChild `json:"children"`
```
Realmonia, line 34: Is it the children device never going to have their children? I did a quick search, if a logical device is created on the device there can be more layers. Is it not applicable to ebs? Should we let `Children` to be of type `[]*BD` instead?

agent/engine/dockerstate/docker_task_engine_state.go
```diff
@@ -298,13 +300,32 @@ func (state *DockerTaskEngineState) GetAllPendingEBSAttachments() []*apiresource
...
 	for _, v := range state.ebsAttachments {
-		if !v.IsSent() {
+		if !v.IsAttached() {
 			pendingEBSAttachments = append(pendingEBSAttachments, v)
 		}
 	}
 	return pendingEBSAttachments
 }
 
+// GetAllPendingEBSAttachmentsWithKey returns all the ebs volumes managed by ecs on the instance that haven't been found attached on the host as a map
```
Realmonia, line 310: xen-based instance does not have an volume id, would that be a concern?

agent/ebs/watcher.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+type EBSWatcher struct {
+	ctx        context.Context
+	cancel     context.CancelFunc
+	agentState dockerstate.TaskEngineState
+	// TODO: The ebsChangeEvent will be used to send over the state change event for EBS attachments once it's been found and mounted/resize/format.
+	ebsChangeEvent chan<- statechange.Event
+	// TODO: The dataClient will be used to save to agent's data client as well as start the ACK timer. This will be added once the data client functionality have been added
+	// dataClient     data.Client
```
Realmonia, line 34: Q: sounds like a new boltDB bucket? Can you explain what will be persisted to boltDB? What's the key of the boltDB entry? How would we keep the task to the key map in state?

Realmonia (commented):
Thanks for making this change! One question: since Xen and Nitro are both amazon internal instance types, is this feature specifically for ec2 ecosystem, how are we planning to prevent ecsanywhere instances running this as it is embedded into agent state?

agent/ebs/watcher.go
mye956: Yep, that's correct. We'll need a new BoltDB bucket for EBS attachments. As for the key, we'll probably follow suit with how we're storing our ENI attachments into BoltDB where it parse the attachment ARN to obtain the key.  -> https://github.com/aws/amazon-ecs-agent/blob/master/agent/utils/utils.go#L231

Although since volume ID is also guaranteed to be unique, this could be a good option as well. This is still relative TBD but will have follow up PR to include these changes. Thanks!

agent/engine/dockerstate/docker_task_engine_state.go
mye956: No we won't. Xen-based instances should still have a volume ID. The only issue regarding volume ID with Xen-based instances is we can't obtain it within the host instance and hence why we won't validate it for Xen-instances.

ecs-agent/api/resource/ebs_discovery_linux.go
mye956: Yep, that's correct.

ecs-agent/api/resource/resource_attachment.go
```diff
@@ -174,6 +176,22 @@ func (ra *ResourceAttachment) SetSentStatus() {
 	ra.AttachStatusSent = true
 }
 
+// IsAttached checks if the resource attachment has been found attached on the host
+func (ra *ResourceAttachment) IsAttached() bool {
+	ra.guard.RLock()
+	defer ra.guard.RUnlock()
+
+	return ra.Status == status.AttachmentAttached
```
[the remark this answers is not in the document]
mye956: Good question. It should be thread safe here since we're locking the read access for the resource attachment object and then unlocking it at the end (unless I am missing!). This should behave similar to `AttachStatusSent`.

ecs-agent/api/resource/ebs_discovery_linux.go
mye956: Ah, great point! Will update this.

ecs-agent/api/resource/resource_attachment.go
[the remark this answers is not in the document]
Realmonia: I meant the naming of `Status` here. The overall setter is thread safe but the field itself is not (other wise we are not protecting it with locks). Should we make it a private field, or change its naming with "Unsafe" in it?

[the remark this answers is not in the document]
Realmonia: Synced offline, this is not a trivial change and will be done as a follow up

Realmonia (approved):

agent/ebs/watcher.go
fierlion: are we not also saving our EFS/FSx filesystems in state?

agent/engine/dockerstate/docker_task_engine_state.go
fierlion: xen based ECS EC2 instances will not advertise the requisite capability for task attach placement.

ecs-agent/api/resource/ebs_discovery.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+)
+
+const (
+	ebsVolumeDiscoveryTimeout = 5 * time.Second
```
fierlion, line 25: Where did this timeout come from?  Based on our data, 5 seconds is not enough to always give EBS time to handle the attachment.

ecs-agent/api/resource/ebs_discovery_linux.go
fierlion: agree. Can we change the name? -- it's ok to refer to this as `bd` in method receivers.

ecs-agent/api/resource/ebs_discovery.go
mye956: Updated!

ecs-agent/api/resource/resource_attachment.go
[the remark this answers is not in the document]
mye956: Thank you!

Realmonia (approved):

ecs-agent/api/resource/ebs_discovery_linux.go
mye956: Ack, will update the name to `BlockDevice` instead. Thanks!

amogh09 (approved):

Realmonia (approved):

PR #3878, opened 2023-08-29
Model transformer: model reconciliation for agent upgrades

Realmonia:
### Summary

This adds the functionality of model transformer to agent, which detects model incompatibility between the running agent version and the data we persists in boltDB. 

**Note 1: the transformation functions and models introduced in this PR is example/template change and will never be executed!**
**Note 2: this is only for agent upgrades. Support for agent downgrades is not in scope.**

A version check (using `version.Version` in "running" agent and `agent-version` in boltDB metadata bucket) is performed before we load boltDB data to agent state. If a version difference is detected, the loaded json will go through a series of transformation functions, apply those applicable transformation functions, then reach a state that it is compatible with current running agent version then loaded into agent state.

Here, "applicable" transformation functions means that the model is older than the version we introduce a new model (and as a result, anew transformation function). We check the model version against transformation function's threshold version, if it is older (smaller version number), it will go through the transformation function.

### Implementation details

- `transformer` in `ecs-agent` module
  - `transformer` stores transformation functions registered by agent (during data client initialization), and expose method to actually execute transformation functions in a loop (through a slice of function closure).
- `model` and `transtf` in `agent` module
  - `model` contains different models we introduced that are mutually incompatible. `transtf` contains corresponding transformation functions we introduced with models.

I also included a template task transformation function which will never get executed now.

### Testing

New unit tests are added. Manual testing done: manually verified following:
- Transformation function executed if a model compatibility is detected and there is a version upgrade
- Transformation function skipped if no model compatibility is detected and there is a version upgrade
- Transformation function skipped if there is no version upgrade

**Log from manual testing**

BoltDB stored agent version is 1.75.0

```
sudo systemctl restart ecs

level=info time=2023-08-30T00:12:52Z msg="Starting Amazon ECS Agent" version="1.75.2" commit="UNKNOWN"
level=info time=2023-08-30T00:12:52Z msg="Registered transformation function with threshold 1.99.0."
level=info time=2023-08-30T00:12:52Z msg="Agent version associated with task model in boltdb 1.75.0 is below threshold 1.99.0. Transformation needed."
level=info time=2023-08-30T00:12:52Z msg="Executing transformation function with threshold 1.99.0."
level=info time=2023-08-30T00:12:52Z msg="Transform associated with version 1.99.0 finished."
```
transformation functions registered and executed, since 1.75.0 < 1.75.2 and a breaking change introduced. 
```
sudo systemctl restart ecs

level=info time=2023-08-30T00:24:23Z msg="Starting Amazon ECS Agent" version="1.75.2" commit="UNKNOWN"
level=info time=2023-08-30T00:24:23Z msg="Registered transformation function with threshold 1.99.0."
```
transformation functions registered but not executed, since it's 1.75.2 = 1.75.2, and models have been updated since last load.

Model transformer: model reconciliation for agent upgrades

### Licensing

### DIFF
--- a/agent/data/transformationfunctions/tasktf.go
+++ b/agent/data/transformationfunctions/tasktf.go
@@ -1,0 +1,69 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package transformationfunctions
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/aws/amazon-ecs-agent/agent/data/models"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/modeltransformer"
+)
+
+// RegisterTaskTransformationFunctions calls all registerTaskTransformationFunctions<x_y_z> in ascending order.
+// (from lower threshold version to higher threshold version) thresholdVersion is the version we introduce a breaking change in.
+// All versions below threshold version need to go through that specific transformation function
+func RegisterTaskTransformationFunctions(t *modeltransformer.Transformer) {
+	registerTaskTransformationFunction1_x_0(t)
+}
+
+// registerTaskTransformationFunction1_x_0 is a template RegisterTaskTransformation function.
+// It registers the transformation functions that translate the task model from models.Task_1_0_0 to models.Task_1_x_0
+// Future addition to transformation functions should follow the same pattern. This current performs noop
+// TODO: edit this function when introducing first actual transformation function, and add unit test
+func registerTaskTransformationFunction1_x_0(t *modeltransformer.Transformer) {
+	thresholdVersion := "1.0.0" // this assures it never actually gets executed
+	t.AddTaskTransformationFunctions(thresholdVersion, func(dataIn []byte) ([]byte, error) {
+		logger.Info(fmt.Sprintf("Executing transformation function with threshold %s.", thresholdVersion))
+		oldModel := models.Task_1_0_0{}
+		newModel := models.Task_1_x_0{}
+		var intermediate map[string]interface{}
+
+		// Load json to old model (so that we can capture some fields before it is deleted)
+		err := json.Unmarshal(dataIn, &oldModel)
+		if err != nil {
+			return nil, err
+		}
+
+		// Load json to intermediate model to process
+		err = json.Unmarshal(dataIn, &intermediate)
+		if err != nil {
+			return nil, err
+		}
+
+		// Actual process to process
+		delete(intermediate, "ENIs")
+		modifiedJSON, err := json.Marshal(intermediate)
+		if err != nil {
+			return nil, err
+		}
+		err = json.Unmarshal(modifiedJSON, &newModel)
+		newModel.NetworkInterfaces = oldModel.ENIs
+		dataOut, err := json.Marshal(&newModel)
+		logger.Info(fmt.Sprintf("Transform associated with version %s finished.", thresholdVersion))
+		return dataOut, err
+	})
+	logger.Info(fmt.Sprintf("Registered transformation function with threshold %s.", thresholdVersion))
+}
--- a/ecs-agent/modeltransformer/transformer.go
+++ b/ecs-agent/modeltransformer/transformer.go
@@ -1,0 +1,108 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package modeltransformer
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+)
+
+const (
+	modelTypeTask = "Task"
+)
+
+// Transformer stores transformation functions for all types of objects.
+// Transform<type> will execute a series of transformation functions to make it compatible with current agent version.
+// add<type>TransformationFunctions will add more <type> transformation functions to the transformation functions chain.
+// Add other transformation functions as needed. e.g. ContainerTransformationFunctions.
+// Add corresponding Transform<Type> and Add<Type>TransformationFunctions while adding other transformation functions.
+type Transformer struct {
+	taskTransformFunctions []*TransformFunc
+}
+
+// TransformFunc contains the threshold version string for transformation function and the transformationFunction itself.
+// During upgrade, all models from versions below threshold version should execute the transformation function.
+type TransformFunc struct {
+	version  string
+	function interface{}
+}
+
+func NewTransFormer() *Transformer {
+	t := &Transformer{}
+	return t
+}
+
+// GetNumberOfTransformationFunctions returns the number of transformation functions given a model type
+func (t *Transformer) GetNumberOfTransformationFunctions(modelType string) int {
+	switch modelType {
+	case modelTypeTask:
+		return len(t.taskTransformFunctions)
+	default:
+		return 0
+	}
+}
+
+// TransformTask executes the transformation functions when version associated with model in boltdb is below the
+func (t *Transformer) TransformTask(version string, data []byte) ([]byte, error) {
+	var err error
+	// execute transformation functions sequentially and skip those not applicable
+	for _, transformFunc := range t.taskTransformFunctions {
+		if closure, ok := transformFunc.function.(func([]byte) ([]byte, error)); ok {
+			if checkVersionSmaller(version, transformFunc.version) {
+				logger.Info(fmt.Sprintf("Agent version associated with task model in boltdb %s is below threshold %s. Transformation needed.", version, transformFunc.version))
+				data, err = closure(data)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				logger.Info(fmt.Sprintf("Agent version associated with task model in boltdb %s is bigger or equal to threshold %s. Skipping transformation.", version, transformFunc.version))
+				continue
+			}
+		} else {
+			return nil, fmt.Errorf("transformation function for version %s is not valid", transformFunc.version)
+		}
+	}
+	return data, err
+}
+
+// AddTaskTransformationFunctions adds the transformationFunction to the handling chain
+func (t *Transformer) AddTaskTransformationFunctions(version string, transformationFunc interface{}) {
+	t.taskTransformFunctions = append(t.taskTransformFunctions, &TransformFunc{
+		version:  version,
+		function: transformationFunc,
+	})
+}
+
+// IsUpgrade checks whether the load of a persisted model to running agent is an upgrade
+func (t *Transformer) IsUpgrade(runningAgentVersion, persistedAgentVersion string) bool {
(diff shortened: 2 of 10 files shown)


ecs-agent/modeltransformer/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+// Add corresponding Transform<Type> and Add<Type>TransformationFunctions while adding other transformation functions.
+type Transformer struct {
+	taskTransformFunctions []*TransformFunc
+}
+
+// TransformFunc contains the threshold version string for transformation function and the transformationFunction itself.
+// During upgrade, all models from versions below threshold version should execute the transformation function.
+type TransformFunc struct {
+	version  string
+	function interface{}
```
danehlim, line 41: Could we use a more specific type here, perhaps using Generics? That way we can provide more type information upfront and avoid type assertions such as on L64.

ecs-agent/modeltransformer/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+func (t *Transformer) GetNumberOfTransformationFunctions(modelType string) int {
+	switch modelType {
+	case modelTypeTask:
+		return len(t.taskTransformFunctions)
+	default:
+		return 0
+	}
+}
+
+// TransformTask executes the transformation functions when version associated with model in boltdb is below the
```
danehlim, line 59: Seems like this comment is incomplete.

ecs-agent/modeltransformer/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// TransformFunc contains the threshold version string for transformation function and the transformationFunction itself.
+// During upgrade, all models from versions below threshold version should execute the transformation function.
+type TransformFunc struct {
+	version  string
+	function interface{}
+}
+
+func NewTransFormer() *Transformer {
```
danehlim, line 44: nit: Prefer `NewTransformer()` over `NewTransFormer()`

ecs-agent/modeltransformer/transformer.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+// TransformTask executes the transformation functions when version associated with model in boltdb is below the
+func (t *Transformer) TransformTask(version string, data []byte) ([]byte, error) {
+	var err error
+	// execute transformation functions sequentially and skip those not applicable
+	for _, transformFunc := range t.taskTransformFunctions {
+		if closure, ok := transformFunc.function.(func([]byte) ([]byte, error)); ok {
+			if checkVersionSmaller(version, transformFunc.version) {
```
danehlim, line 65: Just curious, is there any specific reasoning or preference to call `checkVersionSmaller` here instead of `t.IsUpgrade`?

ecs-agent/modeltransformer/transformer.go
Realmonia: Yep I have considered it, but the challenge I see here is the types we would like to include in generics are defined in agent module and registered to transformer that is in ecs-agent module. Using it can cause a circular dependency

ecs-agent/modeltransformer/transformer.go
Realmonia: oops it seems to be a mis-edit. I will fix it!

ecs-agent/modeltransformer/transformer.go
Realmonia: Yep seems like a typo 🤦🏻‍♂️

ecs-agent/modeltransformer/transformer.go
Realmonia: The intent I created `IsUpgrade` is to provide utility function to code in /agent repo to check version, and there will be an `IsDowngrade` in the future; and `checkVersionSmaller` is for all version checks. I think using either is fine, and with the parameter naming in each function (`runningAgentVersion` vs `persistedAgentVersion` and `version` vs `threshold`) I prefer to use `checkVersionSmaller` here.

ecs-agent/modeltransformer/transformer.go
Realmonia: Synced offline, in previous implementation I defined transformation functions to have a signature like `func([]byte, *Task_1_0_0, *Task_1_10_0) ([]byte, error)`, but the current implementation we should be able to just use a concrete type.

danehlim (dismissed):

ecs-agent/modeltransformer/transformer.go
```diff
@@ -0,0 +1,106 @@
...
+	})
+}
+
+// IsUpgrade checks whether the load of a persisted model to running agent is an upgrade
+func (t *Transformer) IsUpgrade(runningAgentVersion, persistedAgentVersion string) bool {
+	return checkVersionSmaller(persistedAgentVersion, runningAgentVersion)
+}
+
+func checkVersionSmaller(version, threshold string) bool {
+	versionParams := strings.Split(version, ".")
```
chienhanlin, line 94: We can add input validation before line 94, and return false if an input string is invalid.

ecs-agent/modeltransformer/transformer.go
```diff
@@ -0,0 +1,106 @@
...
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+)
+
+const (
+	modelTypeTask = "Task"
+)
+
+// Transformer stores transformation functions for all types of objects.
+// Transform<type> will execute a series of transformation functions to make it compatible with current agent version.
```
chienhanlin, line 29: Could we mention downgrade is not supported here? Thanks.

ecs-agent/modeltransformer/transformer_test.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	newModel.TestFieldId = oldModel.TestFieldId
+	newModel.TestFieldContainerIds = oldModel.TestFieldContainerIds
+	newModel.TestFieldTaskVCpu, _ = strconv.Atoi(oldModel.TestFieldTaskVCpu)
+	dataOut, err := json.Marshal(&newModel)
+	return dataOut, err
+}
+
+func TestAddTaskTransformationFunctionsAndTransformTask(t *testing.T) {
```
chienhanlin, line 82: Could we add negative test cases? Thanks.

ecs-agent/modeltransformer/transformer.go
```diff
@@ -0,0 +1,106 @@
...
+		} else {
+			logger.Info(fmt.Sprintf("Agent version associated with task model in boltdb %s is bigger or equal to threshold %s. Skipping transformation.", version, transformFunc.version))
+			continue
+		}
+	}
+	return data, err
+}
+
+// AddTaskTransformationFunctions adds the transformationFunction to the handling chain
+func (t *Transformer) AddTaskTransformationFunctions(version string, transformationFunc transformationFunctionClosure) {
```
chienhanlin, line 81: Q. Should we add input validation, such as validating Agent version is not an empty string, before add it to the function list?

agent/data/transformationfunctions/tasktf.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		// Load json to intermediate model to process
+		err = json.Unmarshal(dataIn, &intermediate)
+		if err != nil {
+			return nil, err
+		}
+
+		// Actual process to process
+		delete(intermediate, "ENIs")
```
chienhanlin, line 57: Could we add a comment here to explain the reason of ENIs deletion? For instance, readers can take a look on `agent/data/models/task_models.go` as a reference. Thanks.

agent/data/transformationfunctions/tasktf.go
```diff
@@ -0,0 +1,69 @@
...
+func RegisterTaskTransformationFunctions(t *modeltransformer.Transformer) {
+	registerTaskTransformationFunction1_x_0(t)
+}
+
+// registerTaskTransformationFunction1_x_0 is a template RegisterTaskTransformation function.
+// It registers the transformation functions that translate the task model from models.Task_1_0_0 to models.Task_1_x_0
+// Future addition to transformation functions should follow the same pattern. This current performs noop
+// TODO: edit this function when introducing first actual transformation function, and add unit test
+func registerTaskTransformationFunction1_x_0(t *modeltransformer.Transformer) {
+	thresholdVersion := "1.0.0" // this assures it never actually gets executed
```
chienhanlin, line 37: nit. As it is a fixed value, should we define it as a constant in this file? 
```
const {
   thresholdVersion  = "1.0.0"
}
```

chienhanlin:
Thanks a lot for contributing this transformation logic to unblock our future & ongoing works!
In addition to the comments left on each line, I still have some questions/comments listed as follows
1. Can we update the PR template [here](https://github.com/aws/amazon-ecs-agent/blob/master/.github/PULL_REQUEST_TEMPLATE.md) to ask contributors to update the transformation files/functions if any breaking change is introduced by their PR? 
2. Have we tested it on windows, AL2 ARM, AL2023 ARM and ECS Anywhere?
3. Have we validated it with TestUpdateContainerAgent on platforms that support [update-container-agent](https://docs.aws.amazon.com/cli/latest/reference/ecs/update-container-agent.html) locally?
4. Could we add and highlight "downgrade Agent version is not supported" in the summary section?
5. As downgraded is not supported, should we create an announcement or an issue to notify customers?
Thanks!

ecs-agent/modeltransformer/transformer.go
Realmonia: I am not sure if we need an input validation here, since the write is done by agent too (in another save metadata operation)?

ecs-agent/modeltransformer/transformer.go
Realmonia: sure

ecs-agent/modeltransformer/transformer_test.go
Realmonia: What negative test cases are we looking for here? I think the core thing we want to test for transformer is to make sure correct transformation functions are applied, i.e. I think skip one, skip two and not skip is kinda the positive/negative cases

ecs-agent/modeltransformer/transformer.go
Realmonia: Similar as above, this is not a user input but the code we write, I would not think doing such validation is required? What's your thought?

agent/data/transformationfunctions/tasktf.go
Realmonia: Actually, `ENIs` is not necessarily to be removed. This is a simple solution to copy all other fields over from old model to new model through json operations. `ENIs` info is actually persisted in `NetworkInterface` field. Besides, I do have the comment like
```
// It registers the transformation functions that translate the task model from models.Task_1_0_0 to models.Task_1_x_0
```
for reasoning of the function. Not sure if that's sufficient?

Realmonia:
> Thanks a lot for contributing this transformation logic to unblock our future & ongoing works! In addition to the comments left on each line, I still have some questions/comments listed as follows
> 
> 1. Can we update the PR template [here](https://github.com/aws/amazon-ecs-agent/blob/master/.github/PULL_REQUEST_TEMPLATE.md) to ask contributors to update the transformation files/functions if any breaking change is introduced by their PR?
> 2. Have we tested it on windows, AL2 ARM, AL2023 ARM and ECS Anywhere?
> 3. Have we validated it with TestUpdateContainerAgent on platforms that support [update-container-agent](https://docs.aws.amazon.com/cli/latest/reference/ecs/update-container-agent.html) locally?
> 4. Could we add and highlight "downgrade Agent version is not supported" in the summary section?
> 5. As downgraded is not supported, should we create an announcement or an issue to notify customers?
>    Thanks!

Thanks for the input. I think one thing I am not making clear in the description is the transformation functions and models introduced here are just placeholder and will always be skipped with current implementation. Therefore I didn't do 2/3/5. I will do 1 but maybe after this one is pushed, and will do 4!

ecs-agent/modeltransformer/transformer.go
chienhanlin: I think adding a validation is a better practice, and can handle corrupted/modified data.

ecs-agent/modeltransformer/transformer.go
chienhanlin: Thanks!

ecs-agent/modeltransformer/transformer_test.go
chienhanlin: Was thinking of test cases like
* If a transformation function has a bug
* If data is corrupted 

And TransformTask() should return an error here (line 68-71 in ecs-agent/modeltransformer/transformer.go).
```
data, err = transformFunc.function(data)
    if err != nil {
        return nil, err
   }
```

ecs-agent/modeltransformer/transformer.go
Realmonia: Updated.

ecs-agent/modeltransformer/transformer_test.go
Realmonia: added

danehlim (approved):

chienhanlin (approved):

PR #3883, opened 2023-08-31
Move periodic timeout implementation to wsclient library.

RichaGangwar:
### Summary
Move the periodic timeout implemented by ACS and TACS to common websocket client library.

### Implementation details
1. Added the disconnect timer to wsclient library.
2. Removed the disconnect timer from TACS handler.
3. Remove the disconnect timer from ACS session. 

### Testing

Move the periodic timeout implemented by ACS and TACS to common websocket client library. 

### Licensing


ecs-agent/wsclient/client.go
```diff
@@ -565,3 +589,42 @@ func permissibleCloseCode(err error) bool {
...
+		"URL":                    cs.URL,
+		"ConnectTime":            startTime.Format(dateTimeFormat),
+		"ExpectedDisconnectTime": startTime.Add(disconnectTimeout).Format(dateTimeFormat),
+	})
+
+	timer := time.AfterFunc(maxConnectionDuration, func() {
+		err := cs.CloseClient(startTime, maxConnectionDuration)
+		cs.MetricsFactory.New(metricName).Done(err)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Attempted disconnecting; client already closed. %s", err))
```
danehlim, line 611: This seems like a duplicate log statement of the one on L626. Can we remove this and rely on the L626 log statement instead?

ecs-agent/wsclient/client_test.go
```diff
@@ -422,3 +445,33 @@ func TestCtxCancel(t *testing.T) {
...
+	timer, err := cs.Connect(mockDisconnectTimeoutMetricName, 10*time.Second, 2*time.Second)
+	require.NoError(t, err)
+	defer timer.Stop()
+	assert.True(t, cs.IsReady(), "expected websocket connection to be ready")
+
+	go func() {
+		messageError <- cs.ConsumeMessages(ctx)
+		cs.Close()
+	}()
+
```
danehlim, line 472: I'm thinking we could probably make this test stronger by starting a timer here, stopping the timer once we've received a message on the `errChan` and `messageError` channels, and then asserting that the duration of the timer >= max disconnect time. 

Currently the test asserts that the connection was closed but would still pass even if it was closed right away (instead of in 10-12 seconds which is what we should expect).

ecs-agent/wsclient/client.go
```diff
@@ -565,3 +589,42 @@ func permissibleCloseCode(err error) bool {
...
+
+// closeClient will attempt to close the provided client, retries are not recommended
+// as failure modes for this are when client is not found or already closed.
+func (cs *ClientServerImpl) CloseClient(startTime time.Time, timeoutDuration time.Duration) error {
+	logger.Warn(("Closing connection"), logger.Fields{
+		"ConnectionStartTime":  startTime.Format(dateTimeFormat),
+		"MaxDisconnectionTime": startTime.Add(timeoutDuration).Format(dateTimeFormat),
+	})
+	err := cs.WriteCloseMessage()
+	if err != nil {
```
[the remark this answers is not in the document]
RichaGangwar: When i looked through the code for writecontrol, there seems to be x number of possible errors in the workflow. It is impossible to capture all of them. I would say, since we are not doing any error check here, but just logging it and returning the error, we should probably remove the log statement "client already closed" and let the error say for itself.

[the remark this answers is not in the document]
danehlim: Sure sounds good! If possible, let's use logger.Fields here as well.

ecs-agent/wsclient/client.go
```diff
@@ -565,3 +589,42 @@ func permissibleCloseCode(err error) bool {
...
+			logger.Warn(fmt.Sprintf("Attempted disconnecting; client already closed. %s", err))
+		}
+	})
+	return timer
+}
+
+// closeClient will attempt to close the provided client, retries are not recommended
+// as failure modes for this are when client is not found or already closed.
+func (cs *ClientServerImpl) CloseClient(startTime time.Time, timeoutDuration time.Duration) error {
+	logger.Warn(("Closing connection"), logger.Fields{
```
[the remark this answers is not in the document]
danehlim: Original L628 (now L626) is still missing URL information. 

nit: Also prefer to use `logger.Fields` on original L626 (now L624) in favor of `fmt.Sprintf`

ecs-agent/wsclient/client_test.go
```diff
@@ -412,13 +432,55 @@ func TestCtxCancel(t *testing.T) {
...
+	// Using time difference to ensure that the disconnect was
+	go func() {
+		startTime := time.Now()
+		messageError <- cs.ConsumeMessages(ctx)
+		endTime := time.Now()
+		timeDiff := endTime.Sub(startTime)
+		assert.True(t, timeDiff > disconnectTimeout,
+			"ConsumeMessages should be alive for more than disconnectTimeout.")
+		cs.Close()
```
danehlim, line 479: Is calling `cs.Close` necessary here?

ecs-agent/wsclient/client_test.go
```diff
@@ -412,13 +432,55 @@ func TestCtxCancel(t *testing.T) {
...
+	defer timer.Stop()
+	assert.True(t, cs.IsReady(), "expected websocket connection to be ready")
+
+	// Using time difference to ensure that the disconnect was
+	go func() {
+		startTime := time.Now()
+		messageError <- cs.ConsumeMessages(ctx)
+		endTime := time.Now()
+		timeDiff := endTime.Sub(startTime)
+		assert.True(t, timeDiff > disconnectTimeout,
```
danehlim, line 477: Shouldn't this be check be >= ?

ecs-agent/wsclient/client_test.go
```diff
@@ -412,13 +432,55 @@ func TestCtxCancel(t *testing.T) {
...
+			"ConsumeMessages should be alive for more than disconnectTimeout.")
```
danehlim, line 478: I'm not sure that this statement is accurate. ConsumeMessages can be alive for the amount captured in disconnectTimeout, or potentially more (as much as disconnectTimeoutJitter longer). Perhaps a more accurate statement would be something like "ConsumeMessages should be not be closed before disconnectTimeout has elapsed"

ecs-agent/wsclient/client_test.go
RichaGangwar: Sure. Will remove cs.Close(). Its not necessary.  
For https://github.com/aws/amazon-ecs-agent/pull/3883#discussion_r1312235646, there is URL information in statements above (line: 618 and error case line 624), I don't think we need to add it to every line. I will add the logger.fields in the log statements. But since it's not a blocking comment, I would prefer to merge this change and take it up in a later PR.

ecs-agent/wsclient/client_test.go
```diff
@@ -42,7 +44,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-const dockerEndpoint = "/var/run/docker.sock"
+const (
+	dockerEndpoint                  = "/var/run/docker.sock"
```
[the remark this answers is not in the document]
RichaGangwar: This is something which was preexisting actually. But sure, let me take a look up if it impacts and if it can be something random. Since, this is not blocking, I would pick it up in a later CR, if thats okay.

ecs-agent/tcs/handler/handler_test.go
```diff
@@ -441,9 +440,6 @@ func TestClientReconnectsAfterInactiveTimeout(t *testing.T) {
 	// it would continue to reconnect and test will be in forever loop.
 	assert.False(t, websocket.IsCloseError(err, websocket.CloseAbnormalClosure),
 		"Read from closed connection should produce an io.EOF error")
-
-	assert.Equal(t, err.Error(), context.DeadlineExceeded.Error(), "Context deadline exceeded error expected.")
```
danehlim: nit (non-blocking): Just curious, is there a specific reason why this assert statement was removed from this test?

danehlim (approved):

Realmonia (approved):

ecs-agent/tcs/handler/handler_test.go
RichaGangwar: Yes, the test was flaky. The return code differs based on when the ctx is cancelled. If the context is cancelled when we are in client.consumeMessages, it results in non-nil error. But when the context is cancelled in starttelemetrysession(), it results in a nil error. Another way to test would be to test for either of the condition to be true. Let me try that out and i can re-add both the conditions in a later PR.

PR #3884, opened 2023-08-31
Add volume metrics for CSI client

xxx0624:
### Summary

This CSI client encapsulates all things needed to communicate with the CSI driver daemon. As part of the client, this method `GetVolumeMetrics` can be used to get the given EBS volume stats. With this PR, we introduce a new struct to include more stats.

### Implementation details

A new struct is added which include the `Capacity` and `Used` metrics for EBS volume.

### Testing

Run `go mod vendor` for both `agent/` and `ecs-agent/`.

Update GetVolumeMetrics in the CSI client.

### Licensing

### DIFF
--- a/ecs-agent/csiclient/csi_client.go
+++ b/ecs-agent/csiclient/csi_client.go
@@ -18,6 +18,7 @@ const (
 	PROTOCOL = "unix"
 )
 
+// csiClient encapsulates all CSI methods.
 type csiClient struct {
 	csiSocket string
 }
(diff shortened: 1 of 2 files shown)


fierlion (approved):

mye956 (commented):
Will these be used somewhere within the `/agent` directory in a future PR (or if it is already being used)? If so, we probably should also do a `go mod vendor`.

xxx0624:
> Will these be used somewhere within the `/agent` directory in a future PR (or if it is already being used)? If so, we probably should also do a `go mod vendor`.

Yes (not used for now). And I've run this for both agent/ and ecs-agent/

mye956 (approved):

PR #3886, opened 2023-09-05
Fix for gMSA s3 test

as14692:
### Summary
gMSA Linux fetches task's credentialspec from an S3 location. The S3Client session was previously created with the instance's region assuming the instance and S3 region would be the same. This resulted in failure of GMSA functional Linux test in few regions. This change modifies NewS3Client to create S3Client session with S3 bucket region. This will allow functional GMSA Linux test to run in all regions, without creating region specific test buckets and objects.

### Implementation details
* Updated NewS3Client

* Updated mocks
* Updated unit tests

### Testing
* Unit tests are modified
* Manually verified by creating a gMSA task and verifying the S3 bucket region
* Ran functional test in multiple regions (us-east-1, us-west-1)

* Add fix for the failing functional gMSA S3 test. 

### Licensing


singholt:
Thanks for making this change! This not only helps our functional tests but also has customer facing value where their ECS container instance and credential spec s3 object are not in the same region. 🎉

singholt (dismissed):

agent/taskresource/credentialspec/credentialspec_linux.go
```diff
@@ -435,7 +435,7 @@ func (cs *CredentialSpecResource) handleS3CredentialspecFile(originalCredentialS
 		return
 	}
 
-	s3Client := cs.s3ClientCreator.NewS3Client(cs.region, iamCredentials)
+	s3Client, err := cs.s3ClientCreator.NewS3Client(bucket, cs.region, iamCredentials)
```
singholt, line 438: can we handle this err here

as14692: Done!

singholt (approved):

mythri-garaga (approved):

PR #3887, opened 2023-09-05
Move ACS session to ecs-agent module and refactor

danehlim:
### Summary

Move ACS session to ecs-agent module and refactor associated code.

### Implementation details

- `agent/acs/handler/acs_handler.go` -> `ecs-agent/acs/session/session.go`
   - Various code enhancement changes and making the ACS session more general such that it makes less assumptions about the underlying Agent making use of the session. Among the most notable are:
      - Session's `Start` method and other related methods it calls now take in a `context.Context` in favor of storing a `context.Context` as a struct field in the whole session itself
      - Session's `Start` method also now returns the error code returned by the context, which provides meaningful information about why the Session ended (as opposed to current behavior, where the method always returns `nil`)
      - Pass in the following pre-initialized as session struct fields instead of determining what they are from within the session itself: `agentVersion`, `agentHash`, `dockerVersion`, `inactiveInstanceCB()` function, `PayloadMessageHandler`, `CredentialsMetadataSetter`, `ENIHandler`, `ResourceHandler`, `ManifestMessageIDAccessor`, `SequenceNumberAccessor`, `TaskComparer`, `TaskStopper`, `metrics.EntryFactory`, and `*wsclient.WSClientMinAgentConfig`
- `agent/acs/handler/acs_handler_test.go` -> `ecs-agent/acs/session/session_test.go`   
   - Update ACS session tests accordingly based off of above changes to ACS session
   - General cleanup/quality of life updates to tests (e.g., add additional tests, use assert package for test assertions instead of if statement blocks, use defined test constants instead of hard coding test values at the test level, remove initialization and usage of some unnecessary variables, update some code comments to be more clear, etc.)
- `agent/app/agent.go`
   - Make corresponding changes based off of above changes to ACS session (e.g., update what is passed into the session, how the return value of its `Start` method is handled)

### Testing

Move ACS session to ecs-agent module and refactor

### Licensing


ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+
+	// The below is required to trigger the first connection to ACS.
+	sendEmptyMessageOnChannel(connectToACS)
+
+	// Loop continuously until context is closed/canceled.
+	for {
+		select {
+		case <-connectToACS:
+			logger.Debug("Received connect to ACS message. Attempting connect to ACS")
+			// Start a session with ACS
```
timj-hh, line 183: nit: add .

agent/app/agent.go
Realmonia, line 1: nit non blocking: I feel like this file has been doing a lot of thing, adding a huge `startACSSession` function makes it more heavy. Would you think moving line 1002 - 1056 to some place under /acs/handler directory makes it simpler?

agent/app/agent.go
```diff
@@ -997,29 +999,67 @@ func (agent *ecsAgent) startACSSession(
...
-		deregisterInstanceEventStream,
-		agent.containerInstanceARN,
-		agent.credentialProvider,
-		agent.dockerClient,
+	inactiveInstanceCB := func() {
+		// If the instance is inactive (i.e., was deregistered), send an event to the event stream
+		// for the same.
+		err := deregisterInstanceEventStream.WriteToEventStream(struct{}{})
+		if err != nil {
+			seelog.Debugf("Failed to write to deregister container instance event stream, err: %v", err)
```
Realmonia, line 1007: Should we use logger for all new log statement so we have more structured logs? Is `debug` the common level for this type of error?

agent/app/agent.go
```diff
@@ -997,29 +999,67 @@ func (agent *ecsAgent) startACSSession(
...
+		}
+	}
+
+	dockerVersion, err := taskEngine.Version()
+	if err != nil {
+		if err != nil {
+			seelog.Debugf("Failed to get docker version from task engine, err: %v", err)
```
Realmonia, line 1014: Similar here, do we want `debug`? And what's the impact of missing docker version?

ecs-agent/acs/session/session_test.go
```diff
@@ -0,0 +1,1328 @@
...
+		// Start should never return unless the context is canceled.
+		ended <- true
+	}()
+
+	// WaitGroup is necessary to wait for HandleResourceAttachment to be called in separate goroutine before exiting
+	// the test.
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+	resourceHandler.EXPECT().HandleResourceAttachment(gomock.Any()).Do(func(arg0 interface{}) {
+		defer wg.Done() // decrement WaitGroup counter now that HandleResourceAttachment function has been called
```
timj-hh, line 1205: nit: comment capitalization and punctutation

ecs-agent/acs/session/session_test.go
```diff
@@ -0,0 +1,1328 @@
...
+
+var inactiveInstanceError = errors.New("InactiveInstanceException")
+var noopFunc = func() {}
+var testCreds = credentials.NewStaticCredentials("test-id", "test-secret", "test-token")
+var testMinAgentConfig = &wsclient.WSClientMinAgentConfig{
+	AcceptInsecureCert: true,
+	AWSRegion:          "us-west-2",
+	DockerEndpoint:     "unix:///var/run/docker.sock",
+	IsDocker:           true,
+}
```
timj-hh, line 185: thank you for all the robust testing! should this be moved to testconst or a common_test file to decrease the size here?

agent/app/agent.go
danehlim: > Should we use logger for all new log statement so we have more structured logs?

Sure, we can use logger defined in ecs-agent module for new log statements in this file. I will make this change in the next revision (assuming there are other blocking changes that need to be made). 

> Is debug the common level for this type of error?

Debug is the level that is currently being used for this scenario. See current corresponding code here:
https://github.com/aws/amazon-ecs-agent/blob/dev/agent/acs/handler/acs_handler.go#L186-L195

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+	// sendCredentialsURLParameterName is the name of the URL parameter
+	// in the ACS URL that is used to indicate if ACS should send
+	// credentials for all tasks on establishing the connection.
+	sendCredentialsURLParameterName = "sendCredentials"
+	inactiveInstanceExceptionPrefix = "InactiveInstanceException"
+
+	// ACS protocol version spec:
+	// 1: default protocol version
+	// 2: ACS will proactively close the connection when heartbeat ACKs are missing
+	acsProtocolVersion = 2
```
Realmonia, line 64: Just for my understanding. Is this something introduced by ACS and we need to pass in setting up connection (and we just add this to const to avoid magic number in code)? When was it introduced and was there a transition in our system? Is there a consideration that ACS does not use this by default?

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+		doctor:                         doctor,
+		eniHandler:                     eniHandler,
+		manifestMessageIDAccessor:      manifestMessageIDAccessor,
+		taskComparer:                   taskComparer,
+		sequenceNumberAccessor:         sequenceNumberAccessor,
+		taskStopper:                    taskStopper,
+		resourceHandler:                resourceHandler,
+		addUpdateRequestHandlers:       addUpdateRequestHandlers,
+		backoff:                        backoff,
+		sendCredentials:                true,
```
Realmonia, line 156: Q: is this value different for fargate?

agent/app/agent.go
danehlim: > what's the impact of missing docker version?
The impact of missing docker version is dockerVersion key and value not being present in the ACS WebSocket URL. This matches the behavior of the current corresponding code here: https://github.com/aws/amazon-ecs-agent/blob/dev/agent/acs/handler/acs_handler.go#L374-L376

> Similar here, do we want debug?
Since missing dockerVersion is not critically blocking (i.e., ACS WebSocket URL can be generated without it), I decided to use log level DEBUG here. I don't have a strong preference for this though - do you think there is a more fitting log level to use here in this case?

ecs-agent/acs/session/session_test.go
```diff
@@ -0,0 +1,1328 @@
...
+	mockWsClient.EXPECT().AddRequestHandler(gomock.Any()).AnyTimes()
+	mockWsClient.EXPECT().Connect(gomock.Any(), gomock.Any(), gomock.Any()).
+		Return(time.NewTimer(wsclient.DisconnectTimeout), nil).AnyTimes()
+	mockWsClient.EXPECT().Serve(gomock.Any()).Do(func(interface{}) {
+		if addUpdateRequestHandlersCalled {
+			cancel()
+		}
+	})
+	mockWsClient.EXPECT().WriteCloseMessage().Return(nil).AnyTimes()
+	mockWsClient.EXPECT().Close().Return(nil).AnyTimes()
```
timj-hh, line 1254: I see more or less this batch of `EXPECT()`s in most tests, could/should they be moved to another method?

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+	query := url.Values{}
+	query.Set("clusterArn", s.cluster)
+	query.Set("containerInstanceArn", s.containerInstanceARN)
+	query.Set("agentHash", s.agentHash)
+	query.Set("agentVersion", s.agentVersion)
+	query.Set("seqNum", "1")
+	query.Set("protocolVersion", strconv.Itoa(acsProtocolVersion))
+	if s.dockerVersion != "" {
+		dockerVersionValue := s.dockerVersion
+		if s.dockerVersion != "containerd" {
```
Realmonia, line 347: Huh so if I understand correctly Fargate now passes "containerd" to this? I don't have a solid idea about this now and might be an AI for future, but I feel like in case we want to get adoption metrics for containerd version this implementation can be blocking. In TACS I had an extra parameter as container runtime and check that.

ecs-agent/acs/session/session.go
danehlim: `sendCredentials` is used and behaves the same way across ECS/EC2 and ECS/Fargate launch types (similar to other struct fields such as `backoff`).

agent/app/agent.go
Realmonia: Maybe warn or error? I see the code in later part and I am indifferent, the only impact might just be adoption metrics. I would not think this will typically happen so maybe an error.

ecs-agent/acs/session/session_test.go
```diff
@@ -0,0 +1,1328 @@
...
+		heartbeatJitter:        10 * time.Millisecond,
+		disconnectTimeout:      30 * time.Millisecond,
+		disconnectJitter:       10 * time.Millisecond,
+		backoff: retry.NewExponentialBackoff(connectionBackoffMin, connectionBackoffMax,
+			connectionBackoffJitter, connectionBackoffMultiplier),
+	}
+
+	start := time.Now()
+	err := acsSession.Start(ctx)
+	assert.Equal(t, context.Canceled, err)
```
timj-hh, line 734: bonus: maybe we could assert something to make sure no more ACS data got set? Like `sendCredentials` is true on both reconnects (set in `session.go` L255)

ecs-agent/acs/session/session.go
danehlim: > Is this something introduced by ACS and we need to pass in setting up connection (and we just add this to const to avoid magic number in code)

Yes, it is introduced by ACS and we need to pass this in when setting up the connection if we want to use protocol version 2. You are correct that the const is for readability purposes and to provide some context with the surrounding comments.

> When was it introduced and was there a transition in our system?

Agent transitioned to use protocol version 2 in Agent version 1.52.0. Here is the corresponding PR: https://github.com/aws/amazon-ecs-agent/pull/2837

Protocol version 2 itself was introduced likely some time in the weeks / months prior to that transition.

> Is there a consideration that ACS does not use this by default?

ACS defaults to protocol version 1 if protocol version is not specified.

---

If you need more information about ACS protocol version, please let me know and I can point you in the right direction!

ecs-agent/acs/session/session.go
Realmonia: Cool I guess I am curious if this is ever going to change, but keeping it is fine for me.

ecs-agent/acs/session/session.go
Realmonia: Thanks! This is good enough and is very helpful!

agent/app/agent.go
danehlim: Hmm got it, I'll change to level WARN in the next revision. Thanks for bringing this up!

ecs-agent/acs/session/session.go
danehlim: It gets changed on L255 (i.e., it is `true` for the first connection and `false` for all subsequent connections).

ecs-agent/acs/session/session.go
timj-hh: that is correct, I think the future goal should be to pass in a `daemon` attribute (containerd or docker) and a corresponding `runtimeVersion`. Getting containerd version could be a follow up task but it is unclear if this data is already exposed to ACS. Putting the API link here for brevity ([Ref](https://github.com/containerd/containerd/blob/main/client.go#L754-L770))

agent/app/agent.go
danehlim: If we move it to a function in `acs/handler` directory, we would need to pass in a `*ecsAgent` to that function since `startACSSession` function is currently a method of `ecsAgent` struct. But this struct is currently private, so it would also need to be made public, which I'm not sure we want to do. If further discussions do result in us wanting go through the overhead to move it and are ok with the implications, it can probably be done. But I would call that out of scope of this pull request for now.

ecs-agent/acs/session/session.go
danehlim: Discussed offline - changes required for getting adoption metrics for containerd version in the future if we do want to are small and will be out of scope of this PR. They will be included as part of any PR in the future to include containerd version as part of the websocket URL for ACS and TACS connections.

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+					logger.Info("Interrupted waiting for reconnect delay to elapse; Expect session to close")
+				}
+			} else {
+				// No need to delay reconnect - reconnect immediately.
+				logger.Info("Reconnecting to ACS immediately without waiting")
+				sendEmptyMessageOnChannel(connectToACS)
+			}
+		case <-ctx.Done():
+			logger.Info("ACS session ended (context closed)", logger.Fields{
+				loggerfield.Reason: ctx.Err(),
```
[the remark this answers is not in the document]
danehlim: The naming is just to be more specific (i.e., "loggerfield" tells you the purpose of the field whereas "field" is very general and without prior knowledge or further inspection I wouldn't know what it's used for exactly), and I've seen this naming convention used elsewhere in our repository.

ecs-agent/acs/session/session.go
danehlim: Will do, thanks!

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+	backoffResetTimer := time.AfterFunc(
+		retry.AddJitter(s.heartbeatTimeout, s.heartbeatJitter), func() {
+			// If we do not have an error connecting and remain connected for at
+			// least 1 or so minutes, reset the backoff. This prevents disconnect
+			// errors that only happen infrequently from damaging the reconnect
+			// delay as significantly.
+			s.backoff.Reset()
+		})
+	defer backoffResetTimer.Stop()
+
```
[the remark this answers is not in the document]
danehlim: Would prefer to leave the empty line in this case, similar to how we leave an empty line between the defer statement on L298 and L300.

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+func (s *session) reconnectDelay(acsError error) (time.Duration, bool) {
+	if isInactiveInstanceError(acsError) {
+		logger.Info("Container instance is deregistered")
+		s.inactiveInstanceCB()
+		return s.inactiveInstanceReconnectDelay, true
+	}
+	if shouldReconnectWithoutBackoff(acsError) {
+		// ACS has closed the connection for valid reasons. Example: periodic disconnect.
+		// No need to wait/backoff to reconnect.
+		logger.Info("ACS WebSocket connection closed for a valid reason")
```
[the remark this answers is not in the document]
danehlim: I used level INFO since that is currently being used for this scenario. See current corresponding code here:
https://github.com/aws/amazon-ecs-agent/blob/dev/agent/acs/handler/acs_handler.go#L180

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+			s.backoff.Reset()
+		})
+	defer backoffResetTimer.Stop()
+
+	return client.Serve(ctx)
+}
+
+func (s *session) reconnectDelay(acsError error) (time.Duration, bool) {
+	if isInactiveInstanceError(acsError) {
+		logger.Info("Container instance is deregistered")
```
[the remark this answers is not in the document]
danehlim: Good question. Although level DEBUG is currently being used for this scenario, I felt this log message is similar in importance/significance to the log statement on L322. Since that log statement uses level INFO, I decided to have this log statement here use log level INFO as well.

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+	query.Set("agentVersion", s.agentVersion)
+	query.Set("seqNum", "1")
+	query.Set("protocolVersion", strconv.Itoa(acsProtocolVersion))
+	if s.dockerVersion != "" {
+		dockerVersionValue := s.dockerVersion
+		if s.dockerVersion != "containerd" {
+			dockerVersionValue = "DockerVersion: " + dockerVersionValue
+		}
+		query.Set("dockerVersion", dockerVersionValue)
+	}
```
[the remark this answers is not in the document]
danehlim: I'll update L346-L349 to be in its own method. Thanks for calling out!

ecs-agent/acs/session/session_test.go
danehlim: I typically don't capitalize or add period at the end if a code comment is in-line with a line of code (i.e., not on its own separate line), as I consider it akin to more of a "bullet point". If you have a strong preference for this though, I don't mind updating it to match your preference.

ecs-agent/acs/session/session_test.go
danehlim: No problem, thanks for taking a look at the pull request!

Typically for test constants that are used across multiple files, I move those to be in testconst package. However, since these are only used in session testing, I have left them here in the session test file (similar to other component-specific test constants in other test files under `ecs-agent/acs/session` directory).

ecs-agent/acs/session/session_test.go
danehlim: They are not the exact same across every test and can differ depending on the test. Since there are too many differences to take into account depending on the test, IMO writing such a method would add more complexity than simplicity.

ecs-agent/acs/session/session_test.go
danehlim: I don't think we need to check `sendCredentials` is as expected for this test (or every test) as this test is not testing that and we have a specific test for that (i.e., `TestSessionCorrectlySetsSendCredentials`).

agent/app/agent.go
Realmonia: Cool sounds reasonable

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,348 @@
...
+	query.Set(sendCredentialsURLParameterName, strconv.FormatBool(s.sendCredentials))
```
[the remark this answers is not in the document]
danehlim: Followed up offline, clarified that the question is that do we need the `sendCredentialsURLParameterName` variable or could we just pass in hardcoded "sendCredentials" string here directly?

While we could pass in a hardcoded string directly, I find having the variable name (as well as the [code comments surrounding the variable declaration](https://github.com/aws/amazon-ecs-agent/pull/3887/commits/c7e9aa296b271d998223c54d82984e4ceef97edb#diff-d3a79e42cf00660321f5c33e6b487ae242198f3996ac36254ea22d9e7eb9ecddR58)) to be more informative to the reader.


Without jumping into the implementation of the `.Set` method, `query.Set("sendCredentials", strconv.FormatBool(s.sendCredentials))` with no other context tells me that it takes in a string as the first parameter but doesn't really say much about what that string is used for or its purpose.

On the other hand, the below tells me the string it’s taking in is the parameter name for `sendCredentials` in the URL:
`query.Set(sendCredentialsURLParameterName, strconv.FormatBool(s.sendCredentials))`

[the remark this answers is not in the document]
danehlim: Discussed offline, will update to be something like the following instead:

```
// This indicates if ACS should send credentials for all tasks on establishing the connection.
query.Set("sendCredentials", strconv.FormatBool(s.sendCredentials))
```

ecs-agent/acs/session/session.go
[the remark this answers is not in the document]
timj-hh: ah, thank you for the clarification on these two points

ecs-agent/acs/session/session_test.go
timj-hh: no strong preference here either, just wanted to call it out

ecs-agent/acs/session/session.go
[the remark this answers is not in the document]
timj-hh: I do feel like `field` stays consistent with other shared components of ACS, and you can discern its usage as it is always coupled with `logger`.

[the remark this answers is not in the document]
danehlim: Sure, will remove the loggerfield alias in the next revision. Thanks!

ecs-agent/acs/session/session_test.go
timj-hh: that is true about `sendCredentials` being out of scopr, but should we assert that the first failed `DiscoverPollEndpoint()` in this test results in no connection to ACS? Or, that the second call results in a stable connection to ACS?

ecs-agent/acs/session/session_test.go
timj-hh: ack, this may be a good idea if adding more tests grows cumbersome later down the road

ecs-agent/acs/session/session_test.go
danehlim: We implicitly know that the second call results in a stable connection (whereas the first call didn't) since expectation of the .Connect method being called on L707-711 is met and the context is then canceled (which is asserted on L734).

ecs-agent/acs/session/session_test.go
timj-hh: I see your point, it may get somewhat difficult to follow these tests with so much testing setup but it is non-blocking.

As a follow-up, is there a reason you hardcoded an `IAMRoleCredentialsMessage` for `sampleRefreshCredentialsMessage` instead of using the hard type provided by `ecsacs`?

If not, would it be easier to use a real `ecsacs.IAMRoleCredentialsMessage` in case its struct fields change?

ecs-agent/acs/session/session_test.go
timj-hh: ahh, thank you for the explanation - It may be worth adding this as a comment in the test.

ecs-agent/acs/session/session_test.go
danehlim: > As a follow-up, is there a reason you hardcoded an IAMRoleCredentialsMessage for sampleRefreshCredentialsMessage instead of using the hard type provided by ecsacs?
> 
> If not, would it be easier to use a real ecsacs.IAMRoleCredentialsMessage in case its struct fields change?

This was pulled in from the current `agent/acs/handler/acs_handler_test.go` file. Ultimately it is because the channel used by the test fake ACS server to read in messages is a `string` channel, not a `ecsacs.IAMRoleCredentialsMessage` channel. If we were to make it a channel for a specific type of message, then tests such as `TestStartSessionHandlesAttachResourceMessages` or `TestSessionDoesntLeakGoroutines` would break since those tests needs to have the server able to read messages that are not refresh credentials messages.

ecs-agent/acs/session/session_test.go
danehlim: No worries and sure, will add a small note for that.

ecs-agent/acs/session/session_test.go
timj-hh: gotcha, thanks for explaining 👍

Realmonia (approved):

ecs-agent/acs/session/session_test.go
```diff
@@ -0,0 +1,1329 @@
...
+	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
+	mock_wsclient "github.com/aws/amazon-ecs-agent/ecs-agent/wsclient/mock"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/golang/mock/gomock"
+	"github.com/gorilla/websocket"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+)
+
+const (
```
timj-hh, line 53: is it possible to leave a comment explaining why we should use string primitives instead of ecsacs message objects?

timj-hh (approved):

ecs-agent/acs/session/session_test.go
```diff
@@ -0,0 +1,1329 @@
...
+	rolecredentials "github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
+	mock_credentials "github.com/aws/amazon-ecs-agent/ecs-agent/credentials/mocks"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/doctor"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/eventstream"
+	metricsfactory "github.com/aws/amazon-ecs-agent/ecs-agent/metrics"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry"
+	mock_retry "github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry/mock"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
+	mock_wsclient "github.com/aws/amazon-ecs-agent/ecs-agent/wsclient/mock"
+	"github.com/aws/aws-sdk-go/aws/credentials"
```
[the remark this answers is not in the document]
danehlim: There was a discussion on import ordering in a previous PR: https://github.com/aws/amazon-ecs-agent/pull/3765#discussion_r1245960754

To stay consistent with the results of that discussion, I will keep import ordering here as is.

ecs-agent/acs/session/session.go
```diff
@@ -11,13 +11,341 @@
...
+	"github.com/aws/amazon-ecs-agent/ecs-agent/api"
+	rolecredentials "github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/doctor"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/metrics"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/ttime"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
+	"github.com/aws/aws-sdk-go/aws/credentials"
```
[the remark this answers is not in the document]
danehlim: There was a discussion on import ordering in a previous PR: https://github.com/aws/amazon-ecs-agent/pull/3765#discussion_r1245960754

To stay consistent with the results of that discussion, I will keep import ordering here as is.

ecs-agent/acs/session/session_test.go
```diff
@@ -0,0 +1,1329 @@
...
+	parsed, err := url.Parse(wsurl)
+	assert.NoError(t, err, "should be able to parse URL")
+	assert.Equal(t, "/ws", parsed.Path, "wrong path")
+	assert.Equal(t, testconst.ClusterName, parsed.Query().Get("clusterArn"), "wrong cluster")
+	assert.Equal(t, testconst.ContainerInstanceARN, parsed.Query().Get("containerInstanceArn"),
+		"wrong container instance")
+	assert.Equal(t, agentVersion, parsed.Query().Get("agentVersion"), "wrong agent version")
+	assert.Equal(t, agentGitShortHash, parsed.Query().Get("agentHash"), "wrong agent hash")
+	assert.Equal(t, "DockerVersion: "+dockerVersion, parsed.Query().Get("dockerVersion"), "wrong docker version")
+	assert.Equalf(t, "true", parsed.Query().Get("sendCredentials"),
```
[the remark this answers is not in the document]
danehlim: Will take up this nit in a followup PR. Thanks!

ecs-agent/acs/session/session_test.go
danehlim: Will take up this non-blocking comment in a followup PR. Thanks!

PR #3892, opened 2023-09-08
Exclude docker from dependabot auto upgrade

Realmonia:
### Summary

This PR adds `"github.com/docker/docker"` to dependabot auto upgrade list.

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,3 +12,6 @@ updates:
     # the bot stuck at trying to upgrade it; (2) sdk change has higher risk of breaking the agent so probably safer to
     # upgrade manually.
     - dependency-name: "github.com/aws/aws-sdk-go"
+    # docker is explicitly excluded from bot auto upgrade because we need to do manual qualification and cooperate with
+    # amazon linux to release.
+    - dependency-name: "github.com/docker/docker"


Yiyuanzzz (approved):

.github/dependabot.yml
```diff
@@ -12,3 +12,6 @@ updates:
     # the bot stuck at trying to upgrade it; (2) sdk change has higher risk of breaking the agent so probably safer to
     # upgrade manually.
     - dependency-name: "github.com/aws/aws-sdk-go"
+    # docker is explicitly excluded from bot auto upgrade because we need to do manual qualification and cooperate with
```
chienhanlin, line 15: Q. Would we miss vulnerability alerts from dependabot if docker is excluded?

Realmonia: Even if we have vulnerabilities we still need to do qualification before we release it :( we should look for a better mechanism to identify vulnerability instead of dependabot. Moreover, dependabot in our repo is configured to keep dependency to latest, instead of alarming us when there is vulnerability. Say for now, we have a 24.x.x PR in our queue. It would not tell us if there is some issue with 20.x.x we are using.

singholt: We only qualify the dockerd RPM right? not the docker client updates in agent?

Realmonia: The version here is the docker sdk version which does not necessarily need to match docker client version on instance. We do want to make sure that's compatible and that is what qualification pipeline does. Closing this.

PR #3896, opened 2023-09-11
Skip Task resource accounting for Fargate 1.3.0 launch type

prateekchaudhry:
### Summary

ECS backend resource accounting can be different for Fargate (1.3.0) than for ECS/EC2 or ECS-Anywhere. This can cause issues with task resource accounting in ECS Agent, if it is not in sync with backend accounting. This behavior can be turned off for Fargate tasks and all resource availability can be delegated to Fargate backend instead.

### Implementation details

* ACS Payload has field `LaunchType` which takes value `FARGATE` for Fargate tasks. For such tasks, do not queue them up and launch them up straightaway. Their resources never get accounted in HostResourceManager.
* HostResourceManager release resources is not a problem as it is an idempotent release. If a task does not exist in HostResourceManager's accounting, this operation will be a no-op.

### Testing

* Added Integ tests to validate HostResourceManager still consumes resources for non fargate tasks, and does not consume resources for fargate tasks
* Manual testing
Both `EC2 and External` -
	- Queued 1 task after the other, enough memory available for 2nd task
		- 2nd task starts without waiting for 1st task to stopped

	- Queued 1 task, Stopped it with stop timeout of 30s, started new task with conflicted resource (CPU)
		- 2nd task starts after 1st task goes to stopped

New tests cover the changes: 
Yes

Skip Task resource accounting for Fargate 1.3.0 launch type
### Licensing

### DIFF
--- a/agent/engine/engine_unix_integ_test.go
+++ b/agent/engine/engine_unix_integ_test.go
@@ -1381,3 +1381,66 @@ func TestHostResourceManagerStopTaskNotBlockWaitingTasks(t *testing.T) {
 
 	waitFinished(t, finished, testTimeout)
 }
+
+// Test to make sure Host Resource Manager does not account when Fargate tests get started
+func TestHostResourceManagerNoAccountingFargateLaunchType(t *testing.T) {
+	testTimeout := 1 * time.Minute
+	taskEngine, done, _ := setupWithDefaultConfig(t)
+	defer done()
+
+	stateChangeEvents := taskEngine.StateChangeEvents()
+
+	taskArn := fmt.Sprintf("IntegTaskArn-%d", i)
+	testTask := createTestTask(taskArn)
+	testTask.Memory = int64(768)
+	testTask.LaunchType = "FARGATE"
+
+	// create container
+	A := createTestContainerWithImageAndName(baseImageForOS, fmt.Sprintf("A-%d", i))
+	A.EntryPoint = &entryPointForOS
+	A.Command = []string{"trap shortsleep SIGTERM; shortsleep() { sleep 6; exit 1; }; sleep 10"}
+	A.Essential = true
+	A.StopTimeout = uint(6)
+	testTask.Containers = []*apicontainer.Container{
+		A,
+	}
+
+	// Stop task payloads from ACS for the tasks
+	stopTask := createTestTask(fmt.Sprintf("IntegTaskArn-%d", i))
+	stopTask.DesiredStatusUnsafe = apitaskstatus.TaskStopped
+	stopTask.Containers = []*apicontainer.Container{}
+
+	// goroutine to schedule tasks
+	go func() {
+		taskEngine.AddTask(task)
+		time.Sleep(2 * time.Second)
+
+		// single managedTask which should have started
+		assert.Equal(t, 1, len(taskEngine.(*DockerTaskEngine).managedTasks), "exactly one task should be running")
+
+		// stopTask - stop running task, this task will go to STOPPING due to trap handler defined and STOPPED after 6s
+		taskEngine.AddTask(stopTask)
+	}()
+
+	finished := make(chan interface{})
+
+	// goroutine to verify task running order and verify assertions
+	go func() {
+		// 1st task goes to RUNNING
+		verifyContainerRunningStateChange(t, taskEngine)
+		verifyTaskIsRunning(stateChangeEvents, tasks)
+
+		time.Sleep(2500 * time.Millisecond)
+
+		// At this time, task stopTask is received, and SIGTERM sent to task
+		// but the task is still RUNNING due to trap handler
+		assert.Equal(t, apitaskstatus.TaskRunning, task.GetKnownStatus(), "fargate task known status should be RUNNING")
+		assert.Equal(t, apitaskstatus.TaskStopped, task.GetDesiredStatus(), "fargate task desired status should be STOPPED")
+		// Verify resources are properly released in host resource manager
+		assert.False(t, taskEngine.(*DockerTaskEngine).hostResourceManager.checkTaskConsumed(task.Arn), "fargate task resources should not be consumed")
+		time.Sleep(2 * time.Second)
+		close(finished)
+	}()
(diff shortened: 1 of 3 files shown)


agent/engine/engine_unix_integ_test.go
```diff
@@ -1381,3 +1381,98 @@ func TestHostResourceManagerStopTaskNotBlockWaitingTasks(t *testing.T) {
...
+			Name:       "TestHostResourceManagerRandomLaunchTypeBehavior",
+			LaunchType: "RaNdOmStrInG",
+		},
+		{
+			Name:       "TestHostResourceManagerEmptyLaunchTypeBehavior",
+			LaunchType: "",
+		},
+	}
+
+	for _, tc := range testCases {
```
amogh09, line 1413: How about running each test case as a subtest with `t.Run`?

agent/engine/engine_unix_integ_test.go
```diff
@@ -1381,3 +1381,98 @@ func TestHostResourceManagerStopTaskNotBlockWaitingTasks(t *testing.T) {
...
+		testTimeout := 1 * time.Minute
+		taskEngine, done, _ := setupWithDefaultConfig(t)
+		defer done()
+
+		stateChangeEvents := taskEngine.StateChangeEvents()
+
+		taskArn := fmt.Sprintf("IntegTaskArn")
```
amogh09, line 1420: `Sprintf` is redundant here.

agent/engine/engine_unix_integ_test.go
```diff
@@ -1381,3 +1381,98 @@ func TestHostResourceManagerStopTaskNotBlockWaitingTasks(t *testing.T) {
...
+		testTask := createTestTask(taskArn)
+		testTask.Memory = int64(768)
+		testTask.LaunchType = tc.LaunchType
+
+		// create container
+		A := createTestContainerWithImageAndName(baseImageForOS, fmt.Sprintf("A"))
```
amogh09, line 1426: `Sprintf` is redundant here. Also can we please use a more descriptive name than `A`?

agent/engine/engine_unix_integ_test.go
```diff
@@ -1381,3 +1381,98 @@ func TestHostResourceManagerStopTaskNotBlockWaitingTasks(t *testing.T) {
...
+		A.EntryPoint = &entryPointForOS
+		A.Command = []string{"trap shortsleep SIGTERM; shortsleep() { sleep 6; exit 1; }; sleep 10"}
+		A.Essential = true
+		A.StopTimeout = uint(6)
+		testTask.Containers = []*apicontainer.Container{
+			A,
+		}
+
+		// Stop task payloads from ACS for the tasks
+		stopTask := createTestTask(fmt.Sprintf("IntegTaskArn"))
```
amogh09, line 1436: `Sprintf` is redundant.

agent/engine/engine_unix_integ_test.go
```diff
@@ -1381,3 +1381,98 @@ func TestHostResourceManagerStopTaskNotBlockWaitingTasks(t *testing.T) {
...
+
+			// At this time, stopTask is received, and SIGTERM sent to task
+			// but the task is still RUNNING due to trap handler
+			assert.Equal(t, apitaskstatus.TaskRunning, testTask.GetKnownStatus(), "task known status should be RUNNING")
+			assert.Equal(t, apitaskstatus.TaskStopped, testTask.GetDesiredStatus(), "task desired status should be STOPPED")
+			// Verify resources are properly consumed in host resource manager, and not consumed for Fargate
+			if tc.LaunchType == "FARGATE" {
+				assert.False(t, taskEngine.(*DockerTaskEngine).hostResourceManager.checkTaskConsumed(testTask.Arn), "fargate task resources should not be consumed")
+			} else {
+				assert.True(t, taskEngine.(*DockerTaskEngine).hostResourceManager.checkTaskConsumed(testTask.Arn), "non fargate task resources should not be consumed")
```
amogh09, line 1470: Typo in "non fargate task resources should not be consumed"? Should be "non fargate task resources should be consumed"?

agent/engine/task_manager.go
```diff
@@ -285,8 +285,9 @@ func (mtask *managedTask) waitForHostResources() {
 		return
 	}
 
-	if !mtask.IsInternal && !mtask.engine.hostResourceManager.checkTaskConsumed(mtask.Arn) {
+	if !mtask.IsLaunchTypeFargate() && !mtask.IsInternal && !mtask.engine.hostResourceManager.checkTaskConsumed(mtask.Arn) {
```
amogh09, line 288: Is this the only change we need? The PR description says "Their resources never get accounted in HostResourceManager.". What prevents a Fargate task's resources from being accounted in HostResourceManager?

agent/engine/engine_unix_integ_test.go
prateekchaudhry: Thanks! let me update with a follow up

agent/engine/task_manager.go
prateekchaudhry: All tasks have a managedTask, and go through [overseeTask](https://github.com/aws/amazon-ecs-agent/blob/08f9dc120c9a23c3e5c7f969e48398a243066819/agent/engine/task_manager.go#L197) on start up. It is a linear path to here on receiving a Start task request on an already started instance. So this should be the only needed change.

This change prevents queuing up task to go through [waitingTaskQueue](https://github.com/aws/amazon-ecs-agent/blob/08f9dc120c9a23c3e5c7f969e48398a243066819/agent/engine/docker_task_engine.go#L392-L413) processing for Fargate tasks, and hence they will never be accounted by HostResourceManager

amogh09: Thanks!

amogh09 (approved):

yinyic (approved):

amogh09 (dismissed):

yinyic (approved):

amogh09 (approved):

PR #3897, opened 2023-09-11
Add CNI to ecs-agent/netlib/model and update dependencies

Realmonia:
### Summary

This PR introduces CNI models from Fargate agent implementation to ecs-agent module. Later ecs agent will be integrate with those models.

### Implementation details

Besides mentioned change, I also updated some dependency version to match Fargate's dependency versions. Notably:
* github.com/Microsoft/hcsshim from 0.8.25 to 0.9.9
* github.com/stretchr/testify from 1.7.0 to 1.8.1
* github.com/vishvananda/netlink from 1.1.1-0.20201029203352-d40f9887b852 to 1.2.1-beta.2
* golang.org/x/sys 0.6.0 to 0.7.0
* golang.org/x/tools 0.6.0 to 0.7.0

github.com/vishvananda/netlink upgrade to is a beta version but it has been out for a year and fargate is using it. And in our codebase, we are simply using LinkByName and LinkList methods that are not modified.

One dependency update we want to do but contains a breaking change is `github.com/containernetworking/plugins`. In the upgrade from 0.9.1 to 1.3.0, it now depends on `containerd/cgroup` 1.1.0, which contains a breaking change. It is later fixed in v3.0.2 (thanks @mythri-garaga and @singholt), but to integrate with that version we need to change non trivial code in agent. This will be done in a follow up PR.

Related links: https://github.com/aws/amazon-ecs-agent/pull/3623 https://github.com/containerd/cgroups/pull/296

`SetV2NDstPortAndDeviceName`'s function signature is changed. For Fargate to consume this, it needs to do data client operations before entering this function.

Also removed `vpctunnel_linux_integ_test` due to the different plugin path/setup in agent. For `amazon-ecs-agent`, currently integration test does not build CNI plugins, while Fargate does. These types of integration tests should be added when the code is actually integrated with agent.

### Testing

New tests cover the changes:  Yes. Added unit tests.

Note that - since the added changes are imported from Fargate agent, we do not build CNI plugins in unit/integ test; we can change this behavior (make the integ test builds cni plugins too), or move it to functional tests when we integrate with netlib but will call it out of scope for this PR. A list of missing/removed tests due to this limitation follows:
* `func (c *cniClient) Version(plugin string) (string, error)` test in `client.go`. Agent currently checks the `VERSION` value in git submodule instead of calling `--version` command using built plugin.
* `func SetV2NDstPortAndDeviceName(taskENI *networkinterface.NetworkInterface, dstPort uint16)` test in `vpctunnel_config.go` as mentioned in earlier section of this PR description.

Passed Github workflow.

Add CNI to ecs-agent/netlib/model and update dependencies

### Licensing


ecs-agent/netlib/model/ecscni/eni_config.go
```diff
@@ -0,0 +1,89 @@
...
+package ecscni
+
+import (
+	"fmt"
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/networkinterface"
+)
+
+// ENIConfig contains all the information needed to invoke the eni plugin
+type ENIConfig struct {
```
samjkon, line 23: The new version of the ENI model is renamed to NetworkInterface. Shall we make the corresponding change here as well?

ecs-agent/netlib/model/ecscni/eni_config.go
```diff
@@ -0,0 +1,89 @@
...
+	CNIConfig
+	// ENIID is the id of ec2 eni
+	ENIID string `json:"eni"`
```
samjkon, line 26: I have a feeling we can get rid of this field. I do not recall any of the CNI plugins using this, but I could be wrong.

ecs-agent/netlib/model/ecscni/nsutil_linux.go
```diff
@@ -0,0 +1,154 @@
...
+
+type netnsutil struct {
+}
+
+func NewNetNSUtil() NetNSUtil {
+	return &netnsutil{}
+}
+
+// NewNetNS create a new network namespace with given path
+func (*netnsutil) NewNetNS(nspath string) error {
```
samjkon, line 64: I understand that tests for these functions are missing in the original repo. As an improvement, can we add some tests for these here?

Realmonia: Sure I will take a look

ecs-agent/netlib/model/ecscni/eni_config.go
Realmonia: Will update!

ecs-agent/netlib/model/ecscni/eni_config.go
Realmonia: Will take a look

Realmonia: Just checked, I think the model here is specifically for `vpc-eni` plugin and there is no reference to this field in this plugin; it uses MAC instead. So I think we are safe to remove it.

ecs-agent/netlib/model/ecscni/eni_config.go
samjkon: Right I think that makes sense. I am OK keeping it as is.

ecs-agent/netlib/model/ecscni/eni_config.go
samjkon: Since we decided we are not renaming the ENIConfig object, I think this field will not look out of place. Thanks for taking a look.

ecs-agent/netlib/model/ecscni/nsutil_linux.go
Realmonia: Added!

samjkon (dismissed):

ecs-agent/netlib/model/ecscni/appmesh_config.go
```diff
@@ -0,0 +1,79 @@
...
+
+package ecscni
+
+import (
+	"fmt"
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/appmesh"
+)
+
+// AppMeshConfig contains the information needed to incoke the appmesh CNI plugin.
```
mye956, line 22: nit: is there a typo here with `incoke`?

Realmonia: should be `invoke` I think..

ecs-agent/netlib/model/ecscni/appmesh_config.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (amc *AppMeshConfig) String() string {
+	return fmt.Sprintf("%s, ignored uid: %s, ignored gid: %s, ingress port: %s, "+
+		"egress port: %s, app ports: %v, ignored egress ips: %v, ignored egress ports: %v",
+		amc.CNIConfig.String(), amc.IgnoredUID, amc.IgnoredGID,
+		amc.ProxyIngressPort, amc.ProxyEgressPort, amc.AppPorts,
+		amc.EgressIgnoredIPs, amc.EgressIgnoredPorts)
+}
+
+func (amc *AppMeshConfig) InterfaceName() string {
```
mye956, line 62: nit(non-blocking): seems like these are getters. it might make it a bit more intuitive to include a `Get` prefix but I'm also fine with keeping it as is

ecs-agent/netlib/model/ecscni/appmesh_config.go
```diff
@@ -0,0 +1,79 @@
...
+	// Not required for app mesh plugin as no particular interface is set up in this
+	// plugin. The plugin sets up some iptables filters, that's all. However, CNI requires
+	// us to set it. Setting it to "eth0" just to satisfy that constraint.
+	return "eth0"
+}
+
+func (amc *AppMeshConfig) NSPath() string {
```
mye956, line 69: nit(non-blocking): seems like these are getters. it might make it a bit more intuitive to include a `Get` prefix but I'm also fine with keeping it as is

ecs-agent/netlib/model/ecscni/netconfig.go
```diff
@@ -0,0 +1,109 @@
...
+func (cc *CNIConfig) String() string {
+	return fmt.Sprintf("ns: %s, version: %s, plugin: %s",
+		cc.NetNSPath, cc.CNISpecVersion, cc.CNIPluginName)
+}
+
+// ContainerID returns a plaintext identifier for a container. In our case we do not make use
+// of this field, although it is required to include a non-empty value for it since the
+// CNI framework enforces it. Hence we return a fixed string.
+func (cc *CNIConfig) ContainerID() string {
+	return "container-id"
```
mye956, line 70: Just curious, could we just use constants for `ContainerID` and `NetworkName` instead of having dedicated functions to return fixed values?

Realmonia: This will be further overridden in other CNI config types.

ecs-agent/netlib/model/ecscni/appmesh_config.go
mye956: Gotcha, makes sense. Thanks!

ecs-agent/netlib/model/ecscni/netconfig.go
mye956: I see, thanks!

ecs-agent/netlib/model/ecscni/appmesh_config.go
mye956: Since there will be a follow up PR, I'm happy for this to be done in that one.

mye956 (approved):

mye956 (approved):

samjkon (approved):

PR #3900, opened 2023-09-12
Add NodeStage to internal CSI implementation

fierlion:
### Summary
This adds NodeStage to the internal/minimal EBS CSI Driver.  This code comes directly from https://github.com/kubernetes-sigs/aws-ebs-csi-driver and is used to create a minimal Node-specific driver.  

### Implementation details
The driver is updated to use the `public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-csi-ebs` base image which contains all requisite mount utilities in the path while still keeping the overall driver to ~50mb in size.  The NodeStage implementation adds only the files and dependencies which are directly tied to the EBS volume Mount/format/fs resize. 

### Testing
Manual test process:
- attach an ebs volume to EC2 host
- built CSI Driver Image using Dockerfile
- started driver as an ecs-agent Managed Daemon
- called `NodeStage` from ecs-agent-based csiclient with hard coded device name and volume ID
- validated the mount succeeds with xfs filesystem type.

Update Minimal CSI Driver with NodeStage method.

### Licensing


ecs-agent/daemonimages/csidriver/go.mod
```diff
@@ -0,0 +1,31 @@
+module github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver
```
sparrc, line 1: IIUC the ecs volume plugin reuses the vendor directory of it's parent directory (ecs-init). Would it be possible to do that here so we don't have another independent vendor/ tree to maintain? I believe this would be the fourth in this repo.

Realmonia: +1, not sure why we have an extra go.mod here? Can't we just have it in ecs-agent's go.mod?

ecs-agent/daemonimages/csidriver/driver/mount.go
```diff
@@ -14,18 +14,39 @@
...
+	FormatAndMountSensitiveWithFormatOptions(source string, target string, fstype string, options []string, sensitiveOptions []string, formatOptions []string) error
+	IsCorruptedMnt(err error) bool
+	GetDeviceNameFromMount(mountPath string) (string, int, error)
+	MakeFile(path string) error
+	MakeDir(path string) error
 	PathExists(path string) (bool, error)
+	NeedResize(devicePath string, deviceMountPath string) (bool, error)
+	Unpublish(path string) error
+	Unstage(path string) error
+	NewResizeFs() (Resizefs, error)
```
Realmonia, line 41: I don't think it's very common that type A's interface requires another type B's initializer to be implemented. Also since mountutils's methods are included in the interface, can't we just use the `NewResizeFs()` there when we create a `ResizeFs`? Why do we need this wrapper that couple these two types?

ecs-agent/daemonimages/csidriver/driver/mount_linux.go
```diff
@@ -17,9 +17,79 @@
...
+}
+
+// This function is mirrored in ./sanity_test.go to make sure sanity test covered this block of code
+// Please mirror the change to func MakeFile in ./sanity_test.go
 func (m *NodeMounter) PathExists(path string) (bool, error) {
 	return mountutils.PathExists(path)
 }
+
+func (m *NodeMounter) NeedResize(devicePath string, deviceMountPath string) (bool, error) {
+	return mountutils.NewResizeFs(m.Exec).NeedResize(devicePath, deviceMountPath)
```
Realmonia, line 71: Would this mean everytime we call this function we need to initialize a new `ResizeFs`? Is it necessary?

ecs-agent/daemonimages/csidriver/driver/node.go
```diff
@@ -106,3 +311,55 @@ func (d *nodeService) NodeGetVolumeStats(ctx context.Context, req *csi.NodeGetVo
 		},
 	}, nil
 }
+
+func recheckParameter(context map[string]string, key string, fsConfigs map[string]fileSystemConfig, fsType string) (value string, err error) {
+	v, ok := context[key]
+	if ok {
+		// This check is already performed on the controller side
+		// However, because it is potentially security-sensitive, we redo it here to be safe
```
Realmonia, line 319: If they are executing the same code and nothing changed, how recheck makes it more secure?

ecs-agent/daemonimages/csidriver/driver/node.go
```diff
@@ -15,22 +15,224 @@ package driver
...
+	}()
+
+	devicePath, ok := req.PublishContext[DevicePathKey]
+	if !ok {
+		return nil, status.Error(codes.InvalidArgument, "Device path not provided")
+	}
+
+	partition := ""
+	if part, ok := volumeContext[VolumeAttributePartition]; ok {
+		if part != "0" {
```
Realmonia, line 142: Since it is a string, is it possible to have something like "00" here which actually means the same thing?

ecs-agent/daemonimages/csidriver/driver/node.go
```diff
@@ -15,22 +15,224 @@ package driver
...
+			partition = part
+		} else {
+			klog.InfoS("NodeStageVolume: invalid partition config, will ignore.", "partition", part)
+		}
+	}
+
+	source, err := d.findDevicePath(devicePath, volumeID, partition)
```
Realmonia, line 149: Can you add a comment here about if the partition is the `""`, what does that mean?

ecs-agent/daemonimages/csidriver/driver/node.go
```diff
@@ -15,22 +15,224 @@ package driver
...
+	}
+
+	exists, err := d.mounter.PathExists(target)
+	klog.InfoS("NodeStageVolume: path exists:", "exists", exists)
+	if err != nil {
+		klog.InfoS("NodeStageVolume: path exists:", "err", err)
+		msg := fmt.Sprintf("failed to check if target %q exists: %v", target, err)
+		return nil, status.Error(codes.Internal, msg)
+	}
+	// When exists is true it means target path was created but device isn't mounted.
```
Realmonia, line 161: When will this happen? Another volume had been mounted to this path and was unmounted?

Realmonia (commented):
Can you add a brief description about what "stage" is in the process of an ebs mounting?

.gitignore
```diff
@@ -9,6 +9,7 @@ _bin/
 *.orig
 /agent/version/_version.go
 /ecs-init/version/version.go
+/ecs-agent/daemonimages/csidriver/tarfiles
```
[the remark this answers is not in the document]
fierlion: these are the built daemon images -- we need to ignore them.

ecs-agent/daemonimages/csidriver/driver/node.go
mye956: nit(non-blocking): perhaps we can add a bit more context of what we're checking for here

ecs-agent/daemonimages/csidriver/driver/node_test.go
```diff
@@ -1,5 +1,5 @@
-//go:build linux && unit
```
xxx0624, line 1: shouldn't we need the unit tag?

xxx0624 (commented):
Btw do we need to make updates in Makefile - https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/daemonimages/csidriver/Makefile to somehow use the docker file? or that will be used already?

We depend on the https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/daemonimages/csidriver/Makefile to build the tar file.

ecs-agent/daemonimages/csidriver/driver/node.go
```diff
@@ -15,22 +15,224 @@ package driver
...
 	"k8s.io/klog/v2"
 
+	"github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/driver/internal"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/util"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/volume"
 )
 
+const (
+	// default file system type to be used when it is not provided
+	defaultFsType = FSTypeExt4
```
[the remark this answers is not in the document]
fierlion: good point -- updating

PR #3912, opened 2023-09-16
Add network data client to enable boltDB operations in ecs-agent module.

Realmonia:
### Summary

This PR does some refactor to existing data client, as well as creating another data client interface in ecs-agent module, to enable network library in ecs-agent module to do boltDB related operations.

### Implementation details

- Create a new data client interface (called `NetworkDataClient`) in ecs-agent module
- Add network related operations to this interface, then implement it with a concrete type `Client`. This new `Client` type includes field from previous data client type, such as `bolt.DB` and `modeltransformer.Transformer`, as well as a `DBAccessor` type, which contains `PutObject`, `GetObject`, `Walk` that is previously private to `agent/data` package
- In agent module's `Client` interface, embed `NetworkDataClient` interface in its interface definition.
- In agent module's `client` concrete type that implemented `Client`, embed `Client` type defined in `ecs-agent/data`, so `client` type does not needs to implement any method newly introduced in `NetworkDataClient`

### Testing

New tests cover the changes:  N/A. No new code besides interfaces are introduced.

Did following manual testing:
* Add following test code to `NetworkDataClient` interface in `ecs-agent/data/client.go`
```
const NetworkBucketName = "network"

type NetworkDataClient interface {
	NetworkNoopFunc() error
}

func (c *Client) NetworkNoopFunc() error {
	return c.DB.Batch(func(tx *bolt.Tx) error {
		b := tx.Bucket([]byte(NetworkBucketName))
		return c.Accessor.PutObject(b, "testKey", "testVal")
	})
}
```
* In `agent/data/client`, add `NetworkBucketName` to its bucket list
```
	buckets = []string{
		imagesBucketName,
		containersBucketName,
		tasksBucketName,
		eniAttachmentsBucketName,
		metadataBucketName,
		generaldata.NetworkBucketName,
	}
```
* Invoke `NetworkNoopFunc` in agent client setup code
```
func New(dataDir string) (Client, error) {
	...
        err = dbClient.NetworkNoopFunc()
	...
	return dbClient, nil
}
```
Then restart agent. We then see both `network` bucket and the expected key value pair shows up

image

Add network data client to enable boltDB operations in ecs-agent module.

### Licensing

### DIFF
--- a/agent/data/client.go
+++ b/agent/data/client.go
@@ -21,6 +21,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/api/task"
 	"github.com/aws/amazon-ecs-agent/agent/data/transformationfunctions"
 	"github.com/aws/amazon-ecs-agent/agent/engine/image"
+	generaldata "github.com/aws/amazon-ecs-agent/ecs-agent/data"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/modeltransformer"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/networkinterface"
 	bolt "go.etcd.io/bbolt"
@@ -97,8 +98,7 @@ type Client interface {
 
 // client implements the Client interface using boltdb as the backing data store.
 type client struct {
-	db          *bolt.DB
-	transformer *modeltransformer.Transformer
+	generaldata.Client
 }
 
 // New returns a data client that implements the Client interface with boltdb.
@@ -144,12 +144,15 @@ func setup(dataDir string) (*client, error) {
 		return nil, err
 	}
 	return &client{
-		db:          db,
-		transformer: transformer,
+		generaldata.Client{
+			Accessor:    generaldata.DBAccessor{},
+			DB:          db,
+			Transformer: transformer,
+		},
 	}, nil
 }
 
 // Close closes the boltdb connection.
 func (c *client) Close() error {
-	return c.db.Close()
+	return c.DB.Close()
 }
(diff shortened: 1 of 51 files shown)


ecs-agent/data/client.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/pkg/errors"
+	bolt "go.etcd.io/bbolt"
+)
+
+type NetworkDataClient interface {
+}
+
+type Client struct {
+	Accessor    DBAccessor
```
samjkon, line 29: What is the benefit of having `DBAccessor`? Can we make `PutObject`/`GetObject` directly methods of `Client`?

Realmonia: My initial thought is to separate the direct DB operations from the interfaces that agent will call like `SaveTask` to prevent agent direct use. Originally those methods are not exported, I was thinking of honoring this tradition. Do you feel this is reasonable?

agent/data/client.go
```diff
@@ -91,14 +92,16 @@ type Client interface {
 	// GetMetadata gets the value of a certain kind of metadata.
 	GetMetadata(string) (string, error)
 
+	// NetworkDataClient contains networking related data operations
+	generaldata.NetworkDataClient
```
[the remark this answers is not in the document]
Realmonia: Uhh yeah you are right. My original thought was as we will still need the concrete type embed to avoid multiple boltDB or dataclient instance, whether or not avoid the interface embedding here is not important. But from a testing/mocking perspective avoid embedding does make sense to me.

samjkon (approved):

sparrc (approved):

PR #3911, opened 2023-09-15
Process attachment payload message for EBS attach

xxx0624:
### Summary

This PR is to update the current attachment payload process to support the `AttachmentType` for EBS attach which will use `AmazonElasticBlockStorage` as the attachment type specifically. `AmazonElasticBlockStorage` is the only one we support so far and all other attachment types will be ignored.

### Implementation details

- A new field `AttachmentType` is added to the existing `ResourceAttachment`.
- The attachment payload responder will support to parse the `AttachmentType` from ACS and verify all required properties are present.

### Testing

New tests cover the changes: yes

### Description for the changelog

Feature - support AttachmentType for the attachment payload for EBS Attach

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #3914, opened 2023-09-20
Add default wait timeout for attachments from ACS

xxx0624:
### Summary

Add a default wait timeout for handling attachment payload message from ACS if no waitTimeout is provided in the message. This can protect Agent from waiting for the attachment ready forever. 

### Implementation details

Add a the default wait timeout in `attachment_resource_responder` when processing the attachment payload message.

### Testing

Enhancement - add a default wait timeout for attachment payload messages.

### Licensing

### DIFF
--- a/ecs-agent/acs/session/attach_resource_responder.go
+++ b/ecs-agent/acs/session/attach_resource_responder.go
@@ -33,6 +33,8 @@ import (
 
 const (
 	AttachResourceMessageName = "ConfirmAttachmentMessage"
+	// DefaultAttachmentWaitTimeoutInMs is the default timeout, 5 minutes, for handling the attachments from ACS.
+	DefaultAttachmentWaitTimeoutInMs = 300000
 )
 
 type ResourceHandler interface {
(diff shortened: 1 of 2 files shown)


ecs-agent/acs/session/attach_resource_responder.go
```diff
@@ -78,8 +80,17 @@ func (r *attachResourceResponder) handleAttachMessage(message *ecsacs.ConfirmAtt
...
-		time.Duration(aws.Int64Value(message.WaitTimeoutMs)) * time.Millisecond)
+	// Set a default wait timeout (5m) for the attachment message from ACS if not provided.
+	// For example, the attachment payload for the EBS attach might not have the property.
+	waitTimeoutMs := aws.Int64Value(message.WaitTimeoutMs)
+	if waitTimeoutMs == 0 {
+		waitTimeoutMs = DefaultAttachmentWaitTimeoutInMs
+		logger.Debug("No waitTimeoutMs is received, set to the default timeout",
+			logger.Fields{
+				"WaitTimeoutMs": waitTimeoutMs,
+			})
```
samjkon, line 91: Minor: Instead of logging the wait timeout value inside the if block, can we log it after the block?

ecs-agent/acs/session/attach_resource_responder.go
```diff
@@ -194,7 +205,7 @@ func validateAttachmentAndReturnProperties(message *ecsacs.ConfirmAttachmentMess
 		return attachmentProperties, nil
 	}
 
-	// For "EphemeralStorage" and "ElasticBlockStorage", ACS is using resourceType to indicate its attachment type.
+	// For legacy EBS volumes("EphemeralStorage" and "ElasticBlockStorage"), ACS is using resourceType to indicate its attachment type.
```
samjkon, line 208: nit: For my understanding, is it appropriate referring to these storage means as legacy? Are we planning to deprecate them at any point already?

ecs-agent/acs/session/attach_resource_responder.go
xxx0624: Yep, either looks good to me. I can move this.

ecs-agent/acs/session/attach_resource_responder.go
xxx0624: > Are we planning to deprecate them at any point already?

I don't think so. I think we will continue to support both but won't change them much.

I will confirm with @fierlion  since this comment is from this PR https://github.com/aws/amazon-ecs-agent/pull/3911/files/9597f968e447852d88a3fb7d75a7a550fba24ffb#r1327869451

ecs-agent/acs/session/attach_resource_responder.go
xxx0624: Updated. (sry forgot to have this in a separate commit)

samjkon (approved):

mye956 (approved):

ecs-agent/acs/session/attach_resource_responder.go
fierlion: non-blocking: My comment's intent was to remove the words EphemeralStorage altogether.  This is a nit but we should follow up and fix this comment.

fierlion (approved):

PR #3919, opened 2023-09-21
Fierlion/merge go mod [WIP]

fierlion:
### Summary
This adds NodeStage to the internal/minimal EBS CSI Driver. This code comes directly from https://github.com/kubernetes-sigs/aws-ebs-csi-driver and is used to create a minimal Node-specific driver.

### Implementation details
The driver is updated to use the public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-csi-ebs base image which contains all requisite mount utilities in the path while still keeping the overall driver to ~50mb in size. The NodeStage implementation adds only the files and dependencies which are directly tied to the EBS volume Mount/format/fs resize.

### Testing
Manual test process:

- attach an ebs volume to EC2 host
- built CSI Driver Image using Dockerfile
- started driver as an ecs-agent Managed Daemon
- called NodeStage from ecs-agent-based csiclient with hard coded device name and volume ID
- validated the mount succeeds with xfs filesystem type.

Update Minimal CSI Driver with NodeStage method.

### Licensing


ecs-agent/daemonimages/csidriver/driver/node_linux.go
```diff
@@ -18,10 +18,151 @@ package driver
...
+
+	if !strings.HasPrefix(resolved, "/dev") {
+		return "", fmt.Errorf("resolved symlink for %q was unexpected: %q", p, resolved)
+	}
+
+	return resolved, nil
+}
+
+// IsBlock checks if the given path is a block device
+func (d *nodeService) IsBlockDevice(fullPath string) (bool, error) {
```
xxx0624, line 156: nit: same util method here - https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/daemonimages/csidriver/util/utils_linux.go#L22

xxx0624 (approved):
LGTM

PR #3922, opened 2023-09-25
Changes to support Node Stage and Capability Advertisement

fierlion:
### Summary
In order to mount and format an attached EBS volume, the ECS Agent requires the EBS CSI Driver (which we are launching here as the first Managed Daemon).  This CSI Driver is a privileged container with access to the host's devices and filesystem.
The current changes include:
- CSI Driver definition (managed daemon import)
- CSI Driver Image Load as part of new capability advertisement: `storage.ebs-task-volume-attach`
- EBS Watcher changes to both start the CSI Driver container on-demand as the first EBS-Backed attachment arrives and also to call CSI NodeStage with that attachment's fields.
- Various necessary changes to allow the watcher to use the taskEngine to manage state events and load state of the driver

### Implementation details

The ebs-csi-driver is the first managed daemon.  We're hard-coding it in several places, and we'll need to update these to support future managed daemons -- this work is still TODO.  

We're also hard-coding the managed daemon Import, but have put that behind a validation that the CSI Driver Image .tar file is in its proper place before we add it to our daemons in state.  `storage.ebs-task-volume-attach` capability will only be advertised if this tar file is in place and its import succeeds.

The ebs watcher (/agent/ebs/watcher.go) is where the core of these new changes coalesce.  This watcher will create a DaemonTask for the CSI Driver and inject it into the agent to be managed as a task.  The watcher will then use the running task's CSI interface to stage an attached EBS volume.  This includes both mount and format, as well as resize, all of which are handled by the driver's `NodeStage`.  

Once the driver successfully completes its mount, it sends an attachment state change message.  Another TODO is unravelling the ENI from the Attachment -- currently we've got ENIs and Attachments tightly coupled and there's work to decouple these.  For now, since the attachment payload ends up just being the attachment id and its state update, we are wrapping out change up in an ENI package so the agent handles it correctly.  

### Testing
Added new tests to cover the changes to the taskEngine and the ebs watcher.  
Also ran multiple stages of manual tests:
- test to check that the EBS CSI driver Managed Daemon import/load/startup all succeed
- test to check that a hard-coded attachment payload can trigger a volume mount/format via CSI NodeStage
- test that the successful staging will send an Attachment State Change event.

Add CSI NodeStage integration for EBS Attachments

### Licensing


ecs-agent/manageddaemon/managed_daemon.go
```diff
@@ -80,11 +82,60 @@ func NewManagedDaemon(
...
 }
 
+var ImportAll = defaultImportAll
+
 // ImportAll function will parse/validate all managed daemon definitions
 // defined in /var/lib/ecs/deps/daemons and will return an array
 // of valid ManagedDeamon objects
-func ImportAll() ([]*ManagedDaemon, error) {
-	return []*ManagedDaemon{}, nil
+func defaultImportAll() ([]*ManagedDaemon, error) {
```
mye956, line 90: Just curious, is there a specific reasoning why we're using a var to refer to the this function instead of declaring it to be global and calling it directly?

agent/app/agent_capability.go
```diff
@@ -499,6 +506,40 @@ func (agent *ecsAgent) appendServiceConnectCapabilities(capabilities []*ecs.Attr
...
 
+func (agent *ecsAgent) appendEBSTaskAttachCapabilities(capabilities []*ecs.Attribute) []*ecs.Attribute {
+	// todo update to import multiple daemons and append capabilities
+	// for now load only EBS CSI Driver daemon
+	daemonDefinitions, err := md.ImportAll()
+	if err != nil {
+		logger.Error(fmt.Sprintf("Daemon import failure: %s", err))
+		return capabilities
+	}
+	if len(daemonDefinitions) == 0 {
```
mye956, line 517: Q: Will `daemonDefinitions` ever be empty AND we don't get an error from `md.ImportAll()`?

agent/engine/daemonmanager/daemon_manager_linux.go
```diff
@@ -111,7 +129,9 @@ func (dm *daemonManager) CreateDaemonTask() (*apitask.Task, error) {
 	// directly, and the object only contains healthcheck, all other fields will be written as empty/nil
 	// in the result string. This will override the configurations that comes with the container image
 	// (CMD for example)
-	rawConfig = fmt.Sprintf("{\"Healthcheck\":%s}", string(rawHealthConfig))
+	//daemonUIDString := strconv.Itoa(daemonUID)
```
mye956, line 132: nit: remove commented out code

ecs-init/config/common.go
```diff
@@ -247,6 +247,11 @@ func CgroupMountpoint() string {
 	return cgroupMountpoint
 }
 
+// MountDirectoryEBS returns the location on disk where EBS volumes will be mounted
+func MountDirectoryEBS() string {
+       return directoryPrefix + "/mnt/ecs/ebs"
```
mye956, line 252: Q: is `directoryPrefix` set somewhere? Perhaps we should add a comment of what a full directory path here for reference.

ecs-agent/manageddaemon/managed_daemon.go
fierlion: yes -- this is a pattern to allow us to override it in our testing (so we don't end up actually importing this unless we want to) see the agent/app/agent_test.go for where we override it with a dummy function.

agent/app/agent_capability.go
fierlion: yes -- I'm updating this to be a `logger.Warn` as this is possible if there are no definitions to import.

ecs-init/config/common.go
fierlion: yes it's set in release.go
/ecs-init/config/release.go:   `directoryPrefix = ""`

(this is only set if we are building with an alternate base directory)

agent/ebs/watcher.go
```diff
@@ -131,6 +175,66 @@ func (w *EBSWatcher) overrideDeviceName(foundVolumes map[string]string) {
...
+		timeoutCtx, cancelFunc := context.WithTimeout(w.ctx, nodeStageTimeout)
+		defer func() {
+			log.Infof("cancelFunc called in node stage timeout")
+			cancelFunc()
+		}()
+		select {
+		case <-timeoutCtx.Done():
+			if timeoutCtx.Err() != nil {
+				log.Warnf("Context canceled %vs", timeoutCtx.Err())
+			}
```
amogh09, line 213: Why do we need this branch? I don't see why `timeoutCtx` would ever get cancelled? Won't we be running `default` always?

agent/ebs/watcher.go
```diff
@@ -156,14 +260,19 @@ func (w *EBSWatcher) notifyAttachedEBS(volumeId string) {
 		log.Warnf("State change event has already been emitted for EBS volume: %v.", ebs.EBSToString())
 		return
 	}
-
+	// We found an EBS volume which has the expiration time set in future and
+	// needs to be acknowledged as having been 'attached' to the Instance
+	if err := w.sendEBSStateChange(ebs); err != nil {
+		// skip logging status sent error as it's redundant and doesn't really indicate a problem
+		if !strings.Contains(err.Error(), ebsStatusSentMsg) {
```
amogh09, line 267: I don't see where `sendEBSStateChange` might return an error with `ebsStatusSentMsg` as its prefix. The only error it returns is `return errors.New("ebs watcher send EBS state change: nil volume")`.

agent/ebs/watcher.go
```diff
@@ -198,3 +307,27 @@ func (w *EBSWatcher) handleEBSAckTimeout(volumeId string) {
 		w.removeEBSAttachment(volumeId)
 	}
 }
+
+func (w *EBSWatcher) sendEBSStateChange(ebsvol *apiebs.ResourceAttachment) error {
+	if ebsvol == nil {
+		return errors.New("ebs watcher send EBS state change: nil volume")
+	}
+	go w.emitEBSAttachedEvent(ebsvol)
```
amogh09, line 315: We want to call this method asynchronously because `w.taskEngine.StateChangeEvents() <- attachmentChange` might block? Not blocking this PR but have we ever considered making `taskEngine.stateChangeEvents` buffered?

agent/ebs/watcher_test.go
```diff
@@ -65,6 +68,9 @@ func TestHandleEBSAttachmentHappyCase(t *testing.T) {
 	ctx := context.Background()
 	taskEngineState := dockerstate.NewTaskEngineState()
 	mockDiscoveryClient := mock_ebs_discovery.NewMockEBSDiscovery(mockCtrl)
+	mockTaskEngine := mock_engine.NewMockTaskEngine(mockCtrl)
+	mockTaskEngine.EXPECT().GetEbsDaemonTask().Return(nil).AnyTimes()
+	mockTaskEngine.EXPECT().GetDaemonManagers().Return(nil).AnyTimes()
```
amogh09, line 73: nit: how about keeping expectations stricter by specifying the exact number of calls expected instead of `AnyTimes`?

agent/engine/daemonmanager/daemon_manager_linux.go
```diff
@@ -79,16 +90,23 @@ func (dm *daemonManager) CreateDaemonTask() (*apitask.Task, error) {
...
 	for _, mp := range dm.managedDaemon.GetMountPoints() {
-		err := mkdirAllAndChown(mp.SourceVolumeHostPath, daemonMountPermission, daemonUID, os.Getegid())
-		if err != nil {
+		if err := mkdirAllAndChown(mp.SourceVolumeHostPath, daemonMountPermission, daemonUID, os.Getegid()); err != nil {
 			return nil, err
 		}
 		var bindOptions = dockermount.BindOptions{}
 
 		if mp.PropagationShared {
 			// https://github.com/moby/moby/blob/master/api/types/mount/mount.go#L52
```
amogh09, line 99: I know this is not a part of this PR but we should use permalinks so that it's tied to a commit ID instead of branch name which points to the tip of the branch.

agent/engine/daemonmanager/daemon_manager_linux.go
```diff
@@ -79,16 +90,23 @@ func (dm *daemonManager) CreateDaemonTask() (*apitask.Task, error) {
...
-			bindOptions.Propagation = dockermount.PropagationShared
+			bindOptions = dockermount.BindOptions{Propagation: dockermount.PropagationShared}
+			logger.Info(fmt.Sprintf("inside Propagation: %s", bindOptions.Propagation),
```
amogh09, line 101: "inside Propagation" reads like a debug log. How about we either make this a DEBUG level log or add a bit more context to the log if we want to keep it at INFO level?

agent/engine/daemonmanager/daemon_manager_linux.go
```diff
@@ -111,7 +129,8 @@ func (dm *daemonManager) CreateDaemonTask() (*apitask.Task, error) {
 	// directly, and the object only contains healthcheck, all other fields will be written as empty/nil
 	// in the result string. This will override the configurations that comes with the container image
 	// (CMD for example)
-	rawConfig = fmt.Sprintf("{\"Healthcheck\":%s}", string(rawHealthConfig))
+	rawConfig = fmt.Sprintf("{\"Healthcheck\":%s, \"User\":\"0\"}", string(rawHealthConfig))
```
amogh09, line 132: How about using the `daemonUID` constant here instead of hardcoding `0`?

agent/engine/daemonmanager/daemon_manager_linux.go
```diff
@@ -111,7 +129,8 @@ func (dm *daemonManager) CreateDaemonTask() (*apitask.Task, error) {
...
+	logger.Info(fmt.Sprintf("Raw Config: %s", rawConfig))
```
amogh09, line 133: Can we either make this a DEBUG log or add some more context to it if we want to keep it INFO?

agent/engine/docker_task_engine.go
```diff
@@ -1140,36 +1142,11 @@ func (engine *DockerTaskEngine) AddTask(task *apitask.Task) {
...
-					field.TaskID: task.GetID(),
-					field.Error:  err,
-				})
-				task.SetKnownStatus(apitaskstatus.TaskStopped)
-				task.SetDesiredStatus(apitaskstatus.TaskStopped)
-				engine.emitTaskEvent(task, err.Error())
-				return
-			}
-		}
+		logger.Info("Task is EBS TaskAttachEnabled")
```
amogh09, line 1149: I think we should add the task ARN to this log otherwise it's not clear which task is EBS enabled.

agent/ebs/watcher.go
```diff
@@ -112,6 +134,28 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
...
+	if w.taskEngine.GetEbsDaemonTask() != nil {
+		log.Infof("engine ebs CSI driver is running with taskID: %v", w.taskEngine.GetEbsDaemonTask().GetID())
+	} else {
+		if ebsCsiDaemonManager, ok := w.taskEngine.GetDaemonManagers()["ebs-csi-driver"]; ok {
+			if csiTask, err := ebsCsiDaemonManager.CreateDaemonTask(); err != nil {
+				// fail attachment and return
+				log.Errorf("Unable to start ebsCsiDaemon for task %s in the engine: error: %s", csiTask.GetID(), err)
+				csiTask.SetKnownStatus(apitaskstatus.TaskStopped)
+				csiTask.SetDesiredStatus(apitaskstatus.TaskStopped)
+				w.taskEngine.EmitTaskEvent(csiTask, err.Error())
```
amogh09, line 147: For my information, does this send an event to the backend?

ecs-agent/manageddaemon/managed_daemon.go
```diff
@@ -80,11 +82,60 @@ func NewManagedDaemon(
...
+	// TODO parse taskdef json files in parameterized dir ie /deps/daemons
+	// TODO validate that each daemon's layers are loaded or that daemon has a corresponding image tar
+	ebsCsiTarFile := filepath.Join(imageTarPath, "ebs-csi-driver", "ebs-csi-driver.tar")
+	if _, err := os.Stat(ebsCsiTarFile); err != nil {
+		return []*ManagedDaemon{}, nil
```
amogh09, line 95: Should we add a warning log here?

agent/app/agent.go
```diff
@@ -461,6 +439,12 @@ func (agent *ecsAgent) doStart(containerChangeEventStream *eventstream.EventStre
 		imageManager.AddImageToCleanUpExclusionList(agent.serviceconnectManager.GetLoadedImageName())
 	}
 
+	if csiDM, ok := agent.daemonManagers["ebs-csi-driver"]; ok {
```
amogh09, line 442: How about adding a constant for "ebs-csi-driver" as it is used in so many places?

agent/ebs/watcher.go
```diff
@@ -131,6 +175,66 @@ func (w *EBSWatcher) overrideDeviceName(foundVolumes map[string]string) {
 	}
 }
 
+// assumes CSI Driver Managed Daemon is running else call will timeout
+func (w *EBSWatcher) StageAll(foundVolumes map[string]string) error {
```
amogh09, line 179: We should add a test for this function.

agent/ebs/watcher.go
fierlion: messages on the engine's task events channel are consumed at the other end by task/attachment message handlers.  This event ends up making a call to ecs `submit-attachment-state-changes` where an attachment looks like 
{ 
         "[attachmentArn](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AttachmentStateChange.html#ECS-Type-AttachmentStateChange-attachmentArn)": "string",
         "[status](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_AttachmentStateChange.html#ECS-Type-AttachmentStateChange-status)": "string"
      }     
https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_SubmitAttachmentStateChanges.html

ecs-agent/manageddaemon/managed_daemon.go
fierlion: This is an expected result --
If there are no managed daemons tars in the import dir, then the import all should return an empty array.
Ultimately `ImportAll` will check for a <daemon>.json daemon config and a corresponding <daemon>.tar image file and will import whichever have both.

agent/ebs/watcher.go
```diff
@@ -69,6 +88,9 @@ func (w *EBSWatcher) Start() {
 			if len(pendingEBS) > 0 {
 				foundVolumes := apiebs.ScanEBSVolumes(pendingEBS, w.discoveryClient)
 				w.overrideDeviceName(foundVolumes)
+				if err := w.StageAll(foundVolumes); err != nil {
+					log.Errorf("stage error: %s", err)
```
mye956, line 92: We should probably want to skip this tick iteration if we do encounter an error. Perhaps we can add a `continue` here.

agent/ebs/watcher.go
```diff
@@ -131,6 +175,66 @@ func (w *EBSWatcher) overrideDeviceName(foundVolumes map[string]string) {
...
 }
 
+// assumes CSI Driver Managed Daemon is running else call will timeout
+func (w *EBSWatcher) StageAll(foundVolumes map[string]string) error {
+	for volID, deviceName := range foundVolumes {
+		// get volume details from attachment
+		ebsAttachment, _ := w.agentState.GetEBSByVolumeId(volID)
+		if ebsAttachment.IsSent() {
+			log.Warnf("State change event has already been emitted for EBS volume: %v.", ebsAttachment.EBSToString())
+			return nil
```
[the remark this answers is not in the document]
mye956: +1, we probably just want to skip this iteration

agent/ebs/watcher.go
```diff
@@ -131,6 +175,66 @@ func (w *EBSWatcher) overrideDeviceName(foundVolumes map[string]string) {
...
+		}
+		hostPath := ebsAttachment.GetAttachmentProperties(apiebs.SourceVolumeHostPathKey)
+		filesystemType := ebsAttachment.GetAttachmentProperties(apiebs.FileSystemTypeName)
+
+		// CSI NodeStage stub equired fields
+		stubSecrets := make(map[string]string)
+		stubVolumeContext := make(map[string]string)
+		stubMountOptions := []string{}
+		stubFsGroup, _ := strconv.ParseInt("123456", 10, 8)
```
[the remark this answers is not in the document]
mye956: +1 perhaps we can add a comment explain what these are (or it might also be helpful if they're made as a constant)

agent/ebs/watcher.go
mye956: (non-blocking question/idea! feel free to ignore if this isn't a good idea/takes too much) thinking about it now, would it also make sense if we return a list of volume IDs that were staged successfully and then pass it onto NotifyAttached? In that case we'd would log that there was a stage error for specific volumes instead if there were any errors overall (might also make it a bit easier to debug)

agent/ebs/watcher.go
```diff
@@ -131,6 +175,66 @@ func (w *EBSWatcher) overrideDeviceName(foundVolumes map[string]string) {
...
+		if ebsAttachment.IsSent() {
+			log.Warnf("State change event has already been emitted for EBS volume: %v.", ebsAttachment.EBSToString())
+			return nil
+		}
+		if ebsAttachment.HasExpired() {
+			log.Warnf("EBS status expired, no longer tracking EBS volume: %v.", ebsAttachment.EBSToString())
+			return nil
+		}
+		if ebsAttachment.IsAttached() {
+			return nil
```
mye956, line 192: nit(non-blocking): perhaps we can also log that the attachment has already been attached on a debug level

agent/engine/daemonmanager/daemon_manager_linux.go
fierlion: the user must be root in order to perform the mount. (I'd previously tested with daemonUID).  will update this with a managed daemon runas user -- that's my miss.

agent/ebs/watcher.go
fierlion: I removed this log because it gets very noisy.  adding this back and will update all to Debug level

agent/ebs/watcher.go
fierlion: another thought would be to let the attachments themselves handle this.  we update the attachment status only after successful stage and we don't send unless the attachment status is marked attached.

amogh09: Yeah I think we should process notify for the ones that succeeded.

agent/ebs/watcher.go
```diff
@@ -106,12 +124,34 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
...
+	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil {
+		log.Debugf("engine ebs CSI driver is running with taskID: %v", runningCsiTask.GetID())
+	} else {
+		if ebsCsiDaemonManager, ok := w.taskEngine.GetDaemonManagers()[md.EbsCsiDriver]; ok {
+			if csiTask, err := ebsCsiDaemonManager.CreateDaemonTask(); err != nil {
+				// fail attachment and return
+				log.Errorf("Unable to start ebsCsiDaemon for task %s in the engine: error: %s", csiTask.GetID(), err)
+				csiTask.SetKnownStatus(apitaskstatus.TaskStopped)
+				csiTask.SetDesiredStatus(apitaskstatus.TaskStopped)
+				w.taskEngine.EmitTaskEvent(csiTask, err.Error())
```
amogh09, line 143: If `ebsCsiDaemonManager.CreateDaemonTask()` fails with an error then `csiTask` shouldn't be valid and shouldn't be used IMO. It hasn't been added to the task engine so we don't need to update its status anyway, I think? 

For `EmitTaskEvent`, what information is needed? Maybe we can get it by making another method call to `ebsCsiDaemonManager`?

agent/ebs/watcher.go
amogh09: Managed daemon tasks would be internal, right? IIUC `EmitTaskEvent` does not send an event for internal tasks as the `api.NewTaskStateChangeEvent(task, reason)` returns a `ErrShouldNotSendEvent{task.Arn}` error.

https://github.com/aws/amazon-ecs-agent/blob/e446222b214ce790f4af7b7169342fce67ecf707/agent/engine/docker_task_engine.go#L970C2-L970C57

https://github.com/aws/amazon-ecs-agent/blob/e446222b214ce790f4af7b7169342fce67ecf707/agent/api/statechange.go#L117-L119

Am I missing something? How will `EmitTaskEvent` call result in an attachment change event?

agent/ebs/watcher.go
```diff
@@ -106,12 +124,34 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
...
+				return err
+			} else {
+				w.taskEngine.SetDaemonTask(md.EbsCsiDriver, csiTask)
+				w.taskEngine.AddTask(csiTask)
+				log.Infof("task_engine: Added EBS CSI task to engine")
+			}
+		} else {
+			log.Errorf("CSI Driver is not Initialized")
```
amogh09, line 151: Do we need to tell the backend that we failed to handle the EBS volume attachment?

amogh09: Or this shouldn't ever happen in practice and we just need to have some code to error the handle? I think the former?

agent/ebs/watcher.go
```diff
@@ -131,6 +171,61 @@ func (w *EBSWatcher) overrideDeviceName(foundVolumes map[string]string) {
...
+			continue
+		}
+		if ebsAttachment.IsAttached() {
+			log.Debugf("EBS status is already attached, skipping: %v.", ebsAttachment.EBSToString())
+			continue
+		}
+		hostPath := ebsAttachment.GetAttachmentProperties(apiebs.SourceVolumeHostPathKey)
+		filesystemType := ebsAttachment.GetAttachmentProperties(apiebs.FileSystemTypeName)
+
+		// CSI NodeStage stub equired fields
```
amogh09, line 194: nit: typo in "equired".

agent/ebs/watcher.go
```diff
@@ -131,6 +171,61 @@ func (w *EBSWatcher) overrideDeviceName(foundVolumes map[string]string) {
...
+		stubMountOptions := []string{}
+		// note that the numbers '123456', '10' and '8' here are dummy data
+		// we don't use the fsGroup for now
+		stubFsGroup, _ := strconv.ParseInt("123456", 10, 8)
+		publishContext := map[string]string{"devicePath": deviceName}
+		// call CSI NodeStage
+		timeoutCtx, cancelFunc := context.WithTimeout(w.ctx, nodeStageTimeout)
+		defer func() {
+			cancelFunc()
+		}()
```
amogh09, line 206: nit: we can rewrite this as `defer cancelFunc()`

agent/ebs/watcher.go
```diff
@@ -143,27 +238,28 @@ func (w *EBSWatcher) notifyAttachedEBS(volumeId string) {
 	// TODO: Add the EBS volume to data client
 	ebs, ok := w.agentState.GetEBSByVolumeId(volumeId)
 	if !ok {
-		log.Warnf("Unable to find EBS volume with volume ID: %v within agent state.", volumeId)
+		log.Debugf("Unable to find EBS volume with volume ID: %v within agent state.", volumeId)
```
amogh09, line 241: Why is this not a warning?

amogh09: Looks like an error to me.

agent/ebs/watcher.go
fierlion: (removed)

agent/ebs/watcher.go
```diff
@@ -106,12 +123,31 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
...
 	}
 
+	// start EBS CSI Driver Managed Daemon
+	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil {
+		log.Debugf("engine ebs CSI driver is running with taskID: %v", runningCsiTask.GetID())
+	} else {
+		if ebsCsiDaemonManager, ok := w.taskEngine.GetDaemonManagers()[md.EbsCsiDriver]; ok {
+			if csiTask, err := ebsCsiDaemonManager.CreateDaemonTask(); err != nil {
+				// fail attachment and return
+				log.Errorf("Unable to start ebsCsiDaemon for task %s in the engine: error: %s", csiTask.GetID(), err)
```
amogh09, line 139: `CreateDaemonTask` can return `nil` task on error and calling `csiTask.GetID()` will panic in that case.

agent/engine/docker_task_engine.go
```diff
@@ -1230,6 +1212,25 @@ func (engine *DockerTaskEngine) GetTaskByArn(arn string) (*apitask.Task, bool) {
...
+}
+
+func (engine *DockerTaskEngine) SetDaemonTask(daemonName string, task *apitask.Task) {
+	engine.daemonTasksLock.Lock()
+	defer engine.daemonTasksLock.Unlock()
+	engine.daemonTasks[daemonName] = task
+}
+
+func (engine *DockerTaskEngine) GetDaemonManagers() map[string]dm.DaemonManager {
+	return engine.daemonManagers
```
amogh09, line 1231: It is not thread-safe for the user of this method to do any read/writes on the returned map. We should return a copy and document that this method returns a copy.

agent/engine/task_manager.go
```diff
@@ -587,7 +587,7 @@ func getContainerEventLogFields(c api.ContainerStateChange) logger.Fields {
 	return f
 }
 
-func (mtask *managedTask) emitTaskEvent(task *apitask.Task, reason string) {
+func (mtask *managedTask) EmitTaskEvent(task *apitask.Task, reason string) {
```
amogh09, line 590: Why do we want to export this method? Where do we use it?

ecs-agent/csiclient/csi_client.go
```diff
@@ -30,10 +30,8 @@ import (
 const (
 	protocol        = "unix"
 	fsTypeBlockName = "block"
-
-	ImageName      = "ebs-csi-driver"
-	SocketName     = "csi-driver.sock"
-	SocketHostPath = "/var/run/ecs/"
+	SocketName      = "csi-driver.sock"
+	SocketHostPath  = "/var/run/ecs/"
```
amogh09, line 34: These are Linux specific and already defined in `csi_client_linux.go`...

ecs-agent/manageddaemon/managed_daemon.go
```diff
@@ -80,11 +82,60 @@ func NewManagedDaemon(
...
+	}
+	// found the EBS CSI tar file -- import
+	ebsManagedDaemon := NewManagedDaemon(EbsCsiDriver, "latest")
+	// add required mounts
+	ebsMounts := []*MountPoint{
+		&MountPoint{
+			SourceVolumeID:       "agentCommunicationMount",
+			SourceVolume:         "agentCommunicationMount",
+			SourceVolumeType:     "host",
+			SourceVolumeHostPath: "/var/run/ecs/ebs-csi-driver/",
```
amogh09, line 105: Probably not a blocker right now but these paths are Linux specific.

agent/engine/docker_task_engine.go
amogh09: But it's not trivial to copy `DaemonManager`... hmmm

amogh09: Since we have a thread-safe `GetDaemonTask` method maybe we can get rid of this method.

agent/ebs/watcher.go
fierlion: added the continue and a TODO to update stage all to return only successfully staged.

agent/ebs/watcher.go
fierlion: the backend has a timeout handler.  We can short-circuit that with an attachment update, but that's not part of the immediate plans.

agent/engine/docker_task_engine.go
fierlion: this is a different map -- we have a 1:1 mapping of daemon:daemonManager.  The daemonManager has a CreateDaemonTask() method which will create the DaemonTask to run.  

Presently this import only happens once as we start up and register capabilities, so thread safety isn't as much of a concern -- I could wrap this in a read lock as it will not be updated after the initial daemon manager import.  In future this could change if we allow for in-place daemon add.

agent/ebs/watcher.go
fierlion: added TODO

agent/engine/docker_task_engine.go
amogh09: ah right this is the managers map. Cool.

amogh09 (approved):

mye956 (approved):

PR #3928, opened 2023-09-26
Add AgentConfigAccessor interface to ecs-agent/

danehlim:
### Summary

Add `AgentConfigAccessor` interface to ecs-agent module. This interface abstraction allows for access to an agent config via a set of designated methods and will be used in upcoming future pull requests.

For clarity, this pull request just adds files to the ecs-agent module and does NOT change the agent module in any way to use these files.

### Implementation details

* Add `AgentConfigAccessor` interface to `ecs-agent/config/interface.go`
* Add mockgen command to generate a mock for the interface in `ecs-agent/config/generate_mocks.go`
* Generate mocks for `AgentConfigAccessor` interface 

### Testing

Add AgentConfigAccessor interface to ecs-agent/ 

### Licensing


mythri-garaga (approved):

ecs-agent/config/interface.go
```diff
@@ -0,0 +1,62 @@
...
+	// External returns whether the Agent is running on external compute
+	// capacity (i.e., outside of AWS).
+	External() bool
+	// InstanceAttributes returns a map of key/value pairs representing
+	// attributes to be associated with the instance within the ECS service.
+	// They are used to influence behavior such as launch placement.
+	InstanceAttributes() map[string]string
+	// NoInstanceIdentityDocument returns whether the Agent should register
+	// the instance with instance identity document.
+	NoInstanceIdentityDocument() bool
```
prateekchaudhry, line 41: nit: What does the prefix 'No' signify here? - `NoInstanceIdentityDocument`. Can the name be changed to something like `InstanceIdentityDocumentNeeded` which might make it more clear what the method returns? Otherwise, maybe we can update the comment above with what does it will mean when this returns `true`, i.e. if the method returns true, should Agent register with Instance Identity Document.

danehlim: The 'No' signifies the instance identity document is not available or needed (e.g., when Agent is running on external compute capacity). This mirrors the [agent config field NoIID defined here](https://github.com/aws/amazon-ecs-agent/blob/master/agent/config/types.go#L285-L289). I believe we want the 'No' in the name because the zero/default value of values of type `bool` is `false`, which matches our expectation (i.e., by default, unless otherwise specified, we would expect instance identity document to be available or needed).

I will update the comment on L39/L40 to clarify what it means when the `NoInstanceIdentityDocument` method returns true. Thanks for your suggestion!

prateekchaudhry (approved):

mythri-garaga (approved):

PR #3931, opened 2023-09-27
Implement AgentConfigAccessor interface in agent/

danehlim:
### Summary

Create struct `agentConfigAccessor` in agent module that implements `AgentConfigAccessor` interface defined in ecs-agent module. The aforementioned interface was introduced in this previous pull request: https://github.com/aws/amazon-ecs-agent/pull/3928

For clarity, this pull request adds files to the agent module but Agent does NOT currently use them in any meaningful way. These files being "plugged into power" will be done in future pull request(s) and is out of scope of this pull request. 

### Implementation details

* Add `agentConfigAccessor` struct to `agent/config/config_accessor.go` such that `agentConfigAccessor` struct implements the `AgentConfigAccessor` interface defined in ecs-agent module
* Add tests for the above to `agent/config/config_accessor_test.go`

### Testing

Implement AgentConfigAccessor interface in agent/

### Licensing

### DIFF
--- a/agent/config/config_accessor.go
+++ b/agent/config/config_accessor.go
@@ -1,0 +1,84 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package config
+
+import (
+	"errors"
+)
+
+// agentConfigAccessor struct implements AgentConfigAccessor interface defined in ecs-agent module.
+type agentConfigAccessor struct {
+	cfg *Config
+}
+
+// NewAgentConfigAccessor creates a new agentConfigAccessor.
+func NewAgentConfigAccessor(cfg *Config) (*agentConfigAccessor, error) {
+	if cfg == nil {
+		return nil, errors.New("unable to create new agent config accessor due to passed in config value being nil")
+	}
+
+	return &agentConfigAccessor{cfg: cfg}, nil
+}
+
+func (aca *agentConfigAccessor) APIEndpoint() string {
+	return aca.cfg.APIEndpoint
+}
+
+func (aca *agentConfigAccessor) AWSRegion() string {
+	return aca.cfg.AWSRegion
+}
+
+func (aca *agentConfigAccessor) Cluster() string {
+	return aca.cfg.Cluster
+}
+
+func (aca *agentConfigAccessor) DefaultClusterName() string {
+	return DefaultClusterName
+}
+
+func (aca *agentConfigAccessor) External() bool {
+	return aca.cfg.External.Enabled()
+}
+
+func (aca *agentConfigAccessor) InstanceAttributes() map[string]string {
+	return aca.cfg.InstanceAttributes
+}
+
+func (aca *agentConfigAccessor) NoInstanceIdentityDocument() bool {
+	return aca.cfg.NoIID
+}
+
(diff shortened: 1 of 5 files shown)


agent/config/config_accessor.go
```diff
@@ -0,0 +1,78 @@
...
+
+package config
+
+// agentConfigAccessor struct implements AgentConfigAccessor interface defined in ecs-agent module.
+type agentConfigAccessor struct {
+	cfg *Config
+}
+
+// NewAgentConfigAccessor creates a new agentConfigAccessor.
+func NewAgentConfigAccessor(cfg *Config) *agentConfigAccessor {
```
amogh09, line 22: Should we assert that `cfg` is not `nil` and fail early in the constructor if it is `nil`?

agent/config/config_accessor.go
```diff
@@ -0,0 +1,78 @@
...
+func (aca agentConfigAccessor) DefaultClusterName() string {
+	return DefaultClusterName
+}
+
+func (aca agentConfigAccessor) External() bool {
+	return aca.cfg.External.Enabled()
+}
+
+func (aca agentConfigAccessor) InstanceAttributes() map[string]string {
+	return aca.cfg.InstanceAttributes
```
amogh09, line 49: Should we return a copy of data structures instead of a reference to prevent user of this interface from modifying the Agent's view of the config? Or do we trust the users?

agent/config/config_accessor.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// NewAgentConfigAccessor creates a new agentConfigAccessor.
+func NewAgentConfigAccessor(cfg *Config) *agentConfigAccessor {
+	return &agentConfigAccessor{
+		cfg: cfg,
+	}
+}
+
+func (aca agentConfigAccessor) APIEndpoint() string {
```
singholt, line 28: if `NewAgentConfigAccessor` is returning `*agentConfigAccessor`, all the interface methods' signature should be

```
func (aca *agentConfigAccessor) Method() {...}
```
?

agent/config/config_accessor.go
danehlim: Will add this in. Thanks for the suggestion!

agent/config/config_accessor.go
danehlim: Thanks for calling out! I will update this.

agent/config/config_accessor.go
danehlim: Synced up offline, we would trust the users. We could make deep copies of data structures just as a fail-safe, but this comes at a cost (i.e., methods that return data structures could run at O(N) instead of O(1)). Amogh just wanted to raise the fact that any updates made by the user of the interface will reflect on the Agent’s view of the config. We acknowledge this and will intentionally leave as-is given the time complexity tradeoff and that we trust users.

agent/config/config_accessor.go
```diff
@@ -0,0 +1,84 @@
...
+func (aca *agentConfigAccessor) NoInstanceIdentityDocument() bool {
+	return aca.cfg.NoIID
+}
+
+func (aca *agentConfigAccessor) OSFamily() string {
+	return GetOSFamily()
+}
+
+func (aca *agentConfigAccessor) OSType() string {
+	return OSType
```
singholt, line 67: things that are not in the config, do we need these getter methods? or should them be in the config?

danehlim: We need these getters as these things are defined in agent module in the `agent/config/` directory. The `AgentConfigAccessor` interface will be used in ecs-agent module and that interface requires access to such agent-level information. Thus, we need this implementation to have the corresponding methods to access thest things. 

`OSType` and `GetOSFamily()` seem to be determined based off go build flags such as the ones defined [here](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/config/parse_unsupported.go#L1-L2) (as an example). Since `OSType` and `GetOSFamily()` are determined initially and shouldn't be modified after, I don't think they should be in the `Config` struct itself but rather their separate own things in the agent/config/ directory, as we would want to keep the determined values and not have those values be configurable.

amogh09 (approved):

singholt (approved):

PR #3932, opened 2023-09-27
Refactor AWS EBS CSI driver build

sparrc:
### Summary

- use multi-stage docker build file to build binary and make the container
- rework Makefile to utilize docker
- remove unused scripts and container/ dir
- use same golang version as rest of agent repo

### Testing

all Makefile targets tested to run on macos and linux

NA

### Licensing


ecs-agent/daemonimages/csidriver/Dockerfile
```diff
@@ -1,8 +1,28 @@
...
-FROM public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-csi-ebs:latest.2 AS linux-amazon
-COPY csi-driver /bin/aws-ebs-csi-driver
-ENTRYPOINT ["/bin/aws-ebs-csi-driver"]
+ARG GO_VERSION CSI_DRIVER_VERSION BUILD_DATE
+FROM public.ecr.aws/docker/library/golang:${GO_VERSION} as build
+
+WORKDIR /go/src/
+ADD . /go/src/
+
+RUN CGO_ENABLED=0 go build -ldflags "-X github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/version.version=$CSI_DRIVER_VERSION" -ldflags "-X github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/version.buildDate=$BUILD_DATE" -o /aws-ebs-csi-driver .
```
amogh09, line 20: Can we use the existing `make bin/csi-driver` target to build?

ecs-agent/daemonimages/csidriver/Dockerfile
```diff
@@ -1,8 +1,28 @@
...
 
+# use minimal eks base for csi driver image to add required mount utils
+FROM public.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-csi-ebs:latest.2
```
amogh09, line 23: Might be a stupid question but does the `latest.2` tag work for all Linux platforms - AL2 and AL2023?

ecs-agent/daemonimages/csidriver/Makefile
```diff
@@ -1,27 +1,33 @@
...
+CSI_DRIVER_VERSION="v$(shell cat CSI_DRIVER_VERSION)"
 BUILD_DATE=$(shell date -u -Iseconds)
+GO_VERSION=$(shell cat ../../../GO_VERSION)
 
-OS?=linux
-ARCH?=amd64
+all: tar
 
-.PHONY: bin/csi-driver
-bin/csi-driver:
```
amogh09, line 10: Should we leave this make target here in case developers want to build a local binary?

ecs-agent/daemonimages/csidriver/Dockerfile
sparrc: seems `latest.2` is the same as `latest-al2`: https://gallery.ecr.aws/eks-distro-build-tooling/eks-distro-minimal-base-csi-ebs

could maybe switch to latest-al2 to make it more clear

ecs-agent/daemonimages/csidriver/Dockerfile
sparrc: this is a bit complicated since the Makefile depends on the parent directories of this repo being present, and when we setup the Dockerfile we can only copy in the current directory tree (via ADD)

ecs-agent/daemonimages/csidriver/Makefile
sparrc: personally I'd prefer developers just default to the dockerfile to avoid complicated logic around handling different local developer OS and cpu architectures

sparrc: done!

ecs-agent/daemonimages/csidriver/Dockerfile
sparrc: yes done

ecs-agent/daemonimages/csidriver/Makefile
```diff
@@ -1,27 +1,39 @@
...
-	CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) go build -ldflags "\
+bin/aws-ebs-csi-driver:
+	CGO_ENABLED=0 go build -ldflags "\
 		-X \"github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/version.version=$(CSI_DRIVER_VERSION)\" \
 		-X \"github.com/aws/amazon-ecs-agent/ecs-agent/daemonimages/csidriver/version.buildDate=$(BUILD_DATE)\"" \
-		-o ./bin/csi-driver ./
+		-o bin/aws-ebs-csi-driver .
+
+image:
+	docker build --build-arg GO_VERSION=$(GO_VERSION) -t aws-ebs-csi-driver:make .
```
xxx0624, line 28: Can we use this image ref - `ebs-csi-driver:latest` which is hardcoded in fargate side?

sparrc: thx for pointing out, fixed

ecs-agent/daemonimages/csidriver/Makefile
```diff
@@ -1,27 +1,40 @@
...
+# express or implied. See the License for the specific language governing
+# permissions and limitations under the License.
+.PHONY: all image mockgen clean
+
+CSI_DRIVER_VERSION="v$(shell cat CSI_DRIVER_VERSION)"
 BUILD_DATE=$(shell date -u -Iseconds)
+GO_VERSION=$(shell cat ../../../GO_VERSION)
 
-OS?=linux
-ARCH?=amd64
```
xxx0624, line 7: We also need to support ARM

amogh09 (approved):

ecs-agent/daemonimages/csidriver/Makefile
sparrc: this is automatically determined by the build environment, we don't cross-compile our production builds

xxx0624: TIL. Ok, as you can see the arch is used here `CGO_ENABLED=0 GOOS=$(OS) GOARCH=$(ARCH) go build -ldflags`. As long as this can be detected automatically, then it should be fine. Thanks.

sparrc: Explicitly setting GOARCH I think can be risky in a docker build situation, since the `docker pull` for the eks csi base image will pull down whatever arch the build env is. So if you set GOARCH=x86_64 inside a aarch64 build env, you'd end up with an x86_64 cross-compiled binary trying to run inside an aarch64 container, which would obviously fail to ever start.

xxx0624 (approved):

PR #3941, opened 2023-10-01
ECS AppArmor Support (aka Ubuntu 22 support)

SreeeS:
### Summary

This change will introduce `ecs-default` apparmor profile which adds support for ubuntu22 platforms as with Ubuntu 22, Ubuntu defaults both to docker 20.10.x+ and CgroupsV2. Creating ECS’s cgroups require extra [DBUS](https://manpages.ubuntu.com/manpages/xenial/man5/apparmor.d.5.html)permissions in CgroupsV2. This `ecs-default` apparmor profile will provide the required permissions.

### Implementation details

The changs are in ecs-init:
* engine.go is updated with `PreStartAppArmor` which checks if the host is apparmor supported and `loadDefaultProfile` if supported
* app-armor.go is added to load the `ecs-default` profile:
  * checks of the profile is already loaded. If not it will create and write to the file and load the profile using `apparmor_parser`.
 * docker_config.go is updated to `hostConfig.SecurityOpt` if the host is apparmor supported
 * Removed `ecs-init/config/development.go` as it is no longer used by our team, it is interfering with our ecs-int development process for debugging and running ecs-init.

### Testing

* Added unit tests for engine.go and apparmor.go changes
* Run complete functional test suite to make sure there is no availability risk
* `bot/test` - tests run as part of git hub PR workflow

Add apparmor support

This PR is intended to support apparmor supported hosts. This PR can potentially break customers using a custom profile and the ecs-default profile permissions are insufficient (has been tested for availability risk). Given that currently we are not supporting apparmor, this change should not break existing customer.

### Licensing


ecs-init/apparmor/apparmor.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// LoadDefaultProfile ensures the default profile to be loaded with the given name.
+// Returns nil error if the profile is already loaded.
+func LoadDefaultProfile(profileName string) error {
+	yes, err := isProfileLoaded(profileName)
+	if err != nil {
+		return err
+	}
+	if yes {
```
danehlim, line 72: Micro-optimization: L72-L74 can be placed before L69-L71 such that in happy case only one check is performed (I assume we expect the happy case to be more common than the error case).

ecs-init/apparmor/apparmor_utils.go
```diff
@@ -0,0 +1,68 @@
...
+	"io"
+	"os"
+	"strings"
+
+	exec "golang.org/x/sys/execabs"
+)
+
+// NOTE: This code is copied from <github.com/docker/docker/profiles/apparmor>.
+//       If you plan to make any changes, please make sure they are also sent
+//       upstream.
```
danehlim, line 33: Is there any significant reason we don't just consume the upstream code directly?

If there is such a reason and we need to keep this file, do we want to add some unit tests for this file? The test coverage for it is currently at 0%.

ecs-init/config/development.go
```diff
@@ -1,51 +0,0 @@
-//go:build development
```
danehlim, line 1: What's the purpose of this file? Why are we removing it as part of this pull request?

ecs-init/apparmor/apparmor_utils.go
sparrc: The only reason we dont consume from upstream is that the functions are all private, so we can't import it

danehlim: In github.com/docker/docker/profiles/apparmor. I could only find the corresponding `isLoaded` function, which seems to be public:
https://github.com/moby/moby/blob/master/profiles/apparmor/apparmor.go#L102

Could you point to where `load` and `aaParser` are copied from? Were they copied from `github.com/docker/docker/pkg/aaparser` instead of github.com/docker/docker/profiles/apparmor? In the former I did find a function `LoadProfile` that looks like it may correspond to `load`, but that function is also public: https://github.com/moby/moby/blob/master/pkg/aaparser/aaparser.go#L30

ecs-init/config/development.go
sparrc: This file could be left in, the main problem is the "dev" make target automatically added the 'development' tag, which was also removed in this commit so I guess that pretty much solves it.

I don't mind if we just leave it though those commits are from 8.5 years ago and I've not encountered anyone using this in my 5 years on the team...

ecs-init/apparmor/apparmor_utils.go
danehlim: Ah gotcha. If that is the case, could we consume the upstream code that containerd copied from in the first place and made private? That way we remain in sync with any upstream updates/bug fixes that are made, and we can trust that any functions we use were tested upstream before being released. As a result, we would not have to maintain a mirror file nor write unit tests for the same.

SreeeS: I agree, updated to consume from:
`github.com/docker/docker/pkg/aaparser`
`github.com/docker/docker/profiles/apparmor`

ecs-init/config/development.go
SreeeS: Good call out!! Updated summary with the details.

ecs-init/apparmor/apparmor.go
```diff
@@ -0,0 +1,95 @@
...
+
+profile ecs-default flags=(attach_disconnected,mediate_deleted) {
+  #include <abstractions/base>
+
+  network inet, # Allow IPv4 traffic
+  network inet6, # Allow IPv6 traffic
+
+  capability net_admin, # Allow network configuration
+  capability sys_admin, # Allow ECS Agent to invoke the setns system call
+  capability dac_override, # Allow ECS Agent to file read, write, and execute permission
```
danehlim, line 28: Why was this line added in the recent PR revisions? This seems to make the profile overly-permissive and could unnecessarily compromise security. If the profile needs additional permissions, could we add those specific permissions instead of bypassing all file read, write, and execute permission checks? Also, for the additional permissions we do need to add (if any), do we have internal sign off for them?

SreeeS: We have added `capability dac_override` as we were seeing AppArmor denials:
`

SreeeS: > do we have internal sign off for them?

Yes

Also, we have already tried using [file access rules](https://gitlab.com/apparmor/apparmor/-/wikis/AppArmor_Core_Policy_Reference#file-access-rules) as you mentioned but they don't seem to be effective in our case. At this point we can go with this capability and improve it in future iterations.

danehlim (approved):

mythri-garaga (approved):

PR #3949, opened 2023-10-04
Add instance health check for EBS Daemon

amogh09:
### Summary

Add instance health check for EBS Daemon. 

Currently there is only one instance health check supported by ECS Agent that is the Container Runtime Health Check. This health check checks the health of Docker container runtime on the container instance. This PR introduces a new instance health check that checks the health of EBS Daemon. 

EBS Daemon Health Check sends a request to EBS Daemon to get its node service capabilities. If a successful response is received then the health status is concluded to be "OK" or "IMPAIRED" otherwise.

Note that this PR only defines the health check but does not integrate it with Agent. The integration will be done in a future PR.

### Implementation details

* Define a new health check named `ebsCSIDaemonHealthcheck` in `agent/doctor/ebs_csi_runtime_healthcheck.go` file that implements `doctor.Healthcheck` interface. The struct type is not exported. A constructor function `NewEBSCSIDaemonHealthCheck` is provided to construct the health check. The `RunCheck` method of this interface performs the health check as described in the summary above.
* A new and generic `type HealthCheckStatusTracker struct` is added that implements logic for tracking the health status of any instance health check. All health checks need to maintain the current and last status, last two health check times, and the last time when status changed. This struct implements logic to track these values. Any health checks that are added in the future can reuse this struct to get the status tracking functionality for free. However, this PR does not change the existing Docker container runtime health check to use this new type in order to reduce the scope of changes. 
* A new `EBSDaemon` health check type is added to `ecs-agent/doctor/healthcheck.go`.

### Testing

In addition to the new unit tests, I integrated the new health check with Agent locally along with additional changes to start EBS Daemon. I then checked the Agent logs and confirmed that results of the new `EBSDaemon` health check were being sent to the backend in addition to the existing health check results for Container runtime. The contents of the health check results were as expected.

Add instance health check for EBS Daemon

No

### Licensing


agent/doctor/ebs_csi_runtime_healthcheck.go
```diff
@@ -0,0 +1,70 @@
...
+		csiClient:                csiClient,
+		requestTimeout:           requestTimeout,
+		HealthCheckStatusTracker: statustracker.NewHealthCheckStatusTracker(),
+	}
+}
+
+// Performs a health check for EBS CSI Daemon by sending a request to it to get
+// node capabilities. If EBS CSI Daemon is not started yet then returns OK trivially.
+func (e *ebsCSIDaemonHealthcheck) RunCheck() doctor.HealthcheckStatus {
+	ctx, cancel := context.WithTimeout(context.Background(), e.requestTimeout)
```
fierlion, line 53: do we need to capture the timeout error here too?
```
select {
case <-ctx.Done():
    return ctx.Err()
}
```

amogh09: The RPC call should respect the context and return if the context is Done. We shouldn't have to do that ourselves. 

I did a small test to verify this. I tweaked EBS CSI Driver locally so that it sleeps for 3 seconds before responding. And then ran the test below. 
```
func TestMy(t *testing.T) {
    client := NewCSIClient("/tmp/csi-driver.sock")
    start := time.Now()
    ctx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
    defer cancel()
    res, err := client.NodeGetCapabilities(ctx)
    t.Log("elapsed", time.Now().Sub(start))
    require.NoError(t, err)
    t.Log("resp", res)
}
```

With a 10 second timeout - 
```
=== RUN   TestMy
    csi_client_linux_test.go:37: elapsed 3.00334989s
    csi_client_linux_test.go:39: resp capabilities:<rpc:<type:STAGE_UNSTAGE_VOLUME
> > capabilities:<rpc:<type:GET_VOLUME_STATS > >
--- PASS: TestMy (3.00s)
```

With a 2 second timeout - 
```
=== RUN   TestMy
1696466970351342503 [Error] logger=structured msg="Could not get EBS CSI node capabilities" error="rpc error: code = DeadlineExceeded desc = context deadline exceeded
"
    csi_client_linux_test.go:37: elapsed 2.0015858s
    csi_client_linux_test.go:38:
                Error Trace:    /home/ec2-user/go/src/github.com/aws/amazon-ecs-age
nt/ecs-agent/csiclient/csi_client_linux_test.go:38
                Error:          Received unexpected error:
                                rpc error: code = DeadlineExceeded desc = context deadline exceeded
                Test:           TestMy
--- FAIL: TestMy (2.00s)
```

With a 1 second timeout - 
```
=== RUN   TestMy
    csi_client_linux_test.go:37: elapsed 1.000499175s
    csi_client_linux_test.go:38:
                Error Trace:    /home/ec2-user/go/src/github.com/aws/amazon-ecs-agent/ecs-agent/csiclient/csi_client_linux_test.go:38
                Error:          Received unexpected error:
                                rpc error: code = DeadlineExceeded desc = context deadline exceeded
                Test:           TestMy
--- FAIL: TestMy (1.00s)
```

Note that the CSI call takes no longer than the timeout.

fierlion (approved):

mye956 (approved):

PR #3965, opened 2023-10-11
Add missing changes from EBS Task Attach feature branch

fierlion:
### Summary
This change bridges the diff between dev and https://github.com/aws/amazon-ecs-agent/tree/feature/EBSTaskAttach 
The main changes are to add the EBS resource handler to the agent, and to update the watcher's state handling.

### Implementation details
The changes came from a rebase of the feature branch against dev, taking dev as desired.   The changes to the watcher and to the task engine state include a couple of bugfixes to our EBS state handling, where previously there was a chance that a notification came before a successful NodeStage.

This change will allow us to start testing from dev branch rather than our feature branch.

### Testing
Ran tests locally to launch 5 successive EBS-backed tasks.  All tasks' EBS volume mounts succeed and tasks are started as expected.

Bugfixes for EBS Task Attach.

### Licensing

### DIFF
--- a/agent/ebs/watcher.go
+++ b/agent/ebs/watcher.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"path/filepath"
 	"strconv"
 	"time"
 
@@ -36,6 +37,7 @@ import (
 
 const (
 	nodeStageTimeout = 2 * time.Second
+	hostMountDir     = "/mnt/ecs/ebs"
 )
 
 type EBSWatcher struct {
@@ -82,11 +84,7 @@ func (w *EBSWatcher) Start() {
 			if len(pendingEBS) > 0 {
 				foundVolumes := apiebs.ScanEBSVolumes(pendingEBS, w.discoveryClient)
 				w.overrideDeviceName(foundVolumes)
-				if err := w.StageAll(foundVolumes); err != nil {
-					log.Errorf("stage error: %s", err)
-					continue
-				}
-				// TODO only notify attached for volumes that are successfully staged
+				w.StageAll(foundVolumes)
 				w.NotifyAttached(foundVolumes)
 			}
 		case <-w.ctx.Done():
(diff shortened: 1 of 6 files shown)


agent/ebs/watcher.go
```diff
@@ -84,7 +86,6 @@ func (w *EBSWatcher) Start() {
 				w.overrideDeviceName(foundVolumes)
 				if err := w.StageAll(foundVolumes); err != nil {
```
mye956, line 87: I think we can probably remove this check altogether since we'll never return an error in StageAll.

agent/ebs/watcher.go
```diff
@@ -219,14 +225,12 @@ func (w *EBSWatcher) StageAll(foundVolumes map[string]string) error {
 			&stubFsGroup)
 
 		if err != nil {
-			log.Errorf("Failed to initialize EBS volume: error: %s", err)
-			continue
```
mye956, line 223: we probably want to keep this `continue` here since we don't want to set the status to attached if we encounter an error after staging

agent/ebs/watcher.go
```diff
@@ -219,14 +225,12 @@ func (w *EBSWatcher) StageAll(foundVolumes map[string]string) error {
...
+			errors = append(errors, fmt.Errorf("Failed to initialize EBS volume ID: %v: error: %s", ebsAttachment.EBSToString(), err))
```
amogh09, line 228: nit: it's better to wrap an error with `%w` directive over `%s` as the former allows catching nested errors while maintaining the same string representation as `%s`.

agent/ebs/watcher.go
mye956: also, any issues if we keep the log call here as well? that way we don't have to do an additional check if there are any errors after the StageAll call finishes in L.88

agent/ebs/watcher.go
```diff
@@ -244,6 +248,10 @@ func (w *EBSWatcher) notifyAttachedEBS(volumeId string) {
 		log.Errorf("Unable to find EBS volume with volume ID: %v within agent state.", volumeId)
 		return
 	}
+	if !ebs.IsAttached() {
```
amogh09, line 251: `notifyAttachedEBS` might be called for a volume that's not attached? That sounds odd to me.

agent/engine/dockerstate/dockerstate_test.go
```diff
@@ -203,6 +203,52 @@ func TestAddPendingEBSAttachment(t *testing.T) {
 
 }
 
+func TestAddPendingEBSAttachmentExclusion(t *testing.T) {
```
mye956, line 206: nit: perhaps we can add a comment on what this will be testing for. a bit unclear on what we mean by exclusion here

agent/ebs/watcher.go
fierlion: Currently we're calling both "stage all" and "notify attached" for all our found attachments.  These will only execute if the attachment is in the right state -- so we only stage unstaged, un-notified volumes and we only notify on staged volumes.  The rest are no-ops.

mye956 (approved):

amogh09 (approved):

PR #3966, opened 2023-10-13
Move httpclient to ecs-agent/ and minor refactor

danehlim:
### Testing

New tests cover the changes: 

### Licensing


agent/acs/updater/updater.go
```diff
@@ -79,7 +80,7 @@ func NewUpdater(cfg *config.Config, state dockerstate.TaskEngineState, dataClien
 	taskEngine engine.TaskEngine) *updater {
 	return &updater{
 		config:     cfg,
-		httpclient: httpclient.New(updateDownloadTimeout, false),
+		httpclient: httpclient.New(updateDownloadTimeout, false, version.String(), config.OSType),
```
yinyic, line 83: nit - this may give out the impression that we are referring to http version (since this is an http client), so consider using a more specific package identifier like `agentVersion` to avoid confusion.

Same for the other occurrences

yinyic (dismissed):

hozkaya2000 (dismissed):

agent/acs/updater/updater.go
danehlim: Good point, thanks for calling this subtle point out! I will revise accordingly.

hozkaya2000 (approved):

PR #3970, opened 2023-10-17
Fix loading CSI driver container from state if it exists

mye956:
### Summary
This PR will address the issue where agent won't be able to recognize an existing running CSI driver daemon task after an agent restart. Because of this, when we get a new EBS attachment payload, agent will start up a new CSI driver daemon task even if there's already an existing running CSI driver task. While all of the managed tasks are properly populated back into the task engine during the setup of agent start, the list of managed daemon task aren't. To fix this, we'll check if there's an existing non-stopped CSI driver daemon task right after we load all of the managed tasks from BoltDB on agent restart.

### Implementation details
Implemented the following:
* `agent/api/task/task.go`: New functionality called `IsManagedDaemonTask` where it checks if the task is a managed daemon if it satisfy the following: 
  * Was internally made by agent
  * Not stopped (i.e. not terminal)
  * Has container of type `ContainerManagedDaemon`
  * If all of the above satisfy, then the task will be added to the map of daemon tasks within `loadTask`
* `agent/api/container/container.go`: 
  * New functionality called `IsManagedDaemonContainer` where it checks if a container is of type `ContainerManagedDaemon`.  This will be called in `IsManagedDaemonTask` as part of the managed daemon task check. 
  * New functionality called `GetImageName` where it gets the image name without the tags of a container
* `agent/api/container/containertype.go`: Added `ContainerManagedDaemon` to the map of container types that's used for encoding and decoding `ContainerType` objects
* `agent/engine/data.go`: Modified `loadTask` to check if any of the tasks are considered a managed daemon task. If a task is then we add it to the map of managed daemon tasks in our task engine.
* `agent/ebs/watcher.go`: Added an extra condition that we should also be creating a new CSI driver daemon task if the existing CSI driver daemon task has stopped (i.e. terminal).

(Note: There should only be at most one of each managed daemon task running on one instance at a time; i.e. multiple managed daemon tasks can be running on one instance as long as they're not the same one)

### Testing
Added new unit tests.

Manual testing:
Ran an EBS-backed task which then created the CSI driver daemon task. After a bit, I restarted agent with the CSI driver task still running and was able to confirm that agent was able to properly load back the existing CSI driver daemon task.
```
level=debug time=2023-10-18T17:11:51Z msg="Task is not internally made"
level=debug time=2023-10-18T17:11:51Z msg="Task is not a managed daemon task: arn:aws:ecs:us-west-2::task/evm-test-gamma/9bb5422fecda42088bcc3451a8d37dbb" module=data.go
level=debug time=2023-10-18T17:11:51Z msg="Container is running and is internal"
level=debug time=2023-10-18T17:11:51Z msg="Successfully set managed daemon task for: ebs-csi-driver" module=data.go
level=info time=2023-10-18T17:11:51Z msg="Cluster was successfully restored" cluster="evm-test-gamma"
level=debug time=2023-10-18T17:11:51Z msg="Loading pause container tarball:" image="/images/amazon-ecs-pause.tar"
level=debug time=2023-10-18T17:11:51Z msg="Inspecting container image: " image="amazon/amazon-ecs-pause:0.1.0"
```
Ran another EBS-backed task and agent didn't create a new CSI driver daemon task as well as reused the existing task.
```
[ec2-user@ip-172-31-35-197 amazon-ecs-agent]$ docker ps
CONTAINER ID   IMAGE                            COMMAND                  CREATED         STATUS                   PORTS     NAMES
953d6d9ddf96   mreferre/yelb-db:0.5             "docker-entrypoint.s…"   2 minutes ago   Up 2 minutes                       ecs-evm-db-6-db-fee385a6f78fe3a1af01
f09b5e9f3526   amazon/amazon-ecs-agent:latest   "/agent"                 3 minutes ago   Up 3 minutes (healthy)             ecs-agent
b3ede243b73a   ebs-csi-driver:latest            "/bin/ebs-csi-driver…"   6 minutes ago   Up 6 minutes                       ecs---ecs-managed-ebs-csi-driver-a48dcaeed3cfdfdafb01
```

New tests cover the changes:Yes

Fix local agent state for CSI driver daemon task

### Licensing

### DIFF
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -3687,3 +3687,32 @@ func (task *Task) HasActiveContainers() bool {
 	}
 	return false
 }
+
+// IsManagedDaemonTask will check if a task is a non-stopped managed daemon task
+// TODO: Somehow track this on a task level (i.e. obtain the managed daemon image name from task arn and then find the corresponding container with the image name)
+func (task *Task) IsManagedDaemonTask() (string, bool) {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	// We'll want to obtain the last known non-stopped managed daemon task to be saved into our task engine.
+	// There can be an edge case where the task hasn't been progressed to RUNNING yet.
+	if !task.IsInternal || task.KnownStatusUnsafe.Terminal() {
+		return "", false
+	}
+
+	for _, c := range task.Containers {
+		if c.IsManagedDaemonContainer() {
+			imageName := c.GetImageName()
+			return imageName, true
+		}
+	}
+	return "", false
+}
+
+func (task *Task) IsRunning() bool {
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+	taskStatus := task.KnownStatusUnsafe
+
+	return taskStatus == apitaskstatus.TaskRunning
+}
--- a/agent/ebs/watcher.go
+++ b/agent/ebs/watcher.go
@@ -126,7 +126,8 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
 	}
 
 	// start EBS CSI Driver Managed Daemon
-	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil {
+	// We want to avoid creating a new CSI driver task if there's already one that's not been stopped.
+	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil && !runningCsiTask.GetKnownStatus().Terminal() {
 		log.Debugf("engine ebs CSI driver is running with taskID: %v", runningCsiTask.GetID())
 	} else {
 		if ebsCsiDaemonManager, ok := w.taskEngine.GetDaemonManagers()[md.EbsCsiDriver]; ok {
@@ -191,7 +192,7 @@ func (w *EBSWatcher) stageVolumeEBS(volID, deviceName string) error {
 	}
 	attachmentMountPath := ebsAttachment.GetAttachmentProperties(apiebs.SourceVolumeHostPathKey)
 	hostPath := filepath.Join(hostMountDir, attachmentMountPath)
-	filesystemType := ebsAttachment.GetAttachmentProperties(apiebs.FileSystemTypeName)
+	filesystemType := ebsAttachment.GetAttachmentProperties(apiebs.FileSystemKey)
 	// CSI NodeStage stub required fields
 	stubSecrets := make(map[string]string)
 	stubVolumeContext := make(map[string]string)
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -1508,3 +1508,16 @@ func (c *Container) GetContainerPortRangeMap() map[string]string {
 	defer c.lock.RUnlock()
 	return c.ContainerPortRangeMap
 }
+
+func (c *Container) IsManagedDaemonContainer() bool {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.Type == ContainerManagedDaemon
+}
+
+func (c *Container) GetImageName() string {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	containerImage := strings.Split(c.Image, ":")[0]
+	return containerImage
+}
--- a/agent/api/container/container_test.go
+++ b/agent/api/container/container_test.go
@@ -130,9 +130,49 @@ func TestIsInternal(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		t.Run(fmt.Sprintf("IsInternal shoukd return %t for %s", tc.internal, tc.container.String()),
+		t.Run(fmt.Sprintf("IsInternal should return %t for %s", tc.internal, tc.container.String()),
 			func(t *testing.T) {
 				assert.Equal(t, tc.internal, tc.container.IsInternal())
+			})
+	}
+}
+
+func TestIsManagedDaemonContainer(t *testing.T) {
+	testCases := []struct {
+		container       *Container
+		internal        bool
+		isManagedDaemon bool
+	}{
+		{&Container{}, false, false},
+		{&Container{Type: ContainerNormal, Image: "someImage:latest"}, false, false},
+		{&Container{Type: ContainerManagedDaemon, Image: "someImage:latest"}, true, true},
+	}
+
+	for _, tc := range testCases {
+		t.Run(fmt.Sprintf("IsManagedDaemonContainer should return %t for %s", tc.isManagedDaemon, tc.container.String()),
+			func(t *testing.T) {
+				assert.Equal(t, tc.internal, tc.container.IsInternal())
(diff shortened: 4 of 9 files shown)


agent/api/task/task.go
```diff
@@ -3723,3 +3723,23 @@ func (task *Task) HasActiveContainers() bool {
...
+	task.lock.RLock()
+	defer task.lock.RUnlock()
+
+	if !task.IsInternal {
+		return "", false
+	}
+
+	for _, c := range task.Containers {
+		containerStatus := c.GetKnownStatus()
+		if containerStatus.IsRunning() && c.IsInternal() {
```
sparrc, line 3737: why does the container need to be running for us to check if it's name matches the "managed daemon" names?

sparrc: couldn't this be a race condition where the agent gets a new EBS task before the existing managed daemon task containers have gone to running?

sparrc:
I think the concept here would break when future managed daemons are added. For this problem don't we specifically need to figure out if we have the "ebs-csi-driver" task running? Not _any_ managed daemon task?

agent/api/task/task.go
mye956: good question, we don't want to run into a scenario where the task/container was stopped but we still add it to the list of manged daemon tasks. otherwise agent won't be able to create/start a new CSI driver task prior to staging EBS volumes since it'll think that there's already an existing active CSI driver task

mye956:
> I think the concept here would break when future managed daemons are added. For this problem don't we specifically need to figure out if we have the "ebs-csi-driver" task running? Not _any_ managed daemon task?

Hmm, I don't quite follow here. If we were to introduce a new managed daemon then it could be added to the set of managed daemon images and also check for that as well right on top of the CSI driver. You're completely right that the goal for this PR is to check if there's already an CSI driver task. Was also thinking of long term as well when we have more managed daemon tasks.

agent/api/task/task.go
mye956: > couldn't this be a race condition where the agent gets a new EBS task before the existing managed daemon task containers have gone to running?

we should be taking care of this when handling an EBS attachment payload -> https://github.com/aws/amazon-ecs-agent/blob/dev/agent/ebs/watcher.go#L132
we first check that if there's already an existing CSI driver task then we shouldn't create another. 

Thinking about it now, we should also include if there's a CSI driver task is being created

fierlion: One other thought about removing the IsRunning check: Managed Daemons have a container restart policy of `on failure` which implies that they may be running or stopped.  We should only ever have one container per daemon so we can track its health regardless of its process status.  We'll depend on the container restart policy to handle the daemon lifecycle and shouldn't launch redundant daemons.

ecs-agent/manageddaemon/managed_daemon.go
```diff
@@ -33,6 +33,12 @@ const (
 	defaultApplicationLogMount            = "applicationLogMount"
 )
 
+var (
```
fierlion, line 36: We shouldn't hard code this -- the daemons should be dynamic.  There may or may not be any number of managed daemons.  
A thought: can you wire these in from the managed daemon ImportAll path?  
every time the agent starts up it will create an array of managed daemons.  we can use the names of these daemons as our source list for checking against state.

mye956: per offline, we'll always take the image name at face value (i.e. no need to have a hardcoded list of managed daemon images)

agent/api/task/task.go
```diff
@@ -3687,3 +3687,33 @@ func (task *Task) HasActiveContainers() bool {
 	}
 	return false
 }
+
+// IsManagedDaemonTask will check if a task is a non-stopped managed daemon task
+// TODO: We'll probably want to also clean up all of the STOPPED managed daemon task from
```
fierlion, line 3692: nit -- the todo sentence seems to be missing the last part...
"from..." 
from what?

agent/api/task/task.go
```diff
@@ -3687,3 +3687,33 @@ func (task *Task) HasActiveContainers() bool {
...
+	// There can be an edge case where the task hasn't been progressed to RUNNING yet.
+	taskStatus := task.KnownStatusUnsafe
+	if !task.IsInternal || taskStatus >= apitaskstatus.TaskStopped {
+		return "", false
+	}
+
+	for _, c := range task.Containers {
+		if c.IsManagedDaemonContainer() {
+			imageName := c.GetImageName()
+			return imageName, true
```
fierlion, line 3707: if a managed daemon in future has more than a single container we'll need to figure out which container is THE container.  This might be an essential container, or else we'll need to enforce a one-daemon-container per daemon task rule.  We might want to add a TODO to track this as a task-level field in future.  
https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/daemonmanager/daemon_manager_linux.go#L131C3-L131C103
The daemons are currently very specifically focused on a single image as base.

agent/ebs/watcher.go
```diff
@@ -126,7 +126,9 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
 	}
 
 	// start EBS CSI Driver Managed Daemon
-	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil {
+	// We want to avoid creating a new CSI driver task if there's already one that's not been stopped.
+	// TODO: Include unit tests
+	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil && !runningCsiTask.GetKnownStatus().Terminal() {
```
fierlion, line 131: this ties into the check above in `IsManagedDaemonTask`?  can we use the Terminal Check above also?
```
if !task.IsInternal || taskStatus >= apitaskstatus.TaskStopped {
```

agent/ebs/watcher.go
```diff
@@ -126,7 +126,9 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
 	}
 
 	// start EBS CSI Driver Managed Daemon
-	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil {
+	// We want to avoid creating a new CSI driver task if there's already one that's not been stopped.
+	// TODO: Include unit tests
```
fierlion, line 130: I think I see a test for this below in `TestHandleEBSAttachmentWithExistingCSIDriverTask` and `TestHandleEBSAttachmentWithStoppedCSIDriverTask` -- do you still need this TODO?

agent/api/task/task.go
mye956: no longer needed, same todo is in `data.go`

agent/ebs/watcher.go
mye956: good catch

agent/ebs/watcher.go
mye956: sure

fierlion (approved):

sparrc (approved):

agent/api/task/task.go
mye956: added

PR #3971, opened 2023-10-19
Add support for finding EBS devices on Xen instances

amogh09:
### Summary

Add support to `EBSDiscoveryClient` for discovering EBS devices on Xen instances. An EBS device can appear on a Xen instance with `sd` or `xvd` prefix of its name interchanged as compared to its name on the AWS backend. This PR updates the device discovery logic of `EBSDiscoveryClient` accordingly. 

### Implementation details

* A new function `matchXenBlockDevice` is added that checks if a block device matches a device name on Xen instances as per the matching logic explained in the summary. 
* `parseLsblkOutput` function is updated with a new `hasXenSupport bool` parameter. When this parameter is `true` the function will call `matchXenBlockDevice` function to match block devices in addition to the regular matching logic. 
* `EBSDiscoveryClient` is updated with a new `hasXenSupport bool` field and its constructor function is updated to take any number of options of type `EBSDiscoveryClientOption`. A new `WithXenSupport` function is added that returns a `EBSDiscoveryClientOption` that enables Xen support on an `EBSDiscoveryClient`.

### Testing

Add support for finding EBS devices on Xen instances

  no

### Licensing

### DIFF
--- a/ecs-agent/api/resource/ebs_discovery_linux.go
+++ b/ecs-agent/api/resource/ebs_discovery_linux.go
@@ -56,7 +56,7 @@ func (api *EBSDiscoveryClient) ConfirmEBSVolumeIsAttached(deviceName, volumeID s
 	}
 
 	expectedVolumeId := strings.ReplaceAll(volumeID, "-", "")
-	actualDeviceName, err := parseLsblkOutput(&lsblkOut, deviceName, expectedVolumeId)
+	actualDeviceName, err := parseLsblkOutput(&lsblkOut, deviceName, expectedVolumeId, api.HasXenSupport())
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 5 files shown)


ecs-agent/api/resource/ebs_discovery_linux.go
```diff
@@ -78,13 +83,55 @@ func (api *EBSDiscoveryClient) ConfirmEBSVolumeIsAttached(deviceName, volumeID s
 //		   }
 //		]
 //	 }
-func parseLsblkOutput(output *LsblkOutput, deviceName string, volumeId string) (string, error) {
+//
+// If hasXenSupport is true then, in addition to matching a device by its serial, this function
+// matches a device by its name disregarding "sd" or "xvd" prefix (whichever is present).
```
fierlion, line 88: very tiny nit -- we don't disregard the sd/xvd.  We expect at least one to be present.  This reads a bit like we don't care if either exists.  I do see that the code does this explicit check:
```
if hasXenPrefix(block.Name) && hasXenPrefix(deviceName) {
```

fierlion (dismissed):

mye956 (dismissed):

ecs-agent/api/resource/ebs_discovery_linux.go
amogh09: Yep. Updated in [034e758](https://github.com/aws/amazon-ecs-agent/pull/3971/commits/034e758b7e4f32b75cafe821fa28969e5a22a073).

xxx0624 (approved):

fierlion (approved):

mye956 (approved):

PR #3975, opened 2023-10-20
Fixing task volumes in task payload for EBS-backed tasks

mye956:
### Summary
This PR will address the current issue where if we launch an EBS-backed task, the list of volumes we get from the payload to be invalid in the eyes of agent. An example:
```
"Volumes": [{
            "DockerVolumeConfiguration": null,
            "EbsVolumeConfiguration": null,
            "EfsVolumeConfiguration": null,
            "FsxWindowsFileServerVolumeConfiguration": null,
            "Host": null,
            "Name": "someName",
            "Type": null
 }]
```
By default if the volume type is empty or nil, agent will fall back on volume type `host` and then look to try to parse whatever is in the `Host` field of the volume object for a host volume configuration. However, since everything except the volume name is empty, it will fail with the following error:
```
taskStatus="STOPPED" taskReason="UnrecogniedTaskError: Error loading task - invalid volume: empty volume configuration" taskKnownSentStatus="NONE" taskPullStartedAt="0001-01-01T00:00:00Z" taskPullStoppedAt="0001-01-01T00:00:00Z"
```
which isn't desired for EBS-backed tasks since the EBS volume is in the task attachments.

To accommodate this, we'll mark the corresponding volume object with the same volume name as the EBS volume attachment object to be of type `attachment`. This volume object will ultimately be replaced when we're parsing the EBS volume from the list of task attachments.

### Implementation details
Implemented new functionality in:
* `agent/acs/session/payload_responder.go`: 
  * New functionality called `hasEBSAttachment` to check if the task payload that we've received has EBS attachments within the list task attachments. If there are then we'll return the volume name. 
  * New functionlality called `initializeAttachmentTypeVolume` to find a volume object within the list of task volumes by volume name and then initialize/change the type to `attachment`. This task volume object will ultimately be replaced by the volume defined in the list of task attachments. This is currently only applicable to EBS volumes.
*  `agent/api/task/task.go` -> `agent/api/task/task_attachment_handler.go`: Moved functionality to clean up and remove all non-EBS volume configuration with the same volume name as EBS volume configuration within the list of task volumes.
* `agent/api/task/taskvolume.go`: Added a new task volume type called `attachment` for EBS volumes found in the list of task attachments within the payload.

### Testing
Unit tests

Manual testing:
Ran an EBS-backed task and it was able to successfully start one up.

Before changing the task volume type
```
"Volumes": [{
            "DockerVolumeConfiguration": null,
            "EbsVolumeConfiguration": null,
            "EfsVolumeConfiguration": null,
            "FsxWindowsFileServerVolumeConfiguration": null,
            "Host": null,
            "Name": "ebs1",
            "Type": null
        }],
```

After changing the task volume type
```
"volumes": [{
        "name": "ebs1",
        "type": "attachment"
    }]
```

EBS-backed task was started up
```
CONTAINER ID   IMAGE                            COMMAND                  CREATED              STATUS                    PORTS     NAMES
3e9e05f65c17   mreferre/yelb-db:0.5             "docker-entrypoint.s…"   About a minute ago   Up About a minute                   ecs-evm-db-6-db-9aa3bedbfcb8f1ec9201
4df58e21e339   amazon/amazon-ecs-agent:latest   "/agent"                 25 minutes ago       Up 25 minutes (healthy)             ecs-agent
e137050d8a00   ebs-csi-driver:latest            "/bin/ebs-csi-driver…"   5 hours ago          Up 5 hours                          ecs---ecs-managed-ebs-csi-driver-d2d7c8ddd1f08c91bc01
```

Mountpoint for the task
```
"Mounts": [
            {
                "Type": "bind",
                "Source": "/mnt/ecs/ebs/7bcd9e3ff65a489b9e1c43bfd6592384_vol-0501243308b3d16bc",
                "Destination": "/var/stuff",
                "Mode": "",
                "RW": true,
                "Propagation": "rprivate"
            }
```

New tests cover the changes: 

Fix invalid task volumes field for EBS-backed task payload

### Licensing

### DIFF
--- a/agent/acs/session/payload_responder.go
+++ b/agent/acs/session/payload_responder.go
@@ -22,6 +22,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine"
 	"github.com/aws/amazon-ecs-agent/agent/eventhandler"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
+	apiresource "github.com/aws/amazon-ecs-agent/ecs-agent/api/resource"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
@@ -106,6 +107,16 @@ func (pmHandler *payloadMessageHandler) addPayloadTasks(payload *ecsacs.PayloadM
 			allTasksOK = false
 			continue
 		}
+
+		// Note: If we receive an EBS-backed task, we'll also received an incomplete volume configuration in the list of Volumes
+		// To accomodate this, we'll first check if the task IS EBS-backed then we'll mark the corresponding Volume object to be
+		// of type "attachment". This volume object will be replaced by the newly created EBS volume configuration when we parse
+		// through the task attachments.
+		volName, ok := hasEBSAttachment(task)
+		if ok {
+			initializeAttachmentTypeVolume(task, volName)
+		}
+
 		apiTask, err := apitask.TaskFromACS(task, payload)
 		if err != nil {
 			pmHandler.handleInvalidTask(task, err, payload)
@@ -305,4 +316,27 @@ func isTaskStatusStopped(status apitaskstatus.TaskStatus) bool {
 // isTaskStatusNotStopped returns true if the task status != STOPPED.
 func isTaskStatusNotStopped(status apitaskstatus.TaskStatus) bool {
 	return status != apitaskstatus.TaskStopped
+}
+
+func hasEBSAttachment(acsTask *ecsacs.Task) (string, bool) {
+	// TODO: This will only work if there's one EBS volume per task. If we there is a case where we have multi-attach for a task, this needs to be modified
+	for _, attachment := range acsTask.Attachments {
+		if *attachment.AttachmentType == apiresource.EBSTaskAttach {
+			for _, property := range attachment.AttachmentProperties {
+				if *property.Name == apiresource.VolumeNameKey {
+					return *property.Value, true
+				}
+			}
+		}
+	}
+	return "", false
+}
+
+func initializeAttachmentTypeVolume(acsTask *ecsacs.Task, volName string) {
+	for _, volume := range acsTask.Volumes {
+		if *volume.Name == volName && volume.Type == nil {
+			newType := "attachment"
+			volume.Type = &newType
+		}
+	}
 }
--- a/agent/api/task/task_attachment_handler.go
+++ b/agent/api/task/task_attachment_handler.go
@@ -110,6 +110,7 @@ func handleTaskAttachments(acsTask *ecsacs.Task, task *Task) error {
 			task.ServiceConnectConfig = scHandler.(*ServiceConnectAttachmentHandler).scConfig
 		}
 		if len(ebsVolumeAttachments) > 0 {
+			ebsVolumes := make(map[string]bool)
 			for _, attachment := range ebsVolumeAttachments {
 				ebs, err := taskresourcevolume.ParseEBSTaskVolumeAttachment(attachment)
 				if err != nil {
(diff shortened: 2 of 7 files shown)


agent/acs/session/payload_responder.go
```diff
@@ -306,3 +317,26 @@ func isTaskStatusStopped(status apitaskstatus.TaskStatus) bool {
 func isTaskStatusNotStopped(status apitaskstatus.TaskStatus) bool {
 	return status != apitaskstatus.TaskStopped
 }
+
+func hasEBSAttachment(acsTask *ecsacs.Task) (string, bool) {
+	// TODO: This will only work if there's one EBS volume per task. If we there is a case where we have multi-attach for a task, this needs to be modified
```
fierlion, line 322: nit just leave this comment at "// TODO: This will only work if there's one EBS volume per task. "

agent/acs/session/payload_responder.go
```diff
@@ -106,6 +107,16 @@ func (pmHandler *payloadMessageHandler) addPayloadTasks(payload *ecsacs.PayloadM
 			allTasksOK = false
 			continue
 		}
+
+		// Note: If we receive an EBS-backed task, we'll also received an incomplete volume configuration in the list of Volumes
+		// To accomodate this, we'll first check if the task IS EBS-backed then we'll mark the corresponding Volume object to be
```
fierlion, line 112: This issue is ebs-specific today but will probably be true for other volumes in future.  
Can we just say "// To accommodate new volume patterns, we'll first check if the attachment properties include volume configuration then we'll mark the corresponding volume object to be of type attachment to designate source of truth.

agent/api/task/task_attachment_handler.go
```diff
@@ -120,8 +121,17 @@ func handleTaskAttachments(acsTask *ecsacs.Task, task *Task) error {
 					Type:   apiresource.EBSTaskAttach,
 					Volume: ebs,
 				}
+				ebsVolumes[ebs.VolumeName] = true
 				task.Volumes = append(task.Volumes, taskVolume)
 			}
+			// We're removing all incorrect volume configuration that were intially passed over from ACS
```
fierlion, line 127: nit: "all incorrect volume configurations"

agent/api/task/taskvolume.go
```diff
@@ -78,6 +79,9 @@ func (tv *TaskVolume) UnmarshalJSON(b []byte) error {
 		return tv.unmarshalFSxWindowsFileServerVolume(intermediate["fsxWindowsFileServerVolumeConfiguration"])
 	case apiresource.EBSTaskAttach:
 		return tv.unmarshalEBSVolume(intermediate["ebsVolumeConfiguration"])
+	case AttachmentType:
+		seelog.Warn("Obtaining the volume configuration from task attachments.")
```
fierlion, line 83: is this an expected path?  if so can this be Debug or Info level?

fierlion (approved):

agent/api/task/taskvolume.go
mythri-garaga: Trying to understand, why is this no-op for Attachment type? will this be updated in future PR?

agent/api/task/task_attachment_handler.go
amogh09: We are saying that the task payload we get from ECS backend is incorrect? That reads very odd to me. Is there a better way to put it?

agent/api/task/taskvolume.go
mye956: Good question, so we want this to be a noop for volume of type `attachment` since it'll end being removed and replaced by the newly created volume configuration that we make when handling the task attachments. 

We originally thought to try to have a special edge case for volumes defined in the list of task attachments here in `taskvolume.go` but that would mean we'd have to change around some of the existing edge cases and can get messy.

agent/api/task/task_attachment_handler.go
mye956: er sorry perhaps we can say "incomplete" here

agent/acs/session/payload_responder.go
```diff
@@ -106,6 +107,16 @@ func (pmHandler *payloadMessageHandler) addPayloadTasks(payload *ecsacs.PayloadM
...
+		// of type "attachment". This volume object will be replaced by the newly created EBS volume configuration when we parse
+		// through the task attachments.
+		volName, ok := hasEBSAttachment(task)
+		if ok {
+			initializeAttachmentTypeVolume(task, volName)
+		}
```
amogh09, line 118: Is this tested in any unit tests? The test update seems to be in the input and not in assertions?

agent/api/task/task_attachment_handler.go
amogh09: Even incomplete sounds wrong... non-blocking but let's rethink to prevent future confusion.

mythri-garaga (approved):

agent/acs/session/payload_responder.go
fierlion: +1

mye956: Not directly, I was hoping to do this in a follow up. A lot of the testing was mostly done through manual testing.

amogh09 (approved):

agent/api/task/task_attachment_handler.go
mye956: completely agree

PR #3984, opened 2023-10-25
Load Managed Daemon images in background

amogh09:
### Summary

Currently Agent loads Managed Daemon images synchronously on startup which causes a delay in startup. This PR addresses the startup delay by making Managed Daemon images load in background.

**Behavior changes -** 
* Managed Daemon image loading is still started on Agent startup but main thread no longer blocks on the image loading.
* Capability for EBS Task Attach now relies on the image file's existence on the host rather than image load success. This is a trade-off we are making to speed up startup time. 
* EBS Watcher is updated to ensure that image load for EBS Managed Daemon has finished before it starts a Managed Daemon task.
* EBS Managed Daemon task start is moved from EBS Watcher's EBS attachment payload handler to its tick handler. Note that an EBS Managed Daemon task can only be started after its image has been loaded. It is preferred to move the task start to tick handler for the following reasons. 
   1. Tick handler is called repeatedly so there is no need to block it on image load and a tick can simply be skipped if the image is not loaded yet.
   1. Attachment payload handler is triggered by Agent receiving a new attachment payload from ECS backend, so its invocation is non-deterministic and concurrent. If we make the handler block on image load completion then we would need to acquire a lock to prevent multiple EBS Managed Daemon tasks from getting created. It is worth nothing that this issue exists in current design also but would be exacerbated by asynchronous image loading that's introduced by this PR. 

### Implementation details

* A new function `loadManagedDaemonImagesAsync` is added to `app` package. This function loops through all registered Managed Daemons and starts loading the image for each. It adds the image to image cleanup exclusion list on successful image load. 
* `imageManager.AddImageToCleanUpExclusionList` method is updated to acquire the struct's RW lock to make the method thread-safe now that it's called concurrently. 
* `appendEBSTaskAttachCapabilities` function in `app` package is updated to add EBS Task attach capability based on EBS Managed Daemon's image file being found on the host instead of image load success. 
* EBS Managed Daemon task creation code is removed from EBS Watcher's `HandleEBSResourceAttachment` method. A new `tick` method is added for handling watcher ticks. In addition to how the tick is handled currently, `tick` method also calls a new `daemonRunning` method to check if EBS Managed Daemon task is running and does nothing if the Daemon task is not running.
* `daemonRunning` method checks if a Daemon task exists and its state. If the task has not reached RUNNING state then the method returns `true` if task is already running and `false` if not running yet. Otherwise (either the task has not initialized or has stopped or is stopping) the method starts a new EBS Managed Daemon task. 
* A new `ImageExists` method is added to `DaemonManager` and an implementation is added for Linux. The implementation checks if the managed daemon's image tar file exists on the host.

### Testing

Unit tests are added for all the changes.

Manual tests performed - 
* Checked that EBS backed tasks run as expected. Ran multiple tasks.
* Checked that Agent is able to start a new EBS Managed Daemon if an existing one is stopped.
* Restarted the Agent and checked that Agent is able to pick up an existing and running EBS Managed Daemon's state upon restart.

Load managed daemon images in background to speed up startup time.

no

### Licensing


agent/ebs/watcher.go
```diff
@@ -92,6 +93,94 @@ func (w *EBSWatcher) Start() {
...
+		logger.Debug("EBS Managed Daemon task is pending transitioning to running", logger.Fields{
+			field.TaskID:      csiTask.GetID(),
+			field.KnownStatus: csiTask.GetKnownStatus(),
+		})
+		return false
+	}
+
+	// Task is not running or about to run, we need to start a new one.
+
+	// Log some information.
```
fierlion, line 135: nit is this comment necessary (probably a leftover?)

agent/app/agent.go
```diff
@@ -751,6 +747,37 @@ func (agent *ecsAgent) constructVPCSubnetAttributes() []*ecs.Attribute {
...
+	logger.Debug(fmt.Sprintf("Will load images for %d Managed Daemons", len(daemonManagers)))
+	for _, daemonManager := range daemonManagers {
+		go agent.loadManagedDaemonImage(daemonManager, imageManager)
+	}
+}
+
+// Loads Managed Daemon image and adds it to image cleanup exclusion list upon success.
+func (agent *ecsAgent) loadManagedDaemonImage(dm dm.DaemonManager, imageManager engine.ImageManager) {
+	logger.Info("Starting to load Managed Daemon image", logger.Fields{
+		field.ImageRef: dm.GetManagedDaemon().GetImageRef(),
```
fierlion, line 764: nit: we're referencing this imageref a few times -- would it be useful to grab it once as we enter the function and use it?
```
daemonImageRef := dm.GetManagedDaemon().GetImageRef()
```
then use that everywhere below...

agent/ebs/watcher.go
```diff
@@ -92,6 +93,94 @@ func (w *EBSWatcher) Start() {
...
+	}
+	if csiTask != nil && csiTask.GetKnownStatus() < status.TaskRunning {
+		logger.Debug("EBS Managed Daemon task is pending transitioning to running", logger.Fields{
+			field.TaskID:      csiTask.GetID(),
+			field.KnownStatus: csiTask.GetKnownStatus(),
+		})
+		return false
+	}
+
+	// Task is not running or about to run, we need to start a new one.
```
mye956, line 133: nit: I think we can leave out the `or about to run` part. We should only be starting a new one if there's no CSI task yet or if the existing one that's saved to the task engine has stopped.

agent/ebs/watcher.go
```diff
@@ -125,29 +214,6 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
 		})
 	}
```
sparrc, line 216: Should we be calling the daemonRunning() function here where this code has been deleted?

agent/app/agent_test.go
```diff
@@ -1859,3 +1860,49 @@ func TestWaitUntilInstanceInServicePolling(t *testing.T) {
...
+
+func TestLoadManagedDaemonImage(t *testing.T) {
+	tcs := []struct {
+		name                         string
+		setDaemonManagerExpectations func(*mock_daemonmanager.MockDaemonManager)
+		setImageManagerExpectations  func(*mock_engine.MockImageManager)
+	}{
+		{
+			name: "no exclusion list update if image load fails",
+			setDaemonManagerExpectations: func(mdm *mock_daemonmanager.MockDaemonManager) {
```
fierlion, line 1872: I love this pattern.  I'll be using it in future

fierlion (approved):

agent/ebs/watcher.go
amogh09: How would the watcher then handle the case when EBS Daemon task has been created but is not running yet?

agent/ebs/watcher.go
```diff
@@ -125,29 +214,6 @@ func (w *EBSWatcher) HandleEBSResourceAttachment(ebs *apiebs.ResourceAttachment)
...
 
-	// start EBS CSI Driver Managed Daemon
-	// We want to avoid creating a new CSI driver task if there's already one that's not been stopped.
-	if runningCsiTask := w.taskEngine.GetDaemonTask(md.EbsCsiDriver); runningCsiTask != nil && !runningCsiTask.GetKnownStatus().Terminal() {
```
mye956, line 130: nit: since we've moved the CSI creation to now only be created if we have found an EBS volume at every tick, we probably should update the comments about this

agent/ebs/watcher.go
amogh09: No. `HandleEBSResourceAttachment` no longer managed the daemon task.

agent/ebs/watcher.go
amogh09: Sure. I will update.

agent/app/agent.go
amogh09: Sure. I will update.

agent/ebs/watcher.go
amogh09: Ah you are right. I will update the comment.

agent/ebs/watcher.go
amogh09: These comments are deleted. Do you mean we should make the description of `daemonRunning` method more detailed? 
```go
// Checks if EBS Daemon Task is running and starts a new one it if it's not.
func (w *EBSWatcher) daemonRunning() bool {
```

agent/app/agent.go
```diff
@@ -751,6 +747,37 @@ func (agent *ecsAgent) constructVPCSubnetAttributes() []*ecs.Attribute {
 	}
 }
 
+// Loads Managed Daemon images for all Managed Daemons registered on the Agent.
+// The images are loaded in the background. Successfully loaded images are added to
+// imageManager's cleanup exclusion list.
+func (agent *ecsAgent) loadManagedDaemonImagesAsync(imageManager engine.ImageManager) {
+	daemonManagers := agent.getDaemonManagers()
```
mye956, line 754: Just curious, have we tried using the the existing `GetDaemonManagers` in our task engine? Curious if we could just use that instead. https://github.com/aws/amazon-ecs-agent/blob/dev/agent/engine/docker_task_engine.go#L1230

agent/ebs/watcher.go
mye956: oh er sorry I was referring to line 199

agent/app/agent.go
amogh09: That's a good find! I wasn't aware of that. 

It seems that the underlying map pointed to by `ecsAgent.daemonManagers` and `engine.daemonManagers` is the same, so it doesn't matter which one we use here. 
https://github.com/aws/amazon-ecs-agent/blob/dev/agent/app/agent.go#L346-L348

I prefer using the one from `ecsAgent` type as the image load method is added to `ecsAgent` type. Does that sound okay?

agent/ebs/watcher.go
amogh09: Ah right. I missed that. Will update. Thanks!

agent/app/agent.go
mye956: yep yep sounds good to me!

mye956 (dismissed):

agent/ebs/watcher.go
amogh09: Updated in https://github.com/aws/amazon-ecs-agent/pull/3984/commits/a56ea6195b4378a019d0016a40996720c4ece7a1

agent/app/agent.go
amogh09: Updated in https://github.com/aws/amazon-ecs-agent/pull/3984/commits/a56ea6195b4378a019d0016a40996720c4ece7a1

agent/ebs/watcher.go
amogh09: Updated in https://github.com/aws/amazon-ecs-agent/pull/3984/commits/a56ea6195b4378a019d0016a40996720c4ece7a1

agent/ebs/watcher.go
amogh09: Updated in https://github.com/aws/amazon-ecs-agent/pull/3984/files/c5b6385c0d2aa96c14aa81f88c4c0f4c5cbc86ff..3d68cadba4c9dede99e56c0600199cd5daad3f85

agent/app/agent.go
amogh09: Thanks!

mye956 (approved):

agent/ebs/watcher.go
mye956: thanks!

fierlion (approved):

PR #3929, opened 2023-09-26
Collect volume metrics and volume clean up for EBS-backed tasks

mye956:
### Summary
This PR will introduce the functionality to collect volume metrics for EBS-backed tasks via the CSI driver as well as task clean up for EBS-backed task.

#### Changes made to collect Volume metrics in EBS-backed Tasks
Implemented `getEBSVolumeMetrics()` and `fetchEBSVolumeMetrics()` where it'll collect the volume metrics via `NodeGetVolumeStats` from the CSI driver for each EBS-backed tasks. The `getEBSVolumeMetrics()` will be invoked when getting all of the other task and instance metrics in `GetInstanceMetrics()`. 

`IsEBSTaskAttachEnabled()` has also been implemented and will scan for any EBS volume configuration within the list of volumes of a task. Although it's being called in `AddTask()` within a check, it will temporarily always return false for all tasks (i.e. consider it not EBS-backed task) in order to avoid invoking the CSI driver functionality. This will be fixed and addressed in another PR.

List of files changed:
* agent/api/task/task.go: Added functionality to `IsEBSTaskAttachEnabled()`
* agent/engine/docker_task_engine.go: Temporarily assume that all tasks are not EBS-backed, Will be fixed in a future PR
* agent/stats/engine.go
* agent/stats/engine_test.go
* agent/stats/engine_unix.go: Essentially calls upon the CSI driver to run `NodeGetVolumeStats` on all EBS volumes attached to EBS-backed tasks
* agent/stats/engine_unix_test.go
* agent/stats/engine_windows.go
* ecs-agent/csiclient/csi_client.go

#### Changes made to clean up EBS-backed tasks
Implemented `UnstageVolumes` to unmount the host mountpoint for all EBS volumes on a EBS-backed task. This will happen on task clean up.

List of files changed:
* agent/engine/task_manager.go

### Testing
Added new unit test.

`TestFetchEBSVolumeMetrics`: Test that the new `fetchEBSVolumeMetrics` function is working as intended. As a TODO, we'll need to add a unhappy test case as well.

TODO: Add unit test for `NodeUnstage` in a follow up PR.

Manual testing:
We've manually mounted/staged an EBS volume onto a testing EC2 instance with agent running and tried calling the `getEBSVolumeMetrics` functionality. 
Agent logs to get volume metrics
```
level=debug time=2023-09-26T04:14:54Z msg="Fetching EBS volume metrics..."
[...]

PR #3985, opened 2023-10-25
Addressing TODOs for NodeGetVolumeMetrics and NodeUnstageVolume

mye956:
### Summary
This PR will address all of the leftover TODOs for `NodeGetVolumeMetrics` and `NodeUnstageVolume` functionality for EBS-backed tasks.

### Implementation details
* `agent/engine/task_manager.go`: Changed `UnstageVolumes` to return a list of any errors that were encountered when trying to unstage a volume via the CSI driver. We'll then log all encountered errors. This will be helpful for having unit tests since otherwise we wouldn't know whether or not there were any issues.
* `agent/taskresource`: Added new getters within the `Volume` interface. This is to avoid typecasting the volume within the `NodeGetVolumeMetrics` as mentioned in [#3929](https://github.com/aws/amazon-ecs-agent/pull/3929).
* `agent/stats/engine_unix.go`: No longer type casting the task volume within `fetchEBSVolumeMetrics` in order to obtain the EBS volume and will instead use the getters implemented in `agent/taskresource`

Related PR: https://github.com/aws/amazon-ecs-agent/pull/3929

### Testing
New unit tests:
* TestFetchEBSVolumeMetricsHappy
* TestFetchEBSVolumeMetricsUnhappy
* TestFetchEBSVolumeMetricsNoTimeoutError
* TestFetchEBSVolumeMetricsTimeoutError
* TestUnstageVolumesHappy
* TestUnstageVolumesUnhappy
* TestUnstageVolumesNoTimeoutError
* TestUnstageVolumesTimeoutError

Added more unit testing for NodeUnstageVolume and NodeGetVolumeMetrics as well as address other leftover TODOs  for EBS-backed task

### Licensing

### DIFF
--- a/agent/engine/task_manager_test.go
+++ b/agent/engine/task_manager_test.go
@@ -44,11 +44,14 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/sighandlers/exitcodes"
 	"github.com/aws/amazon-ecs-agent/agent/statechange"
 	"github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
+	taskresourcevolume "github.com/aws/amazon-ecs-agent/agent/taskresource/volume"
+	apiresource "github.com/aws/amazon-ecs-agent/ecs-agent/api/attachment/resource"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
 	apierrors "github.com/aws/amazon-ecs-agent/ecs-agent/api/errors"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
 	mock_credentials "github.com/aws/amazon-ecs-agent/ecs-agent/credentials/mocks"
+	mock_csiclient "github.com/aws/amazon-ecs-agent/ecs-agent/csiclient/mocks"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/eventstream"
 	ni "github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/networkinterface"
 	mock_ttime "github.com/aws/amazon-ecs-agent/ecs-agent/utils/ttime/mocks"
@@ -2226,4 +2229,64 @@ func TestTaskWaitForHostResources(t *testing.T) {
 	// Verify arn2 got dequeued
 	topTask, err = taskEngine.topTask()
 	assert.Error(t, err)
+}
+
+func TestUnstageVolumes(t *testing.T) {
+	tcs := []struct {
+		name      string
+		err       error
+		numErrors int
+	}{
+		{
+			name:      "Success",
+			err:       nil,
+			numErrors: 0,
+		},
+		{
+			name:      "Failure",
+			err:       errors.New("unable to unstage volume"),
+			numErrors: 1,
+		},
+		{
+			name:      "TimeoutFailure",
+			err:       errors.New("rpc error: code = DeadlineExceeded desc = context deadline exceeded"),
+			numErrors: 1,
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			defer mockCtrl.Finish()
+			ctx, cancel := context.WithCancel(context.TODO())
+			defer cancel()
+			mtask := &managedTask{
+				Task: &apitask.Task{
+					ResourcesMapUnsafe:  make(map[string][]taskresource.TaskResource),
+					DesiredStatusUnsafe: apitaskstatus.TaskRunning,
+					Volumes: []apitask.TaskVolume{
+						{
+							Name: taskresourcevolume.TestVolumeName,
+							Type: apiresource.EBSTaskAttach,
+							Volume: &taskresourcevolume.EBSTaskVolumeConfig{
+								VolumeId:             taskresourcevolume.TestVolumeId,
+								VolumeName:           taskresourcevolume.TestVolumeId,
+								VolumeSizeGib:        taskresourcevolume.TestVolumeSizeGib,
+								SourceVolumeHostPath: taskresourcevolume.TestSourceVolumeHostPath,
+								DeviceName:           taskresourcevolume.TestDeviceName,
+								FileSystem:           taskresourcevolume.TestFileSystem,
+							},
+						},
+					},
+				},
+				ctx:                      ctx,
+				resourceStateChangeEvent: make(chan resourceStateChange),
+			}
+			mockCsiClient := mock_csiclient.NewMockCSIClient(mockCtrl)
+			mockCsiClient.EXPECT().NodeUnstageVolume(gomock.Any(), "vol-12345", "/mnt/ecs/ebs/taskarn_vol-12345").Return(tc.err).Times(1)
+
+			errors := mtask.UnstageVolumes(mockCsiClient)
+			assert.Len(t, errors, tc.numErrors)
+		})
+	}
 }
--- a/agent/stats/engine_unix_test.go
+++ b/agent/stats/engine_unix_test.go
@@ -18,6 +18,7 @@ package stats
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
@@ -30,6 +31,7 @@ import (
 	apiresource "github.com/aws/amazon-ecs-agent/ecs-agent/api/attachment/resource"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/csiclient"
+	mock_csiclient "github.com/aws/amazon-ecs-agent/ecs-agent/csiclient/mocks"
 	ni "github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/networkinterface"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/tcs/model/ecstcs"
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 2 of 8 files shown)


agent/engine/task_manager_test.go
```diff
@@ -2227,3 +2231,157 @@ func TestTaskWaitForHostResources(t *testing.T) {
...
+		resourceStateChangeEvent: make(chan resourceStateChange),
+	}
+
+	timeoutctx, timeoutcancel := context.WithTimeout(ctx, 50*time.Millisecond)
+	defer timeoutcancel()
+
+	mockCsiClient := mock_csiclient.NewMockCSIClient(mockCtrl)
+	mockCsiClient.EXPECT().NodeUnstageVolume(gomock.Any(), "vol-12345", "/mnt/ecs/ebs/taskarn_vol-12345").Do(func(ctx context.Context, volID, hostPath string) {
+		time.Sleep(5 * time.Millisecond)
+	}).Return(nil).MinTimes(1)
```
amogh09, line 2339: nit: Table-driven testing might help reduce duplicate code.

agent/engine/task_manager_test.go
```diff
@@ -2227,3 +2231,157 @@ func TestTaskWaitForHostResources(t *testing.T) {
...
+	defer timeoutcancel()
+
+	mockCsiClient := mock_csiclient.NewMockCSIClient(mockCtrl)
+	mockCsiClient.EXPECT().NodeUnstageVolume(gomock.Any(), "vol-12345", "/mnt/ecs/ebs/taskarn_vol-12345").Do(func(ctx context.Context, volID, hostPath string) {
+		time.Sleep(50 * time.Millisecond)
+	}).Return(fmt.Errorf("rpc error: code = DeadlineExceeded desc = context deadline exceeded")).MinTimes(1)
+
+	errors := mtask.UnstageVolumes(mockCsiClient)
+
+	require.Error(t, timeoutctx.Err())
```
amogh09, line 2385: Where is `timeoutctx` used?

agent/engine/task_manager_test.go
```diff
@@ -2227,3 +2231,157 @@ func TestTaskWaitForHostResources(t *testing.T) {
...
+		ctx:                      ctx,
+		resourceStateChangeEvent: make(chan resourceStateChange),
+	}
+
+	timeoutctx, timeoutcancel := context.WithTimeout(ctx, 5*time.Millisecond)
+	defer timeoutcancel()
+
+	mockCsiClient := mock_csiclient.NewMockCSIClient(mockCtrl)
+	mockCsiClient.EXPECT().NodeUnstageVolume(gomock.Any(), "vol-12345", "/mnt/ecs/ebs/taskarn_vol-12345").Do(func(ctx context.Context, volID, hostPath string) {
+		time.Sleep(50 * time.Millisecond)
```
amogh09, line 2380: Why is this sleep needed?

agent/stats/engine_unix_test.go
```diff
@@ -207,7 +213,13 @@ func TestFetchEBSVolumeMetrics(t *testing.T) {
 	engine.cluster = defaultCluster
 	engine.containerInstanceArn = defaultContainerInstance
 	engine.client = mockDockerClient
-	engine.csiClient = csiclient.NewDummyCSIClient()
+
+	mockCsiClient := mock_csiclient.NewMockCSIClient(mockCtrl)
+	mockCsiClient.EXPECT().GetVolumeMetrics(gomock.Any(), "vol-12345", "/mnt/ecs/ebs/taskarn_vol-12345").Return(&csiclient.Metrics{
+		Used:     15 * 1024 * 1024 * 1024,
+		Capacity: 20 * 1024 * 1024 * 1024,
+	}, nil).MinTimes(1)
```
amogh09, line 221: Is the exact number of `GetVolumeMetrics` calls deterministic? We may want to use `Times(<that number>)` if so to make the assertion stricter.

agent/stats/engine_unix_test.go
```diff
@@ -235,3 +247,176 @@ func TestFetchEBSVolumeMetrics(t *testing.T) {
...
+					VolumeSizeGib:        "10",
+					SourceVolumeHostPath: "taskarn_vol-12345",
+					DeviceName:           "/dev/nvme1n1",
+					FileSystem:           "ext4",
+				},
+			},
+		},
+	}
+
+	resolver.EXPECT().ResolveTaskByARN("t1").AnyTimes().Return(t1, nil)
```
amogh09, line 276: Why do we need `AnyTimes()`? Is the exact number of calls non-deterministic?

agent/stats/engine_unix_test.go
```diff
@@ -235,3 +247,176 @@ func TestFetchEBSVolumeMetrics(t *testing.T) {
...
+	mockDockerClient.EXPECT().Stats(gomock.Any(), gomock.Any(), gomock.Any()).Return(nil, nil).AnyTimes()
```
amogh09, line 277: Let's avoid using `AnyTimes()` unless necessary.

agent/stats/engine_unix_test.go
```diff
@@ -235,3 +247,176 @@ func TestFetchEBSVolumeMetrics(t *testing.T) {
...
+	actualMetrics := engine.fetchEBSVolumeMetrics(t1, "t1")
+
+	require.NoError(t, timeoutctx.Err())
+	assert.Len(t, actualMetrics, 1)
+	assert.Equal(t, actualMetrics, expectedMetrics)
+}
+
+// TestFetchEBSVolumeMetricsTimeoutError tests that there was a timeout error when fetching volume metrics for an EBS volume via the CSI client. The expected behavior is
+// to have no volume metrics withn the output of fetchEBSVolumeMetrics and return with a context deadline exceeded error.
+func TestFetchEBSVolumeMetricsTimeoutError(t *testing.T) {
```
amogh09, line 375: nit: might want to use table-driven tests to reduce duplicate code.

agent/engine/task_manager_test.go
```diff
@@ -2227,3 +2231,157 @@ func TestTaskWaitForHostResources(t *testing.T) {
 	topTask, err = taskEngine.topTask()
 	assert.Error(t, err)
 }
+
+func TestUnstageVolumesHappy(t *testing.T) {
```
fierlion, line 2235: in future we can reference the "happy" case as just Test<method> and "sad" cases can be named Test<method> + some designation specific to their failure mode.  (I see that you do this below: TestUnstageVolumesNoTimeoutError)

agent/stats/engine_unix_test.go
fierlion: (I'll also take that feedback on my own prs)

agent/engine/task_manager_test.go
mye956: good call, thanks!

agent/engine/task_manager_test.go
mye956: sure, it might be more helpful using a table as well for testing as @amogh09 suggested!

agent/engine/task_manager_test.go
mye956: the sleep is needed here to try to mock a NodeUnstageVolume call taking longer than the timeout duration.

agent/engine/task_manager_test.go
mye956: oh! good catch

agent/stats/engine_unix_test.go
mye956: oh yeah that's a better idea, thanks!

agent/stats/engine_unix_test.go
mye956: nope I don't think so, will update this to use `Times(<that number>)`. thanks!

agent/engine/task_manager_test.go
mye956: actually nevermind, we're using it to check if we hit a context deadline exceeded error.

agent/engine/task_manager_test.go
amogh09: I think it's the `Return(fmt.Errorf("rpc error: code = DeadlineExceeded desc = context deadline exceeded")).MinTimes(1)` that's making `NodeStageVolume` return that error and not this sleep. Timeout duration used by NodeUnstageVolume seems to be hardcoded to 3 seconds which is much larger than 50ms sleep that we have here. 
https://github.com/aws/amazon-ecs-agent/blob/f1bea74554c7f2e5ab08c8fc0436c7a0d7a3ff25/agent/engine/task_manager.go#L63C2-L63C22

agent/engine/task_manager_test.go
amogh09: I think it's unused as we only use it in an assertion.

agent/stats/engine_unix_test.go
mye956: done, great suggestion!

agent/stats/engine_unix_test.go
mye956: done,this was really helpful in figuring out what exactly is needed and what is not 😄

amogh09 (dismissed):

agent/taskresource/volume/dockervolume.go
```diff
@@ -152,16 +155,52 @@ func (cfg *EFSVolumeConfig) Source() string {
 	return cfg.DockerVolumeName
 }
 
+func (cfg *EFSVolumeConfig) GetType() string {
+	return EFSVolumeType
+}
+
+func (cfg *EFSVolumeConfig) GetVolumeId() string {
+	return ""
```
fierlion, line 163: I worry about adding something like this when there's an existing ID for EFS which we might want to return 
```
"fileSystemId": "fs-1234"
```

mye956: I agree. I recall this was sort of the reason why I originally opted to type-cast since not all volumes types will have fields like volume ID, device name, etc.

agent/taskresource/volume/dockervolume.go
```diff
@@ -152,16 +155,52 @@ func (cfg *EFSVolumeConfig) Source() string {
...
 func (cfg *DockerVolumeConfig) Source() string {
 	return cfg.DockerVolumeName
 }
 
+func (cfg *DockerVolumeConfig) GetType() string {
+	return DockerVolumeType
+}
+
+func (cfg *DockerVolumeConfig) GetVolumeId() string {
+	return ""
```
fierlion, line 180: ditto here -- is there a unique identifier we should return for a docker volume besides its name (which I believe is unique per host)?

mye956: ^ same response as above

mye956: I assume the docker volume name?

fierlion (approved):

amogh09 (approved):

PR #3990, opened 2023-10-25
Add unit tests for watcher StageAll method

fierlion:
### Summary
The ebs watcher StageAll() method calls CSI NodeStage for found EBS attachments.  This adds new unit tests to cover the success and failure cases for StageAll.

Add unit tests for EBS watcher StageAll method.

### Licensing


agent/ebs/watcher_test.go
```diff
@@ -300,6 +304,104 @@ func TestHandleDuplicateEBSAttachment(t *testing.T) {
...
+func TestStageAll(t *testing.T) {
+	tcs := []struct {
+		name                     string
+		setTaskStateExpectations func(*mock_dockerstate.MockTaskEngineState)
+		setCSIClientExpectations func(*mock_csiclient.MockCSIClient)
+		expectedErrorStubs       []string
+		expectedAttachmentStatus attachment.AttachmentStatus
+	}{
+		{
+			name: "StageAll succeeds with expected EBS CSI Driver fields",
```
fierlion, line 316: note: I'll add a case with an error in staging followed by success so we can test that we process past errors in our array.

agent/ebs/watcher_test.go
```diff
@@ -300,6 +304,104 @@ func TestHandleDuplicateEBSAttachment(t *testing.T) {
...
+					taskresourcevolume.TestVolumeId, gomock.Any(),
+					filepath.Join(hostMountDir, taskresourcevolume.TestSourceVolumeHostPath),
+					taskresourcevolume.TestFileSystem, gomock.Any(), gomock.Any(),
+					gomock.Any(), gomock.Any(), gomock.Any()).Return(errors.New("test")).Times(1)
+			},
+			expectedErrorStubs:       []string{"Failed to initialize EBS volume ID"},
+			expectedAttachmentStatus: attachment.AttachmentNone,
+		},
+		{
+			name: "StageAll returns expected error when already attached",
```
amogh09, line 348: I think there is no error returned in this case? I am curious why we say "returns expected error".

agent/ebs/watcher_test.go
```diff
@@ -300,6 +304,104 @@ func TestHandleDuplicateEBSAttachment(t *testing.T) {
...
+			setCSIClientExpectations: func(mcc *mock_csiclient.MockCSIClient) {},
+			expectedErrorStubs:       nil,
+			expectedAttachmentStatus: attachment.AttachmentAttached,
+		},
+		{
+			name: "StageAll fails with missing volume ID",
+			setTaskStateExpectations: func(ts *mock_dockerstate.MockTaskEngineState) {
+				ts.EXPECT().GetEBSByVolumeId(taskresourcevolume.TestVolumeId).Return(nil, false).Times(1)
+				ebsResourceAttachment := getEBSResourceAttachment()
+				ts.EXPECT().GetEBSByVolumeId(taskresourcevolume.TestVolumeId).Return(ebsResourceAttachment, true).Times(1)
```
amogh09, line 364: What happens when the second call to `GetEBSVolumeId` returns an error too?

agent/ebs/watcher_test.go
```diff
@@ -300,6 +304,104 @@ func TestHandleDuplicateEBSAttachment(t *testing.T) {
...
+			defer mockCtrl.Finish()
+
+			ctx := context.Background()
+
+			taskEngineState := mock_dockerstate.NewMockTaskEngineState(mockCtrl)
+			tc.setTaskStateExpectations(taskEngineState)
+			mockDiscoveryClient := mock_ebs_discovery.NewMockEBSDiscovery(mockCtrl)
+			mockTaskEngine := mock_engine.NewMockTaskEngine(mockCtrl)
+			mockTaskEngine.EXPECT().GetDaemonTask(md.EbsCsiDriver).Return(nil).AnyTimes()
+			mockTaskEngine.EXPECT().GetDaemonManagers().Return(nil).AnyTimes()
```
amogh09, line 383: I think `StageAll` does not interact with task engine at all? I am curious why we need these expectations.

agent/ebs/watcher_test.go
```diff
@@ -300,6 +304,104 @@ func TestHandleDuplicateEBSAttachment(t *testing.T) {
...
+
+			watcher := newTestEBSWatcher(ctx, taskEngineState, mockDiscoveryClient, mockTaskEngine, mockCsiClient)
+			foundVolumes := map[string]string{taskresourcevolume.TestVolumeId: taskresourcevolume.TestDeviceName}
+			errors := watcher.StageAll(foundVolumes)
+			assert.Equal(t, len(tc.expectedErrorStubs), len(errors))
+			if len(tc.expectedErrorStubs) != len(errors) {
+				t.Errorf("mismatched error count")
+			} else {
+				for i, err := range tc.expectedErrorStubs {
+					assert.True(t, strings.Contains(fmt.Sprintf("%s", errors[i]), err))
```
amogh09, line 396: nit: you can also use `assert.ErrorContains` instead of combining `assert.True` and `strings.Contains`.

agent/ebs/watcher_test.go
fierlion: specifically in this test the second call is only for the assertion which is set up in the test itself.
```
			assert.Equal(t, tc.expectedAttachmentStatus, ebsAttachment.GetAttachmentStatus())
```

agent/ebs/watcher_test.go
fierlion: will clean these up.

agent/ebs/watcher_test.go
fierlion: correct -- will change the description.  When already attached we just return `nil`.

agent/ebs/watcher_test.go
amogh09: oh I see 😅

amogh09 (approved):

agent/ebs/watcher_test.go
```diff
@@ -299,6 +299,125 @@ func TestHandleDuplicateEBSAttachment(t *testing.T) {
...
+			expectedAttachmentStatus: attachment.AttachmentNone,
+		},
+		{
+			name: "StageAll multiple volumes returns a single error and continues if first volume stage fails",
+			setTaskStateExpectations: func(ts *mock_dockerstate.MockTaskEngineState) {
+				ts.EXPECT().GetEBSByVolumeId("testVolId").Return(nil, false).Times(1)
+				ebsResourceAttachment := getEBSResourceAttachment()
+				ts.EXPECT().GetEBSByVolumeId(taskresourcevolume.TestVolumeId).Return(ebsResourceAttachment, true).Times(2)
+			},
+			foundVolumes: map[string]string{"testVolId": "testDeviceName", taskresourcevolume.TestVolumeId: taskresourcevolume.TestDeviceName},
```
fierlion, line 377: because the map is unordered, we can't guarantee that the volumes are processed in order so this is the best I could come up with to test that we process beyond an early failure.  The test doesn't require ordering, just an output of a single attached volume and a single error where no volumes enter the case in attached state.

mye956 (approved):

PR #4001, opened 2023-10-27
Fixing watcher tests with new CSI daemon task creation workflow

mye956:
### Summary
This PR will fix and "prettify" all of the skipped watcher tests with the new watcher changes introduced in [#3984](https://github.com/aws/amazon-ecs-agent/pull/3984).

### Implementation details
Now using a test table instead of individual test functions for most of the handle EBS attachment test cases.

* `agent/ebs/watcher_test.go`: New functionality called `getEBSAttachmentProperties` to obtain a new `AttachmentProperties` object with test values

### Testing
Unit tests for the following scenario:
* Handle EBS attachment with new CSI driver task
* Handle already expired EBS attachment
* Handle duplicate EBS attachments
* Handle EBS attachment with an existing CSI driver task that's running
* Handle EBS attachment with an existing CSI driver task that's stopped

Fix and prettify EBS watcher unit tests

### Licensing

### DIFF
--- a/agent/ebs/watcher_test.go
+++ b/agent/ebs/watcher_test.go
@@ -73,7 +73,6 @@ func newTestEBSWatcher(ctx context.Context, agentState dockerstate.TaskEngineSta
 // TestHandleEBSAttachmentHappyCase tests handling a new resource attachment of type Elastic Block Stores
 // The expected behavior is for the resource attachment to be added to the agent state and be able to be marked as attached.
 func TestHandleEBSAttachmentHappyCase(t *testing.T) {
-	t.Skip()
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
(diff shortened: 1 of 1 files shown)


agent/ebs/watcher_test.go
```diff
@@ -70,233 +70,318 @@ func newTestEBSWatcher(ctx context.Context, agentState dockerstate.TaskEngineSta
...
-	assert.True(t, ok)
-	assert.True(t, ebsAttachment.IsAttached())
+			assert.Len(t, taskEngineState.(*dockerstate.DockerTaskEngineState).GetAllEBSAttachments(), tc.expectedNumSavedAttachments)
+
+			ebsAttachment, ok := taskEngineState.(*dockerstate.DockerTaskEngineState).GetEBSByVolumeId(taskresourcevolume.TestVolumeId)
+			if !tc.shouldTick {
+				assert.False(t, ok)
+			} else {
+				// Note: There are cases where we need to tick more than once.
+				// Ex: Tick 1 -> Create a CSI driver task if it doesn't exist/isn't running, Tick 2 -> Find/Validate/Stage EBS volume
```
amogh09, line 373: In future maybe we should have separate unit tests for `tick` and `HandleEBSResourceAttachment` methods. That should make the tests simpler I think.

agent/ebs/watcher_test.go
```diff
@@ -70,233 +70,318 @@ func newTestEBSWatcher(ctx context.Context, agentState dockerstate.TaskEngineSta
...
+					md.EbsCsiDriver: mockDaemonManager,
+				}
+				gomock.InOrder(
+					mockTaskEngine.EXPECT().GetDaemonTask(md.EbsCsiDriver).Return(nil).Times(1),
+					mockTaskEngine.EXPECT().GetDaemonTask(md.EbsCsiDriver).Return(&task.Task{
+						Arn:               "arn:aws:ecs:us-east-1:012345678910:task/some-task-id",
+						KnownStatusUnsafe: status.TaskRunning,
+					}).Times(1),
+				)
+				mockTaskEngine.EXPECT().GetDaemonManagers().Return(daemonManagers).Times(1)
```
amogh09, line 101: nit: `Times(1)` is redundant as by default gomock sets the number of expected calls to 1.

amogh09 (dismissed):

agent/ebs/watcher_test.go
mye956: Good point, I was thinking more of a max amount of times it should be called. Perhaps I can use `MaxTimes()` as well.

agent/ebs/watcher_test.go
mye956: Yeah that's true, I originally wanted to test also test the general workflow as well. Especially in the test case where we already have an existing CSI task that's running or stopped.

amogh09: Yeah I understand the motivation. But setting up the mocks a certain way to make that happen is equivalent to directly calling the two methods with arguments for the different test cases. :)

agent/ebs/watcher_test.go
amogh09: `.Times(1)` (and not having any `.TImes(1)` since 1 is the default) will make the mock check that the method was invoked one and only once.

amogh09 (approved):

agent/ebs/watcher_test.go
mye956: Yep yep completely agree! I'll add this as a TODO.

agent/ebs/watcher_test.go
fierlion: TIL

agent/ebs/watcher_test.go
fierlion: (+1 I'd like to see this too.)

fierlion (approved):

PR #4003, opened 2023-10-27
Add an integration test for EBS CSI Client's GetVolumeMetrics method

amogh09:
### Summary

Add an integration test for `CSIClient`'s `GetVolumeMetrics` method. This method calls a CSI Driver (EBS in this case) to fetch volume metrics for a volume. The test expects a CSI Driver to be up and running. New make targets are added to start and stop EBS CSI Driver for testing. The test assumes that the host has a single EBS volume attached that is mounted at '/' path. 

### Implementation details

* Added `TestNodeVolumeStats` integration test that finds the volume ID of a single EBS volume on the host and then gets metrics for the volume using a `CSIClient`. It asserts that non-zero metrics are returned successfully. The test is skipped if `lsblk` version is older than 2.27 as those versions don't support JSON output format.
* Added some new make targets. 
   * `ebs-csi-driver`: builds EBS CSI Driver executable. 
   * `start-ebs-csi-driver`: starts an EBS CSI Driver process.
   * `stop-ecs-csi-driver`: stops running EBS CSI Driver processes.
* Updated the existing `run-integ-tests` target to make it start an EBS CSI Driver process before running the integration tests and stopping the driver process after the tests have finished. 

### Testing

Ran `make run-integ-tests` on an EC2 instance. EBS CSI Driver was successfully started and stopped by the make target. `TestNodeVolumeStats` passed. 

CI checks will also run the new test.

Add integration test for `CSIClient`'s `GetvolumeMetrics` method.

no

### Licensing

### DIFF
--- a/ecs-agent/csiclient/csi_client_linux_integ_test.go
+++ b/ecs-agent/csiclient/csi_client_linux_integ_test.go
@@ -1,0 +1,60 @@
+//go:build linux && integration
+// +build linux,integration
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+package csiclient
+
+import (
+	"context"
+	"encoding/json"
+	"os/exec"
+	"testing"
+	"time"
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/api/attachment/resource"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Tests that CSIClient can fetch EBS volume metrics from EBS CSI Driver.
+//
+// This test assumes that there is only one EBS Volume attached to the host that's mounted
+// on "/" path. It also assumes that an EBS CSI Driver is running and listening on
+// /tmp/ebs-csi-driver.sock socket file.
+//
+// A typical environment for this test is an EC2 instance with a single EBS volume as its
+// root device volume.
+func TestNodeVolumeStats(t *testing.T) {
+	const timeoutDuration = 5 * time.Second
+
+	// Find the root volume using lsblk
+	lsblkCtx, lsblkCtxCancel := context.WithTimeout(context.Background(), timeoutDuration)
+	defer lsblkCtxCancel()
+	output, err := exec.CommandContext(lsblkCtx, "lsblk", "-o", "NAME,SERIAL", "-J").CombinedOutput()
+	require.NoError(t, err)
+	var lsblkOut resource.LsblkOutput
+	err = json.Unmarshal(output, &lsblkOut)
+	require.NoError(t, err)
+	require.True(t, len(lsblkOut.BlockDevices) > 0)
+
+	// Get metrics for the root volume from EBS CSI Driver.
+	volumeID := lsblkOut.BlockDevices[0].Serial
+	csiClient := NewCSIClient("/tmp/ebs-csi-driver.sock")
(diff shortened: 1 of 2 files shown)


mye956 (approved):

ecs-agent/csiclient/csi_client_linux_integ_test.go
```diff
@@ -0,0 +1,91 @@
...
+	// There is no serial reported for EBS volumes on Xen devices.
+	// Skip the test for such instances.
+	volumeID := lsblkOut.BlockDevices[0].Serial
+	if volumeID == "" {
+		t.Skip("Test not supported on xen instances")
+	}
+
+	// Get metrics for the root volume from EBS CSI Driver.
+	csiClient := NewCSIClient("/tmp/ebs-csi-driver.sock")
+	getVolumeCtx, getVolumeCtxCancel := context.WithTimeout(context.Background(), timeoutDuration)
```
mye956, line 64: nit(non-blocking): perhaps we can have the same timeout as we have in our actual GetVolumeMetrics call. https://github.com/aws/amazon-ecs-agent/blob/master/agent/stats/engine_unix.go#L102. Although if this is too short and causes flakiness, please ignore!

amogh09: Yep. Updated!

mye956 (approved):

ecs-agent/csiclient/csi_client_linux_integ_test.go
```diff
@@ -0,0 +1,91 @@
...
+	defer getVolumeCtxCancel()
+	metrics, err := csiClient.GetVolumeMetrics(getVolumeCtx, volumeID, "/")
```
xxx0624, line 66: nit - IIRC the implementation of `GetVolumeMetrics` doesn't care the value of the `volumeID`. As long as it's not empty, it will pass. So w can also run this test on xen.

xxx0624 (dismissed):

ecs-agent/csiclient/csi_client_linux_integ_test.go
amogh09: Oh that'd be great. Let me try that.

ecs-agent/csiclient/csi_client_linux_integ_test.go
```diff
@@ -0,0 +1,91 @@
...
+// Tests that CSIClient can fetch EBS volume metrics from EBS CSI Driver.
+//
+// This test assumes that there is only one EBS Volume attached to the host that's mounted
+// on "/" path. It also assumes that an EBS CSI Driver is running and listening on
+// /tmp/ebs-csi-driver.sock socket file.
+//
+// A typical environment for this test is an EC2 instance with a single EBS volume as its
+// root device volume.
+func TestNodeVolumeStats(t *testing.T) {
+	skipIfLsblkVerIsOld(t)
```
fierlion, line 43: nit: I love this method name.  Should it be `skipForUnsupportedLsblk(t)` and the supported lsblk version could be a const in this file so we could update?
```
lsblkRequredMajorVersion = 2
lsblkRequireMinorVersion = 2
lsblkRequiredPatchVersion = 27
```

fierlion (approved):

ecs-agent/csiclient/csi_client_linux_integ_test.go
amogh09: Oh you mean I can pass any random string? Let me try that.

amogh09: Right that works! I will update the PR. Thanks!

amogh09: Updated.

ecs-agent/csiclient/csi_client_linux_integ_test.go
amogh09: Sure! I made the update. I have added the `const`s in the `skipForUnsupportedLsblk` function itself as they are only used by that function. We can move them out when more functions need them.

ecs-agent/csiclient/csi_client_linux_integ_test.go
xxx0624: Yes exactly!

xxx0624 (approved):

fierlion (approved):

PR #4011, opened 2023-11-01
Implement task/container state change metadata getters in agent

timj-hh:
### Summary
This change implements two interfaces that were previously defined in this PR [#3943] and adds a unit test for each. These structs will be necessary for unified FE work later on, allowing the agent to pass this agnostic interface but access a proprietary container / task model.

### Implementation details
- `ContainerMetadataGetter` and `TaskMetadataGetter` interfaces are implemented as private structs in the `/agent/api` module. There is a public constructor and getter methods for each field of the struct.
- `TestTaskStateChangeMetadataGetter` and `TestContainerStateChangeMetadataGetter` create instances of these respective structs and validate that they return the correct values.

### Testing
In addition to two new passing tests, `make test` and `make run-integ-tests`

 Implement task/container state change metadata getters in agent

### Licensing

### DIFF
--- a/agent/api/statechange.go
+++ b/agent/api/statechange.go
@@ -26,9 +26,9 @@ import (
 	apitaskstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 	ni "github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/networkinterface"
-	"github.com/pkg/errors"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/pkg/errors"
 )
 
 // ContainerStateChange represents a state change that needs to be sent to the
--- a/agent/api/metadata_getter.go
+++ b/agent/api/metadata_getter.go
@@ -1,0 +1,90 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package api
+
+import (
+	"time"
+
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
+	apitask "github.com/aws/amazon-ecs-agent/agent/api/task"
+)
+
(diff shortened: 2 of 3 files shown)


agent/api/statechange.go
```diff
@@ -377,3 +377,72 @@ func (ts TaskStateChange) GetEventType() statechange.EventType {
...
+}
+
+// GetContainerRuntimeID returns the Docker ID of the container.
+func (csc *containerMetadataGetter) GetContainerRuntimeID() string {
+	return csc.container.GetRuntimeID()
+}
+
+// GetContainerIsEssential returns whether the container is essential
+// or not.
+func (csc *containerMetadataGetter) GetContainerIsEssential() bool {
```
RichaGangwar, line 410: Nit: cmg might be better name than csc.

timj-hh: ack, see new update

agent/api/statechange_test.go
```diff
@@ -361,3 +362,57 @@ func TestNewManagedAgentChangeEvent(t *testing.T) {
...
+	}
+
+	metadataGetter := NewTaskMetadataGetter(task)
+
+	change := &ecs.TaskStateChange{
+		MetadataGetter: metadataGetter,
+	}
+
+	assert.NotNil(t, change.MetadataGetter)
+	assert.Equal(t, change.MetadataGetter.GetTaskIsNil(), false)
```
danehlim, line 389: The last 2 arguments should be swapped here, right? (i.e., `false` is the expected and `change.MetadataGetter.GetTaskIsNil()` is the actual)

Same comment applies for all other lines calling `assert.Equal`

agent/api/statechange.go
```diff
@@ -377,3 +377,72 @@ func (ts TaskStateChange) GetEventType() statechange.EventType {
 func (AttachmentStateChange) GetEventType() statechange.EventType {
 	return statechange.AttachmentEvent
 }
+
+// Implementation of the ContainerStateChange ContainerMetadataGetter Interface.
+type containerMetadataGetter struct {
```
danehlim, line 382: I'm thinking it might be better to put these metadata getters (i.e., `containerMetadataGetter`, `taskMetadataGetter`) in new file `agent/api/metadata_getter.go` and their tests in new file `agent/api/metadata_getter_test.go` for organizational purposes (i.e., they will still belong to the `api` package defined in agent module, but categorized into these files separate from `statechange.go` and `statechange_test.go` and thus can be found easier). What do you think?

timj-hh: My understanding is that statechange.go would be moved entirely to `ecs-agent`, and the remaining methods were helpers for those structs. These metadata setters are a part of the statechange struct.

But I am happy to move them, since I will already push a new rev I can do this as well.

agent/api/statechange_test.go
timj-hh: ack

RichaGangwar (approved):

danehlim (approved):

PR #4019, opened 2023-11-07
Decouple ENI in ecs-agent/ AttachmentStateChange

danehlim:
### Summary

Update AttachmentStateChange in ecs-agent module to take into account the ENI decoupling from https://github.com/aws/amazon-ecs-agent/pull/3969 (since that pull request did not update the AttachmentStateChange type that was added to ecs-agent module as part of https://github.com/aws/amazon-ecs-agent/pull/3943).

### Implementation details

Same as summary.

### Testing

New tests cover the changes: no, but existing test modified

Decouple ENI in ecs-agent/ AttachmentStateChange

No

### Licensing

### DIFF
--- a/agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/ecs/statechange.go
+++ b/agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/ecs/statechange.go
@@ -18,6 +18,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/aws/amazon-ecs-agent/ecs-agent/api/attachment"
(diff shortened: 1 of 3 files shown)


timj-hh (approved):
nice, thanks for picking this up!

RichaGangwar (commented):
Do we need to update the mocks or are we good there?

danehlim:
@RichaGangwar 
> Do we need to update the mocks or are we good there?

Thanks for taking a look! No we should not need to update any mocks here since no interfaces that we generate mocks from were modified. We can also confirm this by running `make gogenerate` Makefile target and seeing that no diff is produced.

RichaGangwar (approved):

PR #3109, opened 2021-12-17
DiscoverPollEndpoint: lengthen cache ttl and improve resiliency

sparrc:
The acs/tacs endpoints that ecs agent connects have never changed
before, so the current behavior of calling the API every 20 minutes is
not necessary. Lengthening this cache TTL to 12 hours will reduce load
on the ECS service and reduce the chance of customers being throttled.

Additionally, the LRU cache we were using automatically evicted any
value once it expired. This means that this API call could become a
source of complete ecs agent failure in the event of an LSE, in the
event that agent threw away it's cached endpoint and then failed to get
a new one via DiscoverPollEndpoint.

By keeping expired values in the cache, we can fallback to use the
expired value in the event that the DiscoverPollEndpoint API is failing,
thus keeping agent connected to ACS and TACS.

### Testing

- functional tests

New tests cover the changes: yes, ttl cache unit tests

- manually tested scenario where DiscoverPollEndpoint is failing and our cache has expired, and saw expected output: 
```
level=info time=2021-12-17T22:13:17Z msg="Error calling DiscoverPollEndpoint. Using cached but expired endpoint as a fallback. error=foo error endpoint=https://ecs-a-5.us-west-2.amazonaws.com/ telemetryEndpoint=https://ecs-t-5.us-west-2.amazonaws.com/ containerInstanceARN=arn:aws:ecs:us-west-2:039193556298:container-instance/usw2/66a059b9d0a440b78ec327e2d272de00" module=client.go
level=info time=2021-12-17T22:13:17Z msg="Establishing a Websocket connection to https://ecs-t-5.us-west-2.amazonaws.com/ws?agentHash=defd1ea8&agentVersion=1.57.1&cluster=usw2&containerInstance=arn%3Aaws%3Aecs%3Aus-west-2%3A039193556298%3Acontainer-instance%2Fusw2%2F66a059b9d0a440b78ec327e2d272de00&dockerVersion=20.10.7" module=client.go
level=info time=2021-12-17T22:13:17Z msg="Connected to TCS endpoint" module=handler.go
```

### Description for the changelog

Enhancement - DiscoverPollEndpoint: lengthen cache ttl and improve resiliency

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4020, opened 2023-11-08
Add ECS client to ecs-agent module

danehlim:
### Summary

Add ECS client to the ecs-agent module, as well as some related functionality.

For clarity, this pull request just includes changes to the ecs-agent module and does NOT change the agent module in any way to use these files (except for a few existing vendored files which are not directly a part of the ECS client). In a future pull request, we will raise changes in the agent module to consume the ECS client from the ecs-agent module and then remove the resulting unused ECS client related files from the agent module.

### Implementation details

- Add ECS client (i.e., `ecsClient` struct) to `ecs-agent/api/ecs/client/ecs_client.go` and test it in `ecs-agent/api/ecs/client/ecs_client_test.go`. This struct implements the `ECSClient` interface defined in `ecs-agent/api/ecs/interface.go`
- Use functional options design pattern to provide callers flexible ways to configure the ECS client. Implement and test pertinent options in the following files:
   - `ecs-agent/api/ecs/client/ecs_client_option.go`
   - `ecs-agent/api/ecs/client/ecs_client_option_test.go`
- Make default dial timeout, default dial keep alive, and default TLS handshake timeout constants in `ecs-agent/httpclient/httpclient.go` public so that they can be used in `ecs-agent/api/ecs/client/ecs_client.go`
- Create logger field constants for credentials ID, container instance ARN, attribute name, attribute value, endpoint, telemetry endpoint, and service connect endpoint in `ecs-agent/logger/field/constants.go` for use in `ecs-agent/api/ecs/client/ecs_client.go`
- Copy over the following files:
   - `ecs-agent/api/ecs/client/retry_handler.go` from `agent/api/ecsclient/retry_handler.go`
   - `ecs-agent/api/ecs/client/retry_handler_test.go` from `agent/api/ecsclient/retry_handler_test.go`
   - `ecs-agent/api/ecs/client/utils.go` from `agent/api/ecsclient/utils.go`
   - `ecs-agent/api/ecs/client/utils_amd64_test.go` from `agent/api/ecsclient/utils_amd64_test.go`
   - `ecs-agent/api/ecs/client/utils_arm64_test.go` from `agent/api/ecsclient/utils_arm64_test.go`

### Testing

Add ECS client to ecs-agent module

No

### Licensing

### DIFF
--- a/ecs-agent/api/ecs/client/ecs_client.go
+++ b/ecs-agent/api/ecs/client/ecs_client.go
@@ -1,0 +1,801 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package ecsclient
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"runtime"
+	"strings"
+	"time"
+
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/api/ecs"
+	ecsmodel "github.com/aws/amazon-ecs-agent/ecs-agent/api/ecs/model/ecs"
+	apierrors "github.com/aws/amazon-ecs-agent/ecs-agent/api/errors"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/async"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/config"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials/instancecreds"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/ec2"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/httpclient"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/utils"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/utils/retry"
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/credentials"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/docker/docker/pkg/meminfo"
+)
+
+const (
+	ecsMaxImageDigestLength     = 255
+	ecsMaxContainerReasonLength = 255
+	ecsMaxTaskReasonLength      = 1024
+	ecsMaxRuntimeIDLength       = 255
+	defaultPollEndpointCacheTTL = 12 * time.Hour
+	azAttrName                  = "ecs.availability-zone"
+	cpuArchAttrName             = "ecs.cpu-architecture"
+	osTypeAttrName              = "ecs.os-type"
+	osFamilyAttrName            = "ecs.os-family"
+	// RoundtripTimeout should only time out after dial and TLS handshake timeouts have elapsed.
+	// Add additional 2 seconds to the sum of these 2 timeouts to be extra sure of this.
+	RoundtripTimeout = httpclient.DefaultDialTimeout + httpclient.DefaultTLSHandshakeTimeout + 2*time.Second
+	// Below constants are used for SetInstanceIdentity retry with exponential backoff.
+	setInstanceIdRetryTimeOut         = 30 * time.Second
+	setInstanceIdRetryBackoffMin      = 100 * time.Millisecond
+	setInstanceIdRetryBackoffMax      = 5 * time.Second
+	setInstanceIdRetryBackoffJitter   = 0.2
+	setInstanceIdRetryBackoffMultiple = 2
+)
+
+// ecsClient implements ECSClient interface.
+type ecsClient struct {
+	credentialsProvider              *credentials.Credentials
+	configAccessor                   config.AgentConfigAccessor
+	standardClient                   ecs.ECSStandardSDK
+	submitStateChangeClient          ecs.ECSSubmitStateSDK
+	ec2metadata                      ec2.EC2MetadataClient
+	httpClient                       *http.Client
+	pollEndpointCache                async.TTLCache
+	isFIPSDetected                   bool
+	shouldExcludeIPv6PortBinding     bool
+	sascCustomRetryBackoff           func(func() error) error
+	stscAttachmentCustomRetryBackoff func(func() error) error
+}
+
+// NewECSClient creates a new ECSClient interface object.
+func NewECSClient(
+	credentialsProvider *credentials.Credentials,
+	configAccessor config.AgentConfigAccessor,
+	ec2MetadataClient ec2.EC2MetadataClient,
+	agentVer string,
+	options ...ECSClientOption) (ecs.ECSClient, error) {
+
+	client := &ecsClient{
+		credentialsProvider: credentialsProvider,
+		configAccessor:      configAccessor,
+		ec2metadata:         ec2MetadataClient,
+		httpClient:          httpclient.New(RoundtripTimeout, configAccessor.AcceptInsecureCert(), agentVer, configAccessor.OSType()),
+		pollEndpointCache:   async.NewTTLCache(&async.TTL{Duration: defaultPollEndpointCacheTTL}),
+	}
+
+	// Apply options to configure/override ECS client values.
+	for _, opt := range options {
+		opt(client)
+	}
+
+	ecsConfig := newECSConfig(credentialsProvider, configAccessor, client.httpClient, client.isFIPSDetected)
+	s, err := session.NewSession(&ecsConfig)
+	if err != nil {
+		return nil, err
+	}
+
+	if client.standardClient == nil {
+		client.standardClient = ecsmodel.New(s)
+	}
+	if client.submitStateChangeClient == nil {
+		client.submitStateChangeClient = newSubmitStateChangeClient(&ecsConfig)
+	}
+
+	return client, nil
+}
+
+func newECSConfig(
+	credentialsProvider *credentials.Credentials,
+	configAccessor config.AgentConfigAccessor,
+	httpClient *http.Client,
+	isFIPSEnabled bool) aws.Config {
+	var ecsConfig aws.Config
+	ecsConfig.HTTPClient = httpClient
+	ecsConfig.Credentials = credentialsProvider
+	ecsConfig.Region = aws.String(configAccessor.AWSRegion())
+	// We should respect the endpoint given (if any) because it could be the Gamma or Zeta endpoint of ECS service which
+	// don't have the corresponding FIPS endpoints. Otherwise, when the host has FIPS enabled, we should tell SDK to
+	// pick the FIPS endpoint.
+	if configAccessor.APIEndpoint() != "" {
+		ecsConfig.Endpoint = aws.String(configAccessor.APIEndpoint())
+	} else if isFIPSEnabled {
+		ecsConfig.UseFIPSEndpoint = endpoints.FIPSEndpointStateEnabled
+	}
+	return ecsConfig
+}
+
+// CreateCluster creates a cluster from a given name and returns its ARN.
+func (client *ecsClient) CreateCluster(clusterName string) (string, error) {
+	resp, err := client.standardClient.CreateCluster(&ecsmodel.CreateClusterInput{ClusterName: &clusterName})
+	if err != nil {
+		logger.Critical("Could not create cluster", logger.Fields{
+			field.Cluster: clusterName,
+			field.Error:   err,
+		})
+		return "", err
+	}
+	logger.Info("Successfully created a cluster", logger.Fields{
+		field.Cluster: clusterName,
+	})
+	return *resp.Cluster.ClusterName, nil
+}
+
+// RegisterContainerInstance calculates the appropriate resources, creates
+// the default cluster if necessary, and returns the registered
+// ContainerInstanceARN if successful. Supplying a non-empty container
+// instance ARN allows a container instance to update its registered
+// resources.
+func (client *ecsClient) RegisterContainerInstance(containerInstanceArn string, attributes []*ecsmodel.Attribute,
+	tags []*ecsmodel.Tag, registrationToken string, platformDevices []*ecsmodel.PlatformDevice,
+	outpostARN string) (string, string, error) {
+
+	clusterRef := client.configAccessor.Cluster()
+	// If our clusterRef is empty, we should try to create the default.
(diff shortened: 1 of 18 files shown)


ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+
+func findMissingAttributes(expectedAttributes, actualAttributes map[string]string) ([]string, error) {
+	missingAttributes := make([]string, 0)
+	var err error
+	for key, val := range expectedAttributes {
+		if actualAttributes[key] != val {
+			missingAttributes = append(missingAttributes, key)
+		} else {
+			logger.Trace("Response contained expected value for attribute", logger.Fields{
+				"key": key,
```
timj-hh, line 328: nit: prefer capital fields?

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+			// Except for RCI, TCP Ports are named as PORTS_TCP in Agent for Host Resources purpose.
+			resource.Name = aws.String("PORTS_TCP")
+		}
+		resourceMap[*resource.Name] = resource
+	}
+	return resourceMap, nil
+}
+
+func getCpuAndMemory() (int64, int64) {
+	memInfo, err := system.ReadMemInfo()
```
timj-hh, line 397: I believe `system.ReadMemInfo()` is deprecated in favor of `meminfo.Read()`

source: https://pkg.go.dev/github.com/moby/moby/pkg/system#ReadMemInfo

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+		})
+	}
+	return attributes
+}
+
+func (client *ecsClient) SubmitTaskStateChange(change ecs.TaskStateChange) error {
+	if change.Attachment != nil && client.stscAttachmentCustomRetryBackoff != nil {
+		retryFunc := func() error {
+			err := client.submitTaskStateChange(change)
+			if err == nil {
```
timj-hh, line 476: nit: we seem to prefer
```
if err != nil {
 // Code
}
return nil
```
Over
```
if err == nil {
  return nil
}
// Code
```

Thoughts?

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+
+func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error {
+	if change.Attachment != nil {
+		// Confirm attachment by submitting attachment state change via SubmitTaskStateChange API (specifically in
+		// the input's Attachments field).
+		var attachments []*ecsmodel.AttachmentStateChange
+		eniStatus := change.Attachment.Status.String()
+		attachments = []*ecsmodel.AttachmentStateChange{
+			{
+				AttachmentArn: aws.String(change.Attachment.AttachmentARN),
```
timj-hh, line 494: nit: If this is the `ni` type I think it is, it seems we prefer to access with the mutex locked `GetAttachmentArn()`?

https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/netlib/model/networkinterface/eniattachment.go#L160

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+		_, err := client.submitStateChangeClient.SubmitTaskStateChange(&ecsmodel.SubmitTaskStateChangeInput{
+			Cluster:     aws.String(client.configAccessor.Cluster()),
+			Task:        aws.String(change.TaskARN),
+			Attachments: attachments,
+		})
+		if err != nil {
+			logger.Warn("Could not submit task state change associated with confirming attachment",
+				logger.Fields{
+					field.Error:     err,
+					"attachmentARN": change.Attachment.AttachmentARN,
```
timj-hh, line 508: Same comment here

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+		ExecutionStoppedAt: change.ExecutionStoppedAt,
+		ManagedAgents:      change.ManagedAgents,
+		Containers:         formatContainers(change.Containers, client.shouldExcludeIPv6PortBinding, change.TaskARN),
+	}
+
+	_, err := client.submitStateChangeClient.SubmitTaskStateChange(&req)
+	if err != nil {
+		logger.Warn("Could not submit task state change", logger.Fields{
+			field.Error:       err,
+			"taskStateChange": change.String(),
```
timj-hh, line 533: Should this be a codified field in the fields package?

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced offline, in ecs-agent module thus far we have preferred uncapitalized fields.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Thanks for calling out, will update this.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced offline, this is not blocking and we will keep as is since it mirrors current existing code/behavior.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced offline, this is a non-blocking nit and we will keep as is since it mirrors current existing code/behavior.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced offline, this is a non-blocking nit and we will keep as is since it mirrors current existing code/behavior.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced offline, so far we have only been creating logger field constants for values that are called more than once. Since this one is only called once, we will leave as is.

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+	agentVer string,
+	osType string,
+	options ...ECSClientOption) (ecs.ECSClient, error) {
+
+	client := &ecsClient{
+		credentialsProvider: credentialsProvider,
+		configAccessor:      configAccessor,
+		ec2metadata:         ec2MetadataClient,
+		httpClient:          httpclient.New(RoundtripTimeout, configAccessor.AcceptInsecureCert(), agentVer, osType),
+		pollEndpointCache:   async.NewTTLCache(&async.TTL{Duration: defaultPollEndpointCacheTTL}),
```
RichaGangwar, line 96: We are setting the default as 12 hours. What happens when the containerd agent does not set any TTL? Currently, we do not expire our entries. We should set the TTL as nil by default and individual agents should set their TTL through the optional api.

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+	})
+	output, err := client.standardClient.DiscoverPollEndpoint(&ecsmodel.DiscoverPollEndpointInput{
+		ContainerInstance: &containerInstanceArn,
+		Cluster:           aws.String(client.configAccessor.Cluster()),
+	})
+	if err != nil {
+		// If we got an error calling the API, fallback to an expired cached endpoint if
+		// we have it.
+		if expired {
+			if output, ok := cachedEndpoint.(*ecsmodel.DiscoverPollEndpointOutput); ok {
```
timj-hh, line 714: If we rely on an expired endpoint the agent may not be able to form a connection with ACS. Is this practice already being done in `agent/`? Wondering why we wouldn't just want to stop the container?

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,802 @@
...
+	stscAttachmentCustomRetryBackoff func(func() error) error
+}
+
+// NewECSClient creates a new ECSClient interface object.
+func NewECSClient(
+	credentialsProvider *credentials.Credentials,
+	configAccessor config.AgentConfigAccessor,
+	ec2MetadataClient ec2.EC2MetadataClient,
+	agentVer string,
+	osType string,
```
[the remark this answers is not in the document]
danehlim: Good call out! I will see if I can factor out the osType parameter here.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: This is already being done in agent module, yes. See original introducing pull request here for rationale: https://github.com/aws/amazon-ecs-agent/pull/3109

> The acs/tacs endpoints that ecs agent connects have never changed
before, so the current behavior of calling the API every 20 minutes is
not necessary. Lengthening this cache TTL to 12 hours will reduce load
on the ECS service and reduce the chance of customers being throttled.
>
> Additionally, the LRU cache we were using automatically evicted any
value once it expired. This means that this API call could become a
source of complete ecs agent failure in the event of an LSE, in the
event that agent threw away it's cached endpoint and then failed to get
a new one via DiscoverPollEndpoint.
>
> By keeping expired values in the cache, we can fallback to use the
expired value in the event that the DiscoverPollEndpoint API is failing,
thus keeping agent connected to ACS and TACS.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Currently in agent module we default to 12 hours. Any calling Agent can override this with a different TTL (including nil TTL) through the `WithDiscoverPollEndpointCacheTTL` option. IMO it's better to stick with the current default that we have and override wherever necessary. In most cases, the point of using a TTL cache is for the TTL , so using a TTL cache with nil TTL seems a bit weird to me for the default case (i.e, using a nil TTL is more of an exception rather than the norm).

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,801 @@
...
+		resourceMap[*resource.Name] = resource
+	}
+	return resourceMap, nil
+}
+
+func getCpuAndMemory() (int64, int64) {
+	memInfo, err := meminfo.Read()
+	mem := int64(0)
+	if err == nil {
+		mem = memInfo.MemTotal / 1024 / 1024 // MiB
```
timj-hh, line 399: This comment may not be entirely accurate. Looking at the source for moby `meminfo.Read()`

https://github.com/moby/moby/blob/v24.0.7/pkg/meminfo/meminfo.go#L8-L10 Calls:

https://github.com/moby/moby/blob/master/pkg/meminfo/meminfo_linux.go#L38-L44

Which converts to and returns mem in KiB. It seems like this is returning GiB?

timj-hh (approved):

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced offline, we will keep a default TTL for the client to use for its TTL cache.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced offline, no further concern here.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Synced up offline. The memory being returned here is in MiB.

ref: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/s2-proc-meminfo

> MemTotal — Total amount of usable RAM, in kibibytes, which is physical RAM minus a number of reserved bits and the kernel binary code.

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,801 @@
...
+		registerRequest.InstanceIdentityDocumentSignature = &instanceIdentitySignature
+		return registerRequest
+	}
+
+	iidRetrieved := true
+	backoff := retry.NewExponentialBackoff(setInstanceIdRetryBackoffMin, setInstanceIdRetryBackoffMax,
+		setInstanceIdRetryBackoffJitter, setInstanceIdRetryBackoffMultiple)
+	ctx, cancel := context.WithTimeout(context.Background(), setInstanceIdRetryTimeOut)
+	defer cancel()
+	err := retry.RetryWithBackoffCtx(ctx, backoff, func() error {
```
RichaGangwar, line 272: This is a new behavior for containerd agent. Was this addressed in the design? I am unsure of the impact of this.

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,801 @@
...
+			})
+			// Force credentials to expire in case they are stale but not expired.
+			client.credentialsProvider.Expire()
+			client.credentialsProvider = instancecreds.GetCredentials(client.configAccessor.External())
+			return apierrors.NewRetriableError(apierrors.NewRetriable(true), attemptErr)
+		}
+		logger.Debug("Successfully retrieved Instance Identity Document")
+		return nil
+	})
+	if err != nil {
```
RichaGangwar, line 288: Under what scenario, do we return an error? AFAIU, the retry will backoff and continue trying. For containerd agent, I am not sure how it plays out. The agent tries once, returns and sets iidRetrieved false.

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -0,0 +1,801 @@
...
+
+	if iidRetrieved {
+		instanceIdentitySignature, err = client.ec2metadata.
+			GetDynamicData(ec2.InstanceIdentityDocumentSignatureResource)
+		if err != nil {
+			logger.Error("Unable to get instance identity signature", logger.Fields{
+				field.Error: err,
+			})
+		}
+	}
```
RichaGangwar, line 304: Why have we not opted for exponential backoff in this api call?

ecs-agent/api/ecs/client/ecs_client.go
danehlim: This was introduced as part of https://github.com/aws/amazon-ecs-agent/pull/3795 and is universally a good  mechanism to have in place to improve resiliency in the case of intermittent upstream failures. We can chat more about this offline if you need further details than what's listed in that pull request's description.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: > Under what scenario, do we return an error?

The function returns an error on L251 if unable to get the instance identity document.

> AFAIU, the retry will backoff and continue trying. For containerd agent, I am not sure how it plays out. The agent tries once, returns and sets iidRetrieved false.

Retry is done with exponential backoff for a maximum of 5 seconds. See L59-L64 and L268-272. Prior to https://github.com/aws/amazon-ecs-agent/pull/3795, Agent only tried once with no retry - so in the event of transient/intermittent upstream failures, this would be a single point of failure in Agent registering with ECS Control Plane.

ecs-agent/api/ecs/client/ecs_client.go
danehlim: We have not ran into any transient failures upstream with the call on L298 to motivate adding a retry mechanism for this call to get instance identity signature on L297-L298.

For the call to get instance identity document on L275 we ran into multiple issues of customers facing transient failures upstream, which was what motivated https://github.com/aws/amazon-ecs-agent/pull/3795.

RichaGangwar (approved):

ecs-agent/api/ecs/client/ecs_client.go
danehlim: Discussed offline, despite not having run into any transient failures upstream with the call to get instance identity signature, we will proactively add the same retry mechanism for this call in a follow up pull request to allow Agent to be more resilient in the event that we do run into any transient failures upstream with this call in the future.

PR #4032, opened 2023-11-16
Consume ECS client from ecs-agent module in agent module

danehlim:
### Summary

In the agent module, consume the [ECS client defined in ecs-agent module](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/api/ecs/client/ecs_client.go).

### Implementation details

- Replace all ECS client and corresponding mocks usages with the ECS client and corresponding mocks defined in ecs-agent module and update tests accordingly where necessary
- Make current `NewContainerMetadataGetter` and `NewTaskMetadataGetter` functions defined in `agent/api/metadata_getter.go` private as they are not to be used outside of the agent module's `api` package
- Write `ToECSAgent` translation methods for agent module level `TaskStateChange`, `ContainerStateChange`, and `AttachmentStateChange` state change types that convert to the state change types defined in the ecs-agent module
   - NOTE: Struct `ProtocolBindIP` and helper functions `buildManagedAgentStateChangePayload` and `buildContainerStateChangePayload` introduced as part of this pull request in `agent/api/statechange.go` are based off the components with the same names currently defined in `agent/api/ecsclient/client.go`
- Define new utils function `Int64PtrToIntPtr` in `ecs-agent/utils/utils.go` for use in `ToECSAgent` translation method
- In `ecs-agent/acs/session/session.go`, refactor `session` struct field `discoverEndpointClient` of type         `api.ECSDiscoverEndpointSDK` such that it is renamed to struct field `ecsClient` with type `ecs.ECSClient` instead
   - Type `api.ECSDiscoverEndpointSDK` is currently being used in ACS session but is no longer necessary now that there exists an ECS client in the ecs-agent module that we are consuming in agent module as part of this pull request
- Add a unit test in `agent/api/task/task_test.go` to assert that container port zero will explicitly not be included in the set of container ports that are exposed for a container
- Delete the following files as they are no longer being used:
   - `agent/api/ecsclient/client.go`
   - `agent/api/ecsclient/client_test.go`
   - `agent/api/ecsclient/retry_handler.go`
   - `agent/api/ecsclient/retry_handler_test.go`
   - `agent/api/ecsclient/utils.go`
   - `agent/api/ecsclient/utils_amd64_test.go`
   - `agent/api/ecsclient/utils_arm64_test.go`
   - `agent/api/generate_mocks.go`
   - `agent/api/interface.go`
   - `agent/api/mocks/api_mocks.go`
   -  `ecs-agent/api/generate_mocks.go`
   -  `ecs-agent/api/interface.go`
   -  `ecs-agent/api/mocks/api_mocks.go`

### Testing

Consume ECS client from ecs-agent module in agent module

No

### Licensing


agent/api/statechange.go
```diff
@@ -255,6 +267,33 @@ func (c *ContainerStateChange) String() string {
...
+	}
+
+	return &ecs.ContainerStateChange{
+		TaskArn:         c.TaskArn,
+		RuntimeID:       aws.StringValue(pl.RuntimeId),
+		ContainerName:   c.ContainerName,
+		Status:          c.Status,
+		ImageDigest:     aws.StringValue(pl.ImageDigest),
+		Reason:          aws.StringValue(pl.Reason),
+		ExitCode:        utils.Int64PtrToIntPtr(pl.ExitCode),
```
fierlion, line 291: Why would we be using an int64 for the exit code in the first place?

agent/api/statechange.go
```diff
@@ -23,14 +23,25 @@ import (
...
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/docker/go-connections/nat"
 	"github.com/pkg/errors"
 )
 
+const (
+	// ecsMaxNetworkBindingsLength is the maximum length of the ecs.NetworkBindings list sent as part of the
+	// container state change payload. Currently, this is enforced only when containerPortRanges are requested.
+	ecsMaxNetworkBindingsLength = 100
```
fierlion, line 42: is there any chance that this might be a behavior change for a customer?  
Is the 100 max limit already enforced in the control plane, or is this the first that we're enforcing [what our docs say](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_NetworkBinding.html)?

agent/api/statechange.go
```diff
@@ -349,16 +388,59 @@ func (change *TaskStateChange) String() string {
...
+		Status:             change.Status,
+		Reason:             change.Reason,
+		PullStartedAt:      change.PullStartedAt,
+		PullStoppedAt:      change.PullStoppedAt,
+		ExecutionStoppedAt: change.ExecutionStoppedAt,
+		MetadataGetter:     newTaskMetadataGetter(change.Task),
+	}
+
+	for _, managedAgentEvent := range change.ManagedAgents {
+		if mgspl := buildManagedAgentStateChangePayload(managedAgentEvent); mgspl != nil {
```
fierlion, line 405: nit: the `buildContainerStateChangePayload` signature returns an err, while the only failure case for the `buildManagedAgentStateChangePayload` seems to be when it shouldn't be reported to backend.  Are there any other error cases (`change.Container.Name` where container might be nil for example)?

agent/api/statechange.go
```diff
@@ -377,3 +459,129 @@ func (ts TaskStateChange) GetEventType() statechange.EventType {
...
+
+	// ContainerPortSet consists of singular ports, and ports that belong to a range, but for which we were not able to
+	// find contiguous host ports and ask docker to pick instead.
+	containerPortSet := change.Container.GetContainerPortSet()
+	// each entry in the ContainerPortRangeMap implies that we found a contiguous host port range for the same
+	containerPortRangeMap := change.Container.GetContainerPortRangeMap()
+
+	for _, binding := range change.PortBindings {
+		hostPort := int64(binding.HostPort)
+		containerPort := int64(binding.ContainerPort)
```
fierlion, line 547: if one or more of the containerPorts has a zero default value, we'll end up hitting the `else` statement below multiple times with the same zero port.  Just want to be sure we're accounting for all the duplicate key scenarios that might come up here.

fierlion: if this is already accounted for, can you add a test below to assert it?

agent/api/statechange.go
danehlim: > is there any chance that this might be a behavior change for a customer?

This is not a behavior change for a customer but rather persisting the current behavior, see current code: https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/ecsclient/client.go#L58-L60

> Is the 100 max limit already enforced in the control plane, or is this the first that we're enforcing [what our docs say](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_NetworkBinding.html)?

The limit was originally introduced as part of this pull request: https://github.com/aws/amazon-ecs-agent/pull/3479

I synced up with Anuj (the original code author of the above) who mentioned that, as of the time the above was implemented, control plane is enforcing the 100 max limit and that Agent is set to fail early in this case. We could remove this existing behavior on our Agent if we wanted to, but I think the drawback of that is that Agent may unnecessarily send container state changes to control plane that will result in failure.

agent/api/statechange.go
danehlim: This is due to how FE model defines type `ContainerStateChange`, see: https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/api/ecs/model/ecs/api.go#L9670

`pl` is of the aforementioned type and the aforementioned type's exit code struct field is of type `*int64`.

agent/api/statechange.go
danehlim: We should not encounter a scenario where multiple port bindings' container port is zero. ECS Control Plane only considers values are between 1 and 65535 as valid for containerPortRange. We also have a check in place in our Agent to ensure that port zero is not included in the set of container ports that are exposed for a container here: https://github.com/aws/amazon-ecs-agent/blob/master/agent/api/task/task.go#L1822

agent/api/statechange.go
danehlim: ACK, looking into this - may discuss more offline.

danehlim: Discussed offline: aligned that we will add a check that returns an error (which will be logged downstream) if the container state change lacks a container name.

agent/api/statechange.go
danehlim: Discussed offline: I will add a unit test for the Agent code I linked in my previous reply to assert that port zero will explicitly not be included in the set of container ports that are exposed for a container.

agent/api/statechange.go
```diff
@@ -377,3 +459,132 @@ func (ts TaskStateChange) GetEventType() statechange.EventType {
 func (AttachmentStateChange) GetEventType() statechange.EventType {
 	return statechange.AttachmentEvent
 }
+
+func buildManagedAgentStateChangePayload(change ManagedAgentStateChange) *ecsmodel.ManagedAgentStateChange {
+	if !change.Status.ShouldReportToBackend() {
```
mye956, line 464: Q: are there any required fields that we need to validate like we do in `buildContainerStateChangePayload`?

fierlion (approved):

agent/api/statechange.go
mye956: Er never mind it looks like ShouldReportToBackend is doing that

danehlim: I don't believe so - currently ManagedAgentStateChange structs are only initialized using the `NewManagedAgentChangeEvent` function defined on L209-L230. Tracing where and how this function is called, we expect that this function and its downstream callers do the necessary validation of a `ManagedAgentStateChange` prior to the point of it being passed into the `buildManagedAgentStateChangePayload` function (i.e., the relevant fields of any given `ManagedAgentStateChange` passed into the `buildManagedAgentStateChangePayload` function should be populated).

mye956 (approved):

fierlion (approved):
approve (post merge)

PR #4042, opened 2023-11-21
Trim managed agent reason + add retries for getting instance identity signature

danehlim:
### Summary

Add two enhancements to ECS client:
1. Trim managed agent reason to a max of 255 characters
   - This is not strictly required (since currently in practice Agent will not send a reason that exceeds this max), but is more of a safeguard in case Agent sends larger managed agent reasons in the future. For reference, we already have similar trimming logic in place for other "reason" fields such as container reason - we will reuse this logic by defining a reusable function
2. Add retries with exponential backoff for retrieving instance identity signature
   - This is a proactive measure in case we run into any intermittent upstream issues when retrieving instance identity signature. For reference, we already have similar retry with exponential backoff logic for retrieving instance identity document

### Implementation details

Same as above.

### Testing

New tests cover the changes: yes (existing test also modified and addresses testing bug)

Trim managed agent reason + add retries for getting instance identity signature

No

### Licensing

### DIFF
--- a/ecs-agent/api/ecs/client/ecs_client.go
+++ b/ecs-agent/api/ecs/client/ecs_client.go
@@ -294,8 +294,22 @@ func (client *ecsClient) setInstanceIdentity(
 	registerRequest.InstanceIdentityDocument = &instanceIdentityDoc
 
 	if iidRetrieved {
-		instanceIdentitySignature, err = client.ec2metadata.
-			GetDynamicData(ec2.InstanceIdentityDocumentSignatureResource)
+		ctx, cancel = context.WithTimeout(context.Background(), setInstanceIdRetryTimeOut)
+		defer cancel()
+		err = retry.RetryWithBackoffCtx(ctx, backoff, func() error {
+			var attemptErr error
+			logger.Debug("Attempting to get Instance Identity Signature")
+			instanceIdentitySignature, attemptErr = client.ec2metadata.
+				GetDynamicData(ec2.InstanceIdentityDocumentSignatureResource)
+			if attemptErr != nil {
+				logger.Debug("Unable to get instance identity signature, retrying", logger.Fields{
+					field.Error: attemptErr,
+				})
+				return apierrors.NewRetriableError(apierrors.NewRetriable(true), attemptErr)
+			}
+			logger.Debug("Successfully retrieved Instance Identity Signature")
+			return nil
+		})
 		if err != nil {
 			logger.Error("Unable to get instance identity signature", logger.Fields{
 				field.Error: err,
@@ -521,7 +535,7 @@ func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error
 		PullStartedAt:      change.PullStartedAt,
 		PullStoppedAt:      change.PullStoppedAt,
 		ExecutionStoppedAt: change.ExecutionStoppedAt,
-		ManagedAgents:      change.ManagedAgents,
+		ManagedAgents:      formatManagedAgents(change.ManagedAgents),
 		Containers:         formatContainers(change.Containers, client.shouldExcludeIPv6PortBinding, change.TaskARN),
 	}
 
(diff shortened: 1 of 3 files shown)


ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -752,6 +766,17 @@ func (client *ecsClient) UpdateContainerInstancesState(instanceARN string, statu
 	return err
 }
 
+func formatManagedAgents(managedAgents []*ecsmodel.ManagedAgentStateChange) []*ecsmodel.ManagedAgentStateChange {
+	var result []*ecsmodel.ManagedAgentStateChange
+	for _, m := range managedAgents {
+		if m.Reason != nil {
+			m.Reason = aws.String(trimString(aws.StringValue(m.Reason), ecsMaxContainerReasonLength))
```
amogh09, line 773: What happens when reason length exceeds the maximum? Does the backend throw a validation error or something like that?

ecs-agent/api/ecs/client/ecs_client_test.go
```diff
@@ -386,6 +386,8 @@ func TestRegisterContainerInstance(t *testing.T) {
 						Return("", errors.New("fake unit test error")),
 					mockEC2Metadata.EXPECT().GetDynamicData(ec2.InstanceIdentityDocumentResource).
 						Return(expectedIID, nil),
+					mockEC2Metadata.EXPECT().GetDynamicData(ec2.InstanceIdentityDocumentSignatureResource).
+						Return("", errors.New("fake unit test error")),
```
amogh09, line 390: Why is this change needed? IIUC other than retries the behavior remains the same?

ecs-agent/api/ecs/client/ecs_client.go
amogh09: > or reference, we already have similar trimming logic in place for other "reason" fields such as container reason

Is there any scope of reusing the existing function? If not, can we add a test for this function?

danehlim: > What happens when reason length exceeds the maximum? Does the backend throw a validation error or something like that?

Correct, error will occur due to length of reason being too large.

> Is there any scope of reusing the existing function? If not, can we add a test for this function?

Good call out, I am thinking it would be better to define a function to trim string values associated with string pointers (i.e., one that does something like `aws.String(trimString(aws.StringValue(INSERT_STRING_POINTER_HERE), INSERT_MAX_LENGTH_HERE))`), and then just reuse that function where we need to. I will do that and include testing of said function.

ecs-agent/api/ecs/client/ecs_client_test.go
danehlim: Thanks for this comment. Firstly (separate from the question at hand), it seems L383-L393 is never getting executed at the moment as L383 is currently never true.. I will address this bug in the next revision. 

Regarding the original question: assuming that L383-L393 DOES get executed, the change is to test the case of getting instance identity signature failing with some error and making sure that we do retry in such a case. The test now expects Agent to first try to retrieve the instance identity signature, mocks an error to be returned from doing so, and then expects Agent to try to retrieve the instance identity signature once again upon encountering that aforementioned error. Without the new retry backoff logic introduced to `ecs-agent/api/ecs/client/ecs_client.go`, this modified test would fail since it now requires that retrieving the instance identity signature is retried on errors.

ecs-agent/api/ecs/client/ecs_client.go
amogh09: Yep that will be great!

ecs-agent/api/ecs/client/ecs_client_test.go
amogh09: Ah I didn't notice the expectation added on L391. LGTM.

amogh09 (approved):

timj-hh (approved):

PR #4048, opened 2023-12-01
Generalize seelog config and add setters

timj-hh:
### Summary
This change generalizes the Seelog/global logger configuration. Currently, the timestamp format is hardcoded in, and it is unwieldy to change values such as the default output file, log format, etc. by hand as they are controlled by environment variables, and a logger `Config` struct. 

### Implementation details
- Adds a new `timestampFormat` field into the private logger Config struct. This will specify the timestamp format used by structured loggers (i.e. RFC 3339). A new default has been added to the value used in `agent` module, so there is no functional change introduced. Custom formatters now consume this Config attribute rather than the direct time format.
- New setter functions have been added to modify the Config struct for consumer facing logger config
 - `SetTimestampFormat`: Set the logger timestamp format. time.Format() will accept a time format struct, or something custom such as `SetConfigLogFile`.
 - `SetConfigMaxFileSizeMB`: Set the max file size of the logger before log rotation
 - `SetConfigLogFormat`: Set the format for logging(logfmt or json)
- `SetConfigLogFile`: Set the output file for logging.

### Testing

`TestLogfmtFormat_Structured_Timestamp` and `TestJSONFormat_Structured_Timestamp` have been introduced to test the ability to pass a custom time format. They will validate the correctness of a custom time format passed in. Existing tests verify that timestamp format remains consistent when untouched.

Generalize seelog config and add setters

No. Most of this PR adds setters, and what was abstracted has been replaced with a default used by the `agent` module.

### Licensing

### DIFF
--- a/ecs-agent/logger/log.go
+++ b/ecs-agent/logger/log.go
@@ -41,19 +41,21 @@ const (
 	DEFAULT_LOGLEVEL_WHEN_DRIVER_SET         = "off"
 	DEFAULT_ROLLOVER_TYPE                    = "date"
 	DEFAULT_OUTPUT_FORMAT                    = logFmt
+	DEFAULT_TIMESTAMP_FORMAT                 = time.RFC3339
 	DEFAULT_MAX_FILE_SIZE            float64 = 10
 	DEFAULT_MAX_ROLL_COUNT           int     = 24
 )
 
 type logConfig struct {
-	RolloverType  string
-	MaxRollCount  int
-	MaxFileSizeMB float64
-	logfile       string
-	driverLevel   string
-	instanceLevel string
-	outputFormat  string
-	lock          sync.Mutex
+	RolloverType    string
+	MaxRollCount    int
+	MaxFileSizeMB   float64
+	logfile         string
+	driverLevel     string
+	instanceLevel   string
+	outputFormat    string
+	timestampFormat string
+	lock            sync.Mutex
 }
 
 var Config *logConfig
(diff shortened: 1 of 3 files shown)


xxx0624 (approved):

JoseVillalta (approved):
LGTM

ecs-agent/logger/log.go
```diff
@@ -234,15 +236,59 @@ func setInstanceLevelDefault() string {
...
+	Config.lock.Lock()
+	defer Config.lock.Unlock()
+
+	os.Setenv(LOG_OUTPUT_FORMAT_ENV_VAR, logFormat)
+	Config.outputFormat = logFormat
+}
+
+// SetConfigMaxFileSizeMB sets the max file size of a log file
+// in Megabytes before the logger rotates to a new file.
+func SetConfigMaxFileSizeMB(maxSizeInMB float64) {
```
amogh09, line 259: The existing `SetLevel` function takes care of reloading the seelog config but the new setter functions do not. I understand that now there are multiple setter functions, a user might want to call the setter functions and then trigger a (re)load manually by calling `InitSeeLog` once but this breaks the current pattern that setter functions take care of reloading the configuration.

amogh09: Why not make the new setter functions call `reloadConfig`?

amogh09: Discussed offline that `reloadConfig` doesn't help with reloading log format.

timj-hh: Synced offline, Some setters will require a reload of the custom formatters, which is not something `reloadConfig` does. If we are always calling `InitSeeLog` after each setter anyways, the access pattern can be abstracted to call `InitSeeLog` manually after modifying desired config attributes

amogh09: Let's update the descriptions of the new setter functions to say that `InitSeeLog` needs to be called for the change to take effect.

amogh09: Can be done in a new PR.

amogh09 (approved):

PR #4049, opened 2023-12-03
Configure GENEVE interface for V2N workloads

samjkon:
### Summary

V2N tasks require creation of GENEVE interfaces in order to serve traffic. V2N tasks are for now only available on FoF. GENEVE interfaces are created on the baremetal hosts inside the task network namespace. A CNI plugin has to be executed during both READY and READY PULL transitions. The first execution is to create the GENEVE interface itself, the second is to create the TAP interface to attach to the microVM.

Few new unit tests are also added for existing code.

### Implementation details

A new function `configureGENEVEInterface` is added in `common_linux.go` file which handles configuration of interfaces of type V2N.

### Testing

`go test -tags unit ./agent/...`
`go test -tage unit ./ecs-agent/...`

Configure GENEVE interface for V2N workloads

No
### Licensing


ecs-agent/netlib/common_test.go
```diff
@@ -115,10 +115,10 @@ func getMultiNetNSMultiIfaceAWSVPCTestData(testTaskID string) (*ecsacs.Task, tas
 		ElasticNetworkInterfaces: enis,
 		Containers: []*ecsacs.Container{
 			{
-				NetworkInterfaceNames: []*string{aws.String(ifName1)},
+				NetworkInterfaceNames: []*string{aws.String(ifName2)},
```
tshan2001, line 118: May I ask why did we swap the order here?

ecs-agent/netlib/common_test.go
```diff
@@ -277,16 +277,16 @@ func getV2NTestData(testTaskID string) (*ecsacs.Task, tasknetworkconfig.TaskNetw
 		ElasticNetworkInterfaces: enis,
 		Containers: []*ecsacs.Container{
 			{
-				NetworkInterfaceNames: []*string{aws.String(primaryIfaceName)},
+				NetworkInterfaceNames: []*string{aws.String(secondaryIfaceName), aws.String(vethIfaceName)},
```
tshan2001, line 280: Same question as above

ecs-agent/netlib/data/network_client.go
```diff
@@ -14,23 +14,61 @@
...
+	"github.com/pkg/errors"
 	bolt "go.etcd.io/bbolt"
 )
 
 const (
 	networkNamespaceBucketName = "networkNamespaceConfiguration"
+	portBucketName             = "port"
+
+	// geneveDstPortMin and geneveDstPortMax specifies the range from which the GENEVE destination port will be selected.
+	// The range is currently set from 6081 to 6151 which gives us 70 ports per baremetal host to access the same subnet.
```
tshan2001, line 35: I thought baremetal host refers to the devices that we use for FoF. Isn't this change focusing on WP instances?

ecs-agent/netlib/data/network_client.go
```diff
@@ -14,23 +14,61 @@
...
+	// are 3 AZs per region with 2 cells. Each cell has 8 instances. We will also establish tunnels between the target
+	// subnet and at least 3 random connectivity providers per cell. This means we'll be able to spread customer
+	// tasks across at least 16 hosts and across 6 tunnels per AZ. Assuming even distribution, then this puts us
+	// at 69 uVMS per host accessing the same subnet (10000 / [6 * 8 * 3]).
+	// We start with 6081 since that is the default destination port for GENEVE interfaces.
+	geneveDstPortMin = 6081
+	geneveDstPortMax = 6151
+
+	// geneveDstPortDefault is the default port to be assigned if not taken already.
+	geneveDstPortDefault = 6081
```
tshan2001, line 49: Should we do `geneveDstPortDefault = geneveDstPortMin` instead since we mentioned that 6081 is the default value AND the value that we start the range with?

ecs-agent/netlib/data/network_client.go
```diff
@@ -94,3 +132,127 @@ func (ndc *networkDataClient) GetNetworkNamespace(netNSName string) (*tasknetwor
...
+	port, err := ndc.assignPort(geneveDstPortMin, geneveDstPortMax, vni, geneveDstPortDefault)
+	metricEntry.Done(err)
+
+	if errNoValue == errors.Cause(err) {
+		ndc.metricsFactory.New(metrics.V2NDestinationPortExhaustedMetricName).Done(err)
+	}
+	return port, err
+}
+
+func (ndc *networkDataClient) assignPort(min, max int, vni string, defaultValue int) (uint16, error) {
```
tshan2001, line 153: Should we place `defaultValue` next to min and max so we can declare all 3 of them as int together?

ecs-agent/netlib/data/network_client.go
```diff
@@ -94,3 +132,127 @@ func (ndc *networkDataClient) GetNetworkNamespace(netNSName string) (*tasknetwor
...
+}
+
+// findRandomUnusedValue returns a random integer value that was previously unused.
+// You may specify a defaultValue if you have a preferred value. If the defaultValue is already used,
+// it will return a random unused value from within the range. An example for when we use the defaultValue
+// is while assigning destination ports for GENEVE interfaces. By default we need to assign port 6081 unless
+// it is used by another existing interface. If 6081 is taken, we need to pick a random port from the specified
+// range of [min, max).
+// The function makes use of a data store bucket to store all the used values.
+func findRandomUnusedValue(bucket *bolt.Bucket, min, max, defaultValue int) (int, error) {
```
tshan2001, line 204: I don't think any of the comments in this file mentioned whether this will be invoked concurrently. Is it possible for multiple clients to be invoking this method and accessing the same DB at the same time? If that's the case, should we add a semaphore to protect shared resources (in this case, pool of available port numbers)?

ecs-agent/netlib/data/network_client_linux_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	require.Equal(t, requiredPortsCount, len(assigned))
+}
+
+// TestReleasePortNumber verifies the ReleaseGeneveDstPort operation releases port numbers
+// so as to be used again.
+func TestReleasePortNumber(t *testing.T) {
+	c, cleanup := newTestClient(t)
+	defer cleanup()
+	vni := "123456"
```
tshan2001, line 60: Why did we use a different vni from the one in `TestNoDuplicatePortNumber`?

ecs-agent/netlib/platform/api.go
```diff
@@ -49,10 +51,21 @@ type API interface {
 	// GetNetNSPath returns the path of a network namespace.
 	GetNetNSPath(netNSName string) string
 
-	ConfigureInterface(ctx context.Context, netNSPath string, iface *networkinterface.NetworkInterface) error
+	// ConfigureInterface configures an interface inside a network namespace
+	// for it to be able to serve traffic.
+	ConfigureInterface(
+		ctx context.Context,
```
tshan2001, line 57: nit: doesn't look gofmt ed. Should we align the spaces here?

ecs-agent/netlib/platform/common_linux.go
```diff
@@ -609,6 +623,71 @@ func (c *common) configureBranchENI(ctx context.Context, netNSPath string, eni *
...
+	switch iface.DesiredStatus {
+	case status.NetworkReadyPull:
+		// Assign destination port and set device name for V2N ENI.
+		if err := ecscni.SetV2NDstPortAndDeviceName(iface, netDAO); err != nil {
+			return err
+		}
+		cniNetConf = NewTunnelConfig(netNSPath, iface, VPCTunnelInterfaceTypeGeneve)
+	case status.NetworkReady:
+		cniNetConf = NewTunnelConfig(netNSPath, iface, VPCTunnelInterfaceTypeTap)
+	case status.NetworkDeleted:
```
tshan2001, line 645: Seems like these two cases are doing the same thing. Instead of 
```
case a:
case b:
```
can we merge them together by doing `case a, b: `?

ecs-agent/netlib/common_test.go
samjkon: The customer can specify the order of interfaces however they wish. This test ensures we do not make any assumptions in our code against that possibility.

ecs-agent/netlib/data/network_client.go
samjkon: We are importing code for firecracker platform also into the shared library to have all the network related code in one place. But we have not made any modification to FoF workflow.

ecs-agent/netlib/platform/common_linux.go
samjkon: The 3rd argument is different for both cases.

ecs-agent/netlib/platform/api.go
samjkon: Thanks for pointing it out. Will fix it.

ecs-agent/netlib/data/network_client.go
```diff
@@ -94,3 +132,127 @@ func (ndc *networkDataClient) GetNetworkNamespace(netNSName string) (*tasknetwor
 
 	return netNS, nil
 }
+
+// AssignGeneveDstPort returns an unused destination port number for GENEVE interfaces.
+// By default for a particular VNI, it will return the default GENEVE destination port - 6081.
+// In case port 6081 is taken by another interface using the same VNI, it will chose a
+// random port from within the pre-configured range.
```
Realmonia, line 139: What's the advantage of choosing a random port instead of a next available port?

ecs-agent/netlib/model/ecscni/vpctunnel_config.go
```diff
@@ -68,12 +71,18 @@ func (c *VPCTunnelConfig) PluginName() string {
 
 // SetV2NDstPortAndDeviceName assigns a destination port to the task ENI and assigns
 // it a device name with the pattern gnv<vni><dst port>.
-func SetV2NDstPortAndDeviceName(taskENI *networkinterface.NetworkInterface, dstPort uint16) {
-	vni := taskENI.TunnelProperties.ID
-	taskENI.TunnelProperties.DestinationPort = dstPort
```
Realmonia, line 73: Sorry I forgot - is the previous code in "unimplemented" state? Asked because this is a behavior change if we already use it.

ecs-agent/netlib/data/network_client.go
samjkon: I have removed implementation of NetworkDataClient from the shared library because the agent is able to inject the same into the network builder.

ecs-agent/netlib/data/network_client.go
samjkon: I have removed implementation of NetworkDataClient from the shared library because the agent is able to inject the same into the network builder.

ecs-agent/netlib/data/network_client.go
samjkon: I have removed implementation of NetworkDataClient from the shared library because the agent is able to inject the same into the network builder.

ecs-agent/netlib/platform/api.go
samjkon: I checked my IDE and looks like I am keeping the indentation as 4 spaces per tab. While viewing other pages on github, we see the same pattern where tab length seems to be 8 spaces instead of 4. So I think we do not have to change anything here. Besides lint checks pass, so I think we are good.

ecs-agent/netlib/data/network_client.go
```diff
@@ -14,23 +14,61 @@
...
+	// We start with 6081 since that is the default destination port for GENEVE interfaces.
+	geneveDstPortMin = 6081
+	geneveDstPortMax = 6151
+
+	// geneveDstPortDefault is the default port to be assigned if not taken already.
+	geneveDstPortDefault = 6081
+)
+
+var (
+	trueValue       = []byte("1")
```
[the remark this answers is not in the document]
samjkon: The reason why it is of type []byte is because we can only checkin byte data into the boltDB.

ecs-agent/netlib/data/network_client.go
samjkon: I have removed implementation of NetworkDataClient from the shared library because the agent is able to inject the same into the network builder.

To answer your question, maintaining a counter is takes more operations than choosing at random.

ecs-agent/netlib/model/ecscni/vpctunnel_config.go
samjkon: No this function was not being used until now.

ecs-agent/netlib/network_builder.go
```diff
@@ -216,25 +219,51 @@ func (nb *networkBuilder) configureNetNSInterfaces(ctx context.Context, netNS *t
 		logger.Debug("Configuring interface", logFields)
 		iface.DesiredStatus = netNS.DesiredState
 
-		err := nb.platformAPI.ConfigureInterface(ctx, netNS.Path, iface)
+		err := nb.platformAPI.ConfigureInterface(ctx, netNS.Path, iface, nb.networkDAO)
 		if err != nil {
-			return err
+			if netNS.DesiredState == status.NetworkDeleted {
+				logger.Error(fmt.Sprintf("Failed to delete interface: %v", err), logFields)
```
xxx0624, line 225: I think you are trying to say - Failed to configure interface instead of `delete`?

ecs-agent/netlib/network_builder.go
```diff
@@ -216,25 +219,51 @@ func (nb *networkBuilder) configureNetNSInterfaces(ctx context.Context, netNS *t
...
 	err := nb.configureNetNSInterfaces(ctx, netNS)
 	if err != nil {
-		return err
+		logger.Error(fmt.Sprintf("Failed to cleanup interfaces in netns: %v", err), logFields)
+		errs = multierror.Append(err, errs)
+	}
+
+	err = nb.platformAPI.DeleteNetNS(netNS.Path)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Failed to network namespace: %v", err), logFields)
```
xxx0624, line 264: Failed to delete network ns?

ecs-agent/netlib/platform/common_linux.go
```diff
@@ -22,6 +22,9 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
+
+	netlibdata "github.com/aws/amazon-ecs-agent/ecs-agent/netlib/data"
```
xxx0624, line 28: nit - remove empty line

ecs-agent/netlib/platform/common_linux.go
```diff
@@ -38,6 +41,7 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/volume"
 
 	"github.com/aws/aws-sdk-go/aws"
+	currentCNITypes "github.com/containernetworking/cni/pkg/types/100"
```
xxx0624, line 44: [non blocking comment] nit - golang pkg name prefer low cases without caps and underscores. so probably currecntcnitypes or cnitypes?

> Good package names are short and clear. They are lower case, with no under_scores or mixedCaps.
https://go.dev/blog/package-names

ecs-agent/netlib/network_builder.go
samjkon: That is a fair suggestion. I can add a field in the log message to state the desired status and say configure here.

ecs-agent/netlib/network_builder.go
samjkon: Thanks for catching.

ecs-agent/netlib/platform/common_linux.go
samjkon: Sure, can do.

ecs-agent/netlib/platform/common_linux.go
samjkon: Makes sense.

Realmonia (approved):

xxx0624 (approved):

tshan2001 (approved):

PR #4051, opened 2023-12-04
Enhance and standardize logger Config setters

timj-hh:
### Summary
This change enhances the access pattern for the Seelog configuration. Environment variables are currently used in `InitSeelog` when the agent is started to pass in logging values. After that, all changes to the logger should be managed by the Config struct.

This is a follow up to #4048.

### Implementation details
- Enhances existing setters by adding basic checks and mutex locks
- Splits the `SetLevel` setter into driver and instance level functions
- Calls `reloadConfig` in each setter
- Abstracts `logLevels` mapping from ECS_LOGLEVEL to Seelog level options to a file wide variable
- Adds a new variable to the Config, `logToStdout` that will decide whether we should duplicate output to stdout or not. Currently, we are outputting to a rolling log file as well as stdout. The latter should be optional.
- Changes the `seelogConfig` method with this `logToStdout` option. 
    - The stdout seelog config setting, and windows platform settings are wrapped in a <filter> tag. On Linux, we do not set the platform tag. If we do not output to stdout then the filter tag will be empty. This will cause logging to fail to apply to outputs. So, we should conditionally omit the filter tag if both of these tags are empty.
- Adds `SetRolloverType` setter to modify the rollover type of the logs, if any. The default behavior has been kept as size in the agent module, and this also falls back to date based rolling. Passing in 'none' allows for no rollover policy.
- Removes `timestampFormat` from the Config object. This var is special in that it is actively used in runtime when a custom log formatter is used. There is a potential data race with the mutex locked `Config` struct, so we can let this variable's setter directly access it without locks - knowing this will likely be used at startup and not runtime.
- Adds and refactors tests as described below

### Testing
Existing tests pass.

New unit tests `TestSeeLogConfig_WithoutStdout`, `TestSeelogConfig_JSONNoStdout`, and `TestSeelogConfig_JSONNoRollover` have been added to validate new logic.

Update logger Config setter access pattern

no model breaking changes

### Licensing

### DIFF
--- a/ecs-agent/logger/log.go
+++ b/ecs-agent/logger/log.go
@@ -44,18 +44,34 @@ const (
 	DEFAULT_TIMESTAMP_FORMAT                 = time.RFC3339
 	DEFAULT_MAX_FILE_SIZE            float64 = 10
 	DEFAULT_MAX_ROLL_COUNT           int     = 24
+	DEFAULT_LOGTO_STDOUT                     = true
 )
 
+// Because timestamp format will be called in the custom formatter
+// for each log message processed, it should not be handled
+// with an explicitly write protected configuration.
+var timestampFormat = DEFAULT_TIMESTAMP_FORMAT
+
+// logLevels is the mapping from ECS_LOGLEVEL to Seelog provided levels.
+var logLevels = map[string]string{
+	"debug": "debug",
+	"info":  "info",
+	"warn":  "warn",
+	"error": "error",
+	"crit":  "critical",
+	"none":  "off",
+}
+
 type logConfig struct {
-	RolloverType    string
-	MaxRollCount    int
-	MaxFileSizeMB   float64
-	logfile         string
-	driverLevel     string
-	instanceLevel   string
-	outputFormat    string
-	timestampFormat string
-	lock            sync.Mutex
+	RolloverType  string
+	MaxRollCount  int
+	MaxFileSizeMB float64
+	logfile       string
+	driverLevel   string
+	instanceLevel string
+	outputFormat  string
+	logToStdout   bool
+	lock          sync.Mutex
 }
 
 var Config *logConfig
@@ -89,7 +105,7 @@ func logfmtFormatter(params string) seelog.FormatterFunc {
 		buf.WriteString(level.String())
 		buf.WriteByte(' ')
 		buf.WriteString("time=")
-		buf.WriteString(context.CallTime().UTC().Format(Config.timestampFormat))
+		buf.WriteString(context.CallTime().UTC().Format(timestampFormat))
 		buf.WriteByte(' ')
 		// temporary measure to make this change backwards compatible as we update to structured logs
 		if strings.HasPrefix(message, structuredTxtFormatPrefix) {
@@ -114,7 +130,7 @@ func jsonFormatter(params string) seelog.FormatterFunc {
 		buf.WriteString(`{"level":"`)
 		buf.WriteString(level.String())
 		buf.WriteString(`","time":"`)
-		buf.WriteString(context.CallTime().UTC().Format(Config.timestampFormat))
+		buf.WriteString(context.CallTime().UTC().Format(timestampFormat))
 		buf.WriteString(`",`)
 		// temporary measure to make this change backwards compatible as we update to structured logs
 		if strings.HasPrefix(message, structuredJsonFormatPrefix) {
@@ -144,14 +160,26 @@ func reloadConfig() {
 }
 
 func seelogConfig() string {
+	driverLogChildren := []string{}
+	platformLogConfig := platformLogConfig()
+	if Config.logToStdout {
+		driverLogChildren = append(driverLogChildren, `	<console />`)
+	}
+	if platformLogConfig != "" {
+		driverLogChildren = append(driverLogChildren, platformLogConfig)
+	}
+
 	c := `
 <seelog type="asyncloop">
-	<outputs formatid="` + Config.outputFormat + `">
+	<outputs formatid="` + Config.outputFormat + `">`
+	if len(driverLogChildren) > 0 {
+		c += `
 		<filter levels="` + getLevelList(Config.driverLevel) + `">
-			<console />`
-	c += platformLogConfig()
-	c += `
+		`
+		c += strings.Join(driverLogChildren, "\n")
+		c += `
 		</filter>`
+	}
 	if Config.logfile != "" {
 		c += `
 		<filter levels="` + getLevelList(Config.instanceLevel) + `">`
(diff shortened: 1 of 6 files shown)


ecs-agent/logger/log.go
```diff
@@ -44,18 +44,34 @@ const (
 	DEFAULT_TIMESTAMP_FORMAT                 = time.RFC3339
 	DEFAULT_MAX_FILE_SIZE            float64 = 10
 	DEFAULT_MAX_ROLL_COUNT           int     = 24
+	DEFAULT_LOGTO_STDOUT                     = true
 )
 
+// Because timestamp format will be called in the custom formatter
+// for each log message processed, it should not be handled
+// with an explicitly write protected configuration.
+var timestampFormat = DEFAULT_TIMESTAMP_FORMAT
```
amogh09, line 53: Why do we need to move `timestampFormat` out of `logConfig`?

ecs-agent/logger/log.go
```diff
@@ -44,18 +44,34 @@ const (
...
-	timestampFormat string
-	lock            sync.Mutex
+	RolloverType  string
+	MaxRollCount  int
+	MaxFileSizeMB float64
+	logfile       string
+	driverLevel   string
+	instanceLevel string
+	outputFormat  string
+	logToStdout   bool
```
amogh09, line 73: Is this a new feature?

ecs-agent/logger/log.go
```diff
@@ -144,25 +160,35 @@ func reloadConfig() {
...
+	}
 	if Config.logfile != "" {
 		c += `
 		<filter levels="` + getLevelList(Config.instanceLevel) + `">`
 		if Config.RolloverType == "size" {
 			c += `
 			<rollingfile filename="` + Config.logfile + `" type="size"
 			 maxsize="` + strconv.Itoa(int(Config.MaxFileSizeMB*1000000)) + `" archivetype="none" maxrolls="` + strconv.Itoa(Config.MaxRollCount) + `" />`
-		} else {
+		} else if Config.RolloverType == "date" {
```
amogh09, line 185: This seems like a behavior change as before not setting `RolloverType` would default to date based rollover but not it won't. Is this intended?

ecs-agent/logger/log.go
```diff
@@ -236,62 +235,115 @@ func setInstanceLevelDefault() string {
...
+	} else {
+		seelog.Error("Driver log level mapping not found")
+	}
+}
+
 // SetConfigLogFile sets the default output file of the logger.
 func SetConfigLogFile(logFile string) {
-	Config.lock.Lock()
-	defer Config.lock.Unlock()
+	if logFile != "" {
```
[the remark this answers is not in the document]
timj-hh: I think so, as it is first set with a default and can be optionally reloaded later on. I think this should be an operator choice.

ecs-agent/logger/log.go
timj-hh: timestampFormat is accessed each time we log something. Because Config is protected by a mutex, a data race can occur where we are trying to R/W at the same time. Two ways we can get around this are:

1. Mutex locking/unlocking Config during each log write(already done at the seelog level with a different mutex lock, this could also be slow)
2. Making timestampFormat independent of Config. The check will make sure timestampFormat is valid and allow quick access.

ecs-agent/logger/log.go
timj-hh: Could you clarify? logToStdout is not necessarily new. It just makes the console receiver optional that is on by default.

Reference https://github.com/cihub/seelog/wiki/Receiver-reference

ecs-agent/logger/log.go
timj-hh: `DEFAULT_ROLLOVER_TYPE` is set to date. I can definitely switch this `if` and `else if` to speed up the access time to the date setting.

ecs-agent/logger/log.go
amogh09: This is a new setting on the logger. 

> Adds a new variable to the Config, logToStdout that will decide whether we should duplicate output to stdout or not. Currently, we are outputting to a rolling log file as well as stdout. The latter should be optional.

That sounds like a new feature to me. Why do we need this in this PR?

timj-hh: When consuming this change, I noticed that all logs were being logged to my desired config file as well as journal logs. This is not ideal and introduces clutter to our log streams. So users should be able to configure where their logs land, if anywhere.

ecs-agent/logger/log.go
amogh09: I think it's a behavior change for the logger as currently the logger defaults to date based rolling if the rollover type is not "size". I'd prefer that we keep the behavior of the logger same as before if this PR is about standardizing the configuration methods for the logger.

timj-hh: synced offline, we can ensure no behavioral change by keeping "date" with a rollingfile config as the default and checking for 'none' to use a simple log file.

ecs-agent/logger/log.go
amogh09: Discussed offline that this is a new feature that we want the logger to have.

ecs-agent/logger/log.go
```diff
@@ -236,62 +235,115 @@ func setInstanceLevelDefault() string {
...
 
-	// Parse unit as string to set as environment variable.
-	strsize := strconv.FormatFloat(maxSizeInMB, 'f', -1, 64)
-	os.Setenv(LOG_MAX_FILE_SIZE_ENV_VAR, strsize)
-	Config.MaxFileSizeMB = maxSizeInMB
+// SetRolloverType sets the logging rollover constraint.
+// This should be either size or date. Logger will roll
+// to a new log file based on this constraint.
+func SetRolloverType(rolloverType string) {
+	if rolloverType == "date" || rolloverType == "size" {
```
amogh09, line 301: "none" should be accepted too.

amogh09: Can we please add unit tests for the setters to make sure our validation works as expected?

ecs-agent/logger/log.go
```diff
@@ -144,21 +160,31 @@ func reloadConfig() {
...
-	c += platformLogConfig()
-	c += `
+		`
+		if Config.logToStdout {
+			c += `	<console />`
+		}
+		c += platformLogConfig
+		c += `
 		</filter>`
+	}
```
amogh09, line 177: If am finding this block a bit hard to read. Do you think the following is equivalent and reads better?
```go
stdoutFilterChildren := []string{}
if Config.logToStdout {
    stdoutFilterChildren = append(stdoutFilterChildren, ` <console />`)    
}
if platformLogConfig != "" {
    stdoutFilterChildren = append(stdoutFilterChildren, platformLogConfig)
}
if len(stdoutFilterChildren) > 0 {
    c += `
    <filter levels="` + getLevelList(Config.driverLevel) + `">
    `
    c += strings.Join(stdoutFilterChildren, "\n")
    c += `
    </filter>`
}
```

timj-hh: ack, will change to this approach

ecs-agent/logger/log.go
timj-hh: good catch, will update with some more unit tests

amogh09 (approved):

ohsoo (approved):

amogh09 (approved):

xxx0624 (approved):

PR #4053, opened 2023-12-08
Volume plugin changes to fix EFS unmount

amogh09:
### Summary

This PR updates Amazon ECS Volume Plugin, that is used for provisioning and deprovisioning EFS volumes on ECS container instances, so that it mounts and unmounts volumes when `Mount` and `Unmount` requests are made to it instead of `Create` and `Remove` requests, respectively. 

For EFS tasks in awsvpc network mode the task network namespace is torn down at task stop which causes EFS volume unmounting to hang when a `Remove` request is made to the plugin for the task's volumes at task cleanup time later. With this change, EFS volumes will be unmounted from the host when the task is being stopped and while task network namespace is still in place and configured. 

After this change, `Create`, `Mount`, `Unmount`, and `Remove` requests will be handled by the plugin as described below. 
* Create - Create mount path on the host and store information about the volume in plugin state but don't mount the volume yet. 
* Mount - If there are no active mounts for the volume then mount it on the host. Add the mount to the volume's set of active mounts in plugin state. 
* Unmount - Remove the mount from the volume's set of active mounts in plugin state. If the volume's active mount count has decreased to zero then unmount the volume from the host. 
* Remove - If the volume is still mounted on the host then unmount it (this is to handle unmounting of volumes created by older versions of the plugin that are still in state). Remove the mount path from the host and remove the volume from plugin's state.

### Implementation details

* Moved code for mounting an EFS volume from `Create` method to `Mount` method of `AmazonECSVolumePlugin`. Added relevant tests in `TestPluginMount` function. Also added roll-back steps to `Mount` method to make sure its effects are atomic.
* Added code for unmounting an EFS volume to `Unmount` method of `AmazonECSVolumePlugin`. Added relevant tests in `TestPluginUnmount` function. Also added roll-back steps to `Unmount` method to make sure its effects are atomic.
* Updated `Remove` method to make it unmount the volume only if it's currently mounted. Added a new `IsMounted` method to `VolumeDriver` interface. Implemented `IsMounted` method for `ECSVolumeDriver` that checks if the volume is present in driver's state. 
* Updated `Remove` method of `ECSVolumeDriver` to swallow unmount errors caused due to the volume not being mounted on the host. The method already had that logic as it swallows errors with "not mounted" in them but during my testing I found out that error "no mount point specified" is also returned by `umount` when the volume is not mounted. This is to make `Remove` method not fail for already unmounted volumes.
* Updated `Create` method of `ECSVolumeDriver` to not check its internal state for volume's mount state before mounting the requested volume. This is because `ECSVolumeDriver` does not track active mounts on the volume and an unmounted but not yet removed volume will reappear in its state if plugin restarts and with the check in place the driver will not allow the `Mount` request handler to perform volume mounts. The check does not server any purpose currently as the driver's internal state is completely controlled by `AmazonECSVolumePlugin` that has its own checks to prevent duplicate mount attempts. In future we should make `ECSVolumeDriver` completely stateless and have all state handling in `AmazonECSVolumePlugin` only but that's out of scope of this PR.

### Testing

Did extensive manual testing with the new plugin. Tested that - 
* awsvpc tasks clean up without issues ✅
* Host mode tasks work fine ✅
* Bridge mode tasks work fine ✅
* New version of the plugin is be able to unmount and remove volumes created by an older version of the plugin ✅
* Volume plugin restart between Unmount and Remove requests for a volume doesn't cause an error during Remove request ✅
* New plugin version is able to handle restarts without any issues  ✅
* Old version of the plugin is compatible with the state file created by the new version. ✅
* Tasks with multiple containers with EFS volumes work as expected ✅
* Task with multiple EFS volumes work as expected ✅
* A task with EFS volumes can be run multiple times on the same container instance without any issues ✅
* Tested that rollback works as expected for Mount request. For this test I deleted the parent directory of the state file directory of the plugin to make state save operations fail. ✅ 

A new functional test has been added that runs an AWSVPC EFS task twice in succession on an instance with a cleanup duration of 1 second. The test fails on instances with current version of the volume plugin but passes on instances with the version of the volume plugin in this PR.

Ran Agent functional tests on instances with a new version of volume plugin installed. All tests passed. 

Added many unit tests. 

Bugfix: Fix EFS unmount hanging issue for awsvpc tasks

### Licensing

### DIFF
--- a/ecs-init/volumes/ecs_volume_driver.go
+++ b/ecs-init/volumes/ecs_volume_driver.go
@@ -127,6 +127,8 @@ func (e *ECSVolumeDriver) Remove(req *driver.RemoveRequest) error {
 
 // Method to check if a volume is currently mounted.
 func (e *ECSVolumeDriver) IsMounted(volumeName string) bool {
+	e.lock.RLock()
+	defer e.lock.RUnlock()
 	_, exists := e.volumeMounts[volumeName]
 	return exists
 }
--- a/ecs-init/volumes/ecs_volume_driver.go
+++ b/ecs-init/volumes/ecs_volume_driver.go
@@ -66,10 +66,6 @@ func (e *ECSVolumeDriver) Create(r *driver.CreateRequest) error {
 	e.lock.Lock()
 	defer e.lock.Unlock()
 
-	if _, ok := e.volumeMounts[r.Name]; ok {
-		return fmt.Errorf("volume already exists")
-	}
-
 	mnt := setOptions(r.Options)
 	mnt.Target = r.Path
 
--- a/ecs-init/volumes/ecs_volume_driver.go
+++ b/ecs-init/volumes/ecs_volume_driver.go
@@ -115,3 +115,9 @@ func (e *ECSVolumeDriver) Remove(req *driver.RemoveRequest) error {
 	seelog.Infof("Unmounted volume %s successfully.", req.Name)
 	return err
 }
+
+// Method to check if a volume is currently mounted.
+func (e *ECSVolumeDriver) IsMounted(volumeName string) bool {
+	_, exists := e.volumeMounts[volumeName]
+	return exists
+}
--- a/ecs-init/volumes/ecs_volume_driver.go
+++ b/ecs-init/volumes/ecs_volume_driver.go
@@ -23,8 +23,16 @@ import (
 	"github.com/cihub/seelog"
 )
 
-// TODO: might be a good idea to check the mount point with a tool like mountpoint instead of matching error message.
-const notMountedErrMsg = "not mounted"
+const (
+	// TODO: might be a good idea to check the mount point with a tool like mountpoint instead of matching error message.
+	notMountedErrMsg = "not mounted"
+
+	// This error is returned when volume is already unmounted
+	// Example -
+	// $ sudo umount -l abc
+	// umount: abc: no mount point specified.
+	noMountPointSpecifiedErrMsg = "no mount point specified"
+)
 
 // ECSVolumeDriver holds mount helper and methods for different Volume Mounts
 type ECSVolumeDriver struct {
@@ -104,7 +112,8 @@ func (e *ECSVolumeDriver) Remove(req *driver.RemoveRequest) error {
 	}
 	err := mnt.Unmount()
 	if err != nil {
-		if strings.Contains(err.Error(), notMountedErrMsg) {
+		if strings.Contains(err.Error(), notMountedErrMsg) ||
+			strings.Contains(err.Error(), noMountPointSpecifiedErrMsg) {
 			seelog.Infof("Unmounting volume %s failed because it's not mounted.", req.Name)
 			delete(e.volumeMounts, req.Name)
 			return nil
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -190,6 +190,11 @@ func deleteMountPath(path string) error {
 func (a *AmazonECSVolumePlugin) Mount(r *volume.MountRequest) (*volume.MountResponse, error) {
 	seelog.Infof("Received mount request %+v", r)
 
+	// Validate the request
+	if len(r.ID) == 0 {
+		return nil, fmt.Errorf("no mount ID in the request")
+	}
+
 	// Acquire write lock
 	a.lock.Lock()
 	defer a.lock.Unlock()
@@ -247,6 +252,11 @@ func (a *AmazonECSVolumePlugin) Mount(r *volume.MountRequest) (*volume.MountResp
 // Unmount implements Docker volume plugin's Unmount Method
 func (a *AmazonECSVolumePlugin) Unmount(r *volume.UnmountRequest) error {
 	seelog.Infof("Received unmount request %+v", r)
+
+	// Validate the request
+	if len(r.ID) == 0 {
+		return fmt.Errorf("no mount ID in the request")
+	}
 
 	// Acquire write lock
 	a.lock.Lock()
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -297,7 +297,6 @@ func (a *AmazonECSVolumePlugin) Unmount(r *volume.UnmountRequest) error {
 	if len(vol.Mounts) == 0 {
 		seelog.Infof("No active mounts left on volume %s, unmounting it", r.Name)
 		if err := volDriver.Remove(&driver.RemoveRequest{Name: r.Name}); err != nil {
-			vol.AddMount(r.ID)
 			seelog.Errorf("Failed to unmount volume %v: %v", r.Name, err)
 			return fmt.Errorf("failed to unmount volume %v: %w", r.Name, err)
 		}
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -268,7 +268,7 @@ func (a *AmazonECSVolumePlugin) Unmount(r *volume.UnmountRequest) error {
 	a.lock.Lock()
 	defer a.lock.Unlock()
 
-	// Find  the volume
+	// Find the volume
 	vol, ok := a.volumes[r.Name]
 	if !ok {
 		seelog.Errorf("Volume %s to unmount is not found", r.Name)
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -336,6 +336,7 @@ func (a *AmazonECSVolumePlugin) Remove(r *volume.RemoveRequest) error {
 	// mounted. This is mainly to unmount volumes created by an older version of the
 	// plugin in which unmounts were not handled by Unmount method.
 	if volDriver.IsMounted(r.Name) {
+		seelog.Infof("Volume %s is currently mounted, unmouting it", r.Name)
 		if err := volDriver.Remove(&driver.RemoveRequest{Name: r.Name}); err != nil {
 			seelog.Errorf("Volume %s removal failure: %v", r.Name, err)
 			return err
--- a/ecs-init/volumes/ecs_volume_plugin.go
+++ b/ecs-init/volumes/ecs_volume_plugin.go
@@ -272,7 +272,10 @@ func (a *AmazonECSVolumePlugin) Unmount(r *volume.UnmountRequest) error {
 
 	// Remove the mount from the volume
 	seelog.Infof("Removing mount %s from volume %s", r.ID, r.Name)
-	vol.RemoveMount(r.ID)
+	if exists := vol.RemoveMount(r.ID); !exists {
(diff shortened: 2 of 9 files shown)


sparrc:
> Handle mounting and unmounting of EFS volumes at container start and stop instead of task start and cleanup, respectively, to fix EFS volume unmounting issue for awsvpc tasks.

bit of a nitpick but for the changelog I would word this as: `Bugfix: <briefly describe bug being fixed here>`

ecs-init/volumes/ecs_volume_driver.go
```diff
@@ -104,7 +112,8 @@ func (e *ECSVolumeDriver) Remove(req *driver.RemoveRequest) error {
 	}
 	err := mnt.Unmount()
 	if err != nil {
-		if strings.Contains(err.Error(), notMountedErrMsg) {
+		if strings.Contains(err.Error(), notMountedErrMsg) ||
+			strings.Contains(err.Error(), noMountPointSpecifiedErrMsg) {
```
sparrc, line 116: was it a separate bug that we were not handling this error before or is this a new error we might encounter due to this change?

ecs-init/volumes/ecs_volume_driver.go
```diff
@@ -115,3 +124,9 @@ func (e *ECSVolumeDriver) Remove(req *driver.RemoveRequest) error {
 	seelog.Infof("Unmounted volume %s successfully.", req.Name)
 	return err
 }
+
+// Method to check if a volume is currently mounted.
+func (e *ECSVolumeDriver) IsMounted(volumeName string) bool {
```
sparrc, line 129: should we have a read lock on this function?

sparrc: actually I see that it's called in the Remove function which would deadlock if this tried to get a read lock

ecs-init/volumes/ecs_volume_driver.go
amogh09: I am not sure when "not mounted" is printed by `umount`. If I manually unmount the volume and then ask the plugin to unmount it then a "no mount point specified" error is encountered here which is why I added this check.

ecs-init/volumes/ecs_volume_driver.go
amogh09: Yeah that's a miss. I will update this method to acquire a read lock. Thanks!

amogh09: I think it wouldn't deadlock as that `Remove` method is of `AmazonECSVolumePlugin` struct and not `ECSVolumeDriver` struct.

amogh09: Updated the method to acquire read lock.

ecs-init/volumes/ecs_volume_driver.go
amogh09: There is a niche case that if the plugin restarts between Unmount and Remove requests for a volume then the volume would reappear in the state of `ECSVolumeDriver` as it doesn't have a concept of active mounts. This error check prevents the driver's `Remove` method from returning an error in that case.

I think a proper fix is to not have any state in `ECSVolumeDriver` but that's out of scope of this PR.

sparrc (dismissed):

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -201,32 +188,148 @@ func deleteMountPath(path string) error {
...
+		// State update failed, so roll back the changes made so far to make state consistent
+		seelog.Errorf("Failed to save volume %s, rolling back changes: %v", r.Name, err)
+		vol.RemoveMount(r.ID)
+		if len(vol.Mounts) == 0 {
+			seelog.Warnf("Rolling back mounting of volume %s", r.Name)
+			if err := volDriver.Remove(&driver.RemoveRequest{Name: r.Name}); err != nil {
+				seelog.Errorf("Volume %s removal failure: %v", r.Name, err)
+			}
+		}
+		a.state.recordVolume(r.Name, vol)
```
yinyic, line 247: Maybe we don't need this line if the known status has failed to be updated in the first place?

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -201,32 +188,148 @@ func deleteMountPath(path string) error {
 
 // Mount implements Docker volume plugin's Mount Method
 func (a *AmazonECSVolumePlugin) Mount(r *volume.MountRequest) (*volume.MountResponse, error) {
```
yinyic, line 190: nit - instead of `a` can we use `p` (for plugin) instead? Feels a little more intuitive that way. Same thing for using `d` for driver

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -201,32 +188,148 @@ func deleteMountPath(path string) error {
...
+	}
+
+	// If there are no more mounts left on the volume then unmount the volume from the host
+	if len(vol.Mounts) == 0 {
+		seelog.Infof("No active mounts left on volume %s, unmounting it", r.Name)
+		if err := volDriver.Remove(&driver.RemoveRequest{Name: r.Name}); err != nil {
+			vol.AddMount(r.ID)
+			seelog.Errorf("Failed to unmount volume %v: %v", r.Name, err)
+			return fmt.Errorf("failed to unmount volume %v: %w", r.Name, err)
+		}
```
yinyic, line 303: While it makes sense to roll back partial creation, it feels less common to roll back partial deletion. 

For creation, caller might give up on failure, which makes more sense for it to be an "all or none" operation as we do not want orphaned resources. Deletion, on the other hand, is generally best effort and okay with partial work. As long as we make sure the unmount and removal are idempotent requests, I don't see much benefit recreating the resources as a rollback

ecs-init/volumes/ecs_volume_plugin.go
amogh09: Yes. I will make this change.

ecs-init/volumes/ecs_volume_plugin.go
amogh09: I agree with you and will make this change. Thanks!

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -201,32 +188,148 @@ func deleteMountPath(path string) error {
...
+		seelog.Errorf("Volume %s's driver type %s not supported: %v", r.Name, vol.Type, err)
+		return nil, fmt.Errorf("Volume %s's driver type %s not supported: %w", r.Name, vol.Type, err)
+	}
+	if volDriver == nil {
+		// This case shouldn't happen normally
+		return nil, fmt.Errorf("no volume driver found for type %s", vol.Type)
+	}
+
+	// Mount the volume on the host if there are no active mounts for the volume.
+	if len(vol.Mounts) == 0 {
```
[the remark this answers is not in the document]
prateekchaudhry: Here and other places, should we check for cases for `len(vol.Mounts) != 0`  and `vol.Mounts[r.id]` might not exist? Or possibly use `IsMounted()` as suggested?

ecs-init/volumes/ecs_volume_plugin.go
```diff
@@ -201,32 +188,148 @@ func deleteMountPath(path string) error {
...
+	}
+	if len(r.ID) == 0 {
+		return fmt.Errorf("no mount ID in the request")
+	}
+
+	// Acquire write lock
+	a.lock.Lock()
+	defer a.lock.Unlock()
+
+	// Find  the volume
```
prateekchaudhry, line 271: nit: possibly extra whitespace

ecs-init/volumes/ecs_volume_plugin.go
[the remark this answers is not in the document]
amogh09: @prateekchaudhry In request handler for Mount request 
1. if `vol.Mounts[r.ID]` exists then the request handler will not mount the volume on the host as `len(vol.Mounts) > 0` and then `vol.AddMount(r.ID)` is basically a no-op as `vol.Mounts` is a set.
2. if `vol.Mounts[r.ID]` does not exist and if `len(vol.Mounts) == 0` then the request handler will mount the volume on the host

What other cases are there?

ecs-init/volumes/ecs_volume_plugin.go
amogh09: `a.state` is stateful and maintains a copy of the volumes in a `map`. When it fails to save the volumes to the file it simply returns the error to the caller without correcting its in-memory state. This call is to make `a.state`'s in-memory state consistent with `a`'s state. 

TBH `a.state` should be stateless and doesn't have an in-memory state at all. Same is true for `ECSVolumeDriver` struct which also has an unnecessary in-memory state where it keeps the volumes. I will add a task to make these two components stateless.

ecs-init/volumes/ecs_volume_plugin.go
amogh09: Actually all methods of `AmazonECSVolumePlugin` use `a` variable to reference a pointer to the struct... I think we should change all of them at once but I want to keep this PR's diff cleaner. Is it okay if I do it in another PR as a follow-up?

ecs-init/volumes/ecs_volume_plugin.go
amogh09: You mean above this comment? I added the newline above it on purpose...

ecs-init/volumes/ecs_volume_plugin.go
amogh09: Made this change.

ecs-init/volumes/ecs_volume_plugin.go
yinyic: Sounds good

ecs-init/volumes/ecs_volume_plugin.go
yinyic: Thanks! Should we remove the `vol.AddMount(r.ID)` (line 300) as well?

ecs-init/volumes/ecs_volume_plugin.go
amogh09: Errr let me fix the double space

amogh09: Fixed

ecs-init/volumes/ecs_volume_plugin.go
amogh09: Yes I missed that. Thank you! Removed.

ecs-init/volumes/ecs_volume_plugin.go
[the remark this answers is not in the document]
prateekchaudhry: Thanks, as discussed offline, `vol.Mounts` is unique for each task, and holds mounts for a task (as referenced from `vol, ok := a.volumes[r.Name]`). It is also the intended source of truth for mounted volumes for tasks here. The case `len(vol.Mounts) != 0` and `vol.Mounts[r.id]` which I was thinking about should not arise.

yinyic (approved):

prateekchaudhry (approved):

PR #4061, opened 2023-12-18
Add ECSClientFactory to create new ECS clients

danehlim:
### Summary

Small "nice to have" change to add an `ECSClientFactory` interface (and implementing struct `ecsClientFactory`) whose purpose is to create/initialize new ECS clients. 

This can help with testing in other code packages for any future tests we may want to add since it allows us to easily test and define what to expect during other packages' creation of ECS client/how the ECS client came to be in the first place.

### Implementation details

- Add `ECSClientFactory` interface to ecs-agent module and use it to create/initialize ECS client in agent module
- Generate mocks for `ECSClientFactory` interface

### Testing

Add ECSClientFactory to create new ECS clients

No

### Licensing


ecs-agent/api/ecs/client/ecs_client_factory.go
```diff
@@ -0,0 +1,61 @@
...
+	NewClient() (ecs.ECSClient, error)
+	// Credentials returns the credentials chain used by the ECS client.
+	Credentials() *credentials.Credentials
+}
+
+type ecsClientFactory struct {
+	credentialsProvider *credentials.Credentials
+	configAccessor      config.AgentConfigAccessor
+	ec2MetadataClient   ec2.EC2MetadataClient
+	agentVer            string
```
timj-hh, line 36: nit: I don't see the point of shortening this variable name, could we use `agentVersion` instead?

Also, could this be gleaned from a ConfigAccessor instead? Why does this need to be in the factory struct?

danehlim: > nit: I don't see the point of shortening this variable name, could we use agentVersion instead?

Sure, if you prefer I can change the variable name `agentVer` to `agentVersion` as part of any future revision addressing blocking changes.

> Also, could this be gleaned from a ConfigAccessor instead? Why does this need to be in the factory struct?

This needs to be in the factory struct because it is currently a function parameter of the `NewECSClient` function. It may be possible to refactor the ConfigAccessor interface and its implementers to take on a new method returning Agent version, but that's out of scope of this pull request.

timj-hh (approved):

ecs-agent/api/ecs/client/ecs_client_factory_test.go
```diff
@@ -0,0 +1,45 @@
...
+	agentVersion := "0.0.0"
+
+	clientFactory, ok := NewECSClientFactory(credentialsProvider, cfgAccessor, ec2MetadataClient,
+		agentVersion).(*ecsClientFactory)
+	assert.True(t, ok)
+	assert.Equal(t, credentialsProvider, clientFactory.credentialsProvider)
+	assert.Equal(t, cfgAccessor, clientFactory.configAccessor)
+	assert.Equal(t, ec2MetadataClient, clientFactory.ec2MetadataClient)
+	assert.Equal(t, agentVersion, clientFactory.agentVer)
+}
```
mye956, line 45: nit(non-blocking): Perhaps we could also test `NewClient()` as well as `Credentials()` calls just to have better test coverage. Happy for this to be a follow up.

ecs-agent/api/ecs/client/ecs_client_factory.go
```diff
@@ -0,0 +1,61 @@
...
+		agentVer:            agentVer,
+		options:             options,
+	}
+}
+
+func (f *ecsClientFactory) NewClient() (ecs.ECSClient, error) {
+	return NewECSClient(f.credentialsProvider, f.configAccessor, f.ec2MetadataClient, f.agentVer, f.options...)
+}
+
+func (f *ecsClientFactory) Credentials() *credentials.Credentials {
```
mye956, line 59: nit(non-blocking):This looks to be a getter function, perhaps we can name it as so? (i.e. GetCredentials)

ecs-agent/api/ecs/client/ecs_client_factory_test.go
danehlim: Thanks for your comment! I'll add this.

ecs-agent/api/ecs/client/ecs_client_factory.go
danehlim: Sure, I can rename the method name from `Credentials` to `GetCredentials`.

mye956 (approved):

ecs-agent/api/ecs/client/ecs_client_factory_test.go
danehlim: Thanks for your comment! I can add this.

timj-hh (approved):

PR #4067, opened 2023-12-20
Bug: Windows numCPU returning incorrectly for values above 64

arun-annamalai:
### Summary

This PR fixes a customer issue wherein Windows contiainer instances with vCPU larger than 64 were reporting incorrect CPU values. 

### Implementation details

The changes are implemented by using MSFT win32 API to get the correct number of CPUs on the instance. 

### Testing

This was tested by launching an EC2 instance with 128 vCPU and ensuring that the ECS console showed the right CPU count in the resources section. 

New tests cover the changes: 
yes, there are unit tests added. 

Bug: Windows numCPU returning incorrectly for values above 64

No
### Licensing

### DIFF
--- a/ecs-agent/utils/utils_windows.go
+++ b/ecs-agent/utils/utils_windows.go
@@ -1,0 +1,42 @@
+//go:build windows
+// +build windows
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
+
+import (
+	"runtime"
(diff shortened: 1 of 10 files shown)


ecs-agent/utils/utils_windows.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"runtime"
+
+	"golang.org/x/sys/windows"
+)
+
+var (
+	win32APIGetAllActiveProcessorCount = windows.GetActiveProcessorCount
+	golangRuntimeNumCPU                = runtime.NumCPU
+)
```
amogh09, line 28: Since we have some private variables, it might be better to have these and `GetNumCPU` function in a runtime specific utils package like `utils/runtime`. WDYT?

amogh09: Or we can do this when we have more utils. I have approved.

amogh09 (approved):

ecs-agent/utils/utils_windows.go
arun-annamalai: The private variables were more so just to mock out functions for the unit testing (no purpose other than that). I am fine either way, there was no dogma/reason I just put it in a generic utils folder. 

From my perspective I would like to keep it as is because I don't see utils/runtime really getting any more functions added to it and this current utils file is sparse as is. 

I am completely fine moving it as well!

yinyic (approved):

PR #4068, opened 2023-12-20
update credentials-fetcher docker file

saikiranakula-amzn:
Remove unwanted dependencies from the dockerfile

### Testing

New tests cover the changes: 

### Licensing

### DIFF
--- a/ecs-agent/daemonimages/credentialsfetcher/Dockerfile
+++ b/ecs-agent/daemonimages/credentialsfetcher/Dockerfile
@@ -14,17 +14,15 @@
 # use minimal amazonlinux:2023 base for credentials-fetcher image
 FROM public.ecr.aws/amazonlinux/amazonlinux:2023-minimal
 
-RUN dnf install -y krb5-workstation jq unzip && mkdir /krb5 && chmod 755 /krb5
+RUN dnf install -y krb5-workstation
 RUN dnf install -y sssd
-RUN dnf install -y realmd
 RUN dnf install -y which
 RUN dnf install -y credentials-fetcher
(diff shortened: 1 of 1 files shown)


amogh09 (approved):

ecs-agent/daemonimages/credentialsfetcher/Dockerfile
```diff
@@ -14,17 +14,15 @@
 # use minimal amazonlinux:2023 base for credentials-fetcher image
 FROM public.ecr.aws/amazonlinux/amazonlinux:2023-minimal
 
-RUN dnf install -y krb5-workstation jq unzip && mkdir /krb5 && chmod 755 /krb5
+RUN dnf install -y krb5-workstation
 RUN dnf install -y sssd
-RUN dnf install -y realmd
```
danehlim, line 19: I'm assuming `RUN dnf install -y krb5-workstation` replaces all the previous commands associated with krb5 that are being removed as part of this pull request (please correct me if my assumption is wrong).

Curious why we are also removing the install realmd command here. Is it also included as part of krb5-workstation?

saikiranakula-amzn: only realmd is removed

danehlim (approved):

ecs-agent/daemonimages/credentialsfetcher/Dockerfile
danehlim: Synced offline, Sai confirmed that realmd is not needed for the use-case.

PR #4076, opened 2024-01-19
Fix netns model creation issue

samjkon:
### Summary

Fixes a bug in the creation of network namespace models for AWSVPC mode tasks which causes the agent to crash in an edge case situation.

### Implementation details

This change the condition which decides if the task requires a single network namespace or multiple. Previously, presence of ENI names in task payload indicated that the task multiple network namespaces, but we encountered a customer that included ENI names in the run task API for single netns tasks. This change also checks container's interface names to make the decision.

### Testing

Manually verified tasks start successful with the fix.
Added tests

New tests cover the changes: 
Yes

Fixes a bug in the creation of network namespace models for AWSVPC mode tasks which causes the agent to crash in an edge case situation.

No
### Licensing

### DIFF
--- a/ecs-agent/netlib/platform/common_linux.go
+++ b/ecs-agent/netlib/platform/common_linux.go
@@ -186,7 +186,8 @@ func (c *common) buildAWSVPCNetworkNamespaces(
 	// This case is identified if the singleNetNS flag is set, or if the ENIs have an empty 'Name' field,
 	// or if there is only on ENI in the payload.
 	if singleNetNS || len(taskPayload.ElasticNetworkInterfaces) == 1 ||
-		aws.StringValue(taskPayload.ElasticNetworkInterfaces[0].Name) == "" {
+		aws.StringValue(taskPayload.ElasticNetworkInterfaces[0].Name) == "" ||
+		len(taskPayload.Containers[0].NetworkInterfaceNames) == 0 {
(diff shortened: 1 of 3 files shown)


xxx0624 (approved):

ecs-agent/netlib/platform/common_linux.go
```diff
@@ -186,7 +186,8 @@ func (c *common) buildAWSVPCNetworkNamespaces(
 	// This case is identified if the singleNetNS flag is set, or if the ENIs have an empty 'Name' field,
 	// or if there is only on ENI in the payload.
 	if singleNetNS || len(taskPayload.ElasticNetworkInterfaces) == 1 ||
-		aws.StringValue(taskPayload.ElasticNetworkInterfaces[0].Name) == "" {
+		aws.StringValue(taskPayload.ElasticNetworkInterfaces[0].Name) == "" ||
+		len(taskPayload.Containers[0].NetworkInterfaceNames) == 0 {
```
xxx0624, line 190: I assume we don't have to check the length of the task containers.

samjkon: A task with no containers is invalid and there are validations for that early on in the code so we should be good to make this assumption.

tshan2001 (approved):

PR #4078, opened 2024-01-25
TMDS bugfix - v3 task metadata request can fail when container metadata update is pending

amogh09:
### Summary

There is a bug in v3 Task Metadata request handler that causes the request to fail with `500 Internal Server Error`. This happens when the request is processed while Agent only has partial information about one of task's containers and the missing metadata is pending an update from Docker. 

When Agent starts a container, the following events happen. 
1. Agent adds the container to task engine state by container's Docker Name. At this point the container exists in task engine state but Docker ID and some other information (like network metadata) about the container are not available.
https://github.com/aws/amazon-ecs-agent/blob/61c8a8c5b48e2efb7193fe61c4bb3cae816f5821/agent/engine/docker_task_engine.go#L1726-L1738
2. Agent asks Docker to start the container and waits for Docker to provide a Docker ID for the started container. 
3. Agent updates the task engine state so that the container is associated by its Docker ID instead of Docker Name.
https://github.com/aws/amazon-ecs-agent/blob/61c8a8c5b48e2efb7193fe61c4bb3cae816f5821/agent/engine/docker_task_engine.go#L1763-L1778
4. Agent updates container metadata in its state when it receives an event from Docker for the container. This also updates the container's state in Agent with its network metadata.
https://github.com/aws/amazon-ecs-agent/blob/61c8a8c5b48e2efb7193fe61c4bb3cae816f5821/agent/engine/task_manager.go#L440

If a v3 Task Metadata request is received between events (1) and (4), then the request fails with `500 Internal Server Error` as the request handler tries to look up the container by its Docker ID and also checks if its network metadata is available.
https://github.com/aws/amazon-ecs-agent/blob/61c8a8c5b48e2efb7193fe61c4bb3cae816f5821/agent/handlers/v3/task_metadata_handler.go#L70-L78 

This is not an issue for v2 task metadata request handler as it does not return network metadata or perform container lookup using the container's Docker ID.

This is not an issue for v4 task metadata request handler as it already handles this edge case. 
https://github.com/aws/amazon-ecs-agent/blob/61c8a8c5b48e2efb7193fe61c4bb3cae816f5821/agent/handlers/v4/tmdsstate.go#L133-L139

This PR updates v3 task metadata request handler so that a container whose Docker ID is not known or whose network metadata is not available in the state yet is skipped from being added to the response. This behavior already exists for containers that have not been added to task engine state yet (before event (1) from events listed above), so with this change the same behavior will continue until the container's network metadata is available (after event (4) from the events listed above).

### Implementation details

* Update v3 task metadata request handler so that it skips containers whose Docker ID or network metadata is not available instead of failing the request. 
* Updated the unit tests.

### Testing

I have an environment where I am able to reproduce the request failure. With this change the failure no longer happens and the container with unavailable network metadata is excluded from the response for v3 task metadata requests. 

New tests cover the changes:  no new tests but existing tests are updated

TMDS bugfix - v3 task metadata request can fail when container metadata is pending an update

### Licensing

### DIFF
--- a/agent/handlers/task_server_setup_test.go
+++ b/agent/handlers/task_server_setup_test.go
@@ -261,84 +261,13 @@ var (
 		ExecutionStoppedAtUnsafe: now,
 		LaunchType:               "EC2",
 	}
-	container1 = &apicontainer.Container{
-		Name:                containerName,
-		Image:               imageName,
-		ImageID:             imageID,
-		DesiredStatusUnsafe: apicontainerstatus.ContainerRunning,
-		KnownStatusUnsafe:   apicontainerstatus.ContainerRunning,
-		CPU:                 cpu,
-		Memory:              memory,
-		Type:                apicontainer.ContainerNormal,
-		KnownPortBindingsUnsafe: []apicontainer.PortBinding{
-			{
-				ContainerPort: containerPort,
-				Protocol:      apicontainer.TransportProtocolTCP,
-			},
-		},
-		NetworkModeUnsafe: bridgeMode,
-		NetworkSettingsUnsafe: &types.NetworkSettings{
-			DefaultNetworkSettings: types.DefaultNetworkSettings{
-				IPAddress: bridgeIPAddr,
-			},
-		},
-	}
-	bridgeContainer = &apicontainer.DockerContainer{
-		DockerID:   containerID,
-		DockerName: containerName,
-		Container:  container1,
-	}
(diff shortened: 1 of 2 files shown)


chienhanlin (approved):

Yiyuanzzz (approved):

amogh09:
Closing because we decided not to change TMDS v3 since it is a legacy TMDS version.

PR #4079, opened 2024-01-27
Certs base

lewayne-aws:
### Summary

With this PR, ecs-agent will pull in a newer ca-certificates package, from the current stable release of Debian, and will no longer attempt to remove the CNNIC certs, which have already been removed upstream

### Implementation details

The `stable-20211201-slim` image that is currently used to source `ca-certificates` is running Debian's `bullseye` release (a.k.a. `oldstable`), and as such, pulls in [ca-certificates version 20210119](https://packages.debian.org/bullseye/ca-certificates). By using a newer stable image (`stable-20240110-slim`), the newer [ca-certificates version 20230311](https://packages.debian.org/bookworm/ca-certificates) will be used.

I also noticed that the certs Dockerfile was performing an extra step to attempt to remove the `CNNIC_ROOT` certificate, however this cert [no longer appears to be present](https://packages.debian.org/bookworm/all/ca-certificates/filelist) in current versions of the package.

### Testing

`make test` was run. There were some upgrade related tests that failed there (`TestCNIPluginVersionUpgrade` mainly), however it appears to be the same result from an unmodified `master` branch.

New tests cover the changes:  no; There is no change in functionality to test

Update ca-certificates
  
No functionality changes were made.

### Licensing


sparrc:
Thanks for updating this @lewayne-aws, looks like there's no difference in the outputted certs config file from the updated container. I think we should merge this to stop doing the CNNIC hackaround and best practice of using an up-to-date container dependency.

Just as an aside, for reference on why to remove the CNNIC cert, see https://blog.mozilla.org/security/2015/03/23/revoking-trust-in-one-cnnic-intermediate-certificate/

I tested by building the container before and after this change and outputting the ca-certificates.conf file:
```
docker build -t certs .
docker run certs:latest cat /etc/ca-certificates.conf >/tmp/ca-certs
```

sparrc (approved):

sparrc:
Our codebuild CI project is having issues finding this source code for some reason, so I've cherry-picked the commits and reopened as https://github.com/aws/amazon-ecs-agent/pull/4153

Issue #4089, opened 2024-02-07
[ECS Anywhere] Latest ecs-agent/ecs-init breaks compatibility, EBS mount error, v1.78.0 - v1.81.0


frehov:
### Summary

Trying to upgrade to latest version 1.81.0, breaks our install on an external instance.

We will be rolling back and pinning to before 1.78.0.

### Description

During an upgrade attempt of our docker version and ecs-agent version, our ssm-documents kept failing.
Running the upgrade manually according to the [ecs-anywhere upgrade guide](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-anywhere-updates.html) is kept failing to start, and inspecting the logs, it had started looping during the PreStart step. with the logs posted below.

Going through the history, this seems to have been introduced from version 1.78.0 with from the inclusion of EBS task volumes, https://github.com/aws/amazon-ecs-agent/commit/b9d174a8b602d436aa0c912bd65860f42f3d7103.

We believe this should be toggleable, or should be hidden behind `config.runningInExternal` at this point in the engine init https://github.com/aws/amazon-ecs-agent/blob/master/ecs-init/engine/engine.go#L146 as it currently is breaking ecs-anywhere (external) instances.

edit: our /mnt folder on the hosts are read-only and cannot be modified, I believe this presents a missing testcase and thus not covered by the testing done in https://github.com/aws/amazon-ecs-agent/pull/3952 which mentions testing with ecs-anywhere-install.sh

### Expected Behavior
Upgrade/installs successfully, and starts up without issue

### Observed Behavior
Upgrades the package, and fails during PreStart of the service with:
`could not create EBS mount directory: mkdir /mnt/ecs: read-only file system`

### Environment Details

Docker info:  
```
Client: Docker Engine - Community
 Version:    25.0.3
 
Server:
 Containers: 2
  Running: 0
  Paused: 0
  Stopped: 2
 Images: 5
 Server Version: 25.0.3
 Storage Driver: overlay2
  Backing Filesystem: xfs
  Supports d_type: true
  Using metacopy: false
  Native Overlay Diff: true
  userxattr: false
 Logging Driver: json-file
 Cgroup Driver: cgroupfs
 Cgroup Version: 1
 Plugins:
  Volume: local
  Network: bridge host ipvlan macvlan null overlay
  Log: awslogs fluentd gcplogs gelf journald json-file local splunk syslog
 Swarm: inactive
[...]

PR #4091, opened 2024-02-09
Use ECS_EBSTA_SUPPORTED env variable to add EBSTA capabilities

prateekchaudhry:
### Summary

This is a fix for issue #4089

As part of ecs-init start up, init tries to create `/mnt/ecs/ebs` directory. However depending on how instances are configured, `/mnt` can be sometimes be read only. This leads to `ecs-agent` container failing to be created with following error in `ecs-init`
```
[ERROR] could not create EBS mount directory: mkdir /mnt/ecs: read-only file system
```

This PR changes this by adding an environment variable `ECS_EBSTA_SUPPORTED` to `ecs-agent`. If set to false, agent will not advertise EBSTA capabilities. This works as follows
* `ecs-init/engine/engine.go` - If the ebs mount directory `/mnt/ecs/ebs` fails to be created, we log this error in `pre-start` and continue
* `ecs-init/docker/docker.go` - Detect if ebs mount directory has been successfully created. If not set an environment variable `ECS_EBSTA_SUPPORTED` as `false` (Set default `true`)
* `agent/app/agent_capability.go` - Check the value of `ECS_EBSTA_SUPPORTED` and do not add EBSTA capabilities if the env var is set false. This is set as needed by the `ecs-init` changes. Agent defaults this to `true` for compatibility with older `ecs-init` versions

### Testing
* Mounting a Readonly file system at '/mnt' for an instance, able to launch agent and tasks
* Run Functional tests on AL2 and AL2023 AMIs with ecs-init artifacts which also test default ebs functionality

New tests cover the changes: 
No new tests added

Bugfix: Use `ECS_EBSTA_SUPPORTED` env variable to add EBSTA capabilities

No

### Licensing

### DIFF
--- a/ecs-init/engine/engine.go
+++ b/ecs-init/engine/engine.go
@@ -143,12 +143,11 @@ func (e *Engine) PreStart() error {
 		return engineError("could not create route to the credentials proxy", err)
 	}
 	// Add the EBS Task Attach host mount point
-	// Skip for External, EBS Task Attach is not supported for External launch type
-	if !config.RunningInExternal() {
-		err = os.MkdirAll(config.MountDirectoryEBS(), mountFilePermission)
-		if err != nil {
-			return engineError("could not create EBS mount directory", err)
-		}
+	err = os.MkdirAll(config.MountDirectoryEBS(), mountFilePermission)
+	if err != nil {
+		// Log error and continue
+		// If directory creation fails, set ECS_EBSTA_SUPPORTED=false in docker/docker.go
+		log.Error("could not create EBS mount directory", err)
 	}
 
 	docker, err := getDockerClient()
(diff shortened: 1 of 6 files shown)


yinyic (approved):

chienhanlin (dismissed):

ecs-init/engine/engine.go
```diff
@@ -143,9 +143,12 @@ func (e *Engine) PreStart() error {
 		return engineError("could not create route to the credentials proxy", err)
 	}
 	// Add the EBS Task Attach host mount point
-	err = os.MkdirAll(config.MountDirectoryEBS(), mountFilePermission)
-	if err != nil {
-		return engineError("could not create EBS mount directory", err)
+	// Skip for External, EBS Task Attach is not supported for External launch type
+	if !config.RunningInExternal() {
```
sparrc, line 147: This bug affects EC2 launch type too doesn't it? I am wondering if we should just log the failure and continue. Can we disable the EBS capability if we fail to create this directory?

sparrc: Also, when running in external we're going to advertise the capability anyways here? even though we failed to make the mnt directory?

sparrc (approved):

singholt (approved):

PR #4107, opened 2024-03-09
Pass clusterARN in task state change struct to account for Fargate FC tasks.

RichaGangwar:
### Summary

This PR fixes an issue in SubmitTaskStateChange API where the clusterARN populated in change request is incorrect. For one specific platform, the clusterARN is different for different tasks.

### Implementation details

1. Add clusterarn to taskstatechange struct.
2. In submitStatechange api, check if the clusterarn in taskstatechange struct is not nil. If not nil, then use the passed clusterarn instead of clusterarn from configaccessor.

### Testing

New tests cover the changes: 

Pass clusterARN in task state change struct.
  

### Licensing

### DIFF
--- a/ecs-agent/api/ecs/client/ecs_client.go
+++ b/ecs-agent/api/ecs/client/ecs_client.go
@@ -506,6 +506,9 @@ func (client *ecsClient) SubmitTaskStateChange(change ecs.TaskStateChange) error
 }
 
 func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error {
+
+	clusterARN := convertTaskToClusterARN(change.TaskARN)
+
 	if change.Attachment != nil {
 		// Confirm attachment by submitting attachment state change via SubmitTaskStateChange API (specifically in
 		// the input's Attachments field).
@@ -519,7 +522,7 @@ func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error
 		}
 
 		_, err := client.submitStateChangeClient.SubmitTaskStateChange(&ecsmodel.SubmitTaskStateChangeInput{
-			Cluster:     aws.String(client.configAccessor.Cluster()),
+			Cluster:     aws.String(clusterARN),
 			Task:        aws.String(change.TaskARN),
 			Attachments: attachments,
 		})
@@ -537,7 +540,7 @@ func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error
 	}
 
 	req := ecsmodel.SubmitTaskStateChangeInput{
-		Cluster:            aws.String(client.configAccessor.Cluster()),
+		Cluster:            aws.String(clusterARN),
 		Task:               aws.String(change.TaskARN),
 		Status:             aws.String(change.Status.BackendStatus()),
 		Reason:             aws.String(trimString(change.Reason, ecsMaxTaskReasonLength)),
@@ -550,7 +553,7 @@ func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error
 
 	_, err := client.submitStateChangeClient.SubmitTaskStateChange(&req)
 	if err != nil {
-		logger.Warn("Could not submit task state change", logger.Fields{
+		logger.Error("Could not submit task state change", logger.Fields{
 			field.Error:       err,
 			"taskStateChange": change.String(),
 		})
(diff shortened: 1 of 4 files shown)


ecs-agent/api/ecs/statechange.go
```diff
@@ -124,6 +126,9 @@ func (c *ContainerStateChange) String() string {
 // String returns a human readable string representation of a TaskStateChange.
 func (change *TaskStateChange) String() string {
 	res := fmt.Sprintf("%s -> %s", change.TaskARN, change.Status.String())
+	if len(change.ClusterARN) != 0 {
+		res += fmt.Sprintf(", ClusterARN: %s", change.ClusterARN)
```
timj-hh, line 130: nit(non-blocking): can we follow the format on L128 "%s -> %s" here?

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -507,8 +507,10 @@ func (client *ecsClient) SubmitTaskStateChange(change ecs.TaskStateChange) error
 
 func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error {
 
-	clusterARN := convertTaskToClusterARN(change.TaskARN)
-
+	clusterARN := client.configAccessor.Cluster()
```
timj-hh, line 510: It seems like the problem here is the cluster arn changing in between tasks, and that does not get updated in the config accessor. Wouldn't this be accessible from the config accessor so long as that were kept up to date? Why do we need an additional field on the TaskStateChange?

RichaGangwar: Firecracker can host multiple tasks on the same instance. So, clusterARN for different tasks will be different since we use task account's clusterARN. We cannot update the configaccessor as we would need to keep updating the member variable every time the API is called for the respective task. It's not that the config accessor is kept up to date in the lifecycle of the task, its just that it would keep differing for different tasks.

ecs-agent/api/ecs/statechange.go
RichaGangwar: Sure. Will fix it if there is any blocking comment.

ecs-agent/api/ecs/client/ecs_client.go
timj-hh: I see, thanks for explaining!

timj-hh (approved):

ecs-agent/api/ecs/statechange.go
```diff
@@ -124,6 +126,9 @@ func (c *ContainerStateChange) String() string {
 // String returns a human readable string representation of a TaskStateChange.
 func (change *TaskStateChange) String() string {
```
danehlim, line 127: nit: Can we update task state change string unit testing to account for case where` ClusterARN` is populated?

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -506,6 +506,11 @@ func (client *ecsClient) SubmitTaskStateChange(change ecs.TaskStateChange) error
 }
 
 func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error {
+
+	clusterARN := client.configAccessor.Cluster()
+	if len(change.ClusterARN) != 0 {
```
danehlim, line 511: nit: Prefer to use `change.ClusterARN != ""` in favor of `len(change.ClusterARN) != 0` for readability.

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -506,6 +506,11 @@ func (client *ecsClient) SubmitTaskStateChange(change ecs.TaskStateChange) error
 }
 
 func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error {
+
```
danehlim, line 509: nit: Unnecessary newline added here - not blocking to this pull request, but please make sure this is removed in a followup.

ecs-agent/api/ecs/client/ecs_client.go
```diff
@@ -519,7 +524,7 @@ func (client *ecsClient) submitTaskStateChange(change ecs.TaskStateChange) error
 		}
 
 		_, err := client.submitStateChangeClient.SubmitTaskStateChange(&ecsmodel.SubmitTaskStateChangeInput{
```
danehlim, line 526: nit: Let's look into adding a unit test for STSC that has as input a `TaskStateChange` with `ClusterARN` field populated. I understand that there is a level of urgency to merging the underlying change of this pull request, so this can be done in a followup pull request if not done in this one.

danehlim (approved):

ecs-agent/api/ecs/client/ecs_client.go
RichaGangwar: Sure. Will change in a follow-up CR.

ecs-agent/api/ecs/client/ecs_client.go
RichaGangwar: Sure.

ecs-agent/api/ecs/client/ecs_client.go
RichaGangwar: Sure.

PR #4140, opened 2024-04-14
Add functionality to pull image manifests to DockerClient

amogh09:
### Summary

This PR adds `PullImageManifest` method to `DockerClient` interface and adds an implementation of the method for `*dockerGoClient`. The method is supposed to get image manifest for the provided image reference and return the manifest as a value of `registry.DistributionInspect` type. 

A new integration test is added for `*dockerGoClient`'s implementation of `PullImageManifest` that checks that the implementation works for public and private registries. For the test, `scripts/setup-test-registry` script is updated to start a private registry on the host in addition to the public registry it already starts. 

### Implementation details

* A new method `PullImageManifest` is added to `DockerClient` interface. 
* The method is implemented for `*dockerGoClient` type. The implementation resolves registry credentials in the same way as `PullImage` and then calls Docker client's [DistributionInspect](https://pkg.go.dev/github.com/docker/docker/client#Client.DistributionInspect) API to fetch the image manifest from the registry. 
* A new integration test `TestImageManifestPullInteg` is added that checks that `PullImageManifest` method of `*dockerGoClient` is able to pull image manifests from public and private registries. `scripts/setup-test-registry` script is updated to start a private registry in addition to the public registry that it already starts. The private registry is used by the test. The private registry uses basic auth credentials to authenticate and authorize requests. The script generates an [htpasswd](https://httpd.apache.org/docs/2.4/programs/htpasswd.html) file which is provided to the registry. The file contains a single basic auth username/password pair where the username is "username" and password is "password". Password in the file is hashed using [bcrypt](https://en.wikipedia.org/wiki/Bcrypt) algorithm as required by the registry container. So, requests to the registry need to include basic auth credentials that match this username and password pair. 
* `sdkclient.Client` interface that is a wrapper around Docker client's API is updated to include `DistributionInspect` method. 

### Testing

Added new unit and integration tests.

Enhancement: Add image manifest pull functionality to DockerClient

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -52,6 +52,7 @@ import (
 	"github.com/docker/docker/api/types/events"
 	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/network"
+	"github.com/docker/docker/api/types/registry"
 	"github.com/docker/docker/api/types/volume"
 )
 
@@ -120,6 +121,10 @@ type DockerClient interface {
 	// ContainerEvents returns a channel of DockerContainerChangeEvents. Events are placed into the channel and should
 	// be processed by the listener.
 	ContainerEvents(context.Context) (<-chan DockerContainerChangeEvent, error)
+
+	// Given an image reference and registry auth credentials, pulls the image manifest
+	// of the image from the registry.
+	PullImageManifest(context.Context, string, *apicontainer.RegistryAuthenticationData) (registry.DistributionInspect, error)
 
 	// PullImage pulls an image. authData should contain authentication data provided by the ECS backend.
 	PullImage(context.Context, string, *apicontainer.RegistryAuthenticationData, time.Duration) DockerContainerMetadata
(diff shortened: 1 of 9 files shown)


agent/dockerclient/dockerapi/docker_client_integ_test.go
```diff
@@ -0,0 +1,128 @@
...
+			}(),
+		},
+		{
+			name:          "private registry auth failure",
+			dockerClient:  supportedClient,
+			imageRef:      "127.0.0.1:51671/busybox:latest",
+			expectedError: "no basic auth credentials",
+		},
+		{
+			name:         "public registry success",
```
mye956, line 85: Q: Should these success scenarios also have auth data to them?

mye956: Ah never mind these are public registries.

agent/dockerclient/dockerapi/docker_client_test.go
```diff
@@ -295,6 +298,143 @@ func TestPullImageECRSuccess(t *testing.T) {
...
+				client := mock_sdkclient.NewMockClient(ctrl)
+				client.EXPECT().
+					DistributionInspect(
+						gomock.Any(), "image", base64.URLEncoding.EncodeToString([]byte("{}"))).
+					Return(testDistributionInspect, nil)
+				f.EXPECT().GetDefaultClient().Return(client, nil)
+			},
+			expectedDistributionInspect: testDistributionInspect,
+		},
+		func() testCase {
```
mye956, line 371: TIL: Didn't know we could do this!

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -329,6 +334,34 @@ func (dg *dockerGoClient) time() ttime.Time {
...
 }
 
+// Pulls image manifest from the registry
+func (dg *dockerGoClient) PullImageManifest(
+	ctx context.Context, imageRef string, authData *apicontainer.RegistryAuthenticationData,
+) (registry.DistributionInspect, error) {
+	// Get auth creds
+	sdkAuthConfig, err := dg.getAuthdata(imageRef, authData)
+	if err != nil {
+		return registry.DistributionInspect{}, wrapPullErrorAsNamedError(imageRef, err)
```
mye956, line 344: Q: Just curious, any reasons why we can't pass in `nil` if there were errors?

mye956 (approved):

Yiyuanzzz (approved):

agent/dockerclient/dockerapi/docker_client_test.go
amogh09: Yes how cool is that!

amogh09: Allows tighter scoping of variables.

agent/dockerclient/dockerapi/docker_client.go
amogh09: Ah that's because this method has a value return type and not a pointer so the compiler won't allow returning `nil`. I just followed the method signature of Docker SDK which is 
```
DistributionInspect(ctx context.Context, imageRef, encodedRegistryAuth string) (registry.DistributionInspect, error)
```.

It is just like how functions that have a `(string, error)` return type would return `"", nil` for error cases and not a `nil`.

PR #4150, opened 2024-04-20
Add retries to DockerClient's manifest pull functionality and add utils/reference package

amogh09:
…a NamedError

### Summary

This is a follow-up to https://github.com/aws/amazon-ecs-agent/pull/4140 which added a new `PullImageManifest` method to `DockerClient` that can be used to pull image manifest from a registry. The changes in this PR are - 
* Add retry logic to the implementation of `PullImageManifest` so that transient issues do not cause transition failure. The retry strategy is the same as that for `PullImage`.
* Update return type of `PullImageManifest` so that it returns a more concrete `NamedError` type instead of the more general `error` type to make it in-line with other methods of the interface. 

In addition to the changes above, this PR also introduces a new `utils/reference` package that offers util functions that deal with container image references. The only function added to the package in this PR is `GetDigestFromImageRef` that can be used to parse a container image reference string and get a digest from it. This package will be used in an upcoming PR.

### Implementation details

* Update `PullImageManifest` method so that it returns a `apierrors.NamedError` instead of `error`.
* Add a new `manifestPullBackoff retry.Backoff` field to `dockerGoClient`. This is the retry strategy for manifest pulls. It is initialized with the same settings as the existing `imagePullBackoff` field that is the retry strategy for image pulls. The reason to use same retry strategy is that both methods call image registries.
* Update the implementation of `*dockerGoClient.PullImageManifest` so that it retries if `DistributionInspect` API fails. Also update the returned errors so that they are of type `apierrors.NamedError`. 
* Add `agent/utils/reference/reference.go` and add `GetDigestFromImageRef` function to it. The function implements logic to parse an image reference string and retrieve a digest from it if available. Add unit tests for this function.

### Testing

Enhancement: Add retries to DockerClient's image pull functionality and add utils/reference package.

### Licensing

### DIFF
--- a/agent/dockerclient/dockerapi/docker_client.go
+++ b/agent/dockerclient/dockerapi/docker_client.go
@@ -124,7 +124,7 @@ type DockerClient interface {
 
 	// Given an image reference and registry auth credentials, pulls the image manifest
 	// of the image from the registry.
-	PullImageManifest(context.Context, string, *apicontainer.RegistryAuthenticationData) (registry.DistributionInspect, error)
+	PullImageManifest(context.Context, string, *apicontainer.RegistryAuthenticationData) (registry.DistributionInspect, apierrors.NamedError)
 
 	// PullImage pulls an image. authData should contain authentication data provided by the ECS backend.
 	PullImage(context.Context, string, *apicontainer.RegistryAuthenticationData, time.Duration) DockerContainerMetadata
@@ -244,6 +244,7 @@ type dockerGoClient struct {
 	ecrTokenCache            async.Cache
 	config                   *config.Config
 	context                  context.Context
+	manifestPullBackoff      retry.Backoff
 	imagePullBackoff         retry.Backoff
 	inactivityTimeoutHandler inactivityTimeoutHandlerFunc
 
@@ -267,11 +268,12 @@ type ImagePullResponse struct {
 
 func (dg *dockerGoClient) WithVersion(version dockerclient.DockerVersion) (DockerClient, error) {
 	versionedClient := &dockerGoClient{
-		sdkClientFactory: dg.sdkClientFactory,
-		version:          version,
-		auth:             dg.auth,
-		config:           dg.config,
-		context:          dg.context,
+		sdkClientFactory:    dg.sdkClientFactory,
+		version:             version,
+		auth:                dg.auth,
+		config:              dg.config,
+		context:             dg.context,
+		manifestPullBackoff: dg.manifestPullBackoff,
 	}
 	// Check if the version is supported
 	_, err := versionedClient.sdkDockerClient()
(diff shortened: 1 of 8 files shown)


agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -337,29 +341,65 @@ func (dg *dockerGoClient) time() ttime.Time {
...
+	})
+
 	if err != nil {
-		err = redactEcrUrls(imageRef, err)
-		return registry.DistributionInspect{}, CannotPullContainerError{err}
+		return registry.DistributionInspect{}, wrapManifestPullErrorAsNamedError(imageRef, err)
+	}
+	if ctxErr := ctx.Err(); ctxErr != nil {
+		// Context was done while waiting for a retry attempt
+		return registry.DistributionInspect{}, wrapManifestPullErrorAsNamedError(imageRef, ctxErr)
```
danehlim, line 377: I noticed for `PullImage` that if context error is due to context's deadline being exceeded, that a `DockerTimeoutError` is used instead of `CannotPullContainerError` ([ref](https://github.com/aws/amazon-ecs-agent/blob/f8c0dd0932853e338c6124ee9e38bb29d3d0fbd1/agent/dockerclient/dockerapi/docker_client.go#L429-L432)). Do we want to do that here as well for consistency (i.e., use `DockerTimeoutError` instead of `CannotPullImageManifestError` in this case)?

amogh09: Good catch! I will update it to match `PullImage`.

agent/dockerclient/dockerapi/docker_client_integ_test.go
```diff
@@ -52,6 +54,9 @@ func TestImageManifestPullInteg(t *testing.T) {
 	if err != nil {
 		t.Skipf("Skipping test due to unsupported Docker version: %v", err)
 	}
+	// Make retries very fast
+	supportedClient.(*dockerGoClient).manifestPullBackoff = retry.NewExponentialBackoff(
```
mye956, line 58: Just curious, how long do manifest pulls usually take? Will there be a risk of flakiness here?

agent/dockerclient/dockerapi/docker_client.go
```diff
@@ -337,29 +341,65 @@ func (dg *dockerGoClient) time() ttime.Time {
...
-		return registry.DistributionInspect{}, wrapPullErrorAsNamedError(imageRef, err)
+		return registry.DistributionInspect{}, wrapManifestPullErrorAsNamedError(imageRef, err)
 	}
 	encodedAuth, err := registry.EncodeAuthConfig(sdkAuthConfig)
 	if err != nil {
-		return registry.DistributionInspect{}, wrapPullErrorAsNamedError(imageRef, err)
+		return registry.DistributionInspect{}, wrapManifestPullErrorAsNamedError(imageRef, err)
 	}
 
-	// Get an SDK Docker Client and call Distribution API
```
mye956, line 351: Q: Isn't this comment still true regarding calling the Distribution API?

danehlim: I initially had a similar thought, but looks like this information was moved to L361.

mye956: Oh I see. Disregard then!

danehlim (approved):

PR #4137, opened 2024-04-08
Add MANIFEST_PULLED internal container and task states

amogh09:
### Summary

This PR adds a new `ContainerState` named `ContainerManifestPulled` and a new `TaskState` called `TaskManifestPulled`. `ContainerManifestPulled` is added between `ContainerStatusNone` and `ContainerPulled` states and similarly `TaskManifestPulled` is added between `TaskStatusNone` and `TaskPulled` states. 

A new state transition function `pullContainerManifest` is added to `DockerTaskEngine` for handling transition of a container to `ContainerManifestPulled` state. The transition function is a no-op currently but will be updated in a future change to pull the manifest of the container image. This is a part of a project to expedite the reporting of container image manifest digests to ECS backend. 

Containers that want to reach steady state are allowed to transition to `ContainerManifestPulled` state under either of the following conditions - 
* If the container image pull depends on ASM ([AWS Secrets Manager](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/private-auth.html)) auth for private registry then ASM auth resource has been created. 
* If the container image pull depends on task execution role then task execution role credentials have been resolved. 
* All other cases.

[Container ordering dependencies](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDependency.html) are disregarded for transition to `ContainerManifestPulled` state. This is so that image manifest digests of all images of a task can be reported as soon as possible. 

`ContainerManifestPulled` state maps to `TaskManifestPulled` state when computing the known status of a task. 

None of the new states generate new state change events as of this PR, so container and tasks transitioning to these new states will not result in any new STSC calls. 

### Implementation details

* Add `ContainerManifestPulled ContainerStatus` constant to `containerstatus` package and update string mappings of `ContainerStatus` so that `ContainerManifestPulled` is mapped to "MANIFEST_PULLED" string. Update tests accordingly.
* Add `TaskManifestPulled TaskStatus` constant to `taskstatus` package and update string mappings of `TaskStatus` so that `TaskManifestPulled` is mapped to "MANIFEST_PULLED" string. Update tests accordingly.
* Add a new `pullContainerManifest` method to `DockerTaskEngine` type. This method is the transition function for transitioning a container to `ContainerManifestPulled` state. The method is currently a no-op. Update `DockerTaskEngine.containerStatusToTransitionFunction` map with a new `apicontainerstatus.ContainerManifestPulled: engine.pullContainerManifest` entry accordingly.
[...]

PR #4152, opened 2024-04-23
Implement transition function for MANIFEST_PULLED state

amogh09:
### Summary

This PR implements a transition function for container MANIFEST_PULLED internal state. This state was added in #4137 but the transition function was left unimplemented. This PR fills that gap. 

The transition function for MANIFEST_PULLED state tries to resolve the image manifest digest for the container image using the following logic and sets the `container.ImageDigest` field with it. 
1. If the container is an internally managed container then skip digest resolution as digest exists only for images pulled from a registry while images used by internally managed containers are not pulled from a registry.
1. If a digest exists in the image reference in task payload then use that.
1. Otherwise, if image pull is not required (due to the configured image pull behavior setting) then inspect the locally available image and get its manifest digest (aka repo digest). 
1. Otherwise, if the Docker engine version on the host supports API version 1.35 or later then pull image manifest from the image registry and get its digest. The dependency on API version 1.35 is because [Distribution API](https://docs.docker.com/engine/api/v1.35/#tag/Distribution/operation/DistributionInspect) used for fetching manifest from registries is only supported on API version 1.30 and does not support v1 registries. Support for v1 registries [was removed with Docker Engine v17.12](https://docs.docker.com/engine/deprecated/#--disable-legacy-registry-override-daemon-option) which was released [with API version 1.35](https://docs.docker.com/engine/api/#api-version-matrix). So, we are restricting this feature to Docker engines 17.12 and later to ensure that we are reliably able to get manifests from registries. A new timeout that can be configured with `ECS_MANIFEST_PULL_TIMEOUT` is used when calling the registry. 

In case of a failure during transition to MANIFEST_PULLED state a suitable error is generated which will become the container's [`ApplyingError`](https://github.com/aws/amazon-ecs-agent/blob/b702281f0f6aa2fba5614050769aa337d1456853/agent/api/container/container.go#L276-L279) as is the case with any other state transition failures. Except for timeout and authentication errors, `CannotPullImageManifestError` error is used to wrap errors during transition to `MANIFEST_PULLED` state. This aligns with the use of `CannotPullContainerError` to wrap errors during transition to `PULLED` state. For example, "CannotPullImageManifestError: Error response from daemon: manifest unknown: Requested image not found" would be the reason for container transition failure if transition to `MANIFEST_PULLED` state failed due to the image reference in the task payload being invalid.

As noted above, a new configuration setting `ECS_MANIFEST_PULL_TIMEOUT` is introduced in this PR that allows users to configure a timeout to be used when calling image registries to pull image manifests. The default value for this timeout is 1 minute and the minimum configurable value is 30 seconds. 

This change adds additional time to task starts as transition to MANIFEST_PULLED state can make an additional blocking call to image registry. In my test environment the additional time ranges from 300ms (ECR) to 900ms (Dockerhub). This is expected. In an upcoming PR we will change image pulls to use the resolved digest which will make image pulls a bit faster and the long-term impact on task start will but cut to 150ms (ECR) to 500ms (Dockerhub). For this reason, this PR is being merged to a short-lived feature branch so that we merge both changes to dev branch as a single commit.

### Implementation details

* Implement `*DockerTaskEngine.pullContainerManifest` which is the transition function for `MANIFEST_PULLED` state as described above. A new helper method `*DockerTaskEngine.setRegistryCredentials` is added which is used to resolve private registry credentials. Code for this method is taken from the existing `pullAndUpdateContainerReference` method that is also refactored to call `setRegistryCredentials` method.
* Update `*dockerImageManager.RecordContainerReference` method that is responsible for setting `ImageDigest` on containers. Now the method sets `ImageDigest` only if it's not pre-populated by the transition to MANIFEST_PULLED state. 
* Introduce a new `ManifestPullTimeout` setting in `Config` and parse its value from `ECS_MANIFEST_PULL_TIMEOUT` environment variable if it is available. Update relevant tests.
* Existing engine tests are updated to take into account the new interactions with docker during transition to `MANIFEST_PULLED` state. 
* A new `PutASMDockerAuthConfig` method is added to `*ASMAuthResource` so that the resource may be set up with fake docker auth config for testing purposes. The method is not used outside of testing. 
 
### Testing

Many new unit and integration tests are added. 

Unit tests - 
* `TestPullContainerManifest` - Tests `*DockerTaskEngine.pullContainerManifest` method for all image pull behavior and container type cases. Checks all interactions with docker. 
* `TestManifestPullTaskShouldContinue` - Tests task engine's handling of tasks with a focus on transition to `MANIFEST_PULLED` state. This test covers success and failure cases for the transition when the task should continue its lifecycle. If image pull behavior is `default` or `prefer-cached` then the task should continue its lifecycle even if there was an error during transition to `MANIFEST_PULLED` state which is the same behavior as transition to `PULLED` state.
* `TestManifestPullFailuresTaskShouldStop` - Similar to `TestManifestPullTaskShouldContinue` above but covers failure cases for the transition when the task should be stopped due to the transition failure. If image pull behavior is `always` or `once` then the task should be stopped if there was an error during transition to `MANIFEST_PULLED` state which is the same behavior as transition to `PULLED` state.
* `TestImagePullRequired` - Tests `*DockerTaskEngine.imagePullRequired` method which is missing unit tests. 
* `TestSetRegistryCredentials` - Tests the newly added `*DockerTaskEngine.setRegistryCredentials` method. Ensures that the method sets registry auth credentials on the container and returns a function that can be called to clear the credentials when they are no longer needed.

Integration tests - 
* `TestManifestPulledDoesNotDependOnContainerOrdering` - This existing test for `MANIFEST_PULLED` state is updated to check that `ImageDigest` field is set for containers after they have transitioned to `MANIFEST_PULLED` state. 
* `TestPullContainerManifestInteg` - This new test checks that `*DockerTaskEngine.pullContainerManifest` method works as expected for all image pull behaviors. That is, the method sets `ImageDigest` if success is expected and returns the right error if failure is expected.

I also performed the following manual tests.
* Ran tasks with images in public ECR, private ECR, public dockerhub, and private dockerhub registries and checked that Agent is able to fetch image manifest digest for the images during transition to `MANIFEST_PULLED` state. 
* Ran tasks on Agent configured with `once`, `default`, `always`, and `prefer-cached` image pull behaviors. Verified in the logs that Agent gets contacts the registry for `default` and `always` image pull behaviors and gets the digest from local image for `once` and `prefer-cached` image pull behaviors if the image is cached. 
* Ran ten tasks with a ECR and then with Dockerhub image with an Agent with changes in this PR and an Agent without the changes in this PR and computed the average task start latency for the tasks. Both Agents were configured to use `always` image pull behavior. Observed an increase in task start time ranging from 300ms (for ECR) to 900ms (for Dockerhub) owing to the new image registry call during transition to `MANIFEST_PULLED` state. This increase in task start time is expected given that a new blocking network call is made to fetch image manifest before the container can be started. 
* Ran a task on an environment with Docker 17.03. Transition to MANIFEST_PULLED skipped digest resolution with `Failed to find a supported API version that supports manifest pulls. Skipping digest resolution.` warning message printed in the logs as expected. 

Feature - Resolve image manifest digest for task containers during transition to `MANIFEST_PULLED` state before container image pulls. In future this early resolution of manifest digest will be used to expedite the reporting of manifest digests to ECS backend.

### Licensing


agent/engine/docker_task_engine.go
```diff
@@ -1238,12 +1240,139 @@ func (engine *DockerTaskEngine) GetDaemonManagers() map[string]dm.DaemonManager
 func (engine *DockerTaskEngine) pullContainerManifest(
 	task *apitask.Task, container *apicontainer.Container,
 ) dockerapi.DockerContainerMetadata {
-	// Currently a no-op
-	logger.Debug("Manifest pull is currently a no-op", logger.Fields{
-		field.TaskID:    task.GetID(),
-		field.Container: container.Name,
-		field.Image:     container.Image,
+	switch container.Type {
```
sparrc, line 1243: Why not use [Container.IsInternal()](https://pkg.go.dev/github.com/aws/amazon-ecs-agent/agent/api/container#Container.IsInternal) function here instead? Are there some "internal" containers that we do actually want to get the digest?

agent/engine/docker_task_engine.go
```diff
@@ -1238,12 +1240,139 @@ func (engine *DockerTaskEngine) GetDaemonManagers() map[string]dm.DaemonManager
...
+
+	var imageManifestDigest digest.Digest
+	digestFromPayload := referenceutil.GetDigestFromImageRef(container.Image)
+	if digestFromPayload != "" {
+		// Digest available in task payload
+		imageManifestDigest = digestFromPayload
+		logger.Info("Found image manifest digest in task payload", logger.Fields{
+			field.TaskARN:       task.Arn,
+			field.ContainerName: container.Name,
+			field.Image:         container.Image,
```
sparrc, line 1265: nit: I guess it's a part of the image field, but could we log the digest as a field here in case we ever make a mistake parsing the digest and need to verify if we got the correct one from the image?

agent/engine/docker_task_engine.go
```diff
@@ -1238,12 +1240,139 @@ func (engine *DockerTaskEngine) GetDaemonManagers() map[string]dm.DaemonManager
...
+				field.Image:         container.Image,
+			})
+			return dockerapi.DockerContainerMetadata{
+				Error: dockerapi.CannotPullImageManifestError{FromError: err},
+			}
+		}
+		if len(imgInspect.RepoDigests) == 0 {
+			// Image was not pulled from a registry, so the user must have cached it on the
+			// host directly. Skip digest resolution for this case as there is no digest.
+			logger.Warn("No repo digest found", logger.Fields{
```
sparrc, line 1283: If this is expected for some situations, maybe it should just be a "Info" message?

agent/engine/docker_task_engine.go
amogh09: I followed the skip logic from `pullContainer`. 
https://github.com/aws/amazon-ecs-agent/blob/b702281f0f6aa2fba5614050769aa337d1456853/agent/engine/docker_task_engine.go#L1251-L1258

I agree that `container.IsInternal()` is a better way to do it. I will make this change.

agent/engine/docker_task_engine.go
amogh09: Good point. I will make this change.

agent/engine/docker_task_engine.go
```diff
@@ -1238,12 +1240,139 @@ func (engine *DockerTaskEngine) GetDaemonManagers() map[string]dm.DaemonManager
...
+	if digestFromPayload != "" {
+		// Digest available in task payload
+		imageManifestDigest = digestFromPayload
+		logger.Info("Found image manifest digest in task payload", logger.Fields{
+			field.TaskARN:       task.Arn,
+			field.ContainerName: container.Name,
+			field.Image:         container.Image,
+		})
+	} else if !engine.imagePullRequired(engine.cfg.ImagePullBehavior, container, task.GetID()) {
+		// Digest should be resolved by inspecting local image
```
sparrc, line 1268: could you add a little more explanation to this comment on why this case exists? IIUC is it only for locally cached images and if agent is configured to prefer cached?

agent/engine/docker_task_engine.go
amogh09: Yeah.. I thought the usecase of a customer manually putting in an image on the host without it going through a pull is fairly niche so I chose warn. But since the usecase is legit I will change it to info.

agent/engine/docker_task_engine.go
amogh09: Yeah I can do that. The idea is that resolved digest should match the digest of the image which would be used for the container. If image pull is not required then Agent attempts to use a locally available image for the container during transition to CREATED state.

agent/engine/docker_task_engine.go
amogh09: Updated.

agent/engine/docker_task_engine.go
amogh09: Updated

agent/engine/docker_task_engine.go
amogh09: Updated.

agent/engine/docker_task_engine.go
amogh09: Updated.

agent/engine/docker_task_engine.go
```diff
@@ -1238,12 +1240,137 @@ func (engine *DockerTaskEngine) GetDaemonManagers() map[string]dm.DaemonManager
...
+			// Image was not pulled from a registry, so the user must have cached it on the
+			// host directly. Skip digest resolution for this case as there is no digest.
+			logger.Info("No repo digest found", logger.Fields{
+				field.TaskARN:       task.Arn,
+				field.ContainerName: container.Name,
+				field.Image:         container.Image,
+			})
+			return dockerapi.DockerContainerMetadata{}
+		}
+		parsedDigest := referenceutil.GetDigestFromImageRef(imgInspect.RepoDigests[0])
```
yinyic, line 1288: Do we always want to get the first item in the `RepoDigests` list? What does it mean when there are multiple items in it?

agent/engine/docker_task_engine.go
```diff
@@ -1486,6 +1582,53 @@ func (engine *DockerTaskEngine) pullAndUpdateContainerReference(task *apitask.Ta
 	return metadata
 }
 
+// Sets image registry auth credentials for the container.
+// Returns a cleanup function that may be called to clear the credentials from the container
```
yinyic, line 1586: Could you help me understand what's the benefit/need for unsetting the credentials?

yinyic: Also, the credentials resources are currently mapped as resource dependencies for container like [here](https://github.com/amogh09/amazon-ecs-agent/blob/master/agent/api/task/task.go#L861-L863). I'm wondering if it's possible to do the same? It's easier and cleaner to manage all resource dependencies in one place.

amogh09: This follows existing logic from image pulls. 
https://github.com/aws/amazon-ecs-agent/blob/b702281f0f6aa2fba5614050769aa337d1456853/agent/engine/docker_task_engine.go#L1419

I think the idea is to improve Agent's security posture by clearing auth credentials from the container as soon as they are not needed (I understand that the credentials might still be available in other components).

amogh09: The `setCredentials` method gets the credentials from the relevant resource dependency of the task if the container requires credentials from ASM. There is no change to how credentials are modeled and managed.

agent/engine/docker_task_engine.go
amogh09: Updated.

agent/engine/docker_task_engine.go
yinyic: > I think the idea is to improve Agent's security posture by clearing auth credentials from the container as soon as they are not needed (I understand that the credentials might still be available in other components).

Okay. I am still not fully convinced why it is necessary to intentionally add this finer control granularity, when task execution role permissions are supposedly granted to all containers in the task equally. But this is non-blocking for this PR.

yinyic: > The setCredentials method gets the credentials from the relevant resource dependency of the task if the container requires credentials from ASM. There is no change to how credentials are modeled and managed.

If I'm reading correctly, the `SetCredentials` assumes that the credentials have already been retrieved remotely (from ASM), and is simply copying the creds from one internal struct to another. The actual retrieval process happens as this resource [dependency](https://github.com/amogh09/amazon-ecs-agent/blob/master/agent/api/task/task.go#L861-L863) gets fulfilled. The ASM credentials resource is currently mapped as a dependency for `ContainerPulled`, should we change it to `ManifestPulled` (or whatever the new container state is named) to ensure that the credentials have been retrieved before attempting to copy them?

amogh09: > If I'm reading correctly, the SetCredentials assumes that the credentials have already been retrieved remotely (from ASM), and is simply copying the creds from one internal struct to another.

Yes that's right. 

> The ASM credentials resource is currently mapped as a dependency for ContainerPulled, should we change it to ManifestPulled (or whatever the new container state is named) to ensure that the credentials have been retrieved before attempting to copy them?

Yes that's right. That change has already been merged. 
https://github.com/aws/amazon-ecs-agent/blob/b702281f0f6aa2fba5614050769aa337d1456853/agent/api/task/task.go#L1052-L1059

yinyic (approved):

sparrc (approved):

PR #4155, opened 2024-04-24
Prioritize ssm creds for linux

sidlovskyy:
### Implementation details
In Linux environment when `ECS_EXTERNAL` variable is set to `true` `RotatingSharedCredentialsProvider` is prioritized over `EC2RoleProvider` so that SSM Agent credentials are prioritized over EC2 Role credentials.

### Testing

Prioritize `RotatingSharedCredentialsProvider` over `EC2RoleProvider` instance credentials provider for external instances in Linux

### Licensing

### DIFF
--- a/ecs-agent/credentials/instancecreds/instancecreds_linux.go
+++ b/ecs-agent/credentials/instancecreds/instancecreds_linux.go
@@ -28,20 +28,30 @@ import (
 // GetCredentials returns the instance credentials chain. This is the default chain
 // credentials plus the "rotating shared credentials provider", so credentials will
 // be checked in this order:
+//
 //  1. Env vars (AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY).
+//
 //  2. Shared credentials file (https://docs.aws.amazon.com/ses/latest/DeveloperGuide/create-shared-credentials-file.html) (file at ~/.aws/credentials containing access key id and secret access key).
+//
 //  3. EC2 role credentials. This is an IAM role that the user specifies when they launch their EC2 container instance (ie ecsInstanceRole (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html)).
+//
 //  4. Rotating shared credentials file located at /rotatingcreds/credentials
+//
+//     In the case of external instance `RotatingSharedCredentialsProvider` provider should be prioritized over
+//     `EC2RoleProvider` one as SSM credentials are going to be ignored in case agent is deployed to EC2
+//     with instance role which doesn't have permissions needed to run ECS Agent
 func GetCredentials(isExternal bool) *credentials.Credentials {
 	mu.Lock()
-	if credentialChain == nil {
-		credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
+	credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
+	if isExternal {
+		credProviders = append(credProviders[:2], append([]credentials.Provider{providers.NewRotatingSharedCredentialsProvider()}, credProviders[2:]...)...)
+	} else {
 		credProviders = append(credProviders, providers.NewRotatingSharedCredentialsProvider())
-		credentialChain = credentials.NewCredentials(&credentials.ChainProvider{
-			VerboseErrors: false,
-			Providers:     credProviders,
-		})
 	}
+	credentialChain = credentials.NewCredentials(&credentials.ChainProvider{
+		VerboseErrors: false,
+		Providers:     credProviders,
+	})
 	mu.Unlock()
 
 	// credentials.Credentials is concurrency-safe, so lock not needed here
--- a/ecs-agent/credentials/instancecreds/instancecreds_linux.go
+++ b/ecs-agent/credentials/instancecreds/instancecreds_linux.go
@@ -42,16 +42,18 @@ import (
 //     with instance role which doesn't have permissions needed to run ECS Agent
 func GetCredentials(isExternal bool) *credentials.Credentials {
 	mu.Lock()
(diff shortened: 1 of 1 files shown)


amogh09:
Thanks for the contribution @sidlovskyy . I see that this change was made for Windows in https://github.com/aws/amazon-ecs-agent/pull/3184. I am trying to find out why Linux was not considered in-scope at that time.

ecs-agent/credentials/instancecreds/instancecreds_linux.go
```diff
@@ -28,15 +28,27 @@ import (
...
 //  2. Shared credentials file (https://docs.aws.amazon.com/ses/latest/DeveloperGuide/create-shared-credentials-file.html) (file at ~/.aws/credentials containing access key id and secret access key).
+//
 //  3. EC2 role credentials. This is an IAM role that the user specifies when they launch their EC2 container instance (ie ecsInstanceRole (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/instance_IAM_role.html)).
+//
 //  4. Rotating shared credentials file located at /rotatingcreds/credentials
+//
+//     In the case of external instance `RotatingSharedCredentialsProvider` provider should be prioritized over
+//     `EC2RoleProvider` one as SSM credentials are going to be ignored in case agent is deployed to EC2
+//     with instance role which doesn't have permissions needed to run ECS Agent
 func GetCredentials(isExternal bool) *credentials.Credentials {
```
amogh09, line 43: Looks like this change makes `GetCredentials` implementation for Linux identical to that of Windows. I think we should merge the two into a single `instancecreds/instancecreds.go` file.

amogh09:
> Thanks for the contribution @sidlovskyy . I see that this change was made for Windows in #3184. I am trying to find out why Linux was not considered in-scope at that time.

I talked to some folks about this. We don't have a good reason as to why #3184 did not make the same change for Linux. It makes sense for the instance credentials fetch logic to be platform agnostic.

ecs-agent/credentials/instancecreds/instancecreds_linux.go
amogh09: In the meanwhile I will reproduce this issue on my end. Once the PR is ready I can verify that the change works by testing it in the same environment.

ecs-agent/credentials/instancecreds/instancecreds_linux.go
```diff
@@ -28,15 +28,27 @@ import (
...
 	mu.Lock()
 	if credentialChain == nil {
 		credProviders := defaults.CredProviders(defaults.Config(), defaults.Handlers())
-		credProviders = append(credProviders, providers.NewRotatingSharedCredentialsProvider())
+		if isExternal {
+			credProviders = append(credProviders[:2], append([]credentials.Provider{providers.NewRotatingSharedCredentialsProvider()}, credProviders[2:]...)...)
```
amogh09, line 48: Why are we using `credProviders[:2]`?

ecs-agent/credentials/instancecreds/instancecreds_linux.go
amogh09: I was able to reproduce the issue on my end.

ecs-agent/credentials/instancecreds/instancecreds_linux.go
amogh09, line 48: nit: instead of making an assumption on the order and length of providers returned by `defaults.CredProviders` (which is an external dependency of ecs-agent) it might be a good idea to define the list of providers explicitly. 

```go
     credProviders := []credentials.Provider{&credentials.EnvProvider{}}
     if isExternal {
         credProviders = append(credProviders,
             rotatingCredsProvider,
             &credentials.SharedCredentialsProvider{})
     } else {
         credProviders = append(credProviders,
             &credentials.SharedCredentialsProvider{},
             rotatingCredsProvider)
     }
     credentialChain = credentials.NewCredentials(&credentials.ChainProvider{
         VerboseErrors: false,
         Providers:     credProviders,
     })
```

ecs-agent/credentials/instancecreds/instancecreds_linux.go
amogh09: Since merging the Windows implementation with Linux will increase the scope of testing, I am okay with making this change for Linux alone in this PR and do the merge as future work (maybe leave a TODO here). 

I noticed that the Windows implementation gives precedence to rotating shared creds over shared credentials file even (since it slots in rotating creds provider at second position in the slice instead of third) which does not match the documentation of the function. 

https://github.com/aws/amazon-ecs-agent/blob/b702281f0f6aa2fba5614050769aa337d1456853/ecs-agent/credentials/instancecreds/instancecreds_windows.go#L28-L72

sidlovskyy:
> > Thanks for the contribution @sidlovskyy . I see that this change was made for Windows in #3184. I am trying to find out why Linux was not considered in-scope at that time.
> 
> I talked to some folks about this. We don't have a good reason as to why #3184 did not make the same change for Linux. It makes sense for the instance credentials fetch logic to be platform agnostic.

Thank you, should I update PR to include platform agnostic `instancecreds.go` that is the same as existing windows implementation?

jhoward-ts:
Hey @amogh09 and all guys - @sidlovskyy has sadly moved on from this project, but my team is still interested in this. Any updates?

mye956:
> Hey @amogh09 and all guys - @sidlovskyy has sadly moved on from this project, but my team is still interested in this. Any updates?

Hello @jhoward-ts! Could we make the following change that @amogh09 has suggested?

> Looks like this change makes GetCredentials implementation for Linux identical to that of Windows. I think we should merge the two into a single instancecreds/instancecreds.go file.

This PR will probably need to rebase as well as run `go mod tidy` and `go mod vendor` within the `agent/` directory. Once done, I'll add the label to run our functional tests again for this change.

PR #4157, opened 2024-04-25
Use resolved digest for image pulls

amogh09:
### Summary

This PR updates image pull logic to use a resolved image manifest digest if one is available. Image manifest digests are resolved during container transition to `MANIFEST_PULLED` state. The change ensures that the pulled image is the same as pointed by the resolved digest. 

### Implementation details

* Add a new method `TagImage` to `DockerClient` interface and its implementation. The method tags an image on the host. The implementation performs retries using a new `ConstantBackoff` strategy that backs off the same duration every time. A constant backoff retry strategy is fine in this case as there is no external service involved. 
* Add a new `ConstantBackoff` backoff strategy under `ecs-agent` module. The strategy returns the same amount of backoff duration regardless of how many times its `Duration` method is called. 
* Update `*DockerTaskEngine.pullAndUpdateContainerReference` method that is used for pulling container images so that it uses the container's `ImageDigest` field to prepare a canonical reference to the image to be pulled. The method tags the pulled image with `Container.Image` if a different image reference was used to pull the image so that image caching and image cleanup continue to work as before. 
* Add a new method `GetCanonicalRef` to `agent/utils/reference` package that returns a canonical image reference given an image reference and a manifest digest. 
* Test updates and new tests.

### Testing

* Added a new integration test named `TestPullContainerWithAndWithoutDigestInteg` to check that `*DockerTaskEngine.pullContainer` can pull images for containers with and without an `ImageDigest` set. 
* Added a new integration test named `TestPullContainerWithAndWithoutDigestConsistency` to check that `*DockerTaskEngine.pullContainer` pulls the same image with or without a digest set and the image can be inspected with `container.Image` field in both cases. 

In addition to the integration tests above, performed the following manual tests. 
* Ran a variety of tasks with Agent configured to use `always` image pull behavior. Checked that all tasks ran as expected. Images were pulled using digests and tagged with the image reference in the task definition. Images were cleaned up without any issues. 
* Ran a variety of tasks with Agent configured to use `once` and then `prefer-cached` image pull behavior. Checked that all tasks ran as expected. Cached images were used in both cases when found. Image cleanup worked as expected with `once` image pull behavior. Image pull is disabled when `prefer-cached` image pull behavior is used. 
* Ran a simple task multiple times with an Agent built with changes in this PR and again with an Agent built against master branch. Both Agents were configured to use `always` image pull behavior to force image pulls. Measured the task average start times (`startedAt` - `createdAt`) and task pull times (`pullStoppedAt` - `pullStartedAt`) for both cases. Changes to resolve image manifest digest in https://github.com/aws/amazon-ecs-agent/pull/4152 caused an additional delay in task start times that ranged from 300ms (ECR) to 900ms (Dockerhub), however, with this PR the image pulls are now slightly faster. Pull time for an image that's already available on the host is reduced from ~700ms (Dockerhub), ~250ms (public ECR), and ~130ms (private ECR) to ~260ms (Dockerhub), ~100ms (public ECR), and ~50ms (private ECR). Combined with changes to resolve image manifest digests (#4152) the overall average increase in task start time in my test environment is ~500ms (Dockerhub) and ~150ms (ECR). 

### Licensing

### DIFF
--- a/agent/engine/docker_task_engine.go
+++ b/agent/engine/docker_task_engine.go
@@ -82,6 +82,7 @@ const (
 	ipamCleanupTmeout                  = 5 * time.Second
 	minEngineConnectRetryDelay         = 2 * time.Second
 	maxEngineConnectRetryDelay         = 200 * time.Second
+	tagImageTimeout                    = 30 * time.Second
 	engineConnectRetryJitterMultiplier = 0.20
 	engineConnectRetryDelayMultiplier  = 1.5
 	// logDriverTypeFirelens is the log driver type for containers that want to use the firelens container to send logs.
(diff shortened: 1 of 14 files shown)


agent/engine/docker_task_engine.go
```diff
@@ -1534,25 +1535,91 @@ func (engine *DockerTaskEngine) pullAndUpdateContainerReference(task *apitask.Ta
...
 		return metadata
 	}
 	pullSucceeded := metadata.Error == nil
+
+	if pullSucceeded && imageRef != container.Image {
+		// Resolved image manifest digest was used to pull the image.
+		// Tag the pulled image so that it can be found using the image reference in the task.
+		ctx, cancel := context.WithTimeout(engine.ctx, tagImageTimeout)
+		defer cancel()
+		err := engine.client.TagImage(ctx, imageRef, container.Image)
```
mye956, line 1583: Q: Will there ever be a case where the image is already tagged here? I'm assume no but was curious if there's a scenario where we need to call `renameImage()`.

agent/engine/docker_task_engine.go
```diff
@@ -1534,25 +1535,91 @@ func (engine *DockerTaskEngine) pullAndUpdateContainerReference(task *apitask.Ta
...
+
+	metadata := engine.client.PullImage(engine.ctx, imageRef, container.RegistryAuthentication, engine.cfg.ImagePullTimeout)
 
 	// Don't add internal images(created by ecs-agent) into imagemanger state
 	if container.IsInternal() {
 		return metadata
 	}
 	pullSucceeded := metadata.Error == nil
+
+	if pullSucceeded && imageRef != container.Image {
```
mye956, line 1578: Q: Will the `imageRef != container.Image` check ever be false? Looking at [pullImage](https://github.com/amogh09/amazon-ecs-agent/blob/a5f8d25f66fbf4f8839d5e7e5fcbc635a4d75c5e/agent/dockerclient/dockerapi/docker_client.go#L374), it doesn't seem like we're updating either of the two? (please correct me if this is wrong!)

agent/utils/reference/reference.go
```diff
@@ -78,3 +78,34 @@ func GetDigestFromRepoDigests(repoDigests []string, imageRef string) (digest.Dig
 
 	return "", fmt.Errorf("found no repo digest matching '%s'", imageRef)
 }
+
+// Given an image reference and a manifest digest string, returns a canonical reference
+// for the image.
+// If the image reference has a digest then the canonical reference will still use the provided
+// manifest digest overwriting the exiting digest in the image reference.
```
mye956, line 85: nit (non-blocking): typo in `exiting`?

agent/engine/docker_task_engine.go
amogh09: It will be false if `container.Image` is already in canonical form like `busybox@sha256:05a79c7279f71f86a2a0d05eb72fcb56ea36139150f0a75cd87e80a4272e4e39` or if  image digest was not resolved for the container (if Docker engine version is too old for digest resolution, for example) and `container.GetImageDigest()` returns an empty string.

mye956: Sync'd offline! Essentially we're making sure that the image ref that we're using to pull is in canonical reference format will be retagged so that it can be referred to by just the image name or <image_name>:latest. Thanks Amogh!

agent/engine/docker_task_engine.go
mye956: Disregard! This is only used in our integ tests.

yinyic, line 1583: If we are retagging the image with mutable tags (such as `nginx:latest`), and we have a subsequent task whose container uses the same image but using digest (such as `nginx:shaXYZ`)

1. will the new task be able to discover that the local `nginx:latest` is identical to `nginx:shaXYZ` and reuse it?
2. will Agent be able to keep the correct image reference counter on `nginx:latest`?

mye956 (approved):

agent/utils/reference/reference.go
amogh09: Will fix this in a follow-up PR right after.

PR #4160, opened 2024-04-25
Integrate restart policy into the container object

sparrc:
### Summary

- Integrates restart policy into the Container object, with a helper function to check if the container restart policy is defined and enabled.
- Initializes the RestartTracker object when unmarshaling the task payload, if the container has a restart policy defined and enabled.
- `go mod vendor` in the agent/ dir to pull in ecs-agent/ shared lib changes.

Exact struct naming is still subject to change but this should roughly match the GA feature.

### Testing

New tests cover the changes: 

NA

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	resourcestatus "github.com/aws/amazon-ecs-agent/agent/taskresource/status"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/api/container/restart"
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
 	apierrors "github.com/aws/amazon-ecs-agent/ecs-agent/api/errors"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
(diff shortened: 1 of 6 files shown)


agent/api/container/container.go
```diff
@@ -334,13 +334,23 @@ type Container struct {
...
 
 type DependsOn struct {
 	ContainerName string `json:"containerName"`
 	Condition     string `json:"condition"`
 }
 
+type RestartPolicy struct {
+	Enabled              bool          `json:"enabled"`
+	IgnoredExitCodes     []int         `json:"ignoredExitCodes"`
+	RestartAttemptPeriod time.Duration `json:"restartAttemptPeriod"`
```
singholt, line 351: what does this mean?

agent/api/container/container.go
```diff
@@ -334,13 +334,23 @@ type Container struct {
...
+	RestartCount  int           `json:"restartCount,omitempty"`
+	LastRestartAt time.Time     `json:"lastRestartAt,omitempty"`
 }
 
 type DependsOn struct {
 	ContainerName string `json:"containerName"`
 	Condition     string `json:"condition"`
 }
 
+type RestartPolicy struct {
```
timj-hh, line 348: should be struct from ecs-agent module?

timj-hh (approved):

singholt (approved):

agent/api/container/container_test.go
```diff
@@ -1334,6 +1336,73 @@ func TestGetCredentialSpec(t *testing.T) {
...
+					Enabled: false,
+				},
+			},
+			restartCount:    0,
+			expectedEnabled: false,
+		},
+		{
+			name: "enabled restart policy",
+			container: &Container{
+				RestartPolicy: &restart.RestartPolicy{
```
mye956, line 1375: Just curious, do we/are there plans to validate RestartPolicy isn't empty if restart policy is enabled when we receive the task payload?

sparrc: That's a good question, we may want to validate that when we initialize the restart trackers, let me sync with Fargate dataplane on how we want to do that

mye956: 👍 thanks Cam! I also have no issues for this to be a follow up

mye956 (approved):

PR #4161, opened 2024-04-26
Additional integration tests for digest resolution changes

amogh09:
### Summary

This PR adds a couple of additional integration tests for changes to enable image digest resolution in ECS Agent. The tests are -
* `TestInvalidImageInteg` - tests that a task with a single container that has an invalid image fails with a `CannotPullImageManifestError` error.
* `TestImageWithDigestInteg` - tests that a task with a single container whose image has a valid digest specified works as expected.

The PR also includes a change to fix flakiness of `TestManifestPulledDoesNotDependOnContainerOrdering` integration test. The test is flaky because it assumes that `public.ecr.aws/docker/library/busybox:1.36.1` image is immutable but it's not. The PR also changes the test to use an image from local test registry instead as using a remote image does not affect the test.

Some minor refactoring changes and a typo fix is also included in this PR.

### Testing

Enhancement: Additional integration tests for digest resolution changes

### Licensing

### DIFF
--- a/agent/engine/engine_integ_test.go
+++ b/agent/engine/engine_integ_test.go
@@ -859,9 +859,6 @@ func TestInvalidImageInteg(t *testing.T) {
 	verifyContainerStoppedStateChangeWithReason(t, taskEngine,
 		"CannotPullImageManifestError: Error response from daemon: manifest unknown: manifest unknown")
 	verifyTaskStoppedStateChange(t, taskEngine)
-
-	err := taskEngine.(*DockerTaskEngine).removeContainer(task, container)
-	require.NoError(t, err, "failed to remove container during cleanup")
 }
 
 // Tests that a task with an image that has a digest specified works normally.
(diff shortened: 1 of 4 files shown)


agent/engine/engine_integ_test.go
```diff
@@ -831,3 +833,74 @@ func TestPullContainerWithAndWithoutDigestConsistency(t *testing.T) {
...
+	// Find image digest
+	versionedClient, err := taskEngine.(*DockerTaskEngine).client.WithVersion(dockerclient.Version_1_35)
+	manifest, manifestPullError := versionedClient.PullImageManifest(
+		context.Background(), localRegistryBusyboxImage, nil)
+	require.NoError(t, manifestPullError)
+	imageDigest := manifest.Descriptor.Digest.String()
+
+	// Prepare a task with image digest
+	container := createTestContainerWithImageAndName(
+		localRegistryBusyboxImage+"@"+imageDigest, "container")
```
prateekchaudhry, line 881: Is there a case to be tested for invalid digest?

amogh09: That should behave the same as `TestInvalidImageInteg` but I will add it nevertheless

hozkaya2000 (dismissed):

agent/engine/engine_integ_test.go
amogh09: Added the test case

prateekchaudhry (approved):

hozkaya2000 (approved):

PR #4162, opened 2024-04-29
Add restart tracker to ecs-agent module

timj-hh:
### Summary
This change adds a new restart module to /api/container/. The restart tracker contains a configurable policy and some helper methods. There is an accompanying test file unit testing the tracker.

### Implementation details
**Restart Tracker**
-  Implement methods `RecordRestart` and `ShouldRestart` to make a restart decision based on passed in container metadata
- Write RecordRestart to iterate restart count.
- Implement unit testing modifying each variable checked in `ShouldRestart`.
- Integrates `ContainerRestartPolicy` as implemented in the ecsacs container module into the restart tracker

### Testing
This component has not been plugged in yet, but all written tests pass.

Add restart module

no

### Licensing

### DIFF
--- a/ecs-agent/api/container/restart/restart_tracker.go
+++ b/ecs-agent/api/container/restart/restart_tracker.go
@@ -1,0 +1,99 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package restart
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
+)
+
+type RestartTracker struct {
+	RestartCount  int       `json:"restartCount,omitempty"`
+	LastRestartAt time.Time `json:"lastRestartAt,omitempty"`
+	restartPolicy RestartPolicy
+	lock          sync.RWMutex
+}
+
+// RestartPolicy represents a policy that contains key information considered when
+// deciding whether or not a container should be restarted after it has exited.
+type RestartPolicy struct {
+	Enabled              bool          `json:"enabled"`
+	IgnoredExitCodes     []int         `json:"ignoredExitCodes"`
+	RestartAttemptPeriod time.Duration `json:"restartAttemptPeriod"`
+}
+
+func NewRestartTracker(restartPolicy RestartPolicy) *RestartTracker {
+	return &RestartTracker{
+		restartPolicy: restartPolicy,
+	}
+}
+
+func (rt *RestartTracker) GetLastRestartAt() time.Time {
+	rt.lock.RLock()
+	defer rt.lock.RUnlock()
+	return rt.LastRestartAt
+}
+
+func (rt *RestartTracker) GetRestartCount() int {
+	rt.lock.RLock()
+	defer rt.lock.RUnlock()
+	return rt.RestartCount
+}
+
+// RecordRestart updates the restart tracker's metadata after a restart has occurred.
+// This metadata is used to calculate when restarts should occur and track how many
+// have occurred. It is not the job of this method to determine if a restart should
+// occur or restart the container.
+func (rt *RestartTracker) RecordRestart() {
+	rt.lock.Lock()
+	defer rt.lock.Unlock()
+	rt.RestartCount++
+	rt.LastRestartAt = time.Now()
+}
+
+// ShouldRestart returns whether the container should restart and a reason string
+// explaining why not. The reset attempt period will be calculated first
+// with LastRestart at, using the passed in startedAt if it does not exist.
+func (rt *RestartTracker) ShouldRestart(exitCode *int, startedAt time.Time,
+	desiredStatus apicontainerstatus.ContainerStatus) (bool, string) {
+	rt.lock.RLock()
+	defer rt.lock.RUnlock()
+
+	if !rt.restartPolicy.Enabled {
+		return false, "restart policy is not enabled"
+	}
+	if desiredStatus == apicontainerstatus.ContainerStopped {
+		return false, "container's desired status is stopped"
+	}
+	if exitCode == nil {
+		return false, "exit code is nil"
+	}
+	for _, ignoredCode := range rt.restartPolicy.IgnoredExitCodes {
+		if ignoredCode == *exitCode {
+			return false, fmt.Sprintf("exit code %d should be ignored", *exitCode)
+		}
+	}
+
+	startTime := startedAt
+	if !rt.LastRestartAt.IsZero() {
+		startTime = rt.LastRestartAt
+	}
+	if time.Since(startTime) < rt.restartPolicy.RestartAttemptPeriod {
+		return false, "attempt reset period has not elapsed"
+	}
+	return true, ""
+}
--- a/ecs-agent/api/container/restart/restart_tracker_test.go
+++ b/ecs-agent/api/container/restart/restart_tracker_test.go
@@ -1,0 +1,176 @@
+//go:build unit
+// +build unit
+
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License"). You may
+// not use this file except in compliance with the License. A copy of the
+// License is located at
+//
+//	http://aws.amazon.com/apache2.0/
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package restart
+
+import (
+	"testing"
+	"time"
+
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestShouldRestart(t *testing.T) {
+	ignoredCode := 0
+	rt := NewRestartTracker(RestartPolicy{
+		Enabled:              false,
+		IgnoredExitCodes:     []int{ignoredCode},
+		RestartAttemptPeriod: 60 * time.Second,
+	})
+	testCases := []struct {
+		name           string
+		rp             RestartPolicy
+		exitCode       int
+		startedAt      time.Time
+		desiredStatus  apicontainerstatus.ContainerStatus
+		expected       bool
+		expectedReason string
+	}{
+		{
+			name: "restart policy disabled",
+			rp: RestartPolicy{
+				Enabled:              false,
+				IgnoredExitCodes:     []int{ignoredCode},
+				RestartAttemptPeriod: 60 * time.Second,
+			},
+			exitCode:       1,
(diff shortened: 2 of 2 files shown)


ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,71 @@
...
+	restartPolicy *api.ContainerRestartPolicy
+}
+
+func NewRestartTracker(restartPolicy *api.ContainerRestartPolicy) *RestartTracker {
+	return &RestartTracker{
+		restartPolicy: restartPolicy,
+	}
+}
+
+func (rt *RestartTracker) GetRestartCount() int {
```
sparrc, line 37: should we put a mutex with a read lock here and and a write lock in RecordRestart()?

timj-hh: can do, have added this

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+	lock          sync.RWMutex
+}
+
+func NewRestartTracker(restartPolicy *api.ContainerRestartPolicy) *RestartTracker {
+	return &RestartTracker{
+		restartPolicy: restartPolicy,
+	}
+}
+
+func (rt *RestartTracker) GetRestartEnabled() bool {
```
danehlim, line 40: Why is this method necessary? Isn't this information only needed by the restart tracker since it is the one making the decision if a restart should be performed or not?

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,91 @@
...
+	api "github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
+
+	"github.com/aws/aws-sdk-go/aws"
+)
+
+type RestartTracker struct {
+	RestartCount  int       `json:"restartCount,omitempty"`
+	LastStartedAt time.Time `json:"lastStartedAt,omitempty"`
+	restartPolicy *api.ContainerRestartPolicy
```
danehlim, line 30: Let's create a non-ACS model struct type for restart policy which doesn't use pointer types for all of its fields as this is what was decided during internal design. That way we also avoid messy lines such as L87. 

For example,
```
if time.Since(startedAt) < rt.restartPolicy.AttemptResetPeriod
```
looks a lot cleaner than

```
if time.Since(startedAt) < time.Duration(*rt.restartPolicy.RestartAttemptPeriod*int64(time.Second))
```

This is common practice in Agent code, see environment file struct type for example:
https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/acs/model/ecsacs/api.go#L1074-L1080
https://github.com/aws/amazon-ecs-agent/blob/dev/agent/api/container/container.go#L355-L358

danehlim (changes requested):

ecs-agent/api/container/restart/restart_tracker.go
timj-hh: The method is not strictly necessary but may be helpful for implementers to check. e.g. if restart policy is configured but not enabled we may choose not to restart.

danehlim: >  e.g. if restart policy is configured but not enabled we may choose not to restart.

But the restart tracker should be the one making that choice right? I don't think we should expose this publicly unless there is a need to (i.e., keep things private unless there is a reason to make them public).

ecs-agent/api/container/restart/restart_tracker.go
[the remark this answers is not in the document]
timj-hh: Done, new rev out.

ecs-agent/api/container/restart/restart_tracker.go
timj-hh: Sure, if need be we can follow this up and re-add. I have removed now.

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,90 @@
...
+type RestartTracker struct {
+	RestartCount  int       `json:"restartCount,omitempty"`
+	LastStartedAt time.Time `json:"lastStartedAt,omitempty"`
+	restartPolicy RestartPolicy
+	lock          sync.RWMutex
+}
+
+// RestartPolicy represents a policy that contains key information considered when
+// deciding whether or not a container should be restarted after it has exited.
+type RestartPolicy struct {
```
danehlim, line 33: nit: Can we move this into a separate file ecs-agent/api/container/restart/policy.go for clearer separation, and consider renaming the files from this PR to ecs-agent/api/container/restart/tracker.go and ecs-agent/api/container/restart/tracker_test.go for more concise naming?

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,90 @@
...
+
+// RecordRestart updates the restart tracker's metadata after a restart has occurred.
+// This metadata is used to calculate when restarts should occur and track how many
+// have occurred. It is not the job of this method to determine if a restart should
+// occur or restart the container. It is expected to receive a startedAt time from the container runtime.
+func (rt *RestartTracker) RecordRestart(startedAt time.Time) {
+	rt.lock.RLock()
+	defer rt.lock.RUnlock()
+	rt.RestartCount++
+	rt.LastStartedAt = startedAt
```
danehlim, line 65: Apologies for the churn and not calling this out earlier:

Based off what was decided during internal design, this field should be named LastRestartAt (not LastStartedAt) and not take in the started at time for the container from the container runtime, right? Otherwise this seems to assume that the container runtime of the calling Agent always updates the started at time for the container when a container is restarted. Per our previous internal discussions, such assumptions about the container runtime of the calling Agent should not be made as the code should be agnostic of architectural/implementation details of any calling Agent.

timj-hh: We can rename the field, but the reason this is an argument to `RecordRestart` is to ensure we call this method with the correct time value. `LastStartedAt` will likely be passed into `ShouldRestart` after the first restart of a container, so we want to make sure that is set when the container would normally get to `RUNNING`, not necessarily the same time we would record a restart.

danehlim: This assumes that started at time of a container will be updated by the underlying container runtime of the calling Agent. Based off previous internal discussions, is that a safe assumption to make?

danehlim: This assumes that started at time of a container will be updated by the underlying container runtime of the calling Agent. Based off previous internal discussions and the fact this code should be agnostic of the calling Agent (and its underlying container runtime), is that a safe assumption to make?

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,90 @@
...
+
+func NewRestartTracker(restartPolicy RestartPolicy) *RestartTracker {
+	return &RestartTracker{
+		restartPolicy: restartPolicy,
+	}
+}
+
+func (rt *RestartTracker) GetLastStartedAt() time.Time {
+	rt.lock.RLock()
+	defer rt.lock.RUnlock()
```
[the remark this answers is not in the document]
timj-hh: It is possible, for agents that implement multiple asynchronous containers that independently restart we may get simultaneous access. I believe it is also common coding practice in `ecs-agent`

ecs-agent/api/container/restart/restart_tracker.go
timj-hh: discussed offline, removed time argument to `RecordRestart` in favor of `time.Now` and adjusted unit tests to round time checks to check within a millisecond of accuracy

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,90 @@
...
+	return rt.RestartCount
+}
+
+// RecordRestart updates the restart tracker's metadata after a restart has occurred.
+// This metadata is used to calculate when restarts should occur and track how many
+// have occurred. It is not the job of this method to determine if a restart should
+// occur or restart the container. It is expected to receive a startedAt time from the container runtime.
+func (rt *RestartTracker) RecordRestart() {
+	rt.lock.RLock()
```
sparrc, line 62: this is a mutating function so this should be `Lock()` not `RLock()`

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if exitCode == nil {
+		return false, "exit code is nil"
+	}
+	for _, ignoredCode := range rt.restartPolicy.IgnoredExitCodes {
+		if ignoredCode == *exitCode {
+			return false, fmt.Sprintf("exit code %d should be ignored", *exitCode)
+		}
+	}
+	if time.Since(startedAt) < rt.restartPolicy.RestartAttemptPeriod {
```
sparrc, line 86: Isn't the restart tracker tracking LastStartedAt? Why not use that timestamp instead of passing in?

ecs-agent/api/container/restart/restart_tracker.go
```diff
@@ -0,0 +1,90 @@
...
+func (rt *RestartTracker) RecordRestart() {
+	rt.lock.RLock()
+	defer rt.lock.RUnlock()
+	rt.RestartCount++
+	rt.LastStartedAt = time.Now()
+}
+
+// ShouldRestart returns whether the container should restart and a reason string
+// explaining why not.
+func (rt *RestartTracker) ShouldRestart(exitCode *int, startedAt time.Time,
```
sparrc, line 70: we should get a read lock here for accessing all of these fields

ecs-agent/api/container/restart/restart_tracker.go
timj-hh: ack, addressed

ecs-agent/api/container/restart/restart_tracker.go
timj-hh: Ack, addressed

ecs-agent/api/container/restart/restart_tracker.go
timj-hh: Per discussion above with Dane, I have updated this to use `LastRestartAt` over `startedAt` if it is passed in.

ecs-agent/api/container/restart/restart_tracker_test.go
```diff
@@ -0,0 +1,176 @@
...
+	assert.Equal(t, "attempt reset period has not elapsed", reason)
+}
+
+func TestRecordRestart(t *testing.T) {
+	rt := NewRestartTracker(RestartPolicy{
+		Enabled:              false,
+		RestartAttemptPeriod: 60 * time.Second,
+	})
+	assert.Equal(t, 0, rt.RestartCount)
+	for i := 1; i < 1000; i++ {
```
danehlim, line 160: nit: I don't think doing this 1000 times is necessary, this means that this loop would repeatedly execute across all test runs. Can we remove this?

ecs-agent/api/container/restart/restart_tracker_test.go
```diff
@@ -0,0 +1,176 @@
...
+		RestartAttemptPeriod: 60 * time.Second,
+	})
+	exitCode := 1
+
+	shouldRestart, reason := rt.ShouldRestart(&exitCode, time.Now().Add(-61*time.Second), apicontainerstatus.ContainerRunning)
+	assert.True(t, shouldRestart)
+
+	// After restarting, we should inform restart decisions with LastRestartedAt instead of the passed in startedAt time.
+	rt.RecordRestart()
+	shouldRestart, reason = rt.ShouldRestart(&exitCode, time.Now().Add(-61*time.Second), apicontainerstatus.ContainerRunning)
```
danehlim, line 149: nit: Best practice would be to save `time.Now().Add(-61*time.Second)` in a startedAt variable, similar to what you have done for `TestShouldRestart`.

danehlim (approved):

sparrc (approved):

tshan2001 (approved):

PR #4166, opened 2024-05-06
Add restart count to TMDE v4 response

sparrc:
### Summary

Update [v4 TMDE](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint-v4.html) with the restart count of the container, if the container has a restart policy enabled.

### Testing

unit tests updated, manual check on restart count being incremented as expected

NA

### Licensing

### DIFF
--- a/agent/handlers/v4/response.go
+++ b/agent/handlers/v4/response.go
@@ -19,6 +19,8 @@ import (
 	"github.com/aws/amazon-ecs-agent/agent/engine/dockerstate"
 	v2 "github.com/aws/amazon-ecs-agent/agent/handlers/v2"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/api/ecs"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
 	ni "github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/networkinterface"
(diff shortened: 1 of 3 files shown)


hozkaya2000 (dismissed):

agent/handlers/v4/response.go
```diff
@@ -121,6 +124,26 @@ func toV4NetworkResponse(
...
+func augmentContainerResponse(
+	containerID string,
+	state dockerstate.TaskEngineState,
+	v4Response *tmdsv4.ContainerResponse,
+) *tmdsv4.ContainerResponse {
+	dockerContainer, ok := state.ContainerByID(containerID)
+	if !ok {
+		// did not find container, continue on and try next container(s)
+		// we don't return error here to avoid disrupting all of a TMDS response
+		// on a single missing container.
```
chienhanlin, line 137: Should we add a log at warning or debug level if the container cannot be found from the state?
Like https://github.com/aws/amazon-ecs-agent/blob/master/agent/handlers/v4/stats_response.go#L41-L46.

sparrc: yes I can add that

hozkaya2000 (approved):

agent/handlers/v4/response.go
chienhanlin: Thanks!

chienhanlin (approved):

PR #4167, opened 2024-05-07
Add BackendStatusString method to ContainerStatus

amogh09:
### Summary

Currently `ContainerStatus` enum and backend recognized container status strings are coupled as `BackendStatus` method that is used to map former to latter returns a `ContainerStatus` and not a `string`. This is not ideal as the coupling restricts the evolution of internal container statuses in Agent. This PR decouples the two by introducing a new `BackendStatusString` method for `ContainerStatus` type that maps the internal container status to a suitable backend recognized status string. Existing `BackendStatus` method is marked as deprecated in this PR.

`BackendStatusString` method is slightly different from `BackendStatus` in that its steady state parameter is optional and defaults to `ContainerRunning`. That will be useful in cases where there is no special steady state for a container.

This decoupling already exists for `TaskStatus` type, this PR extends the same logic to `ContainerStatus`.
https://github.com/aws/amazon-ecs-agent/blob/b702281f0f6aa2fba5614050769aa337d1456853/ecs-agent/api/task/status/taskstatus.go#L69-L78

The new method is not currently used but we plan to use it in the near future. 

### Testing

Enhancement: Decouple internal container status and backend recognized container status by introducing BackendStatusString() method to map internal container status to a backend recognized container status.

### Licensing

### DIFF
--- a/agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status/containerstatus.go
+++ b/agent/vendor/github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status/containerstatus.go
@@ -120,16 +120,26 @@ func (cs *ContainerStatus) BackendStatus(steadyStateStatus ContainerStatus) Cont
 
 // BackendStatusString maps the internal container status in Agent to a backend recognized
 // status string.
-func (cs ContainerStatus) BackendStatusString(steadyStateStatus ContainerStatus) string {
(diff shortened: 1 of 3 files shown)


timj-hh (approved):

Yiyuanzzz (approved):

JoseVillalta:
late comment but this LGTM

PR #4173, opened 2024-05-11
AppArmor profile includes files if they exist

tinnywang:
### Summary
ECS agent fails to start on SUSE Linux 15 because AppArmor can't load the ECS agent profile.
```
level=info time=2024-05-10T23:31:45Z msg="pre-start"
level=info time=2024-05-10T23:31:45Z msg="Successfully created docker client with API version 1.25"
level=info time=2024-05-10T23:31:45Z msg="pre-start: setting up ecs-agent-default AppArmor profile"
level=error time=2024-05-10T23:31:45Z msg="error loading apparmor profile /etc/apparmor.d/ecs-agent-default: running `/sbin/apparmor_parser apparmor_parser -Kr /etc/apparmor.d/ecs-agent-default` failed with output: AppArmor parser error for /etc/apparmor.d/ecs-agent-default in profile /etc/apparmor.d/ecs-agent-default at line 2: Could not open 'tunables/global'\n\nerror: exit status 1"
```
This is happening because SUSE Linux 15 doesn't ship with the AppArmor files the ECS agent profile depends on ([`/etc/apparmor.d/tunables/global`](https://github.com/aws/amazon-ecs-agent/blob/1133dd999145712ad4b8c3076ac398c1be9a898e/ecs-init/apparmor/apparmor.go#L32) and [`/etc/apparmor.d/abstractions/base`](https://github.com/aws/amazon-ecs-agent/blob/1133dd999145712ad4b8c3076ac398c1be9a898e/ecs-init/apparmor/apparmor.go#L35)).

### Implementation details
To fix this, only write `#include` statements to the profile if the corresponding AppArmor files exist.
Always write `@{PROC}=/proc/` to the profile If `/etc/apparmor.d/tunables/global` does not exist, because the profile uses `@{PROC}`.

Prior art: https://github.com/moby/moby/blob/ac71ac1c92326635fffb5423f70891e53dc769e4/profiles/apparmor/apparmor.go#L38-L42

### Testing
Built the .rpm per these [instructions](https://github.com/aws/amazon-ecs-agent/tree/master/packaging/generic-rpm-integrated).

On a SUSE Linux 15 EC2 instance,
1. installed the .rpm with `zypper install -y --allow-unsigned-rpm`
2. started ECS agent with `systemctl start ecs`
3. verified ECS agent was running with `systemctl status ecs`
4. verified that the EC2 instance joined my ECS cluster
5. verified the contents of the AppArmor profile at `/etc/apparmor.d/ecs-agent-default`
```
@{PROC}=/proc/

profile ecs-agent-default flags=(attach_disconnected,mediate_deleted) {

  network inet,
  network inet6,
  network netlink,
  network unix,
  ...
```

On an AL2 EC2 instance,
1. installed the .rpm with `yum install -y`
2. started ECS agent with `systemctl start ecs`
3. verified ECS agent was running with `systemctl status ecs`
4. verified that the EC2 instance joined my ECS cluster
5. verified that the AppArmor profile at `/etc/apparmor.d/ecs-agent-default` did not exist (because AL2 does not support AppArmor)

Built the .deb per these [instructions](https://github.com/aws/amazon-ecs-agent/tree/master/packaging/generic-deb-integrated/debian).

On an Ubuntu 20.04 LTS EC2 instance,
1. installed the .deb with `apt install -y`
2. started ECS agent with `systemctl start ecs`
3. verified ECS agent was running with `systemctl status ecs`
4. verified that the EC2 instance joined my ECS cluster
5. verified the contents of the AppArmor profile at `/etc/apparmor.d/ecs-agent-default`
```
#include <tunables/global>

profile ecs-agent-default flags=(attach_disconnected,mediate_deleted) {

  #include <abstractions/base>

  network inet,
  network inet6,
  network netlink,
  network unix,
  ...
```

Bug - Fixed a bug that could prevent ECS agent from starting on SUSE Linux 15

No

### Licensing


ecs-init/apparmor/apparmor.go
```diff
@@ -28,11 +30,22 @@ const (
...
+	IncludeTunablesGlobal   bool
+	IncludeAbstractionsBase bool
+}
+
 const ecsAgentDefaultProfile = `
+{{if .IncludeTunablesGlobal}}
 #include <tunables/global>
+{{else}}
+@{PROC}=/proc/
+{{end}}
```
amogh09, line 44: Do we know why we include `tunables/global` and what does `@PROC=/proc/` do?

ecs-init/apparmor/apparmor_test.go
```diff
@@ -67,37 +70,90 @@ func TestLoadDefaultProfile(t *testing.T) {
...
 	}
 	defer func() {
 		isProfileLoaded = aaprofile.IsLoaded
 		loadPath = aaparser.LoadProfile
 		createFile = os.Create
+		statFile = os.Stat
 	}()
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			if !apparmor.HostSupports() {
```
mye956, line 78: Just curious, could you provide a bit more context as to why we're removing this check? (sorry not quite clear to me)

ecs-init/apparmor/apparmor.go
tinnywang: `tunables/global` contains global variable definitions that can be used in AppArmor profiles. One of those variables is `@{PROC}`, which is where `procfs` is mounted. We need `@{PROC}` defined because we use it in the agent profile.

Contents of `tunables/global`:
```
# All the tunables definitions that should be available to every profile
# should be included here

include <tunables/home>
include <tunables/multiarch>
include <tunables/proc>
include <tunables/alias>
include <tunables/kernelvars>
include <tunables/xdg-user-dirs>
include <tunables/share>
include <tunables/etc>
include <tunables/run>

# Also, include files in tunables/global.d for site-specific adjustments
include if exists <tunables/global.d>
```

Contents of `tunables/proc`:
```
# @{PROC} is the location where procfs is mounted.
@{PROC}=/proc/

# Also, include files in tunables/proc.d for site-specific adjustments
include if exists <tunables/proc.d>
```

ecs-init/apparmor/apparmor_test.go
tinnywang: `LoadDefaultProfile` reads/writes files in `/etc/apparmor.d`, which only exists from the get-go if the OS ships with AppArmor. The check would normally protect against accessing a directory that doesn't exist, but since we stub out the filesystem functions in the tests, the tests will never fail if the OS doesn't have AppArmor.

mye956 (commented):
Don't forget to change the destination branch to `dev`!

ecs-init/apparmor/apparmor.go
```diff
@@ -87,19 +101,45 @@ func LoadDefaultProfile(profileName string) error {
 		return err
 	}
 
+	includeTunablesGlobal, err := fileExists(filepath.Join(appArmorProfileDir, "tunables/global"))
+	if err != nil {
+		return err
```
mye956, line 106: Q: If we return an error here if `tunables/global` doesn't exists, wouldn't this mean we'll never write @{PROC}=/proc/ to the profile? (please correct me if this is incorrect/the expected behavior!)

tinnywang: If the file doesn't exist, `fileExists` will return `false` and no error, so we will still write the profile.
https://github.com/aws/amazon-ecs-agent/blob/ed042578fe6733eebe7cee4a60ae906414bd9427/ecs-init/apparmor/apparmor.go#L139-L141

These 3 test cases check for this (`statErrors` are the errors `os.Stat` is stubbed to return, and `profileContent` is what's expected to be written to the profile):
https://github.com/aws/amazon-ecs-agent/blob/ed042578fe6733eebe7cee4a60ae906414bd9427/ecs-init/apparmor/apparmor_test.go#L73-L97

mye956: ah I see, thanks Tiffany!

mye956 (approved):

amogh09 (approved):

ecs-init/apparmor/apparmor.go
amogh09: Thanks for explaining!

PR #4186, opened 2024-05-22
Copy correct resolv.conf  for Bottlerocket in debug mode.

hoshank10:
### Summary
 In EC2 debug mode, while setting up the dns config we copy the `/etc/resolv.conf` to `/etc/netns/(NetNSName)/`. This works fine on AL2 hosts, although on Bottlerocket hosts the `/etc/resolv.conf` is symlinked to a dummy stub `resolv.conf`. Instead the correct file to use for Bottlerocket is at `/run/netdog/resolv.conf`. This change copies the correct file to task network namespace dns config for debug mode on bottlerocket. 

### Implementation details
 Check if the file `/run/netdog/resolv.conf` exists on the host and copy it to task network namsepace.

### Testing
 

New tests cover the changes: NA

**Copy correct resolv.conf for Bottlerocket to launch tasks in debug mode.**

**Does this PR include breaking model changes? If so, Have you added transformation functions?** No
  

### Licensing


ecs-agent/netlib/platform/common_linux.go
```diff
@@ -520,7 +526,12 @@ func (c *common) generateNetworkConfigFilesForDebugPlatforms(
 		return errors.Wrap(err, "unable to create hostname file for netns")
 	}
 
-	err = c.copyFile(filepath.Join(netNSDir, ResolveConfFileName), "/etc/resolv.conf", taskDNSConfigFileMode)
+	// Check if the /run/netdog/resolv.conf file exists on the host. For Bottlerocket, instead of /etc/resolv.conf,
+	// this is the file we copy to the task network namespace dns config.
+	if _, err = c.os.Stat(srcBRResolv); err == nil {
```
samjkon, line 531: Is there a better way to detect if the host is using a bottle rocket AMI?

ecs-agent/netlib/platform/common_linux.go
```diff
@@ -63,8 +63,14 @@ const (
 	// It is assigned 100 because it is an unrealistically high
 	// value for interface index.
 	indexHighValue = 100
+
+	// srcBRResolv is the path of resolv.conf file on a Bottlerocket host
+	// that we copy to the task namespace when using debug mode.
+	srcBRResolv = "/run/netdog/resolv.conf"
 )
 
+var srcResolv = "/etc/resolv.conf"
```
samjkon, line 72: Instead of making this a var, I would suggest keeping this a const. You can use a local var in the function instead of creating a global var if needed.

ecs-agent/netlib/platform/common_linux.go
JoseVillalta: I think we should create a bottle-rocket specific platform in Network builder. Maybe not in this PR, but as a todo long-term. I can take that AI.

ecs-agent/netlib/platform/common_linux.go
hoshank10: Ack, will make that change

ecs-agent/netlib/platform/common_linux.go
hoshank10: Agreed. Thanks Jose

amogh09: Is `/run/netdog/resolv.conf` guaranteed to exist on Bottlerocket only and not on any other supported distros? I second @samjkon's comment.

hoshank10: I checked with a Bottlerocket Engineer and he confirmed the `/run/netdog/resolv.conf` will only exist in Bottlerocket.

strategicpause: The risk with this approach is that you're coupling the ECS Agent with special behavior which is OS/distro-specific. Should Bottlerocket change this behavior, or if other OSes introduce different behavior, then this pattern suggests that we should update the ECS Agent to accommodate that behavior. Instead is it possible to introduce a configuration value to override the default `resolv.conf` location? By default the option can point to `/etc/resolv.conf`; should you need to override it, it can instead be configured with whatever distro the ECS Agent is included on.

PR #4188, opened 2024-05-22
Do container restart on exit

sparrc:
### Summary

In the function that handles container change events, we will evaluate if an exiting container has a restart policy defined.

If the container does have a restart policy, we will evaluate if we _should_ restart the container using the [`RestartTracker.ShouldRestart`](https://pkg.go.dev/github.com/aws/amazon-ecs-agent/ecs-agent@v0.0.0-20240509232427-1133dd999145/api/container/restart#RestartTracker) function from the common library.

If we do do a container restart, we record it using the `RestartTracker.RecordRestart` function.

If the restart fails, we proceed with the regular container stop workflow.

### Testing

NA

### Licensing

### DIFF
--- a/agent/engine/task_manager.go
+++ b/agent/engine/task_manager.go
@@ -445,10 +445,12 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	containerKnownStatus := container.GetKnownStatus()
 
 	eventLogFields := logger.Fields{
-		field.TaskID:    mtask.GetID(),
-		field.Container: container.Name,
-		field.RuntimeID: runtimeID,
-		field.Status:    event.Status.String(),
+		field.TaskID:        mtask.GetID(),
+		field.Container:     container.Name,
+		field.RuntimeID:     runtimeID,
+		"changeEventStatus": event.Status.String(),
+		field.KnownStatus:   containerKnownStatus.String(),
+		field.DesiredStatus: container.GetDesiredStatus().String(),
 	}
 
 	if event.Status.String() == apicontainerstatus.ContainerStatusNone.String() ||
@@ -467,9 +469,7 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	// to be known running so it will be stopped. Subsequently ignore these backward transitions
 	mtask.handleStoppedToRunningContainerTransition(event.Status, container)
 	if event.Status <= containerKnownStatus {
-		logger.Debug("Container change is redundant", eventLogFields, logger.Fields{
-			field.KnownStatus: containerKnownStatus.String(),
-		})
+		logger.Debug("Container change is redundant", eventLogFields)
 
 		// Only update container metadata when status stays RUNNING
 		if event.Status == containerKnownStatus && event.Status == apicontainerstatus.ContainerRunning {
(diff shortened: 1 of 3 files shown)


agent/engine/task_manager.go
```diff
@@ -445,10 +445,12 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
 
 	eventLogFields := logger.Fields{
-		field.TaskID:    mtask.GetID(),
-		field.Container: container.Name,
-		field.RuntimeID: runtimeID,
-		field.Status:    event.Status.String(),
+		field.TaskID:        mtask.GetID(),
+		field.Container:     container.Name,
+		field.RuntimeID:     runtimeID,
+		"changeEventStatus": event.Status.String(),
```
tinnywang, line 451: Is there a rule-of-thumb for when logged fields get a constant in the `field` package?

agent/engine/task_manager.go
```diff
@@ -481,6 +481,45 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
 	// Container has progressed its status if we reach here. Make sure to save it to database.
 	defer mtask.engine.saveContainerData(container)
 
+	// If container is transitioning to STOPPED, first check if we should short-circuit
+	// the stop workflow and restart the container.
+	if event.Status == apicontainerstatus.ContainerStopped && container.RestartPolicyEnabled() {
+		exitCode := event.DockerContainerMetadata.ExitCode
+		shouldRestart, reason := container.RestartTracker.ShouldRestart(exitCode, container.GetStartedAt(), container.GetDesiredStatus())
+		if shouldRestart {
+			defer container.RestartTracker.RecordRestart()
```
tinnywang, line 490: Could you explain what happens when a container restart fails? Since `LastRestartAt` is updated even though the container didn't restart, is the next restart attempt delayed?

agent/engine/task_manager.go
sparrc: No particular rules here, I guess just whenever there is a field being used across multiple files/packages we should probably add it to the constants. In this case I don't think this field or the ones below will ever be used outside this particular function so I didn't bother adding constants.

agent/engine/task_manager.go
sparrc: If the "StartContainer" call itself fails, then the container will never be attempted to be restarted again. The restart policy is only triggered on containers that have moved past the "StartContainer" phase into RUNNING. The container also needs to be in the RUNNING state for a minimum time for it to be eligible to be restarted (this is the [RestartAttemptPeriod](https://pkg.go.dev/github.com/aws/amazon-ecs-agent/ecs-agent@v0.0.0-20240509232427-1133dd999145/api/container/restart#RestartPolicy) field)

agent/engine/task_manager.go
```diff
@@ -481,6 +481,45 @@ func (mtask *managedTask) handleContainerChange(containerChange dockerContainerC
...
+			logger.Info("Restarting container", eventLogFields,
```
yinyic, line 491: Do we mean to not include this current restart attempt in the `restartCount` or `lastRestartAt`? It feels a bit odd to have a log that reads "Restarting container, restartCount=0"

sparrc: Yes that was intended, if you log before restarting then you havent restarted yet, and then there is another log line after the restart happens below.

I should maybe change it to increment before logging the "Restarted" log line below though 🤔

yinyic: Yeah the actual container hasn't been restarted yet, but we've begun another "restart attempt" is probably how I feel 
Maybe we can rename to `previousRestartCount`?

tinnywang (approved):

agent/engine/task_manager.go
sparrc: since we will log after doing the restart no matter what, I will actually just delete this log line as it's a bit redundant anyways

tinnywang (approved):

PR #4189, opened 2024-05-23
Enable agent to detect FIPS enabled hosts

harishxr:
### Summary

This pull request enables ECS Agent on linux to detect if FIPS mode is enabled on the host machine and store the detected value.
### Implementation details

* Added a new function `DetectFIPSMode` in the utils package (utils/fips_linux.go). This function reads the `/proc/sys/crypto/fips_enabled` file to check if FIPS mode is enabled.
* Introduced a new variable `isFIPSEnabled` in the config package (config.go), which is initialized during the package initialization using the `IsFIPSEnabled` function.
* Added logic to log the FIPS mode status during initialization for debugging purposes.
* Created unit tests to ensure correct identification of the FIPS mode status.
* Added a stub function `DetectFIPSMode` in the utils package (utils/fips_windows.go). This function will set `isFIPSEnabled` to `false` by default on Windows hosts.

### Testing

* Unit tests were added to the utils package (utils/fips_linux_test.go)
* The unit tests cover the following scenarios:
   * FIPS mode is enabled (/proc/sys/crypto/fips_enabled contains 1).
   * FIPS mode is disabled (/proc/sys/crypto/fips_enabled contains 0).
   * FIPS mode file is non-existent.

Enable agent to detect FIPS enabled hosts

No

### Licensing

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -229,7 +229,7 @@ func NewConfig(ec2client ec2.EC2MetadataClient) (*Config, error) {
 		errs = append(errs, err)
 	}
 	config := &envConfig
-	isFIPSEnabled = utils.DetectFIPSMode()
+	isFIPSEnabled = utils.DetectFIPSMode(utils.FIPSModeFilePath)
 
 	if config.External.Enabled() {
 		if config.AWSRegion == "" {
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -188,6 +188,9 @@ var (
 
 	// CgroupV2 Specifies whether or not to run in Cgroups V2 mode.
 	CgroupV2 = false
+
+	// isFIPSEnabled indicates whether FIPS mode is enabled on the host
+	isFIPSEnabled = false
 )
 
 // Merge merges two config files, preferring the ones on the left. Any nil or
@@ -226,6 +229,7 @@ func NewConfig(ec2client ec2.EC2MetadataClient) (*Config, error) {
 		errs = append(errs, err)
 	}
 	config := &envConfig
+	isFIPSEnabled = utils.DetectFIPSMode()
 
 	if config.External.Enabled() {
 		if config.AWSRegion == "" {
@@ -656,4 +660,8 @@ func (cfg *Config) String() string {
 		cfg.DynamicHostPortRange,
 		cfg.platformString(),
 	)
+}
+
+func IsFIPSEnabled() bool {
+	return isFIPSEnabled
 }
--- a/agent/utils/fips_linux.go
+++ b/agent/utils/fips_linux.go
@@ -23,10 +23,11 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 )
 
-const fipsModeFilePath = "/proc/sys/crypto/fips_enabled"
+const FIPSModeFilePath = "/proc/sys/crypto/fips_enabled"
 
-func DetectFIPSMode() bool {
-	data, err := os.ReadFile(fipsModeFilePath)
+// DetectFIPSMode checks if FIPS mode is enabled based on the provided file path.
+func DetectFIPSMode(filePath string) bool {
+	data, err := os.ReadFile(filePath)
 	if err == nil && strings.TrimSpace(string(data)) == "1" {
 		logger.Debug("FIPS mode detected on the host")
 		return true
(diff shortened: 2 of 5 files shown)


agent/utils/fips_linux_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+// detectFIPSModeWithPath is a helper function for testing
+func detectFIPSModeWithPath(filePath string) bool {
+	data, err := ioutil.ReadFile(filePath)
+	if err == nil && strings.TrimSpace(string(data)) == "1" {
+		return true
+	}
+	return false
+}
+func TestDetectFIPSMode(t *testing.T) {
```
sparrc, line 37: This unit test appears to just be testing detectFIPSModeWithPath instead of DetectFIPSMode so it's somewhat pointless.

Could we maybe make the DetectFIPSMode function take in the path as an argument, and make `fipsModeFilePath` an exported constant?

That way we can unit-test the actual DetectFIPSMode function by passing in the path, and the only downside is the caller will be slightly redundant like `utils.DetectFIPSMode(utils.FIPSModeFilePath)`

harishxr: Addressed in https://github.com/aws/amazon-ecs-agent/pull/4189/commits/421f95a1f0809cc35d3fcfbc4e8ac7567b8aa762

sparrc (approved):

agent/config/config.go
```diff
@@ -654,3 +658,7 @@ func (cfg *Config) String() string {
 		cfg.platformString(),
 	)
 }
+
+func IsFIPSEnabled() bool {
```
singholt, line 662: where is this method called?

agent/utils/fips_linux.go
```diff
@@ -0,0 +1,37 @@
...
+	"strings"
+
+	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
+)
+
+const FIPSModeFilePath = "/proc/sys/crypto/fips_enabled"
+
+// DetectFIPSMode checks if FIPS mode is enabled based on the provided file path.
+func DetectFIPSMode(filePath string) bool {
+	data, err := os.ReadFile(filePath)
```
singholt, line 30: nit: log the err when there's one?

agent/utils/fips_windows.go
```diff
@@ -0,0 +1,28 @@
+//go:build windows
```
singholt, line 1: do we also need a no-op method in an unsupported file? like https://github.com/aws/amazon-ecs-agent/blob/master/agent/utils/utils_unsupported.go#L1-L2

agent/utils/fips_linux_test.go
```diff
@@ -0,0 +1,52 @@
+//go:build !windows && unit
```
sparrc, line 1: nit: this build constraint should probably be `linux` so that it doesn't include macos or others

agent/utils/fips_windows.go
sparrc: +1 an "unsupported" file with build constraint `!linux` would probably be better here so that we exclude macos and others correctly

agent/utils/fips_linux.go
harishxr: Logging added in [3c0c1b6](https://github.com/aws/amazon-ecs-agent/pull/4189/commits/3c0c1b6bccd0207d76aa19d7418ba5949988ad38)

agent/utils/fips_linux_test.go
```diff
@@ -0,0 +1,52 @@
...
+//
+// or in the "license" file accompanying this file. This file is distributed
+// on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+// express or implied. See the License for the specific language governing
+// permissions and limitations under the License.
+
+package utils
+
+import (
+	"io/ioutil"
```
singholt, line 20: nit: ioutil is deprecated: https://pkg.go.dev/io/ioutil

singholt (approved):
dont forget to squash your commits!

agent/utils/fips_linux_test.go
harishxr: Will get it addressed in a follow up commit

sparrc (approved):

PR #4196, opened 2024-05-30
Add new RestartStatsSet metric to TACS container metric payload

sparrc:
### Summary

Adds the new RestartStatsSet metrics to the TACS container metrics payload.

These stats must be gotten from the Container object but added into the stats queue so that diffs can be calculated. For this reason we add in a new "NonDockerContainerStats" object to grab stats from the Container object rather than from docker stats. This object can be easily extended if more stats like RestartCount (stats from the container object rather than docker) arise in the future.

### Testing

- Unit tests added with a container with a restart policy. Restarts are simulated and then we verify that the expected numbers come out of the stats queue.
- Existing unit test was modified for a container _without_ a restart policy. Verify that this results in a nil RestartStatsSet being returned, so that this metric is left out of the payload for a container that doesn't have it enabled.
- stats package tests were also amended with the `t.Parallel` flag as they require sleeping and each test is fully independent. This reduces the package test time from ~60s to ~10s.

NA

### Licensing

### DIFF
--- a/agent/stats/container.go
+++ b/agent/stats/container.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"time"
 
+	apicontainer "github.com/aws/amazon-ecs-agent/agent/api/container"
 	"github.com/aws/amazon-ecs-agent/agent/config"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient"
 	"github.com/aws/amazon-ecs-agent/agent/dockerclient/dockerapi"
@@ -107,6 +108,29 @@ func (container *StatsContainer) collect() {
 	}
 }
 
+func (container *StatsContainer) getApiContainer(dockerID string) (*apicontainer.Container, error) {
+	if container.resolver == nil {
+		return nil, fmt.Errorf("Container ID=%s has no stats metadata resolver", dockerID)
+	}
+	dockerContainer, err := container.resolver.ResolveContainer(dockerID)
+	if err != nil {
+		return nil, err
+	}
+	return dockerContainer.Container, nil
+}
+
+func getNonDockerContainerStats(apiContainer *apicontainer.Container) *NonDockerContainerStats {
+	if apiContainer == nil {
+		return nil
+	}
+	var nonDockerStats *NonDockerContainerStats
+	if apiContainer.RestartPolicyEnabled() {
+		nonDockerStats = &NonDockerContainerStats{}
+		nonDockerStats.restartCount = int64(apiContainer.RestartTracker.GetRestartCount())
+	}
+	return nonDockerStats
+}
+
 func (container *StatsContainer) processStatsStream() error {
 	dockerID := container.containerMetadata.DockerID
 	seelog.Debugf("Collecting stats for container %s", dockerID)
@@ -114,6 +138,11 @@ func (container *StatsContainer) processStatsStream() error {
 		return errors.New("container processStatsStream: Client is not set.")
 	}
 	dockerStats, errC := container.client.Stats(container.ctx, dockerID, dockerclient.StatsInactivityTimeout)
+
+	apiContainer, err := container.getApiContainer(dockerID)
+	if err != nil {
+		seelog.Errorf("Will not be able to get restart stats set, error: %s", err)
+	}
 
 	returnError := false
 	for {
@@ -140,7 +169,7 @@ func (container *StatsContainer) processStatsStream() error {
 				return err
 			}
 
-			if err := container.statsQueue.Add(rawStat); err != nil {
+			if err := container.statsQueue.Add(rawStat, getNonDockerContainerStats(apiContainer)); err != nil {
 				seelog.Warnf("Container [%s]: error converting stats for container: %v", dockerID, err)
 			}
 		}
--- a/agent/stats/queue.go
+++ b/agent/stats/queue.go
@@ -61,11 +61,14 @@ func (queue *Queue) Reset() {
 }
 
 // Add adds a new set of container stats to the queue.
-func (queue *Queue) Add(dockerStat *types.StatsJSON) error {
+func (queue *Queue) Add(dockerStat *types.StatsJSON, nonDockerStats *NonDockerContainerStats) error {
 	queue.setLastStat(dockerStat)
 	stat, err := dockerStatsToContainerStats(dockerStat)
 	if err != nil {
 		return err
+	}
+	if nonDockerStats != nil {
+		stat.restartCount = &nonDockerStats.restartCount
 	}
 	queue.add(stat)
 	return nil
@@ -85,6 +88,7 @@ func (queue *Queue) add(rawStat *ContainerStats) {
 	queueLength := len(queue.buffer)
 	stat := UsageStats{
 		CPUUsagePerc:      float32(nan32()),
+		RestartCount:      rawStat.restartCount,
 		MemoryUsageInMegs: uint32(rawStat.memoryUsage / BytesInMiB),
 		StorageReadBytes:  rawStat.storageReadBytes,
 		StorageWriteBytes: rawStat.storageWriteBytes,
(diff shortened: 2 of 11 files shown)


agent/stats/queue.go
```diff
@@ -186,6 +190,70 @@ func (queue *Queue) GetStorageStatsSet() (*ecstcs.StorageStatsSet, error) {
 	return storageStatsSet, errOut
 }
 
+// GetRestartStatsSet gets the stats set for container restarts
+func (queue *Queue) GetRestartStatsSet() (*ecstcs.RestartStatsSet, error) {
+	return queue.getRestartStatsSet(getRestartCount)
```
singholt, line 195: do we need a read lock in this method since you're computing what the firstStat and lastStat is?

agent/stats/queue.go
```diff
@@ -186,6 +190,70 @@ func (queue *Queue) GetStorageStatsSet() (*ecstcs.StorageStatsSet, error) {
...
+	queue.lock.Lock()
+	defer queue.lock.Unlock()
+
+	var firstStat, lastStat int64
+	firstStat = -1
+
+	queueLength := len(queue.buffer)
+	if queueLength < 2 {
+		// Need at least 2 data points to calculate this.
+		return nil, fmt.Errorf("need at least 2 data points in queue to calculate CW stats set")
```
singholt, line 208: `... to calculate restart stats set`

agent/stats/queue.go
```diff
@@ -186,6 +190,70 @@ func (queue *Queue) GetStorageStatsSet() (*ecstcs.StorageStatsSet, error) {
...
+		// no non-nil stats found, most likely this means that the container
+		// does not have a restart policy set.
+		return nil, fmt.Errorf("No non-nil restart count stats found, not sending RestartStatsSet." +
+			" Most likely the container does not have a restart policy configured")
+	}
+
+	// get the diff in restart count between the first stat and the last stat in the
+	// queue. Examples:
+	//    [ 0 1 3 3 4 5 ] = 5 - 0 = 5 restarts
+	//    [ 0(sent) 1(sent) 3(sent) 4(unsent) 5(unsent) 7(unsent) ] = 7 - 3 = 4 restarts
```
singholt, line 249: can this be tested with a unit/integ test?

singholt (changes requested):

agent/stats/queue.go
sparrc: yes this is tested in TestContainerStatsCollection_WithRestartPolicy, in that unit test the stats are verified, then the queue is reset, which sets the metrics to "sent" status, and then adds more metrics and again verifies the stats.

agent/stats/queue.go
sparrc: We get a write lock inside getRestartStatsSet. I think we probably could get by with a read lock but all of the existing `get*StatsSet` functions use a write lock so I just followed that pattern.

singholt: I saw that too! there's this `GetLastNetworkStatPerSec` that has a read lock. I wonder why that's the only one with a read lock... 🤔 

https://github.com/aws/amazon-ecs-agent/blob/7884674a94ce88281e44a25e60fc89a64c293f8e/agent/stats/queue.go#L156-L158

agent/stats/types.go
```diff
@@ -29,10 +29,18 @@ type ContainerStats struct {
 	memoryUsage       uint64
 	storageReadBytes  uint64
 	storageWriteBytes uint64
+	restartCount      *int64
```
danehlim, line 32: Is there any specific reason this (and `UsageStats.RestartCount` on L65) needs to be of type `*int64`? Can it be of type `uint64` like the struct fields above it?

agent/stats/container.go
```diff
@@ -107,6 +108,34 @@ func (container *StatsContainer) collect() {
 	}
 }
 
+func (container *StatsContainer) getApiContainer(dockerID string) (*apicontainer.Container, error) {
+	if container.resolver == nil {
+		return nil, nil
+	}
+	task, err := container.resolver.ResolveTask(dockerID)
```
danehlim, line 115: If I understand correctly, using `container.resolver.ResolveTask` and then iterating through all the containers of the task seems unnecessary. Could we leverage using `container.resolver.ResolveContainer` instead (directly if possible), which should also result in a more desirable time complexity?

agent/stats/container.go
```diff
@@ -107,6 +108,34 @@ func (container *StatsContainer) collect() {
 	}
 }
 
+func (container *StatsContainer) getApiContainer(dockerID string) (*apicontainer.Container, error) {
+	if container.resolver == nil {
+		return nil, nil
```
danehlim, line 113: Is container's resolver being nil considered an error case? If so, should an error be returned here?

agent/stats/queue.go
```diff
@@ -186,6 +190,70 @@ func (queue *Queue) GetStorageStatsSet() (*ecstcs.StorageStatsSet, error) {
...
+	queue.lock.Lock()
+	defer queue.lock.Unlock()
+
+	var firstStat, lastStat int64
+	firstStat = -1
+
+	queueLength := len(queue.buffer)
+	if queueLength < 2 {
+		// Need at least 2 data points to calculate this.
+		return nil, fmt.Errorf("need at least 2 data points in queue to calculate restart stats set")
```
danehlim, line 208: Are 2 data points strictly required to send container restart metric to TACS? Why consider a queue containing a single stat with a restart count as invalid/an error case?

agent/stats/container.go
sparrc: Unfortunately ResolveContainer returns a DockerContainer object rather than an apicontainer.Container object, which is what we need here.

agent/stats/queue.go
sparrc: this is because the restart count metric is a monotonic counter that is always increasing, so what we send is the diff between two metrics.

If we accepted a single metric, then we might send a huge metric after an agent/instance reboot, for example if our stats queue looked like this before/after a reboot:

```
[0, 10, 13, 21]
(reboot happens)
[21]
```

So in this case we would not want to send a metric of "21" on startup.

agent/stats/types.go
sparrc: yes! the reason it's int instead uint is because the count comes back from the RestartTracker.GetRestartCount function as an int.

The reason it's a pointer is because we want to exclude this metric if the counter doesn't exist, so having it be a pointer tells us if the value is unset, in which case we don't want to send the RestartStatsSet object to TCS.

sparrc: se https://github.com/aws/amazon-ecs-agent/pull/4196/files#diff-020a54a32e6f08ac86c2901d77f7ca89b0935ddf7e974f42f7ab0f5e353a04c9R242

agent/stats/container.go
sparrc: good call, I will change this

agent/stats/queue.go
sparrc: renamed

agent/stats/container.go
danehlim: Doesn't `DockerContainer` itself contain the `Container` type that is needed? Could we make use of this or is it not possible?

https://github.com/aws/amazon-ecs-agent/blob/dev/agent/api/container/container.go#L353

agent/stats/queue.go
danehlim: Got it, thanks for the example and explanation!

agent/stats/types.go
danehlim: Makes sense, thanks!

agent/stats/container.go
sparrc: yes you are right, I will change this

singholt (approved):

danehlim (dismissed):
nit: Can we also add some unit testing to `agent/stats/queue_test.go` for the changes introduced to `agent/stats/queue.go`?

danehlim (approved):

sparrc:
> nit: Can we also add some unit testing to `agent/stats/queue_test.go` for the changes introduced to `agent/stats/queue.go`?

Good idea! I have now added unit tests here....and incidentally this caught a bug that could happen when only a single stat was added after a queue reset.

singholt (approved):

PR #4204, opened 2024-06-06
cap cpu shares to the max allowed value on Linux

singholt:
### Summary

We've lately had reach-outs from customers using very large instance types experiencing runtime errors. This [u-9tb1.112xlarge](https://aws.amazon.com/ec2/instance-types/high-memory/) instance type, for example, has 448 vCPUs. 

Error (truncated in the ECS describe-task API response)
```
CannotStartContainerError: Error response from daemon: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error setting cgroup config for procHooks process: the maximu
```

Customers defining more than 256 vCPUs as their container CPU in an ECS task definition, will see an error during the cgroup creation. 

The Linux kernel defines a min and max value for the CPU share: [ref](https://github.com/torvalds/linux/blob/0bddd227f3dc55975e2b8dfa7fc6f959b062a2c7/kernel/sched/sched.h#L427-L428) (2^1 - 2^18). Currently we have a `minimumCPUShares` that is set to 2. This PR defines a `maximumCPUShares` equal to 262144.

If you give the kernel anything less or more than these min and max values, the kernel will anyway make sure the value is not out of range and automatically clamp it. Systemd has an additional check that prevents the cgroup creation. A similar patch has been merged into the k8s kubelet as well: https://github.com/kubernetes/kubernetes/pull/93248.

In an ECS task definition, the `container.CPU` is a CPU reservation value (not a hard limit). For example, a container with 1 vCPU defined, on the `u-9tb1.112xlarge` instance, is free to take up the entire 448 vCPUs. Whereas the `task.CPU` is a hard-limit, the task will be throttled beyond the defined value.

We use CPU shares in 2 places:
- When `container.CPU` is defined
- When `task.CPU` is not defined

These are explained below in detail. This PR makes changes to both the above places, to account for `maximumCPUShares`.

### Implementation details

1. When the `container.CPU` is defined, cap it before sending it to Docker.

This handles the case where a task definition has container cpu defined. If the container cpu share is more than the max, we make it be the max. This is also consistent with how we bound the container cpu share to the min value today.

2. When the `task.CPU` is not defined, cap it before creating the task cgroup.

If an ECS task definition does not have `task.CPU` defined, the ECS agent sums up the `container.CPU` values and uses it as the task's CPU share. This is handled by the `buildImplicitLinuxCPUSpec` method - implicit because it was not defined explicitly. When customers explicitly define `task.CPU`, we do not create the task cgroup with a cpu share value; we use cpu quota and period ([ref](https://github.com/aws/amazon-ecs-agent/blob/27ca59a0a052c2e8a6e330899f02969ba15dbe66/agent/api/task/task_linux.go#L168-L171)) to hard-limit the task. 

### Testing

Added unit tests to test the new code. Unfortunately, I'm unable to find capacity for a large instance type (>256 vCPU), hence we can't test this on a real instance. The unit tests suffice for testing the new code added.

enhancement:  cap cpu shares to the max allowed value on Linux.

no

### Licensing

### DIFF
--- a/agent/api/task/task_linux.go
+++ b/agent/api/task/task_linux.go
@@ -43,7 +43,11 @@ import (
 
 const (
 	// Reference: http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDefinition.html
-	minimumCPUShare = 2
+	// These min, max values are defined in the kernel:
+	// https://github.com/torvalds/linux/blob/0bddd227f3dc55975e2b8dfa7fc6f959b062a2c7/kernel/sched/sched.h#L427-L428
+	// We need to make sure task and container cgroups have cpu shares within this range.
+	minimumCPUShares = 2
+	maximumCPUShares = 262144
 
 	minimumCPUPercent = 0
 	bytesPerMegabyte  = 1024 * 1024
@@ -178,11 +182,16 @@ func (task *Task) buildImplicitLinuxCPUSpec() specs.LinuxCPU {
 	// aggregate container CPU shares when present
 	var taskCPUShares uint64
 	for _, container := range task.Containers {
-		if container.CPU < minimumCPUShare {
-			taskCPUShares += minimumCPUShare
+		if container.CPU < minimumCPUShares {
+			taskCPUShares += minimumCPUShares
 		} else {
 			taskCPUShares += uint64(container.CPU)
 		}
+	}
+
+	// If the task CPU shares exceed the maximumCPUShares, we set it to be the maximum permitted by the kernel.
+	if taskCPUShares > maximumCPUShares {
+		taskCPUShares = maximumCPUShares
 	}
 
 	return specs.LinuxCPU{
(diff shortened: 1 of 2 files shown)


agent/api/task/task_linux.go
```diff
@@ -178,8 +182,14 @@ func (task *Task) buildImplicitLinuxCPUSpec() specs.LinuxCPU {
...
+		if container.CPU < minimumCPUShares {
+			taskCPUShares += minimumCPUShares
+		} else if container.CPU > maximumCPUShares {
+			// If one of the containers is requesting more than the max permitted CPU shares,
+			// then we set taskCPUShares to be the maximumCPUShares and break out of this loop.
+			// This is to prevent the taskCPUShares to be more than the max CPU shares permitted by the Linux kernel.
+			taskCPUShares = maximumCPUShares
+			break
 		} else {
 			taskCPUShares += uint64(container.CPU)
```
prateekchaudhry, line 194: Another idea of putting a cap could be, we aggregate and then put a cap. Say we have 2 containers each with container.CPU = `200000`. Then aggregated task cpu shares could be out of bounds, if I understand the problem we have now?

singholt: > Say we have 2 containers each with container.CPU = 200000. Then aggregated task cpu shares could be out of bounds, if I understand the problem we have now?

Yes, the sum would be 400000, and hence out of bounds.

singholt: let me think about this. you're suggestion is valid

singholt: right, the example you shared isnt solved by the current code. I'll make the changes, thanks!!

singholt: updated! and also added this example as a unit test case, thanks!

prateekchaudhry (approved):

Yiyuanzzz (approved):

sparrc (approved):

PR #4206, opened 2024-06-07
Aggregate container metadata across restarts

danehlim:
### Summary

Aggregate container metadata that Docker measures against the last time the container was (re)started time across container restarts. This ensures that such container metadata is tracked cumulatively relative to when the container was originally started and does not get reset upon container restart.

Additionally, save relevant aggregation data in ECS Agent's local state DB so that the behavior mentioned above is persisted in the event ECS Agent goes down and comes back up while containers are still running.

_Listed below are container metadata that Docker measures against the last time the container was (re)started:_

<details>

<summary>Container Metadata Tracked Relative to Container’s Last (Re)start Time by Docker</summary>

### Not sourced from Docker `ContainerStats` API:
* Timestamp that the container started at

### Sourced from Docker `ContainerStats` API:
**CPU:**
* Time consumed
* Time consumed per core (Linux only)
* Time spent in kernel mode by tasks of the cgroup for Linux and by all container processes for Windows
* Time spent in user mode by tasks of the cgroup in user mode for Linux and by all container processes for Windows
* Number of periods with CPU throttling active (Linux only)
* Number of periods that the CPU throttling limit is hit (Linux only)
* Time spent during which the container was CPU throttled (Linux only)

**Memory:**
* Peak memory consumed (Linux only)
* Number of times memory usage hits limits (Linux only)
* Peak memory committed (Windows only)

**Block I/O / Storage:**
* For each specific type of I/O operation (Linux only):
    * For each block device:
        * Number of bytes transferred to the device
        * Number of operations performed on the device
        * Time spent by device to actively service I/O requests from the container
        * Time spent by container waiting for I/O operations to execute
        * Number of merged I/O operations performed on the device
        * Time spent by device to both actively and passively service I/O requests from the container
        * Number of sectors transferred as part of I/O operations
* Number of read operations from the container storage device (Windows only)
* Number of bytes read from the container storage device (Windows only)
* Number of write operations to the container storage device (Windows only)
* Number of bytes written to the container storage device (Windows only)

**Networks:**
* For each container network:
    * Number of bytes received
    * Number of packets received
    * Number of errors received (Linux only)
    * Number of incoming packets dropped
    * Number of bytes sent
    * Number of packets sent
    * Number of errors sent (Linux only)
    * Number of outgoing packets dropped

</details>

### Implementation details

- Ensure `Container` started at time is set to be when the container was first started (and is unchanged across container restarts)
- When a `StatsContainer` is processing a Docker stat and that Docker stat is detected to be the first stat received from Docker after a container restart, cache:
   1. the time which this detection occurred
   1. the last Docker stat from before the restart 
-  Prior to adding a Docker stat for a container to a stats `Queue`, first check if the container has ever restarted before. If it has, then first aggregate metadata from that stat which Docker tracks relative to the container’s last (re)start time with the corresponding metadata from the last Docker stat added to the queue before the most recent container restart
- Leverage existing ECS Agent mechanism that saves `Container` data to BoltDB to store required container restart aggregation data in Agent's local state DB
   - Provide Agent's data client to Agent's stats engine so that whenever aggregation data for a container is updated, it can also be saved in BoltDB 
   - Make `Container` started at time field public so that Agent can marshal/unmarshal this field to/from JSON when Agent saves state
- Add generic `MaxNum` function to `utils` package in ecs-agent module that returns the max value between two numbers. This functionality is required in multiple places for the aggregation logic

### Testing

Manually tested by building and running a custom ECS Agent with the changes from this pull request and having container restart policy enabled. Verified that container metadata is aggregated as expected upon container restarts and that in the event of ECS Agent restart the container restart aggregation data is persisted in and loaded from ECS Agent's local state DB.

Aggregate container metadata across restarts

No

### Licensing

### DIFF
--- a/agent/api/container/container.go
+++ b/agent/api/container/container.go
@@ -29,7 +29,6 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger/field"
-
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/cihub/seelog"
 	"github.com/docker/docker/api/types"
@@ -323,9 +322,16 @@ type Container struct {
 	// pause container
 	ContainerTornDownUnsafe bool `json:"containerTornDown"`
 
-	createdAt  time.Time
-	startedAt  time.Time
-	finishedAt time.Time
+	createdAt time.Time
+	// StartedAtUnsafe specifies the started at time of the container.
+	// It is exposed outside this container package so that it is marshalled/unmarshalled in JSON body while
+	// saving state.
+	// NOTE: Do not access StartedAtUnsafe directly. Instead, use `GetStartedAt` and `SetStartedAt`.
+	StartedAtUnsafe time.Time `json:"startedAt,omitempty"`
+	// setStartedAtOnce is used to set the value of the container's started at time only the first time SetStartedAt is
+	// invoked.
+	setStartedAtOnce sync.Once
+	finishedAt       time.Time
 
 	labels map[string]string
 
@@ -342,11 +348,22 @@ type Container struct {
 	// as restart count and last restart time. This is only initialized if the container
 	// has a restart policy defined and enabled.
 	RestartTracker *restart.RestartTracker `json:"restartTracker,omitempty"`
+	// RestartAggregationDataForStatsUnsafe specifies the restart aggregation data used for stats of the container.
+	// It is exposed outside this container package so that it is marshalled/unmarshalled in JSON body while
+	// saving state.
+	// NOTE: Do not access RestartAggregationDataForStatsUnsafe directly. Instead, use
+	// `GetRestartAggregationDataForStats` and `SetRestartAggregationDataForStats`.
+	RestartAggregationDataForStatsUnsafe ContainerRestartAggregationDataForStats `json:"RestartAggregationDataForStats,omitempty"`
 }
 
 type DependsOn struct {
 	ContainerName string `json:"containerName"`
 	Condition     string `json:"condition"`
+}
+
+type ContainerRestartAggregationDataForStats struct {
+	LastRestartDetectedAt     time.Time       `json:"LastRestartDetectedAt,omitempty"`
+	LastStatBeforeLastRestart types.StatsJSON `json:"LastStatBeforeLastRestart,omitempty"`
 }
 
 // DockerContainer is a mapping between containers-as-docker-knows-them and
@@ -670,7 +687,11 @@ func (c *Container) SetStartedAt(startedAt time.Time) {
 	c.lock.Lock()
 	defer c.lock.Unlock()
 
-	c.startedAt = startedAt
+	c.setStartedAtOnce.Do(func() {
+		if c.StartedAtUnsafe.IsZero() {
+			c.StartedAtUnsafe = startedAt
+		}
+	})
 }
 
 // SetFinishedAt sets the timestamp for container's stopped time
@@ -698,7 +719,7 @@ func (c *Container) GetStartedAt() time.Time {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
 
-	return c.startedAt
+	return c.StartedAtUnsafe
 }
 
 // GetFinishedAt sets the timestamp for container's stopped time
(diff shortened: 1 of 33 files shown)


agent/stats/container.go
```diff
@@ -183,3 +201,44 @@ func (container *StatsContainer) terminal() (bool, error) {
 	}
 	return dockerContainer.Container.KnownTerminal(), nil
 }
+
+// syncContainerRestartAggregationData updates a container's restart aggregation data if a container restart has been
+// detected. This method returns true if a restart was detected and returns false otherwise.
+func (container *StatsContainer) syncContainerRestartAggregationData() bool {
+	_, dockerContainerData := container.client.DescribeContainer(container.ctx, container.containerMetadata.DockerID)
+	restartDetected := dockerContainerData.StartedAt.Sub(container.containerMetadata.StartedAt) > 0
```
sparrc, line 209: instead of calling the docker API on every stat, could we use the restart tracker here to determine if a restart happened? like 

```
apicontainer.RestartTracker.GetRestartCount() > 0
```

agent/api/container/container.go
```diff
@@ -342,13 +348,24 @@ type Container struct {
...
+	RestartAggregationDataForStatsUnsafe ContainerRestartAggregationDataForStats `json:"RestartAggregationDataForStats,omitempty"`
 }
 
 type DependsOn struct {
 	ContainerName string `json:"containerName"`
 	Condition     string `json:"condition"`
 }
 
+type ContainerRestartAggregationDataForStats struct {
+	LastRestartDetectedAt     time.Time       `json:"LastRestartDetectedAt,omitempty"`
```
sparrc, line 365: Do we need this field? if we get the api container object could we use the restart tracker's [GetLastRestartAt](https://pkg.go.dev/github.com/aws/amazon-ecs-agent/ecs-agent@v0.0.0-20240528202446-27ca59a0a052/api/container/restart#RestartTracker.GetLastRestartAt) function?

agent/stats/container.go
danehlim: I don't think we can do this because `apicontainer.RestartTracker.GetRestartCount() > 0` tells us if a container restart has ever happened before but not if a restart has just happened.

danehlim: I don't think we can do this because `apicontainer.RestartTracker.GetRestartCount() > 0` tells us if a container restart has ever happened before but not if a restart has just happened (which is what we are interested in here).

agent/api/container/container.go
danehlim: We make a distinction between: 
a. the last time the container was restarted at as recorded in task engine state 
b. the last time a container restart was detected at by the stats engine

This is because stats collection/processing is an asynchronous workflow of Agent. An implication of this is that there is no guarantee that (b) gets updated before (a) gets updated nor any guarantee that (b) gets updated after (a) gets updated.

In the event that (b) gets updated before (a) gets updated, stats collection/processing can continue and not be blocked despite (a) not yet being updated. 

If we were to use (a) as the source of truth and not make the distinction above, we would need to block stats collection/processing on (a) being updated every time a stat is returned by DockerStats API.

danehlim: We make a distinction between: 
a. the last time the container was restarted at as recorded in task engine state 
b. the last time a container restart was detected at by the stats engine

This is because stats collection/processing is an asynchronous workflow of Agent. An implication of this is that there is no guarantee that (b) gets updated before (a) gets updated nor any guarantee that (b) gets updated after (a) gets updated.

In the event that (b) gets updated before (a) gets updated, stats collection/processing can continue and not be blocked despite (a) not yet being updated. 

If we were to use (a) as the source of truth and not make the distinction above, we would need to block stats collection/processing on (a) being updated every time a stat is returned by DockerStats API to Agent.

agent/stats/queue_test.go
```diff
@@ -785,3 +786,97 @@ func TestPerSecNetworkStatSetFailWithOneDatapoint(t *testing.T) {
 	stats, err := queue.GetNetworkStatsSet()
 	require.Errorf(t, err, "Received unexpected network stats set %v", stats)
 }
+
```
sparrc, line 789: Could we add a unit test in here to directly test the new `AddContainerStat` function?

sparrc (approved):

agent/stats/queue_test.go
danehlim: I will add this in a follow up pull request. The existing `Add` function (which `AddContainerStat` calls) does not have a unit test that directly calls it either, so I will add a unit test for that as well in the follow up. 👍

Yiyuanzzz (approved):

agent/stats/queue_test.go
danehlim: Follow up pull request: https://github.com/aws/amazon-ecs-agent/pull/4208

PR #4216, opened 2024-06-20
Creating Buildspecs for Amazon Linux 2/2023

Ephylouise:
### Creating Buildspecs for Amazon Linux 2/2023

To manage the dependency of the Go version, which needs to be 1.22.0 or greater (as defined in ecs-agent.spec for Amazon Linux builds), Go specifications must be handled in the buildspec files. AL2 uses the YUM package manager and AL2023 uses the DNF package manager. Therefore, the code to manage the Go dependency needs to be different. To account for the dependencies of Amazon Linux versions, a version-specific buildspec file is needed for each. 

### Externally creating 2 new CodeBuild projects as well.

The operating system and image defined in CodeBuild job configurations determine the possible artifacts built by the project. The dev-amd project will not be able to make both Amazon Linux versions. 

During testing, it was discovered that Amazon Linux 2023 can be built on agent-dev-amd CodeBuild project because the image is aws/codebuild/amazonlinux2-x86_64-standard:5.0, which can build Amazon Linux 2023 (despite the misnomer of 'linux2'). However, to build AL2, the image ending in "standard:4.0" should be used. Only one image can be defined for each CodeBuild job. To keep projects decoupled, instead of lumping amzn2023 in with the "generic" RPM and agent.tar, it might be best to create 2 new projects. 

Two CodeBuild jobs called 'agent-dev-amzn2-stephy' and 'agent-dev-amzn2023-stephy' have been created with their respective and specific image requirements. 

### Licensing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -394,7 +394,7 @@ amazon-linux-sources.tgz:
 	cp packaging/amazon-linux-ami-integrated/amazon-ecs-volume-plugin.conf amazon-ecs-volume-plugin.conf
 	cp packaging/amazon-linux-ami-integrated/amazon-ecs-volume-plugin.service amazon-ecs-volume-plugin.service
 	cp packaging/amazon-linux-ami-integrated/amazon-ecs-volume-plugin.socket amazon-ecs-volume-plugin.socket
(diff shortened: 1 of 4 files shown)


Ephylouise:
Accidental master PR.

PR #4217, opened 2024-06-24
[Windows] refactor stats package to fetch EBS Volume stats for Windows

rawahars:
### Summary

Presently, the code path in `stats package` of the agent, which invokes the CSI Client to fetch volume metrics, is limited to Linux platform only. The same code path can be used for Windows as well.

This change refactors the existing code to ensure that the EBS volume stats are collected for Windows as well in addition to Linux.

### Implementation details

Refactor the code from Linux specific files to common platform files.

### Testing

Using unit and integration tests for Windows and Linux.

refactor stats package to fetch EBS Volume stats for Windows

   No

### Licensing


agent/stats/engine.go
```diff
@@ -1077,3 +1079,72 @@ func (engine *DockerStatsEngine) SetPublishServiceConnectTickerInterval(publishS
...
+		}
+	}
+	return metrics
+}
+
+func (engine *DockerStatsEngine) getVolumeMetricsWithTimeout(volumeId, hostPath string) (*csiclient.Metrics, error) {
+	derivedCtx, cancel := context.WithTimeout(engine.ctx, time.Second*1)
+	// releases resources if GetVolumeMetrics finishes before timeout
+	defer cancel()
+	return engine.csiClient.GetVolumeMetrics(derivedCtx, volumeId, hostPath)
```
saurabhc123, line 1149: Is this method implemented for Windows?

saurabhc123 (approved):
LGTM. We just need to ensure that the downstream method within csi client is implemented and tested for Windows.

agent/stats/engine.go
rawahars: Yes the code path is common for both Linux and Windows- https://github.com/aws/amazon-ecs-agent/blob/437e051d75d96f9c5619f8b8742bd84d5b959e9c/ecs-agent/csiclient/csi_client.go#L152

`GetVolumeMetrics` depends upon few methods-
- `NodeGetVolumeStats` from `CSI Driver` which is dependent upon the following-
    - `PathExists` in `mounter` package
    - `IsBlockDevice` in `utils` package
    - `getBlockSizeBytes` from `CSI Driver`
    - `GetMetrics` from `CSI Driver`
        - `getFsInfo`
            - `Info` method from `fs` package

From the list above, the following methods are implemented as part of the PR you raised- https://github.com/aws/amazon-ecs-agent/pull/4214
- `PathExists`
- `getBlockSizeBytes`
- `Info`

`IsBlockDevice` method is part of https://github.com/aws/amazon-ecs-agent/pull/4224.

agent/stats/engine.go
```diff
@@ -1077,3 +1079,72 @@ func (engine *DockerStatsEngine) SetPublishServiceConnectTickerInterval(publishS
 func (engine *DockerStatsEngine) GetPublishMetricsTicker() *time.Ticker {
 	return engine.publishMetricsTicker
 }
+
+func (engine *DockerStatsEngine) getEBSVolumeMetrics(taskArn string) []*ecstcs.VolumeMetric {
+	task, err := engine.resolver.ResolveTaskByARN(taskArn)
+	if err != nil {
+		logger.Error(fmt.Sprintf("Unable to get corresponding task from dd with task arn: %s", taskArn))
```
xxx0624, line 1086: are we ignoring the error intentionally? If so, can we log the error as well?

rawahars: This is from the existing code which has been refactored into platform agnostic file.
Also, the `logger.Error` method does log the error via logrus. Reference- https://github.com/aws/amazon-ecs-agent/blob/437e051d75d96f9c5619f8b8742bd84d5b959e9c/ecs-agent/logger/global.go#L66

xxx0624: I see. IMO technically we can still log the error message here for debug purpose. It's not a blocker for me.

rawahars: So the error is being logged into the log file when we use `logger.Error`. 
Basically, If i understand correctly, `logger` package is an abstraction over actual logger library used in the agent.

xxx0624: Not quite understand how the error will be logged. But given this is for ECS agent, I'll prefer someone from ECS agent team to take a look at this.

mye956: I personally would like to log the actual error as well which we're already doing in other portions of the code file ([ref](https://github.com/aws/amazon-ecs-agent/blob/master/agent/stats/engine.go#L858C4-L861C6)). But not a big blocker for me either.

> So the error is being logged into the log file when we use logger.Error.

So the log message itself will appear as an ERROR (in this case it's `Unable to get corresponding task from dd with task arn: %s`) but I think what @xxx0624 is referring to the actual `err` that we get.

agent/stats/engine.go
mye956: Basically, the idea here is to move what we have in `engine_unix.go` into `engine.go` so that both platforms can use the same implementation (please correct me if this is incorrect). 

>Refactor the code from Linux specific files to common platform files.

Are there any notable changes/refactors that has been made to the existing implementation of `getEBSVolumeMetrics`, `fetchEBSVolumeMetrics`, and `getVolumeMetricsWithTimeout` ? Otherwise, calling it a refactor is a bit misleading IMO.

agent/stats/engine.go
rawahars: Yes that is correct. @mye956 Do we want to log that into the log files?
This would be a change which impacts Linux and therefore, if you feel that is viable, then I can add the same.

agent/stats/engine.go
rawahars: You are correct. This PR just moves the methods from Linux specific files to platform agnostic ones. 
The refactor is from the perspective of moving things from platform to general code files.

If you feel strongly, I can change the commit message and PR title.

mye956: Not a big blocker for me. Was more so of getting clarification.

agent/stats/engine.go
mye956: It would be more of a nice to have but it's also not a blocker in my opinion.

mye956 (approved):

xxx0624 (approved):

PR #4218, opened 2024-06-26
Fix restartAttemptPeriod time.Duration/int parsing

sparrc:
### Summary

The RestartPolicy struct governs the restart policy of the container. Because the API model is an integer, and we had previously made the `restartAttemptPeriod` field a `time.Duration` type, JSON parsing was interpreting a value of `300` as `300 nanoseconds`.

Instead, `restartAttemptPeriod: 300` should be interpreted as `300 seconds`.

This PR changes the agent RestartPolicy struct to use `int` to match the API model, and caste this int to seconds where needed (in the ShouldRestart function).

Unit tests have also been updated to unmarshal the object from a JSON blob, so that issues like this would be caught in unit tests.

### Testing

- unit tests updated to catch this scenario (JSON integer being erroneously marshaled to nanoseconds)
- tested end-to-end in gamma environment

NA

### Licensing

### DIFF
--- a/agent/engine/engine_integ_test.go
+++ b/agent/engine/engine_integ_test.go
@@ -824,7 +824,7 @@ func TestPullContainerWithAndWithoutDigestInteg(t *testing.T) {
 // Tests that pullContainer pulls the same image when a digest is used versus when a digest
 // is not used.
 func TestPullContainerWithAndWithoutDigestConsistency(t *testing.T) {
-	image := "public.ecr.aws/docker/library/alpine:3.19"
+	image := "public.ecr.aws/docker/library/alpine:3.19.1"
 	imageDigest := "sha256:c5b1261d6d3e43071626931fc004f70149baeba2c8ec672bd4f27761f8e1ad6b"
 
 	// Prepare task
--- a/ecs-agent/api/container/restart/restart_tracker_test.go
+++ b/ecs-agent/api/container/restart/restart_tracker_test.go
@@ -17,20 +17,34 @@
 package restart
 
 import (
+	"encoding/json"
+	"fmt"
 	"testing"
 	"time"
 
 	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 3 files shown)


ecs-agent/api/container/restart/restart_tracker_test.go
```diff
@@ -17,20 +17,34 @@
...
 )
 
+// unmarshalStandardRestartPolicy allows us to verify that the restart policy unmarshalled
+// from JSON (as it comes in the task payload) is translated to expected behavior in
+// the restart tracker.
+func unmarshalStandardRestartPolicy(t *testing.T, ignoredCode int) RestartPolicy {
+	rp := RestartPolicy{}
+	jsonBlob := fmt.Sprintf(`{"enabled": true, "restartAttemptPeriod": 60, "ignoredExitCodes": [%d]}`, ignoredCode)
+	err := json.Unmarshal([]byte(jsonBlob), &rp)
+	require.NoError(t, err)
```
danehlim, line 38: > ... verify that the restart policy unmarshalled
> from JSON (as it comes in the task payload) is translated to expected behavior

nit (non-blocking): In addition to requiring no error, we could also require that `rp`'s enabled to be true, restart attempt period to be 60, and ignored exit codes to be `ignoredCode` as further verification of expected behavior.

danehlim (approved):

timj-hh (approved):

agent/engine/engine_integ_test.go
```diff
@@ -824,7 +824,7 @@ func TestPullContainerWithAndWithoutDigestInteg(t *testing.T) {
 // Tests that pullContainer pulls the same image when a digest is used versus when a digest
 // is not used.
 func TestPullContainerWithAndWithoutDigestConsistency(t *testing.T) {
-	image := "public.ecr.aws/docker/library/alpine:3.19"
+	image := "public.ecr.aws/docker/library/alpine:3.19.1"
```
sparrc, line 827: unrelated to this change, fix for failing integ test

danehlim (approved):

PR #4220, opened 2024-06-27
Building Amazon Linux RPMs and a Windows executable

Ephylouise:
### Summary

The following changes were made to ensure that the project can build all necessary artifacts before deprecating a legacy component. It specifically modifies the buildspecs to include build capabilities for the full ECS-init RPM/DEB package as well as a Windows executable file. This will ensure that all required images are created during the Agent build process.

The goal is to have the artifacts built from CodeBuild jobs, which run the project's buildspecs, to be used in testing. Providing updates to the buildspecs and minor adjustments to the Makefile will enable CloudFormation and CodeBuild configurations that produce the full package of images in the development artifacts bucket.

### Implementation details

A new make target called amazon-linux-rpm-codebuild was created. It uses similar logic as the amazon-linux-rpm-integrated target, but with minor changes to accommodate the CodeBuild build process. To avoid altering the integrated target, which can be used in Koji builds, the new target was created. The new target is called in the build specifications along with the Windows executable file, which relies on the docker-release target.

A new file was added, called pr-build-amzn.yml to accommodate the dependencies of the Amazon Linux builds. Amazon Linux 2, for example uses package manager yum, while Amazon Linux 2023 uses package manager dnf which are used to install Go, as required versions do not come preinstalled. Combining the Amazon Linux 2023 script with the pr-build.yml was considered, but that would result in a complex artifact build script that would be difficult to maintain and debug if issues arise. To keep the code clean and decoupled, the Amazon Linux RPMs (AL2 amd, AL2 arm, AL23 amd, AL23 arm) are all addressed in the new file.

#### Files changed:

* Makefile
    * Added a new make target amazon-linux-rpm-codebuild
* pr-build-ubuntu.yml
    * Modified the forked testing path adjustment if/then statement to [single brackets] instead of [[double]], which was breaking the build due to the Ubuntu platform syntax expectations
* pr-build.yml
    * Removed a duplicate command to return go version (left one in the install stage)
    * Added WINDOW_EXE path assignment
    * Add make docker-release TARGET_OS="windows" to build amazon-ecs-agent.exe
    * Moved the windows artifact to the current directory
    * Moved the list command from before the renaming function to after it
    * Added windows artifact to files section to be stored in S3
* pr-build-amzn.yml
    * A new buildspecs file specifically to handle Amazon Linux OS within CodeBuild builds.
    * Assigns AL version and CB image variables
    * Add function to determine the Go installation process depending on OS image and version
    * Add make amazon-linux-rpm-integrated to build AL rpms
    * Rename rpm based on architecture
    * Add RPM and Build Log to artifacts files 

Building the Amazon Linux RPMs will require 4 new CodeBuild (CB) projects, one for each Amazon Linux variant and architecture combination. This is due to the build process used to create the artifacts stored in the S3 bucket, as the Amazon Linux builds require separate image configurations.

* The existing Agent build process utilizes 4 CB projects that run on different images:
    * agent-dev-amd (amazonlinux2-x86_64-standard:5.0)
    * agent-dev-arm (amazonlinux2-aarch64-standard:3.0)
    * agent-dev-ubuntu-amd (Linux EC2 with external docker registry)
    * agent-dev-ubuntu-arm (ARM with external docker registry)
* The new CB jobs for the Amazon Linux RPMs will also require their own images, as follows:
    * agent-dev-amzn2-arm (amazonlinux2-aarch64-standard:2.0) - Amazon Linux 2 ARM
    * agent-dev-amzn2-amd (amazonlinux2-x86_64-standard:4.0) - Amazon Linux 2 AMD
    * agent-dev-amzn2023-arm (amazonlinux2-aarch64-standard:3.0) - Amazon-Linux 2023 ARM
    * agent-dev-amzn2023-amd (amazonlinux2-x86_64-standard:5.0) - Amazon-Linux 2023 AMD
* It's worth noting that while the AL2023 images match the existing dev-amd/dev-arm CB images, they require separate buildspecs files due to differences in dependency handling.

### Testing

Testing was performed in a separate AWS account using CodeBuild projects and webhooks to validate the modifications. The existing Agent CodeBuild jobs and the 4 new proposed projects were replicated in the separate account. The result was successful, with the following new artifacts and associated build logs landing in the expected S3 bucket:

* amazon-ecs-agent.exe 
*  amazon-ecs-init-1.82.4-1.amzn2.aarch64.rpm 
* amazon-ecs-init-1.82.4-1.amzn2.x86_64.rpm
* amazon-ecs-init-1.82.4-1.amzn2023.aarch64.rpm
*  amazon-ecs-init-1.82.4-1.amzn2023.x86_64.rpm 
* Build logs for each new project, named build_<amzn-version>_<arch>.log“

### Licensing


Ephylouise:
The commit history on this PR #4220 was rather long. I used `git rebase` to squash several commits and consolidate the PR.

sparrc (dismissed):

buildspecs/pr-build-amzn.yml
```diff
@@ -0,0 +1,120 @@
...
+
+env:
+  variables:
+    # Github username of the forked repo on which to make builds
+    GITHUBUSERNAME: aws
+
+phases:
+  install:
+    commands:
+      # Same buildspec used for different Amazon Linuc versions - detect verison to use in Go installation
```
singholt, line 11: typo: ~~Linuc~~ Linux

buildspecs/pr-build-amzn.yml
```diff
@@ -0,0 +1,120 @@
...
+           x86_64)
+             architecture="amd64"
+           ;;
+           aarch64)
+             architecture="arm64"
+           ;;
+        esac
+
+      # Identify the correct AL2023 version to use in the Go installation
+      - AL23_VERSION="$(curl -s https://al2023-repos-us-west-2-de612dc2.s3.dualstack.us-west-2.amazonaws.com/core/releasemd.xml | xmllint --xpath "string(//root/releases/release[last()]/@version)" -)"
```
singholt, line 44: Can you add a comment like: https://github.com/aws/amazon-ecs-ami/blob/f7a406dcea7467061452a6da77b5543ff25d4c46/generate-release-vars.sh#L95-L97

incase this stops working, we would know where to look for the right source URL.

buildspecs/pr-build-amzn.yml
```diff
@@ -0,0 +1,120 @@
...
+      - sudo rm -rf /root/.goenv
+      - sudo rm -rf /usr/local/go/bin/go 
+
+      # Install dependencies and define variables based on the build image used
+      # Amazon Linux 2023 uses package manager DNF, while older versions use YUM
+      # Set the appropriate AL23 version echo string to include in build log
+      - |
+        if [[ "$amzn_version" = "amzn2023" ]]; then
+          sudo dnf --releasever="$AL23_VERSION" update -y
+          sudo yum install -y golang
```
singholt, line 58: should we install a specific go version? especially the one in https://github.com/aws/amazon-ecs-agent/blob/master/GO_VERSION.

buildspecs/pr-build-amzn.yml
Ephylouise: Good catch. I will fix that.

buildspecs/pr-build-amzn.yml
Ephylouise: Yes, great idea.

buildspecs/pr-build-amzn.yml
Ephylouise: Good point. Installing go needs to be handled differently for Amazon Linux, but we should still be specifying the version based on the GO_VERSION file. Currently testing the update with the CodeBuild projects now and will update this PR accordingly after. Thank you.

sparrc: Should we actually install the latest version of golang that Amazon Linux has available? The reason being that this is what Amazon Linux will be doing when it builds ecs-agent in their koji build servers, so we might actually want to mimic that behavior in our codebuild jobs.

singholt: thats a good point. I noticed the other workflows install a specific Go version: https://github.com/aws/amazon-ecs-agent/blob/86d635bcde50bd95a17c8b59db67cfdca20badff/buildspecs/pr-build.yml#L23. 

I think we should align all pr workflow files irrespective of what we end up doing, I am ok with either options.

Ephylouise: Oh I understand what you are saying. Do you have any resource that I could check how the Koji builds install Go? The originally installation I had was installing the latest version of Go, but not necessarily the latest version available on Amazon Linux. [AL23](https://docs.aws.amazon.com/linux/al2023/ug/go.html), for example, does update Go with its releases, but will not always be in "lockstep" with the upstream Go releases. 

Perhaps checking for the latest available version of Go for each AL version would be best. For example:
`# Define the latest Go version available in Amazon Linux package repositories
- GOVERSION=$(yum list golang --showduplicates | awk -F'.' '{print $1"."$2}' | sort -r | head -1)`

As the team and Agent build veterans, which of these would be best? 

1. Install the latest Go version (original code)
2. Install the Go version defined in the GO_VERSION file
3. Install the latest Go version available in the AL releases. 

The answer likely lies in how Koji build servers handle Go. I will keep digging, but would value your input.

singholt (approved):

sparrc (approved):

PR #4224, opened 2024-06-28
added the implementation for IsBlockDevice method in CSI Driver for Windows

rawahars:
### Summary

Presently, we were using the empty implementation of the IsBlockDevice method for Windows. In this change, we will add the same.

In order to accomplish the same, we use `GetVolumeNameForVolumeMountPointW` Win32 API. Reference: https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getvolumenameforvolumemountpointw

We will call the API to fetch the Volume GUID. If the mount path is invalid, this method would fail and therefore, we can be certain that the mount path is not the block device mount path.

### Implementation details

Added the implementation to invoke Win32 API for `GetVolumeNameForVolumeMountPointW` to figure out if the device is block device or not. Also, since the implementation depended upon `golang.org/x/sys/windows`, the same were vendor'ed in.

### Testing

A sample binary was created with the same changes as the method implementation in this PR. That was tested for block and non-block devices.

added the implementation for IsBlockDevice method in CSI Driver for Windows

No

### Licensing


saurabhc123 (approved):
LGTM.

xxx0624 (approved):

ecs-agent/daemonimages/csidriver/util/utils_windows.go
```diff
@@ -22,10 +22,56 @@ limitations under the License.
...
+	funcGetVolumeNameForVolumeMountPointW func(a ...uintptr) (r1 uintptr, r2 uintptr, lastErr error) = procGetVolumeNameForVolumeMountPointW.Call
+)
+
+// IsBlockDevice checks if the given path is a block device on Windows.
 func IsBlockDevice(fullPath string) (bool, error) {
-	// TODO
-	return false, errors.New("not supported")
+	// Ensure the fullPath ends with a backslash.
+	// This is as per the API documentation- https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getvolumenameforvolumemountpointw#parameters
+	if !strings.HasSuffix(fullPath, "\\") {
```
mye956, line 50: Q: From the doc, it seems like only one `\` is needed (for example, "Y:\MountX\"). Any reason why we're appending 2?

rawahars: The second `\` is for escaping purposes. The string when read in Go would have escaped `\` and therefore, we need to cater to the same.

mye956 (approved):

xxx0624 (approved):

PR #4233, opened 2024-07-03
Changing Requires to Wants to make ECS more resilient and check if docker service is running in ECS service file

mye956:
### Summary
The goal of this PR is to make the ECS service more resilient in regards to when docker is up and available. There can be cases where docker goes down during instance bootstrapping (e.g. upgrades/restarts) and ECS service is trying to start up. However, it can't since docker is down and will fail due to a service dependency error which isn't taken into consideration for the restart policy that we defined. In other words, since the ECS service never starts up it's not consider a failure. To try to address this inconsistent race condition, we will be changing `Requires` to `Wants` and will have a check if docker is up and running during PreStart for the ECS service.

The reason why we're switching over to `Wants` is because it's less restrictive than `Requires`. Essentially, if the units within the `ecs.service` file are not found or fail to start, the current unit(ECS) will continue to function. This is the recommended way to configure most dependency relationships. Again, this implies a parallel activation unless modified by other directives. ([ref](https://www.digitalocean.com/community/tutorials/understanding-systemd-units-and-unit-files))

In addition, we'll also have the ECS service to check if docker is running as part of its PreStart so that it can fail a bit faster. Otherwise, it will fail once ECS reaches its Start phase where it tries to start up the ECS Agent container.

### Implementation details
* Changed `Requires` to `Wants`. These were made in the following files:
  *  `packaging/amazon-linux-ami-integrated/ecs.service`
  * `packaging/generic-deb-integrated/debian/ecs.service`
  * `packaging/generic-rpm-integrated/ecs.service`
* Added check if docker is up and running within `packaging/amazon-linux-ami-integrated/ecs.service` via `systemctl is-active docker`
  * If it's not, we will exit with error code 1 (Note: We've defined ECS to ignore exit code 5)
  * We won't include this check in the other service files as the location of systemctl can vary across distros. Instead, we will look to change our public docs regarding installing ECS on non-AL AMIs ([ref](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-install.html))

### Testing

Built a custom AMI off of an older ami (specifically `20220304`) where docker would do an auto-upgrade.

Load tested with 16600 instances and all were able to start up ECS after a docker upgrade
```
NEW INSTANCES LAUNCHED = 100 ; TOTAL INSTANCES LAUNCHED = 16600
TOTAL INSTANCES REGISTERED = 16600 / 16600
```

A few other edge cases that were tested:
* Stopping both ECS + Docker -> Start back up ECS
Behavior: Once ECS tried starting back up, docker will also start back up 
* Stopping just Docker while ECS is running
Behavior:ECS service will fail. After a bit, ECS will restart which will also bring docker back up
* Disabling just Docker while ECS is running
Behavior: ECS service will fail. After a bit, ECS will restart which will also bring docker back up
* Run ECS service on a vanilla AL2 AMI
Behavior: ECS service will never reach a running state and will fail due to docker not being present when it tries starting up the ECS Agent docker container. It will retry starting up every 10 second which is expected based on what we’ve defined in our systemd service unit file for ECS.

Old Behavior:
* Stopping both ECS + Docker -> Start back up ECS
Behavior: Once ECS starts back up, docker will also start back up
* Stopping just Docker while ECS is running
Behavior: ECS will also go down and will remain down. It will only start back up if a user manually starts it back up.
* Disabling just Docker while ECS is running
Behavior: ECS will also go down and will remain down. It will only start back up if a user manually starts it back up.
* Run ECS on a vanilla AL2 AMI
Behavior: ECS fails to start up at all.

Enhancement: Change ECS service file from requires to wants and checking if docker is running during ECS service PreStart

### Licensing

### DIFF
--- a/packaging/amazon-linux-ami-integrated/ecs.service
+++ b/packaging/amazon-linux-ami-integrated/ecs.service
@@ -15,7 +15,7 @@
 [Unit]
 Description=Amazon Elastic Container Service - container agent
 Documentation=https://aws.amazon.com/documentation/ecs/
-Requires=docker.service
+Wants=docker.service
 After=docker.service
 After=cloud-final.service
 
@@ -25,6 +25,7 @@ Restart=on-failure
 RestartPreventExitStatus=5
 RestartSec=10s
 EnvironmentFile=-/etc/ecs/ecs.config
+ExecStartPre=/bin/bash -c 'if [ $(/usr/bin/systemctl is-active docker) != "active" ]; then exit 1; fi'
 ExecStartPre=/usr/libexec/amazon-ecs-init pre-start
 ExecStart=/usr/libexec/amazon-ecs-init start
 ExecStop=/usr/libexec/amazon-ecs-init stop
--- a/packaging/generic-deb-integrated/debian/ecs.service
+++ b/packaging/generic-deb-integrated/debian/ecs.service
@@ -15,7 +15,7 @@
 [Unit]
 Description=Amazon Elastic Container Service - container agent
 Documentation=https://aws.amazon.com/documentation/ecs/
-Requires=docker.service
+Wants=docker.service
 After=docker.service
 
 [Service]
--- a/packaging/generic-rpm-integrated/ecs.service
+++ b/packaging/generic-rpm-integrated/ecs.service
@@ -15,7 +15,7 @@
 [Unit]
 Description=Amazon Elastic Container Service - container agent
 Documentation=https://aws.amazon.com/documentation/ecs/
-Requires=docker.service
+Wants=docker.service
 After=docker.service
 
 [Service]


yinyic:
Could we also add to the testing section what the current (old) behavior is for comparison? Otherwise the change looks good. Thanks Mike!

packaging/amazon-linux-ami-integrated/ecs.service
```diff
@@ -25,6 +25,7 @@ Restart=on-failure
 RestartPreventExitStatus=5
 RestartSec=10s
 EnvironmentFile=-/etc/ecs/ecs.config
+ExecStartPre=/usr/bin/systemctl -q is-active docker
```
prateekchaudhry, line 28: Can this be verified that this will not be 5 for inactive? Because of `RestartPreventExitStatus=5` policy

packaging/generic-deb-integrated/debian/ecs.service
```diff
@@ -24,6 +24,7 @@ Restart=on-failure
 RestartSec=10s
 EnvironmentFile=-/var/lib/ecs/ecs.config
 EnvironmentFile=-/etc/ecs/ecs.config
+ExecStartPre=/usr/bin/systemctl -q is-active docker
```
prateekchaudhry, line 27: `/usr/bin/systemctl` does this path ever change with platform type? 🤔

packaging/amazon-linux-ami-integrated/ecs.service
mye956: On the following [page](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#Process%20Exit%20Codes), it looks like it could IF "The program is not installed."? However, this should have been hit in the case where I tried with a vanilla AL AMI where docker isn't installed. For reference, this is the specific exit code:

```
Jul 03 16:23:25 ip-172-31-46-143.us-west-2.compute.internal systemd[1]: ecs.service: control process exited, code=exited status=3
```

mye956: I guess, we can just always exit with 1 (or just any other number other than 5)

mye956: Will do something along the following: `/bin/bash -c 'if [ $(/usr/bin/systemctl is-active docker) != "active" ]; then exit 1; fi'`

packaging/generic-deb-integrated/debian/ecs.service
mye956: Sync'd offline. We'll remove the ExecStartPre from all of the `ecs.service` files except for the one in `amazon-linux-ami-integrated`.

prateekchaudhry (approved):

yinyic (approved):

ryzr:
Any chance this is related?
https://github.com/aws/amazon-ecs-agent/issues/4270

My ecs-agent service suddenly stopped on all of my instances and doesn't attempt to restart. Not sure if possibly due to SSM automatically applying updates?

juangc-wavyssa:
Same issue here. How can we force the agent to use 1.84.0? We are having this issue for all our 49 ECS.Happened on last two sundays in all ours ECS, and we had to reboot all the agents manually,  and yesterday, only on one ECS.

danehlim:
@ryzr @juangc-wavyssa thank you for reaching out. Our team is aware of this issue and currently in the process of addressing it as part of https://github.com/aws/amazon-ecs-agent/pull/4233

danehlim:
@ryzr @juangc-wavyssa thank you for reaching out. Our team is aware of this issue and currently in the process of addressing it as part of https://github.com/aws/amazon-ecs-agent/pull/4277.

PR #4240, opened 2024-07-10
UpdateDesiredStatus for task stop verification ACK

danehlim:
### Summary

Update ECS Agent to update containers' (desired) statuses when stopping a task as part of responding to a task stop verification ACK received from ACS.

ECS Agent does not call [`UpdateDesiredStatus`](https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/api/task/task.go#L2620) when [stopping tasks in response to task stop verification ACKs sent from ACS](https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/acs/session/task_stop_verification_ack_responder.go#L35-L48). This does not align with [how tasks are stopped in response to ACS payload messages to stop a task](https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/engine/task_manager.go#L432-L433), which calls `UpdatedDesiredStatus` (which also updates the container statuses).

For supplementary information on when ACS may send a task stop verification ACK to ECS Agent, please refer to [the original pull request for full task sync feature](https://github.com/aws/amazon-ecs-agent/pull/2191).

### Implementation details

- Call `UpdateDesiredStatus` after setting desired status for the task [here](https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/acs/session/task_stop_verification_ack_responder.go#L41) (refer to Summary above)
- Remove redundant call to `AddTask`
- Add logic to save task state to Agent local state DB when setting task's status to STOPPED as part of processing task stop verification ACK from ACS

### Testing

Manually test using a custom ECS Agent built with the changes in this pull request against an internal bug repro environment. With the custom ECS Agent, bug behavior is no longer observed.

New tests cover the changes: existing tests updated (additional verification of containers' desired statuses being updated when stopping a task in response to a task stop verification ACK)

UpdateDesiredStatus for task stop verification ACK

No

### Licensing

### DIFF
--- a/agent/acs/session/task_stop_verification_ack_responder.go
+++ b/agent/acs/session/task_stop_verification_ack_responder.go
@@ -14,6 +14,7 @@
 package session
 
 import (
+	"github.com/aws/amazon-ecs-agent/agent/data"
 	"github.com/aws/amazon-ecs-agent/agent/engine"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
@@ -23,12 +24,14 @@ import (
 // taskStopper implements the TaskStopper interface defined in ecs-agent module.
 type taskStopper struct {
 	taskEngine engine.TaskEngine
+	dataClient data.Client
 }
 
 // NewTaskStopper creates a new taskStopper.
-func NewTaskStopper(taskEngine engine.TaskEngine) *taskStopper {
+func NewTaskStopper(taskEngine engine.TaskEngine, dataClient data.Client) *taskStopper {
 	return &taskStopper{
 		taskEngine: taskEngine,
+		dataClient: dataClient,
 	}
 }
 
@@ -39,7 +42,13 @@ func (ts *taskStopper) StopTask(taskARN string) {
 			loggerfield.TaskARN: task.Arn,
 		})
 		task.SetDesiredStatus(apitaskstatus.TaskStopped)
-		ts.taskEngine.AddTask(task)
+		task.UpdateDesiredStatus()
+		if err := ts.dataClient.SaveTask(task); err != nil {
+			logger.Error("Failed to save data for task", logger.Fields{
+				loggerfield.TaskARN: task.Arn,
+				loggerfield.Error:   err,
+			})
+		}
 	} else {
 		logger.Debug("Task from task stop verification ACK not found on the instance", logger.Fields{
 			loggerfield.TaskARN: taskARN,
--- a/agent/acs/session/task_manifest_responder_test.go
+++ b/agent/acs/session/task_manifest_responder_test.go
@@ -34,11 +34,14 @@ import (
 )
 
 const (
-	initialSeqNum = 11
-	nextSeqNum    = 12
-	taskARN1      = "arn1"
-	taskARN2      = "arn2"
-	taskARN3      = "arn3"
+	initialSeqNum  = 11
+	nextSeqNum     = 12
+	taskARN1       = "arn1"
+	taskARN2       = "arn2"
+	taskARN3       = "arn3"
+	containerName1 = "name1"
+	containerName2 = "name2"
+	containerName3 = "name3"
 )
 
 var expectedTaskManifestAck = &ecsacs.AckRequest{
--- a/agent/acs/session/task_stop_verification_ack_responder_test.go
+++ b/agent/acs/session/task_stop_verification_ack_responder_test.go
@@ -18,11 +18,14 @@ package session
 import (
 	"testing"
 
+	"github.com/aws/amazon-ecs-agent/agent/api/container"
 	"github.com/aws/amazon-ecs-agent/agent/api/task"
+	"github.com/aws/amazon-ecs-agent/agent/data"
 	mock_engine "github.com/aws/amazon-ecs-agent/agent/engine/mocks"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
 	acssession "github.com/aws/amazon-ecs-agent/ecs-agent/acs/session"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/session/testconst"
+	apicontainerstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/container/status"
 	apitaskstatus "github.com/aws/amazon-ecs-agent/ecs-agent/api/task/status"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/metrics"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/wsclient"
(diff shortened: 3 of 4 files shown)


agent/acs/session/task_stop_verification_ack_responder.go
```diff
@@ -39,6 +39,7 @@ func (ts *taskStopper) StopTask(taskARN string) {
 			loggerfield.TaskARN: task.Arn,
 		})
 		task.SetDesiredStatus(apitaskstatus.TaskStopped)
+		task.UpdateDesiredStatus()
 		ts.taskEngine.AddTask(task)
```
aaithal, line 43: I am wondering what is the point here of the `AddTask()` call. If you follow the code path, for a task that exists in the engine, this method is essentially a nop, especially since the task and all of its containers' desired states have already been updated to STOPPED. 

wdyt about getting rid of that (line 43)?

agent/acs/session/task_stop_verification_ack_responder.go
```diff
@@ -39,6 +39,7 @@ func (ts *taskStopper) StopTask(taskARN string) {
...
 	} else {
```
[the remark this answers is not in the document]
danehlim: Saving to disk shouldn't necessary in this case since desired status is STOPPED.

For reference, see: https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/session/payload_responder.go#L254-L256

agent/acs/session/task_stop_verification_ack_responder.go
danehlim: Agreed that L41-L42 and L43 are redundant alongside each other. 

What do you think about getting rid of L41 and L42 instead? That way, we let the task manager be responsible for stopping tasks (which is what the ACS payload message responder does) and not have this fork in logic in how tasks are stopped. L43 also ensures that task state is saved in Agent's local state DB when desired status changes.

yinyic: I agree. All the other places we call `AddTask()` are when we ingest a new task. Therefore it doesn't seem to be the right call here and might incur unintended side effects (if `AddTask() has any assumptions of the task being "new"). 

I do want to make sure that Agent is able to gracefully handle the case where the task somehow is not known. Although this should not be possible technically because Agent has to put a task into `TaskStopVerification` in the first place, for the task to show up in `TaskStopVerificationAck`, might still be a good idea to have a check/filter.

aaithal: > What do you think about getting rid of L41 and L42 instead?

`AddTask` only updates the task state if it's a task that agent hasn't seen yet (or has been removed from its state engine): https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/engine/docker_task_engine.go#L1184

So, with just line 43, I think we still have a leak. 

If we can update the desired status for the task and just drop it on the ACS channel to handle it, then I think your line of reasoning holds dane. Otherwise, I prefer 41-42 over 43 (and getting rid of it).

danehlim: > All the other places we call AddTask() are when we ingest a new task. 

Could you clarify what you mean by "new task" here? My understanding is that `AddTask` does get called for tasks that Agent knows about (e.g., Agent receives an ACS payload message to stop a task that is running).

> I do want to make sure that Agent is able to gracefully handle the case where the task somehow is not known.

Following the code path for`AddTask`, this is actually already done by the task engine, see: https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/docker_task_engine.go#L2646-L2652

danehlim: @aaithal 

> If we can update the desired status for the task and just drop it on the ACS channel to handle it, then I think your line of reasoning holds dane

Perhaps I may be misunderstanding the code, but isn't this already done? If `exists` evaluates to true in the line of code that you linked, then this line of code still gets executed:
https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/engine/docker_task_engine.go#L1206

Following that code path emits the desired status for the task on the ACS channel here: https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/engine/docker_task_engine.go#L2658-L2664

yinyic: Originally I was thinking that `AddTask()` should only be used for new task ingestion, but like you said, payload message handler calls it for all tasks, stopping or starting ([code](https://github.com/aws/amazon-ecs-agent/blob/master/agent/acs/session/payload_responder.go#L253)), which has been working fine, so that means AddTask() is able to handle a task that has an existing status of RUNNING/RUNNING, but a new status of RUNNING/STOPPED.

I'm fine with removing L41-42 in favor of L43

aaithal: Yeah, I think you're right. I missed line 1206 there. If we think `AddTask()` means that we have one single way through which agent is applying the stop logic, then I'm fine with it. Can you test it once though with the repro scenario (remove 41-42 and keep 43 around)?

prateekchaudhry: @danehlim How does [UpdateDesiredStatus](https://github.com/aws/amazon-ecs-agent/blob/0a1a88d83d7ee34c593f663979b8abd920f55481/agent/api/task/task.go#L2620), which updates the container to stopped, get called in that code path? If we go with L43 route?

danehlim: > Can you test it once though with the repro scenario (remove 41-42 and keep 43 around)?

Will do, the plan is to make sure any changes to the initial revision of this pull request are tested against internal bug repro scenario before merging this pull request. I will also update the pull request description to include that this testing is done.

danehlim: Thanks much for the discussion @aaithal and @yinyic !

danehlim: L41 is actually necessary otherwise Agent may not know that task needs to be stopped. Discussed and aligned offline with @aaithal and @yinyic that we will go with removing L43 and keeping L41-L42. In addition to this, we will add logic to save task state to local state DB when its status is updated to STOPPED, since the AddTask call (which includes saving to local state DB logic) is being removed.

agent/acs/session/task_stop_verification_ack_responder.go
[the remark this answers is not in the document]
danehlim: UPDATE: Based on above discussions and since we are aligned call to AddTask is being removed, we will add logic to save task state here as part of this pull request. See https://github.com/aws/amazon-ecs-agent/pull/4240#discussion_r1674534285 for reference.

aaithal (approved):

yinyic (approved):

PR #4251, opened 2024-07-23
Bugfix: Agent fails to resolve digest for a locally available image for images with docker.io prefix

amogh09:
### Summary

ECS Agent has a bug that prevents it from resolving image manifest digest for locally cached images if the image reference in the task payload has a `docker.io` or `docker.io/library` prefix which is optional. 

Agent resolves image manifest digest for a locally cached image by inspecting the image (equivalent of `docker image inspect <image-reference>`) and then finding a repoDigest element whose repository name matches that of the image reference. However, if the image reference has `docker.io` or `docker.io/library` prefix then Docker doesn't add the prefix to the repoDigest while Agent does not take that into account. 

The change in this PR makes Agent normalize image reference from the task payload and repoDigests when comparing them to fix this bug.

### Implementation details

* Update `GetDigestFromRepoDigests` in `utils/reference` package to parse image references and repoDigests using `ParseNormalizedNamed` function that normalizes image references in addition to parsing them. 

### Testing

Unit tests for `GetDigestFromRepoDigests` have been updated to cover the missing cases. I also ran many tasks with and without `docker.io` and `docker.io/library` prefixes on an instance with `IMAGE_PULL_BEHAVIOR` set to `once`. All tasks ran without issues.

Bugfix: Fix digest resolution for locally cached images when image reference in the task payload contains `docker.io` or `docker.io/library` prefix.

No

### Licensing

### DIFF
--- a/agent/utils/reference/reference.go
+++ b/agent/utils/reference/reference.go
@@ -42,7 +42,7 @@ func GetDigestFromImageRef(imageRef string) digest.Digest {
 // and returns the repo digest's digest.
 func GetDigestFromRepoDigests(repoDigests []string, imageRef string) (digest.Digest, error) {
 	// Parse image reference
-	ref, err := reference.Parse(imageRef)
+	ref, err := reference.ParseNormalizedNamed(imageRef)
 	if err != nil {
 		return "", fmt.Errorf("failed to parse image reference '%s': %w", imageRef, err)
 	}
@@ -55,7 +55,7 @@ func GetDigestFromRepoDigests(repoDigests []string, imageRef string) (digest.Dig
 
 	// Find a repo digest matching imageRef and return its digest
 	for _, repoDigest := range repoDigests {
-		repoDigestRef, err := reference.Parse(repoDigest)
+		repoDigestRef, err := reference.ParseNormalizedNamed(repoDigest)
 		if err != nil {
 			logger.Error("Error in parsing repo digest. Skipping it.", logger.Fields{
 				"repoDigest": repoDigest,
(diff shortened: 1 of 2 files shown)


Yiyuanzzz (approved):

agent/utils/reference/reference.go
```diff
@@ -42,20 +42,14 @@ func GetDigestFromImageRef(imageRef string) digest.Digest {
...
-	if !ok {
-		return "", fmt.Errorf(
-			"failed to parse image reference '%s' as a named reference, it was parsed as '%v'",
-			imageRef, ref)
-	}
 
 	// Find a repo digest matching imageRef and return its digest
 	for _, repoDigest := range repoDigests {
-		repoDigestRef, err := reference.Parse(repoDigest)
+		repoDigestRef, err := reference.ParseNormalizedNamed(repoDigest)
```
tinnywang, line 52: [`ParseNormalizedName`](https://pkg.go.dev/github.com/distribution/reference#ParseNormalizedNamed) returns `reference.Named`, so will the type assertion on line 60 ever be true?

tinnywang: Nevermind, I see `ParseNormalizedName` calls `Parse` calls [`getBestReferenceType`](https://github.com/distribution/reference/blob/main/reference.go#L373), which can return a canonical reference.

tinnywang (approved):

agent/utils/reference/reference.go
amogh09: Yeah `Canonical` implements `Named`.

PR #4275, opened 2024-08-08
Validate interface name passed in ECS_OFFHOST_INTROSPECTION_INTERFACE_NAME

sparrc:
### Summary

ECS_OFFHOST_INTROSPECTION_INTERFACE_NAME allows system administrators to input an interface name in the ecs.config file. This user input is then passed into an `exec` iptables command here: https://github.com/aws/amazon-ecs-agent/blob/4480f9d7d788dd9e9100df3b685d173bdd4b41c2/ecs-init/exec/iptables/iptables.go#L206-L222

ecs.config is a protected file that only the root user has write access to, so the threat is limited. But we should still validate that this is a valid linux interface name anyways, so that this env var cannot be used to inject an unexpected command into the middle of this iptables command.

### Testing

Enhancement: Validate ECS_OFFHOST_INTROSPECTION_INTERFACE_NAME is a valid interface name.

### Licensing

### DIFF
--- a/ecs-init/exec/iptables/iptables.go
+++ b/ecs-init/exec/iptables/iptables.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/aws/amazon-ecs-agent/ecs-init/exec"
 	log "github.com/cihub/seelog"
@@ -54,6 +55,8 @@ const (
 	ipv4ZeroAddrInHex                     = "00000000"
 	loopbackInterfaceName                 = "lo"
 	fallbackOffhostIntrospectionInterface = "eth0"
+
+	ifNameSize = 16
 )
 
 var (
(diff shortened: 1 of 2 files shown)


ecs-init/exec/iptables/iptables.go
```diff
@@ -213,10 +214,38 @@ func getBlockIntrospectionOffhostAccessInputChainArgs() []string {
...
+// https://github.com/torvalds/linux/blob/d7e78951a8b8b53e4d52c689d927a6887e6cfadf/net/core/dev.c#L1056-L1079
+func checkValidIfname(name string) error {
+	if name == "" {
+		return fmt.Errorf("Invalid ifname [%s]: empty string not allowed", name)
+	}
+	if len(name) >= 16 {
+		return fmt.Errorf("Invalid ifname [%s]: length of name must be less than 16 chars", name)
+	}
+	if name == "." || name == ".." {
+		return fmt.Errorf("Invalid ifname [%s]: name or '.' or '..' not allowed", name)
```
singholt, line 229: return fmt.Errorf("Invalid ifname [%s]: ~~name or~~ '.' or '..' not allowed", name)

ecs-init/exec/iptables/iptables.go
```diff
@@ -213,10 +214,38 @@ func getBlockIntrospectionOffhostAccessInputChainArgs() []string {
...
+	}
+
+	for _, char := range name {
+		if char == '/' || char == ':' || unicode.IsSpace(char) {
+			return fmt.Errorf("Invalid ifname [%s]: Space characters, ':', and '/' not allowed", name)
```
singholt, line 234: nit: ~S~space characters

ecs-init/exec/iptables/iptables.go
sparrc: done

ecs-init/exec/iptables/iptables.go
sparrc: reworded

singholt (approved):

danehlim (approved):

PR #4277, opened 2024-08-09
Adding PartOf in ECS Service's Docker Service dependency

hozkaya2000:
### Summary
Update ECS Service to be closer to pre-1.85.0 behavior while maintaining critical aspects of resiliency introduced here: https://github.com/aws/amazon-ecs-agent/pull/4233

### Implementation details
Added `PartOf=docker.service` to the following files:
`packaging/amazon-linux-ami-integrated/ecs.service`
`packaging/generic-deb-integrated/debian/ecs.service`
`packaging/generic-rpm-integrated/ecs.service`

This addition ensures that `ecs.service` stops when docker is stopped manually and does not automatically come back up. It will also ensure that restarting docker will restart `ecs.service` as it used to before 1.85.0. 

#### Setup
1. Created a custom ecs-init rpm with the additions. 
2. Then, using this rpm, created a custom AMI using an old AMI: `amzn2-ami-ecs-hvm-2.0.20230428-x86_64-ebs` as a template.
#### Behavior
Verified that behavior was changed to pre-1.85.0 behavior [without reverting resiliency]:
* Services start automatically on instance launch, able to register to cluster
* When ecs is stopped manually, it doesn't restart by itself
* When ecs is stopped manually, it doesn't restart when docker is restarted
* When docker is stopped manually, ecs is stopped in response
* When docker is started from a stopped state, ecs is not started
* When docker is restarted while running manually, ecs restarts if it was running
#### Resiliency
Simulated race condition that would cause the issue mentioned/addressed in this pull request: https://github.com/aws/amazon-ecs-agent/pull/4233. 

Simulated a docker socket failure while launching ecs with the following command on a test instance with the specified custom AMI:
`sudo systemctl start ecs & sudo systemctl restart docker.socket & `

This simulates a failure because occasionally, since these are running in parallel, `ecs.service` will fail when `docker.socket` is restarting right while ecs is checking for the docker dependency. 

The result was a success, as `ecs.service` came back up automatically. 

In comparison, this old AMI without the updated package, in the same failure scenario, was not able to bring `ecs.service` back up:

#### Load Testing

Launched 15000 instances with the specified custom AMI and verified that all 15000 of them were able to register to ECS. 

Revert ecs.service behavior while maintaining resiliency

### Licensing


danehlim (approved):

singholt:
Please update this commit's message: https://github.com/aws/amazon-ecs-agent/pull/4277/commits/e7c5fc39be8d5505b3623e3e8d2b8a615acc6e2b. It makes one think you are replacing `Wants` with `PartOf`, which isn't the case.

And don't forget to squash your commits before merging, thanks!

singholt (approved):

danehlim (approved):

PR #4278, opened 2024-08-10
Refactor engine integration test functions to be usable by other packages

tinnywang:
### Summary
Refactor helper functions for integration tests in the `engine` package so that they can be used by integration tests in other packages.

This PR does not add new integration tests demonstrating how the helper functions will be used outside the `engine` package. They will added in a separate PR.

### Implementation details
- https://github.com/aws/amazon-ecs-agent/pull/4278/commits/3d91b4907bd7fe95ab81ef8559fb15ebf052d4a9 exports helper functions in the `engine` package.
- https://github.com/aws/amazon-ecs-agent/pull/4278/commits/52c1d6f70783e6d9cec298a54e34578f00649b04 renames `common*_test.go` files, where the helper functions are defined, to `common*_testutil.go`. This is necessary because `*_test.go` files are only compiled when their packages are being `go test`ed, so symbols in `*_test.go` files aren't visible to other packages.
- https://github.com/aws/amazon-ecs-agent/pull/4278/commits/2ed99db16046c3a4ed1c57e6c48a0a8aaaa68956 updates helper function usage in Linux integration test files (capitalizing function names now that they are exported).
- https://github.com/aws/amazon-ecs-agent/pull/4278/commits/60180c12d892df4800735f63d4222d2cd82dc6ea changes the [`Setup`](https://github.com/aws/amazon-ecs-agent/blob/master/agent/engine/common_integ_test.go#L229) function to return the [`dockerapi.DockerClient`](https://github.com/aws/amazon-ecs-agent/blob/c41d8e1b4abad6ece0093bc7d2b42fe5a21c46f8/agent/engine/common_integ_test.go#L236) used by the `TaskEngine`. This allows tests to access the `TaskEngine`'s underlying Docker client without having to do
  ```
  dockerClient := taskEngine.(*DockerTaskEngine).client
  ```
  which is not possible for integration tests outside the `engine` package.
- https://github.com/aws/amazon-ecs-agent/pull/4278/commits/41f50089779c45b8ab740ffdc07512f070faa45a updates helper function usage in Windows integration test files (capitalizing function names now that they are exported).
- https://github.com/aws/amazon-ecs-agent/commit/2e3a967a249b7d864002d05d2b0bae173a450016 renames some helper functions to make it clear they're meant to be used in tests.

### Testing
All GitHub checks are passing.

Refactor engine integration test functions to be usable by other packages

### Licensing


singholt (approved):

agent/engine/common_integ_testutil.go
```diff
@@ -55,14 +55,14 @@ func init() {
 	sdkClientFactory = sdkclientfactory.NewFactory(context.TODO(), dockerEndpoint)
 }
 
-func defaultTestConfigIntegTest() *config.Config {
+func DefaultTestConfigIntegTest() *config.Config {
```
amogh09, line 58: I was wondering if we should put the test utils in a separate `engine_test` package to keep them separate from non-test code in `engine` package? WDYT?

amogh09: For example, `engine.Setup` function doesn't seem like a test function.

tinnywang: Unfortunately, this introduces an import cycle. The helper functions use `engine.TaskEngine`, so if they're moved into a separate `engine_test` package, `engine_test` will import `engine`. But `engine` needs to import `engine_test` to access the helper functions in its integration tests.

The import cycle could be broken by moving all the tests out of `engine` and into `engine_test` (`engine` will stop importing `engine_test`), but the tests access a lot of unexported symbols from the `engine` package.

I've renamed some of the ambiguous helper functions to make it clear they're for testing in https://github.com/aws/amazon-ecs-agent/pull/4278/commits/2e3a967a249b7d864002d05d2b0bae173a450016.

amogh09: Thanks!

amogh09 (approved):

singholt (approved):

PR #4286, opened 2024-08-15
Container transition to MANIFEST_PULLED state should happen after network pause container reaches RESOURCES_PROVISIONED state

amogh09:
### Summary

Make awsvpc task containers' transition to MANIFEST_PULLED state depend on the network pause container reaching RESOURCES_PROVISIONED (RUNNING state for the pause container) state. This is because transition to MANIFEST_PULLED state may involve an image repository call which needs to be routed through the network pause container for some setups. 

Also update existing task engine tests for awsvpc tasks to ensure that for an awsvpc task, any `DockerClient` method calls that may call an image repository, happen after the network pause container has reached RESOURCES_PROVISIONED state. This should prevent similar bugs from happening in the future.

### Testing

Updated unit tests. 

Ran an `awsvpc` network mode task and observed in Agent logs that the pause container reaches RESOURCES_PROVISIONED state before the task container transitions to MANIFEST_PULLED state.

Bugfix: AWSVPC task containers' transition to MANIFEST_PULLED state should depend on the network pause container reaching RESOURCES_PROVISIONED state.

### Licensing


Yiyuanzzz (approved):

agent/engine/docker_task_engine_linux_test.go
```diff
@@ -58,7 +58,6 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/appmesh"
 	ni "github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/networkinterface"
-
```
singholt, line 61: nit: this newline was intentionally added, to distinguish between internal and external 3rd party imports.

singholt (approved):

agent/engine/docker_task_engine_linux_test.go
amogh09: Ah right. Fixed it.

singholt (approved):

Yiyuanzzz (approved):

PR #4304, opened 2024-08-20
Fix Windows integration tests

tinnywang:
### Summary
Be not deceived by the falsehood of GitHub status checks. Windows integration tests aren't actually running and passing. The [PowerShell script](https://github.com/aws/amazon-ecs-agent/blob/dev/scripts/run-integ-tests.ps) uses Bash syntax to set the `CGO_ENABLED` environment variable, which PowerShell doesn't understand, but the script doesn't fail fast because it doesn't have the PowerShell equivalent of `set -e`. As a result, the tests never run but the script returns successfully.
https://github.com/aws/amazon-ecs-agent/blob/53aebd9fe52695d4ad8a3cd8d7007cfdeee164b3/scripts/run-integ-tests.ps1#L62

### Implementation details
- [292c4d3](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/292c4d33fa48acd5133c0cd2b89d5c0dd4db2c40) sets [`$ErrorActionPreference = 'Stop'`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.4#erroractionpreference), the PowerShell equivalent of `set -e`, to make the script fail fast.
- [4c3d855](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/4c3d8556fe26ca8c10ebb3eadd937e9aafa79ba3) updates the `go test` command to use PowerShell syntax.
- [626361f](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/626361fbe4368f174a87150235a9adf2f17e8601) fixes compilation errors in the `agent` package (undefined symbol usage).
- [d742ef1](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/d742ef16fb5c228841366d9bf482c7a57d9118db) fixes compilation errors in the `ecs-agent` package (types not implementing interfaces).
- [e0a70d9](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/e0a70d9153b1dcfbba5742dbe3edcca8990b3784) makes [`TestImageManifestPullInteg`](https://github.com/aws/amazon-ecs-agent/blob/78a2bf0c7d3ebd3a13de3ac733af46dfb3816b18/agent/dockerclient/dockerapi/docker_client_integ_test.go) only run for Linux. We can't test image manifest pulls on Windows because we don't run a local registry, so images are always cached on the machine running the tests and never get pulled.
- [8040c20](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/8040c20a3bb16aa5909858a16fd7a06639e290ca) skips task/container manifest pulled assertions if the tests are running on Windows and moves manifest pull tests into Unix integration test file. We can't test image manifest pulls on Windows because we don't run a local registry, so images are always cached on the machine running the tests and never get pulled.
- [62438de](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/62438deb1f602bb7aabaae7a5e15cd1a2bc7e32e) increases timeouts on tests that were flaking on Windows Server 2019.
- [a2894c6](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/a2894c660a97cb5b5102f8b962ed082fc19e2847) removes the retry in the [PowerShell script](https://github.com/aws/amazon-ecs-agent/blob/dev/scripts/run-integ-tests.ps) because it will always fail. The [`TestIntegImageCleanup*`](https://github.com/aws/amazon-ecs-agent/blob/dev/agent/engine/docker_image_manager_integ_test.go) tests assert that images exist, remove the images, and then assert that the images do not exist. So the images are deleted after the first run and don't exist when the second run starts and expects them to exist. For successful retires, the entire script needs to be rerun so that images can be rebuilt. Note that the [linux Make target](https://github.com/aws/amazon-ecs-agent/blob/dev/Makefile#L176-L178) doesn't retry the `go test` command.

### Testing
- Ran `.\scripts\run-integ-tests.ps1` on Windows Server 2016, 2019, and 2022 and verified that the tests ran and passed.
- Verified that the [PowerShell script](https://github.com/aws/amazon-ecs-agent/blob/dev/scripts/run-integ-tests.ps) returned a non-zero exit code if any command failed.
- GitHub status checks passed after actually running the tests.

### Licensing


scripts/run-integ-tests.ps1
```diff
@@ -59,10 +54,7 @@ Invoke-Expression "${PSScriptRoot}\..\misc\exec-command-agent-test\build.ps1"
 # Run the tests
 $cwd = (pwd).Path
 try {
-  $env:ECS_LOGLEVEL = 'debug'; CGO_ENABLED=1 go test -race -tags integration -timeout=40m -v ../agent/engine ../agent/stats ../agent/app
-  if (${LastExitCode} -ne 0) {
-    $env:ECS_LOGLEVEL = 'debug'; CGO_ENABLED=1 go test -race -tags integration -timeout=40m -v ../agent/engine ../agent/stats ../agent/app
-  }
+  $env:GO111MODULE = 'auto'; $env:CGO_ENABLED = 1; $env:ECS_LOGLEVEL = 'debug'; go test -race -tags integration -timeout=40m -v ./agent/... ./ecs-agent/...
```
amogh09, line 57: Why do we need `CGO_ENABLED = 1`?

agent/engine/engine_windows_integ_test.go
```diff
@@ -279,64 +253,6 @@ func TestLocalHostVolumeMount(t *testing.T) {
 	assert.Equal(t, "empty-data-volume", strings.TrimSpace(string(data)), "Incorrect file contents")
 }
 
-// TestStartStopUnpulledImage ensures that an unpulled image is successfully
-// pulled, run, and stopped via docker.
-func TestStartStopUnpulledImage(t *testing.T) {
```
amogh09, line 284: Why are we removing this test?

agent/engine/engine_unix_integ_test.go
```diff
@@ -296,6 +297,390 @@ func TestStartStopUnpulledImageDigest(t *testing.T) {
...
+
+	for _, behavior := range imagePullBehaviors {
+		t.Run(fmt.Sprintf("%v", behavior), func(t *testing.T) {
+			cfg := DefaultTestConfigIntegTest()
+			cfg.ImagePullBehavior = behavior
+			cfg.DockerStopTimeout = 100 * time.Millisecond
+			taskEngine, done, _, _ := SetupIntegTestTaskEngine(cfg, nil, t)
+			defer done()
+
+			first := createTestContainerWithImageAndName(testRegistryImage, "first")
```
amogh09, line 316: If we use a Windows image from public ECR, is it trivial to support this test on Windows?

agent/engine/common_integ_testutil.go
```diff
@@ -355,6 +372,10 @@ func VerifyTaskStatus(status apitaskstatus.TaskStatus, taskARN string, testEvent
 // This method queries the TestEvents struct to check a Task Status.
 // This method will block if there are no more stateChangeEvents from the DockerTaskEngine but is expected
 func VerifyContainerStatus(status apicontainerstatus.ContainerStatus, ARNcontName string, testEvents *TestEvents, t *testing.T) {
+	if runtime.GOOS == "windows" && status == apicontainerstatus.ContainerManifestPulled {
+		t.Log("Not expecting image manifest pulls on Windows")
```
amogh09, line 376: Let's also add why this won't happen on Windows. That is, because our integration test setup on Windows does not involve image pulls and uses cached images instead.

agent/engine/engine_unix_integ_test.go
tinnywang: I've changed `testRegistryImage` to point the Windows Server image from MCR in https://github.com/aws/amazon-ecs-agent/pull/4304/commits/725e2b785003dc099eea5e25cc9fa21b9e10c9f3. (We're already using it in `TestStartStopUnpulledImage` and `TestStartStopUnpulledImageDigest`.)

tinnywang: I've changed `testRegistryImage` to point the Windows Server image from MCR in https://github.com/aws/amazon-ecs-agent/pull/4304/commits/725e2b785003dc099eea5e25cc9fa21b9e10c9f3. (We're already using it in `TestStartStopUnpulledImage` and `TestStartStopUnpulledImageDigest`.)

scripts/run-integ-tests.ps1
```diff
@@ -37,15 +39,8 @@ if (-Not ($dockerImages -like "*$BaseImageName*")) {
...
 Invoke-Expression "docker tag $BaseImageName amazon-ecs-ftest-windows-base:make"
 
-$imgDigest = Invoke-Expression "docker image inspect --format `"{{.RepoDigests}}`" $BaseImageName"
-# The template returns an array so its [$BaseImageDigest]. Remove the '[' and ']'
-$imgDigest = $imgDigest.SubString(1, $imgDigest.Length - 2)
-
-$ProgramFiles="C:\Program Files\Amazon\ECS"
-$env:BASE_IMAGE_NAME=$BaseImageName
-$env:BASE_IMAGE_NAME_WITH_DIGEST=$imgDigest
-
```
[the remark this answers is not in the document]
tinnywang: Added it back in [4e02991](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/4e02991541dc094511ad6cd14c402a3523cfc194) and [725e2b7](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/725e2b785003dc099eea5e25cc9fa21b9e10c9f3). I initially removed the test because it was failing, and I thought that was because it was related to image pulls. A closer inspection revealed the failures were caused by missing state change assertions.

agent/engine/engine_windows_integ_test.go
tinnywang: Added it back in [4e02991](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/4e02991541dc094511ad6cd14c402a3523cfc194) and [725e2b7](https://github.com/aws/amazon-ecs-agent/pull/4304/commits/725e2b785003dc099eea5e25cc9fa21b9e10c9f3). I initially removed the test because it was failing, and I thought that was because it was related to image pulls. A closer inspection revealed the failures were caused by missing state change assertions.

scripts/run-integ-tests.ps1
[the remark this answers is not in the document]
tinnywang: It was unused after I removed the Windows tests that used the env var.

agent/engine/engine_unix_integ_test.go
amogh09: Thanks! So do we still need to move `TestManifestPulledDoesNotDependOnContainerOrdering` and other tests to this unix specific test file?

amogh09 (approved):

agent/dockerclient/dockerapi/docker_client_integ_test.go
```diff
@@ -1,5 +1,5 @@
-//go:build integration
-// +build integration
+//go:build !windows && integration
```
amogh09, line 1: I think we should rename this file to have `unix` in its name to follow the convention.

agent/engine/common_windows_integ_testutil.go
```diff
@@ -0,0 +1,43 @@
...
+	return createTestContainerWithImageAndName(testBaseImage, "windows")
+}
+
+// GetLongRunningCommand returns the command that keeps the container running for the container
+// that uses the default integ test image (amazon-ecs-ftest-windows-base:make for windows)
+func GetLongRunningCommand() []string {
+	return []string{"ping", "-t", "localhost"}
+}
+
+// TODO: implement this
```
oldschool-engineer, line 48: Please implement this method.

mcregan23 (approved):

mcregan23 (changes requested):
+1 to what Tom said. Might as well implement it now so it doesn't get forgotten about

amogh09 (approved):

singholt (approved):

agent/engine/common_windows_integ_testutil.go
```diff
@@ -45,7 +45,7 @@ func GetLongRunningCommand() []string {
 	return []string{"ping", "-t", "localhost"}
 }
 
-// TODO: implement this
 func isDockerRunning() bool {
-	return true
+	_, err := os.Stat("//./pipe/docker_engine")
```
oldschool-engineer, line 49: I don't think the os.Stat implementation will work.  In my testing it always returned the same result regardless of whether docker is running or not.
Here's an alternative implementation that does work.

```
package main

import (
    "flag"
    "golang.org/x/sys/windows/svc/mgr"
    "log"
)

func isDockerRunning() bool {
    var svcName = "docker"
    flag.StringVar(&svcName, "name", svcName, "name of the service")
    flag.Parse()
    svcMgmt, err := mgr.Connect()
    if err != nil {
        log.Fatalf("Error connecting to service manager: %v", err)
    }

    dockerSvc, err := svcMgmt.OpenService(svcName)
    if err != nil {
        log.Fatalf("Error opening service %q: %v", svcName, err)
    }

    dockerStatus, err := dockerSvc.Query()
    if err != nil {
        log.Fatalf("Error querying docker service: %v", err)
    }
    log.Printf("State:%v, ProcessId:%v, Win32ExitCode: %v", dockerStatus.State, dockerStatus.ProcessId, dockerStatus.Win32ExitCode)
    return dockerStatus.Win32ExitCode == 0
}

func main() {
    dockerStatus := isDockerRunning()
    if dockerStatus != true {
        log.Fatalf("Docker is not running")
    }
    log.Printf("Docker is running")
}
```
```
2024/08/30 13:55:25 State:4, ProcessId:4500, Win32ExitCode: 0
2024/08/30 13:55:25 Docker is running
```
```
2024/08/30 13:55:04 State:1, ProcessId:0, Win32ExitCode: 1067
2024/08/30 13:55:04 Docker is not running
```

tinnywang: I have minimal knowledge/experience with Windows development, so bear with me while I explain my testing process...

To exercise the `os.Stat` implementation of `isDockerRunning`, I added a test
```
func TestIsDockerRunning(t *testing.T) {
	assert.True(t, isDockerRunning())
}
```
which I ran with
```
go test -tags integration  -count 1 -v ./agent/engine/... -run "TestIsDockerRunning"
```
I ran `get-service docker` to check if docker was already running. It was, and the test passed.
I ran `stop-service docker` to stop docker. The test failed when I reran it, as was expected.
I ran `start-service docker` to restart docker. The test passed.

Am I not starting/stopping docker correctly? I got the same results when I swapped out the `os.Stat` implementation with the service manager implementation.

oldschool-engineer: Based on your test results, it seems that Go is handling it properly if run as part of the integration test framework.  Given your test results, I'm OK with your earlier implementation since it is very simple and it works.  (Plus, the implementation I provided should be modified to include deferred closures to the service manager and service instance handle.)

My concern was whether Go's `os.Stat` method can differentiate between named pipes and regular on-disk files.  Named pipes are accessed via file system drivers (NPFS), but they are are not part of the actual on-disk file system, which is (typically) NTFS.

Running the code snippet, as-is, on Windows in a stand-alone way doesn't seem to work right, however.  That said, connecting through service manager requires the unit tests to be run as an administrator.  


---

> Am I not starting/stopping docker correctly?

There is more than one way to start/stop services on Windows.  start-service docker and stop-service docker is the PowerShell way to start/stop docker and is perfectly acceptable.

oldschool-engineer (approved):

PR #4305, opened 2024-08-20
Add a field 'ENIID' for task network configuration

xxx0624:
### Summary

This PR is to add a new field `ENIID` in the NetworkInterface of the task network config which can be used in shared handler for network faults. In the handler, we can have the read-write lock per ENI ID which is unique across tasks to avoid multiple clients/requests from manipulating the same resource at the same time.

### Implementation details

Add the new field in `ecs-agent/tmds/handlers/v4/state/response.go`

### Testing

New tests cover the changes: 

Yes. 

Feature - add an field in NetworkInterface object for internal use only.

### Additional Information

### Licensing


tshan2001 (approved):
LGTM

amogh09 (approved):

PR #4309, opened 2024-08-26
Add read/write lock in the fault handler

xxx0624:
### Summary

This PR is to add a field / RWMutex in the fault handler which can help to avoid race condition where multiple clients would manipulate the same network resource at the same time.

### Implementation details

- Add the read/write lock in the [FaultHandler](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go#L41C6-L41C18)
- Add a new method to create the [handler](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go#L41C6-L41C18) for faults
- Add more validations in [validateTaskNetworkConfig](https://github.com/aws/amazon-ecs-agent/blob/5589a96639d216db755ddde6fb45764376889a84/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go#L704-L712) to make sure the required field has non-empty value.

### Testing

New tests cover the changes: 
No. Existing unit test will cover this.

Feature: Add RWMutex in the fault handler.

### Additional Information

### Licensing

### DIFF
--- a/agent/handlers/v4/tmdsstate.go
+++ b/agent/handlers/v4/tmdsstate.go
@@ -161,7 +161,6 @@ func (s *TMDSAgentState) getTaskMetadata(v3EndpointID string, includeTags bool) 
 					Path: task.GetNetworkNamespace(),
 					NetworkInterfaces: []*tmdsv4.NetworkInterface{
 						{
-							ENIID:      "eni-fake-id",
 							DeviceName: "ethx",
 						},
 					},
--- a/agent/handlers/v4/tmdsstate.go
+++ b/agent/handlers/v4/tmdsstate.go
@@ -161,7 +161,8 @@ func (s *TMDSAgentState) getTaskMetadata(v3EndpointID string, includeTags bool) 
 					Path: task.GetNetworkNamespace(),
 					NetworkInterfaces: []*tmdsv4.NetworkInterface{
 						{
-							DeviceName: "",
+							ENIID:      "eni-fake-id",
+							DeviceName: "ethx",
 						},
 					},
 				},
--- a/agent/handlers/v4/tmdsstate.go
+++ b/agent/handlers/v4/tmdsstate.go
@@ -161,7 +161,12 @@ func (s *TMDSAgentState) getTaskMetadata(v3EndpointID string, includeTags bool) 
 					Path: task.GetNetworkNamespace(),
 					NetworkInterfaces: []*tmdsv4.NetworkInterface{
 						{
-							DeviceName: "ethx",
+							// TODO: fetch the correct device name.
+							// We are exposing this information via AgentState to facilitate the fault injection
+							// handler to start/stop/check network faults.
+							// Use 'eth0'(a fake value) for existing fault injection related unit tests for now and
+							// it will be updated in the future.
+							DeviceName: "eth0",
 						},
 					},
 				},
--- a/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
+++ b/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"sync"
 
 	"github.com/aws/amazon-ecs-agent/ecs-agent/api/ecs/model/ecs"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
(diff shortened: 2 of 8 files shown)


ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -264,6 +297,84 @@ func generateNetworkBlackHolePortTestCases(name string, expectedHappyResponseBod
...
+				fmt.Sprintf("failed to get task metadata due to internal server error for container: %s", endpointId)),
+			setAgentStateExpectations: func(agentState *mock_state.MockAgentState) {
+				agentState.EXPECT().GetTaskMetadata(endpointId).Return(state.TaskResponse{
+					TaskResponse:          &v2.TaskResponse{TaskARN: taskARN},
+					FaultInjectionEnabled: true,
+					TaskNetworkConfig: &state.TaskNetworkConfig{
+						NetworkMode:       awsvpcNetworkMode,
+						NetworkNamespaces: nil,
+					},
+				}, nil)
```
mye956, line 314: nit (non-blocking): might be a good idea to add a `Times(1)` for these test cases as well if want to make things consisent. wdyt?

mye956 (approved):

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
xxx0624: Sounds good. I can add it in a follow up change if no major concern in this PR.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -206,12 +238,21 @@ func (h *FaultHandler) StartNetworkLatency() func(http.ResponseWriter, *http.Req
...
 		if err != nil {
 			return
 		}
 
+		// the network latency fault would be injected to given elastic network
+		// interface/ENI.
+		eniID := taskMetadata.TaskNetworkConfig.
+			NetworkNamespaces[0].
+			NetworkInterfaces[0].
+			ENIID
```
amogh09, line 251: Why not use netns path here also?

xxx0624: Updated.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -42,18 +43,36 @@ const (
 )
 
 type FaultHandler struct {
-	// TODO: Mutex will be used in a future PR
-	// mu             sync.Mutex
+	// mutexMap is used to avoid multiple clients to manipulate same resource at same
+	// time. The 'key' is the ENI ID or the network namespace path and 'value' is the
```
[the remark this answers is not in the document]
xxx0624: Updated.

mye956 (approved):

agent/handlers/v4/tmdsstate.go
```diff
@@ -161,7 +161,7 @@ func (s *TMDSAgentState) getTaskMetadata(v3EndpointID string, includeTags bool)
 					Path: task.GetNetworkNamespace(),
 					NetworkInterfaces: []*tmdsv4.NetworkInterface{
 						{
-							DeviceName: "",
+							DeviceName: "ethx",
```
amogh09, line 164: Why this change?

amogh09: Thanks!

xxx0624: Some context:
The `DeviceName` is required for network latency / network packet loss faults. We are exposing this information (and other task network config) via AgentState to facilitate the fault injection handler to start/stop/check network fault.

Why we need a non-empty value here which is fake: because in this PR, we introduce a new check ([link](https://github.com/aws/amazon-ecs-agent/blob/02cb57659a882844033d0c18c7e666609347fb61/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go#L688-L691)) to see if the deviceName is empty. So if that's empty, these [unit test](https://github.com/aws/amazon-ecs-agent/blob/78a2bf0c7d3ebd3a13de3ac733af46dfb3816b18/agent/handlers/task_server_setup_test.go#L3695-L3708) will fail.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -416,11 +473,17 @@ func (h *FaultHandler) CheckNetworkPacketLoss() func(http.ResponseWriter, *http.
 
 		// Obtain the task metadata via the endpoint container ID
 		// TODO: Will be using the returned task metadata in a future PR
-		_, err = validateTaskMetadata(w, h.AgentState, requestType, r)
+		taskMetadata, err := validateTaskMetadata(w, h.AgentState, requestType, r)
```
tshan2001, line 476: Do we need this in all check methods? Since we're only calling `tc q` and check if `"kind":"netem"` exists, we don't need to know the network namespace path

xxx0624: for awsvpc mode, all commands should be running in the given task network namespace. So we also need this check here.

amogh09 (approved):

tshan2001 (approved):

mye956 (approved):

PR #4310, opened 2024-08-26
Add shared handler for network packet loss fault

tshan2001:
### Summary
Add shared handler for network packet loss fault

### Implementation details
* Add NetworkPacketLossRequest type and corresponding validation    
* Add Start, Stop, and Check for Network Packet Loss
* Added unit tests for the methods added

### Testing
New tests cover the changes:

`StartNetworkPacketLoss()`, `StopNetworkPacketLoss()`, `CheckNetworkPacketLoss()`

Feature: Add network packet loss handler

### Additional Information

### Licensing


ecs-agent/tmds/handlers/fault/v1/types/types.go
```diff
@@ -102,29 +102,39 @@ func (request NetworkLatencyRequest) ValidateRequest() error {
...
+type NetworkPacketLossRequest struct {
+	LossPercent *uint64 `json:"LossPercent"`
+	// Sources is a list including IPv4 addresses or IPv4 CIDR blocks.
+	Sources []*string `json:"Sources"`
 }
 
-func (request NetworkLatencyRequest) ToString() string {
+// ValidateRequest validates required fields are present and its value.
+func (request NetworkPacketLossRequest) ValidateRequest() error {
+	if request.LossPercent == nil {
```
xxx0624, line 125: Can we check its value as well? this should be percentage - https://docs.aws.amazon.com/fis/latest/userguide/fis-actions-reference.html#task-network-packet-loss

tshan2001: Thanks for the call out. I think we can assert the following: 0 <= request.LossPercent <= 100, which validates that this is a percentage (without the percentage sign)

xxx0624: it should be 1 for the minimum value

tshan2001: ACK. Just updated the PR with new revision. Added check to ensure value is within [1, 100], and added 3 unhappy test cases: when the value is -1 (fails with unable to store -1 in uint64), when the value is 101, and when the value is 0 (both fails with invalid val)

mye956 (approved):

xxx0624 (approved):

PR #4314, opened 2024-08-29
Adding os/exec wrapper for testability

tshan2001:
### Summary
Adding a wrapper for `os/exec` so that we can add mock for the fault injection features that use linux `tc`

### Implementation details
A new interface `Cmd` is added, which wraps around several features of `os/exec`

### Testing
`go generate` was ran to create the mock file and passed.

Adding wrapper and mock for `os/exec`

### Additional Information

### Licensing


ecs-agent/utils/execwrapper/exec.go
```diff
@@ -0,0 +1,91 @@
...
+	Wait() error
+	KillProcess() error
+	AppendExtraFiles(...*os.File)
+	Args() []string
+	SetIOStreams(io.Reader, io.Writer, io.Writer)
+}
+
+type cmdWrapper struct {
+	*exec.Cmd
+}
```
amogh09, line 45: Why do we need to create this new `struct`? Shouldn't `exec.Cmd` already implement `Cmd` interface?

tshan2001: The `cmdWrapper` interface is not exactly the same, we also added a couple helper functions to help with common used functionalities such as `SetIOStreams()`. Without having this extra struct, we won't be able to make the helper functions non-static (we would get error `cannot define new methods on non-local type exec.Cmd`)

ecs-agent/utils/execwrapper/exec.go
```diff
@@ -0,0 +1,91 @@
+package execwrapper
```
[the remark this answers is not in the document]
mye956: +1, ditto in the mocks as well ([ref](https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/api/attachment/resource/generate_mocks.go#L14))

xxx0624 (dismissed):

xxx0624 (approved):

mye956 (approved):

PR #4322, opened 2024-09-05
Add check network packet loss implementation

tshan2001:
### Summary
Adding implementation for `CheckNetworkPacketLoss()` and corresponding unit tests. Also updating the `os/exec` wrapper to include 2 more helper methods to facilitate mocks in unit testing.

### Implementation details
After the TMDS server receives the request to check for network packet loss, the following Linux command will be executed:
```
tc -j q
```
The output format will be a byte array of a json string. The output will then be unmarshalled and we will check whether the following exists:
```
{
"kind":"netem"
...
}
```

### Testing
Unit tests for the TMDS package was run.
```
 % go test -tags unit -v -run TestCheckNetworkPacketLoss /workplace/tianzes/amazon-ecs-agent/ecs-agent/tmds/handlers/fault/v1/handlers
...
--- PASS: TestCheckNetworkPacketLoss (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_success-running (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_success-not-running (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_success-not-running-but-latency-is-running (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_only-one-ip-exists-not-running (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_unknown_request_body (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_failed_to_unmarshal_json (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_malformed_request_body_1 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_malformed_request_body_2 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_incomplete_request_body_1 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_incomplete_request_body_2 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_incomplete_request_body_3 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_invalid_LossPercent_in_the_request_body_1 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_invalid_LossPercent_in_the_request_body_2 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_invalid_LossPercent_in_the_request_body_3 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_invalid_IP_value_in_the_request_body_1 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_invalid_IP_CIDR_block_value_in_the_request_body_2 (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_task_lookup_fail (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_task_metadata_fetch_fail (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_task_metadata_unknown_fail (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_fault_injection_disabled (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_invalid_network_mode (0.00s)
    --- PASS: TestCheckNetworkPacketLoss/check_network_packet_loss_empty_task_network_config (0.00s)
```

New tests cover the changes: 
2 unhappy test cases were added:
1. Call was successful but there's no fault running
2. Call was successful but got internal error when processing the Linux command
3. Call was successful, but there's network latency fault running instead of network packet loss fault

### Manual Testing
Launched a Fargate Instance with the changes. Launched a task with ecs-exec enabled.
```
# First curl the TMDS endpoint from the task without any fault on the instance
sh-5.2# curl -X GET ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":10, "Sources":["192.168.0.1"]}'
{"Status":"not-running"}

# Log into the instance as 'su', manually inject a fault with 10% loss with IP '192.168.0.1' associated
root # nsenter --net=/var/run/netns/8aacba9e7ef74030a9310de01f3ca80e-028c5a03d94b tc qdisc add dev eth1 root handle 1: prio priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2
root # nsenter --net=/var/run/netns/8aacba9e7ef74030a9310de01f3ca80e-028c5a03d94b tc qdisc add dev eth1 parent 1:1 handle 10: netem loss 10%
root # nsenter --net=/var/run/netns/8aacba9e7ef74030a9310de01f3ca80e-028c5a03d94b tc filter add dev eth1 protocol ip parent 1:0 prio 1 u32 match ip dst 192.168.0.1 flowid 1:1

# Curl the TMDS endpoint with the same config, result should be running
sh-5.2# curl -X GET ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":10, "Sources":["192.168.0.1"]}'
{"Status":"running"}

# Now change the lossPercent in payload, result should be not-running
sh-5.2# curl -X GET ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":6, "Sources":["192.168.0.1"]}'
{"Status":"not-running"}

# Now add an additional IP address 10.1.1.1, and curl the endpoint with both IPs, result should be running
root # nsenter --net=/var/run/netns/8aacba9e7ef74030a9310de01f3ca80e-028c5a03d94b tc filter add dev eth1 protocol ip parent 1:0 prio 1 u32 match ip dst 10.1.1.1 flowid 1:1
sh-5.2# curl -X GET ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":10, "Sources":["192.168.0.1", "10.1.1.1"]}'
{"Status":"running"}

# If we only curl one of the endpoint, result should be running as well:
sh-5.2# curl -X GET ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":10, "Sources":["192.168.0.1"]}'
{"Status":"running"}

# Lastly, if we add an unfiltered ip in the payload, result should be not-running
sh-5.2# curl -X GET ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":10, "Sources":["192.168.0.1", "10.1.1.1", "1.1.1.1"]}'
{"Status":"not-running"}

# Edge case: manually add a network latency fault, and curl the check packet loss endpoint. The result should be not running
root # nsenter --net=/var/run/netns/8aacba9e7ef74030a9310de01f3ca80e-028c5a03d94b tc q show dev eth1
qdisc prio 1: root refcnt 9 bands 3 priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2
qdisc netem 10: parent 1:1 limit 1000 delay 100ms
sh-5.2# curl -X GET ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":10, "Sources":["192.168.0.1", "10.1.1.1", "1.1.1.1"]}'
{"Status":"not-running"}

```

Add check network packet loss implementation 

### Additional Information

### Licensing


ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -692,3 +721,132 @@ func validateTaskNetworkConfig(taskNetworkConfig *state.TaskNetworkConfig) error
...
+
+	interfaceName := taskMetadata.TaskNetworkConfig.NetworkNamespaces[0].NetworkInterfaces[0].DeviceName
+	lossPercent := request.LossPercent
+	networkMode := taskMetadata.TaskNetworkConfig.NetworkMode
+	containers := taskMetadata.Containers
+	ipSources := request.Sources
+	// If task's network mode is awsvpc, we need to run nsenter to access the task's network namespace.
+	nsenterPrefix := ""
+	if networkMode == "awsvpc" {
+		nsenterPrefix = fmt.Sprintf(nsenterCommandString, containers[0].DockerName)
```
xxx0624, line 738: IIRC we can get the full path from task metadata.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -692,3 +721,132 @@ func validateTaskNetworkConfig(taskNetworkConfig *state.TaskNetworkConfig) error
 
 	return nil
 }
+
+// checkPacketLossFault checks if there's existing network-packet-loss fault running.
+func (h *FaultHandler) checkPacketLossFault(taskMetadata *state.TaskResponse, request types.NetworkPacketLossRequest) (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
```
xxx0624, line 728: can we do one more thing that when the it timed out, we should cancel the command execution?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: Yes, the full path will be read from the metadata. This is to concatenate the `nsenter --net=` prefix

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: We use `CommandContext()` from `os/exec`, so the command execution will be cancelled when it times out.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -42,20 +46,27 @@ const (
...
+)
+
 type FaultHandler struct {
 	// mutexMap is used to avoid multiple clients to manipulate same resource at same
 	// time. The 'key' is the the network namespace path and 'value' is the RWMutex.
 	// Using concurrent map here because the handler is shared by all requests.
 	mutexMap       sync.Map
 	AgentState     state.AgentState
 	MetricsFactory metrics.EntryFactory
+	OsExecWrapper  execwrapper.Exec
```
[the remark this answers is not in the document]
tshan2001: Yes, will switch to lowercase and add this as a part of FaultHandler's `New()`

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -692,3 +722,139 @@ func validateTaskNetworkConfig(taskNetworkConfig *state.TaskNetworkConfig) error
...
+		return false, nil
+	}
+
+	return true, nil
+}
+
+// runExecCommand wraps around the execwrapper, providing a convenient way of running any Linux command
+// and getting the result in both stdout and stderr.
+func (h *FaultHandler) runExecCommand(ctx context.Context, linuxCommandString string) ([]byte, error) {
+	cmdExec := h.osExecWrapper.CommandContext(ctx, "/bin/sh", "-c", linuxCommandString)
```
mye956, line 818: Not quite sure if the `/bin/sh` path exists within the agent container. Will need to double check this. Is it possible to make the call without this just like with the standard `os/exec` go library? (e.g. `exec.Command("ls -al")`)

tshan2001: Yes. I just pushed a new revision that removes "/bin/sh" from the command. Also re-ran all the manual testing to make sure that it still works.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -692,3 +722,140 @@ func validateTaskNetworkConfig(taskNetworkConfig *state.TaskNetworkConfig) error
...
+	// We will then unmarshall the json string and evaluate the fields of it.
+	tcCheckInjectionCommandComposed := nsenterPrefix + fmt.Sprintf(tcCheckInjectionCommandString, interfaceName)
+	fmt.Println(tcCheckInjectionCommandComposed)
+	cmdOutput, err := h.runExecCommand(ctx, tcCheckInjectionCommandComposed)
+	if err != nil {
+		return false, errors.New("failed to check network-packet-loss-fault: " + string(cmdOutput[:]) + err.Error())
+	}
+	// Log the command output to better help us debug.
+	logger.Info(fmt.Sprintf("%s command result: %s", tcCheckInjectionCommandComposed, string(cmdOutput[:])))
+	var outputUnmarshalled []map[string]interface{}
```
mye956, line 753: Q: Instead of a parsing/unmarshaling the output into a `map[string]interface{}`, I wonder if we can parse it into a struct and then get only the needed json keys from the output? (if it's not possible/output is not deterministic then please disregard).

Just ran `tc -j q show dev eth0 parent 1:1` locally and it seems to output the following:
```
[{
    "kind": "netem",
    "handle": "10:",
    "parent": "1:1",
    "options": {
        "limit": 1000,
        "loss-random": {
            "loss": 0.5,
            "correlation": 0
        },
        "ecn": false,
        "gap": 0
    }
}]
```

tshan2001: `tc -j q show dev eth0` does not always have all the fields. For example, when running on a host that doesn't have fault injected, it can have the following outout:
```
{"kind":"pfifo_fast","handle":"0:","parent":":1","options":{"bands":3,"priomap":[1,2,2,2,1,2,0,0,1,1,1,1,1,1,1,1],"multiqueue":false}}]
```
There's no loss-random field here. Also, when specifying `parent 1:1`, the output can also be `[]`.

Unmarshalling it into a `map[string]interface{}` and indexing the json struct iteratively would allow us to better catch the `nil` when parsing it

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -692,3 +722,140 @@ func validateTaskNetworkConfig(taskNetworkConfig *state.TaskNetworkConfig) error
...
+	}
+
+	return true, nil
+}
+
+// runExecCommand wraps around the execwrapper, providing a convenient way of running any Linux command
+// and getting the result in both stdout and stderr.
+func (h *FaultHandler) runExecCommand(ctx context.Context, linuxCommandString string) ([]byte, error) {
+	commandArray := strings.Split(linuxCommandString, " ")
+	cmdExec := h.osExecWrapper.CommandContext(ctx, commandArray[0], commandArray[1:]...)
```
mye956, line 819: Q: Should there be a `cmdExec.Run()` call somewhere in this method?

tshan2001: `cmdExec.CombinedOutput()` has output signature `([]byte, error)`, which is the same as the output of this method. We return the result of `cmdExec.CombinedOutput()` directly. CombinedOutput() is a generic method is os/exec. It runs the command and gives the stdout and stderr result at the same time (whereas Output() will only give stdout when exit code was 0, otherwise stdout will be nil)

mye956: Ah I see now

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -39,17 +41,25 @@ import (
...
+	delayMilliseconds                    = 123456789
+	jitterMilliseconds                   = 4567
+	lossPercent                          = 6
+	taskARN                              = "taskArn"
+	awsvpcNetworkMode                    = "awsvpc"
+	deviceName                           = "eth0"
+	invalidNetworkMode                   = "invalid"
+	tcLatencyFaultExistsCommandOutput    = `[{"kind":"netem","handle":"10:","parent":"1:1","options":{"limit":1000,"delay":{"delay":0.1,"jitter":0,"correlation":0},"ecn":false,"gap":0}}]`
+	tcLossFaultExistsCommandOutput       = `[{"kind":"netem","handle":"10:","dev":"eth0","parent":"1:1","options":{"limit":1000,"loss-random":{"loss":0.06,"correlation":0},"ecn":false,"gap":0}}]`
+	tcLossFaultDoesNotExistCommandOutput = `[{"kind":"dummyname"}]`
```
xxx0624, line 57: probably one more test case without `kind`

PR #4323, opened 2024-09-06
Create support for managed instances in netlib

welladev:
### Summary

Add support for managed instances and host mode networking

### Implementation details

We added a new platform type in netlib for managed instances. Since the host ENI is not added in the ACS payload we need to fetch all information about the host ENI and create a network interface object to return to the caller.
### Testing

Manually tested on a developer owned machine changing the dependency of the ecs agent to use the most up to date commit sha.

New tests cover the changes: 
no

Feature - Add support for managed instances

### Additional Information

### Licensing

### DIFF
--- a/ecs-agent/netlib/platform/api.go
+++ b/ecs-agent/netlib/platform/api.go
@@ -33,6 +33,9 @@ type API interface {
 		taskID string,
 		taskPayload *ecsacs.Task) (*tasknetworkconfig.TaskNetworkConfig, error)
 
+	// HandleHostMode returns error if host mode is not enabled for the platform.
+	HandleHostMode() error
+
 	// CreateNetNS creates a network namespace with the specified path.
 	CreateNetNS(netNSPath string) error
 
--- a/ecs-agent/netlib/platform/common.go
+++ b/ecs-agent/netlib/platform/common.go
@@ -15,6 +15,7 @@ package platform
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/ecscni"
@@ -28,6 +29,8 @@ const (
 	FirecrackerDebugPlatform = "ec2-debug-firecracker"
 	WarmpoolPlatform         = "warmpool"
 	FirecrackerPlatform      = "firecracker"
+	ManagedPlatform          = "managed-instance"
+	ManagedDebugPlatform     = "ec2-debug-managed-instance"
 )
 
 // executeCNIPlugin executes CNI plugins with the given network configs and a timeout context.
@@ -83,4 +86,9 @@ func (c *common) interfacesMACToName() (map[string]string, error) {
 	}
 
 	return macToName, nil
+}
+
+// HandleHostMode by default we do not want to support host mode.
+func (c *common) HandleHostMode() error {
+	return errors.New("invalid platform for host mode")
 }
--- a/ecs-agent/netlib/platform/common_linux.go
+++ b/ecs-agent/netlib/platform/common_linux.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/aws/amazon-ecs-agent/ecs-agent/acs/model/ecsacs"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/api/ecs/model/ecs"
+	"github.com/aws/amazon-ecs-agent/ecs-agent/credentials/instancecreds"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/logger"
 	netlibdata "github.com/aws/amazon-ecs-agent/ecs-agent/netlib/data"
 	"github.com/aws/amazon-ecs-agent/ecs-agent/netlib/model/appmesh"
@@ -41,6 +42,8 @@ import (
 	"github.com/aws/amazon-ecs-agent/ecs-agent/volume"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
 	cnitypes "github.com/containernetworking/cni/pkg/types/100"
 	cnins "github.com/containernetworking/plugins/pkg/ns"
 	"github.com/pkg/errors"
@@ -63,6 +66,8 @@ const (
 	// It is assigned 100 because it is an unrealistically high
 	// value for interface index.
 	indexHighValue = 100
+
+	metadataRetries = 5
 )
 
 // common will be embedded within every implementation of the platform API.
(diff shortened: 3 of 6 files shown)


ecs-agent/netlib/platform/common_linux.go
```diff
@@ -120,6 +125,14 @@ func NewPlatform(
 				common: commonPlatform,
 			},
 		}, nil
+	case ManagedPlatform:
+		config := aws.NewConfig().WithMaxRetries(metadataRetries)
+		config.Credentials = instancecreds.GetCredentials(false)
+		ec2Client := ec2metadata.New(session.New(), config)
```
JoseVillalta, line 131: Does this call always succeed?

ecs-agent/netlib/platform/managed_linux.go
```diff
@@ -0,0 +1,137 @@
...
+	netInt.KnownStatus = status.NetworkReady
+	defaultNameSpace, err := tasknetworkconfig.NewNetworkNamespace(netNSName, "", 0, nil, netInt)
+	if err != nil {
+		return nil, err
+	}
+	defaultNameSpace.KnownState = status.NetworkReady
+	defaultNameSpace.DesiredState = status.NetworkReady
+	return []*tasknetworkconfig.NetworkNamespace{defaultNameSpace}, nil
+}
+
```
JoseVillalta, line 134: Can we add a comment that explains why this is method is a no-op?

JoseVillalta (commented):
Left a couple of comments, otherwise LGTM

ecs-agent/netlib/platform/managed_linux.go
welladev: But it technically is not a no op as we are building a ENI with local info

ecs-agent/netlib/platform/common_linux.go
welladev: This is the same call that the top level client would do in initialization there technically is the possibility of failure which is why we have 5 retries

ecs-agent/netlib/platform/managed_linux.go
welladev: Update with proper comment

ecs-agent/netlib/platform/managed_linux.go
```diff
@@ -0,0 +1,138 @@
...
+// HandleHostMode is a no op since we are allowing Host Mode for managed instances
```
JoseVillalta, line 135: It's a no-op because Host Mode does not require network interface configuration. No need to invoke CNI plugins.

JoseVillalta: Can we modify the comment to say that?

welladev: done

JoseVillalta (approved):
LGTM

JoseVillalta (approved):

ecs-agent/netlib/platform/common.go
```diff
@@ -28,6 +29,7 @@ const (
 	FirecrackerDebugPlatform = "ec2-debug-firecracker"
 	WarmpoolPlatform         = "warmpool"
 	FirecrackerPlatform      = "firecracker"
+	ManagedPlatform          = "managed-instance"
```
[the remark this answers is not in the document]
welladev: done

JoseVillalta (approved):

JoseVillalta (dismissed):

ecs-agent/netlib/platform/api.go
```diff
@@ -33,6 +33,9 @@ type API interface {
 		taskID string,
 		taskPayload *ecsacs.Task) (*tasknetworkconfig.TaskNetworkConfig, error)
 
+	// HandleHostMode returns error if host mode is not enabled for the platform
```
timj-hh, line 36: nit: end comments with `.`

timj-hh: Same comment on L91 of `ecs-agent/netlib/platform/common.go`

ecs-agent/netlib/platform/managed_linux.go
```diff
@@ -0,0 +1,138 @@
...
+func (m *managedLinux) ConfigureServiceConnect(
+	ctx context.Context,
+	netNSPath string,
+	primaryIf *networkinterface.NetworkInterface,
+	scConfig *serviceconnect.ServiceConnectConfig,
+) error {
+	return m.common.configureServiceConnect(ctx, netNSPath, primaryIf, scConfig)
+}
+
+// buildDefaultNetworkNamespace generate a default network namespace for host mode where we do not need to create new namespaces
```
timj-hh, line 92: nit: generates*

timj-hh: also awkward wording;

> where we do not need to create new namespaces

we will only use one default namespace? are namespaces created/provided by users?

ecs-agent/netlib/platform/managed_linux.go
```diff
@@ -0,0 +1,138 @@
...
+			{
+				Primary:        aws.Bool(true),
+				PrivateAddress: aws.String(privateIpv4),
+			},
+		},
+		SubnetGatewayIpv4Address:     aws.String(subNet),
+		MacAddress:                   aws.String(macAddress),
+		DomainNameServers:            []*string{},
+		DomainName:                   []*string{},
+		PrivateDnsName:               aws.String("default"),
```
timj-hh, line 116: nit: should abstract this string `default` to a const

ecs-agent/netlib/platform/managed_linux.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	netNSName := networkinterface.NetNSName(taskID, "default")
+	netInt, _ := networkinterface.New(hostENI, "default", nil, macToNames)
+	netInt.Default = true
+	netInt.DesiredStatus = status.NetworkReady
+	netInt.KnownStatus = status.NetworkReady
+	defaultNameSpace, err := tasknetworkconfig.NewNetworkNamespace(netNSName, "", 0, nil, netInt)
+	if err != nil {
+		return nil, err
```
timj-hh, line 128: let's log the error here for debugging reasons

timj-hh (commented):
Thanks for the change! I believe you will have to run `go mod tidy && go mod vendor` in the `agent` directory to propagate these changes. We should also add some unit testing to this file to match the other `netlib/platform` files

ecs-agent/netlib/platform/managed_linux.go
welladev: host mode requires default namespace since we are using the default eni

ecs-agent/netlib/platform/managed_linux.go
welladev: we will return the error and it will propagate and logged in the two package

welladev: Refactored and added logs for errors for debugging

ecs-agent/netlib/platform/managed_linux.go
welladev: created const for defaultArg

ecs-agent/netlib/platform/managed_linux.go
welladev: updated function documentation for clarity

ecs-agent/netlib/platform/api.go
welladev: done

timj-hh (approved):

JoseVillalta (approved):

PR #4328, opened 2024-09-11
wip

singholt:
### Testing

New tests cover the changes: 

### Additional Information

### Licensing


mye956 (dismissed):

agent/engine/host_resource_manager.go
```diff
@@ -139,7 +139,10 @@ func (h *HostResourceManager) consume(taskArn string, resources map[string]*ecs.
 		logger.Info("Resources successfully consumed, continue to task creation", logger.Fields{"taskArn": taskArn})
 		return true, nil
 	}
-	logger.Info("Resources not consumed, enough resources not available", logger.Fields{"taskArn": taskArn})
+	logger.Info("Resources not consumed, enough resources not available", logger.Fields{
+		"taskArn":  taskArn,
+		"resource": failedResourceKey,
```
prateekchaudhry, line 144: Just to note, there can be cases when more than one resources are not available. This log statement will print the first one `h.consumable` exits one. I think can be still okay to log it, although which resource gets logged is probably not deterministic.

prateekchaudhry (dismissed):

agent/engine/host_resource_manager.go
singholt: good callout!

singholt: I incorporated this feedback, now we log all non-consumable resources. please take a look

prateekchaudhry (approved):

agent/engine/host_resource_manager.go
prateekchaudhry: lgtm, thank you!

PR #4325, opened 2024-09-09
Add fault-injection capability

harishxr:
### Summary

Added support to advertise fault injection capability (ecs.capability.fault-injection) to ecs agent when the required network packages are present on the host.
### Implementation details

* Updated agent capabilities to include the fault inject capability.
* Added `checkNetworkTooling()` to check if the required network packages are available on the host for the capability to be advertised.
* Adjusted unit tests to verify fault injection capability based on available network tooling.
### Testing

* Added unit tests to ensure capability is enabled only when network tooling packages are available.
* Installed `iproute-tc` and `iptables` package on the host and ran `aws ecs describe-container-instances --cluster <cluster-name> --container-instances <instance-id>`
```
{
                    "name": "ecs.capability.fault-injection"
                },
```
* The `ecs.capability.fault-injection` is enabled when the required packages are present on the host.
* If required packages are not present, following message is found in `ecs.agent.log`
```
level=warn time=2024-09-09T21:17:08Z msg="Fault injection capability not enabled: Required network tools (iptables, tc) are missing" module=agent_capability.go
```

New tests cover the changes: yes

### Description for the changelog

* Added fault injection capability to ECS Agent.
### Additional Information

**Does this PR include breaking model changes? If so, Have you added transformation functions?**
  

**Does this PR include the addition of new environment variables in the README?**

### Licensing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4336, opened 2024-09-16
Prevent reporting fault-injection capability from being reported

amogh09:
### Summary

#4325 added a new fault-injection capability. The feature is still in the works so we don't want to start reporting this capability just yet. This PR prevents the capability from being reported. 

### Testing

Prepared all necessary tooling dependencies for fault-injection capability and ran Agent. Verified in `describe-tasks` response that the capability is not reported. 

Stop reporting fault-injection capability

### Additional Information

No

### Licensing

### DIFF
--- a/agent/app/agent_capability.go
+++ b/agent/app/agent_capability.go
@@ -314,8 +314,8 @@ func (agent *ecsAgent) capabilities() ([]*ecs.Attribute, error) {
 		capabilities = removeAttributesByNames(capabilities, externalUnsupportedCapabilities)
 	}
 
-	// add fault-injection capabilities if applicable
-	capabilities = agent.appendFaultInjectionCapabilities(capabilities)
+	// TODO add fault-injection capabilities if applicable
+	// capabilities = agent.appendFaultInjectionCapabilities(capabilities)
 
 	return capabilities, nil
 }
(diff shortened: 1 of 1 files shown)


xxx0624 (dismissed):

mye956 (approved):

agent/app/agent_capability.go
```diff
@@ -543,11 +543,14 @@ func (agent *ecsAgent) appendEBSTaskAttachCapabilities(capabilities []*ecs.Attri
 	return capabilities
 }
 
+// TODO Remove linter directive below when the function becomes used
```
danehlim, line 546: If we are not already, can we also make sure to track this TODO as a task internally (separate from this code comment)? Just want to ensure this removal does not get missed nor left behind in the "sea of lingering TODOs" currently in the code base. Thanks!

danehlim (approved):

agent/app/agent_capability.go
amogh09: Yep we are tracking internally.

amogh09: I will be submitting a PR to a feature branch with that change once this is merged.

Yiyuanzzz (approved):

PR #4337, opened 2024-09-16
Return explicit error message when required request body is missing

xxx0624:
### Summary

This PR is to return an explicit error message when the required request body is missing for a fault injection request.

### Implementation details

- Add one check before we decode it from the request (in `decodeRequest` [code ref](https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go#L507))

### Testing

- Added new unit test.
- Manual test to verify the new error message.
```
// before the PR
sh-5.2# curl -X GET $ECS_AGENT_URI/fault/v1/network-blackhole-port  -v
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying 169.254.170.2:80...
* Connected to 169.254.170.2 (169.254.170.2) port 80
> GET /api/605a5eac95e4408fb27c87d43a135be0-3147973833/fault/v1/network-blackhole-port HTTP/1.1
> Host: 169.254.170.2
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< X-Rate-Limit-Duration: 1
< X-Rate-Limit-Limit: 0.20
< X-Rate-Limit-Request-Forwarded-For: 
< X-Rate-Limit-Request-Remote-Addr: 169.254.172.2:46042
< Date: Mon, 16 Sep 2024 21:56:55 GMT
< Content-Length: 15
< 
* Connection #0 to host 169.254.170.2 left intact
{"Error":"EOF"}sh-5.2#


// after the PR
sh-5.2# curl -X GET $ECS_AGENT_URI/fault/v1/network-blackhole-port  -v
Note: Unnecessary use of -X or --request, GET is already inferred.
*   Trying 169.254.170.2:80...
* Connected to 169.254.170.2 (169.254.170.2) port 80
> GET /api/605a5eac95e4408fb27c87d43a135be0-3147973833/fault/v1/network-blackhole-port HTTP/1.1
> Host: 169.254.170.2
> User-Agent: curl/8.5.0
> Accept: */*
> 
< HTTP/1.1 400 Bad Request
< Content-Type: application/json
< X-Rate-Limit-Duration: 1
< X-Rate-Limit-Limit: 0.20
< X-Rate-Limit-Request-Forwarded-For: 
< X-Rate-Limit-Request-Remote-Addr: 169.254.172.2:33716
< Date: Mon, 16 Sep 2024 22:08:04 GMT
< Content-Length: 44
< 
* Connection #0 to host 169.254.170.2 left intact
{"Error":"required request body is missing"}sh-5.2#
```

New tests cover the changes: 

Yes

- Feature: Return an explicit error message when required request body is missing for fault injection requests.

### Additional Information

### Licensing

### DIFF
--- a/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
+++ b/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
@@ -79,12 +79,12 @@ func (h *FaultHandler) StartNetworkBlackholePort() func(http.ResponseWriter, *ht
 		requestType := fmt.Sprintf(startFaultRequestType, types.BlackHolePortFaultType)
 
 		// Parse the fault request
-		err := decodeRequest(w, &request, requestType, r, true)
+		err := decodeRequest(w, &request, requestType, r)
 		if err != nil {
 			return
 		}
 		// Validate the fault request
-		err = validateRequest(w, request, requestType, true)
+		err = validateRequest(w, request, requestType)
 		if err != nil {
 			return
 		}
@@ -126,12 +126,12 @@ func (h *FaultHandler) StopNetworkBlackHolePort() func(http.ResponseWriter, *htt
 		requestType := fmt.Sprintf(stopFaultRequestType, types.BlackHolePortFaultType)
 
 		// Parse the fault request
-		err := decodeRequest(w, &request, requestType, r, true)
+		err := decodeRequest(w, &request, requestType, r)
 		if err != nil {
 			return
 		}
 		// Validate the fault request
-		err = validateRequest(w, request, requestType, true)
+		err = validateRequest(w, request, requestType)
 		if err != nil {
 			return
 		}
@@ -177,12 +177,12 @@ func (h *FaultHandler) CheckNetworkBlackHolePort() func(http.ResponseWriter, *ht
 		requestType := fmt.Sprintf(checkStatusFaultRequestType, types.BlackHolePortFaultType)
 
 		// Parse the fault request
-		err := decodeRequest(w, &request, requestType, r, true)
+		err := decodeRequest(w, &request, requestType, r)
 		if err != nil {
 			return
 		}
 		// Validate the fault request
-		err = validateRequest(w, request, requestType, true)
+		err = validateRequest(w, request, requestType)
 		if err != nil {
 			return
 		}
@@ -226,13 +226,13 @@ func (h *FaultHandler) StartNetworkLatency() func(http.ResponseWriter, *http.Req
 		var request types.NetworkLatencyRequest
 		requestType := fmt.Sprintf(startFaultRequestType, types.LatencyFaultType)
 		// Parse the fault request
-		err := decodeRequest(w, &request, requestType, r, true)
+		err := decodeRequest(w, &request, requestType, r)
 		if err != nil {
 			return
 		}
 
 		// Validate the fault request
-		err = validateRequest(w, request, requestType, true)
+		err = validateRequest(w, request, requestType)
 		if err != nil {
 			return
 		}
@@ -274,12 +274,12 @@ func (h *FaultHandler) StopNetworkLatency() func(http.ResponseWriter, *http.Requ
 		requestType := fmt.Sprintf(stopFaultRequestType, types.LatencyFaultType)
 
 		// Parse the fault request
-		err := decodeRequest(w, &request, requestType, r, true)
+		err := decodeRequest(w, &request, requestType, r)
 		if err != nil {
 			return
 		}
 		// Validate the fault request
-		err = validateRequest(w, request, requestType, true)
+		err = validateRequest(w, request, requestType)
 		if err != nil {
 			return
 		}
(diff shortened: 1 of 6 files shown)


ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -39,7 +39,7 @@ const (
 	stopFaultRequestType        = "stop %s"
 	checkStatusFaultRequestType = "check status %s"
 	invalidNetworkModeError     = "%s mode is not supported. Please use either host or awsvpc mode."
-	faultInjectionEnabledError  = "fault injection is not enabled for task: %s"
+	faultInjectionEnabledError  = "enableFaultInjection is not enabled for task: %s"
```
amogh09, line 42: nit: I personally prefer the older error message as it is not coupled to the name of the task definition parameter but describes the issue in general. No strong opinion though.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -501,11 +501,19 @@ func (h *FaultHandler) CheckNetworkPacketLoss() func(http.ResponseWriter, *http.
 	}
 }
 
-// decodeRequest will translate/unmarshal an incoming fault injection request into one of the network fault structs
-func decodeRequest(w http.ResponseWriter, request types.NetworkFaultRequest, requestType string, r *http.Request) error {
+// decodeRequest will log the request and then translate/unmarshal an incoming fault injection request into
+// one of the network fault structs if the request body is required.
+func decodeRequest(w http.ResponseWriter, request types.NetworkFaultRequest, requestType string,
+	r *http.Request, requiredRequestBody bool) error {
```
amogh09, line 507: Why do we need `requiredRequestBody` parameter? If any function is calling `decodeRequest` then it should always expect a request body?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -501,11 +501,19 @@ func (h *FaultHandler) CheckNetworkPacketLoss() func(http.ResponseWriter, *http.
...
 	logRequest(requestType, r)
+
 	jsonDecoder := json.NewDecoder(r.Body)
 	if err := jsonDecoder.Decode(request); err != nil {
+		// The request has empty body. Respond an explicit message if that's required.
+		if err == io.EOF && requiredRequestBody {
```
amogh09, line 513: Does `io.EOF` always signal an empty body or could that happen for more cases such as connection issues?

As per the documentation of `http.Request` struct - 

> 	// For client requests, a nil body means the request has no
>	// body, such as a GET request. The HTTP Client's Transport
>	// is responsible for calling the Close method.

https://pkg.go.dev/net/http#Request

So should we check `r.Body == nil` instead?

mye956: >So should we check r.Body == nil instead?
+1

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
xxx0624: I was thinking to decode the request for future uses (because the client can still include the request payload if that's optional). But yes you are correct. I can remove this when request body is not required.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
xxx0624: I don't have a strong preference as well. I like the new one a bit more because if customers see this message they can know what changes need to be done in their tasks.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
xxx0624:
```
	// Body is the request's body.
	//
	// For client requests, a nil body means the request has no
	// body, such as a GET request. The HTTP Client's Transport
	// is responsible for calling the Close method.
	//
	// For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
	//
	// Body must allow Read to be called concurrently with Close.
	// In particular, calling Close should unblock a Read waiting
	// for input.
	Body io.ReadCloser
```

xxx0624: The body is always non-nil so we should not check if that is nil.

amogh09: Thanks for checking @xxx0624 . That makes sense to me.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
xxx0624: Removed this param.

amogh09 (approved):

tshan2001 (approved):

mye956 (approved):

PR #4342, opened 2024-09-17
Get default network interfaces on host mode

xxx0624:
### Summary

This PR is to add an util method (which is equivalent to `ip route show`) to get the device name of the first default network interface on the underlying instance. No use as of now but it would be used for fault injection later on host mode.

### Implementation details

- Add two more methods in the [netlink wrapper](https://github.com/aws/amazon-ecs-agent/blob/master/ecs-agent/utils/netlinkwrapper/netlink_linux.go) to support 1. to get all routes ([netlink.RouteList](https://github.com/vishvananda/netlink/blob/a01829657b53687bcde90c030e2013e482783847/route_linux.go#L1170-L1173) is equivalent to `ip route show`) and 2. to get link/interface info by link index.
- Add a new packge in [tmds util](https://github.com/aws/amazon-ecs-agent/tree/master/ecs-agent/tmds/utils/) to get the device name of the first default network interface.

### Testing

Manual testing on one EC2 instance to verify if the util method works

```
root@ip-10-194-21-204:~/cc# go run b.go
Found default interface: ens5
Found default interface: ens6

Found first default interface: ens5, <nil>
root@ip-10-194-21-204:~/cc# ip route show default
default via 10.194.21.1 dev ens5 proto dhcp src 10.194.21.204 metric 100 
default via 10.194.21.1 dev ens6 proto dhcp src 10.194.21.8 metric 200 
root@ip-10-194-21-204:~/cc#

root@ip-10-194-21-204:~/cc# cat b.go
package main

import (
        "fmt"
        "github.com/vishvananda/netlink"
)

// DefaultNetInterfaceName returns the device name of the first default network interface
// available on the instance. If none exist, an empty string and nil will be returned.
func DefaultNetInterfaceName() (string, error) {
        routes, err := netlink.RouteList(nil, netlink.FAMILY_ALL)
        if err != nil {
                return "", err
        }

        // Iterate over all routes
        firstDefaultNetInterfaceName := ""
        for _, route := range routes {
                if route.Gw == nil {
                        // A default route has a gateway. If it doesn't, skip it.
                        continue
                }

                if route.Dst == nil || route.Dst.String() == "0.0.0.0/0" || route.Dst.String() == "::/0" {
                        // Get the link (interface) associated with the default route
                        link, err := netlink.LinkByIndex(route.LinkIndex)
                        if err != nil {
                                fmt.Printf("Error: %v\n", err)
                                continue
                        } else {
                                if firstDefaultNetInterfaceName == "" {
                                        firstDefaultNetInterfaceName = link.Attrs().Name
                                }
                                fmt.Printf("Found default interface: %v\n", link.Attrs().Name)
                        }
                }
        }
        return firstDefaultNetInterfaceName, nil
}

func main() {
        res, err := DefaultNetInterfaceName()
        fmt.Printf("\nFound first default interface: %v, %v\n", res, err)
}

root@ip-10-194-21-204:~/cc#
```

New tests cover the changes: 

Yes

### Additional Information

### Licensing


ecs-agent/tmds/utils/netconfig/netconfig_linux.go
```diff
@@ -0,0 +1,56 @@
...
+// available on the instance. If none exist, an empty string and nil will be returned.
+func DefaultNetInterfaceName(netlinkClient netlinkwrapper.NetLink) (string, error) {
+	routes, err := netlinkClient.RouteList(nil, netlink.FAMILY_ALL)
+	if err != nil {
+		return "", err
+	}
+
+	// Iterate over all routes
+	for _, route := range routes {
+		if route.Gw == nil {
```
xxx0624, line 37: FYI - learn this from the 3rd party lib - https://github.com/tailscale/tailscale/blob/main/net/netmon/interfaces_linux.go#L168-L193

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
```diff
@@ -0,0 +1,56 @@
...
+			// A default route has a gateway. If it doesn't, skip it.
+			continue
+		}
+
+		if route.Dst == nil || route.Dst.String() == "0.0.0.0/0" || route.Dst.String() == "::/0" {
```
xxx0624, line 42: FYI - default route will have a dst of 0.0.0.0/0 or ::/0 - https://networklessons.com/tag/default-route#:~:text=The%20default%20route%20(0.0.,or%20advertised%20in%20routing%20protocols.

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
```diff
@@ -0,0 +1,56 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
mye956, line 1: Q: Should we have the following here?
```
//go:build linux
// +build linux
```

ecs-agent/tmds/utils/netconfig/netconfig_linux_test.go
```diff
@@ -0,0 +1,169 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
mye956, line 1: ditto but for windows

ecs-agent/utils/netlinkwrapper/netlink_linux.go
```diff
@@ -38,3 +40,11 @@ func (nl *netLink) LinkByName(name string) (netlink.Link, error) {
 func (nl *netLink) LinkSetUp(link netlink.Link) error {
 	return netlink.LinkSetUp(link)
 }
+
```
mye956, line 43: nit (non-blocking): maybe we can add a comment here saying this is equivalent to running `ip route show`

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
```diff
@@ -0,0 +1,56 @@
...
+		if route.Dst == nil || route.Dst.String() == "0.0.0.0/0" || route.Dst.String() == "::/0" {
+			// Get the link (interface) associated with the default route
+			link, err := netlinkClient.LinkByIndex(route.LinkIndex)
+			if err != nil {
+				logger.Warn("Not able to get the associated network interface by the index", logger.Fields{
+					field.Error: err,
+					"LinkIndex": route.LinkIndex,
+				})
+			} else {
+				return link.Attrs().Name, nil
```
mye956, line 51: nit(non-blocking): might be helpful to add a log on a debug level that the device name of the default network interface was found

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
tshan2001: I saw we had the following below:
```
//go:build !windows
// +build !windows
```
Is this equivalent to specifying to only build for Linux?

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
```diff
@@ -0,0 +1,56 @@
...
+		if route.Gw == nil {
+			// A default route has a gateway. If it doesn't, skip it.
+			continue
+		}
+
+		if route.Dst == nil || route.Dst.String() == "0.0.0.0/0" || route.Dst.String() == "::/0" {
+			// Get the link (interface) associated with the default route
+			link, err := netlinkClient.LinkByIndex(route.LinkIndex)
+			if err != nil {
+				logger.Warn("Not able to get the associated network interface by the index", logger.Fields{
```
tshan2001, line 46: Q: what's the difference between Warn and Error?

ecs-agent/utils/netlinkwrapper/netlink_linux.go
```diff
@@ -38,3 +40,11 @@ func (nl *netLink) LinkByName(name string) (netlink.Link, error) {
...
+func (nl *netLink) RouteList(link netlink.Link, family int) ([]netlink.Route, error) {
```
tshan2001, line 44: Why do we need these methods here to be non-static? Seems like the variable `nl` is not used. Should we use `return nl.RouteList(link, family)`?

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
xxx0624: I will move the tags to the top. Generally these two are not equivalent because there could be bsd and darwin and both are not windows. I will update it to be linux explicitly.

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
xxx0624: IMO error is something fatal to the operation and warn means the process can continue even this happens.

ecs-agent/tmds/utils/netconfig/netconfig_linux.go
xxx0624: Added.

ecs-agent/tmds/utils/netconfig/netconfig_windows.go
```diff
@@ -0,0 +1,25 @@
+// Copyright Amazon.com Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
xxx0624: Updated

ecs-agent/tmds/utils/netconfig/netconfig_linux_test.go
xxx0624: Updated

ecs-agent/utils/netlinkwrapper/netlink_linux.go
xxx0624: Yes added.

ecs-agent/utils/netlinkwrapper/netlink_linux.go
xxx0624: Mainly for testing purpose to have this internal only object.

mye956 (approved):

tshan2001 (approved):

amogh09 (approved):

PR #4343, opened 2024-09-17
[WIP] Use execution creds ID in ECR auth cache key

sparrc:
### Testing

New tests cover the changes: 

### Additional Information

### Licensing


agent/dockerclient/dockerauth/ecr.go
```diff
@@ -52,7 +55,8 @@ const (
 
 // String formats the cachKey as a string
 func (key *cacheKey) String() string {
-	return fmt.Sprintf("%s-%s-%s-%s", key.roleARN, key.region, key.registryID, key.endpointOverride)
+	return fmt.Sprintf("%s%s-%s-%s-%s",
```
amogh09, line 58: Should we have a hyphen between credentialsID and roleARN?

agent/dockerclient/dockerauth/ecr.go
```diff
@@ -91,6 +95,7 @@ func (authProvider *ecrAuthProvider) GetAuthconfig(image string,
 	// containers pull with the same role can be cached
 	if authData.GetPullCredentials() != (credentials.IAMRoleCredentials{}) {
 		key.roleARN = authData.GetPullCredentials().RoleArn
+		key.credentialsID = authData.GetPullCredentials().CredentialsID
```
amogh09, line 98: This will likely lead to more getAuthorzationToken calls from customer accounts as I think there will be no cache hits. Throttling limit for this API is "Each supported Region: 500 per second". Should we be worried about this?

agent/dockerclient/dockerauth/ecr.go
sparrc: This way we don't change the cache key if there's no credentials ID to add (for example when we dont use task execution role). I dont _think_ this cache is stored to disk so probably doesn't matter, but seemed safer to do this without the dash anyways.

agent/dockerclient/dockerauth/ecr.go
sparrc: I think the biggest risk here would be for customers who run very large instances with lots of tasks on a single instance (and those tasks are also using a task execution role). Currently this customer might be seeing a high rate of cache hits if all of their tasks reuse the same task execution role.

On the other hand, hopefully a customer such as this already has some mitigations in place to not need as many image pulls. Also, 500 tps is a very high limit so I wouldn't think this would come up often.

I think that if the ECS backend is telling us that the credentials are unique (via a unique credentials ID), then the right thing to do is for the agent to refresh creds.

yinyic (approved):

agent/dockerclient/dockerauth/ecr.go
singholt: Can the cache key be just the `credentialsID`? If the credentialsID is the same, there's a cache hit, else we get credentials from ECR.

sparrc: It can be, although we only get credentialsID when we're using the task execution role, we dont get them when we use the EC2 instance role, so we'd need to keep the rest of the fields anyways. Using only credentials ID for when we have it would fork the behavior a bit more than I wanted to change it.

agent/dockerclient/dockerauth/ecr.go
amogh09: That makes sense to me. Thanks.

amogh09 (dismissed):

hozkaya2000 (approved):

yinyic (approved):

PR #4363, opened 2024-09-24
Add network-latency fault implementation

tshan2001:
### Summary
Adding implementation for the 3 APIs for network-latency fault, including `StartNetworkLatency()`, `StopNetworkLatency()`, and `CheckNetworkLatency()`, as well as corresponding unit tests.

### Implementation details
The network-latency fault implementation is similar to that of the network-packet-loss fault.
The following Linux command will be called:
```
# Start
<nsenterPrefix> tc qdisc add dev <interfaceName> root handle 1: prio priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2
<nsenterPrefix> "tc qdisc add dev <interfaceName> parent 1:1 handle 10: netem delay <latency>ms <jitter>ms
# Stop
Share the same helper method as StopNetworkPacketLoss()
# Check
Share the same helper method as CheckNetworkPacketLoss()
```

### Testing
Unit tests for the TMDS package was run.
```
 % go test -tags unit -v -run TestStartNetworkLatency /workplace/tianzes/amazon-ecs-agent/ecs-agent/tmds/handlers/fault/v1/handlers
...
--- PASS: TestStartNetworkLatency (0.00s)

 % go test -tags unit -v -run TestStopNetworkLatency /workplace/tianzes/amazon-ecs-agent/ecs-agent/tmds/handlers/fault/v1/handlers
...
--- PASS: TestStopNetworkLatency (0.00s)

 % go test -tags unit -v -run TestCheckNetworkLatency /workplace/tianzes/amazon-ecs-agent/ecs-agent/tmds/handlers/fault/v1/handlers
...
--- PASS: TestCheckNetworkLatency (0.00s)
```

New tests cover the changes: 
Besides existing test cases for the handler, also added the following cases specifically for the start endpoint:
1. When there doesn't exist a fault on the instance
2. When there exists a packet loss fault
3. When there exists a latency fault
4. When the request contains an unknown field but the rest of the payload is proper.

### Manual Testing
```
# First curl the check latency endpoint. Since we haven't injected anything yet, response should be not running
 % curl -X GET ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
{"Status":"not-running"}

# Now curl the start endpoint
 % curl -X PUT ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
{"Status":"running"}

# Curl the check endpoint again to confirm that the fault was running. Note that the IP address in the check payload is different from the one in the start payload. This shouldn't matter because we won't use it. But if we agree that we won't need it we should take an AI to remove it from the check and stop payload.
 % curl -X GET ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
{"Status":"running"}

# Curl the start endpoint again. This time it should report that there's already a latency fault running.
 % curl -X PUT ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
{"Error":"There is already one network latency fault running"}

# Curl the stop endpoint with arbitrary IP address
 % curl -X DELETE ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
{"Status":"stopped"}

# Finally, curl the check endpoint again. The result should be not-running
 % curl -X GET ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
 {"Status":"not-running"}

# Additional check:
# First start a network packet loss fault
 % curl -X PUT ${ECS_AGENT_URI}/fault/v1/network-packet-loss --data '{"lossPercent":6, "Sources":["192.168.0.1", "10.1.1.1", "25.168.10.2"]}'
{"Status":"running"}
# Now attempt to start a latency fault. The conflict should be reported
 % curl -X PUT ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
{"Error":"There is already one network packet loss fault running"}

# Now confirm that the fault is working
# First start a latency fault with IP address 8.8.8.8, with 1000ms delay and 100ms jitter.
 % curl -X PUT ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":1000, "JitterMilliseconds":100, "Sources":["8.8.8.8"]}'
{"Status":"running"}
# Now ping the ip address
 % ping 8.8.8.8 -D
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
...
[1727218069.113409] 64 bytes from 8.8.8.8: icmp_seq=20 ttl=57 time=1101 ms
[1727218070.117507] 64 bytes from 8.8.8.8: icmp_seq=21 ttl=57 time=1075 ms
[1727218071.070307] 64 bytes from 8.8.8.8: icmp_seq=22 ttl=57 time=1027 ms
[1727218072.067608] 64 bytes from 8.8.8.8: icmp_seq=23 ttl=57 time=1023 ms
[1727218073.097244] 64 bytes from 8.8.8.8: icmp_seq=24 ttl=57 time=1051 ms
# Now remove the fault
 % curl -X DELETE ${ECS_AGENT_URI}/fault/v1/network-latency --data '{"DelayMilliseconds":100, "JitterMilliseconds":10, "Sources":["192.168.0.1"]}'
{"Status":"stopped"}
# And ping the ip address again
 % ping 8.8.8.8 -D
PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
...
[1727218138.572887] 64 bytes from 8.8.8.8: icmp_seq=4 ttl=57 time=9.29 ms
[1727218139.574341] 64 bytes from 8.8.8.8: icmp_seq=5 ttl=57 time=9.32 ms
[1727218140.575834] 64 bytes from 8.8.8.8: icmp_seq=6 ttl=57 time=9.35 ms
[1727218141.577308] 64 bytes from 8.8.8.8: icmp_seq=7 ttl=57 time=9.32 ms
[1727218142.578780] 64 bytes from 8.8.8.8: icmp_seq=8 ttl=57 time=9.31 ms
# Note the difference in time (~1000ms vs ~9ms)
```

Add network latency fault implementation 

### Additional Information

### Licensing

### DIFF
--- a/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
+++ b/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
@@ -47,7 +47,9 @@ const (
 	invalidNetworkModeError     = "%s mode is not supported. Please use either host or awsvpc mode."
 	faultInjectionEnabledError  = "enableFaultInjection is not enabled for task: %s"
 	requestTimedOutError        = "%s: request timed out"
-	requestTimeoutDuration      = 5 * time.Second
+	// This is our initial assumption of how much time it would take for the Linux commands used to inject faults
+	// to finish. This will be confirmed/updated after more testing.
+	requestTimeoutDuration = 5 * time.Second
 	// Commands that will be used to start/stop/check fault.
 	iptablesNewChainCmd              = "iptables -N %s"
 	iptablesAppendChainRuleCmd       = "iptables -A %s -p %s --dport %s -j DROP"
@@ -59,6 +61,7 @@ const (
 	nsenterCommandString             = "nsenter --net=%s "
 	tcCheckInjectionCommandString    = "tc -j q show dev %s parent 1:1"
 	tcAddQdiscRootCommandString      = "tc qdisc add dev %s root handle 1: prio priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2"
+	tcAddQdiscLatencyCommandString   = "tc qdisc add dev %s parent 1:1 handle 10: netem delay %dms %dms"
 	tcAddQdiscLossCommandString      = "tc qdisc add dev %s parent 1:1 handle 10: netem loss %d%%"
 	tcAddFilterForIPCommandString    = "tc filter add dev %s protocol ip parent 1:0 prio 1 u32 match ip dst %s flowid 1:1"
 	tcDeleteQdiscParentCommandString = "tc qdisc del dev %s parent 1:1 handle 10:"
@@ -536,18 +539,53 @@ func (h *FaultHandler) StartNetworkLatency() func(http.ResponseWriter, *http.Req
 		rwMu.Lock()
 		defer rwMu.Unlock()
 
-		// TODO: Check status of current fault injection
-		// TODO: Invoke the start fault injection functionality if not running
-
-		responseBody := types.NewNetworkFaultInjectionSuccessResponse("running")
-		logger.Info("Successfully started fault", logger.Fields{
+		var responseBody types.NetworkFaultInjectionResponse
+		var httpStatusCode int
+		stringToBeLogged := "Failed to start fault"
+		// All command executions for the start network latency workflow all together should finish within 5 seconds.
+		// Thus, create the context here so that it can be shared by all os/exec calls.
+		ctx, cancel := h.osExecWrapper.NewExecContextWithTimeout(context.Background(), requestTimeoutDuration)
+		defer cancel()
+		// Check the status of current fault injection.
+		latencyFaultExists, packetLossFaultExists, err := h.checkTCFault(ctx, taskMetadata)
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+			responseBody = types.NewNetworkFaultInjectionErrorResponse(fmt.Sprintf(requestTimedOutError, requestType))
+			httpStatusCode = http.StatusInternalServerError
+		} else if err != nil {
+			responseBody = types.NewNetworkFaultInjectionErrorResponse(err.Error())
+			httpStatusCode = http.StatusInternalServerError
+		} else {
+			// If there already exists a fault in the task network namespace.
+			if latencyFaultExists {
+				responseBody = types.NewNetworkFaultInjectionErrorResponse("There is already one network latency fault running")
+				httpStatusCode = http.StatusConflict
+			} else if packetLossFaultExists {
+				responseBody = types.NewNetworkFaultInjectionErrorResponse("There is already one network packet loss fault running")
+				httpStatusCode = http.StatusConflict
+			} else {
+				// Invoke the start fault injection functionality if not running.
+				err := h.startNetworkLatencyFault(ctx, taskMetadata, request)
+				if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+					responseBody = types.NewNetworkFaultInjectionErrorResponse(fmt.Sprintf(requestTimedOutError, requestType))
+					httpStatusCode = http.StatusInternalServerError
+				} else if err != nil {
+					responseBody = types.NewNetworkFaultInjectionErrorResponse(err.Error())
+					httpStatusCode = http.StatusInternalServerError
+				} else {
+					stringToBeLogged = "Successfully started fault"
+					responseBody = types.NewNetworkFaultInjectionSuccessResponse("running")
+					httpStatusCode = http.StatusOK
+				}
+			}
+		}
+		logger.Info(stringToBeLogged, logger.Fields{
 			field.RequestType: requestType,
 			field.Request:     request.ToString(),
 			field.Response:    responseBody.ToString(),
 		})
 		utils.WriteJSONResponse(
 			w,
-			http.StatusOK,
+			httpStatusCode,
 			responseBody,
 			requestType,
 		)
@@ -583,18 +621,51 @@ func (h *FaultHandler) StopNetworkLatency() func(http.ResponseWriter, *http.Requ
 		rwMu.Lock()
 		defer rwMu.Unlock()
 
-		// TODO: Check status of current fault injection
-		// TODO: Invoke the stop fault injection functionality if running
-
-		responseBody := types.NewNetworkFaultInjectionSuccessResponse("stopped")
-		logger.Info("Successfully stopped fault", logger.Fields{
+		var responseBody types.NetworkFaultInjectionResponse
+		var httpStatusCode int
+		stringToBeLogged := "Failed to stop fault"
+		// All command executions for the stop network latency workflow all together should finish within 5 seconds.
+		// Thus, create the context here so that it can be shared by all os/exec calls.
+		ctx, cancel := h.osExecWrapper.NewExecContextWithTimeout(context.Background(), requestTimeoutDuration)
+		defer cancel()
+		// Check the status of current fault injection.
+		latencyFaultExists, _, err := h.checkTCFault(ctx, taskMetadata)
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+			responseBody = types.NewNetworkFaultInjectionErrorResponse(fmt.Sprintf(requestTimedOutError, requestType))
+			httpStatusCode = http.StatusInternalServerError
+		} else if err != nil {
+			responseBody = types.NewNetworkFaultInjectionErrorResponse(err.Error())
+			httpStatusCode = http.StatusInternalServerError
+		} else {
+			// If there doesn't already exist a network-latency fault
+			if !latencyFaultExists {
+				stringToBeLogged = "No fault running"
+				responseBody = types.NewNetworkFaultInjectionSuccessResponse("stopped")
+				httpStatusCode = http.StatusOK
+			} else {
+				// Invoke the stop fault injection functionality if running.
+				err := h.stopTCFault(ctx, taskMetadata)
+				if errors.Is(err, context.DeadlineExceeded) {
+					responseBody = types.NewNetworkFaultInjectionErrorResponse(fmt.Sprintf(requestTimedOutError, requestType))
+					httpStatusCode = http.StatusInternalServerError
+				} else if err != nil {
+					responseBody = types.NewNetworkFaultInjectionErrorResponse(err.Error())
+					httpStatusCode = http.StatusInternalServerError
+				} else {
+					stringToBeLogged = "Successfully stopped fault"
+					responseBody = types.NewNetworkFaultInjectionSuccessResponse("stopped")
+					httpStatusCode = http.StatusOK
+				}
+			}
+		}
+		logger.Info(stringToBeLogged, logger.Fields{
 			field.RequestType: requestType,
 			field.Request:     request.ToString(),
 			field.Response:    responseBody.ToString(),
 		})
 		utils.WriteJSONResponse(
 			w,
-			http.StatusOK,
+			httpStatusCode,
 			responseBody,
 			requestType,
 		)
@@ -630,17 +701,40 @@ func (h *FaultHandler) CheckNetworkLatency() func(http.ResponseWriter, *http.Req
 		rwMu.RLock()
 		defer rwMu.RUnlock()
 
-		// TODO: Check status of current fault injection
-		// TODO: Return the correct status state
-		responseBody := types.NewNetworkFaultInjectionSuccessResponse("running")
-		logger.Info("Successfully checked status for fault", logger.Fields{
+		// Check and return the status of current fault injection.
+		var responseBody types.NetworkFaultInjectionResponse
+		var httpStatusCode int
+		stringToBeLogged := "Failed to check status for fault"
+		// All command executions for the check network latency workflow all together should finish within 5 seconds.
+		// Thus, create the context here so that it can be shared by all os/exec calls.
+		ctx, cancel := h.osExecWrapper.NewExecContextWithTimeout(context.Background(), requestTimeoutDuration)
+		defer cancel()
+		// Check the status of current fault injection.
+		latencyFaultExists, _, err := h.checkTCFault(ctx, taskMetadata)
+		if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+			responseBody = types.NewNetworkFaultInjectionErrorResponse(fmt.Sprintf(requestTimedOutError, requestType))
+			httpStatusCode = http.StatusInternalServerError
+		} else if err != nil {
+			responseBody = types.NewNetworkFaultInjectionErrorResponse(err.Error())
+			httpStatusCode = http.StatusInternalServerError
+		} else {
+			// If there already exists a fault in the task network namespace.
+			if latencyFaultExists {
+				responseBody = types.NewNetworkFaultInjectionSuccessResponse("running")
+				httpStatusCode = http.StatusOK
+			} else {
+				responseBody = types.NewNetworkFaultInjectionSuccessResponse("not-running")
+				httpStatusCode = http.StatusOK
+			}
+		}
+		logger.Info(stringToBeLogged, logger.Fields{
 			field.RequestType: requestType,
 			field.Request:     request.ToString(),
 			field.Response:    responseBody.ToString(),
 		})
 		utils.WriteJSONResponse(
 			w,
-			http.StatusOK,
+			httpStatusCode,
 			responseBody,
 			requestType,
 		)
@@ -702,7 +796,10 @@ func (h *FaultHandler) StartNetworkPacketLoss() func(http.ResponseWriter, *http.
 			} else {
 				// Invoke the start fault injection functionality if not running.
 				err := h.startNetworkPacketLossFault(ctx, taskMetadata, request)
-				if err != nil {
+				if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+					responseBody = types.NewNetworkFaultInjectionErrorResponse(fmt.Sprintf(requestTimedOutError, requestType))
+					httpStatusCode = http.StatusInternalServerError
+				} else if err != nil {
 					responseBody = types.NewNetworkFaultInjectionErrorResponse(err.Error())
 					httpStatusCode = http.StatusInternalServerError
 				} else {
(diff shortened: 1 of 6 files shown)


ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -536,18 +537,50 @@ func (h *FaultHandler) StartNetworkLatency() func(http.ResponseWriter, *http.Req
...
+			// If there already exists a fault in the task network namespace.
+			if latencyFaultExists {
+				responseBody = types.NewNetworkFaultInjectionErrorResponse("There is already one network latency fault running")
+				httpStatusCode = http.StatusConflict
+			} else if packetLossFaultExists {
+				responseBody = types.NewNetworkFaultInjectionErrorResponse("There is already one network packet loss fault running")
+				httpStatusCode = http.StatusConflict
+			} else {
+				// Invoke the start fault injection functionality if not running.
+				err := h.startNetworkLatencyFault(ctx, taskMetadata, request)
```
mye956, line 565: This just came to mind, should we also add a check here whether or not the context has timed out?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -583,18 +616,51 @@ func (h *FaultHandler) StopNetworkLatency() func(http.ResponseWriter, *http.Requ
...
+			httpStatusCode = http.StatusInternalServerError
+		} else {
+			// If there doesn't already exist a network-latency fault
+			if !latencyFaultExists {
+				stringToBeLogged = "No fault running"
+				responseBody = types.NewNetworkFaultInjectionSuccessResponse("stopped")
+				httpStatusCode = http.StatusOK
+			} else {
+				// Invoke the stop fault injection functionality if running.
+				err := h.stopTCFault(ctx, taskMetadata)
```
mye956, line 642: ditto here, should we check if the context has timed out afterwards?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -1082,6 +1171,55 @@ func validateTaskNetworkConfig(taskNetworkConfig *state.TaskNetworkConfig) error
...
+	jitterInMs := aws.Uint64Value(request.JitterMilliseconds)
+
+	// Command to be executed:
+	// <nsenterPrefix> tc qdisc add dev <interfaceName> root handle 1: prio priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2
+	// <nsenterPrefix> "tc qdisc add dev <interfaceName> parent 1:1 handle 10: netem delay <latency>ms <jitter>ms
+	tcAddQdiscRootCommandComposed := nsenterPrefix + fmt.Sprintf(tcAddQdiscRootCommandString, interfaceName)
+	cmdList := strings.Split(tcAddQdiscRootCommandComposed, " ")
+	cmdOutput, err := h.runExecCommand(ctx, cmdList)
+	if err != nil {
+		logger.Error(fmt.Sprintf("'%s' command failed with the following error: '%s'. std output: '%s'. TaskArn: %s",
```
mye956, line 1193: nit(non-blocking): might make it a bit easier to read if we use `logger.Field`

mye956: ditto for the rest of the logger statements (non-blocking/up to you)

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -536,18 +537,50 @@ func (h *FaultHandler) StartNetworkLatency() func(http.ResponseWriter, *http.Req
...
-		// TODO: Invoke the start fault injection functionality if not running
-
-		responseBody := types.NewNetworkFaultInjectionSuccessResponse("running")
-		logger.Info("Successfully started fault", logger.Fields{
+		var responseBody types.NetworkFaultInjectionResponse
+		var httpStatusCode int
+		stringToBeLogged := "Failed to start fault"
+		// All command executions for the start network latency workflow all together should finish within 5 seconds.
+		// Thus, create the context here so that it can be shared by all os/exec calls.
+		ctx, cancel := h.osExecWrapper.NewExecContextWithTimeout(context.Background(), requestTimeoutDuration)
```
xxx0624, line 545: non-blocking: can we have a comment [here](https://github.com/aws/amazon-ecs-agent/blob/cf8c7a6b65043c550533f330b10aef6d0a342214/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go#L50) about this duration that this is just the guess which maybe updated after testing.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -905,92 +914,32 @@ func TestCheckNetworkBlackHolePort(t *testing.T) {
...
-				agentState.EXPECT().GetTaskMetadata(endpointId).Return(state.TaskResponse{}, nil).Times(0)
-			},
-		},
 		{
 			name:               fmt.Sprintf("%s malformed request body 1", name),
 			expectedStatusCode: 400,
 			requestBody: map[string]interface{}{
 				"DelayMilliseconds":  "incorrect-field",
-				"JitterMilliseconds": jitterMilliseconds,
+				"JitterMilliseconds": "incorrect-field",
```
xxx0624, line 924: I think this is intentional that only one of them is incorrect.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -1017,37 +964,56 @@ func generateNetworkLatencyTestCases(name, expectedHappyResponseBody string) []n
...
 			expectedResponseBody: types.NewNetworkFaultInjectionErrorResponse("required parameter Sources is missing"),
 			setAgentStateExpectations: func(agentState *mock_state.MockAgentState) {
-				agentState.EXPECT().GetTaskMetadata(endpointId).
-					Return(state.TaskResponse{}, nil).
-					Times(0)
+				agentState.EXPECT().GetTaskMetadata(endpointId).Return(state.TaskResponse{}, nil).Times(0)
+			},
+		},
+		{
+			name:               fmt.Sprintf("%s incomplete request body 3", name),
```
xxx0624, line 971: 4? since we have 3 below.

btw what's the testing purpose of this one? to make sure it will fail when DelayMilliseconds and JitterMilliseconds are missing?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -1017,37 +964,56 @@ func generateNetworkLatencyTestCases(name, expectedHappyResponseBody string) []n
...
 			setAgentStateExpectations: func(agentState *mock_state.MockAgentState) {
-				agentState.EXPECT().GetTaskMetadata(endpointId).
-					Return(state.TaskResponse{}, nil).
-					Times(0)
+				agentState.EXPECT().GetTaskMetadata(endpointId).Return(state.TaskResponse{}, nil).Times(0)
 			},
 		},
 		{
-			name:               fmt.Sprintf("%s incomplete request body 4", name),
+			name:               fmt.Sprintf("%s invalid LossPercent in the request body 1", name),
```
xxx0624, line 994: is this for latency fault?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -1017,37 +964,56 @@ func generateNetworkLatencyTestCases(name, expectedHappyResponseBody string) []n
...
 			expectedStatusCode: 400,
 			requestBody: map[string]interface{}{
+				"DelayMilliseconds":  -1,
 				"JitterMilliseconds": jitterMilliseconds,
 				"Sources":            ipSources,
 			},
-			expectedResponseBody: types.NewNetworkFaultInjectionErrorResponse("required parameter DelayMilliseconds is missing"),
+			expectedResponseBody: types.NewNetworkFaultInjectionErrorResponse("json: cannot unmarshal number -1 into Go struct field NetworkLatencyRequest.DelayMilliseconds of type uint64"),
```
xxx0624, line 1001: Good catch! Let's have a new test for the case you find.

 I think we should have a check to make sure this is greater than 0. @mye956 @amogh09 thoughts?

xxx0624: https://github.com/aws/amazon-ecs-agent/blame/cf8c7a6b65043c550533f330b10aef6d0a342214/ecs-agent/tmds/handlers/fault/v1/types/types.go#L97-L106

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -1155,22 +1111,334 @@ func generateNetworkLatencyTestCases(name, expectedHappyResponseBody string) []n
...
+			setExecExpectations: func(exec *mock_execwrapper.MockExec, ctrl *gomock.Controller) {
+				ctx, cancel := context.WithTimeout(context.Background(), requestTimeoutDuration)
+				mockCMD := mock_execwrapper.NewMockCmd(ctrl)
+				exec.EXPECT().NewExecContextWithTimeout(gomock.Any(), gomock.Any()).Times(1).Return(ctx, cancel)
+				exec.EXPECT().CommandContext(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(mockCMD)
+				mockCMD.EXPECT().CombinedOutput().Times(1).Return([]byte{}, errors.New("signal: killed"))
+			},
+		},
+		{
+			name:                 "request timed out",
```
xxx0624, line 1157: Could you help me understand why this `tcCommandEmptyOutput` will lead to a request time out?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -1155,22 +1111,334 @@ func generateNetworkLatencyTestCases(name, expectedHappyResponseBody string) []n
...
+			setExecExpectations: func(exec *mock_execwrapper.MockExec, ctrl *gomock.Controller) {
+				ctx, cancel := context.WithTimeout(context.Background(), requestTimeoutDuration)
+				mockCMD := mock_execwrapper.NewMockCmd(ctrl)
+				exec.EXPECT().NewExecContextWithTimeout(gomock.Any(), gomock.Any()).Times(1).Return(ctx, cancel)
+				exec.EXPECT().CommandContext(gomock.Any(), gomock.Any(), gomock.Any()).Times(1).Return(mockCMD)
+				mockCMD.EXPECT().CombinedOutput().Times(1).Return([]byte("["), nil)
+			},
+		},
+		{
+			name:               "os/exec-times-out",
```
xxx0624, line 1136: In this case,won't the client get request time out error?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: Thanks for the catch! Yes, will add a timeout check for all the command executions.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: SGTM. Will refactor the logs in the helper methods

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: Sure thing. Adding a comment in the const section

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: Got it. Changing it back in the next revision

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: I think we can actually remove this test case. If the request fails when any of the DelayMilliseconds or JitterMilliseconds is missing, we don't need to test what's the behavior when both of them are missing

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: Thanks for the catch! The base structure of this method was copied over from the latency test cases, and I forgot to change the name here. Addressing in the next revision

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
xxx0624: Agree. We don't have to check both are missing as long as we have test cases for that only one of them is missing

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: This is because when checking for timeout, we don't look at the error returned by os/exec. Instead, we look at ctx.Err(). In this test case, ctx is initialized with `context.WithTimeout(context.Background(), -1*time.Second)`, since we used -1 as timeout, when checking its error, it will always be timeout

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: I think this can be removed. This was added to accommodate our old way of checking for timeout (look at os/exec output), Now that we look at the context's error directly, we will catch the timeout before we even check for the command execution output

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -630,17 +701,40 @@ func (h *FaultHandler) CheckNetworkLatency() func(http.ResponseWriter, *http.Req
...
+			httpStatusCode = http.StatusInternalServerError
+		} else {
+			// If there already exists a fault in the task network namespace.
+			if latencyFaultExists {
+				responseBody = types.NewNetworkFaultInjectionSuccessResponse("running")
+				httpStatusCode = http.StatusOK
+			} else {
+				responseBody = types.NewNetworkFaultInjectionSuccessResponse("not-running")
+				httpStatusCode = http.StatusOK
+			}
```
mye956, line 728: Q: should we be logging that we've successfully checked the status of the fault here?

xxx0624 (approved):

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -902,49 +911,8 @@ func TestCheckNetworkBlackHolePort(t *testing.T) {
...
-			},
-			expectedResponseBody: types.NewNetworkFaultInjectionSuccessResponse(expectedHappyResponseBody),
-			setAgentStateExpectations: func(agentState *mock_state.MockAgentState) {
-				agentState.EXPECT().GetTaskMetadata(endpointId).
-					Return(happyTaskResponse, nil).
-					Times(1)
-			},
-		},
-		{
-			name:                 fmt.Sprintf("%s no request body", name),
```
xxx0624, line 940: Why do we remove this test case?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -902,49 +911,8 @@ func TestCheckNetworkBlackHolePort(t *testing.T) {
...
-			expectedStatusCode:   400,
-			requestBody:          nil,
-			expectedResponseBody: types.NewNetworkFaultInjectionErrorResponse("required request body is missing"),
```
xxx0624, line 943: FYI - this is used to check if the request body exist and the specific error msg is returned.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
xxx0624: non-blocking - can you add a follow up task for this?

mye956 (approved):

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
xxx0624: Thanks.

IMO we can use `Do()` to mock the timeout case where we can sleep for some time longer than the requestTimeout.

ie  
```
exec.EXPECT().NewExecContextWithTimeout(gomock.Any(), gomock.Any()).Do(func(i, j interface{}){
...// sleep
}).Return(ctx, cancel)
```

xxx0624 (changes requested):

xxx0624 (approved):

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: Confirmed offline. unit64 guarantees that the value will be non-negative. Since TC allows 0 as input, we should allow 0 as input as well. Thus, no action needed for this one.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: SGTM. Addressing this in the next PR

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: Yes. Thanks for the catch! Will address this in the next PR.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: Got it. Thanks for the catch! Will add this back in the next PR.

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: Same as above ^^^. Will add this back in the next PR.

PR #4364, opened 2024-09-24
Adding EnableFaultInjection to task payload

harishxr:
### Summary

This pull request adds `EnableFaultInjection` to the task payload.
### Implementation details

* Introducing `EnableFaultInjection` into the task struct.
* Sets the `EnableFaultInjection` field to true in the task struct when the corresponding AcsTask is not nil.
### Testing

Ran the existing set of tests to verify the change to `EnableFaultInjection`

Adding EnableFaultInjection to task payload
### Additional Information

### Licensing

### DIFF
--- a/agent/acs/session/payload_responder.go
+++ b/agent/acs/session/payload_responder.go
@@ -132,6 +132,12 @@ func (pmHandler *payloadMessageHandler) addPayloadTasks(payload *ecsacs.PayloadM
 			loggerfield.DesiredStatus: apiTask.GetDesiredStatus(),
 		})
 
+		if apiTask.IsFaultInjectionEnabled() {
+			logger.Info("Fault Injection Enabled for task", logger.Fields{
+				loggerfield.TaskARN: apiTask.Arn,
+			})
+		}
+
 		if task.RoleCredentials != nil {
 			// The payload from ACS for the task has credentials for the
 			// task. Add those to the credentials manager and set the
--- a/agent/api/task/task.go
+++ b/agent/api/task/task.go
@@ -299,8 +299,7 @@ type Task struct {
 
 	NetworkNamespace string `json:"NetworkNamespace,omitempty"`
 
-	// TODO: Will need to initialize/set the value in a follow PR
-	FaultInjectionEnabled bool `json:"FaultInjectionEnabled,omitempty"`
+	EnableFaultInjection bool `json:"enableFaultInjection,omitempty"`
 
 	// DefaultIfname is used to reference the default network interface name on the task network namespace
 	// For AWSVPC mode, it can be eth0 which corresponds to the interface name on the task ENI
@@ -321,6 +320,9 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task, 
 	if err := json.Unmarshal(data, task); err != nil {
 		return nil, err
 	}
+
+	// Set the EnableFaultInjection field if present
+	task.EnableFaultInjection = aws.BoolValue(acsTask.EnableFaultInjection)
 
 	// Overrides the container command if it's set
 	for _, container := range task.Containers {
@@ -3771,7 +3773,7 @@ func (task *Task) IsFaultInjectionEnabled() bool {
 	task.lock.RLock()
 	defer task.lock.RUnlock()
 
-	return task.FaultInjectionEnabled
+	return task.EnableFaultInjection
 }
 
 func (task *Task) GetNetworkMode() string {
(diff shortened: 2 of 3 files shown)


agent/api/task/task.go
```diff
@@ -299,8 +299,7 @@ type Task struct {
 
 	NetworkNamespace string `json:"NetworkNamespace,omitempty"`
 
-	// TODO: Will need to initialize/set the value in a follow PR
-	FaultInjectionEnabled bool `json:"FaultInjectionEnabled,omitempty"`
+	EnableFaultInjection bool `json:"enableFaultInjection,omitempty"`
```
mye956, line 302: nit: we should follow the same syntax in terms of JSON field naming. (i.e. EnableFaultInjection)

agent/acs/session/payload_responder.go
```diff
@@ -132,6 +132,12 @@ func (pmHandler *payloadMessageHandler) addPayloadTasks(payload *ecsacs.PayloadM
 			loggerfield.DesiredStatus: apiTask.GetDesiredStatus(),
 		})
 
+		if apiTask.EnableFaultInjection {
```
mye956, line 135: nit: Could we use `IsFaultInjectionEnabled()` instead?

harishxr: We could use `IsFaultInjectionEnabled()` instead, but used `apiTask.EnableFaultInjection` to maintain similarity in the `addPayloadTasks` with similar such checks.

mye956: Er sorry, I meant could we use the following getter function -> https://github.com/aws/amazon-ecs-agent/blob/08341f7b92d1991feac4eb246c87063d19375f9d/agent/api/task/task.go#L3770

agent/api/task/task.go
harishxr: Discussed offline, leaving as `enableFaultInjection`

agent/acs/session/payload_responder.go
mye956: Also btw, the other checks are referring to another task object(ecsacs task) and not the `apitask`

harishxr: Changed in [74c3067](https://github.com/aws/amazon-ecs-agent/commit/74c3067b008ffcddd4db3b61e39d2f7238cecacd)

ecs-agent/tmds/handlers/v4/state/response.go
```diff
@@ -38,7 +38,7 @@ type TaskResponse struct {
 	EphemeralStorageMetrics *EphemeralStorageMetrics `json:"EphemeralStorageMetrics,omitempty"`
 	CredentialsID           string                   `json:"-"`
 	TaskNetworkConfig       *TaskNetworkConfig       `json:"-"`
-	FaultInjectionEnabled   bool                     `json:"-"`
```
amogh09, line 41: I think the current name is intentional.

mye956: +1

agent/api/task/task.go
```diff
@@ -322,6 +321,11 @@ func TaskFromACS(acsTask *ecsacs.Task, envelope *ecsacs.PayloadMessage) (*Task,
 		return nil, err
 	}
 
+	// Set the EnableFaultInjection field if present
+	if acsTask.EnableFaultInjection != nil {
+		task.EnableFaultInjection = aws.BoolValue(acsTask.EnableFaultInjection)
+	}
```
amogh09, line 327: nit: This is equivalent to 
```go
task.EnableFaultInjection = aws.BoolValue(acsTask.EnableFaultInjection)
```
as `aws.BoolValue` function takes care of `nil` handling. We might want to simplify the code.

ecs-agent/tmds/handlers/v4/state/response.go
xxx0624: we have a change for this field in the feature branch

agent/api/task/task.go
harishxr: Fixed in [1d123a7](https://github.com/aws/amazon-ecs-agent/commit/1d123a79076c13fadefcb8e1af64efc3cd99beca)

ecs-agent/tmds/handlers/v4/state/response.go
mye956: Perhaps we should add a comment either in the PR description or in the code denoting that `EnabledFaultInjection` within the Task struct of `agent/api/task` essentially relates/maps to `FaultInjectionEnabled` within the task response struct of `ecs-agent/tmds/handlers/v4/state/`

amogh09 (approved):

mye956 (approved):

PR #4370, opened 2024-09-30
Add fault injection SourcesToFilter field in the request payload

tshan2001:
### Summary
Add fault injection SourcesToFilter field in the request payload.

Also incorporated the following comments from [PR-4363](https://github.com/aws/amazon-ecs-agent/pull/4363)
1. In the request times out unit tests, use `Do(func(){time.Sleep()})` instead of using -1 when creating context with timeout.
2. Add logging when check latency and check packet loss is successful
3. Add back the no-request-body test case for latency fault and packet loss fault

### Implementation details
A new omitempty field `SourcesToFilter` has been added to the network latency and network packet loss request type, allowing.

### Testing
All existing and newly added unit/integ tests passed.

New tests cover the changes: Unit tests have been modified to accommodate this new field. Added a new test case for when the IP address passed in for this field is invalid.

Add fault injection SourcesToFilter field in the request payload

### Licensing

### DIFF
--- a/agent/handlers/task_server_setup_test.go
+++ b/agent/handlers/task_server_setup_test.go
@@ -476,6 +476,8 @@ var (
 
 	ipSources = []string{"52.95.154.1", "52.95.154.2"}
 
+	ipSourcesToFilter = []string{"8.8.8.8"}
+
 	happyBlackHolePortReqBody = map[string]interface{}{
 		"Port":        port,
 		"Protocol":    protocol,
@@ -486,11 +488,13 @@ var (
 		"DelayMilliseconds":  delayMilliseconds,
 		"JitterMilliseconds": jitterMilliseconds,
 		"Sources":            ipSources,
+		"SourcesToFilter":    ipSourcesToFilter,
 	}
 
 	happyNetworkPacketLossReqBody = map[string]interface{}{
-		"LossPercent": lossPercent,
-		"Sources":     ipSources,
+		"LossPercent":     lossPercent,
+		"Sources":         ipSources,
+		"SourcesToFilter": ipSourcesToFilter,
 	}
 )
 
--- a/ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
+++ b/ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
@@ -110,14 +110,18 @@ var (
 		"DelayMilliseconds":  delayMilliseconds,
 		"JitterMilliseconds": jitterMilliseconds,
 		"Sources":            ipSources,
+		"SourcesToFilter":    ipSourcesToFilter,
 	}
 
 	happyNetworkPacketLossReqBody = map[string]interface{}{
-		"LossPercent": lossPercent,
-		"Sources":     ipSources,
+		"LossPercent":     lossPercent,
+		"Sources":         ipSources,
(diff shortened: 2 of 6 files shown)


agent/handlers/task_server_setup_test.go
```diff
@@ -476,6 +476,8 @@ var (
 
 	ipSources = []string{"52.95.154.1", "52.95.154.2"}
 
+	ipSourcesToFilter = []string{"8.8.8.8"}
```
amogh09, line 479: I think we should also add test cases to check that invalid `SourcesToFilter` values cause a validation failure.

tshan2001: This is added to handler_test.go in the shared library. In that new test case we pass in `"SourcesToFilter": []string{"52.95.154.0/33"},`, which will cause a validation failure

amogh09: Thanks. I should have looked at that file.

amogh09 (approved):

ecs-agent/tmds/handlers/fault/v1/types/types.go
```diff
@@ -137,7 +151,15 @@ func (request NetworkPacketLossRequest) ValidateRequest() error {
 	if request.Sources == nil || len(request.Sources) == 0 {
 		return fmt.Errorf(missingRequiredFieldError, "Sources")
 	}
-	return validateNetworkFaultRequestSources(request.Sources)
+	err := validateNetworkFaultRequestSources(request.Sources, "Sources")
```
xxx0624, line 154: non-blocking - we can check the error like this
```
if err := someFunction(); err != nil {
    // handle error
}
```

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
```diff
@@ -110,15 +110,19 @@ var (
 		"DelayMilliseconds":  delayMilliseconds,
 		"JitterMilliseconds": jitterMilliseconds,
 		"Sources":            ipSources,
+		"SourcesToFilter":    ipSourcesToFilter,
 	}
 
 	happyNetworkPacketLossReqBody = map[string]interface{}{
-		"LossPercent": lossPercent,
-		"Sources":     ipSources,
+		"LossPercent":     lossPercent,
```
xxx0624, line 117: Can we have one more happy test case where no `SourcesToFilter` is provided?

xxx0624 (approved):

ecs-agent/tmds/handlers/fault/v1/types/types.go
tshan2001: Addressing this in the next PR: https://github.com/aws/amazon-ecs-agent/pull/4373

ecs-agent/tmds/handlers/fault/v1/handlers/handlers_test.go
tshan2001: This is covered in the "unknown-request-body-no-existing-fault-no-allowlist-filter" case. We pass in an empty array as SourcesToFilter

xxx0624: I see. Thanks!

PR #4372, opened 2024-09-30
Migrate ecs-init to aws-sdk-go-v2

tinnywang:
### Summary
[Migrate](https://aws.github.io/aws-sdk-go-v2/docs/migrating/) the `ecs-init` module to `aws-sdk-go-v2` because `aws-sdk-go` will reach [end of support on July 31, 2025](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-go-v1-on-july-31-2025/).

### Implementation details
- Migrate from [`session.Session`](https://docs.aws.amazon.com/sdk-for-go/api/aws/session/) to [`aws.Config`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config) per the [docs](https://aws.github.io/aws-sdk-go-v2/docs/migrating/#configuration-loading).
- Migrate from [`ec2metadata.EC2Metadata`](https://docs.aws.amazon.com/sdk-for-go/api/aws/ec2metadata/) to [`imds.Client`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/ec2/imds) and update client construction per the [docs](https://aws.github.io/aws-sdk-go-v2/docs/migrating/#client-construction).
- Migrate from [`s3manager.Downloader`](https://docs.aws.amazon.com/sdk-for-go/api/service/s3/s3manager/) to [`manager.Downloader`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/feature/s3/manager) per the [docs](https://aws.github.io/aws-sdk-go-v2/docs/migrating/#s3).
- For the [`endpoints`](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints) package,
  - Per the [docs](https://aws.github.io/aws-sdk-go-v2/docs/migrating/#endpoints), "the [endpoints](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints) package no longer exists in the AWS SDK for Go V2. Each service client now embeds its required AWS endpoint metadata within the client package." However, the client packages only support endpoint resolution (e.g. ECS's [`EndpointResolver`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ecs#NewDefaultEndpointResolver) and [`EndpointResolverV2`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ecs#NewDefaultEndpointResolverV2)). The partition and region metadata needed by `ecs-init` is only available via the internal [`awsrulesfn`](https://github.com/aws/aws-sdk-go-v2/tree/main/internal/endpoints/awsrulesfn) package.
  - To make the partition and region metadata accessible to `ecs-init`, use [`go generate`](https://github.com/aws/amazon-ecs-agent/blob/85a28b923aeb9a6a61e818c8cf4d34892fd49f83/ecs-init/config/awsrulesfn/partitions.gen.go) to copy over relevant SDK files [[1](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.go)][[2](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go)] into an `ecs-init` package.
  - Migrate from [`endpoints.PartitionForRegion`](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#PartitionForRegion) to [`awsrulesfn.GetPartitionForRegion`](https://github.com/aws/amazon-ecs-agent/blob/1291ca2f8dfe91e1fd74e88ecf71b79a6b1d8b8d/ecs-init/config/awsrulesfn/partitions.gen.go#L13-L26), which is our own implementation of [`getPartition`](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go#L33-L57). We do not use `getPartition` because it defaults to the `aws` partition if it cannot find a match for the region. https://github.com/aws/amazon-ecs-agent/blob/cf8c7a6b65043c550533f330b10aef6d0a342214/ecs-init/config/common.go#L141-L143
  - Hardcode in partitions and regions that used to be exported as [constants](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants).
- Add `aws-sdk-go-v2` modules to go.mod in `ecs-init`.
- `go mod tidy && go mod verify && go mod vendor`
- `make gogenerate-init`

### Testing
- GitHub status checks pass.
- Downloaded the `ecs-init` RPM/DEB packages built from this PR. Installed the packages on EC2 instances running AL2, AL2023, and Ubuntu. Verified that Agent started/stopped/restarted when the systemd `ecs-init` service was started/stopped/restarted.

Migrate ecs-init to aws-sdk-go-v2.

### Licensing


ecs-init/config/awsrulesfn/partition.go
```diff
@@ -0,0 +1,76 @@
+package awsrulesfn
```
tinnywang, line 1: [This](https://github.com/aws/amazon-ecs-agent/blob/85a28b923aeb9a6a61e818c8cf4d34892fd49f83/ecs-init/config/awsrulesfn/partitions.gen.go) `go generate` copies [this AWS SDK file](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go) from vendor into our package.

ecs-init/config/awsrulesfn/partitions.go
```diff
@@ -0,0 +1,403 @@
+// Code generated by endpoint/awsrulesfn/internal/partition. DO NOT EDIT.
```
tinnywang, line 1: [This `go generate`](https://github.com/aws/amazon-ecs-agent/blob/85a28b923aeb9a6a61e818c8cf4d34892fd49f83/ecs-init/config/awsrulesfn/partitions.gen.go) copies [this AWS SDK file](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.go) from vendor into our package.

ecs-init/go.mod
```diff
@@ -1,10 +1,16 @@
 module github.com/aws/amazon-ecs-agent/ecs-init
 
-go 1.19
+go 1.21
+
+toolchain go1.22.2
```
danehlim, line 5: What is the use case for having a toolchain version with a different minor version than the module's go version here? (i.e., 22 vs 21)

ecs-init/config/common.go
```diff
@@ -140,14 +139,16 @@ var validDrivers = map[string]struct{}{
...
-		return "", errors.Errorf("could not resolve partition ID for region %q", region)
+	// If the AWS SDK cannot match the region to a partition, it returns the default partition (aws).
+	// It only returns nil if the default partition is undefined.
+	partition := awsrulesfn.GetPartition(region)
+	if partition == nil {
+		return "", errors.Errorf("could not resolve partition for region %q", region)
 	}
 
-	bucketRegion, ok := partitionBucketRegion[regionPartition.ID()]
+	bucketRegion, ok := partitionBucketRegion[partition.Name]
```
danehlim, line 149: Have we confirmed that partition name is guaranteed to always be the same as partition ID? Or has the testing we have done for this pull request include the case where ecs-init downloads Agent from the partitional S3 bucket that stores the Agent in non-aws partition region?

If testing was done only in aws partition, I am concerned we may not have accounted for some potential discrepancy between partition name and ID which would cause L149 here to consistently result in `!ok`. The test case in aws partition would pass even if this discrepancy exists because as you mentioned, ecs-init defaults the region to us-east-1 when call to `GetAgentPartitionBucketRegion` function fails.

ecs-init/go.mod
tinnywang: No need for different versions. I updated my go version and the toolchain directive was automatically added at some point.

ecs-init/config/common.go
tinnywang: > Have we confirmed that partition name is guaranteed to always be the same as partition ID?

[This](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.json) is the source file used to generate [partition metadata](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.go). For all existing partitions, ID and name are the same, but that might not be true for future partitions. If this is a concern, I can implement an alternative to [`getPartition`](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go#L33-L57) that returns [`Partition`](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go#L6-L11), which has ID accessible, instead of [`PartitionConfig`](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go#L14-L21). It could also return `nil` instead of the default aws partition when there's no matching partition for the region.

> Or has the testing we have done for this pull request include the case where ecs-init downloads Agent from the partitional S3 bucket that stores the Agent in non-aws partition region?

I've tested `ecs-init` from this PR in aws (us-west-2) and aws-cn (cn-north-1).

tinnywang: Somewhat unrelated...we always fallback to `us-east-1` as the download region, but does S3 support cross-partition downloads?

danehlim: > [This](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.json) is the source file used to generate [partition metadata](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partitions.go). For all existing partitions, ID and name are the same, but that might not be true for future partitions. If this is a concern, I can implement an alternative to [`getPartition`](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go#L33-L57) that returns [`Partition`](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go#L6-L11), which has ID accessible, instead of [`PartitionConfig`](https://github.com/aws/aws-sdk-go-v2/blob/main/internal/endpoints/awsrulesfn/partition.go#L14-L21). It could also return `nil` instead of the default aws partition when there's no matching partition for the region.

Thanks for confirming! Yeah I am aligned with doing both of these to persist existing behavior and have clearer visibility into when there is no matching partition.

danehlim: > Somewhat unrelated...we always fallback to `us-east-1` as the download region, but does S3 support cross-partition downloads?

As far as I know S3 does not support cross-partition downloads. So in instances where we fallback to `us-east-1` in a non-aws partition, my understanding is that download would fail in this case.

ecs-init/go.mod
danehlim: Got it, I see this line has now been removed in most recent commit. Thanks!

ecs-init/config/common.go
tinnywang: Changes are implemented in https://github.com/aws/amazon-ecs-agent/pull/4372/commits/3f3b0fec1dcfa8750719a755c5f4b2a75268832a.

danehlim (approved):
Thanks much for your work on this migration change!

ecs-init/config/common.go
danehlim: > Somewhat unrelated...we always fallback to us-east-1 as the download region, but does S3 support cross-partition downloads?

Current understanding: S3 cross-partition downloads are possible if the file being downloaded has public read access granted to it (reference steps to do so: https://repost.aws/knowledge-center/read-access-objects-s3-bucket).

mye956 (approved):

PR #4373, opened 2024-09-30
Add fault injection allowlist IP implementation

tshan2001:
### Summary
Add fault injection allowlist IP implementation.

### Implementation details
Now the request payload for network packet loss fault and network latency fault takes in 2 arrays, with the newly added array as an allowlist. IP addresses added to the allowlist will not first be added to a separate filter to ensure that they won't be impacted by the loss/latency filter.

### Testing
Manual testing:
```
# Launch a task with ecs-exec enabled, and connect to the container in the task
# Inject a latency fault with 2 ip addresses, with one of the 2 added to the allowlist list
% curl -X POST ${ECS_AGENT_URI}/fault/v1/network-packet-loss/start --data '{"lossPercent":50, "Sources":["3.5.76.1", "3.5.76.4"], "SourcesToFilter": ["3.5.76.1"]}'
{"Status":"running"}
# Ping the one that's allowlisted
% ping -D 3.5.76.1
PING 3.5.76.1 (3.5.76.1) 56(84) bytes of data.
...
--- 3.5.76.1 ping statistics ---
59 packets transmitted, 59 received, 0% packet loss, time 59305ms
rtt min/avg/max/mdev = 0.468/0.559/1.766/0.181 ms
# We can see that this IP is not impacted by the loss filter
# Ping the one that's not allowlisted
% ping -D 3.5.76.4
PING 3.5.76.4 (3.5.76.4) 56(84) bytes of data.
...
--- 3.5.76.4 ping statistics ---
37 packets transmitted, 20 received, 45.9459% packet loss, time 36781ms
rtt min/avg/max/mdev = 0.382/0.682/3.834/0.840 ms
# We can see that the loss percent is close to 50% (the one we added in the filter)

# Now stop the fault
% curl -X POST ${ECS_AGENT_URI}/fault/v1/network-packet-loss/stop --data '{"lossPercent":50, "Sources":["3.5.76.1", "3.5.76.4"], "SourcesToFilter": ["3.5.76.1"]}'
{"Status":"stopped"}

# Curl the check endpoint
% curl -X POST ${ECS_AGENT_URI}/fault/v1/network-packet-loss/status --data '{"lossPercent":50, "Sources":["3.5.76.1", "3.5.76.4"], "SourcesToFilter": ["3.5.76.1"]}'
{"Status":"not-running"}


# Repeat the same for latency fault:
% curl -X POST ${ECS_AGENT_URI}/fault/v1/network-latency/start --data '{"DelayMilliseconds":1000, "JitterMilliseconds":100, "Sources":["3.5.76.1", "3.5.76.4"], "SourcesToFilter": ["3.5.76.1"]}'
{"Status":"running"}
# Ping the one that's allowlisted
% ping -D 3.5.76.1
PING 3.5.76.1 (3.5.76.1) 56(84) bytes of data.
[1727806969.961276] 64 bytes from 3.5.76.1: icmp_seq=1 ttl=63 time=2.31 ms
[1727806970.960944] 64 bytes from 3.5.76.1: icmp_seq=2 ttl=63 time=0.574 ms
[1727806971.977215] 64 bytes from 3.5.76.1: icmp_seq=3 ttl=63 time=0.513 ms
# Ping the one that's not allowlisted
% ping -D 3.5.76.4
PING 3.5.76.4 (3.5.76.4) 56(84) bytes of data.
[1727806982.874885] 64 bytes from 3.5.76.4: icmp_seq=1 ttl=63 time=967 ms
[1727806983.970733] 64 bytes from 3.5.76.4: icmp_seq=2 ttl=63 time=1063 ms
[1727806984.927355] 64 bytes from 3.5.76.4: icmp_seq=3 ttl=63 time=1015 ms
# We can see the difference in time.

# Now stop the fault
% curl -X POST ${ECS_AGENT_URI}/fault/v1/network-latency/stop --data '{"DelayMilliseconds":1000, "JitterMilliseconds":100, "Sources":["3.5.76.1", "3.5.76.4"], "SourcesToFilter": ["3.5.76.1"]}'
{"Status":"stopped"}

# Curl the check endpoint
% curl -X POST ${ECS_AGENT_URI}/fault/v1/network-latency/status --data '{"DelayMilliseconds":1000, "JitterMilliseconds":100, "Sources":["3.5.76.1", "3.5.76.4"], "SourcesToFilter": ["3.5.76.1"]}'
{"Status":"not-running"}
```

New tests cover the changes: 

Add fault injection allowlist IP implementation.

### Additional Information

### Licensing

### DIFF
--- a/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
+++ b/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
@@ -64,9 +64,9 @@ const (
 	tcAddQdiscRootCommandString      = "tc qdisc add dev %s root handle 1: prio priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2"
 	tcAddQdiscLatencyCommandString   = "tc qdisc add dev %s parent 1:1 handle 10: netem delay %dms %dms"
 	tcAddQdiscLossCommandString      = "tc qdisc add dev %s parent 1:1 handle 10: netem loss %d%%"
-	tcAddFilterForIPCommandString    = "tc filter add dev %s protocol ip parent 1:0 prio 1 u32 match ip dst %s flowid 1:1"
+	tcAllowlistIPCommandString       = "tc filter add dev %s protocol ip parent 1:0 prio 1 u32 match ip dst %s flowid 1:3"
+	tcAddFilterForIPCommandString    = "tc filter add dev %s protocol ip parent 1:0 prio 2 u32 match ip dst %s flowid 1:1"
 	tcDeleteQdiscParentCommandString = "tc qdisc del dev %s parent 1:1 handle 10:"
-	tcDeleteFilterCommandString      = "tc filter del dev %s prio 1"
 	tcDeleteQdiscRootCommandString   = "tc qdisc del dev %s root handle 1: prio"
 )
 
@@ -1231,19 +1231,13 @@ func (h *FaultHandler) startNetworkLatencyFault(ctx context.Context, taskMetadat
 		field.CommandOutput: string(cmdOutput[:]),
 	})
 	// After creating the queueing discipline, create filters to associate the IPs in the request with the handle.
-	for _, ip := range request.Sources {
-		tcAddFilterForIPCommandComposed := nsenterPrefix + fmt.Sprintf(tcAddFilterForIPCommandString, interfaceName, *ip)
-		cmdList = strings.Split(tcAddFilterForIPCommandComposed, " ")
-		_, err = h.runExecCommand(ctx, cmdList)
-		if err != nil {
-			logger.Error("Command execution failed", logger.Fields{
-				field.CommandString: tcAddFilterForIPCommandComposed,
-				field.Error:         err,
-				field.CommandOutput: string(cmdOutput[:]),
-				field.TaskARN:       taskMetadata.TaskARN,
-			})
-			return err
-		}
+	// First redirect the allowlisted ip addresses to band 1:3 where is no network impairments.
+	if err := h.addIPAddressesToFilter(ctx, request.SourcesToFilter, taskMetadata, nsenterPrefix, tcAllowlistIPCommandString, interfaceName); err != nil {
+		return err
+	}
+	// After processing the allowlisted ips, associate the ip addresses in Sources with the qdisc.
+	if err := h.addIPAddressesToFilter(ctx, request.Sources, taskMetadata, nsenterPrefix, tcAddFilterForIPCommandString, interfaceName); err != nil {
+		return err
 	}
 
 	return nil
@@ -1296,19 +1290,13 @@ func (h *FaultHandler) startNetworkPacketLossFault(ctx context.Context, taskMeta
 		field.CommandOutput: string(cmdOutput[:]),
 	})
 	// After creating the queueing discipline, create filters to associate the IPs in the request with the handle.
-	for _, ip := range request.Sources {
-		tcAddFilterForIPCommandComposed := nsenterPrefix + fmt.Sprintf(tcAddFilterForIPCommandString, interfaceName, *ip)
-		cmdList = strings.Split(tcAddFilterForIPCommandComposed, " ")
-		_, err = h.runExecCommand(ctx, cmdList)
-		if err != nil {
-			logger.Error("Command execution failed", logger.Fields{
-				field.CommandString: tcAddFilterForIPCommandComposed,
-				field.Error:         err,
-				field.CommandOutput: string(cmdOutput[:]),
-				field.TaskARN:       taskMetadata.TaskARN,
-			})
-			return err
-		}
+	// First redirect the allowlisted ip addresses to band 1:3 where is no network impairments.
+	if err := h.addIPAddressesToFilter(ctx, request.SourcesToFilter, taskMetadata, nsenterPrefix, tcAllowlistIPCommandString, interfaceName); err != nil {
+		return err
+	}
+	// After processing the allowlisted ips, associate the ip addresses in Sources with the qdisc.
+	if err := h.addIPAddressesToFilter(ctx, request.Sources, taskMetadata, nsenterPrefix, tcAddFilterForIPCommandString, interfaceName); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -64,6 +64,7 @@ const (
 	tcAddQdiscRootCommandString      = "tc qdisc add dev %s root handle 1: prio priomap 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2"
 	tcAddQdiscLatencyCommandString   = "tc qdisc add dev %s parent 1:1 handle 10: netem delay %dms %dms"
 	tcAddQdiscLossCommandString      = "tc qdisc add dev %s parent 1:1 handle 10: netem loss %d%%"
+	tcAllowlistIPCommandString       = "tc filter add dev %s protocol ip parent 1:0 prio 1 u32 match ip dst %s flowid 1:3"
 	tcAddFilterForIPCommandString    = "tc filter add dev %s protocol ip parent 1:0 prio 1 u32 match ip dst %s flowid 1:1"
```
xxx0624, line 68: Let's use prio (priority) of 2 for IP in the disallow list and use prio of 1 for IP in the allow list. Because the allowlist takes precedence.

xxx0624: > The protocol field describes protocol that the filter will be applied to. We will only discuss case of ip protocol. The preference field (priority can be used alternatively) sets the priority of currently defined filter. This is important, since you can have several filters (lists of rules) with different priorities. Each list will be passed in the order the rules were added, then list with lower priority (higher preference number) will be processed. The parent field defines the CBQ tree top (e.g. 1:0), the filter should be attached to.

https://tldp.org/HOWTO/Adv-Routing-HOWTO/lartc.adv-filter.u32.html

xxx0624: and I think we may need to remove the prio 2 handler - https://github.com/aws/amazon-ecs-agent/blob/dev/ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go#L1348-L1360

tshan2001: I think we should just delete the root handle. Here's what happens when we try to delete a filter that doesn't exist:
```
# Filter that doesn't exist:
% tc filter del dev eth0 prio 2
Error: Filter with specified priority/protocol not found.

# Filter that exists:
% tc filter del dev eth0 prio 1
# No output as expected
```

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -1231,8 +1231,24 @@ func (h *FaultHandler) startNetworkLatencyFault(ctx context.Context, taskMetadat
...
+		cmdList = strings.Split(tcAllowlistIPCommandComposed, " ")
+		_, err = h.runExecCommand(ctx, cmdList)
+		if err != nil {
+			logger.Error("Command execution failed", logger.Fields{
+				field.CommandString: tcAllowlistIPCommandComposed,
+				field.Error:         err,
+				field.CommandOutput: string(cmdOutput[:]),
+				field.TaskARN:       taskMetadata.TaskARN,
+			})
+			return err
```
amogh09, line 1246: Just confirming - this error does not get added to the response sent back to the client, right?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -1296,8 +1312,24 @@ func (h *FaultHandler) startNetworkPacketLossFault(ctx context.Context, taskMeta
...
+		}
+	}
+	// After processing the allowlisted ips, associate the ip addresses in Sources with the qdisc.
```
amogh09, line 1330: Is this identical to the code block added to `startNetworkLatencyFault` method? If so, maybe we should create a helper function? These methods are already quite heavy.

xxx0624 (approved):
LGTM

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
```diff
@@ -1345,22 +1377,6 @@ func (h *FaultHandler) stopTCFault(ctx context.Context, taskMetadata *state.Task
...
-			field.Error:         err,
-			field.CommandOutput: string(cmdOutput[:]),
-			field.TaskARN:       taskMetadata.TaskARN,
-		})
-		return err
-	}
-	logger.Info("Command execution completed", logger.Fields{
-		field.CommandString: tcDeleteFilterCommandComposed,
-		field.CommandOutput: string(cmdOutput[:]),
-	})
```
amogh09, line 1363: What was this for and why is it no longer needed?

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: The Linux command output won't be added to the response, but we will add the err to the response

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: Thanks for the callout. Will add this to a helper function

ecs-agent/tmds/handlers/fault/v1/handlers/handlers.go
tshan2001: Originally we run 3 commands to delete a latency/packet loss fault:
1. tc qdisc del dev <interfaceName> parent 1:1 handle 10:
2. tc filter del dev <interfaceName> prio 1
3. tc qdisc del dev <interfaceName> root handle 1: prio
Now that we're using 2 different prio's (1 for allowlist, 2 for ips to be added to the filter), in step 2, we would need to do either of the following:
1. Delete both prio1 and prio2
2. Don't delete any of them. Everything will be cleaned up by the third command (when the root is deleted).
Both options work the same. After discussing this with Xing, we decided to go with option 2. When we delete prio that doesn't exist, there will be extra effort needed to handle the Linux error. When the customer doesn't add a filter, this will be triggered. Thus, we will simply rely on deleting the root handle to remove fault

amogh09: Thanks for the explanation. Sounds good!

amogh09 (approved):

xxx0624 (approved):

